PR #462, opened 2019-03-07
Fix warnings

tobim:


libvast/vast/logger.hpp
```diff
@@ -68,9 +68,9 @@ auto id_or_name(T&& x) {
       return caf::detail::pretty_type_name(typeid(value_type));
   } else {
     if constexpr (has_ostream_operator<T>)
-      return x;
+      return std::forward<T&&>(x);
```
mavam: Should this be `std::forward<T>` as opposed to `T&&`?

libvast/vast/uri.hpp
```diff
@@ -50,7 +50,7 @@ namespace vast {
 /// >         |           |            |            |        |
 /// >      scheme     authority       path        query   fragment
 /// >         |   _____________________|__
-/// >        / \ /                        \
+/// >        / \ /                        ⧵
```
mavam: Elegant "fix"!

libvast/vast/logger.hpp
tobim: duh. Thanks for paying attention.

PR #465, opened 2019-03-13
Allow filtering table slices based on IDs

Neverlord:


Neverlord:
This PR depends on https://github.com/vast-io/vast/pull/466. Once that one got merged I'll rebase this PR to get it past Jenkins.

libvast/test/table_slice.cpp
```diff
@@ -115,4 +116,66 @@ TEST(random integer slices) {
...
+TEST(select prefix) {
+  auto sut = zeek_full_conn_log_slices.front();
+  sut.unshared().offset(100);
+  auto select_result = select(sut, make_ids({{0, 150}}));
+  REQUIRE_EQUAL(select_result.size(), 1u);
+  CHECK_EQUAL(select_result[0]->rows(), 50u);
+  CHECK_EQUAL(to_events(*select_result[0]), to_events(*sut, 0, 50));
+}
+
+TEST(select off by one prefix) {
```
mavam:
```suggestion
TEST(select off-by-one prefix) {
```

libvast/vast/table_slice.hpp
```diff
@@ -143,6 +143,21 @@ expected<std::vector<table_slice_ptr>>
 make_random_table_slices(size_t num_slices, size_t slice_size,
                          record_type layout, id offset = 0, size_t seed = 0);
 
+/// Selects all rows in `xs` with event IDs in `selection` and appends produced
+/// table slices to `result`. Cuts `xs` into multiple slices if `selection`
+/// produces gaps.
+/// @pre `xs != nullptr`
+void select(std::vector<table_slice_ptr>& result, const table_slice_ptr& xs,
+            const ids& selection);
```
mavam: Looks like this function is an implementation detail. If so, it could go into the `cpp` file instead.

Otherwise, documenting the parameters via `@param` would be helpful.

libvast/vast/table_slice.hpp
```diff
@@ -143,6 +143,21 @@ expected<std::vector<table_slice_ptr>>
...
+
+/// Selects all rows in `xs` with event IDs in `selection`. Cuts `xs` into
+/// multiple slices if `selection` produces gaps.
+/// @returns new table slices of the same implementation type as `xs` from
+///          `selection`.
+/// @pre `xs != nullptr`
```
mavam: It'd be nice to have complete documentation whenever possible: (1) a description, (2) arguments, (3) return value, (4) contracts. (2) is missing here.

libvast/test/table_slice.cpp
```diff
@@ -115,4 +116,66 @@ TEST(random integer slices) {
   CHECK_LESS_EQUAL(*highest, 200);
 }
 
+TEST(select all) {
+  auto sut = zeek_full_conn_log_slices.front();
```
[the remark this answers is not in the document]
Neverlord: It's common testing jargon. `XUT = X-Under-Test`. For example, SUT = System Under Test, AUT = Actor Under Test, etc. Here: Slice Under Test.

I usually name the main variable in a scope like this.

libvast/test/table_slice.cpp
Neverlord: That's what I had at first. The problem is that clang-format is "too clever" here and formats this as `off - by - one`. I figured it's more readable this way. :/

libvast/test/table_slice.cpp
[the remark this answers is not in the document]
mavam: Okay, fine with me.

But we just don't have many parts of the code base yet that use that style. The vast majority uses `x` and `xs`. This growing inconsistency still bothers me.

libvast/vast/table_slice.hpp
Neverlord: This is not an implementation detail, see commit message for a68a07c. I'll add the params.

libvast/src/table_slice.cpp
```diff
@@ -110,6 +112,70 @@ make_random_table_slices(size_t num_slices, size_t slice_size,
...
+      last_id = id;
+    } else {
+      ++last_id;
+    }
+    VAST_ASSERT(id >= xs->offset());
+    auto row = id - xs->offset();
+    VAST_ASSERT(row < xs->rows());
+    for (size_t column = 0; column < xs->columns(); ++column) {
+      if (!builder->add(xs->at(row, column))) {
+        VAST_ERROR(__func__, "failed to add column", column, "in row", row,
```
mavam:
```suggestion
        VAST_ERROR(__func__, "failed to add data at column", column, "in row", row,
```

You're not really adding a column here. 😉

libvast/src/table_slice.cpp
```diff
@@ -110,6 +112,70 @@ make_random_table_slices(size_t num_slices, size_t slice_size,
...
+                   "to the builder:", builder->add(xs->at(row, column)));
```
mavam: I wouldn't add the same expression `builder->add(..)` twice.

Neverlord: It's nonsense anyway, because `builder->add` returns a boolean. Will fix.

PR #466, opened 2019-03-13
Add scaffold for using test logs in unit tests without parsing them over and over again

Neverlord:
The new tool `gen-vast-slices` allows us to convert our log artifacts into table slices *once* while compiling `libvast_test`. The unit test than simply deserialize the slices instead of parsing them over and over again.

### DIFF
--- a/libvast_test/src/events.cpp
+++ b/libvast_test/src/events.cpp
@@ -28,6 +28,16 @@
 #include "vast/concept/printable/vast/data.hpp"
 #include "vast/concept/printable/vast/event.hpp"
 
+// Pull in the auto-generated serialized table slices.
+
+namespace artifacts::logs::zeek {
+
+extern char conn_buf[];
+
+extern size_t conn_buf_size;
+
+} // namespace artifacts::logs::zeek
+
 namespace fixtures {
 
 namespace {
@@ -108,6 +118,8 @@ std::vector<table_slice_ptr> events::zeek_dns_log_slices;
 std::vector<table_slice_ptr> events::zeek_http_log_slices;
 std::vector<table_slice_ptr> events::bgpdump_txt_slices;
 // std::vector<table_slice_ptr> events::random_slices;
+
+std::vector<table_slice_ptr> events::zeek_full_conn_log_slices;
 
 std::vector<event> events::ascending_integers;
 std::vector<table_slice_ptr> events::ascending_integers_slices;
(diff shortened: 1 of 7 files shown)


libvast_test/src/events.cpp
```diff
@@ -293,7 +305,22 @@ events::events() {
   SANITY_CHECK(zeek_dns_log, zeek_dns_log_slices);
   SANITY_CHECK(zeek_http_log, zeek_http_log_slices);
   SANITY_CHECK(bgpdump_txt, bgpdump_txt_slices);
-  //SANITY_CHECK(random, const_random_slices);
+  // SANITY_CHECK(random, const_random_slices);
+  // Read the full Zeek conn.log.
+  // TODO: port remaining slices to new deserialization API and replace
+  //       this hard-coded starting offset
+  id offset = 100000;
```
mavam: This one could move down closer to its usage, e.g., right before the `for` loop.

libvast_test/vast/test/fixtures/events.hpp
```diff
@@ -63,7 +65,13 @@ struct events {
   }
 
   std::vector<table_slice_ptr> copy(std::vector<table_slice_ptr> xs);
+
+  auto take_n(const std::vector<table_slice_ptr>& xs, size_t n) {
```
mavam: How about calling that function simply `take`? The `n` parameter seems to be sufficient to convey that it's about multiple slices.

libvast_test/vast/test/fixtures/events.hpp
```diff
@@ -63,7 +65,13 @@ struct events {
...
+    VAST_ASSERT(n <= xs.size());
+    auto first = xs.begin();
+    auto last = first + n;
+    return std::vector<table_slice_ptr>{first, last};
```
mavam: Does anything speak against returning a `span<table_slice_ptr>` here instead?

tools/gen-vast-slices/README.md
```diff
@@ -0,0 +1,23 @@
+# gen-vast-slices
+
+The **gen-vast-slices** tool parses supported log formats from STDIN into table
+slices and prints them in serialized form to STDOUT.
```
mavam:
```suggestion
slices and prints them in serialized, *binary* form to STDOUT.
```

I would also add that the binary format comes from `caf::binary_serializer`.

tools/gen-vast-slices/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+Our primary use case is to create a `.cpp` with an array that contains the
+serialized form of a table slice vector. Unit tests than can quickly
```
mavam:
```suggestion
serialized form of a table slice vector. Unit tests then can quickly
```

tools/gen-vast-slices/README.md
```diff
@@ -0,0 +1,23 @@
...
+deserialize the vector instead of parsing files from the file system.
```
mavam:
```suggestion
deserialize the vector instead of performing filesystem I/O.
```

tools/gen-vast-slices/README.md
Neverlord: That reads like the output to STDOUT is binary, which isn't the case.

libvast_test/vast/test/fixtures/events.hpp
Neverlord: I would not expect a function named `take` to return a *view* into the data. Besides, for unit tests I think operating on owning containers makes life easier. There's no reason to optimize memory allocations here.

tools/gen-vast-slices/README.md
Neverlord: I took a slightly different wording, hope that's a good compromise.

PR #467, opened 2019-03-13
Add erase functionality to ARCHIVE and store

Neverlord:


libvast/src/system/archive.cpp
```diff
@@ -147,6 +147,11 @@ archive(archive_type::stateful_pointer<archive_state> self, path dir,
             namespace defs = defaults::system;
             self->delayed_send(self, defs::telemetry_rate,
                                telemetry_atom::value);
+          },
+          [=](erase_atom, ids erase) {
```
mavam: `s/erase/xs`? The `erase_atom` already defines the function. A second `erase` seems redudant.

libvast/vast/store.hpp
```diff
@@ -48,11 +48,13 @@ class store {
   /// @relates lookup
   virtual std::unique_ptr<lookup> extract(const ids& xs) const = 0;
 
+  /// Erases events from the store.
```
mavam: Missing documentation:
```cpp
/// @param xs The set of IDs to erase.
/// @returns An error iff the operation failed.
```

PR #468, opened 2019-03-13
Add listen option to reader commands 

tobim:
The first two commits are done so that the third and main change can be done in a natural way. The default input source for a concrete reader can be set by "overriding" the `defaults` struct from the `reader` base class:
```
struct my_reader : single_layout_reader {
  struct defaults {
    constexpr static inputs input = inputs::inet;
    constexpr static const char* uri = ":12345/udp";
  };
};
```

- [x] Update documentation
- [x] Add changelog entry


libvast/vast/defaults.hpp
```diff
@@ -27,53 +25,52 @@ namespace vast::defaults {
...
-extern size_t max_events;
+constexpr size_t max_events = 0;
 
 /// Maximum flow lifetime before eviction.
-extern size_t max_flow_age;
+constexpr size_t max_flow_age = 60;
 
 /// Number of concurrent flows to track.
-extern size_t max_flows;
+constexpr size_t max_flows = 1 << 20;
```
mavam: While we're changing this: can we use SI literals here for improved clarity?

libvast/vast/defaults.hpp
```diff
@@ -27,53 +25,52 @@ namespace vast::defaults {
...
-extern size_t cutoff;
+constexpr size_t cutoff = std::numeric_limits<size_t>::max();
 
 /// Flow table expiration interval.
-extern size_t flow_expiry;
+constexpr size_t flow_expiry = 10;
 
 /// Flush to disk after that many packets.
-extern size_t flush_interval;
+constexpr size_t flush_interval = 10000;
```
mavam:
```suggestion
constexpr size_t flush_interval = 10'000;
```

libvast/vast/defaults.hpp
```diff
@@ -86,34 +83,35 @@ caf::atom_value table_slice_type(caf::actor_system& sys,
...
-extern caf::atom_value table_slice_type;
+constexpr caf::atom_value table_slice_type = caf::atom("default");
 
 /// Maximum size for sources that generate table slices.
-extern size_t table_slice_size;
+constexpr size_t table_slice_size = 100;
 
 /// Maximum number of events per INDEX partition.
-extern size_t max_partition_size;
+constexpr size_t max_partition_size = 1 << 20;
```
mavam: As above: SI literals would help here.

libvast/vast/concept/parseable/vast/endpoint.hpp
```diff
@@ -32,11 +33,7 @@ struct endpoint_parser : parser<endpoint_parser> {
     using namespace parser_literals;
     auto hostname = +(alnum | chr{'-'} | chr{'_'} | chr{'.'});
     auto host = hostname ->* [&](std::string x) { e.host = std::move(x); };
-    auto port = (':' >> u16) ->* [&](uint16_t x) { e.port = x; };
-    auto p
-      = (host >> ~port)
-      | port
-      ;
+    auto p = (host >> ~(':' >> parsers::port)) | ':' >> parsers::port;
```
mavam: A port parser has the form `x/T` where `x` is the number and `T` the port type. This is our ASCII data representation. Is that intentional? I could imagine, because you may want to be able to define UDP, but I'm not sure if this changes the syntax on the command line. In any case, we should probably update the unit tests (`endpoint.cpp`) as well if the suffix is now mandatory.

If the syntax for passing an endpoint changes, this is also a user-facing change needs documentation in the changelog.

libvast/src/format/mrt.cpp
```diff
@@ -291,13 +290,13 @@ reader::reader(caf::atom_value table_slice_type)
     {"old_state", count_type{}},
     {"new_state", count_type{}},
   }}.name("mrt::bgp4mp::state_change");
+  if (input)
```
mavam: Is it a problem to move from a non-initialized `unique_ptr`?

libvast/vast/format/reader.hpp
```diff
@@ -29,6 +31,14 @@ class reader {
 public:
   // -- member types -----------------------------------------------------------
 
+  enum class inputs { file, inet };
```
mavam: How about using the two categories `file` and `net`. The `i` in `inet` suggests "Internet" to me. I think all we want is to distinguish between file and network-based input.

libvast/vast/defaults.hpp
tobim: It used to be defined with SI literals in the `.cpp`. I changed it for two reasons:
1. We should not do `using namespace ...` in the header, and when I have to choose between `vast::si_literals::operator""_Mi(1)` and `1 << 20` I choose the latter.
2. Pulling in an extra header for a single trivial symbol feels wasteful.

libvast/vast/concept/parseable/vast/endpoint.hpp
tobim: I should have run the tests before opening this PR...
The parser should still be able to accept the original syntax. This is a mistake on my part and i am looking into possible solutions atm.

libvast/src/format/mrt.cpp
tobim: Not it isn't. I intended to call `reset()` here instead of assigning directly.

libvast/vast/format/reader.hpp
tobim: The suggestive nature is intentional, I wanted to convey that this can be used with ipv4/6. This naming is pretty prevalent in posix APIs (e.g `AF_INET`). Another value we might add here in the far future is `zigbee` which is also a network, but does not use the same addressing scheme.

libvast/vast/defaults.hpp
mavam: I see the namespace argument, but I find `1 << 20` to cryptic for someone who might not be a seasoned coder. Maybe we can write it out instead and use `'` delimiters?

libvast/vast/format/reader.hpp
mavam: Convinced. I'm fine with calling it `inet` now.

libvast/src/format/mrt.cpp
Neverlord: A default-constructed `unique_ptr` is null according to the standard. So there are no surprises from unitialized pointers.

libvast/vast/system/reader_command.hpp
```diff
@@ -41,23 +45,49 @@ caf::message reader_command(const command& cmd, caf::actor_system& sys,
...
+  if (uri) {
+    endpoint ep;
+    if (!vast::parsers::endpoint(*uri, ep))
+      return caf::make_message(
+        make_error(vast::ec::parse_error, "unable to parse endpoint", *uri));
+    Reader reader{table_slice};
+    auto slice_size = get_or(options, "table-slice-size",
+                             defaults::system::table_slice_size);
+    auto& mm = sys.middleman();
+    auto src = mm.spawn_broker(datagram_source<Reader>, ep.port.number(),
```
mavam: Before executing this statement, you now need to check whether `ep.port.type() == port::udp`. It should fail for all other types (until we have support for raw TCP sockets).

tobim: True. I would accept `port::unknown` as well though.

mavam: And what would `unknown` map to? TCP or UDP?

tobim: As long as we only have UDP, it is ok to assume that is what the user meant. The correct way to handle inputs without `port_type` would be to ask the `Reader` for the default protocol that it expects and set it to that. Inside the `inet` case, we should have a switch statement with a branch for every implemented type and return error by default. Does that sound good to you?

mavam: > As long as we only have UDP, it is ok to assume that is what the user meant.

I can see that point. I would just not see a user actually typing `42/?` instead of `42/udp`. That's why I was suggesting to avoid `unknown` in the first place.

> he correct way to handle inputs without `port_type` would be to ask the `Reader` for the default protocol that it expects and set it to that.

Yes. I actually think we should make the `/T` suffix optional in the parser. If no port type is provided in the input, it will be parsed as `unknown`. It will not break anything in the data parser, because the port parser is tried after all other numeric parsers. But it liberates us from using an unconventional endpoint syntax. Users should still be able to write `1.2.3.4:8888` for this to work and `1.2.3.4:8888/udp` if they want/have to be specific.

tobim: > I can see that point. I would just not see a user actually typing 42/? instead of 42/udp. That's why I was suggesting to avoid unknown in the first place.

The endpoint parser does not rely on the port parser specifically to allow port specified through just the number. See below.

> Yes. I actually think we should make the `/T` suffix optional in the parser. If no port type is provided in the input, it will be parsed as `unknown`. It will not break anything in the data parser, because the port parser is tried after all other numeric parsers.

I tried that yesterday and it actually broke the expression parser, so I decided to just special case the endpoint parser.

mavam: > The endpoint parser does not rely on the port parser specifically to allow port specified through just the number.

Okay, good.

> I tried that yesterday and it actually broke the expression parser

Eeek, I wonder where my mental model is broken. The expression parser should use the data parser, and the data parser should try all numeric types before trying ports. We'll have to investigate what goes wrong there in a different context.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+- The `import` subcommand gained the `--listen,l` option to receive input
```
```suggestion
- The `import` command gained the `--listen,l` option to receive input
```

I would consider subcommands an implementation detail. It's all the same for the user.

tobim:
Rebased and ready from my side.

PR #469, opened 2019-03-14
Add try-catch block around call to std::imbue()

tobim:


libvast/src/system/accountant.cpp
```diff
@@ -109,7 +109,12 @@ void accountant_state::command_line_heartbeat() {
   if (logger && logger->verbosity() >= CAF_LOG_LEVEL_INFO
       && accumulator.node.events > 0) {
     std::ostringstream oss;
-    oss.imbue(std::locale(""));
+    try {
+      oss.imbue(std::locale(""));
+    } catch (const std::exception& e) {
+      VAST_DEBUG_ANON("accoutant failed to set locale the statistics output:",
```
Neverlord: Why are you using `VAST_DEBUG_ANON` here? `VAST_DEBUG(self, ...)` should be fine in this context. Also the sentence is off. Did you mean "... to set the locale for statistics ..."?

tobim: The state type didn't have a `self` member. That is rectified with the follow up commits.

libvast/src/system/accountant.cpp
```diff
@@ -112,8 +112,8 @@ void accountant_state::command_line_heartbeat() {
     try {
       oss.imbue(std::locale(""));
     } catch (const std::exception& e) {
-      VAST_DEBUG_ANON("accoutant failed to set locale the statistics output:",
-                      e.what());
+      VAST_DEBUG(self,
```
mavam: Should this perhaps be a warning?

libvast/vast/system/accountant.hpp
```diff
@@ -81,6 +70,20 @@ using accountant_type = caf::typed_actor<
   caf::reacts_to<telemetry_atom>>;
 // clang-format on
 
+/// @relates accountant
+struct accountant_state {
+  static constexpr const char* name = "accountant";
+  accountant_type::stateful_pointer<accountant_state> self;
```
Neverlord: It seems like this pointer is never initialized. The state can have a constructor that takes the self pointer during construction.

libvast/src/system/accountant.cpp
tobim: Imho warnings should be reserved for problems that affect the user in more serious way. For example if data could not be retrieved. I don't think the statistics formatting warrants that.

mavam: Good point.

libvast/vast/system/accountant.hpp
tobim: For some reason I thought this would work even without an explicitly defined constructor. Alas, it does not.

PR #470, opened 2019-03-15
Add --disable-accounting option to root command

tobim:


mavam:
Jenkins doesn't seem to be happy.

tobim:
> Jenkins doesn't seem to be happy.

This seems to be cause by the latest refactoring in caf.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+- Introduced the --disable-accounting option to the vast command.
```
```suggestion
- The new `--disable-accounting` option can shut off periodic gathering of system telemetry.
```

You used backticks for the option in the other PR, so I figured we should be consistent. Also, I don't think we need to prefix entries in the `Added` section with `"Introduced..."`, so I went ahead and described the actual feature, as opposed to that we we just added it.

PR #461, opened 2019-03-06
Add rudimentary aging strategy

Neverlord:
- [x] Add new abstraction for query processing
- [x] Implement new ERASER based on the new abstraction
- [x] Add and implement erase handler in the ARCHIVE
- [x] Implement `store::erase`
- [x] Spawn ERASER when spinning up VAST components

PR #471, opened 2019-03-15
Implement segment_store::erase

Neverlord:
Sorry for one big commit. I had to hand-pick the code from `story/ch965`.


mavam:
Looks like the review in my first PR didn't make it in yet, e.g., [this comment](https://github.com/vast-io/vast/pull/461/files#r264742754).

Neverlord:
Sorry for jumping the gun on opening this PR. I think I've addresses all relevant review comments from #461 now, so this should be good for another review round.

mavam:
I think a major comment remains still unaddressed (at least I could not find it addressed by skimming your commit messages):

> There seems to be an invalid transition possible: after going through `collect hits` -> `request more hits` -> `no`, I end up in `idle`. But I would not go into `await query id` from there again, right?

What's your take on it?

Neverlord:
> What's your take on it?

I think your comment relates to the query processor or the eraser. However, this PR only adds an implementation for `segment_store::erase`.

mavam:
Ah, okay. I'll dive into this PR then.

libvast/src/segment.cpp
```diff
@@ -11,24 +11,38 @@
...
-#include "vast/detail/byte_swap.hpp"
-#include "vast/detail/narrow.hpp"
-
 namespace vast {
 
 using namespace binary_byte_literals;
 
+ids segment::meta_data::flat_slice_ids() const {
+  ids result;
+  auto concat = [&](const ids& x) { result |= x; };
```
mavam: This seems rather heavy-weight. Why not just append bits to `result`?

libvast/src/segment_store.cpp
```diff
@@ -177,27 +170,135 @@ std::unique_ptr<store::lookup> segment_store::extract(const ids& xs) const {
...
+    auto slice_ids = sref.meta().slice_ids();
+    // Check whether we can drop the entire segment.
+    if (std::all_of(slice_ids.begin(), slice_ids.end(), is_subset_of_xs)) {
+      erased_events += drop(sref);
+      return;
+    }
+    auto segment_id = sref.id();
+    // Get all slices in the segment and generate a new segment that contains
+    // only what's left after dropping the selection.
+    auto segment_ids = sref.meta().flat_slice_ids();
```
mavam: Something like this seems less heavy-weight:
```cpp
for (auto [begin, end] : slice_ids(sref.meta())) {
  segment_ids.append_bits(false, begin - segment_ids.size());
  segment_ids.append_bits(true, end - begin);
}
```

libvast/src/segment_store.cpp
```diff
@@ -177,27 +170,135 @@ std::unique_ptr<store::lookup> segment_store::extract(const ids& xs) const {
...
+    std::vector<table_slice_ptr> slices;
+    if (auto maybe_slices = sref.lookup(segment_ids)) {
```
mavam: I find the `maybe_*` style a bit confusing. Can we write it like this? 

```cpp
if (auto slices = sref.lookup(segment_ids))) {
  if (slices->empty()) {
    VAST_WARNING(this, "got no slices after lookup for segment", segment_id,
                 "=> erases entire segment!");
    erased_events += drop(sref);
    return;
  } else {
    VAST_WARNING(this, "was unable to get table slice for segment",
                 segment_id, "=> erases entire segment!");
    erased_events += drop(sref);
    return;
  }
  ...
```

It would only require replacing `slices` with `*slices` within `if (slices)`.

libvast/src/segment_store.cpp
```diff
@@ -177,27 +170,135 @@ std::unique_ptr<store::lookup> segment_store::extract(const ids& xs) const {
...
+        return;
+      }
+    } else {
+      VAST_WARNING(this, "was unable to get table slice for segment",
+                   segment_id, "=> erases entire segment!");
+      erased_events += drop(sref);
+      return;
+    }
+    VAST_ASSERT(slices.size() > 0);
+    auto keep_mask = ~xs;
```
mavam: Can you add a short (single-line suffices) description of the algorithm you're implementing below using `keep_mask`?

libvast/src/segment_store.cpp
```diff
@@ -177,27 +170,135 @@ std::unique_ptr<store::lookup> segment_store::extract(const ids& xs) const {
...
+    // Remove stale state.
+    segments_.erase_value(segment_id);
+    // Create a new segment from the remaining slices.
+    segment_builder tmp_builder;
+    segment_builder* builder = &tmp_builder;
+    if constexpr (std::is_same_v<decltype(sref), segment_builder&>) {
+      // If `update` got called with a builder then we simply use that by
+      // resetting it and filling it with new content. Otherwise, we fill
+      // `tmp_builder` instead and replace the the segment `sref` in the next
+      // `if constexpr` block.
```
mavam: This is very useful comment. Can you add it to the definition of the function? It would be helpful there to know that the lambda is actually generic.

libvast/src/segment_store.cpp
```diff
@@ -177,27 +170,135 @@ std::unique_ptr<store::lookup> segment_store::extract(const ids& xs) const {
...
+    }
+    // Flush the new segment and remove the previous segment.
+    if constexpr (std::is_same_v<decltype(sref), segment&>) {
+      auto new_segment = builder->finish();
+      auto filename = segment_path() / to_string(new_segment->id());
+      if (auto err = save(nullptr, filename, new_segment))
+        VAST_ERROR(this, "failed to persist the new segment");
+      auto stale_filename = segment_path() / to_string(segment_id);
+      // Schedule deletion of the segment file when releasing the chunk.
+      sref.chunk()->add_deletion_step([=] { rm(stale_filename); });
```
mavam: I'm a bit hesitant to drill into the API of `chunk` for what we're trying to do here. Admittedly, I don't see another good way of solving it. @tobim, do you have an idea?

libvast/test/bitmap_algorithms.cpp
```diff
@@ -0,0 +1,35 @@
...
+  CHECK(contains(make_ids({{10, 20}}), make_ids({{11, 20}})));
+  CHECK(contains(make_ids({{10, 20}}), make_ids({{10, 19}})));
+  CHECK(contains(make_ids({{10, 20}}), make_ids({{10, 19}})));
+}
+
+TEST(contains not) {
+  CHECK(!contains(make_ids({{10, 20}}), make_ids({{9, 19}})));
+  CHECK(!contains(make_ids({{10, 20}}), make_ids({{11, 21}})));
+  CHECK(!contains(make_ids({{10, 20}}), make_ids({5, 15, 25})));
+}
```
mavam: Good, I like your dedication to writing tests. 👊

libvast/test/segment_store.cpp
```diff
@@ -16,47 +16,288 @@
...
+  fixture() {
+    store = segment_store::make(directory / "segments", 512_KiB, 2);
+    if (store == nullptr)
+      FAIL("segment_store::make failed to allocate a segment store");
+    segment_path = store->segment_path();
+    // Approximates an ID range for [0, max_id) with 100, because
+    // `make_ids({{0, max_id}})` unfortunately leads to performance
+    // degradations.
+    everything = make_ids({{0, 100}});
+    // Check that ground truth is as we expect.
```
mavam:
```suggestion
    // Check that ground truth is what we expect.
```

libvast/test/segment_store.cpp
```diff
@@ -16,47 +16,288 @@
...
+
+TEST(erase slice part from persisted segment) {
+  put_cold(zeek_conn_log_slices);
+  erase(make_ids({{10, 14}}));
+  auto slices = get(everything);
+  REQUIRE_EQUAL(slices.size(), 4u);
+  CHECK_SLICE(slices[0], 0, 0);
+  CHECK_SLICE(slices[1], 1, 0, 2);
+  CHECK_SLICE(slices[2], 1, 6, 2);
+  CHECK_SLICE(slices[3], 2, 0);
```
mavam: Once again, awesome testing!

libvast/src/segment.cpp
Neverlord: Can you explain what's heavyweight about this and how I would achieve semantically identical but more efficient code using append member functions?

Neverlord: Nevermind, `visit_ids` gets the axe anyways.

libvast/src/segment_store.cpp
Neverlord: This code needs to run when running the destructor of the chunk. So I think `add_deletion_step` is the most natural and direct way to express this.

libvast/src/segment_store.cpp
```diff
@@ -177,27 +170,135 @@ std::unique_ptr<store::lookup> segment_store::extract(const ids& xs) const {
...
+  auto is_subset_of_xs = [&](const ids& ys) {
+    // Compare using rank(), because the bitmaps might have different sizes.
+    return contains(xs, ys);
+  };
+  // Counts number of total erased events for user-facing output.
+  uint64_t erased_events = 0;
+  // Convenience function for updating given segment by pruning all IDs in `xs`
+  // from it.
+  // @returns `true` if the entire segment got erased, `false` otherwise.
+  auto update = [&](auto& sref) {
```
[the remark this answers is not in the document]
Neverlord: I renamed the argument to `seg`. However, factoring this function any more doesn't make much sense to me. This lambda function implements the body of the `for`-loop, so this lambda pretty much *is* the implementation of `segment_store::erase` and does not contain re-usable pieces that make sense elsewhere.

libvast/src/segment_store.cpp
Neverlord: `slices` already is a variable in the outer scope that we fill with whatever `seg.lookup` returns.

PR #473, opened 2019-03-18
Monitor importer from the source process

tobim:


libvast/src/system/source_command.cpp
```diff
@@ -151,6 +153,8 @@ caf::message source_command(const command& cmd, caf::actor_system& sys,
           self->send(importer, subscribe_atom::value, flush_atom::value, self);
         else
           stop = true;
+      } else {
+        VAST_DEBUG(&cmd, "received unexpected DOWN from", msg.source);
```
mavam: This looks more like an assertion. Can we add one?

tobim: done.

PR #475, opened 2019-03-22
Add import --num option and enable type query integration test

tobim:


libvast/src/system/default_application.cpp
```diff
@@ -58,10 +58,12 @@ default_application::default_application() {
...
                 opts()
                   .add<caf::atom_value>("table-slice,t", "table slice type")
-                  .add<bool>("node,n",
-                             "spawn a node instead of connecting to one")
                   .add<bool>("blocking,b",
-                             "block until the IMPORTER forwarded all data"));
+                             "block until the IMPORTER forwarded all data")
+                  .add<bool>("node,N",
+                             "spawn a node instead of connecting to one")
+                  .add<size_t>("num,n",
```
mavam: I find `num` too short for a long option. 😉

How about using `max-events` instead? That also works for import and export, describes the intent better as well. Only downside, it doesn't begin with the same letter as the short option. I think that's fine because most people still find `-n` very idiomatic for a number of events.

libvast/vast/system/datagram_source.hpp
```diff
@@ -129,9 +130,15 @@ caf::behavior datagram_source(datagram_source_actor<Reader>* self,
...
         st.mgr->out().push(std::move(slice));
       };
-      auto [err, produced] = st.reader.read(capacity * table_slice_size,
-                                            table_slice_size, push_slice);
+      auto events = detail::opt_min(st.remaining, capacity * table_slice_size);
+      auto [err, produced] = st.reader.read(events, table_slice_size,
+                                            push_slice);
       t.stop(produced);
+      if (st.remaining) {
+        *st.remaining -= produced;
```
mavam: Maybe add an assertion that `produced <= *st.remaining`.

libvast/vast/system/source.hpp
```diff
@@ -177,18 +193,27 @@ caf::behavior source(caf::stateful_actor<source_state<Reader>>* self,
...
-                                         table_slice_size, push_slice);
+      auto events = detail::opt_min(st.remaining, num * table_slice_size);
+      auto [err, produced] = st.reader.read(events, table_slice_size,
+                                            push_slice);
       // TODO: if the source is unable to generate new events (returns 0)
       //       then we should trigger CAF to poll the source after a
       //       predefined interval of time again, e.g., via delayed_send
       t.stop(produced);
+      if (st.remaining)
+        *st.remaining -= produced;
```
mavam: Same assertion mentioned above could be here as well.

libvast/src/system/default_application.cpp
tobim: Agreed.

mavam:
A changelog entry is still missing. I know, easy to overlook, but we need to train ourselves...if need be with a [cattle prod](https://en.wikipedia.org/wiki/Cattle_prod) at the office. 😉

PR #479, opened 2019-04-06
Reorganize program options in a hierarchical way

Neverlord:
Sort command options into a well-defined structure.
For example, running `vast export -n 20 pcap -w some.file` emits the
following configuration:

```
export {
  max-events = 20
  pcap {
    write = "some.file"
  }
}
```

Putting the snippet above into `vast.conf` and running the command
without CLI options has the same effect.

The `defaults.hpp` header was re-organized so that the names match the
nested names. Further, all subcommands now use classes with static
members, because we need these classes as template parameter for various
functions, e.g., `reader_command`.

Note that users can omit the category on the command line. For example,
`vast export --max-events=20 pcap --write=some.file` is still equivalent
to the previous invocation.


mavam:
Overall, this looks really good! I just skimmed this from a high level and let @tobim to perform an actual review.

- Why did `vast::endpoint` vanish? There are also printers/parsers that should go with it, if we really have no more use for it.
- We should document the user-facing changes in the changelog, e.g., the renaming from `--dir` to `--directory`
- We need to discuss how to keep this in sync with https://docs.tenzir.com.
- We do we need a `global` category? Can't this just be empty so that we'd start with, e.g., `system` as first level in the hierarchy?

Neverlord:
> Why did `vast::endpoint` vanish?

It didn't, I've only removed its `.cpp` file. The class itself and its printers/parser are still used.

> We should document the user-facing changes in the changelog.

I'll double-check what actually changed for users. The changes I made to manpage file had nothing to do with my changes, the manpage was wrong before.

> We do we need a global category?

This is just the default category. Several commands don't add custom options but they still need to define something as category. All commands that actually add options override it.

mavam:
> It didn't, I've only removed its .cpp file. The class itself and its printers/parser are still used.

Got it.

>  Several commands don't add custom options but they still need to define something as category.

Okay, good to know.

libvast/src/system/spawn_sink.cpp
```diff
@@ -38,48 +40,61 @@ namespace vast::system {
...
+template <class Writer, class Defaults>
 maybe_actor spawn_generic_sink(caf::local_actor* self, spawn_arguments& args) {
   if (!args.empty())
     return unexpected_arguments(args);
-  VAST_UNBOX_VAR(out, detail::make_output_stream(args.options));
+  std::string category = Defaults::category;
+  auto& options = args.options;
+  auto output = get_or(options, category + ".write", Defaults::write);
+  auto uds = get_or(options, category + ".uds", false);
+  VAST_UNBOX_VAR(out, detail::make_output_stream(output, uds));
```
tobim: This could use the convenience overload of `make_output_stream`.

libvast/src/system/default_application.cpp
```diff
@@ -35,16 +35,30 @@
 #include "vast/system/pcap_writer_command.hpp"
 #endif
 
+#define READER(name)                                                           \
+  reader_command<format::name ::reader, defaults::import ::name>, #name
```
tobim: Removing the whitespace works for me. Is this a gcc workaround?
Same on the other macros.

Neverlord: This is simply how clang-format does it.

libvast/vast/defaults.hpp
```diff
@@ -22,66 +22,181 @@
...
-constexpr int64_t pseudo_realtime_factor = 0;
+} // namespace import
 
-/// Number of bytes to keep per event.
-constexpr size_t cutoff = std::numeric_limits<size_t>::max();
+// -- constants for the export command and its subcommands ---------------------
 
-/// Flow table expiration interval.
-constexpr size_t flow_expiry = 10;
+// Unfortunately, `export` is a reserved keyword. The trailing `_` exists only
```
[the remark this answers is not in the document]
Neverlord: According to [cppreference](https://en.cppreference.com/w/cpp/keyword), `import` is going to be similar to `override`, i.e., an identifier with special meaning in some places.

libvast/src/system/spawn_sink.cpp
Neverlord: Good catch! Will do.

PR #484, opened 2019-04-12
Allow adding new components to the NODE

Neverlord:
The NODE actor in VAST starts all components. However, we need a way to
add new components as extensions. This set of changes adds a static map
to the NODE actor's state that downstream projects can extend during
startup.

### DIFF
--- a/libvast/vast/system/node.hpp
+++ b/libvast/vast/system/node.hpp
@@ -13,20 +13,40 @@
 
 #pragma once
 
+#include <map>
 #include <string>
 
 #include <caf/actor.hpp>
 #include <caf/event_based_actor.hpp>
 #include <caf/stateful_actor.hpp>
 
+#include "vast/aliases.hpp"
 #include "vast/command.hpp"
 #include "vast/filesystem.hpp"
+#include "vast/system/spawn_arguments.hpp"
 #include "vast/system/tracker.hpp"
 
 namespace vast::system {
 
+struct node_state;
+
+using node_actor = caf::stateful_actor<node_state>;
+
 /// State of the node actor.
 struct node_state {
+  // -- member types ----------------------------------------------------------
+
+  /// Spawns a component (actor) for the NODE with given spawn arguments.
+  using component_factory = maybe_actor (*)(node_actor*, spawn_arguments&);
+
+  /// Identifies a command and its parent by name. The two names in conjunction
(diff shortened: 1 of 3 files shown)


libvast/vast/system/node.hpp
```diff
@@ -13,20 +13,40 @@
...
 /// State of the node actor.
 struct node_state {
+  // -- member types ----------------------------------------------------------
+
+  /// Spawns a component (actor) for the NODE with given spawn arguments.
+  using component_factory = maybe_actor (*)(node_actor*, spawn_arguments&);
+
+  /// Identifies a command and its parent by name. The two names in conjunction
+  /// uniquely identify a component, e.g., by disambiguating "import pcap" from
+  /// "export pcap".
```
mavam: Does this approach actually work? If `pcap` had a subcommand `foo`, then we may have `(pcap, foo)` twice in there, once for `import` and once for `export`. It seems that storing the full path `(import, pcap, foo)` needs to be the map key. Or using a recursive structure.

Neverlord: It does work "for now", but I see your point. A hierarchical mapping is easier to extend in the future.

Neverlord:
@mavam I'm using a simple map now that maps the full command name to component factories. A hierarchical data structure would introduce a lot of extra complexity by introducing a recursive map of string to variant containing either the same recursive map type again or the factory function. With this approach, all we need is to call `full_name` on the command and then we can look that up in one shot. It's also much easier to work with when adding new factories.

libvast/vast/system/node.hpp
```diff
@@ -44,7 +59,7 @@ struct node_state {
   tracker_type tracker;
 
   /// Stores how many components per label are active.
-  std::unordered_map<std::string, size_t> labels;
+  std::map<std::string, size_t> labels;
```
tobim: Why this change? Seems unrelated to the PR...

Neverlord: That's true. I don't remember why I changed it. Probably because this map is small and the hash map doesn't pay off for small maps. Would you like me to change it back?

tobim: I don't think it makes a difference either way. I asked out of curiosity. Let's keep it and merge.

Neverlord: @mavam then it's up to you to merge or add another review round. :)

PR #486, opened 2019-04-26
Add support for JSON as import format

tobim:
This PR adds support for ingesting LDJSON.

Todo:
- [x] Changelog entry.
- [x] Update documentation.
- [x] Unit testing.


libvast/vast/concept/parseable/vast/time.hpp
```diff
@@ -143,21 +144,28 @@ struct ymdhms_parser : vast::parser<ymdhms_parser> {
...
+    auto sec = parsers::real_opt_dot.with(
+      [](auto x) { return x >= 0.0 && x <= 60.0; });
+    auto time_divider = '+'_p | 'T';
+    // clang-format off
     auto p = year >> '-' >> mon
-        >> ~('-' >> day >> ~('+' >> hour >> ~(':' >> min >> ~(':' >> sec))));
+              >> ~('-' >> day
+                >> ~(time_divider >> hour
+                  >> ~(':' >> min
+                    >> ~(':' >> sec >> ~parsers::chr{'Z'}))));
```
mavam: `parser::chr` has a `char` as attribute. Don't you want just a literal parser, which has no attribute?

libvast/vast/format/json.hpp
```diff
@@ -41,4 +53,202 @@ class writer : public printer_writer<event_printer>{
...
+
+template <class F>
+struct convert {
+  convert(F&& f) : f_{std::forward<F>(f)} {
+  }
+
+  bool operator()(vast::json::number n, const vast::count_type&) const {
+    auto c = vast::count(n);
+    return f_(std::move(c));
+  }
```
mavam: Newlines between the functions would be nice.

libvast/vast/format/json.hpp
```diff
@@ -41,4 +53,202 @@ class writer : public printer_writer<event_printer>{
...
+  bool operator()(vast::json::number n, const vast::timespan_type&) const {
+    std::chrono::duration<vast::json::number> x{n};
+    auto t = std::chrono::duration_cast<vast::timespan>(x);
+    return f_(std::move(t));
```
mavam: Since `t` is an arithmetic type, the move feels a bit too heavy for `number` as well as for `boolean`.

libvast/vast/format/json.hpp
```diff
@@ -41,4 +53,202 @@ class writer : public printer_writer<event_printer>{
...
+    auto layout = selector(*xs);
+    if (!layout)
+      return make_error(vast::ec::parse_error, "unable to get a layout");
+    auto bptr = builder(*layout);
+    if (bptr == nullptr)
+      return make_error(vast::ec::parse_error, "unable to get a builder");
+    for (auto& field : layout->fields) {
+      auto i = xs->find(field.name);
+      if (i == xs->end()) {
+        bptr->add(vast::make_data_view(caf::none));
```
mavam: Mind adding a comment for why we add `none` when we cannot find a field?

libvast/vast/system/reader_command.hpp
```diff
@@ -106,10 +140,14 @@ caf::message reader_command(const command& cmd, caf::actor_system& sys,
...
     Reader reader{slice_type, std::move(*in)};
     VAST_INFO(reader, "reads data from", *file);
-    auto src = sys.spawn(source<Reader>, std::move(reader), factory, slice_size,
-                         max_events);
-    return source_command(cmd, sys, std::move(src), options, first, last);
+    src = sys.spawn(source<Reader>, std::move(reader), factory, slice_size,
+                    max_events);
+  }
+  if (schema) {
+    caf::scoped_actor self{sys};
```
mavam: What's the rationale for spinning up a `scoped_actor` here? Would `anon_send` perhaps suffice?

libvast/vast/system/reader_command.hpp
```diff
@@ -59,13 +84,23 @@ caf::message reader_command(const command& cmd, caf::actor_system& sys,
...
-                           defaults::system::table_slice_size);
+  // Supply an alternate schema, if requested.
+  expected<vast::schema> schema{caf::none};
+  {
+    auto sc = caf::get_if<std::string>(&options, category + ".schema");
+    auto sf = caf::get_if<std::string>(&options, category + ".schema-file");
+    if (sc && sf)
+      return make_message(
+        make_error(ec::invalid_configuration,
+                   "had both schema and schema-file provided"));
```
mavam: Good error handling!

libvast/vast/format/json.hpp
```diff
@@ -41,4 +53,202 @@ class writer : public printer_writer<event_printer>{
...
+      return false;
+    return f_(std::move(a));
+  }
+  bool operator()(vast::json::array a, const vast::vector_type& v) const {
+    vast::vector d;
+    auto run = [&](auto value) {
+      d.push_back(std::move(value));
+      return true;
+    };
+    auto c = convert<decltype(run)>{std::move(run)};
```
mavam: Would `convert{run}` do the trick as well? A lambda doesn't need to be moved and C++17 class template deduction probably obviates the need for `decltype(run)`.

libvast/vast/format/json.hpp
```diff
@@ -41,4 +53,202 @@ class writer : public printer_writer<event_printer>{
...
+}
+
+template <class F>
+struct convert {
+  convert(F&& f) : f_{std::forward<F>(f)} {
+  }
+
+  bool operator()(vast::json::number n, const vast::count_type&) const {
+    auto c = vast::count(n);
+    return f_(std::move(c));
```
mavam: Does anything speak against simplifying this to `f_(vast::count(n))`?

libvast/vast/format/json.hpp
```diff
@@ -41,4 +53,202 @@ class writer : public printer_writer<event_printer>{
...
+    return f_(std::move(p));
+  }
+  bool operator()(vast::json::string s, const vast::address_type&) const {
+    vast::address a;
+    if (!vast::parsers::addr(s, a))
+      return false;
+    return f_(std::move(a));
+  }
+  bool operator()(vast::json::array a, const vast::vector_type& v) const {
+    vast::vector d;
```
mavam: You could call `d.reserve(a.size())` here. Also, what is `d` supposed to stand for? Does anything speak against `xs`?

libvast/vast/format/json.hpp
```diff
@@ -41,4 +53,202 @@ class writer : public printer_writer<event_printer>{
...
+    auto run = [&](auto value) {
```
mavam: What does `run` suppose to convey to the reader? That it's a run (= sequence) of elements? Otherwise I'd call this lambda simply `push_back`.

libvast/vast/format/json.hpp
```diff
@@ -41,4 +53,202 @@ class writer : public printer_writer<event_printer>{
...
+    if (t)
+      result.add(*t);
+    return result;
+  }
+
+  static const char* name() {
+    return "json-reader";
+  }
+
+  caf::optional<vast::record_type> t = caf::none;
```
mavam: Maybe use `layout` instead of `t` for a clearer name?

libvast/vast/format/json.hpp
```diff
@@ -41,4 +53,202 @@ class writer : public printer_writer<event_printer>{
...
+protected:
+  caf::error read_impl(size_t max_events, size_t max_slice_size,
+                       consumer& f) override;
+
+private:
+  using iterator_type = std::string_view::const_iterator;
+
+  // void patch(std::vector<data>& xs);
+
+  Selector selector;
```
mavam:
```suggestion
  Selector selector_;
```

Plus all the changes that would go with this.

libvast/vast/format/json.hpp
```diff
@@ -41,4 +53,202 @@ class writer : public printer_writer<event_printer>{
...
+    return f_(std::move(t));
+  }
+  bool operator()(vast::json::boolean b, const vast::boolean_type&) const {
+    return f_(std::move(b));
+  }
+  bool operator()(vast::json::number n, const vast::port_type&) const {
+    auto p = vast::port(n);
+    return f_(std::move(p));
+  }
+  bool operator()(vast::json::string s, const vast::address_type&) const {
```
mavam:
```suggestion
  bool operator()(const vast::json::string& s, const vast::address_type&) const {
```

libvast/vast/format/json.hpp
```diff
@@ -41,4 +53,202 @@ class writer : public printer_writer<event_printer>{
...
+  bool operator()(vast::json::number n, const vast::count_type&) const {
+    auto c = vast::count(n);
+    return f_(std::move(c));
+  }
+  bool operator()(vast::json::number n, const vast::timespan_type&) const {
+    std::chrono::duration<vast::json::number> x{n};
+    auto t = std::chrono::duration_cast<vast::timespan>(x);
+    return f_(std::move(t));
+  }
+  bool operator()(vast::json::string s, const vast::timestamp_type&) const {
```
mavam:
```suggestion
  bool operator()(const vast::json::string& s, const vast::timestamp_type&) const {
```

libvast/vast/format/json.hpp
```diff
@@ -41,4 +53,202 @@ class writer : public printer_writer<event_printer>{
...
+}
+
+template <class Selector>
+const char* reader<Selector>::name() const {
+  return Selector::name();
+}
+
+template <class F>
+struct convert {
+  convert(F&& f) : f_{std::forward<F>(f)} {
```
mavam:
```suggestion
  convert(F f) : f_{f} {
```

For functions, we could use this simpler form. But I'm also fine with the way it is right now.

libvast/vast/format/json.hpp
mavam: There is another viable overload for `(json::number, `timestamp_type)`, e.g., for the typical Zeek floating point values representing the time since UNIX epoch.

libvast/vast/format/json.hpp
```diff
@@ -41,4 +53,202 @@ class writer : public printer_writer<event_printer>{
...
+    auto p = vast::port(n);
+    return f_(std::move(p));
+  }
+  bool operator()(vast::json::string s, const vast::address_type&) const {
+    vast::address a;
+    if (!vast::parsers::addr(s, a))
+      return false;
+    return f_(std::move(a));
+  }
+  bool operator()(vast::json::array a, const vast::vector_type& v) const {
```
mavam: Do you want to capture `a` by value here? You don't seem to move it or its elements.

libvast/vast/format/json.hpp
tobim: I'd prefer to keep the rvalue version, this way we don't need to worry about the size of the capture.

libvast/vast/format/json.hpp
tobim: The type parameter must be passed explicitly because this code is first instantiated with the lamba from `read_impl()`. It tries to create the same type from the current instantiation and does not do any deduction.
The move can be omitted if I change the constructor to take `F` by value, but then we get a copy of the closure. I'd prefer not to have it.

libvast/vast/format/json.hpp
```diff
@@ -149,51 +149,65 @@ const char* reader<Selector>::name() const {
...
+    auto t = std::chrono::duration_cast<timespan>(x);
+    return f_(t);
   }
-  bool operator()(vast::json::number n, const vast::timespan_type&) const {
-    std::chrono::duration<vast::json::number> x{n};
-    auto t = std::chrono::duration_cast<vast::timespan>(x);
-    return f_(std::move(t));
+
+  bool operator()(json::number s, const timestamp_type&) const {
+    std::chrono::seconds tmp{size_t(s)};
```
mavam: This throws away the floating point seconds. This doesn't:
```cpp
auto secs = detail::narrow_cast<double>(s);
auto since_epoch = std::chrono::duration_cast<timespan>(secs);
return f_(timestamp{since_epoch});
```

libvast/test/format/json.cpp
```diff
@@ -0,0 +1,79 @@
...
+                        {"host", vast::string_type{}},
+                        {"uri", vast::string_type{}},
+                        {"version", vast::string_type{}},
+                        {"user_agent", vast::string_type{}},
+                        {"request_body_len", vast::count_type{}},
+                        {"response_body_len", vast::count_type{}},
+                        {"status_code", vast::count_type{}},
+                        {"status_msg", vast::string_type{}},
+                        {"tags", vast::vector_type{vast::string_type{}}},
+                        {"resp_fuids", vast::vector_type{vast::string_type{}}}}
```
mavam: Do we need the `vast::` prefixes here, given that you have `using namespace vast` above?

libvast/test/format/json.cpp
```diff
@@ -0,0 +1,79 @@
...
+{"ts":"2011-08-14T20:37:05.912842Z","uid":"CyOGV53NZM39U6MpVh","id.orig_h":"147.32.84.165","id.orig_p":1391,"id.resp_h":"137.254.16.78","id.resp_p":80,"trans_depth":1,"method":"GET","host":"dl.javafx.com","uri":"/javafx-cache.jnlp","version":"1.1","user_agent":"JNLP/6.0 javaws/1.6.0_26 (b03) Java/1.6.0_26","request_body_len":0,"response_body_len":0,"status_code":304,"status_msg":"Not Modified","tags":[]})__";
+
+} // namespace
+
+TEST(json reader) {
+  using reader_type = format::json::reader<format::json::default_selector>;
+  reader_type reader{defaults::system::table_slice_type,
+                     std::make_unique<std::istringstream>(
+                       std::string{http_log})};
+  vast::schema s;
```
mavam:
```suggestion
  schema s;
```

libvast/src/format/json.cpp
```diff
@@ -57,19 +57,19 @@ struct convert {
     return port{detail::narrow_cast<port::number_type>(n)};
   }
 
-  expected<data> operator()(json::number s, const timespan_type&) const {
+  expected<data> operator()(json::number s, const timestamp_type&) const {
     auto secs = std::chrono::duration<json::number>(s);
```
mavam: I like this much better than relying on `double_seconds`!

scripts/integration/integration.py
```diff
@@ -434,7 +440,7 @@ def main():
     parser = argparse.ArgumentParser(
         description='Test runner',
         formatter_class=argparse.ArgumentDefaultsHelpFormatter)
-    # TODO: add verbose mode to print pasteable commands
+    # TODO: add leveled logging (--verbose n)
```
mavam: See #489 for this.

PR #489, opened 2019-04-30
Polish integration test script

mavam:


scripts/integration/integration.py
```diff
@@ -486,14 +497,39 @@ def main():
...
+    parser.add_argument(
+        '-v',
+        '--verbosity',
+        default='DEBUG',
+        help='Set the logging verbosity')
 
     args = parser.parse_args()
+
+    # Setup logging.
+    LOGGER.setLevel(logging.DEBUG)
```
tobim: It is not clear how this setter interacts with the one on the `StreamHandler` or which one takes precedence. Also should this not be `args.verbosity`?

scripts/integration/integration.py
```diff
@@ -486,14 +497,39 @@ def main():
...
+    plain_formatter = logging.Formatter(fmt)
+    formatter = colored_formatter if sys.stdout.isatty() else plain_formatter
+    ch = logging.StreamHandler()
+    ch.setLevel(args.verbosity)
+    ch.setFormatter(formatter)
+    LOGGER.addHandler(ch)
+    class ShutdownHandler(logging.Handler):
+        def emit(self, record):
+            logging.shutdown()
+            sys.exit(1)
```
tobim: Why do we need `sys.exit()` in this handler? Isn't there a possible race with `handle_exit_signal()`?
Please add a comment in case this is intentional.

scripts/integration/integration.py
```diff
@@ -486,14 +497,39 @@ def main():
...
+    LOGGER.addHandler(sh)
+
     if not args.set:
         args.set = Path(__file__).resolve().parent / 'default_set.yaml'
     args.set = args.set.resolve()
+    LOGGER.debug(f'resolved test set path to {args.set}')
 
     test_file = open(args.set, 'r')
-    test_dict = yaml.load(test_file)
+    test_dict = yaml.full_load(test_file)
```
tobim: Interesting. Does this work around that annoying github warning?

scripts/integration/integration.py
mavam: The idea is that a the `critical` log-level automatically terminates the process, e.g., when something fails badly internally. I looks like we could add this instead of `sys.exit(1)`:
```
    signal_subprocs(signal.SIGINT)
    sys.exit(1)
```

...or slightly tweak the shutdown process with a unified function, factored from `handle_exit_signal`.

scripts/integration/integration.py
mavam: This verbosity level is the global level. It could certainly be `args.verbosity`. My understanding is that the individual stream handlers have their own filtering that precedes the filtering of the global log level.

scripts/integration/integration.py
mavam: Yes, that's the issue. I don't know how github tracks this, though. If they go just by used version, then we need to do a bit more than this fix.

mavam:
[ch5942]

scripts/integration/integration.py
tobim: I think in this case we can modify `handle_exit_signal` to:
```python
  signal_subprocs(signum)
  LOGGER.critical('Got signal {signum}, shutting down')
```

scripts/integration/integration.py
mavam: I've added a new commit to update the version in the `requirements.txt`. This should fix the warning.

mavam:
@tobim This PR is now ready for the final review.

PR #492, opened 2019-05-02
Add support for Suricata JSON input

tobim:


libvast/src/format/json.cpp
```diff
@@ -145,15 +145,22 @@ caf::error add(table_slice_builder& builder, const vast::json::object& xs,
...
+      VAST_WARNING_ANON("json-reader", "could not convert", field.name, ":",
+                        to_string(v));
       return x.error();
-    if (!builder.add(make_data_view(*x)))
+    }
+    if (!builder.add(make_data_view(*x))) {
+      VAST_WARNING_ANON("json-reader", "could not convert", field.name, ":",
+                        to_string(v));
       return make_error(ec::type_clash, field.name, ":", to_string(v));
+    }
```
mavam: Could we use the selector name for the logging subject?

libvast/test/format/json.cpp
```diff
@@ -140,4 +145,21 @@ TEST(json reader) {
...
+  std::vector<table_slice_ptr> slices;
+  auto add_slice = [&](table_slice_ptr ptr) {
+    slices.emplace_back(std::move(ptr));
+  };
+  auto [err, num] = reader.read(2, 5, add_slice);
+  CHECK_EQUAL(err, caf::none);
+  CHECK_EQUAL(num, 2);
+  CHECK_EQUAL(slices[0]->rows(), 2);
+  CHECK_EQUAL(slices[0]->columns(), 25);
+  CHECK(slices[0]->at(0, 19) == count{4520});
```
mavam: You could use `view<count>{4520}` to signal the reader the LHS type, but `count` gets promoted also.

libvast/vast/format/json/suricata.hpp
```diff
@@ -0,0 +1,102 @@
...
+#include "vast/format/multi_layout_reader.hpp"
+#include "vast/json.hpp"
+#include "vast/schema.hpp"
+#include "vast/table_slice_builder.hpp"
+#include "vast/view.hpp"
+
+namespace vast::format::json {
+
+struct suricata {
+  caf::optional<vast::record_type> operator()(const vast::json::object& j) {
```
mavam: Using `xs` instead of `j` may help in understanding the container aspect of the JSON object. Just a nit, though, feel free to ignore.

libvast/vast/format/json/suricata.hpp
```diff
@@ -0,0 +1,102 @@
...
+    auto path = j.find("event_type");
```
mavam: Why is this variable called `path`? I would find simple iterator naming more intuitive.

```suggestion
    auto i = j.find("event_type");
```

(My hunch is that this is from the Zeek streaming JSON, which uses a `_path` field.)

libvast/vast/format/json/suricata.hpp
```diff
@@ -0,0 +1,102 @@
...
+    if (path == j.end())
+      return caf::none;
+    auto p = caf::get_if<std::string>(&path->second);
```
mavam:
```suggestion
    auto event_type = caf::get_if<json::string>(&path->second);
```

libvast/vast/format/json/suricata.hpp
```diff
@@ -0,0 +1,102 @@
...
+    if (!p)
```
mavam: If the value is not a string, something is really off. This warrants a warning in my opinion.

libvast/vast/format/json/suricata.hpp
```diff
@@ -0,0 +1,102 @@
...
+      return caf::none;
+    auto it = types.find(*p);
+    if (it == types.end())
+      return caf::none;
+    auto type = it->second;
+    if (j.size() > type.fields.size()) {
+      extend(type, j);
+    }
```
mavam: No braces needed.

libvast/vast/format/json/suricata.hpp
```diff
@@ -0,0 +1,102 @@
...
+    if (path == j.end())
+      return caf::none;
+    auto p = caf::get_if<std::string>(&path->second);
+    if (!p)
+      return caf::none;
+    auto it = types.find(*p);
+    if (it == types.end())
+      return caf::none;
+    auto type = it->second;
+    if (j.size() > type.fields.size()) {
```
mavam: It would be nice to document briefly what we are doing here, i.e., what does `extend` do?

libvast/vast/format/json/suricata.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+  vast::schema schema() const {
+    vast::schema result;
+    for (auto& [key, value] : types)
+      result.add(value);
+    return result;
+  }
+
+  static const char* name() {
+    return "suricata-json-reader";
```
mavam:
```suggestion
"suricata-eve-reader"
```

Just an idea, take it or leave it.

libvast/vast/format/json/suricata.hpp
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  std::unordered_map<std::string, vast::record_type> types;
```
mavam:
```suggestion
  std::unordered_map<std::string, record_type> types;
```

(This applies to all occurrences in this patch.)

libvast/vast/format/json/suricata.hpp
```diff
@@ -0,0 +1,102 @@
...
+    if (it == types.end())
+      return caf::none;
+    auto type = it->second;
+    if (j.size() > type.fields.size()) {
+      extend(type, j);
+    }
+    return type;
+  }
+
+  // TODO ch5152: Needs implementation
```
mavam: It'd be nice to document the expected behavior already now and move this todo in the body of the implementation.

libvast/vast/type.hpp
```diff
@@ -650,6 +650,17 @@ struct alias_type final : nested_type<alias_type> {
 
 // -- free functions ----------------------------------------------------------
 
+/// @relates record_type
+template <typename... Rs>
+record_type concat(const Rs&... rs) {
```
mavam: Please document every function that's part of the public API.

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🎁 VAST gained support for importing suricata json logs. Specify the
+  `suricata` format to the `import` command and you're in business.
```
mavam: I like being in business! 😄 I still think we should write the changelog in a very clear and sober style. Here's an alternate suggestion.

> VAST gained support for importing Suricata JSON logs. The `import` command has a new `suricata` format that can ingest EVE JSON output.

libvast/vast/format/json/suricata.hpp
tobim: Your guess is correct. It was a copy & paste.

libvast/vast/format/json/suricata.hpp
tobim: I think just "suricata-reader" is probably the best option, I don't think we will ever support other suricata log formats.

scripts/integration/default_set.yaml
```diff
@@ -121,6 +121,12 @@ tests:
         input: data/json/conn.log.json.gz
       - command: export -N ascii 'duration > 6s'
       - command: export -N ascii '#time >= 2011-08-15T03:48'
+  Node suricata alert:
+    tags: [node, import-export, suricata, eve]
+    steps:
+      - command: import -N -b suricata
```
mavam:
```suggestion
      - command: import -N suricata
```

`-N` implies `-b`, no?

libvast/vast/format/json.hpp
```diff
@@ -52,9 +53,10 @@ class writer : public printer_writer<event_printer> {
 /// @param builder The builder to add the JSON object to.
 /// @param xs The JSON object to add to *builder.
 /// @param layout The record type describing *xs*.
+/// @param name The name of the reader used for logging.
 /// @returns An error iff the operation failed.
 caf::error add(table_slice_builder& builder, const vast::json::object& xs,
-               const record_type& layout);
+               const record_type& layout, const std::string_view name);
```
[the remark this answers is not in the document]
tobim: We'd have to drop the log statement in case of a field not present in the incoming JSON object, but that is probably ok.

libvast/vast/format/json.hpp
```diff
@@ -172,20 +174,22 @@ caf::error reader<Selector>::read_impl(size_t max_events, size_t max_slice_size,
     if (!xs)
       return make_error(ec::type_clash, "not a json object");
     auto layout = selector_(*xs);
-    if (!layout)
-      return make_error(ec::parse_error, "unable to get a layout");
+    if (!layout) {
+      VAST_INFO(this, "unable to get a layout");
```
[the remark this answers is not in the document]
tobim: I removed that log statement. Printing that from the selector makes more sense, because it has more context.

scripts/integration/default_set.yaml
tobim: In a way it does. I removed all `-b`s from commands with `-N`.

PR #494, opened 2019-05-08
Fix behavior of numeric parser for partial input

mavam:
Previously, integral parsers with a maximum digit specification failed when the input contained further digits. This is no longer considered a failure: the parser returns successfully with whatever is has parsed until it reached its digit limit.

### DIFF
--- a/libvast/test/parseable.cpp
+++ b/libvast/test/parseable.cpp
@@ -35,6 +35,7 @@
 #include "vast/si_literals.hpp"
 
 using namespace std::string_literals;
+using namespace std::string_view_literals;
 using namespace vast;
 using namespace vast::parser_literals;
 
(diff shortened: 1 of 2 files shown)


libvast/test/parseable.cpp
```diff
@@ -437,14 +437,22 @@ TEST(signed integral with digit constraints) {
...
+  MESSAGE("partial match with additional digit");
   auto str = "12345"sv;
   auto f = str.begin();
   auto l = str.end();
   CHECK(p.parse(f, l, x));
   REQUIRE(f + 1 == l);
   CHECK_EQUAL(*f, '5');
   CHECK_EQUAL(x, 1234);
+  MESSAGE("partial match with non-digits character");
+  str = "6789x"sv;
```
tobim: I think this test should cover the case of an integral with less than `MaxDigits` digits. So `"678x"sv` would make more sense to me.

PR #497, opened 2019-05-13
Fix default for suricata command

tobim:


scripts/integration/default_set.yaml
```diff
@@ -128,6 +128,7 @@ tests:
     steps:
       - command: import -N suricata
         input: data/suricata/eve.json.gz
```
Neverlord: Isn't defining `input` obsolete now?

Neverlord: Oh, I just noticed that you are running two import commands now. Is this importing the same data? I don't see an updated reference file. Then I think we could split this into two tests. Because in case one of the two commands fails to import than the other masks that failure.

PR #501, opened 2019-05-22
Initialize value indexes at construction

tobim:
This fixes a segfault that happens if a lookup runs on an index that did not receive any data yet.

### DIFF
--- a/libvast/src/value_index.cpp
+++ b/libvast/src/value_index.cpp
@@ -20,7 +20,7 @@ namespace vast {
 
 // -- value_index --------------------------------------------------------------
 
-value_index::value_index(vast::type x) : type_{std::move(x)} {
+value_index::value_index(vast::type t) : type_{std::move(t)} {
   // nop
 }
 
@@ -112,7 +112,8 @@ caf::error inspect(caf::deserializer& source, value_index_ptr& x) {
 
 string_index::string_index(vast::type t, size_t max_length)
   : value_index{std::move(t)},
-    max_length_{max_length} {
+    max_length_{max_length},
+    length_{base::uniform(10, std::log10(max_length) + !!(max_length % 10))} {
 }
 
 caf::error string_index::serialize(caf::serializer& sink) const {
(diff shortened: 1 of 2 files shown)


libvast/src/value_index.cpp
```diff
@@ -482,13 +464,18 @@ sequence_index::sequence_index(vast::type t, size_t max_size)
...
 caf::error sequence_index::serialize(caf::serializer& sink) const {
-  return caf::error::eval(
-    [&] { return value_index::serialize(sink); },
-    [&] { return sink(elements_, size_, max_size_, value_type_); }
-  );
+  return caf::error::eval([&] { return value_index::serialize(sink); },
+                          [&] {
+                            return sink(elements_, size_, max_size_,
+                                        value_type_);
+                          });
```
mavam: `clang-format` needs to be slapped on the fingers here.

Neverlord: Can you go into detail? This looks good to me and seems to follow our style guide.

mavam: In comparison to what we had before, this just looks torn apart.

On a separate note, I don't know why this part of the code got even touched.

tobim: > On a separate note, I don't know why this part of the code got even touched.

No idea, but I might have touched it by accident.

May I suggest adjusting `.clang-format` like this:
```
PenaltyBreakAssignment: 10
PenaltyBreakBeforeFirstCallParameter: 30
PenaltyBreakString: 40
PenaltyExcessCharacter: 100
PenaltyReturnTypeOnItsOwnLine: 5
```
The function here would be formated like this:
```
caf::error sequence_index::serialize(caf::serializer& sink) const {
  return caf::error::eval(
    [&] { return value_index::serialize(sink); },
    [&] { return sink(elements_, size_, max_size_, value_type_); });
}
```
I don't know if the closing parentheses can be moved back to its own line with another option, but I think I actually like it better this way anyway.

tobim: Also note that this change to the formatting options would not affect any other code in this PR.
And looking at behaviour returning functions it is actually quite an improvement because the lambdas don't get aligned after the opening curly any more.

PR #502, opened 2019-05-22
Add --config option to root command

tobim:


libvast/src/system/configuration.cpp
```diff
@@ -89,6 +89,17 @@ caf::error configuration::parse(int argc, char** argv) {
   std::vector<std::string> caf_args;
   std::move(caf_opt, command_line.end(), std::back_inserter(caf_args));
   command_line.erase(caf_opt, command_line.end());
+  auto config_opt = std::find_if(command_line.begin(), command_line.end(),
+                                 [](const std::string& arg) {
+                                   return arg == "-c" || arg == "--config";
```
Neverlord: Doing it this way means that the `--config` parameter behaves differently than all other parameters. The CAF option parser accepts `-c <arg>`, `-c<arg>` (no space) or `--config=<arg>`. This code block now accepts `-c <arg>` or `--config <arg>`.

I think we should stick to the behavior of CAF's parser. In order to not hand-roll our own parsing, we could rely on `caf::config_option_set::parse` instead. That way, the parsing remains a black box and our code can't run out of sync with the way CAF treats other parameters.

PR #506, opened 2019-06-07
Report iterator position with command invocation

Neverlord:
In case of parsing of command fails, we still need the original iterator position to figure out *where* things went wrong. This set of changes puts the error state into the `invocation` for this purpose.

### DIFF
--- a/libvast/src/command.cpp
+++ b/libvast/src/command.cpp
@@ -113,9 +113,9 @@ command::invocation parse(const command& root, command::argument_iterator first,
 }
 
 caf::message run(command::invocation& invocation, caf::actor_system& sys) {
-  auto& cmd = *invocation.target;
   if (!invocation)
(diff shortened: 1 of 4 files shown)


libvast/src/command.cpp
```diff
@@ -103,17 +104,18 @@ caf::error parse(command::invocation& result, const command& cmd,
...
   command::invocation result;
   if (auto err = parse(result, root, first, last))
-    return err;
+    result.error = std::move(err);
   return result;
 }
 
 caf::message run(command::invocation& invocation, caf::actor_system& sys) {
   auto& cmd = *invocation.target;
+  if (!invocation)
```
tobim: Shoudln't this be checked before reaching into `*invocation`?

Neverlord: Yep, good catch. Makes more sense to check the error first.

PR #507, opened 2019-06-07
Fix quoted string parser for empty strings 

tobim:


libvast/test/parseable.cpp
```diff
@@ -179,56 +179,64 @@ TEST(quoted string) {
...
-  MESSAGE("no escaped chars");
-  auto str = "'foobar'"s;
+  MESSAGE("empty string");
+  auto str = "''"s;
   auto f = str.begin();
   auto l = str.end();
   CHECK(p(f, l, attr));
-  CHECK(attr == "foobar");
-  CHECK(f == l);
+  CHECK_EQUAL(attr, "");
```
mavam:
```suggestion
  CHECK(attr.empty());
```

PR #510, opened 2019-06-12
Externalize hard-coded C++ schemas

tobim:
TODO:
- [x] Transform suricata schema into a schema file.
- [x] Changelog entry.


libvast/vast/event_types.hpp
```diff
@@ -0,0 +1,38 @@
...
+  /// @returns true on success or false if registry was alread initialized.
+  static bool init(schema s);
+
+  /// Retrieves a pointer to the system-wide type registry.
+  /// @returns nullptr if registry is not initialized.
+  static const schema* get();
+
+private:
+  static schema& get_impl();
+  static bool initialized;
```
tobim: `initalized` should probably only be visible from the source file.

libvast/vast/system/reader_command.hpp
```diff
@@ -70,21 +71,35 @@ caf::message reader_command(const command& cmd, caf::actor_system& sys,
...
+    if (update) {
+      if (!schema || schema->empty()) {
+        reader_schema = *update;
+      } else {
+        reader_schema = *schema;
+        reader_schema.update(*update);
+      }
+      schema = &reader_schema;
+    }
+    if (!update && update.error() != caf::no_error)
```
mavam:
```suggestion
   else  if (update.error() != caf::no_error)
```

Plus additional braces.

libvast/vast/system/reader_command.hpp
```diff
@@ -70,21 +71,35 @@ caf::message reader_command(const command& cmd, caf::actor_system& sys,
...
+      return caf::make_message(ec::invalid_configuration, "");
```
mavam: Is there anything to do for an `else` branch?

libvast/vast/system/reader_command.hpp
```diff
@@ -70,21 +71,35 @@ caf::message reader_command(const command& cmd, caf::actor_system& sys,
...
-      schema = load_schema_file(*sf);
-    if (!schema && schema.error() != caf::none)
-      return make_message(std::move(schema.error()));
+    auto update = [&]() -> caf::expected<vast::schema> {
+      if (sc)
+        return to<vast::schema>(*sc);
+      if (sf)
+        return load_schema_file(*sf);
+      return caf::no_error;
+    }();
```
mavam: Mind explaining (and documenting via a comment) why you created an anonymous function and invoke it directly?

mavam: Mind explaining (and documenting via a comment) why you created an anonymous function and invoke it directly?

libvast/src/schema.cpp
```diff
@@ -183,4 +185,13 @@ bool convert(const schema& s, json& j) {
   return true;
 }
 
+caf::expected<schema> load_schema_file(const path& sf) {
+  if (sf.empty())
+    return make_error(ec::filesystem_error, "");
```
mavam:
```suggestion
    return make_error(ec::filesystem_error, "empty path");
```

mavam: Actually, isn't `load_contents` taking care of the case when loading from file doesn't work?

libvast/vast/schema.hpp
```diff
@@ -79,4 +79,6 @@ class schema : detail::equality_comparable<schema> {
 
 bool convert(const schema& s, json& j);
 
+caf::expected<schema> load_schema_file(const path& sf);
```
mavam: You could consider to name this function `load_schema` with two overloads, one with a `const path& file` argument and the other with `const vector<path>& dirs`.

libvast/src/schema.cpp
```diff
@@ -194,4 +194,36 @@ caf::expected<schema> load_schema_file(const path& sf) {
   return to<schema>(*str);
 }
 
+caf::expected<schema>
+load_schema_dirs(const std::vector<std::string>& schema_paths) {
+  vast::schema types;
+  for (const auto& d : schema_paths) {
```
mavam:
```suggestion
  for (const auto& dir : schema_paths) {
```

In my eyes, `d` is not a common idiomatic variable name.

libvast/src/schema.cpp
```diff
@@ -194,4 +194,36 @@ caf::expected<schema> load_schema_file(const path& sf) {
...
+    auto schema_dir = path{d};
+    if (!exists(schema_dir))
+      break;
+    vast::schema directory_schema;
+    for (auto f : directory(schema_dir)) {
+      if (f.extension() == ".schema" && exists(f)) {
+        auto k = f.kind();
+        switch (k) {
```
mavam:
```suggestion
        switch (f.kind()) {
```

libvast/src/schema.cpp
```diff
@@ -194,4 +194,36 @@ caf::expected<schema> load_schema_file(const path& sf) {
...
+          default:
+            break;
+          case path::regular_file:
+          case path::symlink: {
+            auto schema = load_schema_file(f);
+            if (!schema)
+              return schema.error();
+            if (auto merged = schema::merge(directory_schema, *schema))
+              directory_schema = *merged;
```
mavam:
```suggestion
              directory_schema = std::move(*merged);
```

libvast/vast/event_types.hpp
mavam: Yeah, but then `get_impl()` can equally go into an anonymous namespace in the implementation.

vast/vast.cpp
```diff
@@ -74,6 +76,20 @@ int main(int argc, char** argv) {
     return EXIT_FAILURE;
   }
   caf::actor_system sys{cfg};
+  // Load event types.
+  using string_list = std::vector<std::string>;
+  if (auto schema_paths = caf::get_if<string_list>(&cfg,
+                                                   "system.schema-paths")) {
```
mavam: This feels a bit bulky in all one line. Mind splitting it up?

vast/vast.cpp
```diff
@@ -74,6 +76,20 @@ int main(int argc, char** argv) {
...
+    if (auto schema = load_schema_dirs(*schema_paths)) {
+      event_types::init(*std::move(schema));
+    } else {
+      std::cerr << "failed to read schema dirs: " << to_string(schema.error())
+                << std::endl;
```
mavam: I think we can use the logger here already.

schema/suricata.schema
```diff
@@ -0,0 +1,232 @@
...
+  pcap_cnt: count,
+  src_ip: addr,
+  src_port: port,
+  dest_ip: addr,
+  dest_port: port,
+  proto: string,
+  community_id: string
+}
+
+type suricata.flow_ = record{
```
mavam: What is the reason for the trailing `_` in the type name?

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🔄 Default schema definitions for certain `import` formats changed from
+  hard-coded to runtime evaluated. The default location of the schema
```
```suggestion
  hard-coded to runtime-evaluated. The default location of the schema
```

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ Every entry has a category for which we use the following visual abbreviations:
...
+  definition files is $prefix/share/vast/schema. Currently this is used for the
+  suricata json log reader.
```
```suggestion
  Suricata JSON log reader.
```

tobim:
> Aside from my rather mechanical suggestions, I wonder why the Suricata unit test remains still disabled.

Because I think unit tests should never do IO.

libvast/vast/system/reader_command.hpp
```diff
@@ -117,6 +117,8 @@ caf::message reader_command(const command& cmd, caf::actor_system& sys,
       }
     }
     Reader reader{slice_type};
+    if (schema)
+      reader.schema(*schema);
```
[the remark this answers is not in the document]
tobim: In case we don't have a schema supplied via `--schema` or `--schema-file`, `schema` is a pointer onto the global `event_types`. Maybe `Reader::schema(vast::schema)` should get a `const&` overload that we can use for the unmodified variant instead?

libvast/src/schema.cpp
tobim: This is just code moved from `reader_command.hpp` to reuse if from `load_schema_dirs()`
`load_contents()` doesn't seem to handle that case correctly, I think it would just return an empty string.

schema/suricata.schema
tobim: It is only a component of a type, but not a full type that gets written by suricata. I'm thinking of renaming it to `suricata.component.flow`.

libvast/vast/system/reader_command.hpp
tobim: The final `else` means we take the global schema, but that is how we initialized thinks at the beginning.

libvast/vast/schema.hpp
```diff
@@ -50,9 +50,16 @@ class schema : detail::equality_comparable<schema> {
   /// @returns `true` on success.
   bool add(const type& t);
 
+  /// Updates the schema with definitions from another.
+  /// @param new_def The definitions to add or replace.
+  void update(const schema& new_def);
```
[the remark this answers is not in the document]
tobim: `merge()` can fail, so turning that into an in-place operation would make the algorithm more expensive. I will change `update()` to a static.

libvast/vast/system/reader_command.hpp
tobim: This is to avoid pre-declaring `caf::expected<vast::schema> update;`. For me, this feels more maintainable. It's a stylistic choice, therefore I don't think it warrants a comment in the code.

libvast/vast/system/reader_command.hpp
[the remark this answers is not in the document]
mavam: In this case, I'd say let's go for value semantics. A schema is a list of `type`s and they have reference semantics. We're also not on the critical path.

mavam:
> Because I think unit tests should never do IO.

Okay, that's a valid point. But I don't see the relationship to this PR.

To move things forward, I'm going to merge the PR. Could you make sure that we have a story to re-enable the test, after the necessary refactoring?

tobim:
> > Because I think unit tests should never do IO.
> 
> Okay, that's a valid point. But I don't see the relationship to this PR.

The test cannot complete without loading the schema definitions, the point of this PR was to move the definition from hard-coded c++ to a runtime parsed proprietary syntax.

> To move things forward, I'm going to merge the PR. Could you make sure that we have a story to re-enable the test, after the necessary refactoring?

I think we have 2 options here:
1. Rely on integration testing for the reader functioning.
2. Use `xxd` to convert the schema files to headers and parse them in a test fixture.

mavam:
Thanks for elaborating, I know see the conundrum. Your proposition (2) strikes me as the most practical one. We have an integration test already, so it looks like we're good for now.

PR #513, opened 2019-06-15
Allow writers to consume batches of events

Neverlord:
This tweak adds a new virtual function to `writer` that allows subtypes to get the entire batch at once. It makes no difference to VAST at the moment, but I'm going to need the new member function for tackling `Arrow` export (since I need to see the entire batch all at once to create an Arrow table slice [record batch] from it).

### DIFF
--- a/libvast/vast/system/sink.hpp
+++ b/libvast/vast/system/sink.hpp
@@ -72,6 +72,9 @@ caf::behavior sink(caf::stateful_actor<sink_state<Writer>>* self,
   if (max_events > 0) {
     VAST_DEBUG(self, "caps event export at", max_events, "events");
     st.max_events = max_events;
+  } else {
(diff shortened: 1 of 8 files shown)


libvast/vast/system/sink.hpp
```diff
@@ -80,26 +83,35 @@ caf::behavior sink(caf::stateful_actor<sink_state<Writer>>* self,
     }
   );
   return {
-    [=](const std::vector<event>& xs) {
+    [=](std::vector<event>& xs) {
```
mavam: Why did you drop constness here? Is this just temporary until writers will work with table slices? If so, then I'm fine with this convenience change to resize the batch later.

Neverlord: Yes, that's only a workaround for now. Eventually we'll switch to table slices.

PR #514, opened 2019-06-17
Add write overload to the writer for table slices

Neverlord:
This refactoring simply allows us to port the EXPORTER to table slices
while independently updating writer implementations.


Neverlord:
Good idea! Done.

PR #516, opened 2019-06-18
Implement an evaluate function for table slices

Neverlord:
Currently, the EXPORTER uses an `event_evaluator` for performing the candidate check. If we wish to switch to a table-slice based API then we first need this new function for checking candidates in table slices.

### DIFF
--- a/libvast/src/expression_visitors.cpp
+++ b/libvast/src/expression_visitors.cpp
@@ -572,15 +572,23 @@ bool table_slice_row_evaluator::operator()(const predicate& p) {
 }
 
 bool table_slice_row_evaluator::operator()(const attribute_extractor& e,
-                                          const data& d) {
+                                           const data& d) {
   // FIXME: perform a transformation on the AST that replaces the attribute
   // with the corresponding function object.
   if (e.attr == system::type_atom::value)
     return evaluate(slice_.layout().name(), op_, d);
-  // TODO: implement me
-  // if (e.attr == system::time_atom::value)
-  //   return evaluate(event_.timestamp(), op_, d);
-  return false;
+  // Find the column with attribute 'time'.
+  auto pred = [](auto& x) {
+    return caf::holds_alternative<timestamp_type>(x.type)
+           && has_attribute(x.type, "time");
+  };
+  auto& fs = slice_.layout().fields;
+  auto i = std::find_if(fs.begin(), fs.end(), pred);
+  if (i == fs.end())
+    return false;
+  // Compare timestamp at given cell.
+  auto pos = static_cast<size_t>(std::distance(fs.begin(), i));
+  return evaluate_view(slice_.at(row_, pos), op_, make_view(d));
 }
 
 bool table_slice_row_evaluator::operator()(const type_extractor&, const data&) {
--- a/libvast/src/expression_visitors.cpp
+++ b/libvast/src/expression_visitors.cpp
@@ -577,20 +577,20 @@ bool table_slice_row_evaluator::operator()(const attribute_extractor& e,
   // with the corresponding function object.
   if (e.attr == system::type_atom::value)
     return evaluate(slice_.layout().name(), op_, d);
-  // Find the column with attribute 'time'.
-  if (e.attr != system::time_atom::value)
-    return false;
-  auto pred = [](auto& x) {
-    return caf::holds_alternative<timestamp_type>(x.type)
-           && has_attribute(x.type, "time");
-  };
-  auto& fs = slice_.layout().fields;
-  auto i = std::find_if(fs.begin(), fs.end(), pred);
-  if (i == fs.end())
-    return false;
-  // Compare timestamp at given cell.
-  auto pos = static_cast<size_t>(std::distance(fs.begin(), i));
-  return evaluate_view(slice_.at(row_, pos), op_, make_view(d));
+  if (e.attr == system::time_atom::value) {
+    auto pred = [](auto& x) {
+      return caf::holds_alternative<timestamp_type>(x.type)
+             && has_attribute(x.type, "time");
+    };
+    auto& fs = slice_.layout().fields;
+    auto i = std::find_if(fs.begin(), fs.end(), pred);
+    if (i == fs.end())
+      return false;
+    // Compare timestamp at given cell.
+    auto pos = static_cast<size_t>(std::distance(fs.begin(), i));
+    return evaluate_view(slice_.at(row_, pos), op_, make_view(d));
+  }
+  return false;
 }
 
 bool table_slice_row_evaluator::operator()(const type_extractor&, const data&) {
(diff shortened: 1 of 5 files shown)


libvast/src/expression_visitors.cpp
```diff
@@ -535,6 +537,73 @@ bool event_evaluator::operator()(const data_extractor& e, const data& d) {
...
+bool table_slice_row_evaluator::operator()(const key_extractor&, const data&) {
+  die("key extractor should have been resolved at this point");
+}
+
+bool table_slice_row_evaluator::operator()(const data_extractor& e,
+                                           const data& d) {
+  // Our layout is always flat. Hence, the offset needs to have a depth of
+  // exactly one since it simply represents the column.
+  if (e.offset.size() != 1 || e.type != slice_.layout())
+    return false;
```
mavam: It looks like this could be an assertion, but I haven't fully checked the caller path.

libvast/src/view.cpp
```diff
@@ -192,4 +192,77 @@ bool type_check(const type& t, const data_view& x) {
...
+                          //        != rhs->end();
+                          return false;
+                        },
+                        [](auto lhs, view<set> rhs) {
+                          // TODO: implement me
+                          // return std::find(rhs->begin(), rhs->end(), lhs)
+                          //        != rhs->end();
+                          return false;
+                        }),
+                      x, y);
```
mavam: This looks pretty crammed when coding only in 2/3 of the column width. Mind factoring the visitor?

libvast/test/expression_evaluation.cpp
```diff
@@ -153,4 +156,23 @@ TEST(evaluation - schema) {
   CHECK(caf::holds_alternative<caf::none_t>(*ast_resolved));
 }
 
+TEST(evaluation - table slice rows) {
+  // Calling the fixture ctor makes sure the slices are available.
+  fixtures::events dummy;
```
mavam: Isn't the ctor called upon test entry? The comment unfortunately doesn't help me understand what's going on.

libvast/vast/expression_visitors.hpp
```diff
@@ -165,6 +165,41 @@ struct event_evaluator {
   relational_operator op_;
 };
 
+/// Evaluates a single row of a table slice over a [resolved](@ref
+/// type_extractor) expression.
```
mavam: You could add a `@relates` here since the implementation looks very familiar to one that I already know.

libvast/vast/view.hpp
```diff
@@ -450,4 +450,14 @@ data materialize(data_view xs);
 /// @returns `true` if *t* is a valid type for *x*.
 bool type_check(const type& t, const data_view& x);
 
+/// Evaluates a data predicate.
+/// @param lhs The LHS of the predicate.
+/// @param op The relational operator.
+/// @param rhs The RHS of the predicate.
+/// @note This function unfortunately needs to use the `_view` suffix because
+///       it otherwise produces ambiguous function calls with the `evaluate`
+///       function in `data.hpp`.
```
mavam: This one could be related to other evaluation functions.

libvast/src/expression_visitors.cpp
```diff
@@ -572,15 +572,23 @@ bool table_slice_row_evaluator::operator()(const predicate& p) {
...
-  return false;
+  // Find the column with attribute 'time'.
+  auto pred = [](auto& x) {
+    return caf::holds_alternative<timestamp_type>(x.type)
+           && has_attribute(x.type, "time");
+  };
+  auto& fs = slice_.layout().fields;
+  auto i = std::find_if(fs.begin(), fs.end(), pred);
+  if (i == fs.end())
+    return false;
```
mavam: You are missing a check for `if (e.attr == system::time_atom::value) { /* diff here */}`. Then `return false` at the very end of the function. This way, new attributes can be added easily.

libvast/src/view.cpp
```diff
@@ -204,9 +204,27 @@ bool evaluate_view(const data_view& lhs, relational_operator op,
...
+                            auto pred = [&](const auto& rhs_element) {
+                              return caf::visit(equals_lhs, rhs_element);
+                            };
+                            return std::find_if(rhs->begin(), rhs->end(), pred)
+                                   != rhs->end();
+                            return false;
+                          } else {
+                            // Default case.
+                            return false;
+                          }
```
mavam: It's really hard to read this crammed code with so many line breaks. Mind factoring the visitor into a separate function?

libvast/src/view.cpp
Neverlord: Not at all.

libvast/src/expression_visitors.cpp
Neverlord: I think an assertion should only check for program invariants or function preconditions.

libvast/test/expression_evaluation.cpp
Neverlord: Only if the fixture actually inherits from `fixtures::events`. I did that instead to make the dummies disappear.

libvast/src/expression_visitors.cpp
Neverlord: Whoops. 🙈

libvast/src/view.cpp
```diff
@@ -192,6 +192,49 @@ bool type_check(const type& t, const data_view& x) {
...
+  }
+
+  bool operator()(view<address> lhs, view<subnet> rhs) {
+    return rhs.contains(lhs);
+  }
+
+  bool operator()(view<subnet> lhs, view<subnet> rhs) {
+    return rhs.contains(lhs);
+  }
+};
```
mavam: There are some issues with the argument signatures:
1. Sometimes you use a non-const reference, other times a value.
2. Since this is not a mutating function, references should be const
3. I would replace the value with const references or value. If the predicate doesn't get inlined, it might be a good idea to use const references to avoid unnecessary bumping of the ref-count in the type-erased container views.

libvast/src/expression_visitors.cpp
```diff
@@ -578,6 +578,8 @@ bool table_slice_row_evaluator::operator()(const attribute_extractor& e,
   if (e.attr == system::type_atom::value)
     return evaluate(slice_.layout().name(), op_, d);
   // Find the column with attribute 'time'.
+  if (e.attr != system::time_atom::value)
+    return false;
```
mavam: I was actually suggestion to use `==` instead of `!=`, because this makes adding new paths for new attributes easier:

```cpp
if (e.attr == system::time_atom::value) {
  // ...
}
if (e.attr == system::other_upcoming_attribute::value) {
  // ...
}
return false;
```

libvast/src/expression_visitors.cpp
mavam: Isn't this a precondition? It doesn't make sense to call the visitor for nested types in my eyes.

Neverlord:
@mavam I've integrated your feedback, except for the `@relates` suggestion. So far we use `@relates` to document free functions that belong to a particular type, which is the use case this feature is intended for: https://www.star.bnl.gov/public/comp/sofi/doxygen/commands.html#cmdrelates. Relating free functions to other free functions seems odd to me, especially since the function names already transfer that relation in my opinion.

mavam:
I guess I've been using `@relates` and `@see` interchangeably. I agree that `@relates` is not the right directive here. We don't use a lot of `@see` at this point, but we should keep it in mind for the future to facilitate finding related API functionality.

PR #518, opened 2019-06-22
Port SINK to the table slice API

Neverlord:
This set of changes allows the EXPORTER to send table slices to the SINK and drops the conversion to `vector<event>`.

The two new free functions `split` and `truncate` that operate on table slices could be implemented more efficiently as virtual member functions, but I think we should not convolute the `table_slice` interface more than necessary. Those functions are only needed if a table slice exceeds the number of requested events, so they are probably not in a critical path. Should either one of these functions show up in a profile while benchmarking VAST then we can still move those functions into the `table_slice` interface and implement them efficiently for each table slice type.

### DIFF
--- a/libvast/vast/system/sink.hpp
+++ b/libvast/vast/system/sink.hpp
@@ -31,6 +31,7 @@
 #include "vast/system/atoms.hpp"
 #include "vast/system/instrumentation.hpp"
 #include "vast/system/query_status.hpp"
+#include "vast/table_slice.hpp"
 
 namespace vast::system {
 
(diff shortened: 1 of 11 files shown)


libvast/vast/system/sink.hpp
```diff
@@ -83,8 +84,8 @@ caf::behavior sink(caf::stateful_actor<sink_state<Writer>>* self,
     }
   );
   return {
-    [=](std::vector<event>& xs) {
```
tobim: This is also the only use of `event` in the file.

Neverlord:
@tobim all done. It's great that you keep an eye open for unnecessary includes! Avoiding those helps us to bring down the build time little by little. 👍

PR #520, opened 2019-06-25
Fix schema loading for relocated installations

tobim:
This PR implements 2 changes:
1. The `schema` folder will now be looked up relative to the location of the loaded executable.
2. The `--schema-paths` does not overwrite the default path any more.


libvast/src/detail/process.cpp
```diff
@@ -0,0 +1,62 @@
...
+ * LICENSE file found in the top-level directory of this distribution and at  *
+ * http://vast.io/license. No part of VAST, including this file, may be       *
+ * copied, modified, propagated, or distributed except according to the terms *
+ * contained in the LICENSE file.                                             *
+ ******************************************************************************/
+
+#include "vast/detail/process.hpp"
+
+#include <dlfcn.h>
+#if __linux__
```
mavam: We have `VAST_*` macros for the OS definitions. Let's use these for all OS-based preprocessor stuff.

libvast/src/detail/process.cpp
```diff
@@ -0,0 +1,62 @@
...
+#  include <limits.h>
+#  include <sys/stat.h>
+#  include <unistd.h>
+#endif
+
+namespace vast::detail {
+
+namespace {
+
+caf::optional<vast::path> objectpath_dynamic(const void* addr) {
```
mavam: Why do we need to qualify `path` with `vast::`?

libvast/src/detail/process.cpp
```diff
@@ -0,0 +1,62 @@
...
+#if __linux__
+  struct stat sb;
+  auto self = "/proc/self/exe";
+  if (lstat(self, &sb) == -1)
+    return caf::none;
+  auto size = sb.st_size ? sb.st_size + 1 : PATH_MAX;
+  std::vector<char> buf(size);
+  if (readlink(self, buf.data(), size) == -1) {
+    return caf::none;
+  }
```
mavam: No braces needed.

libvast/src/detail/process.cpp
```diff
@@ -0,0 +1,62 @@
...
+#endif
+}
+
+} // namespace
+
+caf::optional<vast::path> objectpath(const void* addr) {
+  auto result = objectpath_dynamic(addr);
+  if (!result)
+    result = objectpath_static();
+  return result;
```
mavam: This reads more intuitively to me:
```cpp
if (auto result = objectpath_dynamic(addr))
  return result;
return objectpath_static();
```

libvast/vast/detail/process.hpp
```diff
@@ -0,0 +1,34 @@
...
+ ******************************************************************************/
+
+#pragma once
+
+#include "vast/filesystem.hpp"
+#include <caf/optional.hpp>
+
+namespace vast::detail {
+
+/// Get the path of a shared library.
```
mavam: We usually use third person singular to describe what a function does. Here, `Finds` or `Locates` seems more appropriate to describe what this function does.

Also, a complete documentation includes a description, contains Doxygen directives for all arguments, and describes the return value. Here, `addr` is not covered.

scripts/integration/default_set.yaml
```diff
@@ -1,12 +1,16 @@
+config-file: vast.conf
+
 fixtures:
   BiTopologyTester:
     enter: | # python
       node0 = Server(self.cmd,
-                     ['-e', ':{}'.format(VAST_PORT),
+                     [f'--config-file={self.config_file}',
+                      '-e', ':{}'.format(VAST_PORT),
```
mavam: Probably an oversight: format string interpolation and `.format` calls in the same statement.

vast/vast.cpp
```diff
@@ -77,18 +79,23 @@ int main(int argc, char** argv) {
...
   using string_list = std::vector<std::string>;
-  auto schema_paths = caf::get_if<string_list>(&cfg, "system.schema-paths");
-  if (schema_paths) {
-    if (auto schema = load_schema(*schema_paths)) {
-      event_types::init(*std::move(schema));
-    } else {
-      VAST_ERROR_ANON("failed to read schema dirs:", to_string(schema.error()));
-      return EXIT_FAILURE;
-    }
+  if (auto user_dirs = caf::get_if<string_list>(&cfg, "system.schema-paths"))
```
mavam: `schema_dirs` and `user_dirs` don't work well together for my taste. If we have `user_dirs`, then I'm expecting `system_dirs`, `default_dirs`, `program_dirs`, or some sort of other `*_dirs` that describes the shipped dirs.

mavam:
Please also update the changelog.

libvast/vast/detail/process.hpp
tobim: That comment is outdated, I will replace it with a correct one.

libvast/src/detail/process.cpp
```diff
@@ -24,39 +25,37 @@ namespace vast::detail {
...
   std::vector<char> buf(size);
-  if (readlink(self, buf.data(), size) == -1) {
-    return caf::none;
-  }
+  if (readlink(self, buf.data(), size) == -1)
+    return make_error(ec::unspecified, "readlink() returned with error");
   return path{buf.data()};
 #else
-  return caf::none;
+  return make_error(ec::unimplemented, "");
```
mavam: Good call to improve the specificity of the error code. However, either just return the error code or add a non-empty string, but `""` adds more noise than benefit.

tobim: I just assumed the second argument was mandatory.

PR #521, opened 2019-06-27
Add and apply .cmake-format 

tobim:


libvast/CMakeLists.txt
```diff
@@ -1,162 +1,157 @@
...
+    src/time.cpp
+    src/timestamp_synopsis.cpp
+    src/to_events.cpp
+    src/type.cpp
+    src/uuid.cpp
+    src/value.cpp
+    src/value_index.cpp
+    src/value_index_factory.cpp
+    src/view.cpp
+    src/wah_bitmap.cpp)
```
Neverlord: This change breaks our [`add_type`](https://github.com/vast-io/vast/blob/master/scripts/add_type) script. I'm using this quite extensively, so to me that's more important than having a coherent CMake style TBH. 😕

Neverlord: Maybe we can settle on excluding such `src` blocks via `format off/on` comments?

tobim: This fails only because the closing ')' is on the same line as the last parameter, right? Why don't we just extend to tools parsing logic to accept that style.

PR #522, opened 2019-06-29
Add support for new data predicates

mavam:
A *data predicate* is a short-hand for a predicate of the following form:

- LHS: type extractor
- op:  `==`
- RHS: data

The relationship between LHS and RHS are such that the type extractor gets dynamically created for a given data instance. 

This features makes quick point queries much simpler. When the input only consists of data, we can assume that the user wants to say "give me everything that equals this". That's what type extractors are for. Without being more specific, we can also assume that equality comparison is the desired operation.


mavam:
Jenkins failed twice on FreeBSD with this error message:

```
[ 98%] Linking CXX executable ../bin/vast-test
CMakeFiles/vast-test.dir/test/expression.cpp.o: file not recognized: File format not recognized
c++: error: linker command failed with exit code 1 (use -v to see invocation)
*** Error code 1
```

mavam:
I've rebased the branch onto current master because the changelog has moved forward in the meantime. Nothing has changed otherwise.

tobim:
I'll hold off on merging until the Jenkins issue is sorted.

PR #523, opened 2019-06-29
Fix false positive results from the ARCHIVE

Neverlord:


libvast/src/system/archive.cpp
```diff
@@ -86,7 +86,9 @@ archive(archive_type::stateful_pointer<archive_state> self, path dir,
               VAST_DEBUG(self, "dismisses query for inactive sender");
               return make_error(ec::no_error);
             }
-            std::vector<event> result;
+            using receiver_type = caf::typed_actor<
+              caf::reacts_to<table_slice_ptr>>;
+            auto dst = caf::actor_cast<receiver_type>(self->current_sender());
```
mavam: I find `dst` for "destination" a bit generic. How about calling this variable `client` or `requester` for added semantics?

libvast/src/system/archive.cpp
```diff
@@ -96,10 +98,10 @@ archive(archive_type::stateful_pointer<archive_state> self, path dir,
                 // ... or an error occured.
                 return {done_atom::value, std::move(slice.error())};
               }
-              using receiver_type = caf::typed_actor<
-                caf::reacts_to<table_slice_ptr>>;
-              self->send(caf::actor_cast<receiver_type>(self->current_sender()),
-                         *slice);
+              // The slice may contain entries that are not selected by xs.
+              auto sub_slices = select(*slice, xs);
+              for (auto& sub_slice : sub_slices)
```
mavam: Could this be condensed into a single line?
```cpp
for (auto& subset : select(*slice, xs))
```

mavam:
Also, a bugfix entry in the changelog would be helpful.

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🐞 Fixed false positive results from the `export` command that were caused
+  by the ARCHIVE component returning data unrelated to the issued query.
+
```
mavam: Thanks for adding the entry. However, it doesn't fit the style in which we write the changelog. If you look at the other bug entries, we always describe the bug first, with a complete sentence. Then we describe the fix. For the user, the issue was duplicate results. Since we're writing this for the user, I think this aspect should be the center of the message. Here's a suggestion:
```
- 🐞 Some queries returned duplicate events because the archive did not filter
  the result set properly. This no longer occurs after fixing the table slice
  filtering logic.
```

PR #524, opened 2019-07-01
Port ASCII, CSV and JSON writers to the table slice API

Neverlord:
Also addresses performance issues of the previous implementation.


mavam:
Merging #530 caused a conflict here. Should be easy to fix, though.

libvast/vast/format/ostream_writer.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+  ~ostream_writer() override;
+
+  // -- overrides --------------------------------------------------------------
+
+  caf::expected<void> flush() override;
+
+protected:
+  // Buffer for building lines before syncing writing to out_.
+  std::vector<char> buf_;
```
tobim: Can you add a comment at the beginning why this is necessary and what the difference to `sync_with_stdio(false)` is?

Neverlord: Good point, sorry for not mentioning `sync_with_stdio` in the discussion. I've also checked `master` with `sync_with_stdio(false)` but it makes very little difference in performance. I'll add the comment.

Neverlord:
✅ Resolve conflicts
✅ Clarify design choice with comment (mentioning `sync_with_stdio(false)`)

For full transparency, here's the diff on `master` I've used to check the performance implications of `sync_with_stdio(false)` if you want to see for yourself:

```diff
diff --git a/libvast/vast/format/printer_writer.hpp b/libvast/vast/format/printer_writer.hpp
index 30d35209..757a98dd 100644
--- a/libvast/vast/format/printer_writer.hpp
+++ b/libvast/vast/format/printer_writer.hpp
@@ -35,6 +35,7 @@ public:
  /// @param out The stream where to write to
  explicit printer_writer(std::unique_ptr<std::ostream> out)
    : out_{std::move(out)} {
+      out_->sync_with_stdio(false);
  }

  using format::writer::write;
diff --git a/vast/vast.cpp b/vast/vast.cpp
index d94cb7b6..8b7cac21 100644
--- a/vast/vast.cpp
+++ b/vast/vast.cpp
@@ -38,6 +38,8 @@ using namespace vast;
using namespace vast::system;

int main(int argc, char** argv) {
+  std::cout.sync_with_stdio(false);
+  std::cerr.sync_with_stdio(false);
  // CAF scaffold.
  default_configuration cfg{"vast"};
  if (auto err = cfg.parse(argc, argv)) {
```

I think the change in `printer_writer` is sufficient, but I prefer to err on the safe side. 😉

libvast/vast/format/ostream_writer.hpp
```diff
@@ -89,10 +102,12 @@ class ostream_writer : public writer {
   /// Writes the content of `buf_` to `out_` and clears `buf_` afterwards.
   void write_buf();
 
-  // Buffer for building lines before syncing writing to out_.
+  /// Buffer for building lines before writing to `out_`. Printing into this
+  /// buffer with a `back_inserter` first gives a 4x speedup over printing
```
mavam: What do you mean by "first" here?

libvast/vast/format/ostream_writer.hpp
```diff
@@ -55,6 +59,33 @@ class ostream_writer : public writer {
     buf_.emplace_back(x);
   }
 
+  template <bool IncludeFieldNames, class Printer>
```
mavam: `IncludeFieldNames` appears to be a policy. I would move it into `vast::policy`. This also helps understanding what's going on at the call site: `print<true>` vs. `print<policy::include_field_names>`

PR #525, opened 2019-07-01
Implement generic Community ID computation API

mavam:
This PR adds a the [Community ID](https://github.com/corelight/community-id-spec) to PCAP packet data.

### DIFF
--- a/libvast/test/hash.cpp
+++ b/libvast/test/hash.cpp
@@ -11,7 +11,10 @@
  * contained in the LICENSE file.                                             *
  ******************************************************************************/
 
+#include <iomanip>
+
 #include "vast/concept/hashable/crc.hpp"
+#include "vast/concept/hashable/sha1.hpp"
 #include "vast/concept/hashable/uhash.hpp"
 #include "vast/concept/hashable/xxhash.hpp"
 
@@ -32,7 +35,17 @@ auto inspect(Inspector& f, foo& x) {
   return f(x.a, x.b);
 }
 
-} // namespace <anonymous>
+template <class T, size_t N>
+std::string hexify(const std::array<T, N>& xs) {
+  std::stringstream ss;
+  ss << std::setfill('0') << std::hex;
+  auto ptr = reinterpret_cast<const uint8_t*>(xs.data());
+  for (auto i = 0u; i < N * sizeof(T); ++i)
+    ss << std::setw(2) << static_cast<int>(ptr[i]);
+  return ss.str();
+}
+
+} // namespace
 
 TEST(hashing an inspectable type) {
   using hasher = xxhash32;
@@ -96,4 +109,19 @@ TEST(xxhash zero bytes) {
   xxh32(nullptr, 0);
   xxhash64 xxh64;
   xxh64(nullptr, 0);
+}
+
+TEST(sha1) {
+  // one-shot
+  std::array<char, 2> fortytwo = {'4', '2'};
+  auto digest = uhash<sha1>{}(fortytwo);
+  CHECK_EQUAL(hexify(digest), "92cfceb39d57d914ed8b14d0e37643de0797ae56");
+  // incremental
+  sha1 sha;
+  sha("foo", 3);
+  sha("bar", 3);
+  sha("baz", 3);
+  sha("42", 2);
+  digest = static_cast<sha1::result_type>(sha);
+  CHECK_EQUAL(hexify(digest), "4cbfb91f23be76f0836c3007c1b3c8d8c2eacdd1");
 }
--- a/libvast/test/hash.cpp
+++ b/libvast/test/hash.cpp
@@ -17,11 +17,13 @@
 #include "vast/concept/hashable/sha1.hpp"
 #include "vast/concept/hashable/uhash.hpp"
 #include "vast/concept/hashable/xxhash.hpp"
+#include "vast/detail/coding.hpp"
 
 #define SUITE hash
 #include "vast/test/test.hpp"
 
 using namespace vast;
+using vast::detail::hexify;
 
 namespace {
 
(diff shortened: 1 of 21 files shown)


libvast/vast/community_id.hpp
```diff
@@ -0,0 +1,207 @@
...
+  result += version;
+  result += ':';
+  // Compute a SHA-1 hash over the flow tuple.
+  sha1 hasher;
+  hash_append(hasher, detail::to_network_order(seed));
+  hash_append(hasher, x);
+  auto digest = static_cast<sha1::result_type>(hasher);
+  // Convert the binary digest to plain hex ASCII or to Base64.
+  constexpr auto element_size = sizeof(sha1::result_type::value_type);
+  constexpr auto num_bytes = element_size * digest.size();
```
Neverlord: I'm surprised this compiles. `digest` is not a `constexpr` value, so logically `num_bytes` can't be either.

```suggestion
  auto num_bytes = element_size * digest.size();
```

libvast/vast/community_id.hpp
```diff
@@ -0,0 +1,207 @@
...
+  auto ptr = reinterpret_cast<const uint8_t*>(digest.data());
+  if constexpr (std::is_same_v<Policy, policy::base64>) {
+    result.resize(max_length<Policy>());
+    auto offset = version_prefix_length();
+    auto n = detail::base64::encode(result.data() + offset, ptr, num_bytes);
+    result.resize(offset + n);
+  } else {
```
Neverlord: Shouldn't this block assert on `static_assert(std::is_same_v<Policy, policy::ascii>, "unsupported policy");`? It seems like this assumes the ASCII policy.

libvast/vast/detail/coding.hpp
```diff
@@ -16,42 +16,57 @@
...
-  static constexpr char hex[] = "0123456789ABCDEF";
-  return {hex[(b >> 4) & 0x0f], hex[b & 0x0f]};
+template <class Policy = policy::uppercase, class T>
+constexpr std::pair<char, char> byte_to_hex(T x) {
+  if constexpr (std::is_same_v<Policy, policy::uppercase>)
+    return byte_to_hex(x, "0123456789ABCDEF");
+  else if constexpr (std::is_same_v<Policy, policy::lowercase>)
+    return byte_to_hex(x, "0123456789abcdef");
+  else
+    return {0, 0};
```
Neverlord: I think it would improve robustness to use `static_assert(std::is_same_v<Policy, policy::lowercase>);` in the `else` block rather than returning an arbitrary default.

libvast/vast/icmp.hpp
```diff
@@ -0,0 +1,120 @@
...
+  wru_reply = 140,
+  haad_request = 144,
+  haad_reply = 145,
+};
+
+/// Computes the dual to a given ICMP type.
+/// @param x The ICMP type.
+/// @returns For a request type, the response type - and vice versa.
+/// @relates icmp_type
+inline caf::optional<icmp_type> dual(icmp_type x) {
```
Neverlord: This is not a trivial one-liner, so I think it should go into a `.cpp` file.

libvast/vast/icmp.hpp
```diff
@@ -0,0 +1,120 @@
...
+    case icmp_type::mask_reply:
+      return icmp_type::mask;
+  }
+}
+
+/// Computes the dual to a given ICMP6 type.
+/// @param x The ICMP6 type.
+/// @returns For a request type, the response type - and vice versa.
+/// @relates icmp6_type
+inline caf::optional<icmp6_type> dual(icmp6_type x) {
```
Neverlord: Likewise, a nontrivial function like this should move to a `.cpp` file IMO.

libvast/vast/icmp.hpp
```diff
@@ -0,0 +1,120 @@
...
+ * http://vast.io/license. No part of VAST, including this file, may be       *
+ * copied, modified, propagated, or distributed except according to the terms *
+ * contained in the LICENSE file.                                             *
+ ******************************************************************************/
+
+#pragma once
+
+#include <cstdint>
+
+#include <caf/optional.hpp>
```
Neverlord: We can instead include `caf/fwd.hpp` here when moving the function implementations to a `.cpp` file.

libvast/test/hash.cpp
```diff
@@ -32,7 +35,17 @@ auto inspect(Inspector& f, foo& x) {
   return f(x.a, x.b);
 }
 
-} // namespace <anonymous>
+template <class T, size_t N>
+std::string hexify(const std::array<T, N>& xs) {
+  std::stringstream ss;
```
Neverlord: A stream seems a bit heavyweight for this task and the code. Did you consider using `byte_to_hex` instead?

```c++
  std::string result;
  for (auto x : xs) {
    auto [first, second] = byte_to_hex<policy::lowercase>(x);
    result += first;
    result += second;
  }
  return result;
```

I think a version like this is also easier to follow.

libvast/vast/community_id.hpp
mavam: Actually, [`array<T, N>::size`](https://en.cppreference.com/w/cpp/container/array/size) *is* `constexpr`, so this works. This makes also sense intuitively, because the size is known at compile-time.

libvast/test/hash.cpp
mavam: I've factored `hexify` into a utility function. Now it's easier to reference.

libvast/vast/format/mrt.hpp
```diff
@@ -286,6 +287,7 @@ struct attributes_parser : parser<attributes_parser<TypeCode14Policy>> {
       // can use the combinator style to chain the parser together, as opposed
       // to dispatching manually.
       auto t = f;
+      // clang-format off
```
Neverlord: `clang-format` was disabled on line 268. Did you mean `// clang-format on`?

mavam: Whoops, that comment was a bit overzealous. I've removed it.

PR #526, opened 2019-07-02
Rename #time to #timestamp

Neverlord:
To stay consistent with the type names in schema files, I've also renamed the type `time` to `timestamp`. For example, our Zeek schema now reads:

```
type zeek.conn = record{
  ts: timestamp #timestamp,
  uid: string,
  ...
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🔄 Log files generally have some notion of timestamp for recorded events. To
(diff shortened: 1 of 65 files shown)


mavam:
I appreciate your effort to go one step further. However, in this case I don't think that the type should have been renamed as well. `#timestamp` as attribute makes sense, but the type `time` should not be the same as the attribute, that was also the reason for renaming it in the first place. We had a discussion to rename `time` to `date`, but that hasn't been agreed upon yet. Would you mind only going with the attribute naming for now?

mavam:
Looks like the changelog got out of sync with this PR.

PR #528, opened 2019-07-03
Add CSV import format

tobim:
Todo:
- [x] Add additional `reader_command`
- [x] Implement stubs
- [x] Fix remaining `TODO`s
- [x] Add unit test
- [x] Add Integration tests
- [x] Changelog entry


tobim:
Despite the incomplete task list: This PR can be reviewed by now.

libvast/test/format/csv.cpp
```diff
@@ -0,0 +1,93 @@
...
+2011-08-12T13:14:36.012344Z,147.32.84.165,1041
+,147.32.84.165,1046
+,147.32.84.165,
+,,)__";
+
+} // namespace
+
+FIXTURE_SCOPE(csv_reader_tests, fixtures::deterministic_actor_system)
+
+TEST(csv reader) {
```
Neverlord: I would like to see the boilerplate in a fixture and have more tests (one per tested log).

libvast/vast/concept/parseable/core/rule.hpp
```diff
@@ -50,6 +95,55 @@ class rule_definition : public abstract_rule<Iterator, Attribute> {
 
 } // namespace detail
 
+/// A type-erased parser which can store any other parser. This type exhibits
+/// value semantics and can therefore not be used to construct recursive
+/// parsers.
+template <class Iterator>
+class erased_parser : public parser<erased_parser<Iterator>> {
```
Neverlord: Why did you omit `typed_` in the name? Calling this class `erased_parser` left me wondering "erased... what?" while reading the code and I had to double check here that it's actually just type erasure.

Also, I miss a motivation to this class. It seems like this is functionally equivalent to `rule<>`, so what is the reason to have a new class in the first place? Only to avoid one indirection? If that's the case I would first want to see performance metrics that show a bottleneck with `rule<>`.

libvast/vast/concept/parseable/detail/as_parser.hpp
```diff
@@ -25,7 +25,7 @@ namespace detail {
 
 // -- unary -------------------------------------------------------------------
 
-inline auto as_parser(char c) {
+constexpr inline auto as_parser(char c) {
```
Neverlord: Any `constexpr` function is also `inline`.

```suggestion
constexpr auto as_parser(char c) {
```

libvast/vast/concept/parseable/core/rule.hpp
```diff
@@ -71,25 +165,24 @@ class rule : public parser<rule<Iterator, Attribute>> {
...
+    -> std::enable_if_t<is_parser_v<std::decay_t<
+                          RHS>> && !detail::is_same_or_derived_v<rule, RHS>> {
     make_parser<RHS>(std::forward<RHS>(rhs));
   }
 
-  bool parse(Iterator& f, const Iterator& l, unused_type) const {
+  // This overload must not be available if Attribute == unused_type.
+  template <class Instantiate_on_use = bool>
+  std::enable_if_t<!std::is_same_v<Attribute, unused_type>, Instantiate_on_use>
+  parse(Iterator& f, const Iterator& l, unused_type) const {
```
Neverlord: Do we need the overloads in the first place? All of our parser do a dispatching on `unused` anyways:

```c++
template <class T>
bool parse(Iterator& f, const Iterator& l, T&& x) {
  VAST_ASSERT(*parser_ != nullptr);
  return (*parser_)->parse(f, l, std::forward<T>(x));
}
```

Neverlord:
We get `CMakeFiles/vast-test.dir/test/expression.cpp.o: file not recognized: File format not recognized` on FreeBSD again. As a quickfix, I think we should disable this test suite (by commenting out the entry for `test/expression.cpp` in CMake) for now and address this with a separate story.

libvast/src/format/csv.cpp
```diff
@@ -21,6 +32,205 @@ constexpr char value_printer::separator[];
...
+  using result_type = erased_parser<Iterator>;
+
+  csv_parser_factory(const std::string& set_separator,
+                     table_slice_builder_ptr bptr)
+    : set_separator_{set_separator}, bptr_{std::move(bptr)} {
+    // nop
+  }
+
+  template <class T>
+  struct add_t {
```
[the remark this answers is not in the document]
tobim: I tried to do that, but a templated `operator()` broke the callable_traits deduction. So I chose to workaround that by templating the type instead.

libvast/src/format/csv.cpp
```diff
@@ -21,6 +32,205 @@ constexpr char value_printer::separator[];
...
+    if constexpr (std::is_same_v<T, string_type>) {
+      return +(parsers::any - set_separator_)->*[](std::string x) {
+        return data{std::move(x)};
+      };
+    } else if constexpr (std::is_same_v<T, pattern_type>) {
+      return +(parsers::any - set_separator_)->*[](std::string x) {
+        return data{std::move(x)};
+      };
+    } else if constexpr (std::is_same_v<T, set_type>) {
+      auto set_insert = [](std::vector<Attribute> v) {
```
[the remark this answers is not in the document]
tobim: The action parser does indeed move the arguments to the action parser, and in the implementation of `set_insert` the elements are moved into the set, so const ref would not work.

libvast/src/format/csv.cpp
```diff
@@ -21,6 +32,205 @@ constexpr char value_printer::separator[];
...
+               ->*[](const std::vector<Attribute> v) { return v; };
+    } else if constexpr (has_parser_v<type_to_data<T>>) {
+      using value_type = type_to_data<T>;
+      return make_parser<value_type>{}->*[](value_type x) { return x; };
+    } else {
+      return {};
+    }
+  }
+
+  const std::string& set_separator_;
```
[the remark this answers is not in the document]
tobim: `string_view` would require special treatment in `as_parser` to create a `string_parser`, but that is not in place.

libvast/vast/concept/parseable/core/rule.hpp
tobim: > Also, I miss a motivation to this class. It seems like this is functionally equivalent to rule<>, so what is the reason to have a new class in the first place? Only to avoid one indirection? If that's the case I would first want to see performance metrics that show a bottleneck with rule<>.

This class is actually complementary to `rule<>`. `rule<>` parsers can be combined with other parsers through the `choice_parser` (`|`), but if you try to sequence it with `>>`, you actually end up with a parser that recurses infinitely. That is why I created the `erased_parser`, which cannot represent recursive grammars. It only provides type erasure, nothing else.

tobim: > Why did you omit typed_ in the name? Calling this class erased_parser left me wondering "erased... what?" while reading the code and I had to double check here that it's actually just type erasure.

I'll rename it to `type_erased_parser`.

libvast/src/format/csv.cpp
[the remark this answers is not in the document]
tobim: I changed the argument type for this and the corresponding vector and map insert functions to `std::vector<Attribute>&&`.

libvast/test/format/csv.cpp
tobim: I did as you suggested, It still doesn't check every corner case, but the coverage should be decent.

libvast/vast/concept/parseable/core/rule.hpp
tobim: I followed your suggestion and removed the extra overload for `unused_type`.

libvast/src/format/csv.cpp
```diff
@@ -129,4 +140,271 @@ const char* writer::name() const {
...
+    : set_separator_{set_separator}, bptr_{std::move(bptr)} {
+    // nop
+  }
+
+  struct identity {
+    template <class T>
+    constexpr T operator()(T&& t) const noexcept {
+      return std::forward<T>(t);
+    }
+  };
```
mavam: Per our discussion, I think we can get rid of `identity` if we push the functionality in action parsers.

libvast/vast/defaults.hpp
```diff
@@ -71,6 +71,15 @@ struct bgpdump {
...
 };
 
+/// Contains settings for the csv subcommand.
+struct csv {
+  /// Nested category in config files for this subcommand.
+  static constexpr const char* category = "import.csv";
+
+  /// Path for reading input events.
+  static constexpr auto read = shared::read;
+};
```
mavam: The separator and set-separator could go in here as well.

libvast/src/format/csv.cpp
```diff
@@ -129,4 +140,271 @@ const char* writer::name() const {
...
+
+  template <class T, class F>
+  static add_t<T, F> make_add_t(table_slice_builder_ptr bptr, F f) {
+    return add_t<T, F>{std::move(bptr), std::move(f)};
+  }
+
+  template <class T>
+  result_type operator()(const T& t) const {
+    if constexpr (std::is_same_v<T, string_type>) {
+      return (-+(parsers::any - set_separator_))
```
mavam: This should probably be the standard separator, not the *set*-separator.

libvast/src/format/csv.cpp
```diff
@@ -129,4 +140,271 @@ const char* writer::name() const {
...
+        return data{pattern{std::move(x)}};
+      };
+    } else if constexpr (std::is_same_v<T, set_type>) {
+      auto set_insert = [](std::vector<Attribute>&& xs) {
+        return set(std::make_move_iterator(xs.begin()),
+                   std::make_move_iterator(xs.end()));
+      };
+      return ('{'_p >> (caf::visit(*this, t.value_type) % set_separator_)
+              >> '}'_p)
+               ->*set_insert;
```
mavam: The line breaking makes this hard to read. Mind refactoring it here (and also in similar occurrences below)?

libvast/src/format/csv.cpp
```diff
@@ -129,4 +140,271 @@ const char* writer::name() const {
...
+}
+
+caf::error reader::schema(vast::schema s) {
+  schema_ = std::move(s);
+  for (auto& t : s) {
+    auto r = caf::get_if<record_type>(&t);
+    if (!r)
+      continue;
+    schema_.add(flatten(*r));
+  }
```
mavam: KISS:
```cpp
for (auto& t : s)
  if (auto r = caf::get_if<record_type>(&t))
    schema_.add(flatten(*r));
```

libvast/src/format/csv.cpp
```diff
@@ -129,4 +140,271 @@ const char* writer::name() const {
...
+    } // else skip
+  }
+  return caf::none;
+}
+
+namespace {
+
+template <class Iterator, class Attribute>
+struct container_parser_builder {
+  using result_type = rule<Iterator, Attribute>;
```
mavam: IIRC you wanted to add a comment here that switching to `type_erased_parser` could be a performance optimization.

libvast/src/format/csv.cpp
```diff
@@ -129,4 +140,271 @@ const char* writer::name() const {
...
+
+  explicit container_parser_builder(const std::string& set_separator)
```
mavam:
```suggestion
  explicit container_parser_builder(std::string_view set_separator)
```
Since you're copying the string into the function object member, and from there again into the parser. Alternatively, use a `const std::string&` as type for `set_separator_`.

libvast/src/format/csv.cpp
```diff
@@ -129,4 +140,271 @@ const char* writer::name() const {
...
+               ->*set_insert;
+    } else if constexpr (std::is_same_v<T, vector_type>) {
+      auto vector_insert = [](std::vector<Attribute>&& xs) { return xs; };
+      return ('[' >> (caf::visit(*this, t.value_type) % set_separator_) >> ']')
+               ->*vector_insert;
+    } else if constexpr (std::is_same_v<T, map_type>) {
+      auto map_insert = [](std::vector<std::tuple<Attribute, Attribute>>&& xs) {
+        auto to_pair = [](auto&& tuple) {
+          return std::make_pair(std::get<0>(tuple), std::get<1>(tuple));
+        };
```
mavam: You don't need `to_pair`. The sequence parser can handle `std::pair`.

libvast/src/format/csv.cpp
```diff
@@ -129,4 +140,271 @@ const char* writer::name() const {
...
+        for (auto& x : xs)
+          m.insert(to_pair(std::move(x)));
+        return m;
+      };
+      auto kvp = caf::visit(*this, t.key_type) >> '='
+                 >> caf::visit(*this, t.value_type);
+      return (('{'_p >> (kvp % set_separator_) >> '}')->*map_insert);
+    } else if constexpr (has_parser_v<type_to_data<T>>) {
+      using value_type = type_to_data<T>;
+      auto ws = ignore(*parsers::space);
```
mavam: So are we skipping whitespace already? Probably that's the location that we want to have configurable with the `csv_parser_options` struct.

libvast/src/format/csv.cpp
```diff
@@ -129,4 +140,271 @@ const char* writer::name() const {
...
+          VAST_WARNING_ANON("csv reader failed to parse unexpected enum value",
+                            s);
+          return caf::none;
+        }
+        return std::distance(t.fields.begin(), i);
+      };
+      return (-+(parsers::any - set_separator_))
+               ->*make_add_t<std::string>(bptr_, to_enumeration);
+    } else if constexpr (detail::contains_v<
+                           T, std::tuple<set_type, vector_type, map_type>>) {
```
mavam: This is simpler: `detail::is_any_v<T, set_type, vector_type, map_type>`.

libvast/src/format/csv.cpp
```diff
@@ -129,4 +140,271 @@ const char* writer::name() const {
...
+  VAST_ASSERT(num_fields > 0);
+  auto factory = csv_parser_factory<Iterator>{",", builder};
+  auto result = caf::visit(factory, layout.fields[0].type);
+  for (size_t i = 1; i < num_fields; ++i) {
+    auto p = (caf::visit(factory, layout.fields[i].type));
+    result = (result >> ','
+              >> std::move(p)) /*->* [](std::tuple<bool, bool>&& bb) {
+return std::get<0>(bb) && std::get<1>(bb);
+}*/;
+  }
```
mavam: This looks hard to follow and needs some cleanup.

libvast/vast/format/csv.hpp
```diff
@@ -31,4 +43,47 @@ class writer : public format::ostream_writer {
...
+
+  caf::error read_header(std::string_view line);
+
+  std::unique_ptr<std::istream> input_;
+  std::unique_ptr<detail::line_range> lines_;
+  vast::schema schema_;
+  struct rec_table {
+    record_type type;
+    std::vector<std::string> sorted;
+  };
```
mavam: We usually don't inline types like this. You can put it to the type aliases under `private:`.

libvast/vast/view.hpp
```diff
@@ -445,6 +445,14 @@ data_view make_data_view(const T& x) {
   return make_view(x);
 }
 
+/// @relates view_trait
+template <class T>
+data_view make_data_view(const optional<T>& x) {
+  if (!x)
+    return make_view(caf::none);
+  return make_view(x);
```
mavam: Or simpler: `x ? make_data_view(x) : make_data_view(caf::none)`

mavam: But, what I want to ask you: don't you want `make_data_view(*x)`? Note the deref!

libvast/src/format/csv.cpp
tobim: The string-like types are not handled in this path.

libvast/src/format/csv.cpp
tobim: This parameter was replaced by an option struct.

libvast/src/format/csv.cpp
tobim: Good catch.

libvast/src/format/csv.cpp
```diff
@@ -129,4 +139,265 @@ const char* writer::name() const {
...
+  if (!reset_builder(*layout))
+    return make_error(ec::parse_error, "unable to create a builder for layout");
+  parser_ = make_csv_parser<iterator_type>(*layout, builder_, opt_);
+  if (!parser_)
+    return make_error(ec::parse_error, "unable generate a parser");
+  return caf::none;
+}
+
+caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
+                             consumer& cons) {
```
mavam: Maybe `s/cons/callback`? When someone reads "cons", they usually have a different association. 😉

libvast/vast/concept/parseable/vast/schema.hpp
```diff
@@ -25,6 +25,13 @@ namespace vast {
 struct schema_parser : parser<schema_parser> {
   using attribute = schema;
 
+  // clang-format off
+  static constexpr auto id = +(parsers::alnum
+                           | parsers::chr{'_'}
+                           | parsers::chr{'-'}
+                           | parsers::chr{'.'});
+  // clang-format on
```
mavam: For DSL prettiness, you could write this with `|` aligned under `(`:
```cpp
static constexpr auto id = +( parsers::alnum
	                    | parsers::chr{'_'}
	                    | parsers::chr{'-'}
	                    | parsers::chr{'.'}
                            );
```

mavam:
The last unchecked TODO: a couple of integration tests. This should be hopefully trivial at this point.

libvast/src/format/csv.cpp
```diff
@@ -129,4 +139,265 @@ const char* writer::name() const {
...
+      // clang-format on
+    } else if constexpr (std::is_same_v<T, enumeration_type>) {
+      auto to_enumeration = [t](std::string s) -> caf::optional<Attribute> {
+        auto i = std::find(t.fields.begin(), t.fields.end(), s);
+        if (i == t.fields.end()) {
+          VAST_WARNING_ANON("csv reader failed to parse unexpected enum value",
+                            s);
+          return caf::none;
+        }
+        return enumeration(std::distance(t.fields.begin(), i));
```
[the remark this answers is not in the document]
tobim: As far as I understand, the enums are currently not handled at the query side. The binding can only be done "late" anyway, because the layout is needed to map it to the internal representation.

Extensibility seems to be orthogonal to me. One way that I can come up with now is to maintain an internal enum defnition that sorts the possible values by time of declaration. So if an enum is read in a layout, instead of just putting it into the record verbatim, we look it up by its fqn in an enums table and update the entry there by appending any new values from the parsed definition.

libvast/src/format/csv.cpp
```diff
@@ -21,6 +32,205 @@ constexpr char value_printer::separator[];
...
+          else
+            return caf::none;
+        }
+        return record_type{result_raw};
+      }();
+      if (result)
+        return result->name(r->name());
+    } else if (names.size() == 1 && names[0] == t.name()) {
+      // Hoist naked type into record.
+      // TODO: Maybe this is actually not a good idea?
```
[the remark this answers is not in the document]
tobim: I removed the controversial comment. It would be unnecessary to require layouts with a single field to wrapped into a record.

tobim:
> One other thing I've noticed, the pattern `unbox(to<T>(...))` occurs incredibly frequently. It looks like we should come up with a simplification here. How about `coerce<T>` or `force<T>`? I'm open to other suggestions as well. Since this is technically not related to this PR, we can also address this in a follow-up story.

Sounds reasonable.

libvast/src/format/csv.cpp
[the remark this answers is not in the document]
mavam: > [..] because the layout is needed to map it to the internal representation.

Okay, the layout is usually around when we're dealing with enum values, since they are inside a table slice. That means for a given enum type, we can create a specific value index. But since the numeric value is an implementation detail, and users would actually query by enum string value, it won't be a problem.

mavam:
[ch6670]

PR #532, opened 2019-07-05
Implement rule_ref to allow breaking parser cycles

Neverlord:
The new syntax `ref(x)` allows users to create a non-owning reference to a `rule<Iterator, Attribute>` in order to break cycles.

This means we have to update parser definitions and break cycles manually, but I don't think there's a reasonable strategy to break cycles otherwise (we certainly don't want to implement half of a tracing garbage collector).

Once we settled on an approach / API we can update existing parsers (hence Draft PR).


libvast/vast/concept/parseable/vast/type.hpp
```diff
@@ -118,15 +119,17 @@ struct type_parser : parser<type_parser> {
       return vector_type{std::move(value_type)}.attributes(std::move(attrs));
     };
     auto vector_type_parser
-      = ("vector" >> ws >> '<' >> ws >> type_type >> ws >> '>') ->* to_vector
+      = ("vector" >> ws >> '<' >> ws >> vast::ref(type_type) >> ws >> '>')
```
mavam: Could `using vast::ref` help here to bring `ref` in scope and make the grammar definition as terse as possible?

libvast/test/format/writer.cpp
```diff
@@ -45,8 +45,8 @@ auto first_zeek_conn_log_line = R"__({"ts": 1258531221.486539, "uid": "Pii6cUUq1
 template <class Writer>
 std::vector<std::string> generate(const std::vector<event>& xs) {
   std::string str;
-  auto sb = new caf::containerbuf<std::string>{str};
```
mavam: Wow, a naked `new`...and that in 2019!

libvast/vast/concept/parseable/vast/json.hpp
```diff
@@ -36,20 +38,15 @@ struct json_parser : parser<json_parser> {
     auto lbrace = ws >> '{' >> ws;
     auto rbrace = ws >> '}' >> ws;
     auto delim = ws >> ',' >> ws;
-    auto null = ws >> "null"_p ->* [] { return json::null{}; };
+    auto null = ws >> "null"_p->*[] { return json::null{}; };
```
mavam: All parser actions use whitespace around `->*`, as the DSL considers this operator as a binary operation. Please keep this consistent for now. We can discuss global refactoring as part of a separate PR, if you want to.

mavam:
Also, mind adding a changelog entry that describes the bug plus the fix?

Neverlord:
> Also, mind adding a changelog entry that describes the bug plus the fix?

My understanding is that we keep the changelog for user-facing changes and bugs. This is neither.

libvast/vast/concept/parseable/vast/json.hpp
Neverlord: > Please keep this consistent for now.

Putting whitespaces around `->*` will trigger `clang-format`. Is this really that important to you that you want me to put `// clang-format on/off` around this one line in this function?

libvast/test/format/writer.cpp
Neverlord: Technically, a naked `new` from 2016: https://github.com/vast-io/vast/commit/1b479ee1220ae432523d6a03e86483434515712c

libvast/vast/concept/parseable/vast/type.hpp
Neverlord: Please note that we are already in namespace `vast`. I'm not quite sure why `std::ref` becomes a candidate function in the first place, but without qualifier is gets precedence.

Neverlord: I've added a second overload so that there's always a more specialized `vast::ref` than there's a `std::ref`.

libvast/vast/concept/parseable/vast/json.hpp
mavam: I think readability of the code should always trump the desire to use `clang-format`. To me, `"null"_p->*[] { ..` looks incredibly hard to understand. For all other parsers that we wrote, we have the whitespace style. Why make an exception here?

mavam:
> My understanding is that we keep the changelog for user-facing changes and bugs. This is neither.

To me, a memory leak is a bug that impacts the user experience: if a process runs out of memory because of a program bug, it's certainly important to know. Hence we should mention it in the changelog.

I'm trying to understand where you draw the line. I don't think the amount of memory leaked is relevant here. The fact that we fixed a leak is noteworthy in my opinion. I'd be happy to hear also your opinion, @tobim.

tobim:
> > My understanding is that we keep the changelog for user-facing changes and bugs. This is neither.
> 
> To me, a memory leak is a bug that impacts the user experience: if a process runs out of memory because of a program bug, it's certainly important to know. Hence we should mention it in the changelog.
> 
> I'm trying to understand where you draw the line. I don't think the amount of memory leaked is relevant here. The fact that we fixed a leak is noteworthy in my opinion. I'd be happy to hear also your opinion, @tobim.

I would personally include all changes that have an observable effect during the run time of the program. In this case that would mean that an entry should be added.

PR #533, opened 2019-07-05
 Add root command option to set verbosity

Neverlord:
- Consolidate various initialization steps that we needed in both VAST and Tenzir to streamline the startup sequence and make commands more easily extendible
- Improve user experience by allowing users to change the verbosity via `--verbosity=<level>` instead of `--ca#logger.console-verbosity=<level>`

### DIFF
--- a/libvast/src/command.cpp
+++ b/libvast/src/command.cpp
@@ -18,6 +18,7 @@
 #include <caf/actor_system.hpp>
 #include <caf/actor_system_config.hpp>
 #include <caf/defaults.hpp>
+#include <caf/detail/log_level.hpp>
 #include <caf/make_message.hpp>
 #include <caf/message.hpp>
 #include <caf/settings.hpp>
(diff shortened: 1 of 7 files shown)


libvast/src/command.cpp
```diff
@@ -131,6 +131,10 @@ bool init_config(caf::actor_system_config& cfg, const command::invocation& from,
   };
   // Merge all CLI settings into the actor_system settings.
   merge_settings(from.options, cfg.content);
+  // Allow users to use `system.verbosity` to configure console verbosity.
+  if (auto value = caf::get_if<caf::atom_value>(&from.options,
+                                                "system.verbosity"))
+    cfg.set("logger.console-verbosity", *value);
```
mavam: One additional thought: how about issuing a warning when the requested log level is higher than the compile-time maximum? A silent fallback might be an opportunity for confusion.

PR #536, opened 2019-07-11
Port PCAP and Zeek writer to the table slice API

Neverlord:


libvast/src/format/zeek.cpp
```diff
@@ -554,64 +456,159 @@ writer::writer(path dir) {
...
+    return (*this)(out, count{p.number()});
+  }
+
+  bool print(Iterator& out, const attribute& d) const {
+    auto f = [&](const auto& x) { return (*this)(out, x); };
+    return caf::visit(f, d);
+  }
+
+private:
+  real_printer<real, 6, 6> real_printer_;
```
mavam: This suggests that we're dealing with a stateful parser. I'd say either create a type alias at class scope or even a static variable:

```cpp
static constexpr inline auto zeek_real = real_printer<real, 6, 6>{}:
```

libvast/src/format/zeek.cpp
```diff
@@ -554,64 +456,159 @@ writer::writer(path dir) {
...
+      if (!print(out, *i))
+        return false;
+      for (++i; i != x.end(); ++i) {
+        *out++ = set_separator;
+        if (!print(out, *i))
+          return false;
+      }
+      return true;
+    } else if constexpr (std::is_same_v<T, view<map>>) {
+      VAST_ERROR(this, "cannot print maps in Zeek format");
```
mavam: This seems to suggest that VAST is the problem. But it's Zeek TSV format that cannot handle maps. Therefore, I'd rephrase this as follows:

```
      VAST_ERROR(this, "Zeek TSV format does not support maps");
```

libvast/test/format/pcap.cpp
```diff
@@ -88,9 +90,9 @@ TEST(PCAP read/write 2) {
   auto file = "vast-unit-test-workshop-2011-browse.pcap";
   format::pcap::writer writer{file};
   auto deleter = caf::detail::make_scope_guard([&] { rm(file); });
-  for (auto& e : events)
-    if (!writer.write(e))
-      FAIL("failed to write event");
+  for (auto& slice : events)
+    REQUIRE_EQUAL(writer.write(slice), caf::none);
 }
+*/
```
mavam: (Looks like commenting this one was an oversight.)

PR #537, opened 2019-07-15
Fix precision of printed timestamps

Neverlord:


CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🐞 Previously, timestamps were always printed in millisecond resolution.
+  Hence, it was possible that a query produced no result after copying and
+  pasting a timestamp from a previous query result. Timestamps are now printed
+  in the resolution necessary, down to nanoseconds.
```
mavam: Here's an alternate wording suggestion:

```
Previously, timestamps were always printed in millisecond resolution.
Timestamps were always printed in millisecond resolution, which lead to loss of
precision when the internal representation had a higher resoultion. Timestamps
are now rendered up to nanosecond resolution - the maximum resolution
supported.
```

libvast/vast/concept/printable/std/chrono.hpp
```diff
@@ -145,20 +145,34 @@ struct time_point_printer : printer<time_point_printer<Clock, Duration>> {
...
+    if (!p(out, static_cast<int>(Y), static_cast<unsigned>(M),
+           static_cast<unsigned>(D), static_cast<int>(h.count()),
+           static_cast<int>(m.count()), static_cast<int>(s.count())))
+      return false;
+    // For the sub-second part, we print either '.0' if there's none, or print
+    // down to the lowest resolution necessary (all the way down to ns).
+    *out++ = '.';
+    auto num3 = printers::integral<int, policy::plain, 3>;
+    auto num6 = printers::integral<int, policy::plain, 6>;
+    auto num9 = printers::integral<int, policy::plain, 9>;
```
mavam: Can these be `constexpr`?

libvast/vast/concept/printable/std/chrono.hpp
```diff
@@ -145,20 +145,34 @@ struct time_point_printer : printer<time_point_printer<Clock, Duration>> {
...
-             static_cast<unsigned>(D), static_cast<int>(h.count()),
-             static_cast<int>(m.count()), static_cast<int>(s.count()),
-             static_cast<int>(sub_secs.count()));
+    auto sub_secs = duration_cast<nanoseconds>(t - h - m - s).count();
+    if (!p(out, static_cast<int>(Y), static_cast<unsigned>(M),
+           static_cast<unsigned>(D), static_cast<int>(h.count()),
+           static_cast<int>(m.count()), static_cast<int>(s.count())))
+      return false;
+    // For the sub-second part, we print either '.0' if there's none, or print
+    // down to the lowest resolution necessary (all the way down to ns).
```
mavam: Could (or should) we avoid `.0` entirely if there's no sub-second resolution?

Neverlord: I tried to keep the impact of changes to a minimum, but now that you bring it up: I think from a user perspective it's indeed better to omit the `.0` at the end.

CHANGELOG.md
Neverlord: The first sentence gets repeated in your suggestion. I suppose the "Previously, ..." sentence was an oversight?

mavam: Whoops, right. Here's the proper text:
```
Timestamps were always printed in millisecond resolution, which lead to loss of
precision when the internal representation had a higher resolution. Timestamps
are now rendered up to nanosecond resolution - the maximum resolution
supported.
```

Also, some more explanation for the rewording: what I wanted to change here ist the focus on the bug (= loss of precision). The use case with copy-and-paste confused me initially. I'm sure there's also many other cases where the precision matters.

PR #540, opened 2019-07-18
Enable field-based access of table-slice columns

Neverlord:
The new visitors for row- and column-based access provide a more intuitive API to use table slices programmatically, especially in unit tests.

### DIFF
--- a/libvast/src/table_slice.cpp
+++ b/libvast/src/table_slice.cpp
@@ -13,7 +13,6 @@
 
 #include "vast/table_slice.hpp"
 
-#include <stdexcept>
 #include <unordered_map>
 
 #include <caf/actor_system.hpp>
--- a/libvast/src/table_slice.cpp
+++ b/libvast/src/table_slice.cpp
@@ -106,12 +106,13 @@ table_slice::column_view table_slice::column(size_t index) const {
   return {*this, index};
 }
 
-table_slice::column_view table_slice::column(std::string_view name) const {
+caf::optional<table_slice::column_view>
+table_slice::column(std::string_view name) const {
   auto& fields = header_.layout.fields;
   for (size_t index = 0; index < fields.size(); ++index)
     if (fields[index].name == name)
-      return {*this, index};
-  throw std::logic_error("invalid column name");
+      return column_view{*this, index};
+  return caf::none;
 }
 
 caf::error table_slice::load(chunk_ptr chunk) {
--- a/libvast/src/table_slice.cpp
+++ b/libvast/src/table_slice.cpp
@@ -13,6 +13,7 @@
 
 #include "vast/table_slice.hpp"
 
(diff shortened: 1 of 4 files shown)


libvast/vast/table_slice.hpp
```diff
@@ -103,11 +159,23 @@ class table_slice : public caf::ref_counted {
...
     return header_.layout.fields.size();
   }
 
+  /// @returns a column view for the given `index`.
+  /// @pre `column < columns()`
+  column_view column(size_t index) const;
+
+  /// @returns a view for the column with given `name`.
+  /// @pre there is a c such that `c == column_name(name)`
+  column_view column(std::string_view name) const;
```
mavam: I find the precondition hard to validate from the user perspective. It would require grabbing the layout and making sure one of the fields has the given name. Therefore, I'd change the interface so that `column` returns `optional<column_view>`. When the user is sure that the name exists, the deref is easy to perform. Otherwise the checking is explicit.

It also obviates the need to rely on exceptions for enforcing of preconditions. (For these, we usually use assertions anyway.)

libvast/test/format/pcap.cpp
```diff
@@ -89,12 +89,9 @@ TEST(PCAP read/write 1) {
   auto src_field = slice->at(43, 1);
   auto src = unbox(caf::get_if<view<address>>(&src_field));
   CHECK_EQUAL(src, unbox(to<address>("192.168.1.1")));
-  auto community_id_at = [&](size_t row) {
-    auto id_field = slice->at(row, 5);
-    return unbox(caf::get_if<view<std::string>>(&id_field));
-  };
+  auto id_column = slice->column("community_id");
```
mavam:
```suggestion
  auto community_id_column = slice->column("community_id");
```

libvast/vast/table_slice.hpp
```diff
@@ -40,6 +40,62 @@ class table_slice : public caf::ref_counted {
 
   static constexpr size_type npos = std::numeric_limits<size_type>::max();
 
+  /// Convenience helper for traversing a column.
```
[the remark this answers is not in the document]
Neverlord: What additional benefit would `relates` add here? The view types are member types, so they already are strongly coupled to `table_slice` and listed along the outer class by `Doxygen`.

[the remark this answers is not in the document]
mavam: That's right. I glossed over the fact that these are member types and assumed that they reside in namespace `vast`.

libvast/src/table_slice.cpp
```diff
@@ -13,6 +13,7 @@
 
 #include "vast/table_slice.hpp"
 
+#include <stdexcept>
```
mavam: I don't think we need this include any longer.

PR #542, opened 2019-07-19
Support comments in schema language

mavam:
This PR adds the capability to add comments to the schema language.


mavam:
>  I have only two small nitpicks regarding constexpr/inline. See them more as comments.

Good ones; I've addressed these directly.

Now to your actual comment. I agree that the "automagic" comment skipping is not ideal and confusing. In this PR, I wanted to re-instantiate the old behavior where we actually had comments in the schema language. This ability got lost at some point.

The whitespace parser is indeed a non-intuitive name, because it parsers more than whitespace. It consumes all irrelevant characters that do not pertain to a valid token. In its current version, the parser framework does a poor job at separating this aspect from the actual parsing logic. At some point in the future, we need to be able to address this so that we don't always have to write parsers as `x >> ws >> y >> ws >> z` where `x,y,z` are the meat and `ws` is just whitespace. Instead, one should be able to write `p = x >> y >> z` and specify the skipper when invoking the parser. Getting there requires some fundamental changes in the framework and refactoring of existing parsers, which would be out of scope of this PR.

Long story short, I didn't want to touch the implementation of `type_parser` because I was hoping to address the whole skipping thing first. Until we are there, I've now renamed the `ws` parser to `skp` and updated the points of use.

PR #545, opened 2019-07-23
Fix SFINAE expression in vast::span constructor

Neverlord:


Neverlord:
@tobim Yes, there are more fixes in Tenzir necessary after merging this PR. But this PR is a necessary prerequisite before I can open the PR in Tenzir land.

tobim:
> @tobim Yes, there are more fixes in Tenzir necessary after merging this PR. But this PR is a necessary prerequisite before I can open the PR in Tenzir land.

ah, that's fine then.

PR #547, opened 2019-07-23
Consolidate pcap::connection and community_id::flow

Neverlord:


libvast/vast/flow.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+/// @returns the protocol of a flow tuple.
+/// @param x The flow to extract the protocol from.
+/// @relates flow
+port::port_type protocol(const flow& x);
+
+/// @returns a hash value for a flow tuple.
+/// @param x The flow to compute the hash value from.
+/// @relates flow
+size_t hash(const flow& x);
```
mavam: Could you elaborate on the need for this function? I wonder how much it adds on top of `std::hash<flow>`. This is where I would see the implementation of this going. At least this is what we do with all other types.

libvast/test/flow.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+struct fixture {
+  flow tcp_flow = flow{unbox(to<address>("10.10.0.1")),
+                       unbox(to<address>("10.10.0.2")),
+                       unbox(to<port>("123/tcp")), unbox(to<port>("321/tcp"))};
+
+  flow udp_flow = flow{unbox(to<address>("10.10.0.1")),
+                       unbox(to<address>("10.10.0.2")),
+                       unbox(to<port>("123/udp")), unbox(to<port>("321/udp"))};
+};
```
mavam: Is there a reason why ICMP and ICMP6 are not part of the fixture?

libvast/vast/community_id.hpp
```diff
@@ -51,29 +51,12 @@ namespace community_id {
...
-}
-
-/// Computes a hash of a flow.
+/// Computes a hash of a flow according to the community ID specification.
 /// @param hasher The hash algorithm to use.
 /// @param x The flow to hash.
 /// @relates flow
 template <class Hasher>
-void hash_append(Hasher& hasher, const flow& x) {
+void community_hash_append(Hasher& hasher, const flow& x) {
```
mavam:
```suggestion
void community_id_hash_append(Hasher& hasher, const flow& x) {
```

libvast/vast/flow.hpp
Neverlord: The extra function allows the implementation to move to the `.cpp` file instead. Since the implementation relies on `uhash` and `xxhash`, this avoids including `vast/concept/hashable/uhash.hpp` and `vast/concept/hashable/xxhash.hpp` in `flow.hpp` in order to minimize header dependencies and speeding up compilation.

libvast/test/flow.cpp
Neverlord: The purpose of these two test flows is to have flows that differ only in their port type. Adding more protocols would not result in more coverage and only add redundant checks.

PR #548, opened 2019-07-24
Fix PCAP export

mavam:
Adding the Community IDs to the PCAP packet type changed the offset of the payload. This resulted in the Community ID mistakenly being used as packet data.


mavam:
I did not add a changelog entry, on purpose, because it would only add churn at this point. Consider this more like a hotfix for PCAP Community ID PR #539.

PR #549, opened 2019-07-26
Fix operators > and < for types with binning

tobim:


CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🐞 Queries with the less `<` or greater `>` operators would sometimes produce
+  incorrect results for the `duration` type. The results should now be correct.
```
mavam: "should now be correct" doesn't sound very convincing. Let's be more confident and write this as:

```
- 🐞 Queries with the less `<` or greater `>` operators produced off-by-one
  results for the `duration` when the query contained a finer resolution than the
  index. The operator now works as expected.
```

libvast/vast/bitmap_index.hpp
```diff
@@ -86,7 +86,21 @@ class bitmap_index
...
+    auto binned = binner_type::bin(x);
+    // In case the binning causes a loss of precision, the comparison value
+    // has to be adjusted by 1. E.g. a query for `dat > 1.1` will be
+    // transformed to `dat > 1` by the binner, which would result in a loss
+    // of the value range between 1.0 and 2.0. False positives are filtered
+    // out in the candidate check at a later stage.
+    if constexpr (!std::is_same_v<binner_type, identity_binner>) {
+      if (op == greater) {
+        --binned;
+      }
```
mavam: Removing the braces would reduce a bit of visual noise.

CHANGELOG.md
tobim: done

PR #551, opened 2019-08-01
Port VAST to lastet node_id API

Neverlord:
As of https://github.com/actor-framework/actor-framework/pull/863, `node_id` is now an opaque type. Hence, the accountant can no longer split it into its sub-components and instead has to print the node as a single field.


mavam:
Thanks for the downstream adjustment!

mavam:
@tobim Can you check whether we need to adjust downstream scripts? Perhaps the removal of an intermediate column broke some post-processing logic.

PR #553, opened 2019-08-05
Allow pattern query for type attribute extractors 

tobim:


tobim:
https://jenkins.inet.haw-hamburg.de/job/VAST/job/vast/job/story%252Fch7647/

PR #555, opened 2019-08-19
Fix gcc9 warnings

tobim:


libvast/vast/concept/support/unused_type.hpp
```diff
@@ -25,11 +25,11 @@ struct unused_type : detail::equality_comparable<unused_type>,
...
 
-  unused_type& operator=(const unused_type&) = default;
+  // unused_type& operator=(const unused_type&) = default;
 
-  const unused_type& operator=(const unused_type&) const {
-    return *this;
-  }
+  // const unused_type& operator=(const unused_type&) const {
+  //  return *this;
+  //}
```
mavam: Can you add a comment that explains why these are commented? Otherwise one would have to go through the git history to figure this out.

tobim: this is an oversight, they were meant to be removed. The explicit definition is not necessary because the auto generated ones are the same.

PR #556, opened 2019-08-21
Add CAF and Broker as submodules 

tobim:
Todo:
- [ ] Fix broker cmake integration.


tobim:
Blocked by https://github.com/zeek/broker/issues/51.

PR #560, opened 2019-08-26
Bake full documentation into commands

mavam:


dominiklohmann:
I've rebased this onto master and will continue working on this.

mavam:
Now that commands have native support for documentation, we could rebase this PR and start adding docs, one command at a time, in here.

dominiklohmann:
I've already started with that—the only documentation so far is the man page markdown as documentation for the root command.

mavam:
Ah, okay. How would you like proceed with this PR? I would be happy to write some documentation now that the infrastructure is there for it.

dominiklohmann:
I don't mind you pushing onto this branch directly. You could also just send me the text somewhere, whatever fits you best currently.

mavam:
The last commit still has `(WIP)` in the summary. Are you still working on this before new commits can be added to this branch?

dominiklohmann:
You can safely ignore it, I can always rebase and reorder later. I've got nothing much here that's not yet pushed, mostly small bits I extracted from docs.tenzir.com.

mavam:
I have another suggestion for tweaking the scaffold so that actually writing the documentation becomes a bit easier. Having everything as single string and copying it into the source is a bit error-prone; it's easy to mess up the syntax. How about this:

- Have a new directory `/doc/commands` that contains all the `*.md` files
- Let CMake slurp this files and create a macro from it [*]
- Use the CMake-defined macros in the C++ source, e.g., `VAST_DOC_CMD_START` or `VAST_DOC_CMD_IMPORT_ZEEK`

This makes it also easier to use third-party tools to process the markdown files statically.

[*] That's the simplest design I can think of. Generating extra sources files that get included later on seems a bit too heavy-weight.

dominiklohmann:
Honestly that's fine by me, can do either way. I'd suggest creating files named `doc/cli/<full_name>.md` (e.g., `vast.md` for the root command and `spawn_accountant` for the spawn accountant command).

dominiklohmann:
I've pushed a simple, initial design for reading the documentation files in CMake. By using `configure_file` over passing the variables directly we avoid having very long compiler invocations.

mavam:
Every file we install via `make install` (= that are part of the distribution) should have the banner. I would like include also auto-generated files.

I actually added the banner so that we can have a well-defined one to compare against, e.g., so that we can write a script that creates a file template from this file.

Maybe the location in the top-level dir is simply not the best?!

PR #564, opened 2019-08-28
Enable Cirrus CI

tobim:
- [x] Add macOS build
- [ ] Prettify names


mavam:
The *style* check seems to fail despite the checkmark:

```
git diff -U0 --no-color $(git merge-base origin/master HEAD) | scripts/clang-format-diff.py -p1;
fatal: Not a valid object name origin/master
```

.cirrus.yml
```diff
@@ -0,0 +1,58 @@
+style_task:
+  name: Style (clang-format)
+  container:
+    dockerfile: .ci/debian/Dockerfile
+    cpu: 1
+    memory: 1G
+  clang-format_script:
+    - git diff -U0 --no-color $(git merge-base origin/master HEAD) | scripts/clang-format-diff.py -p1;
```
mavam:
```suggestion
    - git diff -U0 --no-color $(git merge-base origin/master HEAD) | scripts/clang-format-diff.py -p1
```

tobim:
> The _style_ check seems to fail despite the checkmark:
> 
> ```
> git diff -U0 --no-color $(git merge-base origin/master HEAD) | scripts/clang-format-diff.py -p1;
> fatal: Not a valid object name origin/master
> ```

Thanks for pointing that out, it is fixed now.

tobim:
To get really nice names we'd have to stop using the `matrix` feature, because the option alternatives are always added to the name field. This would not be extremely painful, because of the yaml anchor feature, but still a bit more repetitive than what we have now. 
As for `X_task` overriding the name key: This is a known cirrus bug and will sort itself out.

In all, I consider this PR complete.

PR #565, opened 2019-09-02
Improve identification of OpenSSL install root

mavam:
This PR makes it easier to specify a custom install prefix for OpenSSL. Additionally, if there exists an OpenSSL installation from Homebrew on macOS, we'll try to use it by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -417,6 +417,14 @@ if (VAST_ENABLE_ASSERTIONS)
   endif ()
 endif ()
 
+# On macOS, Homebrew installs OpenSSL into /usr/local/opt. We'll use this when
+# it's present.
+if (APPLE)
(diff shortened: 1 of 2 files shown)


dominiklohmann:
Just a small note, because I was looking through this project and have had an issue with this previously: A more reliable and future-proof way to get the installation directory from Homebrew is to run the shell command `brew --prefix openssl`, which will return `/usr/local/opt/openssl` for 99.9% of installations.

tobim:
Thanks for sharing your insight. Imho this is enough of a corner case that it does not warrant shelling out. The workaround added here covers most users, but we should not try to be exhaustive here. If we want to make every homebrew user happy, then we should think about macports too. Otoh, requiring `-DOPENSSL_ROOT_DIR=` is acceptable.

PR #577, opened 2019-09-16
Generate tarballs in CI pipeline

tobim:


mavam:
Looks like `pkg install base64` is missing on the FreeBSD checks.

tobim:
> Looks like `pkg install base64` is missing on the FreeBSD checks.

It is installed now.

PR #579, opened 2019-09-17
Fix UB from not using std::launder where necessary

dominiklohmann:


libvast/test/span.cpp
```diff
@@ -36,7 +37,7 @@ TEST(byte) {
   auto x = span<byte>{&b, 1};
   CHECK_EQUAL(x.size(), 1);
   auto foo = "foo"s;
-  x = span<byte>(reinterpret_cast<byte*>(foo.data()), foo.size());
+  x = span<byte>(reinterpret_cast<byte*>(std::launder(foo.data())), foo.size());
```
dominiklohmann: This one seems unnecessary as `alignof(byte) == alignof(decltype(*foo.data()))`

PR #583, opened 2019-09-18
Store accounting data in VAST native format

tobim:
This PR will change the ACCOUNTANT to relay its data back to the IMPORTER in the form of table slices. That will allow us to query the data flexibly and export it in the desired format.

TODO:
- [x] Have a table_slice_builder in ACCOUNTANT
- [x] Periodically send slices to IMPORTER
- [x] Add integration tests

### DIFF
--- a/integration/default_set.yaml
+++ b/integration/default_set.yaml
@@ -41,7 +41,7 @@ tests:
       - command: export -N ascii 'orig_bytes > 1k && orig_bytes < 1Ki'
       - command: export -N ascii ':string == "OrfTtuI5G4e" || :port == 67/udp'
       - command: export -N ascii '#type == "zeek.conn" && resp_h == 192.168.1.104'
-      - command: export -N ascii '#type != "zeek.conn"'
+      - command: export -N ascii '#type != "zeek.conn" && \#type != "vast.account"'
       - command: export -N ascii '#type != "foobar" && resp_h == 192.168.1.104'
   Node Zeek multiple imports:
     tags: [node, import-export, zeek]
--- a/integration/default_set.yaml
+++ b/integration/default_set.yaml
@@ -95,6 +95,17 @@ tests:
       - command: import -b zeek
         input: data/zeek/conn.log.gz
       - command: export ascii 'resp_h == 192.168.1.104'
+      # import some more to make sure accounting data is in the system.
+      - command: import -b zeek
+        input: data/zeek/conn.log.gz
+      - command: import -b zeek
+        input: data/zeek/conn.log.gz
+      - command: import -b zeek
+        input: data/zeek/conn.log.gz
+      - command: import -b zeek
+        input: data/zeek/conn.log.gz
+      - command: status
+        transformation: jq '.node.index.statistics.layouts | del(."vast.account".count)'
   Server Basic bgpdump:
     tags: [server, import-export, bgpdump]
     fixture: ServerTester
--- a/integration/integration.py
+++ b/integration/integration.py
@@ -9,6 +9,8 @@ import difflib
 import filecmp
 import gzip
 import itertools
+import json
+import jsondiff
 import logging
 import os
 import shlex
@@ -40,6 +42,7 @@ class Fixture(NamedTuple):
 class Step(NamedTuple):
     command: List[str]
     input: Optional[Path]
+    transformation : Optional[str]
 
 class Test(NamedTuple):
     tags: Optional[str]
@@ -77,6 +80,18 @@ def spawn(*popenargs, **kwargs):
     proc = subprocess.Popen(*popenargs, **kwargs)
     CURRENT_SUBPROCS.append(proc)
     return proc
+
+def try_wait(process, timeout):
+    """Wait for a specified time or terminate the process"""
+    try:
+        if process.wait(timeout) is not 0:
+            LOGGER.error(f'{process.args} returned {process.returncode}')
+            return Result.ERROR
+        return Result.SUCCESS
+    except subprocess.TimeoutExpired:
+        LOGGER.error(f'timeout reached, terminating process')
+        process.terminate()
+        return Result.ERROR
 
 def run_flamegraph(args, svg_file):
     """Perform instrumentation and produce an output svg"""
@@ -132,18 +147,16 @@ def empty(iterable):
         return True
     return False
 
+def deduce_format(command):
+    positionals = list(filter(lambda x: x[0] != '-', command))
+    if positionals[0] == 'export':
+        if positionals[1] == 'json':
+            return 'json'
+    if positionals[0] == 'status':
+        return 'json'
+    return 'ascii'
+
 def run_step(basecmd, step_id, step, work_dir, baseline_dir, update_baseline):
-    def try_wait(process, timeout):
-        """Wait for a specified time or terminate the process"""
-        try:
-            if process.wait(timeout) is not 0:
-                LOGGER.error(f'{process.args} returned {process.returncode}')
-                return Result.ERROR
-            return Result.SUCCESS
-        except subprocess.TimeoutExpired:
-            LOGGER.error(f'timeout reached, terminating process')
-            process.terminate()
-            return Result.ERROR
     try:
         stdout = work_dir / f'{step_id}.out'
         stderr = work_dir / f'{step_id}.err'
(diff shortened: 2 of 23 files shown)


libvast/src/system/accountant.cpp
```diff
@@ -179,9 +241,8 @@ accountant_type::behavior_type accountant(accountant_actor* self,
...
             time ts = std::chrono::system_clock::now();
             for (const auto& [key, value] : r) {
-              auto f = [&, key = key](const auto& x) {
-                record(self, key, x, ts);
-              };
+              auto f
+                = [&, key = key](const auto& x) { record(self, key, x, ts); };
               caf::visit(f, value);
             }
           },
```
mavam: This lambda sequence looks like it could benefit from re-formatting. However, you did the right thing by not doing it together with this commit, which makes the changes stand out clearly.

libvast/src/system/accountant.cpp
```diff
@@ -134,21 +175,42 @@ accountant_type::behavior_type accountant(accountant_actor* self,
...
+      VAST_TRACE(self, "was asked for", num, "slices and produced", produced,
+                 ".", st.slice_buffer.size(), "are remaining in buffer");
+    },
+    // done?
+    [](const bool&) { return false; });
   return {[=](announce_atom, const std::string& name) {
             self->state.actor_map[self->current_sender()->id()] = name;
             self->monitor(self->current_sender());
+            if (name == "importer") {
+              VAST_DEBUG(self, "importer is announced");
```
mavam: Just for clarification: does this message still result in our *subject-verb-object* style of logging?

libvast/src/system/accountant.cpp
```diff
@@ -134,21 +175,42 @@ accountant_type::behavior_type accountant(accountant_actor* self,
...
+      auto& st = self->state;
+      size_t produced = 0;
+      while (num-- > 0 && !st.slice_buffer.empty()) {
+        auto& slice = st.slice_buffer.front();
+        produced += slice->rows();
+        out.push(std::move(slice));
+        st.slice_buffer.pop();
+      }
+      VAST_TRACE(self, "was asked for", num, "slices and produced", produced,
+                 ".", st.slice_buffer.size(), "are remaining in buffer");
```
mavam:
```suggestion
                 ";", st.slice_buffer.size(), "are remaining in buffer");
```
Since none of our logging messages and with a period, maybe `;` seems like a more natural separator.

libvast/src/system/accountant.cpp
```diff
@@ -81,6 +99,29 @@ void record(accountant_actor* self, const std::string& key, T x,
...
+                    {"key", string_type{}}, {"value", string_type{}}}
+          .name("vast.account");
+
+    auto slice_type = get_or(sys.config(), "system.table-slice-type",
+                             defaults::system::table_slice_type);
+    st.builder = factory<table_slice_builder>::make(slice_type, layout);
+    VAST_DEBUG(self, "obtained builder");
+  }
+  if (!st.builder->add(ts, to_string(node), actor_id, st.actor_map[actor_id],
+                       key, to_string(x)))
```
mavam: Can this actually fail? If not (because the programmer dictates the schema), then we should probably make this an assertion.

libvast/src/system/accountant.cpp
```diff
@@ -64,12 +68,26 @@ void init(accountant_actor* self, const path& filename) {
...
+                         defaults::system::table_slice_size);
+}
+
+void finish_slice(accountant_actor* self) {
+  auto& st = self->state;
+  // Do nothing if builder has not been created.
+  if (!st.builder)
+    return;
+  auto slice = st.builder->finish();
+  VAST_DEBUG(self, "generated slice with", slice->rows(), "rows");
```
mavam: Maybe add an assertion here that `slice->rows() > 0`?

libvast/vast/fwd.hpp
```diff
@@ -91,6 +91,9 @@ void intrusive_ptr_add_ref(const table_slice*);
 void intrusive_ptr_release(const table_slice*);
 table_slice* intrusive_cow_ptr_unshare(table_slice*&);
 
+void intrusive_ptr_add_ref(const table_slice_builder*);
+void intrusive_ptr_release(const table_slice_builder*);
```
mavam: Just out of curiosity: why are these needed?

libvast/src/system/accountant.cpp
```diff
@@ -41,31 +40,10 @@ namespace {
...
-    auto e = make_error(ec::filesystem_error, "failed to open file:", filename);
-    self->quit(e);
-    return;
-  }
-  file << "timestamp\tnode\taid\tactor_name\tkey\tvalue\n";
-  if (!file)
-    self->quit(make_error(ec::filesystem_error));
-  VAST_DEBUG(self, "kicks off flush loop");
-  self->send(self, flush_atom::value);
 #if VAST_LOG_LEVEL >= CAF_LOG_LEVEL_INFO
```
mavam: This is different from your changes, but I find this pre-processor conditional not very intuitive. To me, it reads like "only perform this when logging is enabled". Can we have a clearer macro for that?

integration/default_set.yaml
```diff
@@ -41,7 +41,7 @@ tests:
       - command: export -N ascii 'orig_bytes > 1k && orig_bytes < 1Ki'
       - command: export -N ascii ':string == "OrfTtuI5G4e" || :port == 67/udp'
       - command: export -N ascii '#type == "zeek.conn" && resp_h == 192.168.1.104'
-      - command: export -N ascii '#type != "zeek.conn"'
+      - command: export -N ascii '#type != "zeek.conn" && \#type != "vast.account"'
```
mavam: I find this more idiomatic: `#type !in {"zeek.conn", "vast.account"}`.

integration/integration.py
```diff
@@ -173,34 +186,72 @@ def try_wait(process, timeout):
             LOGGER.debug('standard error:')
             for line in open(stderr).readlines()[-100:]:
                 LOGGER.debug(f'    {line}')
+            return result
         # Perform baseline update or comparison.
         baseline = baseline_dir / f'{step_id}.ref'
+        output_format = deduce_format(step.command)
```
mavam: Couldn't you just try to attempt to parse the output as JSON with builtin Python tools?

libvast/src/system/accountant.cpp
tobim: This was added in an attempt do debug a problem. I forgot to take it out. I'm thinking of creating a new macro for ad-hoc debugging. One that would be detected by a pre-commit hook and reject the commit. Maybe `clang-tidy` can help with that.

integration/default_set.yaml
tobim: This does currently not work.

integration/integration.py
tobim: What would be the advantage of that approach?

libvast/vast/fwd.hpp
tobim: Are you asking why these functions exist in the first place or why are they forward declared here?

libvast/vast/system/accountant.hpp
```diff
@@ -82,6 +84,12 @@ struct accountant_state {
 
   accountant_state(accountant_type::stateful_base<accountant_state>* self);
   void command_line_heartbeat();
+
+  size_t slice_size;
+  table_slice_builder_ptr builder;
+  using downstream_manager = caf::broadcast_downstream_manager<table_slice_ptr>;
```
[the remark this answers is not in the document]
tobim: `accountant_state` deserves a cleanup. It will be pushed shortly.

libvast/src/system/accountant.cpp
tobim: This was actually possible if `finish_slice()` was called from the exit handler. Rectified.

libvast/src/system/accountant.cpp
tobim: Fixed

libvast/src/system/accountant.cpp
tobim: `clang-format` keeps this entire behavior as it is...

integration/integration.py
mavam: It seemed to be simpler and more maintainable than looking at the command structure. This was just an idea, though, no need to budge.

libvast/src/system/accountant.cpp
mavam: Okay.

libvast/vast/fwd.hpp
mavam: The former.

mavam:
Dang, I noticed one more thing that was missing right after merging: could you add a changelog entry that describes how one can access the accountant data?

libvast/src/system/accountant.cpp
tobim: > To me, it reads like "only perform this when logging is enabled".

That is correct however. We don't want to print the heartbeat if verbosity is set to `WARNING` or lower.

mavam: Okay, I think I understand the technical aspect. But wouldn't you agree that this is a hard-to-read of conveying "logging enabled?"

PR #584, opened 2019-09-23
Remove unused Twitter account from README

dominiklohmann:
As discussed last Wednesday at HQ.


tobim:
Maybe we should add a link to @tenzir_company?

PR #585, opened 2019-09-24
Improve gitignore

dominiklohmann:
Just minor changes.

- Explicitly use a `/` prefix for files that are in the root directory
- Add output files from running `vast` and `vast-test`


tobim:
This would have been a nice opportunity to test the `[skip ci]` tag.

dominiklohmann:
> Could you add `.mypy_cache/` while you're at it?

Done.

> This would have been a nice opportunity to test the `[skip ci]` tag.

Seems to work. Tests were marked as successful immediately after pushing.

PR #594, opened 2019-10-04
Add color to diagnostics when using Ninja

dominiklohmann:


configure
```diff
@@ -108,6 +108,9 @@ while [ $# -ne 0 ]; do
       ;;
     --generator=*)
       CMakeGenerator="$optarg"
+      if [ "$CMakeGenerator" = "Ninja" ]; then
```
mavam: Maybe also support `"ninja"` as lower-case alternative?

dominiklohmann: Not supported by CMake.

CMakeLists.txt
```diff
@@ -79,6 +79,16 @@ set(CMAKE_VERBOSE_MAKEFILE
     false
     CACHE STRING "Show all outputs including compiler lines." FORCE)
 
+# Force colored output (mainly for Ninja)
+option (FORCE_COLORED_OUTPUT "Always produce ANSI-colored output." FALSE)
```
tobim: I would prefer if we set the default to true if Ninja is used as generator. Then you won't need the change in `configure`.

dominiklohmann: Addressed this change.

CMakeLists.txt
```diff
@@ -79,6 +79,20 @@ set(CMAKE_VERBOSE_MAKEFILE
...
+else ()
+  option(FORCE_COLORED_OUTPUT "Always produce ANSI-colored output." OFF)
+endif ()
+if (${FORCE_COLORED_OUTPUT})
+    if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+       add_compile_options(-fdiagnostics-color=always)
+    elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+       add_compile_options(-fcolor-diagnostics)
+    endif ()
+endif ()
```
tobim: Afaik CMake has a different compiler id for apple clang. This can be worked around by using `MATCHES` instead of `STREQUL`.
```
option(VAST_COLOR_NINJA "Enable ANSI-colored output for the Ninja generator." ON)
if ("${CMAKE_GENERATOR}" STREQUAL "Ninja" AND VAST_COLOR_NINJA)
  if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
    add_compile_options(-fdiagnostics-color=always)
  elseif ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
    add_compile_options(-fcolor-diagnostics)
  endif ()
endif ()
```

PR #596, opened 2019-10-09
Add count command

dominiklohmann:


libvast/src/system/count_command.cpp
```diff
@@ -0,0 +1,124 @@
...
+  auto first = invocation.arguments.begin();
+  auto last = invocation.arguments.end();
+  const auto& options = invocation.options;
+  // Read query from input file, STDIN or CLI arguments.
+  std::string query;
+  auto assign_query = [&](std::istream& in) {
+    query.assign(std::istreambuf_iterator<char>{in},
+                 std::istreambuf_iterator<char>{});
+  };
+  if (auto fname = caf::get_if<std::string>(&options, "count.read")) {
```
mavam: This whole dispatching logic seems very similar to `detail::make_(input|output)_stream`. I wonder if we can reuse this logic here.

libvast/src/system/application.cpp
```diff
@@ -64,6 +65,15 @@ auto make_root_command(std::string_view path) {
                                       "definitions"));
 }
 
+auto make_count_command() {
+  return std::make_unique<command>(
+    "count", "count hits for a query without exporting data", "",
+    opts("?count")
+      .add<bool>("node,N", "spawn a node instead of connecting to one")
```
mavam: The whole `-N` thing seems to be an orthogonal concern. I wonder if it can somehow be factored. @tobim, any ideas?

libvast/src/system/count_command.cpp
dominiklohmann: I'm not quite sure I grok `make_input_stream` after looking at it for a bit. Do you mean something along the lines of:

```c++
if (auto is = detail::make_input_stream<detail::count>(options))
  assign_query(*is);
```

libvast/src/system/application.cpp
dominiklohmann: As a side note—currently, the customization point is not executed for `count -N`. We decided to move the customization point out of the start command some time in the future.

libvast/src/system/count_command.cpp
tobim: This code does already exist in `sink_command.cpp:52-90`, `make_input_stream` is something different.

libvast/src/system/application.cpp
tobim: We could move the `-N;--node` option to the root command, and check for it in `spawn_or_connect_to_node()`. That would simplify the commands and that function.

dominiklohmann: I'll get this done in another PR once this is merged.

libvast/src/system/count_command.cpp
tobim: Before we create a separate task and the whole ceremony for this, lets just consolidate the functionality within this PR. To make the request more explicit: Please factor the shared logic between `count_command` and `sink_command` into a dedicated function that will be used by both.

dominiklohmann:
[story/ch8984]

libvast/src/system/count_command.cpp
dominiklohmann: I implemented this as requested and factored it into a shared utility function named `vast::system::read_query` (for the lack of a better name).

PR #599, opened 2019-10-10
Implement vast pivot subcommand

tobim:


libvast/vast/system/read_query.hpp
```diff
@@ -21,6 +21,7 @@ namespace vast::system {
 
 /// Reads query from input file, STDIN or CLI arguments.
 caf::expected<std::string>
-read_query(const command::invocation& invocation, std::string_view option_name);
+read_query(const command::invocation& invocation, std::string_view option_name,
+           size_t argument_offset = 0);
```
mavam: Can you document the parameters?

libvast/vast/system/pivoter.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+  static inline constexpr const char* name = "pivoter";
+
+  // -- constructors, destructors, and assignment operators --------------------
+
+  pivoter_state(caf::event_based_actor* self);
+
+  // -- member variables -------------------------------------------------------
+
+  std::string target;
```
mavam: Can you document this member?

libvast/test/system/pivoter.cpp
```diff
@@ -0,0 +1,143 @@
...
+1258531924.321413	CojK5e1MpFgJnwlp6a	192.168.1.103	68	192.168.1.1	67	udp	dhcp	0.044779	303	300	SF	-	-	0	Dd	1	331	1	328	-	1:oG55uQUH+XuHYHOFV0c+yOutW8E=
+1258531939.613071	Cgq2X52qyXxAAi9avc	192.168.1.102	138	192.168.1.255	138	udp	-	-	-	-	S0	-	-	0	D	1	229	0	0	-	1:4iHhzk49NeoFdK6VHSCw4ruRbsw=
+1258532046.693816	CwF9px2owZEPahqWsg	192.168.1.104	68	192.168.1.1	67	udp	dhcp	0.002103	311	300	SF	-	-	0	Dd	1	339	1	328	-	-
+1258532143.457078	Cu0QHL1w6Dp3Z5y0Pg	192.168.1.102	1170	192.168.1.1	53	udp	dns	0.068511	36	215	SF	-	-	0	Dd	1	64	1	243	-	1:FVMx3YawO69eZmiaMJJbrs6447E=
+1258532203.657268	C37rla3neljCJgPsE2	192.168.1.104	1174	192.168.1.1	53	udp	dns	0.170962	36	215	SF	-	-	0	Dd	1	64	1	243	-	1:79fDvfGNCWV1JBYjXCE5Ov1FuMM=
+1258532331.365294	C4ik1w2JPTX1zO2Ubi	192.168.1.1	5353	224.0.0.251	5353	udp	dns	0.100381	273	0	S0	-	-	0	D	2	329	0	0	-	1:aGi0Bt5ApW6HEEO7wfz+PwvniIU=
+1258532331.365330	CCGBMdPGplqLzQCjg	fe80::219:e3ff:fee7:5d23	5353	ff02::fb	5353	udp	dns	0.100371	273	0	S0	-	-	0	D	2	369	0	0	-	1:JoBDvaK4Tt6BfWSKWPKaJTELr2M=)__";
+
+template <class Reader>
+std::vector<table_slice_ptr> inhale(const char* data) {
```
mavam: This function seems to be repeated several times. We should factor it and make it available through a fixture or a helper function in our unit testing library.

libvast/test/system/pivoter.cpp
```diff
@@ -0,0 +1,143 @@
...
+  caf::actor aut;
+};
+
+} // namespace
+
+FIXTURE_SCOPE(pivoter_tests, fixture)
+
+TEST(count IP point query without candidate check) {
+  MESSAGE("build expression");
+  auto expr = unbox(vast::to<expression>("proto == udp && orig_bytes < 600"));
```
mavam: Do we really need the qualification `vast::` here?

libvast/test/system/pivoter.cpp
```diff
@@ -0,0 +1,143 @@
...
+  CHECK_EQUAL(
+    node_state.invocs[0].arguments[0],
+    "#type == \"pcap.packet\" && community_id in "
+    "[\"1:aWZfLIquYlCxKGuJ62fQGlgFzAI=\", "
+    "\"1:fLbpXGtS1VgDhqUW+WYaP0v+NuA=\", \"1:BY/pbReW8Oa+xSY2fNZPZUB1Nnk=\", "
+    "\"1:tShwwbRwEMd3S8SvqZxGyvKm+1c=\", \"1:4iHhzk49NeoFdK6VHSCw4ruRbsw=\", "
+    "\"1:+igyiyVnNTFDre/V6pYx89+Lgr8=\", \"1:5NWtNjiw4JPUO4fMM0WobJPFeU8=\", "
+    "\"1:oG55uQUH+XuHYHOFV0c+yOutW8E=\", \"1:FVMx3YawO69eZmiaMJJbrs6447E=\", "
+    "\"1:79fDvfGNCWV1JBYjXCE5Ov1FuMM=\", \"1:aGi0Bt5ApW6HEEO7wfz+PwvniIU=\", "
+    "\"1:JoBDvaK4Tt6BfWSKWPKaJTELr2M=\"]");
```
mavam: Since the order doesn't matter, I would think that our `set` type is a more natural fit as opposed to a `vector`.

libvast/src/system/spawn_pivoter.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+maybe_actor spawn_pivoter(node_actor* self, spawn_arguments& args) {
+  VAST_TRACE(VAST_ARG(args));
+  auto target_name = args.invocation.arguments[0];
+  // Parse given expression.
+  VAST_UNBOX_VAR(
+    expr, normalized_and_valided(std::next(args.invocation.arguments.begin()),
+                                 args.invocation.arguments.end()));
+  auto piv = self->spawn(pivoter, self, target_name, std::move(expr));
+  return piv;
```
mavam: Why not just return the expression above?

libvast/src/system/spawn_pivoter.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace vast::system {
+
+maybe_actor spawn_pivoter(node_actor* self, spawn_arguments& args) {
+  VAST_TRACE(VAST_ARG(args));
+  auto target_name = args.invocation.arguments[0];
+  // Parse given expression.
+  VAST_UNBOX_VAR(
+    expr, normalized_and_valided(std::next(args.invocation.arguments.begin()),
+                                 args.invocation.arguments.end()));
```
mavam: I find this hard to understand in this local scope without a comment. Can you add a comment about why you call `std::next`?

libvast/src/system/pivoter.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+#include <caf/atom.hpp>
+#include <caf/event_based_actor.hpp>
+
+namespace vast::system {
+
+namespace {
+
+caf::optional<record_field>
+common_field(const pivoter_state& st, const record_type& indicator) {
```
mavam: Can you document this function?

libvast/src/system/pivoter.cpp
```diff
@@ -0,0 +1,159 @@
...
+#if 0
+  for (auto& t : target.fields) {
+    for (auto& i : indicator.fields) {
+      if (t.name == i.name) {
+        st.cache.insert({indicator, i});
+        return i;
+      }
+    }
+  }
+#else
```
mavam: I would preface the following with a clear comment that this is an interim heuristic and nothing that stays.

libvast/src/system/pivoter.cpp
```diff
@@ -0,0 +1,159 @@
...
+    quit_if_done();
+  });
+  return {[=](vast::table_slice_ptr slice) {
+            auto& st = self->state;
+            auto pivot_field = common_field(st, slice->layout());
+            if (!pivot_field)
+              return;
+            VAST_DEBUG(self, "uses", *pivot_field, "to extract", st.target,
+                       "events");
+            auto pivot_field_name = pivot_field->name;
```
mavam: I don't think this alias adds much; I'd keep `pivot_field->name` in the remaining code.

libvast/src/system/pivoter.cpp
```diff
@@ -0,0 +1,159 @@
...
+            auto column = slice->column(pivot_field_name);
+            std::stringstream sstr;
+            sstr << "#type == \"" << st.target << "\" && " << pivot_field_name
+                 << " in [\"";
```
mavam: I would prefer to construct the expression in a type-safe manner, as opposed to going through strings and parsing. This is more resilient against syntactic changes in the language.

```cpp
disjunction expr(2);
expr[0] = predicate{type_extractor{}, equal, st.target};
expr[1] = ...
```

libvast/src/system/pivoter.cpp
```diff
@@ -0,0 +1,159 @@
...
+            size_t request_size = 0;
+            for (size_t i = 0; i < column->rows(); ++i) {
+              auto data = materialize((*column)[i]);
+              auto x = caf::get_if<std::string>(&data);
+              // Skip if no value
+              if (!x)
+                continue;
+              // Skip if id was already requested
+              if (st.requested_ids.count(*x))
```
mavam:
```suggestion
              if (st.requested_ids.count(*x) > 0)
```

libvast/src/system/pivoter.cpp
```diff
@@ -0,0 +1,159 @@
...
+              if (request_size != 0)
+                sstr << "\", \"";
+              sstr << *x;
+              st.requested_ids.insert(*x);
+              request_size++;
+            }
+            VAST_DEBUG(self, "queries for", request_size, pivot_field_name);
+            if (!request_size) {
+              return;
+            }
```
mavam: Braces not needed.

libvast/src/system/pivoter.cpp
```diff
@@ -0,0 +1,159 @@
...
+            sstr << "\"]";
+            VAST_DEBUG(self, "sends query", sstr.str());
```
mavam: I'd a bit more context about the recipient, e.g., `"sends query to node"` or `"spawns new exporter with query"`.

libvast/vast/system/pivoter.hpp
```diff
@@ -0,0 +1,88 @@
...
+  ///       for the field that represents the edge to be of another type than
+  ///       string.
+  std::unordered_set<std::string> requested_ids;
+
+  /// A cache for the connections between a source type and the target type,
+  /// to avoid multiple computations of those.
+  mutable std::unordered_map<record_type, caf::optional<record_field>> cache;
+
+  /// A tracking counter of spawned exporters. Used for lifetime management.
+  size_t outstanding_requests = 0;
```
mavam:
```suggestion
  size_t running_exporters = 0;
```

The term `request` has a very special meaning in CAF. It confused me here initially.

libvast/vast/system/pivoter.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+  /// Flag that stores if the input source is done sending table slices. Used
+  /// for lifetime management.
+  bool initial_completed = false;
```
mavam: The naming suggests that a noun is missing between the two adjectives, maybe `initial_query_completed`?

CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🎁 The new `pivot` command allows retrieval of related data from the results
+  of a search query. A typical use case for this is to make use of the rich
+  data model provided by an IDS like Suricata or Zeek for a search to export
+  the network traffic that generated these high level events. This feature is
+  currently implemented for Suricata eve.log, Zeek logs and PCAP, and requires
+  the `community_id` field to be available in the former.
+
```
mavam: I took this as a start and tweaked it a bit:

```
- 🎁 The new `pivot` command retrieves data of a related type. It inspects each
  event in a query result to find a related type. If a related type exists in
  the schema, VAST will dynamically create a new query to fetch the contextual
  data according to the type relationship. For example, if two records `T` and
  `U` share the same field `x`, and the user requests to pivot via `T.x == 42`,
  then VAST will fetch all data for `U.x == 42`. An example use case would be
  to pivot from a Zeek or Suricata log entry to the corresponding PCAP packets.
  VAST uses the field `community_id` to pivot between the logs and the packets.
  Pivoting is currently implemented for Suricata, Zeek (with [community ID
  computation](https://github.com/corelight/bro-community-id) enabled), and
  PCAP.
```

libvast/src/system/pivot_command.cpp
```diff
@@ -0,0 +1,222 @@
...
+    = spawn_or_connect_to_node(self, invocation.options, content(sys.config()));
+  if (auto err = caf::get_if<caf::error>(&node_opt))
+    return caf::make_message(std::move(*err));
+  auto& node = caf::holds_alternative<caf::actor>(node_opt)
+                 ? caf::get<caf::actor>(node_opt)
+                 : caf::get<scope_linked_actor>(node_opt).get();
+  VAST_ASSERT(node != nullptr);
+  // Start signal monitor.
+  std::thread sig_mon_thread;
+  auto guard = signal_monitor::run_guarded(sig_mon_thread, sys, 750ms, self);
```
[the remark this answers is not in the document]
tobim: Ad-hoc refactoring within PRs of semi-related changes is a difficult tradeoff. On the one hand, it helps combat the degradation of code quality, but on the other it gets even harder to estimate the amount of work that a feature requires and puts extra cognitive load onto the reviewer that they should not have to deal with in this context.

In this case I just copied those lines from the `sink_command` and did not think much about it. I do however agree that a `make_scoped_signal_monitor()` with this sleep duration as default would be an improvement.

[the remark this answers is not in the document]
mavam: Awareness for technical debt is important. It's indeed a trade-off when it comes to release velocity. At the end of the day it's critical that these decisions are *deliberate*. To make this explicit to the reviewer, a `TODO` comment would greatly help. Internally, we should open another story for refactoring this at a later point in time.

libvast/test/system/pivoter.cpp
tobim: This is true, `libast_test/src/events.cpp` contains another `inhale` function, but that one takes a file name parameter, and is therefore not suitable to be used here. I consider refactoring that as out of scope for this PR though.

libvast/test/system/pivoter.cpp
tobim: I added `using namespace vast;` to the file.

tobim:
Most comments have been addressed, the rest will follow soon.

doc/cli/vast-pivot.md
```diff
@@ -0,0 +1,12 @@
...
+exists in the schema, VAST will dynamically create a new query to fetch the
+contextual data according to the type relationship.
+
+```sh
+vast pivot [options] <type> <expr>
+```
+
+VAST uses the field `community_id` to pivot between logs and packets. Pivoting
+is currently implemented for Suricata, Zeek (with [community ID computation]
+(https://github.com/corelight/bro-community-id) enabled), and PCAP.
```
mavam: I don't think breaking a Markdown link between its two components will render the expected result.

PR #602, opened 2019-10-15
Cleanup VAST output

dominiklohmann:


CHANGELOG.md
```diff
@@ -86,9 +86,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 - 🎁 Configuring how much status information gets printed to STDERR previously
   required obscure config settings. From now on, users can simply use
-  `--verbosity=<level>`, where `<level>` is one of `quiet`, `error`, `warn`,
-  `info`, `debug`, or `trace`. However, `debug` and `trace` are only available
-  on debug builds (otherwise they fall back to `info` output level).
+  `--verbosity=<level>,-v <level>`, where `<level>` is one of `quiet`, `error`,
+  `warn`, `info`, `debug`, or `trace`. However, `debug` and `trace` are only
+  available on debug builds (otherwise they fall back to `info` output level).
```
```suggestion
  available for debug builds (otherwise they fall back to log level `info`).
```

libvast/src/system/connect_to_node.cpp
```diff
@@ -47,6 +49,7 @@ connect_to_node(scoped_actor& self, const caf::settings& opts) {
   auto str = get_or(opts, "system.endpoint", defaults::system::endpoint);
   if (!parsers::endpoint(str, node_endpoint))
     return make_error(ec::parse_error, "invalid endpoint", str);
+  node_endpoint.port.type(port::tcp);
```
mavam: I wonder if the `endpoint` parser should assume TCP by default.

libvast/src/system/exporter.cpp
```diff
@@ -175,9 +172,10 @@ behavior exporter(stateful_actor<exporter_state>* self, expression expr,
   self->set_exit_handler(
     [=](const exit_msg& msg) {
       VAST_DEBUG(self, "received exit from", msg.source, "with reason:", msg.reason);
-      self->send<message_priority::high>(self->state.index, self->state.id, 0);
-      self->send(self->state.sink, sys_atom::value, delete_atom::value);
-      self->send_exit(self->state.sink, msg.reason);
+      auto& st = self->state;
+      self->send<message_priority::high>(st.index, st.id, 0);
```
mavam: Can you add a comment what this does? I think we "abuse" some of the message handlers at the INDEX where 0 means "abort" or "done". Since this is not clear in this context, it warrants a comment.

libvast/src/system/signal_monitor.cpp
```diff
@@ -36,8 +36,10 @@ std::atomic<bool> signals[32];
 extern "C" void signal_handler(int sig) {
   // Catch termination signals only once to allow forced termination by the OS
   // upon sending the signal a second time.
-  if (sig == SIGINT || sig == SIGTERM)
+  if (sig == SIGINT || sig == SIGTERM) {
+    std::cerr << "\rInitiating shutdown... (repeat request to terminate)\n";
```
mavam: I can't remember that we use upper-case for log output.

mavam:
```suggestion
    std::cerr << "\rInitiating graceful shutdown... (repeat request to terminate immediately)\n";
```

libvast/src/system/connect_to_node.cpp
dominiklohmann: I've discussed this with @tobim at HQ and we've decided that'd be out of scope for this issue. Do you want me to change that?

mavam: I'm in full agreement that it's out of scope. I still wanted it to hear your opinion on it, though.

tobim: > I wonder if the `endpoint` parser should assume TCP by default.

The `endpoint` does not delegate to the `port` parser, but uses `u16` instead.  This is a mismatch with the `endpoint` type, which contains a `vast::port`. I think we should change to `port` parser to make the "/protocol" part optional and delegate to that.
The specific semantic of a CAF endpoint only comes in in this function, it should indeed default to `tcp`, but not overwrite if anything other than `unknown` is set, and manually return an error if something other than `tcp` is requested.

tobim: The `endpoint` parser was actually correct, I missed the alternative path using the `port` parser.

PR #603, opened 2019-10-15
Fix crash in exporter

dominiklohmann:
The exporter crashed for `max_events % table_slice_size == 0`.

Note that there are no integration tests for this crash because the output of `export -n <count>` can be indeterministic.


mavam:
This seems to be a user-facing bugfix. We need a changelog entry for such fixes.

PR #604, opened 2019-10-15
Respect the users ARFLAGS

dominiklohmann:
A follow-up to #601: `CMAKE_STATIC_LINKER_FLAGS` are supposed to be passed to the archiver, the name is wrong for backwards compatibility reasons.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🐞 The user environments `LDFLAGS` were erroneously passed to `ar`. Instead,
+  the user environments `ARFLAGS` are now used.
+
 - 🐞 Exporting data with `export -n <count>` crashed when `count` was a
   multiple of the table slice size. The command now works as expected.
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,6 +150,7 @@ endmacro ()
 
 # Respect environment variables.
 set(LDFLAGS "$ENV{LDFLAGS}")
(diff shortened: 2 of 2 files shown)


mavam:
Changelog entry please :)

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🐞 Building VAST now properly respects the environment variables `LDFLAGS`
+  and `ARFLAGS`, instead of passing the `LDFLAGS` to both `ld` and `ar`.
+
```
mavam: Thanks! Just a quick first-time comment. We try to have the following format for bugfixes:

1. First describe the bug as concise as possible.
2. Then describe the fix.

tobim:
> Changelog entry please :)

I would prefer to restrict changelog entries to user facing changes, with "user" meaning the role. If we put in minor things like this the changelog will get very long, and people will lose interest before reading all the way to the bottom.

mavam:
I see your point. However, having a small amount of bugfixes in there creates a "warm fuzzy feeling" that the product is evolving. As long it's not resulting in distraction and is relevant to the user, it's important to not brush these under the carpet. Here, a crash caused by a certain input strikes me as noteworthy information.

dominiklohmann:
> Here, a crash caused by a certain input strikes me as noteworthy information. 

That's #603. This one is just a fix for people who have custom `LDFLAGS` or `ARFLAGS`.

mavam:
Looks great, thanks!

dominiklohmann:
I have rebased this to fix the conflict on recent master.

PR #611, opened 2019-10-17
Add editorconfig file

tobim:


mavam:
I'm not sure if this is working, but is it possible to move this file to our new `.github` repo so that it's available to all repos at our org?

dominiklohmann:
> I'm not sure if this is working, but is it possible to move this file to our new `.github` repo so that it's available to all repos at our org?

That's for Community Health files only (except license and readme).

mavam:
Okay, I just wonder how we'll keep this one in sync with other repos.

tobim:
> Okay, I just wonder how we'll keep this one in sync with other repos.

This file saves a similar purpose as `.clang-format`. It should not be necessary to clone another repo to set a few editor defaults. Maybe we can introduce a canonical copy in a central scaffold repository from which the others are synced, but that feels very over engineered to me.

tobim:
> Lgtm. We don't have any Makefiles anymore, right?

I removed that section before opening the PR, you must have had an older version in the cache.

PR #613, opened 2019-10-21
Update CAF to 0.17.2

dominiklohmann:


libvast/src/error.cpp
```diff
@@ -11,13 +11,20 @@
...
 
-#include "vast/detail/assert.hpp"
-#include "vast/error.hpp"
+#include <sstream>
+#include <string>
+
+// FIXME: workaround for actor-framework/actor-framework#940
+namespace caf {
+std::string to_string(pec x);
+}
```
mavam:
```suggestion
} // namespace caf
```

Would you mind adding concrete guidance on how to fix this in the future? e.g., "remove this forward declaration after upgrading to 17.3"

libvast/test/range_map.cpp
```diff
@@ -15,6 +15,10 @@
 
 #include "vast/test/test.hpp"
 
+// CAF 0.17.2 added a test DSL macro named `inject`, which conflicts with
+// `range_map::inject`
+#undef inject
+
```
mavam: For a follow-up issue: maybe we should not include the DSL unconditionally in `vast/test/test.hpp` instead? Since it's not part of the core unit test suite, we may want to include only where we actually use the DSL.

libvast/src/error.cpp
dominiklohmann: I've addressed this.

libvast/test/range_map.cpp
dominiklohmann: We don't actually use the DSL anywhere. I've hidden it behind the macro `VAST_USE_TEST_DSL` for now.

PR #614, opened 2019-10-21
Generate manpage content dynamically

dominiklohmann:
This works generally, but there's still lot's of omitted information that was in the man page previously. Not sure if we need both the help text (currently incompatible with roff, hence commented out) and the documentation text in there. Some documentation appears to be missing still.

The new `manual` subcommand / option prints the documentation for every command that's in the tree at or after the current command and adds some Markdown headers. Not sure if this should be publicly documented in the changelog.

I am considering switching from `md2man-roff` to `pandoc` as it's much more sophisticated and robust.


libvast/src/command.cpp
```diff
@@ -196,6 +197,16 @@ void render_parse_error(const command& cmd,
   }
 }
 
+void manualtext(const command& cmd, std::ostream& os,
```
mavam: I find this hard on they eyes when reading. We have `helptext`, which barely works, and I'd suggest `mantext` to keep it simple. Otherwise I'd suggest using `_` for clearer function naming.

mavam: Also, could you a small documentation string that tells the user that we are generating Markdown here?

PR #616, opened 2019-10-22
Generate periodic statistics at client

dominiklohmann:


dominiklohmann:
[story/ch9335]

scripts/vast-du
```diff
@@ -0,0 +1,23 @@
+#! /usr/bin/env sh
```
mavam: I think a shebang of `#!/bin/sh` is also POSIX compliant.

scripts/vast-du
```diff
@@ -0,0 +1,23 @@
...
+
+# Usage examples:
+#   vast-du -N import zeek -r libvast_test/artifacts/logs/zeek/conn.log
+#	vast-du -N export ascii '#type ~ /zeek.*/'
+
+set -e
+
+if ! command -v vast >/dev/null 2>&1; then
+	echo "vast not found"
+	exit
```
mavam:
```suggestion
	exit 1
```

dominiklohmann:
I’ve addressed the feedback. The script isn’t quite up to the usual standard, because it’s just what I used for testing locally and we decided it was worth keeping.

PR #617, opened 2019-10-22
Fix memory leak in the zeek writer

tobim:


integration/default_set.yaml
```diff
@@ -78,6 +78,8 @@ tests:
       - command: -N import zeek
         input: data/zeek/dns.log.gz
       - command: -N export ascii 'resp_h == 192.168.1.104'
+      - command: -N export zeek 'resp_h == 192.168.1.104'
+        transformation: sed '/#\(open\|close\)/d'
```
mavam:
```suggestion
        transformation: awk $1 !~ /#(open|close)/'
```

dominiklohmann:
```suggestion
        transformation: awk '!/#\(open\|close\)/'
```

mavam: Less is more!

PR #621, opened 2019-10-25
Calculate node throughput from accounting logs

dominiklohmann:
Edit: [skip ci] for Cirrus in PR titles apparently does not work with force pushes. 🤷‍♂

### DIFF
--- a/scripts/vast-du
+++ b/scripts/vast-du
@@ -19,5 +19,13 @@ else
 	DU="du"
 fi
 
+# Run vast and filter for rate and latency results
 vast "${@}" 2>&1 >/dev/null | grep "rate\|latency"
+# Query vast for average node throughput rate
+if [[ "$*" == *import* ]]; then
+	NODE_THROUGHPUT="$(vast -N export json '#type == "vast.account" && key == "node_throughput.rate"' 2>/dev/null \
+		| jq -s 'map(.value | tonumber) as $rates | $rates | add / ($rates | length)')"
+	printf "node throughput rate is %.2f events/sec\n" "${NODE_THROUGHPUT}"
+fi
+# Print archive and index disk usage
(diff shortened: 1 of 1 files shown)


scripts/vast-du
```diff
@@ -19,5 +19,13 @@ else
 	DU="du"
 fi
 
+# Run vast and filter for rate and latency results
 vast "${@}" 2>&1 >/dev/null | grep "rate\|latency"
+# Query vast for average node throughput rate
+if [[ "$*" == *import* ]]; then
+	NODE_THROUGHPUT="$(vast -N export json '#type == "vast.account" && key == "node_throughput.rate"' 2>/dev/null \
+		| jq -s 'map(.value | tonumber) as $rates | $rates | add / ($rates | length)')"
```
mavam: This computes the mean, which I don't think is robust statistic because it assumes normality. Can report min, max, mean, and median?

dominiklohmann: It's an estimate currently anyways, because `node_throughput.duration` is not necessarily exactly 1 second. 

Do you think reporting min, max, mean, and median exactly makes sense here?

mavam: I'm not sure if I understand your comment. Why does it matter whether the data is a set of estimates? My comment was more about what we are doing with the set.

If we have a uniform or normal distribution, the mean is surely not wrong, but I don't know the shape of the data and therefore would like to report defensively at first.

dominiklohmann: Should've clarified better. The data is already kind of skewed because not every rate was measured over the same duration, it's 1.05s ± 0.05s in practice.

Obviously I can write a small jq script to calculate the data you request right now. Will do in a bit..

dominiklohmann: This has been addressed, but the median is still obviously skewed because the rates are not normalized with their actual duration.

PR #623, opened 2019-10-25
Support equality operator for pattern matching

mavam:
This PR adds the ability to use operator `==` for patterns on the LHS or RHS. The semantics are the same as the match operator `~`.

### DIFF
--- a/libvast/src/pattern.cpp
+++ b/libvast/src/pattern.cpp
@@ -116,6 +116,22 @@ bool operator<(const pattern& lhs, const pattern& rhs) {
   return lhs.str_ < rhs.str_;
 }
 
+bool operator==(const pattern& lhs, std::string_view rhs) {
+  return lhs.match(rhs);
+}
+
+bool operator!=(const pattern& lhs, std::string_view rhs) {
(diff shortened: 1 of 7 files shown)


libvast/src/pattern.cpp
```diff
@@ -116,6 +116,22 @@ bool operator<(const pattern& lhs, const pattern& rhs) {
...
+bool operator==(const pattern& lhs, std::string_view rhs) {
+  return lhs.match(rhs);
+}
+
+bool operator!=(const pattern& lhs, std::string_view rhs) {
+  return !(lhs == rhs);
+}
+
+bool operator==(std::string_view lhs, const pattern& rhs) {
+  return rhs.match(lhs);
```
dominiklohmann: Can you add a TODO for C++20 to remove the symmetric operators?

For now, do you think this should be expressed as `operator==(rhs, lhs)`?

mavam: I didn't know about this feature in C++20. Great to see that equality is finally symmetric. Until we refactor our code base accordingly, I'd say we leave this as is.

PR #624, opened 2019-10-28
Cleanup includes

tobim:
This is done by manually integrating suggestions from iwyu. This change will also simplify running `iwyu` and `clang-tidy` via CMake by disabling these tools for the CAF and broker submodules.

This change also moves the CAF and broker related CMake code into `aux/CMakeLists.txt`.


libvast/src/table_slice.cpp
```diff
@@ -255,7 +255,7 @@ std::pair<table_slice_ptr, table_slice_ptr> split(const table_slice_ptr& slice,
   // Create second table slice.
   select(xs, slice, make_ids({{mid, last}}));
   VAST_ASSERT(xs.size() == 2);
-  return {std::move(xs.front()), std::move(xs.back())};
+  return {xs.front(), xs.back()};
```
dominiklohmann: Changed by accident?

vast.imp
```diff
@@ -0,0 +1,3 @@
+[
+  { ref: "caf.imp" }
+]
```
dominiklohmann: Added to git by accident?

tobim: This is an _iwyu-mapping-file_, together with the referenced `caf.imp` they contain out-of-band directives for telling the tool which header is to be used as the public source of a definition.

lava:
> This change will also simplify running iwyu and clang-tidy via CMake by disabling these tools for the CAF and broker submodules.

How is this done?

tobim:
> > This change will also simplify running iwyu and clang-tidy via CMake by disabling these tools for the CAF and broker submodules.
> 
> How is this done?

The trick is to set the target properties `CXX_CLANG_TIDY` and `CXX_INCLUDE_WHAT_YOU_USE` to the empty list each. See https://cmake.org/cmake/help/latest/prop_tgt/LANG_INCLUDE_WHAT_YOU_USE.html for details.

tobim:
I'll close this for now, parts of it should be salvageable in more focused PRs later. I'd like to follow through with the changes to `fwd.hpp` at least.

PR #632, opened 2019-11-01
Add optimized implementation for string indexes

mavam:
The new index is meant for opaque identifiers that can only be retrieved via equality lookups. If a string type contains the `#index=hash` attribute, the index factory chooses this new hash-based implementation.


mavam:
I haven't had the time yet, but it would be great to have some numbers on the performance delta. Ideally for both space and time.

mavam:
There's a fundamental issue I overlooked in the design: the index does not support *inequality* lookups. It can't because it hashes the input. We'll have to reconsider the usefulness of a space-efficient equality-only string index before proceeding.

libvast/vast/hash_index.hpp
```diff
@@ -0,0 +1,213 @@
...
+        VAST_ASSERT(seeds_.count(materialized) == 0);
+        seeds_.emplace(std::move(materialized), i);
+        unique_digests_.insert(k);
+        return k;
+      };
+      // If we have seen the digest, check whether we also have a known
+      // preimage.
+      // FIXME: do not materialize every time, but this requires the ability to
+      // use key equivalence (C++20).
+      if (auto it = seeds_.find(materialize(x)); it != seeds_.end())
```
mavam: @dominiklohmann is there a way to have C++20 key equivalence already in C++17? I don't want to materialize a `data_view` every time, but simply search based on the view instance.

dominiklohmann: For C++17, set the template parameter for the unordered map manually. For C++20, the `KeyEqual` template param is set to `Hash::transparent_key_equal` instead iff it is available, with `Hash` being another template parameter.

The default values are this:

```c++
class Hash = std::hash<Key>,
class KeyEqual = std::equal_to<Key>>
```

This means that all you need to do is set a custom equality function here:

```c++
// define custom comparator somewhere
auto comp = [](const data& lhs, const data& rhs) -> bool { ... }

// change declaration of seeds map
std::unordered_map<data, size_t, std::hash<data>, decltype(comp)> seeds_;
```

mavam: I'm not sure how the custom comparator will solve this issue. The map contains `data`, but I'd like `find` to work with `data_view`, without having to materialize the views to a full-blown `data` instance.

dominiklohmann: Oh I see. So your issue really is that `find` is hardcoded to use the same `Key` as the map instead of some arbitrary type that can be equality compared with `KeyEqual` with the `Key` being on the lhs. You can't get that right now.

A dirty hack I can imagine to work is having the Key be a union of `data` and `data_view`, and using `data_view` on the rhs of your custom `KeyEqual` function and `data` everywhere else.

dominiklohmann: `boost::unordered_map` has this:

```c++
template<typename CompatibleKey,
         typename CompatibleHash, 
         typename CompatiblePredicate> 
iterator find(CompatibleKey const&,
              CompatibleHash const&, 
              CompatiblePredicate const&);
```

Maybe we should just copy this over and add a `TODO:` comment for switching back to `std::unordered_map` once we arrive at C++20.

mavam: > Maybe we should just copy this over

Copying over the entire implementation of a boost header? All the dependencies, the MPL, etc. will cause too much bloat for the desired gain. I think we have to figure out a different solution. For example, we could switch to `flat_map` for which we have our own implementation.

dominiklohmann:
```c++
template <typename Key, typename T, typename Hash, typename KeyEqual,
          typename Allocator>
auto find_with_precomputed_hash(
    const std::unordered_map<Key, T, Hash, KeyEqual, Allocator> &data,
    std::invoke_result_t<Hash, Key> precomputed_hash) {
  return std::find_if(data.begin(), data.end(),
                      [precomputed_hash = std::move(precomputed_hash),
                       hash_fn = data.hash_function()](const auto &lhs) {
                        return hash_fn(lhs.first) == precomputed_hash;
                      });
}
```

This should work for your purpose just fine.

dominiklohmann: More general approach that should work for any unordered container in the standard library:

```c++
#include <algorithm>
#include <type_traits>

template <typename PrecomputedHash,
          template <typename...> class UnorderedContainer, typename... Ts>
auto find_with_precomputed_hash(UnorderedContainer<Ts...>& container,
                                PrecomputedHash&& precomputed_hash)
  -> decltype(
    std::declval<
      std::invoke_result_t<typename UnorderedContainer<Ts...>::hasher,
                           typename UnorderedContainer<Ts...>::key_type>>()
      == std::declval<PrecomputedHash>(),
    container.begin()) {
  return std::find_if(container.begin(), container.end(),
                      [precomputed_hash
                       = std::forward<PrecomputedHash>(precomputed_hash),
                       hash_fn = container.hash_function()](auto& lhs) {
                        return hash_fn(lhs.first) == precomputed_hash;
                      });
}
```

I actually think this is okay-ish in terms of readability.

See test at: https://godbolt.org/z/jQwQaU

mavam: Isn't this O(n) due to `find_if`?

dominiklohmann: If you're planning to not have any collisions then the `std::unordered_map<...>::find(...)` should not be any better than O(end-begin) either, right?

libvast/src/value_index_factory.cpp
```diff
@@ -29,17 +29,19 @@ namespace {
...
+        return *max_size;
   return default_value;
 }
 
 template <class T>
 optional<base> parse_base(const T& x) {
-  if (auto a = extract_attribute(x, "base")) {
-    if (auto b = to<base>(*a))
-      return *b;
+  if (auto a = find_attribute(x, "base")) {
```
tobim: Regarding the naming of the function: I would prefer not to do any parsing / input validation this late, meaning we should validate `load_schema()` and use an assert here instead.

Then we can rename this function to `extract_base` to synchronize with `extract_max_size`, and drop the `optional` from the return type.

This should probably be done in another PR :-).

libvast/vast/view.hpp
```diff
@@ -169,6 +174,18 @@ struct view_trait<data> {
   using type = data_view;
 };
 
+// -- operators ----------------------------------------------------------------
+
+// FIXME: we cannot use operator == and != here because data has a non-explicit
+// constructor, which results in error all over the code base. Therefore, we
+// work around this by giving this function a name.
```
dominiklohmann: Do _you_ intend to fix this? Otherwise this should be a `TODO` and an associated story in ClubHouse so we don't forget about it.

libvast/vast/hash_index.hpp
```diff
@@ -0,0 +1,256 @@
...
+/// bytes for a more space-efficient representation, at the cost of more false
+/// positives. A separate "satellite structure" keeps track of hash collision
+/// to make the index exact. The additional state to build this satellite
+/// structure only exists during the construction of the index. Upon
+/// descruction, this extra state ceases to exist and it will not be possible
+/// to append further values when deserializing an existing index.
+template <size_t Bytes>
+class hash_index : public value_index {
+  static_assert(Bytes > 0);
+  static_assert(Bytes <= 8);
```
dominiklohmann:
```suggestion
  static_assert(Bytes <= sizeof(decltype(seeds_)::value_type);
```

Not a fan of magic numbers, especially since the size of a `size_t` may vary between platforms.

This additionally requires the assertion to be moved below the declaration of `seeds_`. Alternatively, you can always add a `using-declaration` above for the `size_t` that gives it a proper name.

libvast/vast/hash_index.hpp
```diff
@@ -0,0 +1,256 @@
...
+  auto find_seed(data_view x) const {
+    auto pred = [&](const auto& xs) { return is_equal(xs.first, x); };
+    auto& xs = as_vector(seeds_);
+    return std::find_if(xs.begin(), xs.end(), pred);
+  }
+
+  // Retrieves the unique digest for a given input or generates a new one.
+  caf::optional<key> make_digest(data_view x) {
+    size_t i = 0;
+    do {
```
dominiklohmann: Sometimes, a raw for loop might be the correct way to do things. 

```cpp
for (size_t i = 0; i < max_hash_rounds; ++i)
```

This seems to express the retry-n-times intent much clearer than the do-while loop.

libvast/vast/hash_index.hpp
```diff
@@ -0,0 +1,256 @@
...
+      // We're good to go with: create the set predicates an run the scan.
+      auto in_pred = [&](const digest_type& digest) {
+        auto cmp = [=](auto& k) { return k == digest; };
+        return std::any_of(keys->begin(), keys->end(), cmp);
+      };
+      auto not_in_pred = [&](const digest_type& digest) {
+        auto cmp = [=](auto& k) { return k == digest; };
+        return std::none_of(keys->begin(), keys->end(), cmp);
+      };
+      return op == in ? scan(in_pred) : scan(not_in_pred);
```
dominiklohmann: I'm not convinced that the optimizer can eliminate the unused lambda instantiation here, since it is capturing variables and `scan` takes the predicate by-value since it needs to be called multiple times.

Instead, I'd go with this:

```cpp
auto is_digest = [=](const auto& key) {
  return key == digest;
};
if (op == in)
  scan([&](const digest_type& digest) {
    return std::any_of(keys->begin(), keys->end(), is_digest);
  });
else
  scan([&](const digest_type& digest) {
    return std::none_of(keys->begin(), keys->end(), is_digest);
  }); 
```

There is a similar case just above.

libvast/vast/hash_index.hpp
```diff
@@ -0,0 +1,256 @@
...
+
+  bool immutable() const {
+    return unique_digests_.empty() && !digests_.empty();
+  }
+
+  std::vector<digest_type> digests_;
+  std::unordered_set<key, key_hasher> unique_digests_;
+
+  // TODO: Once we can use C++20 and have a standard library that implements
+  // key equivalcne properly, we can switch to std::unordered_map.
```
dominiklohmann:
```suggestion
  // key equivalence properly, we can switch to std::unordered_map.
```

libvast/src/value_index_factory.cpp
mavam: Fair point, attributes are currently arbitrary string key-value pairs in the schema. Our parser doesn't restrict a type or format for a specific attribute, e.g., `#base`, `#base=foo`, and `#base=42` will parse just fine. Until we tighten this restriction, we have to do parsing late in the game.

Down the line, I would like to see two things:

1. Let attributes map from a string key to `data` instead of to `std::string`.
2. For each attribute we support, decide at parse time which type is supported/expected.

If that sounds reasonable to you, then we should open another story for this.

libvast/vast/value_index.hpp
```diff
@@ -88,15 +88,19 @@ class value_index {
...
 private:
   virtual bool append_impl(data_view x, id pos) = 0;
 
   virtual caf::expected<ids>
   lookup_impl(relational_operator op, data_view x) const = 0;
 
-  ewah_bitmap mask_;
-  ewah_bitmap none_;
-  const vast::type type_;
+  ewah_bitmap mask_;      ///< The position of all values, including nil.
```
tobim: Because of `value_index.cpp#32` this comment is seems to be incorrect.

libvast/src/value_index_factory.cpp
tobim: Sounds good to both.

libvast/src/value_index_factory.cpp
```diff
@@ -13,33 +13,38 @@
...
 
-#include <caf/optional.hpp>
-
 #include "vast/attribute.hpp"
 #include "vast/base.hpp"
 #include "vast/concept/parseable/numeric/integral.hpp"
 #include "vast/concept/parseable/to.hpp"
 #include "vast/concept/parseable/vast/base.hpp"
 #include "vast/detail/type_traits.hpp"
+#include "vast/hash_index.hpp"
```
tobim: The include of `vast/attribute.hpp` can be removed.

libvast/src/view.cpp
```diff
@@ -53,6 +54,40 @@ bool operator<(pattern_view x, pattern_view y) noexcept {
...
 }
 
+bool is_equal(const data& x, const data_view& y) {
+  auto pred
+    = [](const auto& lhs, const auto& rhs) { return is_equal(lhs, rhs); };
+  auto f = detail::overload(
+    [&](const auto& lhs, const auto& rhs) {
+      using lhs_type = std::decay_t<decltype(lhs)>;
+      using rhs_type = std::decay_t<decltype(rhs)>;
+      if constexpr (std::is_same_v<view<lhs_type>, rhs_type>)
```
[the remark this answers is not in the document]
mavam: Without decaying, I get a `const T&` as type, which is not compatible with `view<T>`. Hence the decaying.

Unfortunately `is_equality_comparable` doesn't work, because `data` has a non-explicit constructor. (See below for details.)

libvast/vast/hash_index.hpp
mavam: I should have commented the magic number. Here, I thought that it doesn't make sense keep digests larger than 8 bytes because they exceed the length of a 64-bit integer.

I'm relying on this in `key_hasher`. Probably adding a comment here is what would help?!

libvast/src/value_index_factory.cpp
```diff
@@ -73,6 +78,21 @@ auto add_arithmetic_index_factory() {
   return factory<value_index>::add(T{}, make_arithmetic<T>);
 }
 
+auto add_string_index_factory() {
+  static auto f = [](type x) -> value_index_ptr {
+    if (auto a = find_attribute(x, "index"))
+      if (auto value = a->value)
+        if (*value == "hash"sv)
```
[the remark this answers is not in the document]
mavam: Makes sense. We'll address this in a separate PR.

libvast/vast/view.hpp
mavam: Demoted to `TODO`.

libvast/vast/hash_index.hpp
mavam: It's one of those instances where I went for easier readability at the negligible (potential!) performance pessimization.

Your solution doesn't work because `digest` is the argument of the predicate that gets captured.

mavam: This would work, but I find it much less readable:
```diff
diff --git a/libvast/vast/hash_index.hpp b/libvast/vast/hash_index.hpp
index 8fb603d9..1e7a46cc 100644
--- a/libvast/vast/hash_index.hpp
+++ b/libvast/vast/hash_index.hpp
@@ -225,15 +225,19 @@ private:
       if (!keys)
         return keys.error();
       // We're good to go with: create the set predicates an run the scan.
-      auto in_pred = [&](const digest_type& digest) {
-        auto cmp = [=](auto& k) { return k == digest; };
-        return std::any_of(keys->begin(), keys->end(), cmp);
-      };
+      if (op == in) {
+        auto in_pred = [&](const digest_type& digest) {
+          auto cmp = [=](auto& k) { return k == digest; };
+          return std::any_of(keys->begin(), keys->end(), cmp);
+        };
+        return scan(in_pred);
+      }
+      VAST_ASSERT(op == not_in);
       auto not_in_pred = [&](const digest_type& digest) {
         auto cmp = [=](auto& k) { return k == digest; };
         return std::none_of(keys->begin(), keys->end(), cmp);
       };
-      return op == in ? scan(in_pred) : scan(not_in_pred);
+      return scan(not_in_pred);
     }
     return make_error(ec::unsupported_operator, op);
   }
```

libvast/src/view.cpp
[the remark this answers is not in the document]
dominiklohmann: I see. I suppose this is fine then.

libvast/vast/hash_index.hpp
dominiklohmann: > Here, I thought that it doesn't make sense keep digests larger than 8 bytes because they exceed the length of a 64-bit integer.

Not only that, but if `size_t` is less than 64-bit long on your platform, you start to get other issues. I'd make it `sizeof(size_t)` in both places, and maybe swap out `size_t` for `uint64_t` just to be sure.

dominiklohmann:
The commit ede727e seems to have triggered an ADL correctness issue, causing the wrong operator to be called and thus the wrong type name to be displayed. That should likely be fixed in a separate PR.

mavam:
@dominiklohmann Sorry, I fixed the issue already before seeing your request to do this separately.

libvast/vast/hash_index.hpp
mavam: I've added more documentation. I think `size_t` is the right type for the `Bytes` parameter. It will never carry a hash and also serve in `std::array<T, Bytes>`. Hopefully the new documentation makes this clearer.

dominiklohmann: It's not about the `size_t` for the template parameter `Bytes`, but rather about the value type  that you store the digests in.

Anyways, the comment does the job just as fine as the `sizeof`.

libvast/vast/hash_index.hpp
dominiklohmann: This also works and is even more readable:

```diff
diff --git a/libvast/vast/hash_index.hpp b/libvast/vast/hash_index.hpp
index fd5305e3..41c5a95c 100644
--- a/libvast/vast/hash_index.hpp
+++ b/libvast/vast/hash_index.hpp
@@ -233,15 +233,12 @@ private:
       if (!keys)
         return keys.error();
       // We're good to go with: create the set predicates an run the scan.
-      auto in_pred = [&](const digest_type& digest) {
+      return scan([&](const digest_type& digest) {
         auto cmp = [=](auto& k) { return k == digest; };
-        return std::any_of(keys->begin(), keys->end(), cmp);
-      };
-      auto not_in_pred = [&](const digest_type& digest) {
-        auto cmp = [=](auto& k) { return k == digest; };
-        return std::none_of(keys->begin(), keys->end(), cmp);
-      };
-      return op == in ? scan(in_pred) : scan(not_in_pred);
+        return op == in 
+          ?  std::any_of(keys->begin(), keys->end(), cmp)
+          : std::none_of(keys->begin(), keys->end(), cmp);
+      });
     }
     return make_error(ec::unsupported_operator, op);
   }
```

libvast/vast/hash_index.hpp
mavam: Ah, I missed that. Thanks for clarifying.

libvast/vast/hash_index.hpp
mavam: While it's certainly correct and concise, I'm not sure it it's a performance issues: the predicate we are passing to `scan` gets invoked for every element in the index. The branch predictor will certainly make sure cache misses are minimal, but the extra branching is easy to avoid.

PR #633, opened 2019-11-04
Add support for Apache Arrow

dominiklohmann:
This PR adds support for `export arrow`. The `import arrow` counterpart will follow in a later PR.


dominiklohmann:
After some back and forth with CI, we've decided that it's not worth the trouble to get Arrow installed for Debian/FreeBSD and to just use macOS for continuously testing Arrow support. 

This PR now depends on #639, because Arrow installs Snappy, which is improperly linked on CI.

mavam:
FreeBSD integration tests are still failing due to a missing NumPy dependency:
```
CMake Error at cmake_modules/FindNumPy.cmake:62 (message):
    NumPy import failure:
  
    Traceback (most recent call last):
  
      File "<string>", line 1, in <module>
  
    ModuleNotFoundError: No module named 'numpy'
```

dominiklohmann:
I rebased onto master to fix the conflict in CHANGELOG.md.

doc/cli/vast-export-arrow.md
```diff
@@ -0,0 +1,44 @@
+The Arrow export format renders events in [Apache
+Arrow](https://arrow.apache.org), a development platform for in-memory data
+with bindings for many different programming languages.
+
+For example, the below Python program reads Arrow-formatted data from stdin and
+prints it back in a readable format batch by batch.
+
+```python
```
tobim: Do we have any way to ensure that this doc does not get out of sync with the script in the integration subfolder?

CMakeLists.txt
```diff
@@ -610,6 +610,22 @@ if (GPERFTOOLS_FOUND)
   endif ()
 endif ()
 
+if (NOT ARROW_ROOT_DIR AND VAST_PREFIX)
+    set(ARROW_ROOT_DIR ${VAST_PREFIX})
+endif ()
+find_package(arrow QUIET)
```
tobim: We should make this an opt-out dependency with a `VAST_NO_ARROW` option to disable it explicitly.

doc/cli/vast-export-arrow.md
dominiklohmann: None currently. I'll add a note to the top of the script to update this file as well.

CMakeLists.txt
dominiklohmann: Implemented as suggested.

dominiklohmann:
I've addressed the review feedback given by @tobim, and will add another set of changes to CMake in a bit.

configure
```diff
@@ -138,6 +138,9 @@ while [ $# -ne 0 ]; do
     --disable-exceptions)
       append_cache_entry NO_EXCEPTIONS BOOL true
       ;;
+    --disable-arrow=*)
```
mavam:
```suggestion
    --disable-arrow)
```

dominiklohmann: I had this changed locally already and forgot to add it to the index. It's fixed now. 🙄

CMakeLists.txt
```diff
@@ -610,19 +611,21 @@ if (GPERFTOOLS_FOUND)
...
-    provide_find_module(plasma)
-    string(APPEND VAST_FIND_DEPENDENCY_LIST
-           "\nfind_package(arrow REQUIRED QUIET)"
-           "\nfind_package(plasma REQUIRED QUIET)")
+if (NOT VAST_NO_ARROW)
+  if (NOT ARROW_ROOT_DIR AND VAST_PREFIX)
+      set(ARROW_ROOT_DIR ${VAST_PREFIX})
+  endif ()
+  find_package(arrow QUIET)
+  find_package(plasma QUIET)
```
tobim: Should these be `REQUIRED` now?

CMakeLists.txt
```diff
@@ -610,19 +611,21 @@ if (GPERFTOOLS_FOUND)
...
-  set(VAST_HAVE_ARROW true)
-  if (NOT BUILD_SHARED_LIBS)
-    provide_find_module(arrow)
-    provide_find_module(plasma)
-    string(APPEND VAST_FIND_DEPENDENCY_LIST
-           "\nfind_package(arrow REQUIRED QUIET)"
-           "\nfind_package(plasma REQUIRED QUIET)")
+if (NOT VAST_NO_ARROW)
+  if (NOT ARROW_ROOT_DIR AND VAST_PREFIX)
+      set(ARROW_ROOT_DIR ${VAST_PREFIX})
```
tobim: nit: Indentation

dominiklohmann:
I added support for statically linking Arrow.

This required Arrow to be built without jemalloc. For Homebrew, this patch needs to be applied:

```diff
diff --git a/Formula/apache-arrow.rb b/Formula/apache-arrow.rb
index 571d4f3e48..5082bddd7e 100644
--- a/Formula/apache-arrow.rb
+++ b/Formula/apache-arrow.rb
@@ -40,6 +40,7 @@ class ApacheArrow < Formula
       -DARROW_PROTOBUF_USE_SHARED=ON
       -DARROW_PYTHON=ON
       -DARROW_INSTALL_NAME_RPATH=OFF
+      -DARROW_JEMALLOC=OFF
       -DPYTHON_EXECUTABLE=#{Formula["python"].bin/"python3"}
     ]
 
```

tobim:
I just rediscovered that arrow is not packaged for FreeBSD. We don't need to go down that rabbit hole.

dominiklohmann:
I've created a PR for "fixing" the Homebrew formula: Homebrew/homebrew-core#46691

dominiklohmann:
Alright, so this is basically up for review again. Only thing left to discuss is whether we want to instead use Arrow as a submodule, since it's not available on FreeBSD as a package and we've had some trouble linking statically on macOS.

mavam:
> Only thing left to discuss is whether we want to instead use Arrow as a submodule, since it's not available on FreeBSD as a package and we've had some trouble linking statically on macOS.

Given that we have CAF as submodule as well, why don't we do the same with Arrow? Given the frequent API changes of Arrow and the not-so-widespread package support, I would prefer to closely track the revision that we support. This works best by adding Arrow as submodule.

PR #634, opened 2019-11-04
Adjust Cirrus Build-Pipeline 

0ortmann:
- Use `rsync` over `scp` for artifact upload
- Reduce number of CPU cores used during artifact build.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -40,8 +40,8 @@ style_task:
   integration_test_script: >
     if ! cmake --build build --target integration; then
       tar -czf $(cat ARTIFACT_NAME).tar.gz -C build vast-integration-test
-      scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i upload_key \
-        $(cat ARTIFACT_NAME).tar.gz mule@tenzir.dfn-cert.de:integration/vast/
+      rsync --archive --compress --rsh "ssh -i upload_key -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null" \
+        --progress $(cat ARTIFACT_NAME).tar.gz mule@tenzir.dfn-cert.de:/home/mule/integration/vast/
       echo "Download with: \`scp tenzir.dfn-cert.de:/home/mule/integration/vast/$(cat ARTIFACT_NAME).tar.gz .\`"
       false
     fi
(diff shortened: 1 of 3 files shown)


.cirrus.yml
```diff
@@ -51,14 +51,15 @@ style_task:
     - cmake --build build --target package
   upload_script: >
     if [ "$CIRRUS_BRANCH" = "master" ]; then
-      scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i upload_key build/$(cat PACKAGE_NAME).tar.gz mule@tenzir.dfn-cert.de:artifacts/
+      rsync --archive --compress --rsh "ssh -i upload_key -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null" \
+        --progress build/$(cat PACKAGE_NAME).tar.gz mule@tenzir.dfn-cert.de:/home/mule/artifacts/
```
dominiklohmann:
```suggestion
        --progress "build/$(cat PACKAGE_NAME).tar.gz" mule@tenzir.dfn-cert.de:/home/mule/artifacts/
```

Consider quoting this to avoid an error if the result of `cat PACKAGE_NAME` contains a space—it currently does if a tag contains a space, and we don't want to rely on that. The same reasoning applies above as well.

I recommend installing [shellcheck](https://www.shellcheck.net) so this gets caught automatically by a linter.

.cirrus.yml
```diff
@@ -40,8 +40,8 @@ style_task:
   integration_test_script: >
     if ! cmake --build build --target integration; then
       tar -czf $(cat ARTIFACT_NAME).tar.gz -C build vast-integration-test
-      scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i upload_key \
-        $(cat ARTIFACT_NAME).tar.gz mule@tenzir.dfn-cert.de:integration/vast/
+      rsync --archive --compress --rsh "ssh -i upload_key -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null" \
+        --progress $(cat ARTIFACT_NAME).tar.gz mule@tenzir.dfn-cert.de:/home/mule/integration/vast/
```
dominiklohmann:
```suggestion
        --progress "$(cat ARTIFACT_NAME).tar.gz" mule@tenzir.dfn-cert.de:/home/mule/integration/vast/
```

0ortmann: @dominiklohmann thanks for pointing this out. We should then, to my mind, change all variable resolutions to use double quotes instead.

.cirrus.yml
```diff
@@ -34,21 +34,21 @@ style_task:
       -DCMAKE_BUILD_TYPE="$BUILD_TYPE" -DCPACK_GENERATOR=TGZ
       -DCPACK_PACKAGE_FILE_NAME="$(cat PACKAGE_NAME)"
       -DCPACK_SET_DESTDIR=ON "$CONFIGURE_FLAGS" -G Ninja
-    - cmake --build build --target all
+    - cmake --build build --target all --parallel
   unit_test_script:
-    - cmake --build build --target test
+    - cmake --build build --target test --parallel
```
tobim: The targets `test`, `integration`, `install` and `package` do not benefit from the `--parallel` option.

tobim:
Maybe we can go back up to 8 cores for build parallelism?

0ortmann:
Let's try, @tobim

PR #636, opened 2019-11-06
Remove pcap_reader_command in favor of reader_command

dominiklohmann:
This change removes `pcap_reader_command` in favor of `reader_command` and fixes `import pcap` not working with `--read`.


mavam:
Net negative commits FTW! 🥇

PR #643, opened 2019-11-08
Document new releasing workflow

0ortmann:
Update the documentation of the release workflow.

At the same time, this is a proposal for how we could format and formulate all our documentation / guides in the main repo.


0ortmann:
Merging this despite the known issue of failing Debian integration tests on Cirrus.

PR #647, opened 2019-11-14
Refactor importer initialization

tobim:
The IMPORTER used to be set up with a 2-step initialization process. Some necessary parameters were passed at the call to `spawn` and others are supplied by message from the TRACKER. This PR moves all essential parameters to the spawning invocation.

Additionally, this PR factors the logic of retrieving NODE components into the dedicated functions `get_component(node_actor*)` and `get_node_component(scoped_actor&, actor)` to further decouple the TRACKER from the other actors (it is an implementation detail of the NODE and should not leak through the interface).

TODO:
- [x] Use `get_node_component()` in `source_command`.

### DIFF
--- a/libvast/src/system/sink_command.cpp
+++ b/libvast/src/system/sink_command.cpp
@@ -26,6 +26,7 @@
 #include "vast/system/query_status.hpp"
 #include "vast/system/read_query.hpp"
 #include "vast/system/signal_monitor.hpp"
+#include "vast/system/spawn_at_node.hpp"
 #include "vast/system/spawn_or_connect_to_node.hpp"
 #include "vast/system/tracker.hpp"
 
@@ -48,6 +49,10 @@ namespace vast::system {
 
 caf::message sink_command(const command::invocation& invocation,
                           actor_system& sys, caf::actor snk) {
+  // Get a convenient and blocking way to interact with actors.
+  scoped_actor self{sys};
+  auto guard = caf::detail::make_scope_guard(
+    [&] { self->send_exit(snk, exit_reason::user_shutdown); });
   // Read query from input file, STDIN or CLI arguments.
   auto query = read_query(invocation, "export.read");
   if (!query)
@@ -67,8 +72,6 @@ caf::message sink_command(const command::invocation& invocation,
     *query = to_string(ast);
     VAST_DEBUG(&invocation, "transformed expression to", *query);
   }
-  // Get a convenient and blocking way to interact with actors.
-  scoped_actor self{sys};
   // Get VAST node.
   auto node_opt
     = spawn_or_connect_to_node(self, invocation.options, content(sys.config()));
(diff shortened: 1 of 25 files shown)


libvast/src/system/spawn_consensus.cpp
```diff
@@ -46,22 +48,27 @@ maybe_actor spawn_consensus_raft(caf::local_actor* self, spawn_arguments& args)
       anon_send_exit(consensus, caf::exit_reason::user_shutdown);
     }
   );
-  return caf::actor_cast<caf::actor>(s);
+  return s;
```
dominiklohmann: Can you rename this variable to something other than just `s`, maybe `store`?

libvast/vast/system/importer.hpp
```diff
@@ -154,7 +154,7 @@ using importer_actor = caf::stateful_actor<importer_state>;
 /// @param self The actor handle.
 /// @param dir The directory for persistent state.
 /// @param batch_size The initial number of IDs to request when replenishing.
-caf::behavior importer(importer_actor* self, path dir,
+caf::behavior importer(importer_actor* self, path dir, consensus_type consensus,
```
dominiklohmann: You're missing a Doxygen comment for the new parameter here.

libvast/src/system/sink_command.cpp
```diff
@@ -88,26 +88,19 @@ caf::message sink_command(const command::invocation& invocation,
...
-        auto er = reg.components[id].find("accountant");
-        if (er != reg.components[id].end()) {
-          auto accountant = er->second.actor;
-          self->send(snk, actor_cast<accountant_type>(accountant));
-        }
-      },
-      [&](error& e) { err = std::move(e); });
-  if (err) {
-    return caf::make_message(std::move(err));
+  {
```
dominiklohmann: Can you add a comment on why this explicit scope is necessary (or, if it is not, remove it)?

libvast/vast/system/node_control.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include <array>
+
+namespace vast::system {
+
+template <typename... Ts>
+caf::expected<caf::actor>
+spawn_at_node(caf::scoped_actor& self, caf::actor node, Ts&&... xs) {
+  caf::expected<caf::actor> result = caf::no_error;
+  self->request(node, caf::infinite, xs...)
```
dominiklohmann:
```suggestion
  self->request(node, caf::infinite, std::forward<Ts>(xs)...)
```

libvast/vast/detail/array.hpp
```diff
@@ -0,0 +1,35 @@
...
+template <typename T, std::size_t... Is>
+constexpr std::array<T, sizeof...(Is)>
+make_array(const T& value, std::index_sequence<Is...>) {
+  return {{(static_cast<void>(Is), value)...}};
+}
+
+} // namespace
+
+template <std::size_t N, typename T>
+constexpr std::array<T, N> make_array(const T& value) {
```
dominiklohmann: I'm not a fan of naming this `make_array` when it doesn't behave like the `std::make_array` that was voted into C++20.

libvast/vast/system/node_control.hpp
```diff
@@ -0,0 +1,57 @@
...
+    .receive(
+      [&](const std::string& id, system::registry& reg) {
+        auto populate = [&](auto atom) -> caf::expected<caf::actor> {
+          auto er = reg.components[id].find(to_string(atom));
+          if (er == reg.components[id].end())
+            return make_error(ec::missing_component, to_string(atom));
+          return er->second.actor;
+        };
+        size_t i = 0;
+        (..., void((*result)[i++] = populate(Atoms{})));
```
dominiklohmann: I had to stop and think about this `get_node_component` for a bit. I think it deserves a comment, and maybe `populate` should be renamed to `try_find_registered_actor` or something like that.

dominiklohmann:
Oh, one more thing I forgot to mention during the review: References to elements of structured bindings are not explicitly allowed by the C++17 standard (fixed for C++20, I believe). They happen to work with the gcc version used by CI, but fail to compile with clang8, which is why the CI build is failing on FreeBSD and macOS.

libvast/vast/detail/array.hpp
tobim: I think I'll change the implementation to make it work like the `std::make_array`, that should be able to serve the same purpose and more.

tobim:
> ...  They happen to work with the gcc version used by CI, but fail to compile with clang8, which is why the CI build is failing on FreeBSD and macOS.

I changed to code to not make use of structured bindings in `source_command.cpp`. This fixed the clang issue.

libvast/vast/detail/array.hpp
tobim: Turns out `std::make_array` works differently and can't be used in this case. I renamed to `generate_array`.

PR #648, opened 2019-11-14
Add continuous security analysis

0ortmann:
LGTM integrates with github checks (see also this PR's checks). However, so far it does not analyse C++. That is because we need a custom extraction step, which is just introduced with this branch - and hence not enabled on the main repository.
This PR will enable C++ analysis on master.

### DIFF
--- a/lgtm.yaml
+++ b/lgtm.yaml
@@ -1,6 +1,13 @@
 path_classifiers:
+  docs:
+    - doc
+    - exclude: doc/cli
   test:
     - libvast_test
+    - integration
+    - libvast/test
+  template:
+    - scripts
 extraction:
   cpp:
     prepare:
--- a/lgtm.yaml
+++ b/lgtm.yaml
@@ -18,12 +18,12 @@ extraction:
     configure:
       command:
         - pip3 install cmake
-        - /opt/work/.local/lib/python3.6/site-packages/cmake/data/bin/cmake --version
+        - $(python3 -m site --user-site)/cmake/data/bin/cmake --version
         - git submodule update --init aux/caf
         - git submodule update --init aux/broker/broker
         - git -C aux/broker/broker submodule update --init 3rdparty
-        - /opt/work/.local/lib/python3.6/site-packages/cmake/data/bin/cmake -Bbuild -H. -DCMAKE_INSTALL_PREFIX="$PREFIX"
+        - $(python3 -m site --user-site)/cmake/data/bin/cmake -Bbuild -H. -DCMAKE_INSTALL_PREFIX="$PREFIX"
           -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G Ninja
     index:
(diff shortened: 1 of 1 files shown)


lgtm.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - ln -s /usr/bin/g++-8 $LGTM_WORKSPACE/latest-gcc-symlinks/g++
+      - ln -s /usr/bin/gcc-8 $LGTM_WORKSPACE/latest-gcc-symlinks/gcc
+      - export BUILD_TYPE=Release
+      - export GIT=true
+      - export PATH=$LGTM_WORKSPACE/latest-gcc-symlinks:$PATH
+      - export PREFIX=/opt/tenzir
+    configure:
+      command:
+        - pip3 install cmake
+        - /opt/work/.local/lib/python3.6/site-packages/cmake/data/bin/cmake --version
```
mavam: Those explicit paths seem to be brittle when it comes to version changes. Is there perhaps a cleaner approach based on virtual envs? (This assumes that the environment can be mutated between individual steps.)

0ortmann: @mavam We could resolve the site package location dynamically, such as `$(python -m site --user-site)/cmake/data/bin/cmake --version`

0ortmann:
LGTM integrates with github checks (see also this PR's checks). However, so far it does not analyse C++. That is because we need a custom extraction step, which is just introduced with this branch - and hence not enabled on the main repository.
This PR will enable C++ analysis on master.

0ortmann:
This is the direct link to a successful LGTM run with C++ extraction
https://lgtm.com/logs/3a95781801eace2193259800a06bdf29478085ac/lang:cpp

lgtm.yaml
mavam: Seems better, albeit still noisy.

lgtm.yaml
```diff
@@ -0,0 +1,36 @@
...
+  cpp:
+    prepare:
+      packages:
+        - libssl-dev
+        - python-pip
+    after_prepare:
+      - mkdir -p $LGTM_WORKSPACE/latest-gcc-symlinks
+      - ln -s /usr/bin/g++-8 $LGTM_WORKSPACE/latest-gcc-symlinks/g++
+      - ln -s /usr/bin/gcc-8 $LGTM_WORKSPACE/latest-gcc-symlinks/gcc
+      - export BUILD_TYPE=Release
```
dominiklohmann: Should this be Debug so it looks at all the codepaths that are hidden behind `#if VAST_LOG_LEVEL >= VAST_LOG_LEVEL_DEBUG` as well?

0ortmann: As discussed with the team, we'll go with `Release` for now, as there is no important logic to be conditionally compiled in `Debug` builds. `Debug` affects mostly logging behavior.

PR #649, opened 2019-11-16
Rename lgtm.yaml to lgtm.yml

0ortmann:
LGTM did not pick up our custom `lgtm.yaml` file. That appears to be just because of a naming mismatch.


mavam:
This PR already includes the LGTM check, but it says "LGTM analysis: Python". Don't we want C++ here? Or is that the issue that this PR resolves?

0ortmann:
Yup exactly, currently LGTM behaves as if we had no custom config, and hence the C++ build fails. Only the python build, which works without further configuration, succeeds.
Their documentation describes adding an `lgtm.yml` file -- whereas ours is suffixed `.yaml`. https://lgtm.com/help/lgtm/configuring-lgtm-analysis-project

mavam:
Wow, that's brittle...and hard to catch.

0ortmann:
It does not seem to be a problem with the config file itself (i.e. no formatting issues or likewise). See this link for the evaluation / test-config run: https://lgtm.com/logs/d740bfbf61e729b69369dbc66e1cc9dc3e3e0a0b/lang:cpp

mavam:
[ch8187]

mavam:
One question about the link you posted:

Screenshot 2019-11-16 at 13 00 27

There are "failing" checks for C#, Java, and JavaScript. Would it be possible to restrict the analysis to Python and C++ so that these failed states won't occur in the first place? Or does LGTM attempt to do everything it possible can until it fails?

0ortmann:
That is only visible in the test-configuration tab. It appears we cannot disable those analysis attempts, however they are not used to determine the projects' score.
Screenshot_2019-11-16_15-12-15

PR #650, opened 2019-11-16
Ignore VLAN tags in PCAP import

mavam:
This PR fixes failing PCAP import when the traces contain VLAN tags.

For now, we just strip the tags. We might expose them as packet data in the future.


mavam:
The only question I have is: should this be a bugfix or a feature? 🙂

dominiklohmann:
I'd say this is a bugfix.

PR #654, opened 2019-11-18
Fix pointless checks

mavam:
These were a leftover from the time when span still used a signed size type.


mavam:
This is how LGTM displays the changes: https://lgtm.com/projects/g/tenzir/vast/rev/pr-d58c2575edd32950295d3146d3a4691f36ad9078

PR #663, opened 2019-11-25
Fix CI / build-local script for macOS

0ortmann:
The macOS `build-local` script did not work as intended, due to the usage of the wrong version of `readlink`. 
This PR ensures that `greadlink` from the `coreutils` package is used, if installed.

### DIFF
--- a/.ci/freeBSD/build-local
+++ b/.ci/freeBSD/build-local
@@ -3,6 +3,7 @@
 
 # A script to builds the entire VAST ci pipeline locally on freeBSD. That includes an install, so be careful.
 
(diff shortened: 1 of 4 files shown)


dominiklohmann:
FWIW this failed on my setup because the CI scripts change my `CXX` but don't adjust my `CPPFLAGS` (preprocessor flags) / `CXXFLAGS` (c++ flags) / `LDFLAGS` (linker flags) / `ARFLAGS` (archiver flags) accordingly. I had to unset both `CPPFLAGS` and `CXXFLAGS` for this script to work with my setup.

PR #664, opened 2019-11-25
Install manpage only when built

dominiklohmann:
When the manpage target was available, but not explicitly built, the install target would fail because it tried to always install the manpage. This change fixes this oversight.

Should we install `md2man-roff` in CI and build the manpage using `cmake --build … --target man` right before the targets package or install?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -685,8 +685,8 @@ endif ()
 add_subdirectory(tools)
 add_subdirectory(vast)
 add_subdirectory(doc)
-if (NOT vast_is_subproject AND TARGET vast-man-make-vast.1)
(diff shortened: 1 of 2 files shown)


mavam:
> Should we install md2man-roff in CI and build the manpage using cmake --build … --target man right before the targets package or install?

I would say right before `install`. Everybody who installs the software should technically get the man page, not just those who create a package.

tobim:
> > Should we install md2man-roff in CI and build the manpage using cmake --build … --target man right before the targets package or install?
> 
> I would say right before `install`. Everybody who installs the software should technically get the man page, not just those who create a package.

In my opinion we should build it right after the `all` target. This should be done in another PR though.

PR #665, opened 2019-11-26
Add bit_cast helper function

dominiklohmann:
This is a non-constexpr clone of the std::bit_cast that ships with C++20, which is needed for properly punning types (e.g., instead of using reinterpret_cast to convert a stream buffer to the desired record type).

### DIFF
--- a/libvast/vast/detail/bit_cast.hpp
+++ b/libvast/vast/detail/bit_cast.hpp
@@ -1,0 +1,40 @@
+/******************************************************************************
+ *                    _   _____   __________                                  *
+ *                   | | / / _ | / __/_  __/     Visibility                   *
+ *                   | |/ / __ |_\ \  / /          Across                     *
+ *                   |___/_/ |_/___/ /_/       Space and Time                 *
+ *                                                                            *
+ * This file is part of VAST. It is subject to the license terms in the       *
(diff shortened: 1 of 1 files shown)


dominiklohmann:
I have improved the wording of the comment slightly. For more on how this works, see https://en.cppreference.com/w/cpp/numeric/bit_cast

libvast/vast/detail/bit_cast.hpp
```diff
@@ -0,0 +1,40 @@
...
+/// are unspecified.
+/// TODO: Remove this when we have C++20, which ships with a compiler magic
+///       version of this with constexpr support.
+template <
+  typename To, typename From,
+  typename = std::enable_if_t<
+    (sizeof(To) == sizeof(From))
+    && std::is_trivially_copyable_v<From> && std::is_trivial_v<To>>>
+To bit_cast(const From& src) noexcept {
+  To dst;
```
tobim: The simplified sample implementation of the @jfbastien uses [`remove_const_t` here](https://github.com/jfbastien/bit_cast/blob/c288208c1a68a329a5839f007200ab2ee2b65073/bit_cast.h#L61). Maybe that makes the API a little more convenient?

PR #670, opened 2019-11-28
 Build the man page in Cirrus CI 

tobim:


tobim:
[ch10486]

tobim:
Please don't merge yet, the man page build actually depends on pandoc being availabe in the environment.

.ci/scripts/generate-build-tree
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# e - non-interactive terminals will exit on failures immediately
+# x - print commands before exiting
+set -ex
```
mavam: Is this actually needed for a single-command shell script?

tobim: I removed the `-e` and polished the comment.

dominiklohmann:
You'll need to run `cmake --build ... --target man` manually before package/install.

What about FreeBSD and NixOS? The latter isn't in Cirrus anymore, but we should still provide the man page with these builds if possible.

tobim:
I removed FreeBSD, pandoc is not available pre-built and you don't want to compile that every time.
There is no static build in this repository, so nothing to add for that.

tobim:
> You'll need to run `cmake --build ... --target man` manually before package/install.

I think it will be less surprising if the man page is part of target all.

mavam:
I agree, if pandoc is present, then the user should not have to do anything in addition to build the man page. If it would have a significant effect on build time, I'd be reluctant.

PR #671, opened 2019-11-28
Exit with an error code if command parsing fails

tobim:
This PR also changes the option `--config-file` to `--config`.

### DIFF
--- a/libvast/src/error.cpp
+++ b/libvast/src/error.cpp
@@ -49,6 +49,7 @@ const char* descriptions[] = {
   "missing_subcommand",
   "missing_component",
   "unimplemented",
+  "silent",
 };
 
 void render_default_ctx(std::ostringstream& oss, const caf::message& ctx) {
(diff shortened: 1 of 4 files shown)


libvast/src/error.cpp
```diff
@@ -72,6 +73,9 @@ std::string render(caf::error err) {
   using caf::atom_uint;
   std::ostringstream oss;
   auto category = err.category();
+  if (atom_uint(category) == atom_uint("vast")
+      && static_cast<vast::ec>(err.code()) == ec::silent)
```
mavam:
```suggestion
      && err == ec::silent)
```

I haven't tested this, but I think this is possible.

libvast/src/error.cpp
```diff
@@ -74,7 +74,7 @@ std::string render(caf::error err) {
   std::ostringstream oss;
   auto category = err.category();
   if (atom_uint(category) == atom_uint("vast")
-      && static_cast<vast::ec>(err.code()) == ec::silent)
+      && err == ec::silent)
```
mavam: And now it fits on the previous line as well.

dominiklohmann:
> Adding a new error code seems like a one-off here. Can't we just check for `ec::unspecified` in combination with an empty message?

Wouldn't that still print the !! in the beginning? I like the explicitness of the new error code.

tobim:
Yes, it is intentionally explicit.

mavam:
I'm approving because I can see the argument on the other side. My point was that if we would add a new error code whenever we need some dispatching logic in a different context, we'd abuse the error abstraction. The error code should represent a *class* of errors, not a single one. However, `silent` is generic enough to serve as such a class.

PR #677, opened 2019-12-05
Switch from Cirrus CI to GitHub Actions

dominiklohmann:
After experiencing the delays of low availability for the macOS CI servers from Cirrus, we now plan to use GitHub Actions instead.


.github/workflows/build.yml
```diff
@@ -0,0 +1,67 @@
...
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          submodules: recursive
+          ref: ${{ github.ref }}
+      - uses: actions/setup-python@v1
+        with:
+          python-version: "3.8"
+      - name: Install Ubuntu dependencies
+        if: matrix.os == 'ubuntu-latest'
```
tobim: Is there a reason for switching away from debian?

.github/workflows/build.yml
```diff
@@ -0,0 +1,67 @@
...
+          pip3 install --user cmake
+          ln -sf /usr/local/bin/cmake $(python3 -m site --user-site)/cmake/data/bin/cmake
+      - name: Install macOS dependencies
+        if: matrix.os == 'macos-latest'
+        run: |
+          brew install openssl cmake git ninja python libpcap jq tcpdump rsync pandoc apache-arrow
+      - name: Configure
+        run: |
+          echo "VAST-$(git describe)-$(uname -s)-${{ matrix.configure == '' && 'Release' || 'Debug' }}" > PACKAGE_NAME
+          ./configure --generator=Ninja --package-name=$(cat PACKAGE_NAME) --disable-arrow ${{ matrix.configure }}
```
tobim: Why aren't the existing CI scripts used here?

.github/workflows/build.yml
0ortmann: There is no Debian in github actions.

.github/workflows/build.yml
dominiklohmann: There are three reasons for this:
- The build matrix is easier to read if the configuration is either empty string or `--dev-mode`. 
- It is easier for a devops person without much knowledge about CMake (i.e., @0ortmann) not having to interact with `CMake -B...` manually.
- We recommend using `configure` in our README, so CI should use it, too.

.github/workflows/build.yml
tobim: Ah, I see.

.github/workflows/build.yml
tobim: Thanks for the explanation.

PR #684, opened 2019-12-11
Disable component blacklist in integration tests

dominiklohmann:
With this change, CAF logs (most notably `caf_flow` logs) are no longer disabled for integration tests. Some failures will now be easier to debug than before (e.g., https://github.com/tenzir/vast/runs/336745938).

### DIFF
--- a/integration/default_set.yaml
+++ b/integration/default_set.yaml
@@ -4,14 +4,14 @@ fixtures:
   BiTopologyTester:
     enter: | # python
       node0 = Server(self.cmd,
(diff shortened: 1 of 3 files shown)


dominiklohmann:
Turns out this works for the client side of the integration tests, but that vast.conf file does not get picked up for the server process. That also means we never had trace logs for that.

dominiklohmann:
@mavam fixed the issue. The output of `rg caf_flow` is empty now for all integration test logs.

PR #690, opened 2019-12-19
CMake fixes and tweaks

mavam:


mavam:
`cmake-format` butchers the comments. @dominiklohmann can we avoid that this happens?

dominiklohmann:
You can disable comment markup reflow completely, see `cmake-format --help`:

```
Comment Formatting:
  Override config options affecting comment formatting

  --bullet-char BULLET_CHAR
                        What character to use for bulleted lists
  --enum-char ENUM_CHAR
                        What character to use as punctuation after numerals in
                        an enumerated list
  --enable-markup [ENABLE_MARKUP]
                        enable comment markup parsing and reflow
  --first-comment-is-literal [FIRST_COMMENT_IS_LITERAL]
                        If comment markup is enabled, don't reflow the first
                        comment block in each listfile. Use this to preserve
                        formatting of your copyright/license statements.
  --literal-comment-pattern LITERAL_COMMENT_PATTERN
                        If comment markup is enabled, don't reflow any comment
                        block which matches this (regex) pattern. Default is
                        `None` (disabled).
  --fence-pattern FENCE_PATTERN
                        Regular expression to match preformat fences in
                        comments default=r'^\s*([`~]{3}[`~]*)(.*)$'
  --ruler-pattern RULER_PATTERN
                        Regular expression to match rulers in comments
                        default=r'^\s*[^\w\s]{3}.*[^\w\s]{3}$'
  --canonicalize-hashrulers [CANONICALIZE_HASHRULERS]
                        If true, then insert a space between the first hash
                        char and remaining hash chars in a hash ruler, and
                        normalize it's length to fill the column
```

Not. sure why this happens here. Feel free to add a `.cmake-format` file to the repo, we just haven't had the need for it yet. You can run `cmake-format --dump-config` to get the default config and then tweak from there.

mavam:
[We have a .clang-format](https://github.com/tenzir/vast/blob/master/.cmake-format) already. I've tweaked it to avoid comment reformatting.

 I could not find a way to make `cmake-format` stop reformatting the function invocation. I've purposefully used the layout as in the documentation for [`find_package_handle_standard_args`](https://cmake.org/cmake/help/v3.3/module/FindPackageHandleStandardArgs.html):

```cmake
find_package_handle_standard_args(<PackageName>
  [FOUND_VAR <result-var>]
  [REQUIRED_VARS <required-var>...]
  [VERSION_VAR <version-var>]
  [HANDLE_COMPONENTS]
  [CONFIG_MODE]
  [FAIL_MESSAGE <custom-failure-message>]
  )
```

But our config wants one argument per line.

dominiklohmann:
Just surround it with `# cmake-format: off` and `# cmake-format: on` then I'd say.

mavam:
Uh, does manual disabling scale? We have so many function calls where we use the argument formatting as in the docs.

Even setting ` --max-pargs-hwrap` to a high value (e.g., 10) does the wrong thing, i.e., creating a line that is as long as possible:
```
find_package_handle_standard_args(
  CAF FOUND_VAR CAF_FOUND REQUIRED_VARS CAF_LIBRARIES CAF_INCLUDE_DIRS
  HANDLE_COMPONENTS)
```

dominiklohmann:
For the formatting tool, all these arguments are just strings. There's no special casing for argument groups, because there is no special syntax for it in the CMake language. I'm afraid we can't "fix" this reliably. You can still do the one-off if it annoys you too much, or just keep it as expected by `cmake-format`. 

I don't think it's worth the effort to make sure the CMake looks presentable, because CMake really isn't ever presentable at all.

mavam:
> I don't think it's worth the effort to make sure the CMake looks presentable, because CMake really isn't ever presentable at all.

Fair enough, I've undid my manual changes.

Now Black complains about the `.cmake-format` file. Is there a quick fix for this?

dominiklohmann:
> Now Black complains about the `.cmake-format` file. Is there a quick fix for this?

Run black on it, since it's just a python file. Black rightfully complains since your PR touches it, so it should be checked by CI.

mavam:
@dominiklohmann Commit reworded and force-pushed.

PR #694, opened 2019-12-28
Add script to demultiplex Corelight sensor data

mavam:
This script demultiples the multiple connections a Corelight sensor opens when exporting logs via TCP.


mavam:
[ch11285]

PR #697, opened 2020-01-07
Relocate debug information to a common prefix

dominiklohmann:
For CCache users, we should relocate debug information to a common prefix to speed up rebuilds from a different path, e.g., when using `git-worktree`.

The cost of rewriting debug info paths is relatively low, so I think this should be on by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,11 +196,6 @@ endmacro ()
 set(LDFLAGS "$ENV{LDFLAGS}")
 set(ARFLAGS "$ENV{ARFLAGS}")
 
-# Ninja doesn't colorize compiler diagnostics by default.
-if (CMAKE_GENERATOR STREQUAL "Ninja")
-  set(EXTRA_FLAGS "${EXTRA_FLAGS} -fdiagnostics-color")
-endif ()
-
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -221,6 +216,17 @@ if (NOT CMAKE_CXX_FLAGS)
...
   # heavy code.
   set(EXTRA_FLAGS "${EXTRA_FLAGS} -ftemplate-backtrace-limit=0")
+  # Ninja doesn't colorize compiler diagnostics by default.
+  if (CMAKE_GENERATOR STREQUAL "Ninja")
+    set(EXTRA_FLAGS "${EXTRA_FLAGS} -fdiagnostics-color")
+  endif ()
+  # For clang, relocate debug/file/macro paths to a common prefix.
+  # This massively improves recompilation speed with CCache when working from
+  # different worktrees.
+  if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
tobim: This option is not specific to Clang. You can also shorten the explanation.

dominiklohmann: Done and rebased. I thought this was clang only.

PR #698, opened 2020-01-07
Add FreeBSD install instructions

mavam:


INSTALLATION.md
```diff
@@ -61,4 +61,40 @@ TODO
...
+to the ones in `/etc/rc.d` and `/usr/local/etc/rc.d`) that allows
+[service(8)][service] to manage a `vast` process with the `start`, `stop`, and
+`restart` options. The rc script uses [daemon(8)][daemon] to manage the `vast`
+process.
+
+The following steps install VAST as a persistent service:
+
+1. Copy the rc script to `/usr/local/etc/rc.d`.
+
+   **Note**: this happens automatically when executing the `install` target.
```
tobim:
```suggestion
   **Note**: this happens automatically when running `make install` or `ninja install`.
```
`target` is build system jargon that users may not be familiar with.

INSTALLATION.md
```diff
@@ -61,4 +61,40 @@ TODO
...
+
+2. Add the following line to `/etc/rc.conf`:
```
tobim:
```suggestion
2. Add the following line to `/etc/rc.conf` to start VAST on each boot:
```

INSTALLATION.md
mavam: I was trying to avoid saying `make` and `ninja`. Maybe we should have all examples be with `make` and mention once that they can be replaced with `ninja`?

INSTALLATION.md
mavam: I don't think this is necessary to add here, because the instructions are for adding VAST as a "persistent service".

mavam: I don't think this is necessary to add here, because the instructions are for adding VAST as a "persistent service".

INSTALLATION.md
tobim: That would be acceptable.

INSTALLATION.md
tobim: This is fine then.

PR #700, opened 2020-01-07
Provide a graceful fallback from pkg-config

tobim:


CMakeLists.txt
```diff
@@ -13,23 +24,53 @@ else ()
...
+
+# Force colored output for the Ninja generator
+option(VAST_COLOR_NINJA "Enable ANSI-colored output for the Ninja generator."
+       ON)
+if ("${CMAKE_GENERATOR}" STREQUAL "Ninja" AND VAST_COLOR_NINJA)
+  if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    add_compile_options(-fdiagnostics-color=always)
+  elseif ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
+    add_compile_options(-fcolor-diagnostics)
+  endif ()
```
dominiklohmann: Line 67 and 216 both set the `-fdiagnostics-color` for Ninja. The latter should be removed I think.

tobim: Great catch!

PR #703, opened 2020-01-08
Fix race in index lookup

tobim:


libvast/src/system/index.cpp
```diff
@@ -403,60 +403,60 @@ behavior index(stateful_actor<index_state>* self, const path& dir,
...
-        st.pending.erase(iter);
+        VAST_ASSERT(lookup.partitions.empty());
         VAST_DEBUG(self, "returns without result: no partitions qualify");
-        return no_result();
+        no_result();
+      }
+      VAST_DEBUG(self, "scheduled", qm.size(), "/", hits,
+                 "partitions for query", expr);
+      if (!lookup.partitions.empty()) {
+        auto result = st.pending.emplace(query_id, std::move(lookup));
```
dominiklohmann:
```suggestion
        [[maybe_unused]] auto result = st.pending.emplace(query_id, std::move(lookup));
```

libvast/src/system/index.cpp
```diff
@@ -403,60 +403,60 @@ behavior index(stateful_actor<index_state>* self, const path& dir,
...
+        respond(uuid::nil(), uint32_t{0}, uint32_t{0});
+        self->send(client, done_atom::value);
       };
       // Get all potentially matching partitions.
       auto candidates = st.meta_idx.lookup(expr);
       // Report no result if no candidates are found.
       if (candidates.empty()) {
         VAST_DEBUG(self, "returns without result: no partitions qualify");
-        return no_result();
+        no_result();
```
dominiklohmann: Missing `return` after this?

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🐞 A race condition in the index logic was able to lead to incomplete or empty
+  result sets for `vast export`. This has been fixed.
```
mavam: I think something like `This has been fixed` doesn't provide much insightful information. The fact that we are resolving bug should be obvious from the bug icon. (I found one other occurrence of this wording in the changelog, alas.)

I suggest either removing the sentence or giving an idea *how* the bug has been fixed.

libvast/src/system/index.cpp
```diff
@@ -403,60 +403,60 @@ behavior index(stateful_actor<index_state>* self, const path& dir,
   // simply waits for a worker).
   self->set_default_handler(caf::skip);
   self->state.has_worker.assign(
-    [=](expression& expr) -> result<uuid, uint32_t, uint32_t> {
-      auto& st = self->state;
+    [=](expression& expr) {
+      auto respond = [&](auto&&... xs) {
+        auto mid = self->current_message_id();
+        unsafe_response(self, self->current_sender(), {}, mid.response_id(),
+                        std::forward<decltype(xs)>(xs)...);
```
mavam:
```suggestion
                        std::forward<std::decay_t<decltype(xs)>>(xs)...);
```

dominiklohmann: Are you sure this is necessary? I've always seen it used without the decay.

mavam: Assuming `auto&&` is the same type of forwarding reference as `T&&`, then you need to use the `std::decay`. Otherwise you'll get a reference via `decltype`. If you grep through the code base, you'll see many instances of this (unfortunately necessary) pattern.

mavam: Apologies, I have to rudder back. Thanks for explaining this to me at the whiteboard in more detail, @dominiklohmann!

dominiklohmann: We just concluded at the whiteboard that `decay` is incorrect here and should not be used.

CHANGELOG.md
tobim: I opted for removal. The change log is not the place for detailed information.

PR #705, opened 2020-01-09
Fix configuration file option parsing

tobim:


mavam:
@tobim I think we forgot to change one file in this PR...

PR #708, opened 2020-01-13
Fix a memory leak in the arrow table slice

tobim:


tobim:
> I wonder why `LeakSanitizer` didn't catch this one, considering it's enabled by default with `AddressSanitizer`.

`LeakSanitizer` is not enabled by default on macOS, and we don't build with Arrow on Ubuntu.

PR #711, opened 2020-01-13
Support runtime options in value index factory

mavam:
This PR adds the ability to pass arbitrary context into the value index factory, which allows for evaluating runtime parameters to select the appropriate index implementation. In particular, we pass in the partition size to automatically deduce the number of bytes for our optimized hash index for strings.

### DIFF
--- a/libvast/src/value_index.cpp
+++ b/libvast/src/value_index.cpp
@@ -135,7 +135,7 @@ caf::error inspect(caf::deserializer& source, value_index_ptr& x) {
     x = nullptr;
     return caf::none;
   }
-  x = factory<value_index>::make(std::move(t), options{});
+  x = factory<value_index>::make(std::move(t), caf::settings{});
   if (x == nullptr)
     return make_error(ec::unspecified, "failed to construct value index");
   return x->deserialize(source);
@@ -532,7 +532,7 @@ port_index::lookup_impl(relational_operator op, data_view d) const {
 
 // -- sequence_index -----------------------------------------------------------
 
-sequence_index::sequence_index(vast::type t, options opts)
+sequence_index::sequence_index(vast::type t, caf::settings opts)
   : value_index{std::move(t)}, opts_{std::move(opts)} {
   max_size_
     = caf::get_or(opts_, "max-size", defaults::index::max_container_elements);
--- a/libvast/src/value_index.cpp
+++ b/libvast/src/value_index.cpp
@@ -131,7 +131,7 @@ caf::error inspect(caf::deserializer& source, value_index_ptr& x) {
     x = nullptr;
     return caf::none;
   }
-  x = factory<value_index>::make(std::move(t));
+  x = factory<value_index>::make(std::move(t), options{});
   if (x == nullptr)
     return make_error(ec::unspecified, "failed to construct value index");
   return x->deserialize(source);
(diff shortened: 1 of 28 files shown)


libvast/src/value_index.cpp
```diff
@@ -131,7 +131,7 @@ caf::error inspect(caf::deserializer& source, value_index_ptr& x) {
     x = nullptr;
     return caf::none;
   }
-  x = factory<value_index>::make(std::move(t));
+  x = factory<value_index>::make(std::move(t), options{});
```
tobim: Maybe we should add a comment here:
```
// Persisted value indices shall not be parameterized at re-materialization
// time. Instead, a concrete index is responsible to persist its own metadata.
```

mavam: I would really like to have this sort of guarantee, keeping the options only relevant to the factory. However, we have to support container types as well, which perform internal calls to the factory function and thus need to see the options. Consequently, the container indexes must get the options/settings in the constructor, and they must also store them as persistent state.

mavam: As we figured out together at the HQ, there's a much bigger problem here: since the options can contain data that determines the concrete type of the hierarchy, they *must* be persisted as well. Only then the factory function will correctly default-construct the most derived type.

libvast/src/value_index_factory.cpp
```diff
@@ -74,29 +65,81 @@ auto add_value_index_factory() {
...
+          auto i = opts.find("cardinality");
+          if (i == opts.end())
+            // Default to a 40-bit hash value -> good for 2^20 unique digests.
+            return std::make_unique<hash_index<5>>(std::move(x));
+          auto cardinality = caf::get_if<int_type>(&i->second);
+          VAST_ASSERT(cardinality); // checked in make(x, opts)
+          // caf::settings doesn't support unsigned integers, but the
+          // cardinality is a size_t, so we may get -1 if someone provides
+          // numeric_limits<size_t>::max().
+          if (*cardinality == -1) {
```
tobim: Shouldn't we check for *cardinality < 0 instead?

libvast/vast/value_index.hpp
```diff
@@ -98,9 +104,10 @@ class value_index {
   virtual caf::expected<ids>
   lookup_impl(relational_operator op, data_view x) const = 0;
 
-  ewah_bitmap mask_;      ///< The position of all values excluding nil.
-  ewah_bitmap none_;      ///< The positions of nil values.
-  const vast::type type_; ///< The type of this value index.
+  ewah_bitmap mask_;         ///< The position of all values excluding nil.
+  ewah_bitmap none_;         ///< The positions of nil values.
+  const vast::type type_;    ///< The type of this index.
+  const caf::settings opts_; ///< Runtime context with additional parameters.
```
tobim: The following is intended as a start of discussion, not an outright request for change (depending on the outcome, we can still implement that in a follow up PR):

I'm not super happy about adding a direct relationship between `value_index` and `caf::settings` or any other polymorphic type. From a user point of view it does not make sense that they need to wrap the `length` parameter into a `caf::settings` so they can construct a `string_index`.

An alternative approach might be to leave `value_index` as it is, but to compose the `options` parameter to it by membership in a wrapper type:
```
// TODO: Find a better name ;).
struct value_index_wrapper {
  value_index_ptr index;
  caf::settings options;
};
```
I'm not sure if this really leads to a leaner implementation though.

mavam: Yeah, I'm in agreement that this is not optimal. I don't like the tight coupling either, but I don't see another way to handle the polymorphic *recursive* type `value_index`. The problems are containers like `vector`, `set`, and `table`.

PR #721, opened 2020-01-22
Refactor interface of system::get_node_component

lava:
* Rename get_node_component -> get_node_components
* Change arguments to accept the desired actors as
  strings instead of atoms.
* Change return type to use caf::actor instead of
  caf::expected<caf::actor> to save one level of
  error unwrapping on the call site, as actor is
  already nullable.

### DIFF
--- a/libvast/src/system/source_command.cpp
+++ b/libvast/src/system/source_command.cpp
@@ -78,29 +78,28 @@ caf::message source_command(const command::invocation& invocation,
   auto guard = system::signal_monitor::run_guarded(
     sig_mon_thread, sys, defaults::system::signal_monitoring_interval, self);
   // Get node components.
-  auto components
-    = get_node_component<accountant_atom, importer_atom>(self, node);
+  auto components = get_node_components(self, node, {"accountant", "importer"});
   if (!components)
     return make_message(components.error());
   if (auto& accountant = (*components)[0]) {
     VAST_DEBUG(invocation.full_name, "assigns accountant to source");
-    self->send(src, actor_cast<accountant_type>(*accountant));
+    self->send(src, actor_cast<accountant_type>(accountant));
   }
   // Connect source to importer.
   auto& importer = (*components)[1];
   if (!importer)
-    return make_message(importer.error());
+    return make_message(make_error(ec::missing_component, "importer"));
   VAST_DEBUG(invocation.full_name, "connects to importer");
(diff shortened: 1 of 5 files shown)


libvast/vast/system/node_control.hpp
```diff
@@ -33,22 +33,23 @@ spawn_at_node(caf::scoped_actor& self, caf::actor node, Arguments&&... xs) {
...
+get_node_components(caf::scoped_actor& self, caf::actor node,
+                    const char* const (&names)[N]) {
+  auto result = caf::expected{std::array<caf::actor, N>{}};
   self->request(node, caf::infinite, get_atom::value)
     .receive(
       [&](const std::string& id, system::registry& reg) {
-        auto find_actor = [&](auto atom) -> caf::expected<caf::actor> {
-          auto er = reg.components[id].find(to_string(atom));
+        auto find_actor = [&](const char* name) -> caf::actor {
+          auto er = reg.components[id].find(name);
```
mavam: While you're at it, this check could be a bit modernized:

```cpp
if (auto er = reg.components[id].find(name); er != reg.components[id].end())
  return er->second.actor;
return nullptr;
```

libvast/src/system/source_command.cpp
```diff
@@ -78,37 +78,36 @@ caf::message source_command(const command::invocation& invocation,
...
   if (auto& accountant = (*components)[0]) {
     VAST_DEBUG(invocation.full_name, "assigns accountant to source");
-    self->send(src, actor_cast<accountant_type>(*accountant));
+    self->send(src, actor_cast<accountant_type>(accountant));
   }
   // Connect source to importer.
   auto& importer = (*components)[1];
   if (!importer)
-    return make_message(importer.error());
+    return make_message("couldnt lookup IMPORTER");
```
tobim: `caf::message` works a bit like `std::any`, so in this case we need to make sure it contains an error instead of a string.
```suggestion
    return make_message(make_error(ec::missing_component, "importer"));
);
```

mavam:
Maybe we could use `std::string_view` instead of `const char*` here?

lava:
I updated the PR to use `string_view`. Since it looked a bit weird to accept a `string_view` just to immediately copy it into a new string, I also took the opportunity to enable heterogenous lookup on the maps in the tracker, so maybe we'll even get a free speedup out of this :)

PR #726, opened 2020-01-27
Generalize hash indexes to all data types

mavam:
This makes it possible to use `#index=hash` on all data types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🔄 Hash indexes previously only worked with data of type `string`. They now
+  support all data types.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🔄 Hash indexes previously only worked with data of type `string`. They now
```
tobim: I would prefer to merge this entry with the one for #632, and add a second PR link to the end of that. I think in general we should strive to keep the changelog context free.

mavam:
I've added another unit for sets and merged the changelog entries.

Adding another unit test for set would not add much value in my opinion, as it would only test that the factory dispatches for that type to the hash index. Since hashing the views is orthogonal as well, I would say we have good coverage for now.

PR #727, opened 2020-01-28
Support versioning of persistent data structures

mavam:
This PR adds versioning support of persistent data structures via *flatbuffers*. It addresses the problem of changing serialized on-disk structures with version upgrades. To date, VAST cannot access data of older incompatible versions. Using flatbuffers allows for specifying the binary format in portable manner that supports a robust upgrade path.

There's also a performance benefit. Due to aligned data representation and
ability to perform random access, it plays well with our builtin memmory-mapped archive.


libvast/CMakeLists.txt
```diff
@@ -236,6 +242,56 @@ install(
 
 add_library(vast::libvast ALIAS libvast)
 
+# ----------------------------------------------------------------------------
+# flattbuffers
+# ----------------------------------------------------------------------------
+
+file(GLOB flatbuffers_schemas ${CMAKE_CURRENT_SOURCE_DIR}/vast/fbs/*.fbs)
```
tobim: Globbing means you won't touch the CMakeLists.txt when a new file was added. This might be convenient at first sight, but it also means that existing build directories can't know when a regeneration is needed. This will result in failing builds for anyone that updates master after such an addition was merged.

I recommend to list these files explicitly instead.

libvast/CMakeLists.txt
```diff
@@ -236,6 +242,56 @@ install(
...
+set(flatbuffers_output_path ${CMAKE_CURRENT_BINARY_DIR}/vast/fbs)
+
+add_custom_target(flatbuffers)
+add_dependencies(libvast flatbuffers)
+
+# Translate paths to desired output paths.
+foreach (schema ${flatbuffers_schemas})
+  get_filename_component(basename ${schema} NAME_WE)
```
tobim: We should extract the body of this loop into a function and move it to `cmake/Flatbuffers.cmake`.

libvast/vast/uuid.hpp
```diff
@@ -13,25 +13,24 @@
 
 #pragma once
 
-#include <array>
+#include "vast/byte.hpp"
+#include "vast/detail/operators.hpp"
+#include "vast/span.hpp"
+#include "vast/uuid.hpp"
```
tobim: `operators.hpp` seems unused, and it probably also should not include itself.

libvast/test/uuid.cpp
```diff
@@ -11,27 +11,33 @@
...
-  set.insert(uuid::random());
-  CHECK(set.find(*u) != set.end());
+TEST(pod size) {
+  CHECK(sizeof(uuid) == size_t{16});
+}
+
+TEST(parseable and printable) {
+  auto x = to<uuid>("01234567-89ab-cdef-0123-456789abcdef");
+  REQUIRE(x);
+  CHECK(to_string(*x) == "01234567-89ab-cdef-0123-456789abcdef");
```
tobim: Nit: We should prefer `CHECK_EQUAL` if possible.

libvast/CMakeLists.txt
```diff
@@ -54,6 +54,7 @@ set(libvast_sources
     src/ewah_bitmap.cpp
     src/expression.cpp
     src/expression_visitors.cpp
+    src/fbs/utils.cpp
```
tobim: This file is not added in the same commit. Maybe I'll discover it later?

libvast/CMakeLists.txt
dominiklohmann: I think adding `CONFIGURE_DEPENDS` on the GLOB would be fine as well.

libvast/CMakeLists.txt
mavam: I don't think this would help here, because we are adding targets in here dynamically, one for each file.

libvast/vast/uuid.hpp
mavam: Actually, `detail::totally_orderd<T>` is used here. I've removed the reflexive include.

libvast/CMakeLists.txt
mavam: Yeah, that happened when I factored my longer commit into multiple smaller ones.

mavam:
@tobim after reviewing, this is now ready for merge into the epic/integration branch.

libvast/src/segment.cpp
```diff
@@ -13,112 +13,117 @@
...
-  if (result->header_.version > version) {
-    VAST_ERROR_ANON(__func__, "got newer segment version",
-                    result->header_.version, "instead of", version);
-    return nullptr;
+  return segment_ptr{new segment{std::move(chunk)}, false};
+}
+
+uuid segment::id() const {
+  auto ptr = fbs::GetSegment(chunk_->data());
+  auto data = span<const uint8_t, 16>(ptr->uuid()->Data(), 16);
```
tobim: These `uint8_t` to `byte` and back reinterpretations are unfortunate. Nothing we can do about it I guess.

libvast/src/segment.cpp
```diff
@@ -13,112 +13,117 @@
...
+  return uuid{as_bytes(data)};
+}
+vast::ids segment::ids() const {
+  vast::ids result;
+  auto ptr = fbs::GetSegment(chunk_->data());
```
tobim: Could we make this a `mutable` member variable, or is it possible for the chunk to grow in code outside of this file and the `builder`?

libvast/src/segment_builder.cpp
```diff
@@ -34,38 +35,42 @@ segment_builder::segment_builder() {
...
+  size_t offset;
+  size_t size;
+  auto ptr = builder_.ReleaseRaw(size, offset);
+#if 0
+  VAST_ASSERT(size > offset);
+  auto verifier = flatbuffers::Verifier{ptr + offset, size - offset};
+  VAST_ASSERT(fbs::VerifySegmentBuffer(verifier));
+#endif
+  auto deleter = [=]() { flatbuffers::DefaultAllocator::dealloc(ptr, size); };
+  auto chk = chunk::make(size - offset, ptr + offset, deleter);
```
tobim: This means that the first `offset` bytes are used for flatbuffer internals?

mavam: Yeah, there is some internal bookkeeping in the builder that needs to be skipped. Based on the documentation it's not clear exactly what that bookkeeping is, but Wouter gave green light when I presented this example on the mailing list.

libvast/src/segment.cpp
mavam: Yeah, I'm not a fan either. On the flatbuffers side, there is only `byte` and `ubyte`. For the the type `string`, it's possible to specify a custom C++ class, but not for arithmetic types.

tobim:
> Thanks for the nix patch! Is there a way to specify a minimum version? We require 1.11.0.

As discussed in person, it is not necessary / natural to set a version for each package. Instead, all packages are versioned in combination by the version https://github.com/NixOS/nixpkgs, which we pin to a specific commit at `nix/pinned.nix`.

libvast/src/fbs/utils.cpp
```diff
@@ -33,22 +33,15 @@ caf::expected<Encoding> create_encoding(caf::atom_value x) {
...
-  }
-}
-
 // TODO: this function will boil down to accessing the chunk inside the table
 // slice and then calling GetTableSlice(buf). But until we touch the table
 // slice internals, we use this helper.
-caf::expected<flatbuffers::Offset<TableSlice>>
-create_table_slice(flatbuffers::FlatBufferBuilder& builder, table_slice_ptr x) {
+caf::expected<flatbuffers::Offset<TableSliceBuffer>>
+create_table_slice_buffer(flatbuffers::FlatBufferBuilder& builder,
```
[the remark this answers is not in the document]
tobim: s/defined/declared/

libvast/vast/fbs/utils.hpp
```diff
@@ -0,0 +1,52 @@
...
+create_bytes(flatbuffers::FlatBufferBuilder& builder, const T& x) {
+  static_assert(detail::is_any_v<Byte, int8_t, uint8_t>);
+  auto bytes = as_bytes(x);
+  auto data = reinterpret_cast<const Byte*>(bytes.data());
+  return builder.CreateVector(data, bytes.size());
+}
+
+caf::expected<Encoding> create_encoding(caf::atom_value x);
+
+caf::atom_value make_encoding(Encoding x);
```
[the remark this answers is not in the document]
mavam: I would argue the opposite way: once we have a pointer to a flatbuffer object, we know it's safe to use it because it should only involve creating a view.

libvast/vast/fbs/utils.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include <cstdint>
+
+#include <flatbuffers/flatbuffers.h>
+
+namespace vast::fbs {
+
+// The utility functions in this header have the following naming convention:
+// - `create_*` are functions to convert from VAST to flatbuffers
+// - `make_*` are functions to convert from flatbuffers to VAST
```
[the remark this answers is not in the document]
mavam: I like pack/unpack!

libvast/src/segment.cpp
mavam: We could. I decided to avoid it because having just a chunk as member is kinda like pimpl.

libvast/CMakeLists.txt
mavam: Per our discussion, you convinced me that the extra safety belt makes sense here.

PR #728, opened 2020-01-28
Simplify partition structure

tobim:
This needs to be cleaned and rebased. Do not review yet.

TODO:
- [x] Handle pattern matches in type queries
- [x] Unit tests
- [x] Cleanup partition
- [x] Reorganize commits


libvast/vast/system/indexer_stage_driver.hpp
```diff
@@ -13,31 +13,73 @@
...
+/// First, the table id offset is used to determine the target partition.
+/// The partition is created if it did not exists beforehand. Second, the
+/// layout is used to retrieve the set of downstream slots that the slice shall
+/// be passed on to. If the table entry is missing, it is created by retrieving
+/// the indexer actors from the target partition and associating them to their
+/// matching fields from the layout.
+///
+///  Example for a partition containing 2 types foo and bar with the layouts:
+///
+///     field name: field type # ID (hash of record_field)
```
mavam, line 40: I don't understand the formatting here.

libvast/vast/system/indexer_stage_driver.hpp
```diff
@@ -13,31 +13,73 @@
...
+///     b:          string,    # B
+///     c:          string,    # C
+///     d:          address    # D
+///  }
+///  type bar = record {
+///     a:          int,       # A
+///     b:          domain,    # B'
+///     c:          enum,      # C'
+///     d:          address    # D
+///  }
```
mavam, line 52: I don't think you can map a field `a` that is in different layouts to the same indexer. What if `a` has completely different semantics in `foo` and `bar`?

libvast/src/system/indexer.cpp
```diff
@@ -40,31 +40,33 @@ indexer_state::~indexer_state() {
...
+                 std::move(filename), std::move(column));
   return col.init();
 }
 
-behavior indexer(stateful_actor<indexer_state>* self, path dir,
-                 type column_type, caf::settings index_opts, size_t column,
-                 caf::actor index, uuid partition_id, atomic_measurement* m) {
+caf::behavior
+indexer(caf::stateful_actor<indexer_state>* self, path dir, type column_type,
+        caf::settings index_opts, std::string column, caf::actor index,
```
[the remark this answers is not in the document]
tobim: This parameter has been removed.

libvast/src/system/indexer.cpp
```diff
@@ -40,31 +40,33 @@ indexer_state::~indexer_state() {
...
+        uuid partition_id, atomic_measurement* m) {
   VAST_TRACE(VAST_ARG(dir), VAST_ARG(column_type), VAST_ARG(column));
   VAST_DEBUG(self, "operates for column", column, "of type", column_type);
   if (auto err = self->state.init(
-        self, std::move(dir) / "fields" / std::to_string(column),
-        std::move(column_type), std::move(index_opts), column, std::move(index),
-        partition_id, m)) {
+        self, std::move(dir) / "fields" / column + "-" + to_digest(column_type),
+        std::move(column_type), std::move(index_opts), std::move(column),
```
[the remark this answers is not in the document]
tobim: I think with C++17 it is. This is not relevant any more because the `column` parameter is gone and the `indexer` is now initialized with a full path to its backing file.

libvast/vast/system/indexer_stage_driver.hpp
```diff
@@ -13,31 +13,68 @@
...
 
-/// @relates indexer_stage_driver
-/// A downstream manager type for dispatching data to INDEXER actors.
-using indexer_downstream_manager
-  = caf::broadcast_downstream_manager<table_slice_ptr,
-                                      indexer_stage_filter,
-                                      indexer_stage_selector>;
+  const mapped_type& operator()(const table_slice_ptr& x);
+
+  void erase(std::ptrdiff_t offset);
```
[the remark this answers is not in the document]
tobim: This is not required any more. On a side note: You are right, it would have made sense to use the layout for better robustness.

libvast/src/system/index.cpp
```diff
@@ -300,9 +300,16 @@ void index_state::reset_active_partition() {
...
+  stage->out().register(active);
   active_partition_indexers = 0;
 }
 
+partition* index_state::get_active_partition(const table_slice_ptr& slice) {
+  if (active->capacity() < slice->rows())
+    reset_active_partition();
+  return active.get();
+}
+
```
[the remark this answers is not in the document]
tobim: I don't agree, this function encapsulates a change in the `index_state`, therefore it should be a member of it.

libvast/src/system/partition.cpp
```diff
@@ -54,42 +59,112 @@ partition::~partition() noexcept {
...
+      auto fqf = to_fully_qualified(layout.name(), field);
+      auto j = indexers_.find(fqf);
+      if (j == indexers_.end()) {
+        // Spawn a new indexer.
+        auto k = indexers_.emplace(fqf, wrapped_indexer{});
+        auto& ip = k.first->second;
+        auto pos = std::distance(indexers_.begin(), k.first);
+        ip.indexer = state().make_indexer(column_file(fqf), fqf.type, id(),
+                                          &measurements_[pos]);
+        state_->active_partition_indexers++;
```
mavam, line 144: This variable is never decremented, only reset to 0 upon initialization. Is that intentional?

libvast/src/system/partition.cpp
```diff
@@ -54,42 +59,112 @@ partition::~partition() noexcept {
...
+        ip.slot
+          = out().parent()->add_unchecked_outbound_path<table_slice_column>(
+            ip.indexer);
+        ip.outbound = out().paths().rbegin()->second.get();
```
mavam, line 148: I didn't look up the type without `.get()`, but is it maybe possible to use the smart pointer in the indexer state instead of a raw pointer?

libvast/src/system/partition.cpp
```diff
@@ -98,109 +173,85 @@ caf::actor fetch_indexer(table_indexer& tbl, const data_extractor& dx,
...
-      return nullptr;
     // We know the answer immediately: all IDs that are part of the table.
     // However, we still have to "lift" this result into an actor for the
     // EVALUATOR.
+    ids row_ids;
+    for (auto& [name, ids] : meta_data_.type_ids) {
+      if (evaluate(name, op, x)) {
+        row_ids |= ids;
+      }
+    }
```
mavam, line 194:
```suggestion
    for (auto& [name, ids] : meta_data_.type_ids)
      if (evaluate(name, op, x))
        row_ids |= ids;
```

libvast/src/system/partition.cpp
```diff
@@ -98,109 +173,85 @@ caf::actor fetch_indexer(table_indexer& tbl, const data_extractor& dx,
...
-        [](const auto&, const auto&) {
-          return caf::actor{}; // clang-format fix
-        });
-      auto hdl = caf::visit(v, pred.lhs, pred.rhs);
-      if (hdl != nullptr) {
-        triples.emplace_back(kvp.first, curried(pred), std::move(hdl));
-      }
+evaluation_triples partition::eval(const expression& expr) {
+  evaluation_triples result;
+  // Step #1: Pretend the partition is a table.
```
mavam, line 229: I was looking for `Step #2` somewhere. Does it exist? Otherwise the comment is misleading.

libvast/src/system/partition.cpp
```diff
@@ -98,109 +173,85 @@ caf::actor fetch_indexer(table_indexer& tbl, const data_extractor& dx,
...
+      },
+      [&](const data_extractor& dx, const data& x) {
+        return get_indexer_handle(dx, x);
+      },
+      [](const auto&, const auto&) {
+        return caf::actor{}; // clang-format fix
+      });
+    auto hdl = caf::visit(v, pred.lhs, pred.rhs);
+    if (hdl != nullptr) {
+      result.emplace_back(kvp.first, curried(pred), std::move(hdl));
```
mavam, line 249:
```suggestion
    if (auto hdl = caf::visit(v, pred.lhs, pred.rhs))
      result.emplace_back(kvp.first, curried(pred), std::move(hdl));
```

I couldn't get github to include the brace below.

libvast/test/expression.cpp
```diff
@@ -23,7 +23,7 @@
 #include "vast/concept/parseable/vast/time.hpp"
 #include "vast/concept/printable/to_string.hpp"
 #include "vast/concept/printable/vast/expression.hpp"
-#include "vast/detail/steady_map.hpp"
+#include "vast/detail/stable_map.hpp"
```
mavam, line 26: Seems like this change is already merged to master. Could you rebase?

libvast/vast/qualified_record_field.hpp
```diff
@@ -0,0 +1,65 @@
...
+#include <vector>
+
+namespace vast {
+
+/// A standalone field of an event type, used to address an index column.
+/// Example: { "zeek.conn.id.orig_h", address_type{} }
+struct qualified_record_field
+  : detail::totally_ordered<qualified_record_field> {
+  qualified_record_field(std::string name_, type type_)
+    : name{std::move(name_)}, type{std::move(type_)} {
```
mavam, line 32:
```suggestion
  qualified_record_field(std::string type_name, type field_type)
    : name{std::move(type_name)}, type{std::move(field_type)} {
```

Trailing underscores give the intuition of member variables, which is confusing as argument names.

A proposal: the constructor takes actually the record type name and performs the string concatenation as part of the construction. Then the we can call the `name` member `fully_qualified_name` and everything becomes much clearer.

libvast/vast/qualified_record_field.hpp
```diff
@@ -0,0 +1,65 @@
...
+    // nop
+  }
+
+  vast::record_field to_record_field() const;
+
+  std::string name; ///< The name of the field.
```
mavam, line 38:
```suggestion
  std::string fqn; ///< The field name prepended with the record type name.
```

Or `fully_qualified_name` for more expressiveness.

libvast/vast/qualified_record_field.hpp
```diff
@@ -0,0 +1,65 @@
...
+  operator<(const qualified_record_field& x, const qualified_record_field& y);
+
+  template <class Inspector>
+  friend auto inspect(Inspector& f, qualified_record_field& x) {
+    return f(x.name, x.type);
+  }
+};
+
+qualified_record_field
+to_fully_qualified(const std::string& tn, const record_field& field);
```
mavam, line 54: I don't understand why we need this function. Isn't the idea of `qualified_record_field` to contain the FQN?

libvast/vast/system/index.hpp
```diff
@@ -83,7 +83,7 @@ struct index_state {
 
   /// Stores evaluation metadata for pending partitions.
   using pending_query_map
-    = caf::detail::unordered_flat_map<uuid, evaluation_map>;
+    = caf::detail::unordered_flat_map<uuid, evaluation_triples>;
```
mavam, line 86: Can we change the type of `unordered_flat_map` here as well? Or is there a clear benefit for using it?

libvast/vast/system/indexer_stage_driver.hpp
```diff
@@ -13,31 +13,56 @@
...
+///  }
+///  type bar = record {
+///     a:          int,       # A
+///     b:          domain,    # B'
+///     c:          enum,      # C'
+///     d:          address    # D
+///  }
+///
+///  inbound stream
+///        |                              table_slice{ bar }
```
mavam, line 57:
```suggestion
///        |                              table_slice{ foo }
```

How two `foo` types map to different indexers still needs explanation.

libvast_test/src/actor_system.cpp
```diff
@@ -33,6 +34,14 @@ test_configuration::test_configuration() {
     vast::rm(log_file);
 }
 
+caf::error test_configuration::parse(int argc, char** argv) {
+  for (int i = 0; i < argc; i++) {
+    std::cerr << "argument : " << argv[i] << '\n';
```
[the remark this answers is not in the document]
tobim: This was not intended to stay until the merge. It has been removed.

libvast/vast/system/index_common.hpp
```diff
@@ -0,0 +1,66 @@
...
+  friend bool operator<(const fully_qualified_leaf_field& x,
+                        const fully_qualified_leaf_field& y);
+
+  template <class Inspector>
+  friend auto inspect(Inspector& f, fully_qualified_leaf_field& x) {
+    return f(x.name, x.type);
+  }
+};
+
+struct slice_column {
```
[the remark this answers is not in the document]
tobim: The type is now in its own file, but I did not change it's API yet. I think this should rather be done in a different PR, because it will also affect the `column_index`.

libvast/vast/system/index_common.hpp
```diff
@@ -0,0 +1,66 @@
...
+  table_slice_ptr slice;
+  size_t column;
+};
+
+/// Bundles an offset into an expression under evaluation to the curried
+/// representation of the ::predicate at that position in the expression and
+/// the INDEXER actor responsible for answering the (curried) predicate.
+using evaluation_triple = std::tuple<offset, curried_predicate, caf::actor>;
+
+using evaluation_triples = std::vector<evaluation_triple>;
```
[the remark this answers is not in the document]
tobim: I had to think about that for a while, but in the end, I think I disagree on that one.

Normally, a reader would ask for the definition of `evaluation_tripe`. The plural suffix should not raise a question mark, as it is pretty clear that it means `some_container<evaluation_tripe>`, but the type of container is not interesting at the high level.

libvast/vast/system/partition.hpp
```diff
@@ -147,10 +210,22 @@ class partition {
   /// Stores one table indexer per layout that in turn manages INDEXER actors.
   table_indexer_map table_indexers_;
 
+  /// A map to the indexers
+  caf::detail::unordered_flat_map<fully_qualified_leaf_field, indexer_path>
+    indexers_;
+
+  /// Instrumentation data store, one entry for each INDEXER.
+  std::unordered_map<size_t, atomic_measurement> measurements_;
```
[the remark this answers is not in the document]
tobim: Maybe yes, but the measurements used to reside in the `table_indexer` before the refactoring, so putting them into the partition was logical. I think we should rather move them in a separate PR if we think that is necessary.

libvast/src/system/indexer_downstream_manager.cpp
```diff
@@ -0,0 +1,208 @@
...
+size_t indexer_downstream_manager::buffered() const noexcept {
+  // We sum up the buffers for all partiitons.
+  size_t result = 0;
+  for (auto p : partitions)
+    result += buffered(*p);
+  return result;
+}
+
+size_t indexer_downstream_manager::buffered(partition& p) const noexcept {
+  // We have a central buffer, but also an additional buffer at each path. We
```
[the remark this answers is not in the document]
tobim: The `inbound_` container within the partition is meant, I was hoping this would be clear.
What about "partition level buffer"?

libvast/vast/qualified_record_field.hpp
tobim: I'm don't think I understand the question. This function's purpose is to turn a regular record field into a qualified one by prepending the layout's name to the field name. It could be a constructor, but then again, id does not have to be one.

libvast/src/system/partition.cpp
tobim: It is a `unique_ptr<outbout_path>`, so copy is a no.

libvast/src/system/partition.cpp
tobim: The comment has been refined.

libvast/vast/system/indexer_stage_driver.hpp
mavam: We can just remove this line, and also change `#` to `//` below.

libvast/vast/qualified_record_field.hpp
tobim:
```
qualified_record_field(std::string layout_name, record_field field);
```

libvast/src/system/partition.cpp
tobim: This was an oversight and has since been resolved.

libvast/src/system/indexer_downstream_manager.cpp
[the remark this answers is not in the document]
tobim: ... the buffered table slices in the index, but also...

libvast/src/system/indexer_stage_driver.cpp
```diff
@@ -41,61 +39,20 @@ indexer_stage_driver::~indexer_stage_driver() noexcept {
...
-      VAST_ASSERT(st.active->layouts().empty());
-    } else {
-      st.active->reduce_capacity(slice_size);
-    }
+    // Select the destination.
+    auto part = st.get_or_add_partition(slice);
+    // Inform the meta index.
+    st.meta_idx.add(part->id(), *slice);
+    // Ship it to the partition.
+    part->add(std::move(slice));
```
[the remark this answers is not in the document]
tobim: Will remove.

libvast/vast/system/index.hpp
tobim: Will do.

libvast_test/src/main.cpp
```diff
@@ -94,6 +94,8 @@ int main(int argc, char** argv) {
       break;
     }
   }
+  char* argv_[] = {"--logger.file-verbosity=trace"};
+  caf::test::engine::args(1, argv_);
```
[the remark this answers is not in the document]
tobim: Sorry about the long silence here. This has been removed.

libvast/vast/system/indexer_stage_driver.hpp
tobim: We changed the example so that both types are `foo`, but the second one has move precise field types.

libvast/vast/system/partition.hpp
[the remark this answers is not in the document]
tobim: We converged on doing this later.

libvast/vast/system/partition.hpp
```diff
@@ -42,10 +50,45 @@ class partition {
...
 
+    /// Maps type names to ids.
+    caf::detail::unordered_flat_map<std::string, ids> type_ids;
+
     /// Stores whether we modified `types` after loading it.
     bool dirty = false;
   };
 
+  /// A path that connects the incoming stream of table slices to an indexer.
+  struct indexer_path {
```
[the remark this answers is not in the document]
tobim: I used `wrapped_indexer` instead, because `<actor_name>_state` is a common idiom in our code base and has a different meaning.

libvast/src/qualified_record_field.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "vast/qualified_record_field.hpp"
+
+#include "vast/concept/hashable/uhash.hpp"
+#include "vast/concept/hashable/xxhash.hpp"
+
+namespace vast {
+
+vast::record_field qualified_record_field::to_record_field() const {
+  return {fqn, type};
+}
```
mavam, line 23: This is a good candidate for a free function that fit in our `as_*` syntax.

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🔄 The index specific options `max-partition-size`, `max-resident-partitions`,
+  `max-taste-partitions`, and `max-queries` can now be specified on the command
+  line when starting a Node.
```
mavam, line 15: I don't think we ever used CamelCase to denote components.

```suggestion
  line when starting a node.
```

libvast/src/system/index.cpp
```diff
@@ -284,7 +284,8 @@ void index_state::reset_active_partition() {
   // Persist meta data and the state of all INDEXER actors when the active
   // partition gets replaced becomes full.
   if (active != nullptr) {
-    stage->out().unregister(active.get());
+    if (!stage->out().unregister(active.get()))
+      VAST_ERROR(self, "tried to unregister unknown partition");
```
mavam, line 288: How can this be a system error? I looks like a programming error if this occurs where an assertion would be in place. Otherwise there needs to be a clear comment why this internal logic should be propagated to the user.

libvast/src/system/indexer_downstream_manager.cpp
```diff
@@ -115,15 +121,15 @@ void indexer_downstream_manager::cleanup_partition(set_type::iterator& it) {
...
         VAST_WARNING(self(), "tries to delete a non-existing outbound path");
         // We could just exit now, but maybe something unforseen happened to
-        // that one INDEXER while the rest of the partiton is ok, so we rather
+        // that one INDEXER while the rest of the partiton is ok, so we better
         // try to clean up the others.
         f = paths_.begin();
         continue;
       }
     }
+    about_to_erase(f->second.get(), false, nullptr);
```
mavam, line 130: We usually have verbs for function names. This sounds like a property, so the reader doesn't know what the function is *doing*. Here's a suggestion for what it *could* be doing based on reading the call site:

```suggestion
    schedule_removal(f->second.get(), false, nullptr);
```

libvast/src/system/spawn_index.cpp
```diff
@@ -32,11 +32,12 @@ maybe_actor spawn_index(node_actor* self, spawn_arguments& args) {
...
-                            opt("max-events", sd::max_partition_size),
-                            opt("max-parts", sd::max_in_mem_partitions),
-                            opt("taste-parts", sd::taste_partitions),
-                            opt("max_queries", sd::num_query_supervisors));
+  auto result = self->spawn(
+    index, args.dir / args.label,
+    opt("system.max-partition-size", sd::max_partition_size),
+    opt("system.max-resident-partitions", sd::max_in_mem_partitions),
+    opt("system.max-taste-partitions", sd::taste_partitions),
+    opt("system.max-queries", sd::num_query_supervisors));
```
mavam, line 40: This might be my misunderstanding of the option processing, and maybe @dominiklohmann can fill me in, but why don't we have hierarchical options here, such as `system.index.max-partition-size`?

dominiklohmann: They should be `spawn.index.max-partition-size` to correspond to the `vast spawn index` command.

mavam: Okay. @tobim mind adjusting this accordingly?

dominiklohmann: I think we should talk about this first. As is right now, nobody is using the `vast spawn index` command directly. Never.

It supports these options (also available via `vast spawn index help`):

```cpp
spawn->add_subcommand(
  "index", "creates a new index", "",
  opts()
    .add<size_t>("max-events,e", "maximum events per partition")
    .add<size_t>("max-parts,p", "maximum number of in-memory partitions")
    .add<size_t>("taste-parts,t", "number of immediately scheduled "
                                  "partitions")
    .add<size_t>("max-queries,q", "maximum number of concurrent queries"));
```

As nobody is ever invoking these commands directly, they _might_ actually make more sense in the `system` category, which is what we use for options that are placed immediately after the root command.

libvast/vast/system/instrumentation.hpp
```diff
@@ -94,6 +94,17 @@ struct atomic_measurement : public measurement {
 using atomic_measurement = std::atomic<measurement>;
 #endif
 
+measurement collect(atomic_measurement& am) {
```
dominiklohmann, line 97:
```suggestion
inline measurement collect(atomic_measurement& am) {
```

libvast/src/system/index.cpp
```diff
@@ -237,27 +236,25 @@ void index_state::send_report() {
...
-          double rate = tmp.events * 1'000'000'000.0 / tmp.duration.count();
-          if (rate < min_rate) {
-            min_rate = rate;
-            min = tmp;
-          }
-          if (rate > max_rate) {
-            max_rate = rate;
-            max = tmp;
-          }
+      if (tmp.events > 0) {
```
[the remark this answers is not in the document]
tobim: I moved the locking related stuff into `instrumentation.hpp`.

libvast/src/system/spawn_index.cpp
dominiklohmann: Following our discussion in Slack, these should just stay in the `system` category for now. Resolving this.

tobim: I changed these options in one of the commits in here to `system.*` (with more expressive names too).

The problem with `system.index.max-partition-size` or `spawn.index.max-partition-size` is that the command line parser does not accept fully qualified options at the moment. Maybe it is desirable to change that.

libvast/src/system/indexer_downstream_manager.cpp
tobim: This is a function from CAF that needs to be called. Nothing I can change.

libvast/src/system/index.cpp
tobim: Correctly observed. It has been changed into an assertion.

tobim:
The recent round of comments have been addressed or replied to.

libvast/src/system/index.cpp
```diff
@@ -87,6 +85,8 @@ index_state::index_state(caf::stateful_actor<index_state>* self)
 
 index_state::~index_state() {
   VAST_VERBOSE(self, "tearing down");
+  if (active != nullptr)
+    stage->out().unregister(active.get());
```
mavam, line 89: Should the result of `unregister` also be check in the same was as below?

libvast/vast/table_slice_column.hpp
```diff
@@ -0,0 +1,58 @@
...
+using partition_ptr = std::unique_ptr<partition>;
+
+} // namespace system
+
+struct table_slice_column {
+  table_slice_column() {
+  }
+
+  table_slice_column(table_slice_ptr slice_, size_t column_)
+    : slice{std::move(slice_)}, column{column_} {
```
mavam, line 39: I would prefer not to use `_` as suffix only for member variables.

```suggestion
  table_slice_column(table_slice_ptr slice, size_t col)
    : slice{std::move(slice)}, column{col} {
```

Does this trigger a warning by your compiler? For me it just works.

libvast/test/system/indexer.cpp
```diff
@@ -54,7 +54,12 @@ struct fixture : fixtures::deterministic_actor_system_and_events {
     VAST_ASSERT(std::all_of(slices.begin(), slices.end(), [&](auto& slice) {
       return slice->layout() == layout;
     }));
-    vast::detail::spawn_container_source(sys, std::move(slices), indexer);
+    std::vector<table_slice_column> slice_columns;
+    for (auto& slice : slices) {
+      slice_columns.push_back(table_slice_column{slice, 0u});
+    }
```
mavam, line 60:
```suggestion
    for (auto& slice : slices)
      slice_columns.push_back(table_slice_column{slice, 0u});
```

libvast/vast/table_slice_column.hpp
tobim: It does work for me as well. I'm bit surprised.

PR #729, opened 2020-01-29
Fix check whether expected<T> holds a value

dominiklohmann:
This was missed in #724, since the bug was introduced on master after branching that PR off.

### DIFF
--- a/libvast/vast/value_index.hpp
+++ b/libvast/vast/value_index.hpp
@@ -237,7 +237,7 @@ public:
       } else {
         auto str = caf::get<caf::config_value::string>(i->second);
         auto b = to<base>(str);
-        VAST_ASSERT(b != nullptr); // pre-condition is that this was validated
+        VAST_ASSERT(b); // pre-condition is that this was validated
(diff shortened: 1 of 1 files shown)


dominiklohmann:
I have dismissed @lava's approval.

We just noticed that the code is not covered by unit or integration tests, the branch is never called. I have verified this by setting a breakpoint at `libvast/vast/value_index.hpp:240` and running both test suites. 

@mavam you just recently added the code in question—do you mind giving this a review?

libvast/vast/value_index.hpp
```diff
@@ -237,7 +237,7 @@ class arithmetic_index : public value_index {
       } else {
         auto str = caf::get<caf::config_value::string>(i->second);
         auto b = to<base>(str);
-        VAST_ASSERT(b != nullptr); // pre-condition is that this was validated
+        VAST_ASSERT(b.engaged()); // pre-condition is that this was validated
```
mavam:
```suggestion
        VAST_ASSERT(b); // pre-condition is that this was validated
```

You could also rely on the boolean conversion operator. This is perhaps more idiomatic.

PR #735, opened 2020-01-30
Dont discard backslash characters from qqstr

lava:


libvast/vast/concept/parseable/string/quoted_string.hpp
```diff
@@ -27,13 +27,16 @@ class quoted_string_parser : public parser<quoted_string_parser<Quote, Esc>> {
...
   using attribute = std::string;
 
+  // clang-format off
   static constexpr auto esc = ignore(parsers::ch<Esc>);
   static constexpr auto quote = ignore(parsers::ch<Quote>);
-  static constexpr auto esc_esc = esc >> parsers::ch<Esc>;
   static constexpr auto esc_quote = esc >> parsers::ch<Quote>;
   static constexpr auto str_chr
-    = esc_esc | esc_quote | (parsers::print - quote);
+    = (parsers::print - quote - parsers::ch<Esc>)
```
tobim: can we add `static constexpr auto esc = parsers::ch<Esc>;` and rename the old `esc` to `ignore_esc`?

libvast/test/parseable.cpp
```diff
@@ -228,7 +228,7 @@ TEST(quoted string - missing trailing quote after escaped quote) {
 TEST(quoted string - trailing quote after escaped escape) {
   std::string attr;
   CHECK(parsers::qstr("'foobar\\\\'", attr));
-  CHECK_EQUAL(attr, "foobar\\");
+  CHECK_EQUAL(attr, "foobar\\\\");
```
lava: After our discussions yesterday, I think that this test case was incorrect. But please double-check @tobim

libvast/vast/concept/parseable/string/quoted_string.hpp
```diff
@@ -22,22 +22,51 @@
...
 
   template <class Iterator, class Attribute>
   bool parse(Iterator& f, const Iterator& l, Attribute& x) const {
-    return quoted_str(f, l, x);
+    // clang-format off
+    if constexpr (std::is_same_v<unused_type, Attribute>) {
+      auto esc_esc = ignore_esc >> ignore_esc;
+      auto str_chr = esc_quote | esc_esc | (parsers::print - quote);
+      auto quoted_str = ignore_quote >> *str_chr >> ignore_quote;
+      return quoted_str(f, l, x);
```
mavam:
```suggestion
      return quoted_str(f, l, unused);
```

This is totally optional (and identical), but may communicate to the reader what's happening downstream.

libvast/vast/concept/parseable/string/quoted_string.hpp
```diff
@@ -22,22 +22,51 @@
...
+    } else {
+      // The kleene_parser will append exactly one char per invocation to the
+      // output string, so if we returned nothing from this lambda it would
+      // append a default-initialized null byte.
```
mavam: Wow, great you hunted this down!

libvast/vast/concept/parseable/string/quoted_string.hpp
```diff
@@ -22,22 +22,51 @@
...
+///
+/// For example, using backslash as escape character and double quote as quote
+/// character:
+///
+///    "foo\n"     denotes the string ['f', 'o', 'o', '\', 'n']
+///    "C:\sys32"  denotes the string ['C', ':', '\', 's', 'y', 's', '3', '2']
+///    "\\"        denotes the string ['\', '\']
+///    "\""        denotes the string ['"']
+///    "\\""       is an invalid string (junk after end of string)
+///    "\\\"       is an invalid string (unterminated)
```
dominiklohmann: `@note` can be used to highlight a Doxygen comment as a notice. Otherwise it will appear as the class description for `quoted_string_parser`, which doesn't really fit.

libvast/test/parseable.cpp
tobim: You are right of course.

libvast/vast/concept/parseable/string/quoted_string.hpp
```diff
@@ -22,22 +22,51 @@
...
+/// character are interpreted as character literals.
+///
+/// For example, using backslash as escape character and double quote as quote
+/// character:
+///
+///    "foo\n"     denotes the string ['f', 'o', 'o', '\', 'n']
+///    "C:\sys32"  denotes the string ['C', ':', '\', 's', 'y', 's', '3', '2']
+///    "\\"        denotes the string ['\', '\']
+///    "\""        denotes the string ['"']
+///    "\\""       is an invalid string (junk after end of string)
```
tobim: This is actually valid. The quoted string parser by itself should not mandate that the end of input has to be reached, otherwise it would not be possible to sequence it.

libvast/vast/concept/parseable/string/quoted_string.hpp
```diff
@@ -22,22 +22,51 @@
...
+  static constexpr auto ignore_esc = ignore(parsers::ch<Esc>);
+  static constexpr auto ignore_quote = ignore(parsers::ch<Quote>);
+  static constexpr auto esc_quote = ignore_esc >> quote;
 
   template <class Iterator, class Attribute>
   bool parse(Iterator& f, const Iterator& l, Attribute& x) const {
-    return quoted_str(f, l, x);
+    // clang-format off
+    if constexpr (std::is_same_v<unused_type, Attribute>) {
+      auto esc_esc = ignore_esc >> ignore_esc;
```
tobim: You can move the compile time branch into the definition of `esc_esc`:
```c++
    auto esc_esc = [&]() {
      // clang-format off
      if constexpr (std::is_same_v<unused_type, Attribute>)
        return ignore_esc >> ignore_esc;
      else
        // The kleene_parser will append exactly one char per invocation to the
        // output string, so if we returned nothing from this lambda it would
        // append a default-initialized null byte.
        return (ignore_esc >> ignore_esc) ->* [&] { x.push_back(Esc); return Esc; };
      // clang-format on
    }();
```
And avoid the duplication of the following 3 lines.

mavam: Neat workaround for the new scope introduced by `if constexpr`!

lava:
[ch12457]

libvast/vast/concept/parseable/string/quoted_string.hpp
lava: Thanks, adding a watchpoint was a great suggestion, by putting it on the string size member of the parsed string you can effectively break every time a parser pushes a new character.

PR #738, opened 2020-01-31
Annotate schemas with #index=hash attribute

dominiklohmann:
@mavam there are likely a lot I've missed, this was just from a quick grep in the source tree. We can always add these later on when we encounter them. For some fields we can obviously restrict the size of the hash as well, but that'd require an in-depth investigation into the fields allowed contents.

For the schema files, I've used mostly `:g/^\s\+field_name:/s/,\?$/ #index=hash&` to do the replacement.


dominiklohmann:
[story/ch7210]

PR #740, opened 2020-01-31
Add a nix package expression for VAST

tobim:


CMakeLists.txt
```diff
@@ -494,8 +494,30 @@ if (APPLE)
   endif ()
 endif ()
 
-if (NOT CAF_ROOT_DIR AND EXISTS
-                         ${CMAKE_CURRENT_SOURCE_DIR}/aux/caf/CMakeLists.txt)
+# Try to find the required CAF components first...
+find_package(
+  CAF 0.17.3 EXACT
```
dominiklohmann: This behavior change needs a changelog entry.

Do we need a place where we document how to update to a new version of CAF?

PR #742, opened 2020-01-31
Respect user-provided build type settings

lava:
Don't touch CMAKE_CXX_FLAGS_* if they were manually specified
by the user, either using `cmake -D` or in the `cmake-gui`.

Also, removed a reference to the VAST_DEV_MODE setting, which
doesn't exist anymore.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,16 @@ list(GET version_numbers 1 VERSION_MONTH)
 list(GET version_numbers 1 VERSION_DAY)
 unset(version_numbers)
 
+set(CMAKE_CXX_FLAGS_DEBUG
+    "_UNSET"
+    CACHE STRING "")
+set(CMAKE_CXX_FLAGS_RELEASE
+    "_UNSET"
+    CACHE STRING "")
+set(CMAKE_CXX_FLAGS_RELWITHDEBINFO
+    "_UNSET"
+    CACHE STRING "")
+
 project(
   VAST
   VERSION ${VERSION_YEAR}.${VERSION_MONTH}.${VERSION_DAY}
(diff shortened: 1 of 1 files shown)


lava:
Hm, since we're discussing this anyways, here's what I was wondering about while writing this:

 - Would be a good idea to move these settings into a separate file (i.e. option 3 in https://stackoverflow.com/a/32956785/92560), or if it's better for discoverability to keep everything in one file
- Would it make sense to have a "CI" build type? (which would be something like optimizations enabled + assertions enabled + debug logs + minimal debug symbols)

Any thoughts, @dominiklohmann @tobim ?

dominiklohmann:
I think it's best to embrace modern CMake and remove any mentions of `CMAKE_<LANG>_FLAGS` altogether. By default, `CMAKE_CXX_FLAGS` are composed of the user environments `CXXFLAGS` and `CPPFLAGS`, and the linker also respects both `LDFLAGS` and `ARFLAGS` automatically.

Instead, we should write our own flags variables, and add them to the targets using `target_compile_options` while also respecting the `PUBLIC`, `INTERFACE`, and `PRIVATE` scopes (E.g., warnings are in the private scope).

> * Would it make sense to have a "CI" build type? (which would be something like optimizations enabled + assertions enabled + debug logs + minimal debug symbols)

That's what we have a configure script.

lava:
The modern cmake style looks good, but does it already respect user overwrites by default or would we have to add some scaffolding around that example?

> That's what we have a configure script.

I only see "debug" and "release" mode in the configure script (and they forward to cmake, so i imagine we'd have to introduce a new build type here first, before adding support to the configure script)

dominiklohmann:
> The modern cmake style looks good, but does it already respect user overwrites by default or would we have to add some scaffolding around that example?

User overwrites of `CPPFLAGS`, `CXXFLAGS`, et al. go into `CMAKE_<LANG>_FLAGS` automatically, and as long as our CMake code doesn't touch these we are good to go. For any `<LANG>`, `CMAKE_<LANG>_FLAGS` is always added to the compiler invocations.

> I only see "debug" and "release" mode in the configure script (and they forward to cmake, so i imagine we'd have to introduce a new build type here first, before adding support to the configure script)

Let's discuss this when I'm back in the office tomorrow. I think that's a separate story.

lava:
After some local testing, it seems like the order of flags is `${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_DEBUG} <flags from set_target_compile_options>`, with CXXFLAGS from the environment being ignored:

```
benno@poincare:~/code/cmake/build$ cat ../CMakeLists.txt 
cmake_minimum_required(VERSION 3.13)
project(foo)

set(MY_DEBUG_OPTIONS "--target-compile-options")

add_executable(foo foo.cpp)
target_compile_options(foo 
    PUBLIC  "$<$<CONFIG:Debug>:${MY_DEBUG_OPTIONS}>"
            "$<$<CONFIG:RelWithDebInfo>:${MY_DEBUG_OPTIONS}>"
            "$<$<CONFIG:Release>:${MY_RELEASE_OPTIONS}>"
            "$<$<CONFIG:MinSizeRel>:${MY_RELEASE_OPTIONS}>"
    PRIVATE "${MY_WARNINGS}")

benno@poincare:~/code/cmake/build$ CXXFLAGS=--cxxflags cmake .. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_FLAGS=--cmake_cxx_flags
[...]
benno@poincare:~/code/cmake/build$ make VERBOSE=1
[...]
[ 50%] Building CXX object CMakeFiles/foo.dir/foo.cpp.o
/usr/bin/c++    --cmake_cxx_flags -g   --target_compile_options -o CMakeFiles/foo.dir/foo.cpp.o -c /home/benno/code/cmake/foo.cpp
```

dominiklohmann:
You can add `BEFORE` to `target_compile_options` to prepend options instead of appending them. See https://cmake.org/cmake/help/latest/command/target_compile_options.html.

```cmake
target_compile_options(foo BEFORE
    PUBLIC  "$<$<CONFIG:Debug>:${MY_DEBUG_OPTIONS}>"
            "$<$<CONFIG:RelWithDebInfo>:${MY_DEBUG_OPTIONS}>"
            "$<$<CONFIG:Release>:${MY_RELEASE_OPTIONS}>"
            "$<$<CONFIG:MinSizeRel>:${MY_RELEASE_OPTIONS}>"
    PRIVATE "${MY_WARNINGS}")
```

lava:
The `BEFORE` doesn't make a difference in this case (i guess it only orders the compile flags added by `target_compile_options()`?)

However, the other problem is that (assuming we want to keep the current set of flags) we'd have to set the default value `MY_DEBUG_OPTIONS=-g1` to override it, and then also turn it into a public setting so the user can select a custom value.

Then a user has the option of setting `CMAKE_CXX_FLAGS=-g3` (no effect), setting `CMAKE_CXX_FLAGS_DEBUG=-g3` (no effect) or setting `MY_DEBUG_OPTIONS=-g3` (desired effect), which seems to be a bit confusing.

I think the ideal solution would be if we could change the order of the flags so that we get `<build type flags> <target compile options> <cxxflags>`, but I don't know if cmake supports that.

lava:
Closing this for now; the workaround to having custom debug flags is to also specify a custom build type e.g. `CMAKE_CXX_BUILD_TYPE=Debug_Full CMAKE_CXX_FLAGS=-ggdb3`.

PR #743, opened 2020-01-31
Use short-circuiting in template metaprogramming 

dominiklohmann:
Template machinery short-circuits only when a backtracking step is done. This is why C++17 ships `std::conjunction`, `std::disjunction`, and `std::negation`.

This change makes VAST compile faster, because less template instantiations are needed.

I've only done this on the basis of a grep for `enable_if` so far.

### DIFF
--- a/libvast/vast/coder.hpp
+++ b/libvast/vast/coder.hpp
@@ -648,10 +648,9 @@ private:
   template <class C>
   auto decode(const std::vector<C>& coders, relational_operator op,
               value_type x) const
-  -> std::enable_if_t<
-    is_equality_coder<C>{} || is_bitslice_coder<C>{},
-    bitmap_type
-  > {
+    -> std::enable_if_t<
+      std::disjunction_v<is_equality_coder<C>, is_bitslice_coder<C>>,
+      bitmap_type> {
     VAST_ASSERT(op == equal || op == not_equal);
     base_.decompose(x, xs_);
     auto result = coders[0].decode(equal, xs_[0]);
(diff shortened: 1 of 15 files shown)


dominiklohmann:
> I wonder how much compile time does this save, since it looks like for every instantiation that is saved due to short-circuiting there is an additional instantiation of `std::{con,dis}junction_v` adde

A lot. Let me convince you with a simple example:

```cpp
// -- modern version of is_any_of

template <class T, class U, class... Us>
using is_any_of = std::disjunction<std::is_same<T, U>, std::is_same<T, Us>...>;

template <class T, class U, class... Us>
inline constexpr auto is_any_of_v = is_any_of<T, U, Us...>::value;

// -- old version of is_any_of

template <class T, class... Us>
inline constexpr auto is_any_of_v = (std::is_same_v<T, Us> || ...);
```

For the new version, not all template arguments need to be instantiated. Once the comparison is false for a single value, it short-circuits and returns false. 

For the old version, all template arguments need to be instantiated first, then the disjunction is evaluated.

`Us...` may just be a simple type, which is when this makes no big difference. But consider our parser types, which are sometimes a few hundred lines of nested templates in debug output. For those types, saving an instantiation is a big deal.

> `std::negation` only takes a single template argument, so it seems to be a bit out of place in the discussion of short-circuiting behavior

`std::negation` is absolutely related to all of this, as there is no other way to invert a meta type without instantiating it. Notice that `std::negation` operates on the types themselves, not their underlying values (`type_trait_v` or `type_trait::value` or `type_trait()`).

```cpp
// Error: Syntax is invalid
// template <class T, class U, class... Us>
// using is_none_of = !<std::disjunction<std::is_same<T, U>, std::is_same<T, Us>...>>;

template <class T, class U, class... Us>
using is_none_of =  std::negation<std::disjunction<std::is_same<T, U>, std::is_same<T, Us>...>>;
```

PR #746, opened 2020-02-04
Make `CHECK_EQUAL()` behave the same as `operator==`

lava:
The `CHECK_EQUAL()` macro has a lot of additional behavior inherited from caf, most significantly the ability to compare types that don't define `operator==` (returning a configurable default value), and the automatic unboxing of variant types.

This PR simplifies the behavior by having `CHECK_EQUAL(x, y)` return `true` iff `x == y`.

### DIFF
--- a/libvast_test/vast/test/test.hpp
+++ b/libvast_test/vast/test/test.hpp
@@ -22,6 +22,52 @@
 
 #include <caf/test/unit_test.hpp>
 
+namespace vast::test::detail {
+
+struct equality_compare {
+  template <class T1, class T2>
+  bool operator()(const T1& t1, const T2& t2) {
+    return t1 == t2;
+  }
+};
+
+struct inequality_compare {
+  template <class T1, class T2>
+  bool operator()(const T1& t1, const T2& t2) {
+    return t1 != t2;
+  }
+};
+
+struct greater_compare {
+  template <class T1, class T2>
+  bool operator()(const T1& t1, const T2& t2) {
+    return t1 > t2;
+  }
+};
+
+struct greater_equal_compare {
+  template <class T1, class T2>
+  bool operator()(const T1& t1, const T2& t2) {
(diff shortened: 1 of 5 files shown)


libvast_test/vast/test/test.hpp
```diff
@@ -39,21 +50,29 @@
...
 #define REQUIRE_GREATER_EQUAL CAF_REQUIRE_GREATER_EQUAL
 #define CHECK CAF_CHECK
-#define CHECK_EQUAL CAF_CHECK_EQUAL
-#define CHECK_NOT_EQUAL CAF_CHECK_NOT_EQUAL
-#define CHECK_LESS CAF_CHECK_LESS
-#define CHECK_LESS_EQUAL CAF_CHECK_LESS_OR_EQUAL
-#define CHECK_GREATER CAF_CHECK_GREATER
-#define CHECK_GREATER_EQUAL CAF_CHECK_GREATER_OR_EQUAL
+#define CHECK_EQUAL(x, y)                                                      \
+  CAF_CHECK_FUNC(vast::test::detail::equality_compare, x, y)
```
lava: This variant looked the best to me, but I wanted to get a design review first before creating comparison structs for all the other operators and do the same for `REQUIRE_*`

dominiklohmann: I think this reflects the expected behavior best. 

Can you add parentheses around the macro arguments?

```
#define REQUIRE_EQUAL(x, y) REQUIRE((x) == (y))
```

tobim: Will this preserve the introspection behavior in case of a mismatch? If yes, then this is the way to go. The `REQUIRE_*` stuff will also need the same treatment.

lava: @tobim Yes, the output will be the same as before

tobim:
Thank you.

PR #749, opened 2020-02-06
Don't drop record attributes in CSV parser

lava:


libvast/src/format/csv.cpp
```diff
@@ -192,7 +192,7 @@ reader::make_layout(const std::vector<std::string>& names) {
         return record_type{std::move(result_raw)};
       };
       if (auto result = select_fields())
-        return result->name(r->name());
+        return result->name(r->name()).attributes(r->attributes());
```
mavam: Just wondering: is this something that should maybe happen within `select_fields`?

lava: I'm not sure; from the naming I was assuming that `select_fields` should operate on the individual fields of the record and the part below should operate on the record type itself, but I'm also happy to move it inside the function if you prefer.

mavam: Per our discussion, `select_attribute` should have "subset" semantics and therefore not alter any meta data aside from removing fields.

PR #750, opened 2020-02-06
Continuous import of Zeek logs in VAST is broken

tobim:


libvast/src/detail/line_range.cpp
```diff
@@ -28,18 +31,30 @@ const std::string& line_range::get() const {
...
+  if (!input_)
+    return;
   // Get the next non-empty line.
   while (line_.empty())
     if (std::getline(input_, line_))
       ++line_number_;
     else
       break;
+  // Clear error state if read failure was caused by a timeout
+  if (auto* p = dynamic_cast<fdinbuf*>(input_.rdbuf()))
```
tobim:
```suggestion
  if (!input_ && auto* p = dynamic_cast<fdinbuf*>(input_.rdbuf()))
```

libvast/vast/detail/line_range.hpp
```diff
@@ -36,6 +38,14 @@ class line_range : range_facade<line_range> {
 
   size_t line_number() const;
 
+  void set_read_timeout(std::chrono::milliseconds timeout);
+
+  void unset_read_timeout();
+
+  // Did the last call to `next()` fail due to a timeout failure?
+  // Can only return true if a timeout was set.
+  bool timed_out() const;
```
tobim: This API is a little bit too easy to misuse. I'd like to remove this function and add a variant of `next()` instead:
```
[[nodiscard]] bool next_timeout(std::chrono::milliseconds timeout);
```
Internally, I would set the timeout to the fdinbuf only if it is different from the last timeout. We can do so because we own `input_` exclusively.

We can also remove `set_read_timeout()` and `unset_read_timeout()` from the API.

libvast/src/detail/fdinbuf.cpp
```diff
@@ -11,28 +11,42 @@
...
  ******************************************************************************/
 
-#include <unistd.h>
+#include "vast/detail/fdinbuf.hpp"
+
+#include "vast/detail/assert.hpp"
 
 #include <cstdio>
 #include <cstring>
+#include <optional>
```
tobim: We usually don't repeat includes from the header in the implementation file.

libvast/vast/detail/line_range.hpp
lava: I agree, the `next_timeout()` API looks much nicer.

> Internally, I would set the timeout to the fdinbuf only if it is different from the last timeout. We can do so because we own input_ exclusively.

I'm not sure if this is actually an optimization, this will save a write but introduce an additional load and two comparisons. (not that it really matters tbh, because when we hit this path we actually introduce an additional syscall on the next read(), so a few instructions more or less will be undetectable)

libvast/test/format/zeek.cpp
```diff
@@ -165,25 +190,33 @@ std::string_view conn_log_100_events = R"__(#separator \x09
...
-                         std::string{input})};
+                       std::move(input)};
     std::vector<table_slice_ptr> slices;
     auto add_slice = [&](table_slice_ptr ptr) {
       slices.emplace_back(std::move(ptr));
     };
     auto [err, num] = reader.read(std::numeric_limits<size_t>::max(),
                                   slice_size, add_slice);
-    if (err != ec::end_of_input)
+    if (expect_eof && err != ec::end_of_input)
```
tobim: Slightly less code and slightly easier to read (?):
``` c++
if (err)
  if (!(expect_eof && err == ec::end_of_input))
    FAIL
```

libvast/src/format/zeek.cpp
```diff
@@ -276,8 +277,18 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
...
   while (produced < max_events) {
-    // Advance line range and check for EOF.
-    lines_->next();
+    // Advance line range and check for EOF. (In the first iteration, the
+    // current line is the last line of the header)
+    // If we already have some events, set a read timeout to ensure downstream
+    // sees them even if there are currently no further events in the stream.
+    if (builder_->rows() > 0) {
+      using namespace std::chrono_literals;
+      bool timeout = lines_->next_timeout(500ms); // TODO: Make it configurable.
```
tobim: Do you plan on resolving this within this PR?

libvast/src/format/zeek.cpp
```diff
@@ -276,8 +277,18 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
   size_t produced = 0;
   // Loop until reaching EOF or the configured limit of records.
   while (produced < max_events) {
-    // Advance line range and check for EOF.
-    lines_->next();
+    // Advance line range and check for EOF. (In the first iteration, the
+    // current line is the last line of the header)
+    // If we already have some events, set a read timeout to ensure downstream
```
tobim: Is this an optimization only or do we get wrong behavior if we call finish without adding any data beforehand?

libvast/src/detail/line_range.cpp
```diff
@@ -11,9 +11,12 @@
  * contained in the LICENSE file.                                             *
  ******************************************************************************/
 
-#include "vast/detail/assert.hpp"
 #include "vast/detail/line_range.hpp"
 
+#include "vast/detail/assert.hpp"
+#include "vast/detail/fdinbuf.hpp"
+#include "vast/logger.hpp"
```
tobim: Looks like a leftover from debugging.

libvast/src/format/zeek.cpp
```diff
@@ -21,6 +21,7 @@
 #include "vast/concept/printable/vast/view.hpp"
 #include "vast/detail/assert.hpp"
 #include "vast/detail/escapers.hpp"
+#include "vast/detail/fdinbuf.hpp"
```
tobim: This seems to be unused.

tobim:
BTW, this is looks almost finished to me.

libvast/src/format/zeek.cpp
lava: I can add it if you think it makes sense, but I also wouldn't feel too bad to leave it hard-coded for now since I can't really think of a reason for a user to change the setting, and it's not strictly related to fixing continous import of zeek logs.

libvast/src/format/zeek.cpp
lava: Yes, if the reader returns `0` the source will get stuck (see the TODO in source.hpp) I stopped my investigation into *why* exactly this happens, since I think @dominiklohmann already understands the bug.

But I think even without that bug it makes sense to keep this logic, for consistency with how we also block indefinitely when awaiting header lines.

dominiklohmann: It's a upstream bug in CAF that's causing this failure. Working around it is no small refactoring on our end. Continuous stream sources cannot be stopped when they've produced 0 results, because the done handler is not called when calling advance on the stream source. We're unable to yield control back to the parent actor, which is why the delayed_send approach mentioned in the comment won't work.

It's a good thing you've stopped investigating this. 😅

mavam: Since the source is unusable after it returns 0 once, how about just terminating the source so that we at least get an error?!

dominiklohmann: It isn't strictly unusable, it simply locks until new input is available. We can't know for sure at that point if we're stalling forever, or if new input will soon come.

Let's discuss this on the whiteboard at some point in the future. I've debugged this thoroughly just a few weeks back.

libvast/src/format/zeek.cpp
dominiklohmann: Maybe this can just be a constant in the defaults namespace. No need to add an option for it.

libvast/test/format/zeek.cpp
lava: I think that's missing the error case where we **expect** eof but the reader returns no error.

dominiklohmann: Does it return no error or none? 

That's an important distinction, because `if (err)` will succeed if `err` is `caf::no_error`, which is actually an error, but it won't succeed if `err` is `caf::none`, which means no error occurred.

dominiklohmann: (We just discussed this in person and opted to adapt the error messages instead to make a bit more sense.

mavam:
It feels like a changelog entry should document the bugfix.

PR #751, opened 2020-02-10
Update VAST for Apache Arrow 0.16

dominiklohmann:


dominiklohmann:
@tobim does this require any changes in the nix module?

dominiklohmann:
Note that the CI is running with a cached version of Homebrew. It'll take a day until Apache Arrow 0.16 is available in there, so we should just try to re-run the checks tomorrow.

dominiklohmann:
Because I was wondering why CI still doesn't have Arrow 0.16, I changed it so `brew --version` is ran before we install anything. Turns out that homebrew-core in CI has last been fetched on Feb 8th, so we'll just have to wait a bit more (we can force the update, but I don't really think we generally want to. Having a bit of time before CI breaks is usually a good thing).

PR #752, opened 2020-02-11
Let test target depend on the vast-test executable

tobim:
Due to a quirk in CTest, the `test` target is not a real target, and can not be used with `add_dependencies()` the workaround is to add a dedicated test that builds `vast-test` acts as a custom dependency for all other tests.

Caveat: The generator tools do not inherit the setting from the outer context, so both `Ninja` and `GNU Makefiles` build the dependencies of `test` with their default core counts, which are `NCORES + 2` and `1` respectively!


dominiklohmann:
This is not working. Simple test: Change some `libvast/test/*.cpp` file, then run `make test`. It does not re-run the `vast-test` target.

PR #755, opened 2020-02-13
Use checkout v2, refine workflow trigger

0ortmann:
Trigger `pull_request` uses the last merge commit on the GitHub ref branch as `GITHUB_SHA`: https://help.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-event-pull_request

- refine CI trigger to act on push to master and on pull_request
- use GitHub action checkout v2

### DIFF
--- a/.github/workflows/vast.yml
+++ b/.github/workflows/vast.yml
@@ -1,6 +1,13 @@
 name: "VAST"
 on:
   push:
+    branches:
+    - master
+    paths-ignore:
+    - '**.md'
+    - '!doc/**.md'
+  pull_request:
+    types: [opened, synchronize]
     paths-ignore:
     - '**.md'
     - '!doc/**.md'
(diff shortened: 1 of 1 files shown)


.github/workflows/vast.yml
```diff
@@ -13,9 +20,16 @@ jobs:
...
-      with:
-        ref: ${{ github.ref }}
+
+    - uses: actions/checkout@v2
+
+    - name: Fetch Submodules and Master
+      run: |
+        auth_header="$(git config --local --get http.https://github.com/.extraheader)"
+        git submodule sync --recursive
+        git -c "http.extraheader=$auth_header" -c protocol.version=2 submodule update --init --force --recursive
```
dominiklohmann: We shouldn't need to fetch the submodule content for the style check.

PR #756, opened 2020-02-13
Use a new build type tailored for CI

dominiklohmann:


.github/workflows/vast.yml
```diff
@@ -131,7 +130,7 @@ jobs:
             --build-dir="$BUILD_DIR" \
             --package-name="$PACKAGE_NAME" \
             ${{ matrix.os.extra-flags }} \
-            ${{ matrix.configure.flags }}
+            ${{ github.event.action == 'published' && matrix.configure.flags || matrix.configure.ci-flags }}
```
lava: Does this maybe need brackets? It looks like it built the current PR in release configuration with this conditional:
```
Build type:            Release
Build static:          no
Build static only:     no
Build static runtime:  no
Runtime checks:        no
Log level:             QUIET
```

dominiklohmann: What you're quoting is from the build summary of CAF.

lava: Ah, perfect.

PR #758, opened 2020-02-14
Allow for separating persistent state and log directories

dominiklohmann:
It's currently not possible to have different directories for the persistent state and the log files. 

The `--directory` option is the only way to specify the directory for all state. Inside this dir, VAST creates a directory log that contains all ASCII logs.

This PR adds a new option named `--log-directory` that optionally allows setting the directory for the logs only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
-- 🎁 A new option `--log-directory` has been added, which allows storing the log
-  separate from the persistent state. If unset, this option falls back to using
-  the persistent state directory set using `--directory`.
-  [#758](https://github.com/tenzir/vast/pull/758)
+- 🔄 The option `--directory` has been replaced by `--db-directory` and
+  `log-directory`, which set directories for persistent state and log files
+  respectively. The default log file path has changed from `vast.db/log` to
+  `vast.log`. [#758](https://github.com/tenzir/vast/pull/758)
 
 - 🔄 VAST now supports (and requires) Apache Arrow >= 0.16.
   [#751](https://github.com/tenzir/vast/pull/751)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🎁 A new option `--log-directory` has been added, which allows storing the log
+  separate from the persistent state. If unset, this option falls back to using
+  the persistent state directory set using `--directory`.
+  [#758](https://github.com/tenzir/vast/pull/758)
+
 - 🔄 VAST now supports (and requires) Apache Arrow >= 0.16.
   [#751](https://github.com/tenzir/vast/pull/751)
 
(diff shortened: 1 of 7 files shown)


mavam:
This is useful. However, from a UX perspective I think it's weird to specify both `--directory` and `--log-directory`. More natural IMO would be to have `--log-directory` and `--db-directory`. (But both would default to `--directory`.)

dominiklohmann:
I mean, speaking from a UX perspective, it is usually expected that `--directory` (or `-C`, not `-d` as we have it currently) sets the working directory of the binary.

Adding a separate DB directory can be done, but I'm not sure if it makes any sense.

In any case, the defaults would then have to be changed:
- `system.log-directory` falls back to `sys.directory / vast.db / log`
- `system.db-directory` falls back to `sys.directory / vast.db`

This would be not be a backwards-compatible change.

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🎁 A new option `--log-directory` has been added, which allows storing the log
+  separate from the persistent state. If unset, this option falls back to using
+  the persistent state directory set using `--directory`.
```
mavam: Technically, a log is also persistent, so the differentiation could be a bit clearer. Here's an alternate suggestion:

>  The new option `--log-directory` allows for storing the VAST application logs separate from the database state. If unset, this option falls back to  `--directory`.

mavam:
> [..] it is usually expected that --directory (or -C, not -d as we have it currently) sets the working directory of the binary

Yeah, that actually remains the same. My point was more like that we are now starting to differentiate, maybe we'll even get `--archive-directory`, `--index-directory`, `--schema-directory`, and even more eventually. Because the semantics of specifying them *in addition to* `--director` is confusing (what takes precedence when all a given?), I would argue that mixing them should result in an error: either specify one top-level directory or the more granular options.

dominiklohmann:
Let’s discuss this on Monday. I’d rather change the directory option to work like most other UNIX CLI tools have their directory option work. Especially with regards to our focus on UX.

dominiklohmann:
@mavam I have discussed this with @0ortmann, in order to determine what the natural and expected behavior is for him from a UX standpoint. This is what we converged on:

- Drop `--directory` completely.
- Add `--log-diretory`, and `--state-directory` (naming tbd).
- Change the defaults: `$PWD/log` for `system.log-directory`, and `$PWD/db` for `system.state-directory`.

libvast/src/system/application.cpp
```diff
@@ -72,7 +72,8 @@ auto make_root_command(std::string_view path) {
       .add<std::vector<std::string>>(
         "schema-paths", "list of paths to look for schema files "
                         "([" VAST_INSTALL_PREFIX "/share/vast/schema])")
-      .add<std::string>("directory,d", "directory for persistent state")
+      .add<std::string>("db-directory", "directory for persistent state")
+      .add<std::string>("log-directory", "directory for log files")
```
mavam: Maybe we can use `-D` and `-L` (or lowercase) as shortcuts? I would imagine that this comes in handy for testing.

tobim: Let's keep `-d` as a short option for `--db-directory` to avoid breaking scripts.

mavam: Then `-l` for *l*og directory makes more sense.

dominiklohmann: I've added the short options.

PR #763, opened 2020-02-19
Fix resolution of system schema path

tobim:


libvast/src/detail/process.cpp
```diff
@@ -53,6 +56,8 @@ caf::expected<path> objectpath_static() {
 } // namespace
 
 caf::expected<path> objectpath(const void* addr) {
+  if (addr == nullptr)
+    addr = (const void*) objectpath_dynamic;
```
mavam: Is there a specific reason for the C-style cast?

libvast/vast/detail/process.hpp
```diff
@@ -23,17 +23,6 @@ namespace vast::detail {
...
-/// Locates the path of a shared library or executable.
-/// @returns The filesystem path to the library or executable of the caller of
-///          this function, or error if the resolution fails.
-inline caf::expected<path> objectpath() {
-  struct dummy {
-    static void fn() {
-    }
-  };
-  return objectpath(reinterpret_cast<const void*>(&dummy::fn));
-}
```
mavam: Just for my information: why is this function no longer needed?

mavam: Just for my information: why is this function no longer needed?

tobim: This overload is replaced through the default parameter that was added to the other overload. This was buggy because it turned out that passing an address from within the mmapped region of the executable makes `glibc` to something unexpected.

libvast/src/detail/process.cpp
tobim: I changed it to `reinterpret_cast`.

PR #764, opened 2020-02-20
A potpourri of small tweaks

mavam:
- Make printers `constexpr`
- Add standard-like function `file_size`
- Rewrite `uuid` as byte sequence
- Make `chunk` more flexible
- Fix some warnings


mavam:
@tobim, you've reviewed a good amount of these changes already in #727. This PR factors the generic aspects that won't go into the integration branch.

PR #765, opened 2020-02-21
Revert "Annotate schemas with #index=hash"

lava:
This partially reverts commit 9f656e800283a8fae2ed466678e56b2828049d5c.

When benchmarking, we discovered that the hash index introduces
a huge performance regression for the `vast import suricata`
command, due to a variant type being used as a key in a `flat_map`.

Therefore, we're removing the hash indices until the underlying
problem is fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🔄 Hash indices have been disabled again due to a performance regression.
+  [#765](https://github.com/tenzir/vast/pull/765)
+
 - 🔄 The option `--directory` has been replaced by `--db-directory` and
   `log-directory`, which set directories for persistent state and log files
   respectively. The default log file path has changed from `vast.db/log` to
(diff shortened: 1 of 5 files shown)


tobim:
If we want to disable the hash index completely, we also need to disable the attribute inject heuristic in the Zeek parser.

We should also write a changelog entry so users know why this index is disabled.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+
```
dominiklohmann: That's one empty line too much.

schema/argus.schema
```diff
@@ -10,18 +10,18 @@ type argus.record = record{
   // Standard fields that are always present.
   StartTime: time #timestamp,     // stime
   Flgs: string,                   // flgs
-  Proto: string #index=hash,      // proto
+  Proto: string #id,              // proto
```
dominiklohmann: `:%s/#id,/`

That was a placeholder attribute. We can just revert this commit again when it's fixed, so we don't really need the placeholder.

lava: I'm not sure about this one, I think it's less surprising if a revert is **just** a revert, rather than having revert + fix-ups in the same commit. I can of course remove the `#id` in a separate commit, but then we couldn't cleanly re-apply the original commit after we fixed the underlying performance issues. So as long as the `#id` attribute does no harm, I'd prefer just keeping it as-is.

dominiklohmann: Fine by me as well.

PR #766, opened 2020-02-22
Port segment store to flatbuffers

mavam:
This PR is a followup to #727 and ports the segment to store to use flatbuffers. Since the change involved actually changing the segment flatbuffer schema, the store is now completely stateless and initializes runtime state on startup.

### DIFF
--- a/libvast/src/segment_store.cpp
+++ b/libvast/src/segment_store.cpp
@@ -14,6 +14,7 @@
 #include "vast/segment_store.hpp"
 
 #include "vast/bitmap_algorithms.hpp"
+#include "vast/concept/printable/stream.hpp"
 #include "vast/concept/printable/to_string.hpp"
 #include "vast/concept/printable/vast/error.hpp"
 #include "vast/concept/printable/vast/filesystem.hpp"
@@ -21,10 +22,9 @@
 #include "vast/error.hpp"
 #include "vast/event.hpp"
 #include "vast/fbs/segment.hpp"
+#include "vast/fbs/utils.hpp"
 #include "vast/ids.hpp"
-#include "vast/load.hpp"
 #include "vast/logger.hpp"
-#include "vast/save.hpp"
 #include "vast/segment_store.hpp"
 #include "vast/table_slice.hpp"
 #include "vast/to_events.hpp"
@@ -35,23 +35,25 @@
 
 namespace vast {
 
+// TODO: return expected<segment_store_ptr> for better error propagation.
 segment_store_ptr segment_store::make(path dir, size_t max_segment_size,
                                       size_t in_memory_segments) {
   VAST_TRACE(VAST_ARG(dir), VAST_ARG(max_segment_size),
              VAST_ARG(in_memory_segments));
   VAST_ASSERT(max_segment_size > 0);
-  auto x = std::make_unique<segment_store>(std::move(dir), max_segment_size,
-                                           in_memory_segments);
-  // Materialize meta data of existing segments.
-  if (exists(x->meta_path())) {
-    VAST_DEBUG_ANON(__func__, "loads segment meta data from", x->meta_path());
-    if (auto err = load(nullptr, x->meta_path(), x->segments_)) {
-      VAST_ERROR_ANON(__func__, "failed to unarchive meta data from",
-                      x->meta_path());
-      return nullptr;
-    }
-  }
-  return x;
+  auto result = std::make_unique<segment_store>(
+    std::move(dir), max_segment_size, in_memory_segments);
+  if (auto err = result->register_segments())
+    return nullptr;
+  return result;
+}
+
+segment_store::segment_store(path dir, uint64_t max_segment_size,
+                             size_t in_memory_segments)
+  : dir_{std::move(dir)},
+    max_segment_size_{max_segment_size},
+    cache_{in_memory_segments} {
+  // nop
 }
 
 segment_store::~segment_store() {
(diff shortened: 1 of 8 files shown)


libvast/src/segment_store.cpp
```diff
@@ -352,12 +345,34 @@ void segment_store::inspect_status(caf::settings& dict) {
...
 
-segment_store::segment_store(path dir, uint64_t max_segment_size,
-                             size_t in_memory_segments)
-  : dir_{std::move(dir)},
-    max_segment_size_{max_segment_size},
-    cache_{in_memory_segments} {
-  // nop
+caf::error segment_store::register_segments() {
+  for (auto filename : directory{segment_path()})
+    if (auto err = register_segment(filename))
```
tobim: We should probably _not_ break out of the loop here. An unrelated file that happens to be in this folder should not break the application.

libvast/src/segment_store.cpp
```diff
@@ -352,12 +345,34 @@ void segment_store::inspect_status(caf::settings& dict) {
...
+      return err;
+  return caf::none;
+}
+
+caf::error segment_store::register_segment(const path& filename) {
+  auto chk = chunk::mmap(filename);
+  if (!chk)
+    return make_error(ec::filesystem_error, "failed to mmap chunk", filename);
+  auto s = fbs::GetSegment(chk->data());
```
tobim: `s` needs to be checked for `nullptr`.

libvast/vast/fbs/segment.fbs
```diff
@@ -17,14 +17,17 @@ table Segment {
   /// The version of the segment.
   version: Version;
 
+  /// The contained table slices.
```
tobim: Maybe we should add a magic word before the version so we can check that the file actually belongs to us.

libvast/src/segment_store.cpp
```diff
@@ -352,12 +345,34 @@ void segment_store::inspect_status(caf::settings& dict) {
...
+  num_events_ += s->events();
+  auto segment_uuid = uuid{fbs::as_bytes<16>(*s->uuid())};
+  VAST_DEBUG_ANON(__func__, "found segment", segment_uuid);
```
tobim: Why not
```suggestion
  VAST_DEBUG(this, "found segment", segment_uuid);
```

libvast/src/segment_store.cpp
```diff
@@ -14,17 +14,17 @@
 #include "vast/segment_store.hpp"
 
 #include "vast/bitmap_algorithms.hpp"
+#include "vast/concept/printable/stream.hpp"
```
tobim: Is this really necessary now?

libvast/vast/fbs/segment.fbs
mavam: Good thought, it's actually already there. Have a look at the bottom of the file: the `file_identifier` directive does that natively.

libvast/src/segment_store.cpp
mavam: Why? The operation cannot fail. It just so happens that flatbuffers always returns a pointer.

libvast/src/segment_store.cpp
mavam: If we actually get an unrelated file, things break really bad when we call this a few lines before after mmaping the file:

```cpp
  auto s = fbs::GetSegment(chk->data());
```

I've added an extra check that bails out when the flatbuffer integrity is compromised.

libvast/src/segment_store.cpp
dominiklohmann: Maybe you can add `VAST_ASSERT(s)` in the line below for clarity.

mavam: The thing is that all `Get*` can never be `nullptr`, by design. The function always returns a pointer value.

mavam:
I've rebased the integration branch to master and then this topic branch onto the integration branch.

libvast/vast/fbs/segment.fbs
tobim: I see.

libvast/src/segment_store.cpp
tobim: That was not clear to me, a pointer as return type signals the opposite to me as a reader after all.

libvast/src/segment_store.cpp
```diff
@@ -357,6 +356,9 @@ caf::error segment_store::register_segment(const path& filename) {
   if (!chk)
     return make_error(ec::filesystem_error, "failed to mmap chunk", filename);
   auto s = fbs::GetSegment(chk->data());
+  auto verifier = fbs::make_verifier(chk);
+  if (!fbs::VerifySegmentBuffer(verifier))
+    return make_error(ec::format_error, "flatbuffer integrity check failed");
```
tobim: Nit: the verification call to `fbs::GetSegment()` can be done after the verification.

tobim: I'm also not sure this is a good idea. Consider the case of a large segment store (think of several TB): This verification means that we have to calculate the checksum for every segment during startup, which will probably take a long time.

mavam: > Nit: the call to `fbs::GetSegment()` can be done after the verification.

Okay, I'll move it over.

> I'm also not sure this is a good idea. Consider the case of a large segment store (think of several TB): This verification means that we have to calculate the checksum for every segment during startup, which will probably take a long time.

It's not a checksum. It's just a verification that all offsets are within bounds and don't yield segfaults. It's not a full scan over the data. This is how the layout looks like:

[FBPerson01](https://github.com/mzaks/FlatBuffersSwift/wiki/FlatBuffers-Explained)

It's my understanding that verification only requires the initial table state.

tobim: I see, this is good to go then.

PR #767, opened 2020-02-25
Fail early on invalid regular expressions

dominiklohmann:


dominiklohmann:
I don't think that fit's the current interface of `pattern` and very well. Following that logic, every operator would need to return an `expected<pattern>` as well.

Also, if a conjunction or a disjunction is used in a query in combination with an invalid pattern, should we really abort, or just treat the pattern as "not matched" / "not found" while issuing a warning? Because I assumed the latter to be better UX.

dominiklohmann:
I have reworked this as discussed at the office.

libvast/src/expression_visitors.cpp
```diff
@@ -248,15 +250,24 @@ caf::expected<void> validator::operator()(const predicate& p) {
...
-  if (ex.attr == system::type_atom::get_value()
-      && !(caf::holds_alternative<std::string>(d)
-           || caf::holds_alternative<pattern>(d)))
-    return make_error(ec::syntax_error,
-                      "type attribute extractor requires string or pattern "
-                      "operand",
-                      ex.attr, op_, d);
-  else if (ex.attr == system::timestamp_atom::get_value()
-           && !caf::holds_alternative<time>(d))
+  if (ex.attr == system::type_atom::get_value()) {
```
mavam: It looks like only expressions of the form `#type op RHS` get checked. But we want to validate *all* pattern occurrences, that is, every predicate operand that is a pattern should be checked.

libvast/src/expression_visitors.cpp
```diff
@@ -248,15 +250,24 @@ caf::expected<void> validator::operator()(const predicate& p) {
...
+    if (caf::holds_alternative<pattern>(d)) {
+      try {
+        [[maybe_unused]] auto r = std::regex{caf::get<pattern>(d).string()};
```
mavam:
```suggestion
        [[maybe_unused]] std::regex{caf::get<pattern>(d).string()};
```

Would this work as well?

dominiklohmann: Not for all compilers. 🤷‍♂

libvast/src/expression_visitors.cpp
dominiklohmann: This should be all the places now.

dominiklohmann:
> Yeah, it catches everything. However, I would say it makes more sense to perform a single check in the visitor function for `predicate`, e.g., via `get_if<pattern>(rhs)`. Then this is also more robust when we add new predicate operand types.

I have done the check as you requested, but please note that this will perform worse because it requires another level of indirection. We actually need `caf::get_if<pattern>(caf::get_if<data>(&p.rhs))` here. I know this is not a performance critical path, but I think we should always be aware of the cost of variant visitation.

mavam:
> I know this is not a performance critical path, but I think we should always be aware of the cost of variant visitation.

Exactly, the critical path should be the guideline. Expression validation takes place once and dwarfs in comparison to the following query execution.

Because of that, performance takes a less important role in the implementation. Rather, maintainability, flexibility, and degree of abstraction trump in this setting.

PR #768, opened 2020-02-26
Mention the continuous import fix in the changelog

tobim:


CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🐞 Continuously importing events from a Zeek process with a low rate of
+  emitted events resulted in a long delay until the data would be included
+  in the result set of queries. This is because the import process would
+  buffer up to 10000 events before sending them to the server as a batch.
```
```suggestion
  buffer up to 10,000 events before sending them to the server as a batch.
```

CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🐞 Continuously importing events from a Zeek process with a low rate of
```
mavam: Is this specific to the Zeek reader or to all text-based readers?

lava: Currently it's specific to zeek, although we should probably have a follow-up story to see if the same issue needs to be fixed for all our other readers as well.

mavam: Okay.

PR #783, opened 2020-03-06
Persist the index on each new partition

tobim:


libvast/src/system/index.cpp
```diff
@@ -141,16 +141,28 @@ caf::error index_state::load_from_disk() {
   return caf::none;
 }
 
+caf::error index_state::flush_meta_index() {
+  if (auto err = save(&self->system(), meta_index_filename(), meta_idx))
+    return err;
+  return caf::none;
```
mavam: Why not just `return save(...`?

tobim: 🤦‍♂

PR #784, opened 2020-03-06
Represent VAST type system in flatbuffers

mavam:


tobim:
The _flatbufferized_ definitions of `vast::type` look fine. I didn't manage a full pass over the main commit unfortunately.

mavam:
> Is this supposed to update the CAF submodule?

This PR will be rebased after #766 is merged. Only the last commit is currently relevant.

libvast/vast/fbs/layout.fbs
```diff
@@ -0,0 +1,11 @@
+include "type.fbs";
+
+namespace vast.fbs;
+
+table Layout {
+  name: string;
+  record: RecordType;
+  attributes: [Attribute];
+}
```
mavam: @tobim @dominiklohmann What do you think of a separate type for layouts, like this?

dominiklohmann: Do you think it makes sense to change our layout behavior at the same time as switching to flatbuffers? It might make more sense to introduce this change afterwards.

mavam: I think it can happen independently. But a new type `vast::layout` would be useful only once we've made the change with type names and alias types.

libvast/vast/fbs/type.fbs
```diff
@@ -0,0 +1,106 @@
...
+
+/// A CIDR subnet.
+table SubnetType {}
+
+/// A transport-layer port.
+table PortType {}
+
+/// An enumeration type that can have one specific value.
+table EnumerationType {
+  fields: [string];
```
tobim, line 38: Could you change this to use an `EnumerationField` with a definition reflecting `RecordField`:
```
table EnumerationField {
  name: string;
}
```

I request this because I intend to add a type member in the not too far away future, and making this forwards compatible would be great!

libvast/vast/fbs/type.fbs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+/// A type with a name.
+table AliasType {
+  name: string;
+  type: Type;
+}
+
+/// The sum type of all possible types.
+union TypeVariant {
```
tobim, line 75: `EnumerationType` is missing.

libvast/vast/fbs/layout.fbs
tobim: As discussed, we can keep this.

libvast/vast/fbs/type.fbs
mavam: Sure, adding this table is possible.

It'd be still great to understand the underlying use case that motivates this. Mind explaining this real quick?

tobim: I intend to let `enumeration`s grow into a real sum type instead of adding a separate type to the variant.

mavam: I would rather have a separate `SumType` for this.

mavam: Okay, we resolved this offline. The misunderstanding was that we had different notions of the power of `enum`s. My view was way more limited than, say, Rust or Haskell enums, which are effectively special cases of sum types.

PR #785, opened 2020-03-09
Add vast-unit-test dir to gitignore

dominiklohmann:


dominiklohmann:
> Isn't this directory typically in the build directory? If not, maybe something else is wrong.

Only if you run `cmake --build <build-dir> --target test` or `cd <build-dir> && make test`. However, if you want to run just a single test suite, the command you run usually is `vast-test -v 4 -s suite`, which creates the `vast-unit-test/` dir in your cwd. I wouldn't consider that a defect. You'd get the same effect with `make -f <build-dir>/Makefile test` as well.

mavam:
> I wouldn't consider that a defect.

I actually would. I think the unit test directory should at least by default not run relative to your CWD, but relative to the build directory.

dominiklohmann:
That would require putting the path to the build directory into the vast-test binary. That is too magic to be considered sane.

mavam:
> That would require putting the path to the build directory into the vast-test binary.

Don't we have already logic in VAST that can find the library/executable (itself)? It seems that this would allow for a robust specification of a relative path.

dominiklohmann:
So then I install vast-test and run it, and it puts the vast-unit-test directory into `$PREFIX`?

mavam:
The `vast-test` binary doesn't get installed and is almost always located in `$BUILD/bin/vast-test`. I could imagine that simply going up one directory works for 99% of the cases.

dominiklohmann:
I honestly don't understand why this even needs to be discussed.

In my opinion, no time should be spent on this issue, considering it only affects developers who run vast-test manually. We have so many more pressing things to work on.

If small annoyances like this need to be discussed as much, they become infeasible to fix as you encounter them.

For us internally, this single line in the `.gitignore` fixes the issue at no runtime and practically no developer cost. An application ignoring its working directory is unexpected behavior, and may even be considered harmful, as I've tried to explain above.

PR #786, opened 2020-03-09
Create pivot graph in type-registry

dominiklohmann:


dominiklohmann:
I'm closing this. The type registry has changed a fair bit since we've worked on this, so let's re-do when we decide to tackle this story again.

PR #787, opened 2020-03-10
Introduce the VERBOSE log level

tobim:


libvast/vast/logger.hpp
```diff
@@ -54,10 +54,11 @@ namespace vast::detail {
...
 template <class T>
 auto id_or_name(T&& x) {
-  static_assert(
-    !std::is_same_v<const char*, std::remove_reference<T>>,
-    "const char* is not allowed for the first argument in a logging statement. "
-    "Supply a component or use VAST_[ERROR|WARNING|INFO|DEBUG]_ANON instead.");
+  static_assert(!std::is_same_v<const char*, std::remove_reference<T>>,
+                "const char* is not allowed for the first argument in a "
+                "logging statement. "
+                "Supply a component or use "
```
dominiklohmann:
```suggestion
                "logging statement. Supply a component or use "
```

libvast/src/system/actor_system.cpp
```diff
@@ -0,0 +1,75 @@
...
+} // namespace
+
+void fixup_logger(const configuration& cfg) {
+  // Reset the logger so we can support the VERBOSE level
+  namespace lg = defaults::logger;
+  struct _hack : public caf::ref_counted {
+    caf::logger::config cfg_;
+  };
+  auto logger = caf::logger::current_logger();
+  auto& cfg_ = reinterpret_cast<_hack*>(logger)->cfg_;
```
dominiklohmann: We should raise an issue in CAF and add a comment above this line with a link to the issue.

libvast/src/system/configuration.cpp
```diff
@@ -27,6 +27,7 @@
 #include <caf/openssl/manager.hpp>
 #endif
 
+#include "vast/defaults.hpp"
```
dominiklohmann: Left in by accident?

libvast/src/system/actor_system.cpp
dominiklohmann: @lava linked this as a better solution: http://bloglitb.blogspot.com/2010/07/access-to-private-members-thats-easy.html

I think this should be used instead to avoid possible alignment issues.

tobim: Both done.

PR #788, opened 2020-03-10
Print index/archive event handling rates for VERBOSE log level

dominiklohmann:
~Blocked by #787. As is this is printing to INFO, which makes a lot of noise.~

For archive this is just the rate, for the index it's the minimum we print. All the other rates for the index are still in the accountant, this just makes it a lot more convenient to get them.


dominiklohmann:
I have added the maximum rate for the index as well per suggestion in the office, but I'm not convinced that the value is very useful.

dominiklohmann:
This is now ready for review @tobim.

PR #790, opened 2020-03-10
Fail integration tests early if jq is not in PATH

dominiklohmann:
Before this change, integration tests failed with an unintelligible error when `jq` was unavailable. This is a quick fix that works only when running the integration tests through the build target integration.


dominiklohmann:
> Good use of `command -v` instead of `which`!

Some day we'll all be using zsh and we can finally use `if (( $+commands[jq] ))` instead.

PR #792, opened 2020-03-10
Fix recently introduced bug in eoi check for expression parser

dominiklohmann:
`parsers::eoi` doesn't seem to play nicely with the rule parser in all build configurations. While fixing that, I've uncovered an issue in our `integration.py` script, which I've worked around for now.

### DIFF
--- a/libvast/vast/concept/parseable/vast/expression.hpp
+++ b/libvast/vast/concept/parseable/vast/expression.hpp
@@ -18,11 +18,10 @@
 #include "vast/concept/parseable/vast/data.hpp"
 #include "vast/concept/parseable/vast/type.hpp"
 #include "vast/data.hpp"
-#include "vast/expression.hpp"
-#include "vast/type.hpp"
-
 #include "vast/detail/assert.hpp"
 #include "vast/detail/string.hpp"
+#include "vast/expression.hpp"
+#include "vast/type.hpp"
 
 namespace vast {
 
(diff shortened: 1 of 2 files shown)


libvast/vast/concept/parseable/vast/expression.hpp
```diff
@@ -165,23 +165,22 @@ struct expression_parser : parser<expression_parser> {
...
-  bool parse(Iterator& f, const Iterator& l, expression& a) const {
-    static auto p = make<Iterator>();
-    return p(f, l, a);
+    if (!p(f, l, x))
+      return false;
+    if (f != l) {
+      VAST_ERROR_ANON("failed to parse expression", std::string(f, l));
+      return false;
+    }
+    return true;
```
mavam:
```suggestion
    return f == l;
```

We usually don't have side effects like logging a parser. This should go into the parent scope.

The error message says `expression` but the signature has an unconstrained template parameter. Is this intentional?

dominiklohmann: > The error message says `expression` but the signature has an unconstrained template parameter. Is this intentional?

We do this for other parsers as well, since the input may be either `expression` or `unused_type`.

> We usually don't have side effects like logging a parser. This should go into the parent scope.

It can't go into the parent scope. This log message shows _where_ the parser failed when parsing an expression. I know this is suboptimal, but getting a generic "parser_error !!" message when the expression parser fails isn't optimal either. I think it's a reasonable tradeoff.

We do have a task somewhere to allow carrying an error message (or at least an errc à la `std::from_chars`) through our parser framework, but that is a larger refactoring.

PR #796, opened 2020-03-11
Use heterogenous lookup for hash index

lava:
Some notes:
 * I tried to get the `transparent_unordered_map` to work, but it breaks serialization due to the different key type. Since the API was a bit weird anyways, I gave up on that approach.
 * The unit tests are actually failing because `hash<data>()` and `hash<data_view>()` give different result, not completely sure why yet.
 * Apart from that, lookups with and without hash index now perform at the same speed in my benchmarking, but hash indices still don't make things *faster*. That said, I don't know if performance was the main motivation for introducing hash indices, and I didn't go beyond the obvious optimizations so there is probably still room for further improvements.
* The hash map is not properly integrated into the build system yet; locally I hard-coded the include to my global CXXFLAGS.


libvast/vast/hash_index.hpp
```diff
@@ -255,9 +251,45 @@ class hash_index : public value_index {
...
+  struct data_hash {
+    size_t operator()(const data& x) const {
+      // The default hash computation for `data` and `data_view` is subtly
+      // different: For `data_view` a hash of the contents of the view is
+      // created (so that the same data compares equal whether it's stored
+      // in a default/msgpack/arrow view), but for `data` the hashing is
+      // forwarded to the actual container classes, which can define their own
+      // hash functions.
+      // To ensure the same method is used consistently, we create a view here.
+      return vast::uhash<vast::xxhash>{}(make_view(x));
```
lava: Note: This feels expensive, but I couldn't measure any performance difference so I thought it's better to be safe than sorry.

lava:
Note: I highly recommend to look at the commits individually when reviewing.

libvast/vast/hash_index.hpp
mavam: Okay, if it pops up in the profiler we can still revisit this.

libvast/vast/view.hpp
```diff
@@ -239,11 +239,6 @@ class container_view_handle
 
   template <class Hasher>
   friend void hash_append(Hasher& h, container_view_handle xs) {
-    if (!xs) {
-      hash_append(h, false);
-      return;
-    }
-    hash_append(h, true);
```
mavam: Maybe add an assertion that `xs` is not null? Otherwise `*xs` below may crash.

libvast/test/view.cpp
```diff
@@ -183,3 +183,22 @@ TEST(container comparison) {
...
+
+  using hash = vast::uhash<vast::xxhash>;
+  CHECK_EQUAL(hash{}(i), hash{}(make_view(i)));
+  CHECK_EQUAL(hash{}(c), hash{}(make_view(c)));
+  CHECK_EQUAL(hash{}(st), hash{}(make_view(st)));
+  CHECK_EQUAL(hash{}(p), hash{}(make_view(p)));
+  CHECK_EQUAL(hash{}(v), hash{}(make_view(v)));
+  CHECK_EQUAL(hash{}(s), hash{}(make_view(s)));
+  CHECK_EQUAL(hash{}(m), hash{}(make_view(m)));
+}
```
dominiklohmann: Missing newline at EOF.

libvast/vast/hash_index.hpp
```diff
@@ -255,9 +251,45 @@ class hash_index : public value_index {
...
+      return x == y;
+    }
+
+    bool operator()(const data& x, const data& y) const {
+      return x == y;
+    }
+
+    bool operator()(const data_view& x, const data_view& y) const {
+      return x == y;
+    }
```
dominiklohmann:
```suggestion
    bool operator()(const auto& x, const auto& y) const {
      return x == y;
    }
```

libvast/test/view.cpp
dominiklohmann: This test seems to fail btw, for the reasons that you explained in the longer comment in the data_hash call operator. The test should be reworked to reflect that.

libvast/vast/hash_index.hpp
dominiklohmann: Nevermind, that's C++20. You could still use templates to reduce it to just a single function, though.

libvast/test/view.cpp
lava: Huh, that's weird, I changed the hash of vector in the same commit to make the test work. I'll have to investigate.

dominiklohmann: I'll test locally first, I might've messed it up on my end also when trying to get the PR to compile.

libvast/vast/view.hpp
lava: Instead I added the hash of `caf::none` that was there originally; after some consideration I felt that it would be weird if the hash function can crash for a valid container_view state, even if we should not hash a container_view in that state.

lava:
> Also, why not add robin-map as a submodule instead? Copying over everything makes keeping track of new releases quite hard.
> 
> Edit: A changelog entry seems to be missing as well.

I updated the PR to add robin-map as a subtree and added a CHANGELOG entry.

libvast/test/view.cpp
lava: The latest version seems to pass unit tests, so I'll resolve this for now.

PR #802, opened 2020-03-23
Set a default request timeout

dominiklohmann:
I'm not sure if this should be made configurable for now. 10s seems to be a reasonable value, and for decent usability 10s already feels like an infinite wait time.

For some of these, we might also want to print an error if the request timed out. Ultimately this is what motivated this change.

### DIFF
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -175,14 +175,24 @@ void collect_component_status(node_actor* self,
   }
 }
 
-caf::message status_command(const command::invocation&, caf::actor_system&) {
+caf::message
+status_command(const command::invocation&, caf::actor_system& sys) {
   auto self = this_node;
   auto rp = self->make_response_promise();
-  self->request(self->state.tracker, infinite, get_atom::value).then(
-    [=](registry& reg) mutable {
-      collect_component_status(self, std::move(rp), reg);
-    }
-  );
+  caf::error err;
+  self
+    ->request(self->state.tracker, defaults::system::initial_request_timeout,
+              get_atom::value)
+    .then(
+      [=](registry& reg) mutable {
+        collect_component_status(self, std::move(rp), reg);
+      },
+      [&](caf::error& status_err) { err = status_err; });
+  if (err) {
+    VAST_ERROR(__func__,
+               "failed to receive status before timeout:", sys.render(err));
+    return caf::make_message(std::move(err));
+  }
   return caf::none;
 }
 
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -189,8 +189,7 @@ status_command(const command::invocation&, caf::actor_system& sys) {
       },
       [&](caf::error& status_err) { err = status_err; });
   if (err) {
-    VAST_ERROR(__func__,
-               "failed to receive status before timeout:", sys.render(err));
+    VAST_ERROR(__func__, "failed to receive status:", sys.render(err));
     return caf::make_message(std::move(err));
   }
   return caf::none;
--- a/libvast/vast/defaults.hpp
+++ b/libvast/vast/defaults.hpp
@@ -333,9 +333,13 @@ constexpr size_t initially_requested_ids = 128;
 constexpr std::chrono::milliseconds telemetry_rate = std::chrono::milliseconds{
   1000};
 
-// Interval between checks whether a signal occured.
+/// Interval between checks whether a signal occured.
 constexpr std::chrono::milliseconds signal_monitoring_interval
   = std::chrono::milliseconds{750};
+
+/// Timeout for initial connections to the node.
+constexpr std::chrono::seconds initial_request_timeout
+  = std::chrono::seconds{10};
 
 } // namespace system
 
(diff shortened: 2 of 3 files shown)


libvast/src/system/evaluator.cpp
```diff
@@ -170,7 +171,7 @@ caf::behavior evaluator(caf::stateful_actor<evaluator_state>* self,
         auto& curried_pred = get<1>(triple);
         auto& indexer = get<2>(triple);
         st.predicate_hits[pos].first += 1;
-        self->request(indexer, caf::infinite, curried_pred)
+        self->request(indexer, defaults::system::request_timeout, curried_pred)
           .then([=](const ids& hits) { self->state.handle_result(pos, hits); },
                 [=](const caf::error& err) {
                   self->state.handle_missing_result(pos, err);
```
lava: Do we have to distinguish between timeouts and `missing result` errors here now?

libvast/src/system/query_supervisor.cpp
```diff
@@ -60,20 +61,21 @@ query_supervisor(caf::stateful_actor<query_supervisor_state>* self,
...
-            if (--num_evaluators == 0) {
-              VAST_DEBUG(self, "collected all results for partition", id);
-              self->state.open_requests.erase(id);
-              // Ask master for more work after receiving the last sub
-              // result.
-              if (self->state.open_requests.empty()) {
-                VAST_DEBUG(self, "collected all results for all partitions");
-                self->send(client, done_atom::value);
-                self->send(master, worker_atom::value, self);
+          self->request(evaluator, defaults::system::request_timeout, client)
```
lava: Do we need to add a new error handler for the timeout case here?

libvast/src/system/spawn_exporter.cpp
```diff
@@ -51,8 +51,10 @@ maybe_actor spawn_exporter(node_actor* self, spawn_arguments& args) {
     caf::anon_send(exp, extract_atom::value);
   // Send the running IMPORTERs to the EXPORTER if it handles a continous query.
   if (has_continuous_option(query_opts)) {
-    self->request(self->state.tracker, caf::infinite, get_atom::value).then(
-      [=](registry& reg) mutable {
+    self
+      ->request(self->state.tracker, defaults::system::request_timeout,
+                get_atom::value)
+      .then([=](registry& reg) mutable {
```
lava: Same here, do we have to add an error handler for the timeout case?

libvast/vast/defaults.hpp
```diff
@@ -352,10 +352,13 @@ constexpr size_t initially_requested_ids = 128;
 constexpr std::chrono::milliseconds telemetry_rate = std::chrono::milliseconds{
   1000};
 
-// Interval between checks whether a signal occured.
+/// Interval between checks whether a signal occured.
 constexpr std::chrono::milliseconds signal_monitoring_interval
   = std::chrono::milliseconds{750};
 
+/// Time after which a request is considered to have failed.
+constexpr std::chrono::seconds request_timeout = std::chrono::seconds{10};
```
lava: This sounds very generic, maybe consider `caf_request_timeout` to make clear which kinds of requests are meant here.

libvast/src/system/evaluator.cpp
dominiklohmann: I'm currently implementing checks for the response handles. The CI failure still has me puzzled, and I want to find out if it is related to a request shutdown in some way, but I fail to reproduce it. (This answer applies to the next two comments as well.)

libvast/vast/defaults.hpp
dominiklohmann: I've been thinking about `node_request_timeout`. I don't like the `caf_` prefix, as it makes it sound like it's a CAF thing.

libvast/src/system/evaluator.cpp
dominiklohmann: I've come to the conclusion that we must keep using `infinite` for at least the evaluator.

mavam:
Does it really make sense to have a single timeout as default? I could imagine that this highly depends on the expected latency.

If we chose a single value, I think calling it `default_timeout` would help the reader in understanding that a timeout has not yet been set specifically for this request.

dominiklohmann:
> Does it really make sense to have a single timeout as default? I could imagine that this highly depends on the expected latency.

This mostly is about having a timeout at all. The actual value may end up being a minute even. The desired UX is "Hey, something's likely wrong on our end, requesting a status should not take a full minute, we're stopping here."

I've been toying around with the implications of the initial change I made on the side today and have done some stress testing under load, and it looks like having such a timeout is really only viable when connecting to the node.

What I'd really like to have is some sort of timeout until the result arrives, and after that no timeout at all, but I'm not sure if CAF allows for that easily.

libvast/src/system/node.cpp
```diff
@@ -175,14 +175,24 @@ void collect_component_status(node_actor* self,
...
+    ->request(self->state.tracker, defaults::system::initial_request_timeout,
+              get_atom::value)
+    .then(
+      [=](registry& reg) mutable {
+        collect_component_status(self, std::move(rp), reg);
+      },
+      [&](caf::error& status_err) { err = status_err; });
+  if (err) {
+    VAST_ERROR(__func__,
+               "failed to receive status before timeout:", sys.render(err));
```
tobim: The message could contain a non-timeout related error as well.
```suggestion
    VAST_ERROR(__func__, "failed to receive status:", sys.render(err));
```

PR #803, opened 2020-03-24
Restrict log file creation to 'vast start'

lava:
Set file verbosity to 'quiet' by default, and do not create a log file and log directory when file verbosity is set to 'quiet'.

Set file verbosity to 'debug' by default only for the 'vast start' subcommand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🔄 The log folder `vast.log/` in the current directory will not be created
+  by default any more. Users must explicitly set the `system.file-verbosity`
+  option if they wish to keep the old behavior.
+  [#803](https://github.com/tenzir/vast/pull/803)
+
 - 🎁 The new `vast import syslog` command allows importing Syslog messages
   as defined in [RFC5424](https://tools.ietf.org/html/rfc5424).
   [#770](https://github.com/knapperzbusch/vast/pull/770)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🔄 The config option `system.log-directory` was deprecated and replaced
+  by the new option `system.log-file`. All logs will now be written to a
+  single file.
+  [#806](https://github.com/tenzir/vast/pull/803)
+
 - 🔄 The log folder `vast.log/` in the current directory will not be created
   by default any more. Users must explicitly set the `system.file-verbosity`
   option if they wish to keep the old behavior.
--- a/libvast/src/command.cpp
+++ b/libvast/src/command.cpp
@@ -390,7 +390,10 @@ bool init_config(caf::actor_system_config& cfg, const command::invocation& from,
     }
     cfg.set("logger.console-verbosity", *value);
   }
-
+  // Allow users to use `system.log-file` to set log filename
+  if (auto fn = caf::get_if<std::string>(&cfg, "system.log-file"))
+    cfg.set("logger.file-name", *fn);
+  // Allow users to specify `verbose` log level.
   auto fixup_verbose_mode = [&](const std::string& option) {
     auto logopt = "logger." + option;
     if (auto verbosity = caf::get_if<caf::atom_value>(&cfg, logopt)) {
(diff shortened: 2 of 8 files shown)


libvast/src/system/default_configuration.cpp
```diff
@@ -31,9 +31,9 @@ default_configuration::default_configuration() {
   set("logger.component-blacklist",
       caf::make_config_value_list(atom("caf"), atom("caf_flow"),
                                   atom("caf_stream")));
-  set("logger.console-verbosity", atom("INFO"));
+  set("logger.console-verbosity", defaults::logger::console_verbosity);
   set("logger.console", atom("COLORED"));
-  set("logger.file-verbosity", atom("DEBUG"));
+  set("logger.file-verbosity", defaults::logger::file_verbosity);
```
dominiklohmann: Does it make any difference that the atoms were previously created from an all uppercase string, and now they're created from an all lowercase string?

libvast/src/command.cpp
```diff
@@ -352,6 +372,10 @@ parse(const command& root, command::argument_iterator first,
 
 bool init_config(caf::actor_system_config& cfg, const command::invocation& from,
                  std::ostream& error_output) {
+  // Adjust default file verbosity for the start command.
+  if (from.name() == "start")
```
lava: Tbh, after working on this I'm not even sure that we really need this special case: It seems perfectly reasonable to say that `vast` will not write logs to a file unless log-file and/or file-verbosity are specified, including `vast start`.

tobim: You are right, it is more customary not to write logfiles unless specified. But think about the usual scenario from a users perspective: You normally don't think about enabling the log file until the moment something goes wrong and you need them, on the other hand, verbose logging is still not that high volume that is poses a problem, so it does not hurt to just turn them on from the start.

libvast/src/system/default_configuration.cpp
lava: It should be fine, we call `caf::to_lowercase()` in all relevant places.

libvast/src/command.cpp
dominiklohmann: > You normally don't think about enabling the log file until the moment something goes wrong and you need them

This hits the nail on the head I think—it's exactly what prompted us to create the verbose log level just recently. I think having it enabled by default is fine.

lava: I'd argue that this is the *developers* perspective rather than the users perspective. A user toying around with vast will most likely not dig through logs to figure out an error. (and if the error is non-repeatable, he will most likely just shrug it off and move on) And creating a log directory does hurt: the original reason for this PR was user complaints after all.

Maybe a better solution to this problem would be to set the default console verbosity to "debug" for non-release builds? Then you could not forget to turn on verbose logs until it is too late.

lava:
> This does not yet seem to handle vast -N, does it? Haven't tested yet, just had a glance at the code.

Yes, by design: The overall goal was that commands issued by the user in his terminal should not pollute the current working directory by creating a vast log directory, which would include `vast -N`. Do you think we should add an exception for `vast -N`?

dominiklohmann:
> Do you think we should add an exception for `vast -N`?

Yes, since it is basically running `vast start` under the hood.

libvast/src/command.cpp
lava: As discussed in slack, I moved the default location of the log file inside `vast.db/`, this way users don't notice it but developers can still look at it if they want to.

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## [Unreleased]
 
+- 🔄 The log folder `vast.log/` in the current directory will not be created
+  by default any more. Users must explicitly set the `system.file-verbosity`
+  option if they wish to keep the old behaviour.
```
dominiklohmann:
```suggestion
  option if they wish to keep the old behavior.
```

libvast/src/command.cpp
```diff
@@ -352,6 +372,10 @@ parse(const command& root, command::argument_iterator first,
 
 bool init_config(caf::actor_system_config& cfg, const command::invocation& from,
                  std::ostream& error_output) {
+  // Adjust default file verbosity for commands starting a node,
+  // i.e. `vast start` and `vast -N ...`.
```
dominiklohmann:
```suggestion
  // i.e., `vast start` and `vast -N ...`.
```

libvast/src/command.cpp
```diff
@@ -222,6 +222,26 @@ void mantext(const command& cmd, std::ostream& os,
...
+    case caf::atom_uint("warning"):
+      return VAST_LOG_LEVEL_WARNING;
+    case caf::atom_uint("info"):
+      return VAST_LOG_LEVEL_INFO;
+    case caf::atom_uint("debug"):
+      return VAST_LOG_LEVEL_DEBUG;
+    case caf::atom_uint("trace"):
+      return VAST_LOG_LEVEL_TRACE;
+  }
+}
```
dominiklohmann: This seems awfully similar to the code below in `init_config(...)` in the same file. This should be deduplicated.

dominiklohmann: The same code also exists in `vast/system/logger.cpp`.

lava:
[ch13224]

vast.conf
```diff
@@ -3,10 +3,10 @@ system {
 ; endpoint = "localhost:42000"
 
 ;; The file system path used for persistent state.
-; db-directory = "vast.db"
+; db-directory = "/var/lib/vast/"
```
dominiklohmann: The commented-out values in the example configuration should reflect the default values, not platform-specific recommendations. As far as I can tell, these are now `"vast.db"` and `"vast.db/log"` for `system.db-directory` and `system.log-file` respectively.

libvast/src/command.cpp
lava: Good spot; deduplicated.

PR #807, opened 2020-03-24
Make the release build more introspectable

tobim:


CMakeLists.txt
```diff
@@ -210,7 +210,7 @@ set(CMAKE_CXX_STANDARD 17)
 
 # Build-type specific flags.
 string(APPEND CMAKE_CXX_FLAGS_DEBUG " -O0")
-string(APPEND CMAKE_CXX_FLAGS_RELEASE " -msse3 -mssse3 -msse4.1 -msse4.2")
+string(APPEND CMAKE_CXX_FLAGS_RELEASE " -msse3 -mssse3 -msse4.1 -msse4.2 -fno-omit-frame-pointer")
```
mavam: Omitting the frame pointer can degrade the performance substantially. Should this really be set in the release build?

tobim: I assume you mean "improve" instead of "degrade".

We really need to measure this in order to learn the impact, but for now keeping it in allows us to use powerful profiling tools on the release builds, and that will provide us immediate value for investigating performance issues in several test deployments.

mavam: > I assume you mean "improve" instead of "degrade".

Actually, no 🙂. Omitting the frame pointer allows for more optimizations (one extra free register for other stuff) and less work has to be done at the instruction level (i.e., no adjustment of `%ebp` etc.).

mavam: > We really need to measure this in order to learn the impact, but for now keeping it in allows us to use powerful profiling tools on the release builds

Fair point.

PR #813, opened 2020-03-30
Add concrete usage examples to docs

dominiklohmann:
Note: This is still heavily work-in-progress, not worth reviewing yet.

Each subcommand currently has a `help` and a `documentation` subcommand. This change adds concrete usage example to _every_ subcommands documentation subcommand.

### DIFF
--- a/doc/cli/vast-export.md
+++ b/doc/cli/vast-export.md
@@ -1,8 +1,38 @@
 The `export` command retrieves a subset of data according to a given query
 expression. The export format must be explicitly specified:
 
-```
+```bash
 vast export [options] <format> [options] <expr>
 ```
 
-The `export` command is the dual to the `import` command.
+This is easiest explained on an example:
+
+```bash
+vast export --max-events=100 --continuous json '#timestamp < 1 hour ago'
+```
+
+The above command outputs line-delimited JSON like this, showing one event per
+line:
+
+```json
+{"ts": "2020-08-06T09:30:12.530972", "nodeid": "1E96ADC85ABCA4BF7EE5440CCD5EB324BEFB6B00#85879", "aid": 9, "actor_name": "pcap-reader", "key": "source.start", "value": "1596706212530"}
+```
+
+The above command signals the running server to export 100 events to the
+`export` command, and to do so continuously (i.e., not matching data that was
+previously imported). Only events that have a field annotated with the
+`#timestamp` attribute will be exported, and only if the timestamp in that field
+is older than 1 hour ago from the current time at the node.
+
+The default mode of operation for the `export` command is historical queries,
+which exports data that was already archived and indexed by the node. The
+`--unified` flag can be used to export both historical and continuous data.
+
+For more information on the query expression, see the [query language
+documentation](https://docs.tenzir.com/vast/query-language/overview).
+
+Some export formats have format-specific options. For example, the `pcap` export
+format has a `--flush-interval` option that determines after how many packets
+the output is flushed to disk. A list of format-specific options can be
+retrieved using the `vast export <format> help`, and individual documentation is
+available using `vast export <format> documentation`.
--- a/doc/cli/vast-pivot.md
+++ b/doc/cli/vast-pivot.md
@@ -1,13 +1,30 @@
-The `pivot` command retrieves data of a related type. It inspects each event in
-a query result to find an event of the requested type. If the related type
-exists in the schema, VAST will dynamically create a new query to fetch the
+The `pivot` command retrieves data of a related type. It inspects each
+event in a query result to find an event of the requested type. If the related
+type exists in the schema, VAST will dynamically create a new query to fetch the
 contextual data according to the type relationship.
 
-```sh
+```bash
 vast pivot [options] <type> <expr>
 ```
 
 VAST uses the field `community_id` to pivot between logs and packets. Pivoting
-is currently implemented for Suricata, Zeek (with
-[community ID computation] (https://github.com/corelight/bro-community-id)
-enabled), and PCAP.
+is currently implemented for Suricata, Zeek (with [community ID
+computation](https://github.com/corelight/bro-community-id) enabled), and PCAP.
+For Zeek specifically, the `uid` field is supported as well.
+
+For example, to get all events of type `pcap.packet` that can be pivoted to over
+common fields from other events that match the query `dest_ip == 72.247.178.18`,
+use this command:
+
+```bash
+vast pivot pcap.packet 'dest_ip == 72.247.178.18'
+```
+
+The `pivot` command is similar to the `explore` command in that they allow for
+querying additional context.
+
+Unlike the `export` command, the output format can be selected using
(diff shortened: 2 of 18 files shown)


tobim:
It might be useful to split the examples into a separate string so we are flexible with regards to rendering that stuff.

dominiklohmann:
> It might be useful to split the examples into a separate string so we are flexible with regards to rendering that stuff.

Note that this would have implications for our docs pipeline, too. Currently, the documentation is read from Markdown files in the docs/cli subdirectory. Expanding this can obviously be done, but we need to be aware that this would increase the amount of necessary work that goes into this.

tobim:
> Note that this would have implications for our docs pipeline, too. Currently, the documentation is read from Markdown files in the docs/cli subdirectory. Expanding this can obviously be done, but we need to be aware that this would increase the amount of necessary work that goes into this.

Yeah, I forgot about that. In that case let's not incur that overhead on ourselves until we have need for it. I also think that it might be more appropriate to keep the documentation in one `.md` file for each command and use suitable tools to work with them.

dominiklohmann:
[ch14389]

lava:
As a general comment, the current documentation seems to rely quite heavily on the presence of a markdown post-processor. This can make the terminal output very cluttered, where the raw markdown is being printed.

I think the readability could be improved a lot for this use case by staying a bit closer to how e.g. manpage look like, while still having valid markdown. As a quick mock-up:

format_comparison

This doesn't need to be changed as part of this PR, but I think something good to keep in mind in the longer term.

mavam:
@lava does the `[x]` style of link creation work with most Markdown processors? It's a different style of linking that seems more appropriate for plain ASCII, LaTeX, etc. I think the goal should be to create the right markup, so that we can embed it at http://docs.tenzir.com. If someone finds the markup too much in the way, it's easy to just pipe the output to `mdv`. I'd rather have GFM-style code highlighting than optimizing for readability of a subset of Markdown.

dominiklohmann:
For the code snippets, if we indent with 4 spaces instead of using triple backticks we get no highlighting on the documentation page. 

For the links we can do this:

```
Some [text with a link][some-slug].

[some-slug]: https://some.url
```

lava:
@mavam I'm not sure that piping to mdv enhances the output:

mvd

mavam:
Is this a modified version of the docs or the one with footnote-style links? It looks like the latter, but my point was that regular links, as well as using triple-backticks for code, doesn't sacrifice readability when piping to `mdv` and friends.

lava:
@mavam This is the original version of the docs, i.e. the left-hand side in the first picture. The right-hand side version actually looks reasonably well when piped through mdv, although still worse than the raw text because the code block is lost as well.

Ironically, mdv seems to convert links to use out-of-line definitions by default, probably that's why it looked similar to the modified version to you. (i'm using the latest `mdv` installed via `pip3 install mdv`)

doc/cli/vast-pivot.md
```diff
@@ -1,13 +1,24 @@
-The `pivot` command retrieves data of a related type. It inspects each event in
-a query result to find an event of the requested type. If the related type
-exists in the schema, VAST will dynamically create a new query to fetch the
+The `vast pivot` command retrieves data of a related type. It inspects each
+event in a query result to find an event of the requested type. If the related
+type exists in the schema, VAST will dynamically create a new query to fetch the
 contextual data according to the type relationship.
```
lava, line 4: The first paragraph is still pretty hard to parse without roughly knowing how pivot works already. In particular, the first two sentences give the impression that pivot is creating a subset of the query result containing only events of the requested type. (and even before that one has to make the mental connection between "a related type" and "the requested type")

The third sentence also immediately raises the question of what happens when the related type does not exist in the schema, which is never answered.

doc/cli/vast-stop.md
```diff
@@ -1 +1,9 @@
-The `stop` command brings down a VAST node.
+The `vast stop` command gracefully brings down a VAST node, and is the analog of
+the `vast start` command.
+
+WHile it is technically possible to shut down a VAST node gracefully by sending
```
lava, line 4: Random nit: `s/WHile/While/`

dominiklohmann:
I know it's been almost half a year since we dropped this for other stuff, but it's finally ready for review.

Note that I've removed some of the options listing from the explore command, since that is mirrored in the helptext already.

doc/cli/vast-export.md
```diff
@@ -5,4 +5,27 @@ expression. The export format must be explicitly specified:
 vast export [options] <format> [options] <expr>
 ```
 
-The `export` command is the dual to the `import` command.
+This is easiest explained on an example:
+
+```
+vast export --max-events=100 --continuous json '#timestamp < 1 hour ago'
+```
```
mavam, line 12: In addition to an example invocation, it would also be nice to have *output* examples here.

doc/cli/vast-start.md
```diff
@@ -1,3 +1,23 @@
...
+The `start` command spins up a VAST node. Starting a node is the first step when
+deploying VAST as a continuously running server. The process runs in the
 foreground and uses standard error for logging. Standard output remains unused.
+
+By default, the `vast start` command creates a `vast.db` directory in the
+current working directory. It is recommended to set the options for the node in
+the `vast.conf` file, such that they are picked up by all client commands as
+well.
+
+The `node` (sometimes called `server` or `server process`) is the core piece of
```
mavam, line 10: Actually, the node is not the same as the server. See [our docs](https://docs.tenzir.com/vast/architecture/system-structure) what I mean. We use "server" when talking about the entire `vast` process that runs via `vast start`. The "client" is the other process to access the server. The "node" is the main container actor inside the server that contains everything else.

mavam: I suggest using the wording as in the referenced docs.

doc/cli/vast-stop.md
```diff
@@ -1 +1,9 @@
-The `stop` command brings down a VAST node.
+The `stop` command gracefully brings down a VAST node, and is the analog of the
+`start` command.
+
+While it is technically possible to shut down a VAST node gracefully by sending
+`SIGINT(2)` to the `vast start` process, it is recommended to use `vast stop` to
+shut down the node, as it works over the wire as well and guarantees a proper
```
mavam, line 6:
```suggestion
shut down the server, as it works over the wire as well and guarantees a proper
```

doc/cli/vast-stop.md
```diff
@@ -1 +1,9 @@
...
+shutdown. The command blocks execution until the node has quit, and returns a
```
mavam, line 7:
```suggestion
shutdown. The command blocks execution until the server has quit, and returns a
```

There are probably a few more node vs. server terminology fixups needed.

dominiklohmann:
> * When did you choose to document the respective short options? Sometimes you write "`--node` / `-N`" and sometimes you only use the long version.

I've moved to using long options everywhere.

PR #814, opened 2020-03-31
List all VAST options in example vast.conf

dominiklohmann:


vast.conf
```diff
@@ -1,116 +1,380 @@
...
-;; The host and port to listen at and connect to.
-; endpoint = "localhost:42000"
+  ; The host and port to listen at and connect to.
+  ;endpoint = "localhost:42000"
+
+  ; The file system path used for persistent state.
+  ;db-directory = "vast.db/"
+
+  ; The file system path used for log files.
+  ;log-file = "vast.db/server.log"
```
mavam: Is this taking the variable `db-directory` or the hard-coded string `vast.db`?

dominiklohmann: It's `<db-directory>/server.log` by default. I should write it like that.

dominiklohmann:
@mavam do you consider this worth a changelog-entry? I don't know whether people are looking at the example configuration regularly.

mavam:
> do you consider this worth a changelog-entry?

No, it's just a small change in logging output.

dominiklohmann:
> No, it's just a small change in logging output.

You're thinking of the wrong PR, this is actually just a change in the example configuration.

mavam:
Yes, sorry, my tabs got messed up. 😅 

I would say this one can also go without changelog entry.

vast.conf
```diff
@@ -1,116 +1,380 @@
...
-;; The host and port to listen at and connect to.
-; endpoint = "localhost:42000"
+  ; The host and port to listen at and connect to.
+  ;endpoint = "localhost:42000"
+
+  ; The file system path used for persistent state.
+  ;db-directory = "vast.db/"
+
+  ; The file system path used for log files.
+  ;log-file = <server.log in db-directory>
```
mavam:
```suggestion
  ;log-file = "${db-directory}/server.log"
```

This seems a bit less ambiguous IMO. Maybe `<db-directory>/server.log`, but I would not combine text and variable in that fashion.

vast.conf
```diff
@@ -1,116 +1,380 @@
...
+
+  ; Number of events to be packaged in a table slice (this is a target value that
```
mavam:
```suggestion
  ; Number of events to be batched in a table slice (this is a target value that
```

vast.conf
```diff
@@ -1,116 +1,380 @@
...
+  ; can be underrun if the source has a low rate).
+  ;table-slice-size = 100
+
+  ; The table slice type (default|arrow).
```
mavam:
```suggestion
  ; The table slice type (default|arrow|msgpack).
```

It's okay to include his in here, even if the OSS version doesn't have it. When selecting this, the error message should "this is a pro feature".

vast.conf
```diff
@@ -1,116 +1,380 @@
...
+  ;table-slice-type = "default"
+
+  ; The size of an index shard.
+  ;max-partition-size = 1000000
+
+  ; The unique ID of this node.
+  ;node-id = "node"
```
mavam:
```suggestion
  ;node-id = "vast"
```

Does this make sense?

vast.conf
dominiklohmann: The angle brackets are consistent with the style used by CAF for values that are determined at runtime. The comment at the very top of the file explains what the angle brackets mean.

vast.conf
dominiklohmann: The default value is node. This PR merely documents the default values.

vast.conf
mavam: Okay, but is `in` a keyword? that confuses me. Having a `/` would make things more readable IMO.

dominiklohmann: As in `<db-directory/server.log>`?

dominiklohmann:
```suggestion
  ;log-file = "<db-directory>/server.log"
```

vast.conf
dominiklohmann: Let's discuss this in tomorrow's standup, as there are some additional options in commands in VAST Pro that are not mentioned here at all yet.

dominiklohmann: I think it's better not to block this PR from being merged because of this. I'm resolving this without adding `msgpack`—for now.

PR #815, opened 2020-03-31
Print a message when expecting input

dominiklohmann:
Prior to this change, the message was printed only after the reader was successfully initialized. When mistakenly starting VAST configured to read input from stdin, but not actually providing any data over stdin, it often looked like VAST was hanging.

### DIFF
--- a/libvast/vast/system/reader_command.hpp
+++ b/libvast/vast/system/reader_command.hpp
@@ -116,10 +116,13 @@ reader_command(const command::invocation& invocation, caf::actor_system& sys) {
     auto in = detail::make_input_stream(*file, uds);
     if (!in)
       return caf::make_message(std::move(in.error()));
+    if (*file == "-")
+      VAST_INFO_ANON("reader-command spawns reader for stdin");
+    else
+      VAST_INFO_ANON("reader-command spawns reader for file", *file);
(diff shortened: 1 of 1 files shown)


libvast/vast/system/reader_command.hpp
```diff
@@ -116,10 +116,13 @@ reader_command(const command::invocation& invocation, caf::actor_system& sys) {
     auto in = detail::make_input_stream(*file, uds);
     if (!in)
       return caf::make_message(std::move(in.error()));
+    if (*file == "-")
+      VAST_INFO_ANON("reader-command spawns reader for stdin");
+    else
+      VAST_INFO_ANON("reader-command spawns reader for file", *file);
```
mavam: Can you move this below the instantiation of `reader`? Then you can use the same logging as in the parent branch, i.e., `VAST_INFO(reader, ...)`.

dominiklohmann: That kinda defeats the point of this PR, as that is where the log message was located at previously, causing it not to get printed until input arrived.

mavam: Ah, I didn't think that the constructor already performs blocking I/O. This needs some refactoring in the future.

PR #816, opened 2020-03-31
Extend PyVAST for fluent command chaining

0ortmann:
- Command chaining via `.`-notation.
- Basic tests
- Python module
- Example code + README


pyvast/vast.py
```diff
@@ -1,55 +1,79 @@
...
+            proc.stdin.close()
+        self.call_stack = []
+        return await proc.communicate()
+
+    def __getattr__(self, name, **kwargs):
+        """Chains every unknown method call to the internal call stack."""
+        if name.startswith("_"):
+            # trim leading underscores to overcome the 'import' keyword
+            name = name[1:]
+        self.call_stack.append(name)
```
tobim, line 69: We should have a way to check if `name` is allowed in this place. We could use a tree represented as a nested dict for that.
```
commands = {
  "export": {
    "ascii": {},
    "csv": {},
    "json": {},
    "zeek": {}
  },
  "import": {
     ...
  },
  ...
}
```
For example.

0ortmann:
@tobim For the arrow support you could simply do it like this:
```py
stdout, stderr = vast.export().arrow("192.168.1.104")
reader = pyarrow.ipc.open_stream(stdout)
table = reader.read_all()
```

pyvast/pyvast/vast.py
```diff
@@ -0,0 +1,79 @@
...
+            proc.stdin.close()
+        self.call_stack = []
+        return await proc.communicate()
+
+    def __getattr__(self, name, **kwargs):
+        """Chains every unknown method call to the internal call stack."""
+        if name.startswith("_"):
+            # trim leading underscores to overcome the 'import' keyword
+            name = name[1:]
+        self.call_stack.append(name)
```
dominiklohmann, line 69: @tobim mentioned that this flexibility comes with the downside breaking stuff like autocompletion when working with PyVAST.

What if instead of deciding on either the dynamic or the hardcoded way, we kept the dynamic way, but additionally hardcode a few methods like `start`, `export` or `import`—the things that are done the most often.

0ortmann: @tobim and I discussed yesterday that for this MVP version of the library it is OK to not have autocompletion in IDEs. We already have a follow-up story for auto-generating a complete command skeleton from the binary (ch14424). That story also covers the ability to fail early and pass error messages from the python library to the user, rather than from the vast process itself (as it is now).

0ortmann:
@dominiklohmann I will update the demo notebook.

pyvast/README.md
```diff
@@ -0,0 +1,82 @@
...
+  print(stdout)
+  ```
+- Import a Zeek log file
+  ```sh
+  # CLI call
+  vast import zeek --read=/path/to/file
+  ```
+  ```py
+  # python wrapper
+  stdout, stderr = vast._import().zeek(read="/path/to/file").exec()
```
dominiklohmann, line 38:
```suggestion
  stdout, stderr = vast.import_().zeek(read="/path/to/file").exec()
```

In our C++ code we had the same issue with the `export` keyword, and usually we opt to add a trailing underscore instead of a leading underscore. Should this be consolidated?

pyvast/setup.py
```diff
@@ -0,0 +1,37 @@
...
+
+setup(
+    author="Tenzir",
+    author_email="engineering@tenzir.com",
+    classifiers=[
+        # https://pypi.org/classifiers/
+        "Development Status :: 3 - Alpha",
+        "License :: OSI Approved :: BSD License",
+        "Operating System :: MacOS :: MacOS X",
+        "Operating System :: POSIX :: Linux",
```
dominiklohmann, line 17: Should we add FreeBSD here?

pyvast/pyvast/vast.py
```diff
@@ -0,0 +1,79 @@
...
+
+        def method(*args, **kwargs):
+            if kwargs:
+                for k, v in kwargs.items():
+                    self.call_stack.append(f"--{k.replace('_','-')}={v}")
```
dominiklohmann, line 74: This does not properly work for flag parameters. The CAF CLI parser does not allow for `--disable-accounting=true`, but rather takes the presence of the flag as `true` and sets it to fault otherwise.

dominiklohmann:
```suggestion
                    if v is True:
                        self.call_stack.append(f"--{k.replace('_','-')}")                    
                    else:
                        self.call_stack.append(f"--{k.replace('_','-')}={v}")
```

pyvast/setup.py
0ortmann: Unsure about such a bold claim. Since VAST is currently not tested on FreeBSD due to GitHub action runner limitations, I would prefer to not list it here.

PR #820, opened 2020-04-03
Add new error code for "out of memory"

lava:
This will be used by some of the new matcher features in vast-pro

### DIFF
--- a/libvast/src/error.cpp
+++ b/libvast/src/error.cpp
@@ -54,6 +54,9 @@ const char* descriptions[] = {
   "out_of_memory",
 };
 
+static_assert(ec{std::size(descriptions)} == ec::ec_count,
+              "Mismatch between number of error codes and descriptions");
+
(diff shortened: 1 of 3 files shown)


libvast/src/error.cpp
```diff
@@ -51,8 +51,12 @@ const char* descriptions[] = {
   "missing_component",
   "unimplemented",
   "silent",
+  "out_of_memory",
 };
 
+static_assert(ec{sizeof(descriptions) / sizeof(*descriptions)} == ec::ec_count,
+              "Mismatch between number of error codes and descriptions");
```
dominiklohmann, line 58:
```suggestion
static_assert(ec{std::size(descriptions)} == ec::ec_count,
              "Mismatch between number of error codes and descriptions");
```

I think this is more idiomatic for modern C++, and also more resilient in case we switch to `std::array` in the future.

PR #821, opened 2020-04-06
Make PyVAST available via PyPI

0ortmann:
- [x] Makefile automates build
- [x] CI automates releases
- [x] CI tested: https://github.com/tenzir/vast/actions/runs/71644365
- [x] PyPI release generated by CI: https://pypi.org/project/pyvast/
- [x] README updated

### DIFF
--- a/pyvast/example/example.py
+++ b/pyvast/example/example.py
@@ -12,7 +12,9 @@ Example script to showcase the VAST CLI wrapper.
 Prerequisites: VAST must be running.
     0. Install VAST
     1. Go to another terminal
-    2. `vast start`
+    2. `vast start` (start a vast node)
+    3. Go to another terminal
+    4. `python -m pip install -r requirements.txt` (install requirements)
 """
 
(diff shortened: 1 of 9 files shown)


pyvast/example/example.py
```diff
@@ -12,7 +12,9 @@
 Prerequisites: VAST must be running.
     0. Install VAST
     1. Go to another terminal
-    2. `vast start`
+    2. `vast start` (start a vast node)
+    3. Go to another terminal
+    4. `python -m pip install -r requirements.txt` (install requirements)
```
mavam, line 17: You have committed a `Makefile`, would it make sense to reference it here?

0ortmann: The `requirements.txt` are specifically meant for the example. They install `pyarrow` and `pandas` to make the example work. The `Makefile` facilitates the build and dist setup for the PyVAST project itself and is not really needed here.

PR #822, opened 2020-04-06
Print schema load path at verbose log level

lava:


libvast/src/schema.cpp
```diff
@@ -205,9 +205,9 @@ caf::expected<schema> load_schema(const path& sf) {
 caf::expected<schema> load_schema(const std::vector<path>& schema_paths) {
   vast::schema types;
   for (const auto& dir : schema_paths) {
+    VAST_VERBOSE_ANON("Looking for schema files in", dir);
```
dominiklohmann, line 208:
```suggestion
  vast::schema types;
  VAST_VERBOSE_ANON("Looking for schema files in", schema_paths);
  for (const auto& dir : schema_paths) {
```

PR #823, opened 2020-04-06
Use a timeout when collecting a components status

dominiklohmann:
This is a bandaid to fix an issue where fetching the archive status takes too long and locks up the node. The real fix must be changing the archive to fix the issue, but having a timeout here is generally a valuable thing. Fetching a status is something that users generally expect to be a quick operation.


dominiklohmann:
I've force-pushed to fix a typo. Waiting for CI to run through before I merge this.

PR #824, opened 2020-04-08
Enable continuous queries in PyVAST

0ortmann:
With the previous implementation it was not possible to make use of VAST's `continuous query` feature. That was due to the use of [subprocess communicate()](https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.asyncio.subprocess.Process.communicate) which reads data from the process' `stdout` until `EOF` is reached. That behavior blocks indefinitely for continuous queries, since they won't close the pipe by design.

Handling the process and it's pipes is now left to the caller.


pyvast/README.md
```diff
@@ -35,7 +35,8 @@ and the analogous `pyvast` commands.
   ```
   ```py
   # python wrapper
-  stdout, stderr = vast.export(max_events=10).json("192.167.1.102").exec()
+  proc = await vast.export(max_events=10).json("192.167.1.102").exec()
+  stdout, stderr = await proc.communicate()
```
mavam, line 39: Just for my understanding: from a users perspective, I could imagine to have something like this:
```python
stdout, stderr =  await vast.export(max_events=10).json("192.167.1.102")
```
What is the reason for `.exec()` and the two-step with `.communicate()`?

0ortmann: That is the fix that allows to use continuous queries with the PyVAST wrapper. The user is presented the process handler `proc` directly, so the user can poll for new lines in `stdout`.
 
The example you highlighted uses `proc.communicate()` which blocks until EOF in stdout is reached. That was the default before this PR. Now it is up to user to wait (`communicate`) or rather poll `proc.stdout.readline`.

mavam: Makes sense, thanks.

PR #825, opened 2020-04-08
Make archive session extraction interruptible

dominiklohmann:


libvast/src/system/archive.cpp
```diff
@@ -77,94 +77,124 @@ archive(archive_type::stateful_pointer<archive_state> self, path dir,
   });
   self->set_down_handler([=](const down_msg& msg) {
     VAST_DEBUG(self, "received DOWN from", msg.source);
-    self->state.active_exporters.erase(msg.source);
+    auto& st = self->state;
+    st.active_exporters.erase(msg.source);
+    while (st.unhandled_ids.front().second == msg.source)
```
mavam, line 82: When using `.front()`, please add an assertion that the container is not empty.

libvast/src/system/archive.cpp
```diff
@@ -77,94 +77,124 @@ archive(archive_type::stateful_pointer<archive_state> self, path dir,
...
-                == 0) {
-              VAST_DEBUG(self, "dismisses query for inactive sender");
-              return make_error(ec::no_error);
+  return {
+    [=](const ids& xs) {
+      VAST_ASSERT(rank(xs) > 0);
+      VAST_DEBUG(self, "got query for", rank(xs),
+                 "events in range [" << select(xs, 1) << ','
+                                     << (select(xs, -1) + 1) << ')');
+      auto requester = caf::actor_cast<receiver_type>(self->current_sender());
```
mavam, line 97: I'm not sure if this is a safe operation. You cannot guarantee that the sender is `receiver_type`.

libvast/src/system/archive.cpp
```diff
@@ -77,94 +77,124 @@ archive(archive_type::stateful_pointer<archive_state> self, path dir,
...
+      self->send(self, xs, requester);
+    },
+    [=](const ids& xs, receiver_type requester) {
+      auto& st = self->state;
+      if (st.active_exporters.count(requester->address()) == 0) {
+        VAST_DEBUG(self, "dismisses query for inactive sender");
```
mavam, line 103: Should this ever happen? If not, I'd promote this to a warning?

libvast/src/system/archive.cpp
```diff
@@ -77,94 +77,124 @@ archive(archive_type::stateful_pointer<archive_state> self, path dir,
...
+          }
+          t.stop(events);
+        },
+        [=](unit_t&, const error& err) {
+          if (err) {
+            VAST_ERROR(self, "got a stream error:", self->system().render(err));
+          }
+        });
+    },
+    [=](exporter_atom, const actor& exporter) {
```
mavam, line 177: Why does the archive need to depend on the exporter?

libvast/src/system/archive.cpp
dominiklohmann: Yes, this does happen. We reviewed this in person.

libvast/src/system/archive.cpp
dominiklohmann: We talked about this in person: `receiver_type` defines a behavior that the caller must necessarily implement. I have added a check to see whether the cast succeeded.

dominiklohmann:
Pivot is working now, please review again @mavam.

PR #826, opened 2020-04-14
Use Homebrew to locate OpenSSL on macOS

dominiklohmann:
`openssl@1.0`, which was aliased to `openssl`, was recently removed from Homebrew. It was usually installed into `${HOMEBREW_PREFIX}/openssl`. There's now only `openssl@1.1`, which is now aliased to `openssl@1.1`, but  still installed into `${HOMEBREW_PREFIX}/openssl@1.1`. 

This PR circumvents this whole mess by just using `brew` to get the prefix directly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -487,11 +487,17 @@ if (VAST_ENABLE_ASSERTIONS)
   endif ()
 endif ()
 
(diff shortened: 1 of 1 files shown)


dominiklohmann:
@lava this does not install anything. It just gets the root directory of the existing installation.

lava:
@dominiklohmann Ah, that makes sense.

PR #827, opened 2020-04-15
Report PCAP packet loss

dominiklohmann:


libvast/vast/format/reader.hpp
```diff
@@ -13,16 +13,17 @@
 
 #pragma once
 
-#include <cstddef>
-#include <string_view>
+#include "vast/defaults.hpp"
+#include "vast/fwd.hpp"
+#include "vast/system/accountant.hpp"
```
tobim, line 18: Why does the `reader` interface need to know about the `accountant`?

libvast/vast/system/source.hpp
```diff
@@ -126,6 +126,12 @@ struct source_state {
   measurement measurement_;
 
   void send_report() {
+    // Send the reader-specific status report to the accountant.
+    if (auto status = reader.status(); !status.empty()) {
+      if (accountant)
+        self->send(accountant, std::move(status));
+    }
```
mavam, line 133:
```suggestion
    if (auto status = reader.status(); !status.empty())
      if (accountant)
        self->send(accountant, std::move(status));
```

Optional: omit the extra braces.

libvast/vast/system/source.hpp
```diff
@@ -126,6 +126,12 @@ struct source_state {
...
+    // Send the source-specific performance metrics to the accountant.
```
mavam, line 134: Would it make sense to consolidate the reporting into a single message?

libvast/src/format/pcap.cpp
```diff
@@ -106,6 +108,27 @@ const char* reader::name() const {
   return "pcap-reader";
 }
 
+vast::system::report reader::status() const {
+  using namespace std::string_literals;
+  auto stats = pcap_stat{};
```
mavam, line 113: I think we should also make sure that we are not reading from a trace here, i.e.,  make sure we are in open live mode. `pcap_stats` probably returns an error if this is not the case, but maybe we want to log those errors as warnings and disentangle them from things that are actual errors.

libvast/vast/format/pcap.hpp
```diff
@@ -101,6 +103,7 @@ class reader : public single_layout_reader {
   size_t snaplen_;
   bool community_id_;
   type packet_type_;
+  mutable pcap_stat last_stats_;
```
mavam, line 106: Is this correctly default-initialized or does it need a `memset`?

libvast/src/format/pcap.cpp
```diff
@@ -106,6 +108,27 @@ const char* reader::name() const {
...
+  if (auto res = pcap_stats(pcap_, &stats); res != 0)
+    return {};
+  uint64_t recv = stats.ps_recv - last_stats_.ps_recv;
+  uint64_t drop = stats.ps_drop - last_stats_.ps_drop;
+  uint64_t ifdrop = stats.ps_ifdrop - last_stats_.ps_ifdrop;
+  // Warn if at least 5% were dropped since the last report.
+  if ((drop + ifdrop) * 20 >= recv)
```
mavam, line 120: Can we make this threshold configurable?

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🎁 Packet drop statistics are now reported to the accountant for the
+  pcap-reader, and are available using the keys `pcap-reader.recv`,
+  `pcap-reader.drop`, and `pcap-reader.ifdrop`. If the number of dropped packets
+  exceeds 5% of the total packets, VAST additionally reports the packet drop on
+  the command line. [#827](https://github.com/tenzir/pull/827)
```
```suggestion
- 🎁 Packet drop statistics are now reported to the accountant for PCAP
  import, and are available using the keys `pcap-reader.recv`,
  `pcap-reader.drop`, and `pcap-reader.ifdrop` in the `vast.statistics` event. If the
  number of dropped packets exceeds a configurable threshold, VAST additionally
  warns about packet drops on the command line.
  [#827](https://github.com/tenzir/pull/827)
```

libvast/vast/system/source.hpp
dominiklohmann: One is a `report` and the other one is a `performance_report`.

libvast/vast/format/reader.hpp
dominiklohmann: Essentially because that's where `system::report` lives, but I can just as easily move this into a separate header.

libvast/vast/system/source.hpp
dominiklohmann:
```suggestion
    if (auto status = reader.status(); !status.empty() && accountant) {
      self->send(accountant, std::move(status));
    }
```

libvast/vast/format/pcap.hpp
dominiklohmann: This is default-initialized in the constructor of the reader.

libvast/src/format/pcap.cpp
dominiklohmann: It's trivial to implement, but do we want to add yet another option to the PCAP reader for this?

libvast/src/format/pcap.cpp
dominiklohmann: `pcap_stats` returns a non-zero exit code when not reading from an interface according to the documentation.

libvast/src/format/pcap.cpp
mavam: In this case, yes, because everybody has different notions of what acceptable packet loss is.

libvast/src/format/pcap.cpp
mavam: Okay, so the question is: if we read from an interface and `pcap_stats` returns an error, do we want to log it (as warning) or it ignore it silently?

dominiklohmann: I think we want to ignore it silently, as this function is very much expected fail, e.g., when not reading from an interface.

PR #830, opened 2020-04-17
Remove assertion from expression tailoring

lava:
Removed an assertion that checked an expression could
not become 'none_t' after running the type resolver on
it.

Since the type resolver is using 'none_t' as a return value
to signal that it was not able to resolve a type, this
condition should not be an error, let alone an assertion
failure.

For example, an expression like 'foo.bar = "baz"' would always
return 'none' if the type to be matched against was not called
foo or had no field called bar.

### DIFF
--- a/libvast/src/expression.cpp
+++ b/libvast/src/expression.cpp
@@ -206,9 +206,7 @@ caf::expected<expression> tailor(const expression& expr, const type& t) {
   auto x = caf::visit(type_resolver{t}, expr);
   if (!x)
     return x.error();
-  *x = caf::visit(type_pruner{t}, *x);
-  VAST_ASSERT(!caf::holds_alternative<caf::none_t>(*x));
-  return std::move(*x);
+  return caf::visit(type_pruner{t}, *x);
 }
 
 namespace {
--- a/libvast/src/system/counter.cpp
+++ b/libvast/src/system/counter.cpp
@@ -52,16 +52,18 @@ void counter_state::init(expression expr, caf::actor index,
   behaviors_[collect_hits] = base.or_else(
     [this](table_slice_ptr slice) {
       // Construct a candidate checker if we don't have one for this type.
-      auto& checker = checkers_[slice->layout()];
-      if (caf::holds_alternative<caf::none_t>(checker)) {
+      auto it = checkers_.find(slice->layout());
+      if (it == checkers_.end()) {
         if (auto x = tailor(expr_, slice->layout())) {
-          checker = std::move(*x);
+          std::tie(it, std::ignore) = checkers_.emplace(
+            vast::record_type{slice->layout()}, std::move(*x));
         } else {
           VAST_ERROR(self_, "failed to tailor expression:",
                      self_->system().render(x.error()));
           return;
         }
       }
+      auto& checker = it->second;
       // Performance candidate checks for all selected rows.
       uint64_t num_hits = 0;
       auto candidates = to_events(*slice, hits_);
--- a/libvast/test/system/exporter.cpp
+++ b/libvast/test/system/exporter.cpp
@@ -213,4 +213,20 @@ TEST(continuous query with importer) {
   CHECK_EQUAL(results.back().id(), 19u);
 }
 
+TEST(continuous query with mismatching importer) {
+  MESSAGE("prepare importer");
+  importer_setup();
+  MESSAGE("prepare exporter for continous query");
+  expr = unbox(to<expression>("foo.bar == \"baz\""));
+  exporter_setup(continuous);
+  send(importer, system::exporter_atom::value, exporter);
+  MESSAGE("ingest conn.log via importer");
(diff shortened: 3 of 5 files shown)


libvast/src/expression.cpp
```diff
@@ -206,9 +206,8 @@ caf::expected<expression> tailor(const expression& expr, const type& t) {
   auto x = caf::visit(type_resolver{t}, expr);
   if (!x)
     return x.error();
-  *x = caf::visit(type_pruner{t}, *x);
-  VAST_ASSERT(!caf::holds_alternative<caf::none_t>(*x));
-  return std::move(*x);
+  auto result = caf::visit(type_pruner{t}, *x);
```
lava, line 209: I thought about inserting

```
  if (caf::holds_alternative<caf::none_t>(*x))
    return make_error(ec::unspecified, "could not resolve type");
```

but its generating a lot of warning-level noise, and I'm also not sure that this is actually an error.

libvast/src/expression.cpp
```diff
@@ -206,9 +206,8 @@ caf::expected<expression> tailor(const expression& expr, const type& t) {
...
+  return result;
```
mavam, line 210:
```suggestion
return caf::visit(type_pruner{t}, *x));
```

This could work.

Tailoring is a two-step process:

1. Resolve all type and key extractors
2. Remove all unneeded parts of the expression for the given type

It seems to me that (2) should be optional. If nothing can be pruned, then the pruning step should return an identity, i.e, the input expression. The output should `none` *iff* the input is *none*. To me, this seems to be the underlying bug.

Does that make sense?

lava: I'm not sure if I understand what "unneeded" means here: From looking at the code, I got the impression that we're trying to remove all parts of the expression that do not apply to the current type, because they would always return "false". (ie. in `foo.bar == "baz" || type.key == "asdf"`)

But then if no parts apply, this would mean that *nothing* is needed and the output should become `none`, no?

mavam: > But then if no parts apply, this would mean that nothing is needed and the output should become none, no?

I think this is right, I had a different notion in mind. So the output *can* be `none` if everything gets "pruned away". Then the expression is effectively empty and will never match anything.

In historical queries, this in an "early abort". In continuous queries, it can't be. I think the reason for the assert comes from considering only the historical case.

We probably have to adjust the call site of `tailor` to do the right thing. But regardless, the assertion should be removed.

libvast/src/system/counter.cpp
```diff
@@ -52,16 +52,18 @@ void counter_state::init(expression expr, caf::actor index,
...
     [this](table_slice_ptr slice) {
       // Construct a candidate checker if we don't have one for this type.
-      auto& checker = checkers_[slice->layout()];
-      if (caf::holds_alternative<caf::none_t>(checker)) {
+      auto it = checkers_.find(slice->layout());
+      if (it == checkers_.end()) {
         if (auto x = tailor(expr_, slice->layout())) {
-          checker = std::move(*x);
+          bool _;
+          std::tie(it, _) = checkers_.emplace(vast::record_type{slice->layout()}, std::move(*x));
```
dominiklohmann, line 59:
```suggestion
          std::tie(it, std::ignore) = checkers_.emplace(vast::record_type{slice->layout()}, std::move(*x));
```

Either that, or just:

```suggestion
          it = checkers_.emplace(vast::record_type{slice->layout()}, std::move(*x)).first;
```

libvast/src/expression.cpp
dominiklohmann, line 210:
```suggestion
  return caf::visit(type_pruner{t}, *x);
```

libvast/test/system/exporter.cpp
```diff
@@ -213,4 +213,20 @@ TEST(continuous query with importer) {
...
+  expr = unbox(to<expression>("foo.bar == \"baz\""));
+  exporter_setup(continuous);
+  send(importer, system::exporter_atom::value, exporter);
+  MESSAGE("ingest conn.log via importer");
+  // Again: copy because we musn't mutate static test data.
+  vast::detail::spawn_container_source(sys, zeek_conn_log_slices, importer);
+  run();
+  MESSAGE("fetch results");
+  auto results = fetch_results();
+  REQUIRE_EQUAL(results.size(), 0u);
```
dominiklohmann, line 229:
```suggestion
  CHECK_EQUAL(results.size(), 0u);
```

No need to abort the other tests because this one is failing.

libvast/src/system/counter.cpp
mavam: The latter one feels less noisy to me.

mavam:
Green light from my end.

dominiklohmann:
Quick question before this gets merged: Should this get a changelog entry?

mavam:
If it's fixing a user-facing error, then yes. Looks like this is the case here.

PR #832, opened 2020-04-20
Add function for actor lookup by label and type

lava:
Add a new function that can look up an actor from a given
label and type, which should uniquely identify any one
actor.

### DIFF
--- a/libvast/vast/system/node_control.hpp
+++ b/libvast/vast/system/node_control.hpp
@@ -33,6 +33,8 @@ spawn_at_node(caf::scoped_actor& self, caf::actor node, Arguments&&... xs) {
   return result;
 }
 
+/// Look up components by category. Returns the first actor of each
+/// category name passed in `names`.
 template <size_t N>
 caf::expected<std::array<caf::actor, N>>
 get_node_components(caf::scoped_actor& self, caf::actor node,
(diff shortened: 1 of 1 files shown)


libvast/vast/system/node_control.hpp
```diff
@@ -55,4 +57,23 @@ get_node_components(caf::scoped_actor& self, caf::actor node,
   return result;
 }
 
+/// Look up a node component by component type and label
```
mavam, line 60: Could you document the function fully? I.e., `@param` and `@returns` as well?

lava: I'd suggest we discuss first whether we really want to establish this as a standard practice. For user-facing API functions I'd definitely say yes, as well as for functions doing obscure things. 

However, there's some balance to consider, since I'd also like to avoid adding trivial comments like

```
    @param node The node containing the component
    @param type The type of the component
    @param label The label of the component
```

all over our codebase.

lava:
@mavam My use case was in the matcher: When starting a matcher you have `category == "matcher"` and `label == "matcher-123`. So when multiple matchers are started, you need to use the label to distinguish between them.

The label is well-known after a matcher is started, so the user can pass it as an argument to specify *which* matcher he's talking about.

libvast/vast/system/node_control.hpp
mavam: Good point. We'll discuss the "documentation style" in a separate context.

PR #833, opened 2020-04-21
Add Bash autocompletion for VAST

lava:


scripts/vast-completions.bash
```diff
@@ -0,0 +1,26 @@
...
+    # We need to special-case '--', since that has special
+    # meaning, e.g.
+    #
+    #    vast pivot --asdf help -> displays error message/help text
+    #    vast pivot -- help     -> runs pivot command
+    #
+    # Technically we should scan the whole COMP_WORDS array
+    # for '--', but that issue should come up rarely, wheras
+    # a '--' at the end happens all the time as the result of
+    # a previous autocompletion.
```
dominiklohmann, line 16: For zsh, this can be done using `if $(( COMP_WORDS[(Ie)--] )); then ... fi`, but I'm not quite sure that's as easy for bash. I still think this should be fixed, shouldn't be much more than a simple loop.

scripts/vast-completions.bash
```diff
@@ -0,0 +1,26 @@
...
+    completion_words[$((${#COMP_WORDS}-1))]="-"
+  fi
+  tmpfile=$(mktemp)
+  ${completion_words[@]} help 2>&1 |  awk '/subcommands:/{x = 1} !/subcommands:/{if (x > 0) print $1} match($0, /--[^\]]*]/) { print substr($0, RSTART, RLENGTH-1) }' >${tmpfile}
+  readarray -t commands < ${tmpfile}
```
dominiklohmann, line 21:
```suggestion
  readarray -t commands < ${tmpfile}
```

This is the only really non-portable line between zsh and bash in the whole script (assuming `autoload bashcompinit`). Can we get rid of `readarray` somehow?

scripts/vast-completions.bash
```diff
@@ -0,0 +1,26 @@
...
+    #    vast pivot --asdf help -> displays error message/help text
+    #    vast pivot -- help     -> runs pivot command
+    #
+    # Technically we should scan the whole COMP_WORDS array
+    # for '--', but that issue should come up rarely, wheras
+    # a '--' at the end happens all the time as the result of
+    # a previous autocompletion.
+    completion_words[$((${#COMP_WORDS}-1))]="-"
+  fi
+  tmpfile=$(mktemp)
```
dominiklohmann, line 19: What does using a temporary file achieve here? We should not need to write anything to disk here, this is just a completion script, and it should work on a read-only filesystem as well.

lava: It's doing the conversion `string with newlines` -> `bash array`, where each array element should be one line of the input string. Please share if you know of a better way to do this :)

dominiklohmann:
```bash
IFS=$'\n' read -rd '' -a bash_array <<<"$string_with_newlines"
```

This should work and is portable between zsh 4+ and bash 3.2+

dominiklohmann: This is portable between zsh and bash.

```bash
IFS=$'\n' y=($(<command that outputs string with newlines>))
```

lava: Nice, added an adapted version (for some reason, without the `$SAVEIFS` workaround both this version and one with `local IFS=...` would break the subsequent `compgen` call; maybe this is a bash bug because IFS is a special built-in variable)

scripts/vast-completions.bash
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
dominiklohmann, line 1:
```suggestion
#!/usr/bin/env bash
```

lava:
@dominiklohmann Oh right, I was missing spaces around the "--" when scanning COMP_WORDS for existing instances. It should be fixed now.

PR #834, opened 2020-04-22
Fix inequality port lookups

mavam:
This PR fixes a bug when performing queries of the form `x != 80/tcp`. The query was falsely evaluated as `x != 80/? && x != ?/tcp`. (The syntax in the second predicate does not yet exist.) The correct evaluation should be `x != 80/? || x != ?/tcp`, i.e., we want to include `80/udp` and `80/?` in the result as well, which requires performing a disjunction of the port type.


mavam:
I've added to more integration tests. The testing you are describing is actually already covered by the unit tests, I think.

PR #835, opened 2020-04-22
Use line reader timeout

dominiklohmann:
This fixes a bug that caused continuous Suricata imports to possibly hang.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,12 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
-- 🐞 Continuously importing JSON, CSV, and Syslog events now properly works at
-  low rates (e.g., when batched over the network).
-  [#835](https://github.com/tenzir/vast/pull/835)
+- 🐞 Fixed a bug that could cause stalled input streams not to forward events to
+  the index and archive components for the JSON, CSV, and Syslog readers, when
+  the input stopped arriving but no EOF was sent. This is a follow-up to
+  [#750](https://github.com/tenzir/vast/pull/750). A timeout now ensures that
+  that the readers continue when some events were already handled, but the input
+  appears to be stalled. [#835](https://github.com/tenzir/vast/pull/835)
 
 - 🐞 Queries of the form `x != 80/tcp` were falsely evaluated as
   `x != 80/? && x != ?/tcp`. (The syntax in the second predicate does not yet
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 Continuously importing JSON, CSV, and Syslog events now properly works at
+  low rates (e.g., when batched over the network).
+  [#835](https://github.com/tenzir/vast/pull/835)
+
 - 🐞 Queries of the form `x != 80/tcp` were falsely evaluated as
   `x != 80/? && x != ?/tcp`. (The syntax in the second predicate does not yet
   exist; it only illustrates the bug.) Port inequality queries now correctly
--- a/libvast/src/format/csv.cpp
+++ b/libvast/src/format/csv.cpp
@@ -410,6 +410,14 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
                              consumer& callback) {
   VAST_ASSERT(max_events > 0);
   VAST_ASSERT(max_slice_size > 0);
+  bool timeout = false;
+  auto next_line = [&] {
+    if (!builder_ || builder_->rows() == 0)
+      lines_->next();
+    else
+      timeout = lines_->next_timeout(
+        vast::defaults::import::shared::partial_slice_read_timeout);
+  };
   if (!parser_) {
     auto p = read_header(lines_->get());
     if (!p)
@@ -417,9 +425,12 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
     parser_ = *std::move(p);
   }
   auto& p = *parser_;
-  size_t produced = 0;
-  lines_->next();
-  for (; produced < max_events; lines_->next()) {
+  next_line();
+  for (size_t produced = 0; produced < max_events; next_line()) {
+    if (timeout) {
+      VAST_DEBUG(this, "reached input timeout at line", lines_->line_number());
+      return finish(callback);
+    }
     // EOF check.
     if (lines_->done())
       return finish(callback, make_error(ec::end_of_input, "input exhausted"));
--- a/libvast/src/format/csv.cpp
+++ b/libvast/src/format/csv.cpp
@@ -410,13 +410,14 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
                              consumer& callback) {
   VAST_ASSERT(max_events > 0);
   VAST_ASSERT(max_slice_size > 0);
-  bool timeout = false;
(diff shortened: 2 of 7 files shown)


libvast/vast/defaults.hpp
```diff
@@ -78,6 +78,10 @@ struct json {
   /// Nested category in config files for this subcommand.
   static constexpr const char* category = "import.json";
 
+  /// Time that the reader waits for new data before it finishes a partial slice.
+  static constexpr std::chrono::milliseconds partial_slice_read_timeout
+    = std::chrono::milliseconds{500};
+
```
tobim, line 84: I don't think we need to define this default for each reader individually, so we could just have one in the `import::shared` namespace.

dominiklohmann: Fair point.

libvast/vast/format/json.hpp
```diff
@@ -156,7 +156,20 @@ caf::error reader<Selector>::read_impl(size_t max_events, size_t max_slice_size,
   VAST_ASSERT(max_events > 0);
   VAST_ASSERT(max_slice_size > 0);
   size_t produced = 0;
-  for (; produced < max_events; lines_->next()) {
+  table_slice_builder_ptr bptr = nullptr;
+  bool timeout = false;
+  auto next_line = [&] {
+    if (!bptr || bptr->rows() == 0)
```
lava, line 162: Maybe we should add a warning here why we need the special case for `rows() == 0`, since it apparently wasnt obvious from reading the same logic in the zeek reader.

dominiklohmann: That's a fair point, but I think since we plan to rework this in the near future anyways with a central place for input management, this does not need to be done.

dominiklohmann:
@tenzir/backend I've adapted this PR to fix the issue for CSV and Syslog as well. Please re-review. Still running some tests locally to verify the issue is fixed for the other formats as well.

libvast/src/format/syslog.cpp
```diff
@@ -95,51 +95,58 @@ const char* reader::name() const {
...
 reader::read_impl(size_t max_events, size_t max_slice_size, consumer& f) {
-  size_t produced = 0;
-  while (produced < max_events) {
+  bool timeout = false;
+  table_slice_builder_ptr bptr = nullptr;
+  auto next_line = [&] {
+    if (!bptr || bptr->rows() == 0)
+      lines_->next();
+    else
+      timeout = lines_->next_timeout(
```
tobim, line 104: `timeout` is never read?

libvast/src/format/csv.cpp
```diff
@@ -410,16 +410,23 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
                              consumer& callback) {
   VAST_ASSERT(max_events > 0);
   VAST_ASSERT(max_slice_size > 0);
+  bool timeout = false;
+  auto next_line = [&] {
+    if (!builder_ || builder_->rows() == 0)
+      lines_->next();
+    else
+      timeout = lines_->next_timeout(
```
tobim, line 418: Same here, don't we need to return if `timeout` is `true`?

libvast/src/format/csv.cpp
```diff
@@ -410,16 +410,23 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
...
     if (!p)
       return p.error();
     parser_ = *std::move(p);
   }
   auto& p = *parser_;
-  size_t produced = 0;
-  lines_->next();
-  for (; produced < max_events; lines_->next()) {
+  next_line();
+  for (size_t produced = 0; produced < max_events; next_line()) {
```
lava, line 429: It looks like we're never checking `timeout` in the csv reader?

libvast/src/format/csv.cpp
```diff
@@ -410,16 +410,23 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
...
   VAST_ASSERT(max_events > 0);
   VAST_ASSERT(max_slice_size > 0);
+  bool timeout = false;
+  auto next_line = [&] {
+    if (!builder_ || builder_->rows() == 0)
+      lines_->next();
+    else
+      timeout = lines_->next_timeout(
+        vast::defaults::import::shared::partial_slice_read_timeout);
+  };
```
lava, line 420: I'd suggest to either pass in timeout as a `bool&` out-parameter or to always `return false` in the first case, passing by implicit out-reference feels a bit odd.

libvast/src/format/csv.cpp
dominiklohmann: Fixed both, I accidentally dropped that when rebasing. Thanks for noticing.

libvast/src/format/csv.cpp
lava: Ok, looks like @tobim was faster than me here :D

libvast/src/format/csv.cpp
dominiklohmann: I don't think this is worth refactoring, since this abstraction must necessarily change soon-ish anyways.

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 Continuously importing JSON, CSV, and Syslog events now properly works at
+  low rates (e.g., when batched over the network).
```
mavam, line 14: Can we have a clear description of the bug first? It's not apparent to me what was the issue, i.e., was the opposite the bug or something related?

dominiklohmann: This ports the fix from #750 over to other readers. The timeout ensures that we continue when some events were already handled, but the input stream stalls.

lava: Actually at "low rates" this is actually less severe, since at least we will eventually get all data. In this case we could just reduce the table slice size to match the input rate. 

The worse bug is when the input is already finished, but we didn't get EOF (e.g. because its going through some pipe indirections), then the data that is already parsed in the partial table slice is just swallowed and is never forwarded to index/archive.

dominiklohmann: I have improved the changelog wording accordingly.

dominiklohmann:
> Looks good to me, although I still think we should have a warning about the "Thou shalt not return 0" bug in a comment somewhere.

@lava see also:
https://github.com/tenzir/vast/blob/9882c2f950fa66f31f79bb791419b9bc7170c638/libvast/vast/system/source.hpp#L185-L188

libvast/vast/format/json.hpp
```diff
@@ -159,13 +159,15 @@ caf::error reader<Selector>::read_impl(size_t max_events, size_t max_slice_size,
...
-    else
-      timeout = lines_->next_timeout(
+      return false;
+    } else {
+      return lines_->next_timeout(
         vast::defaults::import::shared::partial_slice_read_timeout);
+    }
   };
-  for (; produced < max_events; next_line()) {
+  for (; produced < max_events; timeout = next_line()) {
```
tobim, line 170: I would prefer if we could reduce the complexity here to
```
while (produced < max_events) {
  if (!next_line()) {
    VAST_DEBUG(this, "reached input timeout at line", lines_->line_number());
    return finish(cons);
  }

For that to work, the constructor of `line_range` needs to be changed to not call `next()` automatically.
}
```

dominiklohmann: We decided via Slack that this is not worth spending time on right now because the readers will be refactored to operate on chunks in the near future anyways.

PR #838, opened 2020-04-23
Expose VAST_ENABLE_ASSERTIONS as cmake option

lava:


CMakeLists.txt
```diff
@@ -80,6 +80,7 @@ option(VAST_ENABLE_ASSERTIONS "Enable Assertions" "${VAST_ENABLE_ASSERTIONS_DEFA
 option(VAST_RELOCATABLE_INSTALL "Enable relocatable installations" ON)
 option(ENABLE_ZEEK_TO_VAST "Build zeek-to-vast" ON)
 option(VAST_STATIC_EXECUTABLE "Link VAST statically")
+option(VAST_ENABLE_ASSERTIONS "Enable assertions" ON)
```
dominiklohmann, line 83: What's wrong with the same option that's exposed just 4 lines above?

lava: Oh, interesting, I didn't notice that @tobim had already merged the same change into master two days ago when I rebased my topic branch. So I this one can just be closed now.

lava:
Made obsolete since the change was already included in #728

PR #841, opened 2020-04-23
Fix comparison warning

mavam:
This is a follow-up PR that should have been part of #836.


mavam:
Weird. CI didn't connect in this PR. Probably due to the [current github issue](https://www.githubstatus.com/incidents/2y6v4ltq26g7). I've merged this nonetheless because it's easy to reason about.

PR #844, opened 2020-04-28
Keep track of discarded packets

dominiklohmann:


dominiklohmann:
I have opted to add to the existing changelog entry about the drop rates in the unreleased section.

PR #847, opened 2020-04-28
Correct check for user schema in zeek reader

tobim:


libvast/src/format/zeek.cpp
```diff
@@ -475,6 +479,9 @@ caf::error reader::parse_header() {
   }
   // Add #index=hash attribute for fields where it makes sense.
   add_hash_index_attribute(layout_);
```
dominiklohmann, line 481: Should the `#index=hash` attributes be reflected in the schema instead?

libvast/src/format/zeek.cpp
```diff
@@ -475,6 +479,9 @@ caf::error reader::parse_header() {
...
+  for (auto i = 0u; i < layout_.fields.size(); ++i)
+    VAST_DEBUG(this, "     ", i << ')', layout_.fields[i].name << ':',
+               layout_.fields[i].type);
```
dominiklohmann, line 484: Do the spaces in the debug message have a purpose?

libvast/src/format/zeek.cpp
```diff
@@ -445,18 +445,22 @@ caf::error reader::parse_header() {
...
-  if (auto t = schema_.find(path))
-    if (t->name() == path) {
-      if (congruent(type_, *t))
-        type_ = *t;
-      else
-        return make_error(ec::format_error, "incongruent types in schema");
+  if (auto t = schema_.find(layout_.name())) {
+    auto r = caf::get_if<record_type>(t);
+    if (!r)
+      return make_error(ec::format_error, "record_type required");
```
dominiklohmann, line 453: Since error messages are user facing, I think this should be reworded not to mention the internal concept of a `record_type`.

libvast/src/format/zeek.cpp
tobim: I only moved that part down a couple of lines.

libvast/src/format/zeek.cpp
tobim: `record` is a context dependent keyword in the schema language. But I will reword this message to be more informative and understandable.

libvast/src/format/zeek.cpp
tobim: I think they should, but it still makes sense to add this attribute for fields that are not part of the user defined schema. In the current implementation, one would have to supply `#index=default` to avoid these fields to automatically get the hash index.

CHANGELOG.md
```diff
@@ -8,7 +8,12 @@ Every entry has a category for which we use the following visual abbreviations:
 - 🔄 = change
 - 🐞 = bugfix
 
-<!-- ## Unreleased -->
+## Unreleased
+
+- 🐞 The parser for Zeek tsv data used to ignore attributes that were defined
+  for the zeek specific types in the schema files. It has been modified to
+  respect and prefer the specified attributes for the fields that are present
+  in the input data. [#847](https://github.com/tenzir/vast/pull/847)
```
dominiklohmann, line 16:
```suggestion
- 🐞 The parser for Zeek tsv data used to ignore attributes that were defined
  for the Zeek-specific types in the schema files. It has been modified to
  respect and prefer the specified attributes for the fields that are present
  in the input data. [#847](https://github.com/tenzir/vast/pull/847)
```

PR #852, opened 2020-04-30
Fix unit test compilation failure for clang 10

dominiklohmann:


dominiklohmann:
I have updated the commit message accordingly. Will ask upstream via gitter.

PR #853, opened 2020-05-05
Minor build fixes for uncommon configurations

tobim:


dominiklohmann:
A related thing I've noticed: can you removed `libvast/src/system/application.cpp:25`? It breaks the build without PCAP. A conditional include already exists in the same file.

PR #854, opened 2020-05-05
Add a script to inspect the index disk utilization

tobim:


mavam:
Shouldn't this be output from `vast status` or `vast -N -d <vast.db> status`?

tobim:
> Shouldn't this be output from `vast status` or `vast -N -d <vast.db> status`?

Maybe it should, but this version does not depend on a working vast and non-corrupted meta files. Is there a reason not to have both?

mavam:
> Is there a reason not to have both?

The fewer code we have to maintain, the better. So in general I would prefer to avoid duplication of functionality. But having yet another script is not substantial technical debt.

I can see the issue with corruption, but the scenario of having an index but not VAST seems strange. If files are corrupted VAST shouldn't crash either. (Granted, we're not there yet.)

lava:
From an operators perspective I think another value of having a separate tool is that you can trust it to treat the database as `const` and not make the problem worse; running `vast` on an already corrupted database seems risky even after we get to the point where we don't crash anymore.

PR #856, opened 2020-05-06
Flatbufferize meta index state

mavam:
This PR migrates the index state to flatbuffers.

### DIFF
--- a/libvast/src/io/write.cpp
+++ b/libvast/src/io/write.cpp
@@ -1,0 +1,33 @@
+/******************************************************************************
+ *                    _   _____   __________                                  *
+ *                   | | / / _ | / __/_  __/     Visibility                   *
+ *                   | |/ / __ |_\ \  / /          Across                     *
+ *                   |___/_/ |_/___/ /_/       Space and Time                 *
+ *                                                                            *
+ * This file is part of VAST. It is subject to the license terms in the       *
+ * LICENSE file found in the top-level directory of this distribution and at  *
+ * http://vast.io/license. No part of VAST, including this file, may be       *
+ * copied, modified, propagated, or distributed except according to the terms *
+ * contained in the LICENSE file.                                             *
+ ******************************************************************************/
+
+#include "vast/io/write.hpp"
+
+#include "vast/error.hpp"
+#include "vast/filesystem.hpp"
+
+namespace vast::io {
+
+caf::error write(const path& filename, span<const byte> xs) {
+  file f{filename};
+  if (!f.open(file::write_only))
+    return make_error(ec::filesystem_error, "failed open file");
+  size_t bytes_written;
+  if (!f.write(xs.data(), xs.size(), &bytes_written))
+    return make_error(ec::filesystem_error, "failed to write chunk");
+  if (bytes_written != xs.size())
+    return make_error(ec::filesystem_error, "incomplete write");
+  return caf::none;
+}
+
+} // namespace vast::io
--- a/libvast/vast/fbs/utils.hpp
+++ b/libvast/vast/fbs/utils.hpp
@@ -38,10 +38,16 @@ namespace vast::fbs {
 /// @returns The buffer of *builder*.
 chunk_ptr release(flatbuffers::FlatBufferBuilder& builder);
 
-/// Creates a verifier for a chunk.
+/// Creates a verifier for a byte buffer.
 /// @chk The chk to initialize the verifier with.
 /// @param A verifier that is ready to use.
-flatbuffers::Verifier make_verifier(chunk_ptr chk);
+[[deprecated("use make_verifier(as_bytes(xs)) instead")]] flatbuffers::Verifier
+make_verifier(chunk_ptr chk);
+
+/// Creates a verifier for a byte buffer.
+/// @xs The buffer to create a verifier for.
+/// @param A verifier that is ready to use.
+flatbuffers::Verifier make_verifier(span<const byte> xs);
 
 /// Performs a check whether two given versions are equal and returns an error
 /// if not.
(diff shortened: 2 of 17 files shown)


libvast/src/io/write.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+caf::error write(const path& filename, span<const byte> xs) {
+  file f{filename};
+  if (!f.open(file::write_only))
+    return make_error(ec::filesystem_error, "failed open file");
+  size_t bytes_written;
+  if (!f.write(xs.data(), xs.size(), &bytes_written))
+    return make_error(ec::filesystem_error, "failed to write chunk");
+  if (bytes_written != xs.size())
+    return make_error(ec::filesystem_error, "incomplete write");
```
dominiklohmann, line 29: In case of an incomplete write, should we try to remove the written bytes?

libvast/vast/fbs/utils.hpp
```diff
@@ -71,11 +79,57 @@ pack_bytes(flatbuffers::FlatBufferBuilder& builder, const T& x) {
...
+  flatbuffers::Verifier verifier{data, size};
+  if (!verifier.template VerifyBuffer<Flatbuffer>(identifier))
+    return nullptr;
+  return flatbuffers::GetRoot<Flatbuffer>(data);
+}
+
+/// Wraps an object into a flatbuffer.
+template <class T>
+caf::expected<chunk_ptr>
+wrap(T const& x, const char* file_identifier = nullptr) {
```
tobim, line 111: The `file_identifier` thing is not clear to me. Why can it be `nullptr`? If we want to compose builders for nested tables it should happen inside of `pack`, no?

libvast/src/io/write.cpp
mavam: This may not be possible. The call site could remove the corrupted file, though.

libvast/vast/table_slice.hpp
```diff
@@ -207,6 +208,19 @@ class table_slice : public caf::ref_counted {
 
 // -- free functions -----------------------------------------------------------
 
+/// Packs a table slice into a flatbuffer.
+/// @param builder The builder to pack *x* into.
+/// @param x The table slice to pack.
+/// @returns The flatbuffer offset in *builder*.
+caf::expected<flatbuffers::Offset<fbs::TableSliceBuffer>>
+pack(flatbuffers::FlatBufferBuilder& builder, table_slice_ptr x);
```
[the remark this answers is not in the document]
mavam: The idea was that these are customization points and found via ADL. Therefore they should stay in the same namespace as the type definition.

libvast/vast/fbs/utils.hpp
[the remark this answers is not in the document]
mavam: Yes, good point.

mavam: I've added a note in the documentation. Let me know whether that suffices.

libvast/vast/meta_index.hpp
```diff
@@ -81,8 +81,15 @@ class meta_index {
...
 
-  /// The factory function to construct a synopsis structure for a type.
+  /// Settings for the synopsis factory.
   caf::settings synopsis_options_;
 };
 
+// -- flatbuffer ---------------------------------------------------------------
+
+caf::expected<flatbuffers::Offset<fbs::MetaIndex>>
+pack(flatbuffers::FlatBufferBuilder& builder, const meta_index& x);
```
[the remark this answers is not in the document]
mavam: I'll resolve this comment, as the above discussion covers it content-wise. If changes are needed, I remember to apply them here as well.

libvast/vast/fbs/utils.hpp
tobim: The comments clarify it somewhat, but I still have to question why we can't drop the `file_identifier` argument and hard code `"VAST"` when calling `Finish()`?

mavam: Because not *every* flatbuffer has a file identifier. Only those that are *root* buffers, i.e., those that begin the file.

Given a flatbuffer, you can figure out whether it has a `file_identifier`, but not the other way round. It's a limitation of the API because it was not built to support type-conditional packing in a generic way.

tobim: > Because not every flatbuffer has a file identifier. Only those that are root buffers, i.e., those that **begin** the file. (emphasis mine)

So we can write multiple independent flatbuffers into one file instead of defining a `table` that contains all of those and declaring that the root object?

Or to look at this from a different angle: What is the use case for calling `wrap` on an object that is not going to be the top-level item in a file? Currently there is no such use in the code.

mavam: > So we can write multiple independent flatbuffers into one file instead of defining a table that contains all of those and declaring that the root object?

The file identifier has nothing to do with the scenario when multiple flatbuffers are written to the same file. When specified in a *.fbs file, `flatc` will generate special functions for finishing the flatbuffer such that it starts with the identifier. If we used the these generated functions, like `FinishXXX`, then it would be completely transparent. However, `pack` and `unpack` are generic. They can't call a generated function. They can only dispatch on types. Because [there are no traits for this scenario](https://groups.google.com/d/msg/flatbuffers/DzSpIE-6fRA/Zm544dxvAgAJ), it's our burden to add the identifier manually, based on whether we have a buffer that's to be persisted on disk or whether a buffer that just gets composed into another flatbuffer.

> What is the use case for calling wrap on an object that is not going to be the top-level item in a file? Currently there is no such use in the code.

This will be a table slice inside segments, for example.

mavam: Here is a snippet [from the docs](https://google.github.io/flatbuffers/md__schemas.html):

> Typically, a FlatBuffer binary buffer is not self-describing, i.e. it needs you to know its schema to parse it correctly. But if you want to use a FlatBuffer as a file format, it would be convenient to be able to have a "magic number" in there, like most file formats have, to be able to do a sanity check to see if you're reading the kind of file you're expecting.
> 
> Now, you can always prefix a FlatBuffer with your own file header, but FlatBuffers has a built-in way to add an identifier to a FlatBuffer that takes up minimal space, and keeps the buffer compatible with buffers that don't have such an identifier.
> 
> You can specify in a schema, similar to `root_type`, that you intend for this type of FlatBuffer to be used as a file format:

mavam: FWIW, I've [filed a PR](https://github.com/google/flatbuffers/pull/5904) in the flatbuffers repo to fix this.

tobim: > > What is the use case for calling wrap on an object that is not going to be the top-level item in a file? Currently there is no such use in the code.

> This will be a table slice inside segments, for example.

I see now, we pre-construct `TableSlice`s at the source, but they don't need a file identifier, because the `Segment` provides that already.

PR #858, opened 2020-05-06
Make index-du script work on FreeBSD and macOS

dominiklohmann:
This aliases find to gfind if it is available to explicitly use the find from gnu-findutils.

### DIFF
--- a/scripts/index-du
+++ b/scripts/index-du
@@ -5,6 +5,11 @@
 # Abort on error
 set -e
 
+# Make the script work on FreeBSD and macOS iff gnu-findutils is installed.
+if command -v gfind >/dev/null 2>&1; then
+  alias find=gfind
+fi
+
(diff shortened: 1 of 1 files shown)


dominiklohmann:
This works on macOS, but then we run into the same problems again with different flags for `stat`:

```sh
find vast.db -path "*/index/*" -type f -print0 | xargs -0 stat -f "%N %z" | awk '
{
  # Remove the last name component after "-":
  n = split($1, parts, "-")
  s = parts[1]
  if (n > 1)
    for (i = 2; i < n; i++)
      s = s "-" parts[i]
  sizes[s] += $2
}
END {
  for (field in sizes)
    printf "%s\t%d\n", field, sizes[field]
}'
```

It's usually just better to use the g-prefixed tools in scripts like this.

PR #860, opened 2020-05-08
Add systemd unit for vast

0ortmann:
Add a locked-down `systemd` service for running a VAST server node as system service.


systemd/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Prepare the Host System
+
+A user and group called `vast` are required for running the service. On Linux
+machines you can create one as follows.
+
+```sh
+sudo useradd -u 1337 vast   # create a new user called 'vast' with user ID 1337
+sudo groupadd vast          # create a new group called 'vast'
+sudo usermod -aG vast vast  # add the 'vast' user to the 'vast' group
```
tobim, line 15: I didn't test this, but it should be possible to shorten to:
```suggestion
sudo useradd --system --user-group vast
```

systemd/vast.service
```diff
@@ -0,0 +1,45 @@
...
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+# TODO: move to Type=notify as soon as VAST supports native init systems.
+Type=simple
+
+# user + privileges
+User=vast
+Group=vast
```
tobim, line 12:
```suggestion
# We explicitly don't use DynamicUser because that can lead to recursive `chown`s.
# Doing that is pretty slow on some file systems (e.g., xfs).
User=vast
Group=vast
```

PR #861, opened 2020-05-08
Fix pivot_command.cpp build failure without pcap

dominiklohmann:


libvast/src/system/pivot_command.cpp
```diff
@@ -84,13 +87,18 @@ pivot_command(const command::invocation& invocation, caf::actor_system& sys) {
...
     using defaults_t = defaults::export_::pcap;
     std::string category = defaults_t::category;
     auto output = get_or(options, category + ".write", defaults_t::write);
     auto flush = get_or(options, category + ".flush-interval",
                         defaults_t::flush_interval);
     format::pcap::writer w{output, flush};
     writer = sys.spawn(sink<format::pcap::writer>, std::move(w), limit);
+#else
+    return make_message(make_error(ec::invalid_configuration, "pcap support "
+                                                              "unavailable"));
```
mavam, line 100: It would be great if we can always get the same error message when we don't have compiled a certain optional dependency. For now, can we look whether this message is inline with others?

In a separate (cleanup) PR, it may make sense to factor the error message generation for these cases.

dominiklohmann: I don't think we have another place where we need to put an error message if a feature is unavailable.

PR #862, opened 2020-05-09
Refactor meta index

mavam:
The recent index refactoring flattened the partition layout. However, the old hierarchical structure still existed in the meta index. This commit re-synchronizes the architecture.

### DIFF
--- a/libvast/src/meta_index.cpp
+++ b/libvast/src/meta_index.cpp
@@ -33,7 +33,6 @@ void meta_index::add(const uuid& partition, const table_slice& slice) {
              : factory<synopsis>::make(field.type, synopsis_options_);
   };
   auto& part_syn = synopses_[partition];
-  auto add_to_blacklist = true;
   for (size_t col = 0; col < slice.columns(); ++col) {
     // Locate the relevant synopsis.
     auto& field = slice.layout().fields[col];
@@ -44,7 +43,6 @@ void meta_index::add(const uuid& partition, const table_slice& slice) {
       it = part_syn.emplace(std::move(key), make_synopsis(field)).first;
     // If there exists a synopsis for a field, add the entire column.
     if (auto& syn = it->second) {
-      add_to_blacklist = false;
       for (size_t row = 0; row < slice.rows(); ++row) {
         auto view = slice.at(row, col);
         if (!caf::holds_alternative<caf::none_t>(view))
--- a/libvast/src/meta_index.cpp
+++ b/libvast/src/meta_index.cpp
@@ -27,41 +27,31 @@
 namespace vast {
 
 void meta_index::add(const uuid& partition, const table_slice& slice) {
-  auto& part_synopsis = partition_synopses_[partition];
-  auto& layout = slice.layout();
-  if (blacklisted_layouts_.count(layout) == 1)
-    return;
-  auto i = part_synopsis.find(layout);
-  table_synopsis* table_syn;
-  if (i != part_synopsis.end()) {
-    table_syn = &i->second;
-  } else {
-    // Create new synopses for a layout we haven't seen before.
-    i = part_synopsis.emplace(layout, table_synopsis{}).first;
-    table_syn = &i->second;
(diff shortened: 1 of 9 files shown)


libvast/src/meta_index.cpp
```diff
@@ -27,41 +27,31 @@
...
-  VAST_ASSERT(table_syn->size() == slice.columns());
-  for (size_t col = 0; col < slice.columns(); ++col)
-    if (auto& syn = (*table_syn)[col])
+  auto make_synopsis = [&](const record_field& field) -> synopsis_ptr {
+    return has_skip_attribute(field.type)
+             ? nullptr
+             : factory<synopsis>::make(field.type, synopsis_options_);
+  };
+  auto& part_syn = synopses_[partition];
+  auto add_to_blacklist = true;
```
tobim, line 36: `add_to_blacklist` is only assigned to, but never used. The blacklist itself has been removed from the type, so this variable is a leftover?

libvast/vast/qualified_record_field.hpp
```diff
@@ -24,23 +25,33 @@
...
+  /// Retrieves the full-qualified name, i.e., the record typename concatenated
+  /// with the field name.
+  std::string fqn() const {
+    return layout_name + "." + field_name;
   }
 
-  std::string fqn; ///< The field name prepended with the record type name.
-  vast::type type; ///< The type of the field.
+  std::string layout_name; ///< The name of the layout.
+  std::string field_name;  ///< The name of the field.
```
tobim, line 53: We don't use these name parts individually anywhere, so I don't understand the reasoning behind this change. I would also like to make sure this does not come with a negative impact on performance.

mavam: What is the expensive operation you are concerned about? It doesn't change the hashing performance by much, so the only one I could think of is string concatenation when FQN is needed. My intuition is that this is not at all on the hot path, compared to whats going to happen with all the payload data.

> We don't use these name parts individually anywhere

That's not true, just look at `layout_name` in the diff. The separation of these fields is important so that we can use `layout_name` separately. Otherwise it's impossible to get back to the original `record_field`. We have a lot of code that dispatches by type name, which makes `qualified_record_field` much more flexible.

libvast/src/meta_index.cpp
mavam: A leftover, indeed.

libvast/vast/qualified_record_field.hpp
tobim: I missed that use above. See my other comment.

libvast/src/meta_index.cpp
```diff
@@ -160,13 +144,22 @@ std::vector<uuid> meta_index::lookup(const expression& expr) const {
...
-            for (auto& [part_id, part_syn] : partition_synopses_)
-              for (auto& pair : part_syn)
-                if (evaluate(pair.first.name(), x.op, d)) {
+            for (auto& [part_id, part_syn] : synopses_) {
+              for (auto& pair : part_syn) {
+                // TODO: provide an overload for view of evaluate() so that
+                // we can use string_view here. Fortunately type names are
+                // short, so we're probably not hitting the allocator due to
+                // SSO.
+                auto type_name = data{pair.first.layout_name};
```
tobim, line 156: Piggy-backing on existing synopses is a bit risky. This lookup fails if the matching type does not contain any fields for which we can create a synopsis. Consider
```
type foo = record{
  bar: string
}
```

mavam: I'm not sure what you mean by "piggy-backing on existing synopses". The logic is the same as before: as long as *one* synopsis exists for a given type, we have to consider the partition.

For example:
```
type foo = record{
  bar: string,
  ts: time #timestamp
}
```

The first evaluation fails, the second succeeds. Therefor, the partition should be inspected further.

tobim: I was not aware that the keys are inserted even when `make_synopsis` failed.

PR #863, opened 2020-05-10
Flatbufferize index

mavam:
This PR changes the persistent state of the index to be flatbuffer-based.


libvast/vast/chunk.hpp
```diff
@@ -62,22 +65,32 @@ class chunk : public caf::ref_counted {
...
-    auto data = std::data(*ys);
-    // The deleter won't touch the data.
-    using mutable_data = std::decay_t<decltype(*data)>*;
-    return make(std::size(*ys), const_cast<mutable_data>(data), deleter);
+    return make(ys->size(), ys->data(), deleter);
+  }
+
+  // Construct a chunk by copying a range of bytes.
+  template <typename Byte>
+  static chunk_ptr copy(size_type size, const Byte* data) {
```
mavam, line 87:
```suggestion
  static chunk_ptr copy(span<const byte> bytes) {
```

I think this interface gives more flexibility because there exist a lot of utility functions to convert contiguous things into a `span<const byte>`. For example, we also have `as_bytes(T)` for that.

libvast/vast/chunk.hpp
```diff
@@ -166,4 +179,13 @@ class chunk : public caf::ref_counted {
   deleter_type deleter_;
 };
 
+// We need this template in order to be able to send chunks as messages,
+// but we never actually want to have to use it since it will only be invoked
+// if chunk messages would be sent over the network. And we only want to send
+// chunks around internally. So it is declared here but never defined, leading
+// to a linker error if its used.
```
mavam, line 186: Intriguing idea, I like it.

What's the issue with sending chunks over the network, though? Can't we materialize them as an owning byte buffer on the receiver side, using the default allocator?

libvast/vast/fbs/partition.fbs
```diff
@@ -0,0 +1,34 @@
+namespace vast.fbs;
+
+table ValueIndex {
+  /// The full-qualified field name, e.g., "zeek.conn.id.orig_h".
+  qualified_field_name: string;
```
mavam:
```suggestion
  qualified_field_name: string (key);
```

I don't recall if the value indexes are sorted, but  if they are, this gives *O(log n)* access.

libvast/src/system/indexer.cpp
```diff
@@ -25,12 +28,59 @@
...
+  return {
+    [=](caf::stream<table_slice_column> in) {
+      VAST_DEBUG(self, "got a new table slice stream");
+      return caf::attach_stream_sink(
+        self, in,
+        [=](caf::unit_t&) {
+          self->state.idx = factory<value_index>::make(index_type, index_opts);
+          if (!self->state.idx) {
+            VAST_ERROR(self, "failed to construct index");
+            self->quit(); // TODO: choose right error code.
```
mavam:
```suggestion
            self->quit(ec::unspecified); // TODO: choose right error code.
```

Only marginally better. 🙂

libvast/vast/fbs/partition.fbs
```diff
@@ -18,22 +18,38 @@ table QualifiedValueIndex {
   index: ValueIndex;
 }
 
+table TypeIds {
```
mavam, line 21: Could you provide documentation about this type? What are its semantics?

```suggestion
table EventIDs {
```

A "type ID" to me is a single identifier for a type, but this is an event ID.

libvast/vast/fbs/partition.fbs
```diff
@@ -18,22 +18,38 @@ table QualifiedValueIndex {
...
+}
+
 /// The persistent state of the meta index.
+//  TODO: Add a `Version` field.
 table Partition {
   /// The UUID of this partition.
   uuid: UUID;
 
+  /// A human-readable name for this partition
+  name: string;
```
mavam, line 36: Isn't the partition name simply the UUID?

libvast/src/uuid.cpp
```diff
@@ -109,4 +111,22 @@ bool operator<(const uuid& x, const uuid& y) {
...
+  return uuid_builder.Finish();
+}
+
+caf::error unpack(const fbs::UUID& x, uuid& y) {
+  if (x.data()->size() != uuid::num_bytes)
+    return make_error(ec::format_error, "wrong uuid format");
+  span<const byte, uuid::num_bytes> bytes{
+    reinterpret_cast<const byte*>(x.data()->data()), x.data()->size()};
+  y = uuid{bytes};
+  return ec::no_error;
```
mavam, line 129:
```suggestion
  return caf::none;
```

We usually use that, but I like `ec::no_error` better. Maybe we can stick to one style and then do a global sweep in one refactoring commit?

libvast/src/system/indexer.cpp
```diff
@@ -33,13 +33,21 @@
 
 #include <caf/attach_stream_sink.hpp>
 
+#include "caf/binary_serializer.hpp"
+
+// #include "caf/serializer_impl.hpp"
```
mavam: Can this be removed?

libvast/src/system/indexer.cpp
```diff
@@ -69,14 +78,30 @@ caf::behavior indexer(caf::stateful_actor<indexer_state>* self, type index_type,
           self->quit(err);
         });
     },
+    // FIXME: Get rid of one of these handler, i think the bottom one is
+    // currently unused.
```
mavam: Can we get rid of it? Maybe that happens later in this PR.

libvast/src/system/indexer.cpp
```diff
@@ -69,14 +78,30 @@ caf::behavior indexer(caf::stateful_actor<indexer_state>* self, type index_type,
...
+    [=](const curried_predicate& pred) {
+      VAST_DEBUG(self, "got predicate:", pred);
+      return self->state.idx->lookup(pred.op, make_view(pred.rhs));
+    },
     [=](relational_operator op, const data_view& rhs) {
```
mavam: I just realized that the handler here is a view. This is probably UB, because the underlying partition could already be gone.

```suggestion
    [=](relational_operator op, const data& rhs) {
```

libvast/src/system/indexer.cpp
```diff
@@ -69,14 +78,30 @@ caf::behavior indexer(caf::stateful_actor<indexer_state>* self, type index_type,
...
-    }};
+    [=](atom::snapshot, caf::actor receiver) {
+      std::vector<char> buf;
+      caf::binary_serializer bs{
+        nullptr,
+        buf}; // TODO: do we need to pass the current actor system as first arg?
+      inspect(bs, self->state.idx);
+      auto chunk = chunk::make(std::move(buf));
+      auto sender = self->current_sender();
+      self->send(receiver, atom::done_v, chunk);
```
mavam: Ignoring this hunk, as we figured out the promise logic later.

libvast/vast/system/partition.hpp
```diff
@@ -27,32 +29,148 @@
...
+pack(flatbuffers::FlatBufferBuilder& builder, const partition_state& x);
+
+caf::error unpack(const fbs::Partition& x, readonly_partition_state& y);
+
+// TODO: Use typed actors for the partition actors.
+
 /// Spawns a partition.
 /// @param self The partition actor.
 /// @param id The UUID of this partition.
 caf::behavior partition(caf::stateful_actor<partition_state>* self, uuid id);
```
mavam:
```suggestion
caf::behavior active_partition(caf::stateful_actor<partition_state>* self, uuid id);
```

libvast/vast/system/partition.hpp
```diff
@@ -27,32 +29,148 @@
...
 
+/// Spawns a read-only partition.
+/// TODO: Maybe we should just send the path here, then the actual loading of
+/// the chunk can be done asynchronously
+caf::behavior
+readonly_partition(caf::stateful_actor<readonly_partition_state>* self, uuid id,
```
mavam, line 181:
```suggestion
passive_partition(caf::stateful_actor<readonly_partition_state>* self, uuid id,
```

libvast/vast/system/partition.hpp
```diff
@@ -27,32 +29,148 @@
...
+  std::optional<path> persist_path;
+  size_t persisted_indexers;
+  std::map<caf::actor_id, vast::chunk_ptr> chunks;
 };
 
+// TODO: Split this into a `static data` part that can be mmap'ed
+// straight from disk, and an actor-related part. In the ideal case,
+// we want to eventually be able to use the on-disk state without
+// any intermediate deserialzation step, like yandex::mms or cap'n proto.
+struct readonly_partition_state {
```
mavam, line 122:
```suggestion
struct passive_partition_state {
```

mavam (commented):
I only viewed this until 0db6638dba708129df4f3c2df3316336dd321d58. Looks good so far, but I glimpsed over some implementation details because they will change later on. Is it a good idea to go through this on a commit-by-commit basis? Would a file-based review make more sense?

Most notably, I think we could rename `readonly_partition` to `passive_partition` and rename the standard `partition` to `active_partition`. This the terminology in the architecture and what we use to communicate, so it would be nice to use the same in the code.

(By the way, as PR author I can't approve or request changes. 😒)

tools/lsvast/README.md
```diff
@@ -0,0 +1,7 @@
+# lsvast
```
mavam, line 1: Why not use `vast status` for this?

libvast/vast/system/index.hpp
```diff
@@ -28,12 +31,234 @@
...
 #include <unordered_map>
 #include <vector>
 
+#include "caf/response_promise.hpp"
+
 namespace vast::system {
 
+namespace v2 {
+
+// FIXME: move to vast/fwd.hpp
```
mavam, line 46: ☝️

libvast/vast/system/index.hpp
```diff
@@ -28,12 +31,234 @@
...
+  caf::stream_slot stream_slot;
+
+  /// The remaining free capacity of the partition.
+  size_t capacity;
+
+  /// The UUID of the partition.
+  uuid id;
+
+  template <class Inspector>
+  friend auto inspect(Inspector& f, active_partition_state& x) {
```
mavam, line 64: Is this inspector needed?

libvast/vast/chunk.hpp
lava: We can, but we currently dont need to do that. So I think its better to wait until we have a use case for that and then enable it explicitly.

libvast/vast/fbs/partition.fbs
lava: They're not, the order is determined by the order of the `combined_layout`, which is determined by the order in which we saw the table columns.

libvast/vast/fbs/partition.fbs
lava: Right, removed.

libvast/src/uuid.cpp
lava: Sounds good.

libvast/src/system/index.cpp
```diff
@@ -49,17 +56,683 @@
...
+//                                  curried_predicate
+//                   evaluator  -------------------------------> indexer
+//
+//                                            ids
+//                              <------------------------------
+// clang-format on
+
 namespace vast::system {
 
+namespace v2 {
```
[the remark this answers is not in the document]
lava: I'll append it as a separate commit, so we get one consisting *only* of the removal, otherwise i fear the diff will beecome quite hard to read as the diff algorithm searches for similarities between the v1 and v2 index.

libvast/vast/chunk.hpp
mavam: I agree, let's hold off until we have that use case.

libvast/vast/chunk.hpp
lava: Good point, that also fixes the "wrong" order of size and pointer.

libvast/src/system/index.cpp
```diff
@@ -49,17 +56,683 @@
...
+//
+//                                   [indexer]
+//                  (spawns     <----------------     
+//                   evaluator) 
+//
+//                                  curried_predicate
+//                   evaluator  -------------------------------> indexer
+//
+//                                            ids
+//                              <------------------------------
```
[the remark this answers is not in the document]
lava: 1. Are describing the status quo or suggesting a refactoring?
2. Makes sense
3. I just took a look, but removing `curried_predicate` is a somewhat bigger change (requires changing the use of  evaluation_triples as well) that is also functionally a no-op, so I'd prefer to do it as a separate refactoring.

libvast/vast/system/partition.hpp
```diff
@@ -27,8 +32,157 @@
...
+  /// @param x The literal side of the predicate.
+  caf::actor fetch_indexer(const data_extractor& dx, relational_operator op,
+                           const data& x) const;
+
+  /// Retrieves an INDEXER for a predicate with an attribute extractor.
+  /// @param ex The extractor.
+  /// @param op The operator (only used to precompute ids for type queries.
+  /// @param x The literal side of the predicate.
+  caf::actor fetch_indexer(const attribute_extractor& ex,
+                           relational_operator op, const data& x) const;
```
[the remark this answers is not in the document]
lava: This is already the fan-out from a single function that takes a predicate.

libvast/vast/system/partition.hpp
```diff
@@ -27,8 +32,157 @@
...
+  partition_stream_stage_ptr stage;
+
+  /// Tracks whether we already received at least one table slice.
+  bool streaming_initiated;
+
+  /// The combined type of all columns of this partition
+  record_type combined_layout;
+
+  /// Maps qualified fields to indexer actors.
+  //  TODO: Should we use the tsl map here for heterogenous key lookup?
```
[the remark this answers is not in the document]
lava: I don't see how, the tsl map is a hash map and the stable map is not.

But we could probably add heterogenous key lookup into the stable map if we need it.

libvast/src/system/indexer.cpp
lava: It does, I've squashed all related commits so the history shouldnt contain any of these anymore.

libvast/vast/system/partition.hpp
```diff
@@ -27,8 +32,157 @@
...
+
+  /// Persistence-related state
+  filesystem_type fs_actor;
+  caf::response_promise persistence_promise;
+  std::optional<path> persist_path;
+  size_t persisted_indexers;
+  std::map<caf::actor_id, vast::chunk_ptr> chunks;
+};
+
+// TODO: Split this into a `static data` part that can be mmap'ed
```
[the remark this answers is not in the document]
lava: Not a huge fan of this idea tbh, for two reasons. First, semantically an `active_partition` is not an extension of a `passive_partition`, its more like a `passive_partition_builder`. (and you wouldn't let a string_builder inherit from string, etc.)

And second, the fact that they currently contain the same data is just because we didn\t get to optimizing that part yet, ideally they would have completely different memory layouts, one that is optimized for adding new data and one that is takes advantage of knowing *all* the data going into the partition and can be accessed  without deserializing.

libvast/src/system/partition.cpp
```diff
@@ -14,41 +14,604 @@
...
+  caf::binary_deserializer bds(
+    nullptr, reinterpret_cast<const char*>(combined_layout->data()),
+    combined_layout->size());
+  error = bds(state.combined_layout);
+  if (error)
+    return error;
+  // TODO: This condition should be '!=', but then we cant deserialize
+  // in unit tests anymore without creating a bunch of index actors first. :/
+  if (state.combined_layout.fields.size() < indexes->size()) {
+    VAST_INFO(state.self, state.combined_layout.fields.size(), "fields vs",indexes->size(),"indexes");
```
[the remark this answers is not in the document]
lava: Good question, the version in my working tree already has correct indentation so probably I accidentally moved the reformatting into the wrong commit.

libvast/vast/fbs/index.fbs
```diff
@@ -0,0 +1,33 @@
...
+/// The persistent state of the index.
+table Index {
+  /// The version of the index state.
+  version: Version;
+
+  /// Synopsis structures for the contained partitions.
+  meta_index: MetaIndex;
+
+  /// The contained partition UUIDs.
+  partitions: [UUID];
```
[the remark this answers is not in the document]
lava: We *could*, but it seems useful to have a sanity check of the state we expect to be present on-disk. Otherwise we would have no warning if partitions are present on disk but not in the meta index, or the other way round if partitions are suddenly missing after a restart.

libvast/vast/system/partition.hpp
```diff
@@ -27,8 +32,157 @@
...
+
+  /// The first ID in the partition.
+  size_t offset;
+
+  /// The number of events in the partition.
+  size_t events;
+
+  // Stores the deserialized indexers. This is only used when unpacking the
+  // flatbuffer, afterwards the indexer state is moved into the indexer actor.
+  std::vector<std::pair<qualified_record_field, value_index_ptr>> indexer_states;
```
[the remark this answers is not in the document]
lava: I've changed it (not at least because it brings the line length to <80 characters ;) ), but I'm curious what you think the intent is?

libvast/vast/system/partition.hpp
```diff
@@ -27,8 +32,157 @@
...
+  /// A readable name for this partition
+  std::string name;
+
+  /// The first ID in the partition.
+  id offset;
+
+  /// The number of events in the partition.
+  size_t events;
+
+  /// Persistence-related state
```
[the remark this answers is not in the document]
lava: Not sure, the members below are probably less volatile than some of the other state (e.g. event count).

But this subheading was removed anyways in a subsequent commit.

libvast/vast/system/partition.hpp
```diff
@@ -27,8 +32,157 @@
 
 #include <unordered_map>
 
+#include "caf/fwd.hpp"
+
 namespace vast::system {
 
+// TODO: remove this temporary namespace after we have wired all functionality
+// with the new actorized partition.
+namespace v2 {
```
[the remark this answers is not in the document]
lava: Done!

libvast/vast/system/index.hpp
```diff
@@ -28,12 +31,234 @@
...
+  }
+};
+
+/// Accumulates statistics about indexed data.
+struct index_statistics {
+  /// The number of events for a given layout.
+  std::unordered_map<std::string, layout_statistics> layouts;
+
+  template <class Inspector>
+  friend auto inspect(Inspector& f, index_statistics& x) {
```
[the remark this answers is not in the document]
lava: Forwarding the question to @dominiklohmann

libvast/vast/system/indexer.hpp
```diff
@@ -29,6 +31,40 @@
 
 namespace vast::system {
 
+namespace v2 {
+
+// TODO: Create a separate `readonly_indexer_state`, similar to how partitions
+// are handled.
+struct indexer_state {
```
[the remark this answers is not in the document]
lava: Imho the rename only makes sense after the `TODO` above is addressed.

libvast/vast/system/index.hpp
```diff
@@ -28,12 +31,234 @@
...
+
+  // -- constructor ------------------------------------------------------------
+
+  explicit index_state(caf::stateful_actor<index_state>* self);
+
+  // -- persistence ------------------------------------------------------------
+
+  caf::error load_from_disk();
+
+  // void init(const vast::fbs::Index& index);
```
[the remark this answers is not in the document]
lava: Removed

tools/lsvast/README.md
mavam: Okay, makes sense.

libvast/src/system/index.cpp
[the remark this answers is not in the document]
mavam: I think all of these are somewhat "refactory" changes. I'm not sure what you were planning to do within this PR, but "v2" to me kind of suggests the strong split between partitions and index, which seems currently to be done half way with the evaluator remaining outside its target scope.

I'm not sure where you want to draw the line of "in-situ" and "follow-up", but to me (1) feels like it should go in whereas (3) seems to be cosmetic mostly and could therefore follow later.

libvast/vast/system/index.hpp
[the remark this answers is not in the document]
mavam: Turns out this is not, since the functionality of resuming a state-based actor in CAF is currently not supported.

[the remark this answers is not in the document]
mavam: So unless you work with CAF serialization, hashing, or other things we use the inspector API for, you can remove this code.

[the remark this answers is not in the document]
dominiklohmann: This was used for debugging the actor state by making it nicely printable. There's no downside to having these.

libvast/vast/system/partition.hpp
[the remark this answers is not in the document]
mavam: Good arguments, I'm convinced.

libvast/vast/fbs/index.fbs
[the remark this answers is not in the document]
mavam: I see that, makes sense.

libvast/vast/system/partition.hpp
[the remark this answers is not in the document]
mavam: It looks like the record field *maps* to the value index, hence my suggestion to let the type express that better.

libvast/vast/system/index.hpp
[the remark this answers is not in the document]
mavam: Fine by me.

libvast/vast/fbs/version.fbs
```diff
@@ -2,7 +2,12 @@ namespace vast.fbs;
...
+/// the version should get bumped. All version are put into the same enum so
+/// we can deduce the type of a flatbuffer from its version field, acting as
+/// an additional sanity check.
 enum Version : short {
-  v0,
+  Invalid,               // Default value
+  Segment_v0,
+  Index_v0,
+  Partition_v0,
 }
```
lava, line 13: Note to self: We discussed that we instead want to have two fields, `FlatbufferType` and `Version`.

libvast/vast/system/partition.hpp
```diff
@@ -27,8 +32,173 @@
...
+// TODO: Use typed actors for the partition actors.
+
+/// Spawns a partition.
+/// @param self The partition actor.
+/// @param id The UUID of this partition.
+caf::behavior
+active_partition(caf::stateful_actor<active_partition_state>* self, uuid id,
+                 filesystem_type fs, caf::settings index_opts);
+
+/// Spawns a read-only partition.
```
mavam, line 191: Mind documenting the parameters as well? As a user, for example, I'd be curious to know whether `path` needs to be absolute or relative to the index root.

libvast/src/system/index.cpp
```diff
@@ -41,24 +47,692 @@
...
+    return meta_idx.error();
+  VAST_VERBOSE(state.self, "persisting", state.persisted_partitions.size(),
+               " definitely persisted and ", state.unpersisted.size(),
+               " maybe persisted partitions uuids");
+  std::vector<flatbuffers::Offset<fbs::UUID>> partition_offsets;
+  for (auto uuid : state.persisted_partitions) {
+    auto uuid_fb = pack(builder, uuid);
+    if (!uuid_fb)
+      return uuid_fb.error();
+    partition_offsets.push_back(*uuid_fb);
```
mavam, line 347:
```suggestion
    if (auto uuid_fb = pack(builder, uuid));
        partition_offsets.push_back(*uuid_fb);
    else
      return uuid_fb.error();
```

I find this `expected<T>` pattern more idiomatic, as it brings out the if-else character better. But it's a matter of preference.

libvast/src/system/index.cpp
```diff
@@ -41,24 +47,692 @@
...
+  }
+  // We don't know if these will make it to disk before the index and the rest
+  // of the system is shut down (in case of a hard/dirty shutdown), so we just
+  // store everything and throw out the missing partitions when loading the
+  // index.
+  for (auto& kv : state.unpersisted) {
+    auto uuid_fb = pack(builder, kv.first);
+    if (!uuid_fb)
+      return uuid_fb.error();
+    partition_offsets.push_back(*uuid_fb);
```
mavam, line 357: My above note applies to this loop body as well.

libvast/src/system/index.cpp
```diff
@@ -41,24 +47,692 @@
...
+  auto builder = new flatbuffers::FlatBufferBuilder();
+  auto index = pack(*builder, *this);
+  if (!index) {
+    VAST_WARNING(self, "couldnt pack index", index.error());
+    return;
+  }
+  builder->Finish(*index, vast::fbs::file_identifier);
+  auto ptr = builder->GetBufferPointer();
+  auto size = builder->GetSize();
+  auto chunk = vast::chunk::make(size, ptr, [=] { delete builder; });
```
mavam, line 389: A builder can actually relinquish it's contents: see `fbs::release` in `fbs/utils.hpp`. Then you can avoid the scary naked `new`.

libvast/src/system/index.cpp
```diff
@@ -41,24 +47,692 @@
...
+  self
+    ->request(caf::actor_cast<caf::actor>(filesystem), caf::infinite,
+              atom::write_v, index_filename(), chunk)
+    .then([=](atom::ok) { VAST_DEBUG(self, "finished writing to disk"); },
+          [=](caf::error err) {
+            VAST_WARNING(self, "encountered issue writing meta ids", err);
```
mavam, line 395:
```suggestion
            VAST_WARNING(self, "failed to persist index state", err);
```

libvast/src/system/index.cpp
```diff
@@ -41,24 +47,692 @@
...
+    return;
+  }
+  builder->Finish(*index, vast::fbs::file_identifier);
+  auto ptr = builder->GetBufferPointer();
+  auto size = builder->GetSize();
+  auto chunk = vast::chunk::make(size, ptr, [=] { delete builder; });
+  self
+    ->request(caf::actor_cast<caf::actor>(filesystem), caf::infinite,
+              atom::write_v, index_filename(), chunk)
+    .then([=](atom::ok) { VAST_DEBUG(self, "finished writing to disk"); },
```
mavam, line 393:
```suggestion
    .then([=](atom::ok) { VAST_DEBUG(self, "successfully persisted index state"); },
```

I don't like to use the word "disk" because it assumes a specific hardware setting. Let's go with "persisting" instead.

libvast/src/system/index.cpp
```diff
@@ -38,172 +44,145 @@
...
+                        "unsupported index version, either remove the existing "
+                        "vast.db "
+                        "directory or try again with a newer version of VAST");
+    auto meta_idx = index->meta_index();
+    VAST_ASSERT(meta_idx);
+    auto error = unpack(*meta_idx, this->meta_idx);
+    if (error)
+      return error;
+    auto partition_uuids = index->partitions();
+    VAST_ASSERT(partition_uuids);
```
mavam, line 157: is this also valid with 0 partitions?

libvast/src/system/index.cpp
```diff
@@ -38,172 +44,145 @@
...
+      VAST_ASSERT(uuid_fb);
+      vast::uuid partition_uuid;
+      unpack(*uuid_fb, partition_uuid);
+      if (exists(dir / to_string(partition_uuid)))
+        persisted_partitions.insert(partition_uuid);
+      else
+        // TODO: remove the problematic uuid from the meta index if we get here
+        VAST_WARNING(self, "found partition", partition_uuid,
+                     "in the index state but not on disk. This may have been "
+                     "caused b");
```
mavam, line 168: The error message is not quite complete yet.

libvast/src/system/index.cpp
```diff
@@ -38,172 +44,145 @@
...
+    }
+    auto stats = index->stats();
+    if (!stats)
+      return make_error(ec::format_error, "No stats in index flatbuffer");
```
mavam, line 172:
```suggestion
      return make_error(ec::format_error, "no stats in index flatbuffer");
```

libvast/src/system/index.cpp
```diff
@@ -214,212 +193,346 @@ caf::actor index_state::next_worker() {
...
+    if (active_partition.actor != nullptr
+        && active_partition.id == partition_id)
+      part = active_partition.actor;
+    else if (auto it = unpersisted.find(partition_id); it != unpersisted.end())
+      part = it->second;
+    else if (auto it = persisted_partitions.find(partition_id);
+             it != persisted_partitions.end())
+      part = inmem_partitions.get_or_load(partition_id);
+    if (!part)
+      VAST_ERROR("Could not load partition", partition_id,
```
mavam, line 270:
```suggestion
      VAST_ERROR(self, "could not load partition", partition_id,
```

libvast/src/system/index.cpp
```diff
@@ -214,212 +193,346 @@ caf::actor index_state::next_worker() {
...
+// TODO: At some point we should add some more template magic on top of
+// this and turn it into a generic functions that maps
+//
+//   (map from U to A, request param pack R, result handler with param X) ->
+//   expected<map from U to X>
+template <typename Continuation>
+void await_evaluators_then(
+  caf::stateful_actor<index_state>* self, const expression& expr,
+  const std::vector<std::pair<vast::uuid, caf::actor>>& actors,
+  Continuation c) {
```
mavam, line 304:
```suggestion
  Continuation continuation) {
```

Sticks out more that way down below (since `c` is not an idiomatic function name).

libvast/src/system/index.cpp
```diff
@@ -214,212 +193,346 @@ caf::actor index_state::next_worker() {
...
+  struct counter {
+    size_t received;
+    query_map qm;
+  };
+  auto expected = actors.size();
+  auto shared_counter = std::make_shared<counter>();
+  for (auto& [id, actor] : actors) {
+    auto partition_id = id; // Can't use structured binding inside lambda.
```
mavam, line 312:
```suggestion
    auto& partition_id = id; // Can't use structured binding inside lambda.
```

libvast/src/system/index.cpp
```diff
@@ -214,212 +193,346 @@ caf::actor index_state::next_worker() {
...
   }
-  return result;
 }
 
-void index_state::add_flush_listener(caf::actor listener) {
-  VAST_DEBUG(self, "adds a new 'flush' subscriber:", listener);
-  flush_listeners.emplace_back(std::move(listener));
-  detail::notify_listeners_if_clean(*this, *stage);
+path index_state::index_filename(path basename) const {
+  return basename / dir / "index.bin";
```
mavam, line 331: We probably don't need `basename` if we do all I/O through the filesystem actor.

libvast/src/system/index.cpp
```diff
@@ -428,143 +541,182 @@ caf::behavior index(caf::stateful_actor<index_state>* self, const path& dir,
...
+      else
+        VAST_VERBOSE(self, "successfully persisted partition");
+    },
+    // Query handling
+    [=](vast::expression expr) {
+      auto& st = self->state;
+      auto mid = self->current_message_id();
+      auto sender = self->current_sender();
+      auto client = caf::actor_cast<caf::actor>(sender);
+      auto respond = [=](auto&&... xs) {
```
mavam, line 562: Do we still need this hack?

libvast/src/system/index.cpp
```diff
@@ -428,143 +541,182 @@ caf::behavior index(caf::stateful_actor<index_state>* self, const path& dir,
...
+      return self->state.stage->add_inbound_path(in);
+    },
+    [=](accountant_type accountant) {
+      namespace defs = defaults::system;
+      self->state.accountant = std::move(accountant);
+    },
+    [=](atom::status) -> caf::config_value::dictionary {
+      return self->state.status();
+    },
+    [=](atom::subscribe, atom::flush, [[maybe_unused]] caf::actor& listener) {
```
mavam, line 716:
```suggestion
    [=](atom::subscribe, atom::flush, caf::actor& listener) {
```

libvast/src/system/index.cpp
```diff
@@ -428,143 +541,182 @@ caf::behavior index(caf::stateful_actor<index_state>* self, const path& dir,
...
-          }};
+  return {
+    // The default behaviour
+    [=](atom::worker, caf::actor& worker) {
+      auto& st = self->state;
+      st.idle_workers.emplace_back(std::move(worker));
+      self->become(caf::keep_behavior, st.has_worker);
+    },
+    [=](atom::done, uuid partition_id) {
+      VAST_VERBOSE(self, "query for partition", partition_id, "is done");
```
mavam, line 703:
```suggestion
      VAST_VERBOSE(self, "received DONE for partition", partition_id);
```

libvast/src/system/index.cpp
```diff
@@ -428,143 +541,182 @@ caf::behavior index(caf::stateful_actor<index_state>* self, const path& dir,
...
+    },
+    [=](caf::stream<table_slice_ptr> in) {
+      VAST_DEBUG(self, "got a new source");
+      return self->state.stage->add_inbound_path(in);
+    },
+    [=](accountant_type accountant) {
+      namespace defs = defaults::system;
```
mavam, line 710:
```suggestion
```

libvast/src/system/index.cpp
```diff
@@ -428,143 +541,182 @@ caf::behavior index(caf::stateful_actor<index_state>* self, const path& dir,
...
       VAST_DEBUG(self, "got a new source");
       return self->state.stage->add_inbound_path(in);
     },
-    [=](atom::status, status_verbosity v) -> caf::config_value::dictionary {
-      return self->state.status(v);
+    [=](atom::status) -> caf::config_value::dictionary {
+      return self->state.status();
     },
-    [=](atom::subscribe, atom::flush, caf::actor& listener) {
+    [=](atom::subscribe, atom::flush, [[maybe_unused]] caf::actor& listener) {
```
mavam, line 692:
```suggestion
    [=](atom::subscribe, atom::flush, caf::actor& listener) {
```

libvast/src/system/index.cpp
```diff
@@ -428,143 +541,182 @@ caf::behavior index(caf::stateful_actor<index_state>* self, const path& dir,
...
+              self->delegate(caf::actor_cast<caf::actor>(self), query_id,
+                             num_partitions);
+              return;
+            }
+            st.pending.erase(iter);
+            VAST_DEBUG(self, "returns without result: no partitions qualify");
+            self->send(client, atom::done_v);
+            return;
+          }
+          // auto qm = st.launch_evaluators(pqm, iter->second.expression);
```
mavam, line 665: Why is this commented?

mavam (commented):
There seems to be an inconsistency in the logging regarding capitalization. We usually don't write capitalized log messages but I encountered multiple logging macros that started that way. I'd say we do everything lowercase.

libvast/src/system/index.cpp
lava: I saw that part, but I was stumped on how to get the correct deleter for the released buffer?

libvast/src/system/index.cpp
lava: I'm also wondering why this is `actor&` instead of just `actor`.

libvast/src/system/index.cpp
lava: As far as I understand, the purpose of this is to "respond" to the original sender without returning from the function. There's probably a better way to do it, but I couldnt immediately see how.

libvast/src/system/index.cpp
lava: Called it `then` instead, which feels more idiomatic.

libvast/src/system/index.cpp
lava: I hope nobody wants to persist indexer state on blockchainfs anytime soon ;) Changed it anyways.

libvast/src/system/index.cpp
lava: Yes. It was remarkably hard to figure this out from documentation, but I modified one of the unit tests to produce an empty vector and the pointer was still valid.

libvast/src/system/index.cpp
mavam: It's hard to say at the call site. However, since we have the global view and know that we only use the default allocator, it's safe to assume that this works. Maybe we could add assertion that it's indeed the default allocator?

lava: Hm, to write such an assertion I'd have to somehow get the allocator ;)

libvast/src/system/index.cpp
dominiklohmann: Not sure. I added the basename here without thinking about it too much and removed the TODO. Don't know why that TODO existed in the first place.

CHANGELOG.md
```diff
@@ -11,6 +11,18 @@ Every entry has a category for which we use the following visual abbreviations:
...
+  contents of a VAST database directory. See `lsvast --help` for usage
+  instructions.
+  [#863](https://github.com/tenzir/vast/pull/863)
+
+- ⚠️ The persistent storage format of the index now uses flatbuffers.
+  [#863](https://github.com/tenzir/vast/pull/863)
+
+- ⚠️ We refactored the index architecture to improve stability and responsiveness.
+  This includes fixes for several shutdown issues.
+  [#863](https://github.com/tenzir/vast/pull/863)
```
dominiklohmann, line 24:
```suggestion
- 🎁 VAST now ships with a new tool `lsvast` to display information about the
  contents of a VAST database directory. See `lsvast --help` for usage
  instructions. [#863](https://github.com/tenzir/vast/pull/863)

 - ⚠️ The persistent storage format of the index now uses FlatBuffers.
   [#863](https://github.com/tenzir/vast/pull/863)

 - ⚠️ We refactored the index architecture to improve stability and
   responsiveness. This includes fixes for several shutdown issues.
   [#863](https://github.com/tenzir/vast/pull/863)
```

dominiklohmann (approved):
Approving as is as we discussed earlier.

Please apply the suggested re-formatting and spelling fixes to the changelog before merging.

PR #865, opened 2020-05-11
Support \r\n-style newlines in VAST

lava:
Support both Unix-, Windows- and Mac-style line endings
for all input formats, either uniformly or mixed.

NOTE: I'd actually like to run benchmarks for this change before merging, because I'm not sure if the libc-readline has some "magic sauce" that makes it far more performant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🎁 All input parsers now support mixed `\n` and `\r\n` line endings.
+  [#865](https://github.com/tenzir/vast/pull/847)
+
 - 🐞 Fixed a bug that caused `vast import` processes to produce `'default'`
   table slices, despite having the `'arrow'` type as the default.
   [#866](https://github.com/tenzir/vast/pull/866)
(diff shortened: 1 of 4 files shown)


lava:
@mavam Where would the input to the `line_parser` come from in this scenario? It looks like the parser would still have to pull out the line byte-by-byte from the `ifstream`, so i'm not sure where the speedup could come from.

mavam:
> Where would the input to the `line_parser` come from in this scenario?

You'd always work on a contiguous block of memory. Better cache utilization. Also no virtual function calls.

The new API will forward a `chunk` that contains a block of lines to the format parser. Then  we can put the `\r\n` logic simply at the end of an existing parser, put a kleene star around it, and have an efficient parser line-based parser.

mavam:
Switching to this API will still take a bit. @dominiklohmann and I have talked about it extensively and we're looking forward to loop you in on this.

dominiklohmann:
> Switching to this API will still take a bit. @dominiklohmann and I have talked about it extensively and we're looking forward to loop you in on this.

JFYI: Said looping-in has taken place this morning. 

On this PR, I think this is a good way to fix for our current API. I'll let @mavam have the last word on this, but all I'm missing here is a changelog entry.

mavam:
Yep, I think we should mention that we now support `\r\n` and merge.

(Modulo failing checks.)

PR #866, opened 2020-05-11
Various config and default setting fixes

tobim:


libvast/src/command.cpp
```diff
@@ -353,10 +353,11 @@ parse(const command& root, command::argument_iterator first,
...
 bool init_config(caf::actor_system_config& cfg, const command::invocation& from,
                  std::ostream& error_output) {
-  // Adjust default file verbosity for commands starting a node,
+  // Disable file logging for all commands that don't start a node,
   // i.e., `vast start` and `vast -N ...`.
-  if (from.name() == "start" || caf::get_or(from.options, "system.node", false))
-    cfg.set("logger.file-verbosity", defaults::logger::server_file_verbosity);
+  if (!(from.name() == "start"
+        || caf::get_or(from.options, "system.node", false)))
+    cfg.set("logger.file-verbosity", caf::atom("quiet"));
```
lava, line 360: Did you verify that this doesn't create log files for non-server commands? I faintly remember that the log level should be 'quiet' *before* caf is initialized, otherwise I did end up with a log file containing a single line printed during vast startup. But maybe this has already changed in caf in the meantime.

vast.conf
```diff
@@ -306,19 +306,22 @@ logger {
 
   ; Configures the minimum severity of messages written to the log file.
   ; Possible values: quiet, error, warning, info, verbose, debug, trace.
-  ;file-verbosity = "debug"
+  ; File logging is only available for commands that start a node (e.g.,
+  ; vast start). The levels above 'verbose' are usually not available in
```
lava, line 310: Not sure, but intuitively I'd say `s/above/below/`?

tobim: info -> low verbosity; debug -> high verbosity?

libvast/src/command.cpp
tobim: This code runs before CAF is initialized. I just checked to make sure it behaves at expected.

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 Fixed a bug that caused `vast import` processes to produce `'caf'` table
```
dominiklohmann, line 13:
```suggestion
- 🐞 Fixed a bug that caused `vast import` processes to produce `'default'` table
```

We haven't renamed that (yet).

vast.conf
lava: Sure, but info -> higher importance, debug -> lower importance.

But I'll let you decide, this was only a side comment.

tobim:
Sorry for the late addition, please check if the last commit is to your liking.

CHANGELOG.md
tobim: Yeah, I changed it, even though the sentence is pretty confusing this way. Let's not get hung up on this though.

tobim:
I just pushed once more to fix the broken formatting in `make_command_factory()`, no need to re-review the commit.

PR #873, opened 2020-05-14
Add initial version 'vast explore' command

lava:


libvast/src/system/explore_command.cpp
```diff
@@ -0,0 +1,160 @@
...
+  self->send(*explorer, explorer_initial_query_atom::value, *exporter);
+  // Set the explorer as sink for the initial query exporter.
+  self->send(*exporter, system::sink_atom::value, *explorer);
+  // (Ab)use query_statistics as done message.
+  self->send(*exporter, system::statistics_atom::value, *explorer);
+  self->send(*explorer, system::sink_atom::value, writer);
+  self->send(*exporter, system::run_atom::value);
+  caf::error err;
+  auto stop = false;
+  self
```
lava, line 127: It feels like there should be an opportunity here to deduplicate some of the code that is extremely similar to `pivot_command.cpp` and lift it into common abstractions, but nothing immediately came to mind.

libvast/src/system/application.cpp
```diff
@@ -273,6 +285,11 @@ auto make_spawn_command() {
     opts()
       .add<size_t>("segments,s", "number of cached segments")
       .add<size_t>("max-segment-size,m", "maximum segment size in MB"));
+  spawn->add_subcommand(
+    "explorer", "creates a new explorer", "",
+    opts()
+      .add<vast::duration>("before,B", "timebox before each result")
+      .add<vast::duration>("after,A", "timebox after each result"));
```
dominiklohmann, line 292:
```suggestion
      .add<vast::duration>("after,A", "timebox after each result")
      .add<vast::duration>("before,B", "timebox before each result"));
```

libvast/src/system/count_command.cpp
```diff
@@ -41,7 +41,7 @@ namespace vast::system {
 
 caf::message
 count_command(const command::invocation& invocation, caf::actor_system& sys) {
-  VAST_TRACE(invocation);
+  VAST_DEBUG_ANON(invocation);
```
dominiklohmann, line 44:
```suggestion
  VAST_TRACE(invocation);
```

Looks like a leftover debug change.

libvast/src/system/application.cpp
dominiklohmann: Also, do these even need the short options? `vast spawn` is hidden after all.

libvast/src/system/explore_command.cpp
```diff
@@ -0,0 +1,160 @@
...
+caf::message
+explore_command(const command::invocation& invocation, caf::actor_system& sys) {
+  VAST_DEBUG_ANON(invocation);
+  const auto& options = invocation.options;
+  // Get query.
+  if (invocation.arguments.size() != 1) {
+    return caf::make_message(make_error(ec::invalid_query,
+                                        "expected query as first and only "
+                                        "positional argument"));
+  }
```
dominiklohmann, line 57: This is different from all other commands that join the positional arguments with spaces into one query, isn't it?

libvast/src/system/explore_command.cpp
```diff
@@ -0,0 +1,160 @@
...
+  auto query = invocation.arguments.at(0);
+  auto before = caf::get_or(options, "explore.before", vast::duration{0s});
+  auto after = caf::get_or(options, "explore.after", vast::duration{0s});
+  // TODO: Add some way for users to pass "infinity" for either value.
```
dominiklohmann, line 61: Side note: Should be easy to add this as a language feature in the timespan parser and have it return `caf::infinite`.

libvast/src/system/explore_command.cpp
```diff
@@ -0,0 +1,160 @@
...
+  auto out = detail::make_output_stream(
+    std::string{format::json::writer::defaults::write}, false);
+  if (!out)
+    return make_message(out.error());
+  caf::actor writer
+    = sys.spawn(sink<format::json::writer>,
+                format::json::writer{std::move(*out)}, max_events);
+  if (!writer) {
+    return make_message(ec::unspecified, "could not spawn writer");
+  }
```
dominiklohmann, line 79:
```suggestion
  if (!writer)
    return make_message(ec::unspecified, "could not spawn writer");
```

libvast/src/system/explore_command.cpp
```diff
@@ -0,0 +1,160 @@
...
+  auto writer_guard = caf::detail::make_scope_guard(
+    [&] { self->send_exit(writer, caf::exit_reason::user_shutdown); });
```
dominiklohmann, line 81: Can you add a debug or trace message in the scope guard?

libvast/src/system/explore_command.cpp
```diff
@@ -0,0 +1,160 @@
...
+    sig_mon_thread, sys, defaults::system::signal_monitoring_interval, self);
+  // Spawn exporter for the passed query
+  auto spawn_exporter
+    = command::invocation{invocation.options, "spawn exporter", {query}};
+  VAST_DEBUG(&invocation, "spawns exporter with parameters:", spawn_exporter);
+  auto exporter = spawn_at_node(self, node, spawn_exporter);
+  if (!exporter)
+    return caf::make_message(std::move(exporter.error()));
+  auto exp_guard = caf::detail::make_scope_guard(
+    [&] { self->send_exit(*exporter, caf::exit_reason::user_shutdown); });
```
dominiklohmann, line 104: Can you add a debug or trace message in the scope guard?

libvast/src/system/explore_command.cpp
```diff
@@ -0,0 +1,160 @@
...
+  auto& explore_dict = explorer_options["explore"].as_dictionary();
+  explore_dict["before"] = before;
+  explore_dict["after"] = after;
+  auto spawn_explorer = command::invocation{explorer_options, "spawn explorer"};
+  VAST_DEBUG(&invocation, "spawns explorer with parameters:", spawn_explorer);
+  auto explorer = spawn_at_node(self, node, spawn_explorer);
+  if (!explorer)
+    return caf::make_message(std::move(explorer.error()));
+  auto explorer_guard = caf::detail::make_scope_guard(
+    [&] { self->send_exit(*explorer, caf::exit_reason::user_shutdown); });
```
dominiklohmann, line 116: Can you add a debug or trace message in the scope guard?

libvast/src/system/explore_command.cpp
dominiklohmann: Also, this one may as well be renamed to `exporter_guard` to be explicit. I had to stop and think for a bit what `exp_guard` is.

libvast/vast/system/explorer.hpp
```diff
@@ -0,0 +1,83 @@
...
+#include <caf/actor.hpp>
+#include <caf/fwd.hpp>
+
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+
+namespace vast::system {
+
+using explorer_initial_query_atom = caf::atom_constant<caf::atom("expl_initq")>;
```
dominiklohmann, line 30: Please move this to `vast/system/atoms.hpp` and maybe even think about reusing one of the existing atoms instead.

libvast/src/system/node.cpp
```diff
@@ -322,7 +325,8 @@ node_state::spawn_command(const command::invocation& invocation,
     label = *label_ptr;
   } else {
     label = comp_name;
-    const char* multi_instance[] = {"importer", "exporter", "source", "sink"};
+    const char* multi_instance[]
+      = {"importer", "explorer", "exporter", "source", "sink"};
```
dominiklohmann, line 329: Why does the explorer have to be in here when the pivoter is not?

libvast/src/system/explorer.cpp
```diff
@@ -0,0 +1,160 @@
...
+  st.after = after;
+  auto quit_if_done = [=]() {
+    auto& st = self->state;
+    if (st.initial_query_completed && st.running_exporters == 0)
+      self->quit();
+  };
+  self->set_down_handler([=]([[maybe_unused]] const caf::down_msg& msg) {
+    // Only the spawned EXPORTERs are expected to send down messages.
+    auto& st = self->state;
+    st.running_exporters--;
```
dominiklohmann, line 80:
```suggestion
    --st.running_exporters;
```

libvast/vast/system/explorer.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+  explorer_state(caf::event_based_actor* self);
+
+  // Send the results to the sink, after removing duplicates.
+  void forward_results(vast::table_slice_ptr slice);
+
+  // Size of the time box prior to each event.
+  std::optional<vast::duration> before;
+
+  // Size of the time box after each event.
```
dominiklohmann, line 43:
```suggestion
  /// Size of the time box after each event.
```

Naming question: Don't we usually just call this a timespan?

libvast/vast/system/explorer.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+struct explorer_state {
+  static inline constexpr const char* name = "explorer";
+
+  explorer_state(caf::event_based_actor* self);
+
+  // Send the results to the sink, after removing duplicates.
+  void forward_results(vast::table_slice_ptr slice);
+
+  // Size of the time box prior to each event.
```
dominiklohmann, line 40:
```suggestion
  /// Size of the time box prior to each event.
```

libvast/vast/system/explorer.hpp
```diff
@@ -0,0 +1,83 @@
...
+namespace vast::system {
+
+using explorer_initial_query_atom = caf::atom_constant<caf::atom("expl_initq")>;
+
+struct explorer_state {
+  static inline constexpr const char* name = "explorer";
+
+  explorer_state(caf::event_based_actor* self);
+
+  // Send the results to the sink, after removing duplicates.
```
dominiklohmann, line 37:
```suggestion
  /// Send the results to the sink, after removing duplicates.
```

libvast/src/system/spawn_pivoter.cpp
```diff
@@ -21,8 +21,10 @@
 namespace vast::system {
 
 maybe_actor spawn_pivoter(node_actor* self, spawn_arguments& args) {
-  VAST_TRACE(VAST_ARG(args));
+  VAST_DEBUG_ANON(VAST_ARG(args));
```
dominiklohmann, line 24:
```suggestion
  VAST_TRACE(VAST_ARG(args));
```

libvast/src/system/explorer.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+#include <algorithm>
+
+namespace vast::system {
+
+explorer_state::explorer_state(caf::event_based_actor*) {
+  // nop
+}
+
+void explorer_state::forward_results(vast::table_slice_ptr slice) {
```
dominiklohmann, line 39: For the future reader/maintainer, this function could use some comments for every step.

libvast/src/system/application.cpp
lava: I guess we don't need them, but the other spawn subcommands include them as well.

dominiklohmann: Just leave them in. We have a story for removing them and can then do them all in one pass.

libvast/src/system/explore_command.cpp
dominiklohmann: Let's not do this in the scope of the initial version. We can always unify the code later on in a follow-up story.

libvast/src/system/count_command.cpp
lava: This is actually in a separate commit - I think it's low-frequency and high-value debug information, so I've been promoting these where i've noticed them.

libvast/src/system/node.cpp
lava: Good question - did we ever try to spawn to pivoters in parallel? Maybe we forgot to add it but  never noticed.

libvast/vast/system/explorer.hpp
lava: I think timespan is caf vocabulary, VAST has just `vast::duration`. I've changed it nonetheless, since it sounds like a more standard term than time box.

libvast/src/system/explore_command.cpp
lava: Right, updated. Maybe we should start printing deprecation warnings for the forms using something other than a single command-line argument, since we're planning on changing this at some point.

libvast/src/system/explore_command.cpp
lava: Sadly, I think the timespan parser is provided by CAF so its not trivial to change. But maybe it would make sense to change that anyways, due to ch16776.

integration/default_set.yaml
```diff
@@ -277,3 +277,20 @@ tests:
...
+  Explore Zeek With Overlap:
+    tags: [explore, zeek]
+    # This is a stress test for the deduplication: The query selects every
+    # event, and the huge timebox ensures that every event is included in
+    # the context of every other event.
+    steps:
+      - command: -N import zeek
+        input: data/zeek/conn.log.gz
+      - command: -N explore --before=10000h --after=10000h '#type == /zeek.*/'
+        transformation: wc -l
```
lava, line 296: Side comment: This test actually gets hit pretty hard by our current slow json/zeek(?) export bug, taking almost 30 seconds to write 8462 json results on my machine.

libvast/src/system/count_command.cpp
dominiklohmann: Fair enough.

integration/default_set.yaml
dominiklohmann: Can you add that as a comment into the story?

libvast/src/system/explore_command.cpp
dominiklohmann: We could accept strings here and use `to<vast::duration>` from parseable to parse the time.

lava: Good idea - I've added it to the linked story, so we can do it for both the eraser and the explorer uniformly.

PR #875, opened 2020-05-18
Deduce types for heterogenous JSONL import

dominiklohmann:


dominiklohmann:
[ch16779]

libvast/vast/format/json.hpp
```diff
@@ -58,6 +58,8 @@ struct default_selector {
   }
 
   caf::error schema(vast::schema sch) {
+    if (sch.empty())
+      return make_error(ec::invalid_configuration, "no schema provided");
```
mavam, line 62: Should this be checked at the call site already? In this function, we could make it a precondition.

dominiklohmann: Sounds more error prone. `*x.begin()` without a check just shouldn't pass a review.

mavam: Sorry if this was not clear, I'm talking about the *calling site* for the function `json::schema(..)`. That is where this check should go in my opinion. The reason: it's the same for all readers, so it can be factored.

I don't understand `*x.begin()` in the scope of this context.

dominiklohmann: I understood that, and I don't think it generally makes sense to assume non-empty schemas for all input formats. Both source and reader command call `reader.schema` regardless of the format. Some input formats don't require a schema at all (e.g., NetFlow) and simply ignore this call, and they should continue working when using VAST without schemas.

> I don't understand `*x.begin()` in the scope of this context.

I was referring to the line immediately below the code you're commenting on, which is using `*sch.begin()`, which is a code smell to begin with.

libvast/vast/system/type_registry.hpp
```diff
@@ -64,9 +66,12 @@ struct type_registry_state {
   caf::error load_from_disk();
 
   /// Store a new layout in the registry.
-  void insert(vast::record_type layout);
+  void insert(vast::type layout);
 
   /// Get a list of known types from the registry.
+  std::unordered_set<vast::type> lookup() const;
```
mavam, line 72: I know why you called this function `lookup`, but without a parameter, it's misleading. How about calling this function `types()` instead?

libvast/src/system/type_registry.cpp
```diff
@@ -64,13 +65,24 @@ caf::error type_registry_state::load_from_disk() {
...
   if (success)
     VAST_VERBOSE(self, "registered", hint->name());
 }
 
+std::unordered_set<vast::type> type_registry_state::lookup() const {
+  auto result = std::unordered_set<vast::type>{};
+  // TODO: Replace merging logic once libc++ implements unordered_set::merge.
+  // for ([[maybe_unused]] auto& [key, value] : data)
+  //   result.merge(value);
+  for ([[maybe_unused]] auto& [k, v] : data)
```
mavam, line 80: There's also `detail/set_operations.hpp` with some potentially useful set fu.

libvast/src/system/type_registry.cpp
```diff
@@ -82,16 +94,15 @@ type_registry_behavior
...
     VAST_DEBUG(self, "connects to", VAST_ARG(accountant));
     self->state.accountant = caf::actor_cast<accountant_type>(accountant);
     self->send(self->state.accountant, announce_atom::value, self->name());
     self->delayed_send(self, defaults::system::telemetry_rate,
                        telemetry_atom::value);
   }
-  // Load from disk if possible.
-  if (auto err = self->state.load_from_disk())
-    self->quit(std::move(err));
+  // Register an exit handler.
```
mavam, line 105:
```suggestion
  // Register the exit handler.
```

There is only one.

libvast/src/system/type_registry.cpp
```diff
@@ -102,30 +113,56 @@ type_registry(type_registry_actor self, const path& dir) {
...
-            // Retrieve a list of known states for a name.
-            return self->state.lookup(name);
-          }};
+  // Load existing state from disk if possible.
+  if (auto err = self->state.load_from_disk())
+    self->quit(std::move(err));
+  // Load already known types from the event_types singleton.
+  auto schema = vast::event_types::get();
+  if (!schema)
+    VAST_ERROR(self, "was unable to retrieve global event types");
```
mavam, line 122: I think the user is still left wondering what happened with this error message. How about relating it to the schemas on disk, which apparently failed to load?

dominiklohmann: I've made it an assertion. This code only runs during startup and should never fail.

libvast/src/system/type_registry.cpp
dominiklohmann: Sadly not for this use case. It's not too important anyways, the resulting set isn't all too large.

libvast/src/system/type_registry.cpp
```diff
@@ -102,30 +117,52 @@ type_registry(type_registry_actor self, const path& dir) {
...
-              });
-          },
-          [=](std::string name) {
-            // Retrieve a list of known states for a name.
-            return self->state.lookup(name);
-          }};
+  // Load existing state from disk if possible.
+  if (auto err = self->state.load_from_disk())
+    self->quit(std::move(err));
+  // Load already known types from the event_types singleton.
```
mavam, line 123:
```suggestion
  // Load loaded schema types from the singleton.
```

libvast/src/system/type_registry.cpp
```diff
@@ -102,30 +117,52 @@ type_registry(type_registry_actor self, const path& dir) {
...
+    },
+    [=](caf::stream<table_slice_ptr> in) {
+      VAST_TRACE(self, "attaches to", VAST_ARG("stream", in));
+      caf::attach_stream_sink(
+        self, in,
+        [=](caf::unit_t&) {
+          // nop
+        },
+        [=](caf::unit_t&, table_slice_ptr x) {
+          self->state.insert(std::move(x->layout()));
```
mavam, line 149: I know this was there before: but why are we *moving* from an immutable object?

libvast/vast/system/import_command.hpp
```diff
@@ -204,53 +180,71 @@ import_command(const command::invocation& invocation, caf::actor_system& sys) {
   auto guard = system::signal_monitor::run_guarded(
     sig_mon_thread, sys, defaults::system::signal_monitoring_interval, self);
   // Get node components.
-  auto components = get_node_components(self, node, {"accountant", "importer"});
+  auto components = get_node_components(
+    self, node, {"accountant", "type-registry", "importer"});
```
mavam, line 184: Can you add an assertion that the return set size is big enough so that the index access later on will be guaranteed to be valid?

libvast/vast/system/import_command.hpp
```diff
@@ -204,53 +180,71 @@ import_command(const command::invocation& invocation, caf::actor_system& sys) {
...
+  auto accountant = caf::actor_cast<accountant_type>((*components)[0]);
+  auto type_registry = caf::actor_cast<type_registry_type>((*components)[1]);
+  if (!type_registry)
+    return make_message(make_error(ec::missing_component, "type-registry"));
+  // Spawn the source.
+  if (!src)
+    src = sys.spawn(source<Reader>, std::move(*reader), slice_size, max_events,
+                    type_registry, schema ? std::move(*schema) : vast::schema{},
+                    type ? vast::data{std::move(*type)} : vast::data{},
+                    accountant);
```
mavam, line 196: Please refactor the ternary mess. 😉

libvast/vast/system/import_command.hpp
```diff
@@ -204,53 +180,71 @@ import_command(const command::invocation& invocation, caf::actor_system& sys) {
...
+  if (!invocation.arguments.empty()) {
+    auto expr = parse_expression(invocation.arguments.begin(),
+                                 invocation.arguments.end());
+    if (!expr)
+      return make_message(std::move(expr.error()));
+    self->send(src, std::move(*expr));
   }
   // Connect source to importer.
-  auto& importer = (*components)[1];
+  auto& importer = (*components)[2];
```
mavam, line 206: This could be moved to the top, right after getting `components` back.

libvast/vast/system/source.hpp
```diff
@@ -109,29 +115,73 @@ struct source_state {
...
+            vast::data type_restr, accountant_type acc) {
+    // Create the reader.
     name = reader.name();
     new (&reader) Reader(std::move(rd));
+    reader_initialized = true;
     remaining = std::move(max_events);
-    initialized = true;
+    local_schema = std::move(sch);
+    type_restriction = std::move(type_restr);
+    // Set accountant.
```
mavam, line 143: This comment confused me. We are registering ourselves with the accountant, no?

libvast/vast/system/source.hpp
```diff
@@ -109,29 +115,73 @@ struct source_state {
...
+            type_registry_type type_registry, vast::schema sch,
+            vast::data type_restr, accountant_type acc) {
+    // Create the reader.
     name = reader.name();
     new (&reader) Reader(std::move(rd));
+    reader_initialized = true;
     remaining = std::move(max_events);
-    initialized = true;
+    local_schema = std::move(sch);
+    type_restriction = std::move(type_restr);
```
mavam, line 142:
```suggestion
    type_filter = std::move(type_restr);
```

This may need adjusting other stuff as well, but feels much easier to understand.

libvast/vast/system/source.hpp
```diff
@@ -256,7 +316,7 @@ caf::behavior default_source(caf::stateful_actor<source_state<Reader>>* self,
                              Reader reader) {
   auto slice_size = get_or(self->system().config(), "system.table-slice-size",
                            defaults::system::table_slice_size);
-  return source(self, std::move(reader), slice_size, caf::none);
+  return source(self, std::move(reader), slice_size, caf::none, {}, {}, {}, {});
```
mavam, line 318: Maybe add a comment that this is for testing only?

libvast/vast/json.hpp
```diff
@@ -234,6 +235,44 @@ json to_json(const T& x, Opts&&... opts) {
 
 json::object combine(const json::object& lhs, const json::object& rhs);
 
+namespace detail {
+
+template <class F>
+void each_field_impl(const json& x, F f, std::string& prefix) {
```
mavam, line 241: Instead of using  a `std::string` here, you could consider `detail::stack_vector` with a `std::string`, similar to `record_type::each`.

libvast/vast/format/json.hpp
```diff
@@ -53,33 +56,53 @@ caf::error add(table_slice_builder& builder, const vast::json::object& xs,
...
 struct default_selector {
-  caf::optional<record_type> operator()(const vast::json::object&) {
-    return layout;
+  caf::optional<record_type> operator()(const vast::json::object& obj) const {
+    std::vector<std::string> cache_entry;
+    auto build_cache_entry
+      = [&cache_entry](const std::string& prefix, const vast::json&) {
+          cache_entry.emplace_back(prefix);
+        };
+    each_field(vast::json{obj}, build_cache_entry);
```
mavam, line 65: I think we should operate on `obj` *directly*. Otherwise we construct a new object for every invocation of `each_field`.

libvast/vast/format/json.hpp
```diff
@@ -53,33 +56,53 @@ caf::error add(table_slice_builder& builder, const vast::json::object& xs,
...
+    std::sort(cache_entry.begin(), cache_entry.end());
+    for (const auto& [k, v] : type_cache)
+      if (k == cache_entry)
+        return v;
+    // FIXME: This does not work as expected w/ detail::flat_map.
+    // if (auto search_result = type_cache.find(cache_entry);
+    //     search_result != type_cache.end()) {
+    //   VAST_WARNING_ANON(VAST_ARG(search_result->first));
+    //   return search_result->second;
+    // }
```
mavam, line 75: Can you create a follow-up story for this?

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🎁 The JSON import now automatically selects exactly matching types by field
+  name from all known types. The `--type` / `-t` option for `vast import` can
+  now be used with a regular expression to restrict the list of matching types.
```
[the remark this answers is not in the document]
dominiklohmann: I've changed it to this:

```
- 🎁 VAST now keeps track of the types of imported events and can re-use them
  for consecutive imports even when they're not specified in the schema. A list
  of registered types can now be viewed using `vast status` under the key
  `node.type-registry.types`. [#875](https://github.com/tenzir/vast/pull/875)

- 🎁 When importing JSON data without knowing the type of the imported events a
  priori, VAST now supports automatic event type deduction based on the JSON
  object keys in the data. VAST selects a type _iff_ the set of fields match a
  known type. The `--type` / `-t` option to the `import` command restricts the
  matching to the set of types that share the provided prefix. Omitting `-t`
  attempts to match JSON against all known types.
  [#875](https://github.com/tenzir/vast/pull/875)
```

libvast/src/system/type_registry.cpp
dominiklohmann: I'll create a follow-up story for this.

libvast/src/system/type_registry.cpp
```diff
@@ -120,11 +120,11 @@ type_registry(type_registry_actor self, const path& dir) {
   // Load existing state from disk if possible.
   if (auto err = self->state.load_from_disk())
     self->quit(std::move(err));
-  // Load already known types from the event_types singleton.
+  // Load loaded schema types from th singleton.
```
mavam, line 123: `th`?

CHANGELOG.md
```diff
@@ -10,10 +10,17 @@ Every entry has a category for which we use the following visual abbreviations:
...
 ## Unreleased
 
-- 🎁 The JSON import now automatically selects exactly matching types by field
-  name from all known types. The `--type` / `-t` option for `vast import` can
-  now be used as a prefix for selecting matching types. I.e., `-t suricata` will
-  match all types whose name starts with `suricata`.
+- 🎁 VAST now keeps track of the types of imported events and can re-use them
+  for consecutive imports even when they're not specified in the schema. A list
+  of registered types can now be viewed using `vast status` under the key
+  `node.type-registry.types`. [#875](https://github.com/tenzir/vast/pull/875)
```
mavam, line 16: What does this buy me from a user perspective? It's currently written about what happens, not with how the user can benefit from this.

libvast/vast/format/json.hpp
```diff
@@ -59,8 +60,10 @@ struct default_selector {
   caf::optional<record_type> operator()(const vast::json::object& obj) const {
     std::vector<std::string> cache_entry;
     auto build_cache_entry
-      = [&cache_entry](const std::string& prefix, const vast::json&) {
-          cache_entry.emplace_back(prefix);
+      = [&cache_entry](detail::stack_vector<std::string_view, 64>& prefix,
```
mavam, line 63:
```suggestion
      = [&cache_entry](auto& prefix,
```

Because it's a messy type, I don't think it's necessary to expose it to the user.

CHANGELOG.md
dominiklohmann: > even when they're not specified in the schema

This is the benefit: Once an event of type X was imported successfully, it will always be available regardless of the availability of the original schema.

mavam: My point is: I'm taking on the user-perspective and cannot interpret the benefit directly. So we need to communicate the message better (-> rewording).

libvast/vast/format/json.hpp
```diff
@@ -58,6 +58,12 @@ caf::error add(table_slice_builder& builder, const vast::json::object& xs,
 /// @relates reader
 struct default_selector {
   caf::optional<record_type> operator()(const vast::json::object& obj) const {
+    if (type_cache.empty())
+      return caf::none;
+    // Iff there is only one type in the type cache, allow the JSON reader to
+    // use it despite it not being a perfect fit.
```
mavam, line 64:
```suggestion
    // use it despite not being an exact match.
```

libvast/src/system/type_registry.cpp
lava: The pseudo-code above is also wrong, it should be just `result.merge(data)`

PR #876, opened 2020-05-18
Fix typo in vast explore samples

dominiklohmann:


doc/cli/vast-explore.md
```diff
@@ -28,7 +28,7 @@ selection options. For example, this invocation shows all DNS requests captured
 by Zeek up to 60 seconds after a connection to 192.168.1.10:
 
 ```bash
-vast explore -A 60secs --for=zeek.dns 'zeek.conn.id.resp_h == 192.168.1.10'
+vast explore -A 60sec --for=zeek.dns 'zeek.conn.id.resp_h == 192.168.1.10'
```
rolandpeelen, line 31:
```
error: invalid option parameter

-A 60sec --for=zeek.dns zeek.conn.id.resp_h == 192.168.1.10
^~

usage: explore [<parameters>]

explore context around query results

parameters:
  [-h | -? | --help]             prints the help text
  [-A | --after=] <timespan>     include all records up to this much time after each result
  [-B | --before=] <timespan>    include all records up to this much time before each result
```

rolandpeelen: @dominiklohmann - would it throw this error as well for the `--for` not being implemented yet? Because I'm getting the same error

dominiklohmann: Fixed it, it had to be `s` instead of `sec`. We use CAF for parsing this, which is a bit suboptimal.

PR #877, opened 2020-05-19
Add '--by' option to 'vast explore'

lava:


libvast/src/system/application.cpp
```diff
@@ -107,10 +107,11 @@ auto make_explore_command() {
...
     opts("?explore")
-      .add<vast::duration>("after,A", "include all records up to this much"
-                                      " time after each result")
-      .add<vast::duration>("before,B", "include all records up to this much"
-                                       " time before each result"));
+      .add<std::string>("after,A", "include all records up to this much"
+                                   " time after each result")
+      .add<std::string>("before,B", "include all records up to this much"
+                                    " time before each result")
+      .add<std::string>("by", "perform an equi-join on the given field"));
```
dominiklohmann, line 114:
```suggestion
      .add<std::string>("by", "perform an equijoin on the given field"));
```

libvast/src/system/explorer.cpp
```diff
@@ -104,6 +116,18 @@ explorer(caf::stateful_actor<explorer_state>* self, caf::actor node,
         VAST_DEBUG(self, "could not find timestamp field in", layout);
         return;
       }
+      std::optional<table_slice::column_view> by_column;
+      if (st.by) {
+        // Need to pivot from caf::optional to std::optional here, as the
+        // former doesnt support emplace or value assignment.
+        if (auto vopt = slice->column(*st.by); vopt)
```
dominiklohmann, line 123:
```suggestion
        if (auto vopt = slice->column(*st.by))
```

libvast/src/system/explorer.cpp
```diff
@@ -122,16 +146,38 @@ explorer(caf::stateful_actor<explorer_state>* self, caf::actor node,
...
+        auto build_conjunction
+          = [](std::optional<expression>&& lhs,
+               std::optional<expression>&& rhs) -> std::optional<expression> {
+          if (lhs && rhs) {
+            return conjunction{std::move(*lhs), std::move(*rhs)};
+          } else if (lhs) {
+            return std::move(lhs);
+          } else if (rhs) {
+            return std::move(rhs);
+          }
```
dominiklohmann, line 168:
```suggestion
          if (lhs && rhs)
            return conjunction{std::move(*lhs), std::move(*rhs)};
          if (lhs)
            return std::move(lhs);
          if (rhs)
            return std::move(rhs);
```

libvast/src/system/spawn_explorer.cpp
```diff
@@ -27,18 +29,66 @@
...
 
 namespace vast::system {
 
+caf::error explorer_validate_args(const caf::settings& args) {
+  auto before_arg = caf::get_if<std::string>(&args, "explore.before");
+  auto after_arg = caf::get_if<std::string>(&args, "explore.after");
+  auto by_arg = caf::get_if(&args, "explore.by");
+  if (!before_arg && !after_arg && !by_arg)
+    return make_error(ec::invalid_configuration, "At least one of '--before', "
+                                                 "'--after' or '--by' must be "
```
dominiklohmann, line 44:
```suggestion
                                                 "'--after', or '--by' must be "
```

libvast/src/system/spawn_explorer.cpp
```diff
@@ -27,18 +29,66 @@
...
+                        "as duration.");
+    before = *d;
+  }
+  if (after_arg) {
+    auto d = to<vast::duration>(*after_arg);
+    if (!d)
+      return make_error(ec::invalid_argument, "Could not parse", *after_arg,
+                        "as duration.");
+    after = *d;
+  }
```
dominiklohmann, line 61: I've tried coming up with a way to make this less repetitive, but every solution involving a lambda still needs the duplicated error handling, which makes me think this is fine.

libvast/src/system/spawn_explorer.cpp
```diff
@@ -27,18 +29,66 @@
...
+  if (!by_arg) {
+    if (before <= 0s && after <= 0s)
+      return make_error(ec::invalid_argument,
+                        "At least one of '--before' or '--after' must be > 0 "
```
dominiklohmann, line 65:
```suggestion
                        "At least one of '--before', or '--after' must be greater than 0 "
```

libvast/test/system/explorer.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+#include "vast/system/spawn_explorer.hpp"
+#include "vast/time.hpp"
+
+#include <caf/settings.hpp>
+
+using namespace std::chrono_literals;
+
+TEST(explorer config) {
+  { // Specifying no options at all is not allowed.
```
dominiklohmann, line 26:
```suggestion
  MESSAGE("Specifying no options at all is not allowed.");
  {
```

libvast/test/system/explorer.cpp
```diff
@@ -0,0 +1,65 @@
...
+    caf::settings settings;
+    auto& explore = settings["explore"].as_dictionary();
+    CHECK_NOT_EQUAL(vast::system::explorer_validate_args(settings), caf::none);
+  }
+
+  { // Specifying only time is allowed, as long as it is > 0.
```
dominiklohmann, line 32:
```suggestion
  MESSAGE("Specifying only time is allowed, as long as it is > 0.");
  {
```

libvast/test/system/explorer.cpp
```diff
@@ -0,0 +1,65 @@
...
+    auto& explore = settings["explore"].as_dictionary();
+    explore["before"] = "0s";
+    explore["after"] = "0s";
+    CHECK_NOT_EQUAL(vast::system::explorer_validate_args(settings), caf::none);
+    explore["after"] = "10s";
+    std::cerr << caf::to_string(settings) << std::endl;
+    CHECK_EQUAL(vast::system::explorer_validate_args(settings), caf::none);
+  }
+
+  { // Specifying only 'by' is allowed.
```
dominiklohmann, line 43:
```suggestion
  MESSAGE("Specifying only 'by' is allowed.");
  {
```

libvast/test/system/explorer.cpp
```diff
@@ -0,0 +1,65 @@
...
+    caf::settings settings;
+    auto& explore = settings["explore"].as_dictionary();
+    explore["by"] = "0s";
+    CHECK_EQUAL(vast::system::explorer_validate_args(settings), caf::none);
+  }
+
+  { // Malformed input is not allowed.
```
dominiklohmann, line 50:
```suggestion
  MESSAGE("Malformed input is not allowed.");
  {
```

libvast/test/system/explorer.cpp
```diff
@@ -0,0 +1,65 @@
...
+    caf::settings settings;
+    auto& explore = settings["after"].as_dictionary();
+    explore["by"] = "MIP = RE";
+    CHECK_NOT_EQUAL(vast::system::explorer_validate_args(settings), caf::none);
+  }
+
+  { // Specifying all options is fine.
```
dominiklohmann, line 57:
```suggestion
  MESSAGE("Specifying all options is fine.");
  {
```

libvast/src/system/spawn_explorer.cpp
lava: Yeah, I had the same experience, a macro would work but thats not really better :/

libvast/vast/optional.hpp
```diff
@@ -15,10 +15,20 @@
...
 
+#include <optional>
+
 namespace vast {
 
 /// A drop-in replacement for C++17's std::optional.
 using caf::optional;
 
+template <typename T>
+std::optional<T> to_std(caf::optional<T>&& opt) {
```
mavam, line 26:
```suggestion
std::optional<T> to_std(caf::optional<T> opt) {
```

Anything against this?

lava: It would work as well, but it leaves the possibility to accidentally make huge copies without noticing due to user error.

lava:
[ch16776]

libvast/vast/optional.hpp
mavam: I'm fine with starting out with rvalue reference. In general, having values as parameters gives flexibility on the call site. For example, I would want to pass an `optional<int>` by value.

libvast/src/system/explorer.cpp
```diff
@@ -122,16 +146,37 @@ explorer(caf::stateful_actor<explorer_state>* self, caf::actor node,
...
+          auto ci = (*by_column)[i];
+          if (caf::get_if<caf::none_t>(&ci))
+            continue;
+          // TODO: Make `predicate` accept a data_view as well to save
+          // the call to `materialize()`.
+          by_expr = predicate{key_extractor{*st.by}, equal, materialize(ci)};
         }
-        VAST_TRACE(self, "constructed expression", expr);
-        auto query = to_string(expr);
+        auto build_conjunction
```
mavam, line 159:
```suggestion
        auto make_conjunction
```

Pretty much all factory functions have a `make_` prefix in the code base. Using this here as well would improve consistency.

dominiklohmann: Speaking of this, I think this whole helper function could quite easily be made variadic. It's three conjunctions now and it'll only become more in the future.

mavam: The principle of using the least complex mechanism to achieve a goal would guide me to keep this as is. If we need it more flexible later on, then we can refactor still.

dominiklohmann: This works and I don't think it's any harder to read nor is it a lot more complex.

```cpp

template <class Lhs, class... Rhs>
auto make_conjunction(Lhs&& lhs, Rhs&&... rhs) -> std::optional<expression> {
  if constexpr (sizeof...(rhs) == 0)
    return std::forward<Lhs>(lhs);
  else {
    auto rhs_eval = make_conjunction(std::forward<Rhs>(rhs)...);
    if (lhs && rhs_eval)
      return conjunction{*std::forward<Lhs>(lhs), std::move(*rhs_eval)};
    if (lhs)
      return std::forward<Lhs>(lhs);
    return std::move(rhs_eval);
  }
};
```

```cpp
        auto expr = make_conjunction(std::move(before_expr),
            std::move(after_expr), std::move(by_expr));
```

mavam: But this is a big difference! Now the function is recursive and cluttered with `move` and `forward`. The metal toll for the reader/reviewer has significantly increased.

Even though it's pretty, we don't have a use case for it *today*. When we need such heavy machinery, we can still deploy this. Until then, I would prefer to prioritize KISS.

dominiklohmann: I'd argue the readability on the call site improves a lot and is arguably more important to the reader. Anyways, we can revisit this discussion once we have more than three conjunctions at the call site.

This whole verbosity with `optional` could've been avoided with an always-true expression, which is something that we should maybe consider adding. `T` gets expanded to `:bool == T`, so that can't just work. It's just a useful thing to have in general for crafting queries.

lava: At the risk of beating the dead horse, but this is also not yet at its final stage of complexity because we'd still need to add a recursive static_assert that all types are `std::optional<expression>`.

dominiklohmann: You'd just need to check that for Lhs, which is simple enough. 😅

libvast/vast/optional.hpp
lava: If you want to keep using the original optional after the call you can still do that, the copy just needs to be made explicitly:

```
caf::optional<int> copt;
auto sopt = to_std(caf::optional<int>{copt});
```

While its technically not possible to actually pass by value using only this overload, for types like `int` that don't define a separate move constructor, this is ultimately going to reduce to using the copy constructor anyways so the generated code will be the same.

libvast/src/system/spawn_explorer.cpp
```diff
@@ -27,18 +29,67 @@
...
+    return *parsed;
+  };
+  auto before = maybe_parse(
+    caf::get_if<std::string>(&args.invocation.options, "explore.before"));
+  auto after = maybe_parse(
+    caf::get_if<std::string>(&args.invocation.options, "explore.after"));
+  std::optional<std::string> by
+    = to_std(caf::get_if<std::string>(&args.invocation.options, "explore.by"));
+  auto expl = self->spawn(explorer, self, before, after, by);
   return expl;
```
dominiklohmann, line 93:
```suggestion
  return self->spawn(explorer, self, before, after, by);
```

lava: Resolving as I don't want to re-run CI for this, but I changed it in my working tree and promise to add it into the review for adding limits to `vast explore` :crossed_fingers:

PR #880, opened 2020-05-20
Re-enable macOS in CI

dominiklohmann:
We've been bitten by this a lot lately, so this brings it all back to CI.


0ortmann:
Matrix changes LGTM, let's just wait until all tests have run through.

PR #882, opened 2020-05-20
Add limit options for vast explore

lava:


libvast/vast/defaults.hpp
```diff
@@ -149,6 +149,25 @@ struct pcap {
...
+/// Maximum total number of results.
+constexpr size_t max_events = 0;
+
+/// Maximum number of results for the initial query.
+constexpr size_t limit1 = 1000;
+
+/// Maximum number of results for every explored context.
+constexpr size_t limit2 = 1000;
+
+} // namespace explore
```
lava, line 169: These numbers are basically just guesses, not sure if there's a more systematic way to determine good defaults.

dominiklohmann: If I recall correctly we've used 100 for other default max event limits before.

libvast/vast/system/explorer.hpp
```diff
@@ -28,13 +28,24 @@
...
   explorer_state(caf::event_based_actor* self);
 
   /// Send the results to the sink, after removing duplicates.
   void forward_results(vast::table_slice_ptr slice);
 
+  /// Maximum number of events to output.
+  event_limits limits;
+
+  /// Number of events sent to the source.
+  size_t sent;
```
dominiklohmann, line 47:
```suggestion
  size_t num_sent;
```

We usually prefix counter variables with `num_`.

libvast/vast/defaults.hpp
```diff
@@ -149,6 +149,25 @@ struct pcap {
...
+// by `min(max_events, limit1*limit2)`.
+
+/// Maximum total number of results.
+constexpr size_t max_events = 0;
+
+/// Maximum number of results for the initial query.
+constexpr size_t limit1 = 1000;
+
+/// Maximum number of results for every explored context.
+constexpr size_t limit2 = 1000;
```
dominiklohmann, line 167: Do you think there'll ever be more than two stages? in that case, would a `vector<size_t>` make more sense? We could name is `max_events_per_stage`.

libvast/vast/system/explorer.hpp
```diff
@@ -28,13 +28,24 @@
 namespace vast::system {
 
 struct explorer_state {
+  struct event_limits {
+    int64_t total;
+    int64_t per_result;
```
dominiklohmann, line 33: Why are these signed integers?

lava: I originally had used `-1` to mean 'unlimited' but now it doesnt make sense any more.

dominiklohmann: I think we can just use `std::numeric_limits<T>::max` as default/unlimited now.

libvast/src/system/explorer.cpp
```diff
@@ -107,6 +126,9 @@ explorer(caf::stateful_actor<explorer_state>* self, caf::actor node,
         st.forward_results(slice);
         return;
       }
+      // Dont bother making new queries if we'd discard all results anyways.
```
mavam, line 129:
```suggestion
      // Don't bother making new queries if we discard all results anyways.
```

libvast/src/system/spawn_explorer.cpp
```diff
@@ -83,14 +83,18 @@ maybe_actor spawn_explorer(node_actor* self, spawn_arguments& args) {
...
   };
-  auto before = maybe_parse(
-    caf::get_if<std::string>(&args.invocation.options, "explore.before"));
-  auto after = maybe_parse(
-    caf::get_if<std::string>(&args.invocation.options, "explore.after"));
+  const auto& options = args.invocation.options;
+  auto before
+    = maybe_parse(caf::get_if<std::string>(&options, "explore.before"));
+  auto after = maybe_parse(caf::get_if<std::string>(&options, "explore.after"));
   std::optional<std::string> by
```
mavam, line 90:
```suggestion
  auto by
```

libvast/vast/system/explorer.hpp
mavam: Yeah, I think we get "unlimited" for free this way.

doc/cli/vast-explore.md
```diff
@@ -109,6 +109,22 @@ in the query result.
...
+means unlimited. This option only restricts the output of `vast explore`,
+internally more than N results may be processed while preparing the results.
+
+`--limit1=NUMBER`, `--limit2=NUMBER`:
+
+Restricts the number of results returned by the initial query and by every
+follow-up query, respectively. A value of 0 means unlimited.
+
+Note that these limits apply before result deduplication, so the total number
+of returned results can be less than limit1 * limit2, even if more results
```
mavam, line 124:
```suggestion
of returned results can be less than `limit1 * limit2`, even if more results
```

doc/cli/vast-explore.md
```diff
@@ -109,6 +109,22 @@ in the query result.
 
 **Output**
 
+`--max-events=N`:
+
+An invocation of `vast explore` will print at most N events. A value of 0
+means unlimited. This option only restricts the output of `vast explore`,
+internally more than N results may be processed while preparing the results.
+
+`--limit1=NUMBER`, `--limit2=NUMBER`:
```
mavam, line 118: If I understand your choice of numeric suffixes correctly, you chose them because they express the sequencing of the limits. Maybe we can use something that's focusing less on the way the exploration works, but instead how to control the result. My suggestion would be `query-limit` and `context-limit`. I'm not sold on those, but just wanted to start the discussion.

libvast/vast/system/explorer.hpp
lava: I don't mind doing it, but worth noting that this will then be inconsistent with the semantics of `export_::max_events` where 0 is unlimited.

PR #883, opened 2020-05-25
Dont overwrite default options when running a command

lava:
By assigning the keys of `invocation.options`, we would reset the already merged options to their un-merged original state, erasing any values that were merged from the default configuration.

### DIFF
--- a/libvast/src/command.cpp
+++ b/libvast/src/command.cpp
@@ -223,6 +223,23 @@ void mantext(const command& cmd, std::ostream& os,
   os << '\n';
 }
 
+/// Merge settings of `src` into `dst`, overwriting existing values
+/// from `dst` if necessary.
+void merge_settings(const caf::settings& src, caf::settings& dst,
+                    size_t depth = 0) {
+  if (depth > 100) {
+    VAST_ERROR_ANON("Exceeded maximum nesting depth in settings.");
+    return;
+  }
+  for (auto& [key, value] : src) {
+    if (caf::holds_alternative<caf::settings>(value))
+      merge_settings(caf::get<caf::settings>(value), dst[key].as_dictionary(),
+                     depth + 1);
+    else
+      dst.insert_or_assign(key, value);
+  }
+}
+
 } // namespace
 
 command::command(std::string_view name, std::string_view description,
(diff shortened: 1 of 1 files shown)


libvast/src/command.cpp
```diff
@@ -436,9 +436,9 @@ run(const command::invocation& invocation, caf::actor_system& sys,
   if (auto search_result = fact.find(invocation.full_name);
       search_result != fact.end()) {
     auto merged_invocation = invocation;
+    // The original `sys.config()` was merged with the invocation options
+    // by `init_config()` above before this function is called.
     merged_invocation.options = content(sys.config());
-    for (const auto& [key, value] : invocation.options)
-      merged_invocation.options.insert_or_assign(key, value);
     return std::invoke(search_result->second, merged_invocation, sys);
```
dominiklohmann, line 442: This whole change can just be reduced to this:

```cpp
invocation.options = content(sys.config());
return std::invoke(search_result->second, invocation, sys);
```

lava: But `invocation` is const in this context.

dominiklohmann: In that case it might just make sense to have the function take the parameter by value.

lava: As it turns out, we *do* need to merge after all. After the latest update, passing by const& or by value would return in the same number of lines changed.

libvast/src/command.cpp
```diff
@@ -351,24 +351,30 @@ parse(const command& root, command::argument_iterator first,
   return result;
 }
 
+/// Merge settings of `src` into `dst`, overwriting existing values
+/// from `dst` if necessary.
+static void
```
dominiklohmann, line 356: Are we doing `static` over unnamed namespaces officially now?

libvast/src/command.cpp
```diff
@@ -351,24 +351,30 @@ parse(const command& root, command::argument_iterator first,
...
+    VAST_ERROR_ANON("Exceeded maximum nesting depth in settings.");
+    return;
+  }
+  for (auto& [key, value] : src)
+    if (caf::holds_alternative<caf::settings>(value)) {
+      merge_settings(caf::get<caf::settings>(value), dst[key].as_dictionary(),
+                     depth + 1);
+    } else {
+      dst.insert_or_assign(key, value);
+    }
```
dominiklohmann, line 368:
```suggestion
  for (auto& [key, value] : src) {
    if (caf::holds_alternative<caf::settings>(value))
      merge_settings(caf::get<caf::settings>(value), dst[key].as_dictionary(),
                     depth + 1);
    else
      dst.insert_or_assign(key, value);
  }
```

libvast/src/command.cpp
lava: Apparently not ;)

PR #886, opened 2020-05-27
Publish Sysmon schema

dominiklohmann:
@tenzir/backend side note: I've also taken a few minutes to fix some warnings I've encountered and just put them in here. The only warnings left with clang are deprecation warnings for Arrow 0.17 now.


libvast/vast/format/json.hpp
```diff
@@ -104,7 +104,7 @@ struct default_selector {
     return "json-reader";
   }
 
-  detail::flat_map<std::vector<std::string>, record_type> type_cache = {};
+  detail::stable_map<std::vector<std::string>, record_type> type_cache = {};
```
lava, line 107: I have no opinion on the change itself, but this seems to be a bit more than "fixing a warning".

Can we pull this into a separate commit that includes the reasoning in the commit message (and maybe also before/after benchmarking results if performance was the reason for changing the data structure)

integration/default_set.yaml
```diff
@@ -302,3 +302,10 @@ tests:
       - command: -N import zeek
         input: data/zeek/dns.log.gz
       - command: -N explore --by='id.orig_h' --after=1s --before=0s 'zeek.dns.query == "survey.112.2o7.net"'
+  Heteogenous JSONL import:
```
lava, line 305: Missing an 'r'

dominiklohmann:
> Looks good modulo the comments on the flat_map/stable_map change.

That was supposed to be one commit further up. Will rebase in a sec. It's actually a necessity, it was slightly broken.

libvast/vast/format/json.hpp
dominiklohmann: That was supposed to be one commit further up. Will rebase in a sec. It's actually a necessity, it was slightly broken.

lava: Sounds good then, the idea has to be "make it correct" and only then "make it fast".

Let's still take a close look at the next set of benchmark results, to see if this is introducing a performance regression or if it doesnt matter.

PR #892, opened 2020-05-29
Fix stream error state handling in getline_generic

lava:
Fix several issues with the `vast::generic_getline()` function:

  1) Check if sentry construction was succesful, and abort if not.

  2) Set the input streams failbit if none or too many characters
were extracted from the input stream to be consistent with the
behaviour of std::getline.

  3) Always set eofbit when encountering eof to avoid an extra blank
line at the end of the input.

The lack of (1) above could lead to a rather subtle race condition in
combination with line_range as follows:

When input was read from e.g. a socket and the last line before
encountering a read timeout was a blank line, the input stream would
have the eof bit set to signal the read timeout. Since line_range
would see an empty line returned with the input stream still being
"true" (but not .good()), so it would attempt to read another line.

On the next invocation of `getline_generic`, the constructor of the
sentry object will see that the input stream is not `good()` and thus
set `failbit`, making the stream "false". Since the result of the
sentry construction was ignored, the function then continue to happily
operate on the readbuf of a failed input stream.

If in the meantime fresh data arrived from the network, the fdinbuf
would be able to successfully read that data on the next underflow
and clear the read timeout bit, leading to an inconsistent state.
With the last read being successful, the streams error state would
not be reset as it should, leading the source to believe it received
an EOF and to terminate the stream.

Note that (3) above makes this race much more visible, since it will
generate a blank line at the end of *every* input, automatically
satisfying half of the error conditions leading to the race.

### DIFF
--- a/libvast/vast/detail/getline_generic.hpp
+++ b/libvast/vast/detail/getline_generic.hpp
@@ -11,14 +11,17 @@
  * contained in the LICENSE file.                                             *
  ******************************************************************************/
 
+// This file comes from a 3rd party and has been adapted to fit into the VAST
+// code base. Details about the original code:
+//
+// - Site:       https://stackoverflow.com/a/6089413/92560
+// - Author:     user763305 et al.
+// - License:    CC-BY-SA
+
 #pragma once
 
 #include <istream>
 #include <string>
-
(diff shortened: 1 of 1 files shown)


lava:
[ch17115]

PR #895, opened 2020-06-03
Fix odr violations in escapers header

lava:
Noticed after compiling with gcc 10.1 for the first time.


mavam:
> Why are we using function objects over functions here, anyways?

Not sure, to be honest. Nothing speaks against using templated functions here. The current diff looks like the least-invasive change to get out of UB land. 

When it comes to rethinking the design of the escaping facilities, we may want to take a step back and see whether output iterators could be replaced with something more flexible, e.g., `span<byte>`.

PR #896, opened 2020-06-03
Fix use-after-free bug in indexer state

lava:
NOTE: I did not see any mechanism to ensure that index and indexer are destroyed in the correct order, but if one exists it would probably better to fix that mechanism rather than introducing shared_ptr here.


tobim:
This should actually be covered by the protocol between the index and the indexers. The indexers send a `done` message once they are finished processing all incoming table slices, this is then handled by `index_state::decrement_indexer_count()` in the index actor.

libvast/src/system/index.cpp
```diff
@@ -436,6 +436,12 @@ caf::behavior index(caf::stateful_actor<index_state>* self, const path& dir,
   self->set_exit_handler([=](const caf::exit_msg& msg) {
     VAST_DEBUG(self, "received exit from", msg.source,
                "with reason:", msg.reason);
+    auto& st = self->state;
+    if (!st.unpersisted.empty() || st.active_partition_indexers > 0) {
+      auto delay = defaults::index::shutdown_retry_interval;
+      VAST_INFO(self, "delaying exit by", delay, "to wait for outstanding indexers");
+      self->delayed_send(self, delay, msg);
+    }
```
tobim, line 444: Don't we also need to tell the indexers to exit when we get here for the first time?

lava: Good question, I was assuming this only happens when the system shuts down so everything else  should have gotten exit messages as well.

If we can also shutdown the index at runtime then I think we *do* need to kill the indexers here as well (and add an API to `partition` for that, since the index doesnt have direct knowledge of the existing indexers)

I'm wondering if it could happen that a new indexer is about to be launched but it doesnt exist yet when the `exit` message arrives at the index.

tobim:
FYI: I can locally reproduce the issue with
```sh
python integration/integration.py -d it --app build/ci/bin/vast -t "Conn log counting"
```
The current change does not fix the problem.

lava:
@tobim I added a missing `return;` statement so it may be fixed now, although I still cant reproduce this locally on my machine.

tobim:
@tenzir/backend a review would be very welcome.

libvast/src/system/indexer_downstream_manager.cpp
```diff
@@ -56,6 +56,27 @@ size_t indexer_downstream_manager::buffered(caf::stream_slot slot) const
   return 0u;
 }
 
+void indexer_downstream_manager::close() {
+  VAST_DEBUG(self(), "closing downstream manager");
+  if (closing)
+    return;
+  for (auto it = partitions.begin(); it != partitions.end();) {
+    if (buffered(**it) == 0u) {
```
dominiklohmann, line 64: Can this function get a few comments? The double de-referencing scares me.

CMakeLists.txt
```diff
@@ -500,11 +501,13 @@ if (APPLE AND NOT OPENSSL_ROOT_DIR)
...
-  CAF 0.17.5 EXACT
-  COMPONENTS core io test
-  QUIET)
+if (NOT VAST_NO_SYSTEM_CAF)
+  # Try to find the required CAF components first...
+  find_package(
+    CAF 0.17.5 EXACT
+    COMPONENTS core io test
+    QUIET)
+endif ()
```
dominiklohmann, line 510: Should be reflected in the configure file.

libvast/src/system/indexer_downstream_manager.cpp
```diff
@@ -72,10 +93,19 @@ int32_t indexer_downstream_manager::max_capacity() const noexcept {
 void indexer_downstream_manager::register_partition(partition* p) {
   VAST_DEBUG(self(), "registers partition", p->id());
   partitions.insert(p);
+  // Corner case: it is possible that register gets called
```
dominiklohmann, line 96: I don't quite understand this comment.

libvast/src/system/indexer_downstream_manager.cpp
tobim: I added a few interleaved lines, but I did not address the double referencing directly. It's just what you get if you have pointers in a container and iterate over that.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 <!-- ## Unreleased -->
```
dominiklohmann, line 12: This can be uncommented.

libvast/src/system/indexer_downstream_manager.cpp
tobim: I somehow forgot to finish that, updated.

tobim:
I'll wait a bit  with the rebase push so you can verify if your comments are addressed.

libvast/src/system/indexer_downstream_manager.cpp
dominiklohmann: I just wanted to know whether it is a safe assumption that `*it != nullptr` at all times. This could also be expressed as an assertion above this line.

dominiklohmann:
I resolved every item but one. From my side this is ready to be merged, but I cannot reproduce the issue locally, which means I also cannot test it.

libvast/src/system/indexer_downstream_manager.cpp
tobim: Assertions are a great idea. Integrated.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 A use after free bug would sometimes crash the node while it was shutting
+  down. [#896](https://github.com/tenzir/vast/pull/896)
```
lava, line 15: nit: The bug didn't crash the node, it was our asan instrumentation that did ;)

configure
```diff
@@ -56,6 +56,7 @@ Usage: $0 [OPTION]... [VAR=VALUE]...
   Required packages:
     --with-arrow=PATH       path to Apache Arrow install root
     --without-arrow         explicitly disable Apache Arrow integration
+    --without-system-caf    build CAF from the submodule
```
lava, line 59: nit: should be one line lower to be consistent with the arrow options

CMakeLists.txt
```diff
@@ -78,6 +78,7 @@ endif ()
 
 option(VAST_ENABLE_ASSERTIONS "Enable Assertions" "${VAST_ENABLE_ASSERTIONS_DEFAULT}")
 option(VAST_RELOCATABLE_INSTALL "Enable relocatable installations" ON)
+option(VAST_NO_SYSTEM_CAF "Always use the CAF submodule" OFF)
```
lava, line 81: I'd recommend something `VAST_USE_BUNDLED_CAF` to avoid the double negative (same goes for `VAST_NO_ARROW`, but that's not for this PR ;)

libvast/src/system/indexer_downstream_manager.cpp
```diff
@@ -56,6 +56,33 @@ size_t indexer_downstream_manager::buffered(caf::stream_slot slot) const
...
+void indexer_downstream_manager::close() {
+  VAST_DEBUG(self(), "closing downstream manager");
+  if (closing)
+    return;
+  // At this point, we unregister all partitions by...
+  for (auto it = partitions.begin(); it != partitions.end();) {
+    VAST_ASSERT(*it != nullptr);
+    if (buffered(**it) == 0u) {
+      // ... either removing them directly if the buffers are empty,
+      // meaning all table slices have been forwarded to the indexers,...
```
lava, line 68: Is this part an optimization, or necessary for correctness? I was wondering why we need to distinguish between empty/non-empty here and in `unregister_partition()`, but not in `register_partition()` below.

CHANGELOG.md
tobim: Technically true, but that is just luck, the memory location could have been given back to the OS. Also I don't want to get too technical in the changelog.

libvast/src/system/indexer_downstream_manager.cpp
tobim: I believe both instances are purely defensive, I don't want to rely on `emit_batches_impl()` to clean up partitions that are already done from its point of view (although it currently does).

PR #898, opened 2020-06-04
Support /etc/vast/vast.conf as global config

dominiklohmann:
The configuration file is now looked at in the following order:
- Specified via `--config=path/to/vast.conf`
- `vast.conf` in current working directory
- `${INSTALL_PREFIX}/etc/vast/vast.conf`
- `/etc/vast/vast.conf`


dominiklohmann:
I have confirmed this to work in a Debian docker image.

PR #900, opened 2020-06-05
Error if db-directory cannot be created or is not writable

dominiklohmann:


libvast/src/system/spawn_node.cpp
```diff
@@ -36,6 +37,13 @@ spawn_node(caf::scoped_actor& self, const caf::settings& opts) {
   auto db_dir
     = get_or(opts, "system.db-directory", defaults::system::db_directory);
   auto abs_dir = path{db_dir}.complete();
+  if (!exists(abs_dir))
+    if (auto res = mkdir(abs_dir); !res)
+      return make_error(ec::filesystem_error,
+                        "unable to create db-directory:", abs_dir.str());
+  if (::access(abs_dir.str().c_str(), W_OK) != 0)
```
lava, line 44: It might make sense to bundle the fs-specific functions in the existing `filesystem.hpp`, and e.g. add a `path::is_writable()` or `vast::writable()` function.

dominiklohmann: Good idea.

dominiklohmann:
@lava I have re-requested your review after integrating your feedback.

PR #905, opened 2020-06-08
Allow for pivoting over community_id by default

dominiklohmann:
Note that I have not tested this to work yet, as I need to gather appropriate data and schemas first. I will do so before merging this.


dominiklohmann:
From initial testing, this seems to do the job just fine. However, I'm not quite sure if I can replicate @0ortmann's setup, since I have no simple way of getting the matching Suricata and winlogbeat+Sysmon data.

PR #906, opened 2020-06-08
Remove segment store status that timed out regularly

dominiklohmann:
Even when the archive status report timed out, the node kept running this non-interruptible loop. For large archives, this causes the user to experience a hang, even after the status command returned successfully.


tobim:
[ch14725]

PR #907, opened 2020-06-08
Fix container-overflow reported by ASAN

dominiklohmann:


dominiklohmann:
[ch17387]

CMakeLists.txt
```diff
@@ -444,10 +439,7 @@ if (NOT HAVE_CXX_ATOMICS_WITHOUT_LIB)
   endif ()
 endif ()
 
-# Only set our flags if the environment variable CXXFLAGS is not present.
-if (NOT CMAKE_CXX_FLAGS)
-  set(CMAKE_CXX_FLAGS ${EXTRA_FLAGS})
-endif ()
+set(CMAKE_CXX_FLAGS ${EXTRA_FLAGS})
```
lava, line 442: Don't we need to include the old flags as well?
```
set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${EXTRA_FLAGS}")
```

Also, I think we should *prepend* the extra flags so they're still over-ridable by the user.

dominiklohmann: Good catch.

PR #908, opened 2020-06-08
Simplify id space management

tobim:


tobim:
[ch15456]

libvast/src/system/importer.cpp
```diff
@@ -44,79 +44,41 @@ importer_state::~importer_state() {
 }
 
 caf::error importer_state::read_state() {
-  VAST_TRACE("");
-  id_generators.clear();
-  auto file = dir / "available_ids";
+  auto file = dir / "next_id";
   if (exists(file)) {
-    VAST_DEBUG(self, "reads persistent state from", to_string(file));
+    VAST_VERBOSE(self, "reads persistent state from", to_string(file));
```
dominiklohmann, line 51:
```suggestion
    VAST_VERBOSE(self, "reads persistent state from", file);
```

I have not tested this locally, but `to_string` should already be called implicitly here. Same goes for the other two log messages in this function.

libvast/src/system/importer.cpp
```diff
@@ -44,79 +44,41 @@ importer_state::~importer_state() {
...
       return std::move(result.error());
   }
-  std::ofstream available{to_string(dir / "available_ids")};
-  auto i = id_generators.begin();
-  available << i->i << ' ' << i->last;
-  for (++i; i != id_generators.end(); ++i) {
-    available << '\n' << i->i << ' ' << i->last;
-  }
-  VAST_DEBUG(self, "saved", available_ids(), "available IDs");
+  std::ofstream available{to_string(dir / "next_id")};
```
dominiklohmann, line 69: Can you add a `dir.is_writable()` check above this line?

libvast/src/system/importer.cpp
```diff
@@ -179,34 +179,31 @@ behavior importer(importer_actor* self, path dir, archive_type archive,
...
-    [=](status_atom) {
-      return self->state.status();
-    },
+    [=](status_atom) { return self->state.status(); },
     [=](telemetry_atom) {
       self->state.send_report();
       self->delayed_send(self, defs::telemetry_rate, telemetry_atom::value);
-    }
-  };
+    }};
```
dominiklohmann, line 206:
```suggestion
    },};
```

Trailing comma looks better after clang-format.

dominiklohmann:
I think we should remove the consensus module from the repo if it's essentially dead code now.

libvast/src/system/importer.cpp
tobim: I'm against that. `dir` should always be a child of a `vast.db` folder, and that check should only be done for the parent.

tobim:
@dominiklohmann do you mind taking another look at the latest commit. (The rest is only rebase.)

dominiklohmann:
Looks fine from a taking a look at the code only. I'll check this out and run some tests locall.

libvast/src/system/importer.cpp
dominiklohmann: The `vast.db` may already exist or be created by a system service, and the user may be using the wrong `umask`. I don't think we can make this assumption.

tobim:
> I think we should revert #906 as part of this PR. It's useful for monitoring whether this change works.

I'd prefer not to, as the algorithm to build the JSON value is still extremely inefficient. (Will give problems with large archives regardless.)

> I've noticed a crash when setting table-slice-size to 0, which should probably get a nicer-looking error message instead.

Will do.

PR #910, opened 2020-06-09
Improve handling of UTF-8 input

dominiklohmann:


libvast/vast/concept/parseable/string/quoted_string.hpp
```diff
@@ -13,13 +13,14 @@
 
 #pragma once
 
-#include <cassert>
-#include <string>
-
 #include "vast/concept/parseable/core.hpp"
+#include "vast/concept/parseable/string/any.hpp"
```
tobim, line 17: I don't see that used directly here. Leftover?

dominiklohmann: Good catch. I removed the line.

PR #916, opened 2020-06-11
Forcefully emit batches on input timeout error

dominiklohmann:
Note: This does not fix the underlying issue completely, but is a decent workaround that improves the latency between `source` and `importer`.

### DIFF
--- a/libvast/src/format/csv.cpp
+++ b/libvast/src/format/csv.cpp
@@ -430,7 +430,7 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
   for (size_t produced = 0; produced < max_events; timeout = next_line()) {
     if (timeout) {
       VAST_DEBUG(this, "reached input timeout at line", lines_->line_number());
-      return finish(callback);
+      return finish(callback, ec::input_timeout);
     }
     // EOF check.
     if (lines_->done())
--- a/libvast/src/format/csv.cpp
+++ b/libvast/src/format/csv.cpp
@@ -15,6 +15,7 @@
 
 #include "vast/concept/parseable/core.hpp"
 #include "vast/concept/parseable/string.hpp"
+#include "vast/concept/parseable/to.hpp"
 #include "vast/concept/parseable/vast.hpp"
 #include "vast/concept/printable/to_string.hpp"
 #include "vast/concept/printable/vast/type.hpp"
@@ -151,6 +152,11 @@ reader::reader(caf::atom_value table_slice_type, const caf::settings& options,
                               defaults::set_separator);
   opt_.kvp_separator = get_or(options, "import.csv.kvp_separator",
                               defaults::kvp_separator);
+  if (auto read_timeout_arg = caf::get_if<std::string>(&options, "import.read-"
+                                                                 "timeout")) {
+    if (auto read_timeout = to<vast::duration>(*read_timeout_arg))
+      read_timeout_ = *read_timeout;
+  }
 }
 
 void reader::reset(std::unique_ptr<std::istream> in) {
@@ -416,7 +422,7 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
       return false;
     } else {
       return lines_->next_timeout(
-        vast::defaults::import::shared::partial_slice_read_timeout);
+        std::chrono::duration_cast<std::chrono::milliseconds>(read_timeout_));
     }
   };
   if (!parser_) {
@@ -430,7 +436,7 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
   for (size_t produced = 0; produced < max_events; timeout = next_line()) {
     if (timeout) {
       VAST_DEBUG(this, "reached input timeout at line", lines_->line_number());
-      return finish(callback, ec::input_timeout);
+      return finish(callback, ec::timeout);
     }
     // EOF check.
     if (lines_->done())
--- a/libvast/src/format/csv.cpp
+++ b/libvast/src/format/csv.cpp
@@ -156,6 +156,9 @@ reader::reader(caf::atom_value table_slice_type, const caf::settings& options,
                                                                  "timeout")) {
     if (auto read_timeout = to<vast::duration>(*read_timeout_arg))
       read_timeout_ = *read_timeout;
+    else
+      VAST_WARNING(this, "cannot set read-timeout to", *read_timeout_arg,
+                   "as it is not a valid duration");
   }
 }
 
@@ -416,14 +419,15 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
                              consumer& callback) {
   VAST_ASSERT(max_events > 0);
   VAST_ASSERT(max_slice_size > 0);
-  auto next_line = [&] {
+  auto next_line = [&, start = std::chrono::steady_clock::now()] {
+    auto remaining = start + read_timeout_ - std::chrono::steady_clock::now();
+    if (remaining < std::chrono::steady_clock::duration::zero())
+      return true;
     if (!builder_ || builder_->rows() == 0) {
       lines_->next();
       return false;
-    } else {
-      return lines_->next_timeout(
-        std::chrono::duration_cast<std::chrono::milliseconds>(read_timeout_));
     }
+    return lines_->next_timeout(remaining);
   };
   if (!parser_) {
     auto p = read_header(lines_->get());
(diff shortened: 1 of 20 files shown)


dominiklohmann:
[ch17453]

libvast/vast/system/source.hpp
```diff
@@ -225,47 +226,61 @@ source(caf::stateful_actor<source_state<Reader>>* self, Reader reader,
...
         self->quit();
-      };
-      if (st.remaining && *st.remaining == 0)
-        return finish();
+      }
       if (err != caf::none) {
-        if (err != vast::ec::end_of_input)
-          VAST_INFO(self, "completed with message:", render(err));
-        return finish();
+        if (err == vast::ec::input_timeout) {
```
mavam, line 267: Why not call the error code `timeout` so that it can be used more generally?

mavam: Thanks for the tweak 👍

libvast/vast/system/source.hpp
```diff
@@ -231,26 +233,35 @@ source(caf::stateful_actor<source_state<Reader>>* self, Reader reader,
     },
     // get next element
     [=](bool& done, caf::downstream<table_slice_ptr>& out, size_t num) {
+      using namespace std::chrono_literals;
```
dominiklohmann, line 236:
```suggestion
```

libvast/src/format/csv.cpp
```diff
@@ -416,7 +422,7 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
       return false;
     } else {
       return lines_->next_timeout(
-        vast::defaults::import::shared::partial_slice_read_timeout);
+        std::chrono::duration_cast<std::chrono::milliseconds>(read_timeout_));
```
lava, line 425: Using `read_timeout_` here means that we can wait up to **double** the time specified as read_timeout until we actually get the data, which at 10s is a noticable difference.

Also, we should keep in mind that using such a large timeout will increase input lag and it willl increase the likelihood of seeing "double lines" when importing over a socket, that is the last line of some connection w/o newline concatenated with the first line of the next connection. I'm not sure if this is better or worse in the real world than the current behaviour, but we should monitor it and be prepared to separate the two timeouts if it proves to be a problem (i.e. use a short timeout for the line reading and a large one for the `start + read_timeout` check below)

libvast/src/format/pcap.cpp
```diff
@@ -84,6 +86,11 @@ reader::reader(caf::atom_value id, const caf::settings& options,
     = community_id_ ? pcap_packet_type_community_id : pcap_packet_type;
   last_stats_ = {};
   discard_count_ = 0;
+  if (auto read_timeout_arg = caf::get_if<std::string>(&options, "import.read-"
+                                                                 "timeout")) {
+    if (auto read_timeout = to<vast::duration>(*read_timeout_arg))
```
lava, line 91: We should print an error if a read timeout was specified and cannot be parsed instead of silently ignoring it.

libvast/src/format/syslog.cpp
```diff
@@ -103,13 +110,14 @@ reader::read_impl(size_t max_events, size_t max_slice_size, consumer& f) {
       return false;
     } else {
       return lines_->next_timeout(
-        vast::defaults::import::shared::partial_slice_read_timeout);
+        std::chrono::duration_cast<std::chrono::milliseconds>(read_timeout_));
```
lava, line 113: We can get rid of all these duration_casts by making `read_timeout_` be `std::chrono::milliseconds` instead.

(I didn't verify this, but in theory casting from nanoseconds to milliseconds should involve a division by 10000, so it's probably not free to do this cast)

libvast/vast/format/json.hpp
```diff
@@ -147,6 +152,8 @@ class reader final : public multi_layout_reader {
   mutable size_t num_invalid_lines_ = 0;
   mutable size_t num_unknown_layouts_ = 0;
   mutable size_t num_lines_ = 0;
+  vast::duration read_timeout_
```
tobim, line 155: To me this is attribute of every reader independent of the format, so we should add it there. If you do this, please leave the parsing from the options where it is and only pass the `duration` to the base class.

libvast/src/format/csv.cpp
dominiklohmann: Where do you see the doubling up?

libvast/src/format/pcap.cpp
dominiklohmann: Fair enough.

libvast/src/format/syslog.cpp
dominiklohmann: I'd rather change the interface of `next_timeout` to work with every kind of clock as would be expected.

libvast/vast/system/source.hpp
```diff
@@ -316,7 +336,7 @@ source(caf::stateful_actor<source_state<Reader>>* self, Reader reader,
     },
     [=](telemetry_atom) {
       auto& st = self->state;
-      st.send_report();
+      // st.send_report();
```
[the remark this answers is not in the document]
dominiklohmann: You're reviewing an outdated commit here.

libvast/src/format/csv.cpp
tobim: I totally overlooked this :eyes:.

libvast/vast/system/source.hpp
```diff
@@ -178,7 +179,8 @@ struct source_state {
   void send_report() {
     // Send the reader-specific status report to the accountant.
     if (auto status = reader.status(); !status.empty() && accountant)
-      self->send(accountant, std::move(status));
+      if (accountant)
+        self->send(accountant, std::move(status));
```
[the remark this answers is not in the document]
dominiklohmann: It is consistent with our other uses of `send`, though. I'd argue that we can always create a story for removing the `if (handle)` checks before `self->send(handle, ...)` altogether.

libvast/src/format/syslog.cpp
lava: It's not the clock that is the problem, but the duration that the passed number represents. At some point `line_range` has to pass an integer representing milliseconds to `poll()`, so if you pass in something different there is a cost that needs to paid for converting.

libvast/src/format/csv.cpp
dominiklohmann: We talked about this via Slack and it's really a matter of documentation. I'll add a note saying that it can be doubled up in the very worst case.

libvast/src/format/syslog.cpp
dominiklohmann: It doesn't really matter where we do the conversion: It's either the result of `steady_clock::now()` or the parameter for `next_timeout`.

I've added an overload to `next_timeout` that does the conversion for us so that we can use it with every `std::chrono::duration<Rep, Period>`.

libvast/src/format/csv.cpp
dominiklohmann: The read timeout is now dynamically updated: `read_timeout_ - (now - start)`. This appears to work well in practice.

libvast/vast/format/json.hpp
```diff
@@ -223,18 +224,18 @@ caf::error reader<Selector>::read_impl(size_t max_events, size_t max_slice_size,
   size_t produced = 0;
   table_slice_builder_ptr bptr = nullptr;
   bool timeout = false;
-  auto next_line = [&] {
+  auto next_line = [&, start = std::chrono::steady_clock::now()] {
+    auto now = std::chrono::steady_clock::now();
+    if (start + read_timeout_ < now)
```
tobim, line 229:
```suggestion
    auto remaining_time = start + read_timeout_ - now;
    if (remaining_time < 0)
```

libvast/vast/format/json.hpp
```diff
@@ -223,18 +224,18 @@ caf::error reader<Selector>::read_impl(size_t max_events, size_t max_slice_size,
...
+      return true;
     if (!bptr || bptr->rows() == 0) {
       lines_->next();
       return false;
-    } else {
-      return lines_->next_timeout(
-        std::chrono::duration_cast<std::chrono::milliseconds>(read_timeout_));
     }
+    return lines_->next_timeout(read_timeout_ - (now - start));
```
tobim, line 235:
```suggestion
    return lines_->next_timeout(remaining_time);
```

libvast/vast/format/json.hpp
dominiklohmann: The `rhs` has to be `steady_clock::duration::zero()` for this. Made the change.

PR #918, opened 2020-06-11
Move measurement code into indexers

lava:
Make each indexer submit its own measurements to the accountant, as opposed to passing around pointers to atomic measurements to create a centralized report from the index.

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -392,19 +392,6 @@ caf::behavior index(caf::stateful_actor<index_state>* self, const path& dir,
   self->set_exit_handler([=](const caf::exit_msg& msg) {
     VAST_DEBUG(self, "received exit from", msg.source,
                "with reason:", msg.reason);
-    auto& st = self->state;
-    if (!st.unpersisted.empty() || st.active_partition_indexers > 0) {
-      auto delay = defaults::index::shutdown_retry_interval;
-      VAST_INFO(self, "delaying exit by", delay,
-                "to wait for outstanding indexers due to",
-                st.unpersisted.size(), "unpersisted and ",
-                st.active_partition_indexers, "active");
-      // Tell the upstreams that we are stopping.
-      st.stage->shutdown();
-      st.stage->out().close();
-      self->delayed_send(self, delay, msg);
-      return;
-    }
     self->quit(msg.reason);
   });
   // Launch workers for resolving queries.
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -231,48 +231,6 @@ caf::dictionary<caf::config_value> index_state::status() const {
   return result;
 }
 
-void index_state::send_report() {
-  performance_report r;
-  measurement min, max;
-  auto min_rate = std::numeric_limits<double>::infinity();
-  auto max_rate = -std::numeric_limits<double>::infinity();
-  auto append_report = [&](partition& p) {
-    for (size_t i = 0; i < p.measurements_.size(); ++i) {
-      auto tmp = collect(p.measurements_[i]);
-      if (tmp.events > 0) {
-        r.push_back({as_vector(p.indexers_)[i].first.fqn(), tmp});
-        double rate = tmp.events * 1'000'000'000.0 / tmp.duration.count();
-        if (rate < min_rate) {
-          min_rate = rate;
-          min = tmp;
-        }
-        if (rate > max_rate) {
-          max_rate = rate;
-          max = tmp;
-        }
-      }
-    }
-  };
-  if (active)
-    append_report(*active);
-  for (auto& p : unpersisted)
-    append_report(*p.first);
-  if (min.events > 0) {
-    VAST_VERBOSE(self, "handled", min.events, "events at a minimum rate of",
-                 static_cast<uint64_t>(min_rate), "events/sec in",
-                 to_string(min.duration));
-    r.push_back({"index.min", min});
-  }
-  if (max.events > 0) {
-    VAST_DEBUG(self, "handled", max.events, "events at a maximum rate of",
-               static_cast<uint64_t>(max_rate), "events/sec in",
-               to_string(max.duration));
-    r.push_back({"index.max", max});
-  }
-  if (!r.empty())
-    self->send(accountant, std::move(r));
-}
-
 void index_state::reset_active_partition() {
   // Persist meta data and the state of all INDEXER actors when the active
   // partition gets replaced becomes full.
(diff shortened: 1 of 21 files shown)


libvast/src/system/partition.cpp
```diff
@@ -55,6 +55,10 @@ partition::partition(index_state* state, uuid id, size_t max_capacity)
 }
 
 partition::~partition() noexcept {
+  for (auto kv : indexers_) {
+    auto actor = kv.second.indexer;
+    state_->self->send(actor, stop_atom::value);
```
lava, line 60: I was wondering if we could just send `shutdown_atom` here and get rid of the extra `stop_atom`, but I wasnt completely sure if it is guaranteed that no lookups are still in progress when the partition is destroyed here.

libvast/src/system/indexer.cpp
```diff
@@ -106,6 +130,17 @@ indexer(caf::stateful_actor<indexer_state>* self, path filename,
...
+      // is removed as it will keep a reference to this indexer alive;
+      // thus the actor will never quit and block the shutdown of caf.
+      if (!self->state.stop_telemetry)
+        self->delayed_send(self, defaults::system::telemetry_rate,
+                           telemetry_atom::value);
+    },
+    [=](stop_atom) { self->state.stop_telemetry = true; },
     [=](shutdown_atom) {
       self->quit(caf::exit_reason::user_shutdown); // clang-format fix
     },
```
lava, line 146: I noticed that `shutdown_atom` is currently unused, not sure whether we should remove it completely or keep it just in case.

libvast/src/system/partition.cpp
tobim: We might not need to fiddle with the shutdown logic at all if we reduce the scope of the telemetry to the streaming sink. Would it be possible to start the telemetry heartbeat only when `make_sink` is called and then shut it down in the stop handler?

libvast/vast/system/indexer.hpp
```diff
@@ -13,18 +13,20 @@
...
 #include "vast/system/fwd.hpp"
+#include "vast/system/instrumentation.hpp"
 #include "vast/type.hpp"
 #include "vast/uuid.hpp"
 
+#include <caf/actor.hpp>
+#include <caf/event_based_actor.hpp>
+#include <caf/stateful_actor.hpp>
+
+#include <unordered_map>
```
tobim, line 28:
```suggestion
#include <string>
```

libvast/src/system/partition.cpp
lava: I think that's what I originally tried and it caused some trouble, but I'll have to try again because I can't remember what exactly the problem was (and if it was maybe caused by something else)

lava: It worked now, so it looks like the problem was something else.

libvast/src/system/indexer.cpp
```diff
@@ -38,32 +39,50 @@ indexer_state::~indexer_state() {
...
+caf::error
+indexer_state::init(caf::event_based_actor* self, path filename,
+                    type column_type, caf::settings index_opts,
+                    caf::actor index, uuid partition_id, std::string fqn) {
   this->index = std::move(index);
   this->partition_id = partition_id;
-  this->measurement = m;
+  this->fqn = fqn;
+  this->self = self;
+  this->streaming_done = false;
```
tobim, line 50: I think we should only enable this when the index is currently under construction.
```suggestion
```

libvast/src/system/indexer.cpp
```diff
@@ -84,14 +103,21 @@ indexer(caf::stateful_actor<indexer_state>* self, path filename,
     [=](caf::stream<table_slice_column> in) {
       self->make_sink(
         in,
-        [](caf::unit_t&) {
-          // nop
+        [=](caf::unit_t&) {
+          // Assume that exactly one stream is created for each indexer.
+          auto& st = self->state;
```
tobim, line 108:
```suggestion
          auto& st = self->state;
          st.building = true;
```

libvast/src/system/indexer.cpp
```diff
@@ -84,14 +103,21 @@ indexer(caf::stateful_actor<indexer_state>* self, path filename,
...
+            self->delayed_send(self, defaults::system::telemetry_rate,
+                               atom::telemetry_v);
+          }
         },
         [=](caf::unit_t&, const std::vector<table_slice_column>& xs) {
           handle_batch(xs);
         },
         [=](caf::unit_t&, const error& err) {
           auto& st = self->state;
+          st.streaming_done = true;
```
tobim, line 120:
```suggestion
          st.building = false;
```

libvast/src/system/indexer.cpp
```diff
@@ -105,6 +131,16 @@ indexer(caf::stateful_actor<indexer_state>* self, path filename,
     [=](const std::vector<table_slice_column>& xs) {
       handle_batch(xs); // clang-format fix
     },
+    [=](atom::telemetry) {
+      self->state.send_report();
+      // The indexers are relying on caf's reference counting to shut down, so
+      // we stop telemetry data once the table slice stream is finished.
+      // Otherwise this loop would keep a reference to this indexer alive;
+      // preventing the actor from quitting and blocking the shutdown of caf.
+      if (!self->state.streaming_done)
```
tobim, line 140:
```suggestion
      if (self->state.building)
```

libvast/vast/system/indexer.hpp
```diff
@@ -34,19 +36,31 @@ struct indexer_state {
...
+  caf::event_based_actor* self;
+
   uuid partition_id;
 
-  atomic_measurement* measurement;
+  std::string fqn;
+
+  measurement m;
+
+  bool streaming_done;
```
tobim, line 63:
```suggestion
  bool building = false;
```

lava:
@tobim Finally got this to work:

```
tenzir=> SELECT data->'node'->'time_user_seconds', data->'version'->'VAST' FROM benchmark_results WHERE data->>'name' = 'zeek.bench' ORDER BY ingested_at desc LIMIT 5;
 ?column? |               ?column?                
----------+---------------------------------------
 2111.95  | "2020.05.28-0.indexer-measurements.0"
 2085.89  | "2020.05.28-174-g04a45409"
 2070.99  | "2020.05.28-170-gd443a2ef"
 2073.92  | "2020.05.28-170-gd443a2ef"
 2090.99  | "2020.05.28-170-gd443a2ef"
(5 rows)
```

```
tenzir=> SELECT data->'node'->'time_user_seconds', data->'version'->'VAST' FROM benchmark_results WHERE data->>'name' = 'suricata.bench' ORDER BY ingested_at desc LIMIT 5;
 ?column? |               ?column?                
----------+---------------------------------------
 28.26    | "2020.05.28-0.indexer-measurements.0"
 28.43    | "2020.05.28-174-g04a45409"
 28.71    | "2020.05.28-170-gd443a2ef"
 28.3     | "2020.05.28-170-gd443a2ef"
 28.18    | "2020.05.28-170-gd443a2ef"
(5 rows)
```

Of course, now we can finally consider the *actual* questions: Do we expect current benchmarks to show a slowdown for this change or do we need a new scenario; and do we consider the +30 seconds in the zeek benchmark to be statistically significant?

libvast/src/system/index.cpp
```diff
@@ -393,18 +393,9 @@ caf::behavior index(caf::stateful_actor<index_state>* self, const path& dir,
...
-                st.active_partition_indexers, "active");
-      // Tell the upstreams that we are stopping.
-      st.stage->shutdown();
-      st.stage->out().close();
-      self->delayed_send(self, delay, msg);
-      return;
-    }
+    // Tell the upstreams that we are stopping.
+    st.stage->shutdown();
+    st.stage->out().close();
```
tobim, line 398: I think even this might not be necessary any more (RAII should work here). I'll test that.

tobim: According to my tests it is save to remove.

lava: Removed.

PR #920, opened 2020-06-15
Emit batches only when events were produced

dominiklohmann:


dominiklohmann:
[ch17523]

libvast/vast/system/source.hpp
```diff
@@ -265,8 +265,16 @@ source(caf::stateful_actor<source_state<Reader>>* self, Reader reader,
...
-          return force_emit_batches();
+          if (produced > 0) {
+            VAST_DEBUG(self, "hit input timeout and forcefully emits", produced,
+                       "produced events");
+            return force_emit_batches();
+          } else {
+            // This case should never happen. If it does, we hit an internal
+            // application logic error in the reader. CAF might stall out on us
+            // here.
+            VAST_ERROR(self, "hit input timeout, but produced no events");
```
mavam, line 276: If it's an application logic bug, it sounds like an assertion is warranted here, no? At least the source should probably terminate itself.

dominiklohmann: An assertion can lead to data loss here, but I agree that quitting is a good idea.

PR #921, opened 2020-06-15
Allow output format selection for the pivot/explore command

dominiklohmann:


dominiklohmann:
> The `matcher` might need the same treatment when this is bumped to vast-pro.

That's a good point, thanks!

PR #924, opened 2020-06-16
Treat warnings as errors in CI

dominiklohmann:
This enables `-Werror` for all warnings except `-Wdeprecated` (Clang) / `-Wdeprecated-declarations` (GCC)  in CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,7 +286,7 @@ endif ()
 if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
   check_compiler_version(${GCC_MINIMUM_VERSION})
   # Too many false positives.
-  set(EXTRA_FLAGS "${EXTRA_FLAGS} -Wno-maybe-uninitialized")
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -286,7 +286,7 @@ endif ()
 if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
   check_compiler_version(${GCC_MINIMUM_VERSION})
   # Too many false positives.
-  set(EXTRA_FLAGS "${EXTRA_FLAGS} -Wno-maybe-uninitialized")
+  set(EXTRA_FLAGS "${EXTRA_FLAGS} -Wno-maybe-uninitialized -Wno-redundant-move")
```
lava, line 289: Is the comment still accurate? (i.e. are the redundant move warnings actually false positives, or are we disabling them because they're too many to fix right now?)

dominiklohmann: They're really hard to fix in code using templates and can be considered false positives.

E.g., consider this code:

```cpp
template <class T>
auto f(T x) {
  return std::move(x);
}

int main() {
  // does not trigger the warning, moving is not redundant
  f("123"s);
  // triggers the warning, moving is redundant
  f(123);
}
```

PR #928, opened 2020-06-17
Fix build with gperftools

dominiklohmann:
This was an oversight in #927.


dominiklohmann:
This has been cherry-picked already and does no longer need to be merged.

PR #929, opened 2020-06-17
Support aging out data based on a query

dominiklohmann:
Note: For now, this only supports aging out data from the archive, not from the index.

An aging cycle can be manually triggered by running:

```sh
vast send eraser run
```

The newly added options `system.aging-frequency` and `system.aging-query` can be used for configuring the eraser component.


CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🎁 The new options `system.aging-frequency` (default: 24 hours) and
```
mavam, line 14: Nit: I would remove the parameterization from the changelog and focus rather on the description of the mechanism.

dominiklohmann: As a user, I really like seeing new options listed in the changelog directly. I'll prepend another sentence explaining the feature itself more generally.

dominiklohmann: I've improved the wording. Might taking a look again?

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🎁 VAST can now remove existing data from the archive. The new options
+  `system.aging-frequency` (default: 24 hours) and `system.aging-query` allow
+  for configuring a query that runs on a regular schedule to determine which
+  events to age out. An aging cycle can also be manually triggered by running
+  `vast send eraser run`. [#929](https://github.com/tenzir/vast/929)
+
```
mavam, line 19: - 🧬 VAST now supports aging out existing data. This feature currently only
  concerns data in the archive. The options `system.aging-frequency` and
  `system.aging-query` configure a query that runs on a regular schedule to
  determine which events to delete. It is also possible to trigger a
  an aging cycle manually.
  [#929](https://github.com/tenzir/vast/929)

mavam: This is how I would describe the feature. First, it's incomplete and thus should be marked as experimental. Second, I think the usage details should go into the documentation. (BTW, does it need updating?) The reason is that we probably will change the interface to trigger an aging cycles manually. Defaults may likewise change. The PR contains the example config changes, where the user can lookup what's going on. Since the API will likely change, I don't think it's useful to any user to go through a changelog that contains a bunch outdated usage details. Hence the focus on accurate description of the mechanism, not the parameterization.

dominiklohmann: The documentation for this lives outside of this repository, and yes, it does need updating. That can only happen after this change is merged.

I agree on the experimental feature part. I'll update the changelog accordingly in a bit.

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+ - 🧬 VAST now supports aging out existing data. This feature currently only
+   concerns data in the archive. The options `system.aging-frequency` and
+   `system.aging-query` configure a query that runs on a regular schedule to
+   determine which events to delete. It is also possible to trigger a an aging
```
```suggestion
   determine which events to delete. It is also possible to trigger an aging
```

PR #930, opened 2020-06-17
Improvements to VAST logging framework

lava:


libvast/vast/logger.hpp
```diff
@@ -45,33 +45,40 @@ void fixup_logger(const system::configuration& cfg);
...
-    VAST_LOG_IMPL(lvl,                                                         \
-                  m1 << m2 << m3 << m4 << m5 << m6 << m7 << m8 << m9 << m10)
+    VAST_LOG_IMPL(lvl, (m1) << (m2) << (m3) << (m4) << (m5) << (m6) << (m7)    \
+                            << (m8) << (m9) << (m10))
 
 #  define VAST_LOG(...) VAST_PP_OVERLOAD(VAST_LOG_, __VA_ARGS__)(__VA_ARGS__)
 
+// Ensures all args are used and syntactically valid, without evaluating them.
+#  define VAST_LOG_DISCARD_ARGS(...)                                           \
+    { [[maybe_unused]] decltype(__VA_ARGS__, nullptr) unused{}; }
```
dominiklohmann, line 80: The current macro has issues with code like this:

```cpp
if (...) 
  VAST_WARNING(...)
else
  ...
```

Because it gets expanded to this:

```cpp
if (...)
  { [[maybe_unused]] decltype(..., nulltr) unused{}; };
else
  ...
```

The semicolon after the braces is causing issues here.

You can fix this by using this instead:

```suggestion
#  define VAST_LOG_DISCARD_ARGS(...)                                           \
    do { [[maybe_unused]] decltype(__VA_ARGS__, 0) unused{}; } while(false)
```

(Only changed from `nullptr` to `0` so it still fits the same line. Really doesn't matter what we have there.)

PR #933, opened 2020-06-18
Merge current master into application state versioning branch

lava:
Recommended review procedure:
1) Only look at the last two commits
2) Convince yourself that merge commit has the latest master and the previous branch head as parents
3) Review merge resolutions in the last commit


configure
```diff
@@ -204,13 +204,11 @@ while [ $# -ne 0 ]; do
...
       append_cache_entry CAF_ROOT_DIR PATH "$optarg"
       ;;
-<<<<<<< HEAD
-    --with-flatbuffers=*)
-      append_cache_entry FLATBUFFERS_ROOT_DIR PATH "$optarg"
-=======
     --with-bundled-caf)
       append_cache_entry VAST_USE_BUNDLED_CAF BOOL yes
->>>>>>> origin/master
+      ;;
```
tobim, line 209: There is one remaining conlict in this file. Essentially the same as here, but in the `usage` string.

lava: Good catch!

PR #935, opened 2020-06-18
Fix unused-parameter warnings for log-level quiet

dominiklohmann:


libvast/src/system/sink_command.cpp
```diff
@@ -139,7 +139,8 @@ sink_command(const invocation& inv, actor_system& sys, caf::actor snk) {
       },
       [&](performance_report report) {
         // Log a set of named measurements.
-        VAST_DEBUG(inv.full_name, "received performance report");
+        VAST_DEBUG(inv.full_name, "received performance report",
+                   VAST_ARG(report));
```
tobim, line 143: That just means that the data will be written twice if the log level is debug or trace. I'd argue we should rather remove this line entirely and add the `[[maybe_unused]]` attribute to the parameter.

I think the same of the other change.

dominiklohmann: It's not like the debug log added any value here anyways, so I opted to remove the debug log message entirely.

PR #937, opened 2020-06-19
Improve schema error handling

dominiklohmann:
- An invalid schema file no longer causes VAST not to start
- The filename of an invalid schema is printed along with the proper error

Example output after breaking the bundled `suricata.schema`:

```
❯ vast start
2020-06-19T11:46:07.612 load_schema !! parse_error "/Users/dominiklohmann/Desktop/Tenzir/vast/build/share/vast/schema/suricata.schema"
2020-06-19T11:46:07.629 VAST node is listening on localhost:42000
```

We're still far off from getting proper error messages for the schema parser itself, as that's a bigger refactoring of parseable, but this should be a good starting point for users when something's wrong with their schemas.

### DIFF
--- a/libvast/src/schema.cpp
+++ b/libvast/src/schema.cpp
@@ -217,12 +217,16 @@ caf::expected<schema> load_schema(const std::vector<path>& schema_paths) {
           case path::regular_file:
           case path::symlink: {
             auto schema = load_schema(f);
-            if (!schema)
-              return schema.error();
+            if (!schema) {
+              VAST_ERROR_ANON(__func__, render(schema.error()), f);
+              continue;
+            }
             if (auto merged = schema::merge(directory_schema, *schema))
               directory_schema = std::move(*merged);
-            else
-              return make_error(ec::format_error, "type clash in schema");
+            else {
+              VAST_ERROR_ANON(__func__, "type clash in schema");
+              continue;
+            }
           }
         }
       }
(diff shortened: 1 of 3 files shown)


libvast/src/schema.cpp
```diff
@@ -217,12 +217,16 @@ caf::expected<schema> load_schema(const std::vector<path>& schema_paths) {
...
+            if (!schema) {
+              VAST_ERROR_ANON(__func__, render(schema.error()), f);
+              continue;
+            }
             if (auto merged = schema::merge(directory_schema, *schema))
               directory_schema = std::move(*merged);
-            else
-              return make_error(ec::format_error, "type clash in schema");
+            else {
+              VAST_ERROR_ANON(__func__, "type clash in schema");
```
tobim, line 227: In this case, we should also inform the user about the name of the type with a duplicate definition as well as the paths to both schema files.

I'm also not sure if it is really sensible to continue in this case.

dominiklohmann: I've done the necessary changes.

The output is now this for type clashes:

```
2020-06-19T12:29:52.205 failed to read schema dirs: !! format_error: ("type clash: cannot accomodate two types with the same name:", "suricata.component.common") "in schema file" "schema/suricata.schema"
```

dominiklohmann: Note that I cannot know which schema the type we clashed with was from, as that is already in the merged schema.

tobim: Let's make a note to consider switching to a 2-phase ingestion process for this.

dominiklohmann: I'll create a new story.

PR #938, opened 2020-06-19
Add a get subcommand to retrieve events from the archive directly

tobim:
Please note that we do not necessarily need to merge this, it is just a tool to simplify the investigation of latency issues of `vast export`. The actual fix will come in a different PR after the issue is identified.


libvast/src/system/get_command.cpp
```diff
@@ -0,0 +1,120 @@
...
+      return Writer{std::move(*out)};
+    } else {
+      return Writer{};
+    }
+  };
+  auto writer = make_writer();
+  if (!writer)
+    return writer.error();
+  // TODO: Sending one id at a time is overly pessimistic. A smarter algorithm
+  // would request all ids at once and reorder the results for printing.
```
dominiklohmann, line 69: While a valid point, the idea behind this command was to figure out possible performance issues with the archive, and I think doing this 1 by 1 is fine for this purpose. Maybe it should just be documented that this command is not optimized for performance.

libvast/src/system/get_command.cpp
```diff
@@ -0,0 +1,120 @@
...
+  VAST_ASSERT(node != nullptr);
+  auto components = get_node_components(self, node, {"archive"});
+  if (!components)
+    return caf::make_message(std::move(components.error()));
+  auto archive = caf::actor_cast<archive_type>((*components)[0]);
+  using namespace std::string_literals;
+  auto output_format = get_or(inv.options, "get.format", "json"s);
+  self->send(archive, atom::exporter_v, self);
+  caf::error err = caf::none;
+  if (output_format == "ascii")
```
dominiklohmann, line 111: Let's talk about handling the format selection generally, as that's required for #921 as well. The major difference here is that for #921 we need to spawn a sink and a writer for the format, while this only needs a writer.

libvast/src/system/get_command.cpp
tobim: It depends, I would argue that many queries generate contiguous id ranges, but the mechanism here would break them apart. This is useful to analyze worst-case situations, but it can't be used to simulate normal workloads.

libvast/src/system/get_command.cpp
```diff
@@ -0,0 +1,120 @@
...
+  auto& node = caf::holds_alternative<caf::actor>(node_opt)
+                 ? caf::get<caf::actor>(node_opt)
+                 : caf::get<scope_linked_actor>(node_opt).get();
+  VAST_ASSERT(node != nullptr);
+  auto components = get_node_components(self, node, {"archive"});
+  if (!components)
+    return caf::make_message(std::move(components.error()));
+  auto archive = caf::actor_cast<archive_type>((*components)[0]);
+  using namespace std::string_literals;
+  auto output_format = get_or(inv.options, "get.format", "json"s);
```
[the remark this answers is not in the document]
tobim: This command is for developers only, the example `vast.conf` is for users.

tobim:
[ch19115]

libvast/src/system/get_command.cpp
tobim: This is now fixed with the writer factory.

dominiklohmann (changes requested):
This is mostly looking fine, I just have minor stuff:

- zeek-to-vast seems to be missing an include (that used to be available transitively).
- please add to vast.yaml.example, rebase onto master, and use `vast.get.format` over `get.format`.

integration/default_set.yaml
```diff
@@ -70,14 +70,15 @@ tests:
...
-      - command: -N export ascii '#type == "zeek.conn" && resp_h == 192.168.1.104'
-      - command: '-N export ascii "#type != \"zeek.conn\" && #type != \"vast.metrics\""'
-      - command: -N export ascii '#type != "foobar" && resp_h == 192.168.1.104'
+      - command: -N --disable-metrics export ascii 'resp_h == 192.168.1.104'
+      - command: -N --disable-metrics export ascii 'orig_bytes > 1k && orig_bytes < 1Ki'
+      - command: -N --disable-metrics export ascii ':string == "OrfTtuI5G4e" || :port == 67/udp'
+      - command: -N --disable-metrics export ascii '#type == "zeek.conn" && resp_h == 192.168.1.104'
+      - command: '-N --disable-metrics export ascii "#type != \"zeek.conn\" && #type != \"vast.metrics\""'
+      - command: -N --disable-metrics export ascii '#type != "foobar" && resp_h == 192.168.1.104'
+      - command: -N --disable-metrics get --format=json 1 42 7777 9999999999 122
```
dominiklohmann, line 81: Can you add a comment above that the reason for the `--disable-metrics` everywhere is that we're testing the get command here?

dominiklohmann: Please also test a second format that is not the default, so we have a test for `vast.get.format`.

libvast/src/system/application.cpp
```diff
@@ -173,6 +174,12 @@ auto make_export_command() {
   return export_;
 }
 
+auto make_get_command() {
+  return std::make_unique<command>(
+    "get", "extracts the events assiciated with ids", "TODO",
```
dominiklohmann, line 179: The TODO must be addressed.

libvast/src/system/application.cpp
```diff
@@ -173,6 +174,12 @@ auto make_export_command() {
...
+    opts("?get").add<std::string>("format,f", "how to format the output"));
```
dominiklohmann, line 180: Can you align the option description and short option with the other format options?

libvast/src/system/explore_command.cpp
```diff
@@ -56,7 +56,7 @@ caf::message explore_command(const invocation& inv, caf::actor_system& sys) {
                   defaults::explore::max_events_query);
   // Get a local actor to interact with `sys`.
   caf::scoped_actor self{sys};
-  auto s = make_sink(sys, inv.options, output_format);
+  auto s = make_sink(sys, output_format, inv.options);
```
dominiklohmann, line 59: Was there a specific reason for switching the argument order of `make_sink`? This is not a dealbreaker, just curious.

libvast/src/system/get_command.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+caf::error
+run(caf::scoped_actor& self, archive_type& archive, const invocation& inv) {
+  using namespace std::string_literals;
+  auto output_format = get_or(inv.options, "vast.get.format", "json"s);
+  auto writer = factory<format::writer>::make(output_format, inv.options);
+  if (!writer)
+    return writer.error();
+  // TODO: Sending one id at a time is overly pessimistic. A smarter algorithm
+  // would request all ids at once and reorder the results for printing.
```
dominiklohmann, line 58: Can you create a follow-up story for this?

libvast/src/system/get_command.cpp
```diff
@@ -0,0 +1,102 @@
...
+    if (!i)
+      return make_error(ec::parse_error, c, "is not a positive integer");
+    self->send(archive, to_ids(*i));
+    bool waiting = true;
+    self->receive_while(waiting)
+      // Message handlers.
+      ([&](table_slice_ptr slice) { (*writer)->write(*slice); },
+       [&](atom::done, const caf::error& err) {
+         if (err)
+           VAST_WARNING_ANON("get_command received an error:", err);
```
dominiklohmann, line 71:
```suggestion
           VAST_WARNING_ANON("failed to get table slice:", render(err));
```

libvast/src/system/get_command.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+ids to_ids(const count& x) {
+  ids result;
+  result.append_bits(false, x);
+  result.append_bit(true);
+  return result;
+}
+
+caf::error
+run(caf::scoped_actor& self, archive_type& archive, const invocation& inv) {
```
dominiklohmann, line 51:
```suggestion
run(caf::scoped_actor& self, archive_type archive, const invocation& inv) {
```

libvast/src/system/get_command.cpp
```diff
@@ -0,0 +1,102 @@
...
+         waiting = false;
+       });
+  }
+  return caf::none;
+}
+
+} // namespace
+
+caf::message
+get_command(const invocation& inv, [[maybe_unused]] caf::actor_system& sys) {
```
dominiklohmann, line 81:
```suggestion
get_command(const invocation& inv, caf::actor_system& sys) {
```

libvast/src/system/get_command.cpp
```diff
@@ -0,0 +1,102 @@
...
+  // Get VAST node.
+  auto node_opt
+    = spawn_or_connect_to_node(self, inv.options, content(sys.config()));
+  if (auto err = caf::get_if<caf::error>(&node_opt))
+    return caf::make_message(std::move(*err));
+  auto& node = caf::holds_alternative<caf::actor>(node_opt)
+                 ? caf::get<caf::actor>(node_opt)
+                 : caf::get<scope_linked_actor>(node_opt).get();
+  VAST_ASSERT(node != nullptr);
+  auto components = get_node_components(self, node, {"archive"});
```
dominiklohmann, line 93: This should use `find_by_label` instead.

libvast/src/system/get_command.cpp
```diff
@@ -0,0 +1,102 @@
...
+  if (!components)
+    return caf::make_message(std::move(components.error()));
+  auto archive = caf::actor_cast<archive_type>((*components)[0]);
```
dominiklohmann, line 96: An actor cast can fail, so you should check `if (archive)` afterwards, or at least `VAST_ASSERT(archive)`.

libvast/src/system/sink.cpp
```diff
@@ -0,0 +1,128 @@
...
+    self->quit(msg.reason);
+  });
+  return {
+    [=](table_slice_ptr slice) {
+      VAST_DEBUG(self, "got:", slice->rows(), "events from",
+                 self->current_sender());
+      auto& st = self->state;
+      auto now = steady_clock::now();
+      auto time_since_flush = now - st.last_flush;
+#if VAST_LOG_LEVEL >= VAST_LOG_LEVEL_INFO
```
dominiklohmann, line 70: This `#if` can be removed, no other computation is happening and the `if (st.processed == 0)` is trivially removed by any compiler.

libvast/src/system/sink.cpp
```diff
@@ -0,0 +1,128 @@
...
+      if (st.processed == 0) {
+        VAST_INFO(st.name, "received first result with a latency of",
+                  to_string(time_since_flush));
+      }
+#endif
+      auto reached_max_events = [&] {
+        VAST_INFO(self, "reached max_events:", st.max_events, "events");
```
dominiklohmann, line 77:
```suggestion
        VAST_INFO(self, "reached limit of", st.max_events, "events");
```

libvast/src/system/sink.cpp
```diff
@@ -0,0 +1,128 @@
...
+        st.last_flush = now;
+        st.send_report();
+      }
+    },
+    [=](atom::limit, uint64_t max) {
+      VAST_DEBUG(self, "caps event export at", max, "events");
+      if (self->state.processed < max)
+        self->state.max_events = max;
+      else
+        VAST_WARNING(self, "ignores new max_events of", max,
```
dominiklohmann, line 112:
```suggestion
        VAST_WARNING(self, "ignores new limit of", max,
```

libvast/vast/format/make_writer.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include <caf/expected.hpp>
+#include <caf/settings.hpp>
+
+#include <memory>
+#include <string>
+#include <type_traits>
+
+namespace vast::format {
+
+template <class Writer>
```
dominiklohmann, line 33: This is missing documentation. A note on why this returns an `expected<unique_ptr<T>>` would be nice-to-have.

dominiklohmann (changes requested):
I've got a bunch of small requests.

Tested this locally and it works as expected.

libvast/src/system/explore_command.cpp
tobim: I changed it so the order of arguments is the same as for `spawn_sink` and `factory<writer>::make`.

libvast/src/system/get_command.cpp
tobim: This is outside of the node, `find_by_label` only words if you have the `node_state`.

dominiklohmann (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🧬 The `vast get` subcommand has been added. It can be used to retrieve
+  events by their ids from the database.
```
dominiklohmann, line 15: 💣 I want you to imagine @mavam kindly telling you to use active voice in changelog entries.

dominiklohmann (approved):
Approving, but please improve the changelog entry.

PR #939, opened 2020-06-19
Remove Gperftools/tcmalloc code

dominiklohmann:


mavam:
You missed a TCMALLOC define in `config.hpp.in`

mavam:
I would remove the changelog entry. It was totally unused and undocumented feature that existed since the early days. Nobody besides me used it 😅.

PR #940, opened 2020-06-19
Use configuration from /etc/vast

0ortmann:
Vast should use the configuration file from `/etc/vast` on default. This PR adds the config to that location and starts `vast` normally for `systemd` / `Docker`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,8 +8,6 @@ ENV CXX g++-8
 ENV BUILD_TYPE Release
 ENV BUILD_DIR /tmp/src
 ENV DEBIAN_FRONTEND noninteractive
-ARG BRANCH
-ENV BRANCH master
 
 # Compiler and dependency setup
 RUN apt-get -qq update && apt-get -qqy install \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -19,9 +17,18 @@ RUN pip3 install --upgrade pip && pip install --upgrade cmake && \
...
+COPY cmake ./cmake
+COPY doc ./doc
+COPY integration ./integration
+COPY libvast ./libvast
+COPY libvast_test ./libvast_test
+COPY schema ./schema
+COPY scripts ./scripts
+COPY tools ./tools
+COPY vast ./vast
+COPY .clang-format .cmake-format BANNER CMakeLists.txt configure vast.conf ./
```
dominiklohmann, line 31: This seems brittle, if we must keep it in sync with the repository layout manually.

0ortmann: That is just the way it is with Docker. The alternative is to clone the repository, which we did before. I moved away from that since it is very error prone wrt Docker caching.

dominiklohmann: Is there really no way to just copy the whole directory, something like `COPY . ./`? Or even just the output of `git ls-files`.

0ortmann: Copying the whole directory invalidates the benefit you gain from selective copies. We could do that but would have to maintain a `.dockerignore` file then, which prevents unnecessary files from copying.

dominiklohmann: I'll post a poll in Slack.

dominiklohmann: Resolving as per discussion in Slack.

0ortmann:
Resolved the conflict with a rebase on current `master`.

PR #943, opened 2020-06-23
Update Apache Arrow API usage for 0.17

dominiklohmann:


libvast/src/arrow_table_slice.cpp
```diff
@@ -146,10 +147,10 @@ caf::error arrow_table_slice::deserialize(caf::deserializer& source) {
     batch_ = nullptr;
     return caf::none;
   }
-  std::shared_ptr<arrow::RecordBatchReader> reader;
-  auto st = arrow::ipc::RecordBatchStreamReader::Open(&input_stream, &reader);
-  if (!st.ok())
+  auto reader_result = arrow::ipc::RecordBatchStreamReader::Open(&input_stream);
```
mavam, line 150: Is the API following RAII principles or do we also need to invoke `RecordBatchStreamReader::Close` somewhere?

dominiklohmann: It's just a factory function for a ref-counted RecordBatchReader. I agree that the naming is unfortunate, but we cannot change that. It's doing RAII as expected.

dominiklohmann:
> This set of changes looks fairly straight-forward. How did you test that it works, besides unit and integration tests?

I did not. A similar set of changes was applied to the Arrow benchmarks, which I looked at for inspiration. If something was wrong with the Arrow table slice, the integration tests and unit tests would most certainly catch that.

mavam:
Alright, seems like it's good to go.

PR #944, opened 2020-06-23
Use ccache in CI

dominiklohmann:


tobim:
This looks promising, let's see how well it works when we have multiple concurrent runs from different PRs.

PR #946, opened 2020-06-23
Add filesystem actor

mavam:
We will use this for all future filesystem I/O.

It also makes supporting other filesystems beyond POSIX straight-forward.


libvast/src/system/posix_file_system.cpp
```diff
@@ -0,0 +1,80 @@
...
+        ++self->state.stats.mmaps.bytes += chk->size();
+        return chk;
+      } else {
+        ++self->state.stats.mmaps.failed;
+        return nullptr;
+      }
+    },
+    [=](atom::status) {
+      caf::dictionary<caf::config_value> result;
+      result["type"] = "POSIX";
```
dominiklohmann, line 64: Why are you not using `caf::put` here, when you're using it everywhere else?

libvast/src/system/posix_file_system.cpp
```diff
@@ -0,0 +1,80 @@
...
+      auto& ops = put_dictionary(result, "operations");
+      auto put = [&](auto& name, auto& stats) {
+        auto& dict = put_dictionary(ops, name);
+        caf::put(dict, "successful", stats.successful);
+        caf::put(dict, "failed", stats.failed);
+        caf::put(dict, "bytes", stats.bytes);
+      };
+      put("writes", self->state.stats.writes);
+      put("reads", self->state.stats.reads);
+      put("mmaps", self->state.stats.mmaps);
```
dominiklohmann, line 74:
```suggestion
      auto put_stats = [&](auto& name, auto& stats) {
        auto& dict = put_dictionary(ops, name);
        caf::put(dict, "successful", stats.successful);
        caf::put(dict, "failed", stats.failed);
        caf::put(dict, "bytes", stats.bytes);
      };
      put_stats("writes", self->state.stats.writes);
      put_stats("reads", self->state.stats.reads);
      put_stats("mmaps", self->state.stats.mmaps);
```

dominiklohmann: The helper just being named `put` really confused me at first.

libvast/vast/system/file_system.hpp
```diff
@@ -0,0 +1,56 @@
...
+    uint64_t failed = 0;
+    uint64_t bytes = 0;
+  };
+
+  ops_statistics writes;
+  ops_statistics reads;
+  ops_statistics mmaps;
+};
+
+// clang-format off
```
dominiklohmann, line 39: Is this really necessary here?

libvast/src/system/posix_file_system.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <caf/settings.hpp>
+
+namespace vast::system {
+
+file_system_type::behavior_type posix_file_system(
+  file_system_type::stateful_pointer<posix_file_system_state> self, path root) {
+  return {
+    [=](atom::write, const path& filename,
+        chunk_ptr chk) -> caf::result<atom::ok> {
+      if (auto err = io::write(root / filename, as_bytes(chk))) {
```
dominiklohmann, line 33: How does `root / filename` interact with absolute paths? Is it a requirement that `filename` is a relative path? This also applies to the other handlers below.

This should go into the comments for the typed actor behavior in the header, so it's clear what kind of `path` is expected here.

dominiklohmann:
Obviously the failing build must be addressed as well.

libvast/src/system/posix_file_system.cpp
mavam: Can I maybe keep the name `put` and use the more idiomatic way of making the assignment, e.g., `dict["successful"] = stats.successful`?

libvast/vast/system/file_system.hpp
mavam: Yes, the typed actor interface looks not readable otherwise.

libvast/src/system/posix_file_system.cpp
mavam: Yes, all operations should always be relative to the filesystem root. I'll add documentation.

libvast/src/system/posix_file_system.cpp
mavam: This is more idiomatic, no? If possible, I'd go with this syntax.

libvast/src/system/posix_file_system.cpp
dominiklohmann: These are orthogonal imo. The helper being named just `put` when `caf::put` is a thing is just confusing for the reader.

libvast/src/system/posix_file_system.cpp
dominiklohmann: Sure, can do.

libvast/src/system/posix_file_system.cpp
dominiklohmann: In that case, can you return an error of the path is outside the filesystem root?

mavam: I think the user should have the assumption that every path in an operation is relative to the filesystem root, but the user shouldn't know where that root is. I've clarified this in the comments.

dominiklohmann: As is this still allows users to escape the filesystem root by specifying ".." as the path. We should guard against that, I think.

mavam: I don't see the use case. This is not about path sanitization from a malicious standpoint. We'd have to do a lot more to get that right. This would also require going through the code base at many more places beyond this one.

Don't get me wrong, it's not a bad idea, it's just out of scope for this PR.

.github/workflows/vast.yml
```diff
@@ -146,6 +146,23 @@ jobs:
...
           cmake --build "$BUILD_DIR" --target test
 
+      # TODO: Dropbear doesnt allow to select the location of `authorized_keys`,
+      # otherwise we'd want to separate all config from the rest of the system.
+      # TODO: Provision the correct host key instead of turning off host key
+      # verification.
+      - name: Spawn Deluxe Reverse Shell on Failure
+        if: failure()
+        env:
+          REMOTE: rshell@portal.tenzir.com
```
0ortmann, line 156: I'd prefer to have this in a GitHub secret as well.

.github/workflows/vast.yml
```diff
@@ -122,6 +111,17 @@ jobs:
           brew --version
           brew install libpcap tcpdump rsync pandoc apache-arrow pkg-config ninja ccache gnu-sed
 
+      - name: Compile flatbuffers
+        if: matrix.os.tag == 'Ubuntu' || matrix.os.tag == 'macOS'
```
dominiklohmann, line 115: Can't you just do `brew install --HEAD flatbuffers` on macOS?

libvast/vast/system/posix_file_system.hpp
```diff
@@ -0,0 +1,42 @@
...
+ * LICENSE file found in the top-level directory of this distribution and at  *
+ * http://vast.io/license. No part of VAST, including this file, may be       *
+ * copied, modified, propagated, or distributed except according to the terms *
+ * contained in the LICENSE file.                                             *
+ ******************************************************************************/
+
+#pragma once
+
+#include "vast/filesystem.hpp"
+#include "vast/system/file_system.hpp"
```
dominiklohmann, line 17: Can we `s/file_system/filesystem` for consistency and proper spelling everywhere? This affects a few files.

.github/workflows/vast.yml
mavam: This is a platform-agnostic way that adds minimal overhead. I went initially down the rabbit hole of trying to get a PPA for Linux, but that didn't exist.

If you think we need to change the GH workflow, let's do this in a separate PR.

dominiklohmann: I think we should include flatbuffers as a submodule if we're not using a stable version, but you're correct that this belongs in a separate PR.

Let's talk about this again when merging the epic branch into master.

mavam:
@dominiklohmann I removed the last vestiges of `file_system`. Should be ready to go now.

mavam:
@dominiklohmann There is no PPA for 1.12, so we need the source build anyway. But adding *additional* logic for flatbuffers scatters the dependency list in my opinion. This way, we have at least one workflow step for *all* platforms.

dominiklohmann:
> There is no PPA for 1.12, so we need the source build anyway. But adding _additional_ logic for flatbuffers scatters the dependency list in my opinion

There is no additional logic. You remove the check for macOS in the flatbuffers build and add flatbuffers to the brew install line. It'll make our CI run faster, so we should do it.

dominiklohmann:
Also, as far as I can tell, flatbuffers is available as a package for the Ubuntu version we have in CI:

https://packages.ubuntu.com/search?keywords=flatbuffers&searchon=names&suite=all&section=all

mavam:
> flatbuffers is available as a package for the Ubuntu version we have in CI

But that's 1.11. Didn't you want 1.12 earlier?

Anyway, if you tell me what packages to add from the link you posted, I'm fine with the tweak. But it's pretty confusing that there is `flatbuffers-compiler`, `flatbuffers-compiler-dev`, `libflattbuffers-dev`, and `libflatbuffers1`.

> You remove the check for macOS in the flatbuffers build and add flatbuffers to the brew install line. It'll make our CI run faster, so we should do it.

Okay, did that.

PR #947, opened 2020-06-24
Improve CI cache restore keys

dominiklohmann:


.github/workflows/vast.yml
```diff
@@ -121,15 +121,17 @@ jobs:
...
 
+      - name: Export Slug Variables
+        uses: rlespinasse/github-slug-action@v2.x
+
       - name: Fetch ccache Cache
         uses: actions/cache@v2
         with:
           path: .ccache
-          key: ccache-${{ matrix.os.tag }}-${{ matrix.os.compiler }}-${{ matrix.configure.tag }}-${{ github.ref }}-${{ github.sha }}
+          key: ccache-${{ matrix.os.tag }}-${{ matrix.os.compiler }}-${{ matrix.configure.tag }}-${{ env.GITHUB_HEAD_REF_SLUG }}-${{ github.sha }}
```
lava, line 131: I'm wondering where the cache contents get uploaded, dont we need to apply the same transformation there as well?

dominiklohmann: Yeah valid point. Looks like I'll have to find another way to get the slugs, as the action doesn't run on macOS.

dominiklohmann: Found a better way to get them, and also separated the cache lookups for the different trigger event types.

.github/workflows/vast.yml
```diff
@@ -133,25 +133,25 @@ jobs:
           echo ::set-env name=GITHUB_HEAD_REF_SLUG::"$(slug_ref "$GITHUB_HEAD_REF")"
           echo ::set-env name=GITHUB_BASE_REF_SLUG::"$(slug_ref "$GITHUB_BASE_REF")"
 
-      - name: Fetch ccache Cache
+      - name: Fetch ccache Cache (pull_request)
+        if: github.event_name == 'pull_request'
         uses: actions/cache@v2
         with:
           path: .ccache
           key: ccache-${{ matrix.os.tag }}-${{ matrix.os.compiler }}-${{ matrix.configure.tag }}-${{ env.GITHUB_HEAD_REF_SLUG }}-${{ github.sha }}
```
tobim, line 141: Sorry for being late to the party:
According to [the documentation](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables): `GITHUB_HEAD_REF` and `GITHUB_BASE_REF` are only set in forked repos?

It would also be great if these new env-variables would be logged to the console here.

dominiklohmann: They are logged to the console implicitly.

dominiklohmann: And "forked repos" basically means they're only available in `pull_request` events.

PR #948, opened 2020-06-24
Rename the 'default' table slice type to 'caf'

dominiklohmann:
The 'default' table slice type has not been the default for a while now when Apache Arrow is available. The name 'caf' more accurately describes what it does.


dominiklohmann:
I have verified this to work locally using `./configure --without-arrow`.

libvast/vast/caf_table_slice.hpp
```diff
@@ -24,15 +24,15 @@
 namespace vast {
 
 /// The default implementation of `table_slice`.
-class default_table_slice : public table_slice {
+class caf_table_slice : public table_slice {
```
tobim, line 27: Please adjust the comment above.

PR #950, opened 2020-06-24
Make the id space management robust against irregular shutdowns 

tobim:


libvast/vast/system/importer.hpp
```diff
@@ -56,15 +56,26 @@ struct importer_state {
 
   void send_report();
 
-  /// Stores the id offset for the next slice.
-  id next_id = 0u;
+  caf::error bump_boundary();
+
+  /// @returns the next unused id and increments the position by its argument.
+  id next_id(uint64_t advance);
```
mavam, line 62: I feel you are mixing two things in this functions that should be separate, because the call site really confuses me.
1. `next_id()` or `current_id()` to return the next available ID.
2. `advance(n)` to move in the ID space.

libvast/src/system/importer.cpp
```diff
@@ -38,21 +38,36 @@ importer_state::importer_state(caf::event_based_actor* self_ptr)
 }
 
 importer_state::~importer_state() {
-  write_state();
+}
+
+caf::error importer_state::bump_boundary() {
+  while (next_id_ >= id_boundary)
+    id_boundary += 10'000'000;
```
mavam, line 45: Please use a power of 2 for this:
```suggestion
    id_boundary += 10Mi;
```

libvast/vast/system/importer.hpp
```diff
@@ -56,15 +56,26 @@ struct importer_state {
...
 
   /// @returns the number of currently available IDs.
   id available_ids() const noexcept;
 
   /// @returns various status metrics.
   caf::dictionary<caf::config_value> status() const;
 
+  /// Stores the id offset for the next slice.
+  id next_id_ = 0u;
```
mavam, line 71: Please make consistent use of member variable suffixes.

libvast/vast/system/importer.hpp
```diff
@@ -56,15 +56,26 @@ struct importer_state {
...
+
+  /// Stores the id block boundary for persisting the id_space.
+  /// When next_id_ reaches this value, the boundary gets bumped and synchronized
+  /// to the state file on disk. When the importer spawns, it reads the value,
+  /// initializes next_id_ with it, and bumps the boundary immediately.
+  id id_boundary;
```
mavam, line 77: I find this hard to understand, even though I should know what this is about. Neither the concept of "block" nor "boundary" are intuitive to me. Can we simplify the terminology?

libvast/vast/system/importer.hpp
tobim: I started out with that separation, but realized that calling `next_id()` twice without `advance()` in between would be a logic error. For that reason I thought it would be better to force these 2 operations into lockstep by means of the API.

mavam: That's a good point. Then the function just needs another name. How about calling it `assign(n)`? This makes it clear we are using `n` IDs an it's easy to communicate/see that the return value is the starting point for that.

libvast/vast/system/importer.hpp
tobim: I hope the updated implementation makes the domain more explicit and understandable.

libvast/src/system/importer.cpp
```diff
@@ -38,44 +39,73 @@ importer_state::importer_state(caf::event_based_actor* self_ptr)
...
+  } else {
+    VAST_VERBOSE(self, "persisted id block boundary at", current.end);
+  }
   return caf::none;
 }
 
+caf::error importer_state::get_next_block() {
+  using namespace si_literals;
+  while (current.next >= current.end)
+    current.end += 8_Mi;
```
mavam, line 87: How about putting this as a `constexpr` parameters in namespace `defaults`?

libvast/src/system/importer.cpp
```diff
@@ -38,44 +39,73 @@ importer_state::importer_state(caf::event_based_actor* self_ptr)
...
   if (exists(file)) {
     VAST_VERBOSE(self, "reads persistent state from", file);
-    std::ifstream available{to_string(file)};
-    available >> next_id;
-    if (!available.eof())
-      VAST_ERROR(self, "got an invalidly formatted persistence file:", file);
+    std::ifstream state_file{to_string(file)};
+    state_file >> current.end;
+    state_file >> current.next;
+    // TODO:
```
mavam, line 52: What is supposed to happen here?

libvast/src/system/importer.cpp
dominiklohmann: That would suggest that it can be changed, which it cannot. It's only used in a single place, so I don't really see a need for obfuscating this value.

dominiklohmann:
Does this need a changelog entry?

tobim:
> Does this need a changelog entry?

No, this is a follow up for #908.

libvast/src/system/importer.cpp
tobim: I agree with @dominiklohmann, we should not make this configurable.

tobim:
This is ready for a re-review.

libvast/src/system/importer.cpp
```diff
@@ -38,44 +39,72 @@ importer_state::importer_state(caf::event_based_actor* self_ptr)
...
+  return max_id - current.next;
 }
 
 caf::dictionary<caf::config_value> importer_state::status() const {
   caf::dictionary<caf::config_value> result;
   // Misc parameters.
   result.emplace("available-ids", available_ids());
-  result.emplace("next-id", next_id);
+  result.emplace("next-id", current.next);
+  result.emplace("block-boundary", current.end);
```
mavam, line 107: This could be wrapped into a dictionary with key `ids`; maybe this `.` notation even works:

```suggestion
  result["ids.available"] = available_ids();
  result["ids.block.next"] = current.next;
  result["ids.block.end"] = current.end;
```

libvast/src/system/indexer.cpp
```diff
@@ -63,13 +63,18 @@ indexer(caf::stateful_actor<indexer_state>* self, path filename,
...
+               "; ending at:", xs.rbegin()->slice->offset());
     auto t = atomic_timer::start(*self->state.measurement);
     auto events = uint64_t{0};
     for (auto& x : xs) {
       events += x.slice->rows();
       self->state.col.add(x);
     }
     t.stop(events);
+    VAST_DEBUG(self, self->state.col.filename().str(), "added", events,
+               "events");
```
mavam, line 77:
```suggestion
```

libvast/src/system/indexer.cpp
```diff
@@ -63,13 +63,18 @@ indexer(caf::stateful_actor<indexer_state>* self, path filename,
     return {};
   }
   auto handle_batch = [=](const std::vector<table_slice_column>& xs) {
+    VAST_DEBUG(self, self->state.col.filename().str(), "got batch with",
+               xs.size(), "slices; starting at", xs.begin()->slice->offset(),
+               "; ending at:", xs.rbegin()->slice->offset());
```
mavam, line 68:
```suggestion
```

libvast/src/system/importer.cpp
```diff
@@ -38,44 +39,73 @@ importer_state::importer_state(caf::event_based_actor* self_ptr)
...
       return std::move(result.error());
   }
-  std::ofstream available{to_string(dir / "next_id")};
-  available << next_id;
-  VAST_VERBOSE(self, "persisted id space caret at", next_id);
+  std::ofstream state_file{to_string(dir / "current_id_block")};
+  state_file << current.end;
+  if (mode == write_mode::with_next) {
+    state_file << " " << current.next;
+    VAST_VERBOSE(self, "persisted id block [", current.next, ",", current.end,
```
mavam, line 75:
```suggestion
    VAST_VERBOSE(self, "persisted ID block [", current.next, ",", current.end,
```

libvast/src/system/importer.cpp
```diff
@@ -38,44 +39,73 @@ importer_state::importer_state(caf::event_based_actor* self_ptr)
...
+                 ")");
+  } else {
+    VAST_VERBOSE(self, "persisted id block boundary at", current.end);
```
mavam, line 78:
```suggestion
    VAST_VERBOSE(self, "persisted ID block boundary at", current.end);
```

libvast/src/system/importer.cpp
```diff
@@ -38,44 +39,73 @@ importer_state::importer_state(caf::event_based_actor* self_ptr)
...
     VAST_VERBOSE(self, "reads persistent state from", file);
-    std::ifstream available{to_string(file)};
-    available >> next_id;
-    if (!available.eof())
-      VAST_ERROR(self, "got an invalidly formatted persistence file:", file);
+    std::ifstream state_file{to_string(file)};
+    state_file >> current.end;
+    state_file >> current.next;
+    if (!state_file) {
+      VAST_WARNING(self, "did not find next id position from the state file; "
```
mavam, line 53:
```suggestion
      VAST_WARNING(self, "did not find next ID position in state file; "
```

libvast/src/system/importer.cpp
```diff
@@ -38,44 +39,73 @@ importer_state::importer_state(caf::event_based_actor* self_ptr)
...
-  auto file = dir / "next_id";
+  auto file = dir / "current_id_block";
   if (exists(file)) {
     VAST_VERBOSE(self, "reads persistent state from", file);
-    std::ifstream available{to_string(file)};
-    available >> next_id;
-    if (!available.eof())
-      VAST_ERROR(self, "got an invalidly formatted persistence file:", file);
+    std::ifstream state_file{to_string(file)};
+    state_file >> current.end;
```
mavam, line 50: Can you check whether this operation succeeded, e.g., by `if (state_file)`? If this fails already, you should do an hard abort.

libvast/src/system/importer.cpp
```diff
@@ -38,44 +39,73 @@ importer_state::importer_state(caf::event_based_actor* self_ptr)
...
+    state_file >> current.next;
+    if (!state_file) {
+      VAST_WARNING(self, "did not find next id position from the state file; "
+                         "detected an irregular shutdown");
```
mavam, line 54:
```suggestion
                         "irregular shutdown detected");
```

libvast/src/system/indexer.cpp
tobim: Both of these statements have been removed from the change that originally introduced them.

libvast/src/system/importer.cpp
```diff
@@ -38,44 +39,73 @@ importer_state::importer_state(caf::event_based_actor* self_ptr)
...
+                 ")");
+  } else {
+    VAST_VERBOSE(self, "persisted id block boundary at", current.end);
+  }
   return caf::none;
 }
 
+caf::error importer_state::get_next_block() {
+  using namespace si_literals;
+  while (current.next >= current.end)
```
mavam, line 85: `next > end` seems to be a violation of an invariant.

libvast/src/system/importer.cpp
```diff
@@ -38,44 +39,73 @@ importer_state::importer_state(caf::event_based_actor* self_ptr)
...
+    current.end += 8_Mi;
+  return write_state(write_mode::without_next);
+}
+
+id importer_state::next_id(uint64_t advance) {
+  id pre = current.next;
+  current.next += advance;
+  if (current.next >= current.end)
+    get_next_block();
```
mavam, line 94: I don't like that `next` can exceed `end`. This should be an invariant that gets never violated. Here, it's even across function calls and therefore especially brittle. Why don't we check `if (current.next + advance < current.end)`?

tobim: The code has been adjusted to never violate `current.next < current.end` any more.

libvast/src/system/importer.cpp
tobim: Let's keep the comment chain in one place.

PR #951, opened 2020-06-24
Use just a single option for setting the table slice type

dominiklohmann:
We have a `table-slice-type` option available as both `system.table-slice-type` and `import.table-slice-type`. This is quite confusing. Let's make it a single option just like `import.table-slice-type`.

### DIFF
--- a/libvast/src/defaults.cpp
+++ b/libvast/src/defaults.cpp
@@ -22,16 +22,6 @@
 
 namespace vast::defaults::import {
 
-caf::atom_value table_slice_type(const caf::actor_system& sys,
-                                 const caf::settings& options) {
-  // The parameter import.table-slice-type overrides system.table-slice-type.
-  if (auto val = caf::get_if<caf::atom_value>(&options,
-                                              "import.table-slice-type"))
-    return *val;
-  return get_or(sys.config(), "system.table-slice-type",
-                system::table_slice_type);
-}
-
 size_t test::seed(const caf::settings& options) {
   std::string cat = category;
   if (auto val = caf::get_if<size_t>(&options, cat + ".seed"))
--- a/vast.conf
+++ b/vast.conf
@@ -17,9 +17,6 @@ system {
   ; can be underrun if the source has a low rate).
   ;table-slice-size = 100
 
-  ; The table slice type (caf|arrow).
-  ;table-slice-type = 'arrow'
-
   ; The size of an index shard.
   ;max-partition-size = 1000000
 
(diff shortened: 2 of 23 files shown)


lava:
Are the table slice type/size constants only used during import? If not, `system` may be the better name to keep since it will be confusing to have e.g. export-related options in `import.table_slice_size`.

Also, I wonder what the current semantics are if we import with a different table slice type/size than the one configured in the node? Would this result in an error?

dominiklohmann:
> Are the table slice type/size constants only used during import? If not, `system` may be the better name to keep since it will be confusing to have e.g. export-related options in `import.table_slice_size`.

Only during ingestion, which means during import and when adding `vast.metrics` events.

> Also, I wonder what the current semantics are if we import with a different table slice type/size than the one configured in the node? Would this result in an error?

For the type this never mattered, for the size this was fixed by #908.

libvast/src/defaults.cpp
```diff
@@ -22,16 +22,6 @@
...
 
-caf::atom_value table_slice_type(const caf::actor_system& sys,
-                                 const caf::settings& options) {
-  // The parameter import.table-slice-type overrides system.table-slice-type.
-  if (auto val = caf::get_if<caf::atom_value>(&options,
-                                              "import.table-slice-type"))
-    return *val;
-  return get_or(sys.config(), "system.table-slice-type",
-                system::table_slice_type);
-}
```
lava, line 33: If I remember correctly, this function is used so that the "default" table slice type can be changed to `msgpack` in vast-pro. Unless I'm missing something, removing it will change this behaviour?

vast.conf
```diff
@@ -132,7 +125,11 @@ import {
   ; Block until the importer forwarded all data.
   ;blocking = false
 
-  ; Select the table slice type.
+  ; Number of events to be batched in a table slice (this is a target value that
+  ; can be underrun if the source has a low rate).
+  ;table-slice-size = 100
+
+  ; The table slice type (default|arrow).
```
lava, line 132: This comment needs an update once the renaming PR is merged.

dominiklohmann: It already is, I missed this during the rebase actually. Good catch!

libvast/src/defaults.cpp
dominiklohmann: I'll take a look at this one specifically with VAST Pro in mind and will report back.

dominiklohmann: That's not an issue, here's the relevant code snippet:

```cpp
struct pro_configuration : default_configuration {
  pro_configuration() {
    // Adjust defaults.
    put(content, "import.table-slice-type", caf::atom("msgpack"));
  }
};
```

PR #954, opened 2020-06-26
Fix spawn source subcommand

tobim:


libvast/src/system/node.cpp
```diff
@@ -332,14 +335,23 @@ node_state::spawn_command(const invocation& inv,
       VAST_DEBUG(this_node, "auto-generated new label:", label);
     }
   }
+  VAST_DEBUG(this_node, "spawns a", comp_name, "with the label", label);
+  auto spawn_inv = inv;
```
mavam, line 339: Should this be a copy or an alias?

libvast/src/system/node.cpp
```diff
@@ -332,14 +335,23 @@ node_state::spawn_command(const invocation& inv,
...
+  if (comp_name == "source") {
+    auto spawn_opt = caf::get<caf::settings>(spawn_inv.options, "spawn");
+    auto source_opt = caf::get<caf::settings>(spawn_opt, "source");
+    auto import_opt = caf::get<caf::settings>(spawn_inv.options, "import");
+    merge_settings(source_opt, import_opt);
+    put(spawn_inv.options, "import", import_opt);
```
mavam, line 345:
```suggestion
    spawn_inv.options["import"] = import_opt;
```

If this works, it feels way more idiomatic.

libvast/src/detail/line_range.cpp
```diff
@@ -20,7 +20,6 @@ namespace vast {
 namespace detail {
 
 line_range::line_range(std::istream& input) : input_{input} {
-  next(); // prime the pump
```
mavam, line 23: Have you checked that all line-based formats now issue their first call to `next` lazily?

mavam:
I haven't seen any unit or integration tests yet. What was your strategy in this regard?

mavam:
When I test this locally, I get the following warning in `vast start`:
```
2020-06-29T14:09:19.767 spawn_command got an error from spawn_component: none
```

mavam:
I tried `vast spawn source pcap -i em0` and get a segfault:

```
Target 0: (vast) stopped.
(lldb) bt
* thread #2, name = 'caf.multiplexer', stop reason = EXC_BAD_ACCESS (code=1, address=0x368)
  * frame #0: 0x00007fff6e6b3148 libpcap.A.dylib`pcap_stats + 4
    frame #1: 0x0000000100597d7d libvast.2020.dylib`::status() at pcap.cpp:127:18 [opt]
    frame #2: 0x00000001003755b8 libvast.2020.dylib`::send_report() at source.hpp:181:30 [opt]
    frame #3: 0x0000000100374b0e libvast.2020.dylib`::operator()() [inlined] operator() at source.hpp:261:12 [opt]
 ```

**Comment by tobim** [comment]
> When I test this locally, I get the following warning in `vast start`:
> 
> ```
> 2020-06-29T14:09:19.767 spawn_command got an error from spawn_component: none
> ```

I missed that. I think this is because of the eraser. `spawn_node` tries to create one but `spawn_eraser` returns early with `ec::no_error` if no aging query is configured. I believe this should be handled by explicitly not printing a warning in case of `no_error`.

File: libvast/src/detail/line_range.cpp (hunk 3 above)
Comment (tobim) [reply libvast/src/detail/line_range.cpp:23]: I modified `zeek`, `json`, `csv` and `syslog`. I just found one more user of `line_range`: `parser_reader`, but it isn't used...

**Comment by tobim** [comment]
> I tried `vast spawn source pcap -i em0` and get a segfault:

I don't know what's causing this, can you `vast import pcap -i em0`?

File: libvast/src/system/node.cpp (hunk 1 above)
Comment (tobim) [reply libvast/src/system/node.cpp:339]: A copy is required, because the function signature is mandated by the factory, and that requires `inv` to be const.

**Comment by tobim** [comment]
> I haven't seen any unit or integration tests yet. What was your strategy in this regard?

Commands should be covered by integration tests in general. I'm still thinking about how this can be worked into the dsl.

**Comment by mavam** [comment]
I'm already getting a segfault when I run `vast spawn source pcap`, without any params.

Backtrace:
```
    frame #0: 0x000000010042b7b9 libvast.2020.dylib`::spawn_command() [inlined] __tree at __tree:1726:42 [opt]
    frame #1: 0x000000010042b784 libvast.2020.dylib`::spawn_command() [inlined] __tree at __tree:1721 [opt]
    frame #2: 0x000000010042b784 libvast.2020.dylib`::spawn_command() [inlined] map at map:1031 [opt]
    frame #3: 0x000000010042b784 libvast.2020.dylib`::spawn_command() [inlined] map at map:1032 [opt]
  * frame #4: 0x000000010042b784 libvast.2020.dylib`::spawn_command() [inlined] dictionary at dictionary.hpp:89 [opt]
    frame #5: 0x000000010042b784 libvast.2020.dylib`::spawn_command() [inlined] dictionary at dictionary.hpp:89 [opt]
    frame #6: 0x000000010042b784 libvast.2020.dylib`::spawn_command() [inlined] get<caf::dictionary<caf::config_value> > at settings.hpp:61 [opt]
    frame #7: 0x000000010042b765 libvast.2020.dylib`::spawn_command() at node.cpp:341 [opt]
    frame #8: 0x0000000100131cff libvast.2020.dylib`::run() [inlined] __invoke<caf::message (*const &)(const vast::invocation &, caf::actor_system &), vast::invocation &, caf::actor_system &> at type_traits:4425:1 [opt]
    frame #9: 0x0000000100131cf7 libvast.2020.dylib`::run() [inlined] invoke<caf::message (*const &)(const vast::invocation &, caf::actor_system &), vast::invocation &, caf::actor_system &> at functional:2848 [opt]
    frame #10: 0x0000000100131cf7 libvast.2020.dylib`::run() at command.cpp:433 [opt]
 ```

mavam:
> I don't know what's causing this [...]

How did you test invalid parameters on the command line? This always causes segfaults for me, like this one ☝️.

dominiklohmann:
I'll give this a spin to try reproducing the issues locally.

dominiklohmann:
I can reproduce the changes mentioned by @tobim locally on macOS.

The PCAP source spawned by the node hangs after forwarding to the importer once. Sadly we don't have good enough trace logs for the PCAP source, which is what we should start with.

I cannot, however, reproduce the issues mentioned by @mavam. I think something may be messed up on your end.

dominiklohmann:
Upon further investigation, the PCAP source inside the node seems to be down after a while. Sending a status atom to the source returns with `request_receiver_down` after the configured timeout.

libvast/src/system/spawn_source.cpp
```diff
@@ -60,9 +60,16 @@ maybe_actor spawn_generic_source(node_actor* self, spawn_arguments& args,
...
-  auto src = self->spawn(source<Reader>, std::move(reader), slice_size,
-                         max_events, st.type_registry, vast::schema{},
-                         std::move(type_filter), accountant_type{});
+  auto src
+    = self->spawn<caf::detached>(source<Reader>, std::move(reader), slice_size,
+                                 max_events, st.type_registry, vast::schema{},
+                                 std::move(type_filter), accountant_type{});
+  src->attach_functor([=, name = reader.name()](const caf::error& reason) {
+    if (!reason || reason == caf::exit_reason::user_shutdown)
+      VAST_VERBOSE(src, name, "source shuts down");
```
mavam, line 69: Is the output still adhering to our subject-verb-object logging convention?

libvast/vast/system/source.hpp
```diff
@@ -216,21 +221,30 @@ source(caf::stateful_actor<source_state<Reader>>* self, Reader reader,
...
+  VAST_TRACE(VAST_ARG(self));
   // Initialize state.
   auto& st = self->state;
   st.init(self, std::move(reader), std::move(max_events),
           std::move(type_registry), std::move(local_schema),
           std::move(type_filter), std::move(accountant));
+  self->set_down_handler([=](const caf::down_msg& msg) {
+    VAST_VERBOSE(self, "received DOWN from", msg.source);
+    self->state.done = true;
+  });
```
mavam, line 233: Why do we need this handler? The source doesn't monitor any actor, so it cannot receive a DOWN message.

libvast/vast/system/source.hpp
```diff
@@ -307,9 +321,12 @@ source(caf::stateful_actor<source_state<Reader>>* self, Reader reader,
...
       auto& st = self->state;
-      if (st.sink) {
-        self->quit(make_error(ec::logic_error, "does not support multiple "
-                                               "sinks"));
+      if (st.sink.address() != sink.address()) {
+        self->quit(caf::make_error(ec::logic_error, "source does not support "
+                                                    "multiple sinks"));
+      } else if (st.sink) {
+        VAST_WARNING(self, "ignores request from", self->current_sender(),
+                     "to set same sink again");
```
mavam, line 329: How can this happen? Seems like an invariant violation, so we should terminate at this point and not keep running.

libvast/src/system/spawn_source.cpp
dominiklohmann: It's `<actor-id> pcap-reader source shuts down` for `vast spawn source pcap`, so I'd say yes.

libvast/vast/system/source.hpp
dominiklohmann: Leftover from testing where I monitored the node. Let's clean this up once we figured out the remaining shutdown issue.

libvast/vast/system/source.hpp
dominiklohmann: I've hit this once during testing and have never been able to reproduce it since. I'd like to leave it in until I can reason about it better.

libvast/vast/system/source.hpp
mavam: Fair enough; let's leave this comment open until you remove the handler.

libvast/vast/system/source.hpp
mavam: Can we make it an assertion or even hard abort? If it's such a weird error, we should make it more "in your face".

dominiklohmann: I still want to see if I hit this again, which is why this for now logs the address of the sending actor. An assertion doesn't really allow for me to debug this here.

tobim: I put some thought into this after a call with @dominiklohmann. I think an assertion is the right approach here, because in this case, the state of the source does not need to be changed and this alone is no reason to abort the whole process. We still want to see an error in debug and CI builds if it happens, and an assertions gives us exactly that.

libvast/vast/system/source.hpp
dominiklohmann: I removed it.

libvast/vast/system/source.hpp
dominiklohmann: I'd argue it's reasonable to shut down the source, but just the source. The node itself should continue to live after the source shuts down.

dominiklohmann: I'm suggesting this change:

```diff
diff --git a/libvast/vast/system/source.hpp b/libvast/vast/system/source.hpp
index f25fe5cc..b6110a9c 100644
--- a/libvast/vast/system/source.hpp
+++ b/libvast/vast/system/source.hpp
@@ -317,12 +317,9 @@ source(caf::stateful_actor<source_state<Reader>>* self, Reader reader,
       //       implement an anycast downstream manager and use it for the
       //       source, because we mustn't duplicate data.
       auto& st = self->state;
-      if (st.sink.address() != sink.address()) {
+      if (st.sink || st.sink.address() != sink.address()) {
         self->quit(caf::make_error(ec::logic_error, "source does not support "
                                                     "multiple sinks"));
-      } else if (st.sink) {
-        VAST_WARNING(self, "ignores request from", self->current_sender(),
-                     "to set same sink again");
         return;
       }
       st.sink = sink;
```

integration/default_set.yaml
```diff
@@ -343,3 +343,11 @@ tests:
       - command: export ascii 'resp_h == 192.168.1.104'
       - command: send eraser run
       - command: export ascii 'resp_h == 192.168.1.104'
+  Spawn source:
+    tags: [import, spawn-source, zeek]
+    fixture: ServerTester
+    steps:
+      - command: spawn source suricata -r @./data/suricata/eve.json
+      - command: import -b zeek
```
dominiklohmann, line 351: Can you add a comment explaining why this is necessary?

dominiklohmann:
We should add changelog entries for the other fixes we made as well @tobim.

PR #955, opened 2020-06-29
Relax csv header element requirements

tobim:


libvast/src/format/csv.cpp
```diff
@@ -456,7 +459,7 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
       continue;
     }
     if (!p(line))
-      return make_error(ec::type_clash, "unable to parse CSV line");
+      return make_error(ec::type_clash, "unable to parse CSV line", line);
```
dominiklohmann, line 462: For JSON, we skip over invalid lines. Can we do this for CSV as well?

libvast/src/format/csv.cpp
```diff
@@ -15,6 +15,7 @@
 
 #include "vast/concept/parseable/core.hpp"
 #include "vast/concept/parseable/string.hpp"
```
dominiklohmann, line 17: You should be able to remove this include.

libvast/src/format/csv.cpp
```diff
@@ -188,6 +189,7 @@ const char* reader::name() const {
 
 caf::optional<record_type>
 reader::make_layout(const std::vector<std::string>& names) {
+  VAST_TRACE(__func__, VAST_ARG(names));
```
dominiklohmann, line 192: `__func__` is the first parameter of all trace logs by default, no need to specify it here.

libvast/src/format/csv.cpp
tobim: Good idea, I added a status handler to match the JSON reader.

libvast/src/format/csv.cpp
```diff
@@ -395,9 +396,23 @@ make_csv_parser(const record_type& layout, table_slice_builder_ptr builder,
...
+  using namespace std::string_literals;
+  uint64_t num_lines = num_lines_;
+  uint64_t invalid_lines = num_invalid_lines_;
+  if (num_invalid_lines_ > 0)
+    VAST_WARNING(this, "failed to parse", num_invalid_lines_, "of", num_lines_,
+                 "recent lines");
+  num_lines_ = 0;
+  num_invalid_lines_ = 0;
+  return {{name() + ".num-lines"s, num_lines},
+          {name() + ".invalid-lines"s, invalid_lines}};
```
dominiklohmann, line 409:
```suggestion
          {name() + ".invalid-lines"s, invalid_lines},};
```

PR #961, opened 2020-07-01
Add separate workflow for docker image

0ortmann:


.github/workflows/vast-docker.yaml
```diff
@@ -0,0 +1,31 @@
+name: "VAST"
+on:
+  pull_request:
+    types: [opened, synchronize]
+    paths:
+    - Dockerfile
```
dominiklohmann, line 6: Can you also trigger this if this workflow changes?

dominiklohmann:
```suggestion
    - Dockerfile
    - .github/workflows/vast-docker.yaml
```

.github/workflows/vast-docker.yaml
```diff
@@ -0,0 +1,32 @@
+name: "VAST"
```
dominiklohmann, line 1:
```suggestion
name: "VAST Docker"
```

PR #964, opened 2020-07-03
Make spawn_source generic for all readers

dominiklohmann:


libvast/vast/system/spawn_source.hpp
```diff
@@ -14,50 +14,59 @@
...
-/// @returns a handle to the spawned actor on success, an error otherwise
-maybe_actor spawn_zeek_source(node_actor* self, spawn_arguments& args);
+template <class Reader, class Defaults = typename Reader::defaults>
+maybe_actor spawn_source(node_actor* self, spawn_arguments& args) {
+  VAST_TRACE(VAST_ARG("node", self), VAST_ARG(args));
+  auto& st = self->state;
+  auto& options = args.inv.options;
+  // Bail out early for bogus invocations.
+  if (caf::get_or(options, "system.node", false))
+    return make_error(ec::parse_error, "cannot start a local node");
```
mavam, line 41: Can we make the error message more meaningful, i.e., *why* can't we start a local node?

libvast/vast/system/spawn_source.hpp
```diff
@@ -14,50 +14,59 @@
...
+  VAST_UNBOX_VAR(in, detail::make_input_stream<Defaults>(args.inv.options));
+  VAST_UNBOX_VAR(sch, read_schema(args));
```
mavam, line 43: I'm not a fan of these macros, but we can keep them if you like 'em. To me, the make it more opaque what's actually going on here.

libvast/vast/system/spawn_source.hpp
```diff
@@ -14,50 +14,59 @@
...
+  auto table_slice_type = caf::get_or(options, "source.spawn.table-slice-type",
+                                      defaults::import::table_slice_type);
```
mavam, line 45: Wasn't the table slice type supposed to be a top-level option?

libvast/vast/system/spawn_source.hpp
```diff
@@ -14,50 +14,59 @@
...
+  VAST_INFO(self, "spawned a", reader.name(), "source");
+  auto src
+    = self->spawn<caf::detached>(source<Reader>, std::move(reader), slice_size,
+                                 max_events, st.type_registry, vast::schema{},
+                                 std::move(type_filter), accountant_type{});
+  src->attach_functor([=, name = reader.name()](const caf::error& reason) {
+    if (!reason || reason == caf::exit_reason::user_shutdown)
+      VAST_INFO(name, "source shuts down");
+    else
+      VAST_WARNING(name, "source shuts down with error:", reason);
```
mavam, line 66:
```suggestion
      VAST_WARNING(name, "source shut down with error:", reason);
```

libvast/vast/system/spawn_source.hpp
dominiklohmann: I fully agree. We use the same error message everywhere we bail out when `system.node` is set, so I just copied it over here.

libvast/vast/system/spawn_source.hpp
dominiklohmann: It's in the `import` section, and the `spawn.source` section gets merged into the `import` section, so this is correct.

libvast/vast/system/spawn_source.hpp
dominiklohmann: Let's make that a separate story—we should just remove them everywhere. I did not really touch this code, I just moved it to a separate place.

libvast/vast/system/spawn_source.hpp
dominiklohmann: But this is before it has completely shut down, so I think present tense is okay here?

libvast/vast/system/spawn_source.hpp
dominiklohmann: Do you have a suggestion for a better replacement?

mavam: The error sounds like an invalid combination of command line parameters that are not working, so I'd say something along the lines of "invalid command line: 'spawn' in local node is a no-op".

libvast/vast/system/spawn_source.hpp
mavam: My understanding was that `attach_functor` executes after the actor terminated. But [looking at the docs](https://actor-framework.readthedocs.io/en/latest/Actors.html#attaching-cleanup-code-to-actors), all it says is *during*. Since you can attach cleanup code for remote actors as well, it's equivalent to receiving a DOWN message. Therefore, I suggest we treat `attach_functor` as executed after shutdown.

dominiklohmann:
[ch17783]

libvast/vast/system/spawn_source.hpp
dominiklohmann: I've created a story for replacing this everywhere in both VAST and VAST Pro.

libvast/vast/system/spawn_source.hpp
dominiklohmann: Addressed.

libvast/vast/system/spawn_source.hpp
tobim: I think it makes sense to change the messages that are introduced in this pr to the proposed wording directly. A new story for doing this everywhere is nice, but it won't get prioritized for a while.

PR #966, opened 2020-07-06
Add a static binary workflow

tobim:
Adds a new workflow that uses nix to build a static binary. It runs after every push to master and can be manually triggered by the `repository_dispatch` event.

TODO:

- [x] Add a script to trigger the manual event


.github/workflows/static-binary.yml
```diff
@@ -0,0 +1,85 @@
...
+    - name: Upload Artifact to Github
+      uses: actions/upload-artifact@v1
+      with:
+        name: "${{ steps.create_paths.outputs.artifact_name }}"
+        path: "${{ steps.create_paths.outputs.build_dir }}/${{ steps.create_paths.outputs.artifact_name }}"
+
+  push_to_master:
+    if: github.event_name == 'push'
+    runs-on: ubuntu-latest
+    name: Master Static Binary (${{ matrix.args }})
```
dominiklohmann, line 54: I don't think `${{ matrix.args }}` resolves to something that looks good in the left column of the actions tab, right?

.github/workflows/static-binary.yml
```diff
@@ -0,0 +1,85 @@
...
+    - name: Build a Static Binary
+      run: |
+        nix/static-binary.sh --use-head
+
+    - name: Create Paths
+      id: create_paths
+      run: |
+        BUILD_DIR="build"
+        echo "::set-output name=build_dir::${BUILD_DIR}"
+        ARTIFACT_NAME=$(ls "$BUILD_DIR" | grep "vast-pro.*.tar.gz")
```
dominiklohmann, line 78:
```suggestion
        ARTIFACT_NAME=$(ls "$BUILD_DIR" | grep "vast.*.tar.gz")
```

.github/workflows/static-binary.yml
```diff
@@ -0,0 +1,85 @@
...
+
+    - name: Build a Static Binary
+      run: |
+        nix/static-binary.sh --use-head
+
+    - name: Create Paths
+      id: create_paths
+      run: |
+        BUILD_DIR="build"
+        echo "::set-output name=build_dir::${BUILD_DIR}"
```
dominiklohmann, line 77: Same es above: job-level `env`, then access via `${{ env.BUILD_DIR }}`.

CMakeLists.txt
```diff
@@ -82,8 +82,8 @@ option(VAST_ENABLE_ASSERTIONS "Enable Assertions"
 option(VAST_RELOCATABLE_INSTALL "Enable relocatable installations" ON)
 option(VAST_USE_BUNDLED_CAF "Always use the CAF submodule" OFF)
 option(ENABLE_ZEEK_TO_VAST "Build zeek-to-vast" ON)
-option(VAST_STATIC_EXECUTABLE "Link VAST statically")
-option(VAST_USE_JEMALLOC "Use jemalloc instead of libc malloc" OFF)
+option(VAST_STATIC_EXECUTABLE "Link VAST statically" $ENV{VAST_STATIC_EXECUTABLE})
+option(VAST_USE_JEMALLOC "Use jemalloc instead of libc malloc" "${VAST_STATIC_EXECUTABLE}")
```
dominiklohmann, line 86: Changed by mistake? This doesn't look right to me.

cmake/FindZSTD.cmake
```diff
@@ -23,6 +23,13 @@ set(ZSTD_STATIC_LIB_NAME
     ${CMAKE_STATIC_LIBRARY_PREFIX}zstd${ZSTD_STATIC_LIB_SUFFIX})
 
 # First, find via if specified ZTD_ROOT
+if (NOT ZSTD_ROOT)
+  # We look for an environment variable of the same name too, because
+  # that can be provided by a tool like nix-shell or docker. Wo do this
+  # so the user doesn't have to remember to add "-DZSTD_ROOT=..." to their
+  # cmake invocations.
```
dominiklohmann, line 30: Having this for some variables but not all is really confusing and makes maintenance of the CMake way harder. I'm already always afraid to change CMake code because it might break the Nix build.

nix/overlay.nix
```diff
@@ -0,0 +1,34 @@
...
+      "-DCAF_BUILD_STATIC=ON"
+      "-DCAF_BUILD_STATIC_ONLY=ON"
+      "-DOPENSSL_USE_STATIC_LIBS=TRUE"
+      "-DCMAKE_INTERPROCEDURAL_OPTIMIZATION:BOOL=ON"
+      "-DCMAKE_POLICY_DEFAULT_CMP0069=NEW"
+    ];
+    hardeningDisable = [
+      "pic"
+    ];
+    NIX_CFLAGS_COMPILE = "-std=c++17";
```
dominiklohmann, line 28: Do we really need to explicitly set this here again? That's another point where we can diverge from CMake.

nix/pinned.nix
```diff
@@ -0,0 +1,9 @@
+# See https://nixos.wiki/wiki/FAQ/Pinning_Nixpkgs for more information on pinning
+builtins.fetchTarball {
+  # Descriptive name to make the store path easier to identify
+  name = "nixpkgs-unstable-2020-06-07";
+  # Commit hash for nixpkgs-unstable as of date
+  url = https://github.com/NixOS/nixpkgs-channels/archive/029a5de08390bb03c3f44230b064fd1850c6658a.tar.gz;
+  # Hash obtained using `nix-prefetch-url --unpack <url>`
+  sha256 = "03fjkzhrs2avcvdabgm7a65rnyjaqbqdnv4q86qyjkkwg64g5m8x";
+}
```
dominiklohmann, line 9: How do we update this? Is there a workflow that does this for us automatically? Does this mean that updating CAF involves updating this file?

.github/workflows/static-binary.yml
```diff
@@ -0,0 +1,85 @@
+name: "Static Binary"
+on:
+  repository_dispatch:
+    types: static-binary
+  push:
+    branches:
+      - master
+  pull_request:
+    paths:
+      - '.github/workflows/static-binary.yml'
```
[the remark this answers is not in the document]
lava: Not sure we need a release trigger until we have special handling for releases (i.e. uploading tagged binaries as part of the release). The release commit itself will already get built due to the push to master.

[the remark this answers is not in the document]
dominiklohmann: Yes we do, because we want to attach the binary to the release itself.

lava:
>  Add a script to trigger the manual event

The script already exists in event-horizon, maybe it's even fine to leave it there since only tenzir developers will have permissions to trigger manual dispatches anyways. However, it will not be possible to test the manual trigger until the workflow that responds to it lands on master, so not in this PR.

CMakeLists.txt
tobim: The intention is to use the environment variable if the option has not been given explicitly. The order of precedence is now:
explicit option > environment > default (implicitly off because the empty string is treated as False).

The reason I changed these is that I can use nix to set up an environment for a static build like so:
```bash
nix-shell -E 'let pkgs = import ./nix { }; in import ./shell.nix {pkgs = pkgs.pkgsStatic;}'
``` 
The toolchain in this environment is not capable to produce shared libraries, so one has to pass several options to CMake to get a working build tree with a configuration similar to the static build in CI:
```bash
cmake -Bbuild -S . -DVAST_STATIC_EXECUTABLE=ON -DZSTD_ROOT="$(nix-build ./nix --no-out-link -A pkgsStatic.zstd)" -DVAST_USE_JEMALLOC=ON
```

With the changes to CMake adding these options is not required, while the regular build is not affected.

nix/overlay.nix
tobim: Please note that this is for building CAF + VAST in static mode with `-flto`. The compiler complains if it sees different definitions of the same types in its internal passes. This can be removed once we migrate to CAF 0.18.

nix/pinned.nix
tobim: We update this manually when there is a reason to do so. Trying to automate this would actually generate more work for dealing with the generated PRs alone. This is not needed for updating CAF.

.github/workflows/static-binary.yml
```diff
@@ -8,12 +8,17 @@ on:
   pull_request:
     paths:
       - '.github/workflows/static-binary.yml'
+  release:
+    types: published
```
dominiklohmann, line 12: No job runs for `github.event_name == 'release'`

CMakeLists.txt
dominiklohmann: I really don't like how inconsistent this is. Can you create a follow-up story for a) prefixing all CMake variables that the user can set with `VAST_`, and b) for exposing them on the command line like this?

nix/overlay.nix
dominiklohmann:
```suggestion
    # TODO: Remove when updating to CAF 0.18.
    NIX_CFLAGS_COMPILE = "-std=c++17";
```

lava: If we're going to leave a comment, let's also mention *why* this can be removed when updating to CAF 0.18, otherwise people in the future will have to reverse-engineer this conversation.

scripts/trigger-static-build
```diff
@@ -0,0 +1,17 @@
...
+if [ "$#" -lt 1 ]; then
+  >&2 echo "Using 'master' as default branch."
+fi
+
+if [ -z "${GITHUB_TOKEN}" ]; then
+  >&2 echo "A GitHub personal access token with \"Repo\" access must be"
+  >&2 echo "availabe in the environment variable GITHUB_TOKEN."
+fi
+
+./trigger-action static-build "{\"ref\": \"${1:-master}\", \"args\": \"${*:2}\"}"
```
lava, line 17: This is not the latest master version of this script, there were some fixes pushed on friday.

scripts/trigger-action
```diff
@@ -0,0 +1,22 @@
...
+  echo "    availabe in the environment variable GITHUB_TOKEN."
+  echo
+}
+
+if [ "$#" -lt 1 ]; then
+  usage >&2
+  exit 1
+fi
+
+curl -X POST https://api.github.com/repos/tenzir/vast/dispatches \
```
lava, line 19: Maybe we should change the script to take the target repository (i.e. `tenzir/vast` or `tenzir/event-horizon`) as first argument? Then we can use the same script for everything.

tobim: That can probably be extracted from `git remote -v`, but I don't want to get side tracked by this now.

mavam:
The name for the workflow seems still pretty cryptic. Can we strip the technical details?

.github/workflows/static-binary.yml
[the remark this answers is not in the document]
tobim: Release trigger is in. We also have on push to master, so I don't think nightly makes sense.

tobim:
> The name for the workflow seems still pretty cryptic. Can we strip the technical details?

The name is now "VAST Static / Static Binary (_trigger_)".

.github/workflows/static-binary.yml
```diff
@@ -0,0 +1,69 @@
...
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        upload_url: ${{ github.event.release.upload_url }}
+        asset_path: "${{ env.BUILD_DIR }}/${{ steps.create_paths.outputs.artifact_name }}"
+        # The asset name is constant so we can permanently link to
+        # https://github.com/tenzir/vast/releases/latest/download/vast-linux-static.tar.gz
+        # for a build of the latest release.
+        asset_name: "vast-linux-static.tar.gz"
+        asset_content_type: application/gzip
```
dominiklohmann, line 69: That's a great idea, can you make this change for the VAST workflow as well?

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🎁 Starting with this release, installing VAST on any Linux becomes
+  significantly easier: A static binary will be provided with each release on
+  the GitHub releases page. The direct link to the latest version is
+  https://github.com/tenzir/vast/releases/latest/download/vast-linux-static.tar.gz
```
mavam, line 17: This this link may change because of a variety of reasons, I wouldn't hardcode in the changelog. In a year from now, I bet things will be different...but the changelog won't be.

tobim: Fair enough.

dominiklohmann: I don't agree. The changelog should reflect our current set of changes, and this is a user-facing change.

If the link changes again in the future, we will add another changelog entry.

tobim: I think we can add the link to the docs and the release notes, maybe it even qualifies for the blog post, that way it does not necessarily have to be in the changelog entry.

cmake/FindZSTD.cmake
tobim: This is essentially a workaround for a bug in the generated `.pc` files in ZSTD. We can probably get rid of `FindZSTD.cmake` when upgrading to the next Arrow release thanks to https://github.com/apache/arrow/pull/7388.

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🎁 Starting with this release, installing VAST on any Linux becomes
+  significantly easier: A static binary will be provided with each release on
+  the GitHub releases page.
+  [#966](https://github.com/tenzir/vast/966)
```
dominiklohmann, line 17:
```suggestion
- 🎁 Starting with this release, installing VAST on any Linux becomes
  significantly easier: A static binary will be provided with each release on
  the GitHub releases page. [#966](https://github.com/tenzir/vast/pull/966)
```

Notice the missing `pull` in the URL. Also, this fits the same line with the usual width.

.github/workflows/vast.yaml
```diff
@@ -261,5 +263,10 @@ jobs:
         with:
           upload_url: ${{ github.event.release.upload_url }}
           asset_path: "${{ steps.configure_env.outputs.build_dir }}/${{ steps.configure_env.outputs.package_name }}.tar.gz"
+          # The asset names are constant so we can permanently link to
+          # https://github.com/tenzir/vast/releases/latest/download/vast-ubuntu-release-gcc.tar.gz
+          # https://github.com/tenzir/vast/releases/latest/download/vast-ubuntu-release-clang.tar.gz
+          # https://github.com/tenzir/vast/releases/latest/download/vast-darwin-release-appleclang.tar.gz
+          # for builds of the latest release.
           asset_name: "${{ steps.configure_env.outputs.package_name }}.tar.gz"
```
dominiklohmann, line 271:
```suggestion
          asset_name: "${{ steps.configure_env.outputs.publish_name }}.tar.gz"
```

.github/workflows/static-binary.yaml
```diff
@@ -0,0 +1,69 @@
+name: "VAST Static"
+on:
+  repository_dispatch:
+    types: static-binary
+  push:
+    branches:
+      - master
+  pull_request:
+    paths:
+      - '.github/workflows/static-binary.yml'
```
dominiklohmann, line 10:
```suggestion
      - '.github/workflows/static-binary.yaml'
```

PR #967, opened 2020-07-06
Unify source configuration

dominiklohmann:
This is essentially a bugfix; the configuration for sources and datagram sources deviated from the configuration for remotely spawning sources a while back, since the code was untested and took a different code path for configuration.  This change makes remotely spawning a source take the same code path as spawning a local source and thus enables remotely spawning datagram sources.

I don't yet have a good solution for printing the reader name in the source shutdown functor, so if anyone's got an idea for that I'd greatly appreciate that.

### DIFF
--- a/libvast/vast/system/datagram_source.hpp
+++ b/libvast/vast/system/datagram_source.hpp
@@ -164,10 +164,6 @@ datagram_source(datagram_source_actor<Reader>* self,
       self->delayed_send(self, defaults::system::telemetry_rate,
                          atom::telemetry_v);
     },
-    [=](type_registry_type type_registry) {
-      // TODO adapt
-      VAST_DEBUG(self, "sets type-registry to", type_registry);
-    },
     [=](atom::sink, const caf::actor& sink) {
       // TODO: Currently, we use a broadcast downstream manager. We need to
       //       implement an anycast downstream manager and use it for the
@@ -186,11 +182,10 @@ datagram_source(datagram_source_actor<Reader>* self,
         return err;
       return caf::unit;
     },
-    [=](expression& expr) {
-      VAST_INFO(self, "sets filter expression to", expr);
-      self->state.filter = std::move(expr);
+    [=]([[maybe_unused]] expression& expr) {
       // FIXME: Allow for filtering import data.
-      VAST_WARNING(self, "fitler expressions are not currently implemented");
+      // self->state.filter = std::move(expr);
+      VAST_WARNING(self, "does not currently implement filter expressions");
     },
     [=](atom::telemetry) {
       auto& st = self->state;
--- a/libvast/vast/system/datagram_source.hpp
+++ b/libvast/vast/system/datagram_source.hpp
@@ -187,8 +187,10 @@ datagram_source(datagram_source_actor<Reader>* self,
       return caf::unit;
     },
     [=](expression& expr) {
-      VAST_DEBUG(self, "sets filter expression to:", expr);
+      VAST_INFO(self, "sets filter expression to", expr);
       self->state.filter = std::move(expr);
+      // FIXME: Allow for filtering import data.
+      VAST_WARNING(self, "fitler expressions are not currently implemented");
     },
     [=](atom::telemetry) {
       auto& st = self->state;
--- a/libvast/vast/system/import_command.hpp
+++ b/libvast/vast/system/import_command.hpp
@@ -14,168 +14,23 @@
 #pragma once
 
 #include "vast/command.hpp"
-#include "vast/concept/parseable/to.hpp"
-#include "vast/concept/parseable/vast/endpoint.hpp"
-#include "vast/concept/parseable/vast/expression.hpp"
-#include "vast/concept/parseable/vast/schema.hpp"
 #include "vast/defaults.hpp"
-#include "vast/detail/make_io_stream.hpp"
-#include "vast/detail/string.hpp"
-#include "vast/endpoint.hpp"
 #include "vast/error.hpp"
-#include "vast/expression.hpp"
-#include "vast/format/reader.hpp"
 #include "vast/fwd.hpp"
 #include "vast/logger.hpp"
-#include "vast/schema.hpp"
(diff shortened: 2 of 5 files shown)


dominiklohmann:
For reviewing: I verified locally using VAST Pro that this works with the datagram source without any further changes. The below command works as expected in our office.

```
vast spawn source netflow -l :2055/udp
```

libvast/vast/system/import_command.hpp
```diff
@@ -179,36 +155,76 @@ caf::message import_command(const invocation& inv, caf::actor_system& sys) {
...
+    self, node, {"accountant", "type-registry", "importer"});
+  if (!components)
+    return make_message(std::move(components.error()));
+  VAST_ASSERT(components->size() == 3);
+  auto accountant = caf::actor_cast<accountant_type>((*components)[0]);
+  auto type_registry = caf::actor_cast<type_registry_type>((*components)[1]);
+  auto& importer = (*components)[2];
+  if (!type_registry)
+    return make_message(make_error(ec::missing_component, "type-registry"));
+  // Configure source.
```
tobim, line 206: That comment is not very helpful.

libvast/vast/system/import_command.hpp
```diff
@@ -65,11 +64,14 @@ caf::expected<expression> parse_expression(command::argument_iterator begin,
...
-template <class Reader, class Defaults>
-caf::message import_command(const invocation& inv, caf::actor_system& sys) {
-  VAST_TRACE(inv.full_name, VAST_ARG("options", inv.options), VAST_ARG(sys));
+template <class Reader, class Defaults,
+          caf::spawn_options SpawnOptions = caf::spawn_options::no_flags,
+          class Actor>
+caf::expected<caf::actor>
+configure_source(const Actor& self, caf::actor_system& sys,
+                 const invocation& inv, accountant_type accountant,
+                 type_registry_type type_registry, caf::actor importer) {
```
tobim, line 73: Can you add a comment describing what this function does? I also think that the name is not indicative of the effect. Maybe just `spawn_source_impl`?

dominiklohmann: What about `make_source`?

dominiklohmann: I went with `make_source` for now and moved it into its own file.

tobim: I'm OK with that.

libvast/vast/system/make_source.hpp
```diff
@@ -0,0 +1,181 @@
...
+/// @param inv The invocation that prompted the actor to be spawned.
+/// @param accountant A handle to the accountant component.
+/// @param type_registry A handle to the type registry component.
+/// @param importer A handle to the importer component.
+/// @returns a handle to the spawned actor on success, an error otherwise
+template <class Reader, class Defaults,
+          caf::spawn_options SpawnOptions = caf::spawn_options::no_flags,
+          class Actor>
+caf::expected<caf::actor>
+make_source(const Actor& self, caf::actor_system& sys, const invocation& inv,
```
tobim, line 73: `self` is only used twice at the end of the function, we could either move these uses into the calling context or use `caf::anon_send`. That would at least allow us to get rid of the `Actor` template parameter.

dominiklohmann: I'd rather not do this. `anon_send` needlessly obfuscates the logs, and I also don't want to spread out the initialization of the source again. After all, the whole point of this function is to unify the configuration.

libvast/vast/system/spawn_source.hpp
```diff
@@ -42,33 +37,20 @@ maybe_actor spawn_source(node_actor* self, spawn_arguments& args) {
...
-  auto max_events = caf::get_if<size_t>(&options, "source.spawn.max-events");
-  auto type = caf::get_if<std::string>(&options, "source.spawn.type");
-  auto type_filter = type ? std::move(*type) : std::string{};
-  auto schema = get_schema(options, "spawn.source");
-  if (!schema)
-    return schema.error();
-  if (slice_size == 0)
-    return make_error(ec::invalid_configuration, "table-slice-size can't be 0");
-  Reader reader{table_slice_type, options, std::move(in)};
-  VAST_INFO(self, "spawned a", reader.name(), "source");
```
dominiklohmann, line 60: TODO: Find a way to re-introduce this log message.

libvast/vast/system/make_source.hpp
tobim: I'll cede my case in this particular instance. I want to repeat tough, that we won't measure any runtime advantage from monomorphing the code paths around spawning sources and sinks, but it does cost us quite a bit in development time.

libvast/vast/system/datagram_source.hpp
```diff
@@ -187,8 +183,10 @@ datagram_source(datagram_source_actor<Reader>* self,
       return caf::unit;
     },
     [=](expression& expr) {
-      VAST_DEBUG(self, "sets filter expression to:", expr);
+      VAST_INFO(self, "sets filter expression to", expr);
       self->state.filter = std::move(expr);
+      // FIXME: Allow for filtering import data.
+      VAST_WARNING(self, "fitler expressions are not currently implemented");
```
tobim, line 189:
```suggestion
      VAST_WARNING(self, "does currently not implement filter expressions");
```

libvast/vast/system/source.hpp
```diff
@@ -307,8 +307,9 @@ source(caf::stateful_actor<source_state<Reader>>* self, Reader reader,
     },
     [=](expression& expr) {
       VAST_INFO(self, "sets filter expression to", expr);
-      auto& st = self->state;
-      st.filter = std::move(expr);
+      self->state.filter = std::move(expr);
+      // FIXME: Allow for filtering import data.
+      VAST_WARNING(self, "fitler expressions are not currently implemented");
```
tobim, line 312:
```suggestion
      VAST_WARNING(self, "does currently not implement filter expressions");
```

PR #971, opened 2020-07-09
Upload static binaries to GCloud

0ortmann:
With this PR the static binary is additionally uploaded to a public GCS bucket. That enables non-GitHub-users to download the artifact. The link for this upload is: https://storage.googleapis.com/tenzir-public-data/vast-static-builds/vast-2020.06.25-128-gaa76128c-static.tar.gz

### DIFF
--- a/.github/workflows/static-binary.yaml
+++ b/.github/workflows/static-binary.yaml
@@ -17,7 +17,6 @@ jobs:
     name: Static Binary
     env:
       BUILD_DIR: build
-      STATIC_BINARY_TARGET: ${{ github.event.client_payload.args == '' && 'vast' || github.event.client_payload.args }}
     steps:
     - name: Checkout
       if: github.event_name == 'repository_dispatch'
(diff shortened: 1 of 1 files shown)


.github/workflows/static-binary.yaml
```diff
@@ -18,6 +18,8 @@ jobs:
     env:
       BUILD_DIR: build
       STATIC_BINARY_TARGET: ${{ github.event.client_payload.args == '' && 'vast' || github.event.client_payload.args }}
+      PUBLIC_GCS_BUCKET: tenzir-public-data
+      STATIC_BINARY_FOLDER: vast-static-builds
```
dominiklohmann, line 22: These only need to be available to a single step; can we move them there?

.github/workflows/static-binary.yaml
```diff
@@ -29,6 +31,18 @@ jobs:
...
+      with:
+        python-version: '3.7'
+
+    - name: Configure GCloud Credentials
+      uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+      with:
+        version: '275.0.0'
+        service_account_email: ${{ secrets.GCP_SA_EMAIL }}
+        service_account_key: ${{ secrets.GCP_SA_KEY }}
+
```
dominiklohmann, line 45: Can we move this whole block down? We don't need to initialize the GCS stuff if the build fails.

.github/workflows/static-binary.yaml
```diff
@@ -54,6 +68,10 @@ jobs:
         name: "${{ steps.create_paths.outputs.artifact_name }}"
         path: "${{ env.BUILD_DIR }}/${{ steps.create_paths.outputs.artifact_name }}"
 
+    - name: Upload Artifact to GCS
+      run: |
```
dominiklohmann, line 72: Should this run on `master` pushes only?

0ortmann: We upload the artifact to GitHub regardless of the branch. Keep in mind that this workflow only runs for pushes to master and changes to the workflow file. I'm in favor of keeping it as it is, but I also don't have a strong opinion on this. We should unify it, though, for GitHub and GCS uploads.

.github/workflows/static-binary.yaml
0ortmann: Moved the env vars.

.github/workflows/static-binary.yaml
0ortmann: Moved the block.

.github/workflows/static-binary.yaml
dominiklohmann: I'm fine with this uploading for all triggers if we have a way to distinguish the uploaded files. As is this looks like it overrides files with the same name.

0ortmann: yes it does. but file names have git SHAs in them, so they should not clash.

PR #973, opened 2020-07-13
Print importer ID space using strings

dominiklohmann:
This fixes a bug that makes it look like the available IDs counter has overflown. The underlying limitation is that we transport this information using `caf::settings`, which is a `caf::dictionary` of `caf::config_value`s, which can only represent signed 64 bit integers.

### DIFF
--- a/libvast/src/system/importer.cpp
+++ b/libvast/src/system/importer.cpp
@@ -13,6 +13,7 @@
 
 #include "vast/system/importer.hpp"
 
+#include "vast/concept/printable/numeric/integral.hpp"
 #include "vast/concept/printable/to_string.hpp"
 #include "vast/concept/printable/vast/error.hpp"
 #include "vast/concept/printable/vast/filesystem.hpp"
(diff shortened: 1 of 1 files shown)


libvast/src/system/importer.cpp
```diff
@@ -106,11 +107,12 @@ id importer_state::available_ids() const noexcept {
...
-  result.emplace("available-ids", available_ids());
-  result["ids.available"] = available_ids();
-  result["ids.block.next"] = current.next;
-  result["ids.block.end"] = current.end;
+  // TODO: caf::config_value can only represent signed 64 bit integers, which
+  // may make it look like overflow happened in the status report. As an
+  // intermediate workaround, we convert the values to strings.
+  caf::put(result, "ids.available", to_string(available_ids()));
+  caf::put(result, "ids.block.next", to_string(current.next));
+  caf::put(result, "ids.block.end", to_string(current.end));
```
mavam, line 115: Why did you switch back from the idiomatic `xs["key"] = value` to the free-form style assignment?

tobim: The reason is stated in the commit message:

> `operator[]` on `caf::settings` does not group by `.` automatically, while `caf::put` does.

mavam: Hah, I think we may have a few other bugs introduced recently then.

dominiklohmann: I noticed this when querying the result using `jq`, because I had to use `.node."importer-1"."ids.available"` instead of `.node."importer-1".ids.available` as key. Ideally, we would get rid of the dashes as well and use arrays instead, so it's `.node.importers[0].ids.available`, but that's a separate story.

mavam: > Ideally, we would get rid of the dashes as well and use arrays instead

I like this idea.

PR #972, opened 2020-07-13
Use Flatbuffers for Persistent State of Segment Store and Meta Index

lava:
We want to use a well-defined on-disk format for VAST, to enable stable upgrade paths for VAST users.

This PR adds the first step of this effort by changing the segment store and the meta index to use a flatbuffer layout for its persistent on-disk data.

In addition, some preparatory work for more complete flatbuffer integration in the future is included, such as the `Layout` and `Type` type definitions. 

Finally, a new `filesystem` actor is included, which is not strictly related to flatbufferization but will be used in follow-up work to centralize the file reads and writes when loading and storing flatbuffers.

NOTE: This PR adds a hard dependency on `flatbuffers` to VAST.

PR #976, opened 2020-07-13
Update GitHub workflows to run on Ubuntu 20.04

0ortmann:
This PR elevates the Ubuntu version used both in CI and in our Dockerfiles to 20.04.

A recent merge to master (https://github.com/tenzir/vast/pull/972) has corrupted the Docker build due to a missing dependency for `flatbuffers`. This branch also covers the fix of the current master.

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -25,13 +25,13 @@ jobs:
       matrix:
         os:
         - tag: Ubuntu
-          name: ubuntu-latest
+          name: ubuntu-20.04
           extra-flags:
           compiler: GCC
           cc: gcc-9
           cxx: g++-9
         - tag: Ubuntu
-          name: ubuntu-latest
+          name: ubuntu-20.04
           extra-flags:
           compiler: Clang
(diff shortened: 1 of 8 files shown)


.github/workflows/vast.yaml
```diff
@@ -178,7 +178,7 @@ jobs:
           ccache --zero-stats
           # --show-config is a relatively new ccache option and not available
           # in the ccache version installed via apt, so we ignore failure.
```
dominiklohmann, line 180: Can you remove this comment as well?

dominiklohmann:
We agreed to merge this despite the failing CI in the standup; the failing unit tests with gcc-8 / debian-buster will be addressed as part of #978, which addresses the root cause.

Before merging, as one last step, we should adapt the Dockerfile to build and test in separate steps:

```Dockerfile
RUN cmake --build build --parallel
RUN CTEST_OUTPUT_ON_FAILURE=1 cmake --build build --target test
RUN cmake --build build --target integration
RUN cmake --build build --target install
```

PR #978, opened 2020-07-14
Register MessagePack types with factories

dominiklohmann:


integration/default_set.yaml
```diff
@@ -152,11 +152,13 @@ tests:
     steps:
       - command: import -b --table-slice-type=caf zeek
         input: data/zeek/conn.log.gz
-      - command: import -b --table-slice-type=arrow zeek
-        input: data/zeek/conn.log.gz
       - command: import -b --table-slice-type=msgpack zeek
         input: data/zeek/conn.log.gz
       - command: count '#type == "zeek.conn"'
+      - guard: 'export -h 2>&1 | grep -q arrow'
```
mavam, line 158: Could this be reformulated more idiomatically by using the `version` command?

dominiklohmann: Yeah, like `version | jq -re '."Apache Arrow" != null'

dominiklohmann: I've implemented this change.

dominiklohmann: The push somehow doesn't show up in here, I think the GitHub UI is broken.

libvast/CMakeLists.txt
```diff
@@ -244,7 +244,7 @@ if (VAST_USE_OPENSSL)
   target_link_libraries(libvast PUBLIC caf::openssl)
 endif ()
 
-target_link_libraries(libvast PRIVATE flatbuffers::flatbuffers)
+target_link_libraries(libvast PUBLIC flatbuffers::flatbuffers)
```
tobim, line 247: Is there an unavoidable reason that flatbuffers leak into the interface here?

dominiklohmann: Yeah, zeek-to-vast doesn't build without it. I'd use the INTERFACE target, but that's broken unless we bump the CMake version (a lot).

PR #980, opened 2020-07-15
Improve handling of the default schema paths

tobim:
This PR removes the baking to the install prefix from the binary.
 - The path to the lowest priority schema files was already determined relative to the binary location at runtime. Now the help message properly reflects that.
 - The fallback path to `vast.conf` is now also determined by the runtime location.

This fixes 2 bugs: 
 1. `VAST_RELOCATABLE_INSTALL` was not really relocatable because the paths above would not update on relocations.
 2. The `CMAKE_INSTALL_PREFIX` leaked into the static binary, which is relocatable by definition.

In addition, this PR also adds `/etc/vast/schema` to the schema file lookup path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
-- ⚠️ Vast now recognizes /etc/vast/schema as an additional default directory
+- ⚠️ VAST now recognizes `/etc/vast/schema` as an additional default directory
   for schema files. [#980](https://github.com/tenzir/vast/pull/980)
 
 - ⚠️ [Flatbuffers](https://google.github.io/flatbuffers/) is now a required
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- ⚠️ Vast now recognizes /etc/vast/schema as an additional default directory
+  for schema files. [#980](https://github.com/tenzir/vast/pull/980)
+
 - ⚠️ [Flatbuffers](https://google.github.io/flatbuffers/) is now a required
   dependency for VAST. The archive and the segment store use flatbuffers to
   store and version their on-disk persistent state.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,7 +81,7 @@ endif ()
 option(VAST_ENABLE_ASSERTIONS "Enable Assertions"
        "${VAST_ENABLE_ASSERTIONS_DEFAULT}")
 # Setting this option removes configuration dependent absolute paths from the
-# vast installation. Concretely, it enables the dynamic binary and libraries to
+# VAST installation. Concretely, it enables the dynamic binary and libraries to
 # use relative paths for loading their dependencies.
 option(VAST_RELOCATABLE_INSTALL "Enable relocatable installations" ON)
 option(VAST_USE_BUNDLED_CAF "Always use the CAF submodule" OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,6 +80,9 @@ endif ()
 
 option(VAST_ENABLE_ASSERTIONS "Enable Assertions"
        "${VAST_ENABLE_ASSERTIONS_DEFAULT}")
+# Setting this option removes configuration dependent absolute paths from the
+# vast installation. Concretely, it enables the dynamic binary and libraries to
+# use relative paths for loading their dependencies.
 option(VAST_RELOCATABLE_INSTALL "Enable relocatable installations" ON)
 option(VAST_USE_BUNDLED_CAF "Always use the CAF submodule" OFF)
 option(ENABLE_ZEEK_TO_VAST "Build zeek-to-vast" ON)
--- a/libvast/src/system/application.cpp
+++ b/libvast/src/system/application.cpp
@@ -16,6 +16,7 @@
 #include "vast/command.hpp"
 #include "vast/config.hpp"
 #include "vast/detail/assert.hpp"
+#include "vast/detail/process.hpp"
 #include "vast/documentation.hpp"
 #include "vast/format/ascii.hpp"
 #include "vast/format/csv.hpp"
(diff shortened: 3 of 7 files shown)


tobim:
> Can you revert the change to the lookup order? `PREFIX/etc/vast/vast.conf` should come before `/etc/vast/vast.conf` in the ordering, and this PR changes that.

That was intentional. As a user, I would intuitively edit `/etc/vast/vast.conf` and expect my changes to take effect, with the old ordering that expectation would not always be met. To me this is another small bump in the road that might negatively effect my evaluation of VAST.

dominiklohmann:
> That was intentional. As a user, I would intuitively edit `/etc/vast/vast.conf` and expect my changes to take effect, with the old ordering that expectation would not always be met.

I don't get it. The usual behavior for programs is that the more specific a path to a configuration file, the earlier it should be in the lookup path. I think we should discuss this separately within the team (including ops) and have a vote on the expected and desired user experience.

As a side note: if this change was intentional, it should be reflected as such in the changelog.

mavam:
> The usual behavior for programs is that the more specific a path to a configuration file, the earlier it should be in the lookup path.

That's also my understanding. A home config file takes precedence over system-wide installed version. In that sense, `/etc/vast` is system and if the prefix is not `/`, it's more specific than the system.

tobim:
I can only speak of what I'm used to and to some extent expect:

The configuration of a program is normally composed of 3 parts in order of precedence: 
1. The user level config, found in `~/.vast/vast.conf` or `~/.config/vast/vast.conf` (FSB scheme)
2. The system level config in `/etc/vast/vast.conf`
4. The defaults built into the application

What vast does at the moment:
1. `$PWD/vast.conf` (A)
2. `<prefix>/share/vast/vast.conf` (B)
3. `/etc/vast/vast.conf`
4. defaults

> That's also my understanding. A home config file takes precedence over system-wide installed version. In that sense, /etc/vast is system and if the prefix is not /, it's more specific than the system.

This is not the way I think about the issue. `<prefix>/share/vast/` contains settings defined by the upstream or vendor, thus they are the least specific. `/etc/vast/` contains settings by the system administrator, thus they are more specific.

tobim:
I think this change should be discussed in a separate thread though. I'm going to revert the change for this PR.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- ⚠️ Vast now recognizes /etc/vast/schema as an additional default directory
```
dominiklohmann, line 14:
```suggestion
- ⚠️ VAST now recognizes /etc/vast/schema as an additional default directory
```

CMakeLists.txt
```diff
@@ -80,6 +80,9 @@ endif ()
 
 option(VAST_ENABLE_ASSERTIONS "Enable Assertions"
        "${VAST_ENABLE_ASSERTIONS_DEFAULT}")
+# Setting this option removes configuration dependent absolute paths from the
+# vast installation. Concretely, it enables the dynamic binary and libraries to
```
dominiklohmann, line 84:
```suggestion
# VAST installation. Concretely, it enables the dynamic binary and libraries to
```

libvast/src/system/application.cpp
```diff
@@ -72,15 +73,21 @@ auto make_root_command(std::string_view path) {
...
   auto ob
     = opts("?system")
         .add<std::string>("config", "path to a configuration file")
         .add<caf::atom_value>("verbosity,v", "output verbosity level on the "
                                              "console")
-        .add<std::vector<std::string>>("schema-paths",
-                                       "list of paths to look for schema files "
-                                       "([" VAST_INSTALL_PREFIX
-                                       "/share/vast/schema])")
+        .add<std::vector<std::string>>("schema-paths", schema_desc.c_str())
```
dominiklohmann, line 90: Isn't this undefined behavior? What if the underlying memory of the `std::string` gets deallocated?  Or is this `add` member function making a copy?

libvast/src/system/configuration.cpp
```diff
@@ -52,15 +52,20 @@ void initialize_factories() {
...
+  // Use 'vast.conf' instead of generic 'caf-application.ini'.
+  auto config_path_candidates = std::vector{
+    path::current() / "vast.conf",
+  };
+  auto binary = detail::objectpath();
+  if (binary)
+    config_path_candidates.push_back(binary->parent().parent() / "share"
+                                     / "vast" / "schema");
+  config_path_candidates.emplace_back("/etc/vast/vast.conf");
+  for (auto& p : config_path_candidates) {
```
dominiklohmann, line 64:
```suggestion
  for (const auto& p : config_path_candidates) {
```

vast.conf
```diff
@@ -19,8 +19,11 @@ system {
   ; The unique ID of this node.
   ;node-id = "node"
 
-  ; List of paths to look for schema files.
-  ;schema-paths = <["$VAST_INSTALL_PREFIX/share/vast/schema"]>
+  ; List of paths to look for schema files in ascending order of priority.
+  ; Note: Automatically prepended with
+  ;  ["<binary_directory>/../share/vast/schema", "/etc/vast/schema"].
+  ; Use --no-default-schema to turn off this mechanism.
```
dominiklohmann, line 25: In the context of the configuration file, I think it makes more sense to reference the `system.no-default-schema` option. And considering it is right below, I don't think this needs to be  in here at all.

tobim: I'd rather keep it. Most people don't read example config files top to botton.

libvast/src/system/configuration.cpp
tobim:
```
$ git grep 'for (auto&' | wc
    207    1368   13621

$ git grep 'for (const auto&' | wc
     22     171    1568
```
In principle I'm in favor of this suggestion, but I think it is a question of revising the coding guidelines.

libvast/src/system/application.cpp
tobim: The constructor of `config_option` makes a copy. That gets called inside of add after a couple of indirections.

vast.conf
dominiklohmann: Fair enough. I don't have a very strong opinion on this.

PR #982, opened 2020-07-16
Fix multi-config generator support

dominiklohmann:
I verified locally that this works for both gen-vast-slices and the integration test suite.


CMakeLists.txt
```diff
@@ -683,14 +683,6 @@ if (NOT vast_is_subproject AND NOT TARGET configure)
...
 endif ()
 
-set(EXECUTABLE_OUTPUT_PATH
-    ${CMAKE_CURRENT_BINARY_DIR}/bin
-    CACHE PATH "Single directory for all executables")
-
-set(LIBRARY_OUTPUT_PATH
-    ${CMAKE_CURRENT_BINARY_DIR}/lib
-    CACHE PATH "Single directory for all libraries")
-
```
tobim, line 693: Removing these lines means the files go somewhere else, but the multi-config generators still work correctly if they are kept. The `$<TARGET_FILE:>` changes are sufficient.

dominiklohmann: I've added them back.

PR #983, opened 2020-07-21
Link targets for multi-config generators

dominiklohmann:
This allows the use of `./build/bin/vast` to run the latest compiled build configuration for multi-config generators.


dominiklohmann:
I've been starting to think that using `copy_if_different` may be preferred here, since we look at some paths based on the location of the binary.

cmake/Helpers.cmake
```diff
@@ -52,3 +52,22 @@ macro (install_symlink _filepath _sympath)
...
+macro (multi_config_target_symlink _target)
+  get_property(_is_multi_config GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+  if (_is_multi_config)
+    add_custom_command(
+      TARGET "${_target}"
+      POST_BUILD
+      COMMAND "${CMAKE_COMMAND}" -E create_symlink "$<TARGET_FILE:${_target}>"
+              "${EXECUTABLE_OUTPUT_PATH}/${_target}"
+      COMMENT
+        "Linking ${EXECUTABLE_OUTPUT_PATH}/${_target} -> $<TARGET_FILE:${_target}>"
```
tobim, line 70:
```suggestion
              "${EXECUTABLE_OUTPUT_PATH}/$<TARGET_FILE:${_target}>"
      COMMENT
        "Linking ${EXECUTABLE_OUTPUT_PATH}/$<TARGET_FILE:${_target}> -> $<TARGET_FILE:${_target}>"
```

tobim:
> I've been starting to think that using `copy_if_different` may be preferred here, since we look at some paths based on the location of the binary.

I didn't think about that. I guess custom targets are not affected by multi-config generators?

cmake/Helpers.cmake
dominiklohmann: That doesn't make any sense, since the `TARGET_FILE` generator expression returns the absolute path to the target file.

dominiklohmann:
> > I've been starting to think that using `copy_if_different` may be preferred here, since we look at some paths based on the location of the binary.
> 
> I didn't think about that. I guess custom targets are not affected by multi-config generators?

They are not. This whole thing only happens because we're using `EXECUTABLE_OUTPUT_PATH` instead of setting the target locations for the package and install targets only. That variable is deprecated for a good reason. 

I'm open to either changing to `copy_if_different` and merging or closing this until we inevitably re-do our CMake in a more modern style.

dominiklohmann:
I'm closing this. We discussed at the office that there is no good solution with our current CMake setup.

PR #984, opened 2020-07-21
Fix bug in decoding multi-object MsgPack types

mavam:


dominiklohmann:
For the review: note that this is a breaking change for both MessagePack-encoded table slices in the archive and the persisted type-registry state.

mavam:
Great. You can "self-approve" this and merge! :tada:

PR #995, opened 2020-07-29
Restructure vast status

tobim:


libvast/src/system/application.cpp
```diff
@@ -357,9 +357,10 @@ auto make_spawn_command() {
 }
 
 auto make_status_command() {
-  return std::make_unique<command>("status",
-                                   "shows various properties of a topology",
-                                   documentation::vast_status, opts());
+  return std::make_unique<command>(
+    "status", "shows various properties of a topology",
+    documentation::vast_status,
+    opts().add<caf::atom_value>("verbosity,v", "output verbosity level"));
```
dominiklohmann, line 363: Shouldn't this just re-use our global `system.verbosity` setting?

If you want to keep them separate, please introduce a different name to avoid confusion, and don't re-use the same short option.

libvast/src/detail/process.cpp
```diff
@@ -63,4 +130,15 @@ caf::expected<path> objectpath(const void* addr) {
   return objectpath_static();
 }
 
+caf::settings get_status() {
```
dominiklohmann, line 133:
```suggestion
caf::expected<caf::settings> get_status() {
```

Can we make this return an expected settings, such that downstream can handle the error in a unique way, and that the third path can return `make_error(ec::unimplemented)`?

libvast/src/system/importer.cpp
```diff
@@ -106,17 +107,21 @@ id importer_state::available_ids() const noexcept {
...
   // may make it look like overflow happened in the status report. As an
   // intermediate workaround, we convert the values to strings.
-  caf::put(result, "ids.available", to_string(available_ids()));
-  caf::put(result, "ids.block.next", to_string(current.next));
-  caf::put(result, "ids.block.end", to_string(current.end));
+  if (v >= status_verbosity::verbose) {
+    caf::put(s.verbose, "ids.available", to_string(available_ids()));
+    caf::put(s.verbose, "ids.block.next", to_string(current.next));
+    caf::put(s.verbose, "ids.block.end", to_string(current.end));
+  }
```
dominiklohmann, line 120: Can you remove the TODO comment and remove the `to_string` calls?

libvast/vast/concept/parseable/string/any.hpp
```diff
@@ -14,6 +14,7 @@
 #pragma once
 
 #include "vast/concept/parseable/core/parser.hpp"
+#include "vast/concept/parseable/detail/char_helpers.hpp"
```
dominiklohmann, line 17: Nice catch.

libvast/vast/detail/process.hpp
```diff
@@ -25,4 +27,6 @@ namespace vast::detail {
 ///          addr, or error if the resolution fails.
 caf::expected<path> objectpath(const void* addr = nullptr);
 
+caf::settings get_status();
```
dominiklohmann, line 30: Can you add documentation for this function?

libvast/vast/system/source.hpp
```diff
@@ -114,8 +114,10 @@ struct source_state {
   /// Points to the owning actor.
   Self* self;
 
+  size_t count = 0;
```
dominiklohmann, line 117: Can you add documentation for this variable?

libvast/src/detail/process.cpp
dominiklohmann, line 133: Since the output here differs depending on the platform, can you add an integration test that uses `jq '.path.to.rusage | keys'` to test whether all keys are available on all platforms?

dominiklohmann (changes requested):
First review round; this time code only. Still compiling on the side to give feedback to the output itself.

libvast/src/system/posix_filesystem.cpp
```diff
@@ -62,20 +61,24 @@ posix_filesystem(filesystem_type::stateful_pointer<posix_filesystem_state> self,
...
-        dict["bytes"] = stats.bytes;
-      };
-      add_stats("writes", self->state.stats.writes);
-      add_stats("reads", self->state.stats.reads);
-      add_stats("mmaps", self->state.stats.mmaps);
-      return result;
+    [=](atom::status, status_verbosity v) {
+      vast::status s;
+      if (v >= status_verbosity::info) {
+        s.info["filesystem.type"] = "POSIX";
```
dominiklohmann, line 67:
```suggestion
        caf::put(s.info, "filesystem.type", "POSIX");
```

dominiklohmann (changes requested):
From a user POV, I've noticed a few things:

Things are no longer grouped by components. The output is much easier to parse if the info, verbose and debug objects are merged like this: 

```
vast status -v debug | jq '.info * .debug * .verbose'
```

I think we should just do this by default. Having `.info.archive` and `.debug.archive` separated from each other makes no sense as a user.

For the components, the names are doubled as keys. In a future PR, I'd like to see a `components: [...]` array containing all the components, with `components[<index>].name` containing the name. Right now, `<name>.name` is always `"<name>"`.

The helptext of `vast status` should show what unit the memory usage values are in.

libvast/src/system/application.cpp
tobim: If we separate the verbosity from the object it attaches to, then `system.verbosity` and `status.verbosity` are semantically the same. While we could rename this option to `--detail` it seems somewhat arbitrary to me.

Reusing `system.verbosity` would require the user to write `vast -v debug status` instead of `vast status -v debug`.

I kept this consistent with the log level option intentionally. But we could also change it to `--verbose` and `--debug` flags.

dominiklohmann: From a user perspective, I find having options that are named exactly the same do different things depending on their position and take different arguments quite confusing.

mavam: In this case I'm agreeing with @tobim. I would find a confusing that I have to control the output (stdout) through the same option that controls a different channel (stderr).

dominiklohmann: In that case I'd still like to see a different long and short option. Using `verbose,v` for both _will_ confuse users. 

What do you think about `detail`?

mavam: It was my understanding the we have 3 levels, just like with logging: info -> verbose -> debug. So using `-v`, `-vv` and `-vvv` seems easy to remember. (Just to illustrate, I know we don't support multiple options of the same type.)

I'm not against using something like `detail` as a prefix for the different output levels.

dominiklohmann: You're misunderstanding me @mavam. I don't want these options to have the same name, as that conflicts with a future refactoring we have planned for the CLI for allowing options to appear at a later point than the subcommand they relate to. If we are not merging the `system.verbosity` and `status.verbosity` options, we should _rename_ the `status.verbosity` option, e.g., to `status.detail`.

I'd like to see the short option removed altogether. This command is not used a lot, and we should not re-use short options.

mavam: I'm fine with using clearer naming that differentiates from the logging nomenclature. My only concern is that it should be clear that we multiple levels of detail - 3 in fact. So call them `--detail`, `--meticulous`, `--comprehensive`, or what you like. The max level could be `--developer` to make it clear that it has debug-like semantics.

libvast/src/detail/process.cpp
tobim: In this particular case I would rather keep the return type as `caf::settings` for simplicity, the only calling site is in `node.cpp`, and there we have nothing else to do than discard an error.

dominiklohmann: Fair enough.

dominiklohmann (changes requested):
Can you clean up the includes and the guarding `#if`s in `libvast/src/detail/process.cpp`? There are some seemingly unnecessary includes, and they should be moved to the top.

We also shouldn't use both `__linux__` and `VAST_LINUX`, only the latter should be used for guarding includes.

libvast/src/detail/process.cpp
tobim: This is not possible, for example macOS does not expose information about swap.

dominiklohmann (approved):
Thanks! 🙏

I verified locally that this looks as we discussed.

PR #998, opened 2020-07-30
Remove tracker actor and merge with node

mavam:


libvast/src/system/component_registry.cpp
```diff
@@ -0,0 +1,64 @@
...
+                             std::string label) {
+  VAST_ASSERT(comp);
+  if (label.empty())
+    label = type;
+  for (auto& [lab, c] : components_)
+    if (lab == label || c.actor == comp)
+      return false;
+  components_.emplace(std::move(label),
+                      component{std::move(comp), std::move(type)});
+  return true;
```
tobim, line 30:
```suggestion
#if VAST_ENABLE_ASSERTIONS
    for (auto& c : components_)
      VAST_ASSERT(c.second.actor != comp);
#endif
    return components_.emplace(std::move(label),
                             component{std::move(comp), std::move(type)})[1];
```

PR #999, opened 2020-07-31
Allow the '-' in the expression key parser

tobim:


mavam:
Should `-` really be allowed to be at the beginning of a field name? What are the consequences for the language, given that a field can be either on LHS and RHS?

tobim:
[ch18663]

tobim:
@mavam I changed the code so '-' is not allowed as the first character any longer.

The last commit moves the entire implementation of the parser to the `.cpp` file. This is done with the intention to reduce coupling and compile times.

libvast/vast/concept/parseable/vast/expression.hpp
```diff
@@ -49,11 +49,12 @@ struct predicate_parser : parser<predicate_parser> {
     using parsers::chr;
     using namespace parser_literals;
     auto id = +(alnum | chr{'_'} | chr{'-'});
+    auto key_char = alnum | chr{'_'} | chr{'-'} | chr{':'};
     // A key cannot start with ':', othwise it would be interpreted as a type
     // extractor.
```
mavam: Can you also adjust the comment so that it reflects the change?

libvast/vast/concept/parseable/vast/expression.hpp
```diff
@@ -49,11 +49,12 @@ struct predicate_parser : parser<predicate_parser> {
...
-    auto key = !':'_p >> (+(alnum | chr{'_'} | chr{'-'} | chr{':'}) % '.');
+    auto key = !(':'_p | '-') >> (+key_char % '.');
     auto operand
-      = parsers::data ->* to_data_operand
+      = (parsers::data >> !key_char) ->* to_data_operand
```
mavam: Is end-of-input (`eoi`) also "not a key char"?

That is, does it work when the operand is either LHS or RHS?

tobim: `eoi` is not a _char_. If `data` consumed the last character, `key_char` will fail no matter what it is composed of.

libvast/src/concept/parseable/vast/expression.cpp
```diff
@@ -71,8 +71,8 @@ static auto make_predicate_parser() {
   using namespace parser_literals;
   auto id = +(alnum | chr{'_'} | chr{'-'});
   auto key_char = alnum | chr{'_'} | chr{'-'} | chr{':'};
-  // A key cannot start with ':', othwise it would be interpreted as a type
-  // extractor.
+  // A key cannot start with ':' or '-', othwise it would be interpreted as a
+  // type extractor.
```
mavam, line 75: Why would a key/field that starts with `-` be interpreted as *type extractor*?

libvast/test/expression_parseable.cpp
```diff
@@ -39,7 +41,18 @@ TEST(parseable/printable - predicate) {
   CHECK(pred.op == equal);
   CHECK(caf::get<data>(pred.rhs) == data{42u});
   CHECK_EQUAL(to_string(pred), str);
+  // LHS: data, RHS: schema
+  MESSAGE("42 == Foo");
```
mavam, line 45: What about the reverse, e.g., `Foo == 42`?

libvast/src/concept/parseable/vast/expression.cpp
```diff
@@ -51,4 +64,171 @@ predicate predicate_parser::to_data_predicate(data x) {
...
+      dis.emplace_back(std::move(con[0]));
+    else
+      dis.emplace_back(std::move(con));
+    return dis.size() == 1 ? std::move(dis[0]) : expression{dis};
+  };
+  auto ws = ignore(*parsers::space);
+  auto negate_pred = [](predicate p) { return negation{expression{p}}; };
+  auto negate_expr = [](expression expr) { return negation{expr}; };
+  rule<Iterator, expression> expr;
+  rule<Iterator, expression> group;
```
mavam, line 187: Don't you need to *ref* the recursive parsers?

libvast/src/concept/parseable/vast/expression.cpp
```diff
@@ -187,17 +187,19 @@ static auto make_expression_parser() {
...
     | parsers::predicate
     ;
   auto and_or
     = "||"_p  ->* [] { return logical_or; }
     | "&&"_p  ->* [] { return logical_and; }
     ;
   expr
-    = (group >> *(ws >> and_or >> ws >> group) >> ws) ->* to_expr
+    // FIXME: The first use of `group` should be wrapped in a ref, but
+    // if we do that the rule_ref triggers an assertion.
```
mavam, line 201: But without `ref` we have leak, no?

PR #1004, opened 2020-08-06
Tweak some log levels for better visibility

mavam:


libvast/src/format/pcap.cpp
```diff
@@ -248,15 +248,15 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
         return make_error(ec::format_error, "failed to open pcap file ", input_,
                           ": ", std::string{buf});
       }
-      VAST_DEBUG(this, "reads trace from", input_);
+      VAST_INFO(this, "reads trace from", input_);
       if (pseudo_realtime_ > 0)
-        VAST_DEBUG(this, "uses pseudo-realtime factor 1/", pseudo_realtime_);
+        VAST_VERBOSE(this, "uses pseudo-realtime factor 1/", pseudo_realtime_);
```
dominiklohmann, line 253:
```suggestion
        VAST_VERBOSE(this, "uses pseudo-realtime factor 1 /", pseudo_realtime_);
```

This will print `1/ 10` because of the spaces we add for each parameter, so let's add another space.

PR #1005, opened 2020-08-06
Support hard-kill for unresponsive actors

mavam:


libvast/src/system/terminator.cpp
```diff
@@ -54,43 +58,81 @@ caf::behavior terminator(caf::stateful_actor<terminator_state>* self) {
...
-      auto& next = remaining.back();
-      self->monitor(next);
-      self->send_exit(next, caf::exit_reason::user_shutdown);
-    } else if constexpr (std::is_same_v<Policy, policy::parallel>) {
-      // Terminate all actors.
-      for (auto& x : xs) {
-        self->monitor(x);
-        self->send_exit(x, caf::exit_reason::user_shutdown);
+  return {
+    [=](const std::vector<caf::actor>& xs) {
```
mavam, line 62: Note for the reviewers: nothing changed in this handler, `clang-format` just re-indented the whole thing.

libvast/src/system/terminator.cpp
```diff
@@ -54,43 +58,81 @@ caf::behavior terminator(caf::stateful_actor<terminator_state>* self) {
...
+      } else if constexpr (std::is_same_v<Policy, policy::parallel>) {
+        // Terminate all actors.
+        for (auto& x : xs) {
+          self->monitor(x);
+          self->send_exit(x, caf::exit_reason::user_shutdown);
+        }
+      } else {
+        static_assert(detail::always_false_v<Policy>, "unsupported policy");
+      }
+      // Send a reminder for killing all alive actors.
```
mavam, line 99: New changes start at this line.

libvast/src/system/terminator.cpp
```diff
@@ -54,49 +57,89 @@ caf::behavior terminator(caf::stateful_actor<terminator_state>* self) {
...
+    [=](atom::shutdown) {
+      VAST_ASSERT(!self->state.remaining_actors.empty());
+      VAST_WARNING(self, "failed to terminate all actors within 10 mins");
+      VAST_WARNING(self, "initiates hard kill of",
+                   self->state.remaining_actors.size(), "remaining actors");
+      // Kill remaining actors.
+      for (auto& actor : self->state.remaining_actors) {
+        VAST_DEBUG(self, "sends KILL to actor", actor->id());
+        if constexpr (std::is_same_v<Policy, policy::sequential>)
+          self->monitor(actor);
```
lava, line 110: This could use a comment that we only need to monitor in the sequential case because  we have already monitored all actor when killing in parallel.

libvast/src/system/terminator.cpp
```diff
@@ -24,7 +25,9 @@
 namespace vast::system {
 
 template <class Policy>
-caf::behavior terminator(caf::stateful_actor<terminator_state>* self) {
+caf::behavior terminator(caf::stateful_actor<terminator_state>* self,
+                         std::chrono::seconds clean_exit_timeout,
+                         std::chrono::seconds kill_exit_timeout) {
```
lava, line 30: I think "grace period" is the commonly used term for the clean exit timeout.

1 second minimum timeout seems pretty long, using milli- or microsecond precision would feel slightly more future-proof.

mavam: > I think "grace period" is the commonly used term for the clean exit timeout.

I tried to find two names for timeouts that make sense together. If we call the first `grace_period`, how would you call the second, `kill_period`?

mavam:
I agree that it might not be the best to terminate immediately here. But I'm not sure how to integrate this flexibility into the existing API. How would you beef up `shutdown<Policy>(...)`? Use `terminate<Policy>(...)` when the caller should handle a shutdown failure?

mavam:
@lava please take a look at 91bd9ca as an example on how the node would do a shutdown after receiving an error from the terminator. It's a lot more verbose.

libvast/vast/system/shutdown.hpp
```diff
@@ -31,24 +33,52 @@ namespace vast::system {
...
+/// `user_shutdown`, and then waiting for the DOWN. As soon as all actors have
+/// terminated, the calling actor exits with `caf::exit_reason::user_shutdown`.
+/// If an actor does not respond with a DOWN within the provided grace period,
+/// we send out another EXIT message with reason `kill`. If the actor still
+/// does not terminate within the provided timeout, the process aborts hard. If
+/// these failure semantics do not suit your use case, consider using the
+/// function `terminate`, which allows for more detailed control over the
+/// shutdown sequence.
 /// @param self The actor to terminate.
 /// @param xs Actors that need to shutdown before *self* quits.
```
lava, line 47: Let's describe the new options with `@param`, for consistency.

libvast/vast/system/terminate.hpp
```diff
@@ -38,31 +38,46 @@ namespace vast::system {
 /// Performs an asynchronous shutdown of a set of actors by sending an EXIT
 /// message, configurable either in sequential or parallel mode of operation.
 /// As soon as all actors have terminated, the returned promise gets fulfilled.
+/// This function is the lower-level interface for bringing down actors. The
+/// function `shutdown` uses this function internally to implement a more
+/// convenient one-stop solution.
 /// @param self The actor to terminate.
 /// @param xs The actors to terminate.
```
lava, line 45: Same here, let's add `@param` entries for the new entries for consistency.

libvast/vast/system/shutdown.hpp
```diff
@@ -31,24 +33,52 @@ namespace vast::system {
 
 /// Performs an asynchronous shutdown of a set of actors, followed by
 /// terminating the actor in the calling context. The shutdown process runs
-/// either sequentially or in parallel. As soon as all actors have terminated,
-/// the calling actor exits. The shutdown process involves sending an EXIT
-/// message with reason `user_shutdown`.
+/// either sequentially or in parallel, based on the provided policy parameter.
```
lava, line 36: Just a bit of word-smithing, but it feels like you've buried the lede here:

```
Performs an asynchronous shutdown of a set of actors, or terminates the current process if that is not possible.
```

libvast/vast/system/terminator.hpp
```diff
@@ -34,7 +35,14 @@ struct terminator_state {
...
+/// @param self That terminator actor.
+/// @param clean_exit_timeout The timeout after which the terminator sends a
+///        kill exit message to all remaining actors.
+/// @param kill_exit_timeout The timeout after which the terminator gives up
+///        and exits, after having tried to kill remaining actors.
 template <class Policy>
-caf::behavior terminator(caf::stateful_actor<terminator_state>* self);
+caf::behavior terminator(caf::stateful_actor<terminator_state>* self,
+                         std::chrono::seconds clean_exit_timeout,
+                         std::chrono::seconds kill_exit_timeout);
```
lava, line 46: Should we rename these here as well for consistency?

mavam:
Thanks for the eagle eyes. I've integrated all your final suggestions in 6db05d9 and will merge after CI gives green light.

mavam:
I've had to hack around an issue that may be related to https://github.com/actor-framework/actor-framework/issues/1110 in the last commit (da5fbfe). It ain't pretty, but I currently don't know a better way to fix the detached actor issue. We simply can't put a detached actor into our shutdown utility because it won't send us a DOWN message.

I tried coming up with a reproducible CAF example, but didn't get far enough in replicating our logic. Here's my attempt at what point I gave up:

```cpp
#include <caf/all.hpp>

using namespace caf;

using actor_type = typed_actor<
  reacts_to<int>
>;

actor_type::behavior_type aut(actor_type::pointer) {
  return {
    [](int) {
      // nop
    }
  };
}

behavior parent(event_based_actor* self) {
  auto a = self->spawn<linked + detached>(aut);
  self->set_down_handler([=](const down_msg& msg) {
    self->quit(msg.reason);
  });
  self->set_exit_handler([=](const exit_msg& msg) {
    self->monitor(a);
    self->unlink_from(a);
    self->send_exit(a, exit_reason::user_shutdown);
  });
  return {
    [=](int) {
      // nop
    }
  };
}

int main() {
  actor_system_config cfg;
  actor_system sys{cfg};
  scoped_actor self{sys};
  auto a = self->spawn(parent);
  self->monitor(a);
  self->send_exit(a, exit_reason::user_shutdown);
  self->receive([=](const down_msg& msg) { });
}
```

I'm not sure if if it's worth going into this detached actor issue in depth before https://github.com/actor-framework/actor-framework/issues/1110 gets fixed. I already spent a lot of cycles on this but don't see light at the end of the tunnel.

libvast/vast/system/terminator.hpp
```diff
@@ -34,7 +35,14 @@ struct terminator_state {
 };
 
 /// Performs a parallel shutdown of a list of actors.
+/// @param self That terminator actor.
```
lava, line 38: "That"?

vast.conf
```diff
@@ -64,6 +64,10 @@ system {
       ;type = "datagram"
     }
   }
+
+  ; The period to wait until a shutdown sequence finishes cleanly. After the
+  ; elapses, the shutdown procedure escalates into a "hard kill".
```
lava, line 69: > After the elapses

Missing a

Also, let's mention that a value of "0" means infinite, or "never escalate".

libvast/vast/defaults.hpp
```diff
@@ -362,8 +362,12 @@ constexpr std::chrono::milliseconds signal_monitoring_interval
 constexpr std::chrono::seconds initial_request_timeout
   = std::chrono::seconds{10};
 
-/// Timeout to cleanly shutdown actors.
-constexpr std::chrono::seconds shutdown_timeout = std::chrono::hours{1};
+/// Time to wait until an actor exits cleanly.
```
lava, line 365: The descriptions you had below all seemed clearer to me. To quote my favorites:

```
  ; The period to wait until a shutdown sequence finishes cleanly. After the
  ; elapses, the shutdown procedure escalates into a "hard kill".
```
```
/// @param kill_timeout The timeout befor giving and delivering an error to the
///        response promise.
```

libvast/vast/system/shutdown.hpp
```diff
@@ -29,26 +32,58 @@ struct parallel;
...
+/// we send out another EXIT message with reason `kill`. If the actor still
+/// does not terminate within the provided timeout, the process aborts hard. If
+/// these failure semantics do not suit your use case, consider using the
+/// function `terminate`, which allows for more detailed control over the
+/// shutdown sequence.
 /// @param self The actor to terminate.
 /// @param xs Actors that need to shutdown before *self* quits.
+/// @param grace_period The amount of time to wait until all actors terminated
+///        cleanly.
+/// @param kill_timeout The timeout befor giving and calling `abort(3)`.
```
lava, line 51: s/befor giving/before giving up/

Also, we should probably mention that this timeout starts *after* the grace period has ended.

libvast_test/src/node.cpp
```diff
@@ -28,7 +28,8 @@ namespace fixtures {
 
 node::node() {
   MESSAGE("spawning node");
-  test_node = self->spawn(system::node, "test", directory / "node");
+  test_node = self->spawn(system::node, "test", directory / "node",
+                          std::chrono::milliseconds::zero());
```
lava, line 32: Let's add a comment that we're using `zero()` to disable hard kills in unit tests here. (maybe even with a link to the discussion in this review)

libvast/vast/system/terminator.hpp
mavam: Yes, *that* pesky actor. 😅 

(Will change it.)

libvast/src/system/terminator.cpp
```diff
@@ -54,49 +57,101 @@ caf::behavior terminator(caf::stateful_actor<terminator_state>* self) {
...
+      }
+      // Send a reminder for killing all alive actors.
+      if (grace_period > std::chrono::milliseconds::zero())
+        self->delayed_send(self, grace_period, atom::shutdown_v);
+    },
+    [=](atom::shutdown) {
+      VAST_ASSERT(!self->state.remaining_actors.empty());
+      VAST_WARNING(self, "failed to terminate all actors within 10 mins");
+      VAST_WARNING(self, "initiates hard kill of",
+                   self->state.remaining_actors.size(), "remaining actors");
```
dominiklohmann, line 110: These messages can be joined. Also, "10 mins" shouldn't be hardcoded here—the value is configurable now, isn't it?

mavam: Full ACK regarding the hardcoding - this is fixed now.

While they can be joined, I'd like to keep log lines short. It doesn't hurt if you get a multi-line warning in my opinion.

PR #1007, opened 2020-08-06
Make port-encoding for Arrow host-independent

dominiklohmann:


CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 The port-encoding for Arrow-encoded table slices is now host-independent
+  and always uses network host order.
```
mavam, line 15:
```suggestion
- 🐞 The port encoding for Arrow-encoded table slices is now host-independent
  and always uses network-byte order.
```

dominiklohmann: That was the "why don't we have air conditioning in the office"-style changelog entry, dare I say. 🙈

dominiklohmann:
> I think we actually want something slightly different: encode the port as `Struct<UInt16, UInt8>` in Arrow.

We discussed in the office that this is not possible with the current `column_builder_trait`, as it requires state. There is no such thing as a `StructBuilder` that is templated to its members; instead, the StructBuilder receives the builders for its fields during construction, i.e., at runtime.

PR #1010, opened 2020-08-07
Remove set data type

mavam:


libvast/src/system/pivoter.cpp
```diff
@@ -103,17 +103,17 @@ caf::behavior pivoter(caf::stateful_actor<pivoter_state>* self, caf::actor node,
         return;
       VAST_DEBUG(self, "uses", *pivot_field, "to extract", st.target, "events");
       auto column = slice->column(pivot_field->name);
-      auto xs = set{};
+      auto xs = vector{};
```
dominiklohmann, line 106: This looks like a bug to me.

mavam: Only if the input to `xs` is not unique.

dominiklohmann:
```suggestion
      auto xs = detail::stable_set<data>{};
```

And then switch back to insert below.

mavam: Does it need to be an order-preserving set or would an unordered set work as well?

dominiklohmann: I'd argue that for bisecting reasons (if we were to find a bug with this), you should just use what I proposed, as it essentially is `vast::set`. This PR removes the set data type, it should not change the way pivoting works.

mavam: The thing is, I need to create a `data` from the `vector` later on. So if I use a stable set, I need to move all items over into a new container.

I looked at the code again and there is no difference to using a `vector` here, because uniqueness is enforced via the auxiliary structure `requested_ids`. So using `vector` instead of `set` is safe.

dominiklohmann: If you are certain that this does not change the behavior, feel free to go ahead and mark this as resolved.

CHANGELOG.md
```diff
@@ -11,6 +11,14 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- ⚠️ The `set` type has been removed. Experience with the data model showed
+  that there is no strong use case to separate sets from vectors in the core.
+  While this may be useful in programming languages, VAST deals with immutable
+  data where set constraints have been enforced upstream. Users can still
+  attach set semantics to vectors using type attributes, but the operations are
```
dominiklohmann, line 18: How? The changelog should not talk about yet-to-be-implemented features. If it is possible already, show an example such that users have an easy way to migrate their schemas.

libvast/test/parse_data.cpp
```diff
@@ -62,11 +62,8 @@ TEST(data) {
   MESSAGE("vector");
   CHECK_EQUAL(to_data("[]"), vector{});
   CHECK_EQUAL(to_data("[42, 4.2, nil]"), (vector{42u, 4.2, caf::none}));
-  MESSAGE("set");
-  CHECK_EQUAL(to_data("{}"), set{});
-  CHECK_EQUAL(to_data("{42, 42, nil}"), (set{42u, caf::none}));
   MESSAGE("map");
-  CHECK_EQUAL(to_data("{-}"), map{});
+  CHECK_EQUAL(to_data("{}"), map{});
```
dominiklohmann, line 66: This change is not reflected in the changelog.

libvast_test/src/table_slices.cpp
```diff
@@ -60,23 +59,7 @@ table_slices::table_slices() {
     {"vj", vector_type{subnet_type{}}},
     {"vk", vector_type{port_type{}}},
     // {"vl", vector_type{vector_type{count_type{}}}},
-    // {"vm", vector_type{set_type{bool_type{}}}},
     // {"vn", vector_type{map_type{count_type{}, bool_type{}}}},
```
dominiklohmann, line 62: Can you `s/n/m` for all entries here so the it's easier to keep track?

CHANGELOG.md
mavam: What I'm proposing here is that users can do things like `vector<int> #set` to annotate their type. VAST currently does not exploit any of that, so it may be moot for the changelog. I'll add a note on how to migrate schemas.

libvast_test/src/table_slices.cpp
mavam: I'm not sure what you mean here. Could what is supposed to be replaced? The field name?

CHANGELOG.md
dominiklohmann: So I did understand correctly.

Please just remove that part from the changelog. We do not support it yet, and the changelog should only reflect what's actually been changed.

libvast_test/src/table_slices.cpp
dominiklohmann: The variables were named quite consistently, counting up from a to n for the different types. Now that m is gone, n should be renamed to m for all type combinations.

integration/queries/multi_addr.txt
```diff
@@ -1 +1 @@
-:addr in { 192.168.1.1, 192.168.1.255, 192.168.1.255, 192.168.1.255, 192.168.1.255, 192.168.1.255, 192.168.1.255, 192.168.1.1, 192.168.1.255, 192.168.1.1, 192.168.1.1, 192.168.1.1, 224.0.0.251, ff02::fb, 192.168.1.255, 192.168.1.255, 192.168.1.255, 192.168.1.1, 192.168.1.255, 224.0.0.251}
+:addr in [ 192.168.1.1, 192.168.1.255, 192.168.1.255, 192.168.1.255, 192.168.1.255, 192.168.1.255, 192.168.1.255, 192.168.1.1, 192.168.1.255, 192.168.1.1, 192.168.1.1, 192.168.1.1, 224.0.0.251, ff02::fb, 192.168.1.255, 192.168.1.255, 192.168.1.255, 192.168.1.1, 192.168.1.255, 224.0.0.251]
```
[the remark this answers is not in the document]
mavam: The old `set` never enforced uniqueness, so now that this is a `vector`, it's the exact same behavior.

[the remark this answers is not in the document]
mavam: But now that you mention it, we probably should have a story to perform a "uniquing" pass when using `in` with `vector` operands.

libvast/test/parse_data.cpp
mavam: Good point, it's now in there.

CHANGELOG.md
mavam: Let me know if the current changelog suits the bill.

dominiklohmann: It's looking mighty fine to me. Covers everything I'd want from a changelog entry as a user: Simple explanation of the change, and actionable steps for adapting my setup.

libvast_test/src/table_slices.cpp
mavam: Got it; changed.

PR #1013, opened 2020-08-07
Make get_node_components usable for runtime labels

tobim:


libvast/vast/system/node_control.hpp
```diff
@@ -38,7 +39,7 @@ spawn_at_node(caf::scoped_actor& self, caf::actor node, Arguments&&... xs) {
 template <size_t N>
 caf::expected<std::array<caf::actor, N>>
 get_node_components(caf::scoped_actor& self, caf::actor node,
-                    const char* const (&names)[N]) {
+                    const std::string_view (&names)[N]) {
```
dominiklohmann, line 42:
```suggestion
                    const std::array<std::string_view, N> names) {
```

tobim: Brace-enclosed initializer lists don't convert to this type automatically, so it would require extra annotation at the call site. Not desirable imho.

dominiklohmann: Alrighty.

PR #1014, opened 2020-08-07
Support bracketed IPv6 notation

dominiklohmann:


dominiklohmann:
Some thought: This may also be fixed by preferring `vector` over `address` in the `data` parser. It still leaves `[::ff80]` valid, because it interprets the brackets as part of the vector.

I need to think this through a bit. Let's revisit this on Monday.

dominiklohmann:
@tobim and I decided not to pursue this. On current master, the parser for `vast::endpoint` is gone, which removes the need for this.

PR #1015, opened 2020-08-10
Log strerror if file open fails

tobim:


dominiklohmann:
Does it make sense to do this for all places where we use `vast::ec::filesystem_error`? `git grep filesystem_error` shows a few more, most notably in `chunk.cpp`.

tobim:
I guess you're right.

tobim:
@dominiklohmann most low-level calls are encapsulated in the posix abstractions. Changing these means the return value have to be converted to `caf::expected`. I think this warrants a dedicated story.

PR #1016, opened 2020-08-10
Rename vector to list

mavam:


libvast/test/arrow_table_slice.cpp
```diff
@@ -83,7 +83,7 @@ TEST(manual table slice building) {
   // Our schema is `map: list<key: int64, value: int64>, int: int64`.
   auto kvp_fields = std::vector{field("key", int64()), field("value", int64())};
   auto kvp_record = struct_(kvp_fields);
-  auto map_record = list(kvp_record);
+  auto map_record = arrow::list(kvp_record);
```
dominiklohmann, line 86: Can you remove the `using namespace arrow` from this scope entirely while you're at it?

libvast/test/event.cpp
```diff
@@ -45,7 +45,7 @@ struct fixture {
   }
 
   type t;
-  vector r;
+  std::vector<data> r;
```
dominiklohmann, line 48:
```suggestion
  list r;
```

libvast/test/event.cpp
```diff
@@ -55,7 +55,7 @@ FIXTURE_SCOPE(event_tests, fixture)
 
 TEST(basics) {
   CHECK_EQUAL(e.type().name(), "foo");
-  REQUIRE(caf::holds_alternative<vector>(e.data()));
+  REQUIRE(caf::holds_alternative<std::vector<data>>(e.data()));
```
dominiklohmann, line 58:
```suggestion
  REQUIRE(caf::holds_alternative<list>(e.data()));
```

mavam: Here, `list` is the wrong way to think about this. An event is a sequence of `data`, not an instance of `list`. This is why I changed it to `std::vector<data>`. It'll change with the `record` change in `data` as well.

libvast/test/event.cpp
mavam: See my comment below for why going with `std::vector<data>`.

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- ⚠️ The `vector` type has been renamed to `list`.In an effort to streamline
+  the type system vocabulary, we favor `list` over `vector` because it's closer
+  to existing terminology (e.g., Apache Arrow) and easier to type. This change
+  requires updating existing schemas by changing `vector<T>` to `list<T>`. 
+  [#1016](https://github.com/tenzir/vast/pull/1016)
```
dominiklohmann, line 18:
```suggestion
- ⚠️ The `vector` type has been renamed to `list` .In an effort to streamline
  the type system vocabulary, we favor `list` over `vector` because it's closer
  to existing terminology (e.g., Apache Arrow). This change requires updating
  existing schemas by changing `vector<T>` to `list<T>`. 
  [#1016](https://github.com/tenzir/vast/pull/1016)
```

libvast/test/event.cpp
dominiklohmann: I think this is just unnecessarily confusing. A `list` now is exactly a sequence of `data`, because that is quite literally its definition.

mavam: Yes, but it's a value *for a single field*. Here, we are modeling a *sequence of values*, each of which represent a single field.

dominiklohmann: I still disagree with you, but not not enough to argue about this one.

mavam: Fair enough, it'll change soon anyway once `record` is there.

PR #1018, opened 2020-08-10
Close file descriptor by default in 'vast::file'

lava:
The change to "not close by default" was introduced in this commit, I'm not sure if it was intentional or just an oversight.

```
commit c01b22917f3b10fdea6d40ad3d9981dcf36693f3
Author: Matthias Vallentin <vallentin@icir.org>
Date:   Thu May 14 09:39:27 2015 -0700

    Fix bugs in filesystem interface
    
    Also, factor more POSIX wrappers.
```

### DIFF
--- a/libvast/vast/filesystem.hpp
+++ b/libvast/vast/filesystem.hpp
@@ -162,8 +162,6 @@ private:
   std::string str_;
 };
 
-constexpr bool close_on_destruction = true;
-
 /// A simple file abstraction.
 class file {
   file(const file&) = delete;
@@ -195,11 +193,11 @@ public:
   /// Constructs a file from the OS' native file handle type.
   /// @param handle The file handle.
   /// @param p The file path.
-  /// @param close_behavior Whether to close or leave open the file handle upon
+  /// @param close_on_destruction Whether to close or leave open the file handle
+  /// upon
   ///                       destruction.
   /// @pre The file identified via *handle* is open.
-  file(native_type handle, bool close_behavior = close_on_destruction,
-       vast::path p = {});
+  file(native_type handle, bool close_on_destruction = true, vast::path p = {});
 
   file(file&&) = default;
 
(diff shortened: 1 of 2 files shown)


libvast/vast/filesystem.hpp
```diff
@@ -253,7 +251,7 @@ class file {
 
 private:
   native_type handle_;
-  bool close_on_destruction_ = !close_on_destruction;
+  bool close_on_destruction_ = true;
```
tobim, line 254: Why did you not remove the member outright?

libvast/vast/filesystem.hpp
```diff
@@ -195,11 +193,11 @@ class file {
...
   /// @param handle The file handle.
   /// @param p The file path.
-  /// @param close_behavior Whether to close or leave open the file handle upon
+  /// @param close_on_destruction Whether to close or leave open the file handle
+  /// upon
   ///                       destruction.
   /// @pre The file identified via *handle* is open.
-  file(native_type handle, bool close_behavior = close_on_destruction,
-       vast::path p = {});
+  file(native_type handle, bool close_on_destruction = true, vast::path p = {});
```
dominiklohmann, line 200:
```suggestion
  explicit file(native_type handle, bool close_on_destruction = true, vast::path p = {});
```

libvast/vast/filesystem.hpp
lava: As far as I can tell, we can go even further and remove the `native_type` handling completely, including this variable and the constructor.

But I'd like to confirm with @mavam before doing that since he's the original author. So maybe we can merge the "safe" fix now and leave a TODO for the refactoring?

mavam: I think in the scope of this fix, we can remove the member entirely.

mavam: Regarding removing `native_type`, let's wait with that because I don't think it's unreasonable to request the file descriptor from a file abstraction, e.g., to call `flock(2)` or other things on it.

PR #1021, opened 2020-08-17
Update nix dependencies to latest master

lava:
Update dependencies for the static build after switching to a custom caf upstream.

The changes were generated by running the `nix/update.sh` script.

### DIFF
--- a/nix/broker/source.json
+++ b/nix/broker/source.json
@@ -7,5 +7,5 @@
   "fetchSubmodules": true,
   "deepClone": false,
   "leaveDotGit": false,
-  "version": "current"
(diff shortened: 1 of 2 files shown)


dominiklohmann:
Should we be running this from a GitHub Workflow automatically whenever we update the submodule?

lava:
Ideally yes, but but right now I don't have the free cycles to invest in creating and testing a new CI workflow for that.

dominiklohmann:
We have this workflow already for VAST Pro, I'll copy it over when I have some free cycles. I agree that we shouldn't do that _right now_.

PR #1022, opened 2020-08-18
Add an io::save utility function

tobim:


libvast/src/detail/posix.cpp
```diff
@@ -254,21 +255,23 @@ bool read(int fd, void* buffer, size_t bytes, size_t* got) {
...
     ssize_t written;
     do {
       written = ::write(fd, buf + total, bytes - total);
     } while (written < 0 && errno == EINTR);
-    if (written <= 0)
-      return false;
+    if (written < 0)
+      return make_error(ec::filesystem_error, "write(2)", std::strerror(errno));
+    if (written == 0)
+      return make_error(ec::filesystem_error, "write(2) returned 0");
```
mavam, line 269: Could we add a comment here why returning 0 is a special error case?

Unrelated to this change, I wonder what one should do for 0? Try again? Is the combo of 0 and `EINTR` valid?

libvast/src/file.cpp
```diff
@@ -102,8 +102,10 @@ bool file::read(void* sink, size_t bytes, size_t* got) {
   return is_open_ && detail::read(handle_, sink, bytes, got);
 }
 
-bool file::write(const void* source, size_t bytes, size_t* put) {
-  return is_open_ && detail::write(handle_, source, bytes, put);
+caf::error file::write(const void* source, size_t bytes, size_t* put) {
```
mavam, line 105:
```suggestion
[[nodiscard]] caf::error file::write(const void* source, size_t bytes, size_t* put) {
```

mavam: Also, all `void*` + `size_t` APIs predate the `span<bytes>` era. The right API for this function in my opinion looks like this:

```cpp
[[nodiscard]] caf::expected<size_t> file::write(span<const byte> bytes);
```

However, I think we should rather have a separate refactoring story that goes through all APIs of this kind and do the replacement.

libvast/src/io/save.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "vast/error.hpp"
+#include "vast/io/write.hpp"
+#include "vast/path.hpp"
+
+#include <cstdio>
+
+namespace vast::io {
+
+caf::error save(const path& filename, span<const byte> xs) {
+  auto tmp = filename + ".save";
```
mavam, line 25:
```suggestion
  auto tmp = '.' + filename + ".save";
```

libvast/vast/io/save.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include "vast/byte.hpp"
+#include "vast/fwd.hpp"
+#include "vast/span.hpp"
+
+#include "caf/fwd.hpp"
+
+namespace vast::io {
+
+/// Safely writes a buffer to a file on disk.
```
mavam, line 22: The reader doesn't know how to interpret "safely" here. Please provide an adequate description of what happens.

libvast/src/file.cpp
```diff
@@ -0,0 +1,129 @@
...
+file::file(vast::path p) : path_{std::move(p)} {
+}
+
+file::~file() {
+  // Don't close stdin/stdout implicitly.
+  if (path_ != "-")
+    close();
+}
+
+caf::expected<void> file::open(open_mode mode, bool append) {
```
dominiklohmann, line 40: We recently had a discussion for a similar set of functions about `caf::expected<void>` vs. `caf::error`, where the latter won the vote. Should we change this to use `caf::error` as well?

libvast/vast/file.hpp
```diff
@@ -0,0 +1,111 @@
...
+/// The native type of a file.
+#ifdef VAST_POSIX
+  typedef int native_type;
+#else
+  typedef void native_type;
+#endif
+
+  /// The mode in which to open a file.
+  enum open_mode {
+    invalid,
```
dominiklohmann, line 41:
```suggestion
  enum class open_mode {
```

libvast/vast/file.hpp
```diff
@@ -0,0 +1,111 @@
...
+  caf::error write(const void* source, size_t size, size_t* put = nullptr);
+
+  /// Seeks the file forward.
+  /// @param bytes The number of bytes to seek forward relative to the current
+  ///              position.
+  /// @returns `true` on success.
+  bool seek(size_t bytes);
+
+  /// Retrieves the ::path for this file.
+  /// @returns The ::path for this file.
```
dominiklohmann, line 97:
```suggestion
  /// Retrieves the path for this file.
  /// @returns The path for this file.
```

The `::` here is misleading, since the function returns `::vast::path`, not `::path`.

libvast/src/detail/posix.cpp
```diff
@@ -254,21 +255,23 @@ bool read(int fd, void* buffer, size_t bytes, size_t* got) {
...
   auto buf = reinterpret_cast<const uint8_t*>(buffer);
   while (total < bytes) {
     ssize_t written;
     do {
       written = ::write(fd, buf + total, bytes - total);
     } while (written < 0 && errno == EINTR);
-    if (written <= 0)
-      return false;
+    if (written < 0)
+      return make_error(ec::filesystem_error, "write(2)", std::strerror(errno));
```
dominiklohmann, line 267:
```suggestion
      return make_error(ec::filesystem_error, "write(2):", std::strerror(errno));
```

dominiklohmann (commented):
I had some minor remarks. Leaving the actual review to @mavam.

libvast/src/io/save.cpp
```diff
@@ -0,0 +1,33 @@
...
+  if (auto err = write(tmp, xs))
+    return err;
+  if (std::rename(tmp.str().c_str(), filename.str().c_str()) != 0)
+    return make_error(ec::filesystem_error, "rename(2)", std::strerror(errno));
```
mavam, line 29: What happens to temp file when either `write` or `rename` fails?

libvast/src/file.cpp
mavam: Yes, that would be the right thing to do.

libvast/vast/file.hpp
mavam: Let's not introduce changes in this PR that touch a lot of downstream code. This is a reasonable change in general (but then it would make more sense to hoist it into the top-level namespace and call it `file_open_mode`) but stretches the scope here a bit too much.

libvast/src/detail/posix.cpp
tobim: According to the [man page](https://linux.die.net/man/2/write) `errno` may or may not be set if `write` returns 0. This return value should only be possible if `bytes` is 0, but in that case we don't enter the outer loop. The second check exists for totality, but I don't expect this to ever happen in practice.

libvast/src/file.cpp
dominiklohmann: I think we should tackle these as we go. A story like that will never be prioritized.

libvast/src/file.cpp
dominiklohmann: Similar to the `span`ification, I think we should tackle these as we go. These kinda stories never get priority.

libvast/src/system/connect_to_node.cpp
```diff
@@ -31,9 +31,8 @@
 #include "vast/defaults.hpp"
 #include "vast/endpoint.hpp"
 #include "vast/error.hpp"
-#include "vast/filesystem.hpp"
 #include "vast/logger.hpp"
-#include "vast/system/node.hpp"
+//#include "vast/system/node.hpp"
```
tobim, line 35:
```suggestion
```

libvast/src/file.cpp
tobim: This and the other similar suggestions would be nice to have but not of the same priority as the changes introduced by the existing commits. I would prefer to focus on delivery for now.

libvast/src/io/save.cpp
tobim: I am strongly against hiding temporary files, their presence should be easily detectable because they usually require admin attention.

libvast/src/detail/posix.cpp
mavam: Your comment here makes sense, but the code doesn't check for `bytes == 0 && written == 0`.

libvast/src/io/save.cpp
mavam: Looks good now.

libvast/src/detail/posix.cpp
tobim: `written` is initialized to 0, so if the function is called with `bytes` set to 0 as well, we don't even enter the outer loop.

tobim:
@mavam CI is fixed now.

libvast/vast/io/save.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include "caf/fwd.hpp"
+
+namespace vast::io {
+
+/// Writes a buffer to a temporary file and atomically renames it to *filename*
+/// afterwards.
+/// @param filename The file to write to.
+/// @param xs The buffer to read from.
+/// @returns An error if the operation failed.
+caf::error save(const path& filename, span<const byte> xs);
```
mavam, line 27:
```suggestion
[[nodiscard]] caf::error save(const path& filename, span<const byte> xs);
```

libvast/vast/detail/posix.hpp
```diff
@@ -122,16 +124,17 @@ bool close(int fd);
...
 
 /// Wraps `write(2)`.
 /// @param fd The file descriptor to write to.
 /// @param buffer The buffer to read from.
 /// @param bytes The number of bytes to write into *fd* from *buffer*.
 /// @param put If not-nullptr, receives the number of bytes actually read.
-/// @returns `true` on successful reading.
-bool write(int fd, const void* buffer, size_t bytes, size_t* put = nullptr);
+/// @returns `caf::none` on successful write.
+caf::error
```
mavam, line 136:
```suggestion
[[nodiscard]] caf::error
```

libvast/src/io/save.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include "vast/error.hpp"
+#include "vast/io/write.hpp"
+#include "vast/path.hpp"
+
+#include <cstdio>
+
+namespace vast::io {
+
+caf::error save(const path& filename, span<const byte> xs) {
```
mavam, line 24:
```suggestion
[[nodiscard]] caf::error save(const path& filename, span<const byte> xs) {
```

libvast/src/file.cpp
mavam, line 105:
```suggestion
[[nodiscard]] caf::error file::write(const void* source, size_t bytes, size_t* put) {
```

mavam (approved):
Looks good.

I've sprinkled a few more `[[nodiscard]]` attributes over the higher-level functions. Feel free to make this a separate story or do it here right away.

libvast/src/io/save.cpp
dominiklohmann: The [[nodiscard]] needs to only be at the declaration.

tobim:
> I've sprinkled a few more `[[nodiscard]]` attributes over the higher-level functions. Feel free to make this a separate story or do it here right away.

I believe the right way to do this is to add `[[nodiscard]]` to the definition of `caf::error` itself. This is not an option for CAF 0.17 though.

dominiklohmann:
> I believe the right way to do this is to add `[[nodiscard]]` to the ~definition~ declaration of `caf::error` itself. This is not an option for CAF 0.17 though.

`s/definition/declaration`

That's already happening for CAF 0.18. For now, it doesn't hurt to add these in.

tobim:
> That's already happening for CAF 0.18. For now, it doesn't hurt to add these in.

Alrighty.

PR #1023, opened 2020-08-19
Remove unnecessary rlimit adjustments on macOS

dominiklohmann:


mavam (approved):
Looks good content-wise.

It would be nice to get an understanding why this is not longer necessary. What's the current number of default descriptors?

dominiklohmann:
> It would be nice to get an understanding why this is not longer necessary. What's the current number of default descriptors?

Whatever `launchctl limit` returns on macOS for `maxfiles`. On my system, that's 256 soft limit, unlimited hard limit. I'm not quite sure why this was necessary in the first place. It's the job of the launchd configuration file (e.g., `~/Library/LaunchAgents/com.tenzir.vast.plist`) to set the resource limits.

mavam:
256 sounds awefully low and was probably the reason why I introduced this adjustment in the first place. Before merging, I'd like to get a bit more extensive tests that this doesn't re-introduce a regression.

dominiklohmann:
Merging after quick discussion via Slack.

PR #1025, opened 2020-08-21
Correct the use of ::read()

tobim:


libvast/test/filesystem.cpp
```diff
@@ -181,3 +188,36 @@ TEST(file_and_directory_manipulation) {
   CHECK(rm(p.parent()));
   CHECK(!p.parent().is_directory());
 }
+
+FIXTURE_SCOPE(chunk_tests, fixtures::filesystem)
+
+// This test takes almost one minute on macOS 10.14.
```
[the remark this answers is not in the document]
tobim: In general with APFS on my office iMac.

tobim:
@mavam As discussed I figured out how to use the level 2 API correctly on macOS. The code can now be reviewed again.

mavam (approved):
I like the current solution.

One last thing: please add a changelog entry about the bugfix since this was a user-reported error.

CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 Writing files larger than 2 GB on macOS has been corrected.
+  [#1025](https://github.com/tenzir/vast/pull/1025)
+
+- 🐞 Reading files larger than 2 GB has been corrected.
+  [#1025](https://github.com/tenzir/vast/pull/1025)
+
```
mavam, line 19:
```suggestion
- 🐞 Incomplete reads have not been handled properly, which manifested for
  files larger than 2GB. On macOS, writing files larger than 2GB may have
  failed previously. VAST now respects OS-specific constraints on the maximum
  block size.
  [#1025](https://github.com/tenzir/vast/pull/1025)
```

mavam (changes requested):

libvast/vast/detail/posix.hpp
```diff
@@ -109,23 +109,20 @@ struct [[nodiscard]] unix_domain_socket {
 /// @returns `caf::none` on successful closing.
 [[nodiscard]] caf::error close(int fd);
 
-/// Wraps `read(2)`.
+/// Wraps `fread(3)`.
```
dominiklohmann, line 112: This should be reverted I think.

libvast/vast/file.hpp
```diff
@@ -76,17 +76,14 @@ class file {
   /// Reads a given number of bytes in to a buffer.
   /// @param sink The destination of the read.
   /// @param size The number of bytes to read.
-  /// @param got The number of bytes read.
   /// @returns `caf::none` on success.
-  [[nodiscard]] caf::error read(void* sink, size_t size, size_t* got = nullptr);
+  [[nodiscard]] caf::error read(void* sink, size_t size);
```
dominiklohmann, line 80: Is there a specific reason why we're discarding the number of bytes read here? We could just return `expected<size>` from this function and handle the error downstream, which is usually more appropriate.

mavam: Yeah, this looks like an oversight to me.

lava (commented):
I hate to be piling on here, but imho `read/write` are not good names for these functions. There *is* already a well-known function called `read(int, void*, size_t)` that everyone is familiar with, and from looking at the name and signature I'd expect a function called `posix::read()` to be just a thing wrapper around that. As is, everybody glancing at that will assume there is a bug in the calling code because we dont retry in case of a partial read.

I dont mind committing this now to get the bugfix out to DCSO, but imho renaming this to something like `read_all()`/`safe_read()`/`read_n()`/etc. (that can maybe call the actual read internally) should be the near-term goal.

lava:
Addendum to the above: I just realized that the API change is bigger than I initially thought. Usually one passes an **upper bound** to `read()`, i.e. the size of the buffer. After the changes here, the size also becomes a lower bound, i.e. one has to know exactly how many bytes are to be read in order to call this function.

tobim:
> After the changes here, the size also becomes a lower bound, i.e. one has to know exactly how many bytes are to be read in order to call this function.

That is not the case, see https://github.com/tenzir/vast/pull/1025/files#diff-4d6b0fc0d187a8a24760ed369528e14fR272-R273.

libvast/vast/file.hpp
tobim: I followed your suggestion, it was reverted to return the number of bytes that were read.

PR #1026, opened 2020-08-24
Don't overwrite index state after startup error

lava:
Don't overwrite the on-disk state if we fail to restore it, since
the newly written state will surely be empty and overwrite the
existing VAST data on disk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 VAST would overwrite existing on-disk state data when encountering a partial
+  read during startup. This state-corrupting behavior no longer exists.
+  [#1026](https://github.com/tenzir/vast/pull/1026)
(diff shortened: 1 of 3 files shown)


dominiklohmann:
I think this warrants a changelog entry, doesn't it?

lava:
[ch18989]

dominiklohmann (approved):

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 Fixed a bug where VAST would overwrite existing on-disk state data
+  when encountering a partial read during startup.
+  [#1026](https://github.com/tenzir/vast/pull/1026)
```
```suggestion
- 🐞 VAST would overwrite existing on-disk state when encountering a partial
  read during startup. [#1026](https://github.com/tenzir/vast/pull/1026)
```

mavam (dismissed):

PR #1027, opened 2020-08-24
Make shutdown-grace-period available at CLI

mavam:


libvast/src/system/application.cpp
```diff
@@ -98,7 +98,10 @@ auto make_root_command(std::string_view path) {
                                         "definitions")
         .add<std::string>("aging-frequency", "interval between two aging "
                                              "cycles")
-        .add<std::string>("aging-query", "query for aging out obsolete data");
+        .add<std::string>("aging-query", "query for aging out obsolete data")
+        .add<std::string>("shutdown-grace-period",
+                          "time to wait until component shutdown "
+                          "finishes cleanly");
```
dominiklohmann, line 104: I'm fine with this as is, but I also wouldn't mind making clear that the second shutdown mode starts after this period.

```suggestion
                          "timeout for graceful shutdowns; after this period, the server forces a shutdown");
```

Modulo formatting.

dominiklohmann (approved):
See comment. Approving as is, but feel free to improve as I mentioned.

libvast/src/system/application.cpp
mavam: I'm reluctant to put a lot of detail in the option description. I'll add a condensed version.

We all know that we need a short and long description. Currently, the long description is the comment in `vast.conf`.

dominiklohmann: > Currently, the long description is the comment in `vast.conf`.

That's entirely untrue. Most of the comments in `vast.conf` are straight up copies from the option descriptions.

The long description should go into the documentation for the subcommand, if anywhere.

I'm also not sure why we're adding a limit to these. It doesn't hurt to document explicitly.

dominiklohmann: Anyways, I'm fine with your suggestion. 👍

PR #1031, opened 2020-08-27
Prepare changelog for August 2020 release

dominiklohmann:


pyvast/setup.py
```diff
@@ -33,5 +33,5 @@
     python_requires=">=3.7",
     setup_requires=["setuptools", "wheel"],
     url="https://github.com/tenzir/vast",
-    version="2020.07.28",
+    version="2020.08.28",
```
mavam, line 36: Do we have documentation or a script that `setup.py` needs to be updated here manually? Maybe a `prepare-for-release` script that takes a version/date could do this? Not that it would save much time, but it would help making mistakes.

mavam (approved):

pyvast/setup.py
dominiklohmann: There really isn't a way to do this nicely for Python. For VAST itself, we just get the information from `git describe --tags` in CMake.

It's not a big deal if we forget to change this, we can always adapt it later.

tobim (approved):

PR #1034, opened 2020-08-28
Render duration and port as JSON strings

dominiklohmann:
This a) avoids loss of information and b) makes it such that the user can directly use the output value in another query.


libvast/vast/concept/printable/vast/json.hpp
```diff
@@ -78,13 +79,6 @@ struct json_printer : printer<json_printer<TreePolicy, Indent, Padding>> {
       return caf::visit(*this, x);
     }
 
-    bool operator()(const view<port>& x) {
-      // TODO: convert() generates a string, while our unit tests expect a
-      //       number. Eventually, we're probably going to drop the protocol
-      //       part but convert() and print() should be consistent regardless.
```
dominiklohmann, line 84: Note for the reviewer: `convert()` has been able to convert strings to duration/port for a while now.

tobim (approved):
Let's merge this now, even though the port type will eventually be removed.

PR #1035, opened 2020-09-01
Make periodic logging more sensible

tobim:


libvast/src/system/importer.cpp
```diff
@@ -126,6 +127,18 @@ void importer_state::send_report() {
     auto node_throughput = measurement{elapsed, measurement_.events};
     auto r = performance_report{
       {{"importer"s, measurement_}, {"node_throughput"s, node_throughput}}};
+#if VAST_LOG_LEVEL >= VAST_LOG_LEVEL_VERBOSE
+    auto beat = [&](performance_sample s) {
```
dominiklohmann, line 131:
```suggestion
    auto beat = [&](const auto& sample) {
```

We shouldn't needlessly copy here.

libvast/vast/defaults.hpp
```diff
@@ -352,7 +352,7 @@ constexpr size_t initially_requested_ids = 128;
 
 /// Rate at which telemetry data is sent to the ACCOUNTANT.
 constexpr std::chrono::milliseconds telemetry_rate
-  = std::chrono::milliseconds{1000};
+  = std::chrono::milliseconds{10000};
```
dominiklohmann, line 355: I think this deserves a changelog entry.

dominiklohmann (changes requested):
Just a few nits.

What about the index log message counterpart to the archive? Should we move it to debug, too?

dominiklohmann (approved):
LGTM!

PR #1037, opened 2020-09-02
Add an option to disable meta_index writes

tobim:


nix/vast/default.nix
```diff
@@ -17,6 +17,7 @@
 , python3
 , jq
 , tcpdump
+, utillinux
```
mavam, line 20: What's this?

dominiklohmann: A small fix that enables colored log output in the Nix builds. We should cherry-pick it into master.

mavam: Yes please.

libvast/src/system/application.cpp
```diff
@@ -64,7 +64,9 @@ command::opts_builder add_index_opts(command::opts_builder ob) {
                                             "partitions")
     .add<size_t>("max-taste-partitions", "maximum number of immediately "
                                          "scheduled partitions")
-    .add<size_t>("max-queries,q", "maximum number of concurrent queries");
+    .add<size_t>("max-queries,q", "maximum number of concurrent queries")
+    .add<bool>("disable-recoverability", "don't sync meta-index for every new "
+                                         "partipartition");
```
lava, line 69: `s/participation/partition/`

libvast/src/system/index.cpp
```diff
@@ -97,9 +97,9 @@ index_state::~index_state() {
     flush_to_disk();
 }
 
-caf::error
-index_state::init(const path& dir, size_t max_partition_size,
-                  uint32_t in_mem_partitions, uint32_t taste_partitions) {
+caf::error index_state::init(const path& dir, size_t max_partition_size,
+                             uint32_t in_mem_partitions,
+                             uint32_t taste_partitions, bool yolo_mode) {
```
lava, line 102: `s/yolo_mode/delay_flush_until_shutdown/`

lava (approved):
I know this wasnt meant to be committed, but on reviewing this PR I don't really see why not. We can (in principle) recover the index state from the partition data, so delaying the write at the cost of potentially having to manually restore it after unclean shutdowns seems like a trade-off that can be reasonable in some situations.

Especially since the current way is not extremely robust either, when the process gets aborted while writing the index state we end up in the same situation where the index state and the on-disk partitions are out of sync.

libvast/src/system/application.cpp
dominiklohmann:
```suggestion
                                         "partition");
```

libvast/src/system/index.cpp
dominiklohmann:
```suggestion
                             uint32_t taste_partitions,
                             bool delay_flush_until_shutdown) {
```

PR #1038, opened 2020-09-02
Fix the VAST Static README badge

dominiklohmann:
It doesn't make sense to show this badge for repository dispatch events, so we show it for push events only now.


mavam (approved):
🙏

PR #1040, opened 2020-09-03
Merge contents of all configuration files

dominiklohmann:
Please test this on Linux in review.


CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🎁 VAST now merges the contents of all used configuration files instead of
+  using only the most user-specific file.
```
mavam, line 15: Can you please elaborate about the merge order explicitly? It's only intuitively clear, but I think we should be precise how this happens.

libvast/src/system/configuration.cpp
```diff
@@ -51,22 +51,20 @@ void initialize_factories() {
 
 configuration::configuration() {
   detail::add_message_types(*this);
-  // Use 'vast.conf' instead of generic 'caf-application.ini'.
-  auto config_paths = std::vector<path>{};
+  // We must clear the config_file_path first so it does not use
+  // `caf-application.ini` as fallback. Instead we use our own `config_paths`
+  // variable to support multiple configuration files.
```
mavam, line 56: This comment doesn't make sense, especially the "Instead..." part because the first sentence is not a negative one. Could you clarify?

CHANGELOG.md
```diff
@@ -12,7 +12,10 @@ Every entry has a category for which we use the following visual abbreviations:
 ## Unreleased
 
 - 🎁 VAST now merges the contents of all used configuration files instead of
-  using only the most user-specific file.
+  using only the most user-specific file. The file specified using `--config`
+  takes the highest precedence, followed by
+  `${XDG_CONFIG_HOME:-${HOME}/.config}/vast/vast.conf` and
```
```suggestion
  `${XDG_CONFIG_HOME:-${HOME}/.config}/vast/vast.conf` and the compile-time path
```

mavam (approved):

PR #1046, opened 2020-09-07
Add man page header/footer

satta:
This PR ensures that the `manual` subcommand also adds the necessary sections for pandoc to create a full manpage, i.e. `NAME`, section, etc. The main text is taken from the existing (but now seemingly unused) man page template in `doc/`.


libvast/src/command.cpp
```diff
@@ -211,6 +211,32 @@ void doctext(const command& cmd, std::ostream& out) {
   out << std::left << cmd.documentation;
 }
 
+void manheader(std::ostream& os) {
```
mavam, line 214:
```suggestion
void manheader(std::ostream& out) {
```

libvast/src/command.cpp
```diff
@@ -211,6 +211,32 @@ void doctext(const command& cmd, std::ostream& out) {
...
+"`vast` -- manage a VAST topology\n"
+"\n"
+"# OVERVIEW\n"
+"\n"
+"This section describes the VAST system and its components from a user "
+"interaction point of view.\n"
+"\n";
+}
+
+void manfooter(std::ostream& os) {
```
mavam, line 229:
```suggestion
void manfooter(std::ostream& out) {
```

libvast/src/command.cpp
```diff
@@ -211,6 +211,32 @@ void doctext(const command& cmd, std::ostream& out) {
   out << std::left << cmd.documentation;
 }
 
+void manheader(std::ostream& os) {
+  os << "% VAST(1)\n"
+"% Tenzir GmbH\n"
+"\n"
+"# NAME\n"
+"\n"
+"`vast` -- manage a VAST topology\n"
```
mavam, line 220:
```suggestion
"`vast` -- manage a VAST node\n"
```

libvast/src/command.cpp
```diff
@@ -211,6 +211,32 @@ void doctext(const command& cmd, std::ostream& out) {
...
+"\n"
+"# OVERVIEW\n"
+"\n"
+"This section describes the VAST system and its components from a user "
+"interaction point of view.\n"
+"\n";
+}
+
+void manfooter(std::ostream& os) {
+  os << "# ISSUES\n"
```
mavam, line 230:
```suggestion
  out << "# ISSUES\n"
```

libvast/src/command.cpp
```diff
@@ -211,6 +211,32 @@ void doctext(const command& cmd, std::ostream& out) {
   out << std::left << cmd.documentation;
 }
 
+void manheader(std::ostream& os) {
+  os << "% VAST(1)\n"
```
mavam, line 215:
```suggestion
  out << "% VAST(1)\n"
```

mavam, line 215:
```suggestion
  out << "% VAST(1)\n"
```

mavam (changes requested):
Thanks for the patch! Looks good overall. I just changed `os` to `out`, because all other functions in that file use this naming.

One more general question. Should `vast` be in section 1 or 8? When you write `vast start`, it's effectively a daemon, but otherwise it's more like a control command.

satta:
> Thanks for the patch! Looks good overall. I just changed `os` to `out`, because all other functions in that file use this naming.

Okay, great!

> One more general question. Should `vast` be in section 1 or 8? When you write `vast start`, it's effectively a daemon, but otherwise it's more like a control command.

I would suggest section 1. In general, it's an application command, with the daemon just being a special case.

mavam:
> I would suggest section 1. In general, it's an application command, with the daemon just being a special case.

Makes sense. Let's go with section 1 then.

dominiklohmann:
I removed some leftover scraps from our CMake while reviewing this, and re-formatted the code.

From my side this is ready to go, but I'd like to see the content reviewed by both @satta and @mavam.

After building the `vast` target, iff `pandoc` is available, the manpage is generated as `<build_dr>/vast/vast.1`.

After installing, the man-page is available as `<prefix>/share/man/man1/vast.1`.

libvast/src/command.cpp
```diff
@@ -211,17 +211,45 @@ void doctext(const command& cmd, std::ostream& out) {
...
+         "\n"
+         "# NAME\n"
+         "\n"
+         "`vast` -- manage a VAST node\n"
+         "\n"
+         "# OVERVIEW\n"
+         "\n"
+         "This section describes the VAST system and its components from a "
+         "user "
+         "interaction point of view.\n"
```
mavam, line 226: @dominiklohmann can you reformat this again? Looks like `clang-format` wasn't doing the job right.

mavam (commented):
I have no more structural changes. The content isn't quite there yet. I think we should adopt a [common layout](https://en.wikipedia.org/wiki/Man_page#Layout) with NAME, SYNOPSIS, DESCRIPTION, etc.

@satta I'm happy to add a commit to this branch with the content tweaks. Any objections?

satta:
> @satta I'm happy to add a commit to this branch with the content tweaks. Any objections?

No, go ahead! :) I only added the bare minimum, of course you can add more. The reason I didn't add the synopsis, for example,  is that the same information is already in the generated text below, and I didn't want any duplication. If you have a way of fine-tuning this, even better!

mavam (approved):

mavam:
I went ahead and merged this for now. We'll do the layout refactoring in a follow-up PR.

PR #1049, opened 2020-09-08
Pass configured log level to CAF build

lava:


lava:
Note: this shouldnt be merged yet, as a follow-up I discovered that the component blacklisting doesnt properly work by default, so with this change the CAF debug messages will always spam the server log.

dominiklohmann (approved):
I'd be fine with merging this as is. Tested it locally and it seems to work as expected.

PR #1050, opened 2020-09-08
Fix script to find hanging CAF actors from log file

lava:
* Match 'caf_flow' instead of 'caf.flow'
* Use exact match for ids (to prevent id 16 matching 163)
* Print all IDs to make it possible to detect id reusage


mavam (approved):
I didn't test this locally, but if it does the job for you, we can merge this just fine.

PR #1051, opened 2020-09-08
Clarify comment on CWG issue 2137

dominiklohmann:
See also my post on r/cpp regarding this issue: https://www.reddit.com/r/cpp/comments/iotebm


mavam (approved):
Nice comment. It could make sense to move this to the top of `data.hpp`, so that every dev reads this first. In here, it's somewhat buried and hard to appreciate.

dominiklohmann:
I've moved it to the definition of data.hpp using the `@note` annotation.

PR #1052, opened 2020-09-09
Fix status command across process boundaries

dominiklohmann:


lava (approved):

libvast/src/detail/fill_status_map.cpp
```diff
@@ -21,7 +23,18 @@
...
+#include <sstream>
+#include <thread>
+
+namespace {
+
+std::string thread_id() {
+  std::stringstream ss;
+  ss << std::this_thread::get_id();
+  return std::move(ss).str();
+}
```
lava, line 35: Isn't this just `std::to_string(thread_id)`?

dominiklohmann: Nope, the only way to convert it is by going through `operator<<(std::ostream&, const std::thread::id&)`. See also: https://en.cppreference.com/w/cpp/thread/thread/id

PR #1053, opened 2020-09-10
Add pthread-specific TID to debug status

dominiklohmann:


lava (approved):

mavam (commented):
Should this maybe go into `vast::detail::posix`?

dominiklohmann:
> Should this maybe go into `vast::detail::posix`?

I don't see us needing this anywhere else, as this is super specific. Let's just keep it in the anonymous namespace locally. We can always expose it later on should we need the function elsewhere.

PR #1055, opened 2020-09-10
Enable settings conversion to CAF

mavam:


libvast/vast/concept/convertible/to.hpp
```diff
@@ -31,10 +32,20 @@ template <class To, class From, class... Opts>
 auto to(From&& from, Opts&&... opts)
   -> std::enable_if_t<is_convertible<std::decay_t<From>, To>{},
                       caf::expected<To>> {
-  caf::expected<To> x{To()};
-  if (convert(from, *x, std::forward<Opts>(opts)...))
-    return x;
-  return make_error(ec::convert_error);
+  using return_type = decltype(convert(from, std::declval<To&>(), opts...));
```
dominiklohmann, line 35:
```suggestion
  using return_type = decltype(convert(from, std::declval<To&>(), std::forward<Opts>(opts)...));
```

+ reformatting

dominiklohmann (changes requested):
Mostly fine by me except for one small, subtle issue in the code.

Please change the test record for the unit test such that it contains a list that contains at least one different type, e.g., record.

I did not test this locally; the unit tests running through cover this functionality quite well.

dominiklohmann (approved):
Approving as is, just one remark: Why did you move from `caf::settings` to `caf::dictionary<caf::config_value>`? The former is an alias of the latter, but I find `caf::settings` to be the more idiomatic type to enable conversion to.

mavam:
> Why did you move from `caf::settings` to `caf::dictionary<caf::config_value>`?

The reason is that `caf::settings` is just a top-level alias. The real recursive structure is a `config_value` and this is what I'd like the user of the API to remember: a `config_value` can contain a `dictionary` of `config_value`s. Since the recursion occurs in the implementation, I find it cleaning to stick to a single set of vocabulary.

PR #1058, opened 2020-09-15
Improve import batching options

dominiklohmann:
~Not sure if I should split out the last commit into a separate PR. Also still have to write changelog entries.~

### DIFF
--- a/doc/cli/vast-import.md
+++ b/doc/cli/vast-import.md
@@ -18,6 +18,24 @@ The above command signals the running node to ingest (i.e., to archive and index
 for later export) all Suricata events from the Eve JSON file passed via standard
 input.
 
+The import command batches parsed events in table slices. To control the
+batching, the following options are available:
+
+- `import.batch-encoding`: Controls the encoding of table slices. Available
+  options are `msgpack` (row-based) and `arrow` (column-based).
+- `import.batch-size`: Sets an upper bound for the size of every table slice.
+  A table slice is the unit that all components work on, causing this to be a
+  high impact tuning parameter. Decreasing the table slice size causes reduced
+  latency up to the point where messaging overhead becomes important, and
+  increasing it may improve overall performance. Note that this setting does not
+  mean that events are forwarded to the index and archive immediately upon
+  exceeding the given number of events, as the table slices themselves are
+  buffered again. Setting this option to 0 causes the table slice size to be
+  unbounded, leaving control of the batch size to other parameters.
+- `import.batch-timeout`: Sets a timeout for forwarding buffered table slices to
+  the index and archive, and can cause the `import.batch-size` parameter to be
+  underrun.
+
 An optional filter expression allows for importing the relevant subset of
 information only. For example, a user might want to import Suricata Eve JSON,
 but skip over all events of type `suricata.stats`.
--- a/doc/cli/vast-import.md
+++ b/doc/cli/vast-import.md
@@ -18,23 +18,7 @@ The above command signals the running node to ingest (i.e., to archive and index
 for later export) all Suricata events from the Eve JSON file passed via standard
 input.
 
-The import command batches parsed events in table slices. To control the
-batching, the following options are available:
-
-- `import.batch-encoding`: Controls the encoding of table slices. Available
-  options are `msgpack` (row-based) and `arrow` (column-based).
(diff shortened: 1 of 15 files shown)


doc/cli/vast-import.md
```diff
@@ -18,6 +18,24 @@ The above command signals the running node to ingest (i.e., to archive and index
 for later export) all Suricata events from the Eve JSON file passed via standard
 input.
 
+The import command batches parsed events in table slices. To control the
+batching, the following options are available:
```
mavam, line 22:
```suggestion
The import command parses events into table slices or *batches*.
The following options control the batching:
```

doc/cli/vast-import.md
```diff
@@ -18,6 +18,24 @@ The above command signals the running node to ingest (i.e., to archive and index
...
+
+- `import.batch-encoding`: Controls the encoding of table slices. Available
```
mavam, line 24:
```suggestion
- `import.batch-encoding`: Selects the encoding of table slices. Available
```

doc/cli/vast-import.md
```diff
@@ -18,6 +18,24 @@ The above command signals the running node to ingest (i.e., to archive and index
...
+  options are `msgpack` (row-based) and `arrow` (column-based).
+- `import.batch-size`: Sets an upper bound for the size of every table slice.
```
mavam, line 26:
```suggestion
- `import.batch-size`: Sets an upper bound for the number of events per table slice.
```

doc/cli/vast-import.md
```diff
@@ -18,6 +18,24 @@ The above command signals the running node to ingest (i.e., to archive and index
...
+  A table slice is the unit that all components work on, causing this to be a
+  high impact tuning parameter. Decreasing the table slice size causes reduced
+  latency up to the point where messaging overhead becomes important, and
+  increasing it may improve overall performance. Note that this setting does not
+  mean that events are forwarded to the index and archive immediately upon
+  exceeding the given number of events, as the table slices themselves are
+  buffered again. Setting this option to 0 causes the table slice size to be
+  unbounded, leaving control of the batch size to other parameters.
```
mavam, line 34:
```suggestion
  Most components in VAST operate on table slices, which makes the table slice size a fundamental tuning knob on the spectrum of throughput and latency. Small table slices allow for shorter processing times, resulting in more scheduler context switches and a more balanced workload. However, the increased pressure on the scheduler comes at the cost of throughput. A large table slice size allows actors to spend more time processing a block of memory, but makes them yield less frequently to the scheduler. As a result, other actors scheduled on the same thread may have to wait a little longer. 
  
  The `batch-size` merely controls number of events per table slice, but not necessarily the number of events until a component forwards a batch to the next stage in a stream. The [CAF streaming framework](https://actor-framework.readthedocs.io/en/latest/Streaming.html) uses a credit-based flow-control mechanism to determine buffering of tables slices. Setting `batch-size` to 0 causes the table slice size to be "unbounded" and leaves it to other parameters to determine the actual table slice size.
```

doc/cli/vast-import.md
```diff
@@ -18,6 +18,24 @@ The above command signals the running node to ingest (i.e., to archive and index
...
+- `import.batch-timeout`: Sets a timeout for forwarding buffered table slices to
+  the index and archive, and can cause the `import.batch-size` parameter to be
+  underrun.
```
mavam, line 37:
```suggestion
- `import.batch-timeout`: Sets a timeout for forwarding buffered table slices to
  the importer. If the timeout fires before a table slice reaches `import.batch-size`, then the table slice will contain fewer events but ship immediately.
```

🗣️💣

doc/cli/vast-import.md
```diff
@@ -18,6 +18,24 @@ The above command signals the running node to ingest (i.e., to archive and index
...
+
 An optional filter expression allows for importing the relevant subset of
```
mavam, line 39: This paragraph now needs a bit gentler introduction.

mavam (changes requested):

doc/cli/vast-import.md
mavam: Looks like you went with the previous wording in your new commit?

dominiklohmann: You are right, I overlooked this one apparently.

mavam (approved):

PR #1059, opened 2020-09-15
Switch to YAML config files

mavam:
- [x] Rebase onto master after #1055 is merged.
- [x] Rebase onto master after #1067 is merged.


mavam:
Note to the reviewer: this PR is best reviewed commit by commit.

vast.conf.example
```diff
@@ -1,447 +1,411 @@
...
+# The `vast pivot` command extracts related events of a given type.
+pivot:
+  # The output format.
+  # format: "json"
+
+  # For additionally available options, see export.pcap.
+
+# The `vast status` command prints a JSON-formatted status summary of the node.
+status:
+  # No further configuration options are available. The system options apply.
```
dominiklohmann, line 261: This is not true.

vast.conf.example
```diff
@@ -1,447 +1,411 @@
...
+
+  # Excludes listed components from logging.
+  component-blacklist:
+    - caf
+    - caf_flow
+    - caf_stream
+
+#scheduler:
+  # Accepted alternative: "sharing".
+  #policy: "stealing"
```
dominiklohmann, line 363:
```suggestion
  #policy: stealing
```

vast.conf.example
```diff
@@ -1,447 +1,411 @@
...
+
+  # Configures whether the scheduler generates profiling output.
+  #enable-profiling: false
+
+  # Output file for profiler data (only if profiling is enabled).
+  #profiling-output-file: "/dev/null"
```
dominiklohmann, line 369:
```suggestion
  #profiling-output-file: /dev/null
```

libvast/src/system/configuration.cpp
```diff
@@ -79,32 +83,99 @@ caf::error configuration::parse(int argc, char** argv) {
...
+    if (exists(config)) {
+      auto contents = load_contents(config);
+      if (!contents)
+        return contents.error();
+      auto yaml = from_yaml(*contents);
+      if (!yaml)
+        return yaml.error();
+      auto rec = caf::get_if<record>(&*yaml);
+      if (!rec)
+        return caf::make_error(ec::parse_error, "config file not a YAML map");
```
dominiklohmann, line 109: Please improve this error context. From a user's point of view, this does not currently make any sense.

libvast/src/system/configuration.cpp
```diff
@@ -79,32 +83,99 @@ caf::error configuration::parse(int argc, char** argv) {
...
+      auto flat_yaml = flatten(*rec);
+      // Erase all null values because a caf::config_value has no such notion.
+      for (auto i = flat_yaml.begin(); i != flat_yaml.end();) {
+        if (caf::holds_alternative<caf::none_t>(i->second))
+          i = flat_yaml.erase(i);
+        else
+          ++i;
+      }
```
dominiklohmann, line 117: This is faulty behavior w.r.t. the merging behavior below. `null` in a higher-precedence configuration file should override the existing setting.

libvast/src/system/configuration.cpp
```diff
@@ -79,32 +83,99 @@ caf::error configuration::parse(int argc, char** argv) {
...
+                         [](const std::string& x) { return x; });
+    auto str = caf::visit(no_quote_stringify, val);
+    auto result = opt.parse(str);
+    if (!result) {
+      // We now try to parse strings as atom using a regex, since we get
+      // recursive types like lists for free this way. A string-vs-atom type
+      // clash is the only instance we currently cannot distinguish. Everything
+      // else is a true type clash.
+      // (With CAF 0.18, this heuristic will be obsolete.)
+      str = std::regex_replace(str, std::regex("\""), "'");
```
dominiklohmann, line 145: Please rewrite this in terms of `std::transform_if` for three reasons:
1. It is absolutely overcomplicating the code for readers and reviewers.
2. It is unnecessarily slow at runtime, especially the unnecessarily repeated regular expression compilation (these should _always_ be cached).
3. The inclusion of the regex header adds a lot of compile time overhead.

dominiklohmann (changes requested):
I only looked at the code for now, will do the testing locally for a later review round.

Please rename the config file in this PR already and warn if a `vast.conf` file is found to reduce friction for the user.

vast.conf.example
mavam: I just took this from the default. How about we change this as we are restructuring the YAML file and streamline the sections? Then it becomes also easier to see how the transformation from CAF to YAML config took place.

libvast/src/system/configuration.cpp
mavam: So then we need to merge all `data` instances first.

libvast/src/system/configuration.cpp
mavam: Turns out we had `detail::replace_all` already! 🙈

libvast/src/system/configuration.cpp
mavam: Better?

mavam:
CI currently fails due the issue to be fixed in #1067; marking this PR as *blocked* until merged.

vast.conf.example
dominiklohmann: Alrighty.

dominiklohmann (approved):
This is working nicely in practice. I've manually verified the merging logic to work.

PR #1060, opened 2020-09-16
Set the default batch size to 1000

dominiklohmann:


tobim:
@dominiklohmann this is ready now.

tobim (approved):

PR #1061, opened 2020-09-16
Fix display of the acccountant actor's name

dominiklohmann:
When pimpl-ing actor states, we must move the actor name from the state impl to the state pimpl.


mavam (approved):
Approving. For next time: the small fixup would have been nice in a separate commit.

PR #1062, opened 2020-09-17
Fix the static build

tobim:


mavam (approved):
Simple fix; not much to add. If CI runs through, go for the merge.

PR #1063, opened 2020-09-17
Spawn INDEXERs lazily from partition

lava:
Instead of going over the whole partition and spawning all indexer when loading a partition, we `mmap` the partition and only access the relevant part of the memory once we get a query that actually uses this indexer.

### DIFF
--- a/libvast/src/system/partition.cpp
+++ b/libvast/src/system/partition.cpp
@@ -65,18 +65,38 @@ using namespace caf;
 
 namespace vast::system {
 
+caf::actor active_partition_state::indexer_at(size_t position) const {
+  VAST_ASSERT(position < indexers.size());
+  return as_vector(indexers)[position].second;
+}
+
+/// Gets the INDEXER at a certain position in the `indexers` stable map.
+caf::actor passive_partition_state::indexer_at(size_t position) const {
+  VAST_ASSERT(position < indexers.size());
+  auto& indexer = indexers[position];
+  // Deserialize the value index and spawn a passive_indexer lazily when it is
+  // requested for the first time.
+  if (!indexer) {
+    auto qualified_index = flatbuffer->indexes()->Get(position);
+    auto index = qualified_index->index();
+    auto data = index->data();
+    value_index_ptr state_ptr;
+    caf::binary_deserializer sink(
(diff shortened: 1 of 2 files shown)


libvast/vast/system/partition.hpp
```diff
@@ -14,6 +14,7 @@
 #pragma once
 
 #include "vast/aliases.hpp"
+#include "vast/chunk.hpp"
```
dominiklohmann, line 17: We should only need a forward-declaration of `chunk_ptr` here, not the full header, so `vast/fwd.hpp` should suffice, and if it does not, this should be solved in `vast/fwd.hpp` instead.

The `vast/detail/stable_map.hpp` include can be removed I think.

dominiklohmann (approved):

libvast/src/system/partition.cpp
```diff
@@ -551,7 +554,7 @@ passive_partition(caf::stateful_actor<passive_partition_state>* self, uuid id,
   // to queries. The `skip` default handler is used to buffer all messages
   // arriving until then.
   self->set_default_handler(skip);
-  self->send(caf::actor_cast<caf::actor>(fs), atom::read_v, path);
+  self->send(caf::actor_cast<caf::actor>(fs), atom::mmap_v, path);
```
mavam, line 557: It's great to see that we literally only need a one-line change to switch to mmap'ing now!

libvast/src/system/partition.cpp
```diff
@@ -65,19 +65,39 @@ using namespace caf;
...
-namespace {
-
-// The functions in this namespace take PartitionState as template argument
-// because the impelementation is the same for passive and active partitions.
+caf::actor active_partition_state::indexer_at(size_t position) const {
+  VAST_ASSERT(position < indexers.size());
+  return as_vector(indexers)[position].second;
+}
 
 /// Gets the INDEXER at a certain position in the `indexers` stable map.
```
mavam, line 73: I think the comment no longer holds, right?

mavam (approved):
Nice work!

I just noticed one detail in a comment.

libvast/src/system/partition.cpp
lava: Right, updated.

PR #1064, opened 2020-09-17
Install libyaml-cpp for prebuilt Dockerfile

dominiklohmann:


0ortmann:
The GitHub checks did not run the affected step. I built the image locally with the prebuilt vast artifact from this branch. I tested that instantiated containers can be run without errors, VAST now has all required dependencies in place.

0ortmann (approved):

PR #1065, opened 2020-09-17
Use UUID type in Segment flatbuffer

lava:
**NOTE**: This is changing the binary format of the `Segment` flatbuffer. I'm undecided if its still early enough to sneak that in, or if we should use the opportunity to battle-test our upgrade paths and switch update to `Segment v1`.


mavam (approved):
Looks good to me as is. We're still too early for bumping versions right after the merge to master. Let's hold off with that until the next official release.

PR #1067, opened 2020-09-18
Fix release builds using gcc-8

dominiklohmann:
Turns out gcc-8 is unable to explicitly capture a variable in a lambda that was implicitly captured in a surrounding lambda if that variable is then only used in an unevaluated context like decltype in the inner lambda.

I verified that this builds as expected by running `docker build . -t vast`.

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -679,7 +679,7 @@ caf::behavior index(caf::stateful_actor<index_state>* self, filesystem_type fs,
       // below in the same actor context.
       await_evaluation_maps(
         self, iter->second.expression, actors,
-        [self, client, query_id](caf::expected<pending_query_map> maybe_pqm) {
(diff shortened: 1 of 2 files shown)


libvast/src/system/index.cpp
```diff
@@ -688,7 +688,7 @@ caf::behavior index(caf::stateful_actor<index_state>* self, filesystem_type fs,
             self->send(client, atom::done_v);
             return;
           }
-          auto& [query_id, query_state] = *iter;
+          auto& query_state = iter->second;
```
mavam, line 691: Independent of the syntactic change: @lava isn't capturing `iter` in the lambda a potential bug, since it may get invalidated until the lambda is called?

mavam: Sorry, my bad. The parent context also contains an `iter`, but the local one shadows it.

mavam (approved):

PR #1070, opened 2020-09-21
Restrict map to json conversion to string keys

tobim:
This change makes the convert overload from `map` to JSON less generic, i.e. only `std::string` is allowed as key type.


dominiklohmann:
Why is this change necessary and how does it relate to the `null` output of the status command in the static build of VAST Pro?

mavam (approved):
LGTM. This code will get ripped out soon altogether with `vast::json` anyways.

mavam:
I missed the failing unit test. Please address this before merging.

PR #1071, opened 2020-09-22
Make some verbose log messages more user-facing

dominiklohmann:
This is on verbose, so let's write it in a way that makes sense to the user.

### DIFF
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -195,7 +195,6 @@ spawn_component(const invocation& inv, spawn_arguments& args) {
   auto i = node_state::component_factory.find(inv.full_name);
   if (i == node_state::component_factory.end())
     return make_error(ec::unspecified, "invalid spawn component");
-  VAST_VERBOSE_ANON("node spawns the", inv.name(), "component");
   return i->second(self, args);
 }
 
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -195,7 +195,7 @@ spawn_component(const invocation& inv, spawn_arguments& args) {
   auto i = node_state::component_factory.find(inv.full_name);
   if (i == node_state::component_factory.end())
     return make_error(ec::unspecified, "invalid spawn component");
-  VAST_VERBOSE(__func__, VAST_ARG(inv.full_name));
+  VAST_VERBOSE_ANON("node spawns the", inv.name(), "component");
   return i->second(self, args);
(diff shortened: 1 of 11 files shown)


libvast/src/system/node.cpp
```diff
@@ -195,7 +195,7 @@ spawn_component(const invocation& inv, spawn_arguments& args) {
   auto i = node_state::component_factory.find(inv.full_name);
   if (i == node_state::component_factory.end())
     return make_error(ec::unspecified, "invalid spawn component");
-  VAST_VERBOSE(__func__, VAST_ARG(inv.full_name));
+  VAST_VERBOSE_ANON("node spawns the", inv.name(), "component");
```
tobim, line 198: Wouldn't it be nicer if the message read `node spawns an exporter`?

dominiklohmann: If you can find a better phrasing that works for all of these I'm all ears.

```
2020-09-22T10:09:15.142 node spawns the "accountant" component
2020-09-22T10:09:15.143 node spawns the "type-registry" component
2020-09-22T10:09:15.143 node spawns the "archive" component
2020-09-22T10:09:15.144 node spawns the "index" component
2020-09-22T10:09:15.145 node spawns the "importer" component
```

tobim: Remove this log message and add component specific ones in the `spawn_*` functions. The messages for singleton components should be `node spawns the accountant`. For non-singleton components you can write `node spawns an exporter for <query>`.

dominiklohmann: Done. Please check if this matches your expectations.

libvast/src/system/spawn_archive.cpp
```diff
@@ -39,10 +40,13 @@ maybe_actor spawn_archive(node_actor* self, spawn_arguments& args) {
   auto mss
     = 1_MiB
       * get_or(args.inv.options, "max-segment-size", sd::max_segment_size);
-  auto actor = self->spawn(archive, args.dir / args.label, segments, mss);
+  auto handle = self->spawn(archive, args.dir / args.label, segments, mss);
+  if (!handle)
```
tobim, line 44: I don't understand why we need those checks here. I thought local spawn can't fail at runtime?

dominiklohmann: Alright, removed them again.

tobim (approved):
LGTM

PR #1072, opened 2020-09-22
Add event type name to the record batch metadata

tobim:


dominiklohmann (changes requested):
This needs a changelog entry and the man-page for the `export arrow` command must be adapted.

dominiklohmann (approved):

PR #1075, opened 2020-09-22
Add missing options categories

dominiklohmann:
For most commands, this is for completeness only. For the status and spawn sink commands, this fixes setting configuration options from the vast.yaml file.


tobim (approved):
Didn't test locally. The options categories are defined in the same way as for the other commands, I don't see why this shouldn't work as intended.

PR #1076, opened 2020-09-23
Remove inconsistencies from the CMake setup

tobim:


CMakeLists.txt
```diff
@@ -591,28 +598,24 @@ else ()
...
+    if (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/aux/broker/broker/CMakeLists.txt)
+      add_subdirectory(aux/broker)
+      set(broker_dir ${CMAKE_CURRENT_SOURCE_DIR}/aux/broker)
     else ()
       message(
         FATAL_ERROR
-          "Broker not found, either use -DBROKER_ROOT_DIR=... or initialize the aux/broker submodule"
-      )
+          "Broker not found, either use -DBROKER_ROOT_DIR=... or initialize the
+          aux/broker/broker submodule")
```
dominiklohmann, line 612: This line break shouldn't be here.

CMakeLists.txt
```diff
@@ -483,6 +488,8 @@ get_target_property(flatbuffers_dir ${flatbuffers_target}
 get_filename_component(flatbuffers_dir ${flatbuffers_dir} DIRECTORY) # <prefix>
 
 find_package(yaml-cpp 0.6.2 REQUIRED HINTS ${YAML_CPP_ROOT})
+get_filename_component(yaml-cpp_dir "${YAML_CPP_INCLUDE_DIR}" ABSOLUTE)
```
dominiklohmann, line 491: Can you name this `yaml_cpp_dir` instead?

Also, wouldn't it make more sense to use `get_target_property` to get the include directory?

cmake/FindBROKER.cmake
```diff
@@ -18,13 +18,14 @@
 find_library(
   BROKER_LIBRARY
   NAMES broker
-  HINTS ${BROKER_ROOT_DIR}/lib ${BROKER_ROOT_DIR})
+  HINTS ${BROKER_ROOT_DIR} ${VAST_PREFIX}
```
dominiklohmann, line 21: Not sure if I like that the FindBROKER module knows about variables that start with VAST.

dominiklohmann (changes requested):

cmake/FindBROKER.cmake
tobim: I know it's not the cleanest way to implement this, but putting it here reduces the complexity of the call site considerably.

CMakeLists.txt
tobim: `YAML_CPP_INCLUDE_DIR` is already provided by `yaml-cpp-config.cmake`, why should we ignore that?

cmake/FindBROKER.cmake
dominiklohmann: I'd rather have the complexity with a comment. Moving this here makes maintenance in the future harder.

Aside, do we _really_ need a FindBROKER module? Is the one shipping with broker with a supplied HINT at the call side not sufficient?

CMakeLists.txt
dominiklohmann: Can you confirm that works with both 0.6.2 and 0.6.3? Because somewhere in between they reworked their CMake completely w.r.t to packaging. If this solution is portable then yes, keep it.

tobim: We used the same variable before this change. The config file has only one revision: https://github.com/jbeder/yaml-cpp/commits/master/yaml-cpp-config.cmake.in

cmake/FindBROKER.cmake
tobim: `findPackage` does not support `HINT` in legacy mode.

tobim: As discussed via direct messaging, we decided to remove `VAST_PREFIX` from the Broker search path.

dominiklohmann (approved):

PR #1079, opened 2020-09-28
Don't add compile time paths to relocatable binary

tobim:


vast/vast.cpp
```diff
@@ -73,7 +73,7 @@ int main(int argc, char** argv) {
   for (auto& path : cfg.config_files)
     VAST_INFO_ANON("loaded configuration file:", path);
   using string_list = std::vector<std::string>;
-  auto schema_dirs = std::vector<vast::path>{};
+  auto schema_dirs = std::set<vast::path>{};
```
dominiklohmann, line 76: This is missing `#include <set>` at the top.

libvast/vast/schema.hpp
```diff
@@ -83,7 +83,7 @@ bool convert(const schema& s, json& j);
 
 caf::expected<schema> load_schema(const path& sf);
 
-caf::expected<vast::schema> load_schema(const std::vector<path>& data_paths);
+caf::expected<vast::schema> load_schema(const std::set<path>& data_paths);
```
dominiklohmann, line 86: This is missing `#include <set>` at the top.

dominiklohmann (approved):

PR #1080, opened 2020-09-29
Tag and verify (Docker) releases

0ortmann:


.github/workflows/vast.yaml
```diff
@@ -227,6 +232,22 @@ jobs:
         run: |
           cmake --build "$BUILD_DIR" --target package
 
+      - name: Verify Docker Image
+        if: matrix.os.tag == 'Ubuntu' && matrix.os.compiler == 'GCC' && matrix.configure.tag == 'Release'
```
dominiklohmann, line 236: We have this exact check in a few places now. Can you add a `matrix.use_for_docker` key that is true for this case and false for all others?

dominiklohmann (approved):
Looks alright to me as is, just have one small refactoring request for the workflow.

PR #1081, opened 2020-09-29
Improve behavior of get_node_components

dominiklohmann:
The first commit in this branch is cherry-picked from #1080, where we noticed that the build was broken for gcc-8 (again). This PR is essentially a better fix for the same issue.

### DIFF
--- a/libvast/vast/system/node_control.hpp
+++ b/libvast/vast/system/node_control.hpp
@@ -36,21 +36,21 @@ spawn_at_node(caf::scoped_actor& self, caf::actor node, Arguments&&... xs) {
 
 /// Look up components by category. Returns the first actor of each
 /// category name passed in `names`.
-template <size_t N>
-caf::expected<std::array<caf::actor, N>>
+template <class... Names>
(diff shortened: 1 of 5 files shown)


libvast/vast/system/node_control.hpp
```diff
@@ -36,21 +36,21 @@ spawn_at_node(caf::scoped_actor& self, caf::actor node, Arguments&&... xs) {
...
-                    const std::string_view (&names)[N]) {
-  auto result = caf::expected{std::array<caf::actor, N>{}};
-  std::vector<std::string> labels;
-  labels.reserve(N);
-  for (auto name : names)
-    labels.emplace_back(name);
+                    Names&&... names) {
+  static_assert(
+    std::conjunction_v<std::is_constructible<std::string, Names>...>,
+    "names cannot be used for construction a string");
```
tobim, line 45:
```suggestion
    "name parameter cannot be used to construct a string");
```

tobim (changes requested):

tobim (approved):

PR #1083, opened 2020-09-29
Use FlatBuffers unions for versioning

lava:
This commit improves the versioning scheme used for the on-disk
state of VAST by using a union type as primary means for version
inference.

NOTE: This is a breaking change, it will not be possible to load
flatbuffer files created by older versions of VAST.


tools/lsvast/lsvast.cpp
```diff
@@ -119,22 +102,23 @@ struct flatbuffer_deleter {
...
-  auto view = vast::span<const vast::byte>(bytes.data(), bytes.size());
-  auto maybe_flatbuffer = vast::fbs::as_versioned_flatbuffer<T>(view, version);
-  if (!maybe_flatbuffer)
-    return result;
-  return result_t(*maybe_flatbuffer, flatbuffer_deleter<T>(std::move(bytes)));
+  const auto* ptr = flatbuffers::GetRoot<T>(bytes.data());
+  // auto view = vast::span<const vast::byte>(bytes.data(), bytes.size());
+  // auto maybe_flatbuffer = vast::fbs::as_versioned_flatbuffer<T>(view,
+  // version); if (!maybe_flatbuffer)
+  //   return result;
```
dominiklohmann, line 117: Can this be removed?

libvast/vast/fbs/segment.fbs
```diff
@@ -31,6 +31,16 @@ table Segment {
   events: ulong;
 }
 
+namespace vast.fbs;
+
+union SegmentUnion {
```
dominiklohmann, line 36:
```suggestion
union VersionedSegment {
```

libvast/src/system/index.cpp
```diff
@@ -158,15 +157,18 @@ caf::error index_state::load_from_disk() {
...
-        auto partition = fbs::as_versioned_flatbuffer<fbs::Partition>(
-          as_bytes(chunk), fbs::Version::v1);
-        if (!partition) {
-          VAST_WARNING(self, "coult not parse", partition_path,
-                       "as partition:", partition.error());
+        auto partition = fbs::GetPartition(chunk->data());
+        if (partition->versioned_partition_type()
+            != fbs::PartitionUnion::v1_Partition) {
+          // TODO: We could also support v0 partitions here by creating the
+          // missing meta index on-the-fly.
```
dominiklohmann, line 164: Does it make sense to throw out the v0 and make v1 the v0 instead, since they're both happening for the same release? (Unless this does not make the cut)

libvast/src/system/index.cpp
```diff
@@ -158,15 +157,18 @@ caf::error index_state::load_from_disk() {
...
+          VAST_WARNING(self, "found outdated partition flatbuffer");
```
dominiklohmann, line 165:
```suggestion
          VAST_WARNING(self, "found outdated partition", partition_uuid,
                       "of type", partition->GetFullyQualifiedName());
```

libvast/test/flatbuffers.cpp
```diff
@@ -67,14 +72,26 @@ TEST(index roundtrip) {
   // Add some fake statistics
   state.stats.layouts["zeek.conn"] = vast::system::layout_statistics{54931u};
   // Serialize the index.
-  auto expected_fb = vast::fbs::wrap(state, "I000");
-  REQUIRE(expected_fb);
-  auto fb = *expected_fb;
-  auto span = as_bytes(fb);
+  // auto expected_fb = vast::fbs::wrap(state);
```
dominiklohmann, line 75: Is `vast::fbs::wrap` dead code?

libvast/vast/fbs/segment.fbs
dominiklohmann: I think this is generally a better name, because it matches the `versioned_segment` field below. This makes it easier to make the connection in the generated C++ code.

libvast/src/system/partition.cpp
```diff
@@ -576,17 +581,15 @@ passive_partition(caf::stateful_actor<passive_partition_state>* self, uuid id,
...
+      auto partition = fbs::GetPartition(chunk->data());
+      if (partition->versioned_partition_type()
+          != fbs::PartitionUnion::v1_Partition) {
+        VAST_ERROR(self, "got flatbuffer partition with invalid version");
+        return self->quit();
       }
+      auto partition_v1 = partition->versioned_partition_as_v1_Partition();
       self->state.partition_chunk = chunk;
-      self->state.flatbuffer = *partition;
+      self->state.flatbuffer = partition_v1;
```
dominiklohmann, line 592: This can just be merged with the line above, no need for the additional variable.

dominiklohmann (changes requested):
Got some comments, nothing major.

libvast/src/system/partition.cpp
```diff
@@ -576,17 +581,15 @@ passive_partition(caf::stateful_actor<passive_partition_state>* self, uuid id,
...
-      auto partition
-        = fbs::as_versioned_flatbuffer<fbs::Partition>(view, fbs::Version::v1);
-      if (!partition) {
-        VAST_ERROR(self, "failed to parse provided chunk as flatbuffer:",
-                   render(partition.error()));
-        return self->quit(std::move(partition.error()));
+      auto partition = fbs::GetPartition(chunk->data());
+      if (partition->versioned_partition_type()
+          != fbs::PartitionUnion::v1_Partition) {
+        VAST_ERROR(self, "got flatbuffer partition with invalid version");
```
dominiklohmann, line 587:
```suggestion
        VAST_ERROR(self, "found partition with invalid version of type:"
                   partition->GetFullyQualifiedName());
```

libvast/vast/system/partition.hpp
```diff
@@ -149,7 +149,7 @@ struct passive_partition_state {
   vast::chunk_ptr partition_chunk;
 
   /// A typed view into the `partition_chunk`.
-  const fbs::Partition* flatbuffer;
+  const fbs::v1::Partition* flatbuffer;
```
tobim, line 152: I think I would try to keep the version part in the implementation, at least for the reading side. It would be really desirable to be able to load a `vast.db` from the previous version of vast if the change between `N - 1` and `N` permits it.

tobim (changes requested):
In general we should use the outer table for the reading side and do the version dispatch only internally.
Concretely, it should be `unpack(const fbs::Partition& x,...)` but `caf::expected<flattbuffers::Offset<fbs::partition::v1>> pack(...)`
The same applies to the handles of versioned objects that show up in the headers, for example in the `passive_partition_state`.

libvast/vast/system/partition.hpp
lava: After some offline discussion, we agreed that this makes more sense the other way round:

For `pack()` we can return the union type since there's a unique embedding from `partition::v0` into `Partition`, and the caller always performs this work anyways.

For `unpack()` it depends on the exact changes between the versions whether we can trivially unpack a different version, want to emit a warning, require some explicit conversion or just refuse it altogether, so the calling code should just do whatever it wants to do beforehand so the unpack function can focus purely on unpacking the data.

tobim (approved):

dominiklohmann (approved):
Fine from my side. I think I'll do an `s/flatbuffers?/FlatBuffers/g` in a separate PR.

PR #1084, opened 2020-09-29
Make use of new serialize_bytes convenience function

lava:
Use the new serialize_bytes/deserialize_bytes convenience functions to simplify the flatbuffer-related code of the partition actor.


dominiklohmann (approved):
This is great! I re-ran CI because it failed because of a rare, non-deterministic, and unrelated hang. Should run through this time.

PR #1086, opened 2020-09-30
Use a stable set for schema directories

dominiklohmann:


tobim (commented):
There is one more thing we should consider:
With the current design, the user has no way to give a higher priority to one of the predefined directories.
Consider `vast --schema-paths=[/usr/share/vast/schema]`. The given path would still get lower priority than the xdg ones.

dominiklohmann:
> There is one more thing we should consider:
> With the current design, the user has no way to give a higher priority to one of the predefined directories.
> Consider `vast --schema-paths=[/usr/share/vast/schema]`. The given path would still get lower priority than the xdg ones.

That is not true, we still have `--no-default-schema` to disable the default paths.

We can also change the insertion order such that user dirs are put first in the order. I'll gladly make that change, because I think that makes more sense anyways.

tobim:
> That is not true, we still have --no-default-schema to disable the default paths.

I forgot about that.

> We can also change the insertion order such that user dirs are put first in the order. I'll gladly make that change, because I think that makes more sense anyways.

I prefer it the way it is now, we should assume that the user knows what they're doing.

tobim (approved):

PR #1088, opened 2020-10-05
Render errors using vast::render(...)

dominiklohmann:


tobim (approved):
Thanks for making this more consistent!

PR #1089, opened 2020-10-05
Improve the detail::overload visitor helper

dominiklohmann:
In very much non-scientific tests, this version of `detail::overload` compiles slightly faster, and has better diagnostics, because the visitor no longer is an anonymous struct.

Additionally, the brace-list-init constructor allows for using a trailing comma, which makes git-diffs leaner, and improves formatting in a few places.

This change is purely syntactical and does not alter the behavior of the code.

---

I've verified this to work using gcc-8 manually by running `docker build .` from the repository root.

### DIFF
--- a/libvast/src/arrow_table_slice_builder.cpp
+++ b/libvast/src/arrow_table_slice_builder.cpp
@@ -435,7 +435,7 @@ table_slice_builder_ptr arrow_table_slice_builder::make(record_type layout) {
 arrow_table_slice_builder::column_builder_ptr
 arrow_table_slice_builder::make_column_builder(const type& t,
                                                arrow::MemoryPool* pool) {
-  auto f = detail::overload(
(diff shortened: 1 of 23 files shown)


mavam:
Will there ever be a `std::overload` or something? If there's a proposal out there, it might make sense to use that syntax.

dominiklohmann:
> Will there ever be a `std::overload` or something? If there's a proposal out there, it might make sense to use that syntax.

I've only found [p1772r0](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1772r0.pdf), but it doesn't look like much work has happened on it yet. At this point, everyone uses either of the two versions in their code base anyways.

I prefer the style from this PR for the reasons I listed. Allowing for a trailing comma is really nice.

mavam (approved):

PR #1090, opened 2020-10-06
Remove '--disable-recoverability' option

lava:
This option was added as a workaround for the case where the
meta index grew large enough to cause significant slowdown
when changing the active partition.

However, the meta index is not stored in the index flatbuffer
anymore, so we can remove it again.


tobim (changes requested):
I'll approve this once the tests are fixed.

tobim (approved):

PR #1091, opened 2020-10-07
Support native systemd startup notification from VAST

lava:


libvast/src/detail/posix.cpp
```diff
@@ -31,6 +31,7 @@
 #include <string.h>
 #include <unistd.h>
 
+#include <asm-generic/errno-base.h>
```
tobim, line 34: What is this for?

lava: Uhm, no idea what that is and how that got there :confused:

libvast/src/system/start_command.cpp
```diff
@@ -85,6 +86,9 @@ caf::message start_command_impl(start_command_extra_steps extra_steps,
     return caf::make_message(std::move(bound_port.error()));
   auto listen_addr = std::string{host} + ':' + std::to_string(*bound_port);
   VAST_INFO_ANON("VAST node is listening on", listen_addr);
+  // Notify the service manager if it expects an update.
+  if (auto error = systemd::notify_ready())
+    return caf::make_message(std::move(error));
```
tobim, line 91: Just to be extra safe, we should probably do this after starting the `signal_monitor` thread.

tobim (changes requested):
Good work, just one small request.

libvast/src/detail/posix.cpp
```diff
@@ -171,6 +171,33 @@ int uds_recv_fd(int socket) {
...
+  // For abstract domain sockets (i.e., where the first char is '@'), the
+  // terminating NUL byte is not counted towards the length, but any NUL
+  // bytes embedded in the path are. For example, in the extreme case it
+  // would be possible to have two different sockets named NUL and NUL NUL.
+  auto pathlen = destination[0] == '@' ? destination.size()
+                                       : std::strlen(&destination[0]) + 1;
+  msghdr.msg_namelen = offsetof(struct sockaddr_un, sun_path) + pathlen;
+  msghdr.msg_iov = &iovec;
+  msghdr.msg_iovlen = 1;
+  return ::sendmsg(socket, &msghdr, MSG_NOSIGNAL);
```
tobim, line 198: Why do we need `MSG_NOSIGNAL` for sending a single packet?

lava: We use `MSG_NOSIGNAL` because we want to return a regular `caf::error` instead of getting `SIGPIPE` if the socket was shut down for writing. I don't think the number of packets is relevant for this.

Since `MSG_NOSIGNAL` seems to be linux-specific I've slightly refactored this to not hard-code `MSG_NOSIGNAL` but to instead take a generic `flags` argument.

libvast/src/systemd.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <caf/detail/scope_guard.hpp>
+
+#include <cstdlib>
+#include <cstring>
+
+#include <sys/socket.h>
+
+namespace vast::systemd {
+
+#if VAST_LINUX
```
dominiklohmann, line 16: Since this file is not conditionally compiled, does it make sense to rename it to something like `init_system.cpp` so we can provide implementations for launchd and rc.d later on if needed?

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🎁 The `Type=notify` directive is now supported when running VAST under
+  systemd supervision.
```
mavam, line 15: I wouldn't know how to interpret this changelog entry. Can it be written so that non-systemd experts understand it as well? What is a directive? Where does the assignment come from?

libvast/vast/error.hpp
```diff
@@ -85,6 +85,8 @@ enum class ec : uint8_t {
   silent,
   /// Insufficient memory.
   out_of_memory,
+  /// An error occured
```
mavam, line 88: This description needs to be a bit more meaningful.

CHANGELOG.md
mavam: This became clearer when I read the systemd config, by the way.

libvast/src/detail/posix.cpp
tobim: see: https://pubs.opengroup.org/onlinepubs/9699919799/functions/send.html:

> MSG_NOSIGNAL  
&nbsp;&nbsp;Requests not to send the SIGPIPE signal if an attempt to send is made on a **stream-oriented socket** that is no longer connected. The [EPIPE] error shall still be returned.

(emphasis mine).

libvast/src/systemd.cpp
```diff
@@ -0,0 +1,48 @@
...
+  });
+  auto notify_socket_env = ::getenv("NOTIFY_SOCKET");
+  if (!notify_socket_env)
+    return caf::none;
+  VAST_VERBOSE_ANON("notifying systemd at", notify_socket_env);
+  int socket = ::socket(AF_UNIX, SOCK_DGRAM | SOCK_CLOEXEC, 0);
+  if (socket < 0)
+    return make_error(ec::system_error, "failed to create unix socket");
+  if (detail::uds_sendmsg(socket, notify_socket_env, "READY=1\n", MSG_NOSIGNAL)
+      < 0)
```
dominiklohmann, line 34: Should this use the `unix_domain_socket` wrapper that we have instead such that the user (i.e., the highlighted code) does not need to know about the return values?

lava: I don't think so; the `unix_domain_socket` seems to be purpose-built to send and receive file descriptors. Adding `sendmsg()` but not e.g. `send()` or `recvmsg()` as well would make for an odd API.

libvast/vast/error.hpp
lava: Huh, not sure what happened to the rest of the sentence

libvast/src/systemd.cpp
lava: I think it makes sense to have a more general header when we add support for launchd and rc.d, but I'm not sure we need to add it prematurely. We can still introduce it as soon as we actually have something to abstract away from.

libvast/src/detail/posix.cpp
lava: So the real question is if we need `MSG_NOSIGNAL` for a `SOCK_DGRAM` socket.

I didn't really find any sort of explanation by looking at the systemd git history, so this might just be an oversight or a workaround for some ancient kernel bug. Either way, I've removed it since the docs clearly say it should not be necessary.

CHANGELOG.md
lava: I've reworded this a bit to make it clearer this is talking about systemd-related changes.

PR #1093, opened 2020-10-08
Implement concepts and concept resolution

tobim:
And add taxonomies support to the `type_registry`.


libvast/vast/expression.hpp
```diff
@@ -277,6 +277,45 @@ auto inspect(Inspector&f, expression& x) {
   return f(caf::meta::type_name("expression"), x.get_data());
 }
 
+template <class F>
+struct expression_visitor {
```
mavam, line 281:
```suggestion
struct predicate_transformer {
```

Since the scope of this visitor is a little more narrow, this name might reflect the mechanics a bit better.

libvast/vast/taxonomies.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace vast {
+
+using concepts_t = std::unordered_map<std::string, std::vector<std::string>>;
+using models_t = std::unordered_map<std::string, std::vector<std::string>>;
+
+struct taxonomies {
```
mavam, line 30: Why is this plural?

mavam, line 30: Could you add a bit documentation to the types? Plus all the free functions as well?

libvast/vast/taxonomies.hpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include <memory>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace vast {
+
+using concepts_t = std::unordered_map<std::string, std::vector<std::string>>;
+using models_t = std::unordered_map<std::string, std::vector<std::string>>;
```
mavam, line 28:
```suggestion
using concepts_type = std::unordered_map<std::string, std::vector<std::string>>;
using models_type = std::unordered_map<std::string, std::vector<std::string>>;
```

The `_t` strikes me as a C-ism. In C++ we use `_type` if we can, and only resort to `_t` if need be.

libvast/src/taxonomies.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+static expression
+resolve_concepts(const concepts_t& concepts, const expression& orig) {
+  return for_each_predicate(orig, [&](const auto& pred) {
+    if (auto fe = caf::get_if<field_extractor>(&pred.lhs)) {
+      // This algorithm recursivly looks up items form the concepts map and
+      // generates a predicate for every discovered name that is not a concept
+      // itself.
+      disjunction d;
+      std::stack<std::string> s;
```
mavam, line 51:
```suggestion
      std::stack<std::string> stack;
```

libvast/src/taxonomies.cpp
```diff
@@ -0,0 +1,79 @@
...
+      }
+      return expression{d};
+    }
+    return expression{pred};
+  });
+}
+
+expression resolve(const taxonomies& t, const expression& orig) {
+  return resolve_concepts(t.concepts, orig);
+}
```
mavam, line 77:
```suggestion
expression resolve(const taxonomies& t, const expression& e) {
  return resolve_concepts(t.concepts, e);
}
```

libvast/src/taxonomies.cpp
```diff
@@ -0,0 +1,79 @@
...
+      // generates a predicate for every discovered name that is not a concept
+      // itself.
+      disjunction d;
+      std::stack<std::string> s;
+      std::vector<std::string> tried;
+      s.push(fe->field);
+      while (!s.empty()) {
+        auto x = s.top();
+        s.pop();
+        auto i = concepts.find(x);
```
mavam, line 57:
```suggestion
        auto concept = concepts.find(x);
```

libvast/vast/taxonomies.hpp
```diff
@@ -0,0 +1,50 @@
...
+  friend auto inspect(Inspector& f, taxonomies& t) {
+    return f(caf::meta::type_name("taxonomies"), t.concepts, t.models);
+  }
+};
+
+using taxonomies_ptr = std::shared_ptr<const taxonomies>;
+
+caf::error inspect(caf::serializer& sink, const taxonomies_ptr& x);
+
+caf::error inspect(caf::deserializer& source, taxonomies_ptr& x);
```
mavam, line 46: Why do we need `inspect` overloads for the `shared_ptr`?

libvast/src/taxonomies.cpp
```diff
@@ -0,0 +1,79 @@
...
+        tried.push_back(x);
+        if (i != concepts.end()) {
+          for (auto ri = i->second.rbegin(); ri != i->second.rend(); ++ri) {
```
mavam, line 60: Why backwards?

libvast/src/taxonomies.cpp
```diff
@@ -0,0 +1,79 @@
...
+  if (x.use_count() > 1)
+    return make_error(ec::logic_error, "deserializing into object with multiple"
+                                       "owners violates precondition");
+  return source(*x);
+}
+
+static expression
+resolve_concepts(const concepts_t& concepts, const expression& orig) {
+  return for_each_predicate(orig, [&](const auto& pred) {
+    if (auto fe = caf::get_if<field_extractor>(&pred.lhs)) {
```
mavam, line 46: FYI: You have to assume that the expression is normalized here. What if the field is on the RHS?

libvast/src/system/node.cpp
```diff
@@ -362,21 +363,45 @@ node_state::spawn_command(const invocation& inv,
...
+      return;
+    }
+    this_node->monitor(*component);
+    auto okay = this_node->state.registry.add(*component, std::move(comp_type),
+                                              std::move(label));
+    VAST_ASSERT(okay);
+    rp.deliver(*component);
+    return;
+  };
+  if (std::set<std::string>{"counter", "eraser", "exporter", "pivoter"}.count(
```
mavam, line 391: Mind putting this into a local variable? Looks a bit bulky this way to me.

libvast/src/system/node.cpp
```diff
@@ -362,21 +363,45 @@ node_state::spawn_command(const invocation& inv,
...
+        comp_type)
+      > 0u) {
+    if (auto tr_ = this_node->state.registry.find_by_label("type_registry")) {
```
mavam, line 394: We reserve `_` suffixes for member variables. I don't think this is need here; `tr` is not a keyword.

libvast/src/system/node.cpp
```diff
@@ -362,21 +363,45 @@ node_state::spawn_command(const invocation& inv,
...
+      auto tr = caf::actor_cast<type_registry_type>(tr_);
+      this_node
+        ->request(tr, defaults::system::initial_request_timeout, atom::get_v,
```
mavam, line 397: Why not send the type registry a `resolve` atom along with an expression? In my opinion, the resolution algorithms are part of the type registry API.

libvast/src/system/node.cpp
```diff
@@ -362,21 +363,45 @@ node_state::spawn_command(const invocation& inv,
...
-  // Spawn our new VAST component.
-  spawn_arguments args{spawn_inv, this_node->state.dir, label};
-  auto component = spawn_component(spawn_inv, args);
-  if (!component) {
-    if (component.error())
-      VAST_WARNING(__func__,
-                   "failed to spawn component:", render(component.error()));
-    rp.deliver(component.error());
-    return caf::make_message(component.error());
+  auto doit = [=](taxonomies_ptr t) mutable {
```
mavam, line 366:
```suggestion
  auto handle_taxonomies = [=](taxonomies_ptr t) mutable {
```

mavam (changes requested):
Good work so far.

libvast/vast/taxonomies.hpp
tobim: For our current and short-term planned usages of this information don't make use of the name of a taxonomy itself. On the other hand, keeping all concepts in a single map simplifies the resolution algorithm. For those reasons I opted to store all taxonomies combined in one struct.

libvast/src/system/node.cpp
tobim: That does not make sense to me, it just adds unnecessary coupling between `expression` and the `type_registry`. The next functionality that gets models support is `pivot` should that also be implemented at the `type_registry`?

mavam: I don't see any issues with the type registry to implement the necessary operations for pivoting between events. It's a natural extension of it's existing API.

libvast/vast/taxonomies.hpp
mavam: Is your argument "simplicity"? As a user, I'd like to be able to exchange/select taxonomies, but I do not see how this is supposed to be enhanced from here. Why not just have a single taxonomy throughout the code for now, and then add more cross-taxonomy logic later on?

libvast/test/taxonomies.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+using namespace vast;
+
+TEST(concepts - simple) {
+  auto c = concepts_type{{"foo", {"a.fo0", "b.foO", "x.foe"}},
+                         {"bar", {"a.bar", "b.baR"}}};
+  auto t = taxonomies{std::move(c), models_type{}};
+  {
+    MESSAGE("LHS");
+    auto exp = unbox(to<expression>("foo == \"1\""));
```
mavam, line 23: Why do you use a string on the RHS? It has to be escaped and makes it harder on the eyes than necessary.

libvast/test/taxonomies.cpp
```diff
@@ -0,0 +1,49 @@
...
+    auto ref = unbox(to<expression>("\"\" in a.fo0 || \"\" in b.foO || \"\" in "
+                                    "x.foe"));
+    auto result = resolve(t, exp);
+    CHECK_EQUAL(result, ref);
+  }
+}
+
+TEST(concepts - cyclic definition) {
+  auto c = concepts_type{{"foo", {"bar", "a.fo0", "b.foO", "x.foe"}},
+                         {"bar", {"a.bar", "b.baR", "foo"}}};
```
mavam, line 41: Can you write a comment that explains why we are supporting concepts, and what a potential use case is there?

libvast/vast/taxonomies.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace vast {
+
+/// Maps concept names to the fields or concepts that implement them.
+using concepts_type = std::unordered_map<std::string, std::vector<std::string>>;
+
+/// Maps model names to the concepts or models from which they are constituted.
```
mavam, line 30:
```suggestion
/// Maps model names to the concepts or models that implement them.
```

libvast/src/system/node.cpp
```diff
@@ -362,21 +363,50 @@ node_state::spawn_command(const invocation& inv,
...
+      return;
+    }
+    this_node->monitor(*component);
+    auto okay = this_node->state.registry.add(*component, std::move(comp_type),
+                                              std::move(label));
+    VAST_ASSERT(okay);
+    rp.deliver(*component);
+    return;
+  };
+  auto handle_taxonomies = [=](taxonomies_ptr t) mutable {
```
mavam, line 384: I think you have a race here. The pointer is effectively internal state in the type registry. What prevents the registry to mutate it while it's being looked here?

mavam (changes requested):
Let's sync up offline on this. The idea of handing out internal state (albeit immutable) that may change underneath breaks the actor model abstraction.

More generally, I would rather see the type registry encapsulate the operations to be performed on all type-related informations, as opposed to shipping the external state around.

libvast/src/system/node.cpp
tobim: This is does not apply anymore after the latest API change.

libvast/vast/system/type_registry.hpp
```diff
@@ -50,7 +51,7 @@ using type_registry_type = caf::typed_actor<
   caf::replies_to<atom::get>::with<type_set>,
   caf::replies_to<atom::get, std::string>::with<type_set>,
   caf::reacts_to<atom::put, vast::taxonomies>,
```
mavam, line 53: As discussed, I think it would be great if we could make `taxonomies` an implementation detail and not part of the public API.

```suggestion
  caf::reacts_to<atom::put, taxonomies>,
```

libvast/src/system/type_registry.cpp
```diff
@@ -165,9 +165,11 @@ type_registry(type_registry_actor self, const path& dir) {
     },
     [=](atom::put, taxonomies t) {
       VAST_TRACE("");
-      self->state.taxonomies = std::make_shared<const taxonomies>(std::move(t));
+      self->state.taxonomies = std::move(t);
+    },
+    [=](atom::resolve, expression e) {
```
mavam, line 170:
```suggestion
    [=](atom::resolve, const expression* e) {
```

libvast/src/system/node.cpp
```diff
@@ -397,10 +391,11 @@ node_state::spawn_command(const invocation& inv,
     = std::set<std::string>{"counter", "eraser", "exporter", "pivoter"};
   if (query_handlers.count(comp_type) > 0u) {
     if (auto tr = this_node->state.registry.find_by_label("type_registry")) {
+      auto expr = normalized_and_validated(spawn_inv.arguments);
       this_node
         ->request(caf::actor_cast<type_registry_type>(tr),
-                  defaults::system::initial_request_timeout, atom::get_v,
-                  atom::taxonomies_v)
+                  defaults::system::initial_request_timeout, atom::resolve_v,
+                  *expr)
```
mavam, line 398:
```suggestion
                  std::move(*expr))
```

libvast/src/system/node.cpp
```diff
@@ -381,14 +381,8 @@ node_state::spawn_command(const invocation& inv,
...
   };
-  auto handle_taxonomies = [=](taxonomies_ptr t) mutable {
-    if (t) {
-      auto expr = normalized_and_validated(spawn_inv.arguments);
-      if (!expr)
-        rp.deliver(expr.error());
-      auto resolved = resolve(*t, *expr);
-      spawn_inv.arguments = std::vector{to_string(resolved)};
-    }
+  auto handle_taxonomies = [=](expression e) mutable {
```
mavam, line 384:
```suggestion
  auto handle_taxonomies = [=](const expression& e) mutable {
```

mavam (changes requested):
It'd be great if you could fix CI in lockstep with moving forward on features/changes.

libvast/src/taxonomies.cpp
```diff
@@ -21,18 +21,43 @@
...
+      // A prefix of x matches an existing layout.
+      auto field = x.substr(pos + 1);
+      return std::any_of(i->second.value.begin(), i->second.value.end(),
+                         [&](const type& t) {
+                           if (auto r = caf::get_if<record_type>(&t)) {
+                             if (auto f = r->find(field))
+                               return compatible(f->type, op, data);
+                           }
+                           return false;
+                         });
```
mavam, line 48: It would a bit more readable if you factored the predicate into a named lambda.

libvast/vast/type_set.hpp
```diff
@@ -0,0 +1,31 @@
...
+namespace vast {
+
+struct type_set {
+  std::unordered_set<vast::type> value;
+
+  template <class Inspector>
+  friend auto inspect(Inspector& f, type_set& x) {
+    return f(caf::meta::type_name("type_set"), x.value);
+  }
+};
```
mavam, line 29: This looks like a one-off at a first glance. Why not just use the `unordered_set` directly?

mavam (changes requested):

libvast/vast/type_set.hpp
mavam: If you just want to return a list of types, I think even a `vector<type>` would be better suited.

libvast/vast/system/type_registry.hpp
tobim: As discussed in sync, we currently need this until the loading mechanism is in place.

libvast/vast/type_set.hpp
dominiklohmann: We must forward-declare it to define a type ID for it. We don't want to include further headers in `fwd.hpp`. This would work as well:

```c++
// fwd.hpp
namespace vast { struct type_set; }
VAST_DEFINE_TYPE_ID((vast::type_set));

// type_set.hpp
struct type_set : std::unordered_set<vast::type> {
  using unordered_set::unordered_set;
};
```

I think `s/std::unordered_set/std::stable_set` for the choice of container.

We use the same pattern as shown here in multiple places in the code base.

libvast/src/taxonomies.cpp
tobim: I tried and it doesn't. The formatting is the same bar indentation, but we get an additional indirection, so I'd rather keep it like that.

libvast/vast/taxonomies.hpp
tobim: > Why not just have a single taxonomy throughout the code for now, and then add more cross-taxonomy logic later on?

I implemented the feature according to its documentation. If I would follow your suggestion the end result would not work the way the user expects.

> As a user, I'd like to be able to exchange/select taxonomies, but I do not see how this is supposed to be enhanced from here.

This sounds like a useful feature but I would change it slightly to read: I want to set a default taxonomy, so that I don't have to prefix every concept/module name in an expression with it.

mavam (changes requested):
Looks good! Just CI needs a bit of love.

mavam (approved):

PR #1095, opened 2020-10-14
Add a function to parse concepts from a data object

tobim:


libvast/src/taxonomies.cpp
```diff
@@ -25,6 +26,46 @@
 
 namespace vast {
 
+static caf::error to_concepts_impl(const data& d, concepts_type& out) {
```
tobim, line 29: If we keep the error handling as is, this function should be called `extract_concepts` instead. I'm not sure which way we'll want to go with this yet.

libvast/src/taxonomies.cpp
```diff
@@ -25,6 +26,46 @@
...
+  if (const auto& tl = caf::get_if<list>(&d)) {
```
mavam, line 30:
```suggestion
  if (const auto& xs = caf::get_if<list>(&d)) {
```

libvast/src/taxonomies.cpp
mavam: Yeah, I like your proposal better.

libvast/src/taxonomies.cpp
```diff
@@ -25,6 +26,46 @@
...
+    for (const auto& item : *tl) {
+      if (const auto& x = caf::get_if<record>(&item)) {
+        auto n = x->find("concept");
+        if (n == x->end())
+          continue;
+        if (const auto& c = caf::get_if<record>(&n->second)) {
```
mavam, line 36: I think it would be nice to have a separate function that starts here, i.e., a `data` that is supposed to be a concept:

```cpp
caf::error convert(const data&, concept_type&);
```

That could then be used as a sub-routine in this function (e.g., `to<concept_type>(x)`. At that point, it is a conversion. 😉

libvast/src/taxonomies.cpp
```diff
@@ -25,6 +26,46 @@
...
+          auto n = c->find("name");
+          if (n == c->end())
+            return make_error(ec::convert_error, "concept has no name:", *x);
+          auto name = caf::get_if<std::string>(&n->second);
+          if (!name)
+            return make_error(ec::convert_error, "name is not a string:", *n);
+          auto fs = c->find("fields");
+          if (fs == x->end())
+            continue;
```
mavam, line 45: Is an empty list of fields allowed? I would say not.

mavam (changes requested):

libvast/vast/taxonomies.hpp
```diff
@@ -28,8 +28,14 @@ namespace vast {
...
 
+/// Converts a data object and fills it in out.
+caf::error convert(const data& d, concepts_type& out);
+
+/// Extracts a concept definition from a data object.
+caf::error extract_concepts(const data& d, concepts_type& out);
+
 /// Extracts a concept definition from a data object.
-caf::expected<concepts_type> to_concepts(const data& d);
+caf::expected<concepts_type> extract_concepts(const data& d);
```
mavam: Do we need both overloads? I could image that the latter suffices.

mavam (changes requested):
Looks good overall, just small nits.

libvast/vast/taxonomies.hpp
tobim: The former is required to accumulate concepts from multiple sources. In principle we could also create a separate `merge` function for `concepts_type`, but why shouldn't the function already behave the way we need it?

libvast/src/taxonomies.cpp
```diff
@@ -26,42 +26,49 @@
 
 namespace vast {
 
-static caf::error to_concepts_impl(const data& d, concepts_type& out) {
-  if (const auto& tl = caf::get_if<list>(&d)) {
-    for (const auto& item : *tl) {
+caf::error convert(const data& d, concepts_type& out) {
+  if (const auto& c = caf::get_if<record>(&d)) {
```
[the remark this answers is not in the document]
tobim: If `d` is a `const data` then `get_if<record>` will return a `const record&`, that can't be assigned to `auto&`. If we use just `auto` it will create a copy, so the only way to get rid of the const is to use `decltype(auto) c = caf::get_if<record>(&d)`, and I don't like that.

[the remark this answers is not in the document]
mavam: Fair enough.

mavam (approved):

PR #1099, opened 2020-10-16
Move index implementations into separate directory

mavam:
This PR splits the large `value_index.hpp` file into one file per derived class, and moves all derived class into the directory `libvast/vast/index/`.

Not there there are no changes to functionality or type naming.


mavam:
@dominiklohmann I added you as reviewer since you already peaked in here.

dominiklohmann (approved):
Thanks for refactoring! The new test seems to run through in CI as well.

mavam:
[story/ch18827]

PR #1100, opened 2020-10-19
Fix the literal printer for string literals

dominiklohmann:
Without this change, the constructor for the parser that takes a `bool` was preferrred, because implicit conversion from `const char*` to `bool` has a higher precedence than conversion to `std::string`.

This was caught by a warning introduced by clang-11 that is enabled by default: `-Wc++11-narrowing`.

### DIFF
--- a/libvast/vast/concept/printable/core/literal.hpp
+++ b/libvast/vast/concept/printable/core/literal.hpp
@@ -26,7 +26,7 @@ inline auto operator"" _P(char c) {
 }
 
 inline auto operator"" _P(const char* str) {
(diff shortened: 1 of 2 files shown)


mavam (approved):
Approving the fix without testing.

I wonder whether the `literal_printer` ctor should be explicit, though.

dominiklohmann:
> I wonder whether the `literal_printer` ctor should be explicit, though.

I agree that is a sensible change. I've pushed a separate commit. No code breaks because of this change, all instantiations were explicit already.

PR #1103, opened 2020-10-20
Age rotation for old data

lava:
Review instructions: Each commit can be reviewed individually, although 90% of the changes are in the last commit anyways.


libvast/vast/system/disk_monitor.hpp
```diff
@@ -0,0 +1,46 @@
...
+  /// Node handle of the ARCHIVE.
+  archive_type archive;
+
+  /// Node handle of the INDEX.
+  caf::actor index;
+
+  constexpr static const char* name = "disk_monitor";
+};
+
+disk_monitor_type::behavior_type
```
mavam, line 40: A comment here would be nice, just about the rough functionality this actor provides.

libvast/vast/system/disk_monitor.hpp
```diff
@@ -0,0 +1,46 @@
...
+disk_monitor(disk_monitor_type::stateful_pointer<disk_monitor_state> self,
+             size_t highwater, size_t lowwater,
+             std::chrono::seconds scan_interval, const vast::path& dbdir,
```
mavam, line 43:
```suggestion
             std::chrono::seconds scan_interval, const vast::path& db_dir,
```

libvast/src/system/index.cpp
```diff
@@ -741,6 +747,59 @@ index(caf::stateful_actor<index_state>* self, filesystem_type fs, path dir,
     },
     [=](atom::subscribe, atom::flush, const caf::actor& listener) {
       self->state.add_flush_listener(listener);
+    },
+    [=](atom::erase, uuid partition_id) {
+      VAST_VERBOSE(self, "erasing partition", partition_id);
```
mavam, line 752:
```suggestion
  VAST_VERBOSE(self, "erases partition", partition_id);
```

libvast/src/system/spawn_disk_monitor.cpp
```diff
@@ -0,0 +1,63 @@
...
+    = caf::get_or(args.inv.options, "vast.start.disk-budget-high", "0KiB");
+  auto lowater_str
+    = caf::get_or(args.inv.options, "vast.start.disk-budget-low", "0KiB");
+  uint64_t hiwater, lowater;
+  if (parsers::bytesize(hiwater_str, hiwater))
+    return make_error(ec::invalid_argument,
+                      "could not parse disk budget: " + hiwater_str);
+  if (parsers::bytesize(lowater_str, lowater))
+    return make_error(ec::invalid_argument,
+                      "could not parse disk budget: " + lowater_str);
```
mavam, line 42: Shouldn't these `if` statements be inverted? The parsers return `true` on success.

libvast/src/system/spawn_disk_monitor.cpp
```diff
@@ -0,0 +1,63 @@
...
+  if (!hiwater) {
+    VAST_VERBOSE(self, "not spawning disk_monitor because no limit configured");
```
mavam, line 44:
```
  VAST_VERBOSE(self, "doesn't spawn disk_monitor because of missing limit");
```

libvast/src/system/spawn_disk_monitor.cpp
```diff
@@ -0,0 +1,63 @@
...
+  auto db_dir = caf::get_or(args.inv.options, "vast.db-directory",
+                            defaults::system::db_directory);
+  auto abs_dir = path{db_dir}.complete();
+  if (!exists(abs_dir))
+    return make_error(ec::filesystem_error, "could not find database "
+                                            "directory");
+  auto handle = self->spawn(disk_monitor, hiwater, lowater,
+                            defaults::system::disk_scan_interval, abs_dir,
+                            caf::actor_cast<archive_type>(archive), index);
+  VAST_VERBOSE(self, "spawned a disk monitor");
```
mavam, line 59: Maybe DEBUG?

libvast/src/system/spawn_node.cpp
```diff
@@ -78,10 +78,12 @@ spawn_node(caf::scoped_actor& self, const caf::settings& opts) {
                [&](caf::error& e) { result = std::move(e); });
     return result;
   };
-  std::list components
-    = {"type-registry", "archive", "index", "importer", "eraser"};
-  if (accounting)
+  std::list components = {"type-registry", "archive", "index",
+                          "importer",      "eraser",  "disk_monitor"};
+  if (accounting) {
+    VAST_VERBOSE_ANON("pushing accountant");
```
mavam, line 84: This sounds like a DEBUG statement.

mavam: Maybe we can remove it altogether, though.

mavam (changes requested):
Great PR! I only have smaller things to discuss.

Since we're doing out-of-band testing with this PR, I have only looked at the code from a mechanical point of view.

libvast/src/directory.cpp
```diff
@@ -86,4 +86,23 @@ const path& directory::path() const {
   return path_;
 }
 
+// FIXME: Write unit test
```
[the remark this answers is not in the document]
lava: Ah, sorry, I was using the convention of writing `FIXME` for stuff I still want to change before the PR is merged and `TODO` for things that I'd be comfortable fixing at some point in the future.

libvast/src/system/spawn_disk_monitor.cpp
lava: Empirically they don't, unless I've messed something up while doing the PR cleanup. I'll double-check to be sure.

libvast/src/system/spawn_disk_monitor.cpp
lava: I feel like "missing limit" implies a user error or misconfiguration, so I went with a slightly more neutral message.

libvast/src/directory.cpp
```diff
@@ -86,4 +86,23 @@ const path& directory::path() const {
...
+size_t recursive_size(const vast::directory& dir) {
+  size_t size = 0;
+  std::vector<vast::directory> stack{dir};
```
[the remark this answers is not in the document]
lava: `std::stack` doesn't have an initializer list constructor ;)

The only reason for using `std::stack` that I can think of is for a non-const reference in a function signature; to guarantee that the function will only modify the container in a stack-like fashion.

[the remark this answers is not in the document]
mavam: Fair enough. I am not trying to shove it down your throat. 😉 The algorithm complexity is manageable.

libvast/src/system/spawn_disk_monitor.cpp
mavam: Yeah, please double-check. If you look at the framework, all parsers return a `bool` return value; `true` means parsing success and `false` failure.

libvast/src/system/spawn_disk_monitor.cpp
lava: Not sure, all the other components print the same on verbose:
```
2020-10-26T13:36:46.671 node spawned the type-registry
2020-10-26T13:36:46.671 node spawned the archive
2020-10-26T13:36:46.671 node spawned the index
```

mavam: Fair enough, that's convincing enough to leave it as is.

libvast/src/system/spawn_disk_monitor.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <caf/scoped_actor.hpp>
+#include <caf/settings.hpp>
+#include <caf/typed_event_based_actor.hpp>
+
+#include "caf/error.hpp"
+
+namespace vast::system {
+
+maybe_actor
+spawn_disk_monitor(system::node_actor* self, spawn_arguments& args) {
```
dominiklohmann, line 24: All the new options introduced here must be reflected in `vast.yaml.example`.

dominiklohmann (changes requested):
Some comments, mostly on log message wording.

This needs two changelog entries: A `change` for the archive options, and an `experimental feature` for the disk monitor. Please reflect all new and changed options in the example configuration.

The integration test should be disabled on macOS as discussed via Slack. `.DS_Store` files don't play well with how we're currently calculating the directory size and how the test is currently implemented.

libvast/src/system/spawn_disk_monitor.cpp
```diff
@@ -0,0 +1,68 @@
...
+  auto lowater_str
+    = caf::get_or(args.inv.options, "vast.start.disk-budget-low", "0KiB");
+  // TODO: This
+  auto default_seconds
+    = std::chrono::seconds{defaults::system::disk_scan_interval}.count();
+  auto interval = caf::get_or(
+    args.inv.options, "vast.start.disk-budget-check-interval", default_seconds);
+  uint64_t hiwater, lowater;
+  if (!parsers::bytesize(hiwater_str, hiwater))
+    return make_error(ec::invalid_argument,
```
mavam, line 43:
```suggestion
    return make_error(ec::parse_error,
```

libvast/src/system/spawn_disk_monitor.cpp
```diff
@@ -0,0 +1,68 @@
...
+                      "could not parse disk budget: " + hiwater_str);
+  if (!parsers::bytesize(lowater_str, lowater))
+    return make_error(ec::invalid_argument,
```
mavam, line 46:
```suggestion
    return make_error(ec::parse_error,
```

libvast/vast/concept/parseable/vast/si.hpp
```diff
@@ -55,6 +56,7 @@ namespace parsers {
 
 static auto const count = si_parser<vast::count>{};
 static auto const integer = si_parser<vast::integer>{};
+static auto const bytesize = count >> ~ch<'B'>;
```
mavam, line 59: Just to make sure: is the intention that we *both* support `42k` and `42kB`?

libvast/vast/directory.hpp
```diff
@@ -19,6 +19,12 @@
 
 #ifdef VAST_POSIX
 #  include <dirent.h>
+#else
+
+namespace vast {
+struct DIR;
```
mavam, line 25: Should this really go into the public namespace? It seems like an implementation detail.

CHANGELOG.md
```diff
@@ -12,6 +12,17 @@ Every entry has a category for which we use the following visual abbreviations:
...
+- 🧬 A new experimental "disk monitor" feature was added that allows operators
+  to specify a maximum amount of disk space that can be used by VAST. Once
+  the database directory exceeds that limit, the oldest data will be erased
+  from the database to make space. The new command-line options `--disk-quota-high`,
+  `--disk-quota-low` and `--disk-quota-check-interval` were added for that purpose.
+  [#1103](https://github.com/tenzir/vast/pull/1103)
+
+- 🎁 New command-line options `--segments` and `--max-segment-size` were added
+  to control some details of the segments produced by the archive.
+  [#1103](https://github.com/tenzir/vast/pull/1103)
```
mavam, line 24:
```suggestion
- 🧬 A new *disk monitor* component can now monitor the database size and
  delete data that exceeds a specified threshold. Once VAST reaches the maximum
  amount of disk space, the disk monitor deletes the oldest data. The
  command-line options `--disk-quota-high`, `--disk-quota-low`, and
  `--disk-quota-check-interval` control the rotation behavior.
  [#1103](https://github.com/tenzir/vast/pull/1103)

- 🎁 The new command-line options `--segments` and `--max-segment-size` control
  how the archive generates segments.
  [#1103](https://github.com/tenzir/vast/pull/1103)
```

mavam (changes requested):
Looks good from a 30k-foot view now. Just some cosmetic changes.

I didn't check the application logic or test the commit, it would be great if somebody else from @tenzir/backend can take care of that.

CHANGELOG.md
dominiklohmann: I'd replace the second one with this:

> The new options `vast.segments` and `vast.max-segment-size` control how the archive generates segments.

The reason being that we do this everywhere else, and that the options include the hierarchy.

dominiklohmann:
> @mavam
> I didn't check the application logic or test the commit, it would be great if somebody else from tenzir/backend can take care of that.

I've verified this to work locally on macOS. It's fine for an experimental feature, I think.

libvast/vast/concept/parseable/vast/si.hpp
lava: Yes, to be consistent with other command-line tools (e.g. `dd`), and also because there's no ambiguity so it doesn't hurt to accept generously.

libvast/vast/directory.hpp
lava: Can you think of a better alternative? I did't want to put it in the global namespace since the posix `DIR` already lives there when it exists, and I don't even want to think about the ODR can of worms we're going to open when we put this into an unnamed namespace but let `directory` have a `DIR*` member.

mavam: I would put it at least into namespace `vast::detail`. Namespace `vast` should be as stable as possible.

Alternatively, you can do type erasure and use a void pointer, which you then cast with helper functions in the implementation.

integration/default_set.yaml
```diff
@@ -376,3 +398,34 @@ tests:
...
+  # the numbers of the fixture must be set up quite precisely to trigger the
+  # disk monitor at the last import, but on the other hand it will also allow
+  # us to catch changes that drastically change the space used by the on-disk
+  # layout of the VAST data structures.
+  Disk monitor:
+    tags: [disk-quota]
+    fixture: DiskMonitorTester
+    # Disable on MacOS, since the additional `.DS_store` files throw off the size
+    # calculation.
+    condition: version | uname -s | grep Linux
```
dominiklohmann, line 413:
```suggestion
    condition: version | uname -s | grep -v Darwin
```

This matches the above comment better.

CHANGELOG.md
```diff
@@ -12,6 +12,17 @@ Every entry has a category for which we use the following visual abbreviations:
...
 ## Unreleased
 
+- 🧬 A new *disk monitor* component can now monitor the database size and
+  delete data that exceeds a specified threshold. Once VAST reaches the maximum
+  amount of disk space, the disk monitor deletes the oldest data. The
+  command-line options `--disk-quota-high`, `--disk-quota-low`, and
+  `--disk-quota-check-interval` control the rotation behavior.
+  [#1103](https://github.com/tenzir/vast/pull/1103)
+
+- 🎁  The new options vast.segments and vast.max-segment-size control how the
```
dominiklohmann, line 22:
```suggestion
- 🎁  The new options `vast.segments` and `vast.max-segment-size` control how the
```

libvast/src/directory.cpp
```diff
@@ -19,6 +19,8 @@
 #include "vast/detail/posix.hpp"
 #include "vast/detail/string.hpp"
 
+#include "vast/logger.hpp"
```
dominiklohmann, line 22: This doesn't look properly formatted, this should be joined with the group above.

libvast/src/system/disk_monitor.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include "vast/error.hpp"
+#include "vast/fwd.hpp"
+#include "vast/logger.hpp"
+#include "vast/path.hpp"
+#include "vast/system/archive.hpp"
+#include "vast/uuid.hpp"
+
+#include <caf/typed_event_based_actor.hpp>
+
+#include "caf/detail/scope_guard.hpp"
```
dominiklohmann, line 15:
```suggestion
#include <caf/detail/scope_guard.hpp>
```

Again, wrong header grouping. Please run `git format-branch`.

dominiklohmann (changes requested):

libvast/src/system/disk_monitor.cpp
lava: `git format-branch` doesnt catch this.

dominiklohmann: It catches the wrong grouping if and only if you are using angle brackets.

libvast/vast/directory.hpp
lava: But in the posix case, we don't want to have a `detail::DIR*`, we need a `DIR*`. And this is in a header, so we cant just use the namespace. I'm not a fan of using a `void*`, since we *know* the type of the pointer.

If you feel strongly about this, I think the cleanest alternatives would be to either
 - declare a wrapper struct in namespace `detail` and give it a `DIR*` member on posix (ie. one extra indirection)
 - put it in the global namespace after all, or
 - `#define DIR void` in the non-posix case

mavam: I don't have too strong of an opinion, because our hand-rolled filesystem header will evaporate soon after switching  to `std::filesystem`.

If you're not keen on type erasure using the poor-man's void* pimpling, let's just leave it as. 🙂

mavam (approved):

PR #1105, opened 2020-10-20
Enable versioning of table slice encodings

dominiklohmann:
**NOTE:** Do not review until undrafted. I'm still working on this.

#### Remaining tasks before PR is reviewable
- [ ] Switch to using the new table slice API completely
- [ ] Assign offset when reading table slices from segments
- [ ] Fix handling of `VAST_HAVE_ARROW` for builds without Apache Arrow
- [ ] Rebase onto latest master
- [ ] Add changelog entry

#### Remaining tasks before merging
- [ ] Remove now unnecessary includes
- [ ] Sanity-check how `table_slice` is passed around in functions, stream handlers, and actor behaviors
- [ ] Test using gcc-8 (using Dockerfile)
- [ ] Test for correctness
- [ ] Test for performance
- [ ] Update documentation if required
- [ ] Pass reviews

#### Review Instructions
When reviewing, please look at the final APIs first, but review commit-by-commit and make sure to read the commit messages. Some code was added only to be removed later on to ensure that every commit that is not explicitly labeled `WIP:` can be built.


libvast/src/segment.cpp
```diff
@@ -80,19 +86,25 @@ chunk_ptr segment::chunk() const {
...
+    return std::pair{slice.offset(), slice.offset() + slice.rows()};
   };
   auto g = [&](auto buffer) -> caf::error {
     // TODO: bind the lifetime of the table slice to the segment chunk. This
     // requires that table slices will be constructable from a chunk. Until
     // then, we stupidly deserialize the data into a new table slice.
-    table_slice_ptr slice;
-    if (auto err = unpack(*buffer->data_nested_root(), slice))
+    table_slice slice;
+    if (auto err = unpack(*buffer, slice))
```
tobim, line 107: This needs to be changed to use `chunk_.slice()` with the appropriate offset / length pair and construct the `table_slice` from the returned `chunk_ptr`.

tobim (changes requested):

PR #1109, opened 2020-10-22
Fix merging of source status objects

dominiklohmann:

Unlike for other merge operations, for the status we want to recurse deep into the object structure and merge array contents instead of overwriting arrays.  This fixes an issue that caused only the sources' status to show up in the output of `vast status --detailed` that replied last.

I verified this to work locally. We don't have an integration test that covers this.

###  :memo: Checklist

Please run this locally for testing. We do not have an integration test that covers this.

### DIFF
--- a/libvast/src/detail/settings.cpp
+++ b/libvast/src/detail/settings.cpp
@@ -32,7 +32,7 @@ void merge_settings_impl(const caf::settings& src, caf::settings& dst,
       merge_settings_impl(caf::get<caf::settings>(value),
                           dst[key].as_dictionary(), policy, depth + 1);
     } else {
-      if constexpr (std::is_same_v<Policy, policy::deep_tag>) {
+      if constexpr (std::is_same_v<Policy, policy::merge_lists_tag>) {
         if (caf::holds_alternative<caf::config_value::list>(value)) {
           const auto& src_list = caf::get<caf::config_value::list>(value);
           if (caf::holds_alternative<caf::config_value::list>(dst[key])) {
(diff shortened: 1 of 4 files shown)


libvast/vast/detail/settings.hpp
```diff
@@ -15,11 +15,27 @@
 
 #include <caf/settings.hpp>
 
+namespace vast::policy {
+
+struct deep_tag {};
+struct shallow_tag {};
```
tobim, line 21: I fail to see the case for this `shallow` policy. A tree merge algorithm has to be recursive, if it isn't, then it should not be called `merge`.

tobim (changes requested):

libvast/vast/detail/settings.hpp
dominiklohmann: Did you see the discussion in the linked story in ClubHouse?

We use `merge_settings` in other places as well, e.g., when merging configuration options. We probably don't want to merge arrays there.

tobim: I understood it wrongly at first.

How about `merge_lists_tag` and `overwrite_lists_tag` instead?

libvast/src/detail/settings.cpp
```diff
@@ -13,33 +13,57 @@
...
+          const auto& src_list = caf::get<caf::config_value::list>(value);
+          if (caf::holds_alternative<caf::config_value::list>(dst[key])) {
+            auto& dst_list = dst[key].as_list();
+            dst_list.insert(dst_list.end(), src_list.begin(), src_list.end());
+          } else {
+            dst.insert_or_assign(key, src_list);
+          }
+        } else {
+          dst.insert_or_assign(key, value);
+        }
```
tobim, line 46: Not sure this will work correctly:
```suggestion
      if constexpr (std::is_same_v<Policy, policy::deep_tag>) {
        if (caf::holds_alternative<caf::config_value::list>(value) &&
            caf::holds_alternative<caf::config_value::list>(dst[key])) {
          const auto& src_list = caf::get<caf::config_value::list>(value);
          auto& dst_list = dst[key].as_list();
          dst_list.insert(dst_list.end(), src_list.begin(), src_list.end());
        } else {
          dst.insert_or_assign(key, value);
        }
```

tobim (approved):

PR #1110, opened 2020-10-22
Don't enable automatic connections by default

lava:
Disable the `middleman.enable-automatic-connections` option by default to prevent VAST from opening a random public port, and since we're not currently linking multiple VAST instances in a mesh anyways.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 VAST no longer opens a random public port, which used to be enabled in
+  the experimental VAST cluster mode in order to transparently establish a
+  full mesh.
+  [#1110](https://github.com/tenzir/vast/pull/1110)
+
 - 🧬 The query language now comes with support for concepts, the first part of
   taxonomies. Concepts is a mechanism to unify the various naming schemes of
   different data formats into a single, coherent nomenclature.
--- a/libvast/src/system/default_configuration.cpp
+++ b/libvast/src/system/default_configuration.cpp
@@ -31,8 +31,6 @@ default_configuration::default_configuration() {
   set("logger.console-verbosity", defaults::logger::console_verbosity);
   set("logger.file-verbosity", defaults::logger::file_verbosity);
   set("logger.file-format", defaults::logger::file_format);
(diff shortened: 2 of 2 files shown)


libvast/src/system/default_configuration.cpp
```diff
@@ -31,8 +31,8 @@ default_configuration::default_configuration() {
   set("logger.console-verbosity", defaults::logger::console_verbosity);
   set("logger.file-verbosity", defaults::logger::file_verbosity);
   set("logger.file-format", defaults::logger::file_format);
-  // Allow VAST clusters to form a mesh.
-  set("middleman.enable-automatic-connections", true);
+  // Prevent broker from opening a random public port by default.
+  set("middleman.enable-automatic-connections", false);
```
tobim, line 35: Why don't we just remove these lines instead? `false` is the default after all.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 VAST no longer opens a random public port by default.
```
mavam, line 14: This is a somewhat pathetic statement on its own. Could we add some context for the reader for why this happened in the past?

mavam (changes requested):

libvast/src/system/default_configuration.cpp
lava: But then we'd silently reintroduce the bug if the upstream default changes.

CHANGELOG.md
lava: Do you have a suggestion? Something like this maybe?

```
VAST no longer automatically accepts external connections when configured with broker.
```

libvast/src/system/default_configuration.cpp
tobim: This is a question of policy. Either we rely on the defaults or we set everything that affects vast explicitly. So far we've been doing the former.

CHANGELOG.md
mavam: Here's a twist:

```
VAST no longer opens a random public port, which was used in the experimental VAST cluster mode in order to transparently establish a full mesh. 
```

mavam (approved):

PR #1111, opened 2020-10-22
Fix an ODR violation in the terminator

dominiklohmann:

The terminator template is explicitly instantiated in one translation unit, but not in others that use it. The header file must explicitly state that it can be safely relied upon that the terminator is instantiated in some translation unit by using `extern template`.

###  :memo: Checklist

Read [§ 13.9.3 Explicit instantiation](http://eel.is/c++draft/temp.explicit) and make sure I am correct here.

### DIFF
--- a/libvast/vast/system/terminator.hpp
+++ b/libvast/vast/system/terminator.hpp
@@ -45,4 +45,14 @@ caf::behavior terminator(caf::stateful_actor<terminator_state>* self,
                          std::chrono::milliseconds grace_period,
                          std::chrono::milliseconds kill_timeout);
 
+extern template caf::behavior
+terminator<policy::sequential>(caf::stateful_actor<terminator_state>*,
+                               std::chrono::milliseconds,
+                               std::chrono::milliseconds);
+
+extern template caf::behavior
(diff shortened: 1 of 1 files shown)


mavam (approved):

tobim (commented):
Can you clarify why this is an ODR violation? As I see it the definition is only available in one TU?

dominiklohmann:
> Can you clarify why this is an ODR violation? As I see it the definition is only available in one TU?

It has exactly zero definitions in all but that one translation unit that instantiated the templates explicitly, causing linker failures. ODR violations go both ways: Either zero or more than one are both errors.

tobim:
What linker failure are you seeing and why did it pass CI in that case?

dominiklohmann:
We compile with LTO by default. If you disable LTO, you _may_ get a linker error, because some uses of the terminate function cannot see the definitions of the instantiated templates.

dominiklohmann:
After further consultation of [Explicit instantiation](https://en.cppreference.com/w/cpp/language/class_template#Class_template_instantiation), @tobim and I concluded that this is not actually required, because the template definition is not in the header file itself.

PR #1113, opened 2020-10-23
Create a version file for the database dir

lava:

n/t

As you wish.

### DIFF
--- a/libvast/src/layout_version.cpp
+++ b/libvast/src/layout_version.cpp
@@ -1,0 +1,73 @@
+#include "vast/layout_version.hpp"
+
+#include "vast/detail/assert.hpp"
+#include "vast/error.hpp"
+#include "vast/io/read.hpp"
+#include "vast/io/write.hpp"
+#include "vast/logger.hpp"
+
+#include <caf/expected.hpp>
+
+#include <fstream>
+#include <iterator>
+
+namespace vast {
+
+namespace {
+
+const char* descriptions[]{
+  "invalid",
+  "v0",
+};
+
+static_assert(layout_version{std::size(descriptions)} == layout_version::count,
+              "Mismatch between number of layout versions and descriptions");
+
+const char* to_string(layout_version v) {
+  return descriptions[static_cast<uint8_t>(v)];
+}
+
+} // namespace
+
+layout_version read_layout_version(const vast::path& dbdir) {
+  if (!exists(dbdir))
+    return layout_version::invalid;
+  auto versionfile = dbdir / "VERSION";
+  auto contents = io::read(versionfile);
+  if (!contents)
+    return layout_version::invalid;
+  if (contents->empty())
+    return layout_version::invalid;
+  std::string_view sv{reinterpret_cast<const char*>(contents->data()),
+                      contents->size()};
+  // Only read until first newline.
+  if (auto first_newline = sv.find('\n'))
+    sv = std::string_view{sv.data(), first_newline};
+  auto begin = std::begin(descriptions);
+  auto end = std::end(descriptions);
+  auto it = std::find(begin, end, sv);
+  if (it == end)
+    return layout_version::invalid;
+  return static_cast<layout_version>(std::distance(begin, it));
+}
+
+caf::error initialize_layout_version(const vast::path& dbdir) {
+  if (!exists(dbdir))
+    return make_error(ec::filesystem_error,
+                      "db-directory does not exist:", dbdir.str());
+  auto version_path = dbdir / "VERSION";
+  // Do nothing if a VERSION file already exists.
+  if (exists(version_path))
+    return ec::no_error;
+  std::ofstream fs(version_path.str());
+  fs << to_string(layout_version::v0) << std::endl;
+  if (!fs)
+    return make_error(ec::filesystem_error, "could not write version file");
+  return ec::no_error;
+}
+
+std::ostream& operator<<(std::ostream& str, const layout_version& version) {
+  return str << to_string(version);
+}
+
+} // namespace vast
--- a/libvast/src/system/spawn_node.cpp
+++ b/libvast/src/system/spawn_node.cpp
@@ -17,6 +17,7 @@
 #include "vast/concept/parseable/vast/time.hpp"
 #include "vast/defaults.hpp"
 #include "vast/detail/pid_file.hpp"
+#include "vast/layout_version.hpp"
 #include "vast/logger.hpp"
 #include "vast/scope_linked.hpp"
 #include "vast/system/node.hpp"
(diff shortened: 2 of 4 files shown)


libvast/src/layout_version.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include "vast/detail/assert.hpp"
+#include "vast/error.hpp"
+#include "vast/io/read.hpp"
+#include "vast/io/write.hpp"
+#include "vast/logger.hpp"
+
+#include <fstream>
+#include <iterator>
+
+#include "caf/expected.hpp"
```
dominiklohmann, line 12:
```suggestion
#include <caf/expected.hpp>
```

libvast/src/layout_version.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+static_assert(vast::layout_version{std::size(descriptions)}
+                == vast::layout_version::count,
+              "Mismatch between number of layout versions and descriptions");
+
+} // namespace
+
+namespace vast {
+
+// This must be in namespace `vast` so it can be found by overload resolution.
```
dominiklohmann, line 29: That sounds like a bug, this function should be picked up via ADL.

libvast/src/layout_version.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include "vast/io/read.hpp"
+#include "vast/io/write.hpp"
+#include "vast/logger.hpp"
+
+#include <fstream>
+#include <iterator>
+
+#include "caf/expected.hpp"
+
+namespace {
```
dominiklohmann, line 14: Can you move this anonymous namespace inside the `vast` namespace such that you don't need to specify `vast::` everywhere?

libvast/src/layout_version.cpp
```diff
@@ -0,0 +1,74 @@
...
+  auto end = std::end(descriptions);
+  auto it = std::find(begin, end, sv);
+  if (it == end)
+    return layout_version::invalid;
+  return static_cast<layout_version>(std::distance(begin, it));
+}
+
+caf::error initialize_layout_version(const vast::path& dbdir) {
+  if (!exists(dbdir))
+    return make_error(ec::filesystem_error, "directory does not exist");
```
dominiklohmann, line 58:
```suggestion
    return make_error(ec::filesystem_error, "db-directory does not exist:", dbdir.str());
```

libvast/src/system/spawn_node.cpp
```diff
@@ -40,11 +41,23 @@ spawn_node(caf::scoped_actor& self, const caf::settings& opts) {
...
   auto abs_dir = path{db_dir}.complete();
-  if (!exists(abs_dir))
+  if (!exists(abs_dir)) {
     if (auto err = mkdir(abs_dir))
       return make_error(ec::filesystem_error,
                         "unable to create db-directory:", abs_dir.str(),
                         err.context());
+  }
+  // Write VERSION file if it doesnt exist yet. Note that an empty db dir
+  // often already exists before the node is initialized, e.g. when the log
```
dominiklohmann, line 51:
```suggestion
  // often already exists before the node is initialized, e.g., when the log
```

libvast/src/system/spawn_node.cpp
```diff
@@ -40,11 +41,23 @@ spawn_node(caf::scoped_actor& self, const caf::settings& opts) {
...
+  // output is written into the same directory.
+  if (auto err = initialize_layout_version(abs_dir))
+    return err;
+  // TODO: It might more sense to do this check in the places where
+  // we actually rely on the disk layout to do something (e.g.
+  // disk_monitor or index) and not during startup.
+  if (read_layout_version(abs_dir) != layout_version::v0)
+    return make_error(ec::filesystem_error, "wrong or missing layout version "
+                                            "in db dir");
```
dominiklohmann, line 60:
```suggestion
                                            "in db-directory");
```

libvast/vast/layout_version.hpp
```diff
@@ -0,0 +1,60 @@
...
+//     [...]
+//   archive/
+//     segments/
+//       d9513a65-9349-4c47-8e6d-4dcecf99327d - Segment flatbuffers
+//       [...]
+//   importer/
+//     current_id_block                       - The persisted current id block.
+//   type-registry/
+//     type-registry                          - Serialized state of the type
+//     registry
```
dominiklohmann, line 41: This looks like it did not survive text reflow

libvast/src/system/spawn_node.cpp
```diff
@@ -40,11 +41,23 @@ spawn_node(caf::scoped_actor& self, const caf::settings& opts) {
...
                         err.context());
+  }
+  // Write VERSION file if it doesnt exist yet. Note that an empty db dir
+  // often already exists before the node is initialized, e.g. when the log
+  // output is written into the same directory.
+  if (auto err = initialize_layout_version(abs_dir))
+    return err;
+  // TODO: It might more sense to do this check in the places where
+  // we actually rely on the disk layout to do something (e.g.
+  // disk_monitor or index) and not during startup.
```
dominiklohmann, line 57: Can you create a follow-up story for this?

libvast/src/layout_version.cpp
```diff
@@ -0,0 +1,74 @@
...
+                      contents->size()};
+  // Only read until first newline.
+  if (auto first_newline = sv.find('\n'))
+    sv = std::string_view{sv.data(), first_newline};
+  auto begin = std::begin(descriptions);
+  auto end = std::end(descriptions);
+  auto it = std::find(begin, end, sv);
+  if (it == end)
+    return layout_version::invalid;
+  return static_cast<layout_version>(std::distance(begin, it));
```
dominiklohmann, line 53: Can you explain to me how exactly this works? Reading an integer from a file should be quite simple, but why are you using the distance between two iterators here?

dominiklohmann (changes requested):

libvast/src/layout_version.cpp
lava: In the current implementation its not an integer but an enum value - it's using the `descriptions` table to stringify when its written, and using the index inside the description table to destringify when its read.

libvast/src/layout_version.cpp
lava: I think ADL works in the opposite direction? Ie. we could pick up additional functions in `vast::` if the argument is a `vast::layout_version`, but not additional functions from an unnamed namespace.

dominiklohmann: I can't see any longer how it was because you force-pushed, but it seems to work now. Let's resolve this and discuss later when revisiting.

dominiklohmann (approved):
:shipit:

PR #1128, opened 2020-10-29
Replace PID file if process does not exist

tobim:

Change to PID file check so that it automatically gets replaced if the owning process does not exist any more.

###  :memo: Checklist

Please test this change locally.
1. Try to run `vast -d somedir start` 2 times in parallel, without stopping the first process.
2. Try to run `vast -d somedir start` 2 times sequentially; kill the first process before spawning the second.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
-- ⚠️ VAST no longer requires you to manually remove the PID file it was left
-  behind from a previous instance.
+- ⚠️ VAST no longer requires you to manually remove a stale PID file from a
+  no-longer running `vast` process. Instead, VAST prints a warning and
+  overwrites the old PID file.
   [#1128](https://github.com/tenzir/vast/pull/1128)
 
 ## [2020.10.29]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,11 @@ Every entry has a category for which we use the following visual abbreviations:
 - ⚡️ breaking change
 - 🐞 bugfix
 
-<!-- ## Unreleased -->
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


libvast/src/detail/pid_file.cpp
```diff
@@ -52,9 +66,9 @@ caf::error acquire_pid_file(const path& filename) {
                       "failed in flock(2):", strerror(errno));
   }
   // Write the PID in human readable form into the file.
-  auto pid = to_string(process_id());
-  VAST_ASSERT(!pid.empty());
-  if (::write(fd, pid.data(), pid.size()) < 0) {
+  auto pid_string = to_string(pid);
+  VAST_ASSERT(!pid_string.empty());
+  if (::write(fd, pid_string.data(), pid_string.size()) < 0) {
```
mavam, line 71: We could use `io::write` here, now that we are touching this piece of code.

mavam (changes requested):
Looks good. I'd just add a changelog entry.

libvast/src/detail/pid_file.cpp
tobim: We won't be able to use `flock` if we do that.

mavam: Ah, right.

CHANGELOG.md
```diff
@@ -10,7 +10,11 @@ Every entry has a category for which we use the following visual abbreviations:
 - ⚡️ breaking change
 - 🐞 bugfix
 
-<!-- ## Unreleased -->
+## Unreleased
+
+- ⚠️ VAST no longer requires you to manually remove the PID file it was left
+  behind from a previous instance.
```
```suggestion
- ⚠️ VAST no longer requires you to manually remove a stale PID file from a
  no-longer running `vast` process. Instead, VAST prints a warning and
  overwrites the old PID file.
```

mavam (approved):
Ready to merge, modulo a request for the changelog entry.

PR #1130, opened 2020-10-31
Add base class for FlatBuffers root table wrappers

dominiklohmann:

This PR adds a convenience base class for classes that wrap FlatBuffers root tables using `chunk_ptr`, and a corresponding builder. They function as improved versions of the existing `fbs::as_flatbuffer<Root>` and `fbs::release` respectively.

Additionally, the `segment` and `segment_builder` were re-written to use the new utilities.

This is purely a refactoring: The behavior of the segment and segment builder are unchanged.

###  :memo: Checklist

Review commit-by-commit. The commit messages contain a fair bit of information.

### DIFF
--- a/libvast/vast/fbs/table.hpp
+++ b/libvast/vast/fbs/table.hpp
@@ -24,13 +24,24 @@ namespace vast::fbs {
 /// Semi-regular base class for types that wrap a FlatBuffers table.
 /// @tparam Derived The type that wraps the FlatBuffers table for CRTP.
 /// @tparam Root The generated FlatBuffers table root type.
-template <class Derived, class Root>
+/// @tparam FileIdentifier A pointer to the function that returns the
+/// FlatBuffers table's file identifier.
+template <class Derived, class Root, const char* (*FileIdentifier)()>
 class table {
 public:
   // -- types and constants ----------------------------------------------------
 
   using derived_type = Derived;
   using root_type = Root;
+
+  /// A pointer to the function that returns the underlying FlatBuffers table's
+  /// file identifier.
+  inline static constexpr auto file_identifier = FileIdentifier;
+
+  /// Returns the underlying FlatBuffers table's fully qualified table name.
+  static constexpr const char* table_name() {
+    return root_type::GetFullyQualifiedName();
+  }
 
   // -- constructors, destructors, and assignment operators --------------------
 
@@ -44,7 +55,7 @@ public:
     if (chunk) {
       auto verifier = flatbuffers::Verifier{
         reinterpret_cast<const uint8_t*>(chunk->data()), chunk->size()};
-      if (verifier.template VerifyBuffer<root_type>())
+      if (verifier.template VerifyBuffer<root_type>(file_identifier()))
         chunk_ = std::move(chunk);
     }
   }
--- a/libvast/vast/fbs/table.hpp
+++ b/libvast/vast/fbs/table.hpp
@@ -17,6 +17,11 @@
 #include "vast/detail/assert.hpp"
 #include "vast/fwd.hpp"
 
+#include <caf/error.hpp>
+#include <caf/meta/load_callback.hpp>
+#include <caf/meta/save_callback.hpp>
+#include <caf/meta/type_name.hpp>
+
 #include <flatbuffers/flatbuffers.h>
 
 namespace vast::fbs {
@@ -45,7 +50,7 @@ public:
 
   // -- constructors, destructors, and assignment operators --------------------
 
-  /// Default-constructs an invalid FlatBuffers table. This
+  /// Default-constructs an invalid FlatBuffers table.
   /// @post `!*this`
   table() noexcept = default;
 
@@ -71,6 +76,24 @@ public:
   /// Default destructor.
   virtual ~table() noexcept = default;
 
+  // -- concepts ---------------------------------------------------------------
+
+  /// Return a view on the underlying byte buffer.
+  /// @pre `x`
+  friend span<const byte> as_bytes(const derived_type& x) noexcept {
+    VAST_ASSERT(x);
+    return as_bytes(x.chunk_);
+  }
+
+  /// Serialize/deserialize directly from/to the underlying byte buffer.
+  template <class Inspector>
+  friend auto inspect(Inspector& f, derived_type& x) ->
+    typename Inspector::result_type {
+    return f(caf::meta::type_name(table_name()), x.chunk_,
+             caf::meta::load_callback([&] { return x.load(); }),
+             caf::meta::save_callback([&] { return x.save(); }));
+  }
+
   // -- properties -------------------------------------------------------------
 
   /// Access the underlying FlatBuffers root table.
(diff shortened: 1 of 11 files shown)


lava (commented):
Can you maybe give a high-level explanation why you went for CRTP in `vast::table`, and what the benefit of having a separate `table_builder` class is? I would have intuitively expected a `flatbuffer_ptr` to be something `unique_ptr`-like.

dominiklohmann:
> Can you maybe give a high-level explanation why you went for CRTP in `vast::table`, 

The CRTP enables opt-in to concepts like `inspect` and `as_bytes` for the derived type, and makes the builder itself possible.

> and what the benefit of having a separate `table_builder` class is? 

The builder is a nice wrapper around file identifier and releasing of the chunk. It's only purpose is to make it easier to write builders for FlatBuffers types without having to know too much of the FlatBuffers API. 

> I would have intuitively expected a `flatbuffer_ptr` to be something `unique_ptr`-like.

It can't be, some of the types must be copyable.

libvast/vast/fbs/table.hpp
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+private:
+  // -- implementation details -------------------------------------------------
+
+  /// Access the underlying FlatBuffers root table.
+  /// @pre `*this`
+  /// @post `result != nullptr`
+  const root_type* root() const noexcept {
+    VAST_ASSERT(*this);
```
lava, line 161: `VAST_ASSERT(this)`

dominiklohmann: That's wrong. We want to check whether the `chunk_` is set, not the `this` pointer of the wrapping table.

lava: Ah, right. `VAST_ASSERT(chunk_)` might be a bit less confusing.

dominiklohmann: I intentionally used the assertion as is, because it matches the pre-condition. The chunk may not be exposed to the user.

libvast/vast/fbs/table_builder.hpp
```diff
@@ -0,0 +1,113 @@
...
+  using offset_type = typename flatbuffers::Offset<root_type>;
+
+  /// A pointer to the function that returns the underlying FlatBuffers table's
+  /// file identifier.
+  inline static constexpr auto file_identifier = table_type::file_identifier;
+
+  // -- constructors, destructors, and assigmnent operators --------------------
+
+  /// Construct a builder with an initial buffer size.
+  /// @param initial_size The initial buffer size in Bytes.
```
lava, line 45: I'm not really convinced that there's a need for this class: As far as I can tell, its only purpose is to call `builder_.Finish()` with the correct file identifier, but in a non-generic class one would call `builder_.FinishSegment()` and this would be even safer because flatbuffers code will automatically select the correct identifier based on the information in the flatbuffer schema.

As a tradeoff, we have a virtual base class and a hard-to-describe control flow (i.e. user must call `segment_builder::add()` followed by `table_builder::finish()`, and the `create()` method directly accesses the builder from the base class)

And furthermore, it seems like this would be hard to compose, ie. how would I be able to create a non-root level segment flatbuffer as part of some existing builder?

libvast/vast/fbs/table.hpp
```diff
@@ -0,0 +1,171 @@
...
+  /// @post `!*this`
+  table() noexcept = default;
+
+  /// Constructs and verifies a FlatBuffers table from a chunk.
+  /// @note Constructs an invalid table if the chunk fails verification.
+  explicit table(chunk_ptr&& chunk) noexcept {
+    if (chunk) {
+      auto verifier = flatbuffers::Verifier{
+        reinterpret_cast<const uint8_t*>(chunk->data()), chunk->size()};
+      if (verifier.template VerifyBuffer<root_type>(file_identifier()))
```
lava, line 63: I'm not sure we always want to verify by default, at least for partitions we intentionally only access a subset of the full flatbuffer at startup.

libvast/vast/fbs/table.hpp
```diff
@@ -0,0 +1,171 @@
...
+  table& operator=(table&& rhs) noexcept = default;
+
+  /// Default destructor.
+  virtual ~table() noexcept = default;
+
+  // -- concepts ---------------------------------------------------------------
+
+  /// Return a view on the underlying byte buffer.
+  /// @pre `x`
+  friend span<const byte> as_bytes(const derived_type& x) noexcept {
```
lava, line 83: Is this some well-known pattern? I've never seen a class providing free functions for its derived class.

Would it make sense to declare this for the base type?

```
span<const byte> as_bytes(const table& x) noexcept
```

lava (commented):
For the lack of a better word, this code feels very "caf-like" to me ;) In particular, the `table` class feels a bit overloaded:

- on the one hand, it acts a bit like a `shared_pointer<Root>` by overloading `operator*` and `operator->` to pretend its a `Root`, and storing a `vast::chunk` to implement these semantics. [this part is what we originally discussed as `flatbuffer_ptr`, iirc]

- on the other hand, it grants caf inspection support and provides callbacks for serialization/deserialization

- and finally, it acts as a traits class for the `table_builder`.

As a result, the API becomes pretty hard to reason about (ie. what are the requirements on `Derived`?), and the `table` class is a bit cumbersome to use because it has so high opt-in requirements (ie. one needs at least to create a dedicated wrapper that inherits from `fbs::table` and needs a constructor that accepts a `chunk_ptr&&` and forwards this chunk to the parent class.)

libvast/vast/fbs/table.hpp
dominiklohmann: In a call we agreed to add an optional tag type parameter that disables the verification.

libvast/vast/fbs/table.hpp
dominiklohmann: In a call we agreed that this has both advantages and disadvantages; Using `derived_type` here causes an error of the derived type tries to define this function.

We also talked about the relationship between `as_bytes` and FlatBuffers, and agreed that we need it for. `fbs::pack_bytes`. This enables use cases like flattened table slices.

libvast/vast/fbs/table_builder.hpp
dominiklohmann: > I'm not really convinced that there's a need for this class: As far as I can tell, its only purpose is to call `builder_.Finish()` with the correct file identifier, but in a non-generic class one would call `builder_.FinishSegment()` and this would be even safer because flatbuffers code will automatically select the correct identifier based on the information in the flatbuffer schema.
>
> As a tradeoff, we have a virtual base class and a hard-to-describe control flow (i.e. user must call `segment_builder::add()` followed by `table_builder::finish()`, and the `create()` method directly accesses the builder from the base class)


We agreed that this is a stylistic matter mostly. Putting this off until we agreed on all other tasks.

> And furthermore, it seems like this would be hard to compose, ie. how would I be able to create a non-root level segment flatbuffer as part of some existing builder?

We agreed that the table_builder should probably be passed the FlatBuffers builder externally.

dominiklohmann:
I've added the label `blocked` as we've decided to tackle #1105 first, independent of the root type helper.

dominiklohmann:
I'm closing this PR. We can always look back at this in the future, but right now rebasing and refactoring this would cost more time than just writing this from scratch again with the review comments incorporated from the very beginning.

PR #1131, opened 2020-10-31
Remove unused Version FlatBuffers table

dominiklohmann:

This is a breaking change based on #1130. Note that we intend to break this month anyways with the table slice change (#1105), so this shouldn't be too big of a deal. Might as well use the opportunity to remove some legacy fbs code.

This PR
- removes the unused Version FlatBuffers table,
- removes the versioning from the Interval FlatBuffers struct, because that does not really make sense to version, and
- renames TableSliceBuffer to FlatTableSlice, because that name matches the table's actual purpose better.

###  :memo: Checklist

n/t


dominiklohmann:
I've added the label `blocked` as we've decided to tackle #1105 first, independent of the root type helper.

PR #1134, opened 2020-11-02
Generalize splunk-to-vast

mavam:

The `splunk-to-vast` utility converts splunk CIM JSON files into VAST taxonomy syntax. Because we don't want to create one script per tool, we now use one single tool that can be parameterized and extended.

###  :memo: Checklist

One-shot scrutinization.


scripts/taxonomize
```diff
@@ -0,0 +1,120 @@
+#! /usr/bin/env python
+
+"""
+Generates a VAST taxonomy from other tools
+
+Example usage:
+
+   taxonomize --splunk < model.json > splunk.yaml
+   taxonomize --sentinel > sentinel.yaml
```
tobim, line 9: Where is the input coming from? It looks like it requires internet access. The difference in behavior compared to `--splunk` is not intuitive.

Either change the tool to work the same way in both cases or document the difference. Your call.

tobim (changes requested):

scripts/taxonomize
mavam: I clarified this in a comment.

tobim (approved):

PR #1140, opened 2020-11-04
Port table_slice API to FlatBuffers

dominiklohmann:
###  :notebook_with_decorative_cover: Description

Transform the `table_slice` API to use FlatBuffers for persistence, enabling versioning of table slice encodings and sharing ownership between segments and contained table slices.

###  :world_map: Implementation Status

- [x] Cleanup `table_slice` API #1141
- [x] Remove CAF-encoded table slices #1142
- [x] Allow for extending the table slice FlatBuffers table #1143
- [x] Switch to the new `table_slice` API #1148
- [x] Implement new MessagePack-encoded table slices #1157 #1159
- [x] Implement new Arrow-encoded table slices #1160
- [x] Remove legacy table slices and table slice factory #1165

###  :memo: Checklist

- [x] All user-facing changes have changelog entries.
- [x] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [x] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

Because of the sheer size of this change, this PR is split into multiple PRs that will be merged into this epic branch. After all merge commits are through, this PR will contain fixups only, which are best reviewed commit-by-commit.

PR #1141, opened 2020-11-04
Cleanup table_slice API

dominiklohmann:

This is the first PR to be merged into #1140, doing some much-needed cleanup work on the existing `table_slice` API that will make future changes easier.

###  :memo: Checklist

Commit-by-commit is best.


libvast/vast/table_slice_column.hpp
```diff
@@ -14,27 +14,64 @@
...
-    : slice{std::move(slice)}, column{col} {
-    // nop
-  }
-  table_slice_ptr slice;
-  size_t column;
+  /// Construct a view on a column of a table slice.
+  /// @param slice The slice to view.
+  /// @param column The viewed column's index.
+  /// @pre `column < slice->columns()`
+  table_slice_column(table_slice_ptr slice, size_t column) noexcept;
```
tobim, line 42: I think both of those constructors should be `private`. Instead, we should provide
```
static std::optional<table_slice_column>(table_slice_ptr slice, size_t column);
static std::optional<table_slice_column>(table_slice_ptr slice, std::string_view column);
```
Optional or pointer semantics should not creep into the actual `table_slice_column`.

libvast/src/system/explorer.cpp
```diff
@@ -134,23 +135,23 @@ explorer(caf::stateful_actor<explorer_state>* self, caf::actor node,
         VAST_DEBUG(self, "could not find timestamp field in", layout);
         return;
       }
-      std::optional<table_slice::column_view> by_column;
+      std::optional<table_slice_column> by_column;
```
tobim, line 138: This exemplifies my issue with the "nullability" of `table_slice_column`. With the way it is defined right now, the `optional` is redundant.

libvast/vast/table_slice_column.hpp
```diff
@@ -14,27 +14,64 @@
...
+  /// @note This constructor fails if the column name is not included in the
+  /// table slices layout. Check for success with `slice() != nullptr`.
+  table_slice_column(table_slice_ptr slice, std::string_view column) noexcept;
+
+  /// @returns the data at given row.
+  /// @pre `row < size()`
+  data_view operator[](size_t row) const;
 
+  /// @returns the number of rows in the column.
+  size_t size() const noexcept;
```
tobim, line 56: To me `rows()` is semantically clearer.

libvast/src/system/explorer.cpp
```diff
@@ -134,23 +135,23 @@ explorer(caf::stateful_actor<explorer_state>* self, caf::actor node,
...
           return;
         }
       }
       VAST_DEBUG(self, "uses", it->name, "to construct timebox");
-      auto column = slice->column(it->name);
-      VAST_ASSERT(column);
-      for (size_t i = 0; i < column->rows(); ++i) {
-        auto data_view = (*column)[i];
+      auto column = table_slice_column{slice, it->name};
+      VAST_ASSERT(column.slice());
```
tobim, line 152: This is a strange way to test for validity. Imho value types must be valid by construction.

tobim (changes requested):

libvast/vast/table_slice_row.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <caf/meta/type_name.hpp>
+
+#include <cstdint>
+#include <string_view>
+
+namespace vast {
+
+/// Convenience helper for traversing a row of a table slice.
+/// @relates table_slice
+class table_slice_row {
```
tobim, line 28: What is the purpose of this class? I think it would be preferable to use `record_view` instead.

libvast/vast/table_slice.hpp
```diff
@@ -85,12 +84,6 @@ class table_slice : public caf::ref_counted {
   /// @returns an identifier for the implementing class.
   virtual caf::atom_value implementation_id() const noexcept = 0;
 
-  /// @returns the layout for columns in range
-  /// [first_column, first_column + num_columns).
-  record_type
-  layout(size_type first_column, size_type num_columns
-                                 = std::numeric_limits<size_type>::max()) const;
-
```
tobim: This is very good.

libvast/vast/table_slice_row.hpp
dominiklohmann: We have a story for a future expansion that makes me want to keep this as is.

libvast/vast/table_slice_column.hpp
dominiklohmann: Using `size()` and `index()` allows for using `table_slice_row` and `table_slice_column` interchangeably.

libvast/vast/table_slice_column.hpp
dominiklohmann: I'd argue this only really makes sense for the latter constructor. Will change for that one.

libvast/vast/table_slice_row.hpp
tobim: Can you be more specific? What functionality should be based on this?

dominiklohmann: Row-major and column-major iteration can be built on top of these.

tobim: this type would only be useful for row-major iteration, but that iterator should point to `record_view`s.

dominiklohmann: That would require decoding the whole row beforehand, which we don't necessarily want to do. The data access should be lazy through the encoding, regardless of whether it is MessagePack or Arrow.

mavam: Clearly, we don't have the use case fleshed out yet. Until then, I propose we put these utilities in the backlog and add them only when needed. Otherwise we carry more complexity with us that we need.

tobim: We agreed to keep the type around in case we really do need it later. It is much more discoverable inside the tree than somewhere in an old branch.

libvast/vast/table_slice_column.hpp
tobim: As long as we keep the precondition it is ok.

tobim (approved):

dominiklohmann:
I rebased after the approval to get rid of the autosquash commits. I'll merge into the epic branch once the CI is green.

tobim (approved):
Tested locally, the changes work. I did not do a performance analysis however.

dominiklohmann:
Rebased one further time after fixing the remaining stack-use-after-scope issues w.r.t `table_slice::layout()` returning by value.

PR #1143, opened 2020-11-04
Allow for extending the table slice FlatBuffers table

dominiklohmann:

The third PR to be merged into #1140. This set of changes wraps the existing table slice FlatBuffers table in a union, and cleans up the `table_slice.fbs` file accordingly, and adds a utility function for visiting table slice encodings.

###  :memo: Checklist

Commit-by-commit works best for this PR.

### DIFF
--- a/libvast/vast/fbs/table_slice.fbs
+++ b/libvast/vast/fbs/table_slice.fbs
@@ -1,4 +1,4 @@
-namespace vast.fbs;
+namespace vast.fbs.table_slice.legacy;
 
 /// The format of the binary table slice data.
 enum Encoding : byte {
@@ -6,9 +6,8 @@ enum Encoding : byte {
   MessagePack,
 }
 
-namespace vast.fbs.table_slice;
-
-/// A subset of rows of a table.
+/// An encoding for table slices using runtime dispatching for determining the
+/// underlying and unversioned encoding.
 table v0 {
   /// The offset in the 2^64 ID event space.
   offset: ulong;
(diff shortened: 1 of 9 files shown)


libvast/vast/fbs/table_slice.fbs
```diff
@@ -1,5 +1,3 @@
-file_identifier "vTBS";
```
tobim: This was deleted by accident.

libvast/vast/table_slice_visit.hpp
```diff
@@ -0,0 +1,61 @@
...
+/// @param x The FlatBuffers root type for table slices.
+/// @note The handler for invalid table slices takes no arguments. If none is
+/// specified, visig aborts when the table slice encoding is invalid.
+template <class Visitor>
+auto visit(Visitor&& visitor, const fbs::TableSlice* x) noexcept(
+  std::conjunction_v<
+    // Check whether the handler for invalid encodings is noexcept-specified,
+    // if and only if it exists.
+    std::disjunction<std::is_nothrow_invocable<Visitor>,
+                     std::negation<std::is_invocable<Visitor>>>,
```
tobim, line 36: This is kinda hard to understand. We have 3 cases here:
|             | throws | nothrow | absent |
| --------- | --------- | ---------- | --------- |
| nothrow_invocable | false | true | false |
| not invokable | false | false | true |
|    or  | false | true | true |

So it is correct. We could think about switching the order, I assume the default will usually be absent.

tobim (approved):
So far all changes are understandable and reasonable.

libvast/vast/table_slice_visit.hpp
dominiklohmann: It should be about 50-50. Currently there's lots of missing defaults, but that is only because we necessarily use it outside of `table_slice.cpp`. I still agree with changing the order, as that makes it easier to read in my opinion.

Ideally, we can move the visit function inside `table_slice.cpp` in the future and stop exposing the underlying FlatBuffers table at all.

dominiklohmann: I have reversed the order as suggested and rebased to get rid of all fixup! commits.

PR #1149, opened 2020-11-05
Show file size information in lsvast

lava:
###  :memo: Checklist

### DIFF
--- a/tools/lsvast/lsvast.cpp
+++ b/tools/lsvast/lsvast.cpp
@@ -13,6 +13,7 @@
 
 #include "vast/chunk.hpp"
 #include "vast/concept/printable/to_string.hpp"
+#include "vast/concept/printable/vast/type.hpp"
 #include "vast/concept/printable/vast/uuid.hpp"
 #include "vast/directory.hpp"
 #include "vast/fbs/index.hpp"
@@ -22,6 +23,8 @@
 #include "vast/ids.hpp"
 #include "vast/io/read.hpp"
 #include "vast/path.hpp"
+#include "vast/qualified_record_field.hpp"
+#include "vast/type.hpp"
 #include "vast/uuid.hpp"
 
 #include <caf/binary_deserializer.hpp>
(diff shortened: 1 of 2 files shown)


mavam:
There is a lot of code for just formatting. How about using `vast::data` to store the contents and then simply render it as JSON?

tools/lsvast/lsvast.cpp
```diff
@@ -122,162 +225,297 @@ std::ostream& operator<<(std::ostream& out, const vast::fbs::uuid::v0* uuid) {
...
 
 int main(int argc, char** argv) {
-  if (argc != 2) {
-    std::cerr << "Usage: ./lsvast <path/to/vast.db>\n";
+  std::string raw_path;
+  struct formatting_options format;
+  format.print_bytesizes = true;
+  format.verbosity = output_verbosity::verbose;
+  for (int i = 1; i < argc; ++i) {
+    auto arg = argv[i];
```
mavam, line 490:
```suggestion
    auto arg = std::string_view{argv[i]};
```

This may make the subsequent comparisons less C-ish.

tools/lsvast/lsvast.cpp
```diff
@@ -122,162 +225,297 @@ std::ostream& operator<<(std::ostream& out, const vast::fbs::uuid::v0* uuid) {
...
+    if (!strcmp(arg, "-h") || !strcmp(arg, "--human-readable")) {
+      format.print_bytesizes = true;
+      format.human_readable_numbers = true;
+    } else if (!strcmp(arg, "-s") || !strcmp(arg, "--print-bytesizes"))
```
mavam, line 494: Regarding bracing style, we usually do either *all* branches with braces or *none*, but not a mixture.

mavam (commented):
I'm still getting hex output for the number of rows. Am I doing something wrong?

lava:
> I'm still getting hex output for the number of rows. Am I doing something wrong?

No, I didn't realize `std::hex` was sticky after use. Should be fixed now.

tools/lsvast/lsvast.cpp
```diff
@@ -114,170 +179,346 @@ read_flatbuffer_file(vast::path path) {
...
-  std::cout << spaces << "(unknown " << path.str() << ")\n";
+void print_unknown(vast::path path, indentation& indent,
+                   const formatting_options&) {
+  std::cout << indent << "(unknown " << path.str() << ")\n";
 }
 
-void print_vast_db(vast::path vast_db, int indent) {
+std::string print_bytesize(size_t bytes, const formatting_options& formatting) {
+  const char* suffixes[] = {
+    " byte", " KiB", " MiB", " GiB", " TiB", " EiB",
```
mavam, line 237:
```suggestion
    " B", " KiB", " MiB", " GiB", " TiB", " EiB",
```

I just noticed in the output that this is the odd one out. I imagine you wrote it out because it could be mistaken for "billion"? If so, I'd say that's not an issue because all values have a `B` suffix.

mavam (changes requested):
I tested this with my local DB, both using just the dir as argument and `-h`. Looks good. Just one small change request regarding the representation of plain bytes.

tools/lsvast/lsvast.cpp
lava: Actually more due to the confusion potential with "bit", but I don't mind changing it to `B`.

mavam (approved):

PR #1151, opened 2020-11-09
Fix end-quote in static workflow

tobim:

Workflow quoting was wrong.


dominiklohmann (approved):
👀

PR #1158, opened 2020-11-10
Add Winlogbeat schema definitions

mavam:

This PR adds schema definitions for Winlogbeat wrapper events that wrap the raw Sysmon records.

###  :memo: Checklist

In one shot.

### DIFF
--- a/schema/sysmon.schema
+++ b/schema/sysmon.schema
@@ -558,3 +558,153 @@ type sysmon.RawAccessRead = record {
     Device: string,
 }
 
+// The following beats type definitions wrap the above raw sysmon events.
+// They are effectively a carrier of the Sysmon data that comes with some
+// additional meta data.
+//
+// The logstash filter configuration for the below defined Winlogbeat events
+// looks like this:
+//
+//     filter {
+//       if [winlog][provider_name] == "Microsoft-Windows-Sysmon" {
+//         mutate {
+//           add_tag => [ "meta_data" ]
+//           copy => { "[winlog][event_data]" => "event_data" }
+//           copy => { "[cloud][instance][id]" => "[meta_data][instance_id]" }
+//           copy => { "[agent][hostname]" => "[meta_data][hostname]" }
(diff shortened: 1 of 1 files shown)


schema/sysmon.schema
```diff
@@ -558,3 +558,153 @@ type sysmon.RawAccessRead = record {
     Device: string,
 }
 
+// The following beats type definitions wrap the above raw sysmon events.
+// They are effectively a carrier of the Sysmon data that comes with some
+// additional meta data.
+//
```
0ortmann, line 564:
```suggestion
// This configuration assumes that Sysmon events are wrapped in a JSON struct
// that looks as follows:
// {
//   "event_data": { ... },
//   "meta_data": { ... },
// }
// The `event_data` field wraps the Sysmon data. The `meta_data` field should
// contain the fields `instance_id`, `hostname`, and `event_id` to comply with
// this schema.
```

schema/sysmon.schema
```diff
@@ -558,3 +558,153 @@ type sysmon.RawAccessRead = record {
...
+// The logstash filter configuration for the below defined Winlogbeat events
+// looks like this:
```
0ortmann, line 566:
```suggestion
// An example logstash filter configuration that generates `meta_data` and
// `event_data` fields in the JSON output could look like this:
```

The logstash config provided here is not the only valid one that can be used with this schema.

0ortmann (changes requested):
We should make it more clear how the data has to be shaped to use this schema.

0ortmann (approved):
I tested this file with `vast`
```
{
  "VAST": "2020.10.29-74-g46171424",
  "CAF": "0.17.6",
  "Apache Arrow": "1.0.0",
  "PCAP": "libpcap version 1.9.1 (with TPACKET_V3)",
  "jemalloc": "5.2.1-0-gea6b3e973b477b8061e0076bb257dbd7f3faa756"
}
```

The schema is loaded correctly for Sysmon and Winlogbeat, as shown in `vast status --debug`.

PR #1162, opened 2020-11-12
Update VAST Static workflow dependencies

tobim:

Update dependencies.


dominiklohmann (approved):
The workflow runs through and the deprecation warnings appear to be fixed.

PR #1167, opened 2020-11-15
Specify noexcept for the chunk API

dominiklohmann:

This PR adds `noexcept` specifications to the chunk API. Most notably, it requires deletion steps to be nothrow-invocable, because they are invoked in the destructor, where throwing is definitely not what we want to happen.

###  :memo: Checklist

n/t


tobim (approved):
This makes sense. I did not test locally because I would not do anything that is not covered by the integration tests already.

PR #1169, opened 2020-11-15
Avoid temporary allocations in partition selector

tobim:

This removes an unnecessary temporary allocation in the `partition_selector`. The change doesn't bring any performance gains, because the allocator was able to recycle the memory in subsequent calls. But it removes some noise from memory analysis tools like heaptrack.

### DIFF
--- a/libvast/src/system/partition.cpp
+++ b/libvast/src/system/partition.cpp
@@ -187,11 +187,16 @@ evaluate(const PartitionState& state, const expression& expr) {
 
 bool partition_selector::operator()(const vast::qualified_record_field& filter,
                                     const table_slice_column& column) const {
-  VAST_TRACE(VAST_ARG(filter), VAST_ARG(column));
   auto&& layout = column.slice().layout();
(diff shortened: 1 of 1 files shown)


libvast/src/system/partition.cpp
```diff
@@ -187,8 +187,12 @@ evaluate(const PartitionState& state, const expression& expr) {
 bool partition_selector::operator()(const vast::qualified_record_field& filter,
                                     const table_slice_column& x) const {
   auto& layout = x.slice->layout();
-  vast::qualified_record_field fqf{layout.name(), layout.fields.at(x.column)};
-  return filter == fqf;
+  if (filter.layout_name != layout.name())
```
dominiklohmann, line 190: Can you keep the old code commented out here, saying that the below is essentially identical, and add another comment explaining what allocation you're avoiding here?

lava (approved):
Didn't test this, but it doesnt look like it really needs testing.

PR #1170, opened 2020-11-16
Set fallback port for underspecified endpoints

dominiklohmann:

Sets a fallback port of `42000` when the `vast.endpoint` option is set, but does not contain a port.

###  :memo: Checklist

Run `vast -e localhost start` with and without this change to verify this works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- ⚠️ VAST now listens on port 42000 instead of letting the operating system
+  choose the port if the option `vast.endpoint` specifies an endpoint without a
+  port. To restore the old behavior, set the port to 0 explicitly.
+  [#1170](https://github.com/tenzir/vast/pull/1170)
+
 - ⚡️ Archive segments no longer include an additional, unnecessary version
   identifier. We took the opportunity to clean this up bundled with the other
   recent breaking changes. [#1168](https://github.com/tenzir/vast/pull/1168)
--- a/libvast/src/system/start_command.cpp
+++ b/libvast/src/system/start_command.cpp
@@ -61,6 +61,9 @@ caf::message start_command_impl(start_command_extra_steps extra_steps,
   if (!parsers::endpoint(str, node_endpoint))
     return caf::make_message(
       make_error(ec::parse_error, "invalid endpoint", str));
+  // Default to port 42000/tcp if none is set.
+  if (!node_endpoint.port)
+    node_endpoint.port = port{defaults::system::endpoint_port, port::tcp};
   // Get a convenient and blocking way to interact with actors.
   caf::scoped_actor self{sys};
   // Spawn our node.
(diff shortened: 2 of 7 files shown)


libvast/src/system/start_command.cpp
```diff
@@ -61,6 +61,9 @@ caf::message start_command_impl(start_command_extra_steps extra_steps,
   if (!parsers::endpoint(str, node_endpoint))
     return caf::make_message(
       make_error(ec::parse_error, "invalid endpoint", str));
+  // Default to port 42000/tcp if none is set.
```
mavam, line 64: I would add a note that if you pass `0` to CAF, the expected behavior is that it will select a random high-level port. In fact, I would probably put a log message on VERBOSE to let the user know that we're going to select a random port.

mavam (commented):
Makes sense overall.

Is it still possible to explicitly specify `0` as port to get a random port that is available on the system?

libvast/src/system/start_command.cpp
dominiklohmann: Setting 0 explicitly makes VAST use 42000 with this change.

mavam: Shouldn't the logic be as follows?
- if nothing is specified, default to 42000
- if 0 is specified, use 0

dominiklohmann: That doesn't really make sense, port 0 is not a valid port.

tobim: Specifying `0` means the operating system chooses a free port for you, I think we should not prevent that.

dominiklohmann: TIL actually, did not know that. Will take care of it accordingly.

dominiklohmann: I've implemented this accordingly.

CHANGELOG.md
```diff
@@ -12,6 +12,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 VAST now listens on port 42000 instead of letting the operating system
```
mavam, line 15: I would make this a "change".

libvast/vast/concept/parseable/vast/endpoint.hpp
```diff
@@ -34,7 +34,9 @@ struct endpoint_parser : parser<endpoint_parser> {
     auto hostname = +(alnum | chr{'-'} | chr{'_'} | chr{'.'});
     auto host = hostname->*[&](std::string x) { e.host = std::move(x); };
     auto port = (parsers::port->*[&](vast::port x) { e.port = x; })
-                | (u16->*[&](uint16_t x) { e.port = vast::port{x}; });
+                | (u16->*[&](uint16_t x) {
+                    e.port = vast::port{x, vast::port::tcp};
+                  });
```
mavam, line 39: Please also update https://github.com/tenzir/vast/blob/master/libvast/test/endpoint.cpp.

libvast/src/system/start_command.cpp
```diff
@@ -61,6 +61,9 @@ caf::message start_command_impl(start_command_extra_steps extra_steps,
...
+  if (node_endpoint.port.type() == port::unknown)
+    node_endpoint.port = port{defaults::system::endpoint_port, port::tcp};
```
mavam, line 66: I'm not following the logic, maybe a comment is missing, but where is the jump from "unknown port type" to "let's use the defaults?

mavam (changes requested):

libvast/src/system/start_command.cpp
dominiklohmann: I've reimplemented this independent of the internals of the endpoint parser by making the port in an endpoint optional.

mavam: Yes, `optional` semantics make much more sense here.

mavam (commented):
I tested this locally and it works. For example, `vast -e 0.0.0.0 start` no longer opens a random high-level port.

mavam (approved):

PR #1171, opened 2020-11-16
Extend the VAST Static workflow with a "CI" type build

tobim:

This builds a static binary with that is more debugable in case something goes wrong.

###  :memo: Checklist

Commit by commit.

### DIFF
--- a/libvast/src/detail/fdinbuf.cpp
+++ b/libvast/src/detail/fdinbuf.cpp
@@ -17,9 +17,8 @@
 
 #include <cstdio>
 #include <cstring>
+#include <poll.h>
 #include <unistd.h>
-
-#include <sys/poll.h>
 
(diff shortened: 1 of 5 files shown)


libvast/src/detail/fdinbuf.cpp
```diff
@@ -17,10 +17,9 @@
 
 #include <cstdio>
 #include <cstring>
+#include <poll.h>
 #include <unistd.h>
 
-#include <sys/poll.h>
```
dominiklohmann, line 22: Was this supposed to be in here?

tobim: Yes, `::poll()` is used in this file.

dominiklohmann (approved):

PR #1173, opened 2020-11-16
Make rebuiling table slices not a constructor

dominiklohmann:
### 📔 Description

n/t

###  :memo: Checklist

n/t

### DIFF
--- a/libvast/test/format/arrow.cpp
+++ b/libvast/test/format/arrow.cpp
@@ -81,9 +81,7 @@ TEST(arrow batch) {
   std::shared_ptr<arrow::RecordBatch> batch;
   while (reader->ReadNext(&batch).ok() && batch != nullptr) {
     REQUIRE_LESS(slice_id, zeek_conn_log.size());
-    auto slice
-      = table_slice{zeek_conn_log[slice_id], table_slice::encoding::arrow,
-                    table_slice::verify::yes};
+    auto slice = rebuild(zeek_conn_log[slice_id], table_slice::encoding::arrow);
     CHECK_EQUAL(detail::narrow<size_t>(batch->num_rows()), slice.rows());
     CHECK(batch->schema()->Equals(*arrow_schema));
     CHECK_EQUAL(slice, zeek_conn_log[slice_id]);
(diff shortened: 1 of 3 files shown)


libvast/vast/table_slice.hpp
```diff
@@ -203,6 +185,16 @@ class table_slice final {
...
 
+  // -- operations -------------------------------------------------------------
+
+  /// Rebuilds a table slice with a given encoding if necessary.
+  /// @param slice The slice to rebuild.
+  /// @param encoding The encoding to convert to.
+  /// @note This function only rebuilds if necessary, i.e., the new encoding
+  /// is different from the existing one.
+  friend table_slice
+  rebuild_table_slice(table_slice slice, enum encoding encoding) noexcept;
```
mavam, line 196:
```suggestion
  rebuild(table_slice slice, enum encoding encoding) noexcept;
```

I think the arguments and documentation make it clear what's going on here.

libvast/test/format/arrow.cpp
```diff
@@ -81,9 +81,8 @@ TEST(arrow batch) {
   std::shared_ptr<arrow::RecordBatch> batch;
   while (reader->ReadNext(&batch).ok() && batch != nullptr) {
     REQUIRE_LESS(slice_id, zeek_conn_log.size());
-    auto slice
-      = table_slice{zeek_conn_log[slice_id], table_slice::encoding::arrow,
-                    table_slice::verify::yes};
```
mavam, line 86: How come that the new function doesn't have verification? Not that I want it;  just curious.

mavam (approved):
I didn't test this because it looks like code shuffling to me.

libvast/test/format/arrow.cpp
dominiklohmann: There's no need to verify again if we move or copy, and if we rebuild we can already rely on the table slice having the correct offsets, otherwise the FlatBufferBuilder would fail in finish.

libvast/vast/table_slice.hpp
dominiklohmann: Fair enough, will change before merging tomorrow.

PR #1174, opened 2020-11-16
Move the offset into the table slice state

dominiklohmann:

This reduces `sizeof(table_slice)` from 24 to 16.

Weirdly enough, the tests fail if I run the index test suite using `vast-test -v 4 -s index`, but succeed if I run the failing test individually using `vast-test -v 4 -s index -t 'iterable integer query result'`.

This is not a breaking change.

###  :memo: Checklist

Benchmark! I've requested @tobim as a reviewer for this.

### DIFF
--- a/libvast/src/arrow_table_slice.cpp
+++ b/libvast/src/arrow_table_slice.cpp
@@ -636,6 +636,16 @@ table_slice::size_type arrow_table_slice<FlatBuffer>::columns() const noexcept {
   return record_batch()->num_columns();
 }
 
+template <class FlatBuffer>
+id arrow_table_slice<FlatBuffer>::offset() const noexcept {
+  return state_.offset;
+}
+
+template <class FlatBuffer>
+void arrow_table_slice<FlatBuffer>::offset(id offset) noexcept {
+  state_.offset = offset;
+}
+
 // -- data access ------------------------------------------------------------
 
 template <class FlatBuffer>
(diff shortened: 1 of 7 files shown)


mavam:
> Weirdly enough, the tests fail if I run the index test suite using `vast-test -v 4 -s index`, but succeed if I run the failing test individually

This is pathologic for having forgotten to initialize a factory. Since factory initialization is idempotent and process-wide, you may benefit from the side effects of other tests when running the whole test suite.

dominiklohmann:
> This is pathologic for having forgotten to initialize a factory. Since factory initialization is idempotent and process-wide, you may benefit from the side effects of other tests when running the whole test suite.

It's the other way around, though: It works if I run a single test, but fails if I run all index tests.

mavam:
Huh, that is weird.

Maybe @tobim has an idea about what other side effects might be at play?!

tobim:
The problem is in the test directly above. A local function `rebase` modifies the offsets of the table slices in question. Before this PR, the offset is located inside the `table_slice`, so not part of the shared data. In this PR, the offset is part of the shared data, so even though the `table_slice`s are copied, `rebase` is modifying the content of the `events` fixture.

I guess `rebase` should create clones of the slices instead.

tobim:
The fact that ubuntu-clang didn't fail here has me worried. The only explanation I have for that is undefined behavior.

dominiklohmann:
I've pushed a fix that does a deep copy in `rebase` and rebased the branch onto master.

This could really use some cleanup, but I don't want to sink time into this until after we've benchmarked—after all, we'e still unsure whether we want this at all.

dominiklohmann:
I've run some benchmarks on the side today and they confirm the initial numbers reported by @tobim: This is just slightly slower than before, but not by much. The additional API constraints are undesirable. I am closing this PR, we won't be merging this.

PR #1176, opened 2020-11-17
Add missing Suricata schema types and fields

satta:

This PR introduces the currently missing Suricata schema types `nfs`, `tftp`, `snmp` and `ikev2`. It also adds correct and consistent types for the common fields `vlan` and `in_iface`.

###  :memo: Checklist

- [X] All user-facing changes have changelog entries. (n/a)
- [X] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary. (no dedicated Suricata schema documentation there yet)
- [X] The PR description contains instructions for the reviewer, if necessary. (see below)

Please double-check whether the changes correspond to current best practices and whether the schema loads for the relevant VAST versions.


mavam:
Thanks for contributing these definitions!

Since you changed the layout, the integration test baseline needs to be updated. (This is why CI fails.) If you invoke `make integration` in the build directory, you'll see the diff to the baseline. If this looks reasonable, you can invoke `./integration_BUILD_TYPE.sh -u` where `-u` stands for "update baseline."

Thereafter, CI will be green.

satta:
Thanks for the explanation, will do.

It's OK to introduce optional fields where the output might be 'null', right? I did this because, for example, `vlan` and `pcap_id` was defined only in some types, while in fact they are common to all types, just only present when the traffic _has_ VLANs and/or was replayed from pcaps.

mavam:
> It's OK to introduce optional fields where the output might be 'null', right?

Yes, that works. In fact, that's currently how we can have non-breaking updates. The new type can technically be seen as a strict superset of the old type.

mavam (approved):
Nice work! 🙏

PR #1177, opened 2020-11-17
Fix build on gcc-8

dominiklohmann:

gcc-8 fails to detect when a switch exhausts all uses of an enum class. This is usually worked around by calling a `[[noreturn]]` function afterwards. Apparently gcc-8 also fails to recognize that calling a `[[noreturn]]` function in a `constexpr` context is also totally fine, and fails to compile.

###  :memo: Checklist

n/t


mavam (approved):
Not tested myself, but if it passes the custom docker build and test suite, this is good to go.

PR #1181, opened 2020-11-18
Ensure that table slice builders cannot fail

dominiklohmann:

The check in the beginning should've been an assertion from the get-go: it indicates wrong API usage, and is never a result of wrong input.

###  :memo: Checklist

n/t

### DIFF
--- a/libvast/src/arrow_table_slice_builder.cpp
+++ b/libvast/src/arrow_table_slice_builder.cpp
@@ -473,9 +473,9 @@ arrow_table_slice_builder::~arrow_table_slice_builder() noexcept {
 
 table_slice arrow_table_slice_builder::finish(
   [[maybe_unused]] span<const byte> serialized_layout) {
-  // Sanity check.
-  if (column_ != 0)
-    return {};
+  // Sanity check: If this triggers, the calls to add() did not match the number
+  // of fields in the layout.
+  VAST_ASSERT(column_ == 0);
   // Pack layout.
   auto layout_buffer
     = serialized_layout.empty()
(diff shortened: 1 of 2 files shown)


tobim (commented):
What happens in builds without assertions?

dominiklohmann:
> What happens in builds without assertions?

The check is not performed. But since this is a developer rather than a user error anyways, it doesn't really matter.

tobim:
> The check is not performed.

What is the consequence of that?

>  But since this is a developer rather than a user error anyways, it doesn't really matter.

Just because a build is used in production doesn't make it magically bug-free, so it does matter.

dominiklohmann:
> Just because a build is used in production doesn't make it magically bug-free, so it does matter.

You could say the same for every other assertion we have in our code. This is a logic error that is input-independent.

lava:
Technically it *is* removing a safeguard from the assertion-less build. To be conservative, maybe we can leave the `if (column_ != 0)` part in addition to the assertion, and remove it after we have run it a while without running into it so we're reasonably sure there are no lingering bugs where `column_` is left partially filled.

dominiklohmann:
I'd rather use this instead then:

```cpp
if (column_ != 0)
  die("number of calls to 'add()' since last call to 'finish()' is not divisible by number fields in 'layout()' without rest");
```

Thoughts @tobim @lava?

tobim:
I prefer to keep the assertion. My question was not meant in a suggestive way, I'm just curious if we actually run into corrupt state in builds without the assertion.

dominiklohmann:
> I'm just curious if we actually run into corrupt state in builds without the assertion.

We really shouldn't. This is more of a developer than a user error. The calls The integration test suite tests all readers, so this should never occur in release builds only.

Ideally the `table_slice_builder::add()` API would require adding a full row at a time, but we're not quite there yet.

tobim (approved):

PR #1182, opened 2020-11-18
Construct chunks from const pointers

dominiklohmann:

This avoids the need to `const_cast` when using `chunk::make` on existing data. The chunk API already promises not to touch the underlying pointer unless it is safe to do so, so let's just make that clear from the interface of the make function already.

###  :memo: Checklist

n/t

### DIFF
--- a/libvast/src/chunk.cpp
+++ b/libvast/src/chunk.cpp
@@ -100,7 +100,7 @@ chunk_ptr chunk::slice(size_type start, size_type length) const {
 }
 
 chunk::chunk(const void* ptr, size_type size, deleter_type deleter) noexcept
-  : data_{reinterpret_cast<pointer>(const_cast<void*>(ptr))},
+  : data_{static_cast<pointer>(const_cast<void*>(ptr))},
     size_{size},
     deleter_{deleter} {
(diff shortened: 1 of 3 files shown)


libvast/src/chunk.cpp
```diff
@@ -99,8 +99,10 @@ chunk_ptr chunk::slice(size_type start, size_type length) const {
   return make(length, data_ + start, deleter);
 }
 
-chunk::chunk(void* ptr, size_type size, deleter_type deleter) noexcept
-  : data_{reinterpret_cast<value_type*>(ptr)}, size_{size}, deleter_{deleter} {
+chunk::chunk(const void* ptr, size_type size, deleter_type deleter) noexcept
+  : data_{reinterpret_cast<pointer>(const_cast<void*>(ptr))},
```
lava, line 103: We can `static_cast` instead of `reinterpret_cast` here

lava (approved):
More const-correctness is always good

libvast/src/chunk.cpp
dominiklohmann: Nice catch!

PR #1183, opened 2020-11-18
Fix result lifetime of as_record_batch

dominiklohmann:

This fixes a bug in the new `as_record_batch` function.

###  :memo: Checklist

Commit-by-commit feels best. Read the commit descriptions.

### DIFF
--- a/libvast/src/table_slice.cpp
+++ b/libvast/src/table_slice.cpp
@@ -16,7 +16,7 @@
 #include "vast/chunk.hpp"
 #include "vast/defaults.hpp"
 #include "vast/detail/assert.hpp"
-#include "vast/detail/keeper.hpp"
+#include "vast/detail/keep.hpp"
 #include "vast/detail/overload.hpp"
 #include "vast/error.hpp"
 #include "vast/expression.hpp"
@@ -367,7 +367,7 @@ std::shared_ptr<arrow::RecordBatch> as_record_batch(const table_slice& slice) {
         // guarantee for the underlying Arrow Buffer object.
         auto batch = state(encoded, slice.state_)->record_batch();
         auto result = std::shared_ptr<arrow::RecordBatch>{
-          batch.get(), detail::keeper{batch, slice}};
+          batch.get(), detail::keep(batch, slice)};
         return result;
       } else {
         // Rebuild the slice as an Arrow-encoded table slice.
--- a/libvast/src/table_slice.cpp
+++ b/libvast/src/table_slice.cpp
@@ -16,6 +16,7 @@
 #include "vast/chunk.hpp"
 #include "vast/defaults.hpp"
 #include "vast/detail/assert.hpp"
+#include "vast/detail/keeper.hpp"
 #include "vast/detail/overload.hpp"
 #include "vast/error.hpp"
 #include "vast/expression.hpp"
@@ -363,18 +364,14 @@ std::shared_ptr<arrow::RecordBatch> as_record_batch(const table_slice& slice) {
         // lifetime with the chunk.
         auto batch = state(encoded, slice.state_)->record_batch();
         auto result = std::shared_ptr<arrow::RecordBatch>{
-          batch.get(), [slice, batch](arrow::RecordBatch*) noexcept {
-            VAST_IGNORE_UNUSED(slice);
-            VAST_IGNORE_UNUSED(batch);
-          }};
+          batch.get(), detail::keeper{batch, slice}};
         return result;
       } else {
         // Rebuild the slice as an Arrow-encoded table slice.
         auto copy = rebuild(slice, table_slice::encoding::arrow);
         // Bind the lifetime of the copy (and thus the returned Record Batch) to
         // the lifetime of the original slice.
-        slice.chunk_->add_deletion_step(
-          [copy]() noexcept { VAST_IGNORE_UNUSED(copy); });
+        slice.chunk_->add_deletion_step(detail::keeper{copy});
         return as_record_batch(copy);
       }
     },
--- a/libvast/src/table_slice.cpp
+++ b/libvast/src/table_slice.cpp
@@ -372,9 +372,6 @@ std::shared_ptr<arrow::RecordBatch> as_record_batch(const table_slice& slice) {
       } else {
         // Rebuild the slice as an Arrow-encoded table slice.
         auto copy = rebuild(slice, table_slice::encoding::arrow);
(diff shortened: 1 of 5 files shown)


libvast/src/table_slice.cpp
```diff
@@ -359,15 +359,22 @@ std::shared_ptr<arrow::RecordBatch> as_record_batch(const table_slice& slice) {
...
       } else {
         // Rebuild the slice as an Arrow-encoded table slice.
         auto copy = rebuild(slice, table_slice::encoding::arrow);
         // Bind the lifetime of the copy (and thus the returned Record Batch) to
         // the lifetime of the original slice.
-        slice.chunk_->ref();
-        copy.chunk_->add_deletion_step(
-          [=]() noexcept { slice.chunk_->deref(); });
+        slice.chunk_->add_deletion_step(
+          [copy]() noexcept { VAST_IGNORE_UNUSED(copy); });
```
tobim, line 377: Now that we're actually binding the batch to the copy, we don't need to couple these any more, no?

libvast/src/table_slice.cpp
```diff
@@ -359,15 +359,22 @@ std::shared_ptr<arrow::RecordBatch> as_record_batch(const table_slice& slice) {
...
       if constexpr (std::decay_t<decltype(*state(encoded, slice.state_))>::encoding
                     == table_slice::encoding::arrow) {
-        return state(encoded, slice.state_)->record_batch();
+        // Get the record batch first, then create a copy that shares the
+        // lifetime with the chunk.
+        auto batch = state(encoded, slice.state_)->record_batch();
+        auto result = std::shared_ptr<arrow::RecordBatch>{
+          batch.get(), [slice, batch](arrow::RecordBatch*) noexcept {
+            VAST_IGNORE_UNUSED(slice);
+            VAST_IGNORE_UNUSED(batch);
```
tobim, line 368: Capturing `batch` prolongs the lifetime of the `arrow::Buffer` object, correct?

libvast/src/table_slice.cpp
dominiklohmann: We do, because the record batch lives into the copy.

Calling `as_record_batch` does the following:
1. If called with an an Arrow-encoded table slice, jump to step 5.
2. Convert slice to an Arrow-encoded table slice by making a copy.
3. Bind the lifetime of the copy to the lifetime of the original slice.
4. Call `as_record_batch` with the copy.
5. Get the record batch.
6. Bind the lifetime of the record batch to the slice (and thus to the copy if one exists).
7. Return the record batch.

libvast/src/table_slice.cpp
dominiklohmann: Yes, that is correct.

libvast/src/table_slice.cpp
tobim: Yes, my hypothesis is that (3) is unnecessary.

libvast/src/table_slice.cpp
tobim: I think that should be documented in a comment.

libvast/src/table_slice.cpp
dominiklohmann: If you remove (3) then the record batch may be invalid before the table slice is. It's needed so there's a lifetime guarantee both ways. The question is whether we need that.

libvast/src/table_slice.cpp
dominiklohmann: Will do.

libvast/src/table_slice.cpp
tobim: The original slice has no way to access the copy, and neither the other way round. It has nothing to do with the batch.

dominiklohmann: Sounds reasonable. Will remove.

libvast/vast/chunk.hpp
```diff
@@ -77,21 +78,14 @@ class chunk : public caf::ref_counted {
...
-    static_assert(sizeof(Byte) == 1);
-    auto size = span.size();
-    VAST_ASSERT(size > 0);
-    auto chunk = make(size);
-    if (chunk)
-      std::memcpy(chunk->data_, span.data(), size);
-    return chunk;
+  template <typename Byte, size_t Extent>
+  static chunk_ptr copy(span<Byte, Extent> span) {
+    return make(std::vector(span.begin(), span.end()));
```
tobim, line 88: 2 Flies with one stone?

libvast/src/table_slice.cpp
```diff
@@ -363,18 +364,14 @@ std::shared_ptr<arrow::RecordBatch> as_record_batch(const table_slice& slice) {
...
+          batch.get(), detail::keeper{batch, slice}};
         return result;
       } else {
         // Rebuild the slice as an Arrow-encoded table slice.
         auto copy = rebuild(slice, table_slice::encoding::arrow);
         // Bind the lifetime of the copy (and thus the returned Record Batch) to
         // the lifetime of the original slice.
-        slice.chunk_->add_deletion_step(
-          [copy]() noexcept { VAST_IGNORE_UNUSED(copy); });
+        slice.chunk_->add_deletion_step(detail::keeper{copy});
```
tobim: The comment gives a great suggestion for a better name: `bind` or `bind_lifetime`. That would be a free function that returns a `binder`.

tobim (changes requested):
`keeper` needs a better name.

libvast/src/table_slice.cpp
tobim: Now that I'm reading it again `detail::keep(copy)` isn't so bad either.

dominiklohmann: I actually like `keeper` a lot. Not sure if a function like this really adds any value:

```cpp
template <class... Args>
constexpr auto keep(Args&&... args) noexcept {
  return keeper{std::forward<Args>(args)...};
}
```

libvast/vast/chunk.hpp
dominiklohmann: Yes, unless that's a problem. I want to do a second cleanup pass on the chunk API sometime soon anyways, so I could move this change in that upcoming PR as well.

tobim (approved):
Did not run, but the changes look logical.

libvast/src/table_slice.cpp
dominiklohmann: Will implement my above comment as discussed in Slack.

PR #1187, opened 2020-11-19
Remove port type

mavam:

This PR removes the `port` type from the data model. Ports become `count` types. To keep the domain-specific semantics when working with network data, a *type alias* with the name `port` will be the new way to express the same data.

###  :memo: Checklist

File by file.


dominiklohmann:
@mavam this needs the integration test baseline to be updated.

libvast/src/format/zeek.cpp
```diff
@@ -232,35 +232,6 @@ const char* reader::name() const {
...
-    else if (type_.name() == "zeek.dhcp" || type_.name() == "zeek.dns"
-             || type_.name() == "zeek.smnp")
-      protocol = port::udp;
-  }
-  // Assign the deduced proto to all port fields.
-  for (auto& pf : port_fields_) {
-    auto& p = caf::get<port>(xs[pf]);
-    p.type(protocol);
-  }
-}
```
dominiklohmann, line 262: I wonder if removing this function has a big effect on the performance of the Zeek reader. The benchmarks should be much quicker now. Can't believe we've done it this way.

libvast/src/format/pcap.cpp
```diff
@@ -410,6 +416,7 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
 reader::flow_state& reader::state(const flow& x) {
   auto i = flows_.find(x);
   if (i == flows_.end()) {
+    // Why do we construct a separate flow here as opposed to using x? --MV
     auto cf = flow{x.src_addr, x.dst_addr, x.src_port, x.dst_port};
```
dominiklohmann, line 420: As far as I can tell there's no difference really. We shouldn't copy. Can you make this change in a separate commit?

libvast/test/type.cpp
```diff
@@ -671,7 +668,7 @@ TEST(hashable) {
   CHECK_NOT_EQUAL(hash(type{bool_type{}}), hash(type{address_type{}}));
   auto x = record_type{
     {"x", integer_type{}}, {"y", string_type{}}, {"z", list_type{real_type{}}}};
-  CHECK_EQUAL(hash(x), 7434606127152318469ul);
```
dominiklohmann, line 674: Was there a way not to break here by adding a stub type into the type variant? The hash only changes because the indices after port moved one up.

dominiklohmann: To clarify, this whole thing is mostly a breaking change because the hashes changed.

CHANGELOG.md
```diff
@@ -12,6 +12,17 @@ Every entry has a category for which we use the following visual abbreviations:
...
+  transport-layer ports rely on `count` instead. In the schema, VAST ships with
+  a new alias `type port = count` to keep existing schema definitions in tact.
+  However, this is a breaking change because the on-disk format and Arrow data
+  representation changed. Queries with `:port` type extractors no longer work,
+  but will work in the future after re-enabling `:T` for any type identifier.
+  Similarly, the syntax `53/udp` no longer exists; use `count` syntax `53`
+  instead. Since most `port` occurrences do not carry a known transport-layer
+  type, and the type information exists typically in a separate field, removing
+  `port` as native type streamlines the data model.
+
```
dominiklohmann, line 25:
```suggestion
- ⚡️ The `port` type is no longer a first-class type. The new way to represent
  transport-layer ports relies on `count` instead. In the schema, VAST ships with
  a new alias `type port = count` to keep existing schema definitions in tact.
  However, this is a breaking change because the on-disk format and Arrow data
  representation changed. Queries with `:port` type extractors no longer work.
  Similarly, the syntax `53/udp` no longer exists; use `count` syntax `53`
  instead. Since most `port` occurrences do not carry a known transport-layer
  type, and the type information exists typically in a separate field, removing
  `port` as native type streamlines the data model.

```

Let's not make promises in a changelog entry. It's about what changed, not what will change sometime in the future.

dominiklohmann (commented):
Just a first round. The code itself looks fine, but I'll do some `git grep` locally to check whether you missed anything. Also still need to test locally.

libvast/test/type.cpp
mavam: Unfortunately there's no way around this change. You nailed the issue.

libvast/src/format/zeek.cpp
mavam: Yeah, although I have no good intuition how these checks will effect the end-to-end performance. Only measurements will tell...

dominiklohmann (commented):
I ran further tests locally and this is looking alright to me. Tested various queries and they all work as expected.

Please update the docs as noted in the checklist. This also must be rebased onto latest master before merging, and there's a missing newline in the changelog entry.

I'm yet unsure whether we want to bump the db version. Currently, we run into a global buffer overflow according to asan when loading an old db with this version.

mavam:
> Please update the docs as noted in the checklist.

Once approved, I'll pivot to updating docs.

> This also must be rebased onto latest master before merging, and there's a missing newline in the changelog entry.

✅

mavam:
@tenzir/backend new DV version changes implemented as discussed.

libvast/vast/layout_version.hpp
```diff
@@ -15,40 +15,17 @@
...
-//     registry
-
-// clang-format on
-
 namespace vast {
 
+/// This version number defines compatibility of persistent state with with
+/// prior directory layouts and contents. Breaking changes shall bump the
+/// version number. A breaking change includes the structure, sequence and
+/// organization of the database directory itself, as well irreconcilable
```
dominiklohmann: This is missing some word before `irreconcilable`.

dominiklohmann: Also, can you please use `incompatible` instead? I had to look up what `irreconcilable` even means.

dominiklohmann (changes requested):
As discussed via Slack, there's a segfault when the db version mismatches. Just nits otherwise.

dominiklohmann (approved):
Tested locally, works well now.

PR #1189, opened 2020-11-20
Fix a hang when trying to process an invalid query

tobim:

We (I) forgot to deliver the response promise in one place in the `spawn` handler in the node.

###  :memo: Checklist


dominiklohmann:
Nice catch! How did you find this one?

mavam (approved):
I tested this by verifying that `vast export json 'ff02:2'` no longer hangs but returns a parse error instead.

mavam:
@tobim could you add an integration test that a failing query terminates?

PR #1190, opened 2020-11-20
Disable SIMD for nix-based CI builds

tobim:

This turns off simd in Apache Arrow for the CI build type.

### DIFF
--- a/nix/broker/default.nix
+++ b/nix/broker/default.nix
@@ -1,11 +1,11 @@
 { stdenv, lib, fetchgit, cmake, caf, openssl, python, ncurses
-, static ? stdenv.hostPlatform.isMusl
-, linkTimeOptimization ? static }:
+, isStatic ? stdenv.hostPlatform.isStatic
+, linkTimeOptimization ? isStatic }:
 
 let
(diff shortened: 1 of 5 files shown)


mavam:
Can you elaborate why we should do this?

dominiklohmann:
How does this even work? Don't you need to depend on a custom build of Arrow that builds with `ARROW_SIMD_LEVEL=NONE`?

tobim:
> Can you elaborate why we should do this?

We need an x86_64 binary that doesn't use any simd instructions beyond SSE2. Arrow's default is SSE4.2, so a vast linked to that won't work on old hardware.

tobim:
> How does this even work? Don't you need to depend on a custom build of Arrow that builds with `ARROW_SIMD_LEVEL=NONE`?

Nix is a declarative system, if the requested library doesn't exist it gets built on-demand.

dominiklohmann:
> Nix is a declarative system, if the requested library doesn't exist it gets built on-demand.

Thanks for the explanation. Will take a look at this once the CI runs through.

tobim:
[ch20593]

dominiklohmann (approved):
LGTM!

PR #1192, opened 2020-11-20
Improve the chunk API

dominiklohmann:

This improves the chunk API. A lot.

###  :memo: Checklist

Commit-by-commit.


libvast/vast/chunk.hpp
```diff
@@ -27,7 +27,7 @@
 
 namespace vast {
 
-/// A contiguous block of memory; essentially a span of bytes with a deleter.
+/// A contiguous block of memory.
```
mavam, line 30: You could help the reader a little bit more by explaining the copy/reference semantics. For example, *"A reference-counted contiguous block of memory. A chunk supports custom deleters for custom deallocations when the last instance goes out of scope."*

tools/lsvast/lsvast.cpp
```diff
@@ -451,7 +451,8 @@ void print_segment_v0(const vast::fbs::segment::v0* segment,
     indented_scope _(indent);
     size_t total_size = 0;
     for (auto flat_slice : *segment->slices()) {
-      auto chunk = vast::chunk::view(*flat_slice->data());
+      auto chunk = vast::chunk::make(flat_slice->data()->data(),
+                                     flat_slice->data()->size(), {});
```
mavam, line 455: Isn't this way of calling hiding a very important fact, namely that you are exploiting a (purposeful) weakness in the API by passing a no-op deleter? I feel this should be made a bit more explicit to the reader, who may not at all remember the implications of this.

mavam (changes requested):
I didn't test this, since it's only API refactoring and covered by unit tests.

tools/lsvast/lsvast.cpp
dominiklohmann: This is exactly why I introduced `chunk::view` in the first place: To make this explicit. A similar design can be found in Apache Arrow, whose `arrow::Buffer` also has a `Buffer::View` factory function. (Buffer and chunk are functionally identical, but that the buffer takes an allocator object instead of a deleter.)

libvast/vast/chunk.hpp
dominiklohmann:
```suggestion
/// A reference-counted contiguous block of memory. A chunk supports custom
/// deleters for custom deallocations when the last instance goes out of scope.
```

tools/lsvast/lsvast.cpp
mavam: I thought about this and think that the best way to handle this is making a *copy* in the unit tests, where data volume is low. Then we don't ever have to pass `{}` to a deleter. (I'm not sure if it can be checked that the unique function is "engaged", like a `std::function`)

It's important that we don't drill holes into the chunk API that can break the abstraction. We want that users know that they own a chunk when they got instance from somewhere. With `chunk::view`, this can be violated, so we need a safer API.

I'm not sure if the Arrow `Buffer` comparison makes sense here. It seems that it has less tight semantics, as it supports multiple modes of ownership.

dominiklohmann: For the unit tests we're already making copies. But this is `lsvast`, where we shouldn't just copy all table slices in all segments.

mavam: Gotcha. Here, `chunk` feels like the wrong abstraction then. Why not make the stuff downstream work with `span<byte>` if somebody else manages the memory? At least constructing a `table_slice` from a `span<byte>` sounds non-trivial, right?

dominiklohmann: We had a quick call about this and agreed that moving the ownership problem to table slice isn't good either. This problem is very much local to lsvast, since it uses the `vast::fbs::Segment` directly instead of going through `vast::segment`. I'll add a comment above that explains the issue.

dominiklohmann:
@mavam this is ready to review again if you have time.

mavam (approved):

PR #1195, opened 2020-11-23
Detect and handle breaking changes in schemas

dominiklohmann:

When changing schemas such that types are no longer a superset of previous versions of the same type, VAST may end up in a state where it cannot import events of that type.

This change makes the type registry ~forget about incompatible versions of types automatically when a new and incompatible type is registered.~ warn when this happens and ensures that newly registered types get priority over older types when importing data.

###  :memo: Checklist

Commit-by-commit.

### DIFF
--- a/libvast/src/system/type_registry.cpp
+++ b/libvast/src/system/type_registry.cpp
@@ -98,17 +98,20 @@ caf::error type_registry_state::load_from_disk() {
 }
 
 void type_registry_state::insert(vast::type layout) {
-  // Flatten all incoming types first.
-  auto new_layout = flatten(layout);
-  auto& old_layouts = data[new_layout.name()];
+  auto& old_layouts = data[layout.name()];
   // Insert into the existing bucket.
-  auto [hint, success] = old_layouts.insert(std::move(new_layout));
+  auto [hint, success] = old_layouts.insert(std::move(layout));
   if (success) {
     // Check whether the new layout is compatible with the latest, i.e., whether
     // the new layout is a superset of it.
-    if (old_layouts.begin() != hint && is_subset(*old_layouts.begin(), *hint))
-      VAST_WARNING(self, "detected an incompatible version of", hint->name());
-    VAST_DEBUG(self, "registered", hint->name());
+    if (old_layouts.begin() != hint) {
+      if (!is_subset(*old_layouts.begin(), *hint))
+        VAST_WARNING(self, "detected an incompatible layout change for",
+                     hint->name());
+      else
+        VAST_VERBOSE(self, "detected a layout change for", hint->name());
+    }
+    VAST_INFO(self, "registered", hint->name());
   }
   // Move the newly inserted layout to the front.
   std::rotate(old_layouts.begin(), hint, std::next(hint));
--- a/libvast/src/system/type_registry.cpp
+++ b/libvast/src/system/type_registry.cpp
@@ -98,11 +98,48 @@ caf::error type_registry_state::load_from_disk() {
 }
 
 void type_registry_state::insert(vast::type layout) {
-  auto& cont = data[layout.name()].value;
+  auto flattened_layout = flatten(layout);
+  auto& old_layouts = data[flattened_layout.name()].value;
+  // Helper function.
+  auto is_compatible = [](const type& x) noexcept {
+    auto lhs = caf::get_if<record_type>(&x);
+    return [&, lhs](const type& y) noexcept {
+      auto rhs = caf::get_if<record_type>(&y);
+      // If they are not record types, check if they are congruent.
+      if (!lhs || !rhs)
+        return congruent(x, y);
+      // Check whether all fields of rhs exist in lhs, i.e., no new fields were
+      // added.
+      for (auto&& rf : rhs->fields) {
+        if (auto lf = lhs->find(rf.name)) {
+          // If there's a type mismatch, exit early as well.
+          if (!congruent(lf->type, rf.type))
+            return false;
+        } else {
+          // Not all fields of rhs exist in lhs. Exit early.
+          return false;
+        }
+      }
+      // We passed the test for all fields. Yay!
+      return true;
+    };
+  };
+  // Check whether the new layout is compatible with the old one. If it isn't,
+  // forget about old versions of the layout.
+  auto are_all_compatible = std::all_of(old_layouts.begin(), old_layouts.end(),
+                                        is_compatible(flattened_layout));
+  if (!are_all_compatible) {
+    VAST_VERBOSE(self, "detected an incompatible version of",
+                 flattened_layout.name(), "and forgets existing versions",
+                 flattened_layout.name());
+    old_layouts.clear();
+  }
+  // Insert into the existing bucket.
   if ([[maybe_unused]] auto [hint, success]
-      = cont.insert(flatten(std::move(layout)));
-      success)
+      = old_layouts.insert(std::move(flattened_layout));
+      success) {
     VAST_DEBUG(self, "registered", hint->name());
+  }
 }
 
 type_set type_registry_state::types() const {
--- a/libvast/src/system/type_registry.cpp
+++ b/libvast/src/system/type_registry.cpp
@@ -100,35 +100,27 @@ caf::error type_registry_state::load_from_disk() {
 void type_registry_state::insert(vast::type layout) {
   // Flatten all incoming types first.
   auto new_layout = flatten(layout);
-  auto& old_layouts = data[new_layout.name()].value;
-  // Check whether the new layout is compatible with the old one, i.e., whether
-  // the new layout is a superset of all existing layouts. If it isn't, forget
-  // about old versions of the layout.
-  auto is_compatible
-    = [&](const auto& old_layout) { return is_subset(old_layout, new_layout); };
-  auto are_all_compatible
-    = std::all_of(old_layouts.begin(), old_layouts.end(), is_compatible);
-  if (!are_all_compatible) {
-    VAST_WARNING(self, "detected an incompatible version of", new_layout.name(),
-                 "and deletes existing versions");
-    old_layouts.clear();
-  }
+  auto& old_layouts = data[new_layout.name()];
+  // Check whether the new layout is compatible with the latest, i.e., whether
+  // the new layout is a superset of it.
+  if (!old_layouts.empty())
+    if (!is_subset(*old_layouts.begin(), new_layout))
+      VAST_WARNING(self, "detected an incompatible version of",
+                   new_layout.name());
   // Insert into the existing bucket.
-  if ([[maybe_unused]] auto [hint, success]
-      = old_layouts.insert(std::move(new_layout));
-      success) {
+  auto [hint, success] = old_layouts.insert(std::move(new_layout));
+  if (success)
     VAST_DEBUG(self, "registered", hint->name());
-  }
+  // Move the newly inserted layout to the front.
+  std::rotate(old_layouts.begin(), hint, std::next(hint));
 }
 
 type_set type_registry_state::types() const {
-  auto result = std::unordered_set<vast::type>{};
-  // TODO: Replace merging logic once libc++ implements unordered_set::merge.
-  //   result.merge(data);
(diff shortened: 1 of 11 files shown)


schema/test.schema
```diff
@@ -8,6 +8,6 @@ type test.full = record{
  t: time #default="uniform(0,10)",
  d: duration #default="uniform(100,200)",
  a: addr #default="uniform(0,2000000)",
- s: subnet #default="uniform(1000,2000)",
+ u: subnet #default="uniform(1000,2000)",
```
mavam, line 11: I feel that testing the feature by introducing a mismatch between the schemas we ship and the hard-coded ones is difficult to argue. Instead, you could the custom schema in the integration tests, no?

libvast/src/system/type_registry.cpp
```diff
@@ -88,11 +88,48 @@ caf::error type_registry_state::load_from_disk() {
...
+            return false;
+        } else {
+          // Not all fields of rhs exist in lhs. Exit early.
+          return false;
+        }
+      }
+      // We passed the test for all fields. Yay!
+      return true;
+    };
+  };
```
mavam, line 116: Would anything prevent against beefing up the function `congruent` with that logic? Then it could also be truly recursive.

libvast/src/system/type_registry.cpp
```diff
@@ -88,11 +88,48 @@ caf::error type_registry_state::load_from_disk() {
...
+  // Check whether the new layout is compatible with the old one. If it isn't,
+  // forget about old versions of the layout.
+  auto are_all_compatible = std::all_of(old_layouts.begin(), old_layouts.end(),
+                                        is_compatible(flattened_layout));
+  if (!are_all_compatible) {
+    VAST_VERBOSE(self, "detected an incompatible version of",
```
mavam, line 122:
```suggestion
    VAST_WARNING(self, "detected an incompatible version of",
```

Up to you, but I'd see this as something beyond VERBOSE.

mavam (changes requested):
I wonder whether there's a good way to unit-test this?!

schema/test.schema
dominiklohmann: This is a bugfix. There's a second field above also called `s` that has a different type. This triggered the newly introduced mechanism on every start of the type registry.

libvast/src/system/type_registry.cpp
dominiklohmann: I'd rather duplicate the logic and call it `compatible`. This is not about congruency. Note that as is this expects flattened types.

libvast/src/system/type_registry.cpp
dominiklohmann: Fine by me.

dominiklohmann:
> I wonder whether there's a good way to unit-test this?!

I've added a unit test as a separate commit. It isn't beautiful, but it works.

libvast/test/system/type_registry.cpp
```diff
@@ -57,6 +68,15 @@ make_data_b(std::string a, vast::count b, vast::real c, std::string d) {
   return make_data(mock_layout_b, a, b, c, d);
 }
 
+// This one is incompatible with the others.
```
mavam, line 71: In what sense incompatible?

dominiklohmann: I've expanded on this a bit with the most recent commit (c6ce834)

libvast/vast/type.hpp
```diff
@@ -824,6 +824,12 @@ bool compatible(const type& lhs, relational_operator op, const data& rhs);
 /// @relates type data
 bool compatible(const data& lhs, relational_operator op, const type& rhs);
 
+/// Checks whether a type is a superset of another.
```
mavam, line 827: Please explain how you define "superset".

mavam: Also, I would flip the argument and call it `is_subset`, which is the more common way to look at a type: one type is contained (= subset) within another.

libvast/src/type.cpp
```diff
@@ -704,6 +704,29 @@ bool compatible(const data& lhs, relational_operator op, const type& rhs) {
   return compatible(rhs, flip(op), lhs);
 }
 
+bool is_superset(const type& x, const type& y) {
+  auto sub = caf::get_if<record_type>(&x);
+  auto super = caf::get_if<record_type>(&y);
```
mavam, line 709:
```suggestion
  auto sup = caf::get_if<record_type>(&y);
```

For consistency.

libvast/src/type.cpp
```diff
@@ -704,6 +704,29 @@ bool compatible(const data& lhs, relational_operator op, const type& rhs) {
...
+  // If either of the types is not a record type, check if they are
+  // congruent instead.
+  if (!sub || !super)
+    return congruent(x, y);
+  // Check whether all fields of the subset exist in the superset, i.e., no new
+  // fields were added.
```
mavam, line 715: The "no new fields were added" confuses me. It's about checking difference between two given types, no?

libvast/src/type.cpp
```diff
@@ -704,6 +704,29 @@ bool compatible(const data& lhs, relational_operator op, const type& rhs) {
...
+  for (auto&& field : sub->fields) {
```
mavam, line 716:
```suggestion
  for (auto& field : sub->fields) {
```

libvast/src/type.cpp
```diff
@@ -704,6 +704,29 @@ bool compatible(const data& lhs, relational_operator op, const type& rhs) {
...
+    if (auto match = super->find(field.name)) {
+      // Perform the check recursively to support nested record txpes.
```
mavam, line 718:
```suggestion
      // Perform the check recursively to support nested record types.
```

libvast/src/type.cpp
```diff
@@ -704,6 +704,29 @@ bool compatible(const data& lhs, relational_operator op, const type& rhs) {
...
+      if (!is_superset(field.type, match->type))
+        return false;
+    } else {
+      // Not all fields of the subset exist in the superset; exit early.
+      return false;
+    }
+  }
+  // We passed the test for all fields. Yay!
```
mavam, line 726:
```suggestion
```

Unfortunately this not adding much insight, so we can ditch it.

libvast/src/system/type_registry.cpp
```diff
@@ -88,11 +88,28 @@ caf::error type_registry_state::load_from_disk() {
...
+  // the new layout is a superset of all existing layouts. If it isn't, forget
+  // about old versions of the layout.
+  auto is_compatible = [&](const auto& old_layout) {
+    return is_superset(old_layout, new_layout);
+  };
+  auto are_all_compatible
+    = std::all_of(old_layouts.begin(), old_layouts.end(), is_compatible);
+  if (!are_all_compatible) {
+    VAST_WARNING(self, "detected an incompatible version of", new_layout.name(),
+                 "and forgets existing versions");
```
mavam, line 104:
```suggestion
                 "and deletes existing versions");
```

mavam (dismissed):
The logic looks sounds to me. Just small stuff now before merging.

libvast/src/type.cpp
dominiklohmann: We should use `const auto&` or `auto&&` for this, but never `auto&`.

```suggestion
  for (const auto&& field : sub->fields) {
```

libvast/src/type.cpp
dominiklohmann: Not feeling this one, nobody ever said supset.

libvast/src/type.cpp
dominiklohmann: Ah yes, this comment moved over from when this was inline in the type registry, where we had a new and multiple old types.

dominiklohmann:
```suggestion
  // Check whether all fields of the subset exist in the superset.
```

dominiklohmann:
This should be reviewed by @tobim after our Slack call yesterday on the matter. The last commit is the most important one here.

libvast/vast/system/source.hpp
```diff
@@ -164,11 +164,11 @@ struct source_state {
           };
           // First, merge and de-duplicate the local schema with types from the
           // type-registry.
-          types.value.insert(std::make_move_iterator(st.local_schema.begin()),
-                             std::make_move_iterator(st.local_schema.end()));
+          types.insert(std::make_move_iterator(st.local_schema.begin()),
+                       std::make_move_iterator(st.local_schema.end()));
```
tobim, line 168: This and the `clear()` in the next line have the effect that a modified `*.schema` file that is passed to the `--schema-file=` option will not be considered. I know this is not a result of this change, but I just stumbled over this while testing this change.

libvast/src/system/type_registry.cpp
```diff
@@ -86,21 +86,29 @@ caf::error type_registry_state::load_from_disk() {
...
+  // Check whether the new layout is compatible with the latest, i.e., whether
+  // the new layout is a superset of it.
+  if (!old_layouts.empty())
+    if (!is_subset(*old_layouts.begin(), new_layout))
+      VAST_WARNING(self, "detected an incompatible version of",
+                   new_layout.name());
+  // Insert into the existing bucket.
+  auto [hint, success] = old_layouts.insert(std::move(new_layout));
+  if (success)
     VAST_DEBUG(self, "registered", hint->name());
```
tobim, line 101: This can be improved so that `is_subset()` is only called if the new layout was successfully inserted.
```suggestion
  // Insert into the existing bucket.
  auto [hint, success] = old_layouts.insert(std::move(new_layout));
  if (success) {
    if (hint != old_layouts.begin()) {
      if (!is_subset(*old_layouts.begin(), new_layout))
        VAST_WARNING(self, "detected an incompatible version of",
                     new_layout.name());
    }
    VAST_DEBUG(self, "registered", hint->name());
  }
```

tobim (changes requested):

libvast/src/system/type_registry.cpp
```diff
@@ -98,17 +98,20 @@ caf::error type_registry_state::load_from_disk() {
...
-      VAST_WARNING(self, "detected an incompatible version of", hint->name());
-    VAST_DEBUG(self, "registered", hint->name());
+    if (old_layouts.begin() != hint) {
+      if (!is_subset(*old_layouts.begin(), *hint))
+        VAST_WARNING(self, "detected an incompatible layout change for",
+                     hint->name());
+      else
+        VAST_VERBOSE(self, "detected a layout change for", hint->name());
+    }
+    VAST_INFO(self, "registered", hint->name());
```
dominiklohmann, line 114: This is way too noisy on start with an empty db. We originally had this on verbose, and moved it to debug just recently. I'd say move the above "detected a layout change" from verbose to info and move this to debug again.

libvast/vast/system/source.hpp
```diff
@@ -158,22 +158,23 @@ struct source_state {
...
           // First, merge and de-duplicate the local schema with types from the
           // type-registry.
-          types.insert(std::make_move_iterator(st.local_schema.begin()),
-                       std::make_move_iterator(st.local_schema.end()));
-          st.local_schema.clear();
+          schema s;
+          for (auto& type : local_schema)
+            if (auto&& layout = caf::get_if<vast::record_type>(&type))
+              if (is_valid(*layout))
+                s.add(std::move(*layout));
```
dominiklohmann, line 170: Can you drop this commit? It is entirely unrelated from my PR and we said yesterday that we want to do this in a separate PR.

Not quite sure why you've just pushed this on top, because we were already done reviewing the logic and now there needs to be yet another reviewer.

libvast/src/system/type_registry.cpp
tobim: Ah, I disabled the loading of the on-disk types during testing. Should have done a final check...

libvast/vast/system/source.hpp
tobim: I added this because it is **actually required** to fix the reported issue. If we don't give preference to the local schema at the source there is no way it can ever make its way into the type registry without a restart of the node.

tobim (approved):

PR #1197, opened 2020-11-24
Remove unused LZ4 dependency

mavam:

This PR removes the dependency on LZ4, which we no longer need.

###  :memo: Checklist

Commit-by-commit.


dominiklohmann (changes requested):
Looks good to me, but I'm not a fan of the `as_bytes` header. The name really doesn't describe well what it does.

Maybe remove them outright and add this to `span.hpp` or `byte.hpp`:

```
  template <class Buffer,
            class = std::enable_if_t<sizeof(*std::data(std::declval<Buffer>())) == 1>>
  static auto make(const Buffer& buffer)
    -> decltype(std::data(buffer), std::size(buffer), span<const byte>{}) {
  return span{reinterpret_cast<const byte*>(std::data(buffer)), std::size(buffer)};
}
```

libvast/vast/as_bytes.hpp
```diff
@@ -11,28 +11,27 @@
...
-#include "vast/compression.hpp"
-#include "vast/die.hpp"
+#include "vast/byte.hpp"
+#include "vast/detail/type_traits.hpp"
+#include "vast/span.hpp"
 
-namespace vast {
-namespace lz4 {
+#include <type_traits>
+#include <vector>
```
dominiklohmann, line 21: This include should no longer be necessary.

libvast/vast/chunk.hpp
```diff
@@ -124,6 +90,20 @@ class chunk final : public caf::ref_counted {
   /// Avoid the common mistake of binding ownership to a string view.
   static auto make(std::string_view&&) = delete;
 
+  /// Construct a chunk from a byte buffer by copying it.
+  /// @param buffer The byte buffer.
+  /// @returns A chunk pointer or `nullptr` on failure.
+  template <class Buffer>
+  static auto copy(const Buffer& buffer)
```
mavam, line 97: This works for now. Once our chunks support allocators, we could offer a second argument that contains an allocator.

dominiklohmann (approved):
I've done the required fixups necessary to finalize this PR. @mavam and I reviewed each other here, so I am approving his work, and he is approving my work by the means of merging. Let's wait for CI with merging, though.

PR #1200, opened 2020-11-25
Send correct message type when replacing partition

lava:
###  :memo: Checklist

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -747,6 +747,7 @@ index(caf::stateful_actor<index_state>* self, filesystem_type fs, path dir,
     // to be copy constructible.
     [=](atom::replace, uuid partition_id,
         std::shared_ptr<partition_synopsis>& ps) {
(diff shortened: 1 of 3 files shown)


libvast/src/system/index.cpp
```diff
@@ -747,6 +747,7 @@ index(caf::stateful_actor<index_state>* self, filesystem_type fs, path dir,
     // to be copy constructible.
     [=](atom::replace, uuid partition_id,
         std::shared_ptr<partition_synopsis>& ps) {
+      VAST_DEBUG(self, "replacing synopsis for partition", partition_id);
```
dominiklohmann, line 750:
```suggestion
      VAST_DEBUG(self, "replaces synopsis for partition", partition_id);
```

dominiklohmann (approved):
The code change itself is obviously fine modulo a proposed wording fix, but can you add a changelog entry for the fixed memory leak and also append the changelog entry for the shrinking with this PRs number, because that did not work as expected either?

PR #1201, opened 2020-11-26
Ensure get isn't called on a moved-from unique_ptr 

dominiklohmann:

Just a minor fix for a minor issue that I just noticed. The order of this was not well-defined before. It happened to work, but let's not rely on that.

###  :memo: Checklist

Commit-by-commit.

### DIFF
--- a/libvast/vast/chunk.hpp
+++ b/libvast/vast/chunk.hpp
@@ -78,8 +78,8 @@ public:
               std::disjunction<std::is_lvalue_reference<Buffer>,
                                std::is_trivially_move_assignable<Buffer>>>>>
   static auto make(Buffer&& buffer) -> decltype(as_bytes(buffer), chunk_ptr{}) {
-    auto view = as_bytes(buffer);
-    return make(view, [buffer = std::move(buffer)]() noexcept {
+    const auto view = as_bytes(buffer);
+    return make(view, [buffer = std::exchange(buffer, {})]() noexcept {
       static_cast<void>(buffer);
     });
   }
@@ -97,9 +97,9 @@ public:
   template <class Buffer>
   static auto copy(const Buffer& buffer)
     -> decltype(as_bytes(buffer), chunk_ptr{}) {
-    auto view = as_bytes(buffer);
+    const auto view = as_bytes(buffer);
     auto copy = std::make_unique<value_type[]>(view.size());
-    auto data = copy.get();
+    const auto data = copy.get();
     std::memcpy(data, view.data(), view.size());
     return make(data, view.size(), [copy = std::move(copy)]() noexcept {
       static_cast<void>(copy);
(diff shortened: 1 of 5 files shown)


libvast/vast/chunk.hpp
```diff
@@ -78,8 +78,8 @@ class chunk final : public caf::ref_counted {
               std::disjunction<std::is_lvalue_reference<Buffer>,
                                std::is_trivially_move_assignable<Buffer>>>>>
   static auto make(Buffer&& buffer) -> decltype(as_bytes(buffer), chunk_ptr{}) {
-    auto view = as_bytes(buffer);
-    return make(view, [buffer = std::move(buffer)]() noexcept {
+    const auto view = as_bytes(buffer);
+    return make(view, [buffer = std::exchange(buffer, {})]() noexcept {
       static_cast<void>(buffer);
```
lava, line 83: Shouldn't we technically call `buffer.~Buffer()` in the deleter? That seems to be not harder to understand than the `static_cast<void>`, and explicitly deletes the content rather than relying on the fact that the deleter itself gets deleted.

libvast/vast/chunk.hpp
```diff
@@ -78,8 +78,8 @@ class chunk final : public caf::ref_counted {
               std::disjunction<std::is_lvalue_reference<Buffer>,
                                std::is_trivially_move_assignable<Buffer>>>>>
   static auto make(Buffer&& buffer) -> decltype(as_bytes(buffer), chunk_ptr{}) {
```
lava, line 80: Can we maybe document the interface requirements for this function, like `This overload can only be selected for if the argument is an rvalue-reference, has a user-defined move constructor, and an overload of 'as_bytes' exists for this type. This is intended to guard against accidental copies when calling `chunk::make()`

I'm also not really sure that it's a good idea to make the `as_bytes()` part of the signature/overload resolution, as opposed to a static_assert in the function body.

lava (approved):
The actual changes in the review look reasonable, but since this is a bit broader than just ensuring `get()` isn't called on a moved-from unique ptr it felt fair to comment on the chunk API in general as well.

libvast/vast/chunk.hpp
dominiklohmann: We could do this, I don't really mind the change. Feel free to do so in a new PR.

libvast/vast/chunk.hpp
dominiklohmann: > Can we maybe document the interface requirements for this function, like This overload can only be selected for if the argument is an rvalue-reference, has a user-defined move constructor, and an overload of 'as_bytes' exists for this type. This is intended to guard against accidental copies when calling chunk::make()`

Will add a sentence before merging.

> I'm also not really sure that it's a good idea to make the `as_bytes()` part of the signature/overload resolution, as opposed to a static_assert in the function body.

It's required so it's used for SFINAE. There's no other way to do this without `requires`, which requires C++20.

PR #1202, opened 2020-11-26
Fix the length of the abbreviated commit hash

tobim:
VAST uses the output of git describe in its version string, which is baked into the binary. Since v2.11 git estimates the length needed to avoid colisions based on the number of objects in the .git directory. That means that building VAST from a shallow clone would result in a different binary artifact compared to a build from a regular clone.


dominiklohmann (approved):
Thanks for this.

PR #1203, opened 2020-11-26
Avoid reallocations in the segment builder

dominiklohmann:

This is an experimental PR that sets the initial buffer size of the underlying `FlatBufferBuilder` of the `segment_builder` to 110% of `vast.max-segment-size`. The 10% extra is needed because the configured limit is a soft limit, and because there's still some extra scaffolding in the FlatBuffers table that is only added when we call `segment_builder::finish()`.

###  :memo: Checklist

n/t

### DIFF
--- a/libvast/src/segment_builder.cpp
+++ b/libvast/src/segment_builder.cpp
@@ -29,7 +29,8 @@
 
 namespace vast {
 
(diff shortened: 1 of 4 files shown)


lava (approved):
lgtm

PR #1204, opened 2020-11-26
Avoid stream errors for short-lived commands

dominiklohmann:

This avoids stream errors on short-lived commands like `vast -N dump concepts`. These were really non-errors caused by early exits from shutting down the node. We shouldn't print red stuff when there isn't actually an error.

> 2020-11-26T14:40:06.085 archive got a stream error: exit_reason(user_shutdown)
> 2020-11-26T14:40:06.086 index aborts with error: !! exit

###  :memo: Checklist

Run locally before and after the change I've made, verify that the quoted errors are no more.

### DIFF
--- a/libvast/src/system/archive.cpp
+++ b/libvast/src/system/archive.cpp
@@ -191,6 +191,8 @@ archive(archive_type::stateful_pointer<archive_state> self, path dir,
           if (err && err != caf::exit_reason::unreachable) {
             if (err != caf::exit_reason::user_shutdown)
               VAST_ERROR(self, "got a stream error:", render(err));
+            else
+              VAST_DEBUG(self, "got a user shutdown error:", render(err));
             // We can shutdown now because we only get a single stream from the
             // importer.
             self->send_exit(self, err);
--- a/libvast/src/system/archive.cpp
+++ b/libvast/src/system/archive.cpp
@@ -185,9 +185,17 @@ archive(archive_type::stateful_pointer<archive_state> self, path dir,
           t.stop(events);
         },
         [=](unit_t&, const error& err) {
-          if (err) {
-            VAST_ERROR(self, "got a stream error:", self->system().render(err));
+          // We get an 'unreachable' error when the stream becomes unreachable
+          // because the actor was destroyed; in this case we can't use `self`
+          // anymore.
+          if (err && err != caf::exit_reason::unreachable) {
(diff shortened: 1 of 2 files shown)


libvast/src/system/archive.cpp
```diff
@@ -185,9 +185,17 @@ archive(archive_type::stateful_pointer<archive_state> self, path dir,
           t.stop(events);
         },
         [=](unit_t&, const error& err) {
-          if (err) {
-            VAST_ERROR(self, "got a stream error:", self->system().render(err));
+          // We get an 'unreachable' error when the stream becomes unreachable
+          // because the actor was destroyed; in this case we can't use `self`
+          // anymore.
```
lava, line 190: In retrospect I'm not sure anymore that the comment is 100% accurate: Removing the check in the partition gives occasional segfaults due to null pointer dereference of `self`, but a trivial reproduction of the situation did not have the same issue, so maybe this is something specific to the setup of the index/partition. (i was planning to investigate this for the streaming blog post as well, but didn't really get to it yet)

libvast/src/system/archive.cpp
```diff
@@ -185,9 +185,17 @@ archive(archive_type::stateful_pointer<archive_state> self, path dir,
...
+          if (err && err != caf::exit_reason::unreachable) {
+            if (err != caf::exit_reason::user_shutdown)
```
lava, line 192: I think `user_shutdown` might also be used as a default argument when shutting down the stream, not just when its literally the user shutting down the stream, but I don't completely remember tbh. Did you search for this in the CAF source code?

dominiklohmann: Yes, user_shutdown is just the default exit reason. It doesn't make sense to treat it as an error.

libvast/src/system/archive.cpp
dominiklohmann: Should we investigate this separately from this PR?

libvast/src/system/archive.cpp
```diff
@@ -185,9 +185,17 @@ archive(archive_type::stateful_pointer<archive_state> self, path dir,
...
+              VAST_ERROR(self, "got a stream error:", render(err));
```
lava, line 193: I think we should still print the error on `DEBUG`, to be able to distinguish between unreachable and user_shutdown in the logs.

lava (approved):
Looking good

PR #1205, opened 2020-11-26
Make Zeek writer work with all data types

mavam:

This PR lifts the restriction of the Zeek writer to only dump Zeek events. There is no conceptual limitation, but VAST had an assertition that prevented rendering other data than Zeek.

###  :memo: Checklist

Commit-by-commit.

### DIFF
--- a/libvast/src/format/zeek.cpp
+++ b/libvast/src/format/zeek.cpp
@@ -154,7 +154,7 @@ Stream& operator<<(Stream& out, const time_factory& t) {
   return out;
 }
 
-void print_header(const type& t, std::ostream& out) {
+void print_header(const type& t, std::ostream& out, bool show_timestamp_tags) {
   auto path = std::string_view{t.name()};
   auto i = path.find('.');
   if (i != path.npos) {
@@ -165,9 +165,10 @@ void print_header(const type& t, std::ostream& out) {
       << "#set_separator" << separator << set_separator << '\n'
       << "#empty_field" << separator << empty_field << '\n'
       << "#unset_field" << separator << unset_field << '\n'
-      << "#path" << separator << path << '\n'
-      << "#open" << separator << time_factory{} << '\n'
-      << "#fields";
+      << "#path" << separator << path << '\n';
+  if (show_timestamp_tags)
+    out << "#open" << separator << time_factory{} << '\n';
+  out << "#fields";
   auto r = caf::get<record_type>(t);
   for (auto& e : record_type::each{r})
     out << separator << to_string(e.key());
@@ -497,7 +498,8 @@ caf::error reader::parse_header() {
   return caf::none;
 }
 
-writer::writer(path dir) {
+writer::writer(path dir, bool show_timestamp_tags)
+  : show_timestamp_tags_{show_timestamp_tags} {
   if (dir != "-")
     dir_ = std::move(dir);
 }
@@ -588,10 +590,12 @@ class writer_child : public ostream_writer {
 public:
   using super = ostream_writer;
 
-  using super::super;
+  writer_child(std::unique_ptr<std::ostream> out, bool show_timestamp_tags)
+    : super{std::move(out)}, show_timestamp_tags_{show_timestamp_tags} {
+  }
 
   ~writer_child() override {
-    if (out_ != nullptr)
+    if (out_ != nullptr && show_timestamp_tags_)
       *out_ << "#close" << separator << time_factory{} << '\n';
   }
 
@@ -604,6 +608,8 @@ public:
   const char* name() const override {
     return "zeek-writer";
   }
+
+  bool show_timestamp_tags_;
 };
 
 } // namespace
@@ -615,12 +621,12 @@ caf::error writer::write(const table_slice& slice) {
     if (writers_.empty()) {
       VAST_DEBUG(this, "creates a new stream for STDOUT");
       auto out = std::make_unique<detail::fdostream>(1);
-      writers_.emplace(layout.name(),
-                       std::make_unique<writer_child>(std::move(out)));
+      writers_.emplace(layout.name(), std::make_unique<writer_child>(
+                                        std::move(out), show_timestamp_tags_));
     }
     child = writers_.begin()->second.get();
     if (layout != previous_layout_) {
-      print_header(layout, child->out());
+      print_header(layout, child->out(), show_timestamp_tags_);
       previous_layout_ = std::move(layout);
     }
   } else {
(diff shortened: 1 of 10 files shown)


mavam:
@tenzir/backend This is the CI failure:

```
--- /home/runner/work/vast/vast/integration/reference/node-suricata-alert/step_03.ref
510
+++ vast-integration-test/node-suricata-alert/step_03.out
511
@@ -1,7 +1,7 @@
512
-#close	2020-11-26-15-24-33
513
+#close	2020-11-26-15-42-56
514
 #empty_field	(empty)
515
 #fields	timestamp	flow_id	pcap_cnt	vlan	in_iface	src_ip	src_port	dest_ip	dest_port	proto	event_type	community_id	alert.app_proto	alert.action	alert.gid	alert.signature_id	alert.rev	alert.signature	alert.category	alert.severity	alert.source.ip	alert.source.port	alert.target.ip	alert.target.port	flow.pkts_toserver	flow.pkts_toclient	flow.bytes_toserver	flow.bytes_toclient	flow.start	flow.end	flow.age	flow.state	flow.reason	flow.alerted	payload	payload_printable	stream	packet	packet_info.linktype
516
-#open	2020-11-26-15-24-33
517
+#open	2020-11-26-15-42-56
518
 #path	alert
519
 #separator 	
520
 #set_separator	,
```

The problem is obvious: The `#open` and `#close` comments are not deterministic. Should we add a flag to the Zeek export to omit them? They are optional, after all. I was thinking of something along the lines of `vast export zeek --no-meta-timestamps`.

libvast/src/system/spawn_sink.cpp
```diff
@@ -82,7 +82,8 @@ maybe_actor spawn_zeek_sink(caf::local_actor* self, spawn_arguments& args) {
   if (!args.empty())
     return unexpected_arguments(args);
   auto writer = std::make_unique<format::zeek::writer>(
-    get_or(args.inv.options, category + ".write", defaults_t::write));
+    get_or(args.inv.options, category + ".write", defaults_t::write),
+    !caf::get_or(args.inv.options, category + ".no-timestamp-tags", false));
```
lava, line 86: `s/no-timestamp-tags/show-timestamp-tags/`?

Otherwise we're dealing with a triple negative here :D

libvast/vast/format/zeek.hpp
```diff
@@ -268,7 +268,10 @@ class writer : public format::writer {
 
   /// Constructs a Zeek writer.
   /// @param dir The path where to write the log file(s) to.
-  writer(path dir);
+  /// @param show_timestamp_tags Flag to control whether to include comments
+  /// with `#open` and `#close` tags in the output. Setting this flag to
+  /// `false` produces a deterministic output.
+  writer(path dir, bool show_timestamp_tags);
```
lava, line 274: Should we add `= true` as default argument, so the caller doesn't have to know details about the zeek format?

lava (approved):
Looks good to me

libvast/src/format/zeek.cpp
```diff
@@ -155,13 +155,17 @@ Stream& operator<<(Stream& out, const time_factory& t) {
 }
 
 void print_header(const type& t, std::ostream& out) {
-  VAST_ASSERT(detail::starts_with(t.name(), type_name_prefix));
-  auto path = t.name().substr(type_name_prefix.size());
+  auto path = std::string_view{t.name()};
+  auto i = path.find('.');
```
lava, line 159: Should this be `if (path.starts_with(type_name_prefix))`? Or do we actually want to cut arbitrary prefixes here?

libvast/src/system/spawn_sink.cpp
mavam: I thought about that, but that makes only sense if it is disabled by default, because then the flag `--show-timestamp-tags` must be provided to achieve the normal Zeek output. I'd rather disable it selectively.

Not sure how to avoid the triple negation. 😜

libvast/vast/format/zeek.hpp
mavam: I also thought about that and decided against that. I couldn't find a code path where it would make sense to hide this detail at the call site.

libvast/src/format/zeek.cpp
mavam: Example: input: `zeek.conn`
Desired output: `conn`

This is because the standard Zeek TSV format doesn't add `zeek` as we do to the type name. What we want is something like `split<1>(type_name, ".")[1]`, i.e., perform a single split at the first dot.

But now that I think about that, maybe we want to keep the full type name for all non-Zeek types? Or would that not be consistent? 

The Zeek output for Zeek data should definitely look as if Zeek produced it.

libvast/src/system/spawn_sink.cpp
lava: Not sure that I follow, you could have `--show-timestamp-flags=true` be the default and disable them with `--show-timestamp-flags=false` in the integration tests?

libvast/src/format/zeek.cpp
lava: `type_name_prefix` is defined as `"zeek."` currently, so for the example input both would produce the same result. The interesting questions are what happens when there's a dot that's not preceded by "zeek.", or when there are multiple dots. I don't know what the most reasonable behavior for these cases would be.

libvast/src/system/spawn_sink.cpp
mavam: What I meant is that, for a boolean flag, the expectation is usually toggle it without setting it with `--flag=false` or `--flat=true`. Usually it's just `--flag` and the absence of `--flag` implies the opposite. This is why I named the flag `--no-X` instead of `--show-X`. Does that make sense?

I you prefer to avoid these negations, I'm happy to flip things on its head with the changes you proposed.

libvast/src/format/zeek.cpp
mavam: >  The interesting questions are what happens when there's a dot that's not preceded by "zeek."

Right now, we always have the format name before the dot. This should work for all formats, though.

> [..] or when there are multiple dots.

It's currently implemented as "split-1", i.e., split on the first dot. If formats now contain a dot, we'll display some weird things.

Maybe the best logic would be: strip `zeek.` if the event starts like this, and use the literal event name otherwise.

libvast/src/system/spawn_sink.cpp
lava: The `--no-X` flag makes sense in isolation, but it doesn't fit with the rest of our command-line syntax which uses `--enable-X/--disable-X` for these things (`--disable-community-id`, `--enable-metrics`, ...)

But I don't have a too strong opinion on this, if you like the current `--no-timestamps-tag` feel free to just resolve the discussion and merge as-is.

mavam: Okay, I switched to `--disable-timestamp-tags` to fit in the naming scheme.

libvast/src/format/zeek.cpp
mavam: I've implemented as noted above. Does that make now sense to you @lava?

lava: Makes sense to me

PR #1207, opened 2020-11-27
Enforce better documentation for breaking changes

lava:

Sample output:
```
$ ./bin/vast start
!! filesystem_error: "wrong or missing database version in db-directory"2020-11-27T15:43:26.603 Cannot start VAST, breaking changes detected in the database directory

2020-11-27T15:43:26.603 Detailed explanation:
 The dedicated `port` type was removed from VAST and replaced by a typedef for count. To update, adjust all custom schemas containing a field of type 'port' and reimport all data that contained a 'port' field.
```

Note that I have no idea how the file system error message can be printed *before* the log messages when the error hasn't even been created yet, but here we are.

###  :memo: Checklist

### DIFF
--- a/libvast/src/system/spawn_node.cpp
+++ b/libvast/src/system/spawn_node.cpp
@@ -52,11 +52,14 @@ spawn_node(caf::scoped_actor& self, const caf::settings& opts) {
   // output is written into the same directory.
   if (auto err = initialize_db_version(abs_dir))
     return err;
-  // TODO(ch20326): Replace this with a more specific check in the components
-  // that rely on a specific DB version.
-  if (read_db_version(abs_dir) != db_version::latest)
-    return make_error(ec::filesystem_error, "wrong or missing database version "
-                                            "in db-directory");
+  if (auto version = read_db_version(abs_dir); version != db_version::latest) {
+    VAST_INFO_ANON("Cannot start VAST, breaking changes detected in the "
+                   "database directory");
+    auto reasons = describe_breaking_changes_since(version);
+    return make_error(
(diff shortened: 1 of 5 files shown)


libvast/src/system/spawn_node.cpp
```diff
@@ -52,11 +52,14 @@ spawn_node(caf::scoped_actor& self, const caf::settings& opts) {
...
   if (auto err = initialize_db_version(abs_dir))
     return err;
-  // TODO(ch20326): Replace this with a more specific check in the components
-  // that rely on a specific DB version.
-  if (read_db_version(abs_dir) != db_version::latest)
+  if (auto version = read_db_version(abs_dir); version != db_version::latest) {
+    VAST_INFO_ANON("Cannot start VAST, breaking changes detected in the "
+                   "database directory");
+    auto reasons = describe_breaking_changes_since(version);
+    VAST_INFO_ANON("Detailed explanation:\n", reasons);
```
dominiklohmann, line 59: Output-wise, wouldn't it make more sense to format the output as a list of breaking changes starting with dashes?

Also I think downstream should print this error message such that we're not relying on the log level info to be available. The following should work (roughly):

```
    auto reasons = describe_breaking_changes_since(version);
    return make_error(ec::breaking_change, "breaking changes detected since last start of the VAST with the current database directory:", reasons);
```

dominiklohmann (changes requested):
Mostly nits. I really like this change generally.

dominiklohmann (approved):
LGTM!

PR #1212, opened 2020-12-04
Ensure uniqueness of query ids

dominiklohmann:

A small fixup for a very unlikely issue that @lava and I noticed.

###  :memo: Checklist

n/t

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -643,6 +643,10 @@ index(caf::stateful_actor<index_state>* self, filesystem_type fs, path dir,
       }
       // Allows the client to query further results after initial taste.
       auto query_id = uuid::random();
+      // Ensure the query id is unique.
+      while (st.pending.find(query_id) != st.pending.end()
+             || query_id == uuid::nil())
+        query_id = uuid::random();
       auto total = candidates.size();
       auto scheduled = detail::narrow<uint32_t>(
         std::min(candidates.size(), st.taste_partitions));


mavam:
Did you make sure that there is not code path that checks for a nil UUID (0...0)? IIRC we used that to signal a special case to the exporter, but I'm not sure if it's processed.

dominiklohmann:
> Did you make sure that there is not code path that checks for a nil UUID (0...0)? IIRC we used that to signal a special case to the exporter, but I'm not sure if it's processed.

That's actually a great point. The code path exists, is checked, but we don't ensure random doesn't return null here. Will add a second check.

dominiklohmann:
I've implemented the suggested change.

mavam:
I actually meant something different. I don't think that `uuid::random()` can return the nil UUID. Instead, I was thinking that we should make sure that a result containing a nil UUID doesn't trigger a special code path at the EXPORTER.

dominiklohmann:
> I don't think that `uuid::random()` can return the nil UUID.

It can, actually. I'll leave the check in place.

> I was thinking that we should make sure that a result containing a nil UUID doesn't trigger a special code path at the EXPORTER.

The special codepath retrieves `(uuid query_id, uint32_t partitions, uint32_t scheduled)`, and uses `partitions == 0` to check for early exits, much like the index itself does. So we're safe in that regard.

mavam (approved):

PR #1214, opened 2020-12-04
Add support for type-level synopses and a string synopsis

tobim:

This adds support for type-level synopses to the meta index. For each field in a partition, it is possible to decide whether to use a specific synopsis, share one with the other fields of the same type, or not to use a synopsis at all.

A new string synopsis and buffered string synopsis is also introduced.

###  :memo: Checklist

By commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🎁 A new type-level synopsis structure in the meta-index now massively
+  speeds up string queries with very few results.
+  [#1214](https://github.com/tenzir/vast/pull/1214)
+
 - ⚡️ The build configuration of VAST received a major overhaul. Inclusion of
   libvast in other procects via `add_subdirectory(path/to/vast)` is now easily
   possible. The names of all build options were aligned, and the new build
--- a/libvast/src/meta_index.cpp
+++ b/libvast/src/meta_index.cpp
@@ -31,6 +31,8 @@
 #include <caf/binary_deserializer.hpp>
 #include <caf/binary_serializer.hpp>
 
+#include <type_traits>
+
 namespace vast {
 
 void partition_synopsis::shrink() {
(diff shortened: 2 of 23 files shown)


libvast/vast/string_synopsis.hpp
```diff
@@ -0,0 +1,226 @@
...
+  synopsis_ptr shrink() const override {
+    size_t next_power_of_two = 1ull;
+    while (strings_.size() > next_power_of_two)
+      next_power_of_two *= 2;
+    bloom_filter_parameters params;
+    params.p = p_;
+    params.n = next_power_of_two;
+    VAST_DEBUG_ANON("shrinked string synopsis to", params.n, "elements");
+    auto& type = this->type();
+    auto shrinked_synopsis
```
mavam, line 76: [shrink shrank shrunk](https://www.merriam-webster.com/words-at-play/shrink-shrank-shrunk-shrinked) 😅 

> Generally, shrank is the simple past tense form of "shrink" like in "I shrank the shirt in the wash." Shrunk is the past participle being paired with "have" as in "I have shrunk the jeans." There are rarer examples of shrinked and shrunken in literature but not enough to support those usages as standard.

libvast/src/meta_index.cpp
```diff
@@ -177,13 +191,28 @@ std::vector<uuid> meta_index::lookup(const expression& expr) const {
...
+                auto opt = syn->lookup(x.op, make_view(rhs));
+                if (!opt || *opt) {
+                  VAST_DEBUG(this, "selects", part_id, "at predicate", x);
+                  result.push_back(part_id);
+                  break;
+                }
+              } else {
+                // The field has no dedicated synopsis. Check if there is one
+                // for the type in general.
+                auto it = part_syn.type_synopses_.find(field.type);
```
mavam, line 206: This assumes exact type equality. Two string types with different attributes are not equal wrt. `operator==`. You may want to strip attributes here.

Take a look a the `key` function in our `factory` implementations, where we do this as well.

mavam (commented):
Why do we need the class `string_synopsis`? Can't we just use a `buffered_bloom_filter_synopsis` or `buffered_bloom_filter_synopsis<T>` if we want to restrict it to a single type? I would argue the lookup function can have a precondition that it was already type-checked and that it can be fully polymorphic.

A related question: how does that `buffered_string_synopsis` behave differently from the `buffered_address_synopsis`?

vast.yaml.example
```diff
@@ -37,6 +37,8 @@ vast:
   max-taste-partitions: 5
   # The amount of queries that can be executed in parallel.
   max-queries: 10
+  # The false positive rate for lossy structures in the meta index.
+  meta-index-fprate: 0.01
```
mavam, line 41:
```suggestion
  meta-index-fp-rate: 0.01
```

With the other way of writing, it is hard to see the word boundaries.

dominiklohmann (approved):
We reviewed this together in a Slack call. Looks and works great!

libvast/src/system/index.cpp
```diff
@@ -531,6 +531,10 @@ index(index_actor::stateful_pointer<index_state> self,
       } else {
         VAST_ASSERT(active.capacity >= x.rows());
         active.capacity -= x.rows();
+        VAST_TRACE(self, "reduced active partition capacity to",
+                   (std::to_string(active.capacity) + '/'
+                    + std::to_string(self->state.partition_capacity)),
+                   "rows");
```
tobim, line 537: This message has been removed on master in the meantime but a rebase brought it back. Imho, the message has no value. Please drop that commit.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🎁 A new type-level synopsis structure in the meta-index now massively
+  speeds up string queries with very few results.
+  [#1214](https://github.com/tenzir/vast/pull/1214)
```
mavam, line 17: This is another case where I don't see the need for a "feature" changelog. It's a non-functional change.

It would only make sense to introduce a performance-related category where we document performance-affecting changes.

PR #1215, opened 2020-12-04
Adjust logging verbosity in exporter

lava:

Print basic information about the number of total and scheduled partitions on VERBOSE.

Don't print a message for every export on INFO log level (this is especially noticable when using a command that spawns exporters dynamically, like `vast explore` or `vast matcher`).

###  :memo: Checklist

### DIFF
--- a/libvast/src/system/exporter.cpp
+++ b/libvast/src/system/exporter.cpp
@@ -384,15 +384,15 @@ behavior exporter(stateful_actor<exporter_state>* self, expression expr,
           self->send(x, atom::exporter_v, self);
     },
     [=](atom::run) {
-      VAST_INFO(self, "executes query:", to_string(self->state.expr));
+      VAST_VERBOSE(self, "executes query:", to_string(self->state.expr));
       self->state.start = system_clock::now();
       if (!has_historical_option(self->state.options))
         return;
       self->request(self->state.index, infinite, self->state.expr)
         .then(
           [=](const uuid& lookup, uint32_t partitions, uint32_t scheduled) {
-            VAST_DEBUG(self, "got lookup handle", lookup, ", scheduled",
-                       scheduled, '/', partitions, "partitions");
+            VAST_VERBOSE(self, "got lookup handle", lookup, ", scheduled",
+                         scheduled, '/', partitions, "partitions");
             self->state.id = lookup;
             if (partitions > 0) {
               self->state.query.expected = partitions;


libvast/src/system/exporter.cpp
```diff
@@ -384,15 +384,15 @@ behavior exporter(stateful_actor<exporter_state>* self, expression expr,
...
       self->state.start = system_clock::now();
       if (!has_historical_option(self->state.options))
         return;
       self->request(self->state.index, infinite, self->state.expr)
         .then(
           [=](const uuid& lookup, uint32_t partitions, uint32_t scheduled) {
-            VAST_DEBUG(self, "got lookup handle", lookup, ", scheduled",
-                       scheduled, '/', partitions, "partitions");
+            VAST_VERBOSE(self, "got lookup handle", lookup, ", scheduled",
+                         scheduled, '/', partitions, "partitions");
```
mavam, line 395: Are related messages that print status updates also VERBOSE?

libvast/src/system/exporter.cpp
```diff
@@ -384,15 +384,15 @@ behavior exporter(stateful_actor<exporter_state>* self, expression expr,
           self->send(x, atom::exporter_v, self);
     },
     [=](atom::run) {
-      VAST_INFO(self, "executes query:", to_string(self->state.expr));
+      VAST_VERBOSE(self, "executes query:", to_string(self->state.expr));
```
dominiklohmann, line 387: I think we should print the `query_id` here and in all other log messages for the current query so we can actually know which query it is. Alternatively, we can change the name of the actor from `exporter` to `exporter-{query_id}`.

libvast/src/system/exporter.cpp
lava: Some of them are even `INFO` (see below), some are `DEBUG`. I think VERBOSE didn't really exist when most of this was written, so there's not a lot of consistency about this.

libvast/src/system/exporter.cpp
lava: We only get the query id in the `then()` part of the continuation, so we can't print it here.

Changing the actor name will break/be deprecated in CAF in the future afaik, the idea being is that an actor names should roughly correspond to a type name and always be constant.

dominiklohmann: Alright, then just ignore this.

dominiklohmann (approved):

PR #1216, opened 2020-12-08
Fix the bloom filter synopsis deserialization

tobim:
###  :memo: Checklist

### DIFF
--- a/libvast/vast/address_synopsis.hpp
+++ b/libvast/vast/address_synopsis.hpp
@@ -27,6 +27,21 @@
 
 namespace vast {
 
+namespace {
+
+// Because VAST deserializes a synopsis with empty options and
+// construction of an address synopsis fails without any sizing
+// information, we augment the type with the synopsis options.
+type annotate_type(type type, const bloom_filter_parameters& params) {
+  using namespace std::string_literals;
+  auto v = "bloomfilter("s + std::to_string(*params.n) + ','
+           + std::to_string(*params.p) + ')';
+  // Replaces any previously existing attributes.
+  return std::move(type).attributes({{"synopsis", std::move(v)}});
+}
+
+} // namespace
+
 template <class HashFunction>
 synopsis_ptr
 make_address_synopsis(vast::type type, bloom_filter_parameters params,
@@ -73,7 +88,7 @@ public:
     params.p = p_;
     params.n = next_power_of_two;
     VAST_DEBUG_ANON("shrinked address synopsis to", params.n, "elements");
-    auto& type = this->type();
+    auto type = annotate_type(this->type(), params);
     auto shrinked_synopsis
       = make_address_synopsis<xxhash64>(type, std::move(params));
     if (!shrinked_synopsis)
@@ -200,20 +215,15 @@ synopsis_ptr make_address_synopsis(vast::type type, const caf::settings& opts) {
   bloom_filter_parameters params;
   params.n = *max_part_size;
   params.p = defaults::system::address_synopsis_fprate;
-  // Because VAST deserializes a synopsis with empty options and
-  // construction of an address synopsis fails without any sizing
-  // information, we augment the type with the synopsis options.
-  using namespace std::string_literals;
-  auto v = "bloomfilter("s + std::to_string(*params.n) + ','
-           + std::to_string(*params.p) + ')';
-  auto t = type.attributes({{"synopsis", std::move(v)}});
+  auto annotated_type = annotate_type(type, params);
   // Create either a a buffered_address_synopsis or a plain address synopsis
   // depending on the callers preference.
   auto buffered = caf::get_or(opts, "buffer-ips", false);
   auto result
     = buffered
-        ? make_buffered_address_synopsis<HashFunction>(std::move(t), params)
-        : make_address_synopsis<HashFunction>(std::move(t), params);
+        ? make_buffered_address_synopsis<HashFunction>(std::move(type), params)
+        : make_address_synopsis<HashFunction>(std::move(annotated_type),
+                                              params);
   if (!result)
     VAST_ERROR_ANON(__func__,
                     "failed to evaluate Bloom filter parameters:", params.n,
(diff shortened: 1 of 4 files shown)


libvast/vast/address_synopsis.hpp
```diff
@@ -27,6 +27,21 @@
...
+// Because VAST deserializes a synopsis with empty options and
+// construction of an address synopsis fails without any sizing
+// information, we augment the type with the synopsis options.
+type annotate_type(type type, const bloom_filter_parameters& params) {
+  using namespace std::string_literals;
+  auto v = "bloomfilter("s + std::to_string(*params.n) + ','
+           + std::to_string(*params.p) + ')';
+  // Replaces any previously existing attributes.
+  return std::move(type).attributes({{"synopsis", std::move(v)}});
+}
```
mavam, line 41: Not in this PR, but need a better solution for this type tagging.

libvast/vast/bloom_filter.hpp
```diff
@@ -110,7 +110,11 @@ class bloom_filter : detail::equality_comparable<
 
   template <class Inspector>
   friend auto inspect(Inspector& f, bloom_filter& x) {
-    return f(caf::meta::type_name("bloom_filter"), x.hasher_, x.bits_);
+    auto result = f(caf::meta::type_name("bloom_filter"), x.hasher_, x.bits_);
+    // When deserializing into a vector that was already bigger than required,
+    // CAF will reuse the storage but not release the excess afterwards.
+    x.bits_.shrink_to_fit();
```
mavam, line 116: Would it make sense to put this mutation in the calling context, i.e., after deserializing? Since this code path is used for both serialization and deserialization, and potentially by other concepts as well, I'm not sure if it the right place.

libvast/vast/bloom_filter.hpp
```diff
@@ -110,7 +110,11 @@ class bloom_filter : detail::equality_comparable<
...
+    return result;
```
dominiklohmann, line 117: I think this should only be done when loading.

```suggestion
    auto& bits = x.bits_;
    auto load = [&]() -> caf::error {
      // When deserializing into a vector that was already bigger than required,
      // CAF will reuse the storage but not release the excess afterwards.
      bits_.shrink_to_fit();
      return caf::none;
    };
    auto result = f(caf::meta::type_name("bloom_filter"), x.hasher_, bits, load);
    return result;
```

And add `#include <caf/meta/load_callback.hpp>` at the top.

libvast/vast/bloom_filter.hpp
tobim: @dominiklohmann 's suggestion below is the best option here.

libvast/vast/bloom_filter.hpp
tobim: Right idea, but the suggested code looks wrong.

libvast/vast/bloom_filter.hpp
mavam: Okay.

libvast/vast/bloom_filter.hpp
mavam: Sounds like we want to if-constexpr on `Inspector::writes_state`?

libvast/vast/bloom_filter.hpp
```diff
@@ -110,11 +111,14 @@ class bloom_filter : detail::equality_comparable<
...
-    x.bits_.shrink_to_fit();
-    return result;
+    return f(caf::meta::type_name("bloom_filter"), x.hasher_, x.bits_,
+             caf::meta::load_callback([&]() -> caf::error {
+               // When deserializing into a vector that was already bigger than
+               // required, CAF will reuse the storage but not release the
+               // excess afterwards.
+               x.bits_.shrink_to_fit();
+               return caf::none;
+             }));
```
mavam, line 121: Mind splitting this into a more readable form of code? This one-liner is quite bulky to process cognitively.

e.g.,

```cpp
auto load_callback = ...;
return f(..., load_callback);
```

libvast/vast/bloom_filter.hpp
dominiklohmann: Ah, the suggested code missed wrapping the lambda in `caf::meta::load_callback`. 

@mavam That's how the load_callback is implemented.

libvast/vast/address_synopsis.hpp
lava: Indeed, I think we say this in every PR where we touch this code ;)

libvast/vast/address_synopsis.hpp
```diff
@@ -73,7 +88,7 @@ class buffered_address_synopsis : public synopsis {
     params.p = p_;
     params.n = next_power_of_two;
     VAST_DEBUG_ANON("shrinked address synopsis to", params.n, "elements");
-    auto& type = this->type();
+    auto type = detail::annotate_type(this->type(), params);
```
dominiklohmann, line 91: Please rename to `annotated_type`.

libvast/vast/address_synopsis.hpp
```diff
@@ -73,7 +88,7 @@ class buffered_address_synopsis : public synopsis {
     params.p = p_;
     params.n = next_power_of_two;
     VAST_DEBUG_ANON("shrinked address synopsis to", params.n, "elements");
```
dominiklohmann, line 90: shrinked -> shrunk generally, and here to shrinks because it has yet to happen.

dominiklohmann: Everywhere please. :-)

dominiklohmann (approved):
This looks alright to me, modulo a few wording issues, and please factor out the Nix changes into a separate PR. I did not review them.

lava:
The CI on Mac fails with a known issue: https://app.clubhouse.io/tenzir/story/20882/race-condition-in-node-queries-unit-test

As discussed offline, I'm merging this now and will include the final style changes in a follow-up PR.

PR #1220, opened 2020-12-09
Measure time spent in archive and index separately in exporter

dominiklohmann:

This PR measures the time spent waiting per request to index and archive and prints the results to DEBUG after the exporter shuts down.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 The index no longer crashes when too many parallel queries are running.
+  [#1210](https://github.com/tenzir/vast/pull/1210)
+
 - 🎁 On Linux, VAST now contains a set of built-in USDT tracepoints that can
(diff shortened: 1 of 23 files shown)


tobim (changes requested):
To me, the size of this change is a strong indicator that the communication protocol between `index` and `exporter` is neither scalable nor maintainable. Almost all of the effort is spent to implement request tracking through the index to collect some data that is very hard to interpret correctly.

The original need to add more query instrumentation came from the desire to get an answer to "Which component was responsible for the majority of the runtime duration of a call to `vast export`?"
The only measurements that we actually need for that is the (mean, max and overall) amount of time that passes between subsequent requests to the index. The time spent in the archive doesn't even need to be considered, because the absolute query time is already available.

I apologize for not making this clear at the start.

dominiklohmann:
Your change request does not contain a request for changes @tobim. Please clarify.

tobim:
To clarify: This PR is blocked until the disagreement about the general approach is resolved.

dominiklohmann:
Closing as we plan to pursue this in a different way in the future.

PR #1224, opened 2020-12-10
Improve tracing tools support in the static binary

tobim:

This enables jemalloc profiling and changes our library dependencies to build with `-fno-omit-frame-pointer`.

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -35,6 +35,9 @@ in {
     NIX_CFLAGS_COMPILE = "-std=c++17";
     dontStrip = true;
   });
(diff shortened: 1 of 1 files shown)


dominiklohmann:
@tobim can you add review instructions here?

lava:
@dominiklohmann The best way to review will be to download the static binary for this PR (or build it yourself), and use `perf record`/`gdb` or other debug tools of your choice to capture some stack traces, and verify that you get reasonable-looking stack traces including libc, arrow and jemalloc functions. (i also wanted to do this, but didn't get to it yet)

lava (approved):
Approving this since we already tested in production with Andi

PR #1211, opened 2020-12-04
Use typed actors for the index and related actors

dominiklohmann:
###  :notebook_with_decorative_cover: Description

To quote @lava on this matter:

> I think by now we have reached the point where we would have saved more time by making the index a typed actor than we spent debugging problems caused by not doing it.

Let's actually do this.

###  :memo: Checklist

- [x] All user-facing changes have changelog entries.
- [x] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [x] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

File-by-file feels best. A lot of this is cleanup, and going commit-by-commit is just too much content. Ideally we do this via screen sharing so I can explain while we're going over it.

PR #1225, opened 2020-12-10
Fix index worker depletion

tobim:

A bug that causes the removal of index workers in case of false positive candidate partitions is fixed here.

###  :memo: Checklist

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -189,13 +189,14 @@ bool index_state::worker_available() {
   return !idle_workers.empty();
 }
 
-query_supervisor_actor index_state::next_worker() {
-  VAST_ASSERT(worker_available());
-  auto result = std::move(idle_workers.back());
-  idle_workers.pop_back();
-  if (!worker_available())
+std::optional<query_supervisor_actor> index_state::next_worker() {
+  if (!worker_available()) {
     VAST_VERBOSE(self, "waits for query supervisors to become available to "
                        "delegate work; consider increasing 'vast.max-queries'");
+    return std::nullopt;
+  }
+  auto result = std::move(idle_workers.back());
+  idle_workers.pop_back();
   return result;
 }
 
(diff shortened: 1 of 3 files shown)


dominiklohmann (approved):
This needs a changelog entry, but other than that you're good to go. I actually noticed this while working on #1211 as well—and please let this be the last PR we merge before that one.

(Of if you want to do me a favor, wait until we merged #1211)

tobim:
@dominiklohmann I added a second change that I thought makes sense. Please have another look.

libvast/src/system/index.cpp
```diff
@@ -684,46 +683,49 @@ index(index_actor::stateful_pointer<index_state> self,
...
       // Send an evaluate atom to all the actors and collect the returned
       // evaluation triples in a `pending_query_map`, then run the continuation
       // below in the same actor context.
-      auto worker = self->state.next_worker();
       await_evaluation_maps(
         self, iter->second.expression, actors,
         [=](caf::expected<pending_query_map> maybe_pqm) {
           auto& st = self->state;
+          auto drop = [&] {
+            self->state.idle_workers.emplace_back(std::move(*worker));
```
dominiklohmann, line 700: I think we should capture `worker = std::move(*worker)` in the lambda that we pass to `await_evaluation_maps`. This code does not need to know about the `optional` wrapping.

libvast/src/system/index.cpp
```diff
@@ -684,46 +683,49 @@ index(index_actor::stateful_pointer<index_state> self,
         self->send(client, atom::done_v);
         return {};
       }
+      auto worker = self->state.next_worker();
+      if (!worker)
+        return caf::skip;
```
dominiklohmann, line 688: This is great, but please move it back to the very top of the function.

dominiklohmann (approved):
I really like this additional change, just two nits. I did not verify this to work locally.

libvast/src/system/index.cpp
dominiklohmann: Actually, now that I think about it: Don't do that.

PR #1226, opened 2020-12-11
Sanity check in idle handlers of index/archive

dominiklohmann:

This adds idle handlers to the index and archive that sanity-check whether their work queues are empty, and clears them if they are not empty although they should be.

Consider this an experimental change; I am not sure if we want to merge this, but I experimented with it so I might as well put it up as a PR.

###  :memo: Checklist

n/t

### DIFF
--- a/libvast/src/system/archive.cpp
+++ b/libvast/src/system/archive.cpp
@@ -25,6 +25,7 @@
 #include "vast/system/report.hpp"
 #include "vast/table_slice.hpp"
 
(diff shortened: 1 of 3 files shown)


dominiklohmann:
Interestingly, this causes the unit test suite to hang indefinitely. Looks like something isn't quite working in there.

dominiklohmann:
I'll close this PR as I don't plan to follow-up on this currently.

PR #1227, opened 2020-12-11
Add tracing for meta index lookup speed

lava:

Add debug log message and tracepoint for monitoring meta index lookups.

###  :memo: Checklist

### DIFF
--- a/libvast/vast/detail/tracepoint.hpp
+++ b/libvast/vast/detail/tracepoint.hpp
@@ -27,7 +27,7 @@
 //
 // Notable changes from the upstream version:
 //  * Add a written overview to the top of the file.
-//  * Add `VAST_TRACEPOINT_USDT()` convenience macro.
+//  * Add `VAST_TRACEPOINT()` convenience macro.
 //  * Use `VAST_DISABLE_SDT` macro instead of `FOLLY_DISABLE_SDT` to disable
 //    this feature.
 //  * Add pragmas to ignore warnings for GNU extensions when using clang.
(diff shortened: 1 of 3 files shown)


libvast/vast/detail/tracepoint.hpp
```diff
@@ -36,18 +36,18 @@
 
 // # Overview
 //
-// A USDT (userspace software-defined tracepoint/userspace dynamic tracepoint)
-// is a code instrumentation mechanism provided by the linux kernel to allow
-// tracing software to measure and account specific developer-defined events in
-// user space code and libraries.
+// A USDT (userspace statically-defined tracepoint) is a code instrumentation
+// mechanism provided by the linux kernel to allow tracing software to measure
```
tobim, line 40: I think the mechanism originates from DTrace on Solaris. Not sure if the Folly implementation works on other OSes though.

lava: We don't run on Solaris, so for VAST the linux kernel will provide the feature ;) (also the details are different, with dtrace you had a separate `tracepoints.d` file defining your tracepoints and dtrace itself would need to be integrated into the build pipeline generate an object file out of that which can be linked into the final binary, whereas here everything is directly generated inline. Or so they say in the manual, not that I've ever used dtrace personally.)

I've changed the paragraph a bit to include the historical context.

dominiklohmann (approved):
Please add a changelog entry. Other than that this is ready to be merged.

PR #1229, opened 2020-12-14
Track individual sources at the importer

tobim:

Adds the following INFO messages to the node:
```
2020-12-13T23:17:05.668 importer adds zeek-reader source
2020-12-13T23:17:07.384 importer removes zeek-reader source
```

###  :memo: Checklist

By commit.

### DIFF
--- a/libvast/src/system/importer.cpp
+++ b/libvast/src/system/importer.cpp
@@ -90,6 +90,7 @@ public:
   void register_input_path(caf::inbound_path* ptr) override {
     auto& st = driver_.self()->state;
     st.inbound_descriptions[ptr] = std::move(st.inbound_description);
+    st.inbound_description = "anonymous";
     VAST_INFO_ANON("importer adds", st.inbound_descriptions[ptr], "source");
     super::register_input_path(ptr);
   }
@@ -256,6 +257,11 @@ caf::behavior importer(importer_actor* self, path dir, archive_actor archive,
     [=](atom::exporter, const caf::actor& exporter) {
       VAST_DEBUG(self, "registers exporter", exporter);
       return self->state.stg->add_outbound_path(exporter);
+    },
+    [=](caf::stream<importer_state::input_type>& in) {
+      auto& st = self->state;
+      VAST_DEBUG(self, "adds a new source:", self->current_sender());
+      st.stg->add_inbound_path(in);
     },
     [=](caf::stream<importer_state::input_type>& in, std::string desc) {
       auto& st = self->state;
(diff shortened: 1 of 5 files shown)


libvast/src/system/importer.cpp
```diff
@@ -38,6 +38,79 @@
 
 namespace vast::system {
 
+namespace {
+
+class driver : public importer_state::driver_base {
+public:
+  using super = importer_state::driver_base;
+
+  using pointer = importer_actor*;
```
tobim, line 47: Replace by `importer_actor::pointer`.

libvast/src/system/importer.cpp
```diff
@@ -38,6 +38,79 @@
...
+
+  pointer self() const {
+    return self_;
+  }
+
+private:
+  pointer self_;
+};
+
+class manager : public caf::detail::stream_stage_impl<driver> {
```
tobim, line 82: s/manager/stream_stage/g

libvast/src/system/importer.cpp
```diff
@@ -38,6 +38,79 @@
...
+public:
+  using super = caf::detail::stream_stage_impl<driver>;
+
+  manager(importer_actor* self) : caf::stream_manager(self), super(self, self) {
```
tobim, line 86: Add a comment explaining the arguments to `super`. Don't init the `stream_manager`.

libvast/src/system/importer.cpp
```diff
@@ -38,6 +38,79 @@
...
+    uint64_t events = 0;
+    auto t = timer::start(st.measurement_);
+    for (auto& x : xs) {
+      VAST_ASSERT(x.rows() <= static_cast<size_t>(st.available_ids()));
+      auto rows = x.rows();
+      events += rows;
+      x.offset(st.next_id(rows));
+      out.push(std::move(x));
+    }
+    t.stop(events);
```
dominiklohmann, line 67: This PR should not change the measurement logic. The old code ran the timer per table slice, this now runs it once per vector of table slices.

It also doesn't hurt to name the input and output type here explicitly, I think, and to remove them from the `importer_state`, where they shouldn't have been in for a while now.

dominiklohmann (changes requested):

libvast/src/system/importer.cpp
tobim: I don't understand, the duration outside of the loop should be the same as the sum of the durations of the loop body?

And you mean the `using` decls? Shure that can be cleaned up.

dominiklohmann (approved):
I've tested this locally, and implemented the review suggestions myself that originated from a Slack call where we looked at this code together.

PR #1233, opened 2020-12-15
Enhance dump command to show models

mavam:

This PR enhances the `dump` command to show models in addition to concepts.

###  :memo: Checklist

- [ ] Address FIXMEs (empty-string `description` when dumping)

@dominiklohmann:
  - [x] All command-related C++ file-by-file

@tobim:
  - [x] Taxonomy map storage change
  - [x] Integration test suite changes


libvast/src/system/application.cpp
```diff
@@ -134,6 +134,9 @@ auto make_dump_command() {
   dump->add_subcommand("concepts", "print all registered concept definitions",
                        documentation::vast_dump_concepts,
                        opts("?vast.dump.concepts"));
+  dump->add_subcommand("models", "print all registered model definitions",
+                       documentation::vast_dump_concepts,
+                       opts("?vast.dump.concepts"));
```
dominiklohmann, line 139:
```suggestion
  dump->add_subcommand("models", "print all registered model definitions",
                       documentation::vast_dump_models,
                       opts("?vast.dump.models"));
```

And then the command docs are also missing.

dominiklohmann (changes requested):

mavam:
As the guardian of the integration test suite, @tobim may want to have a look at https://github.com/tenzir/vast/pull/1233/commits/89152f1485f063714f9cf3089e3b2d09498fdf4e and https://github.com/tenzir/vast/pull/1233/commits/9ecbfc889402ce0b9bd2af31d3e86cfa38ae081e.

dominiklohmann (approved):
Code-wise this is fine.

Can you change the documentation of `vast dump` itself so it reflects that it dumps both models and concepts?

Please wait for @tobim to review the integration test suite changes. I did not look at them.

mavam:
I'm handing this PR over to @tobim, who is going to investigate the non-determinism during concept dumping.

mavam:
Taking the PR back after internal sync.

integration/integration.py
```diff
@@ -192,15 +192,9 @@ def empty(iterable):
...
+def is_non_deterministic(command):
     positionals = list(filter(lambda x: x[0] != "-", command))
-    if positionals[0] == "export":
-        if positionals[1] == "json":
-            # Pythons JSON parser does not handle ldjson, so we fall back to ascii
-            return "ascii"
-    if positionals[0] == "status":
-        return "json"
-    return "ascii"
+    return bool(re.match(r"exp(ort|lore)|get|pivot", positionals[0]))
```
tobim, line 197:
```suggestion
    return positionals[0] in {"export", "explore", "get", "pivot"}
```

integration/integration.py
```diff
@@ -11,6 +11,7 @@
 import json
 import logging
 import os
+import re
```
tobim, line 14:
```suggestion
```

tobim (changes requested):
Just a suggestions to simplify the python code.

mavam:
> You only take a const std::regex& here, so you should get away with forward-declaring it.

Is there a header for this, like we have for `<iosfwd>`?

> `<regex>` is one of the most expensive headers to include, and I'm not sure why we need it here. 

It makes for a convenient way to control the size of the return value.

>The default value is also crazy expensive. std::regex should never be created on the fly.

We're doing I/O here. And it's happening pretty much once at startup. The performance concerns don't appear to be in relation to the functionality.

dominiklohmann (approved):
This works fine.

Not printing empty descriptions / nested concepts / fields is a strict follow up and should not block the imminent release.

The `<regex>` stuff I'll clean up myself after the release, I created a story for that.

PR #1241, opened 2020-12-16
Fix the -t option in the integration testing tool

tobim:


mavam (changes requested):
It just made regex specification work, so that this works the same as the unit test integration suite and `-t PATTERN` is valid. The PR doesn't contain any comments for why the functionality is (a) demoted to substring search, and (b) divergent to our unit test suite.

mavam (approved):

mavam:
For the record: this "fix" was an oversight.

PR #1243, opened 2020-12-17
Fix installations with relative prefixes

dominiklohmann:

This prohibits `CMAKE_INSTALL_PREFIX` to be a relative path, and changes `configure` to convert arguments to `--prefix` to be absolute.

###  :memo: Checklist

n/t

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,7 +82,7 @@ if (POLICY CMP0042)
   if (APPLE AND NOT DEFINED CMAKE_MACOSX_RPATH)
     set(CMAKE_SKIP_BUILD_RPATH FALSE)
     set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
-    if (NOT VAST_RELOCATABLE_INSTALL)
+    if (NOT VAST_ENABLE_RELOCATABLE_INSTALLATIONS)
       set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
     endif ()
   endif ()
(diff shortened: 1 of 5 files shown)


libvast_test/CMakeLists.txt
```diff
@@ -26,23 +26,24 @@ target_sources(
   libvast_test
   INTERFACE
     "$<${isExe}:$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/main.cpp>>"
-    "$<${isExe}:$<INSTALL_INTERFACE:$<INSTALL_PREFIX>/${CMAKE_INSTALL_DATADIR}/vast/test/main.cpp>>"
+    "$<${isExe}:$<INSTALL_INTERFACE:${_IMPORT_PREFIX}/${CMAKE_INSTALL_DATADIR}/vast/test/main.cpp>>"
```
tobim, line 29: The old code here was correct, see https://cmake.org/cmake/help/v3.19/manual/cmake-generator-expressions.7.html#target-dependent-queries.

Please revert.

tobim (changes requested):

CMakeLists.txt
```diff
@@ -264,19 +264,16 @@ if (VAST_ENABLE_STATIC_EXECUTABLE AND BUILD_SHARED_LIBS)
 endif ()
 
 if (VAST_ENABLE_RELOCATABLE_INSTALLATIONS AND NOT VAST_ENABLE_STATIC_EXECUTABLE)
+  # Note that we cannot use the target property INSTALL_RPATH on
+  # libvast_internal, because it is an interface library. We should probably fix
+  # this by making the config header part of libvast_internal so it is not just
```
tobim, line 269: Neat idea, but header only libs are still INTERFACE, PUBLIC requires at least one TU.

tobim (approved):

CMakeLists.txt
dominiklohmann: We can add a stub TU then. :)

PR #1246, opened 2020-12-17
Add simdjson dependency to libvast

ngrodzitski:

Restore PR with rebased epic.

###  :memo: Checklist


dominiklohmann:
Instead of re-opening, feel free to rebase and force-push next time. We don't mind force pushes before reviews within a PR.

ngrodzitski:
> Instead of re-opening, feel free to rebase and force-push next time. We don't mind force pushes before reviews within a PR.

Ok, I get it.

nix/overlay.nix
```diff
@@ -55,6 +55,11 @@ in {
     configureFlags = old.configureFlags ++ [ "--enable-prof" "--enable-stats" ];
   });
   broker = final.callPackage ./broker {inherit stdenv; python = final.python3;};
+  simdjson = prev.simdjson.overrideAttrs (old: {
```
ngrodzitski, line 58: There are plans for static-lib as the default option for build, but currently simdjson uses shared-lib. So a little adjustment is required.

https://github.com/simdjson/simdjson/issues/1199

ngrodzitski:
@dominiklohmann, @tobim Please, review the changes to nix* I've made.

nix/overlay.nix
```diff
@@ -55,6 +55,11 @@ in {
...
+    cmakeFlags = old.cmakeFlags ++ [
```
tobim, line 59: From the top of my head:
```suggestion
    cmakeFlags = old.cmakeFlags ++ lib.optionals isStatic [
```

tobim:
Not bad figuring the nix stuff out so fast. It's not particularly newcomer-friendly.

mavam:
You almost make that sound like a "feature", @tobim. 😉

dominiklohmann:
Nice to see you getting somewhere—I like how this is looking so far. 👍

One small heads up: The `suricata` selector will receive a slight change with #1259. This shouldn't have any impact on your code at all, but there may be some merge conflicts. I recommend merging master back into your branch after shortly we've merged #1259 to keep track with master.

libvast/vast/format/simdjson.hpp
```diff
@@ -0,0 +1,302 @@
...
+ * http://vast.io/license. No part of VAST, including this file, may be       *
+ * copied, modified, propagated, or distributed except according to the terms *
+ * contained in the LICENSE file.                                             *
+ ******************************************************************************/
+
+#pragma once
+
+#include "vast/concept/hashable/hash_append.hpp"
+#include "vast/concept/hashable/xxhash.hpp"
+// #include "vast/concept/parseable/vast/json.hpp" // TO DELETE?
```
dominiklohmann, line 18: In the very end, yes.

libvast/vast/format/simdjson.hpp
```diff
@@ -0,0 +1,302 @@
...
+#include "vast/fwd.hpp"
+
+#include "vast/defaults.hpp"
+#include "vast/detail/flat_map.hpp"
+#include "vast/detail/line_range.hpp"
+#include "vast/detail/string.hpp"
+#include "vast/error.hpp"
+#include "vast/format/multi_layout_reader.hpp"
+#include "vast/format/ostream_writer.hpp"
+#include "vast/json.hpp" // TO DELETE?
```
dominiklohmann, line 28: Deleting this should be the very last step.

libvast/vast/format/simdjson.hpp
```diff
@@ -0,0 +1,302 @@
...
+//   using defaults = vast::defaults::export_::json;
+
+//   using super = ostream_writer;
+
+//   using super::super;
+
+//   caf::error write(const table_slice& x) override;
+
+//   const char* name() const override;
+// };
```
dominiklohmann, line 53: This you can just remove, as simdjson only concerns the reader.

libvast/vast/system/import_command.hpp
```diff
@@ -120,4 +120,14 @@ caf::message import_command(const invocation& inv, caf::actor_system& sys) {
   return caf::none;
 }
 
+template <class Reader, class SimdjsonReader, class Defaults>
+caf::message
+import_command_json(const invocation& inv, caf::actor_system& sys) {
+  auto use_simdjson = caf::get_if<bool>(
+    &inv.options, Defaults::category + std::string{".simdjson"});
+  if (use_simdjson)
```
dominiklohmann, line 128:
```suggestion
  auto use_simdjson = caf::get_or(
    &inv.options, Defaults::category + std::string{".simdjson"}, false);
  if (use_simdjson)
```

libvast/src/format/simdjson.cpp
```diff
@@ -0,0 +1,449 @@
...
+    else
+      return elem;
+  }
+  return xs;
+}
+
+// MAP Conversions
+template <>
+caf::expected<data>
+type_biassed_convert_impl<::simdjson::dom::object, type_id<map_type>()>(
```
dominiklohmann, line 280:
```suggestion
type_biased_convert_impl<::simdjson::dom::object, type_id<map_type>()>(
```

Just a nit: It's `biased`, not `biassed`.

libvast/src/format/simdjson.cpp
```diff
@@ -0,0 +1,449 @@
...
+type_biassed_convert_impl<integer, type_id<integer_type>()>(integer n,
+                                                            const type&) {
+  return n;
+}
+
+template <>
+caf::expected<data>
+type_biassed_convert_impl<std::string_view, type_id<integer_type>()>(
+  std::string_view s, const type& ) {
+  // simdjson cannot be reused here as it doesn't accept hex numbers.
```
dominiklohmann, line 123: That's okay, the hex number parsers are separate from the `vast::json` class.

libvast/src/format/simdjson.cpp
```diff
@@ -0,0 +1,449 @@
...
+
+template <>
+caf::expected<data>
+type_biassed_convert_impl<count, type_id<real_type>()>(count n, const type&) {
+  return detail::narrow_cast<real>(n);
+}
+
+template <>
+caf::expected<data>
+type_biassed_convert_impl<std::string_view, type_id<real_type>()>(
```
dominiklohmann, line 183: One thing I wondered while reading this is whether this wouldn't be much more readable with one call to `caf::visit` on the type, instead of using the `type_id` to dispatch manually.

libvast/src/format/simdjson.cpp
```diff
@@ -0,0 +1,449 @@
...
+      return make_error(ec::parse_error, "unable to parse",
+                        caf::detail::pretty_type_name(typeid(value_type)), ":",
+                        std::string{j});
+
+    return x;
+  } else {
+    static_cast<void>(j);
+    // ngrodzitski: Once fmt is available message can be improved.
+    // as simdjson has a `ostream<<element_type` which makes it
+    // "serializable" to fmt.
```
dominiklohmann, line 84: If you include `vast/concept/printable/stream.hpp` you can have it already.

dominiklohmann (commented):
This is looking quite good already. I added a few comments. I like where you're going with this.

libvast/vast/format/simdjson.hpp
ngrodzitski: Ok

libvast/src/format/simdjson.cpp
ngrodzitski: Fixed.

libvast/src/format/simdjson.cpp
ngrodzitski: Yes, that is an interesting part, I'll elaborate on this tomorrow, so we can discuss possible solutions.

libvast/test/format/simdjson.cpp
```diff
@@ -0,0 +1,177 @@
...
+  CHECK(slice.at(0, 1) == data{count{424242}});
+
+  CHECK(slice.at(0, 2).is<real>());
+  const auto r = slice.at(0, 2).get_data().get(
+    std::integral_constant<
+      int, caf::detail::tl_index_of<data::types, real>::value>());
+  CHECK(std::abs(r - real{4.2}) < 0.000001);
+
+  CHECK(slice.at(0, 3) == data{integer{-1337}});
+  CHECK_EQUAL(slice.at(0, 4),data{std::string{"0123456789\r\n"}});
```
ngrodzitski, line 122: There is a mismatch with previous json behavior. Simdjson handles \r\n and provides data in non-escaped form, a "\r\n" json-string (size 4 characters) becomes a string of size 2 characters ( 0x0D, ox0A). And that breaks some integrations tests, and maybe similar mismatches causes other tests to fail (currently figuring out).

libvast/src/format/simdjson.cpp
ngrodzitski: Greate, thnx

libvast/src/format/simdjson.cpp
ngrodzitski: I've added implementation details comment https://github.com/tenzir/vast/pull/1246/files#diff-f82ad10248f33876b8ac75b84449eb8d7da03d2d2cf32a9e7683ff3b3c63b938R39-R73

Using `caf::visit` (**as I got it**) would require to put simdjson values (`dom::array`, `dom::object`, `int64_t`, ...) into `caf::variant`, which by itself is already something like switch logic and that would be just to conform the visitor requirements which then would dispatch it once again.

dominiklohmann: You are absolutely correct, i overlooked this. Thanks for the detailed comment!

libvast/test/format/simdjson.cpp
dominiklohmann: I think that's okay. Don't spend too much time on this.

ngrodzitski: Well, it has another catch, export does no escaping and a single record in export results into 2 lines with string containing it  (node-suricata-alert)

ngrodzitski: No escaping for csv-export. Ascii-export is ok

libvast/src/format/simdjson.cpp
mavam: CAF has the notion of a *sum type* that any type can fulfil to opt-in visitation via `caf::visit`. We do that with `vast::type`.

I haven't followed 100% the details, but in case the `dom` type comes with a discriminator to dispatch on, we only have to write the adapter.

libvast/test/format/simdjson.cpp
dominiklohmann: That's by design I think (cc @tenzir/backend), csv should not do escaping of newlines.

mavam: > csv should not do escaping of newlines

The newline is a control character in CSV that gets you to the next record. So when it comes up in strings, it should be escaped. (I'm not sure if that's what you mean.)

dominiklohmann: If you're certain on that front, then we do have a bug here. Let's tackle it independently from this PR, though.

mavam: I think the naive assumption is that one line is one CSV record. However, everybody does CSV differently. The [Wikipedia page](https://en.wikipedia.org/wiki/Comma-separated_values#General_functionality) sums it up well:

>  Simple CSV implementations may prohibit field values that contain a comma or other special characters such as newlines. More sophisticated CSV implementations permit them, often by requiring " (double quote) characters around values that contain reserved characters (such as commas, double quotes, or less commonly, newlines). Embedded double quote characters may then be represented by a pair of consecutive double quotes,[15] or by prefixing a double quote with an escape character such as a backslash (for example in Sybase Central).

So I wouldn't be surprised if some CSV implementation out there does stateful stuff, such as allowing newlines after an opening double quote or so. I think we should try to do things simple and escape, so that we have one-record-is-one-line semantics. If we ever find a user who needs different semantics, we can still add an option to the command line to change that behavior, but let's start with KISS.

dominiklohmann (commented):
I've tested this out locally and it works great, it's really a lot faster already than the current parser in real world usage. Really great work so far!

Code-wise I really like the approach you've taken (and commented) with the conversion from `JsonType` to `VastType`. It certainly does its job well enough, and after having toyed with the code a I bit it is the way to go over implementing a visit-based approach.

Do you need any more input or want us to look at some things in-depth specifically, e.g., do you need further help with the test issues?

ngrodzitski:
@dominiklohmann Currently I'm stuck with "Node suricata alert simdjson" integration test which is compares a csv export results. Except this I'm going to fresh-look the code and that is it for phase 2.

dominiklohmann:
Sounds good. I'll take an in-depth look at the integration test issue tomorrow.

dominiklohmann:
@ngrodzitski I reviewed this at length today, and discussed with @mavam what we want to do with regards to the different escaping behaviors of the old and the new json readers. 

- We consider the current JSON import behavior to be broken with its escaping (it escaped \r\n to the string \r\n, which then got converted to \\r\\n during export).
- We consider the simdjson import behavior to work as expected (no escaping during import, \r\n gets converted to the string \x0D\x0A during export where necessary).
- We agreed that it was not worth the effort to fix the legacy JSON reader, because it will be removed anyways after this work is finished.
- We agreed that the integration test references files between `vast import json --simdjson` and `vast import json|suricata --simdjson` may differ.

- @mavam checked that Zeek does the same form of escaping for control characters (they use lowercase 0x0d over our 0x0D, but we can change that easily if need be).
- I verified that newlines in CSV strings are okay per the specification as long as the strings are quoted, which they are in this case.

To summarize, we can move forward with this PR. Please update the remaining reference files accordingly. You can run `sh build/vast/integration_Debug.sh -u` (where Debug is your build type) to update them all automatically.

Once the CI is green I would like to merge this into the epic branch, do a few fixups and write a changelog entry and merge the epic branch into master. Phase 3 and 4 can then happen in separate PRs that branch off master.

Side note: The performance looks to be greatly improved!

ngrodzitski:
> To summarize, we can move forward with this PR. Please update the remaining reference files accordingly. You can run sh build/vast/integration_Debug.sh -u (where Debug is your build type) to update them all automatically.

Ok, I will do. Nevertheless, I'd like to note that this new-lines tollerance in csv can lead to errors.
Consider the following case:
```
record1 f1 "value\nAAAAAA", f2
record2 f2 "value\nBBBBBB", f2
```

if `\n` becomes a newline, then the output will be the following:
```
record1 f1 "value
AAAAAA", f2
record1 f2 "value
BBBBBB", f2
```

To compare with reference file the output is sorted on line basis:
```
AAAAAA", f2
BBBBBB", f2
record1 f1 "value
record1 f2 "value
```

The catch is that the following wrong output (A with record2 and B with record1)
```
record1 f1 "value
BBBBBB", f2
record2 f2 "value
AAAAAA", f2
```

will be sorted and compared to reference without errors. That is unlikely to be a problem, but it shows that compare method for under such conditions is not reliable.

dominiklohmann:
> Ok, I will do. Nevertheless, I'd like to note that this new-lines tollerance in csv can lead to errors.

That's absolutely correct. We sort the lines in the integration test framework before comparing them, which cannot work with this.

I'll discuss adapting the CSV writer with the team in tomorrow's standup.

mavam:
> We sort the lines in the integration test framework before comparing them, which cannot work with this.

It's possible to deactivate the sorting, which could be an option if the output is guaranteed to be deterministic.

mavam:
> I verified that newlines in CSV strings are okay per the specification as long as the strings are quoted, which they are in this case.

Even though it's legal to have CR/LF unescaped, per [the spec](https://stackoverflow.com/a/566059), we could decide to escape newlines, because most CSV parsers are not that fancy. This would be an option to the `csv` command.

As of right now, I'd go with the simplest path forward.

dominiklohmann:
We'll fix the CSV issue separately from the simdjson-related PRs so you don't need to worry about it.

dominiklohmann:
[ch4981]

ngrodzitski:
I think the reason of `vast: error while loading shared libraries: libsimdjson.so.4: cannot open shared object file: No such file or directory` can be because of installing `libsimdjson-dev` for ubuntu in vast/.github/workflows/vast.yaml which I can't test.

dominiklohmann:
I'll merge this into the epic branch now and fix the remaining CI issue myself in the epic branch, it likely is a permission issue. No big deal.

PR #1247, opened 2020-12-18
Handle Arrow decoder errors gracefully

dominiklohmann:

This fixes a crash when encountering an issue with the on-disk format of serialized Arrow Record Batches or Arrow Schemas.

###  :memo: Checklist

Change random bytes from a segment and export contained table slices until you get the IOError from Arrow and notice that VAST doesn't crash anymore.

### DIFF
--- a/libvast/src/arrow_table_slice.cpp
+++ b/libvast/src/arrow_table_slice.cpp
@@ -613,14 +613,14 @@ template <class FlatBuffer>
 table_slice::size_type arrow_table_slice<FlatBuffer>::rows() const noexcept {
   if (auto&& batch = record_batch())
     return batch->num_rows();
-  return {};
+  return 0;
 }
 
 template <class FlatBuffer>
 table_slice::size_type arrow_table_slice<FlatBuffer>::columns() const noexcept {
   if (auto&& batch = record_batch())
     return batch->num_columns();
-  return {};
+  return 0;
 }
 
 // -- data access ------------------------------------------------------------
--- a/libvast/src/arrow_table_slice.cpp
+++ b/libvast/src/arrow_table_slice.cpp
@@ -95,9 +95,9 @@ private:
 // -- decoding of Arrow column arrays ------------------------------------------
 
 // Safe ourselves redundant boilerplate code for dispatching to the visitor.
-#define DECODE_TRY_DISPATCH(vast_type)                                         \
-  if (auto dt = caf::get_if<vast_type##_type>(&t))                             \
-  return f(arr, *dt)
+#  define DECODE_TRY_DISPATCH(vast_type)                                       \
+    if (auto dt = caf::get_if<vast_type##_type>(&t))                           \
+    return f(arr, *dt)
 
 template <class F>
 void decode(const type& t, const arrow::BooleanArray& arr, F& f) {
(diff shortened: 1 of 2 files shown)


libvast/src/arrow_table_slice.cpp
```diff
@@ -611,29 +611,37 @@ const record_type& arrow_table_slice<FlatBuffer>::layout() const noexcept {
...
+    return batch->num_rows();
+  return {};
 }
 
 template <class FlatBuffer>
 table_slice::size_type arrow_table_slice<FlatBuffer>::columns() const noexcept {
-  return record_batch()->num_columns();
+  if (auto&& batch = record_batch())
+    return batch->num_columns();
+  return {};
```
mavam, line 623:
```suggestion
  return 0;
```

I think we can assume that there exists an implicit conversion from 0 to the `size_type`. This is more idiomatic.

libvast/src/arrow_table_slice.cpp
```diff
@@ -611,29 +611,37 @@ const record_type& arrow_table_slice<FlatBuffer>::layout() const noexcept {
...
 void arrow_table_slice<FlatBuffer>::append_column_to_index(
   id offset, table_slice::size_type column, value_index& index) const {
-  auto f = index_applier{offset, index};
-  auto array = record_batch()->column(detail::narrow_cast<int>(column));
-  decode(layout().fields[column].type, *array, f);
+  if (auto&& batch = record_batch()) {
+    auto f = index_applier{offset, index};
+    auto array = batch->column(detail::narrow_cast<int>(column));
+    decode(layout().fields[column].type, *array, f);
+  }
```
mavam, line 635: I think we should print a warning when this fails.

libvast/src/arrow_table_slice.cpp
```diff
@@ -611,29 +611,37 @@ const record_type& arrow_table_slice<FlatBuffer>::layout() const noexcept {
 
 template <class FlatBuffer>
 table_slice::size_type arrow_table_slice<FlatBuffer>::rows() const noexcept {
-  return record_batch()->num_rows();
+  if (auto&& batch = record_batch())
+    return batch->num_rows();
+  return {};
```
mavam, line 616:
```suggestion
  return 0;
```

See below. 👇

mavam (changes requested):

libvast/src/arrow_table_slice.cpp
dominiklohmann: I'm not quite sure. We already print an error when the decoding fails above.

mavam: Okay, then disregard.

libvast/src/arrow_table_slice.cpp
dominiklohmann: I don't mind either way.

mavam:
How can this be tested?

dominiklohmann:
> How can this be tested?

Change random bytes from a segment and export contained table slices until you get the IOError from Arrow and notice that VAST doesn't crash anymore.

dominiklohmann:
Rebased onto master.

mavam:
Shouldn't we do extra checking in [L588](https://github.com/tenzir/vast/blob/78a12f6409c296220fa3a76632767e5c3015f372/libvast/src/arrow_table_slice.cpp#L588) and [L597](https://github.com/tenzir/vast/blob/78a12f6409c296220fa3a76632767e5c3015f372/libvast/src/arrow_table_slice.cpp#L597)?

mavam (approved):
Turns out the answer is "no."

PR #1248, opened 2020-12-18
Remove check whether config file is a regular file

dominiklohmann:

This removes the check whether a config file is a regular file.

This is unnecessary after 5b4b052, which I accidentally pushed to master directly. Sorry about that one. That commit made it so manually specified config files could be in the same directory as normal config files.

The changelog entry covers both changes.

###  :memo: Checklist

Look at this PR and the referenced commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,13 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 Manually specified configuration files may now reside in the default
+  location directories. Configuration files can now be symlinked.
+  [#1248](https://github.com/tenzir/vast/pull/1248)
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -12,9 +12,13 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 Manually specified configuration files may not reside in the default
+  location directories. Configuration files can now be symlinked.
+  [#1248](https://github.com/tenzir/vast/pull/1248)
```
lava, line 17: `s/not/now/`

lava (approved):
Looks good, except for the fact that the CHANGELOG says the opposite of what it should ;)

CHANGELOG.md
dominiklohmann: Nice catch, corrected.

PR #1249, opened 2020-12-18
Follow up for the CMake refactoring

tobim:

- A few name changes weren't applied globally
- `VAST_ENABLE_ASSERTIONS` was broken.
- Unconditionally enable `-Wundef`
- Move defines back to `config.hpp`
- Synchonize definition names between CMake and C++

###  :memo: Checklist

By commit.


mavam:
The new CMake summary prints this:

```
-- The following features have been enabled:

 * VAST_ENABLE_RELOCATABLE_INSTALLATIONS, enable relocatable installations.
 * VAST_ENABLE_SDT, generate USDT tracepoint instrumentation.
 * VAST_ENABLE_ASSERTIONS, enable assertions.
 * BUILD_SHARED_LIBS, build shared instead of static library.
 * VAST_ENABLE_DEVELOPER_MODE, enables build settings for a nicer development environment.
 * VAST_ENABLE_BUNDLED_SCHEMAS, install bundled schemas with VAST.
 * VAST_ENABLE_ARROW, build with Apache Arrow support.
 * VAST_ENABLE_PCAP, build with PCAP support.
 * VAST_ENABLE_OPENSSL, encrypt network communication.
 * VAST_ENABLE_UNIT_TESTS, build unit tests for libvast.
 * VAST_ENABLE_VAST, build the standalone vast binary.
```

Why does `BUILD_SHARED_LIBS` have no `VAST_` prefix?

libvast/CMakeLists.txt
```diff
@@ -168,105 +172,114 @@ if (NOT VAST_ENABLE_BUNDLED_CAF)
...
+          $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
+      set_target_properties(libcaf_openssl_${_linkage_suffix}
+                            PROPERTIES EXPORT_NAME openssl)
+      add_library(caf::openssl ALIAS libcaf_openssl_${_linkage_suffix})
+      install(
+        TARGETS libcaf_openssl_${_linkage_suffix}
+        EXPORT CAFTargets
+        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```
mavam, line 253: Why do we use `CMAKE_INSTALL_BINDIR` here and below `*FULL_BINDIR`?

libvast/vast/detail/narrow.hpp
```diff
@@ -30,13 +30,11 @@
 
 namespace vast::detail {
 
-#if !VAST_NO_EXCEPTIONS
 /// @relates narrow
 struct narrowing_error : std::runtime_error {
   using super = std::runtime_error;
   using super::super;
 };
-#endif // VAST_NO_EXCEPTIONS
```
mavam, line 39: Have you looked whether there are any `throw` statements guarded by this (or similar) defines?

CMakeLists.txt
```diff
@@ -82,6 +93,7 @@ if (POLICY CMP0042)
   if (APPLE AND NOT DEFINED CMAKE_MACOSX_RPATH)
     set(CMAKE_SKIP_BUILD_RPATH FALSE)
     set(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
+    # TODO: Verify that this is indeed correct.
```
mavam, line 96: Have you done so? 🙂

tobim:
@mavam that variable is defined by CMake itself, we just consider it important enough to include it in the summary.

mavam (approved):
The branch builds fine for me, and since CI is green, I think green-lighting this makes sense.

I do have a few questions though.

mavam:
Can we remove this "noise" in the CMake run?

```
-- CREATING CAF TARGETS
-- Found CAF
-- Found CAF: /usr/local/lib/libcaf_core.dylib;/usr/local/lib/libcaf_io.dylib;/usr/local/lib/libcaf_openssl.dylib (found version "0.17.6") found components: openssl
-- CREATING CAF TARGETS
```

The all-caps comments probably don't need to be in here, right?

mavam:
> that variable is defined by CMake itself, we just consider it important enough to include it in the summary.

Maybe move it to the top/bottom then? Otherwise it looks like a bug.

libvast/CMakeLists.txt
tobim: An oversight from the previous PR that was missed.

FYI: CMake prepends `CMAKE_INSTALL_PREFIX` if the given path is relative. `CMAKE_INSTALL_FULL_*DIR` is calculated in the exact same way, so technically it doesn't make a difference.

libvast/vast/detail/narrow.hpp
tobim: `git grep VAST_NO_EXCEPTIONS` comes up empty in this branch. There are a few `throw` statements in robin-map and function2, and one in the yaml wrapper in `data.cpp`,

libvast/CMakeLists.txt
mavam: Should we at least use a one consistent version?

mavam (commented):
After your latest commit, my only remaining concern would be the left-over TODO.

CMakeLists.txt
tobim: This needs to be done by someone that truely understands how the dynamic loading process on macOS works. I haven't been able to make to connect the dots between the various sources of information on this.

mavam: Fair enough. Given that macOS is predominantly a developer platform, we can move on.

PR #1251, opened 2020-12-23
Fix installation scaffold

tobim:

Another round of fixes for the new CMake scaffold.
- Find Modules must be provided for PUBLIC and INTERFACE targets even when BUILD_SHARED_LIBS is on
- Dependencies are not propagated correctly for into `VASTConfig.cmake`
- The integration testing harness can now be used when linking against build trees

### DIFF
--- a/cmake/VASTProvideFindModule.cmake
+++ b/cmake/VASTProvideFindModule.cmake
@@ -1,7 +1,7 @@
 # Find modules are needed by the consumer in case of a static build, or if the
 # linkage is PUBLIC or INTERFACE.
 macro (provide_find_module name)
-  message(STATUS "Providing cmake module for ${name}")
+  message(VERBOSE "Providing cmake module for ${name}")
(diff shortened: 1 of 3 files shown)


cmake/VASTProvideFindModule.cmake
```diff
@@ -1,12 +1,10 @@
...
 macro (provide_find_module name)
-  if (NOT BUILD_SHARED_LIBS)
-    message(STATUS "Providing cmake module for ${name}")
-    configure_file("${PROJECT_SOURCE_DIR}/cmake/Find${name}.cmake" ${CMAKE_BINARY_DIR} COPYONLY)
-    install(
-      FILES "${CMAKE_BINARY_DIR}/Find${name}.cmake"
-      DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}/cmake/vast"
-      COMPONENT dev)
-  endif ()
+  message(STATUS "Providing cmake module for ${name}")
```
mavam, line 4:
```suggestion
  message(VERBOSE "Providing cmake module for ${name}")
```

mavam (approved):
I don't know how to test the integration harness, but the branch configures and compiles locally on macOS with system Clang.

PR #1252, opened 2020-12-23
Fix the package target

tobim:

CPack omits files and directories that are installed with absolute paths at the `DESTINATION` argument. That means we need to use
``` cmake
install(
  TARGETS libvast
  EXPORT VASTTargets
  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```
instead of
``` cmake
install(
  TARGETS libvast
  EXPORT VASTTargets
  ARCHIVE DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR}
  LIBRARY DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR}
  RUNTIME DESTINATION ${CMAKE_INSTALL_FULL_BINDIR})
```


mavam (approved):
Amazing where your find all this stuff.

tobim:
Let's hope this was the last one.

PR #1230, opened 2020-12-14
Integrate simdjson with VAST

dominiklohmann:
###  :notebook_with_decorative_cover: Description

This PR serves as a merge for base @ngrodzitski's work on integrating simdjson with VAST.

All CI changes have to take place in this PR, because GitHub does not use Workflow changes from forks.

###  :memo: Checklist

- [x] All user-facing changes have changelog entries.
- [x] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [x] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

n/t

PR #701, opened 2020-01-07
Fix Docker Build

0ortmann:
- Fix Docker build
- Update instructions in `INSTALLATION.md`
- Update `./scripts/docker-build`

PR #1257, opened 2021-01-04
Preserve nesting for exporting events in JSON format

tobim:

What the title says. Layout types now keep nested records as-is. The JSON writer reconstructs the layers while printing.

###  :memo: Checklist

- [x] Fix taxonomy integration test.

By commit, but do a quick first pass over the list before reviewing properly, several issues are fixed in later commits.


libvast/src/format/csv.cpp
```diff
@@ -117,16 +117,15 @@ caf::error writer::write(const table_slice& x) {
     append('\n');
     write_buf();
   }
+  auto flat_layout = flatten(layout);
   // Print the cell contents.
   auto iter = std::back_inserter(buf_);
   for (size_t row = 0; row < x.rows(); ++row) {
     append(last_layout_);
-    append(separator);
```
mavam, line 120: Looks like this statement got lost?

libvast/src/format/pcap.cpp
```diff
@@ -503,11 +503,11 @@ caf::error writer::write(const table_slice& slice) {
     return make_error(ec::format_error, "invalid pcap packet type");
   // TODO: consider iterating in natural order for the slice.
   for (size_t row = 0; row < slice.rows(); ++row) {
-    auto payload_field = slice.at(row, 6);
+    auto payload_field = slice.at(row, 6, *pcap_packet_type.at("payload"));
     auto& payload = caf::get<view<std::string>>(payload_field);
     // Make PCAP header.
     ::pcap_pkthdr header;
-    auto ns_field = slice.at(row, 0);
+    auto ns_field = slice.at(row, 0, *pcap_packet_type.at("time"));
```
mavam, line 510: A bit unrelated: I never liked the manual integer offsets to get to a field. Maybe we can factor this logic to happen dynamically when the writer gets instantiated?

mavam (commented):
Looks like you forgot to add `table_slice.hpp` changes? I'm curious what drives the need for the `at` overload.

tobim:
@mavam:
> Looks like you forgot to add table_slice.hpp changes? I'm curious what drives the need for the at overload.

It's in [Cache the type for column based data access](97f5e480d2444020310c5dddafb29b4b40804834)

libvast/src/format/csv.cpp
tobim: The output is unchanged. I just simplified the logic from 
```
print(typename)
print(separator)
print(fields[0])
for (size_t i = 1; i < fields.size(); ++i) {
  print(separator)
  print(fields[i])
}
```
to
```
print(typename)
for (size_t i = 0; i < fields.size(); ++i) {
  print(separator)
  print(fields[i])
}
```
We would need to revert that if we added an option to omit the typename.

mavam:
> It's in Cache the type for column based data access

Ah, got it. The link yields a 404 for me, though,

dominiklohmann:
Nit: Can you add an assertion to the overload that checks whether the given type is correct?

mavam:
Weird CI failure...

dominiklohmann:
> Weird CI failure...

Why? This is exactly #701, this PR just needs to be rebased.

libvast/src/msgpack_table_slice.cpp
```diff
@@ -281,7 +281,7 @@ table_slice::size_type msgpack_table_slice<FlatBuffer>::rows() const noexcept {
 template <class FlatBuffer>
 table_slice::size_type
 msgpack_table_slice<FlatBuffer>::columns() const noexcept {
-  return layout().fields.size();
+  return flatten(layout()).fields.size();
```
dominiklohmann, line 284: This should probably not flatten on every call to columns.

libvast/vast/table_slice.hpp
```diff
@@ -143,6 +143,13 @@ class table_slice final {
   /// @pre `row < rows() && column < columns()`
   data_view at(size_type row, size_type column) const;
 
+  /// Retrieves data by specifying 2D-coordinates via row and column.
+  /// @param row The row offset.
+  /// @param column The column offset.
+  /// @param t The type of the value to be retrieved.
+  /// @pre `row < rows() && column < columns()`
```
dominiklohmann, line 150: The precondition needs to be adapted to include the assertion that checks that the given type is correct.

Additionally, can you expand on the general description of this function to include the motivation? E.g., to be used when the type is already known for faster access.

dominiklohmann (commented):
This PR is great. I've done some performance measurements, and the differences are barely noticeable for Arrow, but show a slight negative impact for MessagePack. We should probably cache the flat layout in there as well as suggested by my previous review.

Additionally, please expand the documentation of the new `table_slice::at` function.

libvast/vast/type.hpp
```diff
@@ -611,6 +611,11 @@ struct record_type final : recursive_type<record_type> {
   /// Constructs a record type from a list of fields.
   record_type(std::initializer_list<record_field> xs);
 
+  /// Calculates the number of basic types that can be found when traversing the
+  /// tree. An faster version of `flatten(*this).fields.size()` or
+  /// `auto rng = each{*this}; std::distance(rng.begin(), rng.end())`
+  size_t leafs() const;
```
mavam, line 617:
```suggestion
  size_t leaves() const;
```

SCNR

libvast/vast/type.hpp
```diff
@@ -610,6 +611,11 @@ struct record_type final : recursive_type<record_type> {
...
```
dominiklohmann, line 617: This function needs a unit test, and should be named `num_leaves` for clarity on the calling side instead.

tobim:
Feedback is integrated and PR is rebased to fix the merge conflict.

libvast/src/format/pcap.cpp
tobim: I think that belongs to another PR.

dominiklohmann (approved):
I tested this again locally. Performance is alright, the loss is irrelevant in the grand scheme of things.

Thanks for working on this, it's a big UX improvement.

dominiklohmann:
@tobim you may have to do some adjustments on this because of #1230. If you wanna pair on the adjustments, feel free to ping me in Slack or here.

PR #1258, opened 2021-01-04
Fix Address Sanitizer issues in vast-test

dominiklohmann:

**Still work-in-progress.**

This PR enables ASan for vast-test and fixes the build with ASan enabled.

###  :memo: Checklist

Commit-by-commit.

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -219,6 +219,10 @@ jobs:
       - name: Run Unit Tests
         env:
           CTEST_OUTPUT_ON_FAILURE: YES
+          # On macOS, there is a possibly false positive container overflow in
+          # caf::detail::get_mac_address that we did not manage to understand
+          # entirely, so we disable that check for macOS only (ch21254).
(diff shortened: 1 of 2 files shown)


lava:
I tried reproducing this for a while, but
  - CI keeps being green after a few tries
  - Local build with g++10, clang++11, using CI and Debug build types, both on master and on this branch [I didn't test all possible combinations but a subset] keep being successful when running in an endless loop with `stress-ng` in the background.

I'm not sure if something changed in the build setup (e.g. a new asan version) or if this is just a statistical fluke, but I think we can merge this and continue investigating when we can observe it again.

dominiklohmann:
@lava and I managed to reproduce this on my setup locally. The issue occurs only on macOS and only with clang-11, and we think it might be a false positive container overflow detection from address sanitizer. Notably, this occurs in Debug builds only. Should this ever come up again we can always take another look at this. For now, we have disabled container overflow detection for macOS in CI.

tobim (approved):
I trust CI with this one.

PR #1263, opened 2021-01-06
Restrict arguments of 'vast::detail::as_printer'.

lava:

While working on an unrelated PR, we discovered that a harmless bit shift of two numbers can be interpreted as the construction of a `sequence_printer` when the namespace vast is visible where the code is instantiated:

```
/usr/include/fmt/core.h:1856
    unsigned int mask = (1 << detail::packed_arg_bits) - 1;
```

This is not ideal, since it is very easy to accidentally step into, and requires a massive debugging effort by the unsuspecting victim.

###  :memo: Checklist

### DIFF
--- a/libvast/vast/concept/printable/detail/as_printer.hpp
+++ b/libvast/vast/concept/printable/detail/as_printer.hpp
@@ -48,6 +48,8 @@ constexpr auto as_printer(T x) -> std::enable_if_t<is_printer_v<T>, T> {
 
 // -- binary ------------------------------------------------------------------
 
+// clang-format off
+
 template <class T>
 constexpr bool is_convertible_to_unary_printer_v =
   std::is_convertible_v<T, std::string>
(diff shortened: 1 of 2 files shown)


libvast/vast/concept/printable/detail/as_printer.hpp
```diff
@@ -48,6 +48,8 @@ constexpr auto as_printer(T x) -> std::enable_if_t<is_printer_v<T>, T> {
 
 // -- binary ------------------------------------------------------------------
 
+// clang-format off
```
dominiklohmann, line 51: Can you move this to around line 90?

dominiklohmann (approved):
Did not test locally, trusting CI on this one. Great change in general.

libvast/vast/concept/printable/detail/as_printer.hpp
lava: That will destroy the formatting of the existing code though, which looks like hand-crafted and which is currently not liked by `clang-format`.

dominiklohmann: Alright, in that case go ahead without this change.

PR #1264, opened 2021-01-06
Print plugin versions

mavam:

The `version` command now prints the plugin version as follows:

```json
{
  "VAST": "2020.12.16-97-gced115d91-dirty",
  "CAF": "0.17.6",
  "Apache Arrow": "2.0.0",
  "PCAP": "libpcap version 1.9.1",
  "jemalloc": null,
  "plugins": {
    "example": "0.0.0-0"
  }
}
```

###  :memo: Checklist

In one shot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Every entry has a category for which we use the following visual abbreviations:
   spawning a new actor that processes the incoming stream of data. The
   directory `plugins/example` contains an example plugin.
   [#1208](https://github.com/tenzir/vast/pull/1208)
+  [#1264](https://github.com/tenzir/vast/pull/1264)
 
 - 🐞 For relocatable installations, the list of schema loading paths does not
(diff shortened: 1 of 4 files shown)


lava:
I fear this might be a bit premature, we probably need to work on the version concept a bit more before presenting it so prominently to users.

In particular, there are several relevant versions: The ABI version, relevant when loading the plugin (changes all the time), the API version of the VAST plugin API that the plugin was written against, relevant when compiling the plugin (expected to change frequently while the feature is experimental, relatively stable afterwards) and the finally the package version of the plugin itself, irrelevant to vast but relevant to any downstream users of the plugin.

The current version is laid out like the last one, with patch and tweak versions, but internally it's more used like the first two. We should probably think about which of these we want to collect and display, and how.

dominiklohmann (approved):
Let's move forward with this as is. We decided to do versioning checks for libvast API and ABI separately. This version number is only user-facing.

PR #1268, opened 2021-01-08
Use the ServerTester fixture for testing dump

dominiklohmann:

We have a known shutdown bug that affects only commands run with `-N`, and every so often the dump command triggers it. Especially so when testing locally, because the CI runners are somewhat slow. Because the issue has received a very low priority as does anything that is specific to `-N`, this commit serves as a workaround to make running the integration test suite locally more reliable.

###  :memo: Checklist

n/t

### DIFF
--- a/vast/integration/default_set.yaml
+++ b/vast/integration/default_set.yaml
@@ -329,9 +329,10 @@ tests:
 
   Taxonomy dump:
     tags: [concepts, models]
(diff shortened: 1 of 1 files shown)


dominiklohmann:
[ch21461]

tobim (approved):
Works locally.

PR #1269, opened 2021-01-08
Only update the release tarball in an actual release build

tobim:

The current download instructions on the installation docs downloads the latest master build:
```
RELEASE=2020.12.16
curl -L -O https://storage.googleapis.com/tenzir-public-data/vast-static-builds/vast-${RELEASE}-static-latest.tar.gz
tar xf vast-${RELEASE}-static-latest.tar.gz
bin/vast version
{
  "VAST": "2020.12.16-0-",
  "CAF": "0.17.6",
  "Apache Arrow": "2.0.0",
  "PCAP": "libpcap version 1.9.1 (with TPACKET_V3)",
  "jemalloc": "5.2.1-0-gea6b3e973b477b8061e0076bb257dbd7f3faa756",
  "plugins": {}
}
```
The `"plugins"` key should not be there.

### DIFF
--- a/.github/workflows/static-binary.yaml
+++ b/.github/workflows/static-binary.yaml
@@ -89,9 +89,16 @@ jobs:
         PUBLIC_GCS_BUCKET: tenzir-public-data
         STATIC_BINARY_FOLDER: vast-static-builds
       run: |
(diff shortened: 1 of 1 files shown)


tobim:
Note: I'm not sure what I did is correct, please check thoroughly.

mavam:
> The "plugins" key should not be there.

Why? We also have `jemalloc` in there if it's not used.

dominiklohmann:
> > The "plugins" key should not be there.
> 
> Why? We also have `jemalloc` in there if it's not used.

You're misunderstanding: Plugins weren't part of the last release, hence why this PR is needed.

@tobim this PR looks alright to me, but I'll let @0snap approve as our GCS expert.

0snap (approved):

PR #1270, opened 2021-01-08
Make table slice encodings printable

dominiklohmann:

As a small bonus, this fixes a subtle issue that caused undefined behavior when the option `vast.import.batch-encoding` was not set.

###  :memo: Checklist

Commit-by-commit.

### DIFF
--- a/libvast/src/table_slice_encoding.cpp
+++ b/libvast/src/table_slice_encoding.cpp
@@ -13,6 +13,8 @@
 
 #include "vast/table_slice_encoding.hpp"
 
+#include "vast/die.hpp"
+
 namespace vast {
 
 std::string to_string(table_slice_encoding encoding) noexcept {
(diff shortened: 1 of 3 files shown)


lava (approved):
Looks good to me assuming CI approves as well.

tobim (commented):
Why did you decide against using the `printable` framework?

dominiklohmann:
> Why did you decide against using the `printable` framework?

Because that requires adding another header whenever we want to print `table_slice_encoding`, which is very error prone.

tobim:
> Because that requires adding another header whenever we want to print `table_slice_encoding`, which is very error prone.

If you go by this logic then all printers should use the `to_string` style. I think consistency is more important here.

dominiklohmann:
We discussed this in Slack, and will make `printable` more convenient to use in the future.

PR #1271, opened 2021-01-08
Make the start command print the endpoint on stdout

tobim:

This is helpful for test scripts that pass in `:0` to get a random free port allocated from the operating system.

###  :memo: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,8 @@ Every entry has a category for which we use the following visual abbreviations:
 - ⚠️ Plugins must now be built against the exact same version of libvast that
   VAST is built against. [#1275](https://github.com/tenzir/vast/pull/1275)
 
-- ⚠️ `vast start` now prints the endpoint it is listening on on `stdout`.
+- ⚠️ `vast start` now prints the endpoint it is listening on on `stdout` when it
+  is invoked with the `--print-endpoint` option.
   [#1271](https://github.com/tenzir/vast/pull/1271)
 
 - 🎁 VAST queries now also accept `nanoseconds`, `microseconds`, `milliseconds`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,8 +26,8 @@ Every entry has a category for which we use the following visual abbreviations:
 - ⚠️ Plugins must now be built against the exact same version of libvast that
   VAST is built against. [#1275](https://github.com/tenzir/vast/pull/1275)
 
-- ⚠️ `vast start` now prints the endpoint it is listening on on `stdout` when it
-  is invoked with the `--print-endpoint` option.
+- ⚠️ `vast start` now prints the endpoint it is listening on, provided the user
+  added the `--print-endpoint` option.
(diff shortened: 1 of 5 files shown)


dominiklohmann:
This needs a changelog entry, and it also needs the start command documentation to be updated accordingly.

mavam (changes requested):
This feels like a not-so-nice hack. Let's try to find an alternative, or at least make it conditional on a given flag that we have to pass.

tobim:
> This feels like a not-so-nice hack. Let's try to find an alternative, or at least make it conditional on a given flag that we have to pass.

This allow us to write:
``` bash
coproc nodefd {
  vast --verbosity=quiet -e :0 start
}
read -u "${nodefd[0]}" endpoint
# node is listening now
vast -e $endpoint status
...
vast -e $endpoint stop
```

To me this seems like very reasonable behavior and definitely not a hack. Note that reading from `stderr` instead is not that easy, because the buffer would eventually fill up and block the logging thread of the node in a `write` operation.

mavam:
I understand where you are coming from, but I disagree that this warrants unconditional printing to stdout. Does anything speak against introducing an option that triggers this behavior, e.g., `--print-bound-port`? It's not a common case but as big consequences for launch tools that process stdout.

Also, how does your patch interact with log level "quiet"?

dominiklohmann:
> This feels like a not-so-nice hack. Let's try to find an alternative, or at least make it conditional on a given flag that we have to pass.

To me, this feels like the proper way to solve this problem. It's the natural way to solve it for improved scriptability.

However, I also agree that this should be behind a flag, e.g., `vast.start.show-port`, and that flag should default to false.

> Also, how does your patch interact with log level "quiet"?

It shouldn't have anything to do with log levels, this is printing to stdout so it can be processed by the caller. It necessarily has to be on a separate channels from regular log output.

CHANGELOG.md
```diff
@@ -12,7 +12,8 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
-- ⚠️ `vast start` now prints the endpoint it is listening on on `stdout`.
+- ⚠️ `vast start` now prints the endpoint it is listening on on `stdout` when it
+  is invoked with the `--print-endpoint` option.
```
```suggestion
- ⚠️ `vast start` now prints on `stdout` the endpoint it is listening on,
  provided the user added the `--print-endpoint` option.
```

vast.yaml.example
```diff
@@ -81,6 +81,9 @@ vast:
 
   # The `vast start` command starts a new VAST server process.
   start:
+    # Prints the endpoint for clients when the server is ready to accept
+    # connections.
```
mavam, line 85:
```suggestion
    # connections. This comes in handy when letting the OS choose an available
    # random port, i.e., when specifying 0 as port value.
```

mavam (approved):
Looks good, just a few word smithing tweaks.

CHANGELOG.md
tobim: I took the liberty to remove "on `stdout`".

PR #1272, opened 2021-01-08
Refactor INDEX to remove 'await_evaluation_maps' construct

lava:
###  :memo: Checklist

### DIFF
--- a/libvast/src/system/query_supervisor.cpp
+++ b/libvast/src/system/query_supervisor.cpp
@@ -38,7 +38,7 @@ namespace {
 
 query_supervisor_state::query_supervisor_state(
   query_supervisor_actor::stateful_pointer<query_supervisor_state> self)
-  : name("query_supervisor-") {
+  : open_requests(0), name("query_supervisor-") {
   name += std::to_string(self->id());
 }
 
@@ -54,31 +54,35 @@ query_supervisor_actor::behavior_type query_supervisor(
       VAST_DEBUG(self, "got a new query for", qm.size(),
                  "partitions:", get_ids(qm));
       VAST_ASSERT(!qm.empty());
-      VAST_ASSERT(self->state.open_requests.empty());
+      VAST_ASSERT(self->state.open_requests == 0);
       for (auto& [id, partition] : qm) {
-        self->state.open_requests.emplace(id, 1 /*qm.size()*/);
+        ++self->state.open_requests;
         // TODO: Add a proper configurable timeout.
-        // TODO: Handle the error case for the `then()` handler.
         self
           ->request(partition, caf::infinite, expr,
                     caf::actor_cast<partition_client_actor>(client))
-          .then([=, id = id](atom::done) {
-            auto& num_evaluators = self->state.open_requests[id];
-            if (--num_evaluators == 0) {
-              VAST_DEBUG(self, "collected all results for partition", id);
-              self->state.open_requests.erase(id);
-              // Ask master for more work after receiving the last sub
-              // result.
-              // TODO: We should schedule a new partition as soon as one has
-              // finished, otherwise each batch will be as slow as the slowest
-              // contained partition.
-              if (self->state.open_requests.empty()) {
-                VAST_DEBUG(self, "collected all results for all partitions");
+          .then(
+            [=, id = id](atom::done) {
+              if (--self->state.open_requests == 0) {
+                VAST_DEBUG(self, "collected all results for partition", id);
+                // Ask master for more work after receiving the last sub
+                // result.
+                // TODO: We should schedule a new partition as soon as the
+                // prvious one has finished, otherwise each batch will be as
+                // slow as the worst case of the batch.
                 self->send(client, atom::done_v);
                 self->send(master, atom::worker_v, self);
               }
-            }
-          });
+            },
+            [=](const caf::error& e) {
+              // TODO: Add a proper error handling path to escalate the error to
+              // the client.
+              VAST_ERROR(self, "encountered error while supervising query", e);
+              if (--self->state.open_requests == 0) {
+                self->send(client, atom::done_v);
+                self->send(master, atom::worker_v, self);
+              }
+            });
       }
     },
   };
--- a/libvast/src/system/query_supervisor.cpp
+++ b/libvast/src/system/query_supervisor.cpp
@@ -62,9 +62,10 @@ query_supervisor_actor::behavior_type query_supervisor(
           ->request(partition, caf::infinite, expr,
                     static_cast<const partition_client_actor&>(client))
           .then(
-            [=, id = id](atom::done) {
+            [=](atom::done) {
               if (--self->state.open_requests == 0) {
-                VAST_DEBUG(self, "collected all results for partition", id);
+                VAST_DEBUG(self, "collected all results for the current batch "
+                                 "of partitions");
                 // Ask master for more work after receiving the last sub
                 // result.
                 // TODO: We should schedule a new partition as soon as the
--- a/libvast/src/system/query_supervisor.cpp
+++ b/libvast/src/system/query_supervisor.cpp
@@ -48,26 +48,27 @@ query_supervisor_actor::behavior_type query_supervisor(
   // Ask master for initial work.
   self->send(master, atom::worker_v, self);
   return {
-    [=](const expression&, const query_map& qm,
+    [=](const expression& expr,
+        const std::vector<std::pair<uuid, partition_actor>>& qm,
         const index_client_actor& client) {
       VAST_DEBUG(self, "got a new query for", qm.size(),
                  "partitions:", get_ids(qm));
       VAST_ASSERT(!qm.empty());
       VAST_ASSERT(self->state.open_requests.empty());
-      for (auto& kvp : qm) {
-        auto& id = kvp.first;
-        auto& evaluators = kvp.second;
-        VAST_DEBUG(self, "asks", evaluators.size(),
-                   "EVALUATOR actor(s) for partition", id);
-        self->state.open_requests.emplace(id, evaluators.size());
-        for (auto& evaluator : evaluators)
-          self->request(evaluator, caf::infinite, client).then([=](atom::done) {
+      for (auto& [id, partition] : qm) {
+        self->state.open_requests.emplace(id, 1 /*qm.size()*/);
+        // TODO: Add a proper configurable timeout.
+        self->request(partition, caf::infinite, expr, client)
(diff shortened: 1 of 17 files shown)


libvast/src/system/query_supervisor.cpp
```diff
@@ -48,26 +48,27 @@ query_supervisor_actor::behavior_type query_supervisor(
...
-      for (auto& kvp : qm) {
-        auto& id = kvp.first;
-        auto& evaluators = kvp.second;
-        VAST_DEBUG(self, "asks", evaluators.size(),
-                   "EVALUATOR actor(s) for partition", id);
-        self->state.open_requests.emplace(id, evaluators.size());
-        for (auto& evaluator : evaluators)
-          self->request(evaluator, caf::infinite, client).then([=](atom::done) {
+      for (auto& [id, partition] : qm) {
+        self->state.open_requests.emplace(id, 1 /*qm.size()*/);
```
tobim, line 59: Even in the old design I don't think there could be more than 1. I think `open_requests` can be replaced by a count, and we can refactor to use `fan_out_request` when we drop compatibility with CAF 0.17.

tobim (commented):
:rocket: This is so much simpler. Thanks for taking this on!

libvast/src/system/query_supervisor.cpp
lava: Good point, updated.

libvast/src/system/query_supervisor.cpp
```diff
@@ -48,33 +48,41 @@ query_supervisor_actor::behavior_type query_supervisor(
...
-              // result.
-              if (self->state.open_requests.empty()) {
-                VAST_DEBUG(self, "collected all results for all partitions");
+      VAST_ASSERT(self->state.open_requests == 0);
+      for (auto& [id, partition] : qm) {
+        ++self->state.open_requests;
+        // TODO: Add a proper configurable timeout.
+        self
+          ->request(partition, caf::infinite, expr,
+                    caf::actor_cast<partition_client_actor>(client))
```
dominiklohmann, line 63:
```suggestion
                    static_cast<partition_client_actor>(client))
```

libvast/src/system/query_supervisor.cpp
```diff
@@ -54,31 +54,35 @@ query_supervisor_actor::behavior_type query_supervisor(
...
-              // result.
-              // TODO: We should schedule a new partition as soon as one has
-              // finished, otherwise each batch will be as slow as the slowest
-              // contained partition.
-              if (self->state.open_requests.empty()) {
-                VAST_DEBUG(self, "collected all results for all partitions");
+          .then(
+            [=, id = id](atom::done) {
+              if (--self->state.open_requests == 0) {
+                VAST_DEBUG(self, "collected all results for partition", id);
```
tobim, line 67: I don't think the capture of `id` makes sense here. If my understanding is correct the debug message will always print the id of the partition that was requested last. I'd say to change the log message to `VAST_DEBUG(self, "collected all result for the current batch of partitions");`.

libvast/src/system/query_supervisor.cpp
tobim: Why?

libvast/vast/system/partition_client_actor.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include <caf/typed_event_based_actor.hpp>
+
+namespace vast::system {
+
+/// The PARTITION CLIENT actor interface.
+using partition_client_actor = caf::typed_actor<
+	// The client receives several sets of ids followed by a final
+	// `atom::done`, which as sent as response message to the
+    // higher-level request.
```
tobim, line 26: nit: indentation.

libvast/vast/system/partition_client_actor.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include "vast/fwd.hpp"
+
+#include <caf/typed_event_based_actor.hpp>
+
+namespace vast::system {
+
+/// The PARTITION CLIENT actor interface.
+using partition_client_actor = caf::typed_actor<
+	// The client receives several sets of ids followed by a final
+	// `atom::done`, which as sent as response message to the
```
tobim, line 25: Why is the `done` that is mentioned here not listed below?

libvast/vast/system/index_client_actor.hpp
```diff
@@ -19,11 +19,17 @@
 
 namespace vast::system {
 
+/// The PARTITION CLIENT actor interface.
+// FIXME: Move into separate header.
+using partition_client_actor = caf::typed_actor<caf::reacts_to<ids>>;
+
```
tobim: Remove?

libvast/vast/system/index_client_actor.hpp
```diff
@@ -19,11 +19,17 @@
...
 /// The INDEX CLIENT actor interface.
 using index_client_actor = caf::typed_actor<
   // Receives ids from the INDEX for partial query hits.
-  caf::reacts_to<ids>,
+  // caf::reacts_to<ids>,
```
tobim: Remove?

libvast/vast/system/index_client_actor.hpp
```diff
@@ -19,11 +19,17 @@
...
   // Receives done from the INDEX when the query finished.
-  caf::reacts_to<atom::done>>;
+  caf::reacts_to<atom::done>>
+  // FIXME: comment
+  ::extend_with<partition_client_actor>;
```
tobim: Do we actually need *both*?

tobim (changes requested):
A bit of cleanup is necessary, but overall it looks good.

I'm not sure of the split between `index_client_actor` and `partition_client_actor`, maybe we can get by with only one.

libvast/src/system/query_supervisor.cpp
dominiklohmann: Because it suffices. `actor_cast` can cast to typed actors whose behaviors aren't subsets, while `static_cast` only works if the target behavior is a subset of the source behavior.

tobim: Good to know.

libvast/vast/system/partition_actor.hpp
```diff
@@ -26,6 +27,8 @@ namespace vast::system {
 /// The PARTITION actor interface.
 using partition_actor = caf::typed_actor<
   // Evaluate the given expression, returning the relevant evaluation triples.
-  caf::replies_to<expression>::with<std::vector<evaluation_triple>>>;
+  // TODO: Passing the `index_client_actor` here is an historical artifact,
```
tobim, line 30:
```suggestion
  // TODO: Passing the `partition_client_actor` here is an historical artifact,
```

libvast/vast/system/partition_actor.hpp
```diff
@@ -16,6 +16,7 @@
 #include "vast/fwd.hpp"
 
 #include "vast/system/evaluation_triple.hpp"
+#include "vast/system/index_client_actor.hpp"
```
tobim, line 19:
```suggestion
```

libvast/test/flatbuffers.cpp
```diff
@@ -243,18 +249,18 @@ TEST(full partition roundtrip) {
...
+  test_expression(x_equals_one, 0);
+  // For the query `#type == "x"`, we expect one indexer (a one-shot indexer
   // for type queries) and one result.
-  test_expression(type_equals_y, 1, 1);
-  // // For the query `#type == "foo"`, we expect one indexer (a one-shot
+  test_expression(type_equals_y, 1);
+  // For the query `#type == "foo"`, we expect one indexer (a one-shot
   // indexer for type queries) and no results.
-  test_expression(type_equals_foo, 1, 0);
+  test_expression(type_equals_foo, 0);
```
tobim, line 263: We don't expect any indexers anymore here.

libvast/src/system/query_supervisor.cpp
lava: Maybe I'm missing something, but aren't we copying the the `id` of the current iteration each time?

tobim (changes requested):
A few more small things.

libvast/vast/system/index_client_actor.hpp
lava: The alternative would be to implement a dummy `done` handler for the partition clients that never gets called. Which is extra confusing because the reader mustn't confuse it with the "outer" `done`-handler that handles the end of the request.

libvast/vast/system/partition_client_actor.hpp
lava: It's not a free message but sent as replay to the initial request. I've reworded the comment to make it more clear.

dominiklohmann: I think in the context of a partition_client_actor that is irrelevant.

libvast/src/system/query_supervisor.cpp
tobim: Yes, each instance of the lambda gets an id, but only the one from the last partition is printed in the log. You could also move the log statement before the condition to change that.

libvast/vast/system/index_client_actor.hpp
tobim: I think I understand it now. It is required because client gets the `done` from the `query_supervisor`. I think we should change the evaluator to not send partial results anymore soon, then we can avoid this complex multi-party protocol.

dominiklohmann: Are you sure about this? The measurements from #1220 have shown that the partial results have quite the impact. If we only send a `vector<ids>` instead of using hand-rolled streaming of `ids` then we lose a fair bit of performance.

I think this should be refactored to use CAF streaming instead of our poor man's version of it.

lava: Either way, I'd suggest doing that as a separate refactoring.

lava: (also, @tobim, that's almost correct but the partition itself also sends a `done` atom [see partition_actor.hpp], it's just that this is sent as a response to the expression request and is therefore not part of the callback interface)

libvast/src/system/query_supervisor.cpp
lava: Ah, right, updated.

tobim: `... for all partitions` is misleading, because is suggests that the index is done with that query, but this message will be shown each time a subset of the candidates is done, e.g. after "request more hits".

tobim (approved):

PR #1273, opened 2021-01-09
Replace radix trie with simpler approach

mavam:

This PR removes the 3rd party radix tree implementation from the code base. The new vector-based structure will probably be faster because (i) it's a single cache-efficient block of memory, and (ii) we no longer copy the *entire* input before attempting a parse operation.

###  :memo: Checklist

File-by-file.


tobim (approved):
Looks fine. I entrust the testing to the CI.

PR #1275, opened 2021-01-11
Use libvast tree hash for plugin API/ABI checks

dominiklohmann:

This is a bit of a hack, but it's better than the nothing we've had
before.  Until libvast is stable and versioned, we will have to require
exact version matches between the libvast loading plugins and the
libvast plugins are linking against.

###  :memo: Checklist

Test the mismatch locally.


CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- ⚠️ Plugins must now link against the exact same version of libvast that VAST
```
lava, line 15: Must now be built against [...]

libvast/src/plugin.cpp
```diff
@@ -48,7 +48,10 @@ std::string to_string(plugin_version x) {
 }
 
 bool has_required_version(const plugin_version& version) noexcept {
-  return plugin::version.major == version.major
+  return !std::strcmp(plugin::version.libvast_version, version.libvast_version)
+         && !std::strcmp(plugin::version.libvast_build_tree_hash,
+                         version.libvast_build_tree_hash)
+         && plugin::version.major == version.major
```
lava, line 54: Do we still need to compare major/minor/patch/tweak version? I thought we were going to use these as the user-facing plugin version.

libvast/vast/plugin.hpp
```diff
@@ -44,6 +44,14 @@ extern "C" struct plugin_version {
   uint16_t minor;
   uint16_t patch;
   uint16_t tweak;
+
+  /// An additional field containing the version of the libvast that is linked
+  /// against. Should compare equal to VAST_VERSION.
+  const char* libvast_version;
```
tobim, line 50: Why are the preprocessor variables called `VAST_` but the ones here `libvast_*`?

libvast/vast/plugin.hpp
```diff
@@ -44,6 +44,14 @@ extern "C" struct plugin_version {
...
+
+  /// An additional field containing the hash of the build tree of the libvast
+  /// that is linked against. Should compare equal to VAST_BUILD_TREE_HASH.
+  const char* libvast_build_tree_hash;
 };
```
mavam, line 55: These should not go into `plugin_version`, I think. They are conceptually different things and I don't think we need to add them in this particular struct. Can't these just be tracked separate from `plugin_version`?

dominiklohmann: Then you'll need to add additional exported functions to the plugin API.

libvast/vast/plugin.hpp
```diff
@@ -68,7 +77,8 @@ class plugin {
 public:
   /// The current version of the plugin API. When registering a plugin, set the
   /// corresponding plugin version in the `VAST_REGISTER_PLUGIN` macro.
-  constexpr static auto version = plugin_version{0, 1, 0, 0};
+  constexpr static auto version
+    = plugin_version{0, 1, 0, 0, VAST_VERSION, VAST_BUILD_TREE_HASH};
```
tobim, line 81: Do we still need the `major` `minor` `patch` and `tweak` members in this struct? I guess it makes sense when we have some stability guarantees?

tobim (commented):
More questions than change requests.

CMakeLists.txt
```diff
@@ -36,6 +36,28 @@ if (NOT IS_ABSOLUTE "${CMAKE_INSTALL_PREFIX}")
       "CMAKE_INSTALL_PREFIX must be an absolute path: ${CMAKE_INSTALL_PREFIX}")
 endif ()
 
+# -- build tree hash -----------------------------------------------------------
+
+file(GLOB_RECURSE hash_files CONFIGURE_DEPENDS "libvast/*" "cmake/*"
+     "CMakeLists.txt")
```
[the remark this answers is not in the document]
dominiklohmann: That's already happening directly below:

```
list(FILTER hash_files EXCLUDE REGEX
     "^${CMAKE_CURRENT_SOURCE_DIR}/libvast/aux/")
```

libvast/vast/plugin.hpp
dominiklohmann: In terms of the plugin, it is libvast that is relevant for us. Within libvast, all macros are prefix with `VAST_`.

libvast/vast/plugin.hpp
dominiklohmann: For now I've opted to keep them, although they're pretty much useless until we make such guarantees. Shall I remove them for now?

libvast/vast/plugin.hpp
mavam: Seems cleaner to me that way, now that you mention it.

libvast/src/plugin.cpp
dominiklohmann: We can always change this later on.

libvast/src/plugin.cpp
```diff
@@ -59,17 +59,30 @@ bool has_required_version(const plugin_version& version) noexcept {
...
-      auto plugin_create = reinterpret_cast<::vast::plugin* (*) ()>(
-        dlsym(library_, "vast_plugin_create"));
-      auto plugin_destroy = reinterpret_cast<void (*)(::vast::plugin*)>(
-        dlsym(library_, "vast_plugin_destroy"));
-      if (plugin_create && plugin_destroy) {
-        instance_ = plugin_create();
-        deleter_ = plugin_destroy;
-      }
+    if (!plugin_version || !has_required_version(plugin_version()))
+      return;
```
lava, line 77: We should print some kind of appropriate error message if loading the plugin fails. (We can also do this in a follow-up PR if you keep working on plugins anyways.)

dominiklohmann: Yeah I had the same thought already, probably moving from `plugin_ptr::plugin_ptr(...)` to `caf::expected<plugin_ptr> plugin_ptr::make(...)`.

lava (approved):
Not tested personally, but looks good to me and reflects our discussions (and since its a new feature there's no chance of regressions :)

PR #1284, opened 2021-01-15
Fix exporting of nested records

tobim:

Fixes quoting of JSON node keys in exported events.


dominiklohmann (approved):
No need to test manually. The change is simple, and looks good to me. The integration test suite is enough for verification.

PR #1282, opened 2021-01-14
Allow for adding plugin source trees into the build process

dominiklohmann:
###  :notebook_with_decorative_cover: Description

- [x] A method to add additional plugin source trees exists, and is utilized for the example plugin (which in turn is moved to the examples/ directory)
- [x] Plugins can specify test suites that get executed when running the test build target
- [x] Plugins can specify additional integration test suites

###  :memo: Checklist

- [x] All user-facing changes have changelog entries.
- [x] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [x] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

Commit-by-commit. Run locally to verify. Copy the example plugin elsewhere on your system to test for issues.

PR #1285, opened 2021-01-15
Load only explicitly specified plugins

dominiklohmann:

This changes the loading of plugins to only consider explicitly specified plugins.

###  :memo: Checklist

File-by-file.

This is based on #1282, but this can be reviewed independently.

### DIFF
--- a/vast/vast.cpp
+++ b/vast/vast.cpp
@@ -92,27 +92,55 @@ int main(int argc, char** argv) {
   // deferred log statements essentially.
   auto plugin_load_errors = std::vector<caf::error>{};
   auto loaded_plugin_paths = std::vector<path>{};
-  for (const auto& dir : plugin_dirs) {
-    for (const auto& file : directory{dir}) {
-      if (file.extension() == ".so" || file.extension() == ".dylib") {
-        if (auto plugin = plugin_ptr::make(file.str().c_str())) {
-          VAST_ASSERT(*plugin);
-          auto has_same_name = [name = (*plugin)->name()](const auto& other) {
-            return !std::strcmp(name, other->name());
-          };
-          if (std::none_of(plugins.begin(), plugins.end(), has_same_name)) {
-            loaded_plugin_paths.push_back(std::move(file));
-            plugins.push_back(std::move(*plugin));
-          } else {
-            std::cerr << "failed to load plugin " << file.str()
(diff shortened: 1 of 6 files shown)


vast/vast.cpp
```diff
@@ -92,27 +92,55 @@ int main(int argc, char** argv) {
...
-        }
+  auto plugin_files
+    = caf::get_or(cfg, "vast.plugins", std::vector<std::string>{});
+  auto load_plugin = [&](path file) {
+#if VAST_MACOS
+    if (file.extension() == "")
+      file = file.str() + ".dylib";
+#else
+    if (file.extension() == "")
+      file = file.str() + ".dylib";
```
dominiklohmann, line 103:
```suggestion
      file = file.str() + ".so";
```

vast/vast.cpp
```diff
@@ -144,11 +172,11 @@ int main(int argc, char** argv) {
...
     cfg.config_files.emplace_back(std::move(cfg.config_file_path));
-  for (auto& path : cfg.config_files)
-    VAST_INFO_ANON("loaded configuration file:", path);
+  for (auto& file : cfg.config_files)
+    VAST_INFO_ANON("loaded configuration file:", file);
   // Print the plugins that were loaded, and errors that occured during loading.
-  for (auto&& path : std::exchange(loaded_plugin_paths, {}))
-    VAST_VERBOSE_ANON("loaded plugin:", std::move(path));
+  for (auto&& file : std::exchange(loaded_plugin_paths, {}))
+    VAST_VERBOSE_ANON("loaded plugin:", std::move(file));
```
mavam, line 179:
```suggestion
  for (auto&& file : loaded_plugin_paths)
    VAST_VERBOSE_ANON("loaded plugin:", file);
```

Why not go for a simpler solution if it's possible...aka KISS? It seems like the gains in performance do not justify the added complexity, but I might be wrong. 

Either way, I don't have strong opinions on this issue, but I just want to remind that the *simplest* solution often suffices..

mavam (approved):

vast/vast.cpp
dominiklohmann: While I generally disagree with not using best practices, I do not hold too strong of an opinion on the matter to not simply make the requested change.

PR #1287, opened 2021-01-18
Rename *-paths to *-dirs options

dominiklohmann:

The `*-paths` option in our `vast.yaml` feel really confusing. A directory is necessarily a path, but not the other way round.

The old options continue to work, but are deprecated.

###  :memo: Checklist

Try it out locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- ⚠️ The options `vast.schema-paths` and `vast.plugin-paths` are now named
+  `vast.schema-dirs` and `vast.plugin-dirs` respectively. The old options are
+  deprecated, and will be removed in a future release.
+  [#1287](https://github.com/tenzir/vast/pull/1287)
+
 - ⚠️ VAST now preserves nested JSON objects in events instead of formatting them
   in a flattened form when exporting data with `vast export json`. The old
   behavior can be enabled with `vast export json --flatten`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Every entry has a category for which we use the following visual abbreviations:
 
 - ⚠️ The options `vast.schema-paths` and `vast.plugin-paths` are now named
   `vast.schema-dirs` and `vast.plugin-dirs` respectively. The old options are
-  deprecated, and will be removed in a future release.
+  deprecated and will be removed in a future release.
   [#1287](https://github.com/tenzir/vast/pull/1287)
 
 - ⚠️ VAST now preserves nested JSON objects in events instead of formatting them
--- a/libvast/src/schema.cpp
+++ b/libvast/src/schema.cpp
@@ -239,9 +239,15 @@ get_schema_dirs(const caf::actor_system_config& cfg,
     else if (const char* home = std::getenv("HOME"))
       result.insert(path{home} / ".local" / "share" / "vast" / "schema");
   }
-  if (auto user_dirs
-      = caf::get_if<std::vector<std::string>>(&cfg, "vast.schema-paths"))
-    result.insert(user_dirs->begin(), user_dirs->end());
+  if (auto dirs = caf::get_if<std::vector<std::string>>( //
+        &cfg, "vast.schema-dirs"))
+    result.insert(dirs->begin(), dirs->end());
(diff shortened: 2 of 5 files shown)


libvast/src/system/application.cpp
```diff
@@ -97,7 +97,7 @@ auto make_root_command(std::string_view path) {
         .add<std::string>("config", "path to a configuration file")
         .add<caf::atom_value>("verbosity", "output verbosity level on the "
                                            "console")
-        .add<std::vector<std::string>>("schema-paths", schema_desc.c_str())
+        .add<std::vector<std::string>>("schema-dirs", schema_desc.c_str())
```
tobim, line 100: The changelog entry says that the old options are deprecated, but this change removes them completely from the cli and they will only work from `vast.yaml`. I think we should strive to keep both in sync.

tobim (commented):
Not much to say, tested and observed the behavior that I commented on below.

CHANGELOG.md
```diff
@@ -12,6 +12,11 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- ⚠️ The options `vast.schema-paths` and `vast.plugin-paths` are now named
+  `vast.schema-dirs` and `vast.plugin-dirs` respectively. The old options are
+  deprecated, and will be removed in a future release.
```
```suggestion
  deprecated and will be removed in a future release.
```

libvast/src/schema.cpp
```diff
@@ -239,9 +239,15 @@ get_schema_dirs(const caf::actor_system_config& cfg,
     else if (const char* home = std::getenv("HOME"))
       result.insert(path{home} / ".local" / "share" / "vast" / "schema");
   }
-  if (auto user_dirs
-      = caf::get_if<std::vector<std::string>>(&cfg, "vast.schema-paths"))
-    result.insert(user_dirs->begin(), user_dirs->end());
+  if (auto dirs = caf::get_if<std::vector<std::string>>( //
```
mavam, line 242: I think it would be nice if we could factor all CLI deprecations into a single function so that we only have to check one place in future releases.

vast.yaml.example
```diff
@@ -15,15 +15,15 @@ vast:
   # working directory of the client. Note that this is disabled by default. If
   # not specified no log files are written for clients at all.
   #client-log-file: "client.log"
-  # List of paths to look for schema files in ascending order of priority.
+  # List of dirs to look for schema files in ascending order of priority.
```
mavam, line 18:
```suggestion
  # List of directories to look for schema files in ascending order of priority.
```

vast/vast.cpp
```diff
@@ -65,9 +65,15 @@ stable_set<path> get_plugin_dirs(const caf::actor_system_config& cfg) {
...
-    result.insert(user_dirs->begin(), user_dirs->end());
+  if (auto dirs = caf::get_if<std::vector<std::string>>( //
+        &cfg, "vast.plugin-dirs"))
+    result.insert(dirs->begin(), dirs->end());
+  if (auto dirs = caf::get_if<std::vector<std::string>>( //
+        &cfg, "vast.plugin-paths")) {
+    VAST_WARNING_ANON(__func__, "encountered deprecated vast.plugin-paths "
+                                "option; use vast.plugin-dirs instead.");
+    result.insert(dirs->begin(), dirs->end());
+  }
```
mavam, line 76: Since we haven't release the plugin API yet, I think we can avoid jumping through the hoop here.

tobim: I didn't even consider that. It makes sense to completely remove `plugin-paths`.

libvast/src/schema.cpp
dominiklohmann: Not sure if that's worth the effort tbh.

dominiklohmann: To clarify, this would ideally be done when parsing the options, but we cannot use the logger at that point. So I'd like to keep it as is currently.

dominiklohmann: Marking this as resolved as per Slack. `git grep deprecated` will suffice for now.

tobim (approved):
Seems to work as intended. It even prints the warning twice for emphasis :stuck_out_tongue: .

dominiklohmann:
> It even prints the warning twice for emphasis 😛 .

Noticed that as well, we seem to read the schemas twice on startup. I'll write a story so we don't forget about that.

PR #1289, opened 2021-01-18
Fix Arrow export for nested records

dominiklohmann:

This fixes a small oversight from the recent unflattening PR.

###  :memo: Checklist

Should be easy to verify locally like this:

```
❯ vast -N import suricata -r vast/integration/data/suricata/eve.json
❯ vast -N export arrow '"" in #type'
```

### DIFF
--- a/libvast/src/format/arrow.cpp
+++ b/libvast/src/format/arrow.cpp
@@ -15,8 +15,6 @@
 
 #if VAST_ENABLE_ARROW
 
-#  include "vast/format/arrow.hpp"
-
 #  include "vast/arrow_table_slice.hpp"
 #  include "vast/arrow_table_slice_builder.hpp"
 #  include "vast/detail/assert.hpp"
@@ -24,6 +22,7 @@
 #  include "vast/detail/fdoutbuf.hpp"
 #  include "vast/detail/string.hpp"
 #  include "vast/error.hpp"
+#  include "vast/format/arrow.hpp"
 #  include "vast/table_slice_builder.hpp"
 #  include "vast/type.hpp"
 
@@ -50,14 +49,16 @@ writer::~writer() {
 
 caf::error writer::write(const table_slice& slice) {
   if (out_ == nullptr)
-    return ec::filesystem_error;
+    return caf::make_error(ec::filesystem_error, "invalid arrow output stream");
   if (!layout(slice.layout()))
-    return ec::unspecified;
+    return caf::make_error(ec::logic_error, "failed to update layout");
   // Get the Record Batch and print it.
   auto batch = as_record_batch(slice);
   VAST_ASSERT(batch != nullptr);
-  if (!current_batch_writer_->WriteRecordBatch(*batch).ok())
-    return ec::filesystem_error;
+  if (auto status = current_batch_writer_->WriteRecordBatch(*batch);
+      !status.ok())
+    return caf::make_error(ec::filesystem_error, "failed to write record batch",
+                           status.ToString());
   return caf::none;
 }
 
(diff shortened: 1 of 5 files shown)


libvast/src/format/arrow.cpp
```diff
@@ -15,15 +15,14 @@
 
 #if VAST_ENABLE_ARROW
 
-#  include "vast/format/arrow.hpp"
-
```
tobim, line 19: Didn't we once have an `IncludeCategory` for this?

libvast/src/format/arrow.cpp
```diff
@@ -50,14 +49,16 @@ writer::~writer() {
 
 caf::error writer::write(const table_slice& slice) {
   if (out_ == nullptr)
-    return ec::filesystem_error;
+    return caf::make_error(ec::filesystem_error, "invalid arrow output stream");
```
tobim, line 52:
```suggestion
    return make_error(ec::filesystem_error, "invalid arrow output stream");
```

libvast/src/format/arrow.cpp
```diff
@@ -50,14 +49,16 @@ writer::~writer() {
...
   if (!layout(slice.layout()))
-    return ec::unspecified;
+    return caf::make_error(ec::logic_error, "failed to update layout");
```
tobim, line 54:
```suggestion
    return make_error(ec::logic_error, "failed to update layout");
```

libvast/src/format/arrow.cpp
```diff
@@ -50,14 +49,16 @@ writer::~writer() {
...
   // Get the Record Batch and print it.
   auto batch = as_record_batch(slice);
   VAST_ASSERT(batch != nullptr);
-  if (!current_batch_writer_->WriteRecordBatch(*batch).ok())
-    return ec::filesystem_error;
+  if (auto status = current_batch_writer_->WriteRecordBatch(*batch);
+      !status.ok())
+    return caf::make_error(ec::filesystem_error, "failed to write record batch",
```
tobim, line 60:
```suggestion
    return make_error(ec::filesystem_error, "failed to write record batch",
```

libvast/src/format/arrow.cpp
tobim: Isn't this actually a logic error?

libvast/src/format/arrow.cpp
tobim: We should introduce an `arrow_error` for these.

tobim (changes requested):

libvast/src/format/arrow.cpp
dominiklohmann: Yes, makes sense to change.

libvast/src/format/arrow.cpp
dominiklohmann: I'd rather make this `ec::unspecified`, as the Arrow Status contains its own category. In the case of this error, it was `Invalid`.

libvast/src/format/arrow.cpp
dominiklohmann: We still do. I think this is a change in recent clang-format versions. Let's just leave it as formatted by clang-format here.

libvast/src/format/arrow.cpp
dominiklohmann: Not doing these. `vast::make_error` is just an alias for `caf::make_error`, and we use the latter more than the former in the code base.

libvast/src/format/arrow.cpp
tobim: That's ok as well, the inner error is what matters.

libvast/src/format/arrow.cpp
tobim: Just for the record:
```
$ git grep 'make_error(' | wc -l
343
$ git grep 'caf::make_error(' | wc -l
24
```

But it seems I have been outvoted on this issue.

tobim (approved):

libvast/src/format/arrow.cpp
dominiklohmann: It's not as easy to grep for just error, but for `caf::error` the numbers were the other way around. ;)

PR #1292, opened 2021-01-19
Dont create parallel meta index inside the index

lava:

See commit messages.

NOTE: After implementing this, I think it might be a good idea to rewrite this once more to give the `persist`-handler the following signature:

```
using active_partition_actor = caf::typed_actor<
    caf::replies_to<atom::persist>::with<atom::done, partition_synopsis>,
    [...]
```

Then the logic in the indexer would be to always forward expressions to all partitions that are currently known as `unpersisted`.

But let's discuss that first.

###  :memo: Checklist

Commit by commit

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -50,6 +50,7 @@
 #include "vast/value_index.hpp"
 
 #include <caf/error.hpp>
+#include <caf/typed_response_promise.hpp>
 
 #include <flatbuffers/flatbuffers.h>
 
@@ -399,6 +400,8 @@ index(index_actor::stateful_pointer<index_state> self,
   self->state.filesystem = std::move(filesystem);
   self->state.dir = dir;
   self->state.partition_capacity = partition_capacity;
+  self->state.sequence_number = 0;
+  self->state.received_sequence_number = 0;
   self->state.taste_partitions = taste_partitions;
   self->state.inmem_partitions.factory().filesystem() = self->state.filesystem;
   self->state.inmem_partitions.resize(max_inmem_partitions);
@@ -418,8 +421,10 @@ index(index_actor::stateful_pointer<index_state> self,
     auto id = uuid::random();
     caf::settings index_opts;
     index_opts["cardinality"] = partition_capacity;
-    auto part = self->spawn(active_partition, id, self->state.filesystem,
-                            index_opts, self->state.meta_idx.factory_options());
+    auto part_sequence_number = ++self->state.sequence_number;
+    auto part = self->spawn(active_partition, id, part_sequence_number,
+                            self->state.filesystem, index_opts,
+                            self->state.meta_idx.factory_options());
     auto slot = self->state.stage->add_outbound_path(part);
     self->state.active_partition.actor = part;
     self->state.active_partition.stream_slot = slot;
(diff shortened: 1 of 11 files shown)


libvast/vast/system/index_actor.hpp
```diff
@@ -38,11 +38,12 @@ using index_actor = caf::typed_actor<
   // Subscribes a FLUSH LISTENER to the INDEX.
   caf::reacts_to<atom::subscribe, atom::flush, wrapped_flush_listener>,
   // Evaluatates an expression.
-  caf::reacts_to<expression>,
+  caf::reacts_to<expression>, caf::reacts_to<expression, size_t>, // private
   // Queries PARTITION actors for a given query id.
   caf::reacts_to<uuid, uint32_t>,
   // Replaces the SYNOPSIS of the PARTITION witht he given partition id.
```
tobim, line 44:
```suggestion
  // Replaces the SYNOPSIS of the PARTITION with the given partition id.
```

libvast/vast/system/index_actor.hpp
```diff
@@ -38,11 +38,12 @@ using index_actor = caf::typed_actor<
   // Subscribes a FLUSH LISTENER to the INDEX.
   caf::reacts_to<atom::subscribe, atom::flush, wrapped_flush_listener>,
   // Evaluatates an expression.
-  caf::reacts_to<expression>,
+  caf::reacts_to<expression>, caf::reacts_to<expression, size_t>, // private
```
tobim, line 41:
```suggestion
  caf::reacts_to<expression>, 
  // private
  caf::reacts_to<expression, size_t>,
```

libvast/vast/meta_index.hpp
```diff
@@ -70,19 +70,13 @@ struct partition_synopsis {
...
-  /// @param slice The table slice to extract data from.
-  /// @param partition The partition ID that *slice* belongs to.
+  /// index. Only used in unit tests.
+  // TODO: Remove this and use `merge()` in unit tests instead.
   void add(const uuid& partition, const table_slice& slice);
 
-  /// Adds new synopses for a partition in bulk. Used when
-  /// re-building the meta index state at startup.
+  /// Adds new synopses for a partition in bulk.
   void merge(const uuid& partition, partition_synopsis&&);
```
tobim, line 78: This function should be called `add`.

libvast/src/system/index.cpp
```diff
@@ -646,26 +656,30 @@ index(index_actor::stateful_pointer<index_state> self,
...
       // the unique owner of the pointer will be the message (which doesnt
       // need it anymore).
       // Semantically we want a unique_ptr here, but caf message types need
       // to be copy constructible.
       VAST_DEBUG(self, "replaces synopsis for partition", partition_id);
       if (!ps.unique()) {
-        VAST_WARNING(self, "ignores partition synopsis that is still in use");
-        // TODO: Should this return caf::skip?
-        return;
+        VAST_DEBUG(self, "skips partition synopsis that is still in use");
```
tobim, line 673: This is not supposed to happen, so it should still be a warning.

tobim (changes requested):
Just a quick first pass, more to come later.

libvast/src/system/index.cpp
```diff
@@ -50,6 +50,7 @@
 #include "vast/value_index.hpp"
 
 #include <caf/error.hpp>
+#include <caf/typed_response_promise.hpp>
```
dominiklohmann, line 53: This can be removed.

dominiklohmann (commented):
Just looked at this out of curiosity. Note that we should not use platform-dependent types like `size_t` in actor interfaces. Instead, use `uint64_t` or, which would be even better, use a distinct type that wraps an unsigned integer instead.

lava:
Closing this in favor of the superior design in https://github.com/tenzir/vast/pull/1296

PR #1295, opened 2021-01-20
Fix potential race condition between evaluator and partition

lava:

This came up while pairing on on unrelated PR with @tobim . I have no proof that the described race can actually happen in practice, but intuitively it makes sense to me that it would.

###  :memo: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 A potential race condition that could lead to a hanging export if
+  a partition was persisted just as it was scanned no longer exists.
+  [#1295](https://github.com/tenzir/vast/pull/1295)
+
 - ⚠️ The options `vast.schema-paths` and `vast.plugin-paths` are now named
   `vast.schema-dirs` and `vast.plugin-dirs` respectively. The old options are
   deprecated and will be removed in a future release.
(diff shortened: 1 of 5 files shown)


tobim (approved):
From looking at the changes, this seems to be correct and I do believe that the issue is real. I think one potential symptom is that a `query_supervisor` would miss a response from the cleaned-up partition and stay in limbo eternally, which in turn would block the corresponding exporter from iterating through the remaining candidate partitions.

tobim (changes requested):
This is missing a changlog entry.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ Every entry has a category for which we use the following visual abbreviations:
 
 ## Unreleased
 
+- 🐞 Removed potential race condition that could lead to a hanging export if
+  a partition was persisted just as it was scanned.
+  [#1295](https://github.com/tenzir/vast/pull/1295)
```
dominiklohmann, line 17:
```suggestion
- 🐞 A potential race condition that could lead to a hanging export if
  a partition was persisted just as it was scanned no longer exists.
  [#1295](https://github.com/tenzir/vast/pull/1295)
```

dominiklohmann:
@lava can you rebase onto master?

lava:
@dominiklohmann Sorry, didn't notice until after I pushed. Done.

dominiklohmann (approved):
No functional changes happened since I last reviewed this. CI should catch rebase mistakes.

lava:
Ignoring the mac CI fault since it is https://app.clubhouse.io/tenzir/story/20882/race-condition-in-node-queries-unit-test again.

PR #1296, opened 2021-01-20
 Dont create parallel meta index inside the index

lava:

See commit messages

###  :memo: Checklist

Look at the whole chain for an overview, then commit by commit.

### DIFF
--- a/libvast/src/partition_synopsis.cpp
+++ b/libvast/src/partition_synopsis.cpp
@@ -1,0 +1,88 @@
+/******************************************************************************
+ *                    _   _____   __________                                  *
+ *                   | | / / _ | / __/_  __/     Visibility                   *
+ *                   | |/ / __ |_\ \  / /          Across                     *
+ *                   |___/_/ |_/___/ /_/       Space and Time                 *
+ *                                                                            *
+ * This file is part of VAST. It is subject to the license terms in the       *
+ * LICENSE file found in the top-level directory of this distribution and at  *
+ * http://vast.io/license. No part of VAST, including this file, may be       *
+ * copied, modified, propagated, or distributed except according to the terms *
+ * contained in the LICENSE file.                                             *
+ ******************************************************************************/
+
+#include "vast/partition_synopsis.hpp"
+
+#include "vast/synopsis_factory.hpp"
+
+namespace vast {
+
+void partition_synopsis::shrink() {
+  for (auto& [field, synopsis] : field_synopses_) {
+    if (!synopsis)
+      continue;
+    auto shrinked_synopsis = synopsis->shrink();
+    if (!shrinked_synopsis)
+      continue;
+    synopsis.swap(shrinked_synopsis);
+  }
+  // TODO: Make a utility function instead of copy/pasting
+  for (auto& [field, synopsis] : type_synopses_) {
+    if (!synopsis)
+      continue;
+    auto shrinked_synopsis = synopsis->shrink();
+    if (!shrinked_synopsis)
+      continue;
+    synopsis.swap(shrinked_synopsis);
+  }
+}
+
+void partition_synopsis::add(const table_slice& slice,
+                             const caf::settings& synopsis_options) {
+  auto make_synopsis = [&](const type& t) -> synopsis_ptr {
+    return has_skip_attribute(t) ? nullptr
+                                 : factory<synopsis>::make(t, synopsis_options);
+  };
+  auto& layout = slice.layout();
+  auto each = record_type::each(layout);
+  auto field_it = each.begin();
+  for (size_t col = 0; col < slice.columns(); ++col, ++field_it) {
+    auto& type = field_it->type();
+    auto add_column = [&](const synopsis_ptr& syn) {
+      for (size_t row = 0; row < slice.rows(); ++row) {
+        auto view = slice.at(row, col, type);
+        if (!caf::holds_alternative<caf::none_t>(view))
+          syn->add(std::move(view));
+      }
+    };
+    auto key = qualified_record_field{layout.name(), *field_it};
+    if (!caf::holds_alternative<string_type>(type)) {
+      // Locate the relevant synopsis.
+      auto it = field_synopses_.find(key);
+      if (it == field_synopses_.end()) {
+        // Attempt to create a synopsis if we have never seen this key before.
+        it = field_synopses_.emplace(std::move(key), make_synopsis(type)).first;
+      }
+      // If there exists a synopsis for a field, add the entire column.
+      if (auto& syn = it->second)
+        add_column(syn);
+    } else { // type == string
+      field_synopses_[key] = nullptr;
+      auto cleaned_type = vast::type{field_it->type()}.attributes({});
+      auto tt = type_synopses_.find(cleaned_type);
+      if (tt == type_synopses_.end())
+        tt = type_synopses_.emplace(cleaned_type, make_synopsis(type)).first;
+      if (auto& syn = tt->second)
+        add_column(syn);
+    }
+  }
+}
+
+size_t partition_synopsis::size_bytes() const {
+  size_t result = 0;
+  for (auto& [field, synopsis] : field_synopses_)
+    result += synopsis ? synopsis->size_bytes() : 0ull;
+  return result;
+}
+
+} // namespace vast
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -606,12 +606,6 @@ index(index_actor::stateful_pointer<index_state> self,
       auto lookup = query_state{query_id, expr, std::move(candidates)};
       auto result = self->state.pending.emplace(query_id, std::move(lookup));
       VAST_ASSERT(result.second);
-      // NOTE: The previous version of the index used to do much more
-      // validation before assigning a query id; in particular it did
-      // evaluate the entries of the pending query map and checked that
-      // at least one of them actually produced an evaluation triple.
-      // However, the query_processor doesnt really care about the id
-      // anyways, so hopefully that shouldnt make too big of a difference.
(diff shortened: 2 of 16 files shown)


libvast/vast/system/partition.hpp
```diff
@@ -107,15 +107,13 @@ struct active_partition_state {
...
-  /// Actor handle of the index actor.
-  index_actor index;
-
   /// Actor handle of the filesystem actor.
   filesystem_actor filesystem;
 
   /// Promise that gets satisfied when the partition state was serialized
   /// and written to disk.
-  caf::typed_response_promise<atom::ok> persistence_promise;
+  caf::typed_response_promise<atom::ok, std::shared_ptr<partition_synopsis>>
```
dominiklohmann, line 115: Not sure why the ok atom is still needed here, just the partition_synopsis should be enough for the reply.

libvast/src/system/partition.cpp
```diff
@@ -483,15 +482,17 @@ active_partition_actor::behavior_type active_partition(
       self->state.streaming_initiated = true;
       return self->state.stage->add_inbound_path(in);
     },
-    [=](atom::persist, const path& part_dir, index_actor index) {
+    [=](atom::persist, const path& part_dir /*, index_actor index*/) {
```
tobim, line 485:
```suggestion
    [=](atom::persist, const path& part_dir) {
```

tobim (changes requested):

libvast/src/system/index.cpp
```diff
@@ -440,10 +440,15 @@ index(index_actor::stateful_pointer<index_state> self,
...
-    self->request(actor, caf::infinite, atom::persist_v, part_dir, self)
+    self->request(actor, caf::infinite, atom::persist_v, part_dir /*, self*/)
       .then(
-        [=](atom::ok) {
+        [=](atom::ok, std::shared_ptr<partition_synopsis>& ps) {
           VAST_DEBUG(self, "successfully persisted partition", id);
+          // Semantically ps is a unique_ptr, and the partition releases its
+          // copy before sending. We use shared_ptr for the transport because
+          // CAF message types must be copy-constructible.
+          VAST_ASSERT(ps.use_count() <= 1);
```
dominiklohmann, line 450: Shouldn't this check whether the shared_ptr is unique exactly? If the use count is 0, then you dereference a nullptr directly below.

libvast/src/system/index.cpp
```diff
@@ -585,6 +590,8 @@ index(index_actor::stateful_pointer<index_state> self,
       }
       // Get all potentially matching partitions.
       auto candidates = self->state.meta_idx.lookup(expr);
+      for (const auto& [id, _] : self->state.unpersisted)
```
dominiklohmann, line 593: Is this a gcc/clang extension that you don't need `[[maybe_unused]]` here?

libvast/src/system/partition.cpp
```diff
@@ -483,15 +482,17 @@ active_partition_actor::behavior_type active_partition(
...
       // Ensure that the response promise has not already been initialized.
       VAST_ASSERT(
         !static_cast<caf::response_promise&>(self->state.persistence_promise)
            .source());
-      self->state.index = index;
+      // self->state.index = index;
```
dominiklohmann, line 490: This line can just be removed.

libvast/src/system/partition.cpp
```diff
@@ -558,15 +559,21 @@ active_partition_actor::behavior_type active_partition(
...
-              self->state.persistence_promise.delegate(
-                self->state.filesystem, atom::write_v,
-                *self->state.persist_path, fbchunk);
+              // TODO: Add a proper timeout.
+              self
+                ->request(self->state.filesystem, caf::infinite, atom::write_v,
+                          *self->state.persist_path, fbchunk)
+                .then(
+                  [=](atom::ok) {
+                    // Relinquish ownership and send the shrinked synopsis to
```
dominiklohmann, line 568: s/shrinked/shunk

libvast/src/system/partition.cpp
```diff
@@ -580,6 +587,8 @@ active_partition_actor::behavior_type active_partition(
     },
     [=](const expression& expr,
         partition_client_actor client) -> caf::result<atom::done> {
+      // TODO: Add a `partition_synopsis::lookup()` function that we can use
+      // here if we still have a partition synopsis.
```
dominiklohmann, line 591: Please elaborate on this comment. It makes sense now, but might not make too much sense as is for the future self.

libvast/src/system/query_supervisor.cpp
```diff
@@ -53,7 +53,12 @@ query_supervisor_actor::behavior_type query_supervisor(
         const index_client_actor& client) {
       VAST_DEBUG(self, "got a new query for", qm.size(),
                  "partitions:", get_ids(qm));
-      VAST_ASSERT(!qm.empty());
+      if (qm.empty()) {
+        // Nothing to do.
+        self->send(client, atom::done_v);
+        self->send(master, atom::worker_v, self);
+        return;
+      }
```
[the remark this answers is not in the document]
dominiklohmann: 👍 and add an assertion here whether `!qm.empty()` with a comment why.

dominiklohmann (changes requested):

tobim (changes requested):
Can you please also remove the now dead code from the meta index and move `partition_synopsis` into its own `hpp` and `cpp`?

libvast/src/system/partition.cpp
lava: shrunken, actually

libvast/src/system/query_supervisor.cpp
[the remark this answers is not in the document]
lava: I'm not sure this is better; this makes the logic in the index much more complicated. (since we have to return `atom::done`, return the worker, and maybe erase the lookup handle if these were the last partitions)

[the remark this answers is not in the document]
dominiklohmann: Can you at least add a comment with a TODO for this question?

libvast/src/system/index.cpp
lava: Good question, I tried a bit locally and I couldn't get gcc/clang to emit a warning at all as long as part of a structured binding is used:

```
benno@poincare:~/code/cpp$ cat unused2.cpp 
#include <iostream>
#include <utility>
#include <vector>


int main() {
  for (auto& [x, _] : std::vector<std::pair<int,int>>{std::make_pair(3,4)})
	 std::cout << x << std::endl;
}
benno@poincare:~/code/cpp$ clang++ -std=c++17 unused2.cpp -Wall -Wextra
benno@poincare:~/code/cpp$ 
```

So I think the other instances of this in the codebase might just be unnecessary.

libvast/src/system/query_supervisor.cpp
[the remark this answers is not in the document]
lava: Sure, added a TODO-comment.

libvast/src/partition_synopsis.cpp
```diff
@@ -0,0 +1,75 @@
+#include "vast/partition_synopsis.hpp"
```
dominiklohmann, line 1: This is missing the banner.

libvast/src/partition_synopsis.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+size_t partition_synopsis::size_bytes() const {
+  size_t result = 0;
+  for (auto& [field, synopsis] : field_synopses_)
+    result += synopsis ? synopsis->size_bytes() : 0ull;
+  return result;
+}
+
+} // namespace vast
```
dominiklohmann, line 88: Missing newline at EOF.

libvast/vast/system/query_supervisor.hpp
```diff
@@ -38,7 +38,9 @@ struct query_supervisor_state {
   size_t open_requests;
 
   // Gives the query_supervisor a unique, human-readable name in log output.
-  std::string name;
+  std::string log_identifier;
+
+  static inline const char* name = "query_supervisor";
```
dominiklohmann, line 43:
```suggestion
  static inline const char* name = "query-supervisor";
```

libvast/vast/partition_synopsis.hpp
```diff
@@ -0,0 +1,43 @@
...
+  size_t size_bytes() const;
+
+  /// Synopsis data structures for types.
+  std::unordered_map<type, synopsis_ptr> type_synopses_;
+
+  /// Synopsis data structures for individual columns.
+  std::unordered_map<qualified_record_field, synopsis_ptr> field_synopses_;
+};
+
+} // namespace vast
```
dominiklohmann, line 43: Missing newline at EOF.

dominiklohmann (commented):
I think there are a few places where we include `vast/meta_index.hpp` where now including `vast/partition_synopsis.hpp` would suffice.

As an aside, shouldn't it really be `vast/system/partition_synopsis.hpp`?

lava:
@dominiklohmann I went through all includes of `meta_index.hpp`

> As an aside, shouldn't it really be vast/system/partition_synopsis.hpp?

I don't think so, this is a data structure and neither `meta_index` nor `table_slice` are in `vast/system/`.

dominiklohmann:
> > As an aside, shouldn't it really be vast/system/partition_synopsis.hpp?
> 
> I don't think so, this is a data structure and neither `meta_index` nor `table_slice` are in `vast/system/`.

Alright, fine by me.

dominiklohmann (approved):
This was reviewed thoroughly and all review feedback was incorporated. LGTM! 🚀

tobim (approved):
Code is in good shape. Can be merged form my point of view.

PR #1297, opened 2021-01-21
Add per partition memory usage information to vast status

tobim:

Collects memory usage statistics from partition actors and active indexers via the `status` message.

###  :memo: Checklist

By commit.

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -321,35 +321,34 @@ index_state::status(status_verbosity v) const {
     put(index_status, "num-cached-partitions", inmem_partitions.size());
     put(index_status, "num-unpersisted-partitions", unpersisted.size());
     auto& partitions = put_dictionary(index_status, "partitions");
-    auto partition_status
-      = [&](const uuid& id, const partition_actor& pa,
-            caf::config_value::list& xs) {
-          deferred = true;
-          self->request(pa, caf::infinite, atom::status_v, v)
-            .then(
-              [=, &xs](const caf::settings& part_status) {
-                auto& ps = xs.emplace_back().as_dictionary();
-                put(ps, "id", to_string(id));
-                if (auto s = caf::get_if<caf::config_value::integer>(
-                      &part_status, "memory-usage"))
-                  req_state->memory_usage += *s;
-                if (v >= status_verbosity::debug)
-                  detail::merge_settings(part_status, ps);
-                // Both handlers have a copy of req_state.
-                if (req_state.use_count() == 2)
-                  deliver(std::move(*req_state));
-              },
-              [=, &xs](const caf::error& err) {
-                VAST_WARNING(self, "failed to retrieve status from", id, ":",
-                             render(err));
-                auto& ps = xs.emplace_back().as_dictionary();
-                put(ps, "id", to_string(id));
-                put(ps, "error", render(err));
-                // Both handlers have a copy of req_state.
-                if (req_state.use_count() == 2)
-                  deliver(std::move(*req_state));
-              });
-        };
+    auto partition_status = [&](const uuid& id, const partition_actor& pa,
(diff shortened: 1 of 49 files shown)


libvast/src/system/index.cpp
```diff
@@ -247,23 +261,59 @@ index_state::status(status_verbosity v) const {
...
     }
-    put(stats_object, "meta-index-bytes", meta_idx.size_bytes());
-  }
-  if (v >= status_verbosity::debug) {
-    // Resident partitions.
+    put(index_status, "meta-index-bytes", meta_idx.size_bytes());
+    put(index_status, "num_active_partitions",
+        active_partition.actor == nullptr ? 0 : 1);
+    put(index_status, "num_cached_partitions", inmem_partitions.size());
+    put(index_status, "num_unpersisted_partitions", unpersisted.size());
```
dominiklohmann, line 268: Can you `s/_/-/g` on all these?

libvast/src/system/partition.cpp
```diff
@@ -704,6 +753,29 @@ partition_actor::behavior_type passive_partition(
...
+      for (size_t i = 0; i < st.indexers.size(); ++i) {
+        if (st.indexers[i])
+          mem_indexers
+            += sizeof(indexer_state)
+               + st.flatbuffer->indexes()->Get(i)->index()->data()->size();
+      }
+      caf::put(result, "memory-usage-indexers", mem_indexers);
+      auto x = st.partition_chunk->incore();
+      if (!x) {
+        caf::put(result, "memory-usage-incore", render(x.error()));
```
dominiklohmann, line 770: We shouldn't render errors inline with results. Shouldn't the whole handler reply with the error instead, possibly enriched with some information about the indexer? Then we can print the error at the outermost status request.

tobim: Then the failure of one indexer would shadow the data from the whole component. I reusing the key in a "value or error" style is best.
The alternatives are:
 - omitting the key, and inserting the error string under the `error` key. 
 - Setting the key to a neutral value and adding the error under the `error` key.
Both have 2 drawbacks: If 2 keys in the same JSON node can fail to report, we can only insert the error for 1 of them. Second, scripts for automatic evaluation might use a `select()` mechanism to retrieve the key from the object. This would silently fail and lead to wrong information in the generated report.

dominiklohmann: Let's put this on hold until we implement `ch20023`.

libvast/vast/chunk.hpp
```diff
@@ -126,6 +126,9 @@ class chunk final : public caf::ref_counted {
   /// @returns The size of the chunk.
   size_type size() const noexcept;
 
+  /// @returns The size of the chunk.
+  caf::expected<chunk::size_type> incore() const noexcept;
```
dominiklohmann, line 130: The comment needs to be updated.

libvast/vast/ewah_bitmap.hpp
```diff
@@ -102,6 +102,8 @@ class ewah_bitmap : public bitmap_base<ewah_bitmap>,
 
   size_type size() const;
 
+  size_t memusage() const;
```
dominiklohmann, line 105: If I recall correctly, then we use `num_bytes` in some places, `size_bytes` in others, and now also `memusage`. Can we consolidate these?

tobim: `num_bytes` is used for different purposes, but `size_bytes()` seems to fulfill the same role as `memusage()`. From my perspective `memusage()` is a more descriptive name for the intent of the function. So I will rename `size_bytes()`.

dominiklohmann: Fine by me.

dominiklohmann (approved):
Works very well on my machine, and the code lgtm as well! 🚀

PR #1301, opened 2021-01-22
Improve plugin developer experience

mavam:

This PR streamlines the DX of `configure` when building with plugins.

###  :memo: Checklist

In one shot.


tobim (approved):
Looks reasonable.

mavam:
We could add that `--with-plugin` can be specified multiple times. (I guess it's the only option that works that way?)

PR #1313, opened 2021-01-25
Move the meta index into a separate actor

dominiklohmann:

This moves the meta index into an actor of its own.

###  :memo: Checklist

Commit-by-commit.


libvast/src/system/default_configuration.cpp
```diff
@@ -18,14 +18,32 @@
...
+  // However, there is a possible deadlock: [...]
+  //  1) Increase `stream.max-batch-delay`. As long as the total cost
+  //  2) Set a finite, small `scheduler.max-throughput`. This should empirically
+  //     be enough, but it relies on the CAF scheduler to actually schedule the
+  //     downstream actor, which empirically can take quite a while at times.
+  //     Also, if this is too small, there is a chance that a very busy stream
+  //     can drown out an actor completely so that no other work can take place.
+  //  3) Increase `scheduler.max-threads`. When there are more threads than
+  //     deadlocked actors, eventually the downstream actors will be able to
+  //     make progress and the deadlock can be resolved.
```
tobim: The formulation in the comment is confusing, do you need to adjust the settings as described in 1,2,3 for the deadlock to surface or to mitigate it?

libvast/src/system/index.cpp
```diff
@@ -654,16 +654,15 @@ index(index_actor::stateful_pointer<index_state> self,
...
       }
-      VAST_DEBUG("index finalized streaming");
     });
   self->set_exit_handler([self](const caf::exit_msg& msg) {
     VAST_DEBUG("{} received EXIT from {} with reason: {}", self, msg.source,
                msg.reason);
     // Flush buffered batches and end stream.
+    self->state.stage->shutdown(); // closes inbound paths
     self->state.stage->out().fan_out_flush();
+    self->state.stage->out().close(); // closes outbound paths
```
tobim, line 670: It would be good if we had a better understanding of what these do and why the order matters.
Maybe it also makes sense to put this sequence into it's own function.

libvast/src/system/partition.cpp
```diff
@@ -407,20 +408,21 @@ active_partition_actor::behavior_type active_partition(
           VAST_DEBUG("{} spawned new indexer for field {} at slot {}", self,
                      field.name, slot);
         }
+        VAST_WARN("pushing column {}  {}", qf,
+                  self->state.stage->out().capacity());
```
tobim, line 412: Remove.

libvast/src/system/default_configuration.cpp
```diff
@@ -18,14 +18,32 @@
...
   using caf::atom;
+  using namespace std::chrono_literals;
   set("logger.component-blacklist",
       caf::make_config_value_list(atom("caf"), atom("caf_flow"),
                                   atom("caf_stream")));
+  // TODO: Move part of this text into the commit message
+  // The `max-batch-delay` is the maximum amount of time that caf waits until
+  // it attempts to send an underfull batch along the stream.
+  // However, there is a possible deadlock: [...]
+  //  1) Increase `stream.max-batch-delay`. As long as the total cost
```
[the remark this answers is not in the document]
tobim: Ah, too late.

libvast/src/system/default_configuration.cpp
```diff
@@ -18,14 +18,43 @@
...
+  //     be enough, but it relies on the CAF scheduler to actually schedule the
+  //     downstream actor, which empirically can take quite a while at times.
+  //     Also, if this is too small, there is a chance that a very busy stream
+  //     can drown out an actor completely so that no other work can take place.
+  //  3) Increase `scheduler.max-threads`. When there are more threads than
+  //     deadlocked actors, eventually the downstream actors will be able to
+  //     make progress and the deadlock can be resolved.
+  // We do a combination of (1) and (2).
+  set("stream.max-batch-delay", caf::timespan{15ms});
+  set("scheduler.max-throughput", 500);
```
dominiklohmann, line 57: These must be reflected in vast.yaml.example.

libvast/src/system/index.cpp
lava: Yes, good point. I actually think the final `force_emit_batches()` is not needed at all, since caf should automatically try to force underfull batches after the `max-batch-delay` of 5ms has run out.

lava: On second thought, I think it would be better to make this refactoring in a separate topic branch: A function in full generality would add an additional call to `shutdown()` to some places, and while I think it is correct it will give me additional peace of mind to know that this can be reverted independently in case it breaks anything.

libvast/src/system/indexer.cpp
```diff
@@ -74,6 +74,7 @@ active_indexer(active_indexer_actor::stateful_pointer<indexer_state> self,
       auto result = caf::attach_stream_sink(
         self, in,
         [=](caf::unit_t&) {
+          VAST_DEBUG("{} initializes stream", self);
```
tobim, line 77: This can still happen hundreds of times per partition.
```suggestion
          VAST_TRACE("{} initializes stream", self);
```

libvast/vast/system/actors.hpp
```diff
@@ -387,4 +400,8 @@ CAF_BEGIN_TYPE_ID_BLOCK(vast_actors, caf::id_block::vast_atoms::end)
 
 CAF_END_TYPE_ID_BLOCK(vast_actors)
 
+using vast_uuid_synopsis_map = std::map<vast::uuid, vast::partition_synopsis>;
+CAF_ALLOW_UNSAFE_MESSAGE_TYPE(std::shared_ptr<vast_uuid_synopsis_map>)
+CAF_ALLOW_UNSAFE_MESSAGE_TYPE(std::shared_ptr<vast::partition_synopsis>)
```
tobim, line 405: I assume these don't get type ids because they should never be sent over the wire? A small comment would go a long way.

I'm also not sure whether this file is the correct place for these statements. Shouldn't we think of this as implementation detail of the synopsis / meta index?

libvast/test/flatbuffers.cpp
```diff
@@ -161,15 +161,15 @@ TEST(empty partition roundtrip) {
...
-  CHECK_EQUAL(candidates[0], state.id);
+  // vast::system::meta_index recovered_meta_idx;
+  // recovered_meta_idx.merge(state.id, std::move(ps));
+  // // Check that lookups still work as expected.
+  // auto candidates = recovered_meta_idx.lookup(vast::expression{
+  //   vast::predicate{vast::field_extractor{".x"},
+  //                   vast::relational_operator::equal, vast::data{0u}},
+  // });
+  // REQUIRE_EQUAL(candidates.size(), 1u);
+  // CHECK_EQUAL(candidates[0], state.id);
```
tobim, line 172: What did this test check, and why is it in this file?
If the test doesn't make sense any more can you remove it and check if the include can also be removed?

tobim (changes requested):
This looks good and I already verified that it works in practice. I have a few minor change requests, nothing major.

libvast/vast/system/actors.hpp
lava: I think they need to be here since they're part of the interface for the `meta_index_actor`, so anyone who wants to send a message to the meta index needs to see these declarations.

tobim (approved):

PR #1315, opened 2021-01-26
Fix usage of simdjson::simdjson_result<T>

dominiklohmann:

Before simdjson 0.8.0, `simdjson::simdjson_result<T>` inherited from `std::pair<T, Error>` publicly, which allowed us to use it in a possibly unintended way. The type no longer decomposes into structured bindings, so we now use `.error()` and `.value()` respectively where appropriate.

###  :memo: Checklist

File-by-file.


tobim (approved):
CI will verify that `v0.7` still works.

PR #1319, opened 2021-01-27
Make sure the parent dir for the test link exists

tobim:

I came across a build error because the directory `/build/share/vast` did not exist when the symlink for the integration test directory was created. I don't know how the order of those targets is determined, but it is unnecessary to rely on implicit assumptions here.


dominiklohmann (approved):
The order does indeed seem to be unspecified. This fix is easier than ensuring a fixed order, so let's go with it.

PR #1323, opened 2021-01-29
Use spdlog for logging

lava:

Switch to spdlog as our logging framework

###  :memo: Checklist

- [x] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary: https://github.com/tenzir/event-horizon/pull/814

Look at the commits to understand the overall structure of the rewrite.

Then,
    - for the mechanical/bulk commits, do some stochastic sampling if they look reasonable
    - for the rest, review as usual


.github/workflows/vast.yaml
```diff
@@ -65,6 +65,9 @@ jobs:
           apt-get -qq update
           apt-get -qqy install git libsnappy-dev libbrotli-dev zlib1g-dev wget liblz4-dev libzstd-dev ninja-build libpcap-dev libssl-dev libbenchmark-dev tcpdump lsb-release libflatbuffers-dev flatbuffers-compiler-dev libyaml-cpp-dev libsimdjson-dev gnupg2 gcc-8 g++-8 python3 python3-pip python3-venv jq apt-transport-https ca-certificates curl gnupg-agent software-properties-common
 
+          # spdlog
+          apt-get -qqy -t buster-backports install libspdlog-dev
```
dominiklohmann, line 69: No need to do this with `-t buster-backports` now, the Dockerfile is running on Buster-backports now generally.

Dockerfile
```diff
@@ -25,6 +25,9 @@ RUN echo "deb http://www.deb-multimedia.org buster main" | tee -a /etc/apt/sourc
   apt-get -qq update
 RUN apt-get -qqy install libflatbuffers-dev flatbuffers-compiler-dev
 
+# spdlog
+RUN apt-get -qqy -t buster-backports install libspdlog-dev
```
dominiklohmann, line 29: Again, you can drop the `-t buster-backports`.

Dockerfile
```diff
@@ -76,14 +79,16 @@ COPY --from=build_type $PREFIX/ $PREFIX/
 RUN apt-get -qq update && apt-get -qq install -y libc++1 libc++abi1 libpcap0.8 \
   openssl libsimdjson4 libyaml-cpp0.6 libasan5 libflatbuffers1 wget gnupg2
 
+# Need to specify backports, since spdlog also has a regular buster package.
+RUN apt-get -qqy -t buster-backports install libspdlog1
```
dominiklohmann, line 83: See above.

configure
```diff
@@ -245,6 +246,9 @@ while [ $# -ne 0 ]; do
     --with-simdjson=*)
       append_cache_entry SIMDJSON_ROOT PATH "$optarg"
       ;;
+    --with-spdlog=*)
+      append_cache_entry spdlog_ROOT PATH "$optarg"
```
dominiklohmann, line 250: The CMake variable appears to be named `SPDLOG_ROOT`.

examples/plugins/example/example.cpp
```diff
@@ -43,23 +43,24 @@ example_actor::behavior_type
 example(example_actor::stateful_pointer<example_actor_state> self) {
   return {
     [=](atom::config, record config) {
-      VAST_TRACE(self, "sets configuration", config);
+      VAST_TRACE("{} sets configuration {}", detail::id_or_name(self), config);
```
dominiklohmann, line 46: Do we always need to pass `id_or_name`, or can we make some use of SFINAE to apply `id_or_name` to every passed actor handle?

libvast/src/logger.cpp
```diff
@@ -68,25 +58,191 @@ int loglevel_to_int(caf::atom_value x, int default_value) {
...
+      level = spdlog::level::warn;
+      break;
+    case VAST_LOG_LEVEL_INFO:
+      level = spdlog::level::info;
+      break;
+    case VAST_LOG_LEVEL_VERBOSE:
+      level = spdlog::level::debug;
+      break;
+    case VAST_LOG_LEVEL_DEBUG:
+      level = spdlog::level::debug;
```
dominiklohmann, line 102: DIdn't we say that this was supposed to be trace? (might be wrong here, but please double-check)

vast.yaml.example
```diff
@@ -9,16 +9,43 @@ vast:
...
+  # vast start). The levels above 'verbose' are usually not available in
+  # release builds.
+  # file-verbosity: debug
+
+  # Mode for console log output generation.
+  # Possible values: always, automatic, never. (default automatic)
+  # console: automatic
+
+  # Format for printing individual log entries to the console.
+  # For a list of valid format specifiers, see spdlog format specification.
```
dominiklohmann, line 37: Can you add a link here?

vast.yaml.example
```diff
@@ -9,16 +9,43 @@ vast:
...
+
+  # Configures the minimum severity of messages written to the log file.
+  # Possible values: quiet, error, warning, info, verbose, debug, trace.
+  # File logging is only available for commands that start a node (e.g.,
+  # vast start). The levels above 'verbose' are usually not available in
+  # release builds.
+  # file-verbosity: debug
+
+  # Mode for console log output generation.
+  # Possible values: always, automatic, never. (default automatic)
```
dominiklohmann, line 33: What does automatic mean?

vast/vast.cpp
```diff
@@ -164,43 +165,47 @@ int main(int argc, char** argv) {
...
     return EXIT_SUCCESS;
   }
-  // Initialize actor system (and thereby CAF's logger).
-  if (!init_config(cfg, *invocation, std::cerr))
+
+  // since the logger is not dependend on caf, this could be moved up
+  auto log_context = vast::create_log_context(*invocation, cfg.content);
+  if (!log_context)
     return EXIT_FAILURE;
+
```
dominiklohmann, line 173: Please strip the newlines from this file.

vast/vast.cpp
```diff
@@ -164,43 +165,47 @@ int main(int argc, char** argv) {
     // success when printing the help/documentation texts.
     return EXIT_SUCCESS;
   }
-  // Initialize actor system (and thereby CAF's logger).
-  if (!init_config(cfg, *invocation, std::cerr))
+
+  // since the logger is not dependend on caf, this could be moved up
```
dominiklohmann, line 169: Please correct the comment and move this to as early as possible in this function, with a comment explaining why this is the earliest possible point.

libvast/vast/logger.hpp
```diff
@@ -15,189 +15,69 @@
...
-#include "vast/detail/pp.hpp"
-#include "vast/detail/type_traits.hpp"
-
-#include <caf/detail/pretty_type_name.hpp>
-#include <caf/logger.hpp>
+#include "vast/error.hpp"
+
+// VAST_INFO -> spdlog::info
+// VAST_VERBOSE -> spdlog::debug
+// VAST_DEBUG -> spdlog::trace
```
dominiklohmann, line 22: See other comment on this: The current implementation has debug -> debug mapping.

libvast/vast/logger.hpp
```diff
@@ -15,189 +15,69 @@
...
+#  define SPDLOG_ACTIVE_LEVEL SPDLOG_LEVEL_CRITICAL
+#elif VAST_LOG_LEVEL == VAST_LOG_LEVEL_QUIET
+#  define SPDLOG_ACTIVE_LEVEL SPDLOG_LEVEL_OFF
+#endif
+
+// Important: keep that below the log level mapping
+#include "vast/detail/logger.hpp"
+#include "vast/detail/logger_formatters.hpp"
+
+#define VAST_TRACE(...) SPDLOG_LOGGER_TRACE(vast::detail::logger(), __VA_ARGS__)
```
dominiklohmann, line 47: We lost the ability to use trace for marking ENTER and EXIT of scopes. Can you add this back?

libvast/vast/detail/logger_formatters.hpp
```diff
@@ -0,0 +1,486 @@
...
+template <class T>
+struct fmt::formatter<caf::expected<T>> {
+  template <typename ParseContext>
+  constexpr auto parse(ParseContext& ctx) {
+    return ctx.begin();
+  }
+
+  template <typename FormatContext>
+  auto format(const caf::expected<T>& item, FormatContext& ctx) {
+    return format_to(ctx.out(), "{}", caf::deep_to_string(item));
```
dominiklohmann, line 339: Shouldn't this go through the formatter for T instead of `caf::deep_to_string`? The same goes for `caf::optional<T>`.

dominiklohmann (changes requested):

.github/workflows/vast.yaml
lava: The problem is that `libspdlog` has packages in both backports and the regular buster, and by default the non-backport one is selected:

```
root@5ab8414bcaf8:/# apt-cache policy libspdlog-dev
libspdlog-dev:
  Installed: (none)
  Candidate: 1:1.3.1-1
  Version table:
     1:1.5.0+ds-4~bpo10+1 100
        100 http://deb.debian.org/debian buster-backports/main amd64 Packages
     1:1.3.1-1 500
        500 http://deb.debian.org/debian buster/main amd64 Packages
root@5ab8414bcaf8:/# apt-get install libspdlog-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following NEW packages will be installed:
  libspdlog-dev
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 176 kB of archives.
After this operation, 651 kB of additional disk space will be used.
Get:1 http://deb.debian.org/debian buster/main amd64 libspdlog-dev amd64 1:1.3.1-1 [176 kB]
[...]
```

vast.yaml.example
lava: Color enabled when writing to a tty, disabled otherwise.

examples/plugins/example/example.cpp
lava: I think we will ultimately solve this by having formatters that can handle the various kinds of actor handles directly, but I would propose to do this as a separate refactoring.

vast.yaml.example
dominiklohmann: Should be explained in the comment.

examples/plugins/example/example.cpp
dominiklohmann: Yes. Ideally we'd get rid of the need to use the `caf::stringification_inspector` (used by `caf::deep_to_string`) as much as possible.

.github/workflows/vast.yaml
dominiklohmann: Can you add a comment about this to the Dockerfile for the future maintainer?

CHANGELOG.md
```diff
@@ -11,6 +11,19 @@ This changelog documents all notable user-facing changes of VAST.
...
+## Unreleased
+
+- 🎁 VAST rotates server logs by default.
+  [#1223](https://github.com/tenzir/vast/pull/1223)
+
+- ⚡️ VAST switched to spdlog as logging library. For users, this means:
+  The `vast.console-format` and `vast.file-format` now must be specified using the
+  spdlog pattern syntax as described
+  `[here](https://github.com/gabime/spdlog/wiki/3.-Custom-formatting#pattern-flags).
+  All settings under `logger.*` are now ignored by VAST, and only the `vast.*`
```
dominiklohmann, line 23: `s/logger/caf.logger`

libvast/vast/detail/logger_formatters.hpp
dominiklohmann:
```suggestion
    if (item)
      return format_to(ctx.out(), "{}", *item);
    else
      return format_to(ctx.out(), "{}", item.error());
```

lava: I think we'd have to agree first how we want to format the nonopt/error case. Let's do this separately.

dominiklohmann: Fine by me.

libvast/vast/logger.hpp
lava: The comment is correct, I updated the implementation.

dominiklohmann (approved):

PR #1325, opened 2021-01-31
Improve access of typed versions of components

dominiklohmann:

This commit
- replaces all usage of the untyped version of get_node_components with
  the typed version,
- removes the untpyed version of get_node_components entirely,
- introduces a new typed find member function to the component registry,
  and
- replaces calls to find_by_label with calls to the new typed find where
  possible.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -90,6 +90,14 @@ auto make_error_msg(ec code, std::string msg) {
 /// Helper function to determine whether a component can be spawned at most
 /// once.
 bool is_singleton(std::string_view type) {
+  // TODO: All of these actor interfaces are strongly typed. The value of `type`
+  // is received via the actor interface of the NODE sometimes, which means that
+  // we cannot just pass arbitrary actors to it. Atoms aren't really an option
+  // either, because `caf::atom_value` was removed with CAF 0.18. We can,
+  // however, abuse the fact that every typed actor has a type ID assigned, and
+  // change the node to work with type IDs over actor names everywhere. This
+  // refactoring will be much easier once the NODE itself is a typed actor, so
+  // let's hold off until then.
   const char* singletons[]
     = {"accountant", "archive", "eraser",       "filesystem",
        "importer",   "index",   "type-registry"};
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -184,8 +184,7 @@ void collect_component_status(node_actor* self,
 caf::message dump_command(const invocation& inv, caf::actor_system&) {
   auto as_yaml = caf::get_or(inv.options, "vast.dump.yaml", false);
   auto self = this_node;
-  auto type_registry = caf::actor_cast<type_registry_actor>(
-    self->state.registry.find_by_label("type-registry"));
+  auto [type_registry] = self->state.registry.find<type_registry_actor>();
   if (!type_registry)
     return caf::make_message(caf::make_error(ec::missing_component, //
                                              "type-registry"));
@@ -498,16 +497,16 @@ node_state::spawn_command(const invocation& inv,
   if (query_handlers.count(comp_type) > 0u
       && !caf::get_or(spawn_inv.options,
                       "vast." + comp_type + ".disable-taxonomies", false)) {
-    if (auto tr = self->state.registry.find_by_label("type-registry")) {
+    if (auto [type_registry] = self->state.registry.find<type_registry_actor>();
+        type_registry) {
       auto expr = normalized_and_validated(spawn_inv.arguments);
       if (!expr) {
         rp.deliver(expr.error());
         return make_message(expr.error());
       }
       self
-        ->request(caf::actor_cast<type_registry_actor>(tr),
-                  defaults::system::initial_request_timeout, atom::resolve_v,
-                  std::move(*expr))
+        ->request(type_registry, defaults::system::initial_request_timeout,
+                  atom::resolve_v, std::move(*expr))
         .then(handle_taxonomies, [=](caf::error err) mutable {
           rp.deliver(err);
           return make_message(err);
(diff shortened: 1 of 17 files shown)


libvast/vast/detail/actor_cast_wrapper.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include "vast/fwd.hpp"
+
+#include <caf/actor_cast.hpp>
+
+#include <utility>
+
+namespace vast::detail {
+
+struct actor_cast_wrapper {
```
mavam, line 24: Can you briefly explain in a comment what this does and/or give an example?

libvast/src/system/node.cpp
```diff
@@ -567,13 +566,13 @@ caf::behavior node(node_actor* self, std::string name, path dir,
     };
     // Terminate the accountant first because it acts like a source and may
     // hold buffered data.
-    if (auto accountant = registry.find_by_label("accountant"))
-      schedule_teardown(std::move(accountant));
+    if (auto [accountant] = registry.find<accountant_actor>(); accountant)
+      schedule_teardown(caf::actor_cast<caf::actor>(std::move(accountant)));
     // Take out the filesystem, which we terminate at the very end.
-    auto filesystem = registry.find_by_label("filesystem");
+    auto [filesystem] = registry.find<filesystem_actor>();
```
mavam, line 572: It feels like you could if-constexpr on the size of the `Ts` and then either return a single value or a tuple, as before. The singleton-wrapping feels a bit awkward.

libvast/src/system/node.cpp
```diff
@@ -519,7 +518,7 @@ caf::behavior node(node_actor* self, std::string name, path dir,
                    std::chrono::milliseconds shutdown_grace_period) {
   self->state.name = std::move(name);
   self->state.dir = std::move(dir);
-  // Initialize component and command factories.
+  // Initialize component and command factories.,
```
mavam, line 521: Nope 😉

libvast/vast/system/node_control.hpp
```diff
@@ -36,43 +37,11 @@ spawn_at_node(caf::scoped_actor& self, caf::actor node, Arguments&&... xs) {
...
-    return caf::actor_cast<Out>(std::forward<In>(in));
-  }
-};
-
 /// Look up components by their typed actor interfaces. Returns the first actor
 /// of each type passed as template parameter.
 template <class... Actors>
 caf::expected<std::tuple<Actors...>>
-get_typed_node_components(caf::scoped_actor& self, const caf::actor& node) {
+get_node_components(caf::scoped_actor& self, const caf::actor& node) {
```
mavam, line 44: I wish we could have a better name for this function, but I can't come up with one...

libvast/vast/system/component_registry.hpp
```diff
@@ -79,6 +82,37 @@ class component_registry {
...
 
+  /// Locates typed components by handle.
+  /// @tparam Handles... The typed actor handles to lookup.
+  /// @returns The respective component actors if found.
+  template <class... Handles>
+  std::tuple<Handles...> find() const {
+    auto normalize = [](std::string in) {
+      // Remove the uninteresting parts of the name:
+      //   vast::system::type_registry_actor -> type_registry
+      in.erase(0, sizeof("vast::system::") - 1);
```
mavam, line 93: This computation requires some assertions, e.g., `starts_with(str, "vast::system::")`.

Unless the registration with labe also works as dual to this transformation, I don't feel comfortable with this method. @tobim?

mavam (commented):
Nice changes from a user-perspective.

The type-name-to-label mapping is a bit fishy and needs to made explicit. The inverse must also exist at the registration site for this to be safe. I would just write a trait that gives you the label for a given component actor type.

libvast/vast/system/component_registry.hpp
dominiklohmann: Note that this transformation code was written by @tobim originally. Let's talk about this tomorrow briefly. I think it's fine as is for now and we can change it up when making the node itself a typed actor.

mavam: Good to know. I think it's actually not complicated to switch from

```cpp
const char* singletons[]
    = {"accountant", "archive", "eraser",       "filesystem",
       "importer",   "index",   "type-registry"};
```

to a type list in the node. My main worry is that these go out of sync and then we're in trouble.

I agree it's best to chat about this briefly.

libvast/src/system/node.cpp
dominiklohmann: It's awkward here, but doing one-offs is just extra maintenance and makes working with it in generic contexts non-trivial. I'd rather always keep it a tuple.

libvast/vast/detail/actor_cast_wrapper.hpp
dominiklohmann:
```cpp
actor_cast_wrapper{}
```

is essentially this from C++20:

```cpp
[]<typename Out>(auto&& in) {
  return caf::actor_cast<Out>(
    std::forward<decltype(in)>(in));
}
```

I'll add that as a comment with a note to replace once switching to C++20.

libvast/src/system/node.cpp
dominiklohmann: I suppose this is what happens if you set `,` as leader. 🙈

libvast/vast/system/component_registry.hpp
tobim: I agree with the general sentiment that these should be generated from the same expressions or checked statically. 

> My main worry is that these go out of sync and then we're in trouble.

When one changes one but not the other during development we'll see some failing unit or integration tests. That won't give a good indication of what is wrong, and makes the dev waste time on debugging.

tobim: @dominiklohmann is it feasible to generate `singletons` from the type ids and centralize (or drop?) the normalization?

libvast/vast/detail/actor_cast_wrapper.hpp
tobim: It's quite annoying that we have to add those one-offs because of a lack of orthogonality in the language.

dominiklohmann: Even more so because these cannot be defined anonymously within functions. 😤

libvast/vast/system/component_registry.hpp
dominiklohmann: Let's do this only after we converted the node to a typed actor interface. It's almost impossible to catch all possible issues with this currently.

dominiklohmann: I added a comment above `singletons` that explains what we should do here in the future.

libvast/vast/system/node_control.hpp
dominiklohmann: Let's not change that within this PR.

libvast/src/system/node.cpp
dominiklohmann: @tobim seems to agree with me here—marking this as resolved.

libvast/vast/system/component_registry.hpp
dominiklohmann: @tobim does my comment suffice? Can you re-review?

mavam: > Let's do this only after we converted the node to a typed actor interface.

Alright, fine with me. Can you take care of creating a corresponding story?

dominiklohmann: Will do before merging this.

dominiklohmann: Two stories exist now as a follow-up. @tobim / @mavam mind approving (unless you want anything else changed)?

tobim (approved):
Reducing the risk for error is always welcome.

PR #1326, opened 2021-02-01
Fix keyserver receive error

0snap:

- Fetch keys via cURL (avoids keyserver problems and IPv6 issues altogether)
- Remove custom key installation where possible (`flatbuffers`)

###  :memo: Checklist

File-by-file.

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -69,8 +69,8 @@ jobs:
           apt-get -qqy -t buster-backports install libspdlog-dev
 
           # clang 9 (c.f. https://apt.llvm.org/)
+          curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
           echo "deb http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main" | tee -a /etc/apt/sources.list
-          apt-key adv --keyserver hkp://pool.sks-keyservers.net:80 --recv-keys 15CF4D18AF4F7421
           apt-get -qq update
           apt-get -qqy install clang-9
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,16 +14,10 @@ ENV BUILD_DIR /tmp/src
 # Compiler and dependency setup
 RUN apt-get -qq update && apt-get -qqy install \
   build-essential gcc-8 g++-8 ninja-build libbenchmark-dev libpcap-dev tcpdump \
-  libssl-dev python3-dev python3-pip python3-venv git-core jq gnupg2 \
-  libyaml-cpp-dev libsimdjson-dev wget
+  libssl-dev python3-dev python3-pip python3-venv git-core jq gnupg2 wget \
+  libyaml-cpp-dev libsimdjson-dev libflatbuffers-dev flatbuffers-compiler-dev
 RUN pip3 install --upgrade pip && pip install --upgrade cmake && \
   cmake --version
-
-# flatbuffers
-RUN echo "deb http://www.deb-multimedia.org buster main" | tee -a /etc/apt/sources.list && \
-  apt-key adv --keyserver hkp://pool.sks-keyservers.net:80 --recv-keys 5C808C2B65558117 && \
-  apt-get -qq update
-RUN apt-get -qqy install libflatbuffers-dev flatbuffers-compiler-dev
 
 # Need to specify backports, since spdlog also has a regular buster package.
 RUN apt-get -qqy -t buster-backports install libspdlog-dev


.github/workflows/vast.yaml
```diff
@@ -63,14 +63,16 @@ jobs:
...
           apt-get -qq update
-          apt-get -qqy install git libsnappy-dev libbrotli-dev zlib1g-dev wget liblz4-dev libzstd-dev ninja-build libpcap-dev libssl-dev libbenchmark-dev tcpdump lsb-release libflatbuffers-dev flatbuffers-compiler-dev libyaml-cpp-dev libsimdjson-dev gnupg2 gcc-8 g++-8 python3 python3-pip python3-venv jq apt-transport-https ca-certificates curl gnupg-agent software-properties-common
+          apt-get -qqy install git libsnappy-dev libbrotli-dev zlib1g-dev wget liblz4-dev libzstd-dev ninja-build libpcap-dev libssl-dev libbenchmark-dev tcpdump lsb-release libflatbuffers-dev flatbuffers-compiler-dev libyaml-cpp-dev libsimdjson-dev gnupg2 gcc-8 g++-8 python3 python3-pip python3-venv jq apt-transport-https ca-certificates curl gnupg-agent software-properties-common dirmngr
 
           # Need to specify backports, since spdlog also has a regular buster package.
           apt-get -qqy -t buster-backports install libspdlog-dev
 
           # clang 9 (c.f. https://apt.llvm.org/)
+          mkdir -p ~/.gnupg
+          echo "disable-ipv6" >> ~/.gnupg/dirmngr.conf
```
dominiklohmann, line 73: Can you add a comment above this line _why_ we want to disable IPv6? This is not covered on `apt.llvm.org` as indicated by the comment.

dominiklohmann (changes requested):
Can you provide review instructions? How can I test that this fixes the issue we've been seeing? Why does it fix the issue we've been seeing?

.github/workflows/vast.yaml
```diff
@@ -69,8 +69,8 @@ jobs:
           apt-get -qqy -t buster-backports install libspdlog-dev
 
           # clang 9 (c.f. https://apt.llvm.org/)
+          curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
```
dominiklohmann, line 72: This now deviates from the link above, so please add a comment why that is. I know now, but the future maintainer of this file may not know from memory.

0snap: The link from the previous discussion is not contained in the file anymore.

dominiklohmann: Yes, but the code we run is not at https://apt.llvm.org, so please update that comment.

0snap: I don't understand your comment. The code is right on the page you linked, except that we use `cURL` instead of `wget` as they do.
> To retrieve the archive signature:
> wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
> # Fingerprint: 6084 F3CF 814B 57C1 CF12 EFD5 15CF 4D18 AF4F 7421

dominiklohmann: That caused issues, right? Please add a comment that explains what those issues were so we don't run into them again in the future.

0snap: No, using `wget` did not cause any issues. We used to call `apt-key adv <keyserver>` directly before this change (see diff to master). That results in `apt-key` to use gpg2 to fetch the key. And that indeterministically errored, probably due to IPv6 issues. Since we now plainly use `cURL` and then pass the already downloaded key to `apt-key` we circumvent the problem with the unresolvable keyserver alltogether.

dominiklohmann:
```suggestion
          # We used to call apt-key adv <keyserver> directly, which makes
          # apt-key use gpg2 to fetch the key.  That indeterministically
          # errored, probably due to IPv6 issues in CI. We now use cURL to
          # fetch the key, and pass the already downloaded key to apt-key. That
          # way, unresolvable keyservers are no longer an issue.
          curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
```

0snap: We would need to put that kind of comment above every place where we pull custom keys. That is two times alone in this workflow, more in the other workflows, Dockerfiles etc. Not using `gpg` / `apt-key` directly is a general learning and can go into our internal docs but there is no single right place to put that kind of comment in the codebase.

dominiklohmann: Having it somewhere is better than having it nowhere at all. Where would you put it?

dominiklohmann (approved):
Approving since we moved the discussion elsewhere.

PR #1328, opened 2021-02-01
Improve and generalize {fmt} formatters

dominiklohmann:

See `logger_formatters.hpp`:
- We now have fallback formatters based on printable and the stringification inspector.
- id_or_name is now applied automatically.
- Some log messages had their formatting corrected.
- render is now automatically called on errors.
- `<unprintable>` is no more.

###  :memo: Checklist

File-by-file.


lava (approved):
I admittedly only sampled the diff, but what I've seen looks reasonable, and with a PR this size I think its more efficient to merge now and maybe follow-up on issues that we overlooked, rather than extending the lifetime of this PR.

PR #1329, opened 2021-02-01
Bump nixpkgs to 449b698

tobim:

This updates the following libs for the vast static build:
- Arrow: 3.0.0
- simdjson: 0.8.1

It also incorporates a bugfix so spdlog propagates its dependency on fmt, which broke the build.


dominiklohmann (approved):
Trusting CI on this one.

PR #1330, opened 2021-02-02
Require fmt to be installed separately from spdlog

dominiklohmann:

This makes {fmt} a hard dependency of VAST, and fixes some of the remaining issues with the new logger.

###  :memo: Checklist

Commit-by-commit.

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -65,8 +65,9 @@ jobs:
           apt-get -qq update
           apt-get -qqy install git libsnappy-dev libbrotli-dev zlib1g-dev wget liblz4-dev libzstd-dev ninja-build libpcap-dev libssl-dev libbenchmark-dev tcpdump lsb-release libflatbuffers-dev flatbuffers-compiler-dev libyaml-cpp-dev libsimdjson-dev gnupg2 gcc-8 g++-8 python3 python3-pip python3-venv jq apt-transport-https ca-certificates curl gnupg-agent software-properties-common
 
-          # Need to specify backports, since spdlog also has a regular buster package.
-          apt-get -qqy -t buster-backports install libspdlog-dev
+          # Need to specify backports explicitly, since spdlog and fmt also have
+          # a regular buster packages.
+          apt-get -qqy -t buster-backports install libspdlog-dev libfmt-dev
 
           # clang 9 (c.f. https://apt.llvm.org/)
           curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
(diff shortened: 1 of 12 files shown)


.github/workflows/vast.yaml
```diff
@@ -65,8 +65,9 @@ jobs:
           apt-get -qq update
           apt-get -qqy install git libsnappy-dev libbrotli-dev zlib1g-dev wget liblz4-dev libzstd-dev ninja-build libpcap-dev libssl-dev libbenchmark-dev tcpdump lsb-release libflatbuffers-dev flatbuffers-compiler-dev libyaml-cpp-dev libsimdjson-dev gnupg2 gcc-8 g++-8 python3 python3-pip python3-venv jq apt-transport-https ca-certificates curl gnupg-agent software-properties-common
 
-          # Need to specify backports, since spdlog also has a regular buster package.
-          apt-get -qqy -t buster-backports install libspdlog-dev
+          # Need to specify backports explicitly, since spdlog and fmt also have
+          # a regular buster packages.
```
lava, line 69: nit: Remove `a`.

dominiklohmann:
[ch22417]

libvast/vast/logger.hpp
```diff
@@ -44,22 +44,28 @@
...
 #  define VAST_TRACE(...)                                                      \
-    VAST_DEBUG(vast::detail::spd_msg_from_args(1, 2, __VA_ARGS__).str(),       \
-               "ENTER", __func__, __VA_ARGS__);                                \
-    auto CAF_UNIFYN(vast_log_trace_guard_) = ::caf::detail::make_scope_guard(  \
-      [=, func_name_ = __func__] { VAST_DEBUG("EXIT {}", func_name_); })
+    auto CAF_UNIFYN(vast_log_trace_guard_)                                     \
+      = [func_name_ = __func__](auto&& format, auto&&... args) {               \
+          VAST_DEBUG(::fmt::format("ENTER {} {}", func_name_,                  \
+                                   std::forward<decltype(format)>(format)),    \
+                     std::forward<decltype(args)>(args)...);                   \
```
lava, line 65: Would this work for an invocation like `VAST_TRACE(1, 2)`, i.e. without format string? Or was this previously unsupported as well?

dominiklohmann: That isn't supported by any of the other logging macros either. It was previously supported with the old CAF logger, but that didn't require format strings.

lava: This never worked like any of the other logging macros anyways, so I wouldn't expect symmetry.

I'm asking because if we require a format string anyways, it seems like this could just be written as 
```
[ func_name_ = __func__](std::string_view format, auto&&... args) {
  VAST_DEBUG("ENTER {} " + format, func_name_, std::forward<decltype(args)>(args)...);
}
```

dominiklohmann: Looks like it should work, let's see.

lava (approved):
Looks good to me. Although I think the stated reasons (ie. verbosity in the include statements) are a bit weak for adding a new hard dependency, in this case it should be fine since we more or less depended on fmt anyways via spdlog, and it is a "nice" dependency to have since it's stable, useful and available everywhere.

The rest of the changes look solid.

PR #1341, opened 2021-02-04
Allow for plugins to assign additional type IDs

dominiklohmann:

This allows plugins to assign additional type IDs rather easily, and detects overlaps when doing it in the recommended way.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/vast/vast.cpp
+++ b/vast/vast.cpp
@@ -105,7 +105,7 @@ int main(int argc, char** argv) {
 #endif
     if (!exists(file))
       return false;
-    if (auto plugin = plugin_ptr::make(file.str().c_str())) {
+    if (auto plugin = plugin_ptr::make(file.str().c_str(), cfg)) {
       VAST_ASSERT(*plugin);
       auto has_same_name = [name = (*plugin)->name()](const auto& other) {
         return !std::strcmp(name, other->name());
--- a/vast/vast.cpp
+++ b/vast/vast.cpp
@@ -140,7 +140,14 @@ int main(int argc, char** argv) {
       }
     }
     if (!plugin_found) {
-      std::cerr << "failed to find plugin: " << plugin_file << std::endl;
+      std::cerr << "failed to load plugin: " << plugin_file;
+      if (plugin_load_errors.empty()) {
+        std::cerr << '\n';
+      } else {
+        std::cerr << "; the following errors occured:\n";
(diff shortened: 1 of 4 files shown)


tobim (changes requested):
I tested this locally by creating a second plugin that uses uses the same type id as the example plugin. This results in a very misleading error message:
```
$ vast start
failed to find plugin: libclash
```
Unfortunately the main function never prints the correct message.

vast/vast.cpp
```diff
@@ -140,7 +140,10 @@ int main(int argc, char** argv) {
       }
     }
     if (!plugin_found) {
-      std::cerr << "failed to find plugin: " << plugin_file << std::endl;
+      std::cerr << "failed to find plugin: " << plugin_file
```
tobim, line 143: The message "failed to find plugin" should only be printed if the plugin really wasn't found. That message should only be added to `plugin_load_errors` only when this is really the case.

tobim (changes requested):

vast/vast.cpp
dominiklohmann: Would you be happy with "failed to find or load plugin" instead?

dominiklohmann: Please check back, I reworded it slightly.

vast/vast.cpp
```diff
@@ -140,7 +140,14 @@ int main(int argc, char** argv) {
       }
     }
     if (!plugin_found) {
-      std::cerr << "failed to find plugin: " << plugin_file << std::endl;
+      std::cerr << "failed to load plugin: " << plugin_file;
```
tobim, line 143: Here is what I think the outer loop should consist of:
```
  for (const auto& plugin_file : plugin_files) {
    if (auto err = load_plugin(plugin_file))
      std::cerr << "failed to load plugin " << plugin_file << ':' << render(err)
                << std::endl;
  }
```
The iterating over candidate directories should not leak out of `load_plugin`.

dominiklohmann: I'm not against that and will implement this now because I need to get this merged, but please for the next time write a follow-up story because that logic change is entirely unrelated to this PR.

dominiklohmann: I pushed a commit. Please review again on the sooner side.

tobim (approved):
Looks good now.

PR #1345, opened 2021-02-04
Cache serialized layouts in slice builders

tobim:
This change adds a local cache for the binary serialized layout to table slice builders. The cached buffer will be used to avoid repeated invocations of the binary serializer.

###  :memo: Checklist

- [x] Apply the same change for the `msgpack_table_slice_builder`.

### DIFF
--- a/libvast/vast/arrow_table_slice_builder.hpp
+++ b/libvast/vast/arrow_table_slice_builder.hpp
@@ -14,6 +14,7 @@
 #pragma once
 
 #include "vast/fwd.hpp"
+
 #include "vast/table_slice.hpp"
 #include "vast/table_slice_builder.hpp"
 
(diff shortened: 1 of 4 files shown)


dominiklohmann (commented):
You requested a review, so I took a look at it. For Arrow, this is fine. It should be quite easy to make the same change for MessagePack.

Do you think it makes sense to encapsulate this functionality in the base class?

libvast/vast/arrow_table_slice_builder.hpp
```diff
@@ -108,6 +109,10 @@ class arrow_table_slice_builder final : public table_slice_builder {
   /// Schema of the Record Batch corresponding to the layout.
   record_type flat_layout_;
 
+  /// The serialized layout can be cached because every builder instance only
+  /// produces slices of a single layout.
+  mutable std::vector<char> serialized_layout_cache_;
```
mavam, line 114:
```suggestion
  mutable std::vector<std::byte> serialized_layout_cache_;
```

YMMV: more idiomatic—but at the cost of a few casts.

dominiklohmann: `char` is the type used by FlatBuffers internally, so we should stay with it here imo.

tobim: @dominiklohmann is right, it has to be `std::vector<char>`, otherwise we need to introduce a bunch of needless casting.

dominiklohmann (approved):
This is fine and covered by CI. The logic is easy to follow.

We should consider abstracting the layout cache logic in the base class `table_slice_builder` sometime in the future.

PR #1346, opened 2021-02-05
Logging adjustments

lava:

1. Change the default log patterns so that day and time zone are not printed for the console logs, and the time zone *is* printed for file logs
2. Provide separate macros `VAST_TRACE()` and `VAST_TRACE_SCOPED()` to allow trace-level logging without creating a scope guard.

###  :memo: Checklist

- [x] All user-facing changes have changelog entries. (not necessary, since the previous log pattern was never released)

Review commit-by-commit. You can actually regard the commits as individual PRs.

### DIFF
--- a/examples/plugins/example/example.cpp
+++ b/examples/plugins/example/example.cpp
@@ -44,7 +44,7 @@ example_actor::behavior_type
 example(example_actor::stateful_pointer<example_actor_state> self) {
   return {
     [self](atom::config, record config) {
-      VAST_TRACE("{} sets configuration {}", self, config);
+      VAST_TRACE_SCOPE("{} sets configuration {}", self, config);
       for (auto& [key, value] : config) {
         if (key == "max-events") {
           if (auto max_events = caf::get_if<integer>(&value)) {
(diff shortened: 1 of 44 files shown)


libvast/src/chunk.cpp
```diff
@@ -192,7 +192,7 @@ chunk::chunk(view_type view, deleter_type&& deleter) noexcept
   : view_{view}, deleter_{std::move(deleter)} {
   auto data = view.data();
   auto sz = view.size();
-  VAST_TRACEPOINT(chunk_make, data, sz);
+  VAST_TRACE_SCOPEPOINT(chunk_make, data, sz);
```
dominiklohmann, line 195: That does not look correct, your replacement sed should've included the opening bracket.

dominiklohmann (commented):
Looks good, just fix up the VAST_TRACEPOINT macro before merging please.

Can you also change the vast.yaml.example accordingly?

libvast/src/chunk.cpp
lava: Whoops, updated.

dominiklohmann (approved):
LGTM

PR #1335, opened 2021-02-03
Make the node a typed actor

dominiklohmann:
###  :notebook_with_decorative_cover: Description

This converts the node actor to the `caf::typed_actor` API.

###  :memo: Checklist

- [x] All user-facing changes have changelog entries.
- [x] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [x] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

File-by-file. I didn't really know how to split this, but all the changes should be self-explanatory.

PR #1347, opened 2021-02-05
Fix misbehavior of queries returning no results

dominiklohmann:

For reasons I admittedly still don't entirely understand, this fixes an error that was introduced with https://github.com/tenzir/vast/pull/1335 for queries that returned no results.

###  :memo: Checklist

 @lava please verify if this fixes the issue you've been seeing.

### DIFF
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -619,13 +619,19 @@ node(node_actor::stateful_pointer<node_state> self, std::string name, path dir,
       auto msg = run(inv, self->system(), node_state::command_factory);
       if (!msg)
         return msg.error();
-      if (msg->match_elements<caf::error>())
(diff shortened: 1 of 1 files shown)


lava (approved):
I've verified locally that this fixes the bug we observed, although I don't understand from the code what the issue was and how this change avoids it. Maybe you can explain a bit in the commit message?

tobim:
I think the real reason is that we're not using `typed_response_promises`. Looking at the code for `spawn_command` I see at least one place where that just returns `caf::none`. I'm surprised that this compiles actually.

dominiklohmann:
I pushed another commit with a slightly better fix and a commit message that explains the issue.

tobim (approved):
Verified that the fix works.

PR #1348, opened 2021-02-05
Simplify select overloads for words

JoeLoser:

Simplify `select` function templates by preferring `if-constexpr` with a predicate instead of SFINAE.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/word.hpp
+++ b/libvast/vast/word.hpp
@@ -422,30 +422,21 @@ static constexpr auto find_prev(T x, detail::word_size_type<T> i)
 template <bool Bit = true, class T>
 static constexpr auto select(T x, detail::word_size_type<T> i)
 -> std::enable_if_t<
-  Bit && detail::is_unsigned_integral_v<T>,
+  detail::is_unsigned_integral_v<T>,
   detail::word_size_type<T>
 > {
   // TODO: make this efficient and branch-free. There is one implementation
   // that counts from the right for 64-bit here:
   // http://graphics.stanford.edu/~seander/bithacks.html
+  const auto pred = [](const auto&... args){
+    if constexpr(Bit) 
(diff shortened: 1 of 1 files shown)


libvast/vast/word.hpp
```diff
@@ -422,30 +422,23 @@ static constexpr auto find_prev(T x, detail::word_size_type<T> i)
...
   // TODO: make this efficient and branch-free. There is one implementation
   // that counts from the right for 64-bit here:
   // http://graphics.stanford.edu/~seander/bithacks.html
+  const auto pred = [](const auto&... args){
+    if constexpr(Bit) {
+      return word<T>::test(args...);
+    }
+    else {
+      return !word<T>::test(args...);
+    }
```
dominiklohmann, line 437: We usually omit the braces if every branch is just a single statement. Not that this is necessarily better, but let's keep the style consistent throughout the code base.

```suggestion
    if constexpr(Bit)
      return word<T>::test(args...);
    else
      return !word<T>::test(args...);
```

dominiklohmann (approved):
Thanks for your contribution! Changes like this are always welcome. I just have one minor request regarding the code style, but other than that this looks ready to be merged.

libvast/vast/word.hpp
JoeLoser: Fair enough. It'd be good to perhaps enforce this with a `clang-tidy` check, but I just fixed it up for the purposes of this commit to be consistent. Thanks 👍

dominiklohmann: We do have clang-tidy somewhere on our roadmap, but so far just didn't get around to doing it.

Thanks for the fixup, and thanks again for your contribution! 🎉

JoeLoser:
CC: @dwms-qlab

PR #1354, opened 2021-02-07
Move options from format to the import subcommand

tobim:

The options `listen`, `read`, `schema`, `schema-file`, `type`, and `uds` can from now on be supplied to the `import` command directly. They can still be used after the format subcommand, but that usage is deprecated.

###  :memo: Checklist

By commit.

### DIFF
--- a/libvast/src/command.cpp
+++ b/libvast/src/command.cpp
@@ -357,6 +357,29 @@ caf::error parse_impl(invocation& result, const command& cmd,
   return parse_impl(result, **i, position + 1, last, target);
 }
 
+void fixup_options(invocation& inv) {
+  using namespace std::string_literals;
+  for (const auto& cmd : {"import", "spawn.source"}) {
+    for (const auto& format :
+         {"csv", "json", "suricata", "syslog", "zeek", "zeek-json"}) {
+      for (const auto& opt :
+           {"listen", "read", "schema", "schema-file", "type", "uds"}) {
+        auto path = "vast."s + cmd + '.' + format + '.' + opt;
+        if (auto x = caf::get_if<std::string>(&inv.options, path)) {
+          // The logger isn't initialized yet.
+#if VAST_LOG_LEVEL >= VAST_LOG_LEVEL_WARNING
+          fmt::print(stderr,
+                     "The option '{}' is deprected for the '{}' subcommand and "
+                     "should be specified at the {} subcommand instead\n",
+                     opt, format, cmd);
+#endif
+          put(inv.options, "vast."s + cmd + '.' + opt, *x);
+        }
+      }
+    }
+  }
+}
+
 caf::expected<invocation>
 parse(const command& root, command::argument_iterator first,
       command::argument_iterator last) {
@@ -380,6 +403,7 @@ parse(const command& root, command::argument_iterator first,
     manfooter(std::cout);
     return caf::no_error;
   }
+  fixup_options(result);
   return result;
 }
 
--- a/libvast/src/command.cpp
+++ b/libvast/src/command.cpp
@@ -359,25 +359,30 @@ caf::error parse_impl(invocation& result, const command& cmd,
 
 void fixup_options(invocation& inv) {
   using namespace std::string_literals;
-  for (const auto& cmd : {"import", "spawn.source"}) {
-    for (const auto& format :
-         {"csv", "json", "suricata", "syslog", "zeek", "zeek-json"}) {
-      for (const auto& opt :
-           {"listen", "read", "schema", "schema-file", "type", "uds"}) {
-        auto path = "vast."s + cmd + '.' + format + '.' + opt;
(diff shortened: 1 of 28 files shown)


tobim:
I guess it would be a good idea to do the same for the sink options `write` and `uds`.

dominiklohmann (commented):
Quoting myself from the story:

> I have another thought on this: Consider this file:
> 
> ```yaml
> vast:
>   import:
>     pcap:
>       interface: en0
>     read: path/to/file.pcap
> ```
> 
> What does `vast import pcap` do in this case?
> 
> Similarly, consider this currently valid file:
> 
> ```
> vast:
>   import:
>     suricata:
>       read: path/to/eve.json
>     zeek
>       read: path/to/conn.log
> ```
> 
> How would you express this with your proposed changes?

I did not look at the code yet, but I would like to see these questions answered before we merge a change like this.

tobim:
> Quoting myself from the story:
> 
> > I have another thought on this: Consider this file:
> > ```yaml
> > vast:
> >   import:
> >     pcap:
> >       interface: en0
> >     read: path/to/file.pcap
> > ```

This is not different from
```yaml
vast:
  import:
    pcap:
      interface: en0
      read: path/to/file.pcap
```
The pcap implementation prefers the interface key in both cases.

> > Similarly, consider this currently valid file:
> > ```
> > vast:
> >   import:
> >     suricata:
> >       read: path/to/eve.json
> >     zeek
> >       read: path/to/conn.log
> > ```
> > 
> > 
> > How would you express this with your proposed changes?

While valid, the utility of such a config is extremely limited. If we really want to allow users to specify their imports declaratively they should either have a separate config file per import process or this has to become a list in a unified config file.

dominiklohmann:
Reasoning checks out. I think it'd be nice if this same PR can also make the changes on the export side so it stays symmetrical.

tobim:
> Reasoning checks out. I think it'd be nice if this same PR can also make the changes on the export side so it stays symmetrical.

Will do after the rebase on top of #1356.

libvast/src/command.cpp
```diff
@@ -357,6 +357,34 @@ caf::error parse_impl(invocation& result, const command& cmd,
...
 
+void fixup_options(invocation& inv) {
+  using namespace std::string_literals;
+  auto move_option_arg = [&](auto cmd, auto format, auto opt) {
+    auto path = "vast."s + cmd + '.' + format + '.' + opt;
+    if (auto x = caf::get_if<std::string>(&inv.options, path)) {
+      // The logger isn't initialized yet.
+#if VAST_LOG_LEVEL >= VAST_LOG_LEVEL_WARNING
+      fmt::print(stderr,
+                 "The option '{}' is deprected for the '{}' subcommand and "
```
dominiklohmann, line 368:
```suggestion
                 "The option '{}' is deprecated for the '{}' subcommand and "
```

libvast/src/command.cpp
```diff
@@ -357,6 +357,34 @@ caf::error parse_impl(invocation& result, const command& cmd,
   return parse_impl(result, **i, position + 1, last, target);
 }
 
+void fixup_options(invocation& inv) {
+  using namespace std::string_literals;
+  auto move_option_arg = [&](auto cmd, auto format, auto opt) {
```
dominiklohmann, line 362: Should this really take things by value?

libvast/src/command.cpp
```diff
@@ -357,6 +357,34 @@ caf::error parse_impl(invocation& result, const command& cmd,
...
+  };
+  for (const auto& cmd : {"import", "spawn.source"})
+    for (const auto& format :
+         {"csv", "json", "suricata", "syslog", "zeek", "zeek-json"})
+      for (const auto& opt :
+           {"listen", "read", "schema", "schema-file", "type", "uds"})
+        move_option_arg(cmd, format, opt);
+  for (const auto& cmd : {"export", "spawn.sink"})
+    for (const auto& format :
+         {"arrow", "ascii", "csv", "json", "null", "pcap", "zeek"})
```
dominiklohmann, line 383: I don't like that we must list this manually here. Is there away we can refactor this so it uses the existing command structure? If there isn't, then we should at least add comments to application.cpp with a hint to check this code as well.

libvast/src/command.cpp
tobim: I think the deduced type should be `const char*`, so yeah.

libvast/src/command.cpp
tobim: I really don't like the idea of adding extra machinery to the command API to specify these relocations. It would require modifications to the options parser to do it properly.

libvast/src/command.cpp
dominiklohmann: Care to make that explicit?

libvast/src/command.cpp
dominiklohmann: Then please add comments to application.cpp.

dominiklohmann (approved):

PR #1355, opened 2021-02-08
Disable the "node queries" unit test

dominiklohmann:

Not really a fix, but has been so annoying in CI recently that I'd rather disable this until we have the time to investigate the issue properly.

###  :memo: Checklist

Discuss whether we really want to do this.


lava (changes requested):
As we discussed on slack, let's instead remove this test completely and replace it with an integration test that ingests the zeek data and performs the two queries.

lava (approved):
lgtm

PR #1361, opened 2021-02-09
Verify valid JSON in integration tests

lava:

Use `jq -e` to check that our json printer returns valid json from queries.

NOTE: I would have liked a "verify-only" mode from jq that does not modify the input at all, but that doesn't seem to exist. Let's hope that the `-c` format remains stable in the long term.

###  :memo: Checklist


dominiklohmann:
> NOTE: I would have liked a "verify-only" mode from jq that does not modify the input at all, but that doesn't seem to exist. Let's hope that the `-c` format remains stable in the long term.

I think we can easily add another step alongside `transformation` in our integration test framework that purely does verification. I'm not quite sure it's worth the effort here, though.

dominiklohmann (approved):
Missed one at line 152 (pre patch line count). Please fix before merging, otherwise this looks good to me.

PR #1362, opened 2021-02-09
Make log rotation configurable

lava:

This PR adds minimal configurability to the rotating file sink used for spdlog.

It adds the option `vast.log-rotation` to turn off rotation, which is required for production use where VAST's log rotation may interfere with log rotation configured in external programs like systemd or logrotated.

The size threshold, number of rotated log files, and whether to use time-based logrotate are not made configurable, since we did not have anyone asking for these features yet.

###  :memo: Checklist

### DIFF
--- a/libvast/src/logger.cpp
+++ b/libvast/src/logger.cpp
@@ -225,7 +225,6 @@ bool setup_spdlog(const vast::invocation& cmd_invocation,
       auto threshold_str
         = detail::get_bytesize(cfg_file, "vast.log-rotation-threshold",
                                defaults::logger::rotate_threshold);
-
       file_sink = std::make_shared<spdlog::sinks::rotating_file_sink_mt>(
         log_file, defaults::logger::rotate_threshold,
         defaults::logger::rotate_files);
--- a/libvast/src/logger.cpp
+++ b/libvast/src/logger.cpp
@@ -26,6 +26,7 @@
 #include <spdlog/async.h>
 #include <spdlog/common.h>
 #include <spdlog/sinks/ansicolor_sink.h>
+#include <spdlog/sinks/basic_file_sink.h>
 #include <spdlog/sinks/null_sink.h>
 #include <spdlog/sinks/rotating_file_sink.h>
 
@@ -205,6 +206,7 @@ bool setup_spdlog(const vast::invocation& cmd_invocation,
   spdlog::init_thread_pool(defaults::logger::queue_size,
                            defaults::logger::logger_threads);
   std::vector<spdlog::sink_ptr> sinks;
+  // Add console sink.
   auto stderr_sink
     = std::make_shared<spdlog::sinks::ansicolor_stderr_sink_mt>(log_color);
   stderr_sink->set_level(vast_loglevel_to_spd(vast_console_verbosity));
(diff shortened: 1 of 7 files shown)


dominiklohmann:
The option should probably be named `vast.disable-log-rotation` with the default being `false` so it's in line with the other boolean options.

vast.yaml.example
```diff
@@ -30,6 +30,10 @@ vast:
   # release builds.
   # file-verbosity: debug
 
+  # Whether to enable automatic log rotation. If set to true, a new log file
+  # will be created when the size of the current log file exceeds 10 MiB.
+  #log-rotation: true
```
mavam, line 35: Is it possible to offer an option to configure the hard-coded 10 MiB limit? As a user, I would want to have this tuning knob right next to the on/off switch.

lava: It is of course possible, but this is not the only tuning knob that spdlog exposes:
 * file size threshold
 * number of rotated log files that should be kept
 * should we use size-based or time-based rotation?
 * for time-based rotation, the interval (daily/weekly/...) and starting point

I left all of this out for now. I'm happy to add a single additional option for the threshold if you want it, but I'd really like to avoid building a "general" solution for this.

mavam: I was not saying that it should be in this PR, but I do think we should give the user a little more flexibility to pass options to spdlog. After all, a pass-through is not a lot of code complexity. Let's discuss what exactly we deem the minimal set of features to make users happy. I agree that we don't need to support the whole enchilada out of the box.

lava: Sounds good, I've created https://app.clubhouse.io/tenzir/story/22662/expose-more-spdlog-options-to-the-user as a follow-up story for that.

libvast/vast/detail/settings.hpp
```diff
@@ -45,4 +45,12 @@ void merge_settings(const caf::settings& src, caf::settings& dst,
 ///   { a = 13 }
 bool strip_settings(caf::settings& xs);
 
+// Returns the amount of bytes specified by a config option.
+// * If the key has a string or integer key giving a size, use that.
+// * If the key doesn't exist, use the provided default value.
+// * If the key exists with a different type, return an error.
+// * If the key exists but cant be parsed as a byte size, return an error.
```
dominiklohmann, line 52:
```suggestion
/// Returns the amount of bytes specified by a config option.
/// * If the key has a string or integer key giving a size, use that.
/// * If the key doesn't exist, use the provided default value.
/// * If the key exists with a different type, return an error.
/// * If the key exists but cant be parsed as a byte size, return an error.
```

libvast/src/logger.cpp
```diff
@@ -213,11 +216,22 @@ bool setup_spdlog(const vast::invocation& cmd_invocation,
...
-      log_file, defaults::logger::rotate_threshold,
-      defaults::logger::rotate_files);
+    bool disable_rotation = caf::get_or(cfg_file, "vast.disable-log-rotation",
+                                        defaults::logger::disable_log_rotation);
+    spdlog::sink_ptr file_sink = nullptr;
+    if (!disable_rotation) {
+      auto threshold_str
+        = detail::get_bytesize(cfg_file, "vast.log-rotation-threshold",
+                               defaults::logger::rotate_threshold);
+
```
dominiklohmann, line 228:
```suggestion
```

vast.yaml.example
```diff
@@ -30,6 +30,13 @@ vast:
   # release builds.
   # file-verbosity: debug
 
+  # Whether to enable automatic log rotation. If set to true, a new log file
+  # will be created when the size of the current log file exceeds 10 MiB.
+  #disable-log-rotation: false
```
dominiklohmann, line 35:
```suggestion
  disable-log-rotation: false
```

vast.yaml.example
```diff
@@ -30,6 +30,13 @@ vast:
...
+
+  # The size limit when a log file should be rotated.
+  #log-rotation-threshold: 10MiB
```
dominiklohmann, line 38:
```suggestion
  log-rotation-threshold: 10MiB
```

dominiklohmann (approved):
Ran this locally and it works as expected. Just had some nits. Please implement before merging.

vast.yaml.example
mavam: Perfect, I think that's the right trade-off for now.

PR #1363, opened 2021-02-10
Turn the accountant into a named source

tobim:

No more
```
imorter adds anonymous source
...
imorter removes anonymous source
```

### DIFF
--- a/libvast/src/system/accountant.cpp
+++ b/libvast/src/system/accountant.cpp
@@ -266,7 +266,8 @@ accountant(accountant_actor::stateful_pointer<accountant_state> self,
       st.actor_map[self->current_sender()->id()] = name;
       self->monitor(self->current_sender());
       if (name == "importer")
-        st.mgr->add_outbound_path(self->current_sender());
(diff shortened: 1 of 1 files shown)


mavam (commented):
This makes sense to me.

However, don't we need to do this at other places as well? Also, couldn't this be obtained from the registry?

dominiklohmann (approved):

dominiklohmann:
> However, don't we need to do this at other places as well? Also, couldn't this be obtained from the registry?

We have this mechanism for other sources already. Not all actors are as named actors in the registry.

mavam:
> We have this mechanism for other sources already. Not all actors are as named actors in the registry.

My point is: we use the accountant at many places. So are all other places correctly named? This seems to be the time to make sure it's the case.

tobim:
We have 2 places in the code that call `add_outbound_path()` to the `importer`. The other one is in `source.hpp` and we're already doing this there.

The integration test timeout is unrelated to this, merging.

PR #1364, opened 2021-02-10
Fix the embedded version string for static

tobim:

A fix for the CMake version tag helper to handle `-DVAST_VERSION_TAG` correctly.


dominiklohmann (approved):
Makes sense. The change is simple. Did not test locally—can you confirm this works for you @tobim?

tobim:
Yes I tested this.

PR #1366, opened 2021-02-10
Introduce atom::internal as first argument for private behaviors

dominiklohmann:

- Adds `atom::internal` as first argument for private behaviors to make them visible at the call site.
- Adds `atom::intel` as a nice-to-have for the matcher plugin—snuck this in here because touching the atoms header is expensive and I don't want to do it twice.

###  :memo: Checklist

File-by-file.


lava (approved):
The change looks good to me assuming that you can fix the CI issues without any major additional work.

The commit message of the second commit should be reworded if possible, since it doesn't make sense outside the context of this PR.

dominiklohmann:
> The commit message of the second commit should be reworded if possible, since it doesn't make sense outside the context of this PR.

Valid point, I reworded it.

> The change looks good to me assuming that you can fix the CI issues without any major additional work.

Unrelated GitHub failure I think.

PR #1369, opened 2021-02-11
Allow plugins to define up to two type ID blocks

dominiklohmann:

In order to avoid cyclic dependencies between type definitions and actor definitions, it is sometimes necessary to define two type ID blocks. This change enables plugins to use up to two type ID blocks.

###  :memo: Checklist

n/t


dominiklohmann:
Note that I considered making use of some variadic macro magic to allow any number of type ID blocks, but I failed to see why you would ever need more than two. We have three for libvast itself for compile time reasons, but that shouldn't ever be an issue for plugins.

tobim (approved):
I don't have a plugin to use this yet, but I trust that you tested this with in-progress work.

PR #1371, opened 2021-02-11
Fix formatting of recent changelog entries

dominiklohmann:


mavam (approved):
No brainer. All good.

PR #1373, opened 2021-02-11
Expand subnet value predicates

mavam:

This PR expands value predicates of type `subnet` as follows: given a subnet `S`, the parser expands it to `:subnet == S` and the normalization then further to `:subnet == S || :addr in S`.

###  :memo: Checklist

File by file.

### DIFF
--- a/libvast/src/concept/parseable/vast/expression.cpp
+++ b/libvast/src/concept/parseable/vast/expression.cpp
@@ -20,7 +20,6 @@
 #include "vast/detail/assert.hpp"
 #include "vast/detail/string.hpp"
 #include "vast/expression.hpp"
-#include "vast/expression_visitors.hpp"
 #include "vast/type.hpp"
 
 namespace vast {
--- a/libvast/src/concept/parseable/vast/expression.cpp
+++ b/libvast/src/concept/parseable/vast/expression.cpp
@@ -55,7 +55,7 @@ static predicate::operand to_data_operand(data x) {
   return x;
 }
 
-static predicate to_data_predicate(data x) {
+static predicate to_value_predicate(data x) {
   auto infer_type = [](auto& d) -> type {
     return data_to_type<std::decay_t<decltype(d)>>{};
   };
@@ -102,7 +102,7 @@ static auto make_predicate_parser() {
   auto ws = ignore(*parsers::space);
   auto pred
     = (operand >> ws >> operation >> ws >> operand)->*to_predicate
-    | parsers::data->*to_data_predicate
+    | parsers::data->*to_value_predicate
     ;
   return pred;
   // clang-format on
--- a/libvast/src/concept/parseable/vast/expression.cpp
+++ b/libvast/src/concept/parseable/vast/expression.cpp
@@ -55,7 +55,7 @@ static predicate::operand to_data_operand(data x) {
   return x;
 }
 
-static predicate to_value_predicate(data x) {
+static predicate to_data_predicate(data x) {
   auto infer_type = [](auto& d) -> type {
     return data_to_type<std::decay_t<decltype(d)>>{};
   };
@@ -102,7 +102,7 @@ static auto make_predicate_parser() {
   auto ws = ignore(*parsers::space);
   auto pred
     = (operand >> ws >> operation >> ws >> operand)->*to_predicate
-    | parsers::data->*to_value_predicate
+    | parsers::data->*to_data_predicate
     ;
   return pred;
   // clang-format on
--- a/libvast/src/concept/parseable/vast/expression.cpp
+++ b/libvast/src/concept/parseable/vast/expression.cpp
@@ -55,9 +55,9 @@ static predicate::operand to_data_operand(data x) {
   return x;
 }
 
-// Expands a predicate with a type extractor, an equality operator, and a
-// corresponding data instance according to the rules of the expression
-// language.
+/// Expands a predicate with a type extractor, an equality operator, and a
+/// corresponding data instance according to the rules of the expression
+/// language.
 struct expander {
   expression operator()(caf::none_t) const {
     return expression{};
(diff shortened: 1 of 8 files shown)


dominiklohmann (changes requested):
There's a problem with the current implementation.

Instead of expanding just value predicates, it expands any expression `:subnet == <subnet>`, making it impossible to query just for subnets by equality. 

In a call with @mavam we discussed two possible options:
- Move the expansion into the parser.
- Make value predicates valid predicates, i.e., don't expand them during parsing at all.

The second one would be a larger refactoring, so we'll likely opt for the first. The documentation should be rewritten accordingly to make clear this only affects value predicates.

libvast/src/expression_visitors.cpp
```diff
@@ -35,6 +35,52 @@
...
+  for (auto& op : d)
+    result.push_back(caf::visit(*this, op));
+  return result;
+}
+expression expander::operator()(const negation& n) const {
+  return {negation{caf::visit(*this, n.expr())}};
+}
+
+expression expander::operator()(const predicate& p) const {
+  auto make_addr_pred
```
dominiklohmann, line 60: This function urgently needs comments above every step. Make it easy for the future maintainer to follow this function. `make_addr_pred` as a function name alone is not sufficient here.

libvast/src/concept/parseable/vast/expression.cpp
```diff
@@ -20,6 +20,7 @@
 #include "vast/detail/assert.hpp"
 #include "vast/detail/string.hpp"
 #include "vast/expression.hpp"
+#include "vast/expression_visitors.hpp"
```
dominiklohmann, line 23: The expander doesn't really belong in `expression_visitor.hpp`. Can you move it to this file, remove the include, and rename it to something more specific like `value_predicate_expander`?

libvast/vast/expression_visitors.hpp
```diff
@@ -28,6 +28,16 @@ namespace vast {
 
 class event;
 
+/// Completes an expression by expanding value predicates to full sub
+/// expressions.
```
dominiklohmann, line 32: Comment does not reflect what this actually does.

libvast/src/concept/parseable/vast/expression.cpp
```diff
@@ -181,16 +181,23 @@ static auto make_expression_parser() {
...
   auto ws = ignore(*parsers::space);
-  auto negate_pred = [](predicate p) { return negation{expression{p}}; };
-  auto negate_expr = [](expression expr) { return negation{expr}; };
+  auto negate_expr = [](expression expr) { return negation{std::move(expr)}; };
   rule<Iterator, expression> expr;
   rule<Iterator, expression> group;
   // clang-format off
+  auto expand = [](data x) {
+    auto pred = to_value_predicate(std::move(x));
+    return caf::visit(expander{}, expression{std::move(pred)});
```
dominiklohmann, line 190: Should we just call this inside of `to_value_predicate`? Why add another lambda on top?

dominiklohmann (commented):
Just some nits. I ran this locally and it works fine on my end.

libvast/src/concept/parseable/vast/expression.cpp
dominiklohmann: Or, the other way around: Should we call `to_value_predicate` inside the expander? Then its name makes sense.

libvast/src/concept/parseable/vast/expression.cpp
```diff
@@ -56,13 +56,69 @@ static predicate::operand to_data_operand(data x) {
...
 
-static predicate to_value_predicate(data x) {
+// Expands a predicate with a type extractor, an equality operator, and a
+// corresponding data instance according to the rules of the expression
+// language.
+struct expander {
+  expression operator()(caf::none_t) const {
+    return expression{};
+  }
+  expression operator()(const conjunction& c) const {
```
dominiklohmann, line 65: Why aren't there empty lines between the member functions?

libvast/src/concept/parseable/vast/expression.cpp
```diff
@@ -56,13 +56,69 @@ static predicate::operand to_data_operand(data x) {
   return x;
 }
 
-static predicate to_value_predicate(data x) {
+// Expands a predicate with a type extractor, an equality operator, and a
+// corresponding data instance according to the rules of the expression
+// language.
```
dominiklohmann, line 60: Is doxygenize a word? If so, please doxygenize the comment by adding another `/`.

libvast/src/concept/parseable/vast/expression.cpp
```diff
@@ -56,13 +56,69 @@ static predicate::operand to_data_operand(data x) {
...
+      return make_disjunction(p, std::move(*addr_pred));
+    if (auto addr_pred = build_addr_pred(p.rhs, p.op, p.lhs))
+      return make_disjunction(p, std::move(*addr_pred));
+    return {p};
+  }
+};
+
+// Expands a data instance in two steps:
+// 1. Convert the data instance x to T(x) == x
+// 2. Apply type-specific expansion that results in a compound expression
```
dominiklohmann, line 111: Please `s/^/\/` here as well.

dominiklohmann (approved):
Approved modulo some minor fixups.

libvast/src/concept/parseable/vast/expression.cpp
mavam: Fixed.

libvast/src/concept/parseable/vast/expression.cpp
mavam: Disagreeing here; this is an implementation detail local to the function. It will never land in documentation.

libvast/src/concept/parseable/vast/expression.cpp
mavam: s.a.

libvast/src/concept/parseable/vast/expression.cpp
dominiklohmann: Tooling disagrees with you. I get a different hint from clangd on autocomplete with three slashes.

mavam: Thank god I'm a human and can make decisions with more context. 😉

mavam: Joking aside, this visitor is in an anonymous namespace. It should not be publicly documented.

mavam: Per our call, turns out the de-facto standard for language servers is to use triple slashes...who would have known.

PR #1375, opened 2021-02-12
Add a build-time option to always load all plugins

dominiklohmann:

This commit adds a `--with-plugin-autoloading` option to `configure` and a corresponding `VAST_ENABLE_PLUGIN_AUTOLOADING` option to CMake that configures VAST such that all plugins built alongside VAST are enabled by default.

###  :memo: Checklist

File-by-file.

I've tested this thoroughly locally with 0, 1, 2, and 3 plugins, and also tested locally that the deduplication works. This is really nice for the plugin development workflow.

### DIFF
--- a/configure
+++ b/configure
@@ -23,68 +23,70 @@ usage="\
 Usage: $0 [OPTION]... [VAR=VALUE]...
 
   Installation options:
-    --prefix=PREFIX         installation directory [/usr/local]
-    --without-relocatable   disable relocatable installations
-    --package-name          sets the name of the produced package
+    --prefix=PREFIX       installation directory [/usr/local]
+    --without-relocatable disable relocatable installations
+    --package-name        sets the name of the produced package
 
   Convenience options:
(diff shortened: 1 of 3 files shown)


configure
```diff
@@ -48,6 +48,7 @@ Usage: $0 [OPTION]... [VAR=VALUE]...
     --with-dscat            build with dscat
     --with-example-plugin   build with example plugin
     --with-plugin=DIR       build with plugin in given directory
+    --enable-all-bundled-plugins always enable all bundled plugins
```
mavam, line 51:
```suggestion
    --load-bundled-plugins always enable all bundled plugins
```

I find "load" a bit clearer. The "all" quantifier is not adding much IMO.

mavam (approved):
Looks good.

configure
mavam: Another option: `--autoload-plugins`

dominiklohmann: We settled on `--with-plugin-autoloading`.

mavam (approved):

PR #1377, opened 2021-02-13
Use detection idiom from standard library

JoeLoser:

Problem:
- Detection idiom traits are hand-rolled in `type_traits.hpp` but they
  are widely available in standard libraries now.

Solution:
- Use detection idiom traits from Library Fundamentals v2 in standard
  library. Remove hand-rolled ones in `type_traits.hpp`.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/concept/support/detail/sequence.hpp
+++ b/libvast/vast/concept/support/detail/sequence.hpp
@@ -18,19 +18,23 @@
 #include "vast/concept/support/unused_type.hpp"
 #include "vast/detail/type_traits.hpp"
 
+#include <experimental/type_traits>
+
 namespace vast::detail {
 
 template <typename T>
 using has_lhs_type_t = typename T::lhs_type;
 
 template <typename T>
-inline constexpr bool has_lhs_type_v = is_detected_v<has_lhs_type_t, T>;
+inline constexpr bool has_lhs_type_v
+  = std::experimental::is_detected_v<has_lhs_type_t, T>;
 
 template <typename T>
 using has_rhs_type_t = typename T::rhs_type;
 
 template <typename T>
-inline constexpr bool has_rhs_type_v = is_detected_v<has_rhs_type_t, T>;
+inline constexpr bool has_rhs_type_v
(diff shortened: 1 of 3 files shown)


dominiklohmann:
Do you know if there's a comprehensive overview available somewhere on implementation support? We currently aim to support these:
- GCC >= 8.0
- Clang >= 8.0
- Apple Clang >= 9.1

JoeLoser:
> Do you know if there's a comprehensive overview available somewhere on implementation support? We currently aim to support these:
> 
> * GCC >= 8.0
> * Clang >= 8.0
> * Apple Clang >= 9.1

I can't speak for `AppleClang`, but both Clang >= 8.0 and GCC >= 8.0 seem to support detection traits from Library Fundamentals v2 (see [Wandbox example taken from CppReference](https://wandbox.org/permlink/O3Fqlie6DHetAEuP)). Note that there is a feature test macro `__cpp_lib_experimental_detect` which should be `>=201505` if a compiler supports the detection idiom API. However, there seems to be a Clang bug as they do indeed support it fully but didn't define the feature test macro.

Below are the relevant commits when they landed in GCC and Clang
[Commit for clang](https://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20161121/178052.html)
[Commit for gcc](https://gcc.gnu.org/legacy-ml/libstdc++/2015-07/msg00002.html)

I filed a [Clang bug](https://bugs.llvm.org/show_bug.cgi?id=49173) for the feature test macro issue.

Since we have CI jobs for `AppleClang`, `GCC`, and `Clang`, and all tests pass, that seems sufficient coverage for me with this change. What do you think?

dominiklohmann:
> Since we have CI jobs for `AppleClang`, `GCC`, and `Clang`, and all tests pass, that seems sufficient coverage for me with this change. What do you think?

Yes. macOS is also only for local development, so we're freely bumping the minimum requirement for it. 

Thanks for the thorough answer, much appreciated.

dominiklohmann (approved):

JoeLoser:
CC: @dwms-qlab

PR #1379, opened 2021-02-14
Add native Sigma support

mavam:

This PR adds native parsing of [Sigma](https://github.com/Neo23x0/sigma) rules to VAST. It is now possible to use a Sigma rule at places where we use a normal VAST expression. For example, this now works:

```sh
vast export json < sigma/rule.yaml
```

###  :memo: Checklist

- [x] Handle wildcards properly

1. Look at the unit tests and the integration tests.
2. Skim the changes file-by-file to get an overview about the architecture.


mavam:
[ch4916]

libvast/src/detail/sigma.cpp
```diff
@@ -0,0 +1,369 @@
...
+#include <regex>
+#include <string>
+#include <tuple>
+#include <vector>
+
+namespace vast::detail::sigma {
+
+// TODO: A lot of code in here is directly copied from
+// src/concept/parseable/expression.cpp. We should factor the implementation in
+// the future.
```
dominiklohmann, line 35: Please write a story for this, otherwise it won't ever get done.

Also, a corresponding comment in the expression parser is needed to make sure we don't make changes to it that aren't mirrored here.

dominiklohmann (commented):
Just had an initial look at it. Will take another look tomorrow.

libvast/src/detail/sigma.cpp
```diff
@@ -0,0 +1,369 @@
...
+// - Regular expressions are case-sensitive by default
+// - You don't have to escape characters except the string quotation
+//   marks '
+caf::optional<pattern> make_pattern(std::string_view str) {
+  auto f = str.begin();
+  auto l = str.end();
+  std::string rx;
+  // FIXME: this is a pretty hand-wavy approach to transforming a glob string
+  // to a valid regex. We need to revisit this once we have actual pattern
+  // support in the query language.
```
dominiklohmann, line 200: This reads like a TODO rather than a FIXME. This also needs a follow-up story.

libvast/src/detail/sigma.cpp
```diff
@@ -0,0 +1,369 @@
...
+    } else if (c == '?') {
+      rx += ".";
+    } else {
+      rx += c;
+    }
+  };
+  // It's only a pattern if it differs from a regular string.
+  if (str == rx)
+    return caf::none;
+  return pattern{std::move(rx)};
```
dominiklohmann, line 239: I think you need to prepend the pattern with `^` and append it with `$`.

libvast/src/detail/sigma.cpp
```diff
@@ -0,0 +1,369 @@
...
+          all = true;
+        } else if (*i == "contains") {
+          op = relational_operator::ni;
+        } else if (*i == "endswith" || *i == "startswith") {
+          // Once we have regex support we should transform a lot of these
+          // modifier in pattern qualifiers, e.g., `endswith` for a value X
+          // should become /X$/.
+          op = relational_operator::ni;
+        } else if (*i == "base64") {
+          // TODO
```
dominiklohmann, line 284: Does it make sense to return `ec::unimplemented` here (and in the other cases below)?

libvast/src/expression.cpp
```diff
@@ -182,6 +182,10 @@ bool operator<(const expression& x, const expression& y) {
 
 // -- free functions -----------------------------------------------------------
 
+expression hoist(const expression& expr) {
+  return caf::visit(hoister{}, expr);
```
dominiklohmann, line 186: Q: Why does this not take the expression by value and moves it into the hoister, if it creates a copy anyways?

This applies to the `normalize` and the other functions below as well.

libvast/src/system/spawn_arguments.cpp
```diff
@@ -35,10 +38,24 @@ normalized_and_validated(std::vector<std::string>::const_iterator begin,
                          std::vector<std::string>::const_iterator end) {
   if (begin == end)
     return caf::make_error(ec::syntax_error, "no query expression given");
-  if (auto e = to<expression>(caf::join(begin, end, " ")))
+  auto str = detail::join(begin, end, " ");
+  // TODO: improve error handling. Right now, we silently try to parse the
+  // query expression as Sigma rule. If it fails, we only print a debug log
+  // entry and move on to parsing the input as VAST expression.
```
dominiklohmann, line 44: Please write a follow-up story for this, it seems rather important.

libvast/vast/concept/parseable/vast/schema.hpp
```diff
@@ -182,13 +182,13 @@ struct symbol_table_parser : parser<symbol_table_parser> {
...
 template <>
-struct parser_registry<symbol_table> {
+struct parser_registry<symbol_map> {
   using type = symbol_table_parser;
 };
 
 namespace parsers {
 
-constexpr auto symbol_table = symbol_table_parser{};
+constexpr auto symbol_map = symbol_table_parser{};
```
dominiklohmann, line 191: I don't understand why this was renamed. Please undo, or rename the parser accordingly.

dominiklohmann (changes requested):
Works fine in practice, but I've had a fair bit of questions on the code itself.

libvast/vast/concept/parseable/vast/schema.hpp
mavam: The reason is that @tobim removed a class called `symbol_table`, which was part of the parser framework. I had to undo this because I'm actually using the `symbol_table` parser here internally. @tobim's `symbol_table_parser` is strictly speaking also not a "classic" symbol table, which is why I renamed it.

@tobim how do you feel about the renaming?

libvast/src/expression.cpp
mavam: Valid point, but slightly tangential. Right now we have a consistent API. If we do the change, we should do it for all free functions.

Mind creating a story?

libvast/vast/concept/parseable/vast/schema.hpp
dominiklohmann: I'm fine with the parser itself being named `symbol_map_parser` as well, e.g., either of these options:
- `parsers::symbol_table` and `symbol_table_parser`
- `parsers::symbol_map` and `symbol_map_parser`

libvast/src/detail/sigma.cpp
mavam: Since patterns are not yet support on field values, this is more or less irrelevant. We need to revisit the pattern generation anyway *after* we can do pattern matching.

dominiklohmann: Please add a comment then.

libvast/src/expression.cpp
dominiklohmann: Can do.

libvast/src/system/spawn_arguments.cpp
mavam: I think the behavior is actually correct here. The question is: maybe we should only fail if we have valid YAML but not a valid Sigma rule? Since this is the auto-detection part: we must go to the next form input after YAML parsing failed *without* terminating VAST.

dominiklohmann: Why not require a `--sigma` flag and support either instead of trying both?

mavam: This might be additive to lock in a specific syntax, but by default I like the UX that VAST tries all known options.

libvast/src/detail/sigma.cpp
mavam: Done.

libvast/src/detail/sigma.cpp
mavam: Story created.

libvast/src/detail/sigma.cpp
mavam: Yes, good idea.

libvast/src/system/spawn_arguments.cpp
mavam: I wrote a story for revisiting this as well.

libvast/vast/concept/parseable/vast/schema.hpp
mavam: Note that we have this now:

```cpp
using symbol_map = std::unordered_map<std::string, type>;
```

This type was called `symbol_table` before because the old symbol table parser class with the same name go removed. So now we're back to the old state, which works just fine in my opinion.

dominiklohmann: I'm not talking about `vast::symbol_map`, but rather about `parsers::symbol_map`.

For all other parsers, it's `X_parser` and `parsers::X`. That's not the case here.

mavam: You're right. Somehow we were barking up the wrong tree. 🐶

dominiklohmann (approved):

PR #1381, opened 2021-02-16
Revert "Fix potential race condition between evaluator and partition"

lava:
This reverts commit 3c932137af3398737d8e8ed18c46c141b2b191dc.

We empirically verified that this fixes a segfault that can occur when running an export containing a partition that is ejected from the LRU cache while the export is going on, but I don't really understand the root cause so far :'(

###  :memo: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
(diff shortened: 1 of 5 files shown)


tobim (approved):
I approved the original commit. I also wonder what is wrong with it, but getting the app to work again can come before the rest of the investigation.

PR #1382, opened 2021-02-16
Enable type extractors to support user defined types

tobim:

This brings back the possibility to run `vast count ":port == 80"`.

###  :memo: Checklist

- [x] Add an integration test.

By commit.


vast/integration/reference/node-zeek-dns-log/step_04.ref
```diff
@@ -0,0 +1 @@
+0
```
mavam, line 1: Can we adjust the baseline to be non-zero? Otherwise it's difficult to separate faulty behavior from expected.

mavam (commented):
Nice work!

I looked at the diff in detail and the implementation checks out. Nothing needs to be changed. I couldn't give it a spin locally right now, but it looks really good.

If somebody else could check out the branch to validate the hands-on behavior, that'd be great. The code is in good shape.

vast/integration/reference/node-zeek-dns-log/step_04.ref
tobim: The idea is to prove that this does not return the same result as the `:port` query in the previous step.

mavam: Ah, that makes sense. A comment about that would be nice.

dominiklohmann (approved):
This works nicely.

Needs a rebase and the docs need to be written and merged _before_ this PR gets merged.

PR #1385, opened 2021-02-17
Render help and documentation on stdout

mavam:
Interestingly, we did this already for the `manual` command. The idea is that users that explicitly request `help` or `documentation` are interested in getting the result as result—not on a side channel.

###  :memo: Checklist

- [x] Update scripts that use this output.


mavam:
I used `git grep '2>&1'` to see if we have any scripts that make use of the output. I only found one occurrence in the bash completions.

dominiklohmann (approved):

PR #1387, opened 2021-02-18
Simplify to_byte overloads

JoeLoser:

Problem:
- `to_byte` with an argument delegates to an impl function template to
  provide a nice error when the type isn't an `unsigned char`. This level
  of indirection is unnecessary.

Solution:
- Use `if constexpr` and `always_false_v` to achieve the same goal.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/byte.hpp
+++ b/libvast/vast/byte.hpp
@@ -25,6 +25,8 @@
 
 #include "vast/fwd.hpp"
 
+#include "vast/detail/type_traits.hpp"
+
 #include <type_traits>
 
 namespace vast {
@@ -94,23 +96,17 @@ constexpr IntegerType to_integer(byte b) noexcept {
   return static_cast<IntegerType>(b);
 }
 
-template <bool E, typename T>
-constexpr byte to_byte_impl(T t) noexcept {
-  static_assert(E, "to_byte(t) must be provided an unsigned char, otherwise "
-                   "data loss may occur. "
-                   "If you are calling to_byte with an integer contant use: "
-                   "to_byte<t>() version.");
-  return static_cast<byte>(t);
-}
-
-template <>
-constexpr byte to_byte_impl<true, unsigned char>(unsigned char t) noexcept {
-  return byte(t);
-}
-
 template <typename T>
 constexpr byte to_byte(T t) noexcept {
-  return to_byte_impl<std::is_same_v<T, unsigned char>, T>(t);
+  if constexpr (std::is_same_v<T, unsigned char>) {
+    return byte(t);
+  } else {
+    static_assert(detail::always_false_v<T>,
+                  "to_byte(t) must be provided an unsigned char, otherwise "
+                  "data loss may occur. "
+                  "If you are calling to_byte with an integer contant use: "
+                  "to_byte<t>() version.");
+  }
 }
 
 template <int I>


mavam (approved):
Makes sense, thanks for digging this one out!

mavam:
This made me wonder: `std::byte` C++17, so we could switch to that if the standard libraries from our compilers support it.

JoeLoser:
> This made me wonder: `std::byte` C++17, so we could switch to that if the standard libraries from our compilers support it.

Good observation. I think the min compilers supported by Vast should support `std::byte` right now.  I'll remove `byte.hpp` in favor of `std::byte` soon 👍

mavam:
Awesome, looking forward to i!

JoeLoser:
> Awesome, looking forward to it!

https://github.com/tenzir/vast/pull/1392 replaces `vast::byte` with `std::byte` as discussed. I think we can also replace `vast::span` with `std::span` for what its worth.

mavam:
> I think we can also replace `vast::span` with `std::span` for what its worth.

As I was writing my comment, I thought about that as well. However, `std::span` is C++20. [Compiler support](https://en.cppreference.com/w/cpp/compiler_support) for `std::span` is available as of GCC 10, Clang 7, and Apple Clang 10. Once we're at GCC 10, we can do this. (I just tried to include `<span>` while using `-std=c++17` and it works.)

JoeLoser:
> > I think we can also replace `vast::span` with `std::span` for what its worth.
> 
> As I was writing my comment, I thought about that as well. However, `std::span` is C++20. [Compiler support](https://en.cppreference.com/w/cpp/compiler_support) for `std::span` is available as of GCC 10, Clang 7, and Apple Clang 10. Once we're at GCC 10, we can do this. (I just tried to include `<span>` while using `-std=c++17` and it works.)

Ah right, `std::span` is C++20.  What's the process/timeline like for updating supported min compiler versions for Vast?  Do you have plans to migrate to using `-std=c++2a`?

dominiklohmann:
> Ah right, `std::span` is C++20. What's the process/timeline like for updating supported min compiler versions for Vast? Do you have plans to migrate to using `-std=c++2a`?

We're currently aiming to support Debian Buster (with Backports), and that only has GCC-8 at the moment.

mavam:
> [..] and that only has GCC-8 at the moment

That said, this is the _only_ requirement at the moment that prevents us from bumping the minimum compiler requirement. In general, we want to support the latest compiler when possible.

But I don't know the timeline. Maybe @satta? What's the Debian roadmap for adopting GCC 10, if any, for buster backports? I [could only find gcc-8](https://backports.debian.org/changes/buster-backports.html) there.

dominiklohmann:
> But I don't know the timeline.

Wait until the next Debian version is considered stable, and then a while until it was adopted so we can actually make the switch. Maybe sometime later this year.

satta:
> But I don't know the timeline. Maybe @satta? What's the Debian roadmap for adopting GCC 10, if any, for buster backports? I [could only find gcc-8](https://backports.debian.org/changes/buster-backports.html) there.

There is no "roadmap" for GCC10 in buster-backports as backports are not developed in an organized, directed fashion -- someone would need to take care of the backport and adopt it. There might be all sorts of gcc-glibc-etc. interdependencies going on that might complicate things, so don't expect anything with buster on it to get GCC10 anytime soon.

PR #1390, opened 2021-02-19
Add vast.log-queue-size option

dominiklohmann:

This makes the logger queue size configurable, and increases its default size from 8192 to 32768.

###  :memo: Checklist

n/t

### DIFF
--- a/libvast/vast/defaults.hpp
+++ b/libvast/vast/defaults.hpp
@@ -209,7 +209,7 @@ constexpr const caf::atom_value console_verbosity = caf::atom("info");
 constexpr const caf::atom_value file_verbosity = caf::atom("debug");
 
 /// Maximum number of log messages in the logger queue.
-constexpr const size_t queue_size = 8'192;
+constexpr const size_t queue_size = 32'768;
 
 /// Number of logger threads.
 constexpr const size_t logger_threads = 1;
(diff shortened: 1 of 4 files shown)


libvast/vast/defaults.hpp
```diff
@@ -209,7 +209,7 @@ constexpr const caf::atom_value console_verbosity = caf::atom("info");
 constexpr const caf::atom_value file_verbosity = caf::atom("debug");
 
 /// Maximum number of log messages in the logger queue.
-constexpr const size_t queue_size = 8'192;
+constexpr const size_t queue_size = 32'768;
```
mavam, line 212: Assuming an average log message size has an upper bound of ~200 bytes, we'd have 6.4MB as buffer. It doesn't hurt to shift this value by few powers of 2, e.g., 2^20.

200 bytes is generous. I looked at `vast start` in debug output and counted an average of 88 bytes:
```
awk '{ sum += length($0) } END { print sum/NR }'  /tmp/log
```

lava: I agree, but I'll approve it anyways for now in the spirit of "good enough". We can always increase the limit later on (even via CLI option now) it its still not enough.

lava (approved):

PR #1394, opened 2021-02-22
Improve API for copying expression utilities

dominiklohmann:

These functions copy by design, so it should be made clear from their interface that they copy. Additionally, this allows for moving the expression into the function, which removes the need to copy entirely.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/detail/sigma.cpp
+++ b/libvast/src/detail/sigma.cpp
@@ -320,7 +320,7 @@ caf::expected<expression> parse_search_id(const data& yaml) {
         }
         auto expr = all ? expression{conjunction(std::move(connective))}
                         : expression{disjunction(std::move(connective))};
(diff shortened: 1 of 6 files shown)


mavam:
Good start. You probably want to take it one step further and go into the respective visitors and change the const-reference lambda capture to non-const lvalues so that you can move individual AST nodes where possible.

mavam (approved):
Good start. You probably want to take it one step further and go into the respective visitors and change the const-reference lambda capture to non-const lvalues so that you can move individual AST nodes where possible.

dominiklohmann:
> You probably want to take it one step further and go into the respective visitors and change the const-reference lambda capture to non-const lvalues so that you can move individual AST nodes where possible.

We can always make that change in a follow-up PR.

PR #1396, opened 2021-02-22
Change the default batch size to 1024

tobim:

Since the default partitions size is a power of 2 it makes sense to do the same for the default batch size.

###  :memo: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,26 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- ⚠️ The default size of table slices (event batches) that is created from
+  `vast import` processes has been changed from 1000 to 1024.
+  [#1396](https://github.com/tenzir/vast/pull/1396)
+
 - 🐞 The disk monitor now correctly erases partition synopses from the meta index.
   [#1450](https://github.com/tenzir/vast/pull/1450)
+
+- ⚡️ A new `VASTRegisterPlugin` CMake function enables easy setup of the build
+  scaffolding required for plugins. Plugins can now be linked statically against
+  VAST. Configure with `--with-static-plugins` or build a static binary to link
+  all plugins built alongside VAST statically. All plugin build scaffoldings
+  must be adapted, older plugins do no longer work.
+  [#1445](https://github.com/tenzir/vast/pull/1445)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- ⚠️ The default batch size has been changed from 1000 to 1024.
```
mavam, line 16: Can you be more specific what you mean by batch? From a user perspective, it can mean a lot things across the data processing pipeline.

dominiklohmann (commented):
I don't think this change makes any sense unless you also change `vast.metrics.self-sink.slice-size`, which, as I've just noticed, always uses the default value, as `libvast/src/accountant/config.cpp` has a spelling error in the key and writes `slice` as `slize` in two places.

CHANGELOG.md
```diff
@@ -22,6 +26,12 @@ This changelog documents all notable user-facing changes of VAST.
   arguments are now disallowed instead of triggering a deprecation warning.
   [#1374](https://github.com/tenzir/vast/pull/1374)
 
+- 🧬 [Sigma](https://github.com/Neo23x0/sigma) rules are now a valid format to
+  represent query expression. VAST parses the `detection` attribute of a rule
+  and translates it into a native query expression. To run a query using a
+  Sigma rule, pass it on standard input, e.g., `vast export json < rule.yaml`.
+  [#1379](https://github.com/tenzir/vast/pull/1379)
```
dominiklohmann, line 33: Please remove.

tobim:
I also changed the slice size of the accountant to a power of 2.

lava (approved):
Looks good to me, the only thing I noticed was a typo in the word "typo" in one of the commit messages :D (`Fix a type in the accountant config `) But I think it's fine to just ignore this one.

vast.yaml.example
```diff
@@ -102,7 +102,7 @@ vast:
     # Configures if and how metrics should be ingested back into VAST.
     self-sink:
       enable: true
-      slice-size: 100
+      slice-size: 128
```
mavam, line 105: This change is not documented in the changelog. Is it not important?

tobim: I'd argue it is not.

PR #1397, opened 2021-02-23
Simplify directory recursive_size function

JoeLoser:

Problem:
- `recursive_size` has its own logic for "recursing" directories to get
  the total size of files, but this can be implemented directly more or
  less using `std::filesystem`.

Solution:
- Use `std::filesystem::recursive_directory_iterator` to do the heavy
  lifting of recursing into directories.

Note:
- This requires linking explicitly with `stdc++fs` flag for GCC < 9.1.0.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/directory.cpp
+++ b/libvast/src/directory.cpp
@@ -23,8 +23,10 @@
 
 #include <caf/streambuf.hpp>
 
+#include <filesystem>
 #include <fstream>
 #include <iterator>
+#include <system_error>
 
 #if VAST_POSIX
 #  include <sys/types.h>
@@ -107,24 +109,24 @@ const path& directory::path() const {
   return path_;
 }
 
-size_t recursive_size(const vast::directory& dir) {
-  size_t size = 0;
-  std::vector<vast::directory> stack{dir};
-  while (!stack.empty()) {
-    auto dir = std::move(stack.back());
-    stack.pop_back();
-    for (auto file : dir) {
-      if (file.is_regular_file()) {
-        if (auto sz = vast::file_size(file)) {
-          VAST_TRACE_SCOPE("{} += {}", file, *sz);
-          size += *sz;
-        }
-      } else if (file.is_directory()) {
-        stack.push_back(vast::directory{file});
-      }
+caf::expected<size_t> recursive_size(const vast::directory& dir) {
+  size_t total_size = 0;
+
+  const auto p = std::filesystem::path{dir.path().str()};
+  std::error_code ec{};
+  auto dir_itr = std::filesystem::recursive_directory_iterator(p, ec);
+  if (ec)
+    return caf::make_error(ec::filesystem_error, ec.message());
+
+  for (const auto& f : dir_itr) {
+    if (f.is_regular_file()) {
+      const auto size = f.file_size();
+      VAST_TRACE("{} += {}", f.path().string(), size);
+      total_size += size;
     }
   }
-  return size;
+
+  return total_size;
 }
 
 std::vector<path>
--- a/libvast/CMakeLists.txt
+++ b/libvast/CMakeLists.txt
@@ -331,6 +331,11 @@ add_library(libvast ${libvast_sources})
 add_library(vast::libvast ALIAS libvast)
 target_link_libraries(libvast PRIVATE libvast_internal)
 
+# GNU prior to 9.1 requires linking with stdc++fs when using std::filesystem library
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.1.0)
+  target_link_libraries(libvast PUBLIC stdc++fs)
+endif()
+
 set_target_properties(
   libvast
   PROPERTIES INTERPROCEDURAL_OPTIMIZATION_RELEASE ON
--- a/libvast/src/system/disk_monitor.cpp
+++ b/libvast/src/system/disk_monitor.cpp
@@ -62,11 +62,15 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
       // see noticeable overhead even on large-ish databases.
       // Nonetheless, if this becomes relevant we should switch to using
       // `inotify()` or similar to do real-time tracking of the db size.
-      auto size = recursive_size(self->state.dbdir);
-      VAST_VERBOSE("{} checks db-directory of size {} bytes", self, size);
-      if (size > self->state.high_water_mark && !self->state.purging) {
-        self->state.purging = true;
-        self->send(self, atom::erase_v);
+      if (const auto size = recursive_size(self->state.dbdir); !size) {
+        VAST_WARN("{} failed to calculate recursive size of {}: {}", self,
+                  self->state.dbdir, size.error());
+      } else {
+        VAST_VERBOSE("{} checks db-directory of size {} bytes", self, *size);
+        if (*size > self->state.high_water_mark && !self->state.purging) {
+          self->state.purging = true;
+          self->send(self, atom::erase_v);
+        }
       }
     },
     [self](atom::erase) {
@@ -114,18 +118,22 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
                         erased_ids)
               .then(
                 [=, sg = shared_guard](atom::done) {
-                  auto sz = recursive_size(self->state.dbdir);
-                  VAST_VERBOSE("{} erased ids from index; {} bytes "
-                               "left on disk",
-                               self, sz);
-                  if (sz > self->state.low_water_mark) {
-                    // Repeat until we're below the low water mark
-                    self->send(self, atom::erase_v);
+                  if (const auto size = recursive_size(self->state.dbdir);
+                      !size) {
+                    VAST_WARN("{} failed to calculate recursive size of {}: {}",
+                              self, self->state.dbdir, size.error());
+                  } else {
+                    VAST_VERBOSE("{} erased ids from index; {} bytes "
+                                 "left on disk",
+                                 self, *size);
+                    if (*size > self->state.low_water_mark) {
+                      // Repeat until we're below the low water mark
+                      self->send(self, atom::erase_v);
+                    }
                   }
                 },
-                [=, sg = shared_guard](caf::error e) {
-                  VAST_WARN("{} failed to erase from archive: {}", self,
-                            render(e));
+                [=, sg = shared_guard](caf::error err) {
+                  VAST_WARN("{} failed to erase from archive: {}", self, err);
                 });
           },
           [=, sg = shared_guard](caf::error e) {
--- a/libvast/vast/directory.hpp
+++ b/libvast/vast/directory.hpp
@@ -13,6 +13,8 @@
 
 #pragma once
 
+#include "vast/fwd.hpp"
+
 #include "vast/config.hpp"
 #include "vast/defaults.hpp"
 #include "vast/detail/iterator.hpp"
@@ -76,7 +78,7 @@ private:
 /// Calculates the sum of the sizes of all regular files in the directory.
 /// @param dir The directory to traverse.
 /// @returns The size of all regular files in *dir*.
-size_t recursive_size(const vast::directory& dir);
+caf::expected<size_t> recursive_size(const vast::directory& dir);
 
 /// Recursively traverses a directory and lists all file names that match a
 /// given filter expresssion.


mavam:
Now you're getting to the meaty pieces! VAST started with its own header in `vast/filesystem.hpp` way before there was a standard library implementation available. In the long run, we want to get rid of all the functionality in there that duplicates things from `std`. So the obvious candidate is `path`. `directory` is also superfluous because it can always be replaced with what you just did: `std::recursive_directory_iterator`.

Our goal is to take one piece at a time from our own header and replace it with the corresponding piece from the standard library. However, this requires a little bit of discussion first, because we use the `caf::expected<T>` as error monad in our code base. We still need to come up with a reasonable proposal on how to take, e.g., `std::create_directories` and lift it such that we can use it like this:

```cpp
if (auto result = attempt<std::create_directories>(path))
  f(*result); // work with result
else
  g(result.error()); // handle error
```

This is just an ad-hoc idea for composition. We still need to discuss with @tenzir/backend how to go after this. Maybe @tobim and @dominiklohmann can lead the discussion from here.

CMakeLists.txt
```diff
@@ -149,6 +149,11 @@ install(TARGETS libvast_internal EXPORT VASTTargets)
 # Require standard C++17
 target_compile_features(libvast_internal INTERFACE cxx_std_17)
 
+# GNU prior to 9.1 requires linking with -lstdc++fs when using std::filesystem library
+if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND CMAKE_CXX_COMPILER_VERSION STRLESS "9.1.0")
```
dominiklohmann, line 153: This should probably use `VERSION_LESS` rather than `STRLESS`.

libvast/src/directory.cpp
```diff
@@ -108,23 +109,18 @@ const path& directory::path() const {
...
-      } else if (file.is_directory()) {
-        stack.push_back(vast::directory{file});
-      }
+  size_t total_size = 0;
+
+  const auto p = std::filesystem::path{dir.path().str()};
+  for (const auto& f : std::filesystem::recursive_directory_iterator(p)) {
+    if (f.is_regular_file()) {
+      const auto size = f.file_size();
+      VAST_TRACE_SCOPE("{} += {}", f, size);
```
dominiklohmann, line 118:
```suggestion
      VAST_TRACE("{} += {}", f, size);
```

While you're at it, please change this as well.

dominiklohmann (commented):
The macOS CI seems to be failing because of a missing formatter, maybe you can adjust the log message to make it work so we're not trying to print internals of `std::filesystem`.

CMakeLists.txt
dominiklohmann: Also, thinking about this some more, this should probably link publicly against the `libvast` target rather than the `libvast_internal` target.

dominiklohmann:
> Our goal is to take one piece at a time from our own header and replace it with the corresponding piece from the standard library. However, this requires a little bit of discussion first, because we use `caf::expected<T>` as error monad in our code base. We still need to come up with a reasonable proposal on how to take, e.g., `std::create_directories` and lift it such that we can use it like this:
> 
> ```c++
> if (auto result = attempt<std::create_directories>(path))
>   f(*result); // work with result
> else
>   g(result.error()); // handle error
> ```
> 
> This is just an ad-hoc idea for composition. We still need to discuss with @tenzir/backend how to go after this. Maybe @tobim and @dominiklohmann can lead the discussion from here.

I talked about this with @mavam earlier and we came to the conclusion that it's best not to wrap the standard interface at all, but rather forward the error to the call-site by manually wrapping it using `caf::make_error(ec::filesystem_error, ...)`. We also don't use the filesystem API in many places—mostly the `lsvast` utility and within a few helper functions.

CMakeLists.txt
JoeLoser: Thanks for reminding me about `VERSION_LESS`. Just switched to that and also applied the linking flag to `libvast` rather than `libvast_internal`. 👍

libvast/src/directory.cpp
JoeLoser: Switched to `VAST_TRACE`.

JoeLoser:
> The macOS CI seems to be failing because of a missing formatter, maybe you can adjust the log message to make it work so we're not trying to print internals of `std::filesystem`.

Yeah, I saw that last night but hadn't  dug too deep into the `spdlog` -> `fmt` code path to see why it was failing on `macOS` only. `fmt` works fine for `std::filesystem::directory_entry` types by delegating to its `string` representation AFAIK. In any case, not a high priority to chase down. Instead, I changed the call site to get a string representation of the underlying path to keep things simple. Another alternative would be to introduce a `fmt` specialization which inherits from `fmt<std::string>` class template. That approach would be useful if we end up logging in more call sites with `std::filesystem` types when we rework `path` and `directory` to use their `std` counterparts.

JoeLoser:
> > Our goal is to take one piece at a time from our own header and replace it with the corresponding piece from the standard library. However, this requires a little bit of discussion first, because we use `caf::expected<T>` as error monad in our code base. We still need to come up with a reasonable proposal on how to take, e.g., `std::create_directories` and lift it such that we can use it like this:
> > ```c++
> > if (auto result = attempt<std::create_directories>(path))
> >   f(*result); // work with result
> > else
> >   g(result.error()); // handle error
> > ```
> > 
> > 
> > This is just an ad-hoc idea for composition. We still need to discuss with @tenzir/backend how to go after this. Maybe @tobim and @dominiklohmann can lead the discussion from here.
> 
> I talked about this with @mavam earlier and we came to the conclusion that it's best not to wrap the standard interface at all, but rather forward the error to the call-site by manually wrapping it using `caf::make_error(ec::filesystem_error, ...)`. We also don't use the filesystem API in many places—mostly the `lsvast` utility and within a few helper functions.

Seems reasonable to use the error code interface of `std::filesystem` and forward it with our own error type where appropriate. That's what I was thinking as well when perusing the code for `path/directory` in `vast`.  As @mavam pointed out, the observation I had is "can we replace these vast types with their std equivalents"?  It seems there isn't a fundamental reason why this shouldn't be possible -- just work that can be done to simplify the codebase.

dominiklohmann:
> It seems there isn't a fundamental reason why this shouldn't be possible -- just work that can be done to simplify the codebase.

Yes, that's exactly what we concluded as well.

Note that we won't be merging this PR until tomorrow; I'd like to avoid merging until after this month's release.

JoeLoser:
> > It seems there isn't a fundamental reason why this shouldn't be possible -- just work that can be done to simplify the codebase.
> 
> Yes, that's exactly what we concluded as well.
> 
> Note that we won't be merging this PR until tomorrow; I'd like to avoid merging until after this month's release.

Sounds like a plan.  The pipeline is passing now across all compilers, so it's ready to be merged once you cut the release. Thanks!

libvast/src/directory.cpp
```diff
@@ -108,23 +109,18 @@ const path& directory::path() const {
...
-          VAST_TRACE_SCOPE("{} += {}", file, *sz);
-          size += *sz;
-        }
-      } else if (file.is_directory()) {
-        stack.push_back(vast::directory{file});
-      }
+  size_t total_size = 0;
+
+  const auto p = std::filesystem::path{dir.path().str()};
+  for (const auto& f : std::filesystem::recursive_directory_iterator(p)) {
```
mavam, line 115: This function can throw, so we should use the overload with a `std::error_code`. This means the function needs to return `expected<size_t>`:
```cpp
std::error_code ec;
auto dir = std::filesystem::recursive_directory_iterator(p, ec);
if (ec)
   return caf::make_error(ec); // TODO: discuss out how to render the error
for (const auto& f : dir)
   // ...
```

@tenzir/backend  can you flesh out this error-handling strawman?

JoeLoser: Valid point. I just switched to use the error code API for `std::filesystem` here. In the case there is an error, I use the message as returned from `std::error_code` to pass through in the `caf::error` type. This gives the caller the most information and can choose what to do with it.

I'm not sure in this case what we're looking to do in the error cases from the caller side in `disk_monitor.cpp`. See what I've done and let me know.

dominiklohmann: Generally, there are three ways to handle errors in actor behaviors:

1. Log using `VAST_ERROR` and continue operation.
2. Return the error to the requesting actor (you can change the result type of the behavior lambda to `caf::result<void>` to be able to return errors.
3. Shut down the actor using `self->quit(error)`.

Now the disk monitor is running in a loop, so (2) doesn't quite make sense because it's requesting itself periodically to perform the check. (3) seems a bit harsh, as there's no need to shut down the disk monitor. So (1) it should be.

dominiklohmann: If you want me to, I can push the necessary adaptions directly onto your branch.

JoeLoser: Thanks for the explanation for error handling.  Do you mind pushing the necessary changes for error handling in the `disk_monitor.cpp` loop?

dominiklohmann: Can do, but might take me a few days—off of work until Monday.

dominiklohmann: I cannot push to your branch, but here are the remaining changes I'd make to this PR:

```diff
diff --git a/libvast/src/system/disk_monitor.cpp b/libvast/src/system/disk_monitor.cpp
index 9b656a4d3..ac38bf2f7 100644
--- a/libvast/src/system/disk_monitor.cpp
+++ b/libvast/src/system/disk_monitor.cpp
@@ -62,7 +62,10 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
       // see noticeable overhead even on large-ish databases.
       // Nonetheless, if this becomes relevant we should switch to using
       // `inotify()` or similar to do real-time tracking of the db size.
-      if (const auto size = recursive_size(self->state.dbdir); size) {
+      if (const auto size = recursive_size(self->state.dbdir); !size) {
+        VAST_WARN("{} failed to calculate recursive size of {}: {}", self,
+                  self->state.dbdir, size.error());
+      } else {
         VAST_VERBOSE("{} checks db-directory of size {} bytes", self, *size);
         if (*size > self->state.high_water_mark && !self->state.purging) {
           self->state.purging = true;
@@ -115,19 +118,22 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
                         erased_ids)
               .then(
                 [=, sg = shared_guard](atom::done) {
-                  if (const auto sz = recursive_size(self->state.dbdir); sz) {
+                  if (const auto size = recursive_size(self->state.dbdir);
+                      !size) {
+                    VAST_WARN("{} failed to calculate recursive size of {}: {}",
+                              self, self->state.dbdir, size.error());
+                  } else {
                     VAST_VERBOSE("{} erased ids from index; {} bytes "
                                  "left on disk",
-                                 self, *sz);
-                    if (*sz > self->state.low_water_mark) {
+                                 self, *size);
+                    if (*size > self->state.low_water_mark) {
                       // Repeat until we're below the low water mark
                       self->send(self, atom::erase_v);
                     }
                   }
                 },
-                [=, sg = shared_guard](caf::error e) {
-                  VAST_WARN("{} failed to erase from archive: {}", self,
-                            render(e));
+                [=, sg = shared_guard](caf::error err) {
+                  VAST_WARN("{} failed to erase from archive: {}", self, err);
                 });
           },
           [=, sg = shared_guard](caf::error e) {
diff --git a/libvast/vast/directory.hpp b/libvast/vast/directory.hpp
index 8f6fae328..5552f37a8 100644
--- a/libvast/vast/directory.hpp
+++ b/libvast/vast/directory.hpp
@@ -13,18 +13,18 @@
 
 #pragma once
 
+#include "vast/fwd.hpp"
+
 #include "vast/config.hpp"
 #include "vast/defaults.hpp"
 #include "vast/detail/iterator.hpp"
 #include "vast/path.hpp"
 
-#include <caf/expected.hpp>
-
 #if VAST_POSIX
 #  include <dirent.h>
 #else
 
-#include <functional>
+#  include <functional>
 
 namespace vast {
 struct DIR;
```

JoeLoser: Just applied your diff -- thanks for sharing.  Not sure why you couldn't push to my branch on my fork. 🤔 

In any case, I applied the diff but kept the `<caf/expected.hpp>` include in there as we return `caf::expected` from `recursive_size` now. Did you intentionally mean to drop that include?

dominiklohmann: Yes, the forward declaration from `fwd.hpp` suffices.

JoeLoser:
ping @mavam @dominiklohmann -- this MR compiles and builds fine locally with your proposed changes in regard to error handling. CI is currently failing for Debian builds due to some broken packages from what I can tell (outside the scope of this MR). 

I'm seeing errors like:

```
Processing triggers for libc-bin (2.28-10) ...
Warning: apt-key output should not be parsed (stdout is not a terminal)
OK
deb http://apt.llvm.org/buster/ llvm-toolchain-buster-9 main
E: Unable to correct problems, you have held broken packages.
Error: Process completed with exit code 100.
```

mavam:
Yeah, looks an unrelated change. Overall, everything seems to be in place. I'll let @dominiklohmann going to do the final merge.

dominiklohmann (approved):

dominiklohmann:
Merged despite the CI failure, which indeed is unrelated.

PR #1398, opened 2021-02-23
Remove dependency on 3rd party function for vast::directory

JoeLoser:

Problem:
- `vast::directory::filter_dir` has a `detail::function` type as a
  function parameter but this dependency on the 3rd party adopted
  `function` type is overkill.

Solution:
- Use `std::function` directly.

Note:
- There is now only one place in the codebase (`chunk.hpp`) that has a
  dependency on the 3rd party `function.hpp`. This is for storing the
  function as a nonstatic data member. It is not trival to use
  `std::function` here since the stored function is not allowed to be
  marked `noexcept`. However, if we remove the `noexcept` requirement on the
  deleter function, this breaks other things.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/directory.cpp
+++ b/libvast/src/directory.cpp
@@ -128,7 +128,7 @@ size_t recursive_size(const vast::directory& dir) {
 }
 
 std::vector<path>
-filter_dir(const path& dir, detail::function<bool(const path&)> filter,
+filter_dir(const path& dir, std::function<bool(const path&)> filter,
            size_t max_recursion) {
   std::vector<path> result;
   if (max_recursion == 0)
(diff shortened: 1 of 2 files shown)


mavam (commented):
This strikes me as reasonable simplification.

However, @dominiklohmann had some specific reasons for why he brought in the "fancy" version. So I'd like to hear his opinion on this proposal.

dominiklohmann:
It's fine to remove the use of the custom function type here.

For `chunk`, which is a very performance-critical piece of code for VAST, we must have a movable function more so than it being `noexcept`-correct. The latter is just a nice bonus, but the fact that `std::function` cannot be moved makes it impossible to use for `chunk`.

dominiklohmann (approved):

PR #1401, opened 2021-02-23
Add recursion limit for calls to flatten()

lava:

Add a recursion limit for flattening records.

NOTE: This is a quite minimalistic implementation of the corresponding story:
  - There are still `make_record()`, `merge()` and `convert()` in the same header that will do unbounded recursion.
  - While implementing I had the impression that it would make sense to have a "maximum record nesting depth" that is applied regardless of how whether the functions are implemented using recursion or not. (ie. now flatten() and unflatten() are asymmetric because the latter is implemented without recursion)

However, in particular the second point would require a bit more comprehensive change to catch all places, so I decided to put up this PR first so we can decide together how much more effort we want to invest.

###  :memo: Checklist

### DIFF
--- a/libvast/src/data.cpp
+++ b/libvast/src/data.cpp
@@ -175,14 +175,19 @@ size_t depth(const record& r) {
 namespace {
 
 template <class Iterator, class Sentinel>
-caf::optional<record>
-make_record(const record_type& rt, Iterator& begin, Sentinel end) {
+caf::optional<record> make_record(const record_type& rt, Iterator& begin,
+                                  Sentinel end, size_t max_recursion) {
+  if (max_recursion == 0) {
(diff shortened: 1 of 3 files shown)


mavam (approved):
Overall, looks good to me. I would suggest that we keep the `max_recursion` parameters consistent throughout the code base.

```sh
$ git grep 'max_recursion = '
libvast/vast/data.hpp:load_yaml_dir(const path& dir, size_t max_recursion = defaults::max_recursion);
libvast/vast/defaults.hpp:constexpr size_t max_recursion = 100;
libvast/vast/directory.hpp:           size_t max_recursion = defaults::max_recursion);
libvast/vast/schema.hpp:            size_t max_recursion = defaults::max_recursion);
```

I have no strong opinion whether we should call it `recursion_depth` or `max_recursion`.

PR #1405, opened 2021-02-24
Use response promises for flush listeners

dominiklohmann:

This is based on #1313—it makes the flush listener work using response promises instead of the old, more complicated setup. This should make it easier to maintain in the future.

@lava feel free to take this branch over at any time or to merge it into #1313 early.

###  :memo: Checklist

File-by-file.


libvast/src/system/index.cpp
```diff
@@ -332,6 +306,34 @@ void index_state::create_active_partition() {
   active_partition.capacity = partition_capacity;
   active_partition.id = id;
   VAST_DEBUG("{} created new partition {}", self, id);
+  // Move responsibility for fulfilling the flush promise to the active
+  // partitions.
+  if (flush_promises)
+    for (auto&& flush_promise : std::exchange(*flush_promises, {}))
```
lava, line 312: I don't quite understand how this works - it looks like all `flush_promises` are cleared out every time we create an active partition? That would not work in case we create several partitions in a row before shutting down.

libvast/src/system/index.cpp
```diff
@@ -657,21 +660,15 @@ index(index_actor::stateful_pointer<index_state> self,
...
       if (err && err != caf::exit_reason::unreachable) {
-        if (err != caf::exit_reason::user_shutdown)
-          VAST_ERROR("{} got a stream error: {}", self, render(err));
-        else
-          VAST_DEBUG("{} got a user shutdown error: {}", self, render(err));
+        VAST_ERROR("{} aborts with error: {}", self, err);
         // We can shutdown now because we only get a single stream from the
         // importer.
-        self->send_exit(self, err);
+        self->quit(err);
```
lava, line 671: Does `quit()` still call the exit handler?

libvast/vast/detail/notifying_stream_manager.hpp
```diff
@@ -13,28 +13,37 @@
...
+  const std::vector<caf::typed_response_promise<atom::flush>>& flush_promises,
+  const caf::stream_manager& manager) {
+  return !flush_promises.empty() && manager.inbound_paths_idle()
+         && manager.out().clean();
+}
+
+inline void notify_flush_listeners(
+  std::vector<caf::typed_response_promise<atom::flush>>& flush_promises) {
+  for (auto&& flush_promise : std::exchange(flush_promises, {}))
+    flush_promise.deliver(atom::flush_v);
```
lava, line 39: Doesn't it make the whole forwarding mechanism irrelevant if the stream stage itself already delivers the promises?

lava (commented):
I didn't review in-depth, but had a few questions after a first reading.

libvast/src/system/index.cpp
dominiklohmann: Ah, that was left in by mistake. Will remove.

libvast/vast/detail/notifying_stream_manager.hpp
dominiklohmann: The vector inside the shared ptr is empty as soon as the first active partition was created.

libvast/src/system/index.cpp
dominiklohmann: It simply moves them over from the `shared_ptr<vector<rp>>` from inside the index to those inside the active partitions, and then takes the last fulfilled rp from the active partitions and fulfills the original promise with it.

lava: Yeah, but how do the promises get back out one active partition to the next? Consider the sequence

```
1. Subscribe arrives in the index
2. An active partition P1 is created, the response promised is moved inside
3. More data arrives, the old active partition is decomissioned and another one P2 is created
4. Stream ends
```

Now P2 should deliver the response promises after its stream is clean, but how does it get them?

dominiklohmann: Step 2 in your sequence is wrong; the `forwarded_flush_promises` are not moved into the active partition, but rather shared with it, and when a new active partition is created it also receives all not yet fulfilled `forwarded_flush_promises`.

dominiklohmann:
Closing; we'll revisit this once we have the partition builder as currently planned.

PR #1406, opened 2021-02-26
Replace ::stat call with std::filesystem equivalents

JoeLoser:

Problem:
- Use of `::stat` system call is discouraged when it can be replaced
  with higher level standard facilities from `std::filesystem`.

Solution:
- Replace `::stat` call used to get the last modified time and file size
  directly with their equivalent functions from `std::filesystem`. This
  also allows us to work with strong types in the `partition_diskstate`
  which is nice.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/system/disk_monitor.cpp
+++ b/libvast/src/system/disk_monitor.cpp
@@ -17,16 +17,14 @@
 #include <filesystem>
 #include <system_error>
 
-#include <sys/stat.h>
-
 namespace vast::system {
 
 namespace {
 
 struct partition_diskstate {
   uuid id;
-  off_t filesize;
-  time_t mtime;
+  std::uintmax_t filesize;
+  std::filesystem::file_time_type mtime;
 };
 
 template <typename Fun>
@@ -101,8 +99,8 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
                                            self->state.dbdir, err));
       // TODO(ch20006): Add some check on the overall structure on the db dir.
       std::vector<partition_diskstate> partitions;
-      for (const auto& file : index_dir) {
-        auto partition = file.path().stem().string();
+      for (const auto& entry : index_dir) {
+        auto partition = entry.path().stem().string();
         if (partition == "index.bin")
           continue;
         uuid id;
@@ -110,11 +108,17 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
           VAST_VERBOSE("{} failed to find partition {}", self, partition);
           continue;
         }
-        // TODO: Wrap a more generic `stat()` using `vast::path`.
-        struct stat statbuf;
-        if (::stat(file.path().c_str(), &statbuf) < 0)
-          continue;
-        partitions.push_back({id, statbuf.st_size, statbuf.st_mtime});
+        if (entry.is_regular_file()) {
+          std::error_code err{};
+          const auto file_size = entry.file_size(err);
+          const auto mtime = entry.last_write_time(err);
+          if (!err && file_size != static_cast<std::uintmax_t>(-1))
+            partitions.push_back({id, file_size, mtime});
+          else
+            VAST_WARN("{} failed to get file size and last write time for "
+                      "partition {}",
+                      self, partition);
+        }
       }
       if (partitions.empty()) {
         VAST_VERBOSE("{} failed to find any partitions to delete", self);


JoeLoser:
@mavam I see one of the integration tests is failing. Can you help me understand why as a result of this change?

tobim (dismissed):
Thank you for trying to improve this code. I'm not sure that a conversion to use the `std::filesystem` API actually achieves that though. 

The filesystem functions must be implemented in terms of syscalls as well, and I'm pretty sure they will both just call `stat` internally. Calling `stat` twice as often as necessary needs a good justification.

Is your goal to port the code to a non-posix platform?

libvast/src/system/disk_monitor.cpp
```diff
@@ -15,6 +15,8 @@
 #include <caf/settings.hpp>
 #include <caf/typed_event_based_actor.hpp>
 
+#include <filesystem>
```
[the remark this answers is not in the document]
JoeLoser: > Thank you for trying to improve this code. I'm not sure that a conversion to use the `std::filesystem` API actually achieves that though.
> 
> The filesystem functions must be implemented in terms of syscalls as well, and I'm pretty sure they will both just call `stat` internally. Calling `stat` twice as often as necessary needs a good justification.
> 
> Is your goal to port the code to a non-posix platform?

I don't have a goal of porting in mind, but not making system calls directly would help in that regard. This was just a drive-by refactoring when reading the code as I noticed the other day that all of `vast::path` and `vast::directory` could be replaced with their `std::filesystem` equivalents. Personally (others may not share this same opinion), I find the `std::filesystem` functions easier to read than C-style code with making a struct that gets filled in by a system call. This is more canonical C++ I'd argue.

If performance is a concern here in the disk monitor, I'm fine closing the MR as this change would result in twice as many `stat` system calls as you pointed out I suspect.

libvast/src/system/disk_monitor.cpp
```diff
@@ -86,11 +88,12 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
...
         }
-        // TODO: Wrap a more generic `stat()` using `vast::path`.
-        struct stat statbuf;
-        if (::stat(file.complete().str().c_str(), &statbuf) < 0)
-          continue;
-        partitions.push_back({id, statbuf.st_size, statbuf.st_mtime});
+        const auto f = std::filesystem::path{file.str()};
+        std::error_code ec{};
+        const auto file_size = std::filesystem::file_size(f, ec);
+        const auto mtime = std::filesystem::last_write_time(f, ec);
```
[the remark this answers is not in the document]
JoeLoser: I can add that, no problem. Let's decide if we want to keep pursuing this MR at all before I make any modifications to it.

libvast/src/system/disk_monitor.cpp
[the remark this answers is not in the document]
lava: On the one hand, performance in the "erase" path is probably not a major concern, since this should only be called relatively infrequently.

On the other hand, doing unnecessary work also does not feel  very C++-like.

From how I understand `std::filesystem`, the intended solution here would probably be to use a directory iterator, since struct `directory_entry` already has member functions `file_size` and `last_write_time`, which are allowed to cache the information. Although as far as I can tell only libcxx currently supports the caching.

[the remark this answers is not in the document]
JoeLoser: > On the one hand, performance in the "erase" path is probably not a major concern, since this should only be called relatively infrequently.
> 
> On the other hand, doing unnecessary work also does not feel very C++-like.
> 
> From how I understand `std::filesystem`, the intended solution here would probably be to use a directory iterator, since struct `directory_entry` already has member functions `file_size` and `last_write_time`, which are allowed to cache the information. Although as far as I can tell only libcxx currently supports the caching.

Good observation regarding the the `directory_entry` caching of file status -- I noticed that the other day after this PR opened, but did not yet mention it here. 

Note that https://github.com/tenzir/vast/pull/1411 changes the surrounding structure here to use `std::filesystem` with a `directory_iterator` such that we are working with `directory_entry` types in the loop. This would give rise in this PR to take advantage of that like you mentioned.

I'd say it's a quality of implementation concern with whether `libc++` or `libstdc++` caches the `directory_entry` status info for paths.  I'm glad that you brought up the erase performance here in the `disk_monitor` -- thanks for providing some context.

With that being said, @lava @dominiklohmann, do we want to move forward with this PR after https://github.com/tenzir/vast/pull/1411 lands?

libvast/src/system/disk_monitor.cpp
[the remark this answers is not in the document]
JoeLoser: Just added a warning 👍

libvast/src/system/disk_monitor.cpp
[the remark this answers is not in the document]
dominiklohmann: I'll leave the decision up to @lava, as he wrote most of the disk monitor and should make this decision.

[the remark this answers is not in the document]
mavam: I agree that @lava should pull the trigger here.

My 2 cents: assume that the caching takes place and work with `directory_entry`.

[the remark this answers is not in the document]
lava: Yes, I agree, I'm fine with merging this after a rebase.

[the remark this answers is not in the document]
JoeLoser: @lava just rebased. Please take a look again. Thanks!

libvast/src/system/disk_monitor.cpp
```diff
@@ -110,11 +108,15 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
...
           continue;
         }
-        // TODO: Wrap a more generic `stat()` using `vast::path`.
-        struct stat statbuf;
-        if (::stat(file.path().c_str(), &statbuf) < 0)
-          continue;
-        partitions.push_back({id, statbuf.st_size, statbuf.st_mtime});
+        std::error_code ec{};
+        const auto file_size = std::filesystem::file_size(file.path(), ec);
+        const auto mtime = std::filesystem::last_write_time(file.path(), ec);
```
lava, line 113: Didn't you want to switch this to `file.file_size(ec)`/`file.last_write_time(ec)` after #1411 has been merged?

JoeLoser: Yes, thanks for the reminder. I will make that change later tonight -- sorry about that.

JoeLoser: CI is passing now with using the `file_size` and `last_write_time` directly as member function calls from the corresponding `directory_entry` type as discussed.

libvast/src/system/disk_monitor.cpp
```diff
@@ -101,20 +99,27 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
...
         if (!parsers::uuid(partition, id)) {
           VAST_VERBOSE("{} failed to find partition {}", self, partition);
           continue;
         }
-        // TODO: Wrap a more generic `stat()` using `vast::path`.
-        struct stat statbuf;
-        if (::stat(file.path().c_str(), &statbuf) < 0)
-          continue;
-        partitions.push_back({id, statbuf.st_size, statbuf.st_mtime});
+
```
mavam, line 111:
```suggestion
```

libvast/src/system/disk_monitor.cpp
```diff
@@ -101,20 +99,27 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
...
+        if (entry.is_regular_file()) {
+          std::error_code err{};
+          const auto file_size = entry.file_size(err);
+          const auto mtime = entry.last_write_time(err);
+          if (!err && file_size != static_cast<std::uintmax_t>(-1))
```
mavam, line 116: I [tried to find](https://en.cppreference.com/w/cpp/filesystem/directory_entry/last_write_time) when `last_write_time` can return a value with all bits set to 1, and why it's an error. At least high-level docs don't hint at an issue. Mind sharing you insight on this check?

libvast/src/system/disk_monitor.cpp
```diff
@@ -101,20 +99,27 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
...
+            partitions.push_back({id, file_size, mtime});
```
mavam, line 117:
```suggestion
            partitions.emplace_back(id, file_size, mtime);
```

mavam (approved):
Looks good, just a few nits.

libvast/src/system/disk_monitor.cpp
JoeLoser: The first part of the `if` check here is to catch the error case for `last_write_time` per https://en.cppreference.com/w/cpp/filesystem/directory_entry/last_write_time where it states:

> The overload taking a std::error_code& parameter sets it to the OS API error code if an OS API call fails

The second part of the `if` check here is to catch the `file_size` error case per https://en.cppreference.com/w/cpp/filesystem/file_size where it states:

> The non-throwing overload returns static_cast<std::uintmax_t>(-1) on errors.

Does that answer your question? I think you may have misread the code as the check isn't for `last_write_time` having all its bits set.

libvast/src/system/disk_monitor.cpp
JoeLoser: Just changed to use `emplace_back` so we get the move construction here. 👍

JoeLoser:
> Looks good, just a few nits.

Thanks, just pushed up the corresponding changes. Let me know if there is anything else. Appreciate the review.

libvast/src/system/disk_monitor.cpp
JoeLoser: We can't do an `emplace_back` actually since the `partition_diskstate` isn't move constructible. Just changed it back to use `push_back` which will copy construct the `partition_diskstate` instances.

libvast/src/system/disk_monitor.cpp
JoeLoser: ping @lava -- how does this look to you?

libvast/src/system/disk_monitor.cpp
mavam: Thanks for providing the references.

> I think you may have misread the code as the check isn't for `last_write_time` having all its bits set.

Yes, my knee-jerk reaction to seeing `static_cast<size_t>(-1)` (or for other unsigned arithmetic types) is that someone uses a lazy way to get the largest possible value, instead of using `numeric_limits<size_t>::max()`.

mavam:
Not quite sure what's up with CI. @dominiklohmann?

dominiklohmann:
> Not quite sure what's up with CI. @dominiklohmann?

It's this weird connection issue we're only seeing in CI again. Let's merge regardless.

I've dismissed @tobim's change request since that was addressed already (please do so yourself if you request changes over just commenting), and approved because @lava told me earlier he was just waiting for that change request to disappear. 🤷‍♂️

PR #1411, opened 2021-03-03
Migrate recursive_size to use std::filesystem

JoeLoser:

Problem:
- `recursive_size` function uses `std::filesystem` in its
  implementation, but its function argument and usage from callers still
  use `vast::path`.

Solution:
- Switch to `std::filesystem::path` for the function argument and
  convert callers to use `std::filesystem` equivalents.
- Fix additional uses of `vast::path` in favor of
  `std::filesystem::path` throughout `disk_monitor` to keep things simple
  rather than converting between the path types where required.

This gets us one step closer to removing `vast::directory` entirely in favor of `std::filesystem` equivalents.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/system/disk_monitor.cpp
+++ b/libvast/src/system/disk_monitor.cpp
@@ -72,7 +72,17 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
         VAST_VERBOSE("{} checks db-directory of size {} bytes", self, *size);
         if (*size > self->state.high_water_mark && !self->state.purging) {
           self->state.purging = true;
-          self->send(self, atom::erase_v);
+          // TODO: Remove the static_cast when switching to CAF 0.18.
+          self
+            ->request(static_cast<disk_monitor_actor>(self), caf::infinite,
+                      atom::erase_v)
+            .then(
+              [] {
+                // nop
+              },
+              [=](const caf::error& err) {
+                VAST_ERROR("{} failed to purge db-directory: {}", self, err);
+              });
         }
       }
     },
@@ -85,8 +95,10 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
       const auto index_dir
         = std::filesystem::directory_iterator(self->state.dbdir / "index", ec);
       if (ec)
-        return caf::make_error(ec::filesystem_error, "db-directory does not "
-                                                     "exist");
+        return caf::make_error(ec::filesystem_error, //
+                               fmt::format("failed to find index in "
+                                           "db-directory at {}: {}",
+                                           self->state.dbdir, ec));
       // TODO(ch20006): Add some check on the overall structure on the db dir.
       std::vector<partition_diskstate> partitions;
       for (const auto& file : index_dir) {
--- a/libvast/src/system/disk_monitor.cpp
+++ b/libvast/src/system/disk_monitor.cpp
@@ -7,13 +7,15 @@
 #include "vast/directory.hpp"
 #include "vast/error.hpp"
 #include "vast/logger.hpp"
-#include "vast/path.hpp"
 #include "vast/system/archive.hpp"
 #include "vast/uuid.hpp"
 
 #include <caf/detail/scope_guard.hpp>
 #include <caf/settings.hpp>
 #include <caf/typed_event_based_actor.hpp>
+
+#include <filesystem>
+#include <system_error>
 
 #include <sys/stat.h>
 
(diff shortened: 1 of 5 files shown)


libvast/src/system/disk_monitor.cpp
```diff
@@ -78,11 +80,12 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
       // have finished or we encountered an error.
       auto shared_guard
         = make_shared_guard([=] { self->state.purging = false; });
-      directory index_dir = self->state.dbdir / "index";
+      auto index_dir
+        = std::filesystem::directory_iterator(self->state.dbdir / "index");
```
dominiklohmann, line 84: Please switch to the error code version here, change the return type of the lambda to `caf::result<void>`, and return an error created using `caf::make_error(ec::filesystem_error, ...)` when the the call failed. All other places where this lambda returned normally can be replaced with `return {};`.

dominiklohmann: The error can then be received on the callsite like this:

Replace:

```
self->send(disk_monitor, atom::erase_v)
``` 

with

```
self->request(disk_monitor, caf::infinite, atom::erase_v).then(
  []() {
    // nop, all good
  },
  [=](const caf::error& err) {
    VAST_ERROR("{} failed to ...: {}`, self, err);
  });
```

dominiklohmann (commented):
Great change, just one missing check for errors that is a bit more complicated. I explained how to return errors from actor behaviors in a comment, and also showed how to check for errors on the receiving end of the messaging.

libvast/src/system/disk_monitor.cpp
JoeLoser: I switched to the `error_code` overload, but I had some issues making the call sites within `disk_monitor.cpp` work with your suggestion. I tried replacing `disk_monitor` with `self->state.db_dir` in the two call sites, but that didn't quite work as the handle. Can you elaborate a bit more on changing the `send` calls to `request` with appropriate error handling?

By the way, to contribute to my branch, it should suffice to do something like:

```
git remote add joe git@github.com:JoeLoser/vast.git
git fetch joe
git checkout -b recursive-size joe/recursive-size
// make changes and commit
git push origin joe/recursive-size
```

I don't mind if you push onto my branch when it makes sense as it may be easier than explaining the code over PR.

dominiklohmann: I pushed the necessary changes. `self` is a stateful pointer to the disk monitor actor, and due to a deficiency in the CAF version we're using (fixed upstream already), the pointer must be dereferenced manually to specify the receiving actor when using `request` over `send`. That you really could not know. 

Fwiw I had to reset my local VAST repository to add the remote successfully, which is probably why I thought I couldn't push last time around.

dominiklohmann: Also, if you ever want a quick intro on working with CAF, feel free to just ask.

libvast/src/system/disk_monitor.cpp
```diff
@@ -72,7 +72,14 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
         VAST_VERBOSE("{} checks db-directory of size {} bytes", self, *size);
         if (*size > self->state.high_water_mark && !self->state.purging) {
           self->state.purging = true;
-          self->send(self, atom::erase_v);
+          self->request(*self, caf::infinite, atom::erase_v)
+            .then(
+              [] {
```
JoeLoser, line 77: Do we need this overload at all? I'd expect `caf` to pass through if there is no match in the overload set in the `then` implementation.

libvast/src/system/disk_monitor.cpp
```diff
@@ -85,8 +92,10 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
       const auto index_dir
         = std::filesystem::directory_iterator(self->state.dbdir / "index", ec);
       if (ec)
-        return caf::make_error(ec::filesystem_error, "db-directory does not "
-                                                     "exist");
+        return caf::make_error(ec::filesystem_error, //
```
JoeLoser, line 95: Did you mean to have `//` here at the end?

libvast/src/system/disk_monitor.cpp
dominiklohmann: It's not strictly needed, but if we ever change the return type of the actor behavior, we want to actually have compile errors for all call sites. If we omit the handler and use the default-constructed one we'll miss out on those errors, which is why I generally prefer specifying both explicitly.

dominiklohmann (approved):

libvast/src/system/disk_monitor.cpp
JoeLoser: Thanks for pushing the commit -- I left two small comments regarding the changes. Glad you got the remote set up -- I suspect it'll be useful in the future 👍 

As for the CAF intro, I'll probably take you up on that in the future once I get a bit more understanding of the codebase and how Vast uses it. So far, it seems to rely on the actors a fair bit.

libvast/src/system/disk_monitor.cpp
dominiklohmann: This is just forcing a line break because clang-format is stupid.

libvast/src/system/disk_monitor.cpp
JoeLoser: Got it; makes sense.

libvast/src/system/disk_monitor.cpp
dominiklohmann: Turns out I actually had to static_cast to the typed actor handle type. The switch to CAF 0.18 cannot come soon enough... If only that refactoring wasn't as big of a change and not a breaking change. 🙄

JoeLoser: Glad you spotted the error quickly -- I just noticed that when building locally and the CI jobs failing as well.

PR #1416, opened 2021-03-04
Fix the static build

dominiklohmann:

The standard library for the static build seems be missing the ostream operator<< overload for some std::filesystem stuff.

**Edit:** Actually it did have them hidden behind implicit conversions, but {fmt} 7 is more strict about `has_formatter` than old versions, and that caused our SFINAE for the fallback formatters not to pick it up.

###  :memo: Checklist

I've manually triggered a static build run for this branch: https://github.com/tenzir/vast/actions/workflows/static-binary.yaml?query=branch%3Atopic%2Ffix-static-build

### DIFF
--- a/libvast/src/system/disk_monitor.cpp
+++ b/libvast/src/system/disk_monitor.cpp
@@ -91,14 +91,14 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
       // have finished or we encountered an error.
       auto shared_guard
         = make_shared_guard([=] { self->state.purging = false; });
-      std::error_code ec{};
+      auto err = std::error_code{};
       const auto index_dir
-        = std::filesystem::directory_iterator(self->state.dbdir / "index", ec);
-      if (ec)
+        = std::filesystem::directory_iterator(self->state.dbdir / "index", err);
+      if (err)
         return caf::make_error(ec::filesystem_error, //
                                fmt::format("failed to find index in "
                                            "db-directory at {}: {}",
-                                           self->state.dbdir, ec));
+                                           self->state.dbdir, err));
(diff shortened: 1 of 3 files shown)


mavam (approved):
Makes sense.

While you're at it, you may want to rename the local variable `ec`, which is confusing wrt. the `ec` enum.

libvast/src/system/disk_monitor.cpp
```diff
@@ -91,14 +91,14 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
       // have finished or we encountered an error.
       auto shared_guard
         = make_shared_guard([=] { self->state.purging = false; });
-      std::error_code ec{};
-      const auto index_dir
-        = std::filesystem::directory_iterator(self->state.dbdir / "index", ec);
-      if (ec)
+      std::error_code code{};
```
mavam, line 94:
```suggestion
      std::error_code err{};
```

This is our go-to name for local error variables that doesn't conflict with types. In longer functions, seeing `code` without the type is harder to reason about than seeing `err`.

Not worth changing, but worth pointing out, perhaps for future occurrences.

dominiklohmann: Meh, `err` doesn't convey whether it's code or category.

dominiklohmann:
Okay, this turned out to be a lot more complicated than I had first anticipated. @mavam, I would like to merge this on the sooner side, so please take another look.

mavam (approved):
Makes sense to dispatch on the {fmt} version and define the trait accordingly.

tobim (approved):
Verified by the manually triggered static build.

dominiklohmann:
I squashed and rewrote the commit message with more detail after the approval.

PR #1417, opened 2021-03-04
RFC: Add a clang-tidy configuration file

tobim:

This is a proposal for which clang-tidy checks can be enabled on VAST.

I'm using a blacklist approach here. Everything is enabled, and checks that don't fit our style are explicitly disabled. That comes with the advantage that newly introduced checks will be discovered automatically. But it is certainly not the way clang-tidy is intended. When a check is explicitly disabled, all redirects to the same check must be disabled as well.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,10 @@
+---
+Checks:                '*,-clang-analyzer-alpha*,-llvmlibc-*,-google-*,-fuchsia-*,-hicpp-*,-bugprone-lambda-function-name,-modernize-use-trailing-return-type,-readability-braces-around-statements,-readability-named-parameter,-readability-uppercase-literal-suffix'
+WarningsAsErrors:      ''
+HeaderFilterRegex:     '/vast/'
+AnalyzeTemporaryDtors: false
+CheckOptions:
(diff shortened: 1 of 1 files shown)


tobim:
I just found this proposal from @lava in another repo:
```
---
# TODO: Enable the following checks for our code base:
#
#     google-build-using-namespace
#     clang-analyzer-optin.cplusplus.VirtualCall
#     clang-analyzer-core.uninitialized.Branch
#
# They contain checks that we would generally like to enable,
# but can not at the time of this writing either due to legacy
# issues present in the codebase or bugs in the checks itself.
Checks:
"\
clang-diagnostic-*,\
clang-analyzer-*,\
google-*,\
-google-readability-braces-around-statements,\
-google-readability-namespace-comments,\
-google-build-using-namespace,\
-google-readability-todo,\
-google-runtime-references,\
-clang-analyzer-optin.cplusplus.VirtualCall,\
-clang-analyzer-core.uninitialized.Branch,\
misc-use-after-move,\
modernize-use-nullptr,\
modernize-use-override,\
readability-redundant-string-cstr\
"
FormatStyle: file
WarningsAsErrors: '*'
...
```

dominiklohmann (approved):

PR #1419, opened 2021-03-05
Replace vast::directory with std::filesystem in segment_store test

JoeLoser:

Problem:
- `segment_store` unit tests uses `vast::directory` which is going away
  soon.

Solution:
- Replace `vast::directory` uses with `std::filesystem`.
- Replace `vast::path` uses with `std::filesystem::path`.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/test/segment_store.cpp
+++ b/libvast/test/segment_store.cpp
@@ -22,10 +22,13 @@
 #include "vast/concept/printable/to_string.hpp"
 #include "vast/concept/printable/vast/uuid.hpp"
 #include "vast/detail/narrow.hpp"
-#include "vast/directory.hpp"
 #include "vast/ids.hpp"
 #include "vast/si_literals.hpp"
 #include "vast/table_slice.hpp"
+
+#include <algorithm>
+#include <filesystem>
+#include <system_error>
 
 using namespace vast;
 using namespace binary_byte_literals;
@@ -37,7 +40,7 @@ struct fixture : fixtures::deterministic_actor_system_and_events {
     store = segment_store::make(directory / "segments", 512_KiB, 2);
     if (store == nullptr)
       FAIL("segment_store::make failed to allocate a segment store");
-    segment_path = store->segment_path();
+    segment_path = std::filesystem::path{store->segment_path().str()};
     // Approximates an ID range for [0, max_id) with 100, because
     // `make_ids({{0, max_id}})` unfortunately leads to performance
     // degradations.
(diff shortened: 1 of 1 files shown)


JoeLoser:
The AppleClang CI job seems to have spuriously failed for this PR but it should be unrelated to these changes. This PR only addresses purely only one specific unit test in isolation. Note the other CI jobs passed just fine @dominiklohmann

libvast/test/segment_store.cpp
```diff
@@ -52,11 +55,14 @@ struct fixture : fixtures::deterministic_actor_system_and_events {
...
-      if (file.is_regular_file())
-        result.emplace_back(std::move(file));
+    std::vector<std::filesystem::path> result;
+    std::error_code err{};
+    auto dir = std::filesystem::directory_iterator{segment_path, err};
+    if (!err) {
+      std::copy_if(std::filesystem::begin(dir), std::filesystem::end(dir),
+                   std::back_inserter(result),
+                   [&](const auto& entry) { return entry.is_regular_file(); });
+    }
```
mavam, line 65:
```suggestion
    auto dir = std::filesystem::directory_iterator{segment_path, err};
    REQUIRE(!err);
    std::copy_if(std::filesystem::begin(dir), std::filesystem::end(dir),
                 std::back_inserter(result),
                 [&](const auto& entry) { return entry.is_regular_file(); });
```

I haven't tested this, but failing on error might be what we want here. Otherwise we return less query results than expected.

JoeLoser: I just changed it to ensure `err` isn't true as you suggested. I'm fine with that approach. We don't have any tests that would get us in a case where `err` would be `true` from what I could tell, so it's not a biggie from my perspective.

dominiklohmann: REQUIRE is basically an assertion, and I think asserting is absolutely correct here. The tests cannot continue if there is an error.

dominiklohmann:
> The AppleClang CI job seems to have spuriously failed for this PR but it should be unrelated to these changes 

We sometimes have repeated connection failures between server and client in CI, which we can never reproduce locally. I suspect it's something with GitHub's networking setup. We usually just re-run these.

libvast/test/segment_store.cpp
dominiklohmann: I took a look at the code some more. This is in a utility function that gets called for tests on invalid stores as well, so not doing anything on error is absolutely correct. The returned vector will then be empty, which is checked for.

dominiklohmann (approved):

PR #1420, opened 2021-03-05
Handle some clang-tidy warnings

dominiklohmann:

This is just me playing with #1417.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,5 +1,17 @@
 ---
-Checks:                '*,-clang-analyzer-alpha*,-llvmlibc-*,-google-*,-fuchsia-*,-hicpp-*,-bugprone-lambda-function-name,-modernize-use-trailing-return-type,-readability-braces-around-statements,-readability-named-parameter,-readability-uppercase-literal-suffix'
+Checks: >
+  *,
+  -clang-analyzer-alpha*,
+  -llvmlibc-*,
+  -google-*,
+  -fuchsia-*,
+  -hicpp-*,
+  -bugprone-lambda-function-name,
+  -modernize-use-trailing-return-type,
+  -readability-braces-around-statements,
+  -readability-named-parameter,
+  -readability-uppercase-literal-suffix,
+  -cppcoreguidelines-pro-type-reinterpret-cast
 WarningsAsErrors:      ''
 HeaderFilterRegex:     '/vast/'
 AnalyzeTemporaryDtors: false
@@ -7,4 +19,3 @@ CheckOptions:
   - key:             readability-implicit-bool-conversion.AllowPointerConditions
     value:           true
 ...
-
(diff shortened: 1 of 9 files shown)


libvast/vast/detail/assert.hpp
```diff
@@ -22,6 +22,7 @@
 #  define VAST_ASSERT(expr)                                                    \
     do {                                                                       \
       if (static_cast<bool>(expr) == false) {                                  \
+        /* NOLINTNEXTLINE */                                                   \
```
tobim, line 25: Did you try using `fmt` instead?
```
fmt::print(stderr, "{}:{}: assertion failed '{}'\n", __FILE__,          \
           __LINE__, #expr);
```

.clang-tidy
```diff
@@ -1,10 +1,21 @@
...
+  -llvmlibc-*,
+  -google-*,
+  -fuchsia-*,
+  -hicpp-*,
+  -bugprone-lambda-function-name,
+  -modernize-use-trailing-return-type,
+  -readability-braces-around-statements,
+  -readability-named-parameter,
+  -readability-uppercase-literal-suffix,
+  -cppcoreguidelines-pro-type-reinterpret-cast
```
tobim, line 14: We should sort the globs and non-globs alphabetically.

libvast/vast/detail/assert.hpp
dominiklohmann: I think this is a bad idea. An assertion macro should call as low level of an API as possible.

.clang-tidy
dominiklohmann: Good idea.

tobim (approved):

libvast/src/detail/load_plugin.cpp
```diff
@@ -62,30 +62,28 @@ load_plugin(path file, caf::actor_system_config& cfg) {
...
+    return std::move(plugin.error());
   };
   auto load_errors = std::vector<caf::error>{};
   // First, check if the plugin file is specified as an absolute path.
-  if (auto plugin = try_load_plugin(file))
+  auto plugin = try_load_plugin(file);
+  if (plugin)
     return std::pair{file, std::move(*plugin)};
-  else if (plugin.error() != caf::no_error)
+  if (plugin.error() != caf::no_error)
```
mavam, line 86: Not a fan of these changes. This is much more concise:

```cpp
if (auto result = compute)))
   f(*result);
else
  g(result.error())
```

dominiklohmann: Your example is alright. It only warns if there is a return statement before `f`, and that cannot be selectively disabled. Let's not argue with the tool here, because you'd need to patch clang-tidy for that.

PR #1422, opened 2021-03-07
Remove vast::rm

JoeLoser:

Problem:
- `rm` function for `vast::path` is implemented in terms of
  `vast::directory` which is going away soon.

Solution:
- Remove `rm` member function from `vast::path` entirely. Replace call
  sites with `std::filesystem::remove_all` to preserve the previous
  behavior of deleting all files if the file is a directory.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/io/save.cpp
+++ b/libvast/src/io/save.cpp
@@ -15,21 +15,33 @@
 
 #include "vast/error.hpp"
 #include "vast/io/write.hpp"
+#include "vast/logger.hpp"
 #include "vast/path.hpp"
 
 #include <cstddef>
 #include <cstdio>
+#include <filesystem>
+#include <system_error>
 
 namespace vast::io {
 
 caf::error save(const path& filename, span<const std::byte> xs) {
   auto tmp = filename + ".tmp";
   if (auto err = write(tmp, xs)) {
-    rm(tmp);
+    std::error_code ec{};
+    if (const auto removed
+        = std::filesystem::remove(std::filesystem::path{tmp.str()}, ec);
+        !removed)
+      VAST_WARN("failed to remove file {} : {}", tmp, ec.message());
     return err;
   }
   if (std::rename(tmp.str().c_str(), filename.str().c_str()) != 0) {
-    rm(tmp);
+    std::error_code err{};
+    if (const auto removed_files
+        = std::filesystem::remove_all(std::filesystem::path{tmp.str()}, err);
+        removed_files == 0)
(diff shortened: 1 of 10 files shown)


libvast/src/io/save.cpp
```diff
@@ -19,17 +19,21 @@
...
+    std::error_code ec{};
+    std::filesystem::remove(std::filesystem::path{tmp.str()}, ec);
     return err;
   }
   if (std::rename(tmp.str().c_str(), filename.str().c_str()) != 0) {
-    rm(tmp);
+    std::error_code err{};
+    std::filesystem::remove_all(std::filesystem::path{tmp.str()}, err);
     return caf::make_error(ec::filesystem_error,
                            "failed in rename(2):", std::strerror(errno));
```
mavam, line 38: Maybe we should just use `ec` instead of `errno`?

libvast/src/system/spawn_node.cpp
```diff
@@ -80,9 +82,15 @@ spawn_node(caf::scoped_actor& self, const caf::settings& opts) {
...
-  actor->attach_functor([=, pid_file = std::move(pid_file)](const caf::error&) {
+  actor->attach_functor([=, pid_file = std::move(pid_file)](
+                          const caf::error&) -> caf::result<void> {
     VAST_DEBUG("node removes PID lock: {}", pid_file.str());
-    rm(pid_file);
+    std::error_code err{};
+    std::filesystem::remove_all(std::filesystem::path{pid_file.str()}, err);
+    if (err)
+      return caf::make_error(ec::filesystem_error,
+                             "unable to remove pid file:", pid_file);
```
mavam, line 92: Let's include `err` in the message so that the user has something to work with.

libvast/src/io/save.cpp
```diff
@@ -19,17 +19,21 @@
...
+#include <system_error>
 
 namespace vast::io {
 
 caf::error save(const path& filename, span<const std::byte> xs) {
   auto tmp = filename + ".tmp";
   if (auto err = write(tmp, xs)) {
-    rm(tmp);
+    std::error_code ec{};
+    std::filesystem::remove(std::filesystem::path{tmp.str()}, ec);
```
mavam, line 31: Since we're not returning `ec` here, let's issue a warning if `remove` fails.

mavam (changes requested):
Thanks for cranking these refactorings out. This is another good one.

I just have small comments on some touch ups.

mavam (approved):

PR #1423, opened 2021-03-07
Remove dependency on vast::directory in segment_store

JoeLoser:

Problem:
- When registering segments in the `segment_store`, `vast::directory` is
  used which is going away soon.

Solution:
- Use `std::filesystem::directory_iterator` instead to remove the need
  for `vast::directory`. As implemented, this creates some extra
  conversions for converting back and forth between path types.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/segment_store.cpp
+++ b/libvast/src/segment_store.cpp
@@ -19,7 +19,6 @@
 #include "vast/concept/printable/vast/filesystem.hpp"
 #include "vast/concept/printable/vast/uuid.hpp"
 #include "vast/detail/overload.hpp"
-#include "vast/directory.hpp"
 #include "vast/error.hpp"
 #include "vast/fbs/segment.hpp"
 #include "vast/fbs/utils.hpp"
@@ -379,9 +378,19 @@ void segment_store::inspect_status(caf::settings& xs,
 }
 
 caf::error segment_store::register_segments() {
-  for (auto filename : directory{segment_path()})
-    if (auto err = register_segment(filename))
-      return err;
+  auto p = std::filesystem::path{segment_path().str()};
+  if (!std::filesystem::exists(p))
+    return caf::none;
+  std::error_code err{};
+  std::filesystem::directory_iterator dir{p, err};
+  if (err)
+    return caf::make_error(ec::filesystem_error,
+                           fmt::format("failed to find segment path {} : {}",
+                                       segment_path(), err.message()));
+  for (const auto& entry : dir)
+    if (entry.exists())
+      if (auto err = register_segment(vast::path{entry.path().string()}))
+        return err;
   return caf::none;
 }
 


libvast/src/segment_store.cpp
```diff
@@ -373,8 +372,15 @@ void segment_store::inspect_status(caf::settings& xs,
 }
 
 caf::error segment_store::register_segments() {
-  for (auto filename : directory{segment_path()})
-    if (auto err = register_segment(filename))
+  std::error_code err{};
+  std::filesystem::directory_iterator dir{
+    std::filesystem::path{segment_path().str()}, err};
```
mavam, line 377:
```suggestion
  std::filesystem::directory_iterator dir{segment_path(), err};
```

I wondered: why not this? ☝️ 

Then I realized that you may want to do the upstream adjustment in a separate PR, given the PR title. Am I interpreting this correctly?

libvast/src/segment_store.cpp
```diff
@@ -373,8 +372,15 @@ void segment_store::inspect_status(caf::settings& xs,
...
+  if (err)
+    return caf::make_error(ec::filesystem_error,
+                           fmt::format("failed to find segment path {} : {}",
+                                       segment_path(), err.message()));
+  for (const auto& entry : dir)
+    if (auto err = register_segment(vast::path{entry.path().string()}))
```
mavam, line 383:
```suggestion
    if (auto err = register_segment(entry.path())
```

(Same here.)

mavam (approved):

libvast/src/segment_store.cpp
```diff
@@ -373,8 +372,15 @@ void segment_store::inspect_status(caf::settings& xs,
 }
 
 caf::error segment_store::register_segments() {
-  for (auto filename : directory{segment_path()})
-    if (auto err = register_segment(filename))
+  std::error_code err{};
+  std::filesystem::directory_iterator dir{
+    std::filesystem::path{segment_path().str()}, err};
+  if (err)
```
mavam, line 378: I just realized after reviewing in the first round: a non-existent directory is not a failure. In fact, if there is no persistent data yet, it's expected. So we may need something along the lines of `if (err != dir_does_not_exist)`.

libvast/src/segment_store.cpp
JoeLoser: In the longer term, we'd certainly like to change the return type of the `segment_path()` function to return a `std::filesystem::path`. I actually went down that approach first of converting all of `segment_store` to use `std::filesystem` types, but the call-def chain got a bit larger than what I wanted to tackle at that time due to calling into other functions from other files with path types, etc if I recall correctly.  It had the same issues with integration tests failing, but I think it's due to the non-existent directory like you pointed out.  That's why I split up the changeset into just this.

So for now, I've taken the approach of the absolute minimal requirement needed to remove the `vast::directory` dependency from `segment_store`.

After this PR lands, I think we can just remove the remaining includes for `directory.hpp` and the associated `directory` class, and `vast::directory` will be gone.

After `directory` is gone, I'd propose going through and trying to surgically convert uses of `vast::path` with their `std::filesystem::path` equivalents.  The hope would be to do this in a fashion that neither introduces a lot of path conversions (back and forth between `vast::path` and `std::filesystem::path` due to existing APIs) or introduces separate overloads in the APIs for `std::filesystem::path` and `vast::path`.

libvast/src/segment_store.cpp
JoeLoser: See https://github.com/tenzir/vast/pull/1423#discussion_r589505641

libvast/src/segment_store.cpp
JoeLoser: I now check at the beginning of the function if the segment directory exists before proceeding. Alternatively, I could perhaps check the `err` value with an error code I construct using a `generic_category` enum with a value of `2` for directory not existing, but I think this is clearer, though slightly less performant.  Note this is the value that `std::filesystem` maps for POSIX return codes to its error category/values for a directory not existing.

What do you think?

libvast/src/segment_store.cpp
mavam: Thanks for shedding light into your thought process.

It sounds like you already have a solid plan how to proceed with the refactoring. I like your proposal. I'm all for doing small PRs (which also helps in bisecting) and treading carefully.

libvast/src/segment_store.cpp
mavam: > [..]  but I think this is clearer, though slightly less performant

Agreed; this is not on the critical path and in this case we benefit from clarity (and maybe even portability).

libvast/src/segment_store.cpp
```diff
@@ -379,9 +378,19 @@ void segment_store::inspect_status(caf::settings& xs,
...
 
 caf::error segment_store::register_segments() {
-  for (auto filename : directory{segment_path()})
-    if (auto err = register_segment(filename))
-      return err;
+  auto p = std::filesystem::path{segment_path().str()};
+  if (!std::filesystem::exists(p))
+    return caf::none;
+  std::error_code err{};
+  std::filesystem::directory_iterator dir{p, err};
```
mavam, line 385: (Just a side note.)

Letting TOCTTOU issues aside, I wonder if we can leverage some invariants along the lines of if (a) a path exists, and (b) it's a directory, then the `directory_iterator` construction will not fail. This would make iterating over directories a little easier because you could write:

```cpp
if (exists && is_directory)
  for (const auto& entry : std::filesystem::directory_iterator{p})
    if (entry.exists())
       ...
```

mavam (approved):

PR #1424, opened 2021-03-07
Handle clang-tidy warning for vast::data

dominiklohmann:

n/t

###  :memo: Checklist

File-by-file.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -22,4 +22,6 @@ CheckOptions:
     value:           true
   - key:             misc-non-private-member-variables-in-classes.IgnoreClassesWithAllMemberVariablesBeingPublic
     value:           true
+  - key:             readability-else-after-return.WarnOnConditionVariables
+    value:           false
 ...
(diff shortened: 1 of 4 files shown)


.clang-tidy
```diff
@@ -22,4 +22,6 @@ CheckOptions:
     value:           true
   - key:             misc-non-private-member-variables-in-classes.IgnoreClassesWithAllMemberVariablesBeingPublic
     value:           true
+  - key:             readability-else-after-return.WarnOnConditionVariables
```
mavam, line 25: The docs say:

> When true, the check will attempt to refactor a variable defined inside the condition of the if statement that is used in the else branch defining them just before the if statement. This can only be done if the if statement is the last statement in its parents scope. Default value is true.

Can you give an example in terms of code?

mavam (approved):

.clang-tidy
dominiklohmann:
```
{
  if (x)
    return y;
  else
    z;
}
```

must be written as this if the rule is enabled:

```
{
  if (x)
    return y;
  z;
}
```

This is even the case is `z` is dependent on `x`. The rule is just overzealous and doesn't match our coding style.

PR #1425, opened 2021-03-08
Add support for plugins in the nix-based build

tobim:

This doesn't support plugins in the static binary yet, because that needs a few adjustments in the loading mechanism, and they need to be baked in properly.

To enable inclusion of plugins one can simply call
```
nix/static-binary.sh plugin otherplugin
```

### DIFF
--- a/.github/workflows/static-binary.yaml
+++ b/.github/workflows/static-binary.yaml
@@ -3,6 +3,11 @@ on:
   repository_dispatch:
     types: static-binary
   workflow_dispatch:
+    inputs:
+      arguments:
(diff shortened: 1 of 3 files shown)


nix/vast/plugins.nix
```diff
@@ -0,0 +1,7 @@
+{
+  example = {
+    src = ../../examples/plugins/example;
+    buildInputs = [];
+    cmakeFlags = [];
+  };
+}
```
dominiklohmann, line 7: It looks like this needs to be written for every plugin. That's a non-starter—we cannot require plugins to be available in the build tree, because we want to support a static build including closed-source plugins.

dominiklohmann (changes requested):

dominiklohmann (approved):
Love to see this.

PR #1431, opened 2021-03-10
Remove vast::path use in segment_store

JoeLoser:

Problem:
- `segment_store` uses `vast::path` which is going away soon.

Solution:
- Use `std::filesystem::path` throughout in `segment_store`.

Note:
- This adds an overload for `save` on both path types. The next step
  would be to fix other callers of `save` to not use `vast::path` and
  instead use `std::filesystem::path`.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/io/save.cpp
+++ b/libvast/src/io/save.cpp
@@ -21,31 +21,34 @@
 #include <cstddef>
 #include <cstdio>
 #include <filesystem>
+#include <string>
 #include <system_error>
 
 namespace vast::io {
 
-caf::error save(const path& filename, span<const std::byte> xs) {
-  auto tmp = filename + ".tmp";
+caf::error
+save(const std::filesystem::path& filename, span<const std::byte> xs) {
+  auto tmp = filename;
+  tmp += ".tmp";
   if (auto err = write(tmp, xs)) {
     std::error_code ec{};
-    if (const auto removed
-        = std::filesystem::remove(std::filesystem::path{tmp.str()}, ec);
-        !removed)
+    if (const auto removed = std::filesystem::remove(tmp, ec); !removed || ec)
       VAST_WARN("failed to remove file {} : {}", tmp, ec.message());
     return err;
   }
-  if (std::rename(tmp.str().c_str(), filename.str().c_str()) != 0) {
-    std::error_code err{};
-    if (const auto removed_files
-        = std::filesystem::remove_all(std::filesystem::path{tmp.str()}, err);
-        removed_files == 0)
-      VAST_WARN("failed to remove any files from directory {} : {}", tmp,
-                err.message());
+  std::error_code err{};
+  std::filesystem::rename(tmp, filename, err);
+  if (err) {
+    std::filesystem::remove(tmp, err);
     return caf::make_error(ec::filesystem_error,
-                           "failed in rename(2):", std::strerror(errno));
+                           fmt::format("failed to rename {} : {}", filename,
+                                       err.message()));
   }
   return caf::none;
+}
+
+caf::error save(const path& filename, span<const std::byte> xs) {
+  return save(std::filesystem::path{filename.str()}, xs);
 }
 
 } // namespace vast::io
--- a/libvast/src/segment_store.cpp
+++ b/libvast/src/segment_store.cpp
@@ -37,8 +37,9 @@
 namespace vast {
 
 // TODO: return expected<segment_store_ptr> for better error propagation.
-segment_store_ptr segment_store::make(path dir, size_t max_segment_size,
-                                      size_t in_memory_segments) {
+segment_store_ptr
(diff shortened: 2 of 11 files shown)


JoeLoser:
@mavam @dominiklohmann looks like the AppleClang CI job is just your usual spurious failure. The other CI jobs passed just fine.

libvast/src/segment_store.cpp
```diff
@@ -351,8 +353,10 @@ caf::error segment_store::flush() {
     return err;
   // Keep new segment in the cache.
   cache_.emplace(seg.id(), seg);
+  [[maybe_unused]] const auto trimmed_filename
+    = filename.string().substr(0, filename.string().size() - 3);
   VAST_DEBUG("{} wrote new segment to {}", detail::pretty_type_name(this),
-             filename.trim(-3));
```
mavam, line 355: The function `path::trim` operated not on characters but instead on path components. Check the docs here: https://github.com/tenzir/vast/blob/master/libvast/vast/path.hpp#L105. Here, `-3` means calling `filesystem::path::parent_path` three times.

The logic should be easy to factor into a free function in namespace detail. However, I don't recall that we use this function aside from prettifying logging output. And in this particular instance, I wonder whether simply calling `filename.parent_path()` is what we want.

libvast/src/io/save.cpp
```diff
@@ -21,31 +21,33 @@
...
 #include <system_error>
 
 namespace vast::io {
 
-caf::error save(const path& filename, span<const std::byte> xs) {
-  auto tmp = filename + ".tmp";
-  if (auto err = write(tmp, xs)) {
+caf::error
+save(const std::filesystem::path& filename, span<const std::byte> xs) {
+  auto tmp = std::filesystem::path{filename.string() + ".tmp"};
```
mavam, line 31:
```suggestion
  auto tmp = filename + ".tmp";
```

See https://en.cppreference.com/w/cpp/filesystem/path/concat.

(Not sure if I missed similar instances of this in the past.)

libvast/src/io/save.cpp
```diff
@@ -21,31 +21,33 @@
...
+  if (auto err = write(vast::path{tmp.string()}, xs)) {
```
mavam, line 32:
```suggestion
  if (auto err = write(tmp, xs)) {
```

Looks like there is an overload for this now.

mavam (changes requested):
Once again nice work.

I just noticed a few kinks this time.

libvast/src/io/save.cpp
JoeLoser: We can't use `concat` or `operator+=` as it mutates the LHS operand. In this case, we need two `std::filesystem::path` variables: the original `filename` and then a new `tmp` file. We can't modify `filename` as we need it for the rename later. Do you see something I'm missing?

libvast/src/io/save.cpp
JoeLoser: Good catch, I just fixed that up.

libvast/src/segment_store.cpp
JoeLoser: Oops, thanks for pointing that out -- definitely an unintended change in behavior.

I just switched to using `filename.parent_path()` here as it's arguably better. If you or @dominiklohmann feel strongly, we can just `filename.parent_path().parent_path().parent_path()` to mimic the old behavior.

libvast/src/io/save.cpp
mavam: Interesting. Now that I looked closely, there is no `operator+` while there exists an `operator/`. Apparently the committee deems it to risky. 🤷 

So yeah, there are basically two ways to concatenate raw paths. Either do it the way you suggested or potentially save an allocation by doing this:

```cpp
auto tmp = filename;
filename += ".tmp";
```

:point_up: This has less characters overall, so while it expands more vertically, it's quicker to understand. However, I don't feel strongly about this change. After all, we're doing I/O immediately afterwards.

libvast/src/segment_store.cpp
mavam: This works in my opinion.

It's also a debug log statement. We don't have a policy of, say, printing all paths relative to the `vast.db` dir in log statements. It's something to consider for the future and only on log levels of INFO and more severe.

mavam (approved):
Technically ready to merge now. I'm just waiting for your opinion on how to do the path concatenation, @JoeLoser.

libvast/src/io/save.cpp
JoeLoser: I just went ahead with your proposed change; a bit shorter and easier to reason about I'd say. 👍

JoeLoser:
> Technically ready to merge now. I'm just waiting for your opinion on how to do the path concatenation, @JoeLoser.

I took your suggestion on doing the path concatenation.  Should be ready to merge now.

PR #1433, opened 2021-03-11
Prune expressions for the meta index lookup

tobim:

We can take advantage of the fact that all string fields are covered by the same type level synopsis. To be more specific: A query like `suricata.smb.host == "foo" || suricata.ssh.host == "foo"` would check whether the string "foo" is included the same type-level string synopsis twice.

This PR adds a preprocessing step that removes predicates with duplicate strings from the expression as a preprocessing step for meta index lookups.

In my local comparisons with 12723 partitions with `eve.log` data I get the a 4 times speedup for the query `vast export null 'net.domain == "alhgeoafh" || net.hostname == "alhgeoafh"'`.
A release build on master finishes the command in ~ 930 ms, this branch is done after ~ 240 ms.

###  :memo: Checklist

Try to reproduce my results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- ⚠️ The time to first response of queries that compare a concept to a string
+  has been reduced noticably. In the particular case of large databases
+  containing many different event types and queries with a high selectivity we
+  were able to measure speedups of up to 5x.
+  [#1433](https://github.com/tenzir/vast/pull/1433)
+
 - 🐞 The JSON parser now accepts data with numerical or boolean values in
   fields that expect strings according the schema.
   [#1439](https://github.com/tenzir/vast/pull/1439)
--- a/libvast/src/system/meta_index.cpp
+++ b/libvast/src/system/meta_index.cpp
@@ -80,12 +80,14 @@ struct pruner {
     for (const auto& operand : connective) {
       const std::string* str = nullptr;
       if (const auto* pred = caf::get_if<predicate>(&operand)) {
-        if (const auto* d = caf::get_if<data>(&pred->rhs)) {
-          if ((str = caf::get_if<std::string>(d))) {
-            if (memo.find(*str) != memo.end())
-              continue;
-            memo.insert(*str);
-            result.emplace_back(*pred);
+        if (!caf::holds_alternative<meta_extractor>(pred->lhs)) {
+          if (const auto* d = caf::get_if<data>(&pred->rhs)) {
+            if ((str = caf::get_if<std::string>(d))) {
+              if (memo.find(*str) != memo.end())
+                continue;
+              memo.insert(*str);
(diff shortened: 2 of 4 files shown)


dominiklohmann:
I'll review this one in practice, I still have a 1TB database with Zeek streaming JSON on my hard drive.

dominiklohmann (commented):
I verified the performance bump: I've seen a reduction from 2s to 0.4s on a similar query. There seems to be no visible performance penalty for regular queries.

I can reproduce the ASan failures in `vast-test` locally. Please fix and add a changelog entry, other than that LGTM.

tobim:
I'm not sure whether I should add a changelog entry for this.

mavam:
@tobim if the performance gains are substantial, it's a nice change item. As a user, I always enjoy reading gains if they are specific and to the point (while broad claims are rather a turnoff).

CHANGELOG.md
```diff
@@ -9,10 +9,17 @@ This changelog documents all notable user-facing changes of VAST.
 ### 🧬 Experimental Features
 ### 🎁 Features
 ### 🐞 Bug Fixes
+### 🚀 Performance Improvements
```
dominiklohmann, line 12: Can we just make this a `Feature`? Otherwise this will get really crowded in the release notes.

libvast/src/system/meta_index.cpp
```diff
@@ -52,9 +53,69 @@ partition_synopsis& meta_index_state::at(const uuid& partition) {
   return synopses.at(partition);
 }
 
+struct pruner {
```
dominiklohmann, line 56: Needs comments.

libvast/src/system/meta_index.cpp
```diff
@@ -52,9 +53,69 @@ partition_synopsis& meta_index_state::at(const uuid& partition) {
...
+        }
+      }
+      if (!str)
+        result.push_back(caf::visit(*this, operand));
+    }
+    return result;
+  }
+};
+
+expression prune_all(expression e) {
```
dominiklohmann, line 96: Needs a comment.

CHANGELOG.md
mavam: I would make it a change, not a feature. Performance is typically considered a non-functional requirement.

tobim: I think performance improvement belongs in there. I'm wondering why the options here aren't aligned with the labels in Github.

mavam: "in there" is ambiguous. Mind clarifying?

dominiklohmann (approved):
Fine by me, but please make the changelog entry either a Feature of a Change instead of introducing a new category.

CHANGELOG.md
tobim: I mean that we should have a "performance" label in the changelog.

dominiklohmann: I disagree. If we want to make this change, let's discuss it separately and make it part of the release prep PR.

tobim: Alrighty.

PR #1434, opened 2021-03-11
Handle race condition during passive partition shutdown

lava:
###  :memo: Checklist

### DIFF
--- a/libvast/src/system/partition.cpp
+++ b/libvast/src/system/partition.cpp
@@ -805,6 +805,12 @@ partition_actor::behavior_type passive_partition(
       // We can safely assert that if we have the partition chunk already, all
       // deferred evaluations were taken care of.
       VAST_ASSERT(self->state.deferred_evaluations.empty());
+      // Don't handle queries after we already received an exit message, while
+      // the terminator is running. Since we require every partition to have at
+      // least one indexer, we can use this to check.
+      if (self->state.indexers.empty())
+        return caf::make_error(ec::system_error, "can not handle query because "
+                                                 "shutdown was requested");
       auto triples = evaluate(self->state, expr);
       if (triples.empty())
         return atom::done_v;


libvast/src/system/partition.cpp
```diff
@@ -805,6 +805,12 @@ partition_actor::behavior_type passive_partition(
       // We can safely assert that if we have the partition chunk already, all
       // deferred evaluations were taken care of.
       VAST_ASSERT(self->state.deferred_evaluations.empty());
+      // Don't handle queries after we already received an exit message, while
+      // the terminator is running. Since we require every partition to have at
+      // least one indexer, we can use this to check.
```
mavam, line 810: I'm not quite sure if I can understand the comment. Are all `indexers` cleared when entering shutdown? (I did not see a call to `clear` in the EXIT handler.)

lava: Yes, correct, they're cleared in the exit handler via `std::exchange(self->state.indexers, {})`.

mavam: Ah, sure, because everybody who reads this knows that this means "clearing a vector". 🤓

mavam (approved):
I did not test this because this is not trivial to mock. That said, the solution makes sense analytically.

libvast/src/system/partition.cpp
lava: Yeah, I think with the benefit of 10 years of hindsight, I would say that `std::move()` should have just been `static_cast<T&&>`, and that the function `std::move()` should have the semantics of `std::exchange()`. But it's too late to change now, so we might as well get used to it :D

dominiklohmann: To be fair, the proposal for `std::take` that essentially is `std::exchange(x, {})` is apparently being rejected because the name of the function is too short. Maybe we'll get it as `co_take` sometime in the future. 🤷‍♂️

PR #1439, opened 2021-03-11
Accept numbers in place of strings in JSON

tobim:

If the schema says string but the data is numerical, we call `to_string` instead of failing.

###  :memo: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- 🐞 The JSON parser now accepts data with numerical or boolean values in
+  fields that expect strings according the schema.
+  [#1439](https://github.com/tenzir/vast/pull/1439)
+
 - 🐞 Data that was ingested before the deprecation of the `#timestamp`
   attribute wasn't exported correctly with newer versions. This is now
   corrected.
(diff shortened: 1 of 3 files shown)


libvast/test/format/json.cpp
```diff
@@ -6,7 +6,7 @@
  *                                                                            *
  * This file is part of VAST. It is subject to the license terms in the       *
  * LICENSE file found in the top-level directory of this distribution and at  *
- * http://vast.io/license. No part of VAST, including this file, may be       *
```
dominiklohmann, line 9: Please revert this change

dominiklohmann (approved):
LGTM, but please address my one comment before merging.

libvast/test/format/json.cpp
tobim: That was unintentional.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- 🐞 The JSON parser now accepts data with numerical or boolean values in
```
mavam, line 16: I would make this a change instead of bug. Our parser behaved correctly beforehand, but now it operates in a "enhanced" mode and accepts inputs more liberally.

dominiklohmann: I disagree. From a users perspective, this is a bug. We do all kinds of conversions here and this one was missing.

mavam: Okay, I have not a particularly strong opinion here.

PR #1444, opened 2021-03-12
Add lsvast integration test suite

lava:

NOTE: This PR is still a bit hacky. I think we should take time for a review/discussion session to discuss how much we want to cleanly integrate this with CMake infrastructure vs. how much we just want something "good enough" to ensure lsvast stops randomly breaking with every release.

###  :memo: Checklist


dominiklohmann (dismissed):
Hacky, but better than no test at all. We discussed possible ways to improve the situation in the future.

vast/CMakeLists.txt
```diff
@@ -65,6 +65,7 @@ file(
       --set \"$base_dir\"/vast_integration_suite.yaml \
       \"$@\"
      $<$<BOOL:${VAST_ENABLE_LSVAST}>:
+     mkdir -p lsvast-integration-test/vast.db
```
dominiklohmann, line 68: Why would this be necessary?

lava: VAST doesn't start if the directory doesn't exist:
```
[benno@L build]$ ./bin/vast -d foo/bar start
unable to create directory "foo/bar": No such file or directory
```

dominiklohmann: So then it should probably just be `mkdir lsvast-integration-test`, because VAST will and should create the DB directory.

vast/CMakeLists.txt
```diff
@@ -58,12 +58,22 @@ file(
      . \"$env_dir/bin/activate\"
      python -m pip install --upgrade pip
      python -m pip install -r \"$base_dir/requirements.txt\"
-     $<$<BOOL:<${VAST_ENABLE_ARROW}>:python -m pip install pyarrow>
+     $<$<BOOL:${VAST_ENABLE_ARROW}>:python -m pip install pyarrow>
      python \"$base_dir/integration.py\" \
       --app \"$app\" \
       --directory vast-integration-test \
+      --set \"$base_dir\"/vast_integration_suite.yaml \
```
dominiklohmann, line 65: Can you remove this line and revert the renaming _or_ adjust the default value set in `integration.py`?

dominiklohmann (commented):
The old `default_set.yaml` is still mentioned in `vast/integration/README.md` and the argument is still passed explicitly (which it should not be), but other than that LGTM.

lava:
I've updated the README. The explicit argument is intentional, since it would be weird for a runner script to rely on the implicit defaults of the integration testing framework.

dominiklohmann (approved):

dominiklohmann:
The failure is unrelated. Merging manually.

PR #1446, opened 2021-03-12
Establish subtyping relationships for type extractors

tobim:

This works already, but the checklist demands a bit more time investment.

###  :memo: Checklist

By commit.


libvast/src/system/meta_index.cpp
```diff
@@ -297,18 +297,25 @@ std::vector<uuid> meta_index_state::lookup_impl(const expression& expr) const {
...
         },
-        [&](const type_extractor& lhs, const data&) -> result_type {
+        [&](const type_extractor& lhs, const data& d) -> result_type {
           auto result = [&] {
             if (caf::holds_alternative<none_type>(lhs.type)) {
               VAST_ASSERT(!lhs.type.name().empty());
               auto pred = [&](auto& field) {
-                return field.type.name() == lhs.type.name();
+                const auto* p = &field.type;
+                while (const auto* a = caf::get_if<alias_type>(p)) {
```
mavam, line 306: Exactly how I envisioned it. I wonder whether we should add our usual protection against overflow by recursion.

My suggestion: write a story for that and add a todo where you'd like to perform the recursion check.

libvast/src/expression_visitors.cpp
```diff
@@ -359,7 +359,15 @@ caf::expected<expression>
 type_resolver::operator()(const type_extractor& ex, const data& d) {
   if (caf::holds_alternative<none_type>(ex.type)) {
     auto matches = [&](const type& t) {
-      return t.name() == ex.type.name() && compatible(t, op_, d);
+      const auto* p = &t;
+      while (const auto* a = caf::get_if<alias_type>(p)) {
```
mavam, line 363: Same comment about recursion (see below) applies here.

mavam (approved):
Nice work. I looked a the code, unit test, and integration test.

I didn't give it a spin yet, though. It'd be great if someone else from @tenzir/backend could play with it for a few mins and try to break it.

libvast/src/system/meta_index.cpp
tobim: There is no recursive call in here.

Regarding the issue at large: I don't think adding recursion limits for all functions that operate on tree types like `type`, `data` or `expression` is a good approach. It would be preferable to me if we could keep it to the functions that construct or alter these types.

dominiklohmann (approved):
Code and tests are fine by me. This seems to work nicely in practice—I've toyed around with this with a few simple tests and the output matched my expectations.

PR #1447, opened 2021-03-15
Don't allow field extractors to match field name suffixes

lava:
###  :memo: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- 🐞 A query for a field or field name suffix that matches multiple fields of
+  different types would erroneously return no results.
+  [#1447](https://github.com/tenzir/vast/pull/1447)
+
 - ⚠️ The default size of table slices (event batches) that is created from
   `vast import` processes has been changed from 1000 to 1024.
   [#1396](https://github.com/tenzir/vast/pull/1396)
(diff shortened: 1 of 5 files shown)


lava:
After thinking about this PR a bit, I think this might break queries like `zeek.* == "xxx"`. Since there's no unit test for these and they would have only very rarely worked previously, I'm not sure if this was supposed to work. But as a user, it seems like something I'd intuitively expect to work, if wildcards are supported at all.

tobim:
> After thinking about this PR a bit, I think this might break queries like `zeek.* == "xxx"`. Since there's no unit test for these and they would have only very rarely worked previously, I'm not sure if this was supposed to work. But as a user, it seems like something I'd intuitively expect to work, if wildcards are supported at all.

The bug that this PR fixes prevented those queries to work as well, so I don't think there is any harm done.

mavam:
> But as a user, it seems like something I'd intuitively expect to work, if wildcards are supported at all.

Yeah, I agree. let's treat this as a separate feature that we don't support yet.

If we stick with component-based suffix matching, we're good for this PR.

tobim (changes requested):
This is still missing a changelog entry, otherwise it is good to go.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- 🐞 Partitions containing multiple fields with the same field name suffix could
+  fail to return some query results due to a bug in the type resolver.
+  [#1447](https://github.com/tenzir/vast/pull/1447)
```
tobim, line 18: This is very technical, the changelog should be more geared towards the user.
How about:
```
A query for a field or field name suffix that matches multiple fields of different
types would erroneously return no results.
```

lava: I think we agreed not to do changelog wordsmithing in the individual PRs in favor of one big wording pass when we rearrange the changlog for the release.

dominiklohmann: To be fair, that's not really wordsmithing but rather changing the content entirely.

tobim (approved):

PR #1449, opened 2021-03-15
Fix usage of CCache inside Docker in CI

dominiklohmann:
~This is more of an experiment than me actually knowing what I'm doing here. 🤷‍♂️~ I understand now. We're using CCache again _and_ are uploading the cache again. 

Review Instructions:

Take a look at all the changes, and verify that CI took less than 50 minutes to run on the last run.

image


tobim (approved):
Thanks for figuring this out!

PR #1450, opened 2021-03-15
Remove erased partitions from the meta index

lava:
###  :memo: Checklist

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -869,6 +869,8 @@ index(index_actor::stateful_pointer<index_state> self,
       }
       self->state.inmem_partitions.drop(partition_id);
       self->state.persisted_partitions.erase(partition_id);
+      self->request(self->state.meta_index, caf::infinite, atom::erase_v,
+                    partition_id);
       self->request(self->state.filesystem, caf::infinite, atom::mmap_v, path)
         .then(
           [=](const chunk_ptr& chunk) mutable {
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -869,8 +869,15 @@ index(index_actor::stateful_pointer<index_state> self,
       }
       self->state.inmem_partitions.drop(partition_id);
       self->state.persisted_partitions.erase(partition_id);
-      self->request(self->state.meta_index, caf::infinite, atom::erase_v,
-                    partition_id);
+      self
+        ->request(self->state.meta_index, caf::infinite, atom::erase_v,
+                  partition_id)
+        .then([](atom::ok) { /* nop */ },
+              [partition_id](const caf::error& err) {
+                VAST_WARN("index encountered an error trying to erase "
+                          "partition {} from the meta index: {}",
+                          partition_id, err);
+              });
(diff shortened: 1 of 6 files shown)


libvast/src/system/index.cpp
```diff
@@ -869,6 +869,8 @@ index(index_actor::stateful_pointer<index_state> self,
       }
       self->state.inmem_partitions.drop(partition_id);
       self->state.persisted_partitions.erase(partition_id);
+      self->request(self->state.meta_index, caf::infinite, atom::erase_v,
+                    partition_id);
```
tobim, line 873: Is the effect of `request` without `then` different from `send`?

dominiklohmann: Only slightly. It allows for adding a `.then` handler that matches either `()` or `(caf::error)`. Internally the only difference is the additional hidden request ID parameter. We should always use `request` over `send` where possible, even for return types of `caf::result<void>`.

tobim: I don't follow your argument here. Why should we prefer `request` over `send` if the latter suffices?

dominiklohmann: Because with `send` we silently ignore possibly returned errors from a `caf::result<void>`.

tobim: So we should use `request().then(...)`. That brings us to the question of what to do in case of an `error`. Sure, we can log it, but the error would come from CAF and we're already in a broken state if we can't communicate with the meta index from here.

dominiklohmann: Exactly. We can log the correct error and propagate it outwards. The necessary measure might be a shutdown.

dominiklohmann (commented):
The change itself is really simple.

Open tasks:
- Add changelog entry
- Handle request reply and log error

libvast/src/system/index.cpp
```diff
@@ -869,8 +869,15 @@ index(index_actor::stateful_pointer<index_state> self,
...
       self->state.persisted_partitions.erase(partition_id);
-      self->request(self->state.meta_index, caf::infinite, atom::erase_v,
-                    partition_id);
+      self
+        ->request(self->state.meta_index, caf::infinite, atom::erase_v,
+                  partition_id)
+        .then([](atom::ok) { /* nop */ },
+              [partition_id](const caf::error& err) {
+                VAST_WARN("index encountered an error trying to erase "
+                          "partition {} from the meta index: '{}",
```
dominiklohmann, line 878: There's a `'` before the error that shouldn't be there. Also, should index not be inferred from the `self` handle?

dominiklohmann (approved):
Fine by me, but please address my one comment.

libvast/src/system/index.cpp
lava: Technically yes, but I didn't want to open a potential can of worms where the meta index is blocked for whatever reason and a stray resume-handler is keeping a reference to the index actor. And since we wanted to move away from `self` as first argument anyways I thought this might be a good place to start.

dominiklohmann: Alrighty.

PR #1451, opened 2021-03-16
Report metrics while idle

dominiklohmann:

Previously, VAST skipped metrics reports for the archive, importer, source, and sink components when they handled 0 events since the last metric report. This made it impossible to distinguish on metrics dashboards whether a component was offline or idle. Instead, VAST now properly reports 0 events while idle.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/system/accountant.cpp
+++ b/libvast/src/system/accountant.cpp
@@ -314,12 +314,16 @@ accountant(accountant_actor::stateful_pointer<accountant_state> self,
       for (const auto& [key, value] : r) {
         self->state->record(key + ".events", value.events, ts);
         self->state->record(key + ".duration", value.duration, ts);
-        auto rate = value.rate_per_sec();
-        if (std::isfinite(rate))
-          self->state->record(key + ".rate", rate, ts);
-        else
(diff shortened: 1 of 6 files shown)


libvast/src/system/accountant.cpp
```diff
@@ -314,12 +314,16 @@ accountant(accountant_actor::stateful_pointer<accountant_state> self,
       for (const auto& [key, value] : r) {
         self->state->record(key + ".events", value.events, ts);
         self->state->record(key + ".duration", value.duration, ts);
-        auto rate = value.rate_per_sec();
-        if (std::isfinite(rate))
-          self->state->record(key + ".rate", rate, ts);
-        else
-          self->state->record(key + ".rate",
-                              std::numeric_limits<decltype(rate)>::max(), ts);
+        if (value.events == 0) {
```
lava, line 317: As discussed on slack, this special case shouldnt be necessary but empirically we observed that we sometimes receive records with 0 events and 0 duration.

(just leaving this comment in case anyone wonders and looks at the PR in the future)

lava (approved):
Looks good to me.

PR #1453, opened 2021-03-16
Add '--disk-budget-check-binary' option to disk monitor

lava:

NOTE: This is still missing  some tests (after this change, we can finally write an integration test for the disk monitor!), but feel free to already review the code.

###  :memo: Checklist


libvast/src/system/disk_monitor.cpp
```diff
@@ -34,21 +36,48 @@ std::shared_ptr<caf::detail::scope_guard<Fun>> make_shared_guard(Fun f) {
 
 } // namespace
 
+caf::expected<void> validate(const disk_monitor_config& config) {
```
mavam, line 39:
```suggestion
caf:error validate(const disk_monitor_config& config) {
```

This might be more natural. In fact, `expected<void>` is a specialization that just wraps an `error`.

libvast/src/detail/process.cpp
```diff
@@ -178,4 +178,26 @@ caf::settings get_status() {
...
+  std::string result;
+  std::array<char, 256> buffer;
+  auto* out = ::popen(command.c_str(), "re");
+  if (!out)
+    return caf::make_error(ec::system_error,
+                           "popen() failed: "s + ::strerror(errno));
+  size_t nread = 0;
+  do {
+    nread = ::fread(buffer.data(), 1, buffer.size(), out);
+    result += std::string_view{buffer.data(), nread};
```
dominiklohmann, line 192: This can throw with `std::bad_alloc`, which would lead to `pclose` not being called. Now obviously we do not handle allocation failures anywhere in the code base. Not sure what the proper course of action is here. Use some form of RAII wrapper, e.g., `std::unique_ptr` with a custom deleter?

libvast/src/system/spawn_disk_monitor.cpp
```diff
@@ -27,24 +27,36 @@ spawn_disk_monitor(node_actor::stateful_pointer<node_state> self,
   if (!archive)
     return caf::make_error(ec::missing_component, "archive");
   auto opts = args.inv.options;
+  std::optional<std::string> command;
+  if (auto cmd = caf::get_if<std::string>(&opts, "vast.start.disk-budget-check-"
+                                                 "binary"))
```
dominiklohmann, line 32:
```suggestion
  if (auto cmd = caf::get_if<std::string>(&opts, //
        "vast.start.disk-budget-check-binary"))
```

If you insert `//` at the end you force a line break so we can grep for the option.

libvast/src/detail/process.cpp
```diff
@@ -178,4 +178,26 @@ caf::settings get_status() {
 #endif
 }
 
+caf::expected<std::string> execute_blocking(const std::string& command) {
+  using namespace std::string_literals;
+  std::string result;
+  std::array<char, 256> buffer;
```
dominiklohmann, line 184: Please add a comment on why this magic number is 256.

vast.yaml.example
```diff
@@ -135,6 +135,13 @@ vast:
     disk-budget-low: 0GiB
     # Seconds between successive disk space checks.
     disk-budget-check-interval: 90
+    # Binary to use for checking the size of the database directory. If left
+    # unset, VAST will recursively add up the size of all files in the
+    # database directory to compute the size. Mainly useful for e.g. compressed
+    # filesystem where raw file size is not the correct metric.
+    # Must be the absolute path to an executable file, which will get passed
+    # the database directory as its first and only argument.
+    # disk-budget-check-binary:
```
dominiklohmann, line 144:
```suggestion
    #disk-budget-check-binary:
```

libvast/vast/system/disk_monitor.hpp
```diff
@@ -42,12 +48,20 @@ struct disk_monitor_state {
...
+/// Contains values for the fields of `disk_monitor_state` with
+/// the same name.
+struct disk_monitor_config {
+  size_t high_water_mark;
+  size_t low_water_mark;
+  std::optional<std::string> scan_binary;
+  std::chrono::seconds scan_interval;
+};
+
+caf::expected<void> validate(const disk_monitor_config&);
```
dominiklohmann, line 63: Please add documentation for this function. Also, this should probably return `caf::error` instead.

libvast/src/system/spawn_disk_monitor.cpp
```diff
@@ -27,24 +27,36 @@ spawn_disk_monitor(node_actor::stateful_pointer<node_state> self,
...
   auto interval = caf::get_or(opts, "vast.start.disk-budget-check-interval",
                               default_seconds);
+  struct disk_monitor_config config
+    = {*hiwater, *lowater, command, std::chrono::seconds{interval}};
+  if (auto valid = validate(config); !valid)
+    return valid.error();
+  if (!*hiwater) {
+    if (command)
+      VAST_WARN("disk check binary specified but no high-water mark, disk "
+                "monitor will not be spawned");
```
dominiklohmann, line 54: Please rewrite this like our other log messages, e.g.,

> 'vast.start.disk-budget-check-binary' is configured, but 'vast.start.disk-budget-high' is unset; disk-monitor will not be spawned

libvast/src/system/spawn_disk_monitor.cpp
```diff
@@ -27,24 +27,36 @@ spawn_disk_monitor(node_actor::stateful_pointer<node_state> self,
...
+    else
+      VAST_VERBOSE("{} not spawning disk_monitor because no limit configured",
+                   self);
```
dominiklohmann, line 57: Please rewrite this like our other log messages, e.g.,

> 'vast.start.disk-budget-high' is unset; disk-monitor will not be spawned

dominiklohmann (changes requested):
This needs a changelog entry and an integration test. The code itself seems to be working fine, I've just had some smaller nits. I tested this with a simple bash script that runs `du` internally.

Open questions:
- The option is not currently exposed on the command line in `application.cpp`. Should we expose it?
- Should we write a section on this feature on docs.tenzir.com/vast that explains why we have this and how to use it?

libvast/src/detail/process.cpp
lava: Not sure we really need to handle this, if `bad_alloc` terminates the process the open file descriptor will also be closed. And the process is expected to terminate by itself, it's not like we still need to send a signal or similar to shut it down.

libvast/src/system/spawn_disk_monitor.cpp
lava: Maybe you have a different clang-format version than me?

```
  if (auto cmd = caf::get_if<std::string>(&opts, //
                                          "vast.start.disk-budget-check-"
                                          "binary"))
```

libvast/src/detail/process.cpp
dominiklohmann: Reasoning sounds alright to me.

libvast/src/system/spawn_disk_monitor.cpp
dominiklohmann: I was using the default config, my bad. You can still fix this by breaking manually before `&opts`.

tobim (approved):

PR #1455, opened 2021-03-17
Avoid using declarative in anonymous namespace in command.cpp

JoeLoser:

Problem:
- Having a using directive in an anonymous namespace doesn't buy us much
  here and can only cause confusion later perhaps when reasoning about
  type or function resolution.

Solution:
- Qualify the two uses in this file with `std::accumulate` to be
  explicit.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/command.cpp
+++ b/libvast/src/command.cpp
@@ -40,18 +40,17 @@ namespace vast {
 
 namespace {
 
-using std::accumulate;
-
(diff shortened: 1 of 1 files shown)


mavam (approved):
Agreed. Much hunch is that it was used to make `clang-format`'s output less vertical.

PR #1457, opened 2021-03-17
Change io::read to use std::filesystem::path

JoeLoser:

Problem:
- `io::read` depends on `vast::path` which is going away soon.

Solution:
- Have `io::read` work with `std::filesystem::path` instead.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/chunk.cpp
+++ b/libvast/src/chunk.cpp
@@ -18,6 +18,7 @@
 #include "vast/error.hpp"
 #include "vast/io/read.hpp"
 #include "vast/io/save.hpp"
+#include "vast/logger.hpp"
 #include "vast/path.hpp"
 
 #include <caf/deserializer.hpp>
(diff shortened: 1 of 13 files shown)


libvast/src/io/read.cpp
```diff
@@ -15,30 +15,36 @@
...
-caf::expected<std::vector<std::byte>> read(const path& filename) {
-  auto size = file_size(filename);
-  if (!size)
-    return size.error();
-  std::vector<std::byte> buffer(*size);
+caf::expected<std::vector<std::byte>>
+read(const std::filesystem::path& filename) {
+  std::error_code err{};
+  const auto size = std::filesystem::file_size(filename, err);
+  if (size == static_cast<std::uintmax_t>(-1))
```
mavam, line 42:
```suggestion
  if (err || size == static_cast<std::uintmax_t>(-1))
```

libvast/src/chunk.cpp
```diff
@@ -141,14 +142,18 @@ caf::error write(const std::filesystem::path& filename, const chunk_ptr& x) {
   return io::save(filename, as_bytes(x));
 }
 
-caf::error read(const path& filename, chunk_ptr& x) {
-  auto size = file_size(filename);
-  if (!size) {
+caf::error read(const std::filesystem::path& filename, chunk_ptr& x) {
+  std::error_code err{};
+  const auto size = std::filesystem::file_size(filename, err);
+  if (size == static_cast<std::uintmax_t>(-1)) {
```
mavam, line 148:
```suggestion
  if (err || size == static_cast<std::uintmax_t>(-1)) {
```

mavam (commented):
Nice cleanup.

I wonder whether we need to consider `err` in some cases a well.

libvast/src/io/read.cpp
lava: According to the docs the code is fine:

> The non-throwing overload returns static_cast<std::uintmax_t>(-1) on errors.

mavam: Thanks for clarifying.

mavam:
It looks like this API change still needs to be ported to `lsvast`.

JoeLoser:
> It looks like this API change still needs to be ported to `lsvast`.

Good call, I didn't have `VAST_ENABLE_LSVAST` CMake option to `ON` when building/running locally, which is why I missed this. Should be fixed now.

dominiklohmann (approved):
Thanks again!

PR #1458, opened 2021-03-17
Fix possibly unhandled exception in disk monitor

dominiklohmann:

This make `recursive_size` return an error instead of throw a (then unhandled) exception. The underlying root cause will need to be eliminated in a separate PR—it's not that simple to fix.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@ This changelog documents all notable user-facing changes of VAST.
   idle and not running components from the metrics.
   [#1451](https://github.com/tenzir/vast/pull/1451)
 
+- 🐞 VAST no longer crashes when the disk monitor tries to calculate the size of
+  the database while files are being deleted. Instead, it will retry after the
+  configured scan interval.
+  [#1458](https://github.com/tenzir/vast/1458)
+
 - ⚡️ A new `VASTRegisterPlugin` CMake function enables easy setup of the build
   scaffolding required for plugins. Plugins can now be linked statically against
   VAST. Configure with `--with-static-plugins` or build a static binary to link
--- a/libvast/src/system/disk_monitor.cpp
+++ b/libvast/src/system/disk_monitor.cpp
@@ -142,6 +142,9 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
                         erased_ids)
               .then(
                 [=, sg = shared_guard](atom::done) {
+                  // TODO: There's a race condition here: We calculate the size
+                  // of the database directory while we might be deleting files
+                  // from it.
                   if (const auto size
                       = detail::recursive_size(self->state.dbdir);
                       !size) {
--- a/libvast/vast/detail/recursive_size.hpp
+++ b/libvast/vast/detail/recursive_size.hpp
@@ -37,8 +37,11 @@ recursive_size(const std::filesystem::path& root_dir) {
   for (const auto& f : dir) {
     if (f.is_regular_file()) {
       const auto size = f.file_size(err);
-      if (err)
+      if (err) {
+        if (err == std::errc::no_such_file_or_directory)
+          continue;
         return caf::make_error(ec::filesystem_error, err.message());
+      }
       VAST_TRACE("{} += {}", f.path().string(), size);
       total_size += size;
     }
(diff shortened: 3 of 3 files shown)


libvast/src/system/disk_monitor.cpp
```diff
@@ -142,6 +142,10 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
                         erased_ids)
               .then(
                 [=, sg = shared_guard](atom::done) {
+                  // TODO: There's a race condition here: We calculate the size
+                  // of the database directory while we might be deleting files
+                  // from it. Both the size calculation and file delete
+                  // operations must go through the filesystem actor instead.
```
mavam, line 148: Let's keep the solution space more open until we have agreed on how to solve the issue internally.

mavam (approved):
Looks good.

I would reduce the scope of the extra comment, though.

libvast/src/system/disk_monitor.cpp
dominiklohmann: Will reword.

CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- 🐞 VAST no longer crashes when the disk monitor tries to calculate the size of
+  the database while files are being deleted. Instead, it will retry after the
+  configured scan interval.
+  [#1458](https://github.com/tenzir/vast/1458)
```
mavam, line 19: The delta to the recent refactoring is almost so small that this doesn't warrant an entry. No strong opinion though. We can also keep it.

dominiklohmann: Let's keep it in. It's a bug fix after all.

mavam (approved):
No further testing needed here, assuming we exhaustively covered all call sites of `file_size`.

libvast/vast/detail/recursive_size.hpp
```diff
@@ -36,7 +36,9 @@ recursive_size(const std::filesystem::path& root_dir) {
     return caf::make_error(ec::filesystem_error, err.message());
   for (const auto& f : dir) {
     if (f.is_regular_file()) {
-      const auto size = f.file_size();
+      const auto size = f.file_size(err);
+      if (err)
+        return caf::make_error(ec::filesystem_error, err.message());
```
tobim, line 41: This should not lead to an error.
```suggestion
        VAST_WARNING("vast is unable to get the size of {}: {}", f, err.message());
```

tobim (changes requested):

libvast/vast/detail/recursive_size.hpp
dominiklohmann: That may be true in the case of the disk monitor, but is it true in the case of a function that determines the size of a directory?

tobim: ok, lets deal with the toctou correctly:
```
if (err == std::make_error_code(std::errc::no_such_file_or_directory)) {
  VAST_WARN(..);
  continue;
} else {
  return make_error(...);
}
```

tobim: Maybe this shouldn't even emit a warning?

dominiklohmann: That's a nice solution. I agree on not warning. Will fix.

tobim: Good! I'm not sure whether `make_error_code` is the correct way to check for a certain error, just fyi.

lava: Imho `VAST_WARN` is far too noisy for this, the "size" of a directory is an ephemeral, best-effort value anyways (the whole directory can be removed the millisecond after this function returns, files can be resized or deleted, etc.). If we want reliable values, we'd implement a different and much more complicated function using filesystem locking etc.

dominiklohmann: Please check back @tobim

tobim (approved):
Leaving the rest to CI.

libvast/vast/detail/recursive_size.hpp
```diff
@@ -36,7 +36,12 @@ recursive_size(const std::filesystem::path& root_dir) {
     return caf::make_error(ec::filesystem_error, err.message());
   for (const auto& f : dir) {
     if (f.is_regular_file()) {
-      const auto size = f.file_size();
+      const auto size = f.file_size(err);
```
JoeLoser, line 39: Sorry about this bug. My fault 😞

dominiklohmann: No worries, we missed it during review. Your PRs are super appreciated!

PR #1459, opened 2021-03-17
Update version and build tree hash on every build

dominiklohmann:

This changes the build scaffolding to rebuild `config.cpp` when necessary instead of only when reconfiguring.

###  :memo: Checklist

File-by-file. Test locally.


cmake/VASTGitVersion.cmake
```diff
@@ -31,3 +33,36 @@ string(REGEX REPLACE "^[0-9]+\\.[0-9]+\\.[0-9]+\\-([0-9]+).*" "\\1"
...
+                     "${VAST_VERSION_TAG}")
+
+file(
+  GLOB_RECURSE
+  hash_files
+  "${CMAKE_CURRENT_LIST_DIR}/../libvast/*.hpp"
+  "${CMAKE_CURRENT_LIST_DIR}/../libvast/*.hpp.in"
+  "${CMAKE_CURRENT_LIST_DIR}/../libvast/*.cpp"
+  "${CMAKE_CURRENT_LIST_DIR}/../cmake/*"
+  "${CMAKE_CURRENT_LIST_DIR}/../CMakeLists.txt")
```
tobim, line 48: Why has the hash computation been moved to this file?
The way this is written it would break if this file is included from the top level CMakeLists.txt, or any other cmake file that is not in the right relative location.

libvast/CMakeLists.txt
```diff
@@ -311,24 +311,33 @@ if (VAST_ENABLE_RELOCATABLE_INSTALLATIONS
...
+file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/update-config.cmake" "\
+  get_filename_component(CMAKE_MODULE_PATH
+    \"${CMAKE_CURRENT_SOURCE_DIR}/../cmake\" ABSOLUTE)
+  include(VASTGitVersion)
+  configure_file(\"${CMAKE_CURRENT_SOURCE_DIR}/src/config.cpp.in\"
+                 \"${CMAKE_CURRENT_BINARY_DIR}/src/config.cpp\")")
+
+add_custom_target(libvast_update_config
+  BYPRODUCTS "${CMAKE_CURRENT_BINARY_DIR}/src/config.cpp"
+  COMMAND ${CMAKE_COMMAND} -P "${CMAKE_CURRENT_BINARY_DIR}/update-config.cmake")
```
tobim, line 339: This new indirection has to be documented in a comment that explicitly states which problem it solves and why this method is appropriate.

tobim (changes requested):

cmake/VASTGitVersion.cmake
dominiklohmann: It had to be moved to we recompute it alongside the version.

> The way this is written it would break if this file is included from the top level CMakeLists.txt, or any other cmake file that is not in the right relative location.

That is not true, we use `CMAKE_CURRENT_LIST_DIR` over `CMAKE_CURRENT_SOURCE_DIR` here to avoid that. It's relative to the location of the `VASTGitVersion.cmake` file this way.

libvast/CMakeLists.txt
dominiklohmann: Fair point, a command is definitely needed here.

dominiklohmann: Done

cmake/VASTGitVersion.cmake
dominiklohmann: Also, fun fact: this already is being included from both the top-level CMakeLists.txt and from a generated .cmake file in some binary directory.

tobim (approved):
Discussed in a direct call.

PR #1465, opened 2021-03-18
Prefer std::unique_ptr in pcap writer

JoeLoser:

Problem:
- The pcap writer uses raw pointers for the `pcap_t` and `pcap_dumper_t`
  member variables which doesn't express ownership semantics to readers
  of the code. It also requires manual cleanup of the resource in the
  destructor.

Solution:
- Use `std::unique_ptr` for both pointers with appropriate custom
  deleters to cleanup the resource.

Note:
- A similar thing can be done for the pcap reader, but then a move
  constructor needs to be defined since there are places in the code
  relying on the pcap reader being movable. Currently, it is trivially
  copyable and hence movable. Making the pcap data member a
  `std::unique_ptr` changes that.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/format/pcap.hpp
+++ b/libvast/vast/format/pcap.hpp
@@ -13,6 +13,8 @@
 
 #pragma once
 
+#include "vast/fwd.hpp"
+
 #include "vast/address.hpp"
 #include "vast/concept/hashable/hash_append.hpp"
 #include "vast/concept/hashable/xxhash.hpp"
@@ -22,7 +24,6 @@
 #include "vast/format/reader.hpp"
 #include "vast/format/single_layout_reader.hpp"
 #include "vast/format/writer.hpp"
-#include "vast/fwd.hpp"
 #include "vast/port.hpp"
 #include "vast/schema.hpp"
 #include "vast/time.hpp"
@@ -31,6 +32,7 @@
 #include <caf/optional.hpp>
 
 #include <chrono>
+#include <memory>
 #include <pcap.h>
 #include <random>
 #include <unordered_map>
@@ -38,6 +40,18 @@
 namespace vast {
 namespace format {
 namespace pcap {
+
+struct pcap_close_wrapper {
+  inline void operator()(pcap_t* handle) const noexcept {
+    ::pcap_close(handle);
+  }
+};
+
+struct pcap_dump_close_wrapper {
+  inline void operator()(pcap_dumper_t* handle) const noexcept {
+    ::pcap_dump_close(handle);
+  }
+};
 
 /// A PCAP reader.
 class reader : public single_layout_reader {
@@ -116,7 +130,7 @@ public:
   /// @param options The configuration options for the writer.
   explicit writer(const caf::settings& options);
 
-  ~writer();
+  ~writer() override = default;
 
   using format::writer::write;
 
@@ -131,8 +145,10 @@ private:
   size_t flush_interval_ = 0;
   size_t snaplen_ = 65535;
   size_t total_packets_ = 0;
-  pcap_t* pcap_ = nullptr;
-  pcap_dumper_t* dumper_ = nullptr;
+  std::unique_ptr<pcap_t, pcap_close_wrapper> pcap_{nullptr,
+                                                    pcap_close_wrapper{}};
+  std::unique_ptr<pcap_dumper_t, pcap_dump_close_wrapper> dumper_{
+    nullptr, pcap_dump_close_wrapper{}};
   std::string trace_;
 };
 
--- a/libvast/src/format/pcap.cpp
+++ b/libvast/src/format/pcap.cpp
@@ -485,24 +485,17 @@ writer::writer(const caf::settings& options) {
   trace_ = get_or(options, "vast.export.write", vast::defaults::export_::write);
 }
 
-writer::~writer() {
-  if (dumper_)
-    ::pcap_dump_close(dumper_);
-  if (pcap_)
-    ::pcap_close(pcap_);
-}
-
 caf::error writer::write(const table_slice& slice) {
   if (!pcap_) {
 #ifdef PCAP_TSTAMP_PRECISION_NANO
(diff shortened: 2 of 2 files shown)


mavam (approved):
Definitely more C++-y!

Overall this works, but some compilers cannot handle the anonymous namespace. I think you can simply drop it. We usually use it for internal linkage, and in a header file, I'm not sure if we get any benefits on that end.

dominiklohmann:
> but some compilers cannot handle the anonymous namespace.

They should not be used in headers at all: See also [§ SF.21 CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#sf21-dont-use-an-unnamed-anonymous-namespace-in-a-header):

> ### SF.21: Don't use an unnamed (anonymous) namespace in a header
> #### Reason
> It is almost always a bug to mention an unnamed namespace in a header file.

dominiklohmann:
> Currently, it is trivially copyable and hence movable

I would consider this a bug. Moving the PCAP reader would likely break it.

JoeLoser:
> > but some compilers cannot handle the anonymous namespace.
> 
> They should not be used in headers at all: See also [§ SF.21 CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#sf21-dont-use-an-unnamed-anonymous-namespace-in-a-header):
> 
> > ### SF.21: Don't use an unnamed (anonymous) namespace in a header
> > #### Reason
> > It is almost always a bug to mention an unnamed namespace in a header file.

Agreed. Originally, I didn't have the lambdas in anonymous namespaces, but GCC was complaining about subobject linkage, so I had just wrapped them in an anonymous namespace to see if it would make GCC happy. Since we treat the warning as an error, I'll revisit this solution so it works on GCC too.

JoeLoser:
> > Currently, it is trivially copyable and hence movable
> 
> I would consider this a bug. Moving the PCAP reader would likely break it.

Keep in mind the `move` is really doing an "efficient" copy of all of the fields in some sense, so I don't think it's an existing bug FWIW. Making one of the data members not copyable (such as `std::unique_ptr` in my case) is what exposes the difference in behavior intended between copy/move here since the reader would no longer be trivially copyable on the move.

dominiklohmann:
> > > Currently, it is trivially copyable and hence movable
> > 
> > 
> > I would consider this a bug. Moving the PCAP reader would likely break it.
> 
> Keep in mind the `move` is really doing an "efficient" copy of all of the fields in some sense, so I don't think it's an existing bug FWIW. Making one of the data members not copyable (such as `std::unique_ptr` in my case) is what exposes the difference in behavior intended between copy/move here since the reader would no longer be trivially copyable on the move.

Trivially copying is a bug—it stores a `pcap_t*` which must not be free'd twice.

JoeLoser:
> > > > Currently, it is trivially copyable and hence movable
> > > 
> > > 
> > > I would consider this a bug. Moving the PCAP reader would likely break it.
> > 
> > 
> > Keep in mind the `move` is really doing an "efficient" copy of all of the fields in some sense, so I don't think it's an existing bug FWIW. Making one of the data members not copyable (such as `std::unique_ptr` in my case) is what exposes the difference in behavior intended between copy/move here since the reader would no longer be trivially copyable on the move.
> 
> Trivially copying is a bug—it stores a `pcap_t*` which must not be free'd twice.

Right, good point. We should fix that in a separate PR.  👍

JoeLoser:
ping @mavam @dominiklohmann -- found a working solution to make GCC happy. The solution is to just use a struct with templated call operators which is the hand-rolled lambda if you will. This avoids the subobject linkage warning.

FWIW, I think this behavior may change in C++20 with default constructible stateless lambdas (see P0624) and then I think GCC wouldn't warn on the case of using lambdas about subobject linkage.

libvast/vast/format/pcap.hpp
```diff
@@ -39,6 +41,22 @@ namespace vast {
 namespace format {
 namespace pcap {
 
+struct pcap_deleter_t {
+  template <class THandle>
+  auto operator()(THandle handle) {
+    if (handle)
+      ::pcap_close(handle);
+  }
+};
```
dominiklohmann, line 50:
```suggestion
struct pcap_close_wrapper {
  inline void operator()(pcap_t* handle) const noexcept {
    ::pcap_close(handle);
  }
};
```

The deleter of a `unique_ptr` is not called when the underlying pointer is `nullptr`. No need to double check.

I think it makes more sense to name the struct `pcap_close_wrapper`, similar to the `actor_cast_wrapper` we already have in our code base. There's also no need to use a template here.

Please apply the same change to the other deleter.

dominiklohmann (changes requested):

libvast/vast/format/pcap.hpp
JoeLoser: Good call on not needing the template. Just adopted your proposed changes, take a look.

dominiklohmann: You're missing the `inline`. Alternatively you can also move the definition to the source files, which might be a good idea in general so we can make libpcap a private dependency sometime in the future.

JoeLoser: Added the `inline` keyword. I don't want to go down moving the definition to the `cpp` file as the deleter needs to be a complete definition when declaring the `unique_ptr` data members.

dominiklohmann (approved):
Nice work—again! I love refactorings that accidentally fix bugs. 🚀

libvast/vast/format/pcap.hpp
dominiklohmann: I might just do this myself in the future. We want to externalize the PCAP reader/writer into a plugin anyways so we don't always depend on libpcap, and it's only natural to make the change then.

JoeLoser: Seems like a reasonable idea; feel free.

PR #1467, opened 2021-03-18
Prepare repository for VAST 2021.03.25-rc1

dominiklohmann:

This prepares everything required for the first release candidate for the release scheduled for 2021-03-25.

###  :memo: Checklist

File-by-file


cmake/VASTConfigVersion.cmake
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.14...3.19 FATAL_ERROR)
 
-set(VAST_VERSION_FALLBACK "2021.02.24-0-")
+set(VAST_VERSION_FALLBACK "2021.03.25-rc1-0-")
```
mavam, line 3: Independent of this PR, but why do we need a trailing dash?

dominiklohmann: We parse this output later on into different components, and the dash is needed so we can set `VAST_VERSION_COMMIT` to an empty string. Not having this here would mean introducing another two regex replace calls below.

dominiklohmann: (Obviously there may be some regex magician out there that can do all this in a single call, but that person is definitely not me.)

mavam: I'm looking at this purely from a users' perspective: a trailing dash in a version string looks like a bug to me.

If it's a matter of adding a tad more regex plumbing (read: small stuff, not larger changes) then it's worth fixing.

dominiklohmann: This is not a user-facing thing though, this is our internal fallback string if Git is unavailable and the user did not specify a version override.

tobim (approved):
The regexes are a bit crude, but this works for now. I'll open a PR to clean it up later.

cmake/VASTConfigVersion.cmake
mavam: > This is not a user-facing thing though

So it can't show up in `vast version`, man pages, or any other content that users consume?

dominiklohmann: Nope, in there the -0- is stripped away.

mavam: Okay, then I'm not concerned.

PR #1470, opened 2021-03-18
Remove dependency on vast::path in zeek writer

JoeLoser:

Problem:
- `zeek` writer uses `vast::path` which is going away soon.

Solution:
- Change `zeek` writer to use `std::filesystem::path` instead.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/format/zeek.cpp
+++ b/libvast/src/format/zeek.cpp
@@ -518,13 +518,9 @@ writer::writer(const caf::settings& options) {
   auto output
     = get_or(options, "vast.export.write", vast::defaults::export_::write);
   if (output != "-")
-    dir_ = std::move(output);
+    dir_ = std::filesystem::path{std::move(output)};
   show_timestamp_tags_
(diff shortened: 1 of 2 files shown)


mavam (approved):
Legit.

JoeLoser:
I did confirm the permissions set from the system call `::mkdir` from `path.cpp` match that of `std::filesystem::create_directory`. Specifically, it uses `std::filesystem::perms::all` from [CppRef](https://en.cppreference.com/w/cpp/filesystem/perms) which is the equivalent options of `S_IRWXU | S_IRWXG | S_IRWXO`, i.e. octal value of `0777`.

dominiklohmann:
We're in a feature freeze period until after the next release, and would like to hold off from merging this until after the scheduled 2021-03-25 release. We're only merging bug fixes and obviously non-breaking chores until then. I've added a `🚧 blocked` label accordingly.

Other than that I think this is ready to merge, we just don't want to take the risk between the release candidate and the actual release. Don't let this hold you off from creating further PRs. 😉

PR #1471, opened 2021-03-18
Prefer std::unique_ptr in pcap reader

JoeLoser:

Problem:
- `pcap` reader uses raw pointers which doesn't express ownership
  semantics to new readers of the code.

Solution:
- Use `std::unique_ptr` with an appropriate custom deleter.
- Add move and move assignment constructors to the pcap reader so it can
  be moved still.

###  :memo: Checklist

### DIFF
--- a/libvast/src/format/pcap.cpp
+++ b/libvast/src/format/pcap.cpp
@@ -89,11 +89,6 @@ void reader::reset(std::unique_ptr<std::istream>) {
   // reader abstraction.
 }
 
-reader::~reader() {
-  if (pcap_)
-    ::pcap_close(pcap_);
-}
-
 caf::error reader::schema(vast::schema sch) {
   return replace_if_congruent({&packet_type_}, sch);
 }
@@ -113,7 +108,7 @@ vast::system::report reader::status() const {
   if (!pcap_)
     return {};
   auto stats = pcap_stat{};
-  if (auto res = pcap_stats(pcap_, &stats); res != 0)
+  if (auto res = pcap_stats(pcap_.get(), &stats); res != 0)
     return {};
   uint64_t recv = stats.ps_recv - last_stats_.ps_recv;
   if (recv == 0)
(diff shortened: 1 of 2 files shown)


JoeLoser:
@dominiklohmann this PR shows the problem that arise when using `std::unique_ptr` in the pcap reader. I won't have more time to dig into this today (or tomorrow likely), so feel free to run with this if you want.

You want to look at `system/make_source.hpp` and how it calls into `system/source.hpp`. I think we can simplify the placement new for the reader in `system/source.hpp` which will avoid moving the pcap reader which is the crux of the problem. Note the ownership of the reader is from `make_source` where we have a `std::unique_ptr` of the right reader type.

dominiklohmann:
I'll see if I have time to look at this tomorrow. Thanks!

dominiklohmann:
Actually just took a stab at this and the fix is just this:

```diff
diff --git a/libvast/vast/format/pcap.hpp b/libvast/vast/format/pcap.hpp
index ae7c049d9..e3aae55e0 100644
--- a/libvast/vast/format/pcap.hpp
+++ b/libvast/vast/format/pcap.hpp
@@ -64,6 +64,11 @@ public:
   reader(const caf::settings& options, std::unique_ptr<std::istream> in
                                        = nullptr);
 
+  reader(const reader&) = delete;
+  reader& operator=(const reader&) = delete;
+  reader(reader&&) noexcept = default;
+  reader& operator=(reader&&) noexcept = default;
+
   void reset(std::unique_ptr<std::istream> in);
 
   ~reader() override = default;
```

clang-tidy also warned about the ignored rule-of-five here. We still have lots of legacy code that doesn't do this quite right.

JoeLoser:
ping @dominiklohmann -- CI is passing now. Is this something you'd like to include in the upcoming release?

dominiklohmann:
Yes, I would consider this a bug fix. The PCAP reader should not be copyable.

libvast/src/format/pcap.cpp
```diff
@@ -282,8 +276,8 @@ caf::error reader::read_impl(size_t max_events, size_t max_slice_size,
       return finish(f, caf::make_error(ec::end_of_input, "reached end of "
                                                          "trace"));
     if (r == -1) {
-      auto err = std::string{::pcap_geterr(pcap_)};
-      ::pcap_close(pcap_);
+      auto err = std::string{::pcap_geterr(pcap_.get())};
+      ::pcap_close(pcap_.get());
       pcap_ = nullptr;
```
dominiklohmann, line 281: This is just `pcap_ = nullptr` to invoke the new deleter.

dominiklohmann (changes requested):

libvast/src/format/pcap.cpp
JoeLoser: Good spot. Just fixed.

dominiklohmann (approved):
Thanks for this!

JoeLoser:
> Thanks for this!

No problem, happy to stumble across unexpected bugs while refactoring 😄

PR #1472, opened 2021-03-18
Use non-throwing std::filesystem functions in the type registry

tobim:

Change 2 use sites of `exists()`.

###  :memo: Checklist

### DIFF
--- a/libvast/src/system/type_registry.cpp
+++ b/libvast/src/system/type_registry.cpp
@@ -77,18 +77,23 @@ caf::error type_registry_state::save_to_disk() const {
 
 caf::error type_registry_state::load_from_disk() {
   // Nothing to load is not an error.
-  if (!exists(dir)) {
+  std::error_code err{};
+  const auto dir_exists = std::filesystem::exists(dir, err);
+  if (err)
+    return caf::make_error(ec::filesystem_error,
+                           fmt::format("failed to find directory {}: {}", dir,
+                                       err.message()));
+  if (!dir_exists) {
     VAST_DEBUG("{} found no directory to load from", self);
     return caf::none;
   }
-  std::error_code err{};
   const auto fname = filename();
-  const auto exists = std::filesystem::exists(fname, err);
+  const auto file_exists = std::filesystem::exists(fname, err);
   if (err)
     return caf::make_error(ec::filesystem_error,
                            fmt::format("failed to find file {}: {}", fname,
                                        err.message()));
-  if (exists) {
+  if (file_exists) {
     auto buffer = io::read(fname);
     if (!buffer)
       return buffer.error();
@@ -201,11 +206,16 @@ type_registry(type_registry_actor::stateful_pointer<type_registry_state> self,
     },
     [self](atom::load) -> caf::result<atom::ok> {
       VAST_DEBUG("{} loads taxonomies", self);
+      std::error_code err{};
       auto dirs = get_schema_dirs(self->system().config());
       concepts_map concepts;
       models_map models;
       for (const auto& dir : dirs) {
-        if (!exists(dir))
+        const auto dir_exists = std::filesystem::exists(dir, err);
+        if (err)
+          VAST_WARN("{} failed to open directory {}: {}", self, dir,
+                    err.message());
+        if (!dir_exists)
           continue;
         auto yamls = load_yaml_dir(dir);
         if (!yamls)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- 🐞 Insufficient permissions for one of the paths in the `schema-dirs` option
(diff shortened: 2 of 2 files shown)


libvast/src/system/type_registry.cpp
```diff
@@ -77,18 +77,23 @@ caf::error type_registry_state::save_to_disk() const {
 
 caf::error type_registry_state::load_from_disk() {
   // Nothing to load is not an error.
-  if (!exists(dir)) {
+  std::error_code err{};
```
dominiklohmann, line 80: Was this picked up via ADL? My goodness.

libvast/src/system/type_registry.cpp
```diff
@@ -201,11 +206,16 @@ type_registry(type_registry_actor::stateful_pointer<type_registry_state> self,
     },
     [self](atom::load) -> caf::result<atom::ok> {
       VAST_DEBUG("{} loads taxonomies", self);
+      std::error_code err{};
       auto dirs = get_schema_dirs(self->system().config());
       concepts_map concepts;
       models_map models;
       for (const auto& dir : dirs) {
-        if (!exists(dir))
+        auto dir_exists = exists(dir, err);
```
dominiklohmann, line 214: Can you switch this to the explicit `std::filesystem::exists` as well?

libvast/src/system/type_registry.cpp
tobim: `dir` is of type `std::filesystem::path`, so what else would you expect?

dominiklohmann (approved):
I did not test locally. This should be covered by CI. Did you verify that this fixes the issue with the systemd setup?

libvast/src/system/type_registry.cpp
dominiklohmann: No surprise we didn't catch this one during review.

mavam: Gotta love ADL. 😒

tobim:
> Did you verify that this fixes the issue with the systemd setup?

Not yet, I wanted to use the CI artifact for that.

dominiklohmann:
> Not yet, I wanted to use the CI artifact for that.

Consider it approved if the issue is fixed ;)

tobim:
It is indeed fixed.

libvast/src/system/type_registry.cpp
```diff
@@ -201,11 +206,16 @@ type_registry(type_registry_actor::stateful_pointer<type_registry_state> self,
     },
     [self](atom::load) -> caf::result<atom::ok> {
       VAST_DEBUG("{} loads taxonomies", self);
+      std::error_code err{};
       auto dirs = get_schema_dirs(self->system().config());
       concepts_map concepts;
       models_map models;
       for (const auto& dir : dirs) {
-        if (!exists(dir))
+        auto dir_exists = std::filesystem::exists(dir, err);
```
JoeLoser, line 214: Nit: can we make this `const`, please?

libvast/src/system/type_registry.cpp
JoeLoser: Ouch. Sorry I missed this when I introduced this change 😞 

Thanks for the fix.

libvast/src/system/type_registry.cpp
```diff
@@ -201,11 +206,16 @@ type_registry(type_registry_actor::stateful_pointer<type_registry_state> self,
...
+        if (err)
```
JoeLoser, line 215: I think we should do `err.clear()`. In the case where more than one directory doesn't exist (and since we continue in the loop), I think we want to warn on just the ones that don't exist rather than erroneously warn on all of the dirs remaining after the first one fails (and thus `err` is set). Does that make sense? 🤔

tobim: `exists()` calls `err.clear()` itself if it didn't fail. See https://en.cppreference.com/w/cpp/filesystem/exists.

libvast/src/system/type_registry.cpp
tobim: No worries, these things happen.

mavam: This is one of those things that are only obvious in hindsight. We probably would have run into the exact same issue.

JoeLoser: Pedantically, can you help me understand why `exists` unqualified would pick up `std::filesystem::exists`? I don't see us do a `using namespace std;` or anything like that. Who's bringing it into scope?

libvast/src/system/type_registry.cpp
JoeLoser: Ah, perfect then. 👍

libvast/src/system/type_registry.cpp
dominiklohmann: See this recent article on (among other things) ADL: https://preshing.com/20210315/how-cpp-resolves-a-function-call/

PR #1473, opened 2021-03-18
Attempt to fix assertition failure in component shutdown handler

mavam:

This PR attempts to fix an assertition failure in the DOWN handler in the node. The invariant should be that components in the registry are always monitored.

###  :memo: Checklist

Commit-by-commit.

### DIFF
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -345,8 +345,8 @@ caf::message kill_command(const invocation& inv, caf::actor_system&) {
           rp.deliver(atom::ok_v);
         },
         [=](const caf::error& err) mutable {
-          VAST_DEBUG("{} failed to terminate component {}: {}", self, label,
-                     err);
+          VAST_WARN("{} failed to terminate component {}: {}", self, label,
+                    err);
           rp.deliver(err);
         });
   }
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -307,7 +307,6 @@ caf::message kill_command(const invocation& inv, caf::actor_system&) {
   if (!component) {
     rp.deliver(caf::make_error(ec::unspecified, "no such component: " + label));
   } else {
-    self->demonitor(component);
     terminate<policy::parallel>(self, component)
       .then(
         [=](atom::done) mutable {
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -476,10 +476,13 @@ node_state::spawn_command(const invocation& inv,
       rp.deliver(component.error());
       return caf::make_message(std::move(component.error()));
     }
-    self->monitor(*component);
-    auto okay = self->state.registry.add(*component, std::move(comp_type),
-                                         std::move(label));
-    VAST_ASSERT(okay);
+    if (self->state.registry.add(*component, std::move(comp_type),
+                                 std::move(label)))
+      self->monitor(*component);
+    else
+      return caf::make_message(caf::make_error(
+        ec::unspecified,
+        fmt::format("{} failed to add component {} to registry", self, label)));
     rp.deliver(*component);
     return caf::make_message(*component);
   };
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -607,16 +607,7 @@ node(node_actor::stateful_pointer<node_state> self, std::string name, path dir,
     for (const auto& label : remaining)
       schedule_teardown(label);
     // Finally, bring down the filesystem.
-    // FIXME: there's a super-annoying bug that makes it impossible to receive a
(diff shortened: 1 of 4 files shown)


libvast/src/system/node.cpp
```diff
@@ -481,10 +486,12 @@ node_state::spawn_command(const invocation& inv,
       rp.deliver(component.error());
       return caf::make_message(std::move(component.error()));
     }
-    self->monitor(*component);
-    auto okay = self->state.registry.add(*component, std::move(comp_type),
-                                         std::move(label));
-    VAST_ASSERT(okay);
+    if (self->state.registry.add(*component, std::move(comp_type), label))
```
dominiklohmann, line 489: Both add and remove in the component registry should probably be marked nodiscard.

mavam: Yep, added.

libvast/src/system/node.cpp
```diff
@@ -320,7 +320,8 @@ caf::message kill_command(const invocation& inv, caf::actor_system&) {
           rp.deliver(atom::ok_v);
         },
         [=](const caf::error& err) mutable {
-          VAST_DEBUG("{} terminated component {}", self, label);
+          VAST_DEBUG("{} failed to terminate component {}: {}", self, label,
```
dominiklohmann, line 351: Should probably be a warning rather than a debug statement.

libvast/src/system/node.cpp
```diff
@@ -597,11 +619,10 @@ node(node_actor::stateful_pointer<node_state> self, std::string name, path dir,
     // after the node exits when the filesystem ref count goes to 0. (A
     // shutdown after the node won't be an issue because the filesystem is
     // currently stateless, but this needs to be reconsidered when it changes.)
-    // // TODO: uncomment when we get a DOWN from detached actors.
```
dominiklohmann, line 593: This line was erroneously removed.

libvast/vast/system/component_registry.hpp
```diff
@@ -47,12 +47,21 @@ class component_registry {
...
+  add(caf::actor comp, std::string type, std::string label = {});
 
   /// Removes a component from the registry.
-  /// @param comp The component to remove
-  /// @returns `true` iff the component was deleted successfully.
-  bool remove(const caf::actor& comp);
+  /// @param label The label of the component.
+  /// @returns The deleted component or a default-constructed component if
+  /// *label* does not identify an existing component.
+  [[nodiscard]] component remove(const std::string& label);
```
dominiklohmann, line 58: API wise I think both remove functions should return a `caf::expected<component>` instead. Relying on special values makes it hard to reason about the code at the call site without having read the comment above this function.

dominiklohmann (commented):
The code looks alright to me. I have a few minor requests. Let's hope this fixes the actual issue.

mavam:
On my local machine, this works as well:
```diff
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -611,16 +611,7 @@ node(node_actor::stateful_pointer<node_state> self, std::string name, path dir,
     for (const auto& label : remaining)
       schedule_teardown(label);
     // Finally, bring down the filesystem.
-    // FIXME: there's a super-annoying bug that makes it impossible to receive a
-    // DOWN message from the filesystem during shutdown, but *only* when the
-    // filesystem is detached! This might be related to a bug we experienced
-    // earlier: https://github.com/actor-framework/actor-framework/issues/1110.
-    // Until it gets fixed, we cannot add the filesystem to the set of
-    // sequentially terminated actors but instead let it implicitly terminate
-    // after the node exits when the filesystem ref count goes to 0. (A
-    // shutdown after the node won't be an issue because the filesystem is
-    // currently stateless, but this needs to be reconsidered when it changes.)
-    // components.push_back(std::move(*filesystem));
+    components.push_back(std::move(*filesystem));
     auto shutdown_kill_timeout = shutdown_grace_period / 5;
     shutdown<policy::sequential>(self, std::move(components),
                                  shutdown_grace_period, shutdown_kill_timeout);
```

libvast/vast/system/component_registry.hpp
mavam: Okay, how do you feel about making them return `caf::no_error` in case nothing is found? I.e., a default-constructed `caf::error`.

dominiklohmann: Or `ec::lookup_error`? It's probably an error if I tried to remove a component but it's not in the registry.

dominiklohmann:
This also still needs a changelog entry

dominiklohmann:
> I might add this as separate commit afterwards.

I think this should go into a separate follow-up PR. I'd rather not merge that change in the freeze period.

dominiklohmann:
This just needs the changelog entry, then it's ready to go imo.

dominiklohmann (approved):

PR #1474, opened 2021-03-19
Remove unnecessary clearing of std::error_code in lsvast

JoeLoser:

Problem:
- `std::error_code` value is cleared explicitly unncessarily. It is not
  necessary since the call to `std::filesystem::is_regular_file` in the
  case above when the path is a directory already clears the
  `std::error_code` in the case of no errors.

Solution:
- Do not clear the `std::error_code` explicitly.

###  :memo: Checklist


JoeLoser:
@tobim

tobim (approved):
Yes!

dominiklohmann:
We can safely merge this I think.

PR #1475, opened 2021-03-19
Remove dependency on vast::path in pcap formatter

JoeLoser:

Problem:
- `pcap` formatter depends on `vast::path` for checking if the input
  exists.

Solution:
- Use `std::filesystem::exists` instead.

Note:
- Did not change `input_` data member to be a `std::filesystem::path`
  type since the pcap formatter allows for the input to be `-` meaning
  `stdin` which isn't well representable using `std::filesystem::path`.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/format/pcap.cpp
+++ b/libvast/src/format/pcap.cpp
@@ -18,7 +18,6 @@
 #  include "vast/ether_type.hpp"
 #  include "vast/format/pcap.hpp"
 #  include "vast/logger.hpp"
-#  include "vast/path.hpp"
 #  include "vast/span.hpp"
 #  include "vast/table_slice.hpp"
 #  include "vast/table_slice_builder.hpp"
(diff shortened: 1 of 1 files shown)


mavam (approved):
Looks good to me.

libvast/src/format/pcap.cpp
```diff
@@ -27,7 +26,7 @@
 #  include <caf/settings.hpp>
 
 #  include <cstddef>
-#  include <pcap.h>
```
JoeLoser, line 30: @dominiklohmann looks like a rebase error in your top-commit. I'll rebase and push to this branch in a minute.

dominiklohmann: Yes sorry, that shouldn't have been removed. Thanks for the quick response. We were just looking at which PRs to merge in what order now that they're unblocked.

We gave you a shoutout in the release notes and the release blog post btw. 😊

JoeLoser: No problem, this PR should be good now.

Thanks for the kind words in the release notes and blog post -- just gave them a read 😊

PR #1480, opened 2021-03-19
Fix the reported VAST version when using version overrides

tobim:

This bug is very recent, so I don't think it needs a changelog entry.


dominiklohmann (changes requested):
I don't think this should be done in here. Can you add the quoting inside config.cpp.in?

tobim:
> I don't think this should be done in here. Can you add the quoting inside config.cpp.in?

You misunderstand, the problem is that the fallback version will be used for `config.cpp` if `VastConfigVersion.cmake` can't execute git. This manifests in the static build.

dominiklohmann (approved):

PR #1482, opened 2021-03-19
Simplify the version extraction

tobim:

This doesn't rely on a bunch of regexes that are hard to understand.


dominiklohmann:
I'd prefer to merge this after the rc2.

dominiklohmann (commented):
This seems to work nicely in practice. I only have a minor complaint, and that is variable naming.

Please use lowercase variables for everything not prefixed with `VAST_` to avoid accidental clashes. Especially things `VERSION_LIST*` should be `VAST_VERSION_LIST*` or `version_list*`—which one I do not care.

dominiklohmann (approved):
Thanks!

PR #1483, opened 2021-03-19
Raise Zstd compression level for Ccache in CI

dominiklohmann:

This should help reduce cache rotations when working on many open PRs at the same time at the cost of slightly increasing the average run duration.

###  :memo: Checklist

n/t

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -25,7 +25,9 @@ env:
   CCACHE_SLOPPINESS: 'file_macro,time_macros'
   CCACHE_DIR: '${{ github.workspace }}/.ccache'
   CCACHE_COMPRESS: true
-  CCACHE_COMPRESSLEVEL: 9
+  # Zstd compression level; up to 19 is supported by all versions.
+  # https://ccache.dev/manual/latest.html#config_compression_level
+  CCACHE_COMPRESSLEVEL: 19
   CCACHE_ABSSTDERR: true
   BUILD_DIR: build
 jobs:


.github/workflows/vast.yaml
```diff
@@ -25,7 +25,9 @@ env:
   CCACHE_SLOPPINESS: 'file_macro,time_macros'
   CCACHE_DIR: '${{ github.workspace }}/.ccache'
   CCACHE_COMPRESS: true
-  CCACHE_COMPRESSLEVEL: 9
+  # Zstd compression level; up to 19 is supported by all versions.
+  # https://ccache.dev/manual/latest.html#config_compression_level
+  CCACHE_COMPRESSLEVEL: 19
```
mavam, line 30: I would imagine that there is a non-linear increase in runtime with higher levels. Can we do some local tests to see what cost-benefit we are getting ourselves into?

dominiklohmann: We only have 5GB of cache total for the repository, and right now every run is using around 1.5 GB. We just need to play around with it to see what makes sense for our workflows, so I suggest we take the binary search approach. I have a gut feeling that 19 will be alright to use, though. I'm planning to re-run CI a few times here to be able to see the cache size and average run time for 100% cache hits—which is where the decompression penalty is the largest.

dominiklohmann: After some re-runs, here are my observations: The worst-case runtime increase seems to be less than 3 minutes. The combined cache size for one CI run went down from ~1.5 GB to ~0.5 GB. I'd say this is definitely worth merging.

Let's talk on Monday about this within @tenzir/backend.

mavam: If we have a little data on the time vs. space dimensions, I'm all for this.

The results seems promising: cache size down by 3x for 3 mins extra. I'd be curious to see how much lower we can go until the gains in space become no longer so pronounced.

dominiklohmann: > I'd be curious to see how much lower we can go until the gains in space become no longer so pronounced.

Technically it depends on the Zstd version that Ccache was built against, so we'd need to check that for all entries in our build matrix. 19 is the highest we can safely use per the linked Ccache docs.

dominiklohmann: > The library supports regular compression levels from 1 up to `ZSTD_maxCLevel()`, which is currently 22. Levels >= 20, labeled `--ultra`, should be used with caution, as they require more memory. The library also offers negative compression levels, which extend the range of speed vs. ratio preferences. The lower the level, the faster the speed (at the cost of compression).

From http://facebook.github.io/zstd/zstd_manual.html

tobim:
I did a local test putting all `.o` files of a debug build tree into a compressed tarball:
```
$ time tar -I zstd -cvf allfiles2.tar.zst -T files.txt

Executed in    8.31 secs   fish           external
   usr time    8.72 secs    0.00 micros    8.72 secs
   sys time    1.94 secs  1003.00 micros    1.93 secs

xo ~/t/b/v/debug $ ls -l allfiles2.tar.zst
-rw-r--r-- 1 tobim users 348,683,636 Mar 22 10:09 allfiles2.tar.zst

$ export ZSTD_CLEVEL=19
$ time tar -I zstd -cvf allfiles2.tar.zst -T files.txt

Executed in  639.77 secs   fish           external
   usr time  640.07 secs    0.00 micros  640.07 secs
   sys time    1.52 secs  846.00 micros    1.52 secs

xo ~/t/b/v/debug $ ls -l allfiles2.tar.zst
-rw-r--r-- 1 tobim users 296,553,108 Mar 22 10:20 allfiles2.tar.zst
```

Compression time increased from 8 seconds to almost 11 minutes. Files size is reduced to by roughly 50 MB or 15%.

dominiklohmann:
@tobim I'm not sure that's a fair comparison. You must measure using `ccache --recompress 19` to get realistic values. If I understand correctly, Ccache is not compressing the whole build directory, but rather individual object files.

tobim:
It seems to be pretty much the same here:
```
xo ~ $ ccache -x                                                                             vast-dev-gcc
Scanning... 100.0% [=====================================================================================]

Total data:              4.6 GB (4.6 GB disk blocks)
Compressed data:         4.5 GB (23.1% of original size)
  - Original data:      19.6 GB
  - Compression ratio: 4.322 x  (76.9% space savings)
Incompressible data:    46.6 MB
xo ~ $                                                                                       vast-dev-gcc
xo ~ $ time ccache --recompress 9                                                            vast-dev-gcc
Recompressing... 100.0% [================================================================================]

Original data:          19.6 GB
Old compressed data:     4.5 GB (23.2% of original size)
  - Compression ratio: 4.312 x  (76.8% space savings)
New compressed data:     4.1 GB (21.1% of original size)
  - Compression ratio: 4.733 x  (78.9% space savings)
Size change:          -402.8 MB

________________________________________________________
Executed in   73.30 secs   fish           external
   usr time  1076.13 secs  560.00 micros  1076.13 secs
   sys time   24.60 secs  236.00 micros   24.60 secs

xo ~ $ ccache -x                                                                             vast-dev-gcc
Scanning... 100.0% [=====================================================================================]

Total data:              4.2 GB (4.2 GB disk blocks)
Compressed data:         4.1 GB (21.1% of original size)
  - Original data:      19.6 GB
  - Compression ratio: 4.743 x  (78.9% space savings)
Incompressible data:    46.6 MB
xo ~ $ time ccache --recompress 19                                                           vast-dev-gcc
Recompressing... 100.0% [================================================================================]

Original data:          19.6 GB
Old compressed data:     4.1 GB (21.1% of original size)
  - Compression ratio: 4.743 x  (78.9% space savings)
New compressed data:     4.0 GB (20.6% of original size)
  - Compression ratio: 4.853 x  (79.4% space savings)
Size change:           -94.3 MB

________________________________________________________
Executed in   21.87 mins   fish           external
   usr time  319.85 mins  607.00 micros  319.85 mins
   sys time    7.25 mins  255.00 micros    7.25 mins

xo ~ $ ccache -x                                                                             vast-dev-gcc
Scanning... 100.0% [=====================================================================================]

Total data:              4.1 GB (4.1 GB disk blocks)
Compressed data:         4.0 GB (20.6% of original size)
  - Original data:      19.6 GB
  - Compression ratio: 4.853 x  (79.4% space savings)
Incompressible data:    46.6 MB
```

dominiklohmann:
The numbers look a lot better on my machine. So the worst case for you locally is a ~20 minute increase when there is a 0% cache hit rate. However, getting empty caches will be much rarer with this change—what do you suggest we do here, @tobim?

tobim:
> However, getting empty caches will be much rarer with this change

My findings disagree with this. Can you share the details of you got yours and post your actual numbers?

dominiklohmann:
> > However, getting empty caches will be much rarer with this change
> 
> My findings disagree with this. Can you share the details of how you got yours

The overall cache size goes down. We mostly get the almost 1 hour long CI runs because the caches were rotated out by GitHub Actions.

With this change, we increase the run duration when not getting cache hits. But on the flipside, we often avoid not having a cache at all, which is when CI right now takes the longest.

You can also see that when re-running the CI within this PR it only takes 9 minutes, which does not match your numbers at all. Files that were matched in the cache do not need to be recompressed.

> and post your actual numbers?

Going from no compression at all on a Debug build for the target `vast` to compression level 9 takes around 25 seconds, going all the way up to 19 takes 3 minutes and 31 seconds. Only did a single run, so obviously the numbers may vary.

dominiklohmann:
Closing this as we did not come to an agreement. We can always revisit this later.

PR #1484, opened 2021-03-22
Avoid using privately linked libpcap publicly

dominiklohmann:

Before this change, we linked against libpcap privately, but used it in our public headers—that's a bug. This change makes it a truly private dependency.

###  :memo: Checklist

File-by-file. Confirm private linkage against `vast-test` and `libvast` targets.

The second commit update the style-check workflow a bit because my local clang-format-11 and clang-format-9 in CI were in disagreement.

### DIFF
--- a/libvast/src/format/pcap.cpp
+++ b/libvast/src/format/pcap.cpp
@@ -10,14 +10,13 @@
 
 #if VAST_ENABLE_PCAP
 
-#  include "vast/format/pcap.hpp"
-
 #  include "vast/community_id.hpp"
 #  include "vast/defaults.hpp"
 #  include "vast/detail/assert.hpp"
 #  include "vast/detail/byte_swap.hpp"
 #  include "vast/error.hpp"
 #  include "vast/ether_type.hpp"
+#  include "vast/format/pcap.hpp"
 #  include "vast/logger.hpp"
 #  include "vast/path.hpp"
 #  include "vast/span.hpp"
@@ -28,15 +27,15 @@
 #  include <caf/settings.hpp>
 
 #  include <cstddef>
+#  include <pcap.h>
 #  include <string>
 #  include <thread>
 #  include <utility>
 
 #  include <netinet/in.h>
 
-namespace vast {
-namespace format {
-namespace pcap {
+namespace vast::format::pcap {
+
 namespace {
 
 template <class... RecordFields>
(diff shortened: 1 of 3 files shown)


libvast/src/format/pcap.cpp
```diff
@@ -28,15 +27,15 @@
...
 
 #  include <cstddef>
+#  include <pcap.h>
 #  include <string>
 #  include <thread>
 #  include <utility>
 
 #  include <netinet/in.h>
 
-namespace vast {
```
JoeLoser, line 37: Nit: I know your local Clang format 11 collapsed these nested namespace, but it may be cleaner to just apply this in a one-shot-go for all files in a separate PR. This has the benefit of being consistent going forward rather than this sort of thing coming up as small diffs in many files over time.  We can ignore the Git revision for the formatting easily to avoid `git blame` and friends for just formatting revisions in the Git history.

libvast/src/format/pcap.cpp
```diff
@@ -60,7 +59,21 @@ inline const auto pcap_packet_type = make_packet_type();
...
+}
+
+void pcap_dump_close_wrapper::operator()(
+  struct pcap_dumper* handle) const noexcept {
+  ::pcap_dump_close(handle);
+}
+
+void pcap_stat_delete_wrapper::operator()(
+  struct pcap_stat* handle) const noexcept {
+  delete handle;
```
JoeLoser, line 75: I'm surprised there isn't an exported C function for cleaning up the `pcap_stat` handle. I briefly looked around and didn't see one, which is odd I'd imagine.

libvast/src/format/pcap.cpp
```diff
@@ -109,22 +122,23 @@ const char* reader::name() const {
 }
 
 vast::system::report reader::status() const {
+  VAST_ASSERT(last_stats_);
```
JoeLoser, line 125: Do we have `VAST_ASSERT_MSG` or something like that so we can provide a useful assert message for this precondition?

libvast/vast/format/pcap.hpp
```diff
@@ -118,7 +120,8 @@ class reader : public single_layout_reader {
   bool community_id_;
   type packet_type_;
   double drop_rate_threshold_;
-  mutable pcap_stat last_stats_;
```
JoeLoser, line 121: Down with `mutable` keyword 😆

dominiklohmann: It's necessary here.

libvast/src/format/pcap.cpp
dominiklohmann: We can always do `last_stats_ && "message"`, but seems to be unnecessary here. The assertion is reported with file and line.

libvast/src/format/pcap.cpp
dominiklohmann: This was actually clang-tidy, not clang-format. The clang-format issue was totally separate.

> We can ignore the Git revision for the formatting easily to avoid git blame and friends for just formatting revisions in the Git history.

It's a matter of policy. Right now we only apply clang-format to the hunks that were changed from the PR merge base.

libvast/src/format/pcap.cpp
dominiklohmann: I verified that this doesn't exist.

libvast/src/format/pcap.cpp
mavam: I think it's still valuable to throw the user a message in the face. It helps us developers to quickly see what invariant got violated, instead having to start digging why, e.g., a to-bool-convertible variable was false.

mavam: Great, thanks.

lava (approved):
Looks good to me, and I verified that I can still import and export pcap data with a VAST built from this PR.

PR #1485, opened 2021-03-23
Fix shutdown with importers running

dominiklohmann:

This fixes a recently introduced bug—the shutdown was not working while imports were still running.

###  :memo: Checklist

n/t

### DIFF
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -580,13 +580,31 @@ node(node_actor::stateful_pointer<node_state> self, std::string name, path dir,
   // Terminate deterministically on shutdown.
   self->set_exit_handler([=](const caf::exit_msg& msg) {
     VAST_DEBUG("{} got EXIT from {}", self, msg.source);
-    std::vector<caf::actor> components;
-    auto schedule_teardown = [&](const std::string& label) {
-      auto component = self->state.registry.remove(label);
-      if (component) {
-        VAST_DEBUG("{} schedules {} for shutdown", self, label);
-        self->demonitor(component->actor);
-        components.push_back(std::move(component->actor));
+    std::vector<caf::actor> scheduled_for_teardown;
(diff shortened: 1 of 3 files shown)


libvast/src/system/node.cpp
```diff
@@ -580,13 +580,31 @@ node(node_actor::stateful_pointer<node_state> self, std::string name, path dir,
...
-        components.push_back(std::move(component->actor));
+    std::vector<caf::actor> scheduled_for_teardown;
+    // TODO: A recent refactoring introduced a bug that caused this function to
+    // only remove by label instead of by type or label. This was not caught in
+    // our integration test suite because we do not currently test continuous
+    // imports/exports in them. We should evaluate how we can test for shutdown
+    // bugs.
+    auto schedule_teardown = [&](const std::string& type_or_label) {
+      if (is_singleton(type_or_label)) {
+        auto component = self->state.registry.remove(type_or_label);
```
lava, line 591: Nit: Exactly the same code is written as
```
        auto component = self->state.registry.remove(type_or_label);
        if (component) {
```
here but
```
if (auto removed = self->state.registry.remove(component)) {
```
below

libvast/src/system/node.cpp
```diff
@@ -595,7 +613,8 @@ node(node_actor::stateful_pointer<node_state> self, std::string name, path dir,
     // Take out the filesystem, which we terminate at the very end.
     auto filesystem = deregister_component(self, "filesystem");
     // Tear down the ingestion pipeline from source to sink.
-    auto pipeline = {"source", "importer", "index", "archive", "exporter"};
+    auto pipeline
+      = {"source", "importer", "index", "archive", "exporter", "sink"};
```
tobim, line 617: Make it explicit to the reader of this code that the order is important.

lava (approved):
Changes look good to me, a regression test would be nice but as the comment already explains this is a much bigger issue we still need to solve.

PR #1494, opened 2021-03-25
Fix type of suiricata.rdp.rdp.reason

dominiklohmann:

As reported by @satta.

This is a very small fix, so I suggest we sneak this into the release.


lava (approved):
Seems consistent with the upstream docs for `reason`:

> The “reason” field will contain a text summary of this code. Possible values:

PR #1495, opened 2021-03-25
Make the disk-monitor a singleton component

tobim:

This fixes a bug that blocks the server process from exiting after a shutdown was initiated.

###  :memo: Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- 🐞 Enabling the disk budget feature no longer prevents the server process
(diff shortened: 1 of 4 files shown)


dominiklohmann (approved):
Confirmed this to work locally with a setup that I also confirmed to hang without this patch. Please move the new log message to DBEUG. The code looks alright to me—thanks for the `s/_/-` in the actor name as well.

tobim:
> thanks for the s/_/- in the actor name as well.

The process for solving this was:
1. Adding `disk-monitor` to the singletons list and `pipeline` -> didn't work
2. Adding the log message -> :bulb: 
3. Changing the actor name

PR #1496, opened 2021-03-25
Fix false comment in example configuration

dominiklohmann:

Thanks @norg for pointing this one out.

###  :memo: Checklist

n/t


tobim (approved):
We could think about changing to option to `enable-log-rotation`, but this alone isn't worth breaking the config file options.

PR #1499, opened 2021-03-25
Add `suricata.rfb` schema type

satta:

Add a new schema record type for Suricata's `rfb` event type.

###  :memo: Checklist

There should not be any required updates to the reference files, at least `build/vast/integration_RelWithDebInfo.sh -u` did not change anything in the repo. Please double-check if that's the case.

### DIFF
--- a/schema/types/suricata.schema
+++ b/schema/types/suricata.schema
@@ -459,6 +459,46 @@ type suricata.rdp = suricata.component.common + record {
   }
 }
 
+type suricata.rfb = suricata.component.common + record {
+  rfb: record {
+    client_protocol_version: record {
+      major: string,
+      minor: string
+    },
+    server_protocol_version: record {
+      major: string,
+      minor: string
+    },
+    authentication: record {
+      security_type: count,
+      vnc: record {
+        challenge: string,
+        response: string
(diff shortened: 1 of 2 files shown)


schema/types/suricata.schema
```diff
@@ -459,6 +459,46 @@ type suricata.rdp = suricata.component.common + record {
   }
 }
 
+type suricata.rfb = suricata.component.common + record {
+  rfb: record {
+    client_protocol_version: record {
+      major: string,
+      minor: string
```
tobim, line 466: Why are the major and minor version fields strings?

satta: Because they are strings in the protocol and could theoretically be non-integers. See https://tools.ietf.org/html/rfc6143#section-7.1.1

tobim: So the `x`s and `y`s in there can technically be any ASCII character. Do you mind adding a comment with these 2 links?

satta: Well, they would be JSON-escaped of course, But yes, they could be, Sure, I can add a comment.

tobim (approved):

PR #1500, opened 2021-03-25
Don't wait for unpersisted partitions when notifying flush listeners

lava:
###  :memo: Checklist

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -292,12 +292,13 @@ void index_state::add_flush_listener(flush_listener_actor listener) {
 
 // The whole purpose of the `-b` flag is to somehow block until all imported
 // data is available for querying, so we have to layer hack upon hack here to
-// achieve this most of the time.
+// achieve this most of the time. This is only used for integration tests.
+// Note that there's still a race condition here if the call to
+// `notify_flush_listeners()` arrives when there's still data en route to
+// the unpersisted partitions.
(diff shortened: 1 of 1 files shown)


tobim (approved):
This check was seemingly incorrect anyway, and anything only related to `-b`shouldn't get a changelog entry.

lava:
I think "incorrect" is a bit strong, let me expand on the tradeoff here. The main thing to keep in mind is that the notifying stream manager may trigger every time a data point is acked by downstream, or when a stream is closed by upstream, if the stage is "clean" at that point in time.

Before this change the following race was possible:

1. Run VAST such that two partitions are created almost at the same time. Ideally both of these get exactly one table slice so we only have two packets in total going through the stream.
2. When the stream is closed ends, both packets are already acked by downstream, so we only get a *single* call to notify_flush_listeners()
3. However, when that call is made the `persist` from one of the partitions is still outstanding, so we don't act in the handler.
4. Shutdown hang

After this change the following race becomes possible:

  1. Run VAST with a very small partition size and table slice size, so that *lots* of partitions finish at the same time and there's a build-up of unfinished partitions.
  3. The client subscribes as a flush listener, the index detects that the stream is clean and forwards the subscription to the active partition
  2. The active partition Y gets the last of its data and notifies the flush listeners
  3. The subsequent query is run including results for partition X -> no hits
  4. The last stream data arrives at partition X

tobim:
Thank you for the writeup. What I meant by "incorrect" is that we shouldn't simply return, but treat the unpersisted partitions in the same way as the active partition.

PR #1502, opened 2021-03-26
Move load_contents out of path.hpp

JoeLoser:

Problem:
- Callers of `load_contents` are forced to construct a `vast::path` even
  if they are already working with a `std::filesystem::path`. This is
  because `load_contents` is coupled to only work with `vast::path`
  currently.

Solution:
- Move `load_contents` into its own header since it is a useful function
  even after `vast::path` is removed. Make it a function template so it
  can easily work on different path types such as
  `std::filesystem::path` and `vast::path`. Eventually, once all callers
  are using `std::filesystem::path`, we can make it a function rather
  than a function template.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/data.cpp
+++ b/libvast/src/data.cpp
@@ -14,6 +14,7 @@
 #include "vast/concept/printable/vast/json.hpp"
 #include "vast/detail/assert.hpp"
 #include "vast/detail/filter_dir.hpp"
+#include "vast/detail/load_contents.hpp"
 #include "vast/detail/narrow.hpp"
 #include "vast/detail/overload.hpp"
 #include "vast/detail/string.hpp"
@@ -576,7 +577,7 @@ caf::expected<data> from_yaml(std::string_view str) {
 }
 
 caf::expected<data> load_yaml(const std::filesystem::path& file) {
-  auto contents = load_contents(vast::path{file.string()});
+  const auto contents = detail::load_contents(file);
   if (!contents)
     return contents.error();
   if (auto yaml = from_yaml(*contents))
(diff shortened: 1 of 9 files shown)


tobim:
Not that I'm against this, but why don't you simply add an overload instead or turning this innocent little function into a template?
```
caf::expected<std::string> load_contents(const std::filesystem::path& p) {
  return load_contents(vast::path{p.string()});
}
```

JoeLoser:
> Not that I'm against this, but why don't you simply add an overload instead or turning this innocent little function into a template?
> 
> ```
> caf::expected<std::string> load_contents(const std::filesystem::path& p) {
>   return load_contents(vast::path{p.string()});
> }
> ```

Sure, we can go that approach.  The main idea is just getting it out of `path.hpp` since callers won't call it with `vast::path` soon whatsoever since `vast::path` will be removed.

JoeLoser:
> > Not that I'm against this, but why don't you simply add an overload instead or turning this innocent little function into a template?
> > ```
> > caf::expected<std::string> load_contents(const std::filesystem::path& p) {
> >   return load_contents(vast::path{p.string()});
> > }
> > ```
> 
> Sure, we can go that approach. The main idea is just getting it out of `path.hpp` since callers won't call it with `vast::path` soon whatsoever since `vast::path` will be removed.

@dominiklohmann @tobim take a look now. I've gone the overloads approach. Note that I wrote it in terms of `vast::path` one calling the `std::filesystem::path` one since `vast::path` will go away soon.

mavam (approved):
Looks good now! 

(For context: putting the least possible pressure on the compiler is something we've been prioritizing internally recently to accelerate CI.)

PR #1504, opened 2021-03-27
Minimize path conversions for chunk mmap and system index

JoeLoser:

See commits one-by-one for the independent, isolated changes. This changes `chunk`'s `mmap` to work on `std::filesystem::path` instead of `vast::path`. When updating the callers in system index, that motivated the change of converting its internal uses of `vast::path` to be `std::filesystem::path`.

###  :memo: Checklist

Commit-by-commit.

### DIFF
--- a/libvast/src/chunk.cpp
+++ b/libvast/src/chunk.cpp
@@ -14,7 +14,6 @@
 #include "vast/io/read.hpp"
 #include "vast/io/save.hpp"
 #include "vast/logger.hpp"
-#include "vast/path.hpp"
 
 #include <caf/deserializer.hpp>
 #include <caf/make_counted.hpp>
@@ -61,7 +60,8 @@ chunk_ptr chunk::mmap(const std::filesystem::path& filename, size_type size,
     std::error_code err{};
     size = std::filesystem::file_size(filename, err);
     if (size == static_cast<std::uintmax_t>(-1)) {
-      VAST_INFO("failed to get file size for filename");
+      VAST_ERROR("failed to get file size for filename {}: {}", filename,
+                 err.message());
       return nullptr;
     }
   }
--- a/libvast/src/chunk.cpp
+++ b/libvast/src/chunk.cpp
@@ -54,16 +54,19 @@ chunk_ptr chunk::make(view_type view, deleter_type&& deleter) noexcept {
   return chunk_ptr{new chunk{view, std::move(deleter)}, false};
 }
 
-chunk_ptr chunk::mmap(const path& filename, size_type size, size_type offset) {
+chunk_ptr chunk::mmap(const std::filesystem::path& filename, size_type size,
+                      size_type offset) {
   // Figure out the file size if not provided.
   if (size == 0) {
-    auto s = file_size(filename);
-    if (!s)
+    std::error_code err{};
+    size = std::filesystem::file_size(filename, err);
+    if (size == static_cast<std::uintmax_t>(-1)) {
+      VAST_INFO("failed to get file size for filename");
       return nullptr;
-    size = *s;
+    }
   }
   // Open and memory-map the file.
-  auto fd = ::open(filename.str().c_str(), O_RDONLY, 0644);
+  auto fd = ::open(filename.c_str(), O_RDONLY, 0644);
   if (fd == -1)
     return {};
   auto map = ::mmap(nullptr, size, PROT_READ, MAP_SHARED, fd, offset);
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -108,19 +108,19 @@ namespace vast::system {
 
 namespace {
 
-caf::error
-extract_partition_synopsis(const vast::path& partition_path,
-                           const vast::path& partition_synopsis_path) {
+caf::error extract_partition_synopsis(
+  const std::filesystem::path& partition_path,
+  const std::filesystem::path& partition_synopsis_path) {
   // Use blocking operations here since this is part of the startup.
-  auto chunk = chunk::mmap(std::filesystem::path{partition_path.str()});
+  auto chunk = chunk::mmap(partition_path);
   if (!chunk)
     return caf::make_error(ec::system_error, "could not mmap partition at "
-                                               + partition_path.str());
+                                               + partition_path.string());
   const auto* partition = fbs::GetPartition(chunk->data());
   if (partition->partition_type() != fbs::partition::Partition::v0)
     return caf::make_error(ec::format_error, "found unsupported version for "
                                              "partition "
-                                               + partition_path.str());
+                                               + partition_path.string());
   const auto* partition_v0 = partition->partition_as_v0();
   VAST_ASSERT(partition_v0);
   partition_synopsis ps;
@@ -134,17 +134,18 @@ extract_partition_synopsis(const vast::path& partition_path,
   auto flatbuffer = ps_builder.Finish();
   fbs::FinishPartitionSynopsisBuffer(builder, flatbuffer);
   auto chunk_out = fbs::release(builder);
-  return io::save(std::filesystem::path{partition_synopsis_path.str()},
+  return io::save(partition_synopsis_path,
                   span{chunk_out->data(), chunk_out->size()});
 }
 
 } // namespace
 
-vast::path index_state::partition_path(const uuid& id) const {
+std::filesystem::path index_state::partition_path(const uuid& id) const {
   return dir / to_string(id);
 }
 
-vast::path index_state::partition_synopsis_path(const uuid& id) const {
+std::filesystem::path
+index_state::partition_synopsis_path(const uuid& id) const {
   return synopsisdir / (to_string(id) + ".mdx");
 }
 
@@ -153,9 +154,10 @@ partition_actor partition_factory::operator()(const uuid& id) const {
   VAST_ASSERT(std::find(state_.persisted_partitions.begin(),
                         state_.persisted_partitions.end(), id)
               != state_.persisted_partitions.end());
-  auto path = state_.partition_path(id);
+  const auto path = state_.partition_path(id);
   VAST_DEBUG("{} loads partition {} for path {}", state_.self, id, path);
-  return state_.self->spawn(passive_partition, id, filesystem_, path);
+  return state_.self->spawn(passive_partition, id, filesystem_,
+                            vast::path{path.string()});
 }
 
 filesystem_actor& partition_factory::filesystem() {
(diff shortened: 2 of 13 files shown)


libvast/src/chunk.cpp
```diff
@@ -54,16 +54,20 @@ chunk_ptr chunk::make(view_type view, deleter_type&& deleter) noexcept {
...
+                      size_type offset) {
   // Figure out the file size if not provided.
   if (size == 0) {
-    auto s = file_size(filename);
-    if (!s)
+    std::error_code err{};
+    size = std::filesystem::file_size(filename, err);
+    if (size == static_cast<std::uintmax_t>(-1)) {
+      VAST_INFO(fmt::format("failed to get file size for filename {}: {}",
+                            filename, err.message()));
```
dominiklohmann, line 65:
```suggestion
      VAST_INFO("failed to get file size for filename {}: {}",
                            filename, err.message());
```

libvast/src/system/index.cpp
```diff
@@ -172,13 +175,15 @@ index_state::index_state(index_actor::pointer self)
 caf::error index_state::load_from_disk() {
   // We dont use the filesystem actor here because this function is only
   // called once during startup, when no other actors exist yet.
-  if (!exists(dir)) {
+  std::error_code err{};
+  const auto file_exists = std::filesystem::exists(dir, err);
+  if (!file_exists) {
```
dominiklohmann, line 180: Still need to check `err` here.

libvast/src/system/index.cpp
```diff
@@ -342,7 +347,9 @@ void index_state::decomission_active_partition() {
   auto part_dir = partition_path(id);
   auto synopsis_dir = partition_synopsis_path(id);
   VAST_DEBUG("{} persists active partition to {}", self, part_dir);
-  self->request(actor, caf::infinite, atom::persist_v, part_dir, synopsis_dir)
+  self
+    ->request(actor, caf::infinite, atom::persist_v,
+              vast::path{part_dir.string()}, vast::path{synopsis_dir.string()})
```
dominiklohmann, line 352: Note: If you want to change these, you need to adapt the `active_partition_actor` type definition in `vast/system/actors.hpp`, and the corresponding lambdas in the `active_partition` function definition return value.

libvast/src/system/index.cpp
```diff
@@ -576,7 +584,7 @@ void index_state::flush_to_disk() {
   auto chunk = fbs::release(builder);
   self
     ->request(caf::actor_cast<caf::actor>(filesystem), caf::infinite,
-              atom::write_v, index_filename(), chunk)
+              atom::write_v, vast::path{index_filename().string()}, chunk)
```
dominiklohmann, line 587: Same as above, except for the filesystem actor.

libvast/src/system/index.cpp
```diff
@@ -853,8 +862,12 @@ index(index_actor::stateful_pointer<index_state> self,
       auto synopsis_path = self->state.partition_synopsis_path(partition_id);
       bool adjust_stats = true;
       if (self->state.persisted_partitions.count(partition_id) == 0u) {
-        if (!exists(path)) {
-          rp.deliver(caf::make_error(ec::logic_error, "unknown partition"));
+        std::error_code err{};
+        const auto file_exists = std::filesystem::exists(path, err);
+        if (!file_exists) {
```
dominiklohmann, line 867: Error should also be checked here.

dominiklohmann (commented):
Another good change. Just a few comments here.

libvast/src/system/index.cpp
JoeLoser: I'm not sure we do based on my reading at [CppRef](https://en.cppreference.com/w/cpp/filesystem/exists). I have checked `err` in previous uses of this function in this codebase, but after some more reading, I don't think it's required. My understanding is the function  will return `false` still but also populate `err` which could be used in log messages for example if desired.

libvast/src/system/index.cpp
JoeLoser: See https://github.com/tenzir/vast/pull/1504#discussion_r602732512

libvast/src/system/index.cpp
JoeLoser: Yeah, I went down that path for a bit this morning, but it seemed a bit more nuanced than that. I'd have to add `CAF_ADD_TYPE_ID` for `std::filesystem::path` and even when I did that and updated the callers appropriately (or so I think), there were still caf errors, which are pretty poor IMO. I'm hoping to change the actors last once all of the surrounding code doesn't use `vast::path` anymore. What do you think? It might be something good for us to pair on when we get to that point.

libvast/src/chunk.cpp
JoeLoser: Fixed, thanks.

libvast/src/system/index.cpp
dominiklohmann: The type ID part is the worst, and we struggle with its requirements all the time. We'll have to think about possibly forward-declaring `std::filesystem::path`, splitting our type ID blocks into multiple (again), or [adding a `caf::inspector_access<std::filesystem::path>` specialization](https://actor-framework.readthedocs.io/en/stable/TypeInspection.html#specializing-inspector-access).

After giving this some thought, I feel like we either pair as you suggest or we do it ourselves internally sometime over the coming weeks.

libvast/src/chunk.cpp
```diff
@@ -54,16 +53,20 @@ chunk_ptr chunk::make(view_type view, deleter_type&& deleter) noexcept {
...
+                      size_type offset) {
   // Figure out the file size if not provided.
   if (size == 0) {
-    auto s = file_size(filename);
-    if (!s)
+    std::error_code err{};
+    size = std::filesystem::file_size(filename, err);
+    if (size == static_cast<std::uintmax_t>(-1)) {
+      VAST_INFO("failed to get file size for filename {}: {}", filename,
+                err.message());
```
mavam, line 64:
```suggestion
      VAST_ERROR("failed to get file size for filename {}: {}", filename,
                 err.message());
```

libvast/src/system/index.cpp
dominiklohmann: I verified that you are indeed correct on this one.

libvast/src/system/index.cpp
dominiklohmann: I've implemented this in https://github.com/tenzir/vast/pull/1508

libvast/src/chunk.cpp
dominiklohmann: This is now the only blocker for this PR, I think.

I wonder whether it makes sense to change the return value of `chunk::mmap` to be `caf::expected<chunk_ptr>`, with the post-condition that if the return value is not an error, it is never nullptr. What do you think @mavam?

mavam: The general guideline should be: if a binary distinction between `nullptr` and correctly created object is not enough information to understand failed construction at the call site, then going through the `expected` monad is reasonable to enrich the error. In other words, if there a multiple causes failure and the call site wants to distinguish them, then `expected` makes sense.

dominiklohmann: I disagree. I personally prefer returning an error here because that can be forwarded to the client process so the user actually sees it.

mavam: My point is different: if you know exactly what failed, you can also generate the error at the call site. There is no value in `expected` in this case. If not, only then `expected` adds value.

dominiklohmann: Then printing here at all leads to duplicate error reporting.

libvast/src/system/index.cpp
JoeLoser: > I've implemented this in #1508

LGTM.  I'll try to incorporate it into the actors likely in a follow-up PR.

libvast/src/system/index.cpp
JoeLoser: > I verified that you are indeed correct on this one.

Thanks for double checking. I'll put up a PR that doesn't check the extra error in the one or two other call sites I've added it in form previous PRs.

libvast/src/chunk.cpp
mavam: Sorry, but now I disagree 🙂 . Let's apply the rationale by looking at the implementation:, which has 3 calls that can fail:

1. `file_size`
2. `open`
3. `mmap`

The call site doesn't know what failed when it receives `nullptr`. So we are in the second scenario that I described. Ergo: I advocate that we *do* need `expected<chunk_ptr>` here.

mavam: (As a result, there should not be logging in here because it will likely be done at the call site.)

JoeLoser: For now, I've changed it to `VAST_ERROR`. I think if we want to change the return value of this API, let's do that in a small, follow-up PR so it can be reviewed in isolation.

In general, should all callers of `std::filesystem::file_size` report an error if we can't get the file size?

FWIW, I don't think callers should be generating the error at the call site as that requires knowledge of the function implementation (e.g. did the file size check fail due to permissions, file not existing, etc...). If we want to go the `expected` route, it should carry the appropriate `err.message()` so the caller can consume it appropriately.

mavam: > For now, I've changed it to `VAST_ERROR`. I think if we want to change the return value of this API, let's do that in a small, follow-up PR so it can be reviewed in isolation.

Agreed.

> In general, should all callers of `std::filesystem::file_size` report an error if we can't get the file size?

Depends on the call site: some times it's a dealbreaker but sometimes it doesn't hurt. In practice we probably want to report this to `ERROR` or `WARNING` log levels.
 
> FWIW, I don't think callers should be generating the error at the call site as that requires knowledge of the function implementation (e.g. did the file size check fail due to permissions, file not existing, etc...). If we want to go the `expected` route, it should carry the appropriate `err.message()` so the caller can consume it appropriately.

I think we're saying the same thing here. The error handling at the call site is uniform, but you still want to know what *exactly* failed if there are multiple options. By consuming the error message, users get actionable insight in what went wrong.

JoeLoser: Yep, we're on the same page @mavam. I'll work on changing this API to return `expected` and introduce a follow-up PR on top of this soon.

Are we all good with this PR for now?

dominiklohmann: All good afaict. I'll let @mavam merge.

JoeLoser: See https://github.com/tenzir/vast/pull/1509 for the follow-up PR. I'll rebase it once this PR lands.

mavam (approved):
@dominiklohmann looked at the individual pieces already. I just looked at the `mmap` factory.

PR #1506, opened 2021-03-29
Switch to 'flat_map' over 'unordered_map' in the meta index

lava:

See commit message.

###  :memo: Checklist

Ideally: Do a benchmark run to ensure no performance regressions.

### DIFF
--- a/libvast/src/system/meta_index.cpp
+++ b/libvast/src/system/meta_index.cpp
@@ -226,7 +226,7 @@ std::vector<uuid> meta_index_state::lookup_impl(const expression& expr) const {
           "{} checked {} partitions for predicate {} and got {} results",
           detail::pretty_type_name(this), synopses.size(), x, result.size());
         // Some calling paths require the result to be sorted.
-        std::sort(result.begin(), result.end());
+        VAST_ASSERT(std::is_sorted(result.begin(), result.end()));
         return result;
       };
       auto extract_expr = detail::overload{
@@ -248,8 +248,7 @@ std::vector<uuid> meta_index_state::lookup_impl(const expression& expr) const {
                 }
               }
             }
-            // Re-establish potentially violated invariant.
-            std::sort(result.begin(), result.end());
+            VAST_ASSERT(std::is_sorted(result.begin(), result.end()));
             return result;
           } else if (lhs.kind == meta_extractor::field) {
             // We don't have to look into the synopses for type queries, just
@@ -278,8 +277,7 @@ std::vector<uuid> meta_index_state::lookup_impl(const expression& expr) const {
                   result.push_back(synopsis.first);
               }
             }
-            // Re-establish potentially violated invariant.
-            std::sort(result.begin(), result.end());
+            VAST_ASSERT(std::is_sorted(result.begin(), result.end()));
             return result;
           }
           VAST_WARN("{} cannot process attribute extractor: {}",
--- a/libvast/src/system/meta_index.cpp
+++ b/libvast/src/system/meta_index.cpp
@@ -31,7 +31,7 @@ namespace vast::system {
 
 size_t meta_index_state::memusage() const {
   size_t result = 0;
-  for (auto& [id, partition_synopsis] : synopses)
+  for (const auto& [id, partition_synopsis] : synopses)
     result += partition_synopsis.memusage();
   return result;
 }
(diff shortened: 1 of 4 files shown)


tobim:
I guess using a map is less bug-prone because we don't need to remember to sort before intersect / union. Can you provide some before / after numbers for the meta index lookup time to make sure this isn't a regression?

JoeLoser:
Does this codebase have something similar to `boost::flat_map` so we get the ordering aspect without it being node-based/slow?

mavam:
> Does this codebase have something similar to boost::flat_map so we get the ordering aspect without it being node-based/slow?

Namespace `detail` is your friend. 🙂 

Check out https://github.com/tenzir/vast/blob/master/libvast/vast/detail/flat_map.hpp.

JoeLoser:
> > Does this codebase have something similar to boost::flat_map so we get the ordering aspect without it being node-based/slow?
> 
> Namespace `detail` is your friend. 🙂
> 
> Check out https://github.com/tenzir/vast/blob/master/libvast/vast/detail/flat_map.hpp.

I'd prefer we use `vast::detail::flat_map` than `std::map` here then.

mavam:
> I'd prefer we use `vast::detail::flat_map` than `std::map` here then.

I'll let @lava illuminate us what guided his choice.

I could imagine that if the workload is write-dominant we're putting more stress on the allocator in `flat_map` than on a tree-based map.

lava:
Using `flat_map` sounds like a good idea, the biggest problem with it is that it doesn't expose an `erase()` method and that the ordering guarantee is only very weakly documented, but these can be fixed. Inserting and erasing will need to reposition most elements, but conventional wisdom is that this is still faster on average than a linked data structure due to caching and prefetching.  Since I have to benchmark anyways for regression testing purposes, I'll probably throw `flat_map` in the mix and pick the fastest of the three.

lava:
Ok, so I finally got around to measuring the performance here. The results were a bit surprising (to me, at least): I created a `{unordered_,flat_,stable_}set` with around 100k random uuids, then copied around 1/16th of all uuids to a vector and afterwards sorted the vector if required:

```
        insert  |  traversal  |  sort  | total
set:   357879488ns 5124310ns     260ns  363004058ns
uset:  253285898ns 2005057ns 5426370ns  260717325ns
fset: 9270310007ns 1115348ns     210ns 9271425565ns
sset: 152230848508ns 1199940ns 5459420ns 152237507868ns

```

* Traversal time for the unordered set is so much faster that its worth it to spend the additional sorting step afterwards. This tradeoff gets worse as more uuids are selected, at 100% the `set` total time becomes ca. an order of magnitude faster than the `unordered_set`
* `flat_set` is again almost an order of magnitude faster to traverse than `unordered_set`, but the `O(n**2)` total insertion time is brutal in this benchmark. (and this is real-world relevant since that scenario happens during startup, where this would introduce a noticeable delay)
* `stable_set` combines the worst of all worlds with `O(n**2)` insertion and `O(n**2)` lookup time.

So based on this I think the optimal choice is the `flat_set` with an additional change to the startup path in VAST to pre-sort the vector and populate the `flat_map` in bulk. During live operations, we only insert single partitions occasionally, so the high insert cost should be less relevant there.

libvast/vast/detail/flat_map.hpp
```diff
@@ -19,7 +19,7 @@ namespace vast::detail {
 
 template <class Key, class T, class Compare>
 struct flat_map_policy {
-  static bool pair_compare(const std::pair<const Key, T>& x, const Key& y) {
+  static bool pair_compare(const std::pair<Key, T>& x, const Key& y) {
```
tobim, line 22: Isn't the key `const` in the `value_type` or a map?

libvast/src/system/meta_index.cpp
```diff
@@ -17,6 +17,7 @@
 #include "vast/expression.hpp"
 #include "vast/fbs/utils.hpp"
 #include "vast/logger.hpp"
+#include "vast/partition_synopsis.hpp"
```
tobim, line 20: nit: I think it's preferable to skip includes that are already pulled in the header of the current source file.

tobim (approved):
With some evidence that `flat_map` provides the best tradeoff for this use case, this looks good now.

libvast/vast/detail/flat_map.hpp
lava: For `std::map` yes, for `detail::vector_map` no:

    // vector_map.hpp
    using value_type = std::pair<Key, T>;

The previous version would actually trigger the copy constructor of `std::pair`:

    /usr/include/c++/10/ext/new_allocator.h:150:4: error: use of deleted function ‘constexpr std::pair<_T1, _T2>::pair(const std::pair<_T1, _T2>&) [with _T1 = const vast::qualified_record_field; _T2 = std::unique_ptr<vast::synopsis>]’

lava: I think it would be a good idea to change the definition of `flat_map` to make a `vector_map` with `const Key` as key type, but I'll leave that to a future refactoring.

PR #1507, opened 2021-03-29
Remove deprecated no-default-schema option

dominiklohmann:

This removes the option entirely after we deprecated it with the last release.

###  :memo: Checklist

File-by-file. The docs were already updated when we deprecated the option.


tobim (approved):
I checked that these are the only places that use this option. Should we add this to the changlog?

dominiklohmann:
> Should we add this to the changlog?

Forgot to push, it's there now.

PR #1508, opened 2021-03-29
Assign a type ID to std::filesystem::path

dominiklohmann:

This makes the required changes for https://github.com/tenzir/vast/pull/1504#discussion_r602731887.

###  :memo: Checklist

n/t

### DIFF
--- a/libvast/vast/system/actors.hpp
+++ b/libvast/vast/system/actors.hpp
@@ -14,11 +14,37 @@
 
 #include <caf/replies_to.hpp>
 
+#include <filesystem>
+
 #define VAST_ADD_TYPE_ID(type) CAF_ADD_TYPE_ID(vast_actors, type)
+
+// NOLINTNEXTLINE(cert-dcl58-cpp)
+namespace std::filesystem {
+
+// TODO: With CAF 0.18, drop this inspector that is—strictly speaking—undefined
+// behavior and replace it with a specialization of caf::inspector_access.
+template <class Inspector>
+typename Inspector::result_type
+inspect(Inspector& f, ::std::filesystem::path& x) {
+  auto str = x.string();
+  if constexpr (std::is_same_v<typename Inspector::result_type, void>) {
+    f(str);
+    if constexpr (Inspector::reads_state)
+      x = {str};
+    return;
+  } else {
+    auto result = f(str);
+    if constexpr (Inspector::reads_state)
+      x = {str};
+    return result;
+  }
+}
+
+} // namespace std::filesystem
 
 namespace vast::system {
 
-/// Helper utility that enables extending typed actor forwartd declarations
+/// Helper utility that enables extending typed actor forward declarations
 /// without including <caf/typed_actor.hpp>.
 template <class... Fs>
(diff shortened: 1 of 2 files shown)


libvast/vast/system/actors.hpp
```diff
@@ -14,11 +14,35 @@
 
 #include <caf/replies_to.hpp>
 
+#include <filesystem>
+
 #define VAST_ADD_TYPE_ID(type) CAF_ADD_TYPE_ID(vast_actors, type)
 
+// NOLINTNEXTLINE(cert-dcl58-cpp)
+namespace std::filesystem {
```
JoeLoser, line 22: Quite ugly that `inspect` needs to be in `std::filesystem` namespace. AFAIK this is in violation of the standard saying this is UB, but if caf requires it...shame on them.

JoeLoser: Lame, but I accept that given that it's treated as a 3rd party lib. When can we upgrade to CAF 0.18? I know you've expressed interest before in it.

dominiklohmann: The migration is a really large amount of work. If nothing urgent comes up, I plan to start working on it sometime this month, but will only spend one day per week on it at most. We will not, however, drop CAF 0.17.6 support for a while, so there'll be a ifdefs for a few months.

tobim (commented):
What about sending using a wrapper for the actor interfaces? That way we can avoid the UB and including `<filesystem>`.
I think we have two options:
```
struct wrapped_path {
  std::filesystem::path p_;
  operator std::filesystem::path() const {
    return p_;
  }

  template <class Inspector>
  friend auto inspect(Inspector& f, wrapped_path& p) {
    return f(caf::meta::type_name("wrapped_path"), p.p_.string());
  }
};
```
or
```
struct wrapped_path : std::filesystem::path {
  template <class Inspector>
  friend auto inspect(Inspector& f, wrapped_path& p) {
    return f(caf::meta::type_name("wrapped_path"), p.string());
  }
};
```

dominiklohmann:
We've experimented with wrapper types before in actor interfaces and that turned out to be very annoying in practice. The errors from compile time actor interface checks are very hard to parse and cost us a lot of time while developing, and both of the proposed solutions change how the type interacts with ADL.

The undefined behavior is something we can drop entirely with CAF 0.18, when we can specialize the inspector access template.

If we want to drop the include in the `actors.hpp` header then I'd rather change the actor interfaces to work on strings rather than paths.

tobim:
> both of the proposed solutions change how the type interacts with ADL.

The idea is to only use the wrapper for sending it over and unwrapping the path immediately in the handler. I wouldn't want to use it for anything else.

> The errors from compile time actor interface checks are very hard to parse and cost us a lot of time while developing

Yes, but changing to `atom::path + std::string` instead will cause the same compilation errors.

dominiklohmann:
As agreed upon in a call with @tobim, I've put a TODO comment above the `inspect` overload that says to replace it once we have CAF 0.18.

tobim (approved):

PR #1509, opened 2021-03-29
 Use expected for chunk::mmap

JoeLoser:

Problem:
- There are multiple ways that `chunk::mmap` can not succeed. In some
  cases, it logs and returns a `nullptr`, and in other cases, just returns
  a `nullptr`. In the former case, this can lead to duplicate logging
  since the callers of `mmap` already log on failure. However, the
  callers lack sufficient information as to exactly which of the calls
  inside `mmap` made it fail.

Solution:
- Change the return type of `chunk::map` to leverage `caf::expected` so
  we can carry the error message back to the callers so it is known
  exactly what failed when trying to perform the mmap operation.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/chunk.cpp
+++ b/libvast/src/chunk.cpp
@@ -53,26 +53,28 @@ chunk_ptr chunk::make(view_type view, deleter_type&& deleter) noexcept {
   return chunk_ptr{new chunk{view, std::move(deleter)}, false};
 }
 
-chunk_ptr chunk::mmap(const std::filesystem::path& filename, size_type size,
-                      size_type offset) {
+caf::expected<chunk_ptr> chunk::mmap(const std::filesystem::path& filename,
+                                     size_type size, size_type offset) {
   // Figure out the file size if not provided.
   if (size == 0) {
     std::error_code err{};
     size = std::filesystem::file_size(filename, err);
-    if (size == static_cast<std::uintmax_t>(-1)) {
-      VAST_ERROR("failed to get file size for filename {}: {}", filename,
-                 err.message());
-      return nullptr;
-    }
+    if (size == static_cast<std::uintmax_t>(-1))
+      return caf::make_error(ec::filesystem_error,
+                             fmt::format("failed to get file size for file"
+                                         "{}: {}",
+                                         filename, err.message()));
   }
   // Open and memory-map the file.
   auto fd = ::open(filename.c_str(), O_RDONLY, 0644);
   if (fd == -1)
-    return {};
+    return caf::make_error(ec::filesystem_error,
+                           fmt::format("failed to open file {}", filename));
   auto map = ::mmap(nullptr, size, PROT_READ, MAP_SHARED, fd, offset);
   ::close(fd);
   if (map == MAP_FAILED)
-    return {};
+    return caf::make_error(ec::filesystem_error,
+                           fmt::format("failed to mmap file {}", filename));
   auto deleter = [=]() noexcept { ::munmap(map, size); };
   return make(map, size, std::move(deleter));
 }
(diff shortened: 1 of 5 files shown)


mavam:
You can rebase this PR now, @JoeLoser.

JoeLoser:
> You can rebase this PR now, @JoeLoser.

Rebased 👍

libvast/src/chunk.cpp
```diff
@@ -53,26 +53,29 @@ chunk_ptr chunk::make(view_type view, deleter_type&& deleter) noexcept {
...
+                                     size_type size, size_type offset) {
   // Figure out the file size if not provided.
   if (size == 0) {
     std::error_code err{};
     size = std::filesystem::file_size(filename, err);
     if (size == static_cast<std::uintmax_t>(-1)) {
-      VAST_ERROR("failed to get file size for filename {}: {}", filename,
-                 err.message());
-      return nullptr;
+      return caf::make_error(ec::filesystem_error,
```
dominiklohmann, line 63: Nit: You can remove the surrounding braces per our style guide.

libvast/src/segment_store.cpp
```diff
@@ -394,16 +394,17 @@ caf::error
 segment_store::register_segment(const std::filesystem::path& filename) {
   auto chk = chunk::mmap(filename);
   if (!chk)
-    return caf::make_error(ec::filesystem_error, "failed to mmap chunk",
-                           filename.string());
+    return caf::make_error(ec::filesystem_error,
+                           fmt::format("failed to mmap chunk {}: {}", filename,
+                                       caf::to_string(chk.error().context())));
```
dominiklohmann, line 399: It suffices to forward the error here.

```suggestion
    return std::move(chk.error());
```

libvast/src/segment_store.cpp
```diff
@@ -427,9 +428,10 @@ caf::expected<segment> segment_store::load_segment(uuid id) const {
              filename);
   auto chk = chunk::mmap(filename);
   if (!chk)
-    return caf::make_error(ec::filesystem_error, "failed to mmap chunk",
-                           filename.string());
-  if (auto segment = segment::make(std::move(chk))) {
+    return caf::make_error(ec::filesystem_error,
+                           fmt::format("failed to mmap chunk {}: {}", filename,
+                                       caf::to_string(chk.error().context())));
```
dominiklohmann, line 433: Same as above: It suffices to forward the error here.

```suggestion
    return std::move(chk.error());
```

libvast/src/system/index.cpp
```diff
@@ -114,9 +114,11 @@ caf::error extract_partition_synopsis(
   // Use blocking operations here since this is part of the startup.
   auto chunk = chunk::mmap(partition_path);
   if (!chunk)
-    return caf::make_error(ec::system_error, "could not mmap partition at "
-                                               + partition_path.string());
-  const auto* partition = fbs::GetPartition(chunk->data());
+    return caf::make_error(
+      ec::filesystem_error,
+      fmt::format("could not mmap partition at {}: {}", partition_path,
+                  caf::to_string(chunk.error().context())));
```
dominiklohmann, line 120:
```suggestion
    return std::move(chunk.error());
```

libvast/vast/chunk.hpp
```diff
@@ -112,8 +112,9 @@ class chunk final : public caf::ref_counted {
   /// @param size The number of bytes to map. If 0, map the entire file.
   /// @param offset Where to start in terms of number of bytes from the start.
   /// @returns A chunk pointer or `nullptr` on failure.
```
dominiklohmann, line 114:
```suggestion
  /// @returns A chunk pointer or an error on failure. The returned chunk
  /// pointer is never `nullptr`.
```

dominiklohmann (changes requested):
Just some very minor requests, other than that this looks really good. I double-checked that you caught all usages of `chunk::mmap`.

JoeLoser:
> Just some very minor requests, other than that this looks really good. I double-checked that you caught all usages of `chunk::mmap`.

Thanks for the review. I believe I've addressed all of the feedback comments; please re-review.

PR #1523, opened 2021-04-02
Clean up filesystem error handling

JoeLoser:

Problem:
- Checking for whether the `error_code` is set or not is superflous as
  the `bool` returned from `std::filesystem::exists` will be set to the
  correct value. As such, checking just the bool is sufficient.

Solution:
- Remove checking of `err` as the test for whether the exist check
  failed or not.

###  :memo: Checklist

File-by-file.


mavam (approved):
This simplification makes sense in both cases; program logic remains intact.

(However, there might be other functions where a filesystem error cannot be assumed to be a valid complement of an existing directory.)

PR #1524, opened 2021-04-04
Use std::filesystem::path for pcap input/trace paths

JoeLoser:

Problem:
- `pcap` uses `std::string` as its data type for the input and trace,
  which conptually represent paths or `stdin/stdout`. Using
  `std::string` doesn't convey this very well.

Solution:
- Change `input` and `trace` types to be `std::filesystem::path`.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/format/pcap.cpp
+++ b/libvast/src/format/pcap.cpp
@@ -82,7 +82,8 @@ reader::reader(const caf::settings& options, std::unique_ptr<std::istream>)
   std::string category = "vast.import.pcap";
   if (auto interface = get_if<std::string>(&options, category + ".interface"))
     interface_ = *interface;
-  input_ = get_or(options, "vast.import.read", vast::defaults::import::read);
+  input_ = std::filesystem::path{
+    get_or(options, "vast.import.read", vast::defaults::import::read)};
   cutoff_ = get_or(options, category + ".cutoff", defaults_t::cutoff);
   max_flows_ = get_or(options, category + ".max-flows", defaults_t::max_flows);
   max_age_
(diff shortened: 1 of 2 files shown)


libvast/src/format/pcap.cpp
```diff
@@ -82,7 +82,8 @@ reader::reader(const caf::settings& options, std::unique_ptr<std::istream>)
   std::string category = "vast.import.pcap";
   if (auto interface = get_if<std::string>(&options, category + ".interface"))
     interface_ = *interface;
-  input_ = get_or(options, "vast.import.read", vast::defaults::import::read);
+  input_ = std::filesystem::path{
+    get_or(options, "vast.import.read", vast::defaults::import::read)};
```
mavam, line 86: Question for @dominiklohmann: would it be okay to include `<filesystem>` in `defaults.hpp`? Or would it blow compile time and we should do it this way?

mavam (commented):
LGTM, just one workflow question directed to @dominiklohmann, our build time guardian.

libvast/src/format/pcap.cpp
dominiklohmann: No for two reasons:
1. We should avoid including that file for compile-time reasons.
2. `vast.import.read` is not necessarily a path.

dominiklohmann (commented):
I'm not a fan of this change. The option `vast import --read=<option> <format>` does not necessarily reflect a path, so this change is semantically wrong.

@tobim can you be the tiebreaker on this one?

JoeLoser:
> I'm not a fan of this change. The option `vast import --read=<option> <format>` does not necessarily reflect a path, so this change is semantically wrong.
> 
> @tobim can you be the tiebreaker on this one?

I'll just close the MR then -- this isn't a change I feel too strongly about. It was just a drive-by thing while in the neighborhood of code.

PR #1527, opened 2021-04-06
Align plugin and library output names

dominiklohmann:

Plugins are able to set their name to anything they want, e.g., `libexample.so` can set its name to `notexample`. This change forces the names to be aligned, which will allow for reliably checking static plugins by name in the future.

Note that this is a *breaking change* for some configurations: When specifying dynamic plugins under `vast.plugins`, the `lib` prefix is now stripped. To load a plugin from any of the usual or explicitly specified directories, use the key `<name>` instead of `lib<name>`.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,14 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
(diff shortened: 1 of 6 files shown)


mavam (approved):
I didn't test this locally, but all the changes look good to me.

PR #1528, opened 2021-04-06
Ignore static plugins when specified in config

dominiklohmann:

This commit changes VAST to ignore plugins specified in the `vast.yaml` when static versions of them are already loaded.

Aside from being the correct thing to do, this fixes an issue with the integration test suite where plugin integration tests specify their name in a `vast.yaml` file used for their tests.

###  :memo: Checklist

File-by-file. Note that this is based on #1527 and cannot work without that change.

### DIFF
--- a/vast/vast.cpp
+++ b/vast/vast.cpp
@@ -50,32 +50,32 @@ int main(int argc, char** argv) {
   }
   // Load plugins.
   auto loaded_plugin_paths = std::vector<std::filesystem::path>{};
-  auto plugin_files_or_names
+  auto plugin_paths_or_names
     = caf::get_or(cfg, "vast.plugins", std::vector<std::string>{});
 #ifdef VAST_ENABLED_PLUGINS
   // If plugins are enabled at compile time to always be loaded, add them here
   // if they were not already configured as part of vast.plugins.
   for (auto&& plugin_names : std::vector<std::string>{VAST_ENABLED_PLUGINS})
-    if (std::none_of(plugin_files_or_names.begin(), plugin_files_or_names.end(),
+    if (std::none_of(plugin_paths_or_names.begin(), plugin_paths_or_names.end(),
                      [&](auto&& x) { return x == plugin_names; }))
-      plugin_files_or_names.push_back(std::move(plugin_names));
+      plugin_paths_or_names.push_back(std::move(plugin_names));
 #endif
   auto& plugins = plugins::get();
   // Check if any of the specified plugins is already loaded as a static
   // plugin, and silently remove it from the list of specified plugins if
   // that's the case.
   for (const auto& plugin : plugins) {
-    if (auto it = std::find_if(plugin_files_or_names.begin(),
-                               plugin_files_or_names.end(),
-                               [&](const auto& plugin_file_or_name) {
-                                 return plugin->name() == plugin_file_or_name;
+    if (auto it = std::find_if(plugin_paths_or_names.begin(),
+                               plugin_paths_or_names.end(),
+                               [&](const auto& plugin_path_or_name) {
+                                 return plugin->name() == plugin_path_or_name;
                                });
-        it != plugin_files_or_names.end())
-      plugin_files_or_names.erase(it);
+        it != plugin_paths_or_names.end())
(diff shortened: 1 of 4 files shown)


vast/vast.cpp
```diff
@@ -50,18 +50,31 @@ int main(int argc, char** argv) {
   }
   // Load plugins.
   auto loaded_plugin_paths = std::vector<std::filesystem::path>{};
-  auto plugin_files
+  auto plugin_files_or_names
```
tobim, line 53: Slightly more exact:
```suggestion
  auto plugin_paths_or_names
```

vast/vast.cpp
```diff
@@ -50,18 +50,31 @@ int main(int argc, char** argv) {
...
+  for (auto&& plugin_names : std::vector<std::string>{VAST_ENABLED_PLUGINS})
+    if (std::none_of(plugin_files_or_names.begin(), plugin_files_or_names.end(),
+                     [&](auto&& x) { return x == plugin_names; }))
+      plugin_files_or_names.push_back(std::move(plugin_names));
 #endif
   auto& plugins = plugins::get();
-  for (const auto& plugin_file : plugin_files) {
+  // Check if any of the specified plugins is already loaded as a static
+  // plugin, and silently remove it from the list of specified plugins if
+  // that's the case.
```
tobim, line 66:
```suggestion
  // that's the case.
  // TODO: Consider sorting the inputs and using `std::set_difference` instead
  // of this custom loop.
```

vast/vast.cpp
```diff
@@ -50,18 +50,31 @@ int main(int argc, char** argv) {
...
+    if (auto it = std::find_if(plugin_files_or_names.begin(),
+                               plugin_files_or_names.end(),
+                               [&](const auto& plugin_file_or_name) {
+                                 return plugin->name() == plugin_file_or_name;
+                               });
+        it != plugin_files_or_names.end())
+      plugin_files_or_names.erase(it);
+  }
+  // Load the specified dynamic plugins.
+  for (const auto& plugin_file : plugin_files_or_names) {
```
tobim, line 77:
```suggestion
  for (const auto& plugin_path_or_name : plugin_path_or_names) {
```

tobim (changes requested):
Looks good in principle, just nits and a comment on the choice of algorithm.

vast/vast.cpp
dominiklohmann: Good call. Will rename tomorrow.

vast/vast.cpp
dominiklohmann: Can do, but I'm not sure it's worth it over the trivial implementation. This is far from a critical piece of code and I doubt there'll ever be a large number of plugins loaded.

dominiklohmann: I pushed it, but this is really damn ugly. Please take a look.

mavam: Note that we have `detail/set_operations.hpp` that make these operations much easier to use.

dominiklohmann: Nothing in here helps for this specific use case, especially because of the custom comparator.

tobim: It seems this got a little out of hand while I was away. I originally asked to include a comment with the intent to inform the reader at a glance of what this statement does semantically.
I think this version is not even correct though, because the inputs need to be sorted for `std::set_difference`.

dominiklohmann: I intend to revert the commit before merging. Just wanted to show you that the comment is nonsensical.

tobim: > Just wanted to show you that the comment is nonsensical.

I'm not convinced of that. The loop code is still doing a set difference, and saying so in a comment is very helpful for the reader.

dominiklohmann: > Consider [...] using `std::set_difference` instead of this custom loop

tobim: `The following loop implements a custom set difference instead of calling the stl implementation because we want to preserve the order as much as possible.`

Or something like that?

dominiklohmann: I pushed something. Please check back.

tobim (approved):

PR #1530, opened 2021-04-07
Fix start command detection for spdlog

dominiklohmann:

When configuring spdlog we do different things for client and server. We distinguish between the two by checking whether the current subcommand we're running is `start`.

However, this returned a false positive for a plugin that had a command `<plugin> start`. This commit changes to checking whether the full name of the command we're running is `start`.

###  :memo: Checklist

n/t

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- 🐞 Custom commands from plugins ending in `start` no longer try to write to
+  the server instead of the client log file.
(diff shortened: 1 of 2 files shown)


tobim (approved):
My understanding of `invocation` allows me to confirm that this change is correct without testing it locally.

lava (approved):
Looks good, I didn't test it myself but I assume you have already checked that the `full_name` is just `start` and not `vast start` in the server case?

dominiklohmann:
> I assume you have already checked that the `full_name` is just `start` and not `vast start` in the server case?

Yes. I verified that `vast start` still writes to the `vast.db/server.log` file.

PR #1533, opened 2021-04-08
Avoid shutdown when config dirs are not readable

dominiklohmann:

This fixes an overzealous check for configuration files.

###  :memo: Checklist

n/t

### DIFF
--- a/libvast/src/system/configuration.cpp
+++ b/libvast/src/system/configuration.cpp
@@ -103,16 +103,16 @@ caf::error configuration::parse(int argc, char** argv) {
     std::error_code err{};
     const auto exists_conf_yaml = std::filesystem::exists(conf_yaml, err);
     const auto exists_conf_yml = std::filesystem::exists(conf_yml, err);
-    if (err)
-      return caf::make_error(ec::filesystem_error,
-                             fmt::format("failed to check if vast yaml config "
-                                         "files existed from directory {}: {}",
-                                         dir, err.message()));
+    if (err) {
+      VAST_WARN("failed to check if vast.yaml file exists in {}: {}", dir,
+                err.message());
+      return caf::none;
+    }
     if (exists_conf_yaml && exists_conf_yml)
       return caf::make_error(
         ec::invalid_configuration,
         "detected both 'vast.yaml' and 'vast.yml' files in " + dir.string());
-    else if (exists_conf_yaml)
+    if (exists_conf_yaml)
       config_files.emplace_back(std::move(conf_yaml));
     else if (exists_conf_yml)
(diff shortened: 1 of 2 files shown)


libvast/src/system/configuration.cpp
```diff
@@ -103,16 +103,16 @@ caf::error configuration::parse(int argc, char** argv) {
...
     const auto exists_conf_yaml = std::filesystem::exists(conf_yaml, err);
     const auto exists_conf_yml = std::filesystem::exists(conf_yml, err);
-    if (err)
-      return caf::make_error(ec::filesystem_error,
-                             fmt::format("failed to check if vast yaml config "
-                                         "files existed from directory {}: {}",
-                                         dir, err.message()));
+    if (err) {
+      VAST_VERBOSE("failed to check if vast.yaml file exists in {}: {}", dir,
+                   err.message());
```
tobim, line 108: I would assume that this is normally not intended.
```suggestion
      VAST_WARN("failed to check if vast.yaml file exists in {}: {}", dir,
                err.message());
```

tobim (approved):
Please change the message severity. Feel free to merge after that.

libvast/src/system/configuration.cpp
dominiklohmann: I intentionally used verbose over warn because this can happen in sandboxed environments and is not usually worth warning about. @0snap what do you think?

tobim: What if `/etc/vast` is accidentally not readable for the `vast` user?

dominiklohmann: Fair point. Will change, then merge after CI is done.

libvast/src/system/configuration.cpp
```diff
@@ -103,16 +103,16 @@ caf::error configuration::parse(int argc, char** argv) {
...
     const auto exists_conf_yaml = std::filesystem::exists(conf_yaml, err);
     const auto exists_conf_yml = std::filesystem::exists(conf_yml, err);
-    if (err)
-      return caf::make_error(ec::filesystem_error,
-                             fmt::format("failed to check if vast yaml config "
-                                         "files existed from directory {}: {}",
-                                         dir, err.message()));
+    if (err) {
+      VAST_WARN("failed to check if vast.yaml file exists in {}: {}", dir,
+                err.message());
```
mavam, line 108: I know I'm late to the party, but I'm missing a bit of user hand-holding in the communication: what should the user do now? Is everything still okay?

dominiklohmann: The error message says what's wrong with quite some detail, so I think this suffices.

PR #1534, opened 2021-04-08
Handle alias types properly for CSV

dominiklohmann:

This fixes a crash with nested aliases in the CSV reader. To reproduce, replace the definition of `port` in `base.schema` like this:

```
type proxy = count
type port = proxy
```

Then import CSV data, e.g., 

```
gunzip -c vast/integration/data/csv/argus-M57-10k-pkts.csv.gz | vast -N import csv
```

###  :memo: Checklist

Commit-by-commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- 🐞 The CSV reader no longer crashes when encountering nested type aliases.
+  [#1534](https://github.com/tenzir/vast/pull/1534)
+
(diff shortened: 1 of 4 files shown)


lava (approved):
Ideally this could use a unit test to ensure we'll not regress on this. But clearly not crashing is better than crashing, so I'm also not opposed to merging it immediately.

dominiklohmann:
> Ideally this could use a unit test to ensure we'll not regress on this. But clearly not crashing is better than crashing, so I'm also not opposed to merging it immediately.

We have this open story about possibly consolidating the data conversions to a given type (ch23655), and I think it makes more sense to add tests in the scope of that story.

I'll still try and write a minimal unit test for this one, but I'm unsure how

PR #1535, opened 2021-04-08
Drop the timezone delta in the ISO8601 time parser

tobim:

Up until now, we shifted time values according to the timezone modifier in the input. This is suboptimal because we're normally interested in normalized timestamps, and UTC is the natural representation for that.

###  :memo: Checklist

Compare the change of the integration test expectations with the original data from the inputs.


CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- ⚠️ The parser for ISO8601 timestamps now ignores timezone offsets. The old
+  behavior of shifting the timestamp according to the offset was problematic,
+  because it was not possible to know what zone the timestamp was relative to.
+  From now on all timestamps in VAST are in the UTC timezone.
+  [#1535](https://github.com/tenzir/vast/pull/1535)
```
lava, line 20: This seems a bit problematic: a time stamp like `15:30+01:00` unambiguously refers to 14:30 UTC, not 15:30 UTC.

So either I'm misunderstanding the changelog entry, or it looks like we'd be breaking ISO8601 compatibility with this change?

dominiklohmann: I don't understand the change either. Isn't the mismatch on the export side rather than on the import side?

tobim: @lava thanks for pointing this out. I must have misinterpreted the ISO definition.

tobim: I guess I can close this then :sweat_smile:

PR #1537, opened 2021-04-08
Fix the timezone shift to UTC for ISO8601 dates

tobim:

One has to subtract positive offsets and add negative ones to arrive at the correct UTC timestamp value. We did that the wrong way up to now.

###  :memo: Checklist

Manually verify that the test expectations are now correct.


lava (approved):
Crazy to think that we had such a bug for so long without noticing. But of course if the tests are testing the wrong thing, that doesn't help :D

PR #1539, opened 2021-04-08
Fix dependency summary for interface libraries

dominiklohmann:

The issue is two-fold:
- The INTERFACE_INCLUDE_DIRECTORIES property may contain multiple paths and even generator expressions
- The LOCATION propery does not work on INTERFACE libraries.

The solution is to create a stub library and get that libraries location, while also making use of a deprecated CMake feature ([CMP0026](https://cmake.org/cmake/help/latest/policy/CMP0026.html)) to get the LOCATION for build targets without in a context where generator expressions may not be used.

In summary: This fixes the dependency summary when using the bundled tsl::robin-map.

###  :memo: Checklist

Read the code, try it out locally. This is absolutely low-priority, please don't waste as much time reviewing this as I spent trying to fix this bug.


dominiklohmann:
Closing this, it introduces another set of issues that I don't want to deal with.

PR #1543, opened 2021-04-12
Fix ODR violation in custom fallback_formatter

dominiklohmann:

This fixes an ODR violation that we've seen with our static binaries.

###  :memo: Checklist

Check that the error does not occur any longer for static builds. I've requested @tobim for review as he has a local Nix setup.

### DIFF
--- a/libvast/vast/detail/logger_formatters.hpp
+++ b/libvast/vast/detail/logger_formatters.hpp
@@ -80,21 +80,25 @@ struct fmt::internal::fallback_formatter<
   T, fmt::format_context::char_type,
   std::enable_if_t<std::conjunction_v<
     std::negation<vast::detail::is_formattable<T>>,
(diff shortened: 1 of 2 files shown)


tobim:
I can already confirm the problem, but the proposed change seems to introduce another.

dominiklohmann:
> but the proposed change seems to introduce another.

Yes, still working on it. I think this may be a bug in our `vast::detail::has_ostream_operator<T>`, which is broken just like any `is_complete<T>` check.

dominiklohmann:
This is ready for review now. No more ODR violation per the last static binary build.

tobim (approved):
We seem to be hitting https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96291 now, this should get resolved when we get gcc 10.3 with the next nixpkgs bump. The original bug is resolved, so I'm accepting this.

PR #1544, opened 2021-04-12
Factor common functionality in component plugin

dominiklohmann:

It turns out that just having a handle to the NODE is already a valuable plugin. Such a plugin can spawn an actor and register itself with the NODE. That's it.

Other plugins, like the analyzer plugin, will then inherit from this plugin.

The component plugin only gives you the node actor and requires that you implement the status subscriber.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/plugin.hpp
+++ b/libvast/vast/plugin.hpp
@@ -100,17 +100,53 @@ public:
   [[nodiscard]] virtual const char* name() const = 0;
 };
 
+// -- component plugin --------------------------------------------------------
+
+/// A base class for plugins that spawn components in the NODE.
+/// @relates plugin
+class component_plugin : public virtual plugin {
+public:
+  /// Creates an actor as a component in the NODE.
+  /// @param node A pointer to the NODE actor handle.
+  /// @returns The actor handle to the NODE component.
+  virtual system::component_plugin_actor
+  make_component(system::node_actor::pointer node) const = 0;
+};
+
 // -- analyzer plugin ----------------------------------------------------------
 
 /// A base class for plugins that hook into the input stream.
-/// @relates plugin
-class analyzer_plugin : public virtual plugin {
+/// @relates component_plugin
+class analyzer_plugin : public virtual component_plugin {
 public:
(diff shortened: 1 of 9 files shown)


libvast/vast/plugin.hpp
```diff
@@ -100,17 +100,59 @@ class plugin {
...
 
 /// A base class for plugins that hook into the input stream.
-/// @relates plugin
-class analyzer_plugin : public virtual plugin {
+/// @relates component_plugin
+class analyzer_plugin : public virtual component_plugin {
 public:
+  /// Gets the ANALYZER actor spawned by the plugin, if it is still running.
+  inline system::analyzer_plugin_actor
+  get_analyzer(system::node_actor::pointer node) const {
```
mavam, line 124: I'm starting to see a lot of `get_*`. Our coding style does not encourage this. What speaks against using just `analyzer(node)` here?

libvast/vast/plugin.hpp
```diff
@@ -100,17 +100,59 @@ class plugin {
...
+  /// @note It is guaranteed that this function is not called while the ANALYZER
+  /// is still running.
   virtual system::analyzer_plugin_actor
   make_analyzer(system::node_actor::pointer node) const = 0;
+
+private:
+  /// A weak handle to the spawned actor handle.
+  mutable caf::weak_actor_ptr weak_handle_ = {};
+
+  /// Indicates that the ANALYZER was spawned at least once.
```
mavam, line 154: Can you expand the comment for why we need this flag?

mavam (approved):
The code looks good to me. I didn't test this locally, though, but it's my understanding your are using this already.

dominiklohmann:
> it's my understanding your are using this already.

Yes. I've ported our proprietary analyzer plugins for testing this. Only thing that really changed is the ownership model: The analyzers are now owned by the NODE directly rather than the IMPORTER.

libvast/vast/plugin.hpp
dominiklohmann: Fine by me. Will rename.

PR #1547, opened 2021-04-13
Fix a possible deadlock in component plugins

dominiklohmann:

Component plugins were spawned before all other components were ready necessarily. This introduced a possible deadlock when trying to get other components from the NODE as part of the initialization of a component plugins actor.

The solution is to simply spawn components from component plugins last.

###  :memo: Checklist

File- by-file. 

Sorry for introducing this one, I overlooked it when developing #1544.


lava (approved):
Looks reasonable; I didn't test but it sounds like dependabot already did that in this case. (also, the PR is just moving code around, so the chance of breakage should be low)

PR #1551, opened 2021-04-16
Use timeouts when controlling a remote node

dominiklohmann:

This commit adds timeouts to the node control functions. We end up deadlocking if the request times out when connecting to a remote node.

This will aid in debugging the deadlock we're experiencing when an analyzer plugin uses `get_node_components` in its `make_analyzer` function.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/system/node_control.cpp
+++ b/libvast/src/system/node_control.cpp
@@ -17,11 +17,13 @@
 namespace vast::system {
 
 caf::expected<caf::actor>
-spawn_at_node(caf::scoped_actor& self, node_actor node, invocation inv) {
+spawn_at_node(caf::scoped_actor& self, const node_actor& node, invocation inv) {
   caf::expected<caf::actor> result = caf::no_error;
-  self->request(node, caf::infinite, atom::spawn_v, std::move(inv))
-    .receive([&](caf::actor actor) { result = std::move(actor); },
-             [&](caf::error err) { result = std::move(err); });
+  self
+    ->request(node, defaults::system::initial_request_timeout, atom::spawn_v,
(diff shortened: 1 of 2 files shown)


libvast/src/system/node_control.cpp
```diff
@@ -17,11 +17,13 @@
...
+spawn_at_node(caf::scoped_actor& self, const node_actor& node, invocation inv) {
   caf::expected<caf::actor> result = caf::no_error;
-  self->request(node, caf::infinite, atom::spawn_v, std::move(inv))
-    .receive([&](caf::actor actor) { result = std::move(actor); },
-             [&](caf::error err) { result = std::move(err); });
+  self
+    ->request(node, defaults::system::initial_request_timeout, atom::spawn_v,
+              std::move(inv))
+    .receive([&](caf::actor& actor) { result = std::move(actor); },
+             [&](caf::error& err) { result = std::move(err); });
```
tobim, line 26: I'm a bit uncomfortable with moving variables out of the frame of a calling function. If CAF doesn't move the arguments into the callback there must be some reason for it. And even if they aren't accessed anymore now, that might change in the future.

dominiklohmann: See here for why this is the correct thing to do: https://github.com/actor-framework/actor-framework/discussions/1220

mavam: Moving variables to the parent scope is common practice for blocking actors. We've been doing this for a while, actually.

tobim (approved):
I trust CI to make sure that this doesn't break in unexpected ways.

PR #1557, opened 2021-04-19
Make it possible to run VAST without user configs

dominiklohmann:

The new option `--disable-default-config-dirs` that may only be supplied on the command line disables the loading of user and system configuration, schema, and plugin directories. We use this option internally when generating VAST's man page and when running integration tests.

###  :memo: Checklist

Commit-by-commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- ⚠️ The new option `--disable-default-config-dirs` disables the loading of user
+  and system configuration, schema, and plugin directories. We use this option
+  internally when generating VAST's man page and when running integration tests.
+  [#1557](https://github.com/tenzir/vast/pull/1557)
+
 - ⚡️ Plugins must define a separate entrypoint in their build scaffolding using
   the argument `ENTRYPOINT` to the CMake function `VASTRegisterPlugin`. If only
   a single value is given to the argument `SOURCES`, it is interpreted as the
--- a/libvast/src/detail/load_plugin.cpp
+++ b/libvast/src/detail/load_plugin.cpp
@@ -26,26 +26,30 @@ namespace {
 stable_set<std::filesystem::path>
 get_plugin_dirs(const caf::actor_system_config& cfg) {
   stable_set<std::filesystem::path> result;
+  const auto disable_default_config_dirs
+    = caf::get_or(cfg, "vast.disable-default-config-dirs", false);
   if (const char* vast_plugin_directories = std::getenv("VAST_PLUGIN_DIRS"))
     for (auto&& path : detail::split(vast_plugin_directories, ":"))
       result.insert({path});
-#if !VAST_ENABLE_RELOCATABLE_INSTALLATIONS
-  result.insert(std::filesystem::path{VAST_LIBDIR} / "vast" / "plugins");
-#endif
-  // FIXME: we technically should not use "lib" relative to the parent, because
-  // it may be lib64 or something else. CMAKE_INSTALL_LIBDIR is probably the
-  // best choice.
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ This changelog documents all notable user-facing changes of VAST.
 
 ## Unreleased
 
+- ⚠️ The new option `--disable-configuration` disables the loading opf user
```
lava, line 16: s/opf/of/

libvast/src/system/configuration.cpp
```diff
@@ -61,6 +61,14 @@ caf::error configuration::parse(int argc, char** argv) {
     for (const auto& [old, new_] : replacements)
       if (option == old)
         option = new_;
+  // Detect when running with --disable-configuration, and remove the option
+  // from the command line.
```
lava, line 65: Why do we need this block? Is it because CAF doesn't support "flag"-like command line options?

lava (dismissed):
Having the option makes sense and is desired, but the current behavior was a bit surprising for me:

```
$ cat config.yaml 
vast:
  endpoint: "localhost:6000"

$ ./bin/vast --disable-configuration --config=config.yaml start
[13:56:51.138] loaded configuration file: "config.yaml"
[13:56:51.265] VAST node is listening on localhost:6000
```

Also I'd intuitively expect the environment variables to be ignored as well with this, since they're just another channel to pass configuration.

libvast/src/system/configuration.cpp
dominiklohmann: The other options only get parsed after we have the command structure, which is too late to affect the plugin configuration. Hence the manual parsing here.

dominiklohmann:
> Also I'd intuitively expect the environment variables to be ignored as well with this, since they're just another channel to pass configuration.

The environment variables are just another workaround because of the same ordering issue we always have when loading plugins: Plugins can modify the command structure, and the command structure is required to parse the command line. Hence why the two variables must continue working, or we break integration tests for external plugins.

lava:
> Similarly, --config still needs to work. Otherwise, the integration tests cannot run at all.

Fine with me, but then we should adapt the option name & description accordingly. Currently the description says that it will also disable user configuration. Something like `--disable-custom-directory-lookup`?

dominiklohmann:
I've reworded to `--bare-mode` per our discussion in Slack.

libvast/src/detail/load_plugin.cpp
```diff
@@ -26,26 +26,30 @@ namespace {
 stable_set<std::filesystem::path>
 get_plugin_dirs(const caf::actor_system_config& cfg) {
   stable_set<std::filesystem::path> result;
+  const auto disable_configuration
```
lava, line 29: s/disable_configuration/bare_mode/

libvast/src/system/application.cpp
```diff
@@ -487,6 +487,9 @@ auto make_root_command(std::string_view path) {
   auto ob
     = opts("?vast")
         .add<std::string>("config", "path to a configuration file")
+        .add<bool>("bare-mode", "disable user and system configuration, schema "
+                                "and plugin  directories lookup (this may only "
+                                "be used on the command line)")
```
lava, line 492: Do we even need to add this here, since its not handled by caf anyways? Anyways there's a double space in "plugin  directories", and the option does not disable user and system configuration.

lava (approved):
Looking good now, module a few nits.

libvast/src/system/application.cpp
dominiklohmann: Yes, otherwise it doesn't show up in `vast help`.

dominiklohmann:
Renamed again to `--disable-default-config-dirs` after further discussion in Slack.

PR #1517, opened 2021-04-01
Implement transforms for VAST

lava:
###  :notebook_with_decorative_cover: Description

This adds an MVP implementation of "transforms" to VAST.

###  :memo: Checklist

- [ ] All user-facing changes have changelog entries.
- [ ] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [x] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

Look at the commit overview first: Roughly the first half consists of related bugfixes and refactorings I encountered while writing the main feature, the transforms implementation itself is split by category.

PR #1563, opened 2021-04-20
Multiple node shutdown fixes

tobim:

- Don't bail when receiving a down message during teardown
- We've been terminating non-essential components after the main
  ones, but now we do it before.
- We don't try to terminate remote actors any more.
- The shutdown sequence used to treat labels as types, which means
  the actor of the label wouldn't be terminated at all and keeping
  the process alive.

###  :memo: Checklist

By commit.

### DIFF
--- a/changelog/unreleased/bug-fixes/1563.md
+++ b/changelog/unreleased/bug-fixes/1563.md
@@ -1,0 +1,4 @@
+The shutdown logic contained a bug that would make the node fail to terminate
+in case a plugin actor is registered at said node.
+
+A race condition in the shutdown logic that caused an assertion was fixed.
--- a/libvast/src/system/node.cpp
+++ b/libvast/src/system/node.cpp
@@ -560,7 +560,7 @@ node_state::spawn_command(const invocation& inv,
 
 node_actor::behavior_type
 node(node_actor::stateful_pointer<node_state> self, std::string name,
-     const std::filesystem::path& dir,
+     std::filesystem::path dir,
      std::chrono::milliseconds shutdown_grace_period) {
   self->state.name = std::move(name);
   self->state.dir = std::move(dir);
@@ -575,13 +575,15 @@ node(node_actor::stateful_pointer<node_state> self, std::string name,
   // Remove monitored components.
   self->set_down_handler([=](const caf::down_msg& msg) {
     VAST_DEBUG("{} got DOWN from {}", self, msg.source);
-    auto actor = caf::actor_cast<caf::actor>(msg.source);
-    auto component = self->state.registry.remove(actor);
-    VAST_ASSERT(component); // All components are in the registry.
-    // Terminate if a singleton dies.
-    if (is_singleton(component->type)) {
-      VAST_ERROR("{} terminates after DOWN from {}", self, component->type);
-      self->send_exit(self, caf::exit_reason::user_shutdown);
+    if (!self->state.tearing_down) {
+      auto actor = caf::actor_cast<caf::actor>(msg.source);
+      auto component = self->state.registry.remove(actor);
+      VAST_ASSERT(component);
+      // Terminate if a singleton dies.
+      if (is_singleton(component->type)) {
+        VAST_ERROR("{} terminates after DOWN from {}", self, component->type);
+        self->send_exit(self, caf::exit_reason::user_shutdown);
+      }
     }
   });
   // Terminate deterministically on shutdown.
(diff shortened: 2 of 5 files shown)


libvast/src/system/node.cpp
```diff
@@ -467,7 +467,11 @@ node_state::spawn_command(const invocation& inv,
   VAST_TRACE_SCOPE("{}", inv);
   using std::begin;
   using std::end;
-  auto self = this_node;
+  auto* self = this_node;
+  if (self->state.tearing_down)
+    return caf::make_message(caf::make_error(caf::sec::request_receiver_down,
```
dominiklohmann, line 472: Doesn't this error code have a special meaning inside CAF? We should be careful using these.

Also this is not an error per se, right? I think we should just return `caf::none` or `caf::no_error` here to indicate that nothing was spawned.

libvast/src/system/node.cpp
```diff
@@ -575,68 +579,76 @@ node(node_actor::stateful_pointer<node_state> self, std::string name,
...
-    std::vector<std::string> remaining;
-    remaining.reserve(registry.components().size());
-    for (const auto& [label, _] : registry.components())
-      remaining.push_back(label);
-    for (const auto& label : remaining)
-      schedule_teardown(label);
-    // Finally, bring down the filesystem.
-    scheduled_for_teardown.push_back(std::move(*filesystem));
+    });
+    std::vector<caf::actor> teardown_aux;
```
dominiklohmann, line 607: I propose changing this algorithm ever so slightly.

1. Remove the core components from the registry in the correct order, putting them into a vector of components (e.g., named `core_components`). Demonitor them.
2. Create a vector of components by removing the remaining components from the registry (e.g., named `aux_components`). Demonitor them.
3. Remove remote actors from the `aux_components` vector.
4. Do shutdown as currently implemented: `aux_components` in parallel, `core_components` sequential after that.

The logic is the same except for the additional removal from the registry, but it probably reads a lot better.

libvast/src/system/node.cpp
```diff
@@ -575,68 +579,76 @@ node(node_actor::stateful_pointer<node_state> self, std::string name,
...
+    for (const auto& [_, comp] : self->state.registry.components()) {
+      self->demonitor(comp.actor);
+      // Ignore remote actors.
+      if (comp.actor->node() != self->node())
+        continue;
+      // Core components are terminated last.
+      if (comp.type == "archive" || comp.type == "filesystem"
+          || comp.type == "importer" || comp.type == "index")
```
dominiklohmann, line 615: Please create some form of `is_core_component(std::string_view)` function.

Also, should the accountant be considered a core component? What about the transformer? Please add a comment to #1517 if necessary.

Maybe "core component" is the wrong name and it should be called "component requires sequential shutdown"?

dominiklohmann (changes requested):
Mostly looking good. Ran this locally and it reduced the shutdown time with plugins a lot. Your reasoning why this fixes the rare assertion failure in CI seems sound to me.

This PR needs two changelog entries:
- `bug-fixes` for the rare assertion failure fix
- `features` for the parallel shutdown of auxiliary actors

libvast/src/system/node.cpp
```diff
@@ -604,37 +603,38 @@ node(node_actor::stateful_pointer<node_state> self, std::string name,
...
+    // Core components are terminated in a second stage, we remove them from the
+    // registry upfront and deal with them later.
+    std::vector<caf::actor> core_components;
+    // The components listed here need to be terminated in sequential order.
+    // The importer needs to shut down first because it might still have
+    // buffered data. The index uses the archive for querying. The filesystem
+    // is needed by all others for the persisting logic.
+    for (const char* name : {"importer", "index", "archive", "filesystem"}) {
+      if (auto comp = registry.remove(name))
+        core_components.push_back(comp->actor);
```
dominiklohmann, line 616: This is missing a `demonitor` call. Also the `else` should probably print an error, or we should even assert here.

tobim: The `demonitor` happens in `core_shutdown_sequence`.
I didn't not include an `else` branch because the exit handler can also be invoked because one of the core components went down for some reason.

dominiklohmann: Sane reasoning.

tobim:
> This PR needs two changelog entries:
> 
>     * `bug-fixes` for the rare assertion failure fix
   -> done.
> 
>     * `features` for the parallel shutdown of auxiliary actors

I doubt this will really be noticeable under normal circumstances. That change was mostly done to speed things up in case on of the components is not responsive.

changelog/unreleased/bug-fixes/1563.md
```diff
@@ -0,0 +1,4 @@
+The shutdown logic contained a bug that would make the node fail to terminate
+in case a plugin actor is registered at said node.
+
+A race condition in the shutdown logic that caused an assertion was fixed.
```
dominiklohmann, line 4: Every paragraph is a changelog entry of its own, I doubt this is what you wanted to achieve here.

tobim: These are 2 separate Bugs that are both fixed in this PR.

dominiklohmann (approved):

PR #1565, opened 2021-04-22
Disable some clang-tidy warnings

lava:
* cppcoreguidelines-macro-usage: Only tells us to 'consider' using
   constexpr, but we tend to do that anyways so in our codebase I've
   only seen false positives so far.

 * llvm-header-guard: Doesn't apply since we use '#pragma once'.

###  :memo: Checklist


.clang-tidy
```diff
@@ -11,9 +11,11 @@ Checks: >
   -cert-err58-cpp,
   -cppcoreguidelines-avoid-magic-numbers,
   -cppcoreguidelines-avoid-non-const-global-variables,
+  -cppcoreguidelines-macro-usage
```
dominiklohmann, line 14: Missing trailing comma

dominiklohmann (approved):

PR #1570, opened 2021-04-23
Fix building VAST within a shallow git tree

dominiklohmann:

This change makes it so we warn instead of fail on git-describe errors, falling back to the specified fallback version. This helps in situation where git and the source tree are available, but git-describe fails, e.g., when working from within a shallow clone.

###  :memo: Checklist

n/t


lava (approved):
Nice!

PR #1579, opened 2021-04-25
Synchronize calls to getenv

dominiklohmann:

As indicated by clang-tidy, getenv and friends are not thread-safe. This adds wrapper functions that synchronize the calls to them.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/detail/env.cpp
+++ b/libvast/src/detail/env.cpp
@@ -8,6 +8,10 @@
 
 #include "vast/detail/env.hpp"
 
+#include "vast/error.hpp"
+
+#include <fmt/format.h>
+
 #include <cstdlib>
 #include <mutex>
 
@@ -15,11 +19,13 @@ namespace vast::detail {
 
 namespace {
 
+// A mutex for locking calls to functions that mutate `environ`. Global to this
+// translation unit.
 auto env_mutex = std::mutex{};
 
 } // namespace
 
-std::optional<std::string> env(const char* var) {
+std::optional<std::string> locked_getenv(const char* var) {
   auto lock = std::scoped_lock{env_mutex};
   // NOLINTNEXTLINE(concurrency-mt-unsafe)
   if (const char* result = ::getenv(var))
@@ -27,10 +33,14 @@ std::optional<std::string> env(const char* var) {
   return {};
 }
 
-bool unset_env(const char* var) {
+caf::expected<void> locked_unsetenv(const char* var) {
   auto lock = std::scoped_lock{env_mutex};
   // NOLINTNEXTLINE(concurrency-mt-unsafe)
-  return ::unsetenv(var) == 0;
+  if (::unsetenv(var) == 0)
+    return {};
+  return caf::make_error( //
+    ec::system_error,
+    fmt::format("failed in unsetenv(3): {}", ::strerror(errno)));
 }
 
 } // namespace vast::detail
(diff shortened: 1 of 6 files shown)


libvast/vast/detail/env.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+#include <string>
+
+namespace vast::detail {
+
+/// A thread-safe wrapper around `::getenv`.
+/// @param var The environment variable.
+/// @returns The environment variables contents, or an empty string if the
+/// variable is unset.
+std::string env(const char* var);
```
mavam, line 19: Returning `std::optional<std::string>` would make it possible to distinguish between an unset and a set variable that is empty.

dominiklohmann: Done.

libvast/vast/detail/env.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+/// A thread-safe wrapper around `::getenv`.
+/// @param var The environment variable.
+/// @returns The copied environment variables contents, or `std::nullopt`.
+[[nodiscard]] std::optional<std::string> env(const char* var);
+
+/// A thread-safe wrapper around `::unsetenv`.
+/// @param var The environment variable.
+/// @returns True on success.
+[[nodiscard]] bool unset_env(const char* var);
```
lava, line 24: Nodiscard seems a bit overzealous here, or if you want to force error handling you should at least return a `caf::error` and/or the proper errno.

libvast/vast/detail/env.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+#include <optional>
+#include <string>
+
+namespace vast::detail {
+
+/// A thread-safe wrapper around `::getenv`.
+/// @param var The environment variable.
+/// @returns The copied environment variables contents, or `std::nullopt`.
+[[nodiscard]] std::optional<std::string> env(const char* var);
```
lava, line 19: I'd propose we stay close to the wrapped functions and call these `getenv/unsetenv`? Or maybe even something like `getenv_mt`/`locked_getenv`/`blocking_getenv`/`threadsafe_getenv` to make it clear at the callsite why we don't just use the standard library functions.

libvast/src/detail/env.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "vast/detail/env.hpp"
+
+#include <cstdlib>
+#include <mutex>
+
+namespace vast::detail {
+
+namespace {
+
+auto env_mutex = std::mutex{};
```
lava, line 18: What do you think in general about addind a `s_` prefix for the name of static data? For these small functions its obvious, but in general i think its nice when access to global data really stands out while reading the source code. In particular once classes get thrown into the mix, where a non-locally declared variable might be either a data member or a global variable.

lava (approved):
Looks good to me except for some bikeshedding below.

libvast/vast/detail/env.hpp
dominiklohmann: I like the `locked_` prefix, will change.

libvast/src/detail/env.cpp
dominiklohmann: I will add a comment above the variable that makes it clear that this is global to this translation unit.

libvast/vast/detail/env.hpp
dominiklohmann: I like the latter. Will do.

PR #1580, opened 2021-04-25
Use proper full install dirs for system config

dominiklohmann:

Before this change, the default build configuration would try to load schemas from `etc/vast/schema` instead of `/usr/local/etc/vast/schema` on macOS, and instead of `/etc/vast/schema` on Linux systems if `VAST_INSTALL_DATADIR` was unset or `CMAKE_INSTALL_DATADIR` set to a relative path instead of an absolute one (it should be relative).

The solution here is to use the full paths instead.

Additionally, there's no need to make the paths configurable, since that ever only affected where VAST looked for the config/schemas/plugins but not where we installed them.

###  :memo: Checklist

File-by-file.

Note that this is based on #1579 to avoid the merge conflict. Other than that the PRs are independent of each other.


libvast/src/schema.cpp
```diff
@@ -238,11 +238,11 @@ get_schema_dirs(const caf::actor_system_config& cfg,
       VAST_ERROR("{} failed to get program path", __func__);
   }
   if (!disable_default_config_dirs) {
-    result.insert(std::filesystem::path{VAST_SYSCONFDIR} / "vast" / "schema");
-    if (auto xdg_config_home = detail::locked_getenv("XDG_CONFIG_HOME"))
+    result.insert(std::filesystem::path{VAST_CONFIGDIR} / "schema");
+    if (auto xdg_config_home = detail::env("XDG_CONFIG_HOME"))
```
tobim, line 242: `detail::locked_getenv` is reverted to `detail::env` in this commit. Probably unintentional?

tobim (commented):
Testing rn.

libvast/src/schema.cpp
dominiklohmann: Yes, rebase mistake. Will fix in a second.

tobim (approved):
Seems to be working as expected.

PR #1581, opened 2021-04-25
Use platform-independent CMAKE_DL_LIBS over -ldl

dominiklohmann:

Make use of the platform-independent `CMAKE_DL_LIBS` over the platform-specific `dl` to link against the libraries containing `dlopen`, `dlsym`, `dlerror`, and `dlclose`.

This fixes an issue with older CMake versions on Linux, where we errored because `dl` may not be linked against the interface library `libvast_internal` that we use for propagating compiler flags. Instead, we now link against `libvast` publicly when building a shared library.

###  :memo: Checklist

File-by-file.


mavam (approved):
If CI enjoys this PR, we can proceed.

PR #1582, opened 2021-04-26
Add a script that automates release prep work

dominiklohmann:

This adds a script that automates all the release prep work. I will demonstrate its usage shortly.

###  :memo: Checklist

n/t

### DIFF
--- a/scripts/create-release
+++ b/scripts/create-release
@@ -1,0 +1,81 @@
+#! /usr/bin/env zsh -euo pipefail
+
+# Check if exactly one argument was passed: the new version.
+if (( $# != 1 )); then
+  >&2 echo "Error: expected new version as only argument"
+  exit 1
+fi
+
+# Check if all required tools are available.
+for binary in 'cmake' 'gh' 'rsync' 'git' 'perl'; do
+  if ! command -v "${binary}" 2>&1 >/dev/null; then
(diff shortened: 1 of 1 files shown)


scripts/create-release
```diff
@@ -0,0 +1,64 @@
+#! /usr/bin/env zsh -euo pipefail
+
+if (( $# != 1 )); then
+  >&2 echo "Error: expected new version as only argument"
+  exit 1
+fi
+
+for binary in 'gh' 'rsync' 'git' 'perl'; do
```
tobim, line 8:
```suggestion
for binary in 'cmake' 'gh' 'rsync' 'git' 'perl'; do
```

tobim (changes requested):
Please add comments to this.

Why does it rename the changlog directory for release candidates? Imho this should only happen for the final release.

I don't understand the reason this uses rsync over `git mv`.

dominiklohmann:
> I don't understand the reason this uses rsync over `git mv`

`git mv` does not merge directories.

> Why does it rename the changlog directory for release candidates? Imho this should only happen for the final release.

We've done it like this for rc2 as well, so if anything we can change that for the next release. I'd consider that a separate story.

tobim (approved):
I trust that you used that script to create the PR for rc3.

PR #1587, opened 2021-04-26
Patch Arrow APT sources URLs

dominiklohmann:
###  :notebook_with_decorative_cover: Description

This is an intermediate fix until the ASF completed the migration to Artifactory. Once that is done, we need to remove the `sed` and rename `apache-arrow-archive-keyring` to `apache-arrow-apt-source`.

###  :memo: Checklist

- [x] All user-facing changes have changelog entries.
- [x] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [x] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

n/t

PR #1588, opened 2021-04-26
Fix the streaming setup for analyzer plugins

tobim:

We should not call plugin related code in the setup of core components. So the analyzer stream setup is done later now.

###  :memo: Checklist


libvast/src/plugin.cpp
```diff
@@ -61,6 +63,16 @@ system::analyzer_plugin_actor analyzer_plugin::analyzer(
...
     return {};
   auto handle = make_analyzer(node);
+  auto [importer] = node->state.registry.find<system::importer_actor>();
+  VAST_ASSERT(importer);
+  node
+    ->request(importer, caf::infinite,
+              static_cast<system::stream_sink_actor<table_slice>>(handle))
+    .then([](caf::outbound_stream_slot<vast::table_slice>&) {},
+          [&](const caf::error& error) {
+            VAST_ERROR("Failed to connect {} to the importer: {}", handle,
```
dominiklohmann, line 73:
```suggestion
            VAST_ERROR("failed to connect {} to the importer: {}", handle,
```

Or even this:

```suggestion
            VAST_ERROR("failed to connect analyzer {} to the importer: {}", name(),
```

libvast/src/plugin.cpp
```diff
@@ -61,6 +63,16 @@ system::analyzer_plugin_actor analyzer_plugin::analyzer(
   if (spawned_once_ || !node)
     return {};
   auto handle = make_analyzer(node);
+  auto [importer] = node->state.registry.find<system::importer_actor>();
+  VAST_ASSERT(importer);
+  node
+    ->request(importer, caf::infinite,
+              static_cast<system::stream_sink_actor<table_slice>>(handle))
+    .then([](caf::outbound_stream_slot<vast::table_slice>&) {},
```
dominiklohmann, line 71:
```suggestion
    .then([](const caf::outbound_stream_slot<table_slice>&) {},
```

dominiklohmann (approved):
Some nits. 

To fix CI this must be rebased onto master now that #1587 is merged.

Approving as-is, but please fix the mentioned things _and_ most importantly add a changelog entry, as this is a bug fix.

PR #1589, opened 2021-04-27
Remove duplicate uses of [[nodiscard]]

dominiklohmann:


lava (approved):
Let's discard the no-discard.

PR #1592, opened 2021-04-27
Test building external plugins in CI

dominiklohmann:

Before this change, we only tested building shared and static plugin libraries alongside VAST in CI. This sets up a CI workflow that downloads a previously created VAST artifact, and builds and tests plugins against that. For now, this is only done for the example plugin.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -1,4 +1,4 @@
-name: "VAST"
+name: VAST
 on:
   push:
     branches:
@@ -16,24 +16,10 @@ on:
   release:
     types: published
 
-env:
-  CMAKE_GENERATOR: Ninja
-  CMAKE_MAKE_PROGRAM: ninja
-  CMAKE_C_COMPILER_LAUNCHER: ccache
-  CMAKE_CXX_COMPILER_LAUNCHER: ccache
-  CCACHE_HASH_DIR: true
-  CCACHE_UNIFY: true
-  CCACHE_SLOPPINESS: 'file_macro,time_macros'
-  CCACHE_DIR: '${{ github.workspace }}/.ccache'
-  CCACHE_COMPRESS: true
-  CCACHE_COMPRESSLEVEL: 9
-  CCACHE_ABSSTDERR: true
-  BUILD_DIR: build
 jobs:
   build-debian:
     name: Debian - ${{ matrix.build.compiler }} ${{ matrix.configure.tag }}
     runs-on: ubuntu-20.04
-    # run everything inside a Debian container on an ubuntu-20.04 host
     container: debian:buster-backports
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/vast.yaml
```diff
@@ -1,4 +1,4 @@
-name: "VAST"
+name: 'VAST'
```
0snap, line 1: There's no need to quote strings behind the colon.

.github/workflows/vast.yaml
```diff
@@ -513,3 +522,94 @@ jobs:
           # for builds of the latest release.
           asset_name: "${{ steps.configure_env.outputs.publish_name }}.tar.gz"
           asset_content_type: application/gzip
+
+  plugins:
+    runs-on: ubuntu-20.04
+    # run everything inside a Debian container on an ubuntu-20.04 host
```
0snap, line 528: this comment should be moved down by one line.

.github/workflows/vast.yaml
```diff
@@ -513,3 +522,94 @@ jobs:
...
+          cmake --build "$BUILD_DIR" --target all --parallel
+      - name: Run Unit Tests
+        run: |
+          cmake --build "$BUILD_DIR" --target test
+      - name: Run Integration Tests
+        id: integration_tests
+        run: |
+          echo "::set-output name=running=true"
+          cmake --build "$BUILD_DIR" --target integration
+          echo "::set-output name=running=false"
```
0snap, line 605: What do we need the `running` output for?

.github/workflows/vast.yaml
```diff
@@ -513,3 +522,94 @@ jobs:
...
+      - name: Upload Integration Test Logs on Failure
+        if: failure()
+        uses: actions/upload-artifact@v2
+        with:
+          name: 'vast-${{ matrix.plugin.target }}-integration-test'
+          path: '${{ env.BUILD_DIR }}/vast-${{ matrix.plugin.target }}-integration-test'
+          if-no-files-found: ignore
+      - name: Install
+        run: |
+          cmake --install "$BUILD_DIR" --prefix "$INSTALL_DIR"
```
0snap, line 615: Why don't we upload the build artifacts to GitHub at the end? The example plugin binary artifact might be a good source to get folks started with running VAST with a plugin (we could add a docs section).

0snap (changes requested):
Looking good, nothing major from my side. I left some questions in the file.

.github/workflows/vast.yaml
dominiklohmann: They are not very portable and other than VAST the libraries are not meant to be built against. I intentionally decided against that.

.github/workflows/vast.yaml
dominiklohmann: Leftover from a previous test, will remove.

.github/workflows/vast.yaml
dominiklohmann: I opted to remove it, it's a pretty meaningless comment.

0snap (approved):
LGTM :rocket:

PR #1593, opened 2021-04-28
Prefix plugin library output names with vast-plugin-

dominiklohmann:

To avoid confusion between the PCAP plugin and libpcap, which both have a library file named `libpcap.so`, we now prefix the plugin library output names with `vast-plugin-`. The PCAP plugin library file is now named `libvast-plugin-pcap.so`.

This is a breaking change: The output name of plugins changed, so plugins specified with a full path in `vast.plugins` must be adapted accordingly.

###  :memo: Checklist

File-by-file.


tobim (approved):
As far as I can see there shouldn't be any more places that need to be touched.

lava (approved):
I tested this locally and verified that it works as expected.

PR #1597, opened 2021-04-28
Allow for plugins to specify only headers as sources

dominiklohmann:

In case a plugin specifies only header files in SOURCES we did not correctly create a stub source file. ~Now we always generate one instead of relying on any heuristics.~ This rewrites the base target for plugins not to have this dependency.

###  :memo: Checklist

I verified locally that this works with a plugin I'm working on. It's an edge case, but I think we should handle this gracefully instead of letting the plugin developer run into a hard-to-understand linker error.

### DIFF
--- a/cmake/VASTRegisterPlugin.cmake
+++ b/cmake/VASTRegisterPlugin.cmake
@@ -126,19 +126,17 @@ function (VASTRegisterPlugin)
     endif ()
   endif ()
 
-  if (NOT PLUGIN_SOURCES)
-    # Create a stub source file with an identfier if no sources except for the
-    # entrypoint exist.
-    string(MAKE_C_IDENTIFIER "${PLUGIN_TARGET}" plugin_identifier)
-    file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/stub.cpp"
-         "void vast_plugin_${plugin_identifier}_stub() {}")
-    set(PLUGIN_SOURCES "${CMAKE_CURRENT_BINARY_DIR}/stub.cpp")
-  else ()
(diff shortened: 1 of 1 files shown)


cmake/VASTRegisterPlugin.cmake
```diff
@@ -126,23 +134,16 @@ function (VASTRegisterPlugin)
...
+  if ("${PLUGIN_ENTRYPOINT}" IN_LIST PLUGIN_SOURCES)
+    list(REMOVE_ITEM PLUGIN_SPURCES "${PLUGIN_ENTRYPOINT}")
   endif ()
 
-  # Create a static library target for our plugin _without_ the entrypoint.
-  add_library(${PLUGIN_TARGET} STATIC ${PLUGIN_SOURCES})
+  # Create an object library target for our plugin _without_ the entrypoint.
+  file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/stub.h" "")
+  list(APPEND PLUGIN_SOURCES "${CMAKE_CURRENT_BINARY_DIR}/stub.h")
+  add_library(${PLUGIN_TARGET} OBJECT ${PLUGIN_SOURCES})
```
tobim, line 146: nit: It seems that `${PLUGIN_TARGET}-static` is not used in case we're building a shared plugin?

dominiklohmann: It is for the unit test binary.

tobim (approved):
I trust that creating object libraries with only headers as source arguments works.

dominiklohmann:
> I trust that creating object libraries with only headers as source arguments works.

It's also tested by CI for the PCAP plugin, which doesn't specify any sources.

PR #1598, opened 2021-04-28
Remove dependency between directory and plugin name

dominiklohmann:

For plugins built alongside VAST there was an undocumented and since recent changes unnecessary restriction that plugin name and directory name containing the plugin's top-level `CMakeLists.txt` file must be the exact same. This commit removes that restriction, and as an added bonus allows the CMake variable `VAST_PLUGINS` to be specified using paths relative to the repository root instead of always requiring absolute paths.

###  :memo: Checklist

File-by-file.


tobim (approved):
Works nicely!

PR #1600, opened 2021-04-29
Set CMAKE_EXPORT_COMPILE_COMMANDS in project scope

dominiklohmann:

A function introduces a new scope in CMake, so for the variable to be picked up
in project-wide scope we need to set it using the `PARENT_SCOPE` option.

###  :memo: Checklist

@tobim test locally.


tobim (approved):
I verified that it works again with this change.

PR #1603, opened 2021-04-29
Hide built-in plugins from version command

dominiklohmann:

We support built-in plugins by using the `VAST_REGISTER_PLUGIN` macro from within a source file that is part of either the `vast` or the `libvast` targets.  These plugins are always loaded and cannot be disabled. This channge hides such plugins from the version output, as they are really not plugins but simply use the plugin framework to add functionality.

The idea for this came up when reviewing the transforms functionality, which has some transformation as such built-in plugins directly as part of `libvast`.

###  :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/plugin.hpp
+++ b/libvast/vast/plugin.hpp
@@ -232,18 +232,27 @@ public:
 /// @relates plugin
 class plugin_ptr final {
 public:
+  /// The display policy of a plugin.
+  enum class visible {
+    yes, ///< Show the plugin in the output of the version command.
+    no,  ///< Hide the plugin in the output of the version command.
+  };
+
   /// Load a plugin from the specified library filename.
   /// @param filename The filename that's passed to 'dlopen'.
   /// @param cfg The actor system config to register type IDs with.
+  /// @param visible The display policy of the plugin.
   static caf::expected<plugin_ptr>
-  make(const char* filename, caf::actor_system_config& cfg) noexcept;
+  make(const char* filename, caf::actor_system_config& cfg,
+       enum visible visible) noexcept;
 
   /// Take ownership of an already loaded plugin.
   /// @param instance The plugin instance.
   /// @param deleter A deleter for the plugin instance.
   /// @param version The version of the plugin.
+  /// @param visible The display policy of the plugin.
   static plugin_ptr make(plugin* instance, void (*deleter)(plugin*),
-                         plugin_version version) noexcept;
+                         plugin_version version, enum visible visible) noexcept;
 
   /// Unload a plugin and its required resources.
   ~plugin_ptr() noexcept;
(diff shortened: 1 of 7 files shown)


mavam:
With the plugin information hidden, how would you figure out whether your binary supports a given feature/plugin or not?

dominiklohmann:
> With the plugin information hidden, how would you figure out whether your binary supports a given feature/plugin or not?

To clarify, "built-in plugins" are not really plugins. They are just static part of the VAST binary or libvast library that are accessible via the plugin API and essentially work like statically linked bundled plugins. Since they're not optional there's no need to check for this, ever.

libvast/vast/plugin.hpp
```diff
@@ -232,18 +232,27 @@ class writer_plugin : public virtual plugin {
 /// @relates plugin
 class plugin_ptr final {
 public:
+  /// The display policy of a plugin.
+  enum class visible {
+    yes, ///< Show the plugin in the output of the version command.
+    no,  ///< Hide the plugin in the output of the version command.
+  };
```
lava, line 239: It feels a bit weird to have the plugin dictate how `vast version` behaves. What do you think about an enum like this

```
enum class plugin_type {
  dynamic,
  static,
  native
}
```

and then letting `vast version` decide itself how it renders each type of plugin?

libvast/vast/plugin.hpp
```diff
@@ -304,6 +317,14 @@ class plugin_ptr final {
 
 #if defined(VAST_ENABLE_STATIC_PLUGINS_INTERNAL)
 
+#  if VAST_ENABLE_STATIC_PLUGINS_INTERNAL == 1
```
lava, line 320: It feels like there should be a cleaner solution to this than having a magic number.

How about adding an additional defines like `VAST_OPTIONAL_PLUGIN`/`VAST_NATIVE_PLUGIN` that are set in the respective build contexts?

Or alternatively we at least could turn the numbers into named constants like
```
#define PLUGIN_VISIBILITY_DYNAMIC 1
```
etc.

lava (commented):
While this approach would work, it feels a little bit too ad-hoc, especially since this not just some header in `detail/` but part of our public plugin API. Let's see if we can find a cleaner solution for this.

libvast/vast/plugin.hpp
dominiklohmann: I made this change, please check back.

libvast/vast/plugin.hpp
dominiklohmann: The logic is a bit more readable now and does not rely on magic constants any more.

lava (approved):
Looks good now. There's still an argument to be made that `ENABLE` in `VAST_ENABLE_STATIC_PLUGINS` is the wrong choice of words (the current semantics are more like `VAST_CURRENT_PLUGIN_TYPE`), but that's just an implementation detail in our CMakeLists so it doesn't need to hold up this PR; we can still change it anytime later if we find the motivation.

PR #1604, opened 2021-04-29
Convert integer to a concrete type

tobim:

This replaces the integer type alias with a struct definition. It is a first step and feasibility analysis for converting the c++ side of the builtin types, i.e. all members of `vast::data`.

Doing so will protect us from implicit conversions between types that have similar underlying representations, and makes it possible to use the same representation for multiple types.

###  :memo: Checklist

By commit.
**EDIT**: This is mostly mechanical, but take a closer look at the parseable and printable changes in the first commit.

### DIFF
--- a/examples/plugins/example/example.cpp
+++ b/examples/plugins/example/example.cpp
@@ -74,8 +74,8 @@ example(example_actor::stateful_pointer<example_actor_state> self) {
       for (auto& [key, value] : config) {
         if (key == "max-events") {
           if (auto max_events = caf::get_if<integer>(&value)) {
-            VAST_VERBOSE("{} sets max-events to {}", self, *max_events);
-            self->state.max_events = *max_events;
+            VAST_VERBOSE("{} sets max-events to {}", self, max_events->value);
+            self->state.max_events = max_events->value;
           }
         }
       }
--- a/examples/plugins/example/example.cpp
+++ b/examples/plugins/example/example.cpp
@@ -6,6 +6,7 @@
 // SPDX-FileCopyrightText: (c) 2021 The VAST Contributors
 // SPDX-License-Identifier: BSD-3-Clause
 
+#include "vast/concept/printable/vast/integer.hpp"
 #include "vast/data.hpp"
 #include "vast/error.hpp"
 #include "vast/logger.hpp"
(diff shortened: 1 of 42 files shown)


examples/plugins/example/example.cpp
```diff
@@ -74,8 +74,8 @@ example(example_actor::stateful_pointer<example_actor_state> self) {
       for (auto& [key, value] : config) {
         if (key == "max-events") {
           if (auto max_events = caf::get_if<integer>(&value)) {
-            VAST_VERBOSE("{} sets max-events to {}", self, *max_events);
-            self->state.max_events = *max_events;
+            VAST_VERBOSE("{} sets max-events to {}", self, max_events->value);
+            self->state.max_events = max_events->value;
```
dominiklohmann, line 78: Should probably just add a formatter specialization for it:

```cpp
template <>
struct fmt::formatter<integer> : fmt::formatter<int64_t> {
  template <class FormatContext>
  auto format(integer x, FormatContext& ctx) {
    return fmt::formatter<int64_t>::format(x.value);
  }
};

Comment (dominiklohmann) [reply examples/plugins/example/example.cpp:78]: Actually, this is just missing a printable include for the new integer type.

File: libvast/test/arrow_table_slice.cpp [hunk 2]
```diff
@@ -74,7 +74,7 @@ enumeration operator"" _e(unsigned long long int x) {
 }
 
 integer operator"" _i(unsigned long long int x) {
-  return static_cast<integer>(x);
```
dominiklohmann, line 77: Nit: This should be `detail::narrow_cast`.

libvast/vast/data/integer.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+  template <class Inspector>
+  friend typename Inspector::result_type inspect(Inspector& f, integer& x) {
+    return f(caf::meta::type_name("vast.integer"), x.value);
+  }
+};
+
+/// @relates integer
+template <class Hasher>
+void hash_append(Hasher& h, const integer& x) {
```
dominiklohmann, line 47: This should be a friend for faster compile times and better ADL.

libvast/vast/system/actor_factory.hpp
```diff
@@ -0,0 +1,75 @@
...
+#pragma once
+
+#include "vast/fwd.hpp"
+
+#include "vast/system/actors.hpp"
+
+namespace vast::system {
+
+template <class Type>
+struct factory_traits {
```
dominiklohmann, line 18: This whole file should likely not be in here.

dominiklohmann (changes requested):
Looking really good except for the few things I commented on that need to be changed.

libvast/vast/system/actor_factory.hpp
tobim: Yes, I should be more careful with `git commit -a`.

dominiklohmann (approved):
LGTM, leaving the rest to CI here.

PR #1606, opened 2021-04-29
Adapt .clang-format for clang-format 12

dominiklohmann:

This was generated by:

```bash
clang-format -dump-config > tmp
mv -f tmp .clang-format
```

Additionally, I changed the value of `SpaceAfterCStyleCast` to `false`, because that wrongly adds spaces in `(*this)(args...)`, which looks really odd.

###  :memo: Checklist

n/t

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,84 +1,189 @@
 ---
+Language:        Cpp
 AccessModifierOffset: -2
 AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: None
+AlignConsecutiveAssignments: None
+AlignConsecutiveBitFields: None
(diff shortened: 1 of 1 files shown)


lava (approved):
It makes sense to me to have this as explicit as possible.

I'm wondering though if we should remove the Qt/Boost/ObjectiveC/Java/etc.-related settings, since they are irrelevant to VAST.

dominiklohmann:
> I'm wondering though if we should remove the Qt/Boost/ObjectiveC/Java/etc.-related settings, since they are irrelevant to VAST.

I like leaving them in so the diff is as small as possible when doing this again for clang-format 13. It doesn't hurt to have the default values in there for now.

PR #1609, opened 2021-04-29
Add overload with message for VAST_ASSERT

lava:

This allows us to write new assertions with an optional message describing the failure.

###  :memo: Checklist


dominiklohmann:
Does it make sense to go through all assertions and if there exists a comment with reasoning right above it to use that as the message?

dominiklohmann (approved):
Approving as-is. Let's add messages whenever we see fit.

PR #1610, opened 2021-04-29
Fix clang-format configuration for older clang-format versions

dominiklohmann:

Apparently clang-format hard-errors when its config file contains unknown values for existing keys, so this is a second attempt at #1606 that just updates the one key that was causing issues.

###  :memo: Checklist

:shrug:


lava (approved):
We overlooked during the last review that the clang-format CI check isn't really running unless we also touch a c++ file, even though the check status will show as green.

This is just reverting an error and we verified that the new option is harmless in that regard, so approving again.

PR #1611, opened 2021-04-29
Add a timeout option to the export command

dominiklohmann:

When running automated VAST exports, we rarely want an infinite timeout.  This adds `vast export --timeout=<duration>` and `vast export --error-on-timeout` options that allow for setting a timeout to the export command. ~If `error-on-timeout` is enabled, cancelled~ Cancelled exports due to a hit timeout result in a non-zero exit code.

###  :memo: Checklist

File-by-file. Test locally.


changelog/unreleased/features/1611.md
```diff
@@ -0,0 +1,4 @@
+The new options `vast export --timeout=<duration>` and `vast export
+--error-on-timeout` allow for setting a timeout for VAST queries. If
+`error-on-timeout` is enabled, cancelled exports result in a non-zero
+exit code.
```
tobim, line 4: If the timeout gets hit then the exit code should ALWAYS be non-zero. The option `error-on-timeout` makes no sense to me.

Otherwise the functionality seems to work and the code looks ok.

tobim (changes requested):

changelog/unreleased/features/1611.md
dominiklohmann: I changed it accordingly, and also fixed the symlink to `vast.yaml.example` from `vast/integration/vast.yaml.example`, which we seemed to have accidentally converted into a hard copy instead.

tobim (approved):
Good feature!

PR #1613, opened 2021-04-30
Remove unflatten operation from type and data

dominiklohmann:

Turns out we did not need this for a while now, and it was dead code outside of tests. So let's just remove it—flattening is now a one-way operation.

###  :memo: Checklist

This just removes dead code, so just check if CI passes I guess. `git grep unflatten` shows no further uses of the word in the code base.


tobim (approved):
Glad that we don't need this.

PR #1614, opened 2021-04-30
Fix bundled robin-map dependency summary

dominiklohmann:


mavam (approved):
I verified that bundled robin-map paths are now displayed correctly.

PR #1616, opened 2021-04-30
Fail if plugin configuration is invalid

dominiklohmann:

Ran into this and did not expect VAST to just start with a default-initialized plugin. Let's abort when the configuration is invalid.

###  :memo: Checklist

It's a very simple change, so just read it and confirm that this is what we want.


mavam (commented):
I like the hard fail much better.

tobim (approved):

PR #1618, opened 2021-04-30
Error when initializing a plugin fails

dominiklohmann:

There's no point in `plugin::initialize()` returning an error if it's never checked. So let's actually check and refuse to start if any plugin fails to initialize.

###  :memo: Checklist

The change is really simple, so looking at the code should suffice.


mavam (approved):
Didn't test, but the diff looks good.

PR #1620, opened 2021-04-30
Cancel previous workflow runs on push

lava:

I've initially restricted this to the long-running PRs that build VAST.

###  :memo: Checklist

### DIFF
--- a/.github/workflows/static-binary.yaml
+++ b/.github/workflows/static-binary.yaml
@@ -30,6 +30,12 @@ jobs:
     env:
       BUILD_DIR: build
     steps:
+    - name: Cancel Previous Runs
+      if: github.event_name == 'pull_request'
+      uses: styfle/cancel-workflow-action@0.9.0
+      with:
+        access_token: ${{ github.token }}
+
     - name: Checkout
       uses: actions/checkout@v2
       with:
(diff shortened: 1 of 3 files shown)


dominiklohmann (commented):
Wouldn't it make a lot more sense to go with the `workflow_run` approach mentioned in the action's README? https://github.com/styfle/cancel-workflow-action#advanced-pull-requests-from-forks

```yaml
name: Cancel Previous Runs
on:
  workflow_run:
    workflows:
      - VAST
      - VAST Docker
      - VAST Static
    types:
      - requested
jobs:
  cancel:
    runs-on: ubuntu-latest
    steps:
    - uses: styfle/cancel-workflow-action@0.9.0
      with:
        workflow_id: ${{ github.event.workflow.id }}
```

lava:
@dominiklohmann Currently I think less than 1% of PRs are triggered from a fork repository, so in general I'm not sure that we gain much by handling this case.

That said, it doesn't make things too much more complicated so I wouldn't mind changing it to a separate workflow, but I'm not sure how to integrate the logic that we only want to abort workflow runs on PRs but not on master pushes with this approach.

dominiklohmann:
I did not think of the `push` case, that's a fair point. Let's keep this as-is then.

dominiklohmann (approved):

PR #1622, opened 2021-05-03
Fix the datagram source

dominiklohmann:

Every so often, we break the DATAGRAM SOURCE actor. This happens because we cannot make the behavior-returning `datagram_source` function returns a typed broker behavior, since that is absolutely broken in CAF 0.17.6 which we're still stuck on, and so we miss changing it when we change the interface of the SOURCE actor. But, as it turns out, we can still define the typed actor and typed broker interfaces, and simply unbox the underlying untyped behaviors before returning them.

Additionally, this aligns the interfaces again so the datagram source works once more.

###  :memo: Checklist

Read code. Test locally with netcat or similar.


lava (approved):
Looks good, this should solve 95% of the problem.

PR #1623, opened 2021-05-03
Bloom filter: caf::optional -> std::optional

JoeLoser:

Problem:
- Bloom filter uses `caf::optional` which is an unncessary dependency
  given that `std::optional` exists and can be readily used.

Solution:
- Replace uses of `caf::optional` with `std::optional`. Similarly,
  replace `caf::none` with `std::nullopt`.
- Add support for formatting `std::optional` with `fmt`.

Note:
- Other parts of the codebase will be converted to use `std::optional`
  in lieu of `caf::optional` in follow-up MRs.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast_test/vast/test/test.hpp
+++ b/libvast_test/vast/test/test.hpp
@@ -143,6 +143,27 @@ struct less_equal_compare {
 
 namespace vast::test {
 
+template <class T>
+T unbox(std::optional<T>&& x) {
+  if (!x)
+    CAF_FAIL("x == none");
+  return std::move(*x);
+}
+
+template <class T>
+T unbox(std::optional<T>& x) {
+  if (!x)
+    CAF_FAIL("x == none");
+  return std::move(*x);
+}
+
+template <class T>
+T unbox(const std::optional<T>& x) {
+  if (!x)
+    CAF_FAIL("x == none");
+  return *x;
+}
+
 // Holds global configuration options passed on the command line after the
 // special -- delimiter.
 extern std::set<std::string> config;
--- a/libvast/src/bloom_filter_parameters.cpp
+++ b/libvast/src/bloom_filter_parameters.cpp
@@ -16,19 +16,20 @@
 
 #include <cmath>
 #include <cstddef>
+#include <optional>
 
 namespace vast {
 
-caf::optional<bloom_filter_parameters> evaluate(bloom_filter_parameters xs) {
+std::optional<bloom_filter_parameters> evaluate(bloom_filter_parameters xs) {
   // Check basic invariants first.
   if (xs.m && *xs.m <= 0)
-    return caf::none;
+    return {};
   if (xs.n && *xs.n <= 0)
-    return caf::none;
+    return {};
   if (xs.k && *xs.k <= 0)
-    return caf::none;
+    return {};
   if (xs.p && (*xs.p < 0 || *xs.p > 1))
-    return caf::none;
+    return {};
   // Test if we can compute the missing parameters.
   static const double ln2 = std::log(2.0);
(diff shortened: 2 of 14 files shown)


libvast_test/vast/test/test.hpp
```diff
@@ -143,6 +143,13 @@ struct less_equal_compare {
 
 namespace vast::test {
 
+template <class T>
+T unbox(std::optional<T> x) {
+  if (!x)
+    CAF_FAIL("x == none");
+  return std::move(*x);
+}
```
dominiklohmann, line 151: Not that performance matters much in unit tests, but there should probably be three overloads for this: Two that take either a `std::optional<T>& x` and `std::optional<T>&& x` that move the result out, and one that takes a `const std::optional<T>& x` that does not.

dominiklohmann (approved):
You're back! 🤩

This looks great. Just have one minor request for `vast::test::unbox(...)`, then we can merge this imo. This is 100% covered by CI.

See also https://github.com/actor-framework/actor-framework/issues/1236: `caf::optional` will likely be deprecated in the future.

libvast_test/vast/test/test.hpp
lava: Why three as opposed to two? Moving out of a mutable reference sounds unusual?

JoeLoser: Just added the additional overloads 👍

JoeLoser:
> You're back! 🤩
> 
> This looks great. Just have one minor request for `vast::test::unbox(...)`, then we can merge this imo. This is 100% covered by CI.
> 
> See also [actor-framework/actor-framework#1236](https://github.com/actor-framework/actor-framework/issues/1236): `caf::optional` will likely be deprecated in the future.

Thanks for sharing the CAF issue.  I noticed use of `caf::variant` as well, but that's a bit more work due to its additional API on top of `std::variant`. That's why I decided to start with `caf::optional` -> `std::optional` first as I don't think `vast` relies on optional references, so it shouldn't be too hard.

libvast_test/vast/test/test.hpp
dominiklohmann: @lava's point is kinda fair, but it depends on the expected semantics of `unbox`—does it always move internally?

dominiklohmann:
> Thanks for sharing the CAF issue. I noticed use of `caf::variant` as well, but that's a bit more work due to its additional API on top of `std::variant`.

Yeah this will be a tough one for us to migrate as well, because we have custom `caf::sum_type_access` specializations for ?`vast::data`, `vast::view`, and `vast::type`. We haven't really started thinking about a migration path yet.

libvast_test/vast/test/test.hpp
JoeLoser: In `CAF`, they just have a value-based overload:

```
template <class T>
T unbox(caf::optional<T> x) {
  if (!x)
    CAF_FAIL("x == none");
  return std::move(*x);
}
```

That's why I just provided this one function template originally. I don't have strong preferences on multiple overloads. I like the simplicity of just always moving the value and taking a copy.

dominiklohmann: I'll leave this up to @lava to decide. I have no strong opinion on this.

lava: I don't care strongly, we're not going to re-use the boxed value in test code anyways. I just thought maybe I missed something that required a third overload.

dominiklohmann:
Btw, we've recently migrated our chat to https://gitter.im/tenzir/vast if you wanna chat or need some quick input on something.

PR #1631, opened 2021-05-04
Simplify is_any_v and are_same_v

JoeLoser:

Problem:
- `is_any_v` and `are_same_v` use `std::disjunction_v` and
  `std::conjunction_v` which incurs an additional variable template
  instantiation.

Solution:
- Rewrite them in terms of fold expressions directly to avoid the
  additional variable template instantiation.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/detail/type_traits.hpp
+++ b/libvast/vast/detail/type_traits.hpp
@@ -104,10 +104,10 @@ inline constexpr bool is_same_or_derived_v = is_same_or_derived<A, B>::value;
 // -- traits -----------------------------------------------------------------
 
 template <class T, class... Ts>
-inline constexpr bool is_any_v = std::disjunction_v<std::is_same<T, Ts>...>;
+inline constexpr bool is_any_v = (std::is_same_v<T, Ts> or ...);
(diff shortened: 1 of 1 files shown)


mavam (commented):
Neat. Can you provide insight why you chose `and` over `&&`, and `or` over `||`? (This would be a new style in the code base.)

dominiklohmann (changes requested):
`std::conjunction` short-circuits the `::value` instantiation, while the fold expression doesn't. 

I am voting against this change.

JoeLoser:
> `std::conjunction` short-circuits the `::value` instantiation, while the fold expression doesn't.
> 
> I am voting against this change.

Ah, I see. I'll close this PR then.

JoeLoser:
> Neat. Can you provide insight why you chose `and` over `&&`, and `or` over `||`? (This would be a new style in the code base.)

Sorry, the stylistic difference wasn't intended -- just a habit from other codebases I work in where we prefer `and` instead of `&&` and `or` instead of `||`.

PR #1632, opened 2021-05-04
Remove using caf::optional in vast namespace

JoeLoser:

Problem:
- There is a using directive for `caf::optional` in `vast/optional.hpp`
  at `vast` namespace level scope, which is not ideal. Some parts of the
  codebase use `optional` unqualified which refers to `caf::optional` due
  to this using directive, which makes refactorings more difficult.

Solution:
- Remove using directive.
- Be explicit about use of `caf::optional` where previously call sites
  referred to `optional` unqualified.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/attribute.cpp
+++ b/libvast/src/attribute.cpp
@@ -15,9 +15,8 @@ namespace vast {
 attribute::attribute(std::string key) : key{std::move(key)} {
 }
 
-attribute::attribute(std::string key, optional<std::string> value)
-  : key{std::move(key)},
-    value{std::move(value)} {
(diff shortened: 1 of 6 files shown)


dominiklohmann (approved):
Apparently the export timeout integration test can actually generate events even though the timeout is set to 0. Wonder what's happening there, that surely looks to be unrelated.

Also, I've repeatedly seen the Example Plugin workflow not start after all its dependencies are done and I am unsure why that is.

Will merge this tomorrow after I created internal tickets to look at these two things.

PR #1634, opened 2021-05-05
Consolidate VAST CI workflows

dominiklohmann:
### :notebook_with_decorative_cover: Description

This will allow us to make the Docker and the Nix jobs required so we always run them. It also removes some duplication.

### :memo: Checklist

- [x] All user-facing changes have changelog entries.
- [x] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [x] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

Read the workflow files. Update branch protection after merging.

PR #1635, opened 2021-05-05
expression: DRY with predicate_transformer

JoeLoser:

Problem:
- The implementation for transforming a `conjunction` or `disjunction`
  into a `result_type` is nearly identical.

Solution:
- Factor out the implementation into a common function template
  parameterized on the input type to not repeat ourselves in the
  implementation.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/expression.hpp
+++ b/libvast/vast/expression.hpp
@@ -274,35 +274,11 @@ struct predicate_transformer {
   }
 
   result_type operator()(const conjunction& c) const {
-    conjunction result;
-    for (auto& op : c) {
-      auto x = caf::visit(*this, op);
-      if constexpr (std::is_convertible_v<result_type, conjunction::value_type>) {
-        result.push_back(std::move(x));
-      } else {
-        if (!x)
-          return x;
-        else
-          result.push_back(std::move(*x));
-      }
-    }
-    return result;
+    return make_result(c);
   }
 
   result_type operator()(const disjunction& c) const {
-    disjunction result;
-    for (auto& op : c) {
-      auto x = caf::visit(*this, op);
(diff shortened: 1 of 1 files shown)


dominiklohmann (approved):
LGTM.

Note that CI will not be happy because I just changed the branch protection rules for #1634, but we can merge this one regardless. All future PR's will need to be rebased after #1634 is merged.

libvast/vast/expression.hpp
```diff
@@ -319,6 +295,24 @@ struct predicate_transformer {
   }
 
   F f;
+
+private:
+  template <class T>
+  result_type make_result(const T& input) const {
```
tobim, line 301: This doesn't even have to be a to be a template, as it essentially just maps the funtion `f` down the tree, so it does nothing that is specific to `conjunction` or `disjunction`. Since both of them are just simple wrappers over `std::vector<expression>`, we can simply return that here and wrap it into the right variant type at the call site. Variables of this type are called `connective` in a few places.

tobim (changes requested):
I think you have the right idea here.

libvast/vast/expression.hpp
JoeLoser: Do you mind if I explore that proposed refactoring in a follow-up PR? I'll need to look at the callers a bit more to understand what's going on. As it stands, I think this PR is an improvement in the area of code, even if it's a function template.

tobim: I realize that this is actually not that trivial because of the support for transformation that can fail.
I accept this patch as is then.

tobim (approved):

PR #1637, opened 2021-05-05
system: caf::variant -> std::variant

JoeLoser:

Problem:
- System commands use `caf::variant` whose functionality exists in the
  standard library. Moreover, `caf::variant` may be deprecated in the
  future.

Solution:
- Replace uses of `caf::variant` with `std::variant`. Similarly, use
  `std::get_if` instead of `caf::get_if`.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/src/system/count_command.cpp
+++ b/libvast/src/system/count_command.cpp
@@ -47,11 +47,11 @@ caf::message count_command(const invocation& inv, caf::actor_system& sys) {
   // Get VAST node.
   auto node_opt
     = system::spawn_or_connect_to_node(self, options, content(sys.config()));
-  if (auto err = caf::get_if<caf::error>(&node_opt))
+  if (auto err = std::get_if<caf::error>(&node_opt))
     return caf::make_message(std::move(*err));
-  auto& node = caf::holds_alternative<node_actor>(node_opt)
-                 ? caf::get<node_actor>(node_opt)
-                 : caf::get<scope_linked<node_actor>>(node_opt).get();
+  const auto& node = std::holds_alternative<node_actor>(node_opt)
+                       ? std::get<node_actor>(node_opt)
+                       : std::get<scope_linked<node_actor>>(node_opt).get();
   VAST_ASSERT(node != nullptr);
   // Start signal monitor.
   std::thread sig_mon_thread;
--- a/libvast/src/system/explore_command.cpp
+++ b/libvast/src/system/explore_command.cpp
@@ -64,11 +64,11 @@ caf::message explore_command(const invocation& inv, caf::actor_system& sys) {
   // Get VAST node.
   auto node_opt
     = system::spawn_or_connect_to_node(self, options, content(sys.config()));
-  if (auto err = caf::get_if<caf::error>(&node_opt))
+  if (auto err = std::get_if<caf::error>(&node_opt))
     return caf::make_message(std::move(*err));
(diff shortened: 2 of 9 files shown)


dominiklohmann:
I would like to discuss this with the team first. We use caf::visit a lot, and some things we use CAF's variants at cannot be easily replaced.

dominiklohmann:
We discussed this in the team, and came to a mixed conclusion:

VAST has the following heavily used types that rely on `caf::sum_type_access`:
- `vast::data` (cannot easily be replaced with `std::variant`)
- `vast::view` (can easily be replaced with `std::variant`)
- `vast::type` (due to be replaced with FlatBuffers for mmap support when loading from disk)

We very often use `caf::visit(<callback>, <data or view>, <type>)` in our code base to double dispatch with the respective type. For the planned FlatBuffers rewrite of `vast::type` we need to introduce a custom visit anyways, or the refactoring will be very much larger than it otherwise needs to.

Do you happen to know whether it is possible to use `std::visit` with custom types that are neither `std::variant` nor inherit from it?

Generally speaking we're fine with merging this as-is since it'll never collide with one of the three listed types.

JoeLoser:
> We discussed this in the team, and came to a mixed conclusion:
> 
> VAST has the following heavily used types that rely on `caf::sum_type_access`:
> 
> * `vast::data` (cannot easily be replaced with `std::variant`)
> * `vast::view` (can easily be replaced with `std::variant`)
> * `vast::type` (due to be replaced with FlatBuffers for mmap support when loading from disk)
> 
> We very often use `caf::visit(<callback>, <data or view>, <type>)` in our code base to double dispatch with the respective type. For the planned FlatBuffers rewrite of `vast::type` we need to introduce a custom visit anyways, or the refactoring will be very much larger than it otherwise needs to.
> 
> Generally speaking we're fine with merging this as-is since it'll never collide with one of the three listed types.

Thanks for some context. From poking around, I came to similar conclusions with those types specializing `caf::sum_type_access`. In general, I'd prefer to use `std::variant` where possible so it makes it obvious where we rely on different behavior/needs to justify the use of `caf::variant`. As a new reader of the code, generally speaking, I prefer sticking with the standard library types when I can to minimize mental overhead in learning a new library's set of types. With that being said, this would put the codebase in a mixed-world of variant use, which does have its own mental hurdles for existing developers. That is, you need to know at times whether you're dealing with a `caf::variant` or `std::variant`. This can be mitigated by a small wrapping API in `vast` to delegate to `std::get_if` or `caf::get_if` based on the variant type, for example. Not sure it's worth it though since the footprint in the codebase for variant usage doesn't seem too big.

> Do you happen to know whether it is possible to use `std::visit` with custom types that are neither `std::variant` nor inherit from it?
According to [CppReference](https://en.cppreference.com/w/cpp/utility/variant/visit) it states that it is unspecified:

> It is unspecified whether other argument types, e.g. a class derived from a std::variant, are supported.

I haven't tried whether it works in practice, but regardless, in general, I don't like relying on unspecified behavior between different compilers.

dominiklohmann (approved):

JoeLoser:
@dominiklohmann @mavam http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2162r2.html may interest you with inheriting from `std::variant`. It's proposed for C++23 and I don't think there will be any opposition to it.

PR #1639, opened 2021-05-06
Add command-line option for VAST log output from unit tests

lava:

With the introduction of spdlog logging, we finally have an easy way to get log output from unit tests without having to fight caf internals.

Note that this is purely a developer-facing feature, so any missing follow-ups (write to file, log format settings, etc.) can be implemented by the affected party as needed.

### :memo: Checklist


tobim (approved):
This is a very good thing to have.

PR #1641, opened 2021-05-07
Decouple type contains trait from caf::detail::type_list

JoeLoser:

Problem:
- `caf::detail::tl_contains` only works with `caf::type_list` types,
  which doesn't lend itself well to future refactorings.
- `caf::detail::tl_contains` is slow since it is defined in terms of
  recursive class templates.

Solution:
- Use `vast::detail::contains_type_v` which works on any class template
  holder rather than just `caf::type_list`.
- Compilation times improve as well due to use of fold expressions
  rather than recursive class templates.

### :memo: Checklist

Commit-by-commit.

### DIFF
--- a/libvast/test/detail/type_traits.cpp
+++ b/libvast/test/detail/type_traits.cpp
@@ -12,8 +12,30 @@
 
 #include "vast/test/test.hpp"
 
+#include <caf/variant.hpp>
+
+#include <variant>
+
 TEST(sum) {
   static_assert(vast::detail::sum<> == 0);
   static_assert(vast::detail::sum<1, 2, 3> == 6);
   static_assert(vast::detail::sum<42, 58> == 100);
+}
+
+namespace {
+template <class... Ts>
+struct fake_list {};
+
+template <template <class...> class TList>
+constexpr auto check() {
+  static_assert(vast::detail::contains_type_v<TList<int, double>, double>);
+  static_assert(not vast::detail::contains_type_v<TList<int, double>, char>);
+}
+} // namespace
+
+TEST(contains_type) {
+  check<std::variant>();
+  check<std::tuple>();
+  check<fake_list>();
+  check<caf::variant>();
 }
--- a/libvast/vast/detail/type_traits.hpp
+++ b/libvast/vast/detail/type_traits.hpp
@@ -155,4 +155,15 @@ using remove_optional_t = typename remove_optional<T>::type;
 template <auto... Values>
 constexpr auto sum = (0 + ... + Values);
 
+template <class T, template <class...> class TList, class... Ts>
+constexpr auto contains_type_impl(TList<Ts...>) {
+  return std::bool_constant<is_any_v<T, Ts...>>{};
+}
+
+template <class TList, class T>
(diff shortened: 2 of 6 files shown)


libvast/test/detail/type_traits.cpp
```diff
@@ -8,13 +8,32 @@
...
 }
 
+template <class... Ts>
+struct fake_list {};
+
+template <template <class...> class TList>
+constexpr auto check() {
+  static_assert(vast::detail::contains_type_v<TList<int, double>, double>);
+  static_assert(not vast::detail::contains_type_v<TList<int, double>, char>);
+}
```
dominiklohmann, line 32: These need to be in anonymous namespaces, otherwise we risk running into (possibly silent) linker errors.

libvast/vast/detail/type_traits.hpp
```diff
@@ -156,4 +156,13 @@ using remove_optional_t = typename remove_optional<T>::type;
 template <auto... Values>
 constexpr auto sum = (0 + ... + Values);
 
+template <class T, template <class...> class TList, class... Ts>
+constexpr auto contains_type(TList<Ts...>) {
+  return std::bool_constant<is_any_v<T, Ts...>>{};
+}
```
dominiklohmann, line 162: It's a bit weird that this is not a struct, which makes it hard to use with other meta-programming facilities. I don't have a particularly hard opinion on this, but if you have an idea how to rewrite this easily as a struct that inherits from `std::bool_constant` than I'd be more than happy to see that change applied.

dominiklohmann (commented):
Generally fine, just have some minor requests.

libvast/vast/detail/type_traits.hpp
JoeLoser: Can you elaborate on the part about hard to use with other meta programming facilities? I just added an alias template `contains_type_t` if you want a type flavor of it (but I don't think it will ever get used in that way, in practice).

I'm not sure why you think it's weird that it's not a struct. With template templates, in deducing the types in the class template, you need to instantiate the type, which you could do in a constructor perhaps. We could maybe do something with a templated constructor and initialize a base class with the value based on `is_any_v` containment. That's some gymnastics for no benefit though IMO.

dominiklohmann: Take for example `disjunction<A, B, C>::value`, which is logically equivalent to `A::value || B::value || C::value`. I cannot use this type trait as part of such a conjunction, disjunction, or negation.

I'm fine with this as-is, except for one minor naming request.

For a given type trait `A`, I expected `A_t` to be `typename A::type` and `A_v` to be `A::value`. The naming of the helper function should not be `A`.

dominiklohmann: Will do should we ever need this. Thanks for going through this elongated review with me. 👍

dominiklohmann (approved):

libvast/vast/detail/type_traits.hpp
JoeLoser: @dominiklohmann https://wandbox.org/permlink/HpHyHTiP4NaN5aiq is an example of what I think you're after.

PR #1644, opened 2021-05-08
Enable duplicate tracking in Bloom filters

mavam:

This PR adds a new *tracking policy* to the Bloom filter, making it possible to
find out whether an addition resulted in a no-op (in case of a duplicate or
collision) or did turn at least one bit to 1.

### :memo: Checklist

In one shot.

### DIFF
--- a/libvast/test/bloom_filter.cpp
+++ b/libvast/test/bloom_filter.cpp
@@ -209,3 +209,16 @@ TEST(bloom filter - simple hasher and partitioning) {
   REQUIRE_EQUAL(err, caf::none);
   CHECK(x == y);
 }
+
+TEST(bloom filter - duplicate tracking) {
+  bloom_filter_parameters xs;
+  xs.m = 1_M;
+  xs.p = 0.1;
+  auto x = vast::test::unbox(
+    make_bloom_filter<xxhash, double_hasher, policy::no_partitioning,
+                      policy::duplicate_tracking>(xs));
+  CHECK(!x.lookup(42));
+  CHECK(x.add(42));
+  CHECK(x.lookup(42));
+  CHECK(!x.add(42));
+}
--- a/libvast/test/bloom_filter.cpp
+++ b/libvast/test/bloom_filter.cpp
@@ -191,7 +191,7 @@ TEST(bloom filter - simple hasher and partitioning) {
   xs.m = 10_M;
   xs.p = 0.001;
   auto x = vast::test::unbox(
-    make_bloom_filter<xxhash, simple_hasher, policy::partitioning>(xs));
+    make_bloom_filter<xxhash, simple_hasher, policy::partitioning::yes>(xs));
   CHECK_EQUAL(x.size(), 10_M);
   CHECK_EQUAL(x.num_hash_functions(), 10u);
   x.add(42);
@@ -204,7 +204,7 @@ TEST(bloom filter - simple hasher and partitioning) {
   std::vector<char> buf;
   auto err = detail::serialize(buf, x);
   REQUIRE_EQUAL(err, caf::none);
-  bloom_filter<xxhash, simple_hasher, policy::partitioning> y;
+  bloom_filter<xxhash, simple_hasher, policy::partitioning::yes> y;
   err = detail::deserialize(buf, y);
   REQUIRE_EQUAL(err, caf::none);
   CHECK(x == y);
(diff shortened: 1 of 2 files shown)


mavam:
[ch25400]

libvast/test/bloom_filter.cpp
```diff
@@ -210,3 +210,16 @@ TEST(bloom filter - simple hasher and partitioning) {
...
   CHECK(x == y);
 }
+
+TEST(bloom filter - duplicate tracking) {
+  bloom_filter_parameters xs;
+  xs.m = 1_M;
+  xs.p = 0.1;
+  auto x = vast::test::unbox(
+    make_bloom_filter<xxhash, double_hasher, policy::no_partitioning,
+                      policy::duplicate_tracking>(xs));
```
dominiklohmann, line 220: If you swap the two policy arguments you get very much undesired behavior. Let's make both enum classes instead with two arguments each.

```
enum class partition_bloom_filter {
  yes,
  no,
};

enum class track_duplicates {
  yes,
  no,
};
```

dominiklohmann (commented):
Just have one slight nit to make it harder to misuse the policy template arguments.

libvast/test/bloom_filter.cpp
mavam: I'm not aware of this idiom. So instead of passing an unconstrained type template, pass an enum?

I assume these should also be in namespace `policy`, right?

mavam: What I like about this is the user side of things: 
```cpp
    if constexpr (tracking_policy == policy::tracking::no) {
      ...
    }
```

Much easier to read than the type-level comparison with `is_same`.

libvast/vast/bloom_filter.hpp
```diff
@@ -65,11 +70,25 @@ class bloom_filter : detail::equality_comparable<
 
   /// Adds an element to the Bloom filter.
   /// @param x The element to add.
+  /// @returns `void` if the tracking policy is `tracking::no`,
+  /// otherwise a `bool` that is `false` iff *x* already exists in the filter.
   template <class T>
-  void add(T&& x) {
+  auto add(T&& x) {
```
dominiklohmann, line 76: Please `[[nodiscard]]` this one.

libvast/vast/bloom_filter.hpp
```diff
@@ -137,16 +156,18 @@ class bloom_filter : detail::equality_comparable<
 /// Constructs a Bloom filter for a given set of parameters.
 /// @tparam HashFunction The hash function to use in the hasher.
 /// @tparam Hasher The hasher type to generate digests.
-/// @tparam PartitioningPolicy The partitioning policy.
+/// @tparam partitioning The partitioning policy.
```
dominiklohmann, line 159:
```suggestion
/// @tparam Partitioning The partitioning policy.
```

dominiklohmann (approved):
Just small things now, then feel free to merge.

libvast/vast/bloom_filter.hpp
mavam: But this can be `void` as well. Is this well-defined?

It's also entirely optionally signaling.

dominiklohmann: The optional signaling is opt-in, so if users enable that I assume they also want to catch this.

The nodiscard is well-defined on void.

libvast/vast/bloom_filter.hpp
```diff
@@ -65,11 +70,25 @@ class bloom_filter : detail::equality_comparable<
...
+        bits_[position(i, digests[i])] = true;
+    }
+    if constexpr (tracking_policy == policy::tracking::yes) {
+      auto unique = false;
+      for (size_t i = 0; i < digests.size(); ++i) {
+        auto bit = bits_[position(i, digests[i])];
+        if (!bit)
+          unique = true;
+        bit = true;
+      }
```
lava, line 89: Do we really need a separate template policy argument **and** duplicated code for such simple functionality? Why not just always return a (discardable) bool here?

mavam: I thought about this as well, but then felt uneasy with the extra branching. Granted, I'm not sure if the bandwidth on the Bloom filter construction site matters.

What would you prefer?

lava: I would be extremely suprised if that turns out to be measurable. The old value needs to be loaded + stored anyways to apply the masking, so the only additional work is an  `or` operation (you could write it as `unique |= (byte == 0)` if you don't trust the compiler), which should be amongst the fastest possible things you can do to a register.

It's not really important, so if you have other things to do I don't mind merging this as it is, just be prepared that there's a high chance that someone will remove the extra code again in the future for better maintainability ;)

mavam: I'm nodding along, this all makes sense. 

I've removed the tracking policy, but kept the "new-style" enum class policy refactoring.

PR #1645, opened 2021-05-08
Simplify hash_append for tuples

JoeLoser:

Problem:
- `hash_append` for tuples delegates to a helper function to hash each
  item in the tuple explicitly.

Solution:
- Use `std::apply` directly to avoid the need for the helper function.

### :memo: Checklist

File-by-file.


mavam (approved):
LGTM.

@dominiklohmann can you merge this is to your liking?

libvast/vast/concept/hashable/hash_append.hpp
```diff
@@ -329,20 +329,11 @@ void hash_append(Hasher& h,
...
-  (hash_append(h, std::get<I>(t)), ...);
-}
-
-} // namespace detail
-
 template <class Hasher, class ...T>
 std::enable_if_t<!detail::is_contiguously_hashable<std::tuple<T...>, Hasher>{}>
 hash_append(Hasher& h, const std::tuple<T...>& t) noexcept {
-  detail::tuple_hash(h, t, std::index_sequence_for<T...>{});
+  std::apply([&h](auto&& x) { hash_append(h, std::forward<decltype(x)>(x)); },
```
dominiklohmann, line 335:
```suggestion
  std::apply([&h](auto&&... xs) { hash_append(h, std::forward<decltype(xs)>(xs)...); },
```

As-is this only works for single-element tuples. Might need additional text reflow for clang-format, but you get the idea.

dominiklohmann (changes requested):

libvast/vast/concept/hashable/hash_append.hpp
lava: It sounds like we're lacking a unit test that hashes a tuple with multiple elements. It might be a good idea to add one for complex refactorings like this so we know the output doesn't change.

(Also, wouldn't the suggested fix result in an endless loop? We never take off the head of the param pack)

JoeLoser: Oops, thanks for spotting the need for the pack.

I'll add some unit tests and fix up the implementation later tonight likely.

dominiklohmann: Thanks! 👍

PR #1648, opened 2021-05-10
Simplify clang-format include categories

dominiklohmann:

Our clang-format include categories did not make much sense for plugins and bundled tools. This changes them so they're more widely applicable and, in a second commit, applies that change to the code base.

This will make plugin development easier because we can now just copy this file over instead of needing to keep them in sync manually.

### :memo: Checklist

Look at first commit in-depth, skim through second commit. The second commit was created semi-automatically; I only removed hunks that contained complex non-whitespace changes.


lava (approved):
I only sampled the changes here but everything looks good as far as I can tell, apart from the new formatting in `posix.cpp` that now groups `sys/` and `caf/` together:

```
#include <caf/expected.hpp>
#include <sys/select.h>
#include <sys/socket.h>
#include <sys/types.h>
#include <sys/un.h>
```

PR #1655, opened 2021-05-11
Add step size to disk monitor

lava:

Add a "step size" setting to the disk monitor to only trigger re-checking of the disk size after deleting N partitions. Useful for situations where computing the disk size is computationally expensive, e.g. on compressed filesystems.

### :memo: Checklist

### DIFF
--- a/libvast/src/system/application.cpp
+++ b/libvast/src/system/application.cpp
@@ -389,7 +389,9 @@ auto make_start_command() {
       .add<std::string>("disk-budget-check-binary",
                         "binary to run to determine current disk usage")
       .add<std::string>("disk-budget-high", "high-water mark for disk budget")
-      .add<std::string>("disk-budget-low", "low-water mark for disk budget"));
+      .add<std::string>("disk-budget-low", "low-water mark for disk budget")
+      .add<size_t>("disk-budget-step-size", "number of partitions to erase "
+                                            "before re-checking size"));
 }
 
 auto make_stop_command() {
--- a/libvast/vast/system/disk_monitor.hpp
+++ b/libvast/vast/system/disk_monitor.hpp
@@ -18,24 +18,32 @@
 
 namespace vast::system {
 
+struct disk_monitor_config {
+  /// When current disk space is above the high water mark, stuff
+  /// is deleted until we get below the low water mark.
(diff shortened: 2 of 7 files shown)


libvast/src/system/application.cpp
```diff
@@ -389,7 +389,9 @@ auto make_start_command() {
       .add<std::string>("disk-budget-check-binary",
                         "binary to run to determine current disk usage")
       .add<std::string>("disk-budget-high", "high-water mark for disk budget")
-      .add<std::string>("disk-budget-low", "low-water mark for disk budget"));
+      .add<std::string>("disk-budget-low", "low-water mark for disk budget")
+      .add<size_t>("disk-budget-step-size", "number of partitions to erase "
+                                            "before re-checking size"));
```
dominiklohmann, line 394: This is missing from the `vast.yaml.example` file.

libvast/vast/system/disk_monitor.hpp
```diff
@@ -55,6 +59,7 @@ struct disk_monitor_state {
 struct disk_monitor_config {
   size_t high_water_mark;
   size_t low_water_mark;
+  size_t step_size;
```
dominiklohmann, line 62: These should have default values, otherwise they may contain garbage if you forget to initialize them.

libvast/vast/system/disk_monitor.hpp
```diff
@@ -27,6 +27,10 @@ struct disk_monitor_state {
   size_t high_water_mark;
   size_t low_water_mark;
 
+  /// How many partitions to delete at once when reaching the high-water
+  /// mark.
+  size_t step_size;
```
dominiklohmann, line 32: Instead of duplicating the config, wouldn't it make more sense to just use the `disk_monitor_config` here?

dominiklohmann (approved):
I just have three small comments _and_ this needs a changelog entry. Other than that this is ready to ship.

I tested this locally with very small partition sizes and a custom check binary that counts partition files. It works very well.

libvast/vast/system/disk_monitor.hpp
lava: Yup, that's much better.

PR #1657, opened 2021-05-12
Remove configure script

lava:

To quote the corresponding story,  the main reasons for removing:

   * The script is called configure but does not act like a standard configure script, leading to confusion for people who just want to plug the build into some automated system with hooks. We usually have to tell them "please just ignore the configure script, we are actually using cmake as build system, yes we know its confusing".
       * It does not operate in the current directory by default
       * It overwrites the contents of any existing directory named build/ by default
       * It does not understand all the same default options that autotools-generated configure scripts have
   * Its easy to forget to add new features to the configure script, leading to a divergence over time
   * It slows down development since we have to invest time in updating it every time we touch the build system.

### :memo: Checklist


.github/workflows/vast.yaml
```diff
@@ -3,7 +3,7 @@ on:
   workflow_dispatch:
     inputs:
       arguments:
-        description: 'Additional arguments to pass, e.g., `--with-plugin=<path/to/plugin>` or `-D<CMake Option>`'
+        description: 'Additional arguments to pass, e.g., `-D<CMake Option>`'
```
dominiklohmann, line 6: This is not true; the Nix build script has a --with-plugin option.

.github/workflows/vast.yaml
```diff
@@ -60,7 +60,7 @@ jobs:
           CC: gcc-8
           CXX: g++-8
         run: |
-          ./configure --with-bundled-caf --without-arrow
+          cmake -B build -DVAST_ENABLE_BUNDLED_CAF=1 -DVAST_ENABLE_ARROW=0
```
dominiklohmann, line 63: Can you use -DVAR:TYPE=VALUE so the proper type shows up in `ccmake`?

E.g.,

`-DVAST_ENABLE_BUNDLED_CAF:BOOL=ON`

.github/workflows/vast.yaml
```diff
@@ -248,13 +248,13 @@ jobs:
...
-              --with-bundled-caf \
+            cmake -B "$BUILD_DIR" \
+              -DCMAKE_INSTALL_PREFIX="${PWD}/opt/vast" \
+              -DCPACK_PACKAGE_FILE_NAME="$PACKAGE_NAME" \
+              -DCPACK_GENERATOR="STRING TGZ"
+              -DVAST_ENABLE_LSVAST=1 \
+              -DVAST_ENABLE_DSCAT=1 \
+              -DVAST_ENABLE_BUNDLED_CAF=1 \
               ${{ matrix.build.extra-flags }} \
               ${{ github.event_name == 'release' && matrix.configure.flags || matrix.configure.ci-flags }}
```
dominiklohmann, line 259: These need to be adapted in the Matrix I think.

changelog/unreleased/breaking-changes/1657.md
```diff
@@ -0,0 +1,4 @@
+The `configure` script was removed. This was a custom script that
+mimicked the functionality of an autotools-based `configure` script
+by writing directly to the cmake cache. Instead, users now must use
+the `cmake` and/or `ccmake` binaries directly to configure VAST.
```
dominiklohmann, line 4: We have a ton of variables in our CMake that are not `mark_as_advanced` for `ccmake`. Can you create a follow-up story for cleaning that up?

scripts/create-release
```diff
@@ -72,7 +72,7 @@ if [[ "${last_rc_version}" == *"-rc"* ]]; then
 fi
 
 # Configure the build and update the local changelog.
-${source_dir}/configure --build-dir="${build_dir}"
+cmake -B build
```
dominiklohmann, line 75:
```suggestion
cmake -B "${build_dir}"
```

dominiklohmann (changes requested):
Love to see this done so much that I am commenting this on my day off. 🎉

.github/workflows/vast.yaml
lava: Sure, although this is the CI config so there is no human to run `ccmake` later on ;)

JoeLoser:
I like this. Is there any plans to remove scripts use of `configure`? E.g. https://github.com/tenzir/vast/blob/a2cb4be879a13cef855da2c1d73083204aed4dff/scripts/clang-build-analyzer#L37 and perhaps others.

lava:
@JoeLoser Absolutely, I used `git grep` to search for any occurences in the scripts but I must have missed this one.

.github/workflows/vast.yaml
```diff
@@ -248,13 +248,13 @@ jobs:
...
             ccache --show-config
-            ./configure \
-              --prefix="${PWD}/opt/vast" \
-              --build-dir="$BUILD_DIR" \
-              --package-name="$PACKAGE_NAME" \
-              --with-lsvast \
-              --with-dscat \
-              --with-bundled-caf \
+            cmake -B "$BUILD_DIR" \
+              -DCMAKE_INSTALL_PREFIX="${PWD}/opt/vast" \
```
dominiklohmann, line 252: No need to set the prefix here for a relocatable build, instead we should install using `cmake --install "${BUILD_DIR}" --prefix "${PWD}/opt/vast"`.

.github/workflows/vast.yaml
```diff
@@ -379,8 +379,8 @@ jobs:
             cxx: clang++
         configure:
           - tag: Release
-            flags: --release
-            ci-flags: --ci-build
+            flags: -DCMAKE_BUILD_TYPE=Release
+            ci-flags: -DCMAKE_BUILD_TYPE=CI -DVAST_ENABLE_ASAN:BOOL=one
```
dominiklohmann, line 383:
```suggestion
            flags: -DCMAKE_BUILD_TYPE:STRING=Release
            ci-flags: -DCMAKE_BUILD_TYPE:STRING=CI -DVAST_ENABLE_ASAN:BOOL=ON
```

.github/workflows/vast.yaml
```diff
@@ -248,13 +248,13 @@ jobs:
...
-            ./configure \
-              --prefix="${PWD}/opt/vast" \
-              --build-dir="$BUILD_DIR" \
-              --package-name="$PACKAGE_NAME" \
-              --with-lsvast \
-              --with-dscat \
-              --with-bundled-caf \
+            cmake -B "$BUILD_DIR" \
+              -DCMAKE_INSTALL_PREFIX="${PWD}/opt/vast" \
+              -DCPACK_PACKAGE_FILE_NAME="$PACKAGE_NAME" \
```
dominiklohmann, line 253:
```suggestion
              -DCPACK_PACKAGE_FILE_NAME:STRING="$PACKAGE_NAME" \
```

.github/workflows/vast.yaml
```diff
@@ -506,14 +506,14 @@ jobs:
...
-            --with-dscat \
-            --with-bundled-caf \
+          cmake -B "$BUILD_DIR" \
+            -DCMAKE_INSTALL_PREFIX="${PWD}/opt/vast" \
+            -DCPACK_PACKAGE_FILE_NAME="$PACKAGE_NAME" \
+            -DCPACK_GENERATOR:STRING=TGZ \
+            -DVAST_PLUGINS:STRING="${PWD}/plugins/pcap" \
+            -DVAST_ENABLE_LSVAST:BOOL=ON \
+            -DVAST_ENABLE_DSCAT:BOOL=ON \
+            -DVAST_ENABLE_BUNDLED_CAF:BOOL=ON \
```
dominiklohmann, line 516: Same as above. We can also extract this into a composite run step for deduplication.

.github/workflows/vast.yaml
```diff
@@ -506,14 +506,14 @@ jobs:
...
-            --package-name="$PACKAGE_NAME" \
-            --with-pcap-plugin \
-            --with-lsvast \
-            --with-dscat \
-            --with-bundled-caf \
+          cmake -B "$BUILD_DIR" \
+            -DCMAKE_INSTALL_PREFIX="${PWD}/opt/vast" \
+            -DCPACK_PACKAGE_FILE_NAME="$PACKAGE_NAME" \
+            -DCPACK_GENERATOR:STRING=TGZ \
+            -DVAST_PLUGINS:STRING="${PWD}/plugins/pcap" \
```
dominiklohmann, line 513: Why do we build with PCAP on macOS but without it on Debian?

scripts/clang-build-analyzer
```diff
@@ -34,8 +34,9 @@ git diff | git -C "${WORKING_DIR}" apply
 # Configure VAST to build with time-traces enabled
 pushd "${WORKING_DIR}"
 on_exit=('popd' "${on_exit[@]}")
-./configure "${@}" --build-dir="${WORKING_DIR}/build" \
-  --extra-flags="-ftime-trace"
+cmake "${@}" -B "${WORKING_DIR}/build" \
+  -DCMAKE_CXX_COMPILER="clang++" \
```
dominiklohmann, line 38: This may break for some setups. Please do not hardcode the compiler in scripts.

scripts/clang-build-analyzer
```diff
@@ -34,8 +34,9 @@ git diff | git -C "${WORKING_DIR}" apply
...
+  -DCMAKE_CXX_FLAGS="-ftime-trace"
```
dominiklohmann, line 39:
```suggestion
  -DVAST_ENABLE_TIME_TRACE:BOOL=ON
```

Dockerfile
```diff
@@ -41,12 +41,11 @@ COPY schema ./schema
...
 COPY vast ./vast
-COPY .clang-format .cmake-format LICENSE LICENSE.3rdparty README.md BANNER CHANGELOG.md CMakeLists.txt configure vast.yaml.example ./
-RUN ./configure \
-    --prefix=$PREFIX \
-    --build-type=$BUILD_TYPE \
-    --log-level=INFO \
-    --generator=Ninja
+COPY .clang-format .cmake-format LICENSE LICENSE.3rdparty README.md BANNER CHANGELOG.md CMakeLists.txt vast.yaml.example ./
+RUN cmake -B build -G Ninja \
+    -DCMAKE_INSTALL_PREFIX="$PREFIX" \
```
dominiklohmann, line 46: Considering there's no need for this to be relocatable inside the Docker container we might want to consider enabling `VAST_ENABLE_RELOCATABLE_INSTALLATIONS:BOOL=OFF`.

Dockerfile
```diff
@@ -41,12 +41,11 @@ COPY schema ./schema
...
+    -DCMAKE_BUILD_TYPE="$BUILD_TYPE" \
+    -DCMAKE_VAST_LOG_LEVEL=INFO
```
dominiklohmann, line 48: The variable for the log level has no `CMAKE_` prefix, but I don't think we should use it anyways, but rather rely on the defaults for the build type.

```suggestion
    -DCMAKE_BUILD_TYPE:STRING="$BUILD_TYPE"
```

README.md
```diff
@@ -84,7 +84,7 @@ git clone --recursive https://github.com/tenzir/vast
 Once you have all dependencies in place, build VAST with the following commands:
 
 ```sh
-./configure
+cmake -B build
```
dominiklohmann, line 87: This should mention `ccmake`, or link to the docs page.

README.md
```diff
@@ -84,7 +84,7 @@ git clone --recursive https://github.com/tenzir/vast
...
 cmake --build build
 cmake --build build --target test
 cmake --build build --target install
```
dominiklohmann, line 90: While you're at it:

```suggestion
cmake --install build [--prefix /path/to/prefix]
```

dominiklohmann (changes requested):

.github/workflows/vast.yaml
lava: I don't disagree, but this PR is about removing the configure script so I'd like to exactly replicate the current behavior, and then do any refactorings in a separate PR.

dominiklohmann: Fair enough.

README.md
dominiklohmann: Should skip this then as well.

scripts/clang-build-analyzer
lava: But `-ftime-trace` is not understood by g++ and results in an error.

dominiklohmann: Hence why we have a `VAST_ENABLE_TIME_TRACE` option.

Dockerfile
lava: I just changed the variable name now to keep the end result the same as before.

.github/workflows/vast.yaml
lava: No idea; probably a copy/pasting mistake but not something I want to touch in this PR.

dominiklohmann (approved):

PR #1659, opened 2021-05-17
Allow for running unit tests in parallel

dominiklohmann:

Before this change, the filesystem unit test fixture used a fixed `vast-unit-test` root directory. Now, we use `vast-unit-test/<suite>` instead to avoid races when running tests in parallel, e.g., via `ctest -j12 -C Debug`.

### :memo: Checklist

Test locally via

```bash
ctest --test-dir <path/to/build> -j12 -C Debug
```

### DIFF
--- a/changelog/unreleased/bug-fixes/1659.md
+++ b/changelog/unreleased/bug-fixes/1659.md
@@ -1,0 +1,1 @@
+Executing VAST's unit test suite in parallel no longer fails.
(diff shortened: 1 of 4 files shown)


lava (approved):
Looks good to me, and at least solves the issue with ctest. (although running two independent instances of `vast-test` in parallel would still fail, but we don't need to solve this in this PR)

PR #1660, opened 2021-05-17
Remove Debian Clang-10 build from VAST CI

dominiklohmann:

We currently have no user for this build, and are already testing Clang via both AppleClang and the latest clang. For GCC we test GCC-8 explicitly on Debian, and latest via the Nix build. That already covers the range we want to support, so this is superfluous.

### :memo: Checklist

n/t


lava (approved):
Looks good.

PR #1662, opened 2021-05-17
Opt-in to transparent lookup for data and data_view

dominiklohmann:

This moves code to where it should've been from the get-go. Needs this in a plugin to avoid code duplication.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/index/hash_index.hpp
+++ b/libvast/vast/index/hash_index.hpp
@@ -8,8 +8,6 @@
 
 #pragma once
 
-#include "vast/concept/hashable/uhash.hpp"
-#include "vast/concept/hashable/xxhash.hpp"
 #include "vast/data.hpp"
 #include "vast/detail/assert.hpp"
 #include "vast/detail/overload.hpp"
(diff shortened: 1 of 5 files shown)


mavam (approved):
This looks good.

Before merging, please add a unit test, for basic types and containers, that demonstrates that the hash digests are actually equal.

libvast/vast/index/hash_index.hpp
```diff
@@ -256,36 +254,9 @@ class hash_index : public value_index {
...
-      static_assert(std::is_same_v<L, data> || std::is_same_v<L, data_view>);
-      static_assert(std::is_same_v<R, data> || std::is_same_v<R, data_view>);
-      return x == y;
-    }
-  };
-
   // We use a robin_map here because it supports heterogenous lookup, which
   // has a major performance impact for `seeds_`, see ch13760.
-  tsl::robin_map<data, size_t, data_hash, data_equal> seeds_;
+  tsl::robin_map<data, size_t> seeds_;
```
lava, line 259: Can you add a static assert here that the "equality" of this map has opted-in to heterogenous lookup? (it was explicit before, but after moving it to a different file it seems easy to change the functor and to accidentally forget this)

PR #1663, opened 2021-05-18
Fix table slice offset deserialization

dominiklohmann:

Before this change, we deserialized the table slice's offset and then copied the table slice, resulting in the offset being reset to `invalid_id`. This caused no issues in practice until now, because no current code uses offsets after sending table slices over the wire. This did, however, just cause a very frustrating debugging session because I relied on this to work for a plugin.

### :memo: Checklist

The change should be obvious.


lava (approved):
Nice catch.

PR #1675, opened 2021-05-21
Properly disable uppercase literal suffix check

dominiklohmann:

Apparently clang-tidy 12 requires the trailing comma here.

### :memo: Checklist

@lava check locally if this fixes it on your end as well.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -22,7 +22,8 @@ Checks: >
   -readability-function-cognitive-complexity,
   -readability-magic-numbers,
   -readability-named-parameter,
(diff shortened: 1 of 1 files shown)


lava (approved):
Works on my machine. (when I use `run-clang-format` actually bother versions work on clang-12, so it seems like this is maybe a bug in clangd?)

dominiklohmann:
> so it seems like this is maybe a bug in clangd?

Might be, I only tested via `clangd`. Regardless, this fix is simple enough.

PR #1676, opened 2021-05-21
Login to GCS in CI only if necessary

dominiklohmann:

Noticed we ran this step both earlier and more often than we should've. Considering it takes around 20 seconds, this really makes a difference.

### :memo: Checklist

See CI run.


lava (approved):
I didn't test, but I noticed the same thing so the change makes sense to me.

PR #1679, opened 2021-05-25
Sort loaded plugins by name

dominiklohmann:

This ensures the version comparison is agnostic to the order of the plugins.

### :memo: Checklist

Connect a local VAST to a VAST running the same plugins specified in a different order and notice VAST no longer warns unnecessarily.

### DIFF
--- a/changelog/unreleased/bug-fixes/1679--plugin-load-order.md
+++ b/changelog/unreleased/bug-fixes/1679--plugin-load-order.md
@@ -1,0 +1,2 @@
+VAST no longer erroneously warns about a version mismatch between client and
+server when their plugin load order differs.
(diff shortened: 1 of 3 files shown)


lava (approved):
I don't have a setup to test, but looking it the code it seems pretty safe to say that it should work. 

Relying on the order being setup during startup is a  very implicit way of doing this, but long-term the issue should resolve itself anyways as soon as record comparisons become order-independent.

dominiklohmann:
> I don't have a setup to test, but looking it the code it seems pretty safe to say that it should work.

A better fix would probably be to use a stable set for this with a custom equal_to function object, but that is far more than we need to fix this before the release. I'll write a story for that.

PR #1680, opened 2021-05-26
Use clang by default in clang-build-analyzer script

dominiklohmann:

Before this change, the `-DVAST_ENABLE_TIME_TRACE` CMake option was silently ignored when the current compiler was not set to `clang`.  On many machines, this led to the paradoxical situation where the `clang-build-analyzer` script by default runs a full build using no `clang` and no build analysis.

Instead, we now default the `CC` and `CXX` to `clang` and `clang++` respectively to still respect custom compilers and compiler wrappers, and check in our CMake code whether the current compiler supports the `-ftime-trace` flag rather than only enabling it when the compiler ID matches `clang`.

### :memo: Checklist

Try locally. Requested @lava for review since he noticed this in the first place.

I do not think this is worth a changelog entry since it does not affect VAST itself.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -301,12 +301,18 @@ if (VAST_ENABLE_TIME_REPORT)
   target_compile_options(libvast_internal INTERFACE "-ftime-report")
 endif ()
 
-cmake_dependent_option(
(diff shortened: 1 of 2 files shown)


lava (approved):
Looks just like we discussed, thanks.

PR #1684, opened 2021-05-26
Improve caching behavior in CI

dominiklohmann:

A recent change to always upload the cache overlooked that the caches were still uploaded with separate names, which is not ideal. We want to have one cache per build configuration and let `ccache` take care of the rest.

### :memo: Checklist

Reason about the change.


lava (approved):
I didn't test it (because its impossible to test), but it looks great!

PR #1685, opened 2021-05-26
Handle arbitrary types in bloom filter synopsis

lava:

The meta index lookup would select type synopses based on the field name being looked up, without verifying that the right hand side of the query also has the corresponding type. This leads to undefined behavior for queries like `s == nil` or `s == 123` (where `s` is a string field) and depending on the compiler settings also segfaults or other kinds of bad behavior.

NOTE: I'm not completely sure if the same issue exists for "normal" synopsis and we just don't notice it because the invalid call to `caf::get<view<T>>()` was not using complex types like `std::string`, or if these were typechecked somewhere else beforehand.

### :memo: Checklist

### DIFF
--- a/libvast/vast/bloom_filter_synopsis.hpp
+++ b/libvast/vast/bloom_filter_synopsis.hpp
@@ -33,6 +33,7 @@ public:
   }
 
   void add(data_view x) override {
+    VAST_ASSERT(caf::holds_alternative<view<T>>(x), "invalid data");
     bloom_filter_.add(caf::get<view<T>>(x));
   }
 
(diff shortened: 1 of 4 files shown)


dominiklohmann (commented):
Generally this looks fine to me, but I would like to refrain from merging this change with such short notice before the release. This does not fix a bug that was just recently introduced, but rather a long-standing issue and we must properly evaluate whether this is the proper fix, or whether we want to look at using transparent hashing for the bloom filter.

libvast/vast/bloom_filter_synopsis.hpp
```diff
@@ -42,12 +43,24 @@ class bloom_filter_synopsis : public synopsis {
...
+        // are non-nil.
+        if (caf::holds_alternative<view<caf::none_t>>(rhs))
+          return {};
+        if (!caf::holds_alternative<view<T>>(rhs))
+          return false;
         return bloom_filter_.lookup(caf::get<view<T>>(rhs));
       case relational_operator::in: {
         if (auto xs = caf::get_if<view<list>>(&rhs)) {
-          for (auto x : **xs)
+          for (auto x : **xs) {
```
tobim, line 56: Sorry for being pedantic: the list itself can be `nil` as well, right?

dominiklohmann: That's already covered by the `get_if`.

dominiklohmann (approved):
Discussed with @tobim. Needs a rebase + a story for switching the result type of the lookup from `optional<bool>` to an enum class with three members.

lava:
I've rebased the PR. The story (in a slightly more general form) exists as https://app.clubhouse.io/tenzir/story/19453/discussion-change-meta-index-api-to-enable-short-circuiting

PR #1689, opened 2021-05-26
Add support for different sink types

lava:

Add support for native journal / syslog output formats

### :memo: Checklist

### DIFF
--- a/libvast/CMakeLists.txt
+++ b/libvast/CMakeLists.txt
@@ -129,6 +129,27 @@ else ()
   set(ARROW_LIBRARY arrow_static)
 endif ()
 
+# -- libsystemd ----------------------------------------------------------------
+
+option(VAST_ENABLE_JOURNALD_LOGGING "Build with support for native journald log output" OFF)
+add_feature_info("VAST_ENABLE_JOURNALD_LOGGING" VAST_ENABLE_JOURNALD_LOGGING
+                 "build with support for native journald log output.")
+if (VAST_ENABLE_JOURNALD_LOGGING)
+  find_package(PkgConfig)
+
+  pkg_check_modules(LIBSYSTEMD REQUIRED libsystemd)
+
+  find_path(SYSTEMD_INCLUDE_DIRS
+    NAMES systemd/sd-journal.h
+    PATHS ${LIBSYSTEMD_INCLUDE_DIRS}
+  )
+
+  find_library(SYSTEMD_LIBRARIES
+    NAMES systemd
+    PATHS ${LIBSYSTEMD_LIBRARY_DIRS}
+  )
+endif ()
+
 # -- log level -----------------------------------------------------------------
 
 # Choose a deafult log level based on build type.
@@ -432,6 +453,11 @@ dependency_summary("FlatBuffers" ${flatbuffers_target} "Dependencies")
 find_package(yaml-cpp 0.6.2 REQUIRED)
 target_link_libraries(libvast PRIVATE yaml-cpp)
 dependency_summary("yaml-cpp" yaml-cpp "Dependencies")
+
+# Link against libsystemd
+if (VAST_ENABLE_JOURNALD_LOGGING)
+  target_link_libraries(libvast PRIVATE ${LIBSYSTEMD_LIBRARIES})
+endif ()
 
 target_compile_definitions(
   libvast
--- a/libvast/CMakeLists.txt
+++ b/libvast/CMakeLists.txt
@@ -134,21 +134,6 @@ endif ()
 option(VAST_ENABLE_JOURNALD_LOGGING "Build with support for native journald log output" OFF)
 add_feature_info("VAST_ENABLE_JOURNALD_LOGGING" VAST_ENABLE_JOURNALD_LOGGING
                  "build with support for native journald log output.")
-if (VAST_ENABLE_JOURNALD_LOGGING)
-  find_package(PkgConfig)
-
-  pkg_check_modules(LIBSYSTEMD REQUIRED libsystemd)
-
-  find_path(SYSTEMD_INCLUDE_DIRS
-    NAMES systemd/sd-journal.h
-    PATHS ${LIBSYSTEMD_INCLUDE_DIRS}
-  )
-
-  find_library(SYSTEMD_LIBRARIES
-    NAMES systemd
-    PATHS ${LIBSYSTEMD_LIBRARY_DIRS}
-  )
-endif ()
 
 # -- log level -----------------------------------------------------------------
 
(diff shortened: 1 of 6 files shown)


libvast/src/logger.cpp
```diff
@@ -207,14 +213,41 @@ bool setup_spdlog(const vast::invocation& cmd_invocation,
...
+    = caf::get_or(cfg_file, "vast.console-sink-type", default_sink_type);
+  auto console_sink = [&]() -> spdlog::sink_ptr {
+    if (sink_type == "stderr") {
+      auto stderr_sink
+        = std::make_shared<spdlog::sinks::ansicolor_stderr_sink_mt>(log_color);
+      return stderr_sink;
+    } else if (sink_type == "journald") {
+#ifndef VAST_ENABLE_JOURNALD_LOGGING
+      std::cerr << "Cannot use 'journald' sink, vast was built without systemd "
+                   "support\n";
```
mavam, line 228:
```suggestion
      std::cerr << "Cannot use 'journald' sink, VAST was built without systemd "
                   "support\n";
```

What's our policy regarding casing of error messages? I think we use lowercase in the logger, but what do we do on `cerr`?

dominiklohmann: We also use lowercase to be consistent with all other uses of it.

libvast/src/logger.cpp
```diff
@@ -207,14 +213,41 @@ bool setup_spdlog(const vast::invocation& cmd_invocation,
...
+#else
+      auto spdlog_sink = std::make_shared<spdlog::sinks::systemd_sink_mt>();
+      return std::static_pointer_cast<spdlog::sinks::sink>(spdlog_sink);
+#endif
+    } else if (sink_type == "syslog") {
+      auto syslog_sink = std::make_shared<spdlog::sinks::syslog_sink_mt>(
+        "vast", /*options = */ 0, LOG_USER, /*enable_formatting = */ true);
+      return std::static_pointer_cast<spdlog::sinks::sink>(syslog_sink);
+    } else {
+      std::cerr << "Illegal vast.console-sink-type " << sink_type << "\n";
```
dominiklohmann, line 239: This error message is not helpful for users.

libvast/src/systemd.cpp
```diff
@@ -15,6 +17,27 @@ namespace vast::systemd {
 
 #if VAST_LINUX
 
+bool connected_to_journal() {
```
dominiklohmann, line 20: Should this return a `caf::error`? After all, there are many ways this can go wrong.

libvast/src/systemd.cpp
```diff
@@ -15,6 +17,27 @@ namespace vast::systemd {
...
+  auto journal_env = detail::locked_getenv("JOURNAL_STREAM");
+  if (!journal_env)
+    return false;
+  size_t device_number = 0;
+  char colon = 0;
+  size_t inode_number = 0;
+  auto parser = parsers::u64 >> parsers::ch<':'> >> parsers::u64;
+  if (parser(*journal_env, device_number, colon, inode_number)) {
```
dominiklohmann, line 28:
```suggestion
  size_t inode_number = 0;
  auto parser = parsers::u64 >> ':' >> parsers::u64;
  if (parser(*journal_env, device_number, inode_number)) {
```

Either just the char, or wrap it explicitly in `unused(<parser>)`.

libvast/src/systemd.cpp
```diff
@@ -15,6 +17,27 @@ namespace vast::systemd {
...
+    VAST_WARN("could not parse systemd environment variable "
+              "$JOURNAL_STREAM='{}'",
+              *journal_env);
+    return false;
+  }
+  struct stat buf = {};
+  // TODO: errno-ify this
+  auto stderrfd = ::fileno(stderr);
+  ::fstat(stderrfd, &buf);
+  return buf.st_dev == device_number && buf.st_ino == inode_number;
```
dominiklohmann, line 38: Is this doable with `std::filesystem` instead?

dominiklohmann (commented):
Really liking this. Left some review comments.

libvast/src/systemd.cpp
lava: I highly doubt it, `std::filesystem` is too high-level to know about device or inode numbers.

libvast/src/systemd.cpp
mavam: Background: a string literal has `unused` as default attribute. The `ch` parsers have a `char` as attribute and are typically only use for values to be extracted.

libvast/CMakeLists.txt
```diff
@@ -140,6 +140,27 @@ if (VAST_ENABLE_ARROW)
   endif ()
 endif ()
 
+# -- libsystemd ----------------------------------------------------------------
+
+option(VAST_ENABLE_JOURNALD_LOGGING "Build with support for native journald log output" OFF)
```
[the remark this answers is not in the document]
dominiklohmann: Alternative: Use `find_package(systemd REQUIRED)` if this is enabled.

libvast/src/systemd.cpp
dominiklohmann: We discussed this; not necessary.

libvast/src/logger.cpp
```diff
@@ -207,14 +213,43 @@ bool setup_spdlog(const vast::invocation& cmd_invocation,
...
+#else
+      auto spdlog_sink = std::make_shared<spdlog::sinks::systemd_sink_mt>();
+      return std::static_pointer_cast<spdlog::sinks::sink>(spdlog_sink);
+#endif
+    } else if (sink_type == "syslog") {
+      auto syslog_sink = std::make_shared<spdlog::sinks::syslog_sink_mt>(
+        "vast", /*options = */ 0, LOG_USER, /*enable_formatting = */ true);
+      return std::static_pointer_cast<spdlog::sinks::sink>(syslog_sink);
+    } else {
+      std::cerr << "Illegal vast.console-sink value '" << sink_type << "', "
```
dominiklohmann, line 239:
```suggestion
      std::cerr << "illegal vast.console-sink value '" << sink_type << "', "
```

dominiklohmann: Might also want to consider using `fmt::print(stderr, ...)` here.

dominiklohmann (approved):
Approving modulo one log fixup and cmake-format.

libvast/src/logger.cpp
lava: Probably a good idea, but there are already a few other instances of `std::cerr`-based logging in the file, so I'll keep it as-is for now.

PR #1699, opened 2021-05-31
Add option for configurable post-start hooks

dominiklohmann:

The new option `vast.start.commands` allows for specifying an ordered list of VAST commands that run inside the node after successful startup. This is useful for declarative deployments that run sources or matchers (via the proprietary Live Matching plugin) in the node.

### :memo: Checklist

Try locally. This can hardly be tested, sadly, since we do not have unit tests for the start command and cannot reliably test remote sources in the integration test framework. I played around with the following:

```yaml
vast:
  start:
    hooks:
      # Requires PCAP plugin
      - spawn source pcap -i en0
      # Requires NetFlow  plugin
      - spawn source netflow -i :4739/udp
      # Requires Live Matching plugin
      - matcher start --match-types=address ips
```

### DIFF
--- a/libvast/src/system/application.cpp
+++ b/libvast/src/system/application.cpp
@@ -379,6 +379,9 @@ auto make_start_command() {
     "start", "starts a node", documentation::vast_start,
     opts("?vast.start")
       .add<bool>("print-endpoint", "print the client endpoint on stdout")
+      .add<std::vector<std::string>>("commands", "an ordered list of commands "
+                                                 "to run inside the node after "
+                                                 "starting")
       .add<size_t>("disk-budget-check-interval", "time between two disk size "
                                                  "scans")
       .add<std::string>("disk-budget-check-binary",
--- a/libvast/src/system/start_command.cpp
+++ b/libvast/src/system/start_command.cpp
@@ -123,7 +123,7 @@ caf::message start_command(const invocation& inv, caf::actor_system& sys) {
         auto current = std::string{};
         while (tokenizer >> std::quoted(current))
           cli.push_back(std::move(current));
-        VAST_INFO("running post-start command [{}]", fmt::join(cli, ", "));
+        VAST_INFO("running post-start command {}", command);
         auto hook_invocation = parse(*root, cli.begin(), cli.end());
         if (!hook_invocation)
           return caf::make_message(hook_invocation.error());
--- a/libvast/src/system/start_command.cpp
+++ b/libvast/src/system/start_command.cpp
@@ -11,6 +11,8 @@
 #include "vast/command.hpp"
 #include "vast/config.hpp"
 #include "vast/data.hpp"
+#include "vast/detail/settings.hpp"
+#include "vast/system/application.hpp"
 #include "vast/systemd.hpp"
 
 #include <caf/actor_system_config.hpp>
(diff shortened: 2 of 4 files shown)


changelog/unreleased/features/1699--post-start-hooks.md
```diff
@@ -0,0 +1,4 @@
+The new option `vast.start.hooks` allows for specifying an ordered list of VAST
+commands that run inside the node after successful startup. This is useful for
+declarative deployments that run sources or matchers (via the proprietary Live
+Matching plugin) in the node.
```
mavam, line 4: The new option `vast.start.commands` allows for specifying an ordered list of
VAST commands that run after successful startup. The effect is the same as
first starting a node, and then using another VAST client to issue commands.
This is useful for commands that have side effects that cannot be expressed
through the config file, e.g., starting a source inside the VAST server that
listens on a socket or reads packets from a network interface.

libvast/src/system/application.cpp
```diff
@@ -379,6 +379,9 @@ auto make_start_command() {
     "start", "starts a node", documentation::vast_start,
     opts("?vast.start")
       .add<bool>("print-endpoint", "print the client endpoint on stdout")
+      .add<std::vector<std::string>>("hooks", "an ordered list of commands to "
```
mavam, line 382:
```suggestion
      .add<std::vector<std::string>>("commands", "an ordered list of commands to "
```

libvast/src/system/start_command.cpp
```diff
@@ -104,6 +106,33 @@ caf::message start_command(const invocation& inv, caf::actor_system& sys) {
   // A single line of output to publish out address for scripts.
   if (caf::get_or(inv.options, "vast.start.print-endpoint", false))
     std::cout << listen_addr << std::endl;
+  if (auto hooks = caf::get_if<std::vector<std::string>>(&inv.options,
+                                                         "vast.start.hooks")) {
```
mavam, line 110:
```suggestion
                                                         "vast.start.commands")) {
```

vast.yaml.example
```diff
@@ -128,6 +128,11 @@ vast:
     # connections. This comes in handy when letting the OS choose an available
     # random port, i.e., when specifying 0 as port value.
     print-endpoint: false
+    # An ordered list of commands to run inside the node after starting.
+    # As an example, this is how to configure an auto-starting PCAP source that
+    # listens on the interface 'en0' and lives inside the VAST node.
+    #hooks:
```
mavam, line 134:
```suggestion
    #commands:
```

libvast/src/system/start_command.cpp
```diff
@@ -104,6 +106,33 @@ caf::message start_command(const invocation& inv, caf::actor_system& sys) {
   // A single line of output to publish out address for scripts.
   if (caf::get_or(inv.options, "vast.start.print-endpoint", false))
     std::cout << listen_addr << std::endl;
+  if (auto hooks = caf::get_if<std::vector<std::string>>(&inv.options,
```
mavam, line 109:
```suggestion
  if (auto start_commands = caf::get_if<std::vector<std::string>>(&inv.options,
```

mavam (changes requested):
I think "hook" is too generic. We can be more specific here: it's really just a sequence of commands and we can call them what they are.

mavam (approved):
This seems like a good middleground to get experience with post-hoc configuration of VAST using commands with side effects.

That said, I cringe a bit in that it could be more declarative to fit the surrounding configuration. I don't think that's a dealbreaker though; this is better than the current state.

@tobim @lava any objections from your side?

PR #1702, opened 2021-06-01
Make unix dgram metrics sink connectionless

tobim:
The previous way of setting up the UDS datagram metrics relied on a fake connection that made it possible to use the iostream abstraction. However this fake connection would get destroyed if the listening end deleted it's socket.

This commit introduces a wrapper around sockaddr_un and the client fd and uses sendto instead of write.

### :memo: Checklist

Code review by file.

Test by running VAST with
```
vast:
  enable-metrics: true
  metrics:
    uds-sink:
      enable: true
      path: "/tmp/vast-metrics.sock"
      type: "datagram"
```
And run and restart `socat UNIX-RECV:/tmp/vast-metrics.sock ` while the VAST node is running.

### DIFF
--- a/libvast/src/detail/posix.cpp
+++ b/libvast/src/detail/posix.cpp
@@ -60,8 +60,24 @@ int uds_accept(int socket) {
   return fd;
 }
 
+uds_datagram_sender::uds_datagram_sender(uds_datagram_sender&& other) noexcept
+  : src_fd{other.src_fd}, dst{other.dst} {
+  // Invalidate the original copy to avoid closing when `other` is destroyed.
+  other.src_fd = -1;
+}
+
+uds_datagram_sender&
+uds_datagram_sender::operator=(uds_datagram_sender&& other) noexcept {
+  src_fd = other.src_fd;
+  // Invalidate the original copy to avoid closing when `other` is destroyed.
+  other.src_fd = -1;
+  dst = other.dst;
+  return *this;
+}
+
 uds_datagram_sender::~uds_datagram_sender() {
-  ::unlink(src_path.c_str());
+  if (src_fd != -1)
+    ::close(src_fd);
 }
 
 caf::expected<uds_datagram_sender>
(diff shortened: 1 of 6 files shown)


libvast/src/detail/posix.cpp
```diff
@@ -60,6 +60,43 @@ int uds_accept(int socket) {
...
+
+caf::expected<uds_datagram_sender>
+uds_datagram_sender::make(const std::string& path) {
+  auto result = uds_datagram_sender{};
+  result.src_fd = ::socket(AF_UNIX, SOCK_DGRAM, 0);
+  if (result.src_fd < 0)
+    return caf::make_error(ec::system_error,
+                           "failed to obtain an AF_UNIX DGRAM socket: {}",
+                           ::strerror(errno));
+  // TODO: Consider creating this inside the database directory instead.
```
dominiklohmann, line 75: This is a must, actually. On hardened setups this will not be writable.

libvast/vast/detail/posix.hpp
```diff
@@ -18,6 +22,27 @@ namespace vast::detail {
 
 enum class socket_type { datagram, stream, fd };
 
+/// Holds the necessary state to send unix datagrams to a destination socket.
+struct uds_datagram_sender {
+  /// The destructor unlinks the src socket.
+  ~uds_datagram_sender();
```
dominiklohmann, line 28: I struggle to understand the contract of this struct. Is it safely copyable? Is it movable? 

Can we express this contract more clearly by using a scope guard for the file descriptor?

libvast/vast/detail/posix.hpp
```diff
@@ -18,6 +22,27 @@ namespace vast::detail {
...
+
+  /// Helper function to create a sender.
+  static caf::expected<uds_datagram_sender> make(const std::string& path);
```
dominiklohmann, line 31: Is this not a `std::filesystem::path`?

libvast/vast/detail/posix.hpp
```diff
@@ -18,6 +22,27 @@ namespace vast::detail {
...
+
+  /// Send the content of `data to `dst`.
+  caf::error send(span<char> data);
```
dominiklohmann, line 34: Should be `span<const std::byte>` to be in line with our other code.

libvast/vast/accountant/config.hpp
```diff
@@ -41,6 +41,13 @@ struct accountant_config {
   self_sink self_sink;
   file_sink file_sink;
   uds_sink uds_sink;
+
+  template <class Inspector>
+  friend auto inspect(Inspector& f, accountant_config& x) {
+    f(caf::meta::type_name("vast::system::accountant_config"),
```
dominiklohmann, line 47:
```suggestion
    f(caf::meta::type_name("vast.system.accountant_config"),
```

We don't have a consistent style here throughout the code base, but this I think we use most often.

libvast/src/system/node.cpp
```diff
@@ -335,6 +335,7 @@ spawn_accountant(node_actor::stateful_pointer<node_state> self,
   auto cfg = to_accountant_config(metrics_opts);
   if (!cfg)
     return cfg.error();
+  VAST_INFO("accountant config is {}", *cfg);
```
dominiklohmann, line 338: That looks like a leftover debug message.

libvast/src/system/accountant.cpp
```diff
@@ -207,13 +224,26 @@ struct accountant_state_impl {
...
+          VAST_INFO("{} could not open {} for metrics: {}", self,
+                    cfg.uds_sink.path, s.error());
+        }
       } else {
-        VAST_INFO("{} could not open {} for metrics: {}", self,
-                  cfg.uds_sink.path, s.error());
+        auto s
+          = detail::make_output_stream(cfg.uds_sink.path, cfg.uds_sink.type);
+        if (s) {
+          VAST_INFO("{} writing metrics to {}", self, cfg.uds_sink.path);
```
dominiklohmann, line 241:
```suggestion
          VAST_INFO("{} writes metrics to {}", self, cfg.uds_sink.path);
```

libvast/src/system/accountant.cpp
```diff
@@ -207,13 +224,26 @@ struct accountant_state_impl {
...
     }
     if (start_uds_sink) {
-      auto s = detail::make_output_stream(cfg.uds_sink.path, cfg.uds_sink.type);
-      if (s) {
-        VAST_INFO("{} writing metrics to {}", self, cfg.uds_sink.path);
-        uds_sink = std::move(*s);
+      if (cfg.uds_sink.type == detail::socket_type::datagram) {
+        auto s = detail::uds_datagram_sender::make(cfg.uds_sink.path);
+        if (s) {
+          VAST_INFO("{} writing metrics to {}", self, cfg.uds_sink.path);
```
dominiklohmann, line 230:
```suggestion
          VAST_INFO("{} writes metrics to {}", self, cfg.uds_sink.path);
```

libvast/src/system/accountant.cpp
```diff
@@ -207,13 +224,26 @@ struct accountant_state_impl {
...
+          // uds_datagram_sink = std::move(*s);
+          uds_datagram_sink = std::make_unique<detail::uds_datagram_sender>(*s);
```
dominiklohmann, line 232: Why do you need the `make_unique` here?

dominiklohmann (changes requested):

libvast/vast/detail/posix.hpp
tobim: Yes it is, but the rest of `posix.hpp` doesn't use filesystem either, and it is better to keep local consistency.

libvast/src/detail/posix.cpp
tobim: I replaced this with a temporary location that should even be available with a very hardened systemd unit.

libvast/vast/detail/posix.hpp
tobim: I changed the implementation to make it clear that it is move only.

libvast/vast/detail/posix.hpp
tobim: The underlying API works with char, I don't see a reason to bring unnecessary casting upon ourselves.

libvast/src/system/accountant.cpp
tobim: The variable is of type `unique_ptr`, and there is no move assigment operator from `value_type`. It could be turned into an optional, but since `uds_datagram_sender` is move only we wouldn't gain anything from that.

libvast/src/detail/posix.cpp
```diff
@@ -60,6 +60,85 @@ int uds_accept(int socket) {
...
+  result.src_fd = ::socket(AF_UNIX, SOCK_DGRAM, 0);
+  if (result.src_fd < 0)
+    return caf::make_error(
+      ec::system_error,
+      "failed to obtain an AF_UNIX DGRAM socket: ", ::strerror(errno));
+  // Create a unique temporary directory for a place to bind the sending side
+  // to. There is no mktemp variant for sockets, so that is unfortunately
+  // necessary.
+  // NOTE: The temporary directory will be removed at the end of this function.
+  char mkd_template[] = u8"/tmp/vast-XXXXXX\0socket";
```
dominiklohmann, line 95: Why the UTF-8 string literal? Can you add a comment for the future reader?

dominiklohmann (approved):
This seems to work fine in a really short test, and the code looks good as well now. Just had one small question.

libvast/src/detail/posix.cpp
tobim: That is a leftover that I forgot to remove.

PR #1703, opened 2021-06-01
Rework the plugin loading logic

dominiklohmann:

This PR contains a set of changes related to how we load plugins.
1. `--plugins` and `--plugin-dirs` options may now be specified on the command line as well as the configuration under `vast.plugins` and `vast.plugin-dirs` respectively.
2. We no longer load static plugins by default.
3. The CMake build option `VAST_ENABLE_PLUGIN_AUTOLOADING` no longer exists.
4. Adding `bundled` to `vast.plugins` causes all bundled plugins, i.e., static or dynamic plugins built alongside VAST, to be loaded.
5. Adding `all` to `vast.plugins` causes all bundled *and* external plugins to be loaded.

### :memo: Checklist

Review code file-by-file.

Regarding the design: I went over this with @mavam and we agreed this was a good UX improvement. I will wait until after the initial review with writing the documentation for this, though.

To test this change I recommend the following steps:

1. Test various combinations of the new plugin option with a bundled plugin in a standard location.
   ```bash
   cmake -B build.vast -D "VAST_PLUGINS=${PWD}/plugins/pcap"
   cmake --build build.vast
   cmake --install build.vast --prefix "${PWD}/install.vast"
   vast -v --plugins=... version
   ```
2. Test various combinations of the new plugin option with an external plugin in a non-standard location and a bundled plugin in a standard location.
   ```bash
   VAST_DIR="${PWD}/install.vast" -S example/plugins/analyzer -B build.analyzer
   cmake --build build.analyzer
   vast -v --plugins=... --plugin-dirs="${PWD}/build.analyzer/lib/vast/plugins" version
   ```
3. Test various combinations of the new plugin option with an external plugin in a non-standard location and a bundled plugin in a standard location.
   ```bash
   cmake --install build.analyzer --prefix "${PWD}/install.vast"
   vast -v --plugins=... version
   ```


vast/vast.cpp
```diff
@@ -37,89 +37,110 @@
...
+                           return x == plugin_name;
+                         }))
+          plugin_paths_or_names.push_back(std::move(plugin_name));
+    // Check if any of the specified plugins is already loaded and remove it
+    // from the list of specified plugins if that's the case. If bare mode is
+    // enabled, unload plugins that were not specified explicitly to avoid
+    // accidentaly running with undesired plugins.
+    for (auto& plugin : plugins) {
+      VAST_ASSERT(plugin.type() != plugin_ptr::type::dynamic,
+                  "dynamic plugins must be loaded at this point");
```
lava, line 75: Is the message correct?

vast/vast.cpp
```diff
@@ -37,89 +37,110 @@
...
+      for (auto&& plugin_name : std::vector<std::string>{VAST_BUNDLED_PLUGINS})
+        if (std::none_of(plugin_paths_or_names.begin(),
+                         plugin_paths_or_names.end(), [&](auto&& x) {
+                           return x == plugin_name;
+                         }))
+          plugin_paths_or_names.push_back(std::move(plugin_name));
+    // Check if any of the specified plugins is already loaded and remove it
+    // from the list of specified plugins if that's the case. If bare mode is
+    // enabled, unload plugins that were not specified explicitly to avoid
+    // accidentaly running with undesired plugins.
```
lava, line 72: accidentally

vast/vast.cpp
```diff
@@ -37,89 +37,110 @@
...
+    = caf::get_or(cfg, "vast.plugins", std::vector<std::string>{});
+  const auto check_and_remove_plugin = [&](const char* name) {
+    auto bundled = std::remove(plugin_paths_or_names.begin(),
+                               plugin_paths_or_names.end(), name);
+    const auto result = bundled != plugin_paths_or_names.end();
+    plugin_paths_or_names.erase(bundled, plugin_paths_or_names.end());
+    return result;
+  };
+  if (check_and_remove_plugin("all")) {
+    // The special plugin name "bundled" enables all plugins, i.e., plugins
```
lava, line 57: s/bundled/all/

vast/vast.cpp
```diff
@@ -37,89 +37,110 @@
...
+    }
+    // Remove all possibly invalidated static plugins.
+    plugins.erase(std::remove_if(plugins.begin(), plugins.end(),
+                                 [](const auto& plugin) {
+                                   return !plugin;
+                                 }),
+                  plugins.end());
+    // Load the specified dynamic plugins.
+    for (const auto& plugin_path_or_name : plugin_paths_or_names) {
+      if (auto loaded_plugin = detail::load_plugin(plugin_path_or_name, cfg)) {
```
lava, line 94: I'm confused, isn't this still in the "else" part, i.e. only hit when we did not specify "all" as plugin name?

libvast/src/system/configuration.cpp
```diff
@@ -70,9 +70,30 @@ caf::error configuration::parse(int argc, char** argv) {
...
+  // needs to happen before the regular parsing of the command line since
+  // plugins may add additional commands.
+  auto is_not_plugin_opt = [](auto& x) {
+    return !detail::starts_with(x, "--plugins=")
+           && !detail::starts_with(x, "--plugin-dirs=");
+  };
+  auto plugin_opt = std::stable_partition(
+    command_line.begin(), command_line.end(), is_not_plugin_opt);
+  auto plugin_args = std::vector<std::string>{};
+  std::move(plugin_opt, command_line.end(), std::back_inserter(plugin_args));
```
lava, line 83: I guess this was supposed to call a function other than `std::move()`?

vast/vast.cpp
```diff
@@ -37,89 +37,110 @@
 #include <string>
 #include <vector>
 
-using namespace vast;
-using namespace vast::system;
+namespace vast {
+
+caf::expected<std::vector<std::filesystem::path>>
+load_plugins(system::configuration& cfg) {
```
lava, line 43: Why do we need this function in addition to `detail::load_plugins()`?

changelog/unreleased/breaking-changes/1703--flexible-plugin-loading.md
```diff
@@ -0,0 +1 @@
+VAST no longer loads static plugins by default.
```
lava, line 1: Can you point me to the corresponding code change that this refers to?

libvast/src/detail/load_plugin.cpp
```diff
@@ -124,4 +124,113 @@ load_plugin(const std::filesystem::path& path_or_name,
...
+    = caf::get_or(cfg, "vast.plugins", std::vector<std::string>{});
+  if (paths_or_names.empty())
+    return loaded_plugin_paths;
+  auto& plugins = plugins::get();
+  auto load_specified_plugins = [&]() -> caf::error {
+    for (const auto& plugin_path_or_name : paths_or_names) {
+      if (auto loaded_plugin = load_plugin(plugin_path_or_name, cfg)) {
+        auto&& [path, plugin] = std::move(*loaded_plugin);
+        loaded_plugin_paths.push_back(std::move(path));
+        plugins.push_back(std::move(plugin));
```
lava, line 141: I know out-references are a code smell, but *implicit* out-references are surely even smellier.

lava (commented):
Just a quick first look.

Maybe you can expand the description a bit to include what the specific issues were that you and @mavam were trying to solve with this redesign, otherwise its hard to tell if it succeeds at that or not.

As a general feeling, it seems like expanding the `bundled` and `all` keywords into a list of plugins could be done outside the routines that handle the actual loading of plugins, to disentangle the code.

vast/vast.cpp
dominiklohmann: Accidental duplication, sorry about that. Removed it.

changelog/unreleased/breaking-changes/1703--flexible-plugin-loading.md
dominiklohmann: https://github.com/tenzir/vast/pull/1703/files#diff-52e07dd1a5a1de790b1d57e81d445a22a0b8bb6791a5f13f44ca61b8ae1ceb2fL85-L86 vs. https://github.com/tenzir/vast/pull/1703/files#diff-4599665b12bc79bea311152e2c968a74ce853bc06ac316c18928272485017251R215-R216

libvast/src/system/configuration.cpp
dominiklohmann: No, this is correct. There is both a `std::move` in `<utility>` and one in `<algorithm>`: https://en.cppreference.com/w/cpp/algorithm/move

libvast/src/detail/load_plugin.cpp
dominiklohmann: I renamed `loaded_plugin_paths` to `result` to make that clearer, and replaced `plugins` with the mutable singleton.

vast/vast.cpp
dominiklohmann: Missing a "not"

dominiklohmann:
> Maybe you can expand the description a bit to include what the specific issues were that you and @mavam were trying to solve with this redesign, otherwise its hard to tell if it succeeds at that or not.

A few things that were all interrelated:
- Do no longer enable static plugins by default to treat both types of bundled plugins equally.
- Make it possible to use `--plugins=` and `--plugin-dirs=` on the command line.
- Add an option to load all bundled plugins
- Add an option to load all installed plugins including external plugins

> As a general feeling, it seems like expanding the `bundled` and `all` keywords into a list of plugins could be done outside the routines that handle the actual loading of plugins, to disentangle the code.

It cannot with the current behavior, unless `all` should also hard fail for libraries that cannot be loaded.

mavam:
> As a general feeling, it seems like expanding the `bundled` and `all` keywords into a list of plugins could be done outside the routines that handle the actual loading of plugins, to disentangle the code.

Mixing `--plugins=all` with `--plugins=foo,bar,baz` may confuse different things. Putting into separate options could be alternative, e.g.,
- Explicit plugins: `--plugins=foo,bar,baz`
- Plugin groups: `--load-bundled-plugins` or `--load-all-plugins` (modulo option naming)

dominiklohmann:
> * Explicit plugins: `--plugins=foo,bar,baz`
> * Plugin groups: `--load-bundled-plugins` or `--load-all-plugins` (modulo option naming)

I'd like to avoid doing that, as adding more options that need special-case handling in the CAF configuration is a pain to maintain. I also think @lava was talking about the code itself more, where we can expand the `all` and `bundled` keywords into a list of plugin file in-place. I actually thought about this some last evening and came up with an idea of how to further refactor this with that in mind.

libvast/src/plugin.cpp
```diff
@@ -40,18 +48,188 @@ std::string to_string(plugin_version x) {
...
+    if (auto path = resolve_plugin_name(plugin_dirs, path_or_name))
+      path_or_name = path->string();
+    else
+      return std::move(path.error());
+  }
+  // Step 6: Deduplicate plugin paths.
+  auto paths = detail::stable_set<std::string>{};
+  paths.insert(std::make_move_iterator(paths_or_names.begin()),
+               std::make_move_iterator(paths_or_names.end()));
+  // Step 7: Load plugins.
```
lava, line 176: I think here would be a natural place to split up the function (with step 4 pulled out) since the parts before and after are unrelated; one function to resolve the list of plugins to be loaded and one that does the loading + ordering.

changelog/unreleased/features/1703--flexible-plugin-loading.md
```diff
@@ -0,0 +1,9 @@
+The options `vast.plugins` and `vast.plugin-dirs` may now be specified on the
+command line as well as the configuration. Use the options `--plugins` and
+`--plugin-dirs` respectively.
+
+The build-time CMake option `VAST_ENABLE_PLUGIN_AUTOLOADING` is superseded by a
```
lava, line 5: Not sure if we should mention VAST_ENABLE_PLUGIN_AUTOLOADING, since it was never a public-facing option.

vast.yaml.example
```diff
@@ -62,8 +62,10 @@ vast:
   # Additional directories to load plugins specified using `vast.plugins` from.
   plugin-dirs: []
   # The plugins to load at startup. For relative paths, VAST tries to find the
-  # files in the specified `vast.plugin-dirs`.
-  # Note: Add `example` or `/path/to/libexample.so` to load the example plugin.
+  # files in the specified `vast.plugin-dirs`. The special values 'bundled' and
+  # 'all' enable autoloading of bundled and all plugins respectively.
```
lava, line 66: This should probably clarify what "all" plugins means, ie. that it will attempt to load all shared libraries matching `libvast-plugin-*` from the configured `plugin-dirs`.

lava (approved):
Looks generally good to me; just a few stylistic comments below.

Also, I think it would be good to capture the motivation for this PR that you explained in a video call somewhere (maybe a new "feature" changelog entry?), since your previous answer just gave a summary of the PR contents.

libvast/src/plugin.cpp
dominiklohmann: I'm not sure how to extract step 4 without duplicating some of its logic in later steps. Let's talk tomorrow about this. Other than that I've incorporated all feedback.

PR #1707, opened 2021-06-03
Use libunwind for the backtrace if available

tobim:

`libunwind` gives us a more readable backtrace, so we should use that if possible.

### :memo: Checklist

By file.

Note: I don't think this warrants a changelog entry, but I can write one if desired.


dominiklohmann (commented):
Pushed a commit that makes this work on macOS and also does some minor fixups.

We should consider installing libunwind in CI and in the Dockerfile.

tobim:
The commits from @dominiklohmann are fine from my point of view.

dominiklohmann (approved):

PR #1712, opened 2021-06-08
Handle quoted strings in CSV parser

dominiklohmann:

This change makes it possible to import CSV data containing quoted strings that contain the configured separator. This works for both header and fields.

### :memo: Checklist

Read code diff. Check provided unit tests.

### DIFF
--- a/libvast/src/format/csv.cpp
+++ b/libvast/src/format/csv.cpp
@@ -293,6 +293,8 @@ struct csv_parser_factory {
 
   template <class T>
   result_type operator()(const T& t) const {
+    [[maybe_unused]] const auto field
+      = parsers::qqstr | +(parsers::any - opt_.set_separator);
     if constexpr (std::is_same_v<T, alias_type>) {
       return caf::visit(*this, t.value_type);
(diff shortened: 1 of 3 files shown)


libvast/src/format/csv.cpp
```diff
@@ -293,6 +293,9 @@ struct csv_parser_factory {
 
   template <class T>
   result_type operator()(const T& t) const {
+    [[maybe_unused]] const auto field
+      = parsers::qqstr
+        | +(parsers::any - opt_.set_separator - opt_.kvp_separator);
```
lava, line 298: Why are we excluding `set_separator` and `kvp_separator` here, as opposed to just `separator` which was excluded in the ad-hoc parsers before?

dominiklohmann: Good spot. It's a leftover from debugging, will fix.

dominiklohmann: Fixed.

lava (approved):
Looks great! I didn't compile it locally, but the unit tests are enough to satisy me that it works as expected and doesn't break anything.

PR #1713, opened 2021-06-08
Relax requirements for stream sink in make_source

dominiklohmann:

Requiring the IMPORTER actor is needlessly strict here. This change allows for using `make_source` in plugins with a custom stream sink.

### :memo: Checklist

n/t


lava (approved):
The best part about using typed actors is that I feel confident approving this based on CI results without grepping through source to see if we might have missed a call site that uses an importer-specific handler.

PR #1716, opened 2021-06-14
Still consider install plugindir in bare mode

tobim:

The contents of the install dir should be considered as vendored. In case of relocatable installations that means the path will not contain unintentional plugins. In case of non-relocatable we don't support running VAST from any place that is not the install location.

We should also look for plugins in the install plugindir last, after the locations set in the runtime configurations.

No changelog entry because `--bare-mode` is an undocumented development feature in the first place.

### DIFF
--- a/libvast/src/plugin.cpp
+++ b/libvast/src/plugin.cpp
@@ -58,7 +58,6 @@ get_plugin_dirs(const caf::actor_system_config& cfg) {
     for (auto&& path : detail::split(*vast_plugin_directories, ":"))
       result.insert({path});
   if (!bare_mode) {
-    result.insert(detail::install_plugindir());
     if (auto home = detail::locked_getenv("HOME"))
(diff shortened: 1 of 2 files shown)


dominiklohmann (commented):
Good catch.

I think the install plugin dir should still have a higher precedence than the user-specified plugin dirs, i.e., moved above the bare mode check rather than below it.

tobim:
> I think the install plugin dir should still have a higher precedence than the user-specified plugin dirs, i.e., moved above the bare mode check rather than below it.

Doesn't that make it unnecessarily complicated to load a modified version of a plugin that is placed somewhere outside the installed location? I.e. If a plugin is installed in both `/custom` and `/vast/install` I would think the user wants to load the custom one.

dominiklohmann (approved):
Pushed a commit with a precedence change for `VAST_PLUGIN_DIRS` as discussed in a call with @tobim. Approving his side of things—please merge if you approve of mine, or make changes as you see fit.

PR #1718, opened 2021-06-15
Fix shutdown hang in sources on SIGTERM/SIGINT

dominiklohmann:

The new TRANSFORMER actor that always resides in the import process must be shut down when the SOURCE shuts down, as the SOURCE is the sole owner of the TRANSFORMER. Before this change, the DATAGRAM SOURCE shut down the stream of the TRANSFORMER, which caused it to quit somehow, but the SOURCE didn't propagate the shutdown to the TRANSFORMER at all, causing the import process to hang until a repeated SIGTERM/SIGINT led to a hard kill.

### :memo: Checklist

Try locally.

### DIFF
--- a/changelog/unreleased/bug-fixes/1718--source-shutdown-hang.md
+++ b/changelog/unreleased/bug-fixes/1718--source-shutdown-hang.md
@@ -1,0 +1,1 @@
+Import processes now shut down gracefully on SIGINT/SIGKILL again.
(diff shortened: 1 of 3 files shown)


lava (approved):
The change looks correct to me in theory; if you also tested it in practice by sending SIGTERM/SIGINT to a running source this should be good to go.

dominiklohmann:
> if you also tested it in practice by sending SIGTERM/SIGINT

I tested all four combinations of the two signals and running/idle sources.

PR #1720, opened 2021-06-15
Implement partition-local stores

lava:
- Add new functions to erase data from segments
- Add a method to create an empty chunk
- Add a new 'plugins::find()' function
- Add new plugin type 'store_plugin'
- Add the ability to use partition-local stores

This adds the ability to use separate store backends for each partition in VAST.

A new "partition-local" backend is provided that just provides a simplified version of the current segment store which is restricted to a single segment.

### :memo: Checklist


libvast/src/system/index.cpp
```diff
@@ -603,18 +634,30 @@ index(index_actor::stateful_pointer<index_state> self, store_actor store,
...
+  if (partition_local_stores)
+    VAST_VERBOSE("{} uses partition-local stores instead of the global archive",
+                 self);
   if (dir != meta_index_dir)
     VAST_VERBOSE("{} uses {} for meta index data", self, meta_index_dir);
   // Set members.
   self->state.self = self;
   self->state.accept_queries = true;
-  self->state.store = std::move(store);
+  self->state.store_plugin_
```
lava, line 645: s/store_plugin_/store_plugin/

libvast/src/system/index.cpp
```diff
@@ -319,15 +333,30 @@ void index_state::create_active_partition() {
...
+  index_opts["partition-local-stores"] = partition_local_stores;
   // These options must be kept in sync with vast/address_synopsis.hpp and
   // vast/string_synopsis.hpp respectively.
   auto synopsis_options = caf::settings{};
   put(synopsis_options, "max-partition-size", partition_capacity);
   put(synopsis_options, "address-synopsis-fp-rate", meta_index_fp_rate);
   put(synopsis_options, "string-synopsis-fp-rate", meta_index_fp_rate);
+  const auto* store_name = store_plugin_->name();
+  auto builder_and_header = store_plugin_->make_store_builder(id);
+  VAST_ASSERT(builder_and_header); // FIXME
```
lava, line 345: Replace by `self->quit()` and return

libvast/src/system/index.cpp
```diff
@@ -603,18 +634,30 @@ index(index_actor::stateful_pointer<index_state> self, store_actor store,
   VAST_VERBOSE("{} initializes index in {} with a maximum partition "
                "size of {} events and {} resident partitions",
                self, dir, partition_capacity, max_inmem_partitions);
+  if (partition_local_stores)
+    VAST_VERBOSE("{} uses partition-local stores instead of the global archive",
```
lava, line 638: s/global archive/archive/

libvast/src/system/global_segment_store.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "vast/plugin.hpp"
+#include "vast/system/archive.hpp"
+#include "vast/system/node.hpp"
+#include "vast/system/node_control.hpp"
+
+namespace vast::system {
+
+// This store plugin wraps the global "archive" so we can
+// use a unified API in the transition period.
+class global_store_plugin final : public virtual store_plugin {
```
lava, line 18: s/global_store_plugin/archive_store_plugin/

libvast/src/system/global_segment_store.cpp
```diff
@@ -0,0 +1,58 @@
...
+public:
+  using store_plugin::builder_and_header;
+
+  // plugin API
+  caf::error initialize(data) override {
+    return {};
+  }
+
+  [[nodiscard]] const char* name() const override {
+    return "global_segment_store";
```
lava, line 28: TODO: Create slack  poll to decide on dashes vs. underscores as naming convention.

libvast/src/system/index.cpp
```diff
@@ -603,18 +634,30 @@ index(index_actor::stateful_pointer<index_state> self, store_actor store,
...
   self->state.self = self;
   self->state.accept_queries = true;
-  self->state.store = std::move(store);
+  self->state.store_plugin_
+    = partition_local_stores
+        ? plugins::find<store_plugin>("local_segment_store")
+        : plugins::find<store_plugin>("global_segment_store");
+  if (!self->state.store_plugin_) {
+    VAST_ERROR("{} could not create store", self);
+    self->quit();
```
lava, line 651: Pass an `error` with the message to quit()

libvast/src/system/index.cpp
```diff
@@ -603,18 +634,30 @@ index(index_actor::stateful_pointer<index_state> self, store_actor store,
...
   // Set members.
   self->state.self = self;
   self->state.accept_queries = true;
-  self->state.store = std::move(store);
+  self->state.store_plugin_
+    = partition_local_stores
+        ? plugins::find<store_plugin>("local_segment_store")
+        : plugins::find<store_plugin>("global_segment_store");
+  if (!self->state.store_plugin_) {
+    VAST_ERROR("{} could not create store", self);
```
lava, line 650: ...could not find store plugin {}

libvast/src/system/local_segment_store.cpp
```diff
@@ -0,0 +1,306 @@
...
+      checkers.emplace_back();
+    } else {
+      auto c = tailor(query.expr, slice.layout());
+      if (!c)
+        return c.error();
+      checkers.emplace_back(prune_meta_predicates(std::move(*c)));
+    }
+  }
+  auto zipped = detail::zip(slices, checkers);
+  caf::visit(detail::overload{
```
lava, line 51: Extract the functor in a separate variable to save ~10 levels of indentation.

libvast/src/system/local_segment_store.cpp
```diff
@@ -0,0 +1,306 @@
...
+                 self->state.deferred_requests.size());
+      for (auto&& [expr, ids, rp] :
+           std::exchange(self->state.deferred_requests, {}))
+        rp.delegate(static_cast<store_actor>(self), std::move(expr),
+                    std::move(ids));
+    });
+  return {
+    // store
+    [self](query query, ids ids) -> caf::result<atom::done> {
+      VAST_WARN("got a query for some ids");
```
lava, line 130: Remove this

libvast/src/system/local_segment_store.cpp
```diff
@@ -0,0 +1,306 @@
...
+            // partition flatbuffer with the changed store header as well.
+            std::filesystem::rename(new_path, old_path, ec);
+            if (ec)
+              VAST_ERROR("failed to erase old data {}", seg.id());
+            self->state.segment = std::move(seg);
+          },
+          [](caf::error& err) {
+            VAST_ERROR("failed to flush archive {}", to_string(err));
+          });
+      return atom::done_v;
```
lava, line 185: Return response promise here and only return `done` after the file was overwritten.

libvast/src/system/local_segment_store.cpp
```diff
@@ -0,0 +1,306 @@
...
+    // store builder
+    [self](
+      caf::stream<table_slice> in) -> caf::inbound_stream_slot<table_slice> {
+      return self
+        ->make_sink(
+          in, [=](caf::unit_t&) {},
+          [=](caf::unit_t&, std::vector<table_slice>& batch) {
+            VAST_TRACE("{} gets batch of {} table slices", self, batch.size());
+            for (auto& slice : batch)
+              if (auto error = self->state.builder->add(slice))
```
lava, line 241: Don't use a builder immediately here, but just add all table slices at once just before calling `finish()`.

libvast/src/system/partition.cpp
```diff
@@ -301,6 +312,15 @@ unpack(const fbs::partition::v0& partition, passive_partition_state& state) {
     return caf::make_error(ec::format_error,
                            "missing 'layouts' field in partition "
                            "flatbuffer");
+  auto store_header = partition.store();
+  // If no store_id is set, use the global store for backwards compatibility.
+  if (store_header && !store_header->id())
+    return caf::make_error(ec::format_error, "missing 'id' field in partition store header");
+  state.store_backend = store_header ? store_header->id()->str()
+                                     : std::string{"global_segment_store"};
+  state.store_header
```
lava, line 321: store_header = as_bytes(*store_header->data());

libvast/src/system/local_segment_store.cpp
```diff
@@ -0,0 +1,306 @@
...
+    return builder_and_header{std::move(builder), std::move(header)};
+  }
+
+  [[nodiscard]] virtual caf::expected<system::store_actor>
+  make_store(span<const std::byte> header) const override {
+    std::string_view sv{reinterpret_cast<const char*>(header.data()),
+                        header.size()};
+    std::filesystem::path path{sv};
+    // TODO: Would it make sense to pass an actor_system& as first arg?
+    return fs_->home_system().spawn(passive_local_store, fs_, path);
```
lava, line 297: Use lazy spawn here

libvast/src/system/spawn_node.cpp
```diff
@@ -131,6 +132,13 @@ spawn_node(caf::scoped_actor& self, const caf::settings& opts) {
              [&](caf::error& err) { error = std::move(err); });
   if (error)
     return error;
+  // Setup the store backends.
+  // TODO: Figure out the correct place to do this.
```
lava, line 136: Re-use the mechanism from the component plugins to remove this loop.

libvast/vast/system/partition.hpp
```diff
@@ -104,6 +104,13 @@ struct active_partition_state {
   /// The number of events in the partition.
   size_t events;
 
+  /// Whether VAST is configured to use partition-local stores.
+  bool partition_local_stores;
+
+  std::string store_backend;
```
lava, line 110: s/store_backend/store_id/

libvast/src/system/local_segment_store.cpp
lava: We currently rely on the builder to be able to call `lookup()` on it.

libvast/src/system/spawn_index.cpp
```diff
@@ -30,18 +30,17 @@ spawn_index(node_actor::stateful_pointer<node_state> self,
   auto opt = [&](std::string_view key, auto default_value) {
     return get_or(args.inv.options, key, default_value);
   };
-  auto [archive, filesystem, accountant]
+  auto [filesystem, accountant, archive]
     = self->state.registry
-        .find<archive_actor, filesystem_actor, accountant_actor>();
-  if (!archive)
-    return caf::make_error(ec::lookup_error, "failed to find archive actor");
+        .find<filesystem_actor, accountant_actor, archive_actor>();
```
tobim, line 35: nit: This is just reordering the list, would be nice to avoid such changes for a better `git blame` experience.

libvast/src/system/spawn_index.cpp
```diff
@@ -30,18 +30,17 @@ spawn_index(node_actor::stateful_pointer<node_state> self,
...
   if (!filesystem)
     return caf::make_error(ec::lookup_error, "failed to find filesystem actor");
   const auto indexdir = args.dir / args.label;
   namespace sd = vast::defaults::system;
   auto handle = self->spawn(
-    index, static_cast<store_actor>(archive), filesystem, indexdir,
+    index, filesystem, archive, indexdir,
     // TODO: Pass these options as a vast::data object instead.
+    opt("vast.partition-local-stores", sd::partition_local_stores),
```
tobim, line 43: This is missing from `application.cpp`.
I would also prefer to use a string option by the name "default-store" instead. The default value should be "legacy-archive". This way we won't need another option when we introduce other store plugins.

libvast/vast/system/index.hpp
```diff
@@ -256,19 +273,20 @@ caf::expected<flatbuffers::Offset<fbs::Index>>
...
 /// @param taste_partitions How many lookup partitions to schedule immediately.
 /// @param num_workers The maximum amount of concurrent lookups.
 /// @param meta_index_fp_rate The false positive rate for the meta index.
 /// @pre `partition_capacity > 0
+//  TODO: Use a settings struct for the various parameters.
 index_actor::behavior_type
-index(index_actor::stateful_pointer<index_state> self, store_actor store,
-      filesystem_actor filesystem, const std::filesystem::path& dir,
+index(index_actor::stateful_pointer<index_state> self,
+      filesystem_actor filesystem, archive_actor archive,
```
tobim, line 288: Prefer to cast in `spawn_index()`:
```suggestion
      filesystem_actor filesystem, store_builder_actor global_store,
```
Can you also align the comments?

tobim (changes requested):
A few minor suggestions regarding the parameters for the index.

libvast/vast/system/index.hpp
lava: Not sure about this one, we certainly want the one and only archive here, not just any store builder.

libvast/src/system/spawn_index.cpp
lava: Agreed on the broader point, but not sure about the details:
 - "default-store" is a bit misleading because there's no way for the imported data to go to non-default stores.
 - I'm not sure the archive should be "legacy"  and the default value at the same time.

How about `store-backend: archive` as the default?

libvast/vast/system/index.hpp
tobim: That would be OK as well, but use the `archive` type in the `index_state` as well then.

libvast/src/system/spawn_index.cpp
tobim: > Agreed on the broader point, but not sure about the details:
> 
>     * "default-store" is a bit misleading because there's no way for the imported data to go to non-default stores.

I expect there to be a way in the future.

> 
>     * I'm not sure the archive should be "legacy"  and the default value at the same time.

Naming it legacy would convey to the user that it is on its way out, but I see your point.

libvast/src/system/disk_monitor.cpp
```diff
@@ -201,8 +201,10 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
                     partition.id)
           .then(
             [=, sg = shared_guard](ids erased_ids) {
-              // TODO: It would be more natural if we could chain these futures,
-              // instead of nesting them.
+              // TODO: This currently only erases data from the global archive.
+              // We should remove this continuation from the disk monitor and
+              // let the partition do this instead, to support partition-local
+              // stores.
```
dominiklohmann, line 207: Do we have a story for this or is this still being fixed in the scope of this PR?

libvast/vast/plugin.hpp
```diff
@@ -258,6 +263,22 @@ class transform_plugin : public virtual plugin {
   make_transform_step(const caf::settings& options) const = 0;
 };
 
+// -- store plugin ------------------------------------------------------------
+
+class store_plugin : public virtual plugin {
```
dominiklohmann, line 268: This is missing a comment and a separate PR against docs.tenzir.com

libvast/vast/plugin.hpp
```diff
@@ -258,6 +263,22 @@ class transform_plugin : public virtual plugin {
...
+public:
+  using builder_and_header = std::pair<system::store_builder_actor, chunk_ptr>;
+
+  /// Create a store builder actor that accepts incoming table slices.
+  [[nodiscard]] virtual caf::expected<builder_and_header>
+  make_store_builder(system::filesystem_actor, const vast::uuid&) const = 0;
```
dominiklohmann, line 274: Please document the individual parameters here.

libvast/vast/plugin.hpp
```diff
@@ -258,6 +263,22 @@ class transform_plugin : public virtual plugin {
...
+
+  /// Create a store actor from the given header. Called when deserializing a
+  /// partition that uses this partition as a store backend.
+  [[nodiscard]] virtual caf::expected<system::store_actor>
+  make_store(system::filesystem_actor, span<const std::byte> header) const = 0;
```
dominiklohmann, line 279: Please document the individual parameters here.

libvast/src/system/disk_monitor.cpp
lava: I've created ch26761 to track this.

libvast/vast/plugin.hpp
lava: I've added the comment; as per our offline discussion we'll do the docs PR as a separate story at a later time.

tobim (changes requested):
This is still preliminary, I'm slowly going through the local store implementation.

libvast/src/system/local_segment_store.cpp
```diff
@@ -0,0 +1,336 @@
...
+  for (const auto& slice : slices) {
+    if (query.expr == expression{}) {
+      checkers.emplace_back();
+    } else {
+      auto c = tailor(query.expr, slice.layout());
+      if (!c)
+        return c.error();
+      checkers.emplace_back(prune_meta_predicates(std::move(*c)));
+    }
+  }
```
tobim, line 50: If I understand it right the first loop exists to make an early exit in case one of the calls to `tailor` fails, but I think we should skip only those slices instead.

tobim (changes requested):
I ran the current version through the full integration test suite and noticed that some of the tests fail that I expected to pass.

For example "Server Zeek conn log" step 1 does not produce any output. I had a brief look at the node log and it seems the evaluator does not produce any ids. I would like to have everything but the disk monitor and `vast get` in a working state before this gets merged.

Another note: We should also have a chat about explore, specifically whether we'll need to change the type of the `store_actor` when we get rid of the global id space.

libvast/vast/system/local_segment_store.hpp
```diff
@@ -0,0 +1,82 @@
...
+  /// Number of events in this store.
+  size_t events = {};
+};
+
+struct passive_store_state {
+  /// Holds requests that did arrive while the segment data
+  /// was still being loaded from disk.
+  using request = std::tuple<vast::query, vast::ids,
+                             caf::typed_response_promise<atom::done>>;
+  std::vector<request> deferred_requests;
```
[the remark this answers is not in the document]
lava: We discussed the review questions in a video call; for documentation purposes I'll write down the answers here.

While `caf::skip` would *most likely* solve the problem according to an experiment we did (and contrary what we assumed from just reading the code), this usage of `caf::skip` will be deprecated with CAF 0.18 and thus we would need to rewrite the actor using `partial_behavior_init` and multiple behaviors to take advantage of caf's internal skipping mechanism.

Also, according to discussions with upstream, skipping may be removed completely in the future, since its easy to accidentally kill performance with it.

libvast/vast/system/local_segment_store.hpp
```diff
@@ -0,0 +1,82 @@
...
+  filesystem_actor fs;
+
+  /// A pointer to the hosting actor.
+  //  We intentionally store a strong pointer here: The store lifetime is
+  //  ref-counted, it should exit after all currently active queries for this
+  //  store have finished, its partition has dropped out of the cache, and it
+  //  received all data from the incoming stream. This pointer serves to keep
+  //  the ref-count alive for the last part, and is reset after the data has
+  //  been written to disk.
+  local_store_actor self;
```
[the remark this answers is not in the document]
lava: It is true, but it shouldn't matter in both directions:

The upstream (ie. the index currently) will always close the stream stage before it is supposed to exit, so this should not extend the lifetime unnecessarily.

In the other direction, we don't want to rely on the ref-count from the upstream stream stage to keep the store alive, since that will usually drop to zero *before* all data from the stream arrived at the segment.

libvast/src/system/local_segment_store.cpp
lava: The root cause of this is us currently lacking a way to return "partial success" from a query. If we just skip, then results might be silently missing without error notification for the user. This behavior is the same as the old (non-local) segment store,

libvast/src/system/local_segment_store.cpp
```diff
@@ -308,7 +308,7 @@ class local_store_plugin final : public virtual store_plugin {
   }
 
   [[nodiscard]] const char* name() const override {
-    return "local-segment-store";
+    return "segment-store";
```
dominiklohmann, line 311: It's a bit weird that the plugin name differs from the class and file names, why is that?

lava: We thought that `local-segment-store` is too technical for a user-facing name, so switched it to just "segment-store". But there already exist files called `segment_store.{cpp,hpp}`, so I thought since the file name is developer-facing it should be fine to keep it as `local_segment_store.hpp` so its easy to remember whats the difference between this and the other file.

Of course, naming is the hardest part, so feel free to suggest a better alternative name.

dominiklohmann: I have no hard opinion on this, was just a bit confused. Thanks for explaining.

For naming suggestions: Isn't `segment` already too technical? You could call it just `store`, or `file-store`. But then again this really is bikeshedding.

lava: Just `store` sounds a bit too generic if we assume that we might have different store backends at some point. Using `file-store` sounds fine to me, personally I'm also starting to like `local-store` (saying that the data is stored on your local hard drive)

dominiklohmann: 👍 for `local-store` from my side. @tobim what do you think?

mavam: How about `event-store`?

dominiklohmann: I don't think `event-store` conveys that it is local to the database.

mavam: From a user's perspective, isn't this just a `partition-store` plugin? At least high-level, the developer/user thinks of the block of events as a (horizontal) partition, so the way the data is stored would naturally be called the `partition-store`.

lava: If I understand you correctly you're saying that `partition-store` would be a more intuitive name for the "category" of store plugins? (ie. what is currently called `store-backend` in this PR) That still leaves the question how to name this particular implementation of a "partition store", which is what we originally discussed here.

tobim: We should look at these names based on the context of what other implementations we might have in the future. Most of those have obvious names. I'm thinking of `parquet-store` of `s3-store`. In that context we can rule out some options thanks to ambiguities: `file-store` is in conflict with `parquet-store`, because parquet is a file too. `local-store` also conflicts with `parquet-store` because both are local. It seems that single properties don't serve as good names here.
I'm equally unenthusiastic about `event-store` and `partition-store` because every implementation will store events at the partition granularity.

From the options we have at the moment `segment-store` seems to be the best, even though the name is not descriptive for operators without implementation knowledge. The big advantage is that it isn't misleading. The opaque name will hopefully prompt users to read the documentation instead of going with some assumptions.

tobim (approved):
I tested this locally and will deploy it on the testbed soon.

libvast/src/system/local_segment_store.cpp
lava: I'll go with `segment-store` for now to get the PR merged; the plugin is still clearly labelled as experimental so there should be no issue renaming this later on if so desired.

PR #1722, opened 2021-06-16
Fix integration tests for external plugins

dominiklohmann:

The bare-mode, which we use for integration tests to avoid side-effects from global configuration files, disabled the addition of `vast.plugin-dirs` to the plugin search paths. However, after recent changes that list also contained the options from both the command line and the environment variable, which were then ignored with bare-mode enabled.

Since we do not read configuration files that were not explicitly specified in bare-mode, it is safe to just read the option `vast.plugin-dirs` even with bare-mode enabled.

### :memo: Checklist

Read the description and the diff. This does not need a changelog entry since we just introduced this bug last week. I changed the CI for the example plugins to catch regressions on this and added a comment on the changed order.

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -655,13 +655,16 @@ jobs:
         working-directory: ${{ env.BUILD_DIR }}
         run: |
           ctest --parallel
+      - name: Run Integration Tests
+        id: integration_tests
+        # We intentionally run the integration tests before installing
+        # them, because that is a use-case we want to explicitly support
+        # for easier plugin development.
+        run: |
+          cmake --build "$BUILD_DIR" --target integration
       - name: Install
         run: |
           cmake --install "$BUILD_DIR"
-      - name: Run Integration Tests
-        id: integration_tests
-        run: |
-          cmake --build "$BUILD_DIR" --target integration
       - name: Upload Integration Test Logs on Failure
         if: failure()
         uses: actions/upload-artifact@v2.2.3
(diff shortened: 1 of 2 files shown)


libvast/src/plugin.cpp
```diff
@@ -54,14 +54,13 @@ detail::stable_set<std::filesystem::path>
 get_plugin_dirs(const caf::actor_system_config& cfg) {
   detail::stable_set<std::filesystem::path> result;
   const auto bare_mode = caf::get_or(cfg, "vast.bare-mode", false);
-  if (!bare_mode) {
+  if (auto dirs = caf::get_if<std::vector<std::string>>( //
+        &cfg, "vast.plugin-dirs"))
+    result.insert(dirs->begin(), dirs->end());
+  if (!bare_mode)
```
tobim, line 60: Can you copy the content of the commit message to a comment above line 57? It looks wrong without the context.

.github/workflows/vast.yaml
```diff
@@ -655,13 +655,16 @@ jobs:
...
         run: |
           ctest --parallel
-      - name: Install
-        run: |
-          cmake --install "$BUILD_DIR"
       - name: Run Integration Tests
         id: integration_tests
+        # We intentionally run the integration tests before installing
+        # them, because that is a use-case we want to explicitly support
+        # for easier plugin development.
```
tobim, line 662:
```suggestion
        # We intentionally run the integration tests before installing,
        # because that is a use-case we want to explicitly support for
        # easier plugin development.
```

tobim (changes requested):

tobim (approved):

PR #1723, opened 2021-06-16
Fix libexecinfo fallback for VAST_ENABLE_BACKTRACE

dominiklohmann:

This fixes setups where `FindBacktrace.cmake` provided `libexecinfo` rather than `libbacktrace`, but both `backtrace.h` and `execinfo.h` are available.

### :memo: Checklist

@lava please check locally if this fixes your issue.


lava (approved):
With this change `execinfo.h` will get a higher priority than `backtrace.h`, ie. even after installing `libbacktrace` it will not be selected by default.

This is a change in behavior, but fine with me.

dominiklohmann:
> With this change `execinfo.h` will get a higher priority than `backtrace.h`, ie. even after installing `libbacktrace` it will not be selected by default.
> 
> This is a change in behavior, but fine with me.

It's not like we can change `FindBacktrace.cmake` and it was just broken before.

PR #1724, opened 2021-06-16
Add support for per-plugin configuration files

dominiklohmann:

This change adds per-plugin configuration files. The following two configurations are equivalent:

```yaml
# <configdir>/vast/vast.yaml
plugins:
  example-analyzer:
    answer: 42
```

```yaml
# <configdir>/vast/plugin/example-analyzer.yaml
answer: 42
```

This allows for unified deployments of plugins and their respective configuration. Plugin configuration no longer sneaks into the main deployments VAST configuration, and uninstalling a plugin no longer requires touching the main configuration file.

### :memo: Checklist

Run locally. Read code file-by-file.

This needs to be rebased after #1724 to make use of the new merge functions.

### DIFF
--- a/libvast/src/plugin.cpp
+++ b/libvast/src/plugin.cpp
@@ -18,6 +18,7 @@
 #include "vast/error.hpp"
 #include "vast/logger.hpp"
 #include "vast/plugin.hpp"
+#include "vast/system/configuration.hpp"
 #include "vast/system/node.hpp"
 
 #include <caf/actor_system_config.hpp>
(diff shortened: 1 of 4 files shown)


libvast/src/plugin.cpp
```diff
@@ -213,30 +214,61 @@ load(std::vector<std::string> bundled_plugins, caf::actor_system_config& cfg) {
...
       }
-    } else {
-      VAST_DEBUG("no configuration found for plugin {}", plugin->name());
-      if (auto err = plugin->initialize(data{}))
-        return caf::make_error(ec::unspecified,
-                               fmt::format("failed to initialize plugin {}: {}",
-                                           plugin->name(), err));
     }
+    // Third, initialize the plugin with the merged configuration.
+    VAST_DEBUG("initializing plugin {} with options: {}", plugin->name(),
```
dominiklohmann, line 266: Move to verbose?

lava (approved):
I verified locally that VAST is loading plugin configuration files as described in the changelog. (although tbf, I didn't have a plugin at hand that actually *uses* its configuration)

It's still a bit weird that it's not possible to use these config files explicitly, but just implicitly by placing them in the correct directory with the correct filename, but it already solves the DCSO use case and is a good first step anyways.

PR #1728, opened 2021-06-18
Fix dependency graph for VASTTargets exports set

dominiklohmann:

This PR fixes two bugs:

1. **Fix dependency graph for VASTTargets exports set**

    For the Unix Makefiles generator, our CMake code generated a false install-time dependency for the VASTTargets export set, causing the installation of libvast-fbs to trigger a rebuild of all other targets in the VASTTargets set. This did not occur for other generators like Ninja. The effects could be seen for example in the static binary CI workflow, which compiled everything in the VASTTargets export set twice.

   Installing the generated FlatBuffers interface library in the same export set fixes the issue. To make that work, the CMake 3.15+ property PUBLIC_HEADER must be manually filled with the generated headers, as adding sources to an interface library is a CMake 3.19+ feature.

   I verified manually that this works for both Unix Makefiles and Ninja, and the effects of this should also immediately be visible in CI.

2. **Use bare-mode for all commands in integration tests**

   Our integration test contains three places where it runs VAST:
   - Running a command
   - Starting a fixture
   - Stopping a fixture

   Before this change, we only added the `--bare-mode` flag when starting the fixture, which caused unexpected results when running the integration tests for some setups.

### :memo: Checklist

Commit-by-commit. Read the rather lengthy commit messages.

I don't think either of this changes is user-facing, so we should not need a changelog entry. I will mention the bug fixes in the release notes for `VAST 2021.06.24-rc2`.

### DIFF
--- a/cmake/VASTRegisterPlugin.cmake
+++ b/cmake/VASTRegisterPlugin.cmake
@@ -62,7 +62,8 @@ function (VASTCompileFlatBuffers)
   else ()
     message(
       FATAL_ERROR
-        "Found FlatBuffers, but neither shared nor static library target exist")
+        "Found FlatBuffers, but neither shared nor static library targets exist"
+    )
   endif ()
 
   if ("${CMAKE_PROJECT_NAME}" STREQUAL "VAST")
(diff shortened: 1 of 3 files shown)


cmake/VASTRegisterPlugin.cmake
```diff
@@ -99,22 +100,32 @@ function (VASTCompileFlatBuffers)
       "      new_include \"\${include}\")\n"
       "file(WRITE \"${desired_file}\" \"\${new_include}\")\n")
     # Compile and rename schema.
-    add_custom_target(
-      "compile-flatbuffers-schema-${basename}"
-      BYPRODUCTS ${desired_file}
+    add_custom_command(
```
tobim, line 103: Explain why we use the additional indirection through the custom command here.

tobim (approved):
Please add the requested comment, feel free to merge once that is done.

PR #1733, opened 2021-06-21
Bundle plugin schemas with the plugin build target

dominiklohmann:

Before this change, plugin schemas and taxonomies were only installed with the build target `all`. Since plugins that bundle schemas usually depend on them, building the plugin should always bundle the schemas.

### :memo: Checklist

Run locally with the PCAP plugin. On master, notice that just running `cmake --build build --target pcap` does not install the bundled concept definitions, while running `cmake --build build --target all` does. This PR fixes that.


tobim (approved):
This should prevent confusing behavior when working on plugins in the future.

PR #1734, opened 2021-06-21
Fix duplication of lists in client command options

dominiklohmann:

We've recently changed VAST to merge lists in its configuration from all possible sources rather than overwriting them. However, in the command-executing `run` function we merged configuration options with the local actor system's options again in case the command invocation happened remotely, which was no longer a no-op now that we merge lists, which in turn led to all list options of client commands being duplicated. This fixes the issue by not merging lists between client and server when doing so.

### :memo: Checklist

Test locally. E.g., view merged options by running with `--verbosity=trace`.


lava (approved):
Code looks safe; reasoning sounds convincing.

PR #1740, opened 2021-06-23
Fix loading of the optional OpenSSL module

dominiklohmann:

Before this change, we never loaded the OpenSSL module despite checking whether any configuration option for using CAF's built-in OpenSSL module are set. This led to an uncaught runtime_error exception when trying to use the (undocumented) options.

For testing, I used the following:

```bash
vast --caf.openssl.certificate=cert.pem --caf.openssl.key=key.pem start
vast --caf.openssl.certificate=cert.pem --caf.openssl.key=key.pem status
```

### :memo: Checklist

Run locally. I don't think we should merge this before the release—it's been broken for months and nobody noticed it, so we should not need to rush this in.

### DIFF
--- a/libvast/src/system/configuration.cpp
+++ b/libvast/src/system/configuration.cpp
@@ -273,7 +273,17 @@ caf::error configuration::parse(int argc, char** argv) {
   // We clear the config_file_path first so it does not use
   // caf-application.ini as fallback during actor_system_config::parse().
   config_file_path.clear();
-  return actor_system_config::parse(std::move(caf_args));
+  auto result = actor_system_config::parse(std::move(caf_args));
+  // Load OpenSSL module if configured to do so.
+#if VAST_ENABLE_OPENSSL
+  const auto use_encryption
+    = !openssl_certificate.empty() || !openssl_key.empty()
+      || !openssl_passphrase.empty() || !openssl_capath.empty()
+      || !openssl_cafile.empty();
+  if (use_encryption)
+    load<caf::openssl::manager>();
+#endif // VAST_ENABLE_OPENSSL
+  return result;
 }
 
 } // namespace vast::system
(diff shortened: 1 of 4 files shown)


libvast/src/system/configuration.cpp
```diff
@@ -273,7 +273,17 @@ caf::error configuration::parse(int argc, char** argv) {
...
+  auto result = actor_system_config::parse(std::move(caf_args));
+  // Load OpenSSL module if configured to do so.
+#if VAST_ENABLE_OPENSSL
+  const auto use_encryption
+    = !openssl_certificate.empty() || !openssl_key.empty()
+      || !openssl_passphrase.empty() || !openssl_capath.empty()
+      || !openssl_cafile.empty();
+  if (use_encryption)
+    load<caf::openssl::manager>();
+#endif // VAST_ENABLE_OPENSSL
```
lava, line 285: It feels a bit weird to actually switch openssl on/off during configuration parsing. Intuitively I'd have expected us to always load the openssl-manager in the constructor along with the other stuff when openssl is enabled, and then use the configuration to decide if we *use* the loaded openssl manager for sending and receiving messages.

If the main concern is about not repeating the very verbose check, maybe we can introduce a single setting `use_openssl` and then assert here that the 5 other values are correctly specified when this is set to true.

lava (approved):
Approving this because it's obviously better than being broken, but I think it's not the ideal solution yet as explained in the review comment.

libvast/src/system/configuration.cpp
dominiklohmann: I tried that first, but if you load the OpenSSL manager then CAF requires you to use it.

lava: Oh, ok, nevermind then.

PR #1742, opened 2021-06-23
Support import filter expressions

dominiklohmann:

We have long advertised this feature in the documentation of the import command, but never gotten around to implementing it. Turns out it really is quite simple to implement with the new `filter` function on table slices.

Here's how you can use it:

```bash
# Do not import Suricata.stats events as they are rather meaningless.
vast import suricata '#type != "suricata.stats"'

# Import only indicators to a matcher whose 'type' field is 'ip' or
# 'ipv6' (requires the proprietary Live Matching plugin).
vast matcher import -t my-ioc csv my-matcher 'type == "ip" || type == "ipv6"'
```

### :memo: Checklist

File-by-file. Give it a spin locally.


libvast/src/system/datagram_source.cpp
```diff
@@ -105,8 +107,23 @@ caf::behavior datagram_source(
...
+            self->state.mgr->out().push(std::move(*filtered_slice));
+          } else {
+            VAST_VERBOSE("{} forwards 0/{} produced {} events after filtering",
+                         self, unfiltered_rows, slice.layout().name());
+          }
+        } else {
+          VAST_VERBOSE("{} forwards {} produced {} events", self,
+                       unfiltered_rows, slice.layout().name());
+          self->state.mgr->out().push(std::move(slice));
+        }
```
tobim, line 126: The log severity for all of those messages has to be `DEBUG`.

tobim (changes requested):
Looks good, but let's avoid flooding `VERBOSE`.

libvast/src/system/datagram_source.cpp
dominiklohmann: Done. I also moved the logic to a function in the source state that is used for both source types.

tobim (approved):

PR #1747, opened 2021-06-28
Fix byproduct listing for update-changelog target

dominiklohmann:

Without this change, running the `clean` target after running the `update-changelog` target causes the `CHANGELOG.md` file in the source directory to be removed, which is most certainly undesired.

### :memo: Checklist

Reading the code should suffice here, the change is really simple.


tobim (approved):
I trust that this change solves the problem.

PR #1748, opened 2021-06-29
Print the import transformer status

tobim:

We didn't request a status message from the import transformer so far. This fixes that oversight.


libvast/src/system/importer.cpp
```diff
@@ -205,7 +216,24 @@ caf::settings importer_state::status(status_verbosity v) const {
   // General state such as open streams.
   if (v >= status_verbosity::debug)
     detail::fill_status_map(importer_status, self);
-  return result;
+  const auto timeout = caf::duration{defaults::system::initial_request_timeout};
```
dominiklohmann, line 219: Please use half the timeout for nested requests.

libvast/src/system/importer.cpp
```diff
@@ -205,7 +216,24 @@ caf::settings importer_state::status(status_verbosity v) const {
...
+  self->request(transformer, timeout, atom::status_v, v)
+    .then(
+      [=, &importer_status](caf::settings& transformer_status) {
+        put(importer_status, "transformer", std::move(transformer_status));
+        // Both handlers have a copy of req_state.
+        if (req_state.use_count() == 2)
+          deliver(std::move(*req_state));
+      },
+      [=, &importer_status](caf::error& err) {
+        VAST_WARN("{} failed to retrieve status from the transformer : {}",
```
dominiklohmann, line 229:
```suggestion
        VAST_WARN("{} failed to retrieve status from the transformer: {}",
```

libvast/src/system/importer.cpp
```diff
@@ -205,7 +216,24 @@ caf::settings importer_state::status(status_verbosity v) const {
...
+                  self, render(err));
+        put(importer_status, "transformer", render(std::move(err)));
```
dominiklohmann, line 231: This should use `fmt::to_string` rather than `render`

dominiklohmann (commented):
Just some smaller comments.

Please make the same adjustment in the status handlers of the SOURCE / DATAGRAM SOURCE actors.

This also needs a changelog entry.

tobim:
This turned into a bit of a refactoring of the status handlers now. I originally hoped to manage this with less code, but at least it should now be easier to write handlers that retrieve additional data from child components.

libvast/vast/system/status.hpp
```diff
@@ -0,0 +1,130 @@
...
+  std::chrono::milliseconds timeout, status_verbosity verbosity, Resp responder,
+  caf::settings& s, std::string_view key) {
+  collect_status(
+    rs, timeout, verbosity, responder,
+    [key, &s](caf::settings& response) {
+      put(s, std::string_view{key}, std::move(response));
+    },
+    [self = rs->self, key, &s](caf::error& err) {
+      VAST_WARN("{} failed to retrieve status for the key {}: {}", self, key,
+                fmt::to_string(err));
```
dominiklohmann, line 125: The to_string here is unnecessary.

libvast/vast/system/status.hpp
```diff
@@ -0,0 +1,130 @@
...
+void collect_status(
+  const std::shared_ptr<status_request_state<Ptr, Result, Extra>>& rs,
+  std::chrono::milliseconds timeout, status_verbosity verbosity, Resp responder,
+  caf::settings& s, std::string_view key) {
+  collect_status(
+    rs, timeout, verbosity, responder,
+    [key, &s](caf::settings& response) {
+      put(s, std::string_view{key}, std::move(response));
+    },
+    [self = rs->self, key, &s](caf::error& err) {
```
dominiklohmann, line 123:
```suggestion
    [self = rs->self, key, &s](const caf::error& err) {
```

libvast/vast/system/status.hpp
```diff
@@ -0,0 +1,130 @@
...
+
+/// The verbosity level of a status request.
+enum class status_verbosity { info, detailed, debug };
+
+template <class Ptr, class Result>
+struct status_request_state_base {
+  // The actor handling the original request.
+  Ptr self;
+  // Promise to the original request.
+  caf::typed_response_promise<Result> promise;
```
dominiklohmann, line 31: We need to be careful here not to access the promise outside of the context of the actor that created it. Can't we just call `self->make_response_promise<Result>` in the wrapping code and remove it from this state?

dominiklohmann (commented):
This is a good abstraction, but we should make sure with the API design that response promises cannot be used from outside of their hosting actor's context.

libvast/vast/system/status.hpp
tobim: The promise is part of the request state object because their lifetime is intentionally coupled. Would you be satisfied if `status_request_state_base` was not copyable?

dominiklohmann: Yes, that would help a lot.

dominiklohmann (approved):

tobim:
Rebased on top of 84140ffec526bb035685bb01d4d6675b60385bfa now.

PR #1750, opened 2021-06-30
Bump nixpkgs and CAF to switch to GCC11 and clang12

tobim:

This updates nixpkgs to db6e089 so we can start using GCC 11 for our static binary build. I had to integrate some fixes in the CAF submodule for that too.

### :memo: Checklist

Download the static binary artifacts and verify that they pass the integration tests.

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -37,6 +37,10 @@ in {
     # at the optimization stage.
     # TODO: Remove when updating to CAF 0.18.
     + lib.optionalString isStatic " -std=c++17";
+    preCheck = ''
+      export LD_LIBRARY_PATH=$PWD/lib
+      export DYLD_LIBRARY_PATH=$PWD/lib
+    '';
   } // lib.optionalAttrs isStatic {
     cmakeFlags = old.cmakeFlags ++ [
       "-DCAF_BUILD_STATIC=ON"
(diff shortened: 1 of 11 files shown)


nix/overlay.nix
```diff
@@ -37,6 +37,10 @@ in {
     # at the optimization stage.
     # TODO: Remove when updating to CAF 0.18.
     + lib.optionalString isStatic " -std=c++17";
+    preCheck = ''
+      export LD_LIBRARY_PATH=$PWD/lib
+      export DYLD_LIBRARY_PATH=$PWD/lib
+    '';
```
dominiklohmann, line 43: Why is this necessary?

tobim: For regular builds - meaning not cross compiling - the unit tests are run after the build but before installing. Nix sets the flag `CMAKE_BUILD_WITH_INSTALL_RPATH=ON`, that means we need to add the library location here so the dynamic linker can find the `.so`s.

> But this is based on the upstream caf package which we just override with a custom version of the sources, so why isn't it taken care of by the upstream package?

Well, upstream is already at CAF 0.18, and that changed the `CMAKE_LIBRARY_OUTPUT_DIRECTORY`, so the upstream exports are different:
```
    export LD_LIBRARY_PATH=$PWD/libcaf_core:$PWD/libcaf_io
    export DYLD_LIBRARY_PATH=$PWD/libcaf_core:$PWD/libcaf_io
```
So we need to override it to work for CAF 0.17.

dominiklohmann: This is a really great reply, thanks for explaining in such detail.

dominiklohmann (approved):

PR #1751, opened 2021-07-01
Remove unused forward declarations and atoms

tobim:

We have a lot of them that aren't used anywhere.


dominiklohmann (approved):

tobim:
Thanks for approving. I decided to keep the atoms `add`, `connect`, `list`, and `start`, because they are generic enough to be useful in plugins.

PR #1752, opened 2021-07-01
Shutdown the filesystem only when no longer used

dominiklohmann:

See comment in the diff. This is not user-facing for VAST, but fixes a persistence bug in the Matcher plugin.

### :memo: Checklist

Read the code, and most importantly, verify locally that everything still works.


dominiklohmann:
[ch25432]

mavam (commented):
Switching to reference counting for the filesystem actor makes sense to me, given that plugins want to use it as well.

Implementation-wise, @lava and @tobim have the final say.

tobim (approved):
Making an exception here makes is acceptable.

PR #1754, opened 2021-07-02
Fix error message about /dev/null-backend on startup

lava:

Don't use `VAST_WARN()` while initializing the logger, which leads to a nasty-looking error message.

Also, fix a parser bug in the journald detection routine that would cause us to always mistakenly return "false".

### :memo: Checklist


mavam (approved):
Didn't test locally, but the changeset makes sense.

dominiklohmann (approved):
This looks good to me. Can you add a user-facing changelog entry before merging?

PR #1755, opened 2021-07-02
Fix bug in bitmap offset computation

lava:
### :memo: Checklist

### DIFF
--- a/libvast/vast/bitmap_algorithms.hpp
+++ b/libvast/vast/bitmap_algorithms.hpp
@@ -411,13 +411,12 @@ public:
          !rng_.done();
          n_ += bits().size(), rng_.next()) {
       VAST_ASSERT(k > 0);
-      if (k > bits().size()) {
-        k -= rank<Bit>(bits());
-      } else {
+      if (k <= bits().size()) {
         i_ = select<Bit>(bits(), k);
         if (i_ != npos)
           break;
       }
+      k -= rank<Bit>(bits());
     }
   }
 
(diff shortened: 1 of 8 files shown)


libvast/vast/bitmap_algorithms.hpp
```diff
@@ -411,13 +411,12 @@ class bitwise_range : public detail::range_facade<bitwise_range<BitRange>> {
          !rng_.done();
          n_ += bits().size(), rng_.next()) {
       VAST_ASSERT(k > 0);
-      if (k > bits().size()) {
-        k -= rank<Bit>(bits());
-      } else {
+      if (k <= bits().size()) {
```
tobim, line 414: I would like to understand why we use `bits().size()` here but decrement by `rank<Bit>(bits())` for the next round. In what circumstance are these 2 values different?

tobim (commented):
I think the fix is correct, but I'm not sure whether there is another problem hiding in that function. Question below.

libvast/vast/bitmap_algorithms.hpp
lava: As far as I understand, `rank<Bit>(bits())` counts the number of 1-bits, ie. how many valid entries we're skipping, while `bits().size()` is the distance between the first and last id in the bitmap. So e.g. a bitmap with entries bits 1 and 1000 set to true would have rank 2 and size 1000.

mavam: I looked at the unit tests and verified that they are written correctly.

To be precise, `rank<Bit>` counts the number of `Bit`-bits where `Bit` is either 0 or 1.

tobim: I checked if one could use `rank<Bit>(bits())` in the loop condition and that works as well. I can't say which approach has better performance characteristics though, so I'd say let's keep it as it is.

tobim (approved):
Awesome work getting to the bottom of this!

PR #1758, opened 2021-07-05
Add Zeek Broker reader plugin

mavam:

This PR adds a new `broker` reader plugin, enabling VAST to be true native
Logger node. This approach takes a slightly different path than the usual way
of writing a Zeek plugin and implementing one's own `WriterBackend`. Here are
the pros and cons of this method.

**Pros**:

- Easy rollout: works out-of-the-box with Zeek, just load a script and log
  remotely
- Works seamless with the Cluster framework: VAST is a first-class logger node
- Performance: no need to "log twice," once to the Logger node and then
  parsing/repacking the data again to ship it to VAST

**Cons**:

- Does not use a standard API and reverse-engineered binary wire format
- No support for Zeek logging filters, if used

Since filtering can also be implemented easily downstream in VAST, we find that
the pros dominate this method. 

### :memo: Checklist

- [x] Replace `poll` with `get(num, timeout)`

File-by-file.


libvast/src/detail/zeekify.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+using namespace std::string_literals;
+
+record_type zeekify(record_type layout) {
+  // The first field is almost exclusively the event timestamp for standard
+  // Zeek logs. Its has the field name `ts`. For streaming JSON, some other
+  // fields, e.g., `_path`, precede it.
+  for (auto& field : layout.fields)
+    if (field.name == "ts")
+      if (auto ts = caf::get_if<time_type>(&field.type)) {
```
dominiklohmann: Please add braces here if the nested if uses braces, otherwise this becomes really hard to follow.

dominiklohmann:
```suggestion
  for (auto& field : layout.fields)
    if (field.name == "ts")
      if (caf::holds_alternative<time_type>(field.type)) {
```

libvast/src/detail/zeekify.cpp
```diff
@@ -0,0 +1,54 @@
...
+    if (!caf::holds_alternative<string_type>(field.type))
+      return false;
+    auto has_name = [&](const auto& name) {
+      return name == field.name;
+    };
+    // TODO: do more than this simple heuristic. For example, we should also
+    // consider zeek.files.conn_uids, which is a set of strings. The inner
+    // index needs to have the #index=hash tag. Moreover, we need to consider
+    // other fields, such as zeek.x509.id instead of uid.
+    static auto ids = make_array("uid", "fuid", "community_id");
```
dominiklohmann, line 43: You use this only here. Please just write `` std::array{"uid", "fuid", "community_id"} `` and remove the newly added `vast/detail/array.hpp`.

plugins/broker/zeek/remote-logging.zeek
```diff
@@ -0,0 +1,18 @@
+# This script activates Broker communication such that it becomes possible to
```
dominiklohmann, line 1: Please add a note to this file explaining how to install it.

dominiklohmann (approved):
LGTM. I just have some smaller things, plus this needs CI to run through. Let's get this in and then test this in our E2E setup outside of this repository.

PR #1760, opened 2021-07-06
Split 'lsvast' source code into separate files

lava:
As the tool continues to grow, it becomes more and more
untenable to keep all of the source code in a single file.

This commit should contains no changes in logic, just a
reorganization of the existing source files.

### :memo: Checklist


dominiklohmann (approved):
I did not look at this in-depth, but the change appears to be simple enough. Thanks for maintaining this!

PR #1761, opened 2021-07-07
Fix a shutdown hang for busy imports

dominiklohmann:

This was a tricky one: We recently switched to in-stream EOF messages for the SOURCE and the IMPORTER actor, and I bisected a hang in the server process on shutdown while particularly busy SOURCEs were running down to that change.

Turns out that while a good idea in theory, in-stream EOF messages only work if you can guarantee that they are received downstream. If downstream drops them you suddenly find yourself stuck during shutdown. Imagine a stream `A -> B -> C`, where `A` and `C` are killed, and `B` only shuts down when `A` sends an EOF message in its stream. Because `B` has no more downstream, it starts buffering up to  `2 * stage->out().capacity()` stream messages. If the EOF arrives from `A` only after the the buffer in `B` is full, `B` drops the message and hangs indefinitely.

This PR _partially_ fixes that issue and ensures that at least the server process shuts down properly. A proper fix to this is much more involved.

### :memo: Checklist

Commit-by-commit. Test locally with various import commands, and stress-test as needed.


dominiklohmann:
[ch26790]

lava:
Can you maybe link the place in the CAF source code where messages are dropped when the buffer is full? I had trouble finding it after a brief look through `buffered_downstream_manager` etc.

The formula `2 * stage->out().capacity()` also looks a bit weird to me, because `capacity()` already measures how many *free* slots the buffer still has left.

dominiklohmann:
Closing in favour of #1771

PR #1767, opened 2021-07-09
Fix Broker plugin build in CI

dominiklohmann:

This fixes the Broker plugin setup in CI.

### :memo: Checklist

Check if CI runs through.


mavam (approved):
Relying on the official instead of bundled CAF is something we want to do in the future anyway. (After upgrading to 0.18)

dominiklohmann:
[ch27043]

nix/zeek-broker/default.nix
```diff
@@ -0,0 +1,62 @@
...
+    "-DCAF_ROOT_DIR=${caf}"
+  ];
+
+  hardeningDisable = lib.optional isStatic "pic";
+
+  doCheck = false;
+  checkTarget = "test";
+
+  meta = with lib; {
+    description = "Zeek's Messaging Library ";
```
mavam, line 56:
```suggestion
    description = "Zeek's Messaging Library";
```

Nit.

tobim (approved):
I verified that this works for both bundled as well as standalone builds of the broker plugin.
I noticed some more uninitialized variable warnings with GCC11, but that can be fixed later.

tobim:
The static-binary build with a bundled broker plugin also doesn't work yet, we can fix that later.

dominiklohmann:
> The static-binary build with a bundled broker plugin also doesn't work yet, we can fix that later.

I've created a story for this deficiency.

tobim:
> > The static-binary build with a bundled broker plugin also doesn't work yet, we can fix that later.
> 
> I've created a story for this deficiency.

I already have a fix, but I don't want to delay CI again.

dominiklohmann:
> I already have a fix, but I don't want to delay CI again.

Feel free to push again, the merge can wait a bit.

PR #1768, opened 2021-07-09
Compile with C++20

tobim:

All supported build systems have recent compiler versions, so it's time to pull the trigger.

Issues that I ran into:
- `this` is no longer implicitly captured with `[=]`
- `concept` is now a keyword
- typed actor states should not have typed actor handles of different actors as their first member variable any more

### :memo: Checklist

Build and run locally.

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -321,10 +321,6 @@ jobs:
       matrix:
         build:
           - extra-flags:
-            compiler: AppleClang
-            cc: clang
-            cxx: clang++
-          - extra-flags:
             compiler: Clang
             cc: clang
             cxx: clang++
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -599,9 +599,9 @@ endif ()
 
 # -- compiler setup ------------------------------------------------------------
 
-set(CLANG_MINIMUM_VERSION 8.0)
-set(APPLE_CLANG_MINIMUM_VERSION 9.1)
-set(GCC_MINIMUM_VERSION 8.0)
+set(CLANG_MINIMUM_VERSION 11.0)
+set(APPLE_CLANG_MINIMUM_VERSION 12.0.5)
+set(GCC_MINIMUM_VERSION 10.0)
 
 if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
   check_compiler_version(${GCC_MINIMUM_VERSION})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,8 +242,8 @@ set_target_properties(libvast_internal PROPERTIES EXPORT_NAME internal)
 add_library(vast::internal ALIAS libvast_internal)
 install(TARGETS libvast_internal EXPORT VASTTargets)
 
-# Require standard C++17
-target_compile_features(libvast_internal INTERFACE cxx_std_17)
+# Require standard C++20.
+target_compile_features(libvast_internal INTERFACE cxx_std_20)
 
 # Increase maximum number of template instantiations.
 target_compile_options(libvast_internal INTERFACE -ftemplate-backtrace-limit=0)
--- a/changelog/unreleased/changes/1768--C++20.md
+++ b/changelog/unreleased/changes/1768--C++20.md
@@ -1,0 +1,3 @@
+From now on VAST is compiled with the C++20 language standard.
+The minimum compiler versions are increased to GCC 10, clang 11,
+and AppleClang 12.0.5.
--- a/libvast/CMakeLists.txt
+++ b/libvast/CMakeLists.txt
@@ -193,15 +193,30 @@ else ()
       set(CAF_BUILD_STATIC_ONLY ON)
     endif ()
     add_subdirectory(aux/caf)
-    set_target_properties(libcaf_core_${_linkage_suffix} PROPERTIES EXPORT_NAME
-                                                                    core)
+    get_target_property(libcaf_core_includes libcaf_core_${_linkage_suffix}
+                        INTERFACE_INCLUDE_DIRECTORIES)
+    get_target_property(
+      libcaf_core_system_includes libcaf_core_${_linkage_suffix}
+      INTERFACE_SYSTEM_INCLUDE_DIRECTORIES)
+    set_target_properties(
+      libcaf_core_${_linkage_suffix}
+      PROPERTIES INTERFACE_SYSTEM_INCLUDE_DIRECTORIES
+                 "${libcaf_core_system_includes};${libcaf_core_includes}"
+                 EXPORT_NAME core)
     add_library(CAF::core ALIAS libcaf_core_${_linkage_suffix})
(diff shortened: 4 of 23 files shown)


tobim:
I'm leaning towards dropping the support for AppleClang. Devs working on mac can easily install clang via homebrew.

dominiklohmann:
> I'm leaning towards dropping the support for AppleClang. Devs working on mac can easily install clang via homebrew.

Yep, I agree. Just gotta provide instructions for @mavam, which essentially boil down to putting this into your `zshrc`:

```zsh
export PATH="$(brew --prefix llvm)/bin:${PATH}"
export LDFLAGS="-Wl,-rpath,$(brew --prefix llvm) ${LDFLAGS}"
export CPPFLAGS"-isystem $(brew --prefix llvm)/include ${CPPFLAGS}"
export CXXFLAGS"-isystem $(brew --prefix llvm)/include/c++/v1 ${CXXFLAGS}"
```

JoeLoser:
Super excited to see this land and start using concepts and requires expressions!

lava:
Adding my research here since this bullet point

> typed actor states should not have typed actor handles of different actors as their first member variable any more

had me confused a bit. As far as I understand, it's due to the new "parenthesized aggregate initialization": This function in CAF's `stateful_actor.hpp`

```
  template <class T>
  typename std::enable_if<std::is_constructible<State, T>::value>::type
  cr_state(T arg) {
    new (&state_) State(arg);
  }
```
was formerly disabled via SFINAE, but now a line like this (which is probably similar to what is used in the implementation of `is_constructible`)

```
    new (&state_) State(arg);
```

is parsed as a parenthesized aggregate initializer, turning the condition to true and leading us to take the wrong branch.

An alternative to switching the order would probably be to add defaulted constructors, but either way should be fine.

mavam:
@dominiklohmann have you tested these shell settings? I am asking because they contain two syntax errors (missing space). I also wonder whether there's a more local way of enabling C++ on macOS that doesn't mess with global compilation flags.

dominiklohmann:
> @dominiklohmann have you tested these shell settings? I am asking because they contain two syntax errors (missing space).

Yes, I use them locally all the time. Sorry for the typos, I missed two equality signs when translating from my usual typeset-foo to export.

You might also need to set CC and CXX to `clang` and `clang++` respectively.

This is what I generally recommend on macOS for building VAST:

```bash
export PATH="$(brew --prefix llvm)/bin:${PATH}"
export CC="$(brew --prefix llvm)/bin/clang"
export CXX="$(brew --prefix llvm)/bin/clang++"
export LDFLAGS="-Wl,-rpath,$(brew --prefix llvm) ${LDFLAGS}"
export CPPFLAGS="-isystem $(brew --prefix llvm)/include ${CPPFLAGS}"
export CXXFLAGS="-isystem $(brew --prefix llvm)/include/c++/v1 ${CXXFLAGS}"

export CMAKE_GENERATOR='Ninja'
export CMAKE_C_COMPILER_LAUNCHER='ccache'
export CMAKE_CXX_COMPILER_LAUNCHER='ccache'
export DOCKER_BUILDKIT=1
```

> I also wonder whether there's a more local way of enabling C++ on macOS that doesn't mess with global compilation flags.

Nothing's stopping you from setting `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` for VAST only, although you will rarely experience issues with using LLVM clang by default over AppleClang. And if you do, you can always use `xcrun make` over `make` (and similar) to force usage of the default dev toolkit configured via `xcode-select` to get back to AppleClang.

lava:
Not patching CAF works in CI because we add `-Wno-error=deprecated` to the CI flags, but if you build the branch locally in debug mode you get a ton of these warnings for almost every file that interacts with CAF's actor system:

```
[ 60%] Building CXX object libvast/CMakeFiles/libvast.dir/src/system/application.cpp.o
In file included from /home/benno/src/vast/master/libvast/aux/caf/libcaf_core/caf/scoped_actor.hpp:24,
                 from /home/benno/src/vast/master/libvast/vast/detail/logger_formatters.hpp:36,
                 from /home/benno/src/vast/master/libvast/vast/logger.hpp:43,
                 from /home/benno/src/vast/master/libvast/vast/format/json.hpp:21,
                 from /home/benno/src/vast/master/libvast/src/system/application.cpp:19:
/home/benno/src/vast/master/libvast/aux/caf/libcaf_core/caf/blocking_actor.hpp: In lambda function:
/home/benno/src/vast/master/libvast/aux/caf/libcaf_core/caf/blocking_actor.hpp:334:15: warning: implicit capture of ‘this’ via ‘[=]’ is deprecated in C++20 [-Wdeprecated]
  334 |     auto cb = [=](receive_cond& rc) mutable {
      |               ^
/home/benno/src/vast/master/libvast/aux/caf/libcaf_core/caf/blocking_actor.hpp:334:15: note: add explicit ‘this’ or ‘*this’ capture
```

dominiklohmann:
> if you build the branch locally in debug mode you get a ton of these warnings for almost every file that interacts with CAF's actor system:

I don't get a single warning using Clang, why would that be different? Did you try with the latest branch and the submodule from the last commit?

lava:
Yes, I used this branch and updated the submodule. Clang doesn't issue a warning for this, so you would not see anything when compiling with clang:

```
benno@bourbaki:~/code/cpp$ cat lambda-this-copy.cpp 
class A {
  void f() {
    int x = [=] {
      return y;
    }();
  }

  int y;
};
benno@bourbaki:~/code/cpp$ clang++ -c lambda-this-copy.cpp -std=c++20
benno@bourbaki:~/code/cpp$ g++ -c lambda-this-copy.cpp -std=c++20
lambda-this-copy.cpp: In lambda function:
lambda-this-copy.cpp:3:13: warning: implicit capture of ‘this’ via ‘[=]’ is deprecated in C++20 [-Wdeprecated]
    3 |     int x = [=] {
      |             ^
lambda-this-copy.cpp:3:13: note: add explicit ‘this’ or ‘*this’ capture
```

tobim:
I think the best way to handle this is to treat bundled CAF headers as system headers. This happens automatically for imported targets in CMake, hence the absence of those warnings when linking against a system provided CAF.

This should be possible in CMake by copying the contents of `INTERFACE_INCLUDE_DIRECTORIES` to `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES`. Check [the docs](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_SYSTEM_INCLUDE_DIRECTORIES.html#prop_tgt:INTERFACE_SYSTEM_INCLUDE_DIRECTORIES).

dominiklohmann:
> This should be possible in CMake

It is indeed. @lava please test if this fixes your issues.

lava (approved):
Things seem to work nicely now. From my testing, cmake is also smart enough to filter out `/usr/include` from the generated `-isystem` directives, avoiding the error described in https://stackoverflow.com/questions/37218953/isystem-on-a-system-include-directory-causes-errors

dominiklohmann (approved):
Let's ship it!

PR #1769, opened 2021-07-09
nix: fix missing flag withPlugin

GTrunSec:
Fixes: missing cmake lib
```
97b7435/build/CMakeFiles/CMakeOutput.log".
vast> See also "/build/vast-97b7435/build/CMakeFiles/CMakeError.log".
error: builder for '/nix/store/j3z1139zyx2jm5i86da0y3fbdh400ply-vast-97b7435-latest-dirty.drv' failed with exit code 1;
       last 10 log lines:
       > -- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE)
       > CMake Error at cmake/VASTRegisterPlugin.cmake:381 (message):
       >   CMAKE_INSTALL_LIBDIR must be a relative path
       > Call Stack (most recent call first):
       >   plugins/broker/CMakeLists.txt:18 (VASTRegisterPlugin)
       >
       >
       > -- Configuring incomplete, errors occurred!
       > See also "/build/vast-97b7435/build/CMakeFiles/CMakeOutput.log".
       > See also "/build/vast-97b7435/build/CMakeFiles/CMakeError.log".
```
- Better Args withPlugin:

```
lib.concatImapStringsSep ";" (pos: x: "plugins/" + x ) [ "broker" "pcap"]                
"plugins/broker;plugins/pcap"
```

### :memo: Checklist

### DIFF
--- a/nix/vast/default.nix
+++ b/nix/vast/default.nix
@@ -74,15 +74,14 @@ stdenv.mkDerivation rec {
   nativeBuildInputs = [ cmake cmake-format ];
   propagatedNativeBuildInputs = [ pkgconfig pandoc ];
   buildInputs = [
-    libpcap
     jemalloc
     libyamlcpp
     simdjson
     spdlog
     robin-map
     libunwind
-    zeek-broker
-  ];
+  ] ++ (if (toString (lib.intersectLists [ "plugins/pcap" ] withPlugins)) == "plugins/pcap" then [ libpcap ] else [])
+   ++ (if (toString (lib.intersectLists [ "plugins/broker" ] withPlugins)) == "plugins/broker" then [ zeek-broker ] else []);
   propagatedBuildInputs = [ arrow-cpp caf flatbuffers ];
 
   cmakeFlags = [
@@ -101,10 +100,12 @@ stdenv.mkDerivation rec {
     "-DVAST_ENABLE_STATIC_EXECUTABLE:BOOL=ON"
     "-DCMAKE_INTERPROCEDURAL_OPTIMIZATION:BOOL=ON"
   ] ++ lib.optional disableTests "-DVAST_ENABLE_UNIT_TESTS=OFF"
-    # Plugin Section
-    ++ lib.optional (withPlugins != [])
-       "-DVAST_PLUGINS=${lib.concatStringsSep ";" withPlugins}"
-    ++ extraCmakeFlags;
+  # Plugin Section
+  ++ lib.optionals (withPlugins != []) [ #Exp: ["plugins/broker" "plugins/pcap"]
+    "-DVAST_PLUGINS=${lib.concatImapStringsSep ";" (pos: x: "${src}/" + x ) withPlugins}"
+    "-DVAST_ENABLE_STATIC_PLUGINS=ON"
+  ]
+  ++ extraCmakeFlags ;
 
   hardeningDisable = lib.optional isStatic "pic";
 
--- a/nix/static-binary.sh
+++ b/nix/static-binary.sh
@@ -15,10 +15,7 @@ target="${STATIC_BINARY_TARGET:-vast}"
 USE_HEAD="off"
 cmakeFlags=""
 # Enable the bundled plugins by default.
-plugins=(
-  "${toplevel}/plugins/broker"
-  "${toplevel}/plugins/pcap"
-)
+plugins=( '"plugins/pcap" "plugins/broker"' )
 
 while [ $# -ne 0 ]; do
   case "$1" in


GTrunSec:
@tobim  The vastPlugin does not work to nix-build.

- lib tree check:
```
 tree ./result/lib/vast 
./result/lib/vast
└── plugins
    ├── libvast-plugin-broker.so
    └── libvast-plugin-pcap.so

1 directory, 2 files
```
Plugin command check:
```
./result/bin/vast import broker                                                                                        
error: unrecognized subcommand                                                                                           
                                                                                                                         
broker                                                                                                                   
^~~~~~                 
```

dominiklohmann:
Thanks for your contribution!

> @tobim The vastPlugin does not work with nix-build. Can you reach out?

@tobim is on holiday this week, so this PR will likely sit until early next week. He's the guy doing all of our Nix stuff, and it's probably best if we let him review this.

nix/vast/default.nix
```diff
@@ -102,10 +101,12 @@ stdenv.mkDerivation rec {
     "-DVAST_ENABLE_STATIC_EXECUTABLE:BOOL=ON"
     "-DCMAKE_INTERPROCEDURAL_OPTIMIZATION:BOOL=ON"
   ] ++ lib.optional disableTests "-DVAST_ENABLE_UNIT_TESTS=OFF"
-    # Plugin Section
-    ++ lib.optional (withPlugins != [])
-       "-DVAST_PLUGINS=${lib.concatStringsSep ";" withPlugins}"
-    ++ extraCmakeFlags;
+  # Plugin Section
+  ++ lib.optionals (withPlugins != []) [#Exp: ["broker" "pcap"]
+    "-DVAST_PLUGINS=${lib.concatImapStringsSep ";" (pos: x: "plugins/" + x ) withPlugins}"
```
dominiklohmann, line 106: If I understand this correctly you prefix `plugins/` automatically. That would break our closed source CI, which passes additional paths that exist outside of this repository.

nix/vast/default.nix
```diff
@@ -102,10 +101,12 @@ stdenv.mkDerivation rec {
...
+    "-DCMAKE_INSTALL_LIBDIR=lib"
```
dominiklohmann, line 107: `lib` should be the default value for `CMAKE_INSTALL_LIBDIR`. Why would you need to specify this manually here?

GTrunSec: `CMAKE_INSTALL_LIBDIR` missing from the Non-static build option of the withPlugins.

nix/vast/default.nix
GTrunSec: Fixed
plugins=('"pcap" "broker"') or "${toplevel}" there nothing different.

nix/vast/default.nix
GTrunSec: > @tobim is on holiday this week, so this PR will likely sit until early next week. He's the guy doing all of our Nix stuff, and it's probably best if we let him review this.

Thanks, I got it. So I did a minimal patch to here which just modified args of withPlugin.
For the Non-static build with Nix,  it might still have a problem with `withPlugin`.  Anyway, I have synced the build to the static.

GTrunSec:
> @tobim is on holiday this week, so this PR will likely sit until early next week. He's the guy doing all of our Nix stuff, and it's probably best if we let him review this.

Thanks, I got it. So I did a minimal patch to here which just modified args of withPlugin.
For the Non-static build with Nix,  it might still have a problem with `withPlugin`.  Anyway, I have synced the build to the static.

nix/vast/default.nix
dominiklohmann: You only hit that codepath if CMAKE_INSTALL_LIBDIR is not a subdirectory of CMAKE_INSTALL_PREFIX, which it always should be.

nix/vast/default.nix
tobim: What @dominiklohmann wrote is correct, we need to support plugins from other paths as well.
I think we should create explicit attributes for every plugin and make a `vastWithPlugins` wrapper derivation, but refactorings like that have low prio at the moment.

nix/vast/default.nix
tobim: https://github.com/tenzir/vast/pull/1786 should fix the core issue, so this workaround will hopefully not be necessary.

nix/vast/default.nix
GTrunSec: How about remove  "plugins/" to plugins=('"plugins/pcap" "plugins/broker" "<otherpath><plguin>"')?

nix/vast/default.nix
GTrunSec: I don't need specific `CMAKE_INSTALL_LIBDIR` anymore. Just ENABLE_STATIC_PLUGIN in default build. https://github.com/tenzir/vast/pull/1769/commits/9069f3b7b5ae10f3e15daf9ac11a4b42b80ef7ec#diff-20342b514d1c1f399318d01baf84c6982f37b7aa789a10a9b944127068e5398eR106

nix/vast/default.nix
tobim: I think we should take a closer look at the problem that this commit solves first. The derivation as written before the change simply added any plugin dependencies as dependencies of VAST itself, which is obviously suboptimal, but the issues goes deeper: One would need to use `vast.overrideAttrs` to add support for plugins with additional dependencies.

A more versatile approach would be to add a `dependencies.nix` file to plugins which need it:
```nix
{ pkgs }:
{
  buildInputs = [ pkgs.libpcap ];
  checkInputs = [ pkgs.tcpdump ];
}
```
We should be able to call this from `vast.nix` and merge all the attributes.

We can't build plugins as separate projects with that approach yet, but it solves at least some of the problems.

GTrunSec: useful advice, Thanks.

PR #1770, opened 2021-07-12
Skip empty config files

dominiklohmann:

Loading an empty config file should not fail with an unintelligable error, but rather function as if the config file did not exist.

### :memo: Checklist

I don't think this needs a changelog entry. Test locally.

### DIFF
--- a/libvast/src/system/configuration.cpp
+++ b/libvast/src/system/configuration.cpp
@@ -176,14 +176,12 @@ caf::error configuration::parse(int argc, char** argv) {
       auto contents = detail::load_contents(config);
       if (!contents)
         return contents.error();
-      // Skip empty config files.
-      if (std::all_of(contents->begin(), contents->end(), [](char ch) {
-            return std::isspace(ch);
-          }))
-        continue;
       auto yaml = from_yaml(*contents);
       if (!yaml)
         return yaml.error();
+      // Skip empty config files.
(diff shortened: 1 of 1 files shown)


libvast/src/system/configuration.cpp
```diff
@@ -176,6 +176,11 @@ caf::error configuration::parse(int argc, char** argv) {
       auto contents = detail::load_contents(config);
       if (!contents)
         return contents.error();
+      // Skip empty config files.
```
mavam, line 179:
```suggestion
      // Skip empty config files.
      if (contents->empty())
        continue;
      // Skip config files only containing whitespace.
```

mavam (approved):
I verified that it works locally.

But, for a 0-byte file, this code path is not triggered. I made a suggestion for factoring this.

libvast/src/system/configuration.cpp
dominiklohmann: > for a 0-byte file, this code path is not triggered

Why would it not be? `all_of` always returns true for empty ranges: https://godbolt.org/z/a5ojGxh93

dominiklohmann: Anyways, pushed a second commit for a separate reason that obsoletes this discussion. See commit message.

mavam: Makes sense. Green light from my side.

PR #1771, opened 2021-07-13
Make the source shutdown instantaneous

dominiklohmann:

This changes the source actor not to do busy work in the processing stage of its stream source so it stays responsive.

### :memo: Checklist

Give it a spin locally with various configurations

### DIFF
--- a/libvast/src/system/source.cpp
+++ b/libvast/src/system/source.cpp
@@ -156,6 +156,22 @@ source(caf::stateful_actor<source_state>* self, format::reader_ptr reader,
     self->state.done = true;
     if (self->state.mgr)
       self->state.mgr->out().push(detail::framed<table_slice>::make_eof());
+    // Spawn a dummy transformer sink.
+    auto dummy = self->spawn(dummy_transformer_sink);
+    self
+      ->request(self->state.transformer, caf::infinite,
+                static_cast<stream_sink_actor<table_slice>>(dummy))
+      .then(
+        [=](caf::outbound_stream_slot<table_slice>) {
+          if (self->state.mgr) {
+            VAST_WARN("{} flushes", self);
+            self->state.mgr->shutdown();
+            self->state.mgr->out().force_emit_batches();
+            self->state.mgr->out().close();
+            self->state.mgr->out().fan_out_flush();
+          }
+        },
+        [](const caf::error&) {});
     self->quit(msg.reason);
   });
   // Spin up the stream manager for the source.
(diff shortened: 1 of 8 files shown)


libvast/vast/system/transformer.hpp
```diff
@@ -51,4 +51,16 @@ transformer_actor::behavior_type
 transformer(transformer_actor::stateful_pointer<transformer_state> self,
             std::string name, std::vector<transform>&&);
 
-} // namespace vast::system
\ No newline at end of file
+/// An actor that hosts a no-op stream sink for table slices, that the SOURCE
+/// and IMPORTER attach to their respective TRANASFORMER actors on shutdown.
```
JoeLoser, line 55: Nit: typo in TRANSFORMER.

dominiklohmann: Thanks!

```suggestion
/// and IMPORTER attach to their respective TRANSFORMER actors on shutdown.
```

dominiklohmann:
[ch27086]

dominiklohmann:
[ch27105]

libvast/src/system/source.cpp
```diff
@@ -166,18 +182,42 @@ source(caf::stateful_actor<source_state>* self, format::reader_ptr reader,
...
+        ->request(caf::actor_cast<source_actor>(self), caf::infinite,
+                  atom::internal_v, atom::run_v, static_cast<uint64_t>(num))
+        .then(
+          [=]() {
+            VAST_DEBUG("{} finished generation of {} messages", self, num);
+          },
+          [=](const caf::error& err) {
+            VAST_WARN("{} failed generation of {} messages: {}", self, num,
+                      err);
+          });
```
lava, line 204: Not sure if the response really adds anything here, wouldn't the run-handler itself already print a warning if it encounters errors? On the other hand, it probably doesn't hurt either, except for readability.

lava (approved):
Overall this looks good to me, and we already extensively discussed the mechanism. I didn't personally test, but the experience from our test bed should be enough imho. So just a few minor comments below.

libvast/src/system/transformer.cpp
```diff
@@ -86,4 +89,33 @@ transformer(transformer_actor::stateful_pointer<transformer_state> self,
...
+        [=](size_t& num_discarded) {
+          num_discarded = 0;
+        },
+        [=](size_t& num_discarded, const std::vector<table_slice>& slices) {
+          const auto num_rows
+            = std::transform_reduce(slices.begin(), slices.end(), size_t{},
+                                    std::plus<>{}, [](const auto& slice) {
+                                      return slice.rows();
+                                    });
+          num_discarded += num_rows;
```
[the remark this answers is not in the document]
dominiklohmann: Not sure what exactly I was thinking here... 🙈

libvast/src/system/source.cpp
dominiklohmann: I don't expect this to ever be triggered, but now that it's in a separate handler having this is good practice. Otherwise, errors will go unreported should we ever decide to add one.

PR #1772, opened 2021-07-13
Fix rendering of table slice encoding

dominiklohmann:

Before this change, the import command favored CAF's string inspector over the more specific to_string, which caused `1` to be printed rather than `arrow`, which did not make much sense in log messages.

### :memo: Checklist

Read it, it's a simple change.


libvast/vast/table_slice_encoding.hpp
```diff
@@ -24,7 +26,25 @@ enum class table_slice_encoding : uint8_t {
...
-std::string to_string(table_slice_encoding encoding) noexcept;
-
 } // namespace vast
+
+namespace fmt {
+
+template <>
+struct formatter<vast::table_slice_encoding> : formatter<std::string_view> {
+  template <typename FormatContext>
+  auto format(const vast::table_slice_encoding& x, FormatContext& ctx) {
```
mavam, line 36:
```suggestion
  auto format(vast::table_slice_encoding x, FormatContext& ctx) {
```

libvast/vast/table_slice_encoding.hpp
```diff
@@ -24,7 +26,25 @@ enum class table_slice_encoding : uint8_t {
...
+    using super = formatter<std::string_view>;
```
mavam, line 37:
```suggestion
    using f = formatter<std::string_view>;
```

I'm only reluctant to use `super` as identifier here because we almost always use it as type alias for the respective base class. Since there's no inheritance here, something simpler (like `f`) might cause less confusion.

mavam (approved):
Assuming you verified the log messages locally, this doesn't warrant further testing.

libvast/vast/table_slice_encoding.hpp
dominiklohmann: It _is_ the base class, hence the name `super`

PR #1773, opened 2021-07-13
Fix Docker image builds for arm64

dominiklohmann:

- `FindSSE.cmake` module that checks for availability of various SSE-related compiler flags.
- Bug fixes in `base64.hpp` to account for the difference between char and signed char on some platforms, e.g., arm64.
- Use GCC-8 specific version of `VAST_DISCARD_ARGS` for all GCC versions; the same issues occured for GCC-10.

### :memo: Checklist

Reason about the changes I've made. Read the code changes carefully.

### DIFF
--- a/libvast/vast/detail/discard.hpp
+++ b/libvast/vast/detail/discard.hpp
@@ -14,9 +14,7 @@
 
 // Ensures all args are used and syntactically valid, without evaluating them.
 
-#if __GNUC__ == 8
-// GCC-8 does weird things here, so we actually evaluate the arguments (when
-// compiling without optimizations).
+#if __GNUC__
 #  define VAST_DISCARD_1(msg) static_cast<void>((msg))
 #else
 #  define VAST_DISCARD_1(msg) static_cast<std::void_t<decltype((msg))>>(0)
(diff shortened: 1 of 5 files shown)


libvast/vast/detail/discard.hpp
```diff
@@ -14,9 +14,7 @@
 
 // Ensures all args are used and syntactically valid, without evaluating them.
 
-#if __GNUC__ == 8
-// GCC-8 does weird things here, so we actually evaluate the arguments (when
-// compiling without optimizations).
+#if __GNUC__
```
mavam, line 17: Isn't this dependent on #1768?

mavam (approved):
The C++ changes look reasonable to me. Glancing over the CMake, I can't find anything weird either.

libvast/vast/detail/discard.hpp
dominiklohmann: We switched to GCC-10 in our Dockerfiles recently and had lots of warnings in Release builds for that. This just enables the alternative discard macro for all versions of GCC instead of just GCC-8.

PR #1777, opened 2021-07-15
Remove /etc as hardcoded sysconfdir from Nix build 

dominiklohmann:

This PR contains two changes:

- Our static binaries behaved differently than our regular relocatable installations: They were looking for the global configuration in `/etc/vast/vast.yaml` rather than the expected `<prefix>/etc/vast/vast.yaml`. This change makes them behave just like regular relocatable installations.
   For users, this means that configuration files at `/etc/vast/vast.yaml` need to be moved accordingly.
- We now install an example config file alongside VAST.

### :memo: Checklist

Read the code, try the static binary from CI.

### DIFF
--- a/.github/workflows/style-check.yaml
+++ b/.github/workflows/style-check.yaml
@@ -35,6 +35,6 @@ jobs:
         pip install --upgrade pip black
         # Note: black fails when it doesn't have to do anything.
         git diff --name-only --no-color --diff-filter=ACM $(git merge-base origin/master HEAD) |
-          grep -v '\(\.json\|\.ipynb\|\.hpp\.in\|\.ref\)$' |
+          grep -v '\(\.json\|\.ipynb\|\.hpp\.in\|\.ref\|.example\)$' |
           2>/dev/null xargs black || true
         git diff --exit-code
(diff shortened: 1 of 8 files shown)


mavam:
Why is the example config all commented out?

dominiklohmann:
> Why is the example config all commented out?

That's intentional: It's an example config file, and if we install it, then we don't want it to mesh with the actual defaults. Unless we find a way to guarantee that defaults and example config are in sync (which we don't right now), I'd rather have everything commented out.

mavam:
I would agree if it's an actual `.yaml` file. But it's an `.example` file that VAST doesn't look at anyway on startup.

The UX of reading a properly syntax-highlighted file and separate the tuning knobs from the content is not to be dismissed, I think.

dominiklohmann:
> I would agree if it's an actual `.yaml` file. But it's an `.example` file that VAST doesn't look at anyway on startup.

Take a look at the changed CMake code in this PR: We install this file to `<prefix>/etc/vast/vast.yaml`.

mavam:
I wonder whether we should simply keep the uncommented example file in the repo and install a transformation (e.g., through `sed`) at the destination. That would make at least maintaining the example config file easier and avoid mistakes.

mavam (approved):
If you can get `awk` installed in Nix, we're in business in CI.

PR #1778, opened 2021-07-15
Disable auto-vectorization in prebuilt Docker images

dominiklohmann:

The following new build options exist:
- `VAST_ENABLE_AUTO_VECTORIZATION` enables/disables all auto-vectorization flags.
- `VAST_ENABLE_SSE_INSTRUCTIONS` enables `-msse`; similar options exist for SSE2, SSE3, SSSE3, SSE4.1, SSE4.2, AVX, and AVX2.

The CI now builds the Docker images hosted on Dockerhub with auto-vectorization disabled. This increases the portability of our hosted Docker images. Users building locally will still have flags enabled that their respective device supports.

### :memo: Checklist

- @tenzir/backend read the CMake code carefully
- @rolandpeelen see if this fixes the issue with AVX2 instructions on your end

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -650,7 +650,9 @@ jobs:
       # We explicitly disable SSE instructions here: The CI runners support it,
       # but we want to support our prebuilt Docker images on older machines as
       # well.
-      VAST_BUILD_OPTIONS: -D VAST_ENABLE_AUTO_VECTORIZATION:BOOL=OFF
+      VAST_BUILD_OPTIONS: >
+        -D VAST_ENABLE_AVX_INSTRUCTIONS:BOOL=OFF
+        -D VAST_ENABLE_AVX2_INSTRUCTIONS:BOOL=OFF
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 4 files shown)


rolandpeelen (approved):

lava:
Did you do any testing on how this impacts performance?

Also, from the description it sounds like only AVX2 support was problematic, so would it make more sense to only turn that off? From a brief glance at Wikipedia, it sounds like all the other features should be supported by basically every CPU made in the last 10 years?

dominiklohmann:
> Did you do any testing on how this impacts performance?

It's only disabled for Docker, which we don't in high-perf scenarios.

> Also, from the description it sounds like only AVX2 support was problematic, so would it make more sense to only turn that off?

We can do that as well, will make that change.

lava (approved):
Changes look good; would still be interesting to know the performance impact so we can display a note somewhere if required, but given that Roland is affected we'd probably merge this anyways even if the impact was high.

dominiklohmann:
> would still be interesting to know the performance impact

None since the last release definitely, because we only added the AVX stuff a week ago.

PR #1784, opened 2021-07-18
Remove explicit deduction guides

JoeLoser:

Problem:
- Some class templates have explicit deduction guides which were needed
  prior to switching to C++20. Now, they are no longer needed since
  `vast` compiles with C++20.

Solution:
- Remove explicit deduction guides.

### :memo: Checklist

File-by-file.


JoeLoser:
Looks like the CI compilers aren't quite ready for this change since they lack full C++20 support.

PR #1786, opened 2021-07-19
Normalize GNUInstallDirs for external plugins

dominiklohmann:

Plugins built against an installed VAST expect relative GNUInstallDirs, which some package managers (looking at you, Nix) stupidly set to absolute paths. This patch externalizes our normalization we do for VAST normally and applies it to all dynamic plugins.

### :memo: Checklist

Test it locally by setting `-D CMAKE_INSTALL_PREFIX=${PWD}/build.install -D CMAKE_INSTALL_LIBDIR=${PWD}/build.install/lib` or similar.


cmake/VASTRegisterPlugin.cmake
```diff
@@ -1,5 +1,48 @@
...
+          "DATAROOT"
+          "DATA"
+          "INFO"
+          "LOCALE"
+          "MAN"
+          "DOC")
+    # Try removing CMAKE_INSTALL_PREFIX with a trailing slash from the full path
+    # to get the correct relative path because some package managers do this
+    # stupid thing where they put absolute paths into variables that are
+    # supposed to be interpreted as relative to the install preifx.
```
tobim, line 26: This comment is too strong. The [documentation](https://cmake.org/cmake/help/v3.21/module/GNUInstallDirs.html) explicitly states:
> ... However, an absolute path is also allowed.

In some cases absolute paths make sense, and we should probably only do this in relocatable mode for plugins too.

dominiklohmann: I toned the comment down, and made the requested change.

tobim (approved):

tobim:
Tested with the static build locally.

PR #1787, opened 2021-07-19
Use bundled CAF by default if submodule is initialized

dominiklohmann:

After recent changes, the VAST configury does not pick up the CAF submodule by default anymore, choosing either a system CAF or returning an error. For a successful build out of the box, `VAST_ENABLE_BUNDLED_CAF` has to be set explicitly if no system CAF is installed. This change makes it so we use the bundled CAF by default when developer mode (i.e., not building VAST as a subproject) is enabled and the submodule is checked out.

### :memo: Checklist

Try locally :)


lava (approved):
I've tested locally that it works.

PR #1790, opened 2021-07-19
Disable SSE for vast-ci Nix build (again)

dominiklohmann:

A recent change caused the vast-ci Nix build not to have SSE disabled. Additionally, this disables AVX instructions for the regular vast Nix build in order not to break compatibility.

### :memo: Checklist

I consider this a bugfix to last week's set of PRs regarding this issue, so no separate changelog entry should be necessary.

To review this, reason about the logic, read the code changes and look at the CI output.

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -709,14 +709,19 @@ jobs:
 
   build-nix:
     needs: cancel-previous-runs
-    name: Nix Static (${{ matrix.args }})
+    name: Nix Static (${{ matrix.nix.target }})
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
-        args:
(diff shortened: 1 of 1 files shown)


tobim (approved):
This is ok, but I think you should be aware that compiling for `x86_64` implies `-msse` and `-msse2`, because every `x86_64` bit ISA support those.

Also, this is good enough as a fix for a regression, but since we have fine-grained control of SIMD levels now we could also just use those with a release build to deploy on certain exotic machines.

dominiklohmann:
> I think you should be aware that compiling for `x86_64` implies `-msse` and `-msse2`,

I am aware. It's only possible to disable these now since they're not available on M1 (which is arm64/v8). We can take the more fine-grained approach somewhere down the road as you said.

PR #1791, opened 2021-07-19
Make has_is_transparent a concept

JoeLoser:

Problem:
- `has_is_transparent` type trait is a bit of boilerplate using
  `std::void_t` which is not needed as of C++20.

Solution:
- Add `transparent` concept and write it using `requires` clause
  instead.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/test/detail/concepts.cpp
+++ b/libvast/test/detail/concepts.cpp
@@ -1,0 +1,24 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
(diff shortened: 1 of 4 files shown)


dominiklohmann (approved):
Confirmed locally that we can use this downstream in our proprietary plugins that used `has_is_transparent` for more than just a static assertion.

JoeLoser:
> Confirmed locally that we can use this downstream in our proprietary plugins that used `has_is_transparent` for more than just a static assertion.

Thanks for confirming. Do you mind fixing up the proprietary plugins after this change?

dominiklohmann:
> Thanks for confirming. Do you mind fixing up the proprietary plugins after this change?

Yes. Since it's usually me updating them I have a pretty clear view of what's affected, and this particular change I was a bit worried about, but luckily it turns out to be okay.

PR #1793, opened 2021-07-19
Make is_container a concept

JoeLoser:

Problem:
- Implementation of `is_container` type trait is a bit noisy due to the
  need of two helper alias templates for checking whether `std::size(t)`
  and `std::data(t)` is well-formed for a given `t`.

Solution:
- Add `container` concept and write it directly in terms of a requires
  expression.
- Rename `container` to `container_t` since the identifier `container`
  cannot be a `concept` entity and a `struct`. In favor of not having
  concepts be prefixed with `is_` (like type traits), rename the struct
  use.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/test/detail/concepts.cpp
+++ b/libvast/test/detail/concepts.cpp
@@ -12,6 +12,7 @@
 
 #include "vast/test/test.hpp"
 
+#include <array>
 #include <type_traits>
 
 TEST(transparent) {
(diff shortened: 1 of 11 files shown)


JoeLoser:
Needs another commit I was working on to remove `std_size_t` and `std_data_t` alias templates in `type_traits.hpp`. Will fix up later tonight.

mavam (approved):
Nice work. This can be merged from my side, but I'll let @tenzir/backend do the final push.

Issue #1779, opened 2021-07-15
Consider use of Pipenv to control Python dependencies (black, cmake-format, etc)


JoeLoser:
In the [style-check CI job](https://github.com/tenzir/vast/blob/7a298624e5e66b32d406ee131eb29c545f601a87/.github/workflows/style-check.yaml), there are some checks which make sure a changeset is formatted correctly with some tools such as `black` and `cmake-format`.

In particular, I often am on the latest version of many tools, but the formatting expected from a project corresponds to an older version of the tool. For example, `black` frequently changes formatting even in minor upgrades.

I recommend providing a `Pipfile` and `Pipfile.lock` to improve the workflow when doing local development. With this, we can pin the exact versions used for these Python modules so I don't run into unexpected formatting differences compared to what CI is using.

tobim:
> With this, we can pin the exact versions used for these Python modules so I don't run into unexpected formatting differences compared to what CI is using.

We already have Nix based tooling to set up a full development environment including dev dependencies. I would strongly suggest we try to integrate CI formatting checks with that instead of investing effort into a problem that is essentially already solved. Having a single tool for dependency management for a polyglot codebase like we have brings a couple of advantages, especially considering python wheels often bundle precompiled C or C++ libraries which might introduce conflicts.

Only if that turns out to be unsuitable we should look at alternatives that focus on the python ecosystem specifically. The ones that come to mind are [Poetry](https://python-poetry.org/), [pip-tools](https://github.com/jazzband/pip-tools), and [pipenv](https://github.com/pypa/pipenv).

dominiklohmann:
It's an issue with many facets. On the one hand, using Nix in CI would solve this perfectly because it truly locks down dependencies. On the other hand, recommending Nix for local development is just not practical, because it increases the barrier to entry by just so much.

I personally agree with the `Pipfile` approach. It also has Dependabot support, so that would make staying on latest a breeze.

One additional approach we could take is to add `format` and `check-format` targets In our CMake, which would then do the correct thing for the user automatically.

dominiklohmann:
For the record: In 0d889e3 we switched to installing clang-format from Linuxbrew in CI for proper formatting of concepts.

PR #1795, opened 2021-07-21
Integral concepts to clean up SFINAE

JoeLoser:

Problem:
- There is a fair amount of SFINAE and/or `static_assert`s used to
  constrain various function templates with integral types which is
  verbose and noisy.

Solution:
- Add `integral` concept and refinement concepts for it:
  `signed_integral` and `unsigned_integral`. These are as-defined in the
  `<concepts>` header that is not yet available in the standard
  libraries for compilers used by Vast.
- Apply these concepts to clean up some SFINAE and `static_assert`
  usage. Notably, this cleans up `vast::word` quite a bit.

### :memo: Checklist

File-by-file.

### DIFF
--- a/.github/workflows/style-check.yaml
+++ b/.github/workflows/style-check.yaml
@@ -5,7 +5,7 @@ env:
 jobs:
   style:
     name: Style Check
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
 
     - uses: actions/checkout@v2
@@ -18,8 +18,8 @@ jobs:
 
     - name: Clang Format
       run: |
-        sudo apt-get -qq update
-        sudo apt-get -qqy install clang-format
+        brew install llvm
+        export PATH="$(brew --prefix llvm)/bin:${PATH}"
         git diff -U0 --no-color $(git merge-base origin/master HEAD) |
           scripts/clang-format-diff.py -p1
 
--- a/libvast/vast/word.hpp
+++ b/libvast/vast/word.hpp
@@ -9,6 +9,7 @@
 #pragma once
 
 #include "vast/detail/assert.hpp"
+#include "vast/detail/concepts.hpp"
 #include "vast/detail/type_traits.hpp"
 
 #include <cstdint>
@@ -17,20 +18,7 @@
 
 namespace vast::detail {
 
-template <class T>
-constexpr bool is_unsigned_integral_v
-  = std::is_unsigned_v<T> && std::is_integral_v<T>;
-
-// Type alias used for SFINAE of free functions below. If we would simply use
-// typename word<T>::size_type, the static_assert would fire. With this work
-// around, we avoid going through word<T> so that overload resolution can
-// proceed.
-template <class T>
-using word_size_type = std::conditional_t<
-  is_unsigned_integral_v<T>,
-  uint64_t,
-  void
->;
+using word_size_type = uint64_t;
 
 } // namespace vast::detail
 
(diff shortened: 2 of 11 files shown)


libvast/vast/word.hpp
```diff
@@ -314,22 +271,14 @@ static constexpr auto rank(T x)
...
 /// @returns *rank_i(x)*.
 /// @pre `i < width`
-template <bool Bit = true, class T>
-static constexpr auto rank(T x, detail::word_size_type<T> i)
--> std::enable_if_t<
-  Bit && detail::is_unsigned_integral_v<T>,
-  detail::word_size_type<T>
-> {
+template <bool Bit = true, detail::unsigned_integral T>
+requires(Bit) static constexpr auto rank(T x, detail::word_size_type<T> i) {
```
dominiklohmann, line 275: Is there some clang-format rule to put this in a line of its own or in the previous line?

dominiklohmann (approved):
This all looks good to me. Trusting CI on this one. Just had one small question regarding formatting.

libvast/vast/word.hpp
JoeLoser: In my experience, the `clang-format` rules for `requires` expressions, concepts,  and `requires requires` is all pretty poor until llvm 12. I use llvm 12.0.1 locally, which may differ for clang-format of this function template (and others).

There aren't many knobs we can set in `.clang-format` yet. In llvm12, they introduced `BreakBeforeConceptDeclaration` which is `true` by default it seems. This is what it up doing locally for me.

If we update the CI compilers even more-so, I'm happy to add that to the `.clang-format` file. I doubt Bullseye is ready for llvm12/gcc11 yet though, otherwise I suspect we would have jumped right to those?

JoeLoser: It seems the style format CI job is failing due to the differences I mentioned (similar in vein to https://github.com/tenzir/vast/issues/1779).

dominiklohmann: I have no issue with updating the style check workflow to use a newer clang-format regardless of the platforms we support. We can easily have higher requirements for developing than for running VAST.

I think the easiest way to fix this is to `brew install llvm` in that workflow to get latest llvm from Linuxbrew, which is pre-installed in the Ubuntu images in GitHub Actions.

Do you mind if I take this branch over and push the changes directly onto this?

JoeLoser: Sure, feel free to run with it for fixing up `clang-format` changes. I just disabled `clang-format` in `concepts.hpp` to avoid the error which should work, but it's not what I want to write. I want code to format concepts and requires etc nicely which comes mostly in `llvm12`. There are a few bugs with formatting in llvm trunk with concepts and friends, but none get hit in this PR.

dominiklohmann: I've made the change, let's see if it runs through.

.github/workflows/style-check.yaml
```diff
@@ -5,7 +5,7 @@ env:
 jobs:
   style:
     name: Style Check
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-llatest
```
JoeLoser, line 8: Nit: typo. `s/llatest/latest`.

libvast/vast/word.hpp
JoeLoser: You made a typo in `latest`.  If we fix that and remove the `clang-format off` section in `concepts.hpp` and it works, I'm good with it.

Otherwise, as-is with this PR, the `clang-format off` toggle in `concepts.hpp` is good enough to pass the style check right now.

dominiklohmann: Eh, seems like there's something wrong with the clang-format bottle for Linuxbrew. I'll take another look in a few minutes.

JoeLoser: No worries. Feel free to do it as a follow-up PR and we can drop that commit for this branch.

dominiklohmann: Installation seems to have worked. Can you take it over from here again?

JoeLoser: Yep. Just removed the explicit `clang-format off` toggle in `concepts.hpp`. We'll see if the style job passes now 🤞 .

Thanks for the help. 😄

PR #1797, opened 2021-07-23
Keep layer-2 framing when reading PCAP payload

mavam:

This PR changes the behavior of reading and writing PCAPs and that VAST no longer strips layer-2 information. VAST preserves the original framing on reading writes out the data with [`DLT_EN10MB`](https://linux.die.net/man/7/pcap-linktype) instead of `DLT_RAW`.

### :memo: Checklist

- [x] Update test baselines

Take a trace with link-layer framing, then:

```
# Import the trace.
vast -N import pcap < trace

# Import the trace, should have link-layer framing.
vast -N export pcap | tcpdump -nle -r -
```


plugins/pcap/main.cpp
```diff
@@ -446,15 +446,19 @@ class reader : public format::single_layout_reader {
...
 #endif
       // Assemble packet.
-      const auto* layer3_ptr = reinterpret_cast<const char*>(layer3.data());
-      auto packet = std::string_view{std::launder(layer3_ptr), layer3.size()};
+      auto make_string_view = [](std::span<const std::byte> bytes) {
+        auto data = reinterpret_cast<const char*>(bytes.data());
+        auto size = bytes.size();
+        return std::string_view{data, size};
+      };
+      auto payload = make_string_view(raw_frame);
```
JoeLoser, line 454: It may be better to add an overload or generalize the `add` function for the builder to work with more contiguous types of characters (e.g. `std::span`) rather than requriing a `std::string_view` only.

mavam: I thought about making that a generic utility, like we have for `as_bytes`. For example, `as_string_view` should work for any byte container. This way, we don't have to force the call site to be templated and pollute the headers.

Would you be on board with `as_string_view`?

JoeLoser: That would work, but I see an inconvenience that the `add` function only works with `std::string_view`. If it worked with spans for example, we wouldn't need to make a `std::string_view` at all.

mavam: I think we can add an overload for byte spans as well, actually.

mavam: The builder functions take a `data_view`, which is variant over concrete views. Unfortunately `std::string_view` is just an alias instead of a strong typedef. Otherwise we could construct a string view from a byte span just fine. We want to convert all elements of `data` and all views to strong typedefs in the medium term. Until then, I'd say we keep the verbosity at the call site.

mavam: @JoeLoser (and @tenzir/backend) what do you think of an `as<T>` facility? With concepts, I think we can constrain it properly. In my eyes, it would make sense to have it as general-purpose, inexpensive (zero-cost, to be precise) conversion of "congruent" types, where accessing the same data is more convenient through a different API.

tobim (changes requested):
I tested this locally and verified that we now get ethertype info when piping `pcap.packets` to `tcpdump`:
```
1573834947.165943 a0:f3:e4:46:8c:8d > b0:33:a6:63:22:dd, ethertype 802.1Q (0x8100), length 287: vlan 34, p 0, ethertype IPv4 (0x0800), 216.169.5.110.56383 > 198.128.153.23.443: Flags [P.], seq 1:218, ack 1, win 2058, options [nop,nop,TS val 2149504310 ecr 1470692848], length 217
```

It would be valuable to check this in the "Pivot with PCAP" integration test. Can you add the flag `-e` to the options for `tcpdump` in the last step?

mavam:
> Can you add the flag `-e` to the options for `tcpdump` in the last step?

The payload of the packet already required a baseline update, which prepended the Ethernet header.

Also, the output of this particular integration test doesn't differ. I tried appending `-e` but it looks like the trace doesn't have Ethernet headers in the first place:

```diff
-1540889849.325075 IP 192.168.168.100.61206 > 72.247.178.18.80: Flags [SEW], seq 1646017262, win 65535, options [mss 1460,nop,wscale 6,nop,nop,TS val 3114244558 ecr 0,sackOK,eol], length 0
+1540889849.325075 ip: 192.168.168.100.61206 > 72.247.178.18.80: Flags [SEW], seq 1646017262, win 65535, options [mss 1460,nop,wscale 6,nop,nop,TS val 3114244558 ecr 0,sackOK,eol], length 0
```

tobim:
That is strange, it does differ for me:
``` diff
-1540889849.325075 IP 192.168.168.100.61206 > 72.247.178.18.80: Flags [SEW], seq 1646017262, win 65535, options [mss 1460,nop,wscale 6,nop,nop,TS val 3114244558 ecr 0,sackOK,eol], length 0
+1540889849.325075 78:7b:8a:b7:92:ea > c0:3f:0e:bd:6f:d2, ethertype IPv4 (0x0800), length 78: 192.168.168.100.61206 > 72.247.178.18.80: Flags [SEW], seq 1646017262, win 65535, options [mss 1460,nop,wscale 6,nop,nop,TS val 3114244558 ecr 0,sackOK,eol], length 0
```

Are you sure you ran it with the correct binary?

mavam:
> Are you sure you ran it with the correct binary?

Maybe `tcpdump` differences are the culprit? Your TCP options also look much more verbose than mine.

tobim:
The delta between our outputs is:
```diff
-ip
+78:7b:8a:b7:92:ea > c0:3f:0e:bd:6f:d2, ethertype IPv4 (0x0800), length 78
```
Which seems to me that your tcpdump implementation doesn't support `-e`. I can accept the change as is for the sake of portability, but in general I don't like to rely on binary output in these tests because it's very cumbersome to verify that it's correct.

tobim (approved):

PR #1798, opened 2021-07-23
Add missing concepts for Suricata events

tobim:
### :memo: Checklist


dominiklohmann:
Is this exempt from the feature freeze? Should we merge this and add an rc2?

tobim:
In my opinion this is very low risk, but I would not want to force an rc2 just for this.

dominiklohmann:
Then let's only merge this if we need another mixup.

lava (approved):
I did not verify in detail that the updated concept fields are correct, but the gist of the PR looks good to me.

PR #1764, opened 2021-07-08
Use a unique version for plugins

dominiklohmann:
### :notebook_with_decorative_cover: Description

This changes plugin versions to be set automatically by the `VASTRegisterPlugin` CMake function. A plugin version consists of three parts:
- The CMake project version of the plugin
- The Git revision of the last commit that touched the plugin
- A `dirty` suffix for uncommited changes to the plugins

The `VAST_REGISTER_PLUGIN` macro no longer takes the plugin version in the C++ code, which may require plugins to be updated.

### :memo: Checklist

- [x] All user-facing changes have changelog entries.
- [x] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [x] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

Run locally to test the dirty indicator and that rebuilds pick up changes without reconfiguration.

PR #1799, opened 2021-07-23
Support unique plugin versions for Nix and Docker

dominiklohmann:

This PR adds the necessary scaffolding to support the new plugin versioning from #1764 for Nix and Docker.

### :memo: Checklist

To be taken over by @tobim for the Nix fixups and the changelog entry, then to be reviewed commit-by-commit.

### DIFF
--- a/nix/static-binary.sh
+++ b/nix/static-binary.sh
@@ -43,10 +43,23 @@ while [ $# -ne 0 ]; do
       USE_HEAD="on"
       ;;
     --with-plugin=*)
-      plugins+=("${optarg}")
+      plugins+=("${optarg%/}")
       ;;
   esac
   shift
+done
+
+plugin_version() {
+  local plugin="$1"
+  local name="${plugin##*/}"
+  local key="VAST_PLUGIN_${name^^}_REVISION"
+  local value="$(git -C "${plugin}" rev-list --abbrev-commit --abbrev=10 -1 HEAD -- "${plugin}")"
+  echo "-D${key}=${value}"
+}
+
+# Get Plugin versions
+for plugin in "${plugins[@]}"; do
+  cmakeFlags="${cmakeFlags} \"$(plugin_version ${plugin})\""
 done
 
 if [ "${USE_HEAD}" == "on" ]; then
--- a/nix/static-binary.sh
+++ b/nix/static-binary.sh
@@ -56,7 +56,7 @@ while [ $# -ne 0 ]; do
       USE_HEAD="on"
       ;;
     --with-plugin=*)
-      plugins+=("${optarg%/}")
+      plugins+=("$(realpath "${optarg}")")
       ;;
   esac
   shift
--- a/nix/static-binary.sh
+++ b/nix/static-binary.sh
@@ -4,6 +4,19 @@
 nix --version
 nix-prefetch-github --version
 
+usage() {
+  printf "usage: %s [options]\n" $(basename $0)
+  echo
+  echo 'options:'
+  echo "    -h,--help               print this message"
+  echo "       --use-head           build from the latest commit instead of your current"
+  echo "                            working copy"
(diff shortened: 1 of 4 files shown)


tobim (approved):
The changes to the docker file and CMake look fine to me.

nix/static-binary.sh
```diff
@@ -43,12 +43,25 @@ while [ $# -ne 0 ]; do
       USE_HEAD="on"
       ;;
     --with-plugin=*)
-      plugins+=("${optarg}")
+      plugins+=("${optarg%/}")
```
lava, line 46: It would probably nice to add a comment documenting that this expects an array of local git repositories, with directory names corresponding to plugin names; it's not super-obvious from looking at the functions below.

(also, while writing this comment I noticed that `usage` in line 39 appears to be undefined.

nix/static-binary.sh
```diff
@@ -43,12 +43,25 @@ while [ $# -ne 0 ]; do
...
       ;;
   esac
   shift
 done
 
+plugin_version() {
+  local plugin="$1"
+  local name="${plugin##*/}"
+  local key="VAST_PLUGIN_${name^^}_REVISION"
+  local value="$(git -C "${plugin}" rev-list --abbrev-commit --abbrev=10 -1 HEAD -- "${plugin}")"
```
lava, line 56: The last argument seems unnecessary? At least in a local test, this version was not able to pick up the correct flags for the analyzer plugin:

```
benno@bourbaki:~/src/vast/reviews/nix$ ./static-binary.sh --with-plugin=../examples/plugins/analyzer
nix (Nix) 2.3.10
nix-prefetch-github, version 4.0.3
rev is e5767db6626854b9d77e1869ab96ccafbe64b7f8
running nix-build --no-out-link -E \'with import /home/benno/src/vast/reviews/nix {};
  pkgsStatic."vast".override {
    versionOverride = "2021.07.29-rc1-4-ge5767db662-dirty";
    withPlugins = [ /home/benno/src/vast/reviews/plugins/broker /home/benno/src/vast/reviews/plugins/pcap ../examples/plugins/analyzer ];
    extraCmakeFlags = [  "-DVAST_PLUGIN_BROKER_REVISION=da276dfc53" "-DVAST_PLUGIN_PCAP_REVISION=da276dfc53" "-DVAST_PLUGIN_ANALYZER_REVISION=" ];
  }\'
[...]
```

lava (changes requested):
There still seems to be a bug in the `git rev-list` invocation, or I misunderstood the usage.

nix/static-binary.sh
tobim: The last argument is necessary, because we want to get the hash of the most recent commit that modified any file below that path. If the path is omitted we would get the equivalent of `git rev-parse HEAD`.

Regarding the failure: The script (and also our CMake) only work with absolute paths to plugins. I'll add support for relative paths to this script. They will be interpreted relative to CWD of the caller.

nix/static-binary.sh
tobim: The `--with-plugin=` option can be given multiple times. I'll fix the usage part.

dominiklohmann (commented):
Note that while this works, one of the remaining deficiencies is that plugin directory name does not necessarily equal plugin target name.

lava (approved):
I tried this locally and it works now as expected.

nix/static-binary.sh
lava: Hm, `git rev-parse HEAD` sounds like a reasonable "revision" for a plugin, I can imagine that it will depend on headers outside of the specific folder (like all of our current plugins) so just the latest change of the plugin itself may be ambiguous. But that's not an issue we need to solve right now.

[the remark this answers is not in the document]
tobim: `--abbrev=` works in combination with `--abbrev-commit`, another small weirdness of the git cli.

PR #1800, opened 2021-07-24
Parseable concepts to simplify SFINAE

JoeLoser:

Problem:
- There is a lot of variable templates with cumbersome metaprogramming
  in defining them.
- There is some verbose SFINAE for constraining various parseable
  concepts.

Solution:
- Add various concepts and apply them to simplify the metaprogramming.

### :memo: Checklist

Commit-by-commit.

### DIFF
--- a/libvast/vast/access.hpp
+++ b/libvast/vast/access.hpp
@@ -8,76 +8,34 @@
 
 #pragma once
 
-#include <type_traits>
-
 namespace vast {
 
 /// Wrapper to encapsulate the implementation of concepts requiring access to
 /// private state.
 struct access {
-  template <class, class = void>
+  template <class>
   struct state;
 
-  template <class, class = void>
+  template <class>
   struct parser;
 
-  template <class, class = void>
+  template <class>
   struct printer;
-
-  template <class, class = void>
-  struct converter;
 };
-
-namespace detail {
-
-struct has_access_state {
-  template <class T>
-  static auto test(T* x) -> decltype(access::state<T>{}, std::true_type());
-
-  template <class>
-  static auto test(...) -> std::false_type;
-};
-
-struct has_access_parser {
-  template <class T>
-  static auto test(T* x) -> decltype(access::parser<T>{}, std::true_type());
-
-  template <class>
-  static auto test(...) -> std::false_type;
-};
-
-struct has_access_printer {
-  template <class T>
-  static auto test(T* x) -> decltype(access::printer<T>{}, std::true_type());
-
-  template <class>
-  static auto test(...) -> std::false_type;
-};
-
-struct has_access_converter {
-  template <class T>
-  static auto test(T* x) -> decltype(access::converter<T>{}, std::true_type());
-
-  template <class>
-  static auto test(...) -> std::false_type;
-};
-
-} // namespace detail
 
 template <class T>
-constexpr bool has_access_state_v
-  = decltype(detail::has_access_state::test<T>(0))::value;
+concept access_state = requires {
+  access::state<T>{};
+};
 
 template <class T>
-constexpr bool has_access_parser_v
-  = decltype(detail::has_access_parser::test<T>(0))::value;
+concept access_parser = requires {
+  access::parser<T>{};
+};
 
 template <class T>
-constexpr bool has_access_printer_v
-  = decltype(detail::has_access_printer::test<T>(0))::value;
-
-template <class T>
-constexpr bool has_access_converter_v
-  = decltype(detail::has_access_converter::test<T>(0))::value;
+concept access_printer = requires {
+  access::printer<T>{};
+};
 
 } // namespace vast
--- a/libvast/vast/access.hpp
+++ b/libvast/vast/access.hpp
@@ -17,7 +17,7 @@ struct access {
   struct state;
 
   template <class>
-  struct parser;
+  struct parser_base;
 
   template <class>
   struct printer;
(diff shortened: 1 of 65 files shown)


libvast/vast/access.hpp
```diff
@@ -8,76 +8,34 @@
 
 #pragma once
 
-#include <type_traits>
-
 namespace vast {
 
 /// Wrapper to encapsulate the implementation of concepts requiring access to
 /// private state.
 struct access {
```
mavam, line 15: Something for the future: this type exists only to give some parsers/printers access to the private state of befriended types. It's redundant to our `inspect` overloads, which we should use instead. `git grep` reveals only `address`, `subnet`, and `pattern` still using this.

libvast/vast/concept/parseable/core/parser.hpp
```diff
@@ -130,7 +119,7 @@ template <class T>
 using is_parser_t = typename is_parser<T>::type;
 
 template <class T>
-constexpr bool is_parser_v = std::is_base_of<parser<T>, T>::value;
+concept is_parser_v = is_parser<std::decay_t<T>>::value;
```
mavam, line 122:
```suggestion
concept parser = is_parser<std::decay_t<T>>::value;
```

Would this be possible, or be name conflict? (Same for `printer`.)

libvast/vast/concept/parseable/to.hpp
```diff
@@ -19,18 +19,17 @@
...
-  -> std::enable_if_t<is_parseable_v<Iterator, To>, caf::expected<To>> {
+requires(parseable<Iterator, To>) auto to(Iterator& f, const Iterator& l)
+  -> caf::expected<To> {
   caf::expected<To> t{To{}};
   if (!parse(f, l, *t))
     return caf::make_error(ec::parse_error);
   return t;
 }
 
 template <class To, class Range>
```
mavam, line 30: `Range` would also make for a great concept, with a specialization for character ranges for this function.

mavam (commented):
All great stuff, thanks!

I wonder how we deal with naming conflicts, e.g., the `parseable` concept vs. the `parser` base class. I found [P1754](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1754r0.pdf) on this, which suggests the following:
- Use a name that is an adjective or abstract noun. (Prefer names like `regular` and `swappable` and `color`. Avoid names like `has constructor` and `red.`)
- Occasionally, use a `_type` suffix as a concession to avoiding name collisions, usually for very general concepts with very common names.

The reason why I'm bringing this up is that `is_parser_v` doesn't really look like a concept. I understand that it makes for a drop-in replacement of the type trait, but as we gain experience with concepts, this is a good discussion opportunity.

libvast/vast/concept/parseable/core/parser.hpp
JoeLoser: Good thought -- it would be a name conflict. I was planning on renaming some things in a follow-up PR. E.g. change the `parser` class type to be `parser_t` and then rename the concept `is_parser_v` to `parser`. Same with `printer`. I thought that would introduce more noise than value for review purposes of this PR.

libvast/vast/access.hpp
JoeLoser: Thanks for the insight. We should clean that up in a follow-up PR then.

libvast/vast/concept/parseable/to.hpp
JoeLoser: Agreed. I'm starting to see more places where good concepts could help as I'm perusing more of the codebase with my C++20 lens in mind.

libvast/vast/concept/parseable/core/parser.hpp
mavam: Nice idea. I agree that the renaming could be a bit noisy in here. Although in general, with clear review instructions, you can make the reviewer look at your PR the way you want...most of the time. 😉 

Regarding the concrete renaming: how about we use something like `basic_parser` or `parser_base`? It's a bit more semantic that `*_t`.

mavam (approved):
We discussed the follow-up renaming in the conversation. With this being part 1 out of 2, green light from my side.

(I'll let @tenzir/backend merge, though, in case I'm missing something obvious with concepts.)

libvast/vast/concept/parseable/core/parser.hpp
JoeLoser: I don't feel strongly on the rename. I like `parser_base` too, so we can go with that.

lava:
Hi, I'm adding a 'blocked'-label since we're currently in feature freeze until our next release on thursday.

Based on a quick glance this looks good though, so if you want to continue conceptifying things feel free to continue off of this branch until then :D

JoeLoser:
> Hi, I'm adding a 'blocked'-label since we're currently in feature freeze until our next release on thursday.
> 
> Based on a quick glance this looks good though, so if you want to continue conceptifying things feel free to continue off of this branch until then :D

Sounds good, thanks for the heads up 👍

JoeLoser:
> We discussed the follow-up renaming in the conversation. With this being part 1 out of 2, green light from my side.
> 
> (I'll let @tenzir/backend merge, though, in case I'm missing something obvious with concepts.)

I just pushed a few commits to do some renaming of the concepts and `parser -> parser_base` to avoid the name conflict.

mavam (approved):
Looked at the new changes and they work out nicely. Looking forward to merge this after Thu!

PR #1803, opened 2021-07-27
Normalize static binary artifact name

tobim:

Until this change, the static binary script used to reuse the vast
version string for the filename of the final tarball. This was
inconsistent with the artifacts of dynamic builds that we do in CI
for Debian and macOS, where the commit hash is omitted for tagged
revisions.

### DIFF
--- a/nix/static-binary.sh
+++ b/nix/static-binary.sh
@@ -7,6 +7,7 @@ nix-prefetch-github --version
 dir="$(dirname "$(readlink -f "$0")")"
 toplevel="$(git -C ${dir} rev-parse --show-toplevel)"
 desc="$(git -C ${dir} describe --tags --long --abbrev=10 --dirty)"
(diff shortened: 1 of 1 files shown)


nix/static-binary.sh
```diff
@@ -83,5 +84,5 @@ tar -C "${result}" \
   --exclude share/vast/test \
   --exclude share/vast/integration \
   --mode='u+w' \
-  -cvzf "$PWD/build/${target}-${desc}-static.tar.gz" \
```
lava, line 86: Is the `-linux` suffix for consistency with the other CI artifacts?

lava (approved):
Tested locally that it omits the commit hash for tagged versions as expected.

nix/static-binary.sh
tobim: Exactly.

PR #1805, opened 2021-07-28
Restore unique IDs for partition-local stores

lava:

Restore the ability to use offsets/ids when working with partition-local stores. This should also make it possible to use `vast explore` and `vast get` in combination with custom store backends.

### :memo: Checklist

Review commit-by-commit. Some points worth considering:

 * Should the query contain `optional<ids>` instead of ids, to distinguish between "no ids given" and "the empty set of ids requested"?

### DIFF
--- a/libvast/test/system/meta_index.cpp
+++ b/libvast/test/system/meta_index.cpp
@@ -46,6 +46,8 @@ partition_synopsis make_partition_synopsis(const vast::table_slice& ts) {
   auto result = partition_synopsis{};
   auto synopsis_opts = caf::settings{};
   result.add(ts, synopsis_opts);
+  result.offset = 0;
+  result.events = ts.rows();
   return result;
 }
 
@@ -159,7 +161,8 @@ struct fixture : public fixtures::deterministic_actor_system_and_events {
     q += hhmmss;
     q += ".0";
     std::vector<uuid> result;
-    auto rp = self->request(meta_idx, caf::infinite, unbox(to<expression>(q)));
+    auto rp = self->request(meta_idx, caf::infinite, vast::atom::candidates_v,
+                            unbox(to<expression>(q)), vast::ids{});
     run();
     rp.receive(
       [&](std::vector<uuid> candidates) { result = std::move(candidates); },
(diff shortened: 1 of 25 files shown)


libvast/vast/system/meta_index.hpp
```diff
@@ -11,11 +11,13 @@
 #include "vast/fwd.hpp"
 
 #include "vast/detail/flat_map.hpp"
+#include "vast/detail/range_map.hpp"
 #include "vast/fbs/index.hpp"
 #include "vast/fbs/partition.hpp"
 #include "vast/ids.hpp"
 #include "vast/partition_synopsis.hpp"
 #include "vast/qualified_record_field.hpp"
+#include "vast/query.hpp"
```
tobim, line 20: unneeded

libvast/vast/system/actors.hpp
```diff
@@ -211,7 +211,7 @@ using index_actor = typed_actor_fwd<
   caf::reacts_to<accountant_actor>,
   // Subscribes a FLUSH LISTENER to the INDEX.
   caf::reacts_to<atom::subscribe, atom::flush, flush_listener_actor>,
-  // Evaluatates an query.
+  // Evaluates an query.
```
tobim, line 213: a query

libvast/vast/system/actors.hpp
```diff
@@ -200,8 +200,8 @@ using meta_index_actor = typed_actor_fwd<
   // Erase a single partition synopsis.
   caf::replies_to<atom::erase, uuid>::with<atom::ok>,
   // Evaluate the expression.
-  caf::replies_to<expression>::with< //
-    std::vector<uuid>>>::unwrap;
+  caf::replies_to<atom::candidates, vast::expression,
```
tobim, line 202: There is nothing to disambiguate, so why the atom?

libvast/test/system/meta_index.cpp
```diff
@@ -173,8 +176,8 @@ struct fixture : public fixtures::deterministic_actor_system_and_events {
 
   auto lookup(meta_index_actor& meta_idx, std::string_view expr) {
     std::vector<uuid> result;
-    auto rp
-      = self->request(meta_idx, caf::infinite, unbox(to<expression>(expr)));
+    auto rp = self->request(meta_idx, caf::infinite, vast::atom::candidates_v,
+                            unbox(to<expression>(expr)), vast::ids{});
```
tobim, line 180: We need a test with Ids and one with Ids and expression too now.

tobim (changes requested):
This seems to be in decent shape already, I think with a little effort we can get this over the line with only one more iteration.

There are still a few TODOs and FIXMEs in the code that should probably all get resolved too.

libvast/vast/system/actors.hpp
lava: As discussed offline, I'd like to eventually get *all* handlers to use an atom, since that can be thought of as the "name" of the handler and it enables easy searching for all call sites and all implementations of this interface.

tobim (approved):
Pair review with @lava and verified that all integration tests pass with the local segment store now.

PR #1809, opened 2021-07-29
Fix another segfault in the partition deletion logic

tobim:

This time the index forgot to check the validity of a requested partition chunk. The relevant error handling below was also slightly modified to log the cause of failed `mmap(2)`s and prevent similar problems in the future.

### :memo: Checklist

By commit.


lava (approved):
The fix looks correct; a unit test might have been nice but also annoying to write this since ultimately requires a system error to trigger and results in a segfault which isn't really testable so our testbed testing should suffice.

libvast/src/chunk.cpp
```diff
@@ -75,11 +75,15 @@ caf::expected<chunk_ptr> chunk::mmap(const std::filesystem::path& filename,
...
                            fmt::format("failed to open file {}", filename));
   auto map = ::mmap(nullptr, size, PROT_READ, MAP_SHARED, fd, offset);
+  auto mmap_errno = errno;
   ::close(fd);
   if (map == MAP_FAILED)
     return caf::make_error(ec::filesystem_error,
-                           fmt::format("failed to mmap file {}", filename));
-  auto deleter = [=]() noexcept { ::munmap(map, size); };
+                           fmt::format("failed to mmap file {}: {}", filename,
+                                       strerror(mmap_errno)));
```
JoeLoser, line 83: We could use `std::strerror` if you prefer. In general, I prefer using the standard library equivalents rather than the POSIX calls directly.

JoeLoser:
> The fix looks correct; a unit test might have been nice but also annoying to write this since ultimately requires a system error to trigger and results in a segfault which isn't really testable so our testbed testing should suffice.

I'd like to see a unit test too. Perhaps we need to build out some testing infrastructure to be able to fake out the `mmap` call and have it return an error so we can test this code. I'm OK with that coming as a follow-up after this release though given the timing of things.

tobim:
> > The fix looks correct; a unit test might have been nice but also annoying to write this since ultimately requires a system error to trigger and results in a segfault which isn't really testable so our testbed testing should suffice.
> 
> I'd like to see a unit test too. Perhaps we need to build out some testing infrastructure to be able to fake out the `mmap` call and have it return an error so we can test this code. I'm OK with that coming as a follow-up after this release though given the timing of things.

I think it shouldn't actually be that hard, we can just mock the filesystem, but we'll have to delay that until after the release.

PR #1811, opened 2021-07-29
Fix compilation with Arrow 5.0

tobim:
### :memo: Checklist

The nix build still uses Arrow 4, so it should be good if CI passes.

### DIFF
--- a/examples/jupyter/requirements.txt
+++ b/examples/jupyter/requirements.txt
@@ -4,5 +4,5 @@ matplotlib>=3.2
 networkx>=2.4
 numpy>=1.18
 pandas>=1.0
-pyarrow>=0.17,<5.0
+pyarrow>=0.17
 scipy>=1.4.1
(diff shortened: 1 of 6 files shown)


examples/jupyter/requirements.txt
```diff
@@ -4,5 +4,5 @@ matplotlib>=3.2
 networkx>=2.4
 numpy>=1.18
 pandas>=1.0
-pyarrow>=0.17,<5.0
+pyarrow>=0.17
```
mavam, line 7: Would it make sense to provide a tight instead of just a lower bound? Explicitly ratcheting up a volatile dependency seems prudent, given that we've been bitten several times with this in the past.

tobim: We should fix this managing the dependencies with [poetry](https://python-poetry.org/) instead.
Related to https://github.com/tenzir/vast/issues/1779.

lava (approved):
Looks good to me; I didn't test it locally but this is mainly to fix CI anyways so a green CI should be enough for this one.

tobim:
... and now I see I forgot to add a changlog entry.

PR #1816, opened 2021-07-30
Rename concepts namespace

JoeLoser:

Problem:
- `vast::detail` namespace is used for concepts, but this is a bit
  awkward in some cases and isn't clear whether something is a concept
  entity or not.

Solution:
- Rename concepts namespace from `vast::detail` to `vast::concepts`.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/test/concepts.cpp
+++ b/libvast/test/concepts.cpp
@@ -8,7 +8,7 @@
 
 #define SUITE concepts
 
-#include "vast/detail/concepts.hpp"
+#include "vast/concepts.hpp"
 
 #include "vast/test/test.hpp"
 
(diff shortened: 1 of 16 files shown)


tobim (changes requested):
Nit: Can you also move the files `concepts.hpp` and `concepts.cpp` out of the `detail` subdirectory?

JoeLoser:
> Nit: Can you also move the files `concepts.hpp` and `concepts.cpp` out of the `detail` subdirectory?

Just moved it to the top-level `vast` and out of `detail`. Does that work for you?

tobim:
> > Nit: Can you also move the files `concepts.hpp` and `concepts.cpp` out of the `detail` subdirectory?
> 
> Just moved it to the top-level `vast` and out of `detail`. Does that work for you?

That's exactly what I wanted.

tobim (approved):
Thanks for the improvement!

PR #1818, opened 2021-07-31
Printable concepts to simplify SFINAE 

JoeLoser:

Problem:
- There are several variable templates with cumbersome metaprogramming
  in defining them.
- There is some verbose SFINAE for constraining various function
  templates around the ideas of being printable.

Solution:
- Add various concepts like `printable` and apply them to simplify the
  metaprogramming and SFINAE.

### :memo: Checklist

File-by-file. The first commit is just a rename of `printer` to  `printer_base`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -297,7 +297,8 @@ target_compile_options(
             # Never treat deprecation warnings as errors.
             -Wno-error=deprecated
             -Wno-error=deprecated-declarations>
-            $<$<CXX_COMPILER_ID:GNU>:-Wno-redundant-move>)
+            $<$<CXX_COMPILER_ID:GNU>:-Wno-redundant-move>
+            $<$<CXX_COMPILER_ID:GNU>:-Wno-error=bool-compare>)
 
 # -- build id ------------------------------------------------------------------
 
--- a/libvast/vast/concept/printable/detail/as_printer.hpp
+++ b/libvast/vast/concept/printable/detail/as_printer.hpp
@@ -36,8 +36,8 @@ constexpr auto as_printer(T x)
   return literal_printer{x};
 }
 
-template <class T>
-constexpr auto as_printer(T x) -> std::enable_if_t<is_printer_v<T>, T> {
+template <printer T>
+constexpr auto as_printer(T x) {
   return x; // A good compiler will elide the copy.
 }
 
@@ -53,9 +53,9 @@ constexpr bool is_convertible_to_unary_printer_v =
 template <class T, class U>
 using is_convertible_to_binary_printer =
   std::bool_constant<
-    (is_printer_v<T> && is_printer_v<U>)
-    || (is_printer_v<T> && is_convertible_to_unary_printer_v<U>)
-    || (is_convertible_to_unary_printer_v<T> && is_printer_v<U>)
+    ( printer<T> && printer<U>)
+    || ( printer<T> && is_convertible_to_unary_printer_v<U>)
+    || (is_convertible_to_unary_printer_v<T> && printer<U>)
   >;
 
 template <class T, class U>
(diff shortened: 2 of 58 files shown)


JoeLoser:
ping @mavam @tobim -- pipeline is passing

libvast/vast/concept/printable/detail/as_printer.hpp
```diff
@@ -69,33 +69,29 @@ template <
...
-  // to merely being convertible to a printer). This prevent statements like `1 << 4`
-  // being parseable as a sequence printer of two literal printers.
-  std::disjunction_v<
-    is_printer<std::decay_t<T>>,
-    is_printer<std::decay_t<U>>>,
+// Require that at least one of the arguments already is a printer (as opposed
+// to merely being convertible to a printer). This prevent statements like `1 << 4`
+// being parseable as a sequence printer of two literal printers.
+template <template <class, class> class Binaryprinter, class T, class U> requires(printer<T> ||  printer<U>)
+constexpr auto as_printer(T&& x, U&& y) ->
```
tobim, line 89: No lines shall exceed the 80 character limit.

CMakeLists.txt
```diff
@@ -297,7 +297,8 @@ target_compile_options(
             # Never treat deprecation warnings as errors.
             -Wno-error=deprecated
             -Wno-error=deprecated-declarations>
-            $<$<CXX_COMPILER_ID:GNU>:-Wno-redundant-move>)
+            $<$<CXX_COMPILER_ID:GNU>:-Wno-redundant-move>
+            $<$<CXX_COMPILER_ID:GNU>:-Wno-error=bool-compare>)
```
tobim, line 301: I believe this would be better solved by not allowing bool as tParam for the `integral_printer`. The dedicated `bool_printer` should be used anyway.

tobim (changes requested):
Mostly formatting issues.

tobim:
Strange compilation error. It works with GCC11, but both GCC10 and clang fail. The curious part is that the fallback formatter that uses the stringification inspector hits the one for printable further down, which should not be the case. I wonder how your changes could have triggered this?

JoeLoser:
> Strange compilation error. It works with GCC11, but both GCC10 and clang fail. The curious part is that the fallback formatter that uses the stringification inspector hits the one for printable further down, which should not be the case. I wonder how your changes could have triggered this?

I'm not sure yet...I don't fully understand the issue. With https://github.com/tenzir/vast/pull/1818/commits/88aaa58814aec7ed554fe46e7e8671c20af32c1d, it was passing fine on Clang and we were just ignoring the warning on GCC. I think it's a GCC bug in eagerly evaluating the body of a concept of the `is_printable` concept or something.

CMakeLists.txt
JoeLoser: That's not quite the problem. The problem is that `integral_printer<T>::print` method is called with a `bool` for its argument. The class template type for the instantiation of `integral_printer` is non-bool.

tobim:
> With that being said, would you be opposed to disabling the warning for now on GCC? All tests passed still...I plan on trying to simplify the logger_formatters.hpp to simplify the metaprogramming which may fix this bug.

i would like to verify that the formatting is still correct for the that path. Can you check that the output of this added line is unchanged:
```diff
diff --git a/libvast/src/system/accountant.cpp b/libvast/src/system/accountant.cpp
index 53d138086..da1538b84 100644
--- a/libvast/src/system/accountant.cpp
+++ b/libvast/src/system/accountant.cpp
@@ -292,6 +292,7 @@ accountant(accountant_actor::stateful_pointer<accountant_state> self,
   self->delayed_send(self, overview_delay, atom::telemetry_v);
   return {
     [self](atom::announce, const std::string& name) {
+      VAST_INFO("{} got announce from {}", self, self->current_sender()->address());
       auto& st = *self->state;
       st.actor_map[self->current_sender()->id()] = name;
       self->monitor(self->current_sender());
```

JoeLoser:
> ```diff
> VAST_INFO("{} got announce from {}", self, self->current_sender()->address());
> ```

Sure. Can you elaborate a bit more on how to verify that the output of your added line is unchanged? Is there a test I can run, or something else?

tobim:
> > ```diff
> > VAST_INFO("{} got announce from {}", self, self->current_sender()->address());
> > ```
> 
> Sure. Can you elaborate a bit more on how to verify that the output of your added line is unchanged? Is there a test I can run, or something else?

Running `vast --enable-metrics start` should do the trick.

JoeLoser:
> > > ```diff
> > > VAST_INFO("{} got announce from {}", self, self->current_sender()->address());
> > > ```
> > 
> > 
> > Sure. Can you elaborate a bit more on how to verify that the output of your added line is unchanged? Is there a test I can run, or something else?
> 
> Running `vast --enable-metrics start` should do the trick.

Here is the output which looks the same.  On top of tree with your added log line, using llvm 12.0.1, the output is:

```c++
./vast --enable-metrics start
[09:27:31.669] loaded configuration file: "/Users/joe/dev_local/vast/build/etc/vast/vast.yaml"
[09:27:32.176] accountant got announce from 10@7817E89FEDA8EF0185063BBD8B5EF70D97BBE700#6723
[09:27:32.177] accountant got announce from 11@7817E89FEDA8EF0185063BBD8B5EF70D97BBE700#6723
[09:27:32.178] accountant got announce from 14@7817E89FEDA8EF0185063BBD8B5EF70D97BBE700#6723
[09:27:32.178] importer adds accountant source
[09:27:32.181] VAST node is listening on localhost:42000
```

On my branch when using the dev Docker file with gcc10, the output is similar:
```
[13:33:18.291] loaded configuration file: "/opt/tenzir/vast/etc/vast/vast.yaml"
[13:33:18.328] accountant got announce from 10@C0388D7211856698956FF8542CE446EBABE7A100#1
[13:33:18.329] accountant got announce from 11@C0388D7211856698956FF8542CE446EBABE7A100#1
[13:33:18.330] accountant got announce from 24@C0388D7211856698956FF8542CE446EBABE7A100#1
[13:33:18.331] importer adds accountant source
[13:33:18.336] VAST node is listening on localhost:42000
```

Given this, I think we're fine just disabling the warning for GCC for now and moving on. What do you think? Does this look right to you?

tobim (approved):
This is good to go since it's confirmed that there is no regression in the log output.

tobim:
Maybe there is a fix for the fallback formatters so that the warning can be enabled again?

I think in the long run we should turn all the alternatives in `vast::data` into strong typedefs and implement fmt formatters for them to avoid these issues.

PR #1822, opened 2021-08-04
Fix int-to-bool conversion warning in integral_parser

JoeLoser:

Problem:
- The result of `std::isxdigit` is an int which triggers an int to
  bool (the return value of this function) warning.
- Calling `std::isxdigit` with a `char` is undefined. Like other
  functions from `<cctype>`, the behavior of `std::isxdigit` is undefined
  if the argument's value is neither representable as an `unsigned char`
  nor equal to `EOF`.

Solution:
- Explicitly compare the return value of `std::isxdigit` with `0` to
  land in a boolean space.
- Convert the argument type to `unsigned char` before calling
  `std::isxdigit`.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/concept/parseable/numeric/integral.hpp
+++ b/libvast/vast/concept/parseable/numeric/integral.hpp
@@ -11,6 +11,7 @@
 #include "vast/concept/parseable/core.hpp"
 #include "vast/detail/coding.hpp"
 
+#include <cctype>
 #include <cstdint>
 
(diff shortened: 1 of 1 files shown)


tobim (approved):
The description at https://en.cppreference.com/w/cpp/string/byte/isxdigit confirms that this change is correct. Thanks!

I wonder why we didn't see the conversion warning in CI.

PR #1824, opened 2021-08-06
Replace enable_if_t with requires in choice printer

JoeLoser:

Problem:
- `choice` printer uses `std::enable_if_t` which is verbose.

Solution:
- Use concepts instead.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/concept/printable/core/choice.hpp
+++ b/libvast/vast/concept/printable/core/choice.hpp
@@ -16,20 +16,20 @@
 namespace vast {
 
 template <class Lhs, class Rhs>
-class choice_printer;
+class choice_printer_t;
 
 template <class>
 struct is_choice_printer : std::false_type {};
 
 template <class Lhs, class Rhs>
-struct is_choice_printer<choice_printer<Lhs, Rhs>> : std::true_type {};
+struct is_choice_printer<choice_printer_t<Lhs, Rhs>> : std::true_type {};
 
 template <class T>
-constexpr bool is_choice_printer_v = is_choice_printer<T>::value;
+concept choice_printer = is_choice_printer<T>::value;
 
 /// Attempts to print either LHS or RHS.
 template <class Lhs, class Rhs>
-class choice_printer : public printer_base<choice_printer<Lhs, Rhs>> {
+class choice_printer_t : public printer_base<choice_printer_t<Lhs, Rhs>> {
 public:
   using lhs_attribute = typename Lhs::attribute;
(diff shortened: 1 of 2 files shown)


mavam:
Paging @tobim regarding the clang-format hack; I haven't followed the formatting guidelines regarding concepts.

@tenzir/backend can we add some rules to concept formatting here 👉https://github.com/tenzir/.github/blob/master/contributing.md ?

dominiklohmann:
I've been out of the loop for the past two weeks, but why are we adding manual clang-format toggles everywhere? Wouldn't it make much more sense to live with what we have now and bulk-reformat once we have clang-format 13?

Regarding contributing.md, all it should say is "use clang-format". Everything else is just wasting the contributor's precious time.

JoeLoser:
> Paging @tobim regarding the clang-format hack; I haven't followed the formatting guidelines regarding concepts.
> 
> @tenzir/backend can we add some rules to concept formatting here 👉https://github.com/tenzir/.github/blob/master/contributing.md ?

I just removed the `clang-format` toggle; it actually wasn't needed here -- sorry. @dominiklohmann

dominiklohmann:
> I just removed the `clang-format` toggle; it actually wasn't needed here -- sorry. @dominiklohmann

No worries. I think we're all a bit confused what the correct thing to do here is.

dominiklohmann (approved):
Thanks as usual! These refactoring are sooooo much appreciated. 🧹

JoeLoser:
> > I just removed the `clang-format` toggle; it actually wasn't needed here -- sorry. @dominiklohmann
> 
> No worries. I think we're all a bit confused what the correct thing to do here is.

Sometimes, `clang-format` chokes hard and formats the function template *terribly* in some cases with `requires` expressions. This, however, was not one of those cases.

dominiklohmann:
We'll update to clang-format 13 as soon as it's out. Hopefully that resolves most of our issues.

We could also do `brew install --HEAD clang-format` in CI, but I'm not sure whether that's a good idea.

JoeLoser:
> We'll update to clang-format 13 as soon as it's out. Hopefully that resolves most of our issues.
> 
> We could also do `brew install --HEAD clang-format` in CI, but I'm not sure whether that's a good idea.

Yeah, I'm hopeful llvm13 will be better. It should release soonish I imagine as they branched llvm14 just recently for dev. I wouldn't recommend installing `HEAD` on a format style checking tool in general on CI as that can be a frustrating dev experience potentially.

PR #1826, opened 2021-08-09
Replace std::enable_if_t with requires expressions

JoeLoser:

Problem:
- There is still use of `std::enable_if_t` which is verbose, gives poor
  error messages, and is slower to compile than concepts.

Solution:
- Convert most uses of `std::enable_if_t` to use concepts or in-place
  `requires` expressions.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/concepts.hpp
+++ b/libvast/vast/concepts.hpp
@@ -21,6 +21,12 @@ concept SameHelper = std::is_same_v<T, U>;
 template <class T, class U>
 concept same_as = SameHelper<T, U> && SameHelper<U, T>;
 
+template <class T, class U>
+concept sameish = same_as<std::decay_t<T>, std::decay_t<U>>;
+
+template <class T, class U>
+concept different = !same_as<T, U>;
+
 template <typename From, typename To>
 concept convertible_to = std::is_convertible_v<From, To> && requires(
   std::add_rvalue_reference_t<From> (&f)()) {
--- a/libvast/vast/detail/bit_cast.hpp
+++ b/libvast/vast/detail/bit_cast.hpp
@@ -21,12 +21,11 @@ namespace vast::detail {
 /// are unspecified.
 /// TODO: Remove this when we have C++20, which ships with a compiler magic
 ///       version of this with constexpr support.
-template <
-  typename To, typename From,
-  typename = std::enable_if_t<
-    (sizeof(To) == sizeof(From))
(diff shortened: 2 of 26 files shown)


libvast/vast/detail/bit_cast.hpp
```diff
@@ -21,12 +21,11 @@ namespace vast::detail {
...
-template <
-  typename To, typename From,
-  typename = std::enable_if_t<
-    (sizeof(To) == sizeof(From))
-    && std::conjunction_v<std::is_trivially_copyable<From>, std::is_trivial<To>>>>
-To bit_cast(const From& src) noexcept {
+template <typename To, typename From>
+  requires(sizeof(To) == sizeof(From)
+           && std::is_trivially_copyable_v<From> && std::is_trivial_v<To>)
+To bit_cast(const From& src)
```
dominiklohmann, line 27: Is this already available as a builtin in `<bits>`?

mavam: Unfortunately `<bits>` is not available yet with our compilers.

libvast/vast/concepts.hpp
```diff
@@ -21,6 +21,12 @@ concept SameHelper = std::is_same_v<T, U>;
 template <class T, class U>
 concept same_as = SameHelper<T, U> && SameHelper<U, T>;
 
+template <class T, class U>
+concept sameish = same_as<std::decay_t<T>, std::decay_t<U>>;
```
dominiklohmann, line 25: This needs to be defined differently: https://stackoverflow.com/questions/58509147/why-does-same-as-concept-check-type-equality-twice

libvast/vast/detail/bit_cast.hpp
JoeLoser: It's not available yet. `__builtin_bit_cast` isn't available until gcc11 either.

libvast/vast/concepts.hpp
JoeLoser: Mind elaborating? We're routing through the `same_as` concept defined right above which goes through `SameHelper` so we check that `T == U` and `U == T`.

dominiklohmann: You are absolutely correct—must've confused SameHelper and same_as yesterday when glancing at the code.

dominiklohmann (approved):

PR #1834, opened 2021-08-10
Fix chunk::make for buffers with SBO

dominiklohmann:

See comment in diff.

### :memo: Checklist

Follow the logic, trust our unit test suite. Or try to come up with something better.

### DIFF
--- a/libvast/vast/chunk.hpp
+++ b/libvast/vast/chunk.hpp
@@ -78,9 +78,12 @@ public:
   /// accidental copies when calling this function.
   /// @returns A chunk pointer or `nullptr` on failure.
   template <class Buffer>
-    requires(!(std::is_lvalue_reference_v<
-                 Buffer> || std::is_trivially_move_assignable_v<Buffer>))
-  static auto make(Buffer&& buffer) -> decltype(as_bytes(buffer), chunk_ptr{}) {
+    requires(!std::is_lvalue_reference_v<Buffer> &&          //
+             !std::is_trivially_move_assignable_v<Buffer> && //
+             requires(const Buffer& buffer) {
+               { as_bytes(buffer) } -> concepts::same_as<view_type>;
+             })
+  static auto make(Buffer&& buffer) -> chunk_ptr {
     // Move the buffer into a unique pointer; otherwise, we might run into
     // issues with small buffer optimizations where the view no longer points
     // to the correct data once we moved the buffer into the deleter.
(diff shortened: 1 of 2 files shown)


libvast/vast/chunk.hpp
```diff
@@ -81,8 +81,12 @@ class chunk final : public caf::ref_counted {
     requires(!(std::is_lvalue_reference_v<
                  Buffer> || std::is_trivially_move_assignable_v<Buffer>))
   static auto make(Buffer&& buffer) -> decltype(as_bytes(buffer), chunk_ptr{}) {
```
lava, line 83: Btw, is there a special reason why we kept the `as_bytes(buffer)` as SFINAE and didn't move it into the `requires` clause?

dominiklohmann: I've pushed the change.

libvast/vast/chunk.hpp
```diff
@@ -78,11 +78,18 @@ class chunk final : public caf::ref_counted {
   /// accidental copies when calling this function.
   /// @returns A chunk pointer or `nullptr` on failure.
   template <class Buffer>
-    requires(!(std::is_lvalue_reference_v<
-                 Buffer> || std::is_trivially_move_assignable_v<Buffer>))
-  static auto make(Buffer&& buffer) -> decltype(as_bytes(buffer), chunk_ptr{}) {
-    const auto view = as_bytes(buffer);
-    return make(view, [buffer = std::exchange(buffer, {})]() noexcept {
+    requires(!std::is_lvalue_reference_v<Buffer> &&          //
+             !std::is_trivially_move_assignable_v<Buffer> && //
```
lava, line 82: I think we can drop the `!is_trivially_move_assignable` requirement now, right?

libvast/vast/chunk.hpp
```diff
@@ -78,11 +78,18 @@ class chunk final : public caf::ref_counted {
...
+             requires(const Buffer& buffer) {
+               { as_bytes(buffer) } -> concepts::same_as<view_type>;
+             })
+  static auto make(Buffer&& buffer) -> chunk_ptr {
+    // Move the buffer into a unique pointer; otherwise, we might run into
+    // issues with small buffer optimizations where the view no longer points
+    // to the correct data once we moved the buffer into the deleter.
```
lava, line 89: The comment seems to be too narrow, even without small buffer optimization there is no requirement that a move constructor must preserve all internal memory addresses.

lava (approved):
Looks good to me, just some comments on tangential issues.

libvast/vast/chunk.hpp
dominiklohmann: I reworded the comment in a separate commit.

libvast/vast/chunk.hpp
dominiklohmann: Good catch. I dropped the requirement in a separate commit.

PR #1836, opened 2021-08-11
Run tests when building with Nix

tobim:

- Add the option to disable some tests in the integration testing framework
- Workaround an issue in `mkDerivation` that prevents running tests for cross builds
- Explicitly disable native plugin and disk monitor tests for now

### :memo: Checklist

Review code Commit by commit. Check the Nix build CI logs.

### DIFF
--- a/nix/vast/default.nix
+++ b/nix/vast/default.nix
@@ -19,41 +19,15 @@
 , jemalloc
 , libunwind
 , zeek-broker
-, python3
-, jq
-, tcpdump
-, utillinux
 , versionOverride ? null
 , withPlugins ? []
 , extraCmakeFlags ? []
-, disableTests ? true
+, disableTests ? false
 , buildType ? "Release"
 }:
 let
   inherit (stdenv.hostPlatform) isStatic;
   isCross = stdenv.buildPlatform != stdenv.hostPlatform;
-
-  py3 = (let
-    python = let
-      packageOverrides = final: prev: {
-        # See https://github.com/NixOS/nixpkgs/pull/96037
-        coloredlogs = prev.coloredlogs.overridePythonAttrs (old: rec {
-          doCheck = !stdenv.isDarwin;
-          checkInputs = with prev; [ pytest mock utillinux verboselogs capturer ];
-          pythonImportsCheck = [ "coloredlogs" ];
-
-          propagatedBuildInputs = [ prev.humanfriendly ];
-        });
-      };
-    in python3.override {inherit packageOverrides; self = python;};
-
-  in python.withPackages(ps: with ps; [
-    coloredlogs
-    jsondiff
-    pyarrow
-    pyyaml
-    schema
-  ]));
 
   src = vast-source;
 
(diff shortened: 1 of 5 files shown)


nix/vast/default.nix
```diff
@@ -108,15 +82,29 @@ stdenv.mkDerivation rec {
...
+  # installCheckInputs = [ py3 jq tcpdump ];
+  # TODO: Investigate why the disk monitor test fails in the build sandbox.
+  # TODO: Native plugins are currently broken with static builds.
   installCheckPhase = ''
-    python ../vast/integration/integration.py --app ${placeholder "out"}/bin/vast
+    python ../vast/integration/integration.py \
+      --app ${placeholder "out"}/bin/vast \
+      --disable "Disk Monitor"
+      --disable "Partition-local Stores"
+      --disable "Transforms"
```
dominiklohmann, line 107: This won't run integration tests for lsvast and for plugins, which have separate integration test suites. Do we have a story for that already, or, is it possible to simply run the CMake build target `integration` here? We could add an option to CMake that skips the virtual env setup.

nix/vast/default.nix
```diff
@@ -108,15 +82,29 @@ stdenv.mkDerivation rec {
 
   hardeningDisable = lib.optional isStatic "pic";
 
-  doCheck = false;
+  doCheck = true;
+  # Concurrent testing is not possible because multiple suites reuse the same
+  # file paths.
```
dominiklohmann, line 87: Which suites do this?

nix/vast/default.nix
```diff
@@ -108,15 +82,29 @@ stdenv.mkDerivation rec {
...
+  enableParallelChecking = false;
+  preCheck = ''
+    export LD_LIBRARY_PATH=$PWD/libvast
+    export DYLD_LIBRARY_PATH=$PWD/libvast
+  '';
```
dominiklohmann, line 92: Why do we need this?

vast/integration/integration.py
```diff
@@ -626,6 +638,11 @@ def main():
         type=Path,
         help="The basedir for the test runs",
     )
+    parser.add_argument(
+        "--dry-run",
```
dominiklohmann, line 642: Should this be named `--list-suites` instead?

dominiklohmann (commented):
Just some questions so far on the code; waiting for CI to check the build logs.

Do we want to upload the logs for failed integration tests in CI just like we do for the regular builds?

nix/vast/default.nix
tobim: For passing this through CMake we would have to make it possible to pass arbitrary options that get passed through to the integration testing script. This would be brittle because the `integration` target invokes it multiple times. I think its better to avoid that complexity and run the tests directly here. I will add support for integration testing of plugins in a follow up. lsvast can be checked with a second call directly here.

nix/vast/default.nix
tobim: I don't know exactly, I just noticed that I get lots of breakage when the tests are run in parallel. We normally never do that.

nix/vast/default.nix
tobim: This is for non-static builds. Nix invokes CMake with the option that rpaths get set to the final locations of the install prefix - `BUILD_WITH_INSTALL_RPATH` or something like that - so this is a workaround for `vast-test`.

vast/integration/integration.py
tobim: I think `--list-suites` doesn't convey that the `--test` `--disable` and `--tag` options would be considered. `--dry-run` is commonly used for showing what would be done without actually doing it, so I think it fits quite well.

nix/vast/default.nix
dominiklohmann: Can you add that as a comment here for the future maintainer?

PR #1837, opened 2021-08-12
Add convenience overload for data

mavam:

This PR is a small addendum to the recent `convert` enhancements.

Take a quick look.


tobim (approved):
As long as we constrain this overload to types that have a layout we should be safe from implicit conversions to `data`.
Thanks for adding a test.

PR #1838, opened 2021-08-12
Use CPM.cmake for dependency version pinning

dominiklohmann:

We experienced some pretty bad behavior with {fmt} 8.x, spdlog 1.9.1 and clang 12.0.1, and with Homebrew only providing latest the development experience on macOS suffered a lot. This patch changes {fmt} and spdlog to be fetched via CPM, unless CPM_USE_LOCAL_PACKAGES is ON or VAST_DEVELOPER_MODE is OFF (which implies the former).

Additionally, this bumps the minimum requirement for {fmt} to 7.1.3 and gets rid of some workarounds for compatibility with older versions.

### :memo: Checklist

Stress test this locally under every possible scenario you can imagine.


libvast/CMakeLists.txt
```diff
@@ -385,23 +385,30 @@ if (VAST_ENABLE_JOURNALD_LOGGING)
...
-    # Forbid unsafe duration_cast usage.
-    FMT_SAFE_DURATION_CAST
-    # Make fmt::internal an alias for fmt::detail. Remove when requiring fmt >= 7.
-    FMT_USE_INTERNAL)
-find_package(fmt 5.2.1 REQUIRED)
-string(APPEND VAST_FIND_DEPENDENCY_LIST "\nfind_package(fmt 5.2.1 REQUIRED)")
-target_link_libraries(libvast PUBLIC fmt::fmt)
-dependency_summary("fmt" fmt::fmt "Dependencies")
+target_compile_definitions(libvast PUBLIC # Forbid unsafe duration_cast usage.
+                                          FMT_SAFE_DURATION_CAST)
```
JoeLoser, line 389: Do we still need this? Looks like `fmt` has `FMT_SAFE_DURATION_CAST` enabled if not explicitly disabled: https://github.com/fmtlib/fmt/blob/master/include/fmt/chrono.h#L22

dominiklohmann: We probably don't, good catch.

dominiklohmann:
This is kind of superseded by #1846. Let's revisit this in the future only if we absolutely must do so.

PR #1839, opened 2021-08-12
Simplify call operators in core printer

JoeLoser:

Problem:
- There are several call operators and some discern between the arity of
  attributes which complicates the overload sets and is verbose.

Solution:
- Reduce number of call operator overloads from `4` to `2` by always
  taking a parameter pack of attributes and just branching on the number
  of attributes in the pack.
- Apply concepts to simplify the SFINAE.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/concept/printable/core/printer.hpp
+++ b/libvast/vast/concept/printable/core/printer.hpp
@@ -40,30 +40,34 @@ struct printer_base {
   }
 
   // FIXME: don't ignore ADL.
-  template <class Range, class Attribute = unused_type>
-  auto operator()(Range&& r, const Attribute& a = unused) const
-  -> decltype(std::begin(r), std::end(r), bool()) {
-    auto out = std::back_inserter(r);
-    return derived().print(out, a);
+  template <class Range, class... TAttributes>
+    requires requires(Range r) {
+      std::begin(r);
+      std::end(r);
+    }
+  auto operator()(Range&& r, const TAttributes&... attributes) const {
+    if constexpr (sizeof...(TAttributes) == 0) {
+      auto out = std::back_inserter(r);
+      return derived().print(out, unused);
+    } else if constexpr (sizeof...(TAttributes) == 1) {
+      auto out = std::back_inserter(r);
+      return derived().print(out, attributes...);
+    } else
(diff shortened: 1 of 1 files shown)


libvast/vast/concept/printable/core/printer.hpp
```diff
@@ -40,30 +40,30 @@ struct printer_base {
...
-  auto operator()(Range&& r, const Attribute& a = unused) const
-  -> decltype(std::begin(r), std::end(r), bool()) {
-    auto out = std::back_inserter(r);
-    return derived().print(out, a);
+  template <class Range, class... TAttributes>
+    requires requires(Range r) {
+      std::begin(r);
+      std::end(r);
+    }
+  auto operator()(Range&& r, const TAttributes&... attributes) const -> bool {
```
tobim, line 48: Nit: this can now be called without an attribute, I guess it would fail at `std::tie` then.
Nit: There is no reason to use a trailing return type any more.

tobim: The same applies to the overload for iterators.

tobim (approved):
At first glance `range` and `iterator` should be useful as dedicated concepts. Opinions?

libvast/vast/concept/printable/core/printer.hpp
JoeLoser: Just pushed a change that adds a constraint the `sizeof...(TAttributes) > 0` to give a better error message if someone were to call this call operator without any attributes.

JoeLoser:
> At first glance `range` and `iterator` should be useful as dedicated concepts. Opinions?

`range` is definitely a useful concept. I plan on making `range` a dedicated concept when I make a similar change like this to simplify the call operators in the `core` parser.

I'll need to see more use cases of our iterator types/requirements beyond core parser to decide what's appropriate to promote to a dedicated concept. 

Does that work for you?

mavam:
> At first glance `range` and `iterator` should be useful as dedicated concepts. Opinions?

If we add these as concepts, I'd be careful to make them drop-in replacements for concepts defined in the standard, e.g., as in https://en.cppreference.com/w/cpp/ranges.

PR #1841, opened 2021-08-12
Simplify pretty_type_name in vast/detail/logger.hpp

JoeLoser:

Problem:
- `pretty_type_name` does different behavior whether `T` is a pointer
  type or not which is unnecessary by only removing the pointer type if
  `T` is a pointer type.

Solution:
- Always use `std::remove_pointer_t<T>` which will leave the `T` as is
  for non-pointer types, thus preserving the same behavior.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/detail/logger.hpp
+++ b/libvast/vast/detail/logger.hpp
@@ -40,10 +40,7 @@ std::shared_ptr<spdlog::logger>& logger();
 
 template <class T>
 auto pretty_type_name(const T&) {
-  if constexpr (std::is_pointer_v<T>)
-    return caf::detail::pretty_type_name(typeid(std::remove_pointer_t<T>));
-  else
-    return caf::detail::pretty_type_name(typeid(T));
+  return caf::detail::pretty_type_name(typeid(std::remove_pointer_t<T>));
 }
 
 template <class T>


JoeLoser:
For what it's worth, I'd recommend `vast` writes its own `pretty_type_name` function in terms of the output of `__PRETTY_FUNCTION__`. The implementation in `caf` is not great and uses `std::type_info` which makes the call sites all use `typeid` which isn't needed in a high quality implementation. I'd prefer an interface like

```
template<class T>
std::string_view pretty_type_name() {
 / / uses __PRETTY_FUNCTION__ to parse out the type name.
}
```

dominiklohmann:
Anything against using this?


```cpp
#include <string_view>

namespace type_name_probe {

/// A templated function whose whole purpose is to embed the name of an
/// explicitly specified template parameter into its signature, which the
/// function returns via platform-specific macros.
template <class T> constexpr std::string_view embed() {
#if defined(_MSC_VER)
  return __FUNCSIG__;
#else
  return __PRETTY_FUNCTION__;
#endif
}

/// A probe with a known type; used for extracting the platform-specific prefix
/// and suffix from the result of embed<T>.
constexpr auto probe = embed<double>();
constexpr auto prefix_length = probe.find("double");
constexpr auto suffix_length =
    probe.size() - prefix_length - std::string_view("double").size();

// Sanity check that the probe did contain the type name.
static_assert(prefix_length != std::string_view::npos,
              "cannot extract typename from function signature");

} // namespace type_name_probe

/// Retrieves the type name of **T** at compile time.
template <class T> constexpr std::string_view type_name() {
  auto name = type_name_probe::embed<T>();
  name.remove_prefix(type_name_probe::prefix_length);
  name.remove_suffix(type_name_probe::suffix_length);
  return name;
}
```

JoeLoser:
> Anything against using this?
> 
> ```c++
> #include <string_view>
> 
> namespace type_name_probe {
> 
> /// A templated function whose whole purpose is to embed the name of an
> /// explicitly specified template parameter into its signature, which the
> /// function returns via platform-specific macros.
> template <class T> constexpr std::string_view embed() {
> #if defined(_MSC_VER)
>   return __FUNCSIG__;
> #else
>   return __PRETTY_FUNCTION__;
> #endif
> }
> 
> /// A probe with a known type; used for extracting the platform-specific prefix
> /// and suffix from the result of embed<T>.
> constexpr auto probe = embed<double>();
> constexpr auto prefix_length = probe.find("double");
> constexpr auto suffix_length =
>     probe.size() - prefix_length - std::string_view("double").size();
> 
> // Sanity check that the probe did contain the type name.
> static_assert(prefix_length != std::string_view::npos,
>               "cannot extract typename from function signature");
> 
> } // namespace type_name_probe
> 
> /// Retrieves the type name of **T** at compile time.
> template <class T> constexpr std::string_view type_name() {
>   auto name = type_name_probe::embed<T>();
>   name.remove_prefix(type_name_probe::prefix_length);
>   name.remove_suffix(type_name_probe::suffix_length);
>   return name;
> }
> ```

Seems reasonable, albeit not the one I carry around in my toolbox. If we want to switch to a better `type_name`, I'd recommend we do that as a follow-up PR and fix all the call sites. I'd want some solid tests on our implementation to see if it works with types in inline namespaces, lambdas, type aliases, etc.

tobim:
I normally use the implementation from https://stackoverflow.com/a/56766138.

dominiklohmann:
> I normally use the implementation from https://stackoverflow.com/a/56766138.

The only difference being that that has the prefixes and suffixes hardcoded, while the solution I posted calculates them once at compile time.

>  I'd recommend we do that as a follow-up PR and fix all the call sites.

Fine by me, although I don't think there's much to test—it really is only being used in log outputs, mostly error messages when a conversion fails.

dominiklohmann (approved):

PR #1844, opened 2021-08-13
Simplify call operators in core parser

JoeLoser:

Problem:
- Core parser has a few overloaded call operators, some with default
  arguments.

Solution:
- Reduce the number of overloads by always accepting a parameter pack
  for the attributes and doing different behavior based on the arity.
- Add `range` concept and apply it to core parser and core printer.

Note:
- We could remove the C-array overload and collapse it into the range
  overload if desired, but it's not worth it given how small the
  implementation of the array overload is.

### :memo: Checklist

File-by-file.

### DIFF
--- a/libvast/vast/concept/parseable/core/parser.hpp
+++ b/libvast/vast/concept/parseable/core/parser.hpp
@@ -9,7 +9,7 @@
 #pragma once
 
 #include "vast/concept/support/unused_type.hpp"
-#include "vast/detail/type_traits.hpp"
+#include "vast/concepts.hpp"
 
 #include <iterator>
 #include <tuple>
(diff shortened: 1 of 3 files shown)


libvast/vast/concept/parseable/core/parser.hpp
```diff
@@ -57,35 +57,36 @@ struct parser_base {
...
-    auto t = std::tie(a0, a1, as...);
-    return operator()(r, t);
-  }
-
-  template <class Iterator, class Attribute>
-  auto operator()(Iterator& f, const Iterator& l, Attribute& a) const
-  -> decltype(*f, ++f, f == l, bool()) {
-    return derived().parse(f, l, a);
+  template <class... TAttributes>
+  auto operator()(concepts::range auto&& r, TAttributes&... attributes) const {
```
lava, line 61: Can we keep the `-> bool` here? I feel like every bit helps with these highly abstract functions.

libvast/vast/concept/parseable/core/parser.hpp
```diff
@@ -57,35 +57,36 @@ struct parser_base {
...
+      auto f = std::begin(r);
+      auto l = std::end(r);
+      return derived().parse(f, l, unused) && f == l;
+    } else if constexpr (sizeof...(TAttributes) == 1) {
+      auto f = std::begin(r);
+      auto l = std::end(r);
+      return derived().parse(f, l, attributes...) && f == l;
+    } else {
+      auto t = std::tie(attributes...);
+      return operator()(r, t);
```
lava, line 72: Would it make sense here to go for symmetry and to write the last case like this?

```
auto f = std::begin(r);
auto l = std::end(r);
auto t = std::tie(attributes...);
return derived().parse(f, l, t);
```

Or can it happen that `operator()(r, t)` calls into a different function here?

lava (approved):
Looks good to me; the transformation seems valid and the new version is certainly easier to read.

Just some stylistic nits below.

libvast/vast/concept/parseable/core/parser.hpp
JoeLoser: Sure, just added it back.

libvast/vast/concept/parseable/core/parser.hpp
JoeLoser: That should be safe. It makes the declarations of `f` and `l` common between all the branches, so it's less code in the end too. Just changed 👍

PR #1848, opened 2021-08-17
Interpret metrics paths relative to the db root

dominiklohmann:

The paths `vast.metrics.{file,uds}-sink.path` were, unlike othe relative paths, interpreted as relative to the current working directory of the VAST server instead of relative to the database directory. This changes the ACCOUNTANT actor to be aware of the database directory.

### :memo: Checklist

File-by-file. Run locally.

### DIFF
--- a/libvast/src/system/accountant.cpp
+++ b/libvast/src/system/accountant.cpp
@@ -52,8 +52,9 @@ struct accountant_state_impl {
 
   // -- constructor, destructors, and assignment operators ---------------------
 
-  accountant_state_impl(accountant_actor::pointer self, accountant_config cfg)
-    : self{self} {
+  accountant_state_impl(accountant_actor::pointer self, accountant_config cfg,
+                        std::filesystem::path root)
+    : self{self}, root{std::move(root)} {
     apply_config(std::move(cfg));
   }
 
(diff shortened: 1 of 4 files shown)


libvast/src/system/accountant.cpp
```diff
@@ -236,8 +240,8 @@ struct accountant_state_impl {
                     cfg.uds_sink.path, s.error());
         }
       } else {
-        auto s
-          = detail::make_output_stream(cfg.uds_sink.path, cfg.uds_sink.type);
+        auto s = detail::make_output_stream(root / cfg.uds_sink.path,
```
tobim, line 243: The root should only be prepended when the `sink_path` is relative.

tobim (changes requested):

libvast/src/system/accountant.cpp
dominiklohmann: That's a no-op. `a / b == b` if `b` is absolute.

dominiklohmann: See https://godbolt.org/z/qzWbW7z63 for a minimal example.

tobim: Surprising, but ok.

tobim (approved):

libvast/src/system/accountant.cpp
dominiklohmann: How is that surprising? What do you expect to happen when you append an absolute path?

PR #1850, opened 2021-08-18
Support native plugins in the static binary

dominiklohmann:

The plugin scaffold has an auto-registration feature, but for static linking this requires the plugin library to be assembled with `-whole-archive` (or similar, depending on the linker and platform). libvast doesn't use that option, so auto-registration fails for the generated static binary because the symbols are omitted.

The solution here is to extract native plugins from the libvast target into a separate library that is linked appropriately into the VAST binary directly, forcing the linker not to omit the symbols required for the auto-registration.

### :memo: Checklist

Run the generated static binary from CI and check if the native plugins works. Or, alternatively, build this locally and patch out the code that removes natives plugins from the output of `vast version`.

This is what it shows for me on this branch:

```
❯ ./build/bin/vast -qq version
{
  "VAST": "2021.07.29-137-gbf1cebad0d-dirty",
  "VAST Build Tree Hash": "06d86f9260a80cc4050610dfd8d76209",
  "CAF": "0.17.6",
  "Apache Arrow": "5.0.0",
  "jemalloc": null,
  "plugins": {
    "delete": "native",
    "example-analyzer": "0.1-ce8bb03806",
    "example-transform": "0.1-3899d5d80e",
    "hash": "native",
    "pcap": "95e3699619",
    "replace": "native",
    "segment-store": "native"
  }
}
```

### DIFF
--- a/.cmake-format
+++ b/.cmake-format
@@ -44,6 +44,12 @@ with section("parse"):
                 "nargs": 1,
             },
         },
+        "vasttargetlinkwholearchive": {
+            "spelling": "VASTTargetLinkWholeArchive",
+            "pargs": {
+                "nargs": 3,
+            },
+        },
     }
 
     # Override configurations per-command where available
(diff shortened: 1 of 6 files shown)


tobim (approved):
Thanks for taking this off my plate! I checked the sized of the produced unstripped static binaries and they are almost the same.

PR #1854, opened 2021-08-19
Align suricata.dhcp with the latest eve.log schema

tobim:
### :memo: Checklist

### DIFF
--- a/schema/types/suricata.schema
+++ b/schema/types/suricata.schema
@@ -97,17 +97,41 @@ type suricata.dcerpc = suricata.component.common + record {
   }
 }
 
+// At the time of writing no canonical documentation exists for dhcp events.
+// The fields can be derived from the logging code in:
+// https://github.com/OISF/suricata/blob/master/rust/src/dhcp/logger.rs
 type suricata.dhcp = suricata.component.common + record {
   dhcp: record {
-    type: string,
+    type: string, // consider switching to enum { request, reply, <unknown> }
     id: count #index=hash,
     client_mac: string #index=hash,
     assigned_ip: addr,
     client_ip: addr,
-    dhcp_type: string,
+    dhcp_type: enum {
+      discover,
+      offer,
+      request,
+      decline,
+      ack,
+      nak,
+      release,
+      inform,
+      unknown
+    },
+    // In requests:
     client_id: string #index=hash,
     hostname: string #ioc,
-    params: list<string>
+    requested_ip: addr,
+    params: list<string>,
+    // In replies:
+    relay_ip: addr,
+    next_server_ip: addr,
+    lease_time: count,
+    rebinding_time: count,
+    renewal_time: count,
+    subnet_mask: addr,
+    routers: list<addr>,
+    dns_servers: list<addr>
   }
 }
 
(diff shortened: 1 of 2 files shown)


schema/types/suricata.schema
```diff
@@ -97,17 +97,39 @@ type suricata.dcerpc = suricata.component.common + record {
...
+      discover,
+      offer,
+      request,
+      decline,
+      ack,
+      nak,
+      release,
+      inform,
+      unknown
+    },
```
dominiklohmann, line 118: Is this type change backwards compatible?

schema/types/suricata.schema
```diff
@@ -97,17 +97,39 @@ type suricata.dcerpc = suricata.component.common + record {
...
+    dhcp_type: enum {
+      discover,
+      offer,
+      request,
+      decline,
+      ack,
+      nak,
+      release,
+      inform,
+      unknown
```
dominiklohmann, line 117: Nit: trailing comma

schema/types/suricata.schema
```diff
@@ -97,17 +97,39 @@ type suricata.dcerpc = suricata.component.common + record {
   }
 }
 
+// https://github.com/OISF/suricata/blob/master/rust/src/dhcp/logger.rs
 type suricata.dhcp = suricata.component.common + record {
   dhcp: record {
-    type: string,
+    type: string, // consider switching to enum { request, reply, <unknown> }
```
dominiklohmann, line 103: Please do so if we're changing types of existing fields anyways.

schema/types/suricata.schema
```diff
@@ -97,17 +97,39 @@ type suricata.dcerpc = suricata.component.common + record {
...
+      inform,
+      unknown
+    },
+    // In requests:
     client_id: string #index=hash,
     hostname: string #ioc,
-    params: list<string>
+    requested_ip: addr,
+    params: list<string>,
+    // In replys:
```
mavam, line 124:
```suggestion
    // In replies:
```

schema/types/suricata.schema
```diff
@@ -97,17 +97,39 @@ type suricata.dcerpc = suricata.component.common + record {
   }
 }
 
+// https://github.com/OISF/suricata/blob/master/rust/src/dhcp/logger.rs
```
dominiklohmann, line 100: Please write this comment in a user-consumable way, since the schema files are user-facing.

schema/types/suricata.schema
dominiklohmann, line 124:
```suggestion
    // In replies:
```

dominiklohmann (changes requested):

schema/types/suricata.schema
tobim: We can't because "<unknown>" is not a valid token in an enum alternatives list.

schema/types/suricata.schema
tobim: Backwards yes, but less forward compatible then the old definition. It is a tradeoff between compatibility and space optimization.

schema/types/suricata.schema
tobim: I forgot to add that, can you live without it so we don't have to retrigger CI again?

dominiklohmann (approved):

PR #1855, opened 2021-08-19
Turn version components into strings

dominiklohmann:
Turns out `08` is not a valid integer literal in C++, which caused the 2021.08.26-rc1 build to fail. :seenoevil:


lava (approved):
If this compiles then it should be fine for now, although at some point we'll probably also need to provide version numbers since with strings it's a bit annoying for downstream users to actually do something depending on a specific version.

PR #1857, opened 2021-08-20
Fix logging in systemd when built without support

tobim:

VAST now only switches to `journald` style logging by default when it is actually supported.

### :memo: Checklist


lava (approved):
Looks good to me.

dominiklohmann:
Please correct the comment in the vast.yaml.example as well and comment out the default value.

dominiklohmann (approved):

PR #1858, opened 2021-08-20
Support quoted non-string fields in the CSV parser

dominiklohmann:

Before this change, only the `field` parser which we used for column names and string fields respected possible quoting and escaping of fields. Now, all non-container types support quoting.

### :memo: Checklist

Read the code. The tests should be self explanatory. The changed tests failed before the parser changes.


lava (approved):
Looks good, and even comes with unit tests!

It would be interesting to have a before/after comparsion of CSV import speed, since there is now an additional branch for every value.

dominiklohmann:
> It would be interesting to have a before/after comparsion of CSV import speed, since there is now an additional branch for every value.

I doubt it'll be too much, there are already many branches in the parser and this will mostly be just a check for a double quote that returns false so we go back to the previous behavior. And considering this is a bug fix I don't really see how we can improve upon this, sadly.

PR #1860, opened 2021-08-23
Bundle an example configuration file with plugins

dominiklohmann:

If present in the plugin source directory, the build scaffolding now automatically installs `<plugin>.yaml.example` files, commenting out every line so the file has no effect. This serves as documentation for operators that can modify the installed file in-place.

### :memo: Checklist

Most of this is refactoring the code we already used for vast.yaml.example into a function. The rest is re-using that for plugins.


cmake/VASTRegisterPlugin.cmake
```diff
@@ -451,6 +500,14 @@ function (VASTRegisterPlugin)
     endif ()
   endif ()
 
+  # Install an example configuration file, if it exists at the plugin project root.
+  if (EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/${PLUGIN_TARGET}.yaml.example")
+    VASTInstallExampleConfiguration(
+      ${PLUGIN_TARGET}
+      "${CMAKE_CURRENT_SOURCE_DIR}/${PLUGIN_TARGET}.yaml.example"
+      "plugin/${PLUGIN_TARGET}.yaml")
+  endif ()
```
tobim, line 509: Instead of doing this automagically, did you consider adding an optional parameter to VASTRegisterPlugin? I.e.:
```
VASTRegisterPlugin(
  TARGET example-analyzer
  ENTRYPOINT example_analyzer.cpp
  # SOURCES
  #   list additional source files here
  TEST_SOURCES tests/example_analyzer.cpp
  SAMPLE_CONFIGURATION example-analyzer.yaml.example)
```

dominiklohmann: I did consider it, but we do it for schemas and integration tests automatically as well so this felt more consistent. If we want to change that, let's change across the board for the next release.

tobim: Fair enough.

tobim (approved):

PR #1861, opened 2021-08-24
Story/ch28078/plugins not an array

dominiklohmann:

This fixes two variations of errors with the following configuration:

```yaml
vast:
  plugins: all
```

First, we did not allow the key to contain a single string. Second, if we also specified the option on the command line we did not merge the options as expected.

### :memo: Checklist

Test locally.


libvast/src/system/configuration.cpp
```diff
@@ -261,8 +261,13 @@ caf::error configuration::parse(int argc, char** argv) {
         return x.error();
     } else {
       // If the option is not relevant to CAF's custom options, we just store
-      // the value directly in the content.
-      put(content, key, value);
+      // the value directly in the content, or append if it's a list.
+      if (auto content_list
+          = caf::get_if<caf::config_value::list>(&content, key)) {
+        content_list->emplace_back(value);
```
tobim, line 267: What if value is a list but "content" is a single value?

tobim (changes requested):

libvast/src/plugin.cpp
```diff
@@ -93,6 +93,10 @@ load(std::vector<std::string> bundled_plugins, caf::actor_system_config& cfg) {
   // Step 1: Get the necessary options.
   auto paths_or_names
     = caf::get_or(cfg, "vast.plugins", std::vector<std::string>{});
```
lava, line 95: I think this would still silently fail if we specify e.g. a number for `vast.plugins`. Probably a more robust approach would be like

```
if (caf::holds_alternative<std::vector<std::string>>(cfg, "vast.plugins))
  // ...
else if (caf::holds_alternative<std::string>(cfg, "vast.plugins"))
  // ...
else
  return make_error();
```

Also, i'm wondering what would happen if someone were to try to write something like this in the config, would this be regarded as a single string containing whitespace and comma?
```
vast:
  plugins: all, matcher
```

libvast/src/system/configuration.cpp
lava, line 267: Do we need a special case here if `value` is also a list, so avoid creating a nested list?

Also, I'm not 100% sure we want to unconditionally merge here, wouldn't this imply that if there is e.g. `plugins: all` configured in some config file, there is no way to override this to *not* load all plugins from the command line?

lava (commented):
I did not test locally yet, but I already had some questions about it.

libvast/src/plugin.cpp
dominiklohmann: > Also, i'm wondering what would happen if someone were to try to write something like this in the config, would this be regarded as a single string containing whitespace and comma?
> 
> ```
> vast:
>   plugins: all, matcher
> ```

Per the YMAL specification that would try to load a single plugin names "all, matcher", which would fail to load because it does not exist.

> I think this would still silently fail if we specify e.g. a number for `vast.plugins`. Probably a more robust approach would be like
> 
> ```
> if (caf::holds_alternative<std::vector<std::string>>(cfg, "vast.plugins))
>   // ...
> else if (caf::holds_alternative<std::string>(cfg, "vast.plugins"))
>   // ...
> else
>   return make_error();
> ```

Can't do that since it's still possible not to specify the key at all. Until we have some form of validation for our config I fear this cannot be implemented properly.

libvast/src/system/configuration.cpp
dominiklohmann: See below in the other feedback.

dominiklohmann: We handle all four possible combinations now, see the last commit.

libvast/src/plugin.cpp
lava: > Can't do that since it's still possible not to specify the key at all

This could be worked around by a `put_missing` in front:
```
  caf::put_missing(opts, "vast.plugins", std::vector<std::string>{});
```

libvast/src/system/configuration.cpp
lava: What about the second question? I see there is e.g. a separate `policy::merge_lists` parameter in `detail::merge_settings`, whereas here we seem to merge unconditionally. Is that intentional? (honest question, I'm not 100% sure off the top of my head how the two are related)

dominiklohmann: > Also, I'm not 100% sure we want to unconditionally merge here, wouldn't this imply that if there is e.g. `plugins: all` configured in some config file, there is no way to override this to _not_ load all plugins from the command line?

There already is no way to override this because we merge lists from configuration files.

PR #1862, opened 2021-08-24
Print memory counts in bytes instead of kB

tobim:
### :memo: Checklist

Please test on macOS.


changelog/unreleased/changes/1862--bytes-please.md
```diff
@@ -0,0 +1,2 @@
+The memory counts in the output of `vast status` from now on use bytes
+instead of kilobytes for the unit.
```
dominiklohmann, line 2: I would label this a bug fix rather than a change, because we used bytes and kilobytes inconsistently in the status output. Additionally, small typo:

```suggestion
The memory counts in the output of `vast status` now represent bytes consistently, as opposed to a mix of bytes and kilobytes. 
```

libvast/src/detail/process.cpp
```diff
@@ -89,8 +90,10 @@ static caf::settings get_settings_mach() {
     return {};
   struct task_basic_info t_info;
   mach_msg_type_number_t t_info_count = TASK_BASIC_INFO_COUNT;
-  task_info(task, TASK_BASIC_INFO, (task_info_t) &t_info, &t_info_count);
-  result["current-memory-usage"] = t_info.resident_size;
+  task_info(task, TASK_BASIC_INFO, (task_info_t)&t_info, &t_info_count);
+  // http://web.mit.edu/darwin/src/modules/xnu/osfmk/man/task_basic_info.html
+  // says the resident set size is counted in pages, so we multiply accordingly.
+  result["current-memory-usage"] = t_info.resident_size * PAGE_SIZE;
```
dominiklohmann, line 96: I can confirm the reported value looks about correct.

Would it make sense to use `getpagesize()` to determine the value at runtime rather than using this macro? Not sure if it can be changed.

dominiklohmann (changes requested):
```
$ vast status documentation 
[...]
The unit of measurement for memory sizes is kilobytes.
[...]
```

Please update that text as well.

libvast/src/detail/process.cpp
tobim: According to https://github.com/giampaolo/psutil/issues/1906 `getpagesize()` is deprecated in Big Sur.

dominiklohmann (approved):

PR #1864, opened 2021-08-25
Verify partition synopsis update and fill in correct information

lava:

- Check for errors when writing partition synopsis
- Fill in missing offset/events information for partition synopses

### :memo: Checklist

Import some data using `VAST 2021.07.29`, then do some queries using a build from this PR to verify that queries work and VAST doesn't crash with assertion failures.

### DIFF
--- a/libvast/src/partition_synopsis.cpp
+++ b/libvast/src/partition_synopsis.cpp
@@ -113,15 +113,13 @@ pack(flatbuffers::FlatBufferBuilder& builder, const partition_synopsis& x) {
   return ps_builder.Finish();
 }
 
-caf::error
-unpack(const fbs::partition_synopsis::v0& x, partition_synopsis& ps) {
-  if (!x.synopses())
-    return caf::make_error(ec::format_error, "missing synopses");
-  if (!x.id_range())
-    return caf::make_error(ec::format_error, "missing id range");
-  ps.offset = x.id_range()->begin();
-  ps.events = x.id_range()->end() - x.id_range()->begin();
-  for (auto synopsis : *x.synopses()) {
+namespace {
+
+// Not publicly exposed because it doesn't fully initialize `ps`.
+caf::error unpack_(
+  const flatbuffers::Vector<flatbuffers::Offset<fbs::synopsis::v0>>& synopses,
+  partition_synopsis& ps) {
+  for (auto synopsis : synopses) {
     if (!synopsis)
       return caf::make_error(ec::format_error, "synopsis is null");
     qualified_record_field qf;
@@ -139,6 +137,19 @@ unpack(const fbs::partition_synopsis::v0& x, partition_synopsis& ps) {
   return caf::none;
 }
 
+} // namespace
+
+caf::error
+unpack(const fbs::partition_synopsis::v0& x, partition_synopsis& ps) {
+  if (!x.id_range())
+    return caf::make_error(ec::format_error, "missing id range");
+  ps.offset = x.id_range()->begin();
+  ps.events = x.id_range()->end() - x.id_range()->begin();
+  if (!x.synopses())
+    return caf::make_error(ec::format_error, "missing synopses");
+  return unpack_(*x.synopses(), ps);
+}
+
 // This overload is for supporting the transition from VAST 2021.07.29
 // to 2021.08.26, where the `id_range` field was added to partition
 // synopsis. It can be removed once these versions are unsupported.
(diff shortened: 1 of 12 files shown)


libvast/src/system/partition.cpp
```diff
@@ -387,6 +387,10 @@ caf::error unpack(const fbs::partition::v0& x, partition_synopsis& ps) {
     return caf::make_error(ec::format_error, "missing partition synopsis");
   if (!x.type_ids())
     return caf::make_error(ec::format_error, "missing type_ids");
+  // The id_range was only added in VAST 2021.08.26, so we fill it
+  // from the data in the partition if it does not exist.
+  if (!x.partition_synopsis()->id_range())
+    return unpack(*x.partition_synopsis(), ps, x.offset(), x.events());
```
tobim, line 393: Wouldn't just assigning them suffice?
```suggestion
  if (!x.partition_synopsis()->id_range()) {
    ps.offset = x.offset();
    ps.events = x.events();
  }
```

lava: I had that version initially, but it also requires permanently removing the check for `id_range` in the regular `unpack()` function, which would need to allow "invalid" flatbuffers trusting that the caller fixes it up later. Which felt less clean than making a separate overload for this.

libvast/src/partition_synopsis.cpp
```diff
@@ -139,4 +139,33 @@ unpack(const fbs::partition_synopsis::v0& x, partition_synopsis& ps) {
   return caf::none;
 }
 
+// This overload is for supporting the transition from VAST 2021.07.29
+// to 2021.08.26, where the `id_range` field was added to partition
+// synopsis. It can be removed once these versions are unsupported.
+caf::error unpack(const fbs::partition_synopsis::v0& x, partition_synopsis& ps,
+                  uint64_t offset, uint64_t events) {
```
dominiklohmann, line 146: Can we express this in terms of the other (or vice versa) to avoid redundancy? This is a lot of duplicated code currently.

libvast/vast/partition_synopsis.hpp
```diff
@@ -47,6 +47,10 @@ struct partition_synopsis {
 
   friend caf::error
   unpack(const fbs::partition_synopsis::v0&, partition_synopsis&);
+
+  friend caf::error
+  unpack(const fbs::partition_synopsis::v0& x, partition_synopsis& ps,
```
dominiklohmann, line 52: Please sprinkle some `[[nodiscard]]` across the code base on these.

libvast/src/partition_synopsis.cpp
lava: I extracted the loop into a separate function.

Also, one nice thing is that this function does not need to be kept in sync, as it will only be used for partition synopses created before now. So any future features only need to be added to the regular `unpack()` overload.

libvast/src/partition_synopsis.cpp
```diff
@@ -139,4 +137,35 @@ unpack(const fbs::partition_synopsis::v0& x, partition_synopsis& ps) {
...
+  // We should not end up in this overload when an id range already exists.
+  VAST_ASSERT(!x.id_range());
+  ps.offset = offset;
+  ps.events = events;
+  if (!x.synopses())
+    return caf::make_error(ec::format_error, "missing synopses");
+  if (auto error = unpack_(*x.synopses(), ps))
+    return error;
+  return caf::none;
```
dominiklohmann, line 168:
```suggestion
  return unpack_(*x.synopses(), ps);
```

libvast/src/partition_synopsis.cpp
```diff
@@ -139,4 +137,35 @@ unpack(const fbs::partition_synopsis::v0& x, partition_synopsis& ps) {
...
+unpack(const fbs::partition_synopsis::v0& x, partition_synopsis& ps) {
+  if (!x.id_range())
+    return caf::make_error(ec::format_error, "missing id range");
+  ps.offset = x.id_range()->begin();
+  ps.events = x.id_range()->end() - x.id_range()->begin();
+  if (!x.synopses())
+    return caf::make_error(ec::format_error, "missing synopses");
+  if (auto error = unpack_(*x.synopses(), ps))
+    return error;
+  return caf::none;
```
dominiklohmann, line 152:
```suggestion
  return unpack_(*x.synopses(), ps);
```

tobim (approved):
We discussed the chosen solution in a call and consider it as acceptable.

PR #1866, opened 2021-08-25
Add tool to regenerate .mdx files

lava:

On a large DB, it can take a lot of time to regenerate the .mdx synopsis files on startup if the previous version somehow got corrupted or outdated. VAST never accesses these files during operation, so this can safely be done *before* restarting VAST to ensure minimal downtime. This tool provides just that functionality, ie. a way to bulk-update all `.mdx` files of a database that does not interfere with the running VAST process.

### :memo: Checklist


dominiklohmann (approved):
Just some smaller comments on the README.

The tool itself seems to work as expected. I tested this by using an older version of VAST to create a database, and then using this tool on the database before running a recent version of VAST. Everything started up as expected.

PR #1872, opened 2021-08-30
Fix possible garbage in status command output

dominiklohmann:

This had a slight race condition: We captured a `string_view` to every component's type, and used that as a key in the status output. However, when small string optimization is used, the addition/removal of components caused the string views to be invalidated. This change makes it so the status handler copies the keys.

### :memo: Checklist

Reasoning about the change should suffice, it's really simple in hindsight.

### DIFF
--- a/libvast/vast/system/status.hpp
+++ b/libvast/vast/system/status.hpp
@@ -139,12 +139,12 @@ void collect_status(
   collect_status(
     rs, timeout, verbosity, responder,
     [key = std::string{key}, &s](caf::settings& response) {
-      put(s, std::string_view{key}, std::move(response));
+      put(s, key, std::move(response));
     },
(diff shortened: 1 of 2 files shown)


libvast/vast/system/status.hpp
```diff
@@ -138,10 +138,10 @@ void collect_status(
   caf::settings& s, std::string_view key) {
   collect_status(
     rs, timeout, verbosity, responder,
-    [key, &s](caf::settings& response) {
+    [key = std::string{key}, &s](caf::settings& response) {
```
tobim, line 141: There might be another minor cleanup possible: Do we actually have to use the explicit conversion to `std::string_view` on lines 142 and 145?

tobim (approved):
I verified that the issue is resolved by testing locally.

PR #1873, opened 2021-08-31
Fix ignoring of remote sources and sinks for status

dominiklohmann:

We erroneously compared the remote actors proxy actor system's node ID rather than the remote actors actual node ID when checking whether an actor is remote.

### :memo: Checklist

Try locally.


tobim (approved):
I locally verified that the problem is fixed.

PR #1884, opened 2021-09-20
Fix possible desync in pending queries map

dominiklohmann:

This changes the INDEX to monitor EXPORTER actors in order to be able to remove pending queries when an EXPORTER goes down unexpectedly without cancelling its running queries. This fixes a mismatch in the pending queries map in the output of `vast status --detailed`, and also stops the index from doing further unnecessary work if an export was unexpectedly killed.

### :memo: Checklist

Commit-by-commit. Play around with it locally by killing processes off.


lava (approved):
I did not have a good way to test this locally, but logically it looks like the fix is correct and should work.

PR #1885, opened 2021-09-21
Tailor expressions in filter operation

dominiklohmann:

We ran into a `vast::die` call due to an unresolved field extractor when using the filter function with a field extractor. This most notably caused import filter like `event_type == "alert"` to fail.

### :memo: Checklist

Commit-by-commit.


libvast/src/table_slice.cpp
```diff
@@ -706,9 +712,7 @@ filter(const table_slice& slice, const expression& expr, const ids& hints) {
   VAST_ASSERT(builder);
   auto flat_layout = flatten(slice.layout());
   auto check = [&](row_evaluator eval) {
-    if (expr == expression{})
-      return true;
-    return caf::visit(eval, expr);
+    return caf::visit(eval, *tailored_expr);
```
lava, line 715: Isn't this inverting the logic for the case of an empty expression? As far as I understand we used to return everything and now we return nothing. (although the tests seem to pass, so I'm probably misreading this somehow)

lava (commented):
The fix looks good, but from the description it sounds like we can very easily add the problematic query as a new test.

libvast/src/table_slice.cpp
dominiklohmann: Yep, and that's why the integration test is failing as well. Working on a fix rn.

lava (approved):
Looks good, thanks for adding the test!

PR #1888, opened 2021-09-23
Type FlatBuffers

dominiklohmann:

This refactors our type system to be based on a single, extensible FlatBuffers table. This comes with multiple advantages:
- `type` is now cache friendly. It's probably safe to say that this was the biggest downside of the old type system. Iterating over the leaves of a a `record_type` is no longer significantly slower than iterating over just the fields.
- `type` is now sliceable from other FlatBuffers table (e.g., Table Slices), and shares its lifetime with the sliced-from underlying chunk.  `type` must no longer be deserialized as its underlying data format is a contiguous buffer now.

I've taken the opportunity to clean up the type API at the same time. This includes the following developer-facing changes:
- The `alias_type` no longer exists. Visiting a `type` steps through aliases transparently.
- ~Attributes are now called tags.~ (reverted late on)
- The `record_type` API no longer operates with keys directly, but rather only with indices and offsets. The functions `record_type::resolve_prefix` and `record_type::resolve_suffix` must be used to convert a key into offsets.
- The `enumeration_type` API allows for explicitly assigning keys to individual fields. This is preparatory work for aligning our data model with Apache Arrow's, which offers the related [`arrow::DictionaryType`](https://arrow.apache.org/docs/cpp/api/datatype.html?highlight=dictionarytype#_CPPv4N5arrow14DictionaryTypeE).
- The functions `record_type::fields` and `record_type::leaves` return iterables over fields and leaves of a record respectively.
- Complex types are no longer default-constructible.
- The new `concrete_type`, `basic_type`, and `complex_type` concepts make writing generic visitors a lot easier.

There exists conversions between types and legacy type in both directions to ease the transition. The conversion from type to legacy type will be removed towards the end of this PR.

The user-facing changes of this PR are hopefully minimal. The conversion from `legacy_type` to `type` when exporting old data from pre-existing databases may incur a small performance overhead, but for newly imported data this should speed up both the import and export pipelines across the board. There are a lot of smaller bug fixes for edge cases with leaf field key handling and alias type resolution throughout the code base.

### :vertical_traffic_light: Status

- [x] Design the new type FlatBuffers table and the basic type API.
- [x] Add support for basic types.
- [x] Add support for type aliases and tags.
- [x] Add support for list types.
- [x] Add support for map types.
- [x] Add support for enumeration types.
- [x] Add support for record types.
- [x] Implement new versions of encoding-specific table slice FlatBuffers tables.
- [x] Convert libvast to operate on new type system directly.
- [x] Convert unit tests, tools, and plugins to operate on new type system directly.
- [x] Remove now dead code from the legacy type API (most operations should be unnecessary now).
- [x] ~Consider implementing a new version of the partition FlatBuffers table.~ Not in this PR.
- [x] ~Consider making `record_type::iterable::iterator` a random access iterator.~ Not in this PR.
- [x] ~Consider making `record_type::leaf_iterable::iterator` a bidirectional iterator.~ Not in this PR.

### :memo: Checklist

1. Look at the new Type FlatBuffers table and at the new `type.hpp` file.
2. Look at the individual commits. Most of them should be self-explanatory.
3. Wait until this PR is out of draft mode.
3. Measure overall performance impact on both pre-existing and newly created databases.


libvast/src/system/meta_index.cpp
```diff
@@ -198,16 +199,17 @@ std::vector<uuid> meta_index_state::lookup_impl(const expression& expr) const {
...
       auto search = [&](auto match) {
         VAST_ASSERT(caf::holds_alternative<data>(x.rhs));
         const auto& rhs = caf::get<data>(x.rhs);
         result_type result;
         for (const auto& [part_id, part_syn] : synopses) {
           for (const auto& [field, syn] : part_syn.field_synopses_) {
             if (match(field)) {
-              auto cleaned_type = vast::legacy_type{field.type}.attributes({});
+              auto cleaned_type = field.type();
+              cleaned_type.prune_metadata();
```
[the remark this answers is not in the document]
dominiklohmann: This is because type synopses only work for basic types. We want to add a function like this for that purpose:

```cpp
friend auto basic_type(const type& in) -> std::optional<type> {
  auto f = detail::overload{
    []<basic_type T>(const T& out) -> std::optional<type> {
      return out;
    },
    []<complex_type T>(const T&) -> std::optional<type> {
      return std::nullopt;
    },
  };
  return caf::visit(f, in);
};
```

libvast/src/system/infer_command.cpp
```diff
@@ -45,54 +45,58 @@ namespace {
...
   VAST_ASSERT(n == 1);
   schema result;
-  result.add(std::move(rec));
+  if (rec)
+    result.add(*rec);
   return result;
 }
 
-legacy_type deduce(simdjson::dom::element e) {
+type deduce(simdjson::dom::element e) {
```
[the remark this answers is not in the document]
dominiklohmann: This is invalid, as extract takes a type already. Not sure what I was thinking here.

libvast/src/arrow_table_slice.cpp
```diff
@@ -645,21 +639,22 @@ class record_batch_decoder final {
...
+  if constexpr (std::is_same_v<FlatBuffer, fbs::table_slice::arrow::v0>) {
+    // This legacy type has to stay; it is deserialized from disk.
+    auto intermediate = legacy_record_type{};
+    if (auto err = fbs::deserialize_bytes(slice_.layout(), intermediate))
+      die("failed to deserialize layout: " + render(err));
+    state_.layout = type::from_legacy_type(intermediate);
+    auto decoder = record_batch_decoder{};
+    state_.record_batch = decoder.decode(slice.schema(), slice.record_batch());
+  } else {
+    state_.layout = type{parent->slice(as_bytes(*slice_.layout()))};
```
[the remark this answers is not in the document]
dominiklohmann: Actually we don't want to do this, as we cannot slice it, so we cannot bind the lifetime correctly. I've added a comment inline.

libvast/vast/type.hpp
```diff
@@ -0,0 +1,1258 @@
...
+
+  /// Resolves a flat index into an offset.
+  [[nodiscard]] offset resolve_flat_index(size_t flat_index) const noexcept;
+
+  /// Resolves a key into an offset.
+  /// @note This only matches on full keys, so the key 'x.y'  matches 'x.y.z'
+  /// but not 'x.y_other.z' .
+  /// @note The key may optionally begin with the type name for aliased types.
+  [[nodiscard]] std::optional<offset>
+  resolve_prefix(std::string_view key) const noexcept;
```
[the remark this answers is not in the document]
dominiklohmann: This is done.

libvast/vast/type.hpp
```diff
@@ -0,0 +1,1258 @@
...
+
+  /// Resolves a key into a list of offsets by suffix matching the given key.
+  /// @note This only matches on full keys, so the key 'y.z' matches 'x.y.z' but
+  /// not 'x.other_y.z'.
+  /// @note The key may optionally begin with the type name for aliased types.
+  [[nodiscard]] std::vector<offset>
+  resolve_suffix(std::string_view key) const noexcept;
```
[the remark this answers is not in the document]
dominiklohmann: This is done.

libvast/src/system/active_partition.cpp
```diff
@@ -384,15 +411,20 @@ active_partition_actor::behavior_type active_partition(
                   self->state.persistence_promise.deliver(error);
                   return;
                 }
+                // TODO: Consider storing indexer chunks by the fully qualified
+                // field instead of just its fully qualified name in a future
+                // partition version. As-is, this breaks if multiple fields with
+                // the same fully qualified name but different types exist in
+                // the same partition.
```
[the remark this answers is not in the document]
dominiklohmann: I created a follow-up story on our internal issue tracker to gather requirements for the next-gen partition FlatBuffers table.

libvast/src/system/explorer.cpp
```diff
@@ -131,21 +131,18 @@ explorer(caf::stateful_actor<explorer_state>* self, node_actor node,
...
             return true;
-        }
         return false;
       };
-      auto it = std::find_if(layout.fields.begin(), layout.fields.end(),
-                             is_timestamp);
-      if (it == layout.fields.end()) {
-        VAST_DEBUG("{} could not find timestamp field in {}", *self, layout);
+      auto leaves = layout.type.leaves();
+      auto it = std::find_if(leaves.begin(), leaves.end(), is_timestamp);
```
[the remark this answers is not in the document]
dominiklohmann: I don't think this needs a separate changelog entry. It's such a minor thing compared to many of the other changes.

libvast/vast/table_slice.hpp
```diff
@@ -40,6 +40,12 @@ class table_slice final {
     yes, ///< Enable FlatBuffers table verification.
   };
 
+  /// A named record type.
+  struct layout final {
+    std::string_view name; ///< The name of the layout.
+    record_type type;      ///< The type of the layout.
+  };
```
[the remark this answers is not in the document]
dominiklohmann: I've removed the layout struct; `table_slice::layout()` now simply returns a `type` that is guaranteed to have a name and contain a `record_type`.

vast/integration/reference/node-suricata-rrdata/step_01.ref
```diff
@@ -1,2 +1,2 @@
...
..."rname": nil, "serial": nil, "refresh": nil, "retry": nil, "expire": nil, "minimum": nil>, "sshfp": <"fingerprint": nil, "algo": nil, "type": nil>, "srv": <"priority": nil, "weight": nil, "port": nil, "name": "xxx">>, <"rrname": "google.com", "rrtype": "MX", "ttl": 599, "rdata": "alt2.aspmx.l.google.com", "soa": <"mname": "foo", "rname": nil, "serial": nil, "refresh": nil, "retry": 42, "expire": nil, "minimum": nil>, "sshfp": <"fingerprint": nil, "algo": nil, "type": nil>, "srv": <"priority": nil, "weight": nil, "port": nil, "name": nil>>, <"rrname": "google.com", "rrtype": "MX", "ttl": 599, "rdata": "alt3.aspmx.l.google.com", "soa": <"mname": nil, "rname": nil, "serial": nil, "refresh": nil, "retry": nil, "expire": nil, "minimum": nil>, "sshfp": <"fingerprint": nil, "algo": nil, "type": nil>, "srv": <"priority": nil, "weight": nil, "port": nil, "name": nil>>, <"rrname": "google.com", "rrtype": "MX", "ttl": 599, "rdata": "alt4.aspmx.l.google.com", "soa": <"mname": nil, "rname": nil, "serial": nil, "refresh": nil, "retry": nil, "expire": nil, "minimum": nil>, "sshfp": <"fingerprint": nil, "algo": nil, "type": nil>, "srv": <"priority": nil, "weight": nil, "port": nil, "name": nil>>]>
```
dominiklohmann, line 2: From peer-review with @tobim: We want to add a workaround to the JSON reader to revert to the old behavior, even though the new one is more consistent. In the future we want to stop flattening entirely, but this requires a redesign of the table slice builder API.

dominiklohmann: This is now implemented.

libvast/src/table_slice_column.cpp
```diff
@@ -27,29 +27,29 @@ table_slice_column::table_slice_column(table_slice_column&&) noexcept = default;
...
+  // nop
 }
 
 std::optional<table_slice_column>
 table_slice_column::make(table_slice slice, std::string_view column) noexcept {
-  auto&& layout = slice.layout();
-  const auto offsets = layout.find_suffix(column);
+  const auto layout = slice.layout().type;
+  // TODO: Doesn't it make way more sense to resolve the column name exactly
+  // here instead of taking the first column that suffix-matches the name?
```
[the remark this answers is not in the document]
dominiklohmann: I've instead taken the liberty to remove this function in its entirety. The call site now has to determine what it wants to do.

libvast/src/system/active_partition.cpp
```diff
@@ -90,12 +95,23 @@ void active_partition_state::notify_flush_listeners() {
...
-  return filter == column.field();
+  const auto rhs_layout = column.slice().layout();
+  const auto rhs_field_offset
+    = rhs_layout.type.resolve_flat_index(column.index());
+  const auto rhs_field = rhs_layout.type.field(rhs_field_offset);
+  // name check
+  const auto lhs = filter.name();
+  // FIXME: use mismatch twice here.
+  if (filter.name() != fmt::format("{}.{}", rhs_layout.name, rhs_field.name))
+    return false;
```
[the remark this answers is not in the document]
dominiklohmann: This is done.

libvast/src/system/source.cpp
```diff
@@ -45,20 +45,17 @@ void source_state::initialize(const type_registry_actor& type_registry,
...
-              if (is_valid(*layout))
+            if (type.name().starts_with(type_filter))
+              if (const auto* layout = caf::get_if<vast::record_type>(&type))
                 merged_schema.add(*layout);
           // Second, filter valid types from all available record types.
           for (auto& type : types)
-            if (auto&& layout = caf::get_if<vast::legacy_record_type>(&type))
-              if (is_valid(*layout))
+            if (type.name().starts_with(type_filter))
+              if (auto&& layout = caf::get_if<vast::record_type>(&type))
```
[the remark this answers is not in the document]
dominiklohmann: This is now implemented.

libvast/vast/type.hpp
```diff
@@ -0,0 +1,1258 @@
...
+    return f(as_bytes(x));
+  }
+
+  /// Integrates with the CAF stringification inspector.
+  /// TODO: Implement for all fmt::is_formattable<T, char>.
+  friend void inspect(caf::detail::stringification_inspector& f, type& x);
+
+  /// Explicitly forbid usage of the CAF binary serializer/deserializer.
+  friend auto inspect(caf::binary_serializer&, type&) = delete;
+  friend auto inspect(caf::binary_deserializer&, type&) = delete;
```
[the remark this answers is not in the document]
dominiklohmann: This is now implemented.

libvast/src/system/type_registry.cpp
```diff
@@ -110,15 +118,22 @@ caf::error type_registry_state::load_from_disk() {
...
-    if (auto error = source(data))
+    std::map<std::string, detail::stable_set<legacy_type>> intermediate = {};
+    if (auto error = source(intermediate))
       return error;
+    for (const auto& [k, vs] : intermediate) {
+      auto entry = type_set{};
+      for (const auto& v : vs)
+        entry.emplace(type::from_legacy_type(v));
+      data.emplace(k, entry);
+    }
```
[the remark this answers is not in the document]
dominiklohmann: I decided against doing this. We shouldn't do too much magic.

libvast/vast/type.hpp
```diff
@@ -825,6 +836,10 @@ class record_type final : public stateful_type_base {
   explicit record_type(std::initializer_list<field_view> fields) noexcept;
   explicit record_type(const std::vector<struct field>& fields) noexcept;
 
+  /// Returns the underlying FlatBuffers table representation.
+  /// @note Include `vast/fbs/type.hpp` to be able to use this function.
```
mavam, line 840: I don't think we need to tell the user what to include in order to use a function.

libvast/vast/type.hpp
```diff
@@ -671,6 +674,10 @@ class list_type final : public stateful_type_base {
     // nop
   }
 
+  /// Returns the underlying FlatBuffers table representation.
+  /// @note Include `vast/fbs/type.hpp` to be able to use this function.
+  [[nodiscard]] const fbs::Type& table() const noexcept;
```
mavam, line 679: To me, this should rather be a free function in the namespace `fbs`, e.g., `template <class T> const T& fbs::get();`. It's an implementation detail and not part of the primary interface the class `type` provides to the user.

tobim (commented):
3180479fbc94dffdbf6fca7953ca4f4929923cac is not necessary. The GCC11 problem can be fixed with `s/nullptr_t/std::nullptr_t/`.

tobim:
The build failure of `type.cpp` under GCC seems to be due to a bug in our version of function2. The [upstream fix](https://github.com/Naios/function2/pull/43/files) doesn't fully solve the problem though.

dominiklohmann:
> [3180479](https://github.com/tenzir/vast/commit/3180479fbc94dffdbf6fca7953ca4f4929923cac) is not necessary. The GCC11 problem can be fixed with `s/nullptr_t/std::nullptr_t/`.

Thanks, will fix.

> The build failure of `type.cpp` under GCC seems to be due to a bug in our version of function2. The [upstream fix](https://github.com/Naios/function2/pull/43/files) doesn't fully solve the problem though.

We can always switch to the slightly less performance `caf::detail::unique_function`, although that is not noexcept-correct. We just cannot use `std::function` because we must be able to move lambda captures for the transformations.

libvast/vast/type.hpp
dominiklohmann: I disagree. `get` is a terrible name, and it is not an implementation detail but rather part of the APIs of `type` and all complex concrete types, because you might want to fall back to using the underlying FlatBuffers API directly.

libvast/vast/type.hpp
dominiklohmann: Fair point. I'll add a comment at the top of the `vast/type.hpp` file though to never include `vast/fbs/type.hpp` in the header, so reviewers don't miss this if IWYU added it by accident.

mavam:
> We can always switch to the slightly less performant `caf::detail::unique_function`, although that is not noexcept-correct.

I would do that. Unless you measure a difference, it even allows us to get rid of a dependency.

dominiklohmann:
> I would do that. Unless you measure a difference, it even allows us to get rid of a dependency.

It doesn't allow us to get rid of a dependency, because `chunk` still depends on it. It's merely a fallback we can use in this one place. For `chunk` we call the deleter function in the destructor, and that really should be noexcept-correct.

libvast/vast/type.hpp
mavam: > but rather part of the APIs of `type`

Calling `.table()` doesn't do any meaningful type operation, it's purely exposing a different representation.

> because you might want to fall back to using the underlying FlatBuffers API directly

We have the same for `pack` and `unpack`. These are flatbuffers-specific CPOs. Here, exposing an implementation detail, which a lot of *different* flatbuffers-backed types do, should be a free function. Whether you call it `get` or something else is irrelevant for this discussion, e.g., `flatbuffer(x)` or `table(x)` are probably nicer.

dominiklohmann: It'd still be majorly confusing, because for a given `record_type x`, `table(rt)` would return the same type as `table(type{rt})`, but with different assumptions that can be made about the resulting type. It's an operation on the type that for concrete complex types resolves some layers transparently.

It's also not a CPO. We can introduce a CPO when we have generic code that calls the CPO via ADL or similar means, but in this case it really isn't that—there's nothing you can do on the result of this function generically.

tobim:
I had to apply the this change to make the integration tests not crash:
```diff
diff --git a/libvast/src/type.cpp b/libvast/src/type.cpp
index 6f1a6dfbd..4fad02571 100644
--- a/libvast/src/type.cpp
+++ b/libvast/src/type.cpp
@@ -1969,7 +1969,7 @@ std::optional<record_type> record_type::transform(
           std::move(replacements.rbegin(), replacements.rend(),
                     std::back_inserter(new_fields));
           ++current;
-        } while (current->first == index);
+        } while (current != end && current->first == index);
       } else if (index_mismatch == index.end()) {
         // The index is a prefix of the target offset for the next
         // transformation, so we recurse one level deeper.
@@ -1982,7 +1982,7 @@ std::optional<record_type> record_type::transform(
             std::move(*sub_result),
           });
         // Check for invalid arguments on the way in.
-        VAST_ASSERT(index != current->first,
+        VAST_ASSERT(current == end || index != current->first,
                     "cannot apply transformations to both a nested record type "
                     "and its children at the same time.");
       } else {
```

dominiklohmann:
> I had to apply the this change to make the integration tests not crash:

That's a nice find, that was certainly UB before.

libvast/src/type.cpp
```diff
@@ -2096,6 +2096,29 @@ record_type::insert_after(std::vector<struct field> fields) noexcept {
 
 std::optional<record_type> record_type::transform(
   std::vector<transformation> transformations) const noexcept {
+  // This function recursively calls do_transform while iterating over all the
+  // leafs of the record type backwards.
```
mavam, line 2100: leaves

libvast/test/type.cpp
```diff
@@ -0,0 +1,960 @@
...
+}
+
+TEST(list_type) {
+  static_assert(concrete_type<list_type>);
+  static_assert(!basic_type<list_type>);
+  static_assert(complex_type<list_type>);
+  const auto t = type{};
+  const auto lit = type{list_type{integer_type{}}};
+  CHECK(lit);
+  CHECK_EQUAL(as_bytes(lit), as_bytes(list_type{integer_type{}}));
```
tobim, line 277: This line triggers the address sanitizer with GCC. Presumably because the destructor of the `list_type` is invoked immediately after `as_bytes` finishes. The span obviously can't extend the lifetime, so I think this should be expected.

I wonder if an rvalue overload `as_bytes(chunk&&) = delete;` would work as a warning system.

dominiklohmann: Nice catch!

This occurs mostly because of how `CHECK_EQUAL` is implemented; it suffers from the same issues as the range-for loop as detailed in https://wg21.link/p2012.

It's very hard to find these kinda issues when macros are involved. We can generally forbid `as_bytes` on rvalues of all types and concrete types, which should help catch these kinda issues.

dominiklohmann: Done. The deleted `as_bytes(list_type&&) = delete` actually found the line you commented on. I wonder why this didn't trigger the address sanitizer for me with clang.

libvast/vast/type.hpp
dominiklohmann: I think #1972 offers a better solution to this. Once we no longer have the chunk we can make this all a bit simpler and remove the `table()` function entirely. It's a future refactoring, though, and I would like to keep it out of this PR, so I am resolving this for now.

libvast/src/type.cpp
```diff
@@ -0,0 +1,2503 @@
...
+      return type{other.name(), enumeration_type{fields}, attributes};
+    },
+    [&](const legacy_list_type& list) {
+      return type{other.name(), list_type{from_legacy_type(list.value_type)},
+                  attributes};
+    },
+    [&](const legacy_map_type& list) {
+      return type{other.name(),
+                  map_type{from_legacy_type(list.key_type),
+                           from_legacy_type(list.value_type)},
```
tobim, line 481:
```suggestion
    [&](const legacy_map_type& map) {
      return type{other.name(),
                  map_type{from_legacy_type(map.key_type),
                           from_legacy_type(map.value_type)},
```

tobim (approved):
This branch has been running on a testbed at Tenzir for the past 15 hours with no unexpected performance changes.

The time to load a partition with the old serialization format has increased slightly increased, but we expect that to return back to the old numbers once the partition uses the new serialization format for its layout.

PR #1892, opened 2021-09-28
Fix response promises for disk monitor deletion

dominiklohmann:

In the testing of the release candidate we noticed the disk monitor not correctly handling deletion in some cases, leading to full disks. These changes are a first attempt at solving the issue.

### :memo: Checklist

- Review code by following the message flow from the `atom::ping` handler of the disk monitor down to the local segment stores.
- @tobim deploy on our testbed and try to reproduce.


libvast/src/system/disk_monitor.cpp
```diff
@@ -106,14 +106,14 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
...
   return {
-    [self](atom::ping) {
-      self->delayed_send(self, self->state.config.scan_interval, atom::ping_v);
+    [self](atom::ping) -> caf::result<void> {
       if (self->state.purging) {
         VAST_DEBUG("{} ignores ping because a deletion is still in "
                    "progress",
                    *self);
-        return;
+        return caf::skip;
```
tobim, line 114: I believe we should not skip in this case. If we do that means the ping will be handled at the moment we reached the `low_water_mark`, and it doesn't make much sense to check in that moment.

libvast/src/system/disk_monitor.cpp
```diff
@@ -133,20 +133,18 @@ disk_monitor(disk_monitor_actor::stateful_pointer<disk_monitor_state> self,
           ->request(static_cast<disk_monitor_actor>(self), caf::infinite,
                     atom::erase_v)
           .then(
-            [] {
+            [=] {
+              self->state.purging = false;
```
tobim, line 137: I think we need to solve this differently. The `erase` handler below returns after scheduling an initial set of partitions, so this `then` handler runs while the deletion is actually just starting. But analyzing the code further I realized that there is an additional problem: In case step size is greater than 1 the `erase` handler seems to turn into a fork bomb. Let's discuss tomorrow.

tobim (changes requested):

dominiklohmann:
The changes you made look fine to me. I can't approve my own PR, but consider your mixups approved @tobim.

tobim (approved):
Approving the parts done by @dominiklohmann.

PR #1893, opened 2021-09-30
Prepare repository for VAST 2021.09.30

dominiklohmann:


libvast/src/system/index.cpp
```diff
@@ -932,23 +934,11 @@ index(index_actor::stateful_pointer<index_state> self,
...
-            it != self->state.monitored_queries.end()) {
-          auto& [_, ids] = *it;
-          ids.erase(query_id);
-          if (ids.empty()) {
-            self->demonitor(sender->address());
-            self->state.monitored_queries.erase(it);
-          }
-        }
-        self->state.pending.erase(query_id);
+        VAST_WARN("{} ignores query {} for zero partitions", *self, query_id);
```
tobim, line 941: This edit makes it impossible to "cancel" a query for a requester that wants to continue with other tasks. While we don't have the use-case right now we should not make such interface changes in a haste.

tobim: We discussed this and came to the conclusion that we can live without this mechanism for now because it isn't used in any index client.

dominiklohmann: We discussed this in a Slack call, and decided to keep this as-is in the PR. We do not have a use case for early cancellation without quitting the receiving actor, and if we want to add such an API back it is really simple to do so.

tobim (approved):

PR #1896, opened 2021-10-01
Story/ch28702/query backlog

tobim:

Before this change the index would push queries back into its own message queue when no worker was available. This means we don't have to skip lots of queries any more whenever the index receives a new message.

The change also introduces a stricter restriction on the number of parallel queries. The old version would interleave the work for all queries, i.e. a worker would switch to a random work package from the pending queries map when it was done with the one before, but now a worker gets assigned to a query until that is completely done.

### :memo: Checklist


dominiklohmann:
@tobim and I collaborated on the earlier. The code looks good to me, but it needs testing (and performance testing!) before merging, and also a changelog entry for the changed behavior of `vast.max-queries` if this turns out to be a performance-relevant change.

PR #1910, opened 2021-10-15
    Explicitly close transformer stream stage after eof

lava:

    Apparently a pending stream handshake can delay the destruction
    of the hosting scheduled actor. For actors that failed to spawn
    successfully, the handshake will never become not pending so we
    need to close the stage manually.

### :memo: Checklist

Review commit-by-commit.

To test, make e.g. the `index/` subdirectory of an existing `vast.db` non-readable. Prior to this PR, `vast` will hang after printing an error, after this PR it should terminate normally.

### DIFF
--- a/changelog/unreleased/bug-fixes/1910--transformer-shutdown.md
+++ b/changelog/unreleased/bug-fixes/1910--transformer-shutdown.md
@@ -1,0 +1,2 @@
+Fixed a bug that would lead to a prevented VAST from shutting
+down properly when the index crashed during startup.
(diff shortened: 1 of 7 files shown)


dominiklohmann (dismissed):
Looks alright to me. ~It's hard to test this manually, can you explain how you verified that this fixes the issue?~ See description.

lava:
As much as I'd like to see this bug fixed, I'll close this PR now until it shows up again in a higher-priority context since investigating the test failures is too involved currently.

PR #1917, opened 2021-10-20
Prefer reading query from stdin if available

dominiklohmann:

A recent change made it so not providing a query causes everything to be exported. However, that made it so providing a query on stdin required the option `vast.export.read` to be set to `"-"`, which was far from intuitive. This change streamlines the option, giving the user an error if conflicting ways to read the query exist and preferring stdin over exporting everything if a query is provided on stdin but not on the command line.

### :memo: Checklist

It's a simple change; reading the code and passing CI should suffice.


libvast/src/system/read_query.cpp
```diff
@@ -88,6 +88,11 @@ read_query(const invocation& inv, std::string_view file_option,
   //   reads the query from stdin, and if none is present, it exports everything
   const auto fname = caf::get_if<std::string>(&inv.options, file_option);
   const bool has_query_cli = inv.arguments.size() > argument_offset;
+  const bool has_query_stdin = [] {
```
lava, line 91: Can we add a comment?

```
// We want to support running queries like `echo ":count == 1" |  vast export json` and like `export json ":count == 1" <query.txt`,
// so if the input is connected to a pipe or regular file we assume that we want to read a query from there.
```

libvast/src/system/read_query.cpp
```diff
@@ -98,13 +103,24 @@ read_query(const invocation& inv, std::string_view file_option,
                     read_query(*fname), *fname));
     if (*fname == "-")
       return read_query(std::cin);
+    if (has_query_stdin)
+      return caf::make_error(
+        ec::invalid_argument,
+        fmt::format("got query '{}' on the command line and '{}' via stdin",
```
lava, line 109: I think the error should be `"got query '{}' on the command line and '{}' as input filename"`, and not try to read from `std::cin`.

lava (approved):
Looks good assuming CI is happy this time; a few minor issues but as discussed I can fixup them myself.

libvast/src/system/read_query.cpp
lava: Adjusted the comment right above instead.

mavam:
Finally, what a pesky issue!

PR #1944, opened 2021-11-02
Be more lenient with empty queries on stdin

lava:
After real-world experience, the previous rules had too many
false positives that caused VAST to break due to some file
descriptor being accidentally attached to stdin.

### :memo: Checklist

### DIFF
--- a/libvast/src/system/read_query.cpp
+++ b/libvast/src/system/read_query.cpp
@@ -114,25 +114,27 @@ read_query(const invocation& inv, std::string_view file_option,
                                          *fname));
     return read_query(*fname);
   }
-  if (has_query_stdin && has_query_cli) {
+  // As a special case we ignore stdin unless we get an actual query
+  // from it, since empirically this is often connected to the remnants
+  // of a long-forgotten file descriptor by some distant parent process
+  // when running in any kind of automated environment like a CI runner.
+  std::string cin_query;
+  if (has_query_stdin) {
     auto query = read_query(std::cin);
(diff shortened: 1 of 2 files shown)


libvast/src/system/read_query.cpp
```diff
@@ -114,13 +114,21 @@ read_query(const invocation& inv, std::string_view file_option,
...
-  if (has_query_cli) {
-    if (has_query_stdin)
+  if (has_query_stdin && has_query_cli) {
+    auto query = read_query(std::cin);
+    // As a special case we ignore stdin unless we get an actual query
+    // from it, since empirically this is often connected to the remnants
+    // of some long-forgotten file descriptor by some distant parent
+    // process when running in any kind of automated environment, e.g. CI.
+    if (!query || query->empty())
+      has_query_stdin = false;
```
dominiklohmann, line 124: One case is missing here: if the query is empty and we do not have a query on stdin, we should still export all. E.g.,

```
echo '' | vast export json
```

dominiklohmann (approved):
As discussed via Slack. Just one minor thing missing here.

libvast/src/system/read_query.cpp
mavam: I'm not sure about this, actually. Is "empty query" the same as "no query?"

dominiklohmann: Did you read the context in Slack? We're facing this issue in CI where there's always something on stdin, so we get a lot of problems and false positives. Not doing this additional check will just cause the same trouble for other invocations.

PR #1947, opened 2021-11-03
Fix deadlock for queries with no candidate partitions

dominiklohmann:
With the introduction of the query backlog we changed how we handle the available query workers. This introduced a subtle bug: Queries from anonymous senders and those that did not qualify for any candidate partitions from the meta index did not return the query worker, causing it to be unavailable permanently. This meant that VAST was unable to export data after handling exactly as many queries of that kind as there are query workers available, resulting in a deadlock.

### :memo: Checklist

I've reproduced this before locally and couldn't after this change. Probably makes sense to follow the logic for the handlers that receive a query worker and a query individually, which makes this fix quite obvious in hindsight.


changelog/unreleased/bug-fixes/1947--query-worker-deadlock.md
```diff
@@ -0,0 +1,3 @@
+Queries that resulted in zero results sometimes caused query workers to become
+stuck, possibly resulting in a deadlock when this affected all query workers.
+This now works as expected again.
```
tobim, line 3: Since this problem was introduced after the last release I would skip this message and only add the PR number to the entry for #1929.

tobim (approved):
Great find!

changelog/unreleased/bug-fixes/1947--query-worker-deadlock.md
dominiklohmann: Makes sense.

PR #1950, opened 2021-11-03
Disable precompiled headers

lava:
We noticed these having an overall negative impact on
build times, since they prevent ccache from caching
the build result.

Since precompiled headers are used for every translation
unit, some minor fixups are needed to fix the build
after disabling them.

### :memo: Checklist

### DIFF
--- a/libvast/CMakeLists.txt
+++ b/libvast/CMakeLists.txt
@@ -338,13 +338,6 @@ target_compile_definitions(libvast_native_plugins
                            PRIVATE VAST_ENABLE_NATIVE_PLUGINS)
 add_library(vast::native_plugins ALIAS libvast_native_plugins)
 
-# Configure precompiled headers, except for ARM macOS because of this CMake bug:
-# https://gitlab.kitware.com/cmake/cmake/-/issues/21825
-if (NOT (CMAKE_SYSTEM_NAME MATCHES "Darwin" AND CMAKE_SYSTEM_PROCESSOR MATCHES
-                                                "arm"))
-  target_precompile_headers(libvast PUBLIC [["vast/fwd.hpp"]] <caf/fwd.hpp>)
-endif ()
-
 target_link_libraries(libvast PRIVATE libvast_internal)
 target_link_libraries(libvast_native_plugins PRIVATE libvast libvast_internal)
 
--- a/libvast/vast/hash/uniquely_represented.hpp
+++ b/libvast/vast/hash/uniquely_represented.hpp
@@ -9,6 +9,7 @@
 #pragma once
 
 #include <array>
+#include <cstdint>
 #include <tuple>
 #include <type_traits>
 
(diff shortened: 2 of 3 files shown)


libvast/CMakeLists.txt
```diff
@@ -340,10 +340,10 @@ add_library(vast::native_plugins ALIAS libvast_native_plugins)
 
 # Configure precompiled headers, except for ARM macOS because of this CMake bug:
 # https://gitlab.kitware.com/cmake/cmake/-/issues/21825
-if (NOT (CMAKE_SYSTEM_NAME MATCHES "Darwin" AND CMAKE_SYSTEM_PROCESSOR MATCHES
-                                                "arm"))
-  target_precompile_headers(libvast PUBLIC [["vast/fwd.hpp"]] <caf/fwd.hpp>)
-endif ()
+#if (NOT (CMAKE_SYSTEM_NAME MATCHES "Darwin" AND CMAKE_SYSTEM_PROCESSOR MATCHES
```
dominiklohmann, line 343: We could also make this a bit smarter by detecting whether the CXX_COMPILER_LAUNCHER target property is set on libvast.

mavam:
I can confirm that disabling pre-compiled headers makes `ccache` work.

libvast/vast/hash/uniquely_represented.hpp
```diff
@@ -38,10 +39,10 @@ struct is_uniquely_represented<std::tuple<Ts...>>
                        && ((0 + ... + sizeof(Ts)) == sizeof(std::tuple<Ts...>))> {
 };
 
-template <class T, size_t N>
+template <class T, std::size_t N>
```
mavam, line 42: Why is that needed here?

lava: I think something in `fwd.hpp` used to pull `size_t` into the global namespace (e.g. the `<algorithm>` header has that effect), so without it it leads to compile errors from some translation units here. The alternative would be including `fwd.hpp`, but this seems more minimal.

libvast/CMakeLists.txt
lava: That's brittle, e.g. on Debian the standard way of using ccache is to install it as a symlink for the compiler binary itself, so it can be used even without cmake support:

```
$ which cc
/usr/lib/ccache/cc
```

If anything, I'd suggest making it an option `VAST_ENABLE_PCH`, but I'm not sure who would be the target audience to enable that option.

libvast/vast/hash/uniquely_represented.hpp
dominiklohmann: `size_t` is defined in `<cstddef>`. It's guaranteed to be the same type as `std::size_t`.

libvast/CMakeLists.txt
dominiklohmann: Then I'd prefer removing the lines altogether.

libvast/vast/hash/uniquely_represented.hpp
```diff
@@ -9,6 +9,7 @@
 #pragma once
 
 #include <array>
+#include <cstdint>
```
mavam, line 12:
```suggestion
#include <cstddef>
```

And then change back `std::size_t` to `size_t`, per the other discussion.

dominiklohmann (approved):

PR #1954, opened 2021-11-04
Add metrics about the index workers

tobim:
This additionally changes the status output to use components labels instead of their type name. This means that all exporters should now be printed.

### :memo: Checklist


dominiklohmann (approved):
We're testing this in a Slack call right now in order to catch a regression, so I'm approving as-is.

PR #1956, opened 2021-11-05
Implement legacy_deserialize

6yozo:
Implement legacy_deserialize

### :memo: Checklist

### :memo: Tasks

- [x] Legacy_deserializer does not work with caf::message(table_slices:323)
- [x] The "schema.hpp" has serialize but is not inspectable. Either we have to make it inspectable or use has_serialize

File by file review


libvast/vast/detail/legacy_deserialize.hpp
```diff
@@ -119,6 +119,22 @@ class legacy_deserializer { // an inspector for CAF inspect
...
   }
 
+#if VAST_CLANG
+  // For whatever reason, clang treats long and unsigned long as separate
+  // types from int64_t and uint64_t respectively, so we need to add another
+  // two overloads.
+
+  result_type apply(unsigned long& x) {
+    static_assert(sizeof(x) == sizeof(uint64_t));
+    return apply(reinterpret_cast<uint64_t&>(x));
```
mavam, line 129:
```suggestion
    return apply(static_cast<uint64_t&>(x));
```

Shouldn't this be a static cast?

dominiklohmann: It can probably be one, yes.

I also still don't quite understand _why_ we need this, and I think we also may have to restrict the `#if` to `VAST_CLANG && VAST_MACOS`.

6yozo: I now compile with Clang 13, and it complains about "class member cannot be redeclared" so we need a more exact condition like `VAST_CLANG` && `VAST_MACOS`. The explanation of the problem can probably be inside stdint.h.

dominiklohmann: Okay, I think we can consider this one resolved.

libvast/src/index/address_index.cpp
```diff
@@ -3,16 +3,16 @@
 //   | |/ / __ |_\ \  / /          Across
 //   |___/_/ |_/___/ /_/       Space and Time
 //
-// SPDX-FileCopyrightText: (c) 2020 The VAST Contributors
+// SPDX-FileCopyrightText: (c) 2021 The VAST Contributors
```
dominiklohmann, line 6: The copyright date here is the file creation date, not the time it was last touched. Please revert.

dominiklohmann (changes requested):
Nice to see the binary deserializer being almost entirely removed. I noticed one small thing when glancing at the code.

libvast/src/index/address_index.cpp
6yozo: Ok. It will be reverted by next push.

libvast_test/src/table_slices.cpp
```diff
@@ -324,8 +319,9 @@ void table_slices::test_message_serialization() {
   auto sink = make_sink();
   CHECK_EQUAL(sink(slice1), caf::none);
   MESSAGE("load content for the second slice from the buffer");
-  auto source = make_source();
-  CHECK_EQUAL(source(slice2), caf::none);
+  // FIXME: legacy_deserializer does not handle caf::message
+  // CHECK_EQUAL(vast::detail::legacy_deserialize(buf, slice2), true);
+  return; // FIXME: Not necessary
```
dominiklohmann, line 324: Just remove this test; it's covered by other tests anyways and not very meaningful.

libvast/src/chunk.cpp
```diff
@@ -187,24 +188,29 @@ caf::error inspect(caf::serializer& sink, const chunk_ptr& x) {
...
-    return caf::none;
+    return true;
   }
   auto buffer = std::make_unique<chunk::value_type[]>(size);
   const auto data = buffer.get();
-  if (auto err = source.apply_raw(size, data)) {
+  if (!source.apply_raw(size, data)) {
     x = nullptr;
-    return caf::none;
+    return false; // TODO originally returned true
```
dominiklohmann, line 208: That was definitely a bug.

libvast/vast/schema.hpp
```diff
@@ -77,6 +78,7 @@ class schema : detail::equality_comparable<schema> {
 
   friend void serialize(caf::serializer& sink, const schema& sch);
   friend void serialize(caf::deserializer& source, schema& sch);
+  friend void serialize(detail::legacy_deserializer& source, schema& sch);
```
dominiklohmann, line 81: Rename to `inspect` so this works with the legacy deserialiser. `serialize` is very much outdated, and predated inspect.

libvast/vast/legacy_type.hpp
```diff
@@ -1105,6 +1111,46 @@ auto make_inspect(caf::detail::type_list<Ts...>) {
...
+    } else {
+      using reference = legacy_type&;
+      using fun = result_type (*)(Inspector&, reference);
+      static fun tbl[] = {make_inspect_fun<Inspector, Ts>()...};
+      if (x.type_tag != invalid_type_id)
+        return tbl[x.type_tag](f, x.x);
+      x.x = legacy_type{};
+    }
+  };
+}
```
tobim, line 1152: The same thing could be achieved without duplication by adding a helper lambda to the original function:
```c++
   constexpr auto ret = []{
     if constexpr (std::is_same_v<result_type, caf::error>)
       return caf::none;
     if constexpr (std::is_same_v<result_type, bool>)
       return true;
     if constexpr (std::is_same_v<result_type, void>)
       return;
   };
```
and `return ret()`.
I think I know WHY it wasn't done like that but it should be reasoned about with comment.

libvast/vast/fwd.hpp
```diff
@@ -149,6 +149,7 @@ namespace detail {
 template <class>
 class framed;
 
+class legacy_deserializer;
```
tobim, line 152: For consistency:
```suggestion
class legacy_deserializer;

```

tobim (changes requested):
Looking good in general, my comments are mostly about style.

libvast/src/value_index.cpp
```diff
@@ -117,7 +119,12 @@ caf::error inspect(caf::serializer& sink, const value_index& x) {
   return x.serialize(sink);
 }
 
-caf::error inspect(caf::deserializer& source, value_index& x) {
+caf::error inspect(caf::deserializer&, value_index&) {
+  return caf::make_error(ec::parse_error, "caf::deserializer inspection "
+                                          "instead of legacy_deserializer");
+}
```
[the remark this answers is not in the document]
dominiklohmann: Also, this is a bug: We don't want to touch all `caf::deserializer` usages, but rather only the `caf::binary_deserializer`.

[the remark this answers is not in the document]
6yozo: In the next commit these `inspect`s will be reverted back to their original version, and also `caf::error deserialize(caf::deserializer& `s.

libvast/src/chunk.cpp
6yozo: Resolved in commit c0311cfd71c165a71bc0e8c4f941fa938572e14d

libvast_test/src/table_slices.cpp
6yozo: Resolved in c0311cfd71c165a71bc0e8c4f941fa938572e14d

libvast/vast/schema.hpp
6yozo: Resolved in c0311cfd71c165a71bc0e8c4f941fa938572e14d

libvast/src/value_index.cpp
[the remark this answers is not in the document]
6yozo: Resolved in c0311cfd71c165a71bc0e8c4f941fa938572e14d

libvast/vast/legacy_type.hpp
6yozo: I believe only the bool version will survive the upgrade to CAF 0.18. 

@dominiklohmann What do you think?

dominiklohmann: I don't care either way. What Tobias suggested is quickly implemented, but I think it can also stay as-is. With CAF 0.18 only the `bool` version will survive as you said.

6yozo: What if I add a TODO comment to the other two version?

dominiklohmann: Works for me. @tobim what do you think?

tobim: I'm fine with it as long as it is explained with a comment.

libvast/src/chunk.cpp
```diff
@@ -199,14 +200,34 @@ caf::error inspect(caf::deserializer& source, chunk_ptr& x) {
...
+    return false;
+  if (size == 0) {
+    x = nullptr;
+    return true;
+  }
+  auto buffer = std::make_unique<chunk::value_type[]>(size);
+  const auto data = buffer.get();
+  if (!source.apply_raw(size, data)) {
+    x = nullptr;
+    return false; // TODO originally returned true
```
dominiklohmann, line 223: Please remove the comment.

libvast/src/system/type_registry.cpp
```diff
@@ -112,9 +112,8 @@ caf::error type_registry_state::load_from_disk() {
     auto buffer = io::read(fname);
     if (!buffer)
       return buffer.error();
-    caf::binary_deserializer source{self->system(), *buffer};
-    if (auto error = source(data))
-      return error;
+    if (!detail::legacy_deserialize(buffer.cvalue(), data))
+      return caf::make_error(ec::parse_error, "data deserialization failed");
```
dominiklohmann, line 116: If this errors the users get to see this message, so I suggest rewording like this:

```suggestion
      return caf::make_error(ec::parse_error, "failed to load type-registry state");
```

libvast/test/detail/legacy_deserialize.cpp
```diff
@@ -0,0 +1,346 @@
...
+using namespace std::chrono_literals;
+using namespace vast;
+
+template <class... Ts>
+  requires(!std::is_rvalue_reference_v<Ts> && ...)
+bool ldes(std::vector<char>& buf, Ts&... xs) {
+  return detail::legacy_deserialize(buf, xs...);
+}
+
+TEST(deserialize_string) {
```
dominiklohmann, line 41: The suite is already named `legacy_deserialize`, so please just name the test case `string`.

That allows for invoking `vast-test -s legacy_deserialize -t string` to run just this one test. The test case names only need to be unique within a single suite.

The same applies for all other test cases below.

libvast/test/detail/legacy_deserialize.cpp
```diff
@@ -0,0 +1,346 @@
...
+  std::vector<char> buf;
+  CHECK_EQUAL(detail::serialize(buf, i), caf::none);
+  auto i2 = ids{};
+  REQUIRE(ldes(buf, i2));
+  CHECK_EQUAL(i, i2);
+}
+
+auto to_addr_view(std::string_view str) {
+  return make_data_view(unbox(to<address>(str)));
+}
```
dominiklohmann, line 292: This has potential ODR issues, and needs to be wrapped in an anonymous namespace like so:


```suggestion
namespace {

auto to_addr_view(std::string_view str) {
  return make_data_view(unbox(to<address>(str)));
}

} // namespace
```

libvast/test/detail/legacy_deserialize.cpp
```diff
@@ -0,0 +1,346 @@
...
+    = legacy_string_type{}.attributes({{"synopsis", "bloomfilter(1,0.1)"}});
+  auto ss = factory<synopsis>::make(lst, caf::settings{});
+  REQUIRE_NOT_EQUAL(ss, nullptr);
+  ss->add(std::string_view{"192.168.0.1"});
+  std::vector<char> buf;
+  CHECK_EQUAL(detail::serialize(buf, ss), caf::none);
+  auto ss2 = factory<synopsis>::make(lst, caf::settings{});
+  REQUIRE(ldes(buf, ss2));
+  CHECK_EQUAL(*ss, *ss2);
+}
```
dominiklohmann, line 346: Not a request to change anything, just wanted to say thanks for writing this really extensive test suite. It's great to see! 🚀

libvast/test/synopsis.cpp
```diff
@@ -80,11 +79,14 @@ FIXTURE_SCOPE(synopsis_tests, fixtures::deterministic_actor_system)
 
 TEST(serialization) {
   factory<synopsis>::initialize();
+  // auto x = synopsis_ptr{};
+  // CHECK_EQUAL(roundtrip(x), x);
+
```
dominiklohmann, line 84: This is probably not supposed to be here, still.

libvast/test/synopsis.cpp
```diff
@@ -80,11 +79,14 @@ FIXTURE_SCOPE(synopsis_tests, fixtures::deterministic_actor_system)
...
   CHECK_ROUNDTRIP(synopsis_ptr{});
-  CHECK_ROUNDTRIP_DEREF(
-    factory<synopsis>::make(legacy_bool_type{}, caf::settings{}));
-  CHECK_ROUNDTRIP_DEREF(
-    factory<synopsis>::make(legacy_time_type{}, caf::settings{}));
+  // TODO CHECK_ROUNDTRIP_DEREF(
+  // TODO factory<synopsis>::make(legacy_bool_type{}, caf::settings{}));
+  // TODO CHECK_ROUNDTRIP_DEREF(
+  // TODO factory<synopsis>::make(legacy_time_type{}, caf::settings{}));
```
dominiklohmann, line 89: Why are these commented out?

dominiklohmann (commented):
I just have some minor things left. Once these are done I think we can merge this. 🚀

libvast/test/synopsis.cpp
6yozo: I have reenabled these checks. I was not sure that they work still, but they do.

libvast/test/synopsis.cpp
6yozo: They really need not be here. Removed them.

libvast/test/detail/legacy_deserialize.cpp
6yozo: Test cases will have a shorter name after next push.

libvast/vast/legacy_type.hpp
6yozo: I will add a comment to the functions to be removed by the next push.

libvast/test/detail/legacy_deserialize.cpp
6yozo: Writing test cases was the least I could do.

dominiklohmann (approved):
LGTM! :shipit:

PR #1957, opened 2021-11-07
Straighten the index actor signature

tobim:
The internal query handler of the index used to return void but
actually sent an unsafe response with the uuid, candidates and
scheduled partitions triple to the requester. The historical
reason for this was to prevent a race condition which happened when
results started flowing to the requester before it received the
initial response. This problem happens when the OS scheduler
suspends the thread between initiating the worker and returning
the triple.

This commit replaces the old workaround with a cleaner approach:
We now use a response promise and deliver it before initiating the
worker. This is enough to guarantee that the messages will arrive
in the expected order at the requester.

Thanks to that the query handler signature now correctly indicates
that the requester should expect a `<uuid, uint32_t, uint32_t>` as
response.

### :memo: Checklist

Verify that the message flow between index, requester and query_supervisor is unchanged.
Safe the unit test changes for last, the mock index implementations had to be refactored because the query handler is more strongly typed now.

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -217,8 +217,8 @@ partition_actor partition_factory::operator()(const uuid& id) const {
 
 // -- query_backlog ------------------------------------------------------------
 
-void query_backlog::emplace(
-  vast::query query, caf::typed_response_promise<uuid, uint32_t, uint32_t> rp) {
+void query_backlog::emplace(vast::query query,
+                            caf::typed_response_promise<query_cursor> rp) {
   auto& q = query.priority == query::priority::normal ? normal : low;
   // TODO: emplace does not work with libc++ <= 12.0. Switch to it once
   // we updated to LLVM 13.
@@ -909,7 +909,7 @@ index(index_actor::stateful_pointer<index_state> self,
       VAST_DEBUG("{} adds flush listener", *self);
       self->state.add_flush_listener(std::move(listener));
     },
-    [self](vast::query query) -> caf::result<uuid, uint32_t, uint32_t> {
+    [self](vast::query query) -> caf::result<query_cursor> {
       if (!self->state.accept_queries) {
         VAST_VERBOSE("{} delays query {} because it is still starting up",
                      *self, query);
@@ -921,12 +921,12 @@ index(index_actor::stateful_pointer<index_state> self,
                               std::move(query), std::move(*worker));
       }
       VAST_VERBOSE("{} pushes query {} to the backlog", *self, query);
-      auto rp = self->make_response_promise<uuid, uint32_t, uint32_t>();
+      auto rp = self->make_response_promise<query_cursor>();
       self->state.backlog.emplace(std::move(query), rp);
       return rp;
     },
-    [self](atom::internal, vast::query query, query_supervisor_actor worker)
-      -> caf::result<uuid, uint32_t, uint32_t> {
+    [self](atom::internal, vast::query query,
+           query_supervisor_actor worker) -> caf::result<query_cursor> {
       // Query handling
       auto sender = self->current_sender();
       auto client = caf::actor_cast<receiver_actor<atom::done>>(sender);
@@ -958,7 +958,7 @@ index(index_actor::stateful_pointer<index_state> self,
         candidates.push_back(self->state.active_partition.id);
       for (const auto& [id, _] : self->state.unpersisted)
         candidates.push_back(id);
-      auto rp = self->make_response_promise<uuid, uint32_t, uint32_t>();
+      auto rp = self->make_response_promise<query_cursor>();
       // Get all potentially matching partitions.
       auto start = std::chrono::steady_clock::now();
       self
@@ -978,7 +978,7 @@ index(index_actor::stateful_pointer<index_state> self,
               VAST_DEBUG("{} returns without result: no partitions qualify",
                          *self);
               self->send(self, atom::worker_v, worker);
-              rp.deliver(query_id, 0u, 0u);
+              rp.deliver(query_cursor{query_id, 0u, 0u});
               self->send(client, atom::done_v);
               return;
             }
@@ -993,7 +993,8 @@ index(index_actor::stateful_pointer<index_state> self,
             auto delta = std::chrono::steady_clock::now() - start;
             VAST_TRACEPOINT(query_meta_index, query_id.as_u64().first, total,
                             delta.count());
-            rp.deliver(query_id, detail::narrow<uint32_t>(total), scheduled);
+            rp.deliver(query_cursor{query_id, detail::narrow<uint32_t>(total),
+                                    scheduled});
             caf::send_as(client, static_cast<index_actor>(self), query_id,
                          scheduled);
           },
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -1016,7 +1016,8 @@ index(index_actor::stateful_pointer<index_state> self,
       auto iter = self->state.pending.find(query_id);
       if (iter == self->state.pending.end()) {
         VAST_WARN("{} drops query for unknown query id {}", *self, query_id);
-        return caf::make_error(ec::lookup_error, "unknown query id");
+        return caf::make_error(ec::lookup_error,
+                               fmt::format("unknown query id: {}", query_id));
       }
       auto& query_state = iter->second;
       if (!sender) {
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -48,6 +48,7 @@
 
 #include <caf/error.hpp>
 #include <caf/response_promise.hpp>
+#include <caf/send.hpp>
 #include <flatbuffers/flatbuffers.h>
 
 #include <algorithm>
@@ -216,8 +217,8 @@ partition_actor partition_factory::operator()(const uuid& id) const {
 
 // -- query_backlog ------------------------------------------------------------
 
-void query_backlog::emplace(vast::query query,
-                            caf::typed_response_promise<void> rp) {
+void query_backlog::emplace(
+  vast::query query, caf::typed_response_promise<uuid, uint32_t, uint32_t> rp) {
   auto& q = query.priority == query::priority::normal ? normal : low;
   // TODO: emplace does not work with libc++ <= 12.0. Switch to it once
   // we updated to LLVM 13.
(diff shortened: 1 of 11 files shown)


libvast/src/system/index.cpp
```diff
@@ -994,8 +977,9 @@ index(index_actor::stateful_pointer<index_state> self,
             auto delta = std::chrono::steady_clock::now() - start;
             VAST_TRACEPOINT(query_meta_index, query_id.as_u64().first, total,
                             delta.count());
-            respond(query_id, detail::narrow<uint32_t>(total), scheduled);
-            rp.delegate(static_cast<index_actor>(self), query_id, scheduled);
+            rp.deliver(query_id, detail::narrow<uint32_t>(total), scheduled);
+            caf::send_as(client, static_cast<index_actor>(self), query_id,
+                         scheduled);
```
dominiklohmann, line 982: Not sure I like this better than the old approach: This is a hand-rolled delegate, except that it does not keep the message ID and the priority.

Can't you create a second response promise before this request-then you're in which you return from the handler, and delegate that here?

libvast/src/system/index.cpp
```diff
@@ -1020,7 +1004,8 @@ index(index_actor::stateful_pointer<index_state> self,
       if (iter == self->state.pending.end()) {
         VAST_WARN("{} drops query for unknown query id {}", *self, query_id);
         self->send(client, atom::done_v);
-        return {};
+        return caf::make_error(ec::lookup_error, "unknown query id");
```
dominiklohmann, line 1007: We're returning an error to the client, so why are we sending `done` as well?

In general, I think sending `done` manually and not as part of a reply is the biggest structural issue with this code.

libvast/src/system/index.cpp
```diff
@@ -904,29 +905,20 @@ index(index_actor::stateful_pointer<index_state> self,
...
       VAST_VERBOSE("{} pushes query {} to the backlog", *self, query);
-      auto rp = self->make_response_promise<void>();
+      auto rp = self->make_response_promise<uuid, uint32_t, uint32_t>();
       self->state.backlog.emplace(std::move(query), rp);
       return rp;
     },
-    [self](atom::internal, vast::query query,
-           query_supervisor_actor worker) -> caf::result<void> {
+    [self](atom::internal, vast::query query, query_supervisor_actor worker)
+      -> caf::result<uuid, uint32_t, uint32_t> {
```
dominiklohmann, line 913: Can you add a struct for this besides the `query` struct?

E.g.,

```cpp
struct query_response {
  uuid id;
  uint32_t num_partitions;
  uint32_t num_scheduled;
};
```

libvast/src/system/index.cpp
```diff
@@ -976,11 +961,9 @@ index(index_actor::stateful_pointer<index_state> self,
...
               VAST_DEBUG("{} returns without result: no partitions qualify",
                          *self);
-              no_result();
-              // TODO: When updating to CAF 0.18, remove the use of the
-              // untyped response promise and call deliver without arguments.
-              auto& untyped_rp = static_cast<caf::response_promise&>(rp);
-              untyped_rp.deliver(caf::unit);
+              self->send(self, atom::worker_v, worker);
+              rp.deliver(query_id, 0u, 0u);
+              caf::anon_send(client, atom::done_v);
```
dominiklohmann, line 966: Why hide the index as the sender?

dominiklohmann (commented):
Good idea to refactor the message flow here. I'm pretty sure your changes are correct, but I have some thoughts on them.

libvast/src/system/index.cpp
mavam: I think a struct would help, but it feels more like a `cursor` or `handle` type object, because `response` suggests that it contains an answer to the query.

dominiklohmann: I like cursor or progress.

mavam: For "progress", a bit more information is missing; "status" is probably most neutral.

dominiklohmann: Status is an already used term that I don't want to overload, so let's go with cursor.

libvast/src/system/index.cpp
tobim: That doesn't work. I already have to return the other `rp` from the outer handler. I tried without returning the extra promise but it the response doesn't arrive at the exporter.

libvast/src/system/index.cpp
tobim: We probably don't have to send a done in this case.

The tricky part of this code is that the `query` handler essentially returns 2 messages, the triple and later the `done` for the first batch. The alternative would be to change the protocol so that the client must send an extra message to request the first batch. Tbh I think it's good that we're avoiding that extra roundtrip, and besides, this PR is certainly not the place to change the protocol.

libvast/src/system/index.cpp
tobim: :slightly_frowning_face: I actually like the fact that this triple does not have a name. I'm reluctant to introduce such a struct here.

libvast/src/system/index.cpp
tobim: I just needed a way to send from a typed actor to `client`. We could also make `client` a typed `receiver_actor<atom::done>`. Another approach would be to skip the `done` if no partitions qualify, but that is a protocol change again and the `query_processor` state machine would have to be adapted.

libvast/src/system/index.cpp
mavam: > I actually like the fact that this triple does not have a name

Can you elaborate on why? When reading the code for the first time, how would you infer the semantics of this triple? The messaging API is the public interface and should be possible to understand *without* looking at the implementation.

tobim: My issue is that this collection of values is very ad-hoc, it doesn't make sense to have it in any other context.

How would you feel about using a more complete comment at the interface declaration in `actors.hpp`? Something ala:
```c++
  // Initiates evaluation of a query.
  // @param query The query to evaluate.
  // @returns <evaluation_handle: uuid, candidate partitions: uint32_t, scheduled partitions: uint32_t>.
  //           evaluation_handle can be used to schedule additional partitions from the candidate set.
  //           candidate_partitions is the number of partitions that qualify for the query.
  //           scheduled_partitions is the number of partitions in the initial evaluation batch.
  caf::replies_to<query>::with<uuid, uint32_t, uint32_t>,
```

mavam: I'd be happy with that. As long as a reader doesn't have to read code and can infer the semantics from proper typing (ideally) or documentation (acceptable), I think we're doing the right thing.

dominiklohmann: Why not just introduce a struct like this, so we can _go-to definition_ at both the sending and the receiving site instead of manually needing to go to the definition in `vast/system/actors.hpp`, finding the correct handler, and then hoping that the documentation exists there? Documentation in that header is also much easier to get out of sync than documentation for a given type.

```cpp
/// Information returned by the index for ongoing queries to allow for pagination.
struct query_cursor {
  /// The handle of the query; needed to schedule additional partitions from the candidate set.
  uuid query_id = {};

  /// The number of partitions that qualify for the query.
  uint32_t num_candidate_partitions = {};

  /// The number of partitions scheduled for the query.
  uint32_t num_scheduled_partitions = {};

  auto inspect(auto& f, query_cursor& x) {
    return f(x.query_id, x.num_candidate_partitions, x.num_scheduled_partitions);
  }
};
```

I really didn't wanna make a fuss out of this, and mostly posted the comment because I thought we'd all agree that naming things is always good. Kinda surprised—I tend to always prefer naming things over using tuples.

libvast/src/system/index.cpp
dominiklohmann: Makes sense not to do such a thing in this PR.

libvast/src/system/index.cpp
dominiklohmann: You're right.

libvast/src/system/index.cpp
```diff
@@ -1010,8 +993,10 @@ index(index_actor::stateful_pointer<index_state> self,
             auto delta = std::chrono::steady_clock::now() - start;
             VAST_TRACEPOINT(query_meta_index, query_id.as_u64().first, total,
                             delta.count());
-            respond(query_id, detail::narrow<uint32_t>(total), scheduled);
-            rp.delegate(static_cast<index_actor>(self), query_id, scheduled);
+            rp.deliver(query_cursor{query_id, detail::narrow<uint32_t>(total),
+                                    scheduled});
+            caf::send_as(client, static_cast<index_actor>(self), query_id,
```
dominiklohmann, line 998: Please add a comment above explaining to the future reader why we're not sending as the index but are spoofing the client instead.

libvast/src/system/index.cpp
```diff
@@ -1028,9 +1013,7 @@ index(index_actor::stateful_pointer<index_state> self,
       auto iter = self->state.pending.find(query_id);
       if (iter == self->state.pending.end()) {
         VAST_WARN("{} drops query for unknown query id {}", *self, query_id);
-        if (client)
-          self->send(client, atom::done_v);
-        return {};
+        return caf::make_error(ec::lookup_error, "unknown query id");
```
dominiklohmann, line 1016:
```suggestion
        return caf::make_error(ec::lookup_error, fmt::format("unknown query id: {}", query_id));
```

libvast/test/system/query_processor.cpp
```diff
@@ -159,7 +162,11 @@ TEST(state transitions) {
   expect((atom::done), from(index).to(aut));
   CHECK_EQUAL(mock_ref().log, expected_log);
   CHECK_EQUAL(mock_ref().results, unsigned{2 + 3 + 6 + 12 + 24});
+  MESSAGE("hi");
   CHECK_EQUAL(mock_ref().state(), system::query_processor::idle);
+  MESSAGE("ho");
+  // self->send_exit(index, ec::no_error);
+  // sched.run();
```
dominiklohmann, line 169: This needs some cleanup.

dominiklohmann (approved):
Just some minor comments. Feel free to merge after addressing them. I think this is a good improvement overall for readability.

libvast/src/system/index.cpp
```diff
@@ -1010,8 +993,13 @@ index(index_actor::stateful_pointer<index_state> self,
             auto delta = std::chrono::steady_clock::now() - start;
             VAST_TRACEPOINT(query_meta_index, query_id.as_u64().first, total,
                             delta.count());
-            respond(query_id, detail::narrow<uint32_t>(total), scheduled);
-            rp.delegate(static_cast<index_actor>(self), query_id, scheduled);
+            rp.deliver(query_cursor{query_id, detail::narrow<uint32_t>(total),
+                                    scheduled});
+            // We "delegate" the first continuation back to self by spoofing
+            // the client as source. This is done so the response gets devliverd
+            // to the correct receipient - the client.
```
dominiklohmann, line 1000:
```suggestion
            // the client as source. This is done so the response gets delivered
            // to the correct recipient: the client.
```

dominiklohmann:
The style check failed anyways, so please excuse my last minute spelling fix request.

PR #1959, opened 2021-11-12
Load static plugins only when enabled

dominiklohmann:
We noticed that for our static builds with plugins it was impossible to disable the plugins, which is a regression and differs from the documentation that says that static plugins must be explicitly enabled. This change makes it so static plugins work as expected again.

### :memo: Checklist

I've tried this locally with a mix of dynamic and static plugins and various configurations, including mixed `all`, `bundled`, and specifically enabled plugins. Either reproduce this locally, or ask me to show you in a video call.


tobim (approved):
I verified that the static binary build behaves as expected now.

PR #1962, opened 2021-11-16
Log the version string with the startup message

tobim:
In this PR we add the version string to the startup log message and add an additional info message when the index component is ready for queries.

### :memo: Checklist

By commit.

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -356,9 +356,11 @@ caf::error index_state::load_from_disk() {
                 std::exchange(synopses, {}))
       .then(
         [this](atom::ok) {
-          VAST_VERBOSE("{} successfully loaded meta index from disk and will "
-                       "start processing queries",
-                       *self);
+          VAST_DEBUG("{} successfully loaded meta index from disk and will "
+                     "start processing queries",
+                     *self);
+          VAST_INFO("{} finished initalizing and is ready to accept queries",
+                    *self);
           this->accept_queries = true;
         },
         [this](caf::error& err) {
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -356,9 +356,6 @@ caf::error index_state::load_from_disk() {
                 std::exchange(synopses, {}))
       .then(
         [this](atom::ok) {
-          VAST_DEBUG("{} successfully loaded meta index from disk and will "
-                     "start processing queries",
-                     *self);
           VAST_INFO("{} finished initalizing and is ready to accept queries",
                     *self);
           this->accept_queries = true;
--- a/libvast/src/system/start_command.cpp
+++ b/libvast/src/system/start_command.cpp
@@ -87,7 +87,7 @@ caf::message start_command(const invocation& inv, caf::actor_system& sys) {
   if (!bound_port)
     return caf::make_message(std::move(bound_port.error()));
   auto listen_addr = std::string{host} + ':' + std::to_string(*bound_port);
-  VAST_INFO("vast-{} is listening on {}", version::version, listen_addr);
+  VAST_INFO("VAST ({}) is listening on {}", version::version, listen_addr);
   // Start signal monitor.
   std::thread sig_mon_thread;
   auto guard = system::signal_monitor::run_guarded(
--- a/libvast/src/system/start_command.cpp
+++ b/libvast/src/system/start_command.cpp
@@ -87,7 +87,7 @@ caf::message start_command(const invocation& inv, caf::actor_system& sys) {
   if (!bound_port)
     return caf::make_message(std::move(bound_port.error()));
   auto listen_addr = std::string{host} + ':' + std::to_string(*bound_port);
-  VAST_INFO("VAST node is listening on {}", listen_addr);
+  VAST_INFO("vast-{} is listening on {}", version::version, listen_addr);
   // Start signal monitor.
   std::thread sig_mon_thread;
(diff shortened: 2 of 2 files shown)


libvast/src/system/start_command.cpp
```diff
@@ -87,7 +87,7 @@ caf::message start_command(const invocation& inv, caf::actor_system& sys) {
   if (!bound_port)
     return caf::make_message(std::move(bound_port.error()));
   auto listen_addr = std::string{host} + ':' + std::to_string(*bound_port);
-  VAST_INFO("VAST node is listening on {}", listen_addr);
+  VAST_INFO("vast-{} is listening on {}", version::version, listen_addr);
```
dominiklohmann, line 90:
```suggestion
  VAST_INFO("VAST ({}) is listening on {}", version::version, listen_addr);
```

libvast/src/system/index.cpp
```diff
@@ -356,9 +356,11 @@ caf::error index_state::load_from_disk() {
                 std::exchange(synopses, {}))
       .then(
         [this](atom::ok) {
-          VAST_VERBOSE("{} successfully loaded meta index from disk and will "
-                       "start processing queries",
-                       *self);
+          VAST_DEBUG("{} successfully loaded meta index from disk and will "
+                     "start processing queries",
+                     *self);
+          VAST_INFO("{} finished initalizing and is ready to accept queries",
```
lava, line 362: If we have an INFO message, I don't think we still need a DEBUG message in the same place.

What's the general idea here, is it to be able to see the message from `systemctl status`?

libvast/src/arrow_table_slice.cpp
```diff
@@ -459,9 +461,20 @@ class row_picker {
...
+      if (field->length() < struct_length || field->length() < row)
+        return false;
+    }
+    return true;
+  }
+
   void operator()(const arrow::StructArray& arr, const legacy_record_type& t) {
     if (arr.IsNull(row_))
       return;
+    if (!valid(arr, row_))
```
lava, line 476: Shouldn't we print an error here, like `database contains corrupted data for id range [xxxx - yyy)`?

Also, is this only in the export path? (Ie. should we be concerned about performance for this check?)

dominiklohmann: Wouldn't it make more sense to make this an assertion rather than trying to verify data here? I think doing data integrity checks should not be part of exporting.

libvast/src/system/index.cpp
tobim: I didn't delete the old message because it contains more information but of course the new one implies the same for those in the know. The new message is worded differently to avoid confusing regular users.

I have now problem with removing the debug message.

libvast/src/arrow_table_slice.cpp
tobim: Do we have a mechanism to rate-limit such error logs? I fear this can become spammy on bulk-exports otherwise.

In general this affects every `at()` into such slices, I think the check is in the right place. We should definitely check the performance of this. In principle we could do test for the error condition during table slice materialization and set a flag in the hidden state instead.

tobim: > Wouldn't it make more sense to make this an assertion rather than trying to verify data here? I think doing data integrity checks should not be part of exporting.

The reality is that databases with incorrect table slices exist. This addition prevents segfaults when they get accessed, an assertion would not help in this case.

dominiklohmann: Don't they only exist for users that used a build that was never released (i.e., from 2-3 weeks ago)?

tobim: > Don't they only exist for users that used a build that was never released (i.e., from 2-3 weeks ago)?

Only if you assume that nobody created their own schemas with `list<record>` or `map<record>`.

dominiklohmann: For measuring in the hot path on export:

```diff
diff --git a/libvast/src/table_slice.cpp b/libvast/src/table_slice.cpp
index 8cf19bfba..cbac8d110 100644
--- a/libvast/src/table_slice.cpp
+++ b/libvast/src/table_slice.cpp
@@ -706,6 +706,7 @@ filter(const table_slice& slice, expression expr, const ids& hints) {
       result.emplace_back(field.type);
     return result;
   }();
+  auto start = std::chrono::steady_clock::now();
   for (auto id : select(selection)) {
     VAST_ASSERT(id >= offset);
     auto row = id - offset;
@@ -718,6 +719,12 @@ filter(const table_slice& slice, expression expr, const ids& hints) {
       }
     }
   }
+  auto stop = std::chrono::steady_clock::now();
+  auto elapsed
+    = std::chrono::duration_cast<std::chrono::duration<double, std::milli>>(
+      stop - start);
+  VAST_ERROR("filter {}; n = {}, elapsed = {}, elapsed / n = {}", __LINE__,
+             rank(selection), elapsed, elapsed / rank(selection));
   if (builder->rows() == 0)
     return std::nullopt;
   if (builder->rows() == slice.rows())
```

dominiklohmann (approved):
Please remove the duplicate log message again, otherwise LGTM.

libvast/src/arrow_table_slice.cpp
dominiklohmann: This turned out to be a costly check (worst case ~8% slower), so we decided to remove the workaround as discussed above.

PR #1966, opened 2021-11-18
Avoid evaluating calls when assertions are disabled

tobim:
Previous to this change the statement `VAST_ASSERT(expr)` would still evaluate `expr` unless the optimizer was able to eliminate the code after inlining.

Of course it is better not to rely on that mechanism, and indeed it sometimes fails even with LTO turned on.

In a simple ingest test on my T14s the suricata parser performance increased by about 16% and exporting with a highly selective query performed a whopping 4 times faster.

### :memo: Checklist

Trivial.

### DIFF
--- a/libvast/vast/detail/assert.hpp
+++ b/libvast/vast/detail/assert.hpp
@@ -33,11 +33,11 @@
 
 #  define VAST_ASSERT_2(expr, msg)                                             \
     do {                                                                       \
-      static_cast<void>(expr);                                                 \
-      static_cast<void>(msg);                                                  \
+      static_cast<void>(sizeof(expr));                                         \
+      static_cast<void>(sizeof(msg));                                          \
     } while (false)
 
-#  define VAST_ASSERT_1(expr) static_cast<void>(expr)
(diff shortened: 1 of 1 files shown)


libvast/vast/detail/assert.hpp
```diff
@@ -33,11 +33,11 @@
...
 #  define VAST_ASSERT_2(expr, msg)                                             \
     do {                                                                       \
-      static_cast<void>(expr);                                                 \
-      static_cast<void>(msg);                                                  \
+      static_cast<void>(sizeof(expr));                                         \
+      static_cast<void>(sizeof(msg));                                          \
     } while (false)
 
-#  define VAST_ASSERT_1(expr) static_cast<void>(expr)
+#  define VAST_ASSERT_1(expr) static_cast<void>(sizeof(expr))
```
dominiklohmann, line 40:
```suggestion
#  define VAST_ASSERT_1(expr) static_cast<void>(sizeof(decltype(expr)))
```

This suppresses some warnings from clang-tidy for reasons I absolutely do not understand. It's necessary above as well.

dominiklohmann (approved):
Obviously fine. Let's merge this soon after the release.

PR #1970, opened 2021-11-19
Make use of CMake 3.22 policies

dominiklohmann:
See CMake changelog and release notes at https://blog.kitware.com/cmake-3-22-0-available-for-download/

Nothing major that affects us, and no policy changes that affect our build either in any of the five build configurations I've tested this with (Debug, RelWithDebInfo, Release, MinSizeRel, CI).

### :memo: Checklist

Don't think this needs a thorough review—it's a pretty simple maintenance change.

This doesn't need a changelog entry either, as it does not raise the minimum version, but rather only the policy opt-in version.


6yozo (approved):
CMake logs in CI show no problem.

tobim (approved):
I'll trust you on this one.

PR #1971, opened 2021-11-19
Remove type name hashing

dominiklohmann:
The legacy type system was the only place that depended on the hashes of the type name annotation in inspect overloads to be included in the hash, and now that we don't have that anymore we can remove it. This should result in a small speedup of hash functions using the hash inspector and make the branch predictor be in a happier place.

### :memo: Checklist

This builds on top of #1888. Really small PR, just need to verify manually that old databases still work.

### DIFF
--- a/libvast/vast/hash/hash_append.hpp
+++ b/libvast/vast/hash/hash_append.hpp
@@ -15,7 +15,6 @@
 
 #include <caf/detail/type_traits.hpp>
 #include <caf/meta/save_callback.hpp>
-#include <caf/meta/type_name.hpp>
 
 #include <array>
 #include <chrono>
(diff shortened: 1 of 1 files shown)


mavam (approved):
This looks innocuous at the local scope. But I wonder about the non-trivial places where this has subtle implications. Have you looked at locations where we hash inspectable types, e.g., as keys in unordered maps?

dominiklohmann:
> Have you looked at locations where we hash inspectable types, e.g., as keys in unordered maps?

Let me summarize the changes relevant for type hashing. Given `legacy_bool_type lbt{}`, `bool_type bt{}`, `legacy_count_type lct{}`, and `count_type ct{}`.

||`hash(lbt) == hash(lct)`|`hash(bt) == hash(ct)`|
|-:|:-:|:-:|
|**Before**|`false`|`false`|
|**After**|`true`|`false`|

Because #1888 removed all uses of the legacy type, including all uses of legacy types in unordered containers, this will not be an issue. We do not persist hashes of types anywhere, otherwise #1888 would've been a breaking change.

I'll wait with merging this until after #1888 is in because I'm still waiting for its review to be completed.

dominiklohmann:
@mavam I double checked which types we use with the hash inspector by simply printing the type name in it, and none of the types are affected by this like `legacy_type` would've been. So we're definitely on the safe side with this change.

PR #1972, opened 2021-11-20
Implement a wrapper for FlatBuffers table pointers

dominiklohmann:
This should help simplify our code a fair bit down the line: A wrapper around a FlatBuffers table pointer that shares the reference count of the chunk it is constructed from, and is thus easier to handle via RAII. A `vast::flatbuffer<T>`  can optionally be constructed from a `flatbuffers::DetachedBuffer&&`, allowing the developer to skip handling the chunk entirely. The major advantage of this compared to using a `vast::chunk` directly is that it is safe to slice a non-root table from a `vast::flatbuffer`, which helps reduce unnecessary pointer chasing.

### :memo: Checklist

Likely best reviewed by @lava, with whom I just recently discussed this in Slack. 

Note that this is a small Saturday ~morning~ project, so ~I've not bothered adding test cases~ test cases are minimal so far. I will add some if the general design is agreed upon. Overall I think that this is something that we could really make a lot of use of, and that will make working with FlatBuffers easier in the future.

### DIFF
--- a/libvast/src/segment.cpp
+++ b/libvast/src/segment.cpp
@@ -34,26 +34,17 @@ namespace vast {
 
 using namespace binary_byte_literals;
 
-caf::expected<segment> segment::make(chunk_ptr chunk) {
-  VAST_ASSERT(chunk != nullptr);
-  // FlatBuffers <= 1.11 does not correctly use '::flatbuffers::soffset_t' over
-  // 'soffset_t' in FLATBUFFERS_MAX_BUFFER_SIZE.
-  using ::flatbuffers::soffset_t;
-  if (chunk->size() >= FLATBUFFERS_MAX_BUFFER_SIZE)
-    return caf::make_error(ec::format_error,
-                           "cannot read segment because its size",
-                           chunk->size(), "exceeds the maximum allowed size of",
-                           FLATBUFFERS_MAX_BUFFER_SIZE);
-  auto s = fbs::GetSegment(chunk->data());
-  VAST_ASSERT(s); // `GetSegment` is just a cast, so this cant become null.
-  if (s->segment_type() != fbs::segment::Segment::v0)
+caf::expected<segment> segment::make(const chunk_ptr& chunk) {
+  auto s = flatbuffer<fbs::Segment>::make(chunk);
+  if (!s)
+    return s.error();
+  if ((*s)->segment_type() != fbs::segment::Segment::v0)
     return caf::make_error(ec::format_error, "unsupported segment version");
-  return segment{std::move(chunk)};
+  return segment{std::move(*s)};
 }
 
 uuid segment::id() const {
-  auto segment = fbs::GetSegment(chunk_->data());
-  auto segment_v0 = segment->segment_as_v0();
+  auto segment_v0 = flatbuffer_->segment_as_v0();
   uuid result;
   if (auto error = unpack(*segment_v0->uuid(), result))
     VAST_ERROR("couldnt get uuid from segment: {}", error);
@@ -62,8 +53,7 @@ uuid segment::id() const {
 
 vast::ids segment::ids() const {
   vast::ids result;
-  auto segment = fbs::GetSegment(chunk_->data());
-  auto segment_v0 = segment->segment_as_v0();
+  auto segment_v0 = flatbuffer_->segment_as_v0();
   for (auto interval : *segment_v0->ids()) {
     result.append_bits(false, interval->begin() - result.size());
     result.append_bits(true, interval->end() - interval->begin());
@@ -72,19 +62,18 @@ vast::ids segment::ids() const {
 }
 
 size_t segment::num_slices() const {
-  auto segment = fbs::GetSegment(chunk_->data());
-  auto segment_v0 = segment->segment_as_v0();
+  auto segment_v0 = flatbuffer_->segment_as_v0();
   return segment_v0->slices()->size();
 }
 
 chunk_ptr segment::chunk() const {
-  return chunk_;
+  return flatbuffer_.chunk();
 }
 
 caf::expected<std::vector<table_slice>>
 segment::lookup(const vast::ids& xs) const {
   std::vector<table_slice> result;
-  auto segment = fbs::GetSegment(chunk_->data())->segment_as_v0();
+  auto segment = flatbuffer_->segment_as_v0();
   if (!segment)
     return caf::make_error(ec::format_error, "invalid segment version");
   VAST_ASSERT(segment->ids()->size() == segment->slices()->size());
@@ -94,7 +83,8 @@ segment::lookup(const vast::ids& xs) const {
   };
   auto g = [&](const auto& zip) {
     auto&& [interval, flat_slice] = zip;
-    auto slice = table_slice{*flat_slice, chunk_, table_slice::verify::yes};
+    // TODO: rework livetime sharing API of table slice.
+    auto slice = table_slice{*flat_slice, chunk(), table_slice::verify::yes};
     slice.offset(interval->begin());
     VAST_ASSERT(slice.offset() == interval->begin());
     VAST_ASSERT(slice.offset() + slice.rows() == interval->end());
@@ -120,7 +110,7 @@ segment::lookup(const vast::ids& xs) const {
 
 caf::expected<std::vector<table_slice>>
 segment::erase(const vast::ids& xs) const {
-  const auto* segment = fbs::GetSegment(chunk_->data())->segment_as_v0();
+  const auto* segment = flatbuffer_->segment_as_v0();
   auto intervals = std::vector(segment->ids()->begin(), segment->ids()->end());
   auto flat_slices
     = std::vector(segment->slices()->begin(), segment->slices()->end());
@@ -133,7 +123,7 @@ segment::erase(const vast::ids& xs) const {
   std::vector<table_slice> result;
   for (unsigned int i = 0; i < segment->slices()->size(); ++i) {
     const auto& flat_slice = segment->slices()->Get(i);
-    auto slice = table_slice{*flat_slice, chunk_, table_slice::verify::yes};
+    auto slice = table_slice{*flat_slice, chunk(), table_slice::verify::yes};
     slice.offset(intervals.at(i)->begin());
     // Expand keep_mask on-the-fly if needed.
     auto max_id = slice.offset() + slice.rows();
@@ -144,7 +134,8 @@ segment::erase(const vast::ids& xs) const {
   return result;
 }
 
-segment::segment(chunk_ptr chk) : chunk_{std::move(chk)} {
+segment::segment(flatbuffer<fbs::Segment> flatbuffer)
+  : flatbuffer_{std::move(flatbuffer)} {
   // nop
 }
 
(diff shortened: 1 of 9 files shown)


libvast/vast/flatbuffer.hpp
```diff
@@ -0,0 +1,350 @@
...
+  static constexpr auto verify_default = verify::no;
+#endif // VAST_ENABLE_ASSERTIONS
+
+  // -- constructors, destructors, and assignment operators -------------------
+
+  /// Constructs a ref-counted FlatBuffers root table that shares the
+  /// lifetime with the chunk it's constructed from.
+  /// @pre *chunk* must hold a valid *Table*.
+  [[nodiscard]] static caf::expected<flatbuffer>
+  make(const chunk_ptr& chunk, enum verify verify = verify_default) noexcept
```
lava, line 55: For root tables, would it make sense to use `chunk_ptr&&` as argument? Or is there a use case where we have to share the chunk with something else?

libvast/vast/flatbuffer.hpp
```diff
@@ -0,0 +1,350 @@
...
+
+  /// Default-constructs a FlatBuffers table.
+  flatbuffer() noexcept = default;
+
+  /// Constructs a ref-counted FlatBuffers root table from a FlatBufferBuilder
+  /// by finishing it.
+  // TODO: FlatBuffers 2.0's reflection allows for inferring the
+  // *file_identifier* from *Table*, which makes this API a lot easier to use.
+  // When upgrading to that version from 1.12 we can enable reflection when
+  // compiling the schemas.
```
lava, line 109: In a call we discussed instead introducing a `flatbuffer_traits<T>::file_identifier` map for all our root types, since the correct file identifier is already determined by `Table`, even though flatbuffers 1.12 lacks the API to figure it out.

libvast/vast/flatbuffer.hpp
```diff
@@ -0,0 +1,350 @@
...
+
+#include <flatbuffers/flatbuffers.h>
+#include <fmt/core.h>
+
+#include <compare>
+
+namespace vast {
+
+/// A wrapper class around a FlatBuffers table that allows for sharing the
+/// lifetime with the chunk containing the table.
```
lava, line 23: Please mention that it is reference-counted and copyable.

I also wonder if it would make sense to make a separate `root_flatbuffer` class instead of the boolean template argument, but that probably goes too far for this PR.

libvast/vast/flatbuffer.hpp
```diff
@@ -0,0 +1,350 @@
...
+      const auto lhs_bytes = as_bytes(lhs);
+      const auto rhs_bytes = as_bytes(rhs);
+      return std::equal(lhs_bytes.begin(), lhs_bytes.end(), rhs_bytes.begin(),
+                        rhs_bytes.end());
+    }
+    return static_cast<bool>(lhs) == static_cast<bool>(rhs);
+  }
+
+  friend std::strong_ordering
+  operator<=>(flatbuffer lhs, flatbuffer rhs) noexcept requires(IsRootTable) {
```
lava, line 200: I'm not sure I understand the semantics of a flatbuffer being "bigger" or "smaller" than another? Is this required somewhere? Also, I think this will lead to counter-intuitive results if you have e.g.

```
table Integer {
  value: uint32;
}
```

since the ordering would not follow the numerical ordering of value.

libvast/vast/flatbuffer.hpp
```diff
@@ -0,0 +1,350 @@
...
+    VAST_ASSERT(&child_table
+                == flatbuffers::GetRoot<ChildTable>(nested_flatbuffer.data()));
+    return flatbuffer<ChildTable, true>{
+      chunk_->slice(as_bytes(nested_flatbuffer))};
+  }
+
+  /// Accesses the underlying chunk.
+  /// @note This is only available for FlatBuffers root table pointers, as
+  /// this operation is fundamentally unsafe for non-root tables, for which
+  /// the table pointer is not at the beginning of the chunk.
```
lava, line 263: Hm, "fundamentally unsafe" sounds a bit strong, there are valid uses for getting the chunk in non-root tables (ie. for computing the memory consumption). But I agree that we can leave it restricted to root tables until we actually need one of these use cases to reduce accidents.

libvast/vast/flatbuffer.hpp
```diff
@@ -0,0 +1,350 @@
...
+
+  friend bool operator==(flatbuffer lhs, flatbuffer rhs) noexcept
+    requires(IsRootTable) {
+    if (&lhs == &rhs)
+      return true;
+    if (lhs && rhs) {
+      const auto lhs_bytes = as_bytes(lhs);
+      const auto rhs_bytes = as_bytes(rhs);
+      return std::equal(lhs_bytes.begin(), lhs_bytes.end(), rhs_bytes.begin(),
+                        rhs_bytes.end());
```
lava, line 194: I'm not sure that a user would expect byte equality here, the closest equivalent is `shared_ptr` which only does pointer equality.

libvast/src/segment.cpp
```diff
@@ -94,7 +83,8 @@ segment::lookup(const vast::ids& xs) const {
   };
   auto g = [&](const auto& zip) {
     auto&& [interval, flat_slice] = zip;
-    auto slice = table_slice{*flat_slice, chunk_, table_slice::verify::yes};
+    // TODO: rework livetime sharing API of table slice.
```
lava, line 86: `lifetime`

libvast/vast/segment.hpp
```diff
@@ -33,7 +34,7 @@ class segment {
   /// Constructs a segment.
   /// @param header The header of the segment.
   /// @param chunk The chunk holding the segment data.
-  static caf::expected<segment> make(chunk_ptr chunk);
+  static caf::expected<segment> make(const chunk_ptr& chunk);
```
lava, line 37: Is it possible to use `chunk_ptr&&` here and `flatbuffer<fbs::Segment>&&` below?

I general I'm wondering if it would be worthwhile to distinguish between owned and shared flatbuffers. I guess will come up when we want to add support to mutable flatbuffers.

Also, the comment mentions a non-existing parameter `header`.

lava (approved):
Looks great, thanks! I didn't personally test, but since this is a pure refactoring having green unit and integration tests should suffice.

libvast/vast/flatbuffer.hpp
dominiklohmann: I've tried this, and it's held back by the fact that this needs the definition of `T` available, which we don't want to include in this header. So I think it's best to leave this as-is for now.

libvast/vast/flatbuffer.hpp
dominiklohmann: I thought about having the `root_flatbuffer` as well, but it would really add a lot of code and as a developer I want the root variant to be the default, not the other way around. I think given that we almost always use `auto` we have an easy time using this in our code. We could probably add this alias for ease of use:

```cpp
template <class Table>
using child_flatbuffer = flatbuffer<Table, false>;
```

But I don't see the immediate win in adding that now. Future uses of this will show what we really need, and we can continually improve this I think.

libvast/vast/flatbuffer.hpp
dominiklohmann: We can maybe make this an opt-in thing via a third template argument that toggles the availability of "behaves like a `vector<byte>` for comparisons."

libvast/vast/flatbuffer.hpp
dominiklohmann: I'll tone down the comment a bit and explain the why a bit more.

libvast/vast/flatbuffer.hpp
dominiklohmann: I intentionally named this `flatbuffer<T>` rather than `flatbuffer_ptr<T>` since this supposed to be use like a shared buffer, not a shared pointer like type.

libvast/vast/segment.hpp
dominiklohmann: > distinguish between owned and shared flatbuffers.

`x.chunk().unique()` does exactly that.

libvast/vast/flatbuffer.hpp
dominiklohmann: I'll take a look at the call sites. At least we should be taking by value to allow the user to move.

libvast/vast/flatbuffer.hpp
dominiklohmann: >  Is this required somewhere

The new type system does the same thing internally because the types need to be sortable to be used in an unordered container. This makes it easy to use that way. I think we can make it an option, though.

libvast/vast/flatbuffer.hpp
```diff
@@ -19,17 +19,38 @@
...
+};
+
+/// Determines whether the FlatBuffers table has buffer- or pointer-like
+/// behavior, i.e., whether comparison operators compare the buffer or the table
+/// pointer.
+enum class flatbuffer_semantics {
+  buffer,  ///< Comparison operators compare the buffer contents bytewise. This
+           ///< is only supported for root tables.
+  pointer, ///< Comparison operators compare the table pointers.
+};
```
lava, line 37: It still feels to me like this complicates the interface for not too much gain (the caller could just call `memcmp()` on any chunk for a buffer-semantics comparison), but if you want to keep it in I won't object further.

It's also a bit weird that by inheriting the `semantics` from the parent, many flatbuffers will end up with the combination of `semantics == buffer && type == child`, which isn't really supported. Maybe it would even make sense to just make a `template<flatbuffer_semantics> bool compare()` function, so that the user has to explicitly select at the call site which semantics he wants.

dominiklohmann: Do we have any table other than `type` that we put into an unordered container as a key? For which tables does the buffer behavior make more sense than the pointer behavior?

Personally I prefer having the buffer behavior everywhere, but I can see where you're coming from. We could also introduce some form of `flatbuffer_base`, and then have a `flatbuffer` and a `flatbuffer_ptr` inherit from that and add the functionality specific to them.

lava: I'm not sure I understand the connection to unordered containers?

dominiklohmann: Take `type` for example. We often use it as a key in unordered containers, so they need to have a well-defined `operator<`. If two types have identical buffer contents, we would want them to compare equal, and that also needs to be reflected in the `operator<`.

I think exposing pointer semantics for the comparison operators is rarely what we want.

libvast/vast/flatbuffer.hpp
```diff
@@ -0,0 +1,389 @@
...
+
+/// A wrapper class around a FlatBuffers table that allows for sharing the
+/// lifetime with the chunk containing the table.
+/// @tparam Table The generated FlatBuffers table type.
+/// @tparam Semantics Determins whether the table behaves like a buffer or a
```
dominiklohmann, line 42: We want to get rid of this and also remove the comparison operators entirely, because we were unsure which is right, so ultimately this should end up being decided on a case-by-case basis in the classes making use of `flatbuffer<T>`.

libvast/vast/flatbuffer.hpp
dominiklohmann: We've decided to straight up remove the comparison operators because we couldn't agree on their semantics.

PR #1973, opened 2021-11-22
Emit version column along with metrics

6yozo:
Emit a metrics event when VAST is started

### :memo: Checklist

Review file by file.

### :memo: Tasks
- [x] Emit metrics on startup
- [x] Emit metrics on shutdown
- [x] Emit version number

### DIFF
--- a/changelog/unreleased/features/1973--version-metrics.md
+++ b/changelog/unreleased/features/1973--version-metrics.md
@@ -1,0 +1,2 @@
+All metrics events now contain the version of VAST. Additionally, VAST now emits
+startup and shutdown metrics at the start and stop of the VAST server.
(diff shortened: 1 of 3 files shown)


6yozo:
[story/sc-29464]

changelog/unreleased/features/1973--version-metrics.md
```diff
@@ -0,0 +1,2 @@
+VAST emits a version column along with metrics. VAST also emits two new metrics:
+startup and shutdown.
```
dominiklohmann, line 2: Please rewrite this in a way that is easier to digest for the user, e.g.,

> All metrics events now contain the version of VAST. Additionally, VAST now emits `startup` and `shutdown` metrics at start and stop of the VAST server.

dominiklohmann (approved):
I verified that this works as expected locally for both the self-sink and the file-sink. No need to test the other sinks as they use the same code path as the file-sink.

I've got just one rewording request for the changelog entry.

This also needs to be rebased now that #1888 is merged.

PR #1974, opened 2021-11-22
Make JSON field selectors configurable

### LINKED ISSUE #1943: Allow for using custom field selectors with the JSON import
### Description

The `import json` command currently matches field names from incoming JSONL data against the field names in record types in the known schemas. This process can optionally be skipped if `--type` is used to restrict the known schemas to a single type.

In practice this means that the `import json` command requires the field names to be always be an exact match to the schema. This is not well-suited for real-world usage.

For the `import suricata` and `import zeek-json` commands the type matching process works differently. They internally use the `field_selector`, which is configured like this:

```cpp
struct suricata_selector_specification {
  // The name of the reader in logs.
  static constexpr auto name = "suricata-reader";
  // The field that contains the event type.
  static constexpr auto field = "event_type";
  // The prefix of the event type.
  static constexpr auto prefix = "suricata";
};

using suricata_selector = field_selector<suricata_selector_specification>;
```

This causes the Suricata import to always find the type `suricata.alert` for data that contains the field `event_type: "alert"`.

We need to expose the field selector more generically instead of hardcoding it for the Suricata and Zeek Streaming JSON types, such that users can use it for other types of data as well.

### Definition of Done

The field selector is configurable at runtime rather than compile time, and the `import json` command has new options to select the event type field and the prefix. I suggest the following new option, with the default behavior being unchanged.

```bash
# exactly mirrors the behavior of `vast import suricata`
vast import json --field-selector=suricata:event_type
```

### Anything else?

This is a feature request from @satta.

dominiklohmann:
This adds a new `--selector=<field>[:<prefix>]` option to the JSON import. The *field* contains the *event name*, which when prefixed with the optional *prefix* is used for selecting the event type.

This previously existed as a hard-coded option for the Zeek Stremaing JSON and Suricata Eve JSON readers, which still exist unchanged.

The following commands are now equivalent:

```sh
# Import Suricata Eve JSON
vast import suricata
vast import json --selector=event_type:suricata

# Import Zeek Streaming JSON
vast import zeek-json
vast import json --selector=_path:zeek
```

### :memo: Checklist

File-by-file. I left some notes as review comments that are helpful to look at.

Note that this is based on #1888 to avoid merge conflicts.

### DIFF
--- a/libvast/vast/format/json.hpp
+++ b/libvast/vast/format/json.hpp
@@ -10,20 +10,15 @@
 
 #include "vast/fwd.hpp"
 
-#include "vast/detail/flat_map.hpp"
 #include "vast/detail/line_range.hpp"
-#include "vast/detail/string.hpp"
 #include "vast/error.hpp"
+#include "vast/format/json/selector.hpp"
 #include "vast/format/multi_layout_reader.hpp"
 #include "vast/format/ostream_writer.hpp"
-#include "vast/hash/hash_append.hpp"
-#include "vast/hash/xxhash.hpp"
-#include "vast/logger.hpp"
 #include "vast/schema.hpp"
 #include "vast/view.hpp"
 
 #include <caf/expected.hpp>
-#include <caf/fwd.hpp>
 #include <caf/settings.hpp>
 
 #include <chrono>
@@ -55,7 +50,6 @@ private:
 
 /// A reader for JSON data. It operates with a *selector* to determine the
 /// mapping of JSON object to the appropriate record type in the schema.
-template <class Selector>
 class reader final : public multi_layout_reader {
 public:
   using super = multi_layout_reader;
@@ -84,7 +78,9 @@ protected:
 private:
   using iterator_type = std::string_view::const_iterator;
 
-  Selector selector_;
+  std::unique_ptr<selector> selector_;
+  std::string reader_name_ = "json-reader";
+
   std::unique_ptr<std::istream> input_;
 
   // https://simdjson.org/api/0.7.0/classsimdjson_1_1dom_1_1parser.html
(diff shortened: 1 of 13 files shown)


libvast/vast/format/json/field_selector.hpp
```diff
@@ -36,44 +44,49 @@ struct field_selector {
...
+                  detail::pretty_type_name(this), field_name_, field);
       return {};
     }
     return it->second;
   }
 
-  caf::error schema(const vast::schema& s) {
+  inline caf::error schema(const vast::schema& s) override {
     for (const auto& t : s) {
-      auto sn = detail::split(t.name(), ".");
```
dominiklohmann, line 47: Note for the reviewer: I fixed a bug where this was broken for event types containing more than one `.` in the name, e.g., where the type prefix is `foo.bar`, the field at the given field name contains `bad` and the event type is `foo.bar.baz`. This was broken before this change.

libvast/vast/format/json.hpp
```diff
@@ -99,129 +95,4 @@ class reader final : public multi_layout_reader {
   mutable size_t num_lines_ = 0;
 };
 
-// -- implementation ----------------------------------------------------------
```
dominiklohmann, line 102: Note for the reviewer: All of this moved into the `.cpp` file because it's no longer templated. I did not make any changes to the logic outside of the reader constructor, which parses the new option and creates the selector now.

mavam:
How could I interpret this with the upcoming schema module changes? Would `prefix` just be the module?

dominiklohmann:
> How could I interpret this with the upcoming schema module changes? Would prefix just be the module?

Yes, exactly.

6yozo (approved):
LGTM.

libvast/vast/format/json/field_selector.hpp
```diff
@@ -18,16 +19,23 @@
...
-struct field_selector {
-  std::optional<vast::type> operator()(const ::simdjson::dom::object& j) {
-    auto el = j.at_key(Specification::field);
-    if (el.error())
+struct field_selector final : selector {
+  /// Constructs a field selector given a field name and a type prefix.
+  /// @pre `!field_name.empty()`
+  inline field_selector(std::string field_name, std::string type_prefix)
+    : field_name_{std::move(field_name)}, type_prefix_{std::move(type_prefix)} {
+    VAST_ASSERT(!field_name_.empty());
```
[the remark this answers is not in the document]
dominiklohmann: Really nice catch. I've addressed this in 76bf8c15ff2ecf28f37289657fdc8bb4fec8d1bb.

PR #1975, opened 2021-11-22
Partition Transforms Refactoring

lava:
Several smaller refactorings that make working with partition transforms more convenient. In particular:

  * Storing the transform configuration in the node state, so that plugins can have access to it.
  * Exposed a function to make a transform object out of a configuration object describing a transform.
  * Fixed a bug where transformed partitions were not correctly added to the index state.
  * Various clean-ups and clang-tidy suggestions.

### :memo: Checklist

### DIFF
--- a/libvast/src/system/partition_transformer.cpp
+++ b/libvast/src/system/partition_transformer.cpp
@@ -25,6 +25,8 @@ namespace vast::system {
 // constructed, we don't have to spawn separate indexer actors and
 // stream data but can just compute everything inline here.
 void partition_transformer_state::add_slice(const table_slice& slice) {
+  // Flatten the layout analogous to the active partition, to ensure
+  // `slice.columns()` corresponds to `layout.fields`.
   const auto& layout = flatten(slice.layout());
   data.events += slice.rows();
   data.synopsis->add(slice, synopsis_opts);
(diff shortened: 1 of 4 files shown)


dominiklohmann:
> * Storing the transform configuration in the node state, so that plugins can have access to it.

Don't you already have this available in the plugin from the actor system configuration?

lava:
>  Don't you already have this available in the plugin from the actor system configuration?

@dominiklohmann As far as I can tell the plugin only gets to see its own configuration from `plugins.plugin_name`, but not the whole actor system config?

dominiklohmann:
> @dominiklohmann As far as I can tell the plugin only gets to see its own configuration from `plugins.plugin_name`, but not the whole actor system config?

Talked about this in a call: `node->system().config()` should contain the whole configuration object known inside the server process when inside `make_component(...)`.

libvast/src/system/partition_transformer.cpp
```diff
@@ -25,10 +25,9 @@ namespace vast::system {
 // constructed, we don't have to spawn separate indexer actors and
 // stream data but can just compute everything inline here.
 void partition_transformer_state::add_slice(const table_slice& slice) {
-  const auto& layout = slice.layout();
+  const auto& layout = flatten(slice.layout());
```
dominiklohmann, line 28: I've been trying to get rid of flatten calls as part of #1888. Can you add a comment that explains why we would need to flatten here so it's easier for me to reason about this change when updating this to make use of the new type system?

dominiklohmann (approved):
Just a minor request, otherwise this is LGTM after the config change. I think we can merge this before #1888; I will make the required changes in there still.

PR #1977, opened 2021-11-23
Improve caching behavior in CI

dominiklohmann:
GitHub recently increased the cache to 10GB per repository[^1], which caused me to look at our CI configuration in that regard, where I noticed that we only accounted for caching for one branch at the same time. This should improve the situation a fair bit. The inline comment explains the reasoning behind the chosen size.

[^1]: https://github.blog/changelog/2021-11-23-github-actions-cache-size-is-now-increased-to-10gb-per-repository/

Additionally, this takes advantage of us using the modified cache action that always uploads the cache by changing by making the key unique for every branch, which in turn makes it so there only is a single cache per pull request.

### :memo: Checklist

This is developer-facing only, so it probably suffices if you confirm the changes I made are sound, and have verified that this is the behavior we want to have.

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -14,8 +14,10 @@ on:
   release:
     types: published
 env:
-  # 5G divided by 3 entries in the Matrix with a little bit of room for error.
-  CCACHE_MAXSIZE: '1600M'
+  # 10G cache total w/ 2 runs per branch and a target size of 325M per branch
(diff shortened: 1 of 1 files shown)


lava (approved):
I didn't read this too deeply tbh, but we should be able to just yolo it and see what happens since the worst case is that builds take longer again because the caching breaks, and in that case we can simply revert this.

PR #1986, opened 2021-11-30
Add an ID to each query

tobim:
This adds a random UUID as identifier for queries. This ID is meant to be used for metrics and logging to improve observability of the query pipeline.

### :memo: Checklist

By commit.


libvast/vast/system/actors.hpp
```diff
@@ -201,16 +201,15 @@ using meta_index_actor = typed_actor_fwd<
   caf::replies_to<atom::merge, std::shared_ptr<std::map<
                                  uuid, partition_synopsis>>>::with<atom::ok>,
   // Merge a single partition synopsis.
-  caf::replies_to<atom::merge, uuid, std::shared_ptr<partition_synopsis>>::with< //
-    atom::ok>,
+  caf::replies_to<atom::merge, uuid,
+                  std::shared_ptr<partition_synopsis>>::with<atom::ok>,
```
dominiklohmann, line 205: Why did you make this formatting change? The forced line break was intentional to put the reply on the next line.

libvast/vast/system/actors.hpp
```diff
@@ -201,16 +201,15 @@ using meta_index_actor = typed_actor_fwd<
...
   // Erase a single partition synopsis.
   caf::replies_to<atom::erase, uuid>::with<atom::ok>,
   // Atomically remove one and merge another partition synopsis
   caf::replies_to<atom::replace, uuid, uuid,
                   std::shared_ptr<partition_synopsis>>::with<atom::ok>,
   // Evaluate the expression.
-  caf::replies_to<atom::candidates, vast::expression,
-                  vast::ids>::with<std::vector<vast::uuid>>>
+  caf::replies_to<atom::candidates, vast::query>::with<std::vector<vast::uuid>>>
```
dominiklohmann, line 212: The comment above is outdated, as this now evaluates a query rather than an expression. Also we could probably say "return candidate partitions for query" rather than "evaluate expression/query", as that is much easier to understand.

libvast/vast/system/exporter.hpp
```diff
@@ -66,7 +66,7 @@ struct exporter_state {
   std::chrono::system_clock::time_point start = {};
 
   /// Stores various meta information about the progress we made on the query.
-  query_status query = {};
+  query_status status = {};
```
dominiklohmann, line 69:
```suggestion
  struct query_status query_status = {};
```

Gives you the best of both worlds and a lot of clarity at all use sites when reading the code.

libvast/vast/query.hpp
```diff
@@ -71,6 +72,8 @@ struct query {
 
   enum class priority { normal, low };
 
+  uuid id = uuid::random();
```
dominiklohmann, line 75: Can you add some comments to the data members of this class explaining what they are for?

Also, I think that the actor ID of the exporter actor may be better to use than generating a new random ID here. Otherwise `vast get 1 2 3` is gonna have three different query IDs, and I'm not sure that's the intended behavior.

dominiklohmann (commented):
Just some notes on the code. Haven't run anything locally yet.

Should we be adding this ID to existing metrics already, or is that part of a separate PR?

libvast/vast/system/actors.hpp
tobim: The forced break exceeded the 80 character limit. I can revert that.

libvast/vast/query.hpp
tobim: An actor can issue multiple queries, but it would not make sense for them to have the same IDs.

PR #1987, opened 2021-11-30
Add metadata elements to the metrics events

tobim:
This change makes it possible to add pairs of strings as metadata to a metrics event.

This also removes code for the defunct server side STDOUT heartbeat and 2 commits adding accountant actor handles to the meta index and partition in preparation of additional metrics.

This depends on #1986.

### :memo: Checklist

- [x] Add a changelog entry.
- [ ] Update the docs.

By commit.

### DIFF
--- a/libvast/vast/system/report.hpp
+++ b/libvast/vast/system/report.hpp
@@ -22,8 +22,10 @@
 namespace vast::system {
 
 /// A set of tags to attach to a metrics event.
-// This is also declared in `actors.hpp`
-using metrics_metadata = std::vector<std::pair<std::string, std::string>>;
+struct metrics_metadata : std::vector<std::pair<std::string, std::string>> {
+  using super = std::vector<std::pair<std::string, std::string>>;
+  using super::super;
+};
 
 struct data_point {
   std::string key;
--- a/libvast/vast/system/report.hpp
+++ b/libvast/vast/system/report.hpp
@@ -21,27 +21,50 @@
 
 namespace vast::system {
 
+/// A set of tags to attach to a metrics event.
+// This is also declared in `actors.hpp`
+using metrics_metadata = std::vector<std::pair<std::string, std::string>>;
+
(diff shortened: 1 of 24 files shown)


libvast/vast/system/report.hpp
```diff
@@ -21,27 +21,50 @@
...
+struct performance_report {
+  std::vector<performance_sample> data = {};
+  metrics_metadata metadata = {};
 
-using performance_report = std::vector<performance_sample>;
-using report = std::vector<data_point>;
+  template <class Inspector>
+  friend typename Inspector::result_type
+  inspect(Inspector& f, performance_report& x) {
+    return f(caf::meta::type_name("preformance_report"), x.data, x.metadata);
```
dominiklohmann, line 66:
```suggestion
    return f(caf::meta::type_name("performance_report"), x.data, x.metadata);
```

libvast/vast/system/report.hpp
```diff
@@ -21,27 +21,50 @@
 
 namespace vast::system {
 
+/// A set of tags to attach to a metrics event.
+// This is also declared in `actors.hpp`
+using metrics_metadata = std::vector<std::pair<std::string, std::string>>;
```
dominiklohmann, line 26:
```suggestion
/// A set of tags to attach to a metrics event.
struct metrics_metadata : std::vector<std::pair<std::string, std::string>> {
  using super = std::vector<std::pair<std::string, std::string>>;
  using super::super;
};
```

Makes for way better error messages. We should use "strong" aliases for anything used in actor interfaces. Also this avoids you having to duplicate the alias in `actors.hpp`.

dominiklohmann (commented):
I think this is a user-facing change: adapting the metrics layout requires a changelog entry and a documentation update.

Can you also expand on _why_ you are making this change? That's not immediately obvious from the PR description, and I think it really should be.

PR #1989, opened 2021-12-03
Add support for xxHash >= 0.8.1

dominiklohmann:
The patch version bump from 0.8 to 0.8.1 removed the `XXH_ACCEPT_NULL_INPUT_POINTER` macro.

### :memo: Checklist

Double-check with the release notes for xxHash 0.8.1: https://github.com/Cyan4973/xxHash/releases/tag/v0.8.1

### DIFF
--- a/libvast/CMakeLists.txt
+++ b/libvast/CMakeLists.txt
@@ -393,7 +393,12 @@ find_package(yaml-cpp 0.6.2 REQUIRED)
 target_link_libraries(libvast PRIVATE yaml-cpp)
 dependency_summary("yaml-cpp" yaml-cpp "Dependencies")
 
-# Link against xxhash.
+# Link against xxHash.
+target_compile_definitions(
+  libvast
+  PUBLIC # Allow null pointer input when hashing data of length greater 0.
+         # TODO: Remove this when requiring xxHash >= 0.8.1.
+         XXH_ACCEPT_NULL_INPUT_POINTER=1)
 find_package(xxhash REQUIRED)
 provide_find_module(xxhash)
 string(APPEND VAST_FIND_DEPENDENCY_LIST "\nfind_package(xxhash REQUIRED)")
--- a/libvast/vast/hash/xxhash.hpp
+++ b/libvast/vast/hash/xxhash.hpp
@@ -30,10 +30,6 @@ static constexpr int xxh_force_memory_access = XXH_FORCE_MEMORY_ACCESS;
 static constexpr int xxh_force_memory_access = 0;
 #endif
 
-/// Allow null pointer input when hashing data of length greater 0.
-static constexpr bool xxh_accept_null_input_pointer
-  = XXH_ACCEPT_NULL_INPUT_POINTER;
-
 class xxh64 {
 public:
   static constexpr detail::endian endian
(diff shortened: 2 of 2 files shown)


libvast/vast/hash/xxhash.hpp
```diff
@@ -30,9 +30,14 @@ static constexpr int xxh_force_memory_access = XXH_FORCE_MEMORY_ACCESS;
 static constexpr int xxh_force_memory_access = 0;
 #endif
 
-/// Allow null pointer input when hashing data of length greater 0.
+/// Allow null pointer input when hashing data of length greater 0. Enabled by
+/// default for xxHash >= 8.0.1.
```
mavam, line 34:
```suggestion
/// Allow null pointer input when hashing data of length greater 0
/// Since xxHash 8.0.1, all functions operate on null pointers, as long as
/// the size parameter is 0.
```

mavam (commented):
This works.

However, I would suggest to invert the logic: remove everything that's dependant on accepting null pointers >= 8.0.1 such that we can remove the legacy code easily in the future.

This means changing the assertion starting 0.8.1 to:
```cpp
VAST_ASSERT(bytes.data() != nullptr || bytes.size() == 0);
```

libvast/vast/hash/xxhash.hpp
```diff
@@ -30,10 +30,6 @@ static constexpr int xxh_force_memory_access = XXH_FORCE_MEMORY_ACCESS;
 static constexpr int xxh_force_memory_access = 0;
 #endif
 
-/// Allow null pointer input when hashing data of length greater 0.
-static constexpr bool xxh_accept_null_input_pointer
-  = XXH_ACCEPT_NULL_INPUT_POINTER;
```
mavam, line 35: Can we do something like this?
```cpp
#ifdef XXHASH < 0.8.1
#define XXH_ACCEPT_NULL_INPUT_POINTER
#endif
```

mavam (commented):
Almost ready to go, except that it's easy to support compatibility with xxHash < 0.8.1, which we should do in this case.

libvast/vast/hash/xxhash.hpp
dominiklohmann: I don't think this changes the logic at all, does it?

mavam: You need to define the constant, actually prior to including the xxHash header, as it changes how null pointers are handled. If you do not define this constant, we may get a crash if we attempt to hash `nullptr`, even if the size parameter is 0.

dominiklohmann: In that case we should let CMake define it. That'd be the proper way to do it.

dominiklohmann: Done in CMake in a separate commit.

mavam:
@tobim please review the CMake piece. I'm clocking out on this PR, LGTM from my side.

libvast/CMakeLists.txt
```diff
@@ -393,7 +393,11 @@ find_package(yaml-cpp 0.6.2 REQUIRED)
 target_link_libraries(libvast PRIVATE yaml-cpp)
 dependency_summary("yaml-cpp" yaml-cpp "Dependencies")
 
-# Link against xxhash.
+# Link against xxHash.
+target_compile_definitions(
```
tobim, line 397: Please add a comment so we don't forget to remove this when we update the minimum version of XXHash.

tobim (approved):

PR #1992, opened 2021-12-06
Emit fine grained metrics for every part of the query pipeline

tobim:
This change adds metrics events that are emitted for each query from the meta index, partition and store components. All new metrics and the preexisting exporter metrics events use the query ID as metadata value to allow cross-referencing.

### :memo: Checklist

Each commit can be reviewed individually in order.
The last commit modifies the interface of the store plugin, which is unfortunate and I would like to discuss alternative approaches.

### DIFF
--- a/changelog/unreleased/features/1987-1992--metrics-metatdata.md
+++ b/changelog/unreleased/features/1987-1992--metrics-metatdata.md
@@ -1,0 +1,16 @@
+Metrics events now optionally contain a metadata field that is a key-value
+mapping of string to string, allowing for finer-grained introspection. For now
+this enables correlation of metrics events and individual queries. A set of new
+metrics for query lookup use this feature to include the query ID. I.e.:
+```
+{
+  "ts": "2021-12-09T12:47:09.079148669",
+  "version": "2021.11.18-269-gba9f97bf77-dirty",
+  "actor": "meta-index",
+  "key": "meta-index.lookup.runtime",
+  "value": 0.070954,
+  "metadata": {
+    "query": "7E18BF00-0C8C-4841-B8A4-C8EAEB9E9203"
+  }
+}
+```
--- a/libvast/src/system/local_segment_store.cpp
+++ b/libvast/src/system/local_segment_store.cpp
@@ -19,6 +19,7 @@
 #include "vast/query.hpp"
 #include "vast/segment_store.hpp"
 #include "vast/system/node_control.hpp"
+#include "vast/system/report.hpp"
 #include "vast/system/status.hpp"
 #include "vast/table_slice.hpp"
 
@@ -26,6 +27,7 @@
 #include <flatbuffers/flatbuffers.h>
 #include <fmt/format.h>
 
+#include <chrono>
 #include <span>
 #include <vector>
 
@@ -36,10 +38,11 @@ namespace {
 // Handler for `vast::query` that is shared between active and passive stores.
 // Precondition: Query type is either `count` or `extract`.
 template <typename Actor>
-caf::result<atom::done> handle_lookup(Actor& self, const vast::query& query,
-                                      const std::vector<table_slice>& slices) {
+caf::expected<size_t> handle_lookup(Actor& self, const vast::query& query,
+                                    const std::vector<table_slice>& slices) {
   const auto& ids = query.ids;
   std::vector<expression> checkers;
+  size_t num_hits = 0ull;
   for (const auto& slice : slices) {
     if (query.expr == expression{}) {
       checkers.emplace_back();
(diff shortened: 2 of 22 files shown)


libvast/src/system/local_segment_store.cpp
```diff
@@ -36,10 +39,11 @@ namespace {
 // Handler for `vast::query` that is shared between active and passive stores.
 // Precondition: Query type is either `count` or `extract`.
 template <typename Actor>
-caf::result<atom::done> handle_lookup(Actor& self, const vast::query& query,
-                                      const std::vector<table_slice>& slices) {
+caf::expected<size_t> handle_lookup(Actor& self, const vast::query& query,
```
tobim, line 42: Document the return value.

libvast/vast/system/local_segment_store.hpp
```diff
@@ -12,6 +12,7 @@
 
 #include "vast/chunk.hpp"
 #include "vast/segment_builder.hpp"
+#include "vast/system/accountant.hpp"
```
tobim, line 15: This include was added erroneously by clangd. The same in a couple of other files.

dominiklohmann:
I'm approving of the changes to the plugin interface. We should consider making it a component plugin in the future, though, but this PR is neither the time nor the place to do it.

libvast/src/system/local_segment_store.cpp
```diff
@@ -36,10 +38,11 @@ namespace {
 // Handler for `vast::query` that is shared between active and passive stores.
 // Precondition: Query type is either `count` or `extract`.
 template <typename Actor>
```
6yozo, line 40: What about something like:

```suggestion
// Precondition: Query type is either `count` or `extract`.
/// @returns The matching/extracted rows.
template <typename Actor>
```

changelog/unreleased/features/1987-1992--metrics-metatdata.md
```diff
@@ -0,0 +1,7 @@
+Metrics events now optionally contain a metadata field that is a key-value
+mapping of string to string, allowing for finer-grained introspection. For now
+this enables correlation of metrics events and individual queries. A set of new
+metrics for query lookup use this feature to include the query ID. I.e.:
+```
+{"ts": "2021-12-09T09:42:16.797478769","version": "2021.11.18-266-gbb60334346","actor": "meta-index","key": "meta-index.lookup.candidates","value": 42,"query":"DBECE02A-680C-4FB7-A806-69F262B2DB0A"}
```
6yozo, line 6: Is it ok to have longer than 80 chars in the changelog, or should we format it multiline like:

```suggestion
{
  "ts": "2021-12-09T09:42:16.797478769",
  "version": "2021.11.18-266-gbb60334346",
  "actor": "meta-index",
  "key": "meta-index.lookup.candidates",
  "value": 42,
  "query": "DBECE02A-680C-4FB7-A806-69F262B2DB0A"
}
```

dominiklohmann: It's not gonna render correctly either way. Every paragraph gets joined into a single line. After building, you can check `build/CHANGELOG.md` to find how the generated changelog file looks.

dominiklohmann, line 6: I don't think this example JSON is even correct, because it's missing the metadata field in which the query field should be nested.

dominiklohmann (commented):
One general question: Did you

changelog/unreleased/features/1987-1992--metrics-metatdata.md
tobim: As explained in the documentation the metadata fields are folded into the top level object for the `file` and `UDS` sinks.

dominiklohmann: Doesn't that add unnecessary complexity?

tobim: I guess it doesn't hurt to make the outputs identical.

I initially did it this way because the `self-sink` is somewhat unused nowadays and I thought the nesting under `metadata` does little to improve the usability.

dominiklohmann: It's alright to me. No need to change it now.

I think we wanted to get rid of the self-sink anyways, so that may be a good thing to do in the near future. This whole `vast.metrics` thing never really worked out the way we wanted to and it makes writing metrics unnecessarily complex.

tobim: I aligned it now anyways, so we still have all options in the future.

6yozo (approved):
We did the review synchronously, paired up with Tobias, and fixed compilation problems.

Good to go.

PR #1996, opened 2021-12-08
Capture the ingest time for new table slices

dominiklohmann:
This adds the ingest time to the table slice API, which is set in the server process alongside the id/offset assignment.

### :memo: Checklist

Commit-by-commit.

### DIFF
--- a/libvast/src/arrow_table_slice_builder.cpp
+++ b/libvast/src/arrow_table_slice_builder.cpp
@@ -550,8 +550,11 @@ table_slice arrow_table_slice_builder::finish() {
   // Create Arrow-encoded table slices. We need to set the import time to
   // something other than 0, as it cannot be modified otherwise. We then later
   // reset it to the clock's epoch.
-  auto arrow_table_slice_buffer = fbs::table_slice::arrow::Createv1(
-    builder_, layout_buffer, schema_buffer, record_batch_buffer, 1337);
+  constexpr int64_t stub_ns_since_epoch = 1337;
+  auto arrow_table_slice_buffer
+    = fbs::table_slice::arrow::Createv1(builder_, layout_buffer, schema_buffer,
+                                        record_batch_buffer,
+                                        stub_ns_since_epoch);
   // Create and finish table slice.
   auto table_slice_buffer
     = fbs::CreateTableSlice(builder_, fbs::table_slice::TableSlice::arrow_v1,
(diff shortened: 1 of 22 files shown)


libvast/src/table_slice.cpp
```diff
@@ -282,6 +282,35 @@ void table_slice::offset(id offset) noexcept {
...
+    },
+    [&](const auto& encoded) noexcept {
+      return state(encoded, state_)->import_time();
+    },
+  };
+  return visit(f, as_flatbuffer(chunk_));
+}
+
+void table_slice::import_time(time import_time) noexcept {
+  VAST_ASSERT(chunk_->unique());
```
dominiklohmann, line 298: We can also "unsure" the chunk here, which is probably the safer option, and appears to be necessary to avoid larger unit test refactorings.

libvast/src/arrow_table_slice_builder.cpp
```diff
@@ -547,9 +547,11 @@ table_slice arrow_table_slice_builder::finish() {
         .ValueOrDie();
   auto record_batch_buffer = builder_.CreateVector(flat_record_batch->data(),
                                                    flat_record_batch->size());
-  // Create Arrow-encoded table slices.
+  // Create Arrow-encoded table slices. We need to set the import time to
+  // something other than 0, as it cannot be modified otherwise. We then later
+  // reset it to the clock's epoch.
   auto arrow_table_slice_buffer = fbs::table_slice::arrow::Createv1(
-    builder_, layout_buffer, schema_buffer, record_batch_buffer);
+    builder_, layout_buffer, schema_buffer, record_batch_buffer, 1337);
```
lava, line 554: `1337` should be a named constant

libvast/vast/fbs/table_slice.fbs
```diff
@@ -32,6 +32,9 @@ table v1 {
 
   /// The Arrow Recod Batch containing all data.
   record_batch: [ubyte];
+
+  /// A timestamp assigned on import, representing a `vast::time` value.
```
lava, line 36: `representing nanoseconds since the unix epoch` may be more useful for the casual reader.

libvast/src/concept/parseable/vast/expression.cpp
```diff
@@ -129,6 +129,13 @@ static auto make_predicate_parser() {
     = (parsers::data >> !(field_char | '.')) ->* to_data_operand
     | "#type"_p  ->* [] { return meta_extractor{meta_extractor::type}; }
     | "#field"_p ->* [] { return meta_extractor{meta_extractor::field}; }
+    // TODO: We didn't agree on the name for the meta extractor in time for the
+    // release, so we need to first come to a conclusion, and then add it back
+    // to the parser and rename the meta extractor to make it a user-facing
+    // change. We still wanted to have dates assigned to table slices and date
+    // ranges assigned to partition synopses as soon as possible to have more
+    // data in databases for upcoming features that can make use of them.
```
lava, line 137: That's pretty verbose, probably all that readers need to care about is that we currently don't expose `#age` in the query language but are planning to do so at a later point in time.

lava (approved):
We already live-tested in a video call that it works as expected; and agreed to add a unit test if possible with small effort or else a story to add one later.

The remaining issues are not really important and more for documentation purposes, feel free to resolve them as you see fit.

dominiklohmann:
> and agreed to add a unit test if possible with small effort or else a story to add one later.

I've opted for the latter, as it's far from trivial. We don't have a unit test spinning up so many components, because we usually use integration tests for that, which happen to be a bad fit for this particular feature.

PR #1997, opened 2021-12-08
Remove the erroneous warning about exceeded merge recursion limit

dispanser:
The warning was actually triggered because the counter was decremented
in every iteration of a loop, instead of only on recursion.

We fix that and make the function argument `const` while we're at it.

### :memo: Checklist

This should be straight forward, one file two lines.

### DIFF
--- a/libvast/src/data.cpp
+++ b/libvast/src/data.cpp
@@ -278,7 +278,7 @@ record flatten(const record& r) {
 namespace {
 
 void merge(const record& src, record& dst, enum policy::merge_lists merge_lists,
-           size_t max_recursion) {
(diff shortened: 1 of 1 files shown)


dominiklohmann (approved):
I verified locally that the output of `vast status --debug` works as expected and doesn't trigger the error anymore. Thanks for fixing! 🚀

mavam:
No bugfix changelog entry?

dominiklohmann:
> No bugfix changelog entry?

We've never hit this bug before, also `status --debug` really isn't user-facing, but rather developer-facing. I'd say this doesn't need one.

PR #2001, opened 2021-12-10
Restore aging

lava:
As it turns out aging was actually used in production, unknown to us. We therefore need to delay removal of this feature until we have implemented an equivalent replacement using compaction.

### :memo: Checklist


lava:
For the docs, I think it should be ok to leave them as-is and not restore the section on aging, since we still plan to remove this in the near future.

dominiklohmann:
> For the docs, I think it should be ok to leave them as-is and not restore the section on aging, since we still plan to remove this in the near future.

I agree with this assessment. We should still remove this as soon as time-based compaction is available, and then provide a guide on how to migrate. We also need to mark experimental things as experimental more clearly in the future.

dominiklohmann (approved):

PR #2003, opened 2021-12-10
Fix minimum import time in partition synopses

dominiklohmann:
Turns out that `min(epoch, t)` is always `epoch`. This fixes that by assigning `time::max()` for the minimum import time.

### :memo: Checklist

Not really user-facing (yet), so no changelog entry needed. The fix is quite easy to understand.

### DIFF
--- a/libvast/src/system/meta_index.cpp
+++ b/libvast/src/system/meta_index.cpp
@@ -282,12 +282,8 @@ std::vector<uuid> meta_index_state::lookup_impl(const expression& expr) const {
                 part_syn.max_import_time,
               };
               auto add = ts.lookup(x.op, caf::get<vast::time>(d));
-              if (!add || *add) {
-                VAST_WARN("yes: part {} w/ range [{}, {}]", part_id,
-                          data{part_syn.min_import_time},
-                          data{part_syn.max_import_time});
+              if (!add || *add)
                 result.push_back(part_id);
-              }
             }
             VAST_ASSERT(std::is_sorted(result.begin(), result.end()));
             return result;
--- a/libvast/src/system/meta_index.cpp
+++ b/libvast/src/system/meta_index.cpp
@@ -276,34 +276,18 @@ std::vector<uuid> meta_index_state::lookup_impl(const expression& expr) const {
           } else if (lhs.kind == meta_extractor::age) {
             result_type result;
             for (const auto& [part_id, part_syn] : synopses) {
-              const auto t = caf::get<time>(d);
-              bool add = false;
-              switch (x.op) {
-                case relational_operator::match:
-                case relational_operator::not_match:
-                case relational_operator::in:
-                case relational_operator::not_in:
-                case relational_operator::ni:
-                case relational_operator::not_ni:
-                case relational_operator::equal:
-                case relational_operator::not_equal:
-                  VAST_ASSERT(false, "unexpected operator");
-                  break;
-                case relational_operator::less:
-                  add = part_syn.min_import_time < t;
-                  break;
-                case relational_operator::less_equal:
-                  add = part_syn.min_import_time <= t;
-                  break;
-                case relational_operator::greater:
-                  add = part_syn.max_import_time > t;
-                  break;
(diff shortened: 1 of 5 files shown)


lava (approved):

dominiklohmann:
@lava feel free to merge by yourself after approving of the additional unit tests.

dominiklohmann:
[sc-29848]

libvast/src/system/meta_index.cpp
```diff
@@ -276,34 +276,18 @@ std::vector<uuid> meta_index_state::lookup_impl(const expression& expr) const {
...
-              }
-              if (add)
+              VAST_ASSERT(part_syn.min_import_time <= part_syn.max_import_time);
+              auto ts = time_synopsis{
+                part_syn.min_import_time,
+                part_syn.max_import_time,
+              };
+              auto add = ts.lookup(x.op, caf::get<vast::time>(d));
+              if (!add || *add) {
+                VAST_WARN("yes: part {} w/ range [{}, {}]", part_id,
```
lava, line 286: That looks like left-over debug code? (also, is this functionally the same as the previous version, or did it include a bug that was fixed by using the time synopsis?)

libvast/test/system/counter.cpp
```diff
@@ -166,4 +168,76 @@ TEST(count IP point query with partition - local stores) {
...
+  auto indexdir = directory / "index2";
+  auto index
+    = self->spawn(system::index, system::accountant_actor{}, fs, archive,
+                  indexdir, "segment-store", defaults::import::table_slice_size,
+                  100, 3, 1, indexdir, 0.01);
+  // Fill the INDEX with 400 rows from the Zeek conn log.
+  auto slices = take(zeek_conn_log_full, 4);
+  for (auto& slice : slices) {
+    slice = slice.unshare();
+    slice.import_time(time::clock::now());
```
lava, line 182: Do we have to set a valid offset here?

libvast/vast/detail/partition_common.hpp
```diff
@@ -50,8 +51,26 @@ fetch_indexer(const PartitionState& state, const meta_extractor& ex,
         row_ids |= ids;
     }
   } else if (ex.kind == meta_extractor::age) {
-    for (const auto& [_, ids] : state.type_ids())
-      row_ids |= ids;
+    // For the passive partition we can trust the eval from the meta index, but
+    // we cannot rely on that for the active partition. We create a time
```
lava, line 55: That was a bit confusing to understand, I'd suggest something like "For a passive partition, this already went through a time synopsis in the meta index, but for the active partition we create an ad-hoc time synopsis here to do the lookup."

libvast/test/system/meta_index.cpp
```diff
@@ -268,6 +285,35 @@ TEST(attribute extractor - type) {
   CHECK_EQUAL(lookup("#type !~ /x/"), ids);
 }
 
+TEST(attribute extractor - age) {
```
lava, line 288: Can you add a comment that half the test data was set to import age 1975 and the other half to 2015 in the test setup?

lava (approved):
Looks good module the unnecessary WARN() log message; the unit tests seem to cover the basic usage.

libvast/test/system/counter.cpp
dominiklohmann: Nope, `slice.unshare()` already takes care of that.

libvast/src/system/meta_index.cpp
dominiklohmann: > also, is this functionally the same as the previous version, or did it include a bug that was fixed by using the time synopsis?

Same behavior, just less code to maintain. I didn't want to duplicate it again, considering how much trouble we had wrapping our head around the seemingly simple comparison yesterday.

libvast/test/system/counter.cpp
lava: Only if the slice had a valid offset to begin with, but it doesn't seem to be relevant for this test anyways.

PR #2005, opened 2021-12-14
Implement an Arrow fast path for row selection

dominiklohmann:
This implements a fast path via Arrow's RecordBatch API for row selection for the `filter` and `select` operations on Table Slices.

### :memo: Checklist

Read the code, which itself is quite simple—and do performance benchmarking. This is more of a proof of concept, and wherever we currently emit log messages we should probably abort because we either ran out of memory or made a logic error.

If this turns out to be a performance-relevant change then we need to add a user-facing changelog entry as well.

Edit: From my own measurements this is only noticeable in very constructed scenarios, i.e., when `filter` returns every second row or so. I don't think this needs a changelog entry.

### DIFF
--- a/libvast/src/arrow_table_slice.cpp
+++ b/libvast/src/arrow_table_slice.cpp
@@ -15,6 +15,7 @@
 #include "vast/detail/overload.hpp"
 #include "vast/die.hpp"
 #include "vast/error.hpp"
+#include "vast/expression.hpp"
 #include "vast/fbs/table_slice.hpp"
 #include "vast/fbs/utils.hpp"
 #include "vast/legacy_type.hpp"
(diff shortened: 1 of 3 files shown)


tobim:
I added a commit that moves the filter optimization into a function in `arrow_table_slice.cpp`. The same will have to be done for `select`, then we can remove `supports_arrow_fast_path` crutch.

tobim:
I don't understand this compiler error and can't reproduce it locally.

dominiklohmann:
> I don't understand this compiler error and can't reproduce it locally.

The CI builds against this branch as if it was merged against master, which has the extractor renamed from `age` to `import_time`. We simply need to rebase the branch.

dominiklohmann:
I'm closing this. We never really followed through on the current implementation, and since we want to change this in the future anyways it doesn't really make sense to invest more time here.

PR #2007, opened 2021-12-15
Rebuild older Arrow table slices in `as_record_batch`

dominiklohmann:
This is a bug fix for transforms, which operated under the assumption that it was always safe to create a new table slice from a record batch. If, however, assumptions about the record batch memory layout change between Arrow table slice encoding versions, this will no longer work, requiring us to rebuild the slice to the latest encoding version before retrieving the underlying record batch.

### :memo: Checklist

This bug is not affecting the current release, so (1) we don't need to merge it before the release and (2) it doesn't need a changelog entry. It is preparatory for upcoming work.

### DIFF
--- a/libvast/src/table_slice.cpp
+++ b/libvast/src/table_slice.cpp
@@ -391,6 +391,14 @@ std::shared_ptr<arrow::RecordBatch> as_record_batch(const table_slice& slice) {
         = std::decay_t<decltype(*state(encoded, slice.state_))>::encoding;
       if constexpr (encoding == table_slice_encoding::arrow
                     || encoding == table_slice_encoding::experimental) {
+        // If we have a record batch, but it is from an older table slice
+        // encoding, we must still rebuild the table slice. Otherwise, creating
+        // a new table slice from the returned record batch leads to undefined
+        // behavior.
+        if (!state(encoded, slice.state_)->is_latest_version) {
+          auto copy = rebuild(slice, encoding);
+          return as_record_batch(copy);
+        }
         // Get the record batch first, then create a copy that shares the
         // lifetime with the chunk and the original record batch. Capturing the
         // chunk guarantees that the table slice is valid as long as the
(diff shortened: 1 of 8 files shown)


dominiklohmann:
One other thing we should probably do is rename the `as_record_batch` function. It's name implies that it'd be a cheap operation, which is only the case most of the time.

libvast/src/table_slice.cpp
```diff
@@ -380,6 +380,14 @@ std::shared_ptr<arrow::RecordBatch> as_record_batch(const table_slice& slice) {
       //                 == table_slice_encoding::arrow) { ... }
       if constexpr (std::decay_t<decltype(*state(encoded, slice.state_))>::encoding
                     == table_slice_encoding::arrow) {
+        // If we have a record batch, but it is from an older table slice
+        // encoding, we must still rebuild the table slice. Otherwise, creating
+        // a new table slice from the returned record batch leads to undefined
+        // behavior.
+        if (!state(encoded, slice.state_)->is_latest_version) {
+          auto copy = rebuild(slice, table_slice_encoding::arrow);
```
dominiklohmann, line 388: We should probably rename the function from `as_record_batch` to something else, as possibly rebuilding is a hidden cost that the name does not imply properly.

mavam: Yes, thanks for remembering this discussion point. Any `as_*` function implies basically "no cost".

dispanser (approved):
Fit for purpose ;). I consider the renaming of `as_record_batch` to be a separate concern.

mavam:
> I consider the renaming of `as_record_batch` to be a separate concern.

Could you create a separate story for that? Then we can be all relieved and postpone this discussion. 😉

dominiklohmann:
A quick Slack vote resulted in me renaming this to `to_record_batch` rather than making that a separate story.

PR #2008, opened 2021-12-15
Introduce an experimental Arrow table slice encoding

dominiklohmann:
The `experimental` format is explicitly unstable, and serves as a playground for upcoming changes to the Arrow encoding. We plan to merge these as a v2 of the current encoding some time in the future, but want to be able to make incremental changes. Please do not use this feature in production, as we explicitly do not guarantee backwards compatibility for it.

To enable permanently for local development, put this into your `vast.yaml` configuration file:

```yaml
vast:
  import:
    batch-encoding: experimental
```

### :memo: Checklist

This format is intentionally undocumented and thus does not require a changelog entry. For reviewing, I suggest running the integration tests locally after changing the default encoding to `arrow-ng`. Most of the code is just a copy, except for the occasional `_ng` suffix to avoid ODR violations.

### DIFF
--- a/libvast/src/experimental_table_slice.cpp
+++ b/libvast/src/experimental_table_slice.cpp
@@ -1,0 +1,744 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2021 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "vast/experimental_table_slice.hpp"
+
+#include "vast/config.hpp"
+#include "vast/detail/byte_swap.hpp"
+#include "vast/detail/narrow.hpp"
+#include "vast/detail/overload.hpp"
+#include "vast/die.hpp"
+#include "vast/error.hpp"
+#include "vast/experimental_table_slice_builder.hpp"
+#include "vast/fbs/table_slice.hpp"
+#include "vast/fbs/utils.hpp"
+#include "vast/legacy_type.hpp"
+#include "vast/logger.hpp"
+#include "vast/value_index.hpp"
+
+#include <arrow/api.h>
+#include <arrow/io/api.h>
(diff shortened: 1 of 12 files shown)


libvast/src/arrow_table_slice.cpp
```diff
@@ -6,9 +6,8 @@
 // SPDX-FileCopyrightText: (c) 2019 The VAST Contributors
 // SPDX-License-Identifier: BSD-3-Clause
 
-#include "vast/arrow_table_slice.hpp"
-
-#include "vast/arrow_table_slice_builder.hpp"
+#include "vast/arrow_ng_table_slice.hpp"
+#include "vast/arrow_ng_table_slice_builder.hpp"
 #include "vast/config.hpp"
```
dispanser: I'm confused; I would have expected the new `arrow_ng_table_slice` to be only imported in the new file(s), not for the existing arrow table slice.

dispanser (commented):
`libvast/src/arrow_table_slice.cpp` seems to contain all the types / names / fields with `_ng` tucked in, whereas `libvast/src/arrow_ng_table_slice.cpp` contains the previous code base. Is that intentional, or did you just flip the files?

libvast/src/arrow_table_slice.cpp
dominiklohmann: The files were just swapped—I renamed in the wrong copy. I already fixed this in a force-push.

dispanser:
> `libvast/src/arrow_table_slice.cpp` seems to contain all the types / names / fields with `_ng` tucked in, whereas `libvast/src/arrow_ng_table_slice.cpp` contains the previous code base. Is that intentional, or did you just flip the files?

much better now, thanks :-)

dispanser:
there's an arrow-specific test suite that should also run for the new experimental table slice at `libvast/test/arrow_table_slice.cpp`

dominiklohmann:
> there's an arrow-specific test suite that should also run for the new experimental table slice at `libvast/test/arrow_table_slice.cpp`

Ported that as well, thanks!

dispanser (approved):
verified the integration tests running & using the new experimental code path.

PR #2009, opened 2021-12-15
Make use of `table_slice::append_column_to_index`

dominiklohmann:
Instead of decoding columns in table slice for every row, we should make use of the existing API for appending a whole column at once. I'm not sure when this got thrown out, but it's been dead code before this change, although we definitely do want to make use of it.

An even better API for this would be to not have a custom function in the table slice API that takes a value index, but rather to return a `detail::generator<data_view>` per column that can be lazily iterated over. That still shouldn't stop us from using this special function before we have a more generic solution, though.

### :memo: Checklist

Measure perf.

### DIFF
--- a/libvast/src/system/indexer.cpp
+++ b/libvast/src/system/indexer.cpp
@@ -59,9 +59,9 @@ active_indexer(active_indexer_actor::stateful_pointer<indexer_state> self,
           // implementation so we're keeping it for now.
           if (self->state.has_skip_attribute)
             return;
-          for (auto& column : columns)
-            for (size_t i = 0; i < column.size(); ++i)
-              self->state.idx->append(column[i], column.slice().offset() + i);
+          for (const auto& column : columns)
+            column.slice().append_column_to_index(column.index(),
+                                                  *self->state.idx);
         },
         [=](caf::unit_t&, const caf::error& err) {
           VAST_TRACE("indexer is closing stream");


dispanser (approved):
I made a quick, hand-wavy test of the import performance using the following setup:

1. an argus csv file containing 10 mio events
2. a running `vast` instance on a clean database: `time vast start`
3. a separate client running the import: 
   `hyperfine 'build/bin/vast import --blocking -r /home/data/tenzir/vast/csv/argus-10m.csv csv' --runs 10`

We watch two metrics: 
1. the CPU time used on the vast server, to quantify the effect on the resource usage for the indexing operation
2. the execution time for the vast client, as execution time measured via hyperfine
rc2 baseline: 10x 10mio argus csv lines input:

## rc2 baseline: 10x 10mio argus csv lines input:

```
Executed in  382.97 secs    fish           external
   usr time   29.47 mins  525.00 micros   29.47 mins
   sys time    0.87 mins  202.00 micros    0.87 mins
```

```
Benchmark 1: build/bin/vast import --blocking -r /home/data/tenzir/vast/csv/argus-10m.csv csv
  Time (mean ± σ):     30.798 s ±  1.803 s    [User: 20.611 s, System: 1.707 s]
  Range (min … max):   28.707 s … 35.465 s    10 runs
```

## branch:

```                       
Executed in  352.93 secs    fish           external
   usr time   16.00 mins  556.00 micros   16.00 mins
   sys time    0.70 mins  222.00 micros    0.70 mins
```


```
Benchmark 1: build/bin/vast import --blocking -r /home/data/tenzir/vast/csv/argus-10m.csv csv
  Time (mean ± σ):     22.350 s ±  0.721 s    [User: 20.106 s, System: 1.647 s]
  Range (min … max):   21.181 s … 23.448 s    10 runs
```

We see a significant improvement in CPU utilization but also in the ingestion rate (events per second)

PR #2013, opened 2021-12-16
Use FlatBuffers for the Type Registry state

dominiklohmann:
This changes the Type Registry state to be persisted using FlatBuffers in a versioned fashion, rather than the old CAF-serialized format. When starting VAST for the first time with this PR the old state is removed automatically.

### :memo: Checklist

This should not be user-facing at all.

I recommend reading the code, and most importantly reviewing the new FlatBuffers table, and then giving it a spin locally. For me this has worked as expected so far.

### DIFF
--- a/libvast/src/system/type_registry.cpp
+++ b/libvast/src/system/type_registry.cpp
@@ -153,6 +153,10 @@ caf::error type_registry_state::load_from_disk() {
         data.emplace(k, entry);
       }
       VAST_DEBUG("{} loaded state from disk", *self);
+      // We save the new state already now before removing the old state just to
+      // be save against crashes.
+      if (auto err = save_to_disk())
+        return err;
       if (!std::filesystem::remove(fname, err) || err)
         VAST_DEBUG("failed to delete legacy type-registry state");
       return caf::none;
--- a/libvast/src/system/type_registry.cpp
+++ b/libvast/src/system/type_registry.cpp
@@ -15,6 +15,8 @@
 #include "vast/detail/legacy_deserialize.hpp"
 #include "vast/error.hpp"
 #include "vast/event_types.hpp"
+#include "vast/fbs/type_registry.hpp"
+#include "vast/flatbuffer.hpp"
 #include "vast/io/read.hpp"
 #include "vast/io/save.hpp"
 #include "vast/legacy_type.hpp"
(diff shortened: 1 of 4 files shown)


libvast/src/system/type_registry.cpp
```diff
@@ -111,27 +131,61 @@ caf::error type_registry_state::load_from_disk() {
...
+                                                "type-registry state");
+      for (const auto& [k, vs] : intermediate) {
+        auto entry = type_set{};
+        for (const auto& v : vs)
+          entry.emplace(type::from_legacy_type(v));
+        data.emplace(k, entry);
+      }
+      VAST_DEBUG("{} loaded state from disk", *self);
+      if (!std::filesystem::remove(fname, err) || err)
+        VAST_DEBUG("failed to delete legacy type-registry state");
```
dominiklohmann, line 158: I wonder if before deleting we should actually persist the new state first. If that fails we shouldn't remove the old state.

dominiklohmann: I considered this a sensible change and went ahead and implemented it.

libvast/src/system/type_registry.cpp
```diff
@@ -111,27 +130,65 @@ caf::error type_registry_state::load_from_disk() {
...
+      return caf::none;
+    }
+  }
+  // Support the new FlatBuffers state.
+  {
+    const auto fname = filename();
+    const auto file_exists = std::filesystem::exists(fname, err);
+    if (err)
+      return caf::make_error(ec::filesystem_error,
+                             fmt::format("failed to find file {}: {}", fname,
```
tobim, line 171: We can make this message a little clearer, the sentence can me misinterpreted to mean that the file was not found.
```suggestion
                             fmt::format("failed while trying to find file {}: {}", fname,
```

libvast/src/system/type_registry.cpp
```diff
@@ -111,27 +130,65 @@ caf::error type_registry_state::load_from_disk() {
...
-      for (const auto& v : vs)
-        entry.emplace(type::from_legacy_type(v));
-      data.emplace(k, entry);
+  // Support the legacy CAF-serialized state, and delete it afterwards.
+  {
+    const auto fname = dir / name;
+    const auto file_exists = std::filesystem::exists(fname, err);
+    if (err)
+      return caf::make_error(ec::filesystem_error,
+                             fmt::format("failed to find file {}: {}", fname,
```
tobim, line 139: We can make this message a little clearer, the sentence can me misinterpreted to mean that the file was not found.
```suggestion
                             fmt::format("failed while trying to find file {}: {}", fname,
```

tobim (approved):
Just a minor nit regarding an error message.

PR #2014, opened 2021-12-16
Implement a selection transform plugin

6yozo:
Create the new selection transform plugin.

### :memo: Checklist

### DIFF
--- a/libvast/src/transform_steps/select.cpp
+++ b/libvast/src/transform_steps/select.cpp
@@ -1,0 +1,80 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2021 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "vast/transform_steps/select.hpp"
+
+#include "vast/concept/parseable/to.hpp"
+#include "vast/concept/parseable/vast/expression.hpp"
+#include "vast/error.hpp"
+#include "vast/expression.hpp"
+#include "vast/logger.hpp"
+#include "vast/plugin.hpp"
+#include "vast/table_slice_builder_factory.hpp"
+
+#include <arrow/type.h>
+#include <caf/expected.hpp>
+
+namespace vast {
+
+select_step::select_step(std::string expr) : expression_(caf::no_error) {
+  auto e = to<vast::expression>(std::move(expr));
+  if (!e) {
+    // TODO: Implement a better way to report errors, for example, a separate
+    // setup() function.
+    VAST_ERROR("the select step cannot use the expression: '{}', reason: '{}'",
+               expr, e.error());
+    expression_ = e;
+    return;
+  }
+  expression_ = normalize_and_validate(*e);
+  if (!expression_) {
+    VAST_ERROR("the select step cannot validate the expression: '{}', reason: "
+               "'{}'",
+               expr, expression_.error());
+  }
+}
+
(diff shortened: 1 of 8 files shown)


libvast/vast/transform_steps/select.hpp
```diff
@@ -0,0 +1,29 @@
...
+class select_step : public generic_transform_step {
+public:
+  select_step(std::string expr);
+
+  caf::expected<table_slice> operator()(table_slice&& slice) const override;
+
+private:
+  std::string unparsed_;
+  mutable expression expression_;
+  mutable bool parsed_;
```
dominiklohmann, line 26: `mutable` usually implies thread-safe access. I'm not sure this is a solution you can feasibly take here. While it works _now_ because the `select_step` itself is only called from a single thread, it's really a bad API because it makes the wrong promises.

Can't we store a `caf::expected<expression>` instead and parse/normalize/tailor in the constructor, and then when applying the check we only need to check whether the expected holds a value (or return its error)?

libvast/src/transform_steps/select.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include "vast/plugin.hpp"
+#include "vast/table_slice_builder_factory.hpp"
+
+#include <arrow/type.h>
+#include <caf/expected.hpp>
+
+namespace vast {
+
+select_step::select_step(std::string expr)
+  : expression_(caf::none) { // FIXME: Is caf::none OK?
```
6yozo, line 25: There are some FIXME comments to clean up

libvast/src/transform_steps/select.cpp
```diff
@@ -0,0 +1,75 @@
...
+  auto e = to<vast::expression>(std::move(expr));
+  if (!e) {
+    expression_ = e;
+    return;
+  }
+  // FIXME: Is normalize useful here?
+  expression_ = normalize_and_validate(*e);
```
lava, line 32: It probably makes sense to print a `VAST_ERROR()` here already if this returns an error, so this is immediately visible on startup. (also long-term we need to think about a better way to report errors, probably a separate `setup()` function)

libvast/src/transform_steps/select.cpp
lava, line 25: I think `caf::no_error` instead of `caf::none` is the default value.

libvast/src/transform_steps/select.cpp
```diff
@@ -0,0 +1,75 @@
...
+caf::expected<table_slice> select_step::operator()(table_slice&& slice) const {
+  if (!expression_)
+    return expression_.error(); // FIXME: Is this message OK?
+  auto tailored_expr = tailor(*expression_, slice.layout());
+  if (!tailored_expr)
+    return tailored_expr.error();
+  auto new_slice = filter(slice, *tailored_expr);
+  if (new_slice)
+    return *new_slice;
+  // FIXME: return an empty slice or caf::error if new_slice is empty?
```
lava, line 44: I think currently we don't allow empty table slices, so it will have to be a `caf::error`.

libvast/test/transform.cpp
```diff
@@ -150,6 +151,15 @@ TEST(replace step) {
   CHECK_EQUAL((*replaced).at(0, 0), vast::data_view{"xxx"sv});
 }
 
+TEST(select step) {
+  auto slice = make_transforms_testdata(vast::table_slice_encoding::msgpack);
+  vast::select_step select_step("index==2");
+  auto selected = select_step.apply(vast::table_slice{slice});
+  REQUIRE_NOERROR(selected);
+  // FIXME Check result!
+  // FIXME Tests for empty slice, no change, multiple rows
```
lava, line 160: What the comment says.

lava (approved):
Nice work, this looks very close already.

I'm approving now and left comments for the FIXME's, so you can resolve them as you go and then merge for yourself afterwards.

PR #2017, opened 2022-01-03
Represent vast `duration_type`s as `arrow::DurationType` in arrow / experimental table slice

dispanser:
Instead of using `Int64Array` to represent our `duration_type`, we now use the `DurationArray` which is the proper arrow data type for durations. This only applies to our `experimental` table slice, and is therefor not considered a breaking change for our customers.

### :memo: Checklist

- [ ] All user-facing changes have changelog: No, because experimental table slices are not supposed to be used yet, so we don't consider the change user-facing.
- [ ] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary: Also not necesssary.

This should be relatively straight forward: please convince yourself that the existing unit tests for experimental table slice properly capture encoding / decoding paths for durations.

### DIFF
--- a/libvast/src/experimental_table_slice.cpp
+++ b/libvast/src/experimental_table_slice.cpp
@@ -139,11 +139,8 @@ template <class TypeClass>
 struct decodable<integer_type, arrow::NumericArray<TypeClass>>
   : std::true_type {};
 
-template <class TypeClass>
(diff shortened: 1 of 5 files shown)


tobim (approved):
I'm confident that the tests would catch any issues with this change.

PR #2019, opened 2022-01-03
Introduce the `#import_time` meta extractor

dominiklohmann:
This introduces the `#import_time` meta extractor to the query language. Use it to query for data based on the timestamp that the data arrived at the VAST server process, e.g.,  `vast export json '#import_time > 1 hour ago'` will export all data that was added in the last hour as NDJSON.

The new extractor behaves like a timestamp in the query language and can be used for comparisons with [time value literals](https://docs.tenzir.com/vast/query-language/values/#time).

### :memo: Checklist

Review commit-by-commit. Test extensively locally. Previous PRs added unit test coverage, so this only adds integration test coverage now that it is exposed to the user.


dispanser (approved):
Verified locally that queries using `#import_time` work as expected.

PR #2020, opened 2022-01-05
Change transform_step API to a builder pattern

6yozo:
Change transform_step API to a builder pattern.

### :memo: Checklist

Review this pull request commit-by-commit


libvast/src/transform.cpp
```diff
@@ -109,57 +111,77 @@ transformation_engine::transformation_engine(std::vector<transform>&& transforms
...
+  return caf::none;
+}
+
+/// Apply relevant transformations to the table slice.
+caf::expected<std::vector<table_slice>> transformation_engine::finish() {
+  // NOTE: It's important that `batch` is kept alive until `create()`
+  // is finished: If a copy was made, `batch` will hold the only reference
+  // to its underlying table slice, but the RecordBatches created by the
+  // transform step will most likely reference some of same underlying data.
+  std::vector<std::shared_ptr<arrow::RecordBatch>> keep_alive{};
```
6yozo, line 152: @dominiklohmann, @lava : Is this keep_alive variable useful and sufficient in this case?

dominiklohmann: I don't think we need this, actually.

Given `auto batch = as_record_batch(slice)` the deleter of `batch` should hold reference to both `slice` and the actually underlying record batch of `slice`, such that the lifetimes between `batch` and `slice` are properly shared. The existence of the batch thus implies that its underlying data is still present.

lava: Did this change in the meantime since we implemented transforms?

dominiklohmann: Aside from a bug fix, no, I don't think so.

lava: Ok, I was asking because we added the comment and implemented the keepalive workaround after observing actual segfaults when using arrow transforms. So I don't really understand the difference, ie. why it was necessary in the old version but not anymore after the rewrite.

dominiklohmann: Do you still have the context for that segfault? I can't imagine why this would've caused it.

6yozo: The `keep_alive` vector has a beneficial effect. The integration test will fail when the following lines are commented out:

```
    for (auto& [rows, offset, layout, batch] : bq)
      keep_alive.push_back(batch);
```

integration test:

```
sh build/vast/integration-Debug.sh -t transforms$ -v DEBUG
```

dominiklohmann: But _why_ do we need it? There's a conflict between the actual reason and my mental model of the ownership, and I worry that we have a bug that's somewhere deeper in VAST that we yet need to understand.

lava: As far as I understand it's exactly for the reason that's described in the comment: When we create a new arrow record batch using `as_record_batch(slice)`, this slice will extend the lifetime of the underlying table slice, with which it may share some memory.

```
// table_slice.cpp
        auto result = std::shared_ptr<arrow::RecordBatch>{
          data,
          [batch = std::move(batch), slice](arrow::RecordBatch*) noexcept {
            static_cast<void>(batch);
            static_cast<void>(slice);
          }};
```

When during some transformation the batch is discarded and a new one is created as result, the table slice loses its last reference count, gets destroyed, and the batch (which still may share some of the memory) points into an unmapped memory region.

dominiklohmann: So let me summarize my understanding:

We have table slice A and record batch B created using `B = as_record_batch(A)`. B references memory owned by A; A and B extend each other's lifetime so the memory is only free'd when both the reference count of A and B equal 0.

The issue occurs when we create a new record batch C that structurally shares memory with B. If A and B drop their reference count to 0 then C may point into free'd memory.

There are two possible solutions here:
1. Manually ensure that A or B is kept alive as long as C is alive.
2. Tie the lifetime of C to either A or B.

We're currently doing (1). That means that if we are ever to return C to an outside context we need to manually ensure that A or B are still alive as long as we use C.

I propose that we switch to doing (2) instead, which causes C's refcount to actually be meaningful. Owning C should imply that its underlying memory is still alive. Imagine what happens if a plugin developer stores C for later use: the obvious assumption would be that this is a valid thing to do, but it most certainly is not. I think the lifetime extension can easily be implemented when invoking the transform step so the plugin developer does not need to know about this.

libvast/vast/transform_step.hpp
```diff
@@ -13,55 +13,33 @@
...
 
+#include <queue>
+
 namespace vast {
 
+using arrow_batch
+  = std::tuple<vast::id, vast::type, std::shared_ptr<arrow::RecordBatch>>;
+using batch_queue = std::deque<arrow_batch>;
+using slice_queue = std::deque<table_slice>;
+using batch_vector = std::vector<arrow_batch>;
```
dominiklohmann, line 24: These aliases should not be made globally available in the namespace `vast`. They are unintuitive and hinder readability.

I would not expect an `arrow_batch` to be anything but an `arrow::RecordBatch`, and the name does not carry that information.

libvast/vast/transform.hpp
```diff
@@ -64,17 +87,29 @@ class transformation_engine {
   explicit transformation_engine(std::vector<transform>&&);
 
   /// Apply relevant transformations to the table slice.
-  caf::expected<table_slice> apply(table_slice&&) const;
+  caf::error add(table_slice&&);
+  caf::expected<std::vector<table_slice>> finish();
```
dominiklohmann, line 91: Missing API docs.

libvast/vast/transform.hpp
```diff
@@ -9,9 +9,22 @@
...
 
+class offset_range {
+public:
+  void add(vast::id offset, table_slice::size_type rows);
+  bool contains(vast::id offset, table_slice::size_type rows);
+  void clear();
+
+private:
+  std::vector<std::pair<vast::id, table_slice::size_type>> ranges_;
+};
```
dominiklohmann, line 26: Conceptually you probably want to use a `std::vector<id_range>` instead of this custom construct.

libvast/vast/transform.hpp
```diff
@@ -26,15 +39,23 @@ class transform {
 
   void add_step(transform_step_ptr step);
 
-  caf::expected<table_slice> apply(table_slice&&) const;
+  // FIXME: Convinence function
+  [[nodiscard]] caf::error add_slice(table_slice&&);
+  // FIXME: Convinence function
+  // The result is not ordered by offset
+  [[nodiscard]] caf::expected<std::vector<table_slice>> finish_slice();
```
dominiklohmann, line 46: Please provide API docs.

Why do we need the `_slice` suffix for these member functions? If the non-suffixed version is private, shouldn't the naming rather be the other way around?

libvast/vast/transform.hpp
```diff
@@ -26,15 +39,23 @@ class transform {
...
 
 private:
-  [[nodiscard]] std::pair<vast::type, std::shared_ptr<arrow::RecordBatch>>
-  apply(vast::type layout, std::shared_ptr<arrow::RecordBatch> batch) const;
+  [[nodiscard]] caf::error add(vast::id offset, vast::type layout,
+                               std::shared_ptr<arrow::RecordBatch> batch);
+
+  [[nodiscard]] caf::expected<batch_vector> finish();
+
+  caf::error process_queue(const transform_step_ptr& step, batch_queue& queue);
```
dominiklohmann, line 58: Please provide API docs.

From the docs it is not immediately clear to me who is owning the batch queue. It looks like the queue is an inout parameter. Should this instead be a member function of the class that actually owns the queue?

libvast/vast/transform.hpp
```diff
@@ -26,15 +39,23 @@ class transform {
...
 
   // Grant access to the transformation engine so it can check the fast path.
   friend class transformation_engine;
```
dominiklohmann, line 61: Per the comment this should no longer be required.

libvast/vast/transform.hpp
```diff
@@ -45,13 +66,15 @@ class transform {
...
-
   /// Triggers for this transform
   std::vector<std::string> event_types_;
+
+  /// Transformed slices
+  batch_queue to_transform_; // FIXME: write doc
+
+  offset_range range_;
+
+  void clear_slices();
```
dominiklohmann, line 77: Missing API docs.

Why would I need to manually clear the slices?

libvast/vast/transform.hpp
```diff
@@ -64,17 +87,29 @@ class transformation_engine {
...
   std::vector<transform> transforms_;
 
   /// event type -> applicable transforms
   std::unordered_map<std::string, std::vector<size_t>> layout_mapping_;
+
+  std::unordered_map<vast::type, slice_queue> to_transform_;
+
+  offset_range range_;
+
+  void clear_slices();
```
dominiklohmann, line 112: Missing API docs.

libvast/vast/transform_step.hpp
```diff
@@ -13,55 +13,33 @@
...
-  /// if neccessary converting the data to a format the step implementation can
-  /// natively handle.
-  [[nodiscard]] caf::expected<table_slice> apply(table_slice&&) const;
-};
-
-using transform_step_ptr = std::unique_ptr<transform_step>;
+  /// Applies the transformation to a record batch with a corresponding vast
+  /// layout.
+  [[nodiscard]] virtual caf::error
+  add(vast::id offset, type layout, std::shared_ptr<arrow::RecordBatch> batch)
```
dominiklohmann, line 35: I still don't understand why the transform step itself needs to know the ID. Conceptually, I think this should be handled in the outer context to reduce the knowledge requirements for the plugin developer.

The only requirement to the transform step should have to be that `finish()` returns at most as many results as the number of calls to `add(slice, layout)` since the last call to `finish()`.

libvast/vast/transform_steps/delete.hpp
```diff
@@ -36,6 +33,8 @@ class delete_step : public generic_transform_step, public arrow_transform_step {
 
   /// The key suffixes of the fields to delete.
   const std::vector<std::string> fields_;
+
+  batch_vector transformed_;
```
dominiklohmann, line 37: Missing API docs.

examples/plugins/transform/transform_plugin.cpp
```diff
@@ -20,31 +20,27 @@ namespace vast::plugins {
...
 public:
   example_transform_step() = default;
 
-  // This handler receives a generic table slice in any encoding. It can be
-  // inefficient to modify table slices on this level, on the other hand
-  // VAST never needs to perform a conversion before calling this transform.
-  [[nodiscard]] caf::expected<table_slice>
-  operator()(table_slice&& slice) const override {
+  /// Applies the transformation to a record batch(arrow encoding) with a
+  /// corresponding vast layout.
```
dominiklohmann, line 28:
```suggestion
  /// Applies the transformation to an Arrow Record Batch with a
  /// corresponding VAST layout.
```

libvast/src/system/exporter.cpp
```diff
@@ -63,8 +63,20 @@ void ship_results(exporter_actor::stateful_pointer<exporter_state> self) {
     st.query_status.cached -= rows;
     st.query_status.requested -= rows;
     st.query_status.shipped += rows;
-    auto transformed = self->state.transformer.apply(std::move(slice));
-    self->anon_send(st.sink, std::move(*transformed));
+    auto add_failed = self->state.transformer.add(std::move(slice));
+    if (add_failed) {
+      VAST_ERROR("add failed {}", add_failed);
```
dominiklohmann, line 68: This error is user-facing, so please use noun-verb-object order, and explain to the user what exactly went wrong.

E.g.,

```c++
VAST_ERROR("exporter failed to apply transformation to {}: {}", slice.layout(), err); 
```

libvast/src/system/exporter.cpp
```diff
@@ -63,8 +63,20 @@ void ship_results(exporter_actor::stateful_pointer<exporter_state> self) {
     st.query_status.cached -= rows;
     st.query_status.requested -= rows;
     st.query_status.shipped += rows;
-    auto transformed = self->state.transformer.apply(std::move(slice));
-    self->anon_send(st.sink, std::move(*transformed));
+    auto add_failed = self->state.transformer.add(std::move(slice));
+    if (add_failed) {
```
dominiklohmann, line 67:
```suggestion
    if (auto err = self->state.transformer.add(std::move(slice))) {
```

This is a common design pattern in our code base.

libvast/src/system/exporter.cpp
```diff
@@ -63,8 +63,20 @@ void ship_results(exporter_actor::stateful_pointer<exporter_state> self) {
...
+    }
+    auto transformed = self->state.transformer.finish();
+    if (!transformed) {
+      VAST_ERROR("discarding data: error in transformation step. {}",
+                 transformed.error());
+      return;
+    }
+    for (auto& t : *transformed) {
+      self->anon_send(st.sink, std::move(t));
+    }
```
dominiklohmann, line 79: We usually omit the braces for single-line scopes.

libvast/src/system/partition_transformer.cpp
```diff
@@ -165,18 +164,27 @@ partition_transformer_actor::behavior_type partition_transformer(
...
+        = self->state.transform
+            ->finish_slice(); // TODO: Call finish after all add in partition.
+      if (!transformeds) {
+        VAST_ERROR("failed to apply transform");
+        return;
+      }
+      for (auto& transformed : *transformeds) {
+        // If the transform is a no-op we may get back the original table slice
+        // that's still mapped as read-only, but in this case we also don't need
+        // to adjust the import time.
```
dominiklohmann, line 182: A better way to check this would be whether the table slice is unique. I'm not sure if we expose that information in the API yet. cc @lava?

libvast/src/system/partition_transformer.cpp
```diff
@@ -165,18 +164,27 @@ partition_transformer_actor::behavior_type partition_transformer(
...
         return;
       }
-      // If the transform is a no-op we may get back the original table slice
-      // that's still mapped as read-only, but in this case we also don't need
-      // to adjust the import time.
-      if (transformed->import_time() != old_import_time)
-        transformed->import_time(old_import_time);
-      self->state.events += transformed->rows();
-      self->state.slices.push_back(std::move(*transformed));
+      auto transformeds
```
dominiklohmann, line 172: I don't think `transformeds` is a word.

libvast/src/system/partition_transformer.cpp
```diff
@@ -165,18 +164,27 @@ partition_transformer_actor::behavior_type partition_transformer(
...
+        = self->state.transform
+            ->finish_slice(); // TODO: Call finish after all add in partition.
+      if (!transformeds) {
+        VAST_ERROR("failed to apply transform");
```
dominiklohmann, line 176: noun-object-verb—think user-facing for error messages.

libvast/src/system/partition_transformer.cpp
```diff
@@ -165,18 +164,27 @@ partition_transformer_actor::behavior_type partition_transformer(
...
-      // If the transform is a no-op we may get back the original table slice
-      // that's still mapped as read-only, but in this case we also don't need
-      // to adjust the import time.
-      if (transformed->import_time() != old_import_time)
-        transformed->import_time(old_import_time);
-      self->state.events += transformed->rows();
-      self->state.slices.push_back(std::move(*transformed));
+      auto transformeds
+        = self->state.transform
+            ->finish_slice(); // TODO: Call finish after all add in partition.
```
dominiklohmann, line 174: What about this TODO?

libvast/src/system/sink.cpp
```diff
@@ -74,13 +74,18 @@ transforming_sink(caf::stateful_actor<sink_state>* self,
         VAST_INFO("{} received first result with a latency of {}",
                   self->state.name, to_string(time_since_flush));
       }
-      auto transformed = self->state.transforms.apply(std::move(slice));
+      auto add_failed = self->state.transforms.add(std::move(slice));
+      if (add_failed) {
+        VAST_ERROR("add failed {}", add_failed);
```
dominiklohmann, line 79: Please make this more user-facing.

libvast/src/system/sink.cpp
```diff
@@ -74,13 +74,18 @@ transforming_sink(caf::stateful_actor<sink_state>* self,
...
+        return;
+      }
+      auto transformed = self->state.transforms.finish();
       if (!transformed) {
         VAST_WARN("discarding slice; error in output transformation: {}",
                   transformed.error());
         return;
       }
-      slice = std::move(*transformed);
+      slice = std::move((*transformed)[0]);
```
dominiklohmann, line 88: This smells like a bug waiting to be introduced to me. This currently works because you know you only get a single slice back, but why not just do a two-step change here:
1. Check if the result was empty
2. Iterate over all slices in the result

libvast/src/system/transformer.cpp
```diff
@@ -39,13 +39,20 @@ transformer_stream_stage_ptr attach_transform_stage(
...
+      auto transformed = self->state.transforms.finish();
       if (!transformed) {
         VAST_ERROR("discarding data: error in transformation step. {}",
                    transformed.error());
         return;
       }
-      out.push(std::move(*transformed));
+      for (auto& t : *transformed) {
+        out.push(std::move(t));
+      }
```
dominiklohmann, line 55: Single-line scope should not have braces

libvast/src/system/transformer.cpp
```diff
@@ -39,13 +39,20 @@ transformer_stream_stage_ptr attach_transform_stage(
         self->send_exit(self, caf::make_error(ec::end_of_input));
         return;
       }
-      auto transformed = self->state.transforms.apply(std::move(x.body));
+      auto add_failed = self->state.transforms.add(std::move(x.body));
+      if (add_failed) {
+        VAST_ERROR("add failed {}", add_failed);
```
dominiklohmann, line 44: Please make the error message user-facing.

libvast/src/system/transformer.cpp
```diff
@@ -39,13 +39,20 @@ transformer_stream_stage_ptr attach_transform_stage(
...
+        return;
```
dominiklohmann, line 45: Other errors above cause the transformer to exit, why is this not the case here?

libvast/src/transform.cpp
```diff
@@ -65,40 +53,100 @@ const std::vector<std::string>& transform::event_types() const {
...
-      == event_types_.end())
-    return std::move(x);
-  // TODO: Use the fast-path overload if all steps are `arrow_transform_step`s.
-  for (const auto& step : steps_) {
-    auto transformed = step->apply(std::move(x));
-    if (!transformed)
-      return transformed.error();
-    x = std::move(*transformed);
+caf::error transform::add_slice(table_slice&& x) {
+  VAST_DEBUG("add slice data to transform {}", name_);
```
dominiklohmann, line 57: noun-verb-object

libvast/src/transform.cpp
```diff
@@ -65,40 +53,100 @@ const std::vector<std::string>& transform::event_types() const {
...
+  range_.add(x.offset(), x.rows());
+  auto batch = as_record_batch(x);
+  return add(x.offset(), x.layout(), batch);
+}
+
+caf::expected<std::vector<table_slice>> transform::finish_slice() {
+  VAST_DEBUG("retrieving result from {} steps of transform {}", steps_.size(),
```
dominiklohmann, line 64: noun-verb-object

libvast/src/transform.cpp
```diff
@@ -65,40 +53,100 @@ const std::vector<std::string>& transform::event_types() const {
...
+
+caf::error
+transform::process_queue(const transform_step_ptr& step, batch_queue& queue) {
+  caf::error failed{};
+  const auto size = queue.size();
+  for (size_t i = 0; i < size; ++i) {
+    auto [offset, layout, batch] = std::move(queue.front());
+    queue.pop_front();
+    // TODO: Add a private method `unchecked_apply()` that the transformation
+    // engine can call directly to avoid repeating the check.
```
dominiklohmann, line 106: Do you still want to do this as part of the PR?

libvast/src/transform.cpp
```diff
@@ -65,40 +53,100 @@ const std::vector<std::string>& transform::event_types() const {
...
+      continue;
+    }
+    auto add_failed = step->add(offset, std::move(layout), std::move(batch));
+    if (add_failed) {
+      failed = add_failed;
+      queue.clear();
+      break;
+    }
+  }
+  // Finish frees up resource inside the plugin.
```
dominiklohmann, line 121: This sounds like something that we should do via RAII. Then you also don't need the `failed` variable and can simply return the error returned from the `step`.

libvast/src/transform.cpp
```diff
@@ -65,40 +53,100 @@ const std::vector<std::string>& transform::event_types() const {
...
+                  std::string{layout.name()})
+        == event_types_.end()) {
+      // The transform does not change slices of unconfigured event types.
+      queue.emplace_back(offset, std::move(layout), std::move(batch));
+      continue;
+    }
+    auto add_failed = step->add(offset, std::move(layout), std::move(batch));
+    if (add_failed) {
+      failed = add_failed;
+      queue.clear();
```
dominiklohmann, line 117: Why do we clear the queue here? Is this behavior documented?

libvast/src/transform.cpp
```diff
@@ -109,57 +157,105 @@ transformation_engine::transformation_engine(std::vector<transform>&& transforms
 }
 
 /// Apply relevant transformations to the table slice.
-caf::expected<table_slice> transformation_engine::apply(table_slice&& x) const {
-  auto offset = x.offset();
-  auto size = x.rows();
+caf::error transformation_engine::add(table_slice&& x) {
+  VAST_DEBUG("transformation engine add");
```
dominiklohmann, line 161: noun-verb-object

libvast/src/transform.cpp
```diff
@@ -109,57 +157,105 @@ transformation_engine::transformation_engine(std::vector<transform>&& transforms
...
+  // Finish frees up resource inside the plugin.
+  auto finished = transform.finish();
+  if (!finished && !failed)
+    failed = std::move(finished.error());
+  if (failed)
+    return failed;
+  for (const auto& b : *finished)
+    queue.push_back(b);
+  return caf::none;
+}
```
dominiklohmann, line 192: This code looks awfully similar to the other `process_queue` function. Why do we need the duplication?

libvast/src/transform.cpp
```diff
@@ -109,57 +157,105 @@ transformation_engine::transformation_engine(std::vector<transform>&& transforms
...
+
+/// Apply relevant transformations to the table slice.
+caf::expected<std::vector<table_slice>> transformation_engine::finish() {
+  VAST_DEBUG("transformation engine finish");
```
dominiklohmann, line 196: noun-verb-object

libvast/src/transform.cpp
```diff
@@ -109,57 +157,105 @@ transformation_engine::transformation_engine(std::vector<transform>&& transforms
...
+  // is finished: If a copy was made, `batch` will hold the only reference
+  // to its underlying table slice, but the RecordBatches created by the
+  // transform step will most likely reference some of same underlying data.
+  std::vector<std::shared_ptr<arrow::RecordBatch>> keep_alive{};
+  std::unordered_map<vast::type, batch_queue> batches{};
+  std::vector<table_slice> result{};
+  auto guard = caf::detail::make_scope_guard([this]() {
+    this->clear_slices();
+  });
+
```
dominiklohmann, line 207: We try to avoid empty lines within functions

libvast/src/transform.cpp
```diff
@@ -109,57 +157,105 @@ transformation_engine::transformation_engine(std::vector<transform>&& transforms
...
+  // NOTE: It's important that `batch` is kept alive until `create()`
+  // is finished: If a copy was made, `batch` will hold the only reference
+  // to its underlying table slice, but the RecordBatches created by the
+  // transform step will most likely reference some of same underlying data.
+  std::vector<std::shared_ptr<arrow::RecordBatch>> keep_alive{};
+  std::unordered_map<vast::type, batch_queue> batches{};
+  std::vector<table_slice> result{};
+  auto guard = caf::detail::make_scope_guard([this]() {
+    this->clear_slices();
+  });
```
dominiklohmann, line 206: Why do we need manual memory management here? Can we use RAII somehow implicitly?

libvast/src/transform.cpp
```diff
@@ -109,57 +157,105 @@ transformation_engine::transformation_engine(std::vector<transform>&& transforms
...
+    auto& bq = batches[layout];
+    for (auto& s : queue) {
+      auto b = as_record_batch(s);
+      bq.emplace_back(s.offset(), layout, b);
+    }
+    queue.clear();
+    for (auto& [offset, layout, batch] : bq)
+      keep_alive.push_back(batch);
+    const auto& indices = matching->second;
+    VAST_INFO("applying {} transforms for received table slice w/ layout {}",
```
dominiklohmann, line 226: noun-verb-object

libvast/src/transform_steps/delete.cpp
```diff
@@ -52,44 +52,34 @@ delete_step::adjust_layout(const vast::type& layout) const {
...
                                        result.status().ToString()));
-  return std::make_pair(std::move(adjusted_layout), result.MoveValueUnsafe());
+  }
+  transformed_.emplace_back(offset, std::move(adjusted_layout),
+                            result.MoveValueUnsafe());
+  return caf::none;
+}
+
+caf::expected<batch_vector> delete_step::finish() {
+  VAST_DEBUG("delete_step finished");
```
dominiklohmann, line 81: noun-verb-object

libvast/src/transform_steps/hash.cpp
```diff
@@ -28,56 +28,17 @@ hash_step::hash_step(const std::string& fieldname, const std::string& out,
...
-  }
-  return builder_ptr->finish();
-}
-
-caf::expected<std::pair<type, std::shared_ptr<arrow::RecordBatch>>>
-hash_step::operator()(type layout,
-                      std::shared_ptr<arrow::RecordBatch> batch) const {
+caf::error hash_step::add(vast::id offset, type layout,
+                          std::shared_ptr<arrow::RecordBatch> batch) {
+  VAST_DEBUG("hash_step add");
```
dominiklohmann, line 33: noun-verb-object

libvast/src/transform_steps/hash.cpp
```diff
@@ -105,7 +68,14 @@ hash_step::operator()(type layout,
...
   auto adjusted_layout = type{*adjusted_layout_rt};
   adjusted_layout.assign_metadata(layout);
-  return std::make_pair(std::move(adjusted_layout), result_batch.ValueOrDie());
+  transformed_.emplace_back(offset, std::move(adjusted_layout),
+                            result_batch.ValueOrDie());
+  return caf::none;
+}
+
+caf::expected<batch_vector> hash_step::finish() {
+  VAST_DEBUG("hash_step finished");
```
dominiklohmann, line 77: noun-verb-object

libvast/src/transform_steps/identity.cpp
```diff
@@ -17,15 +17,16 @@
...
+  transformed_.emplace_back(offset, layout, std::move(batch));
+  return caf::none;
 }
 
-caf::expected<std::pair<vast::type, std::shared_ptr<arrow::RecordBatch>>>
-identity_step::operator()(vast::type layout,
-                          std::shared_ptr<arrow::RecordBatch> batch) const {
-  return std::make_pair(std::move(layout), std::move(batch));
+caf::expected<batch_vector> identity_step::finish() {
+  VAST_DEBUG("identity_step finished");
```
dominiklohmann, line 28: noun-verb-object

libvast/src/transform_steps/project.cpp
```diff
@@ -56,47 +56,35 @@ project_step::adjust_layout(const vast::type& layout) const {
   return std::pair{std::move(adjusted_layout), std::move(flat_index_to_keep)};
 }
 
-caf::expected<table_slice> project_step::operator()(table_slice&& slice) const {
-  // adjust layout
-  const auto& layout = slice.layout();
+caf::error project_step::add(vast::id offset, type layout,
+                             std::shared_ptr<arrow::RecordBatch> batch) {
+  VAST_DEBUG("project_step add");
```
dominiklohmann, line 61: noun-verb-object

libvast/src/transform_steps/replace.cpp
```diff
@@ -25,45 +25,16 @@ replace_step::replace_step(const std::string& fieldname,
   VAST_ASSERT(is_basic(value_));
 }
 
-caf::expected<table_slice> replace_step::operator()(table_slice&& slice) const {
-  const auto& layout = slice.layout();
+caf::error replace_step::add(vast::id offset, type layout,
+                             std::shared_ptr<arrow::RecordBatch> batch) {
+  VAST_DEBUG("replace_step add");
```
dominiklohmann, line 30: noun-verb-object

libvast/src/transform_steps/project.cpp
```diff
@@ -56,47 +56,35 @@ project_step::adjust_layout(const vast::type& layout) const {
...
                                        result.status().ToString()));
-  return std::make_pair(std::move(adjusted_layout), result.MoveValueUnsafe());
+  }
+  transformed_.emplace_back(offset, std::move(adjusted_layout),
+                            result.MoveValueUnsafe());
+  return caf::none;
+}
+
+caf::expected<batch_vector> project_step::finish() {
+  VAST_DEBUG("project_step finished");
```
dominiklohmann, line 86: noun-verb-object

libvast/src/transform_steps/replace.cpp
```diff
@@ -75,31 +46,45 @@ replace_step::operator()(type layout,
...
+  auto field = layout_rt.field(*column_offset);
   auto adjusted_layout_rt = layout_rt.transform(
-    {{*offset,
+    {{*column_offset,
       record_type::assign({{std::string{field.name}, inferred_type}})}});
   VAST_ASSERT(adjusted_layout_rt); // replacing a field cannot fail.
   auto adjusted_layout = type{*adjusted_layout_rt};
   adjusted_layout.assign_metadata(layout);
-  return std::make_pair(std::move(adjusted_layout), std::move(batch));
+
```
dominiklohmann, line 77: Please remove the empty line within the function

libvast/src/transform_steps/replace.cpp
```diff
@@ -75,31 +46,45 @@ replace_step::operator()(type layout,
...
+  transformed_.emplace_back(offset, std::move(adjusted_layout),
+                            std::move(batch));
+  return caf::none;
+}
+
+caf::expected<batch_vector> replace_step::finish() {
+  VAST_DEBUG("replace_step finished");
```
dominiklohmann, line 84: noun-verb-object

libvast/src/transform_steps/select.cpp
```diff
@@ -39,19 +40,36 @@ select_step::select_step(std::string expr) : expression_(caf::no_error) {
   }
 }
 
-caf::expected<table_slice> select_step::operator()(table_slice&& slice) const {
-  if (!expression_)
+caf::error select_step::add(vast::id offset, type layout,
+                            std::shared_ptr<arrow::RecordBatch> batch) {
+  VAST_DEBUG("select_step add");
```
dominiklohmann, line 45: noun-verb-object

libvast/src/transform_steps/select.cpp
```diff
@@ -39,19 +40,36 @@ select_step::select_step(std::string expr) : expression_(caf::no_error) {
...
     return tailored_expr.error();
-  auto new_slice = filter(slice, *tailored_expr);
-  if (new_slice)
-    return *new_slice;
+  }
+  auto new_slice
+    = filter(arrow_table_slice_builder::create(batch, layout), *tailored_expr);
+  if (new_slice) {
+    auto as_batch = as_record_batch(
+      *new_slice); // FIXME: ask: keepalive batch until create is called()
```
dominiklohmann, line 59: Please take care of the FIXME

libvast/src/transform_steps/select.cpp
```diff
@@ -39,19 +40,36 @@ select_step::select_step(std::string expr) : expression_(caf::no_error) {
...
+  auto tailored_expr = tailor(*expression_, layout);
+  if (!tailored_expr) {
+    transformed_.clear();
     return tailored_expr.error();
-  auto new_slice = filter(slice, *tailored_expr);
-  if (new_slice)
-    return *new_slice;
+  }
+  auto new_slice
+    = filter(arrow_table_slice_builder::create(batch, layout), *tailored_expr);
```
dominiklohmann, line 56: Can you add a TODO: comment above this function to replace this with an Arrow-native filter function as soon as we are able to directly evaluate expressions on a record batch?

libvast/test/system/exporter.cpp
```diff
@@ -129,6 +129,7 @@ struct fixture : fixture_base {
 
   void verify(const std::vector<table_slice>& results) {
     auto xs = make_data(results);
+    MESSAGE(xs);
```
dominiklohmann, line 132: This looks like a leftover from debugging.

libvast/test/transform.cpp
```diff
@@ -19,8 +17,17 @@
...
+#include "vast/type.hpp"
 #include "vast/uuid.hpp"
 
+#include <arrow/array/array_base.h>
+#include <arrow/array/array_binary.h>
+#include <arrow/array/data.h>
+
+#include <string_view>
+
+#define SUITE transform
```
dominiklohmann, line 29: This should go above the includes.

libvast/test/transform.cpp
```diff
@@ -122,90 +129,117 @@ struct transforms_fixture {
...
-  REQUIRE_EQUAL(*projected2, expected_slice2);
-  auto not_projected2 = invalid_project_step.apply(vast::table_slice{slice2});
-  REQUIRE_EQUAL(*not_projected2, slice2);
+  REQUIRE_EQUAL(projected->size(), 1ull);
+  REQUIRE_EQUAL(as_table_slice(projected), expected_slice);
+  auto invalid_add_failed = invalid_project_step.add(
+    slice.offset(), slice.layout(), as_record_batch(slice));
+  REQUIRE(!invalid_add_failed);
+  auto not_projected = invalid_project_step.finish();
+  REQUIRE_NOERROR(not_projected); // FIXME: Check it if needed, or REQUIRE_ERROR
```
dominiklohmann, line 179: Please address the FIXME

libvast/test/transform.cpp
```diff
@@ -214,35 +248,46 @@ TEST(transform with multiple steps) {
...
+  CHECK_EQUAL(((*transformed)[0]).at(0, 0), vast::data_view{"xxx"sv});
   auto wrong_layout = vast::type{"stub", testdata_layout};
   wrong_layout.assign_metadata(vast::type{"foo", vast::none_type{}});
   auto builder = vast::factory<vast::table_slice_builder>::make(
     vast::defaults::import::table_slice_type, wrong_layout);
   REQUIRE(builder->add("asdf", "jklo", vast::integer{23}));
   auto wrong_slice = builder->finish();
-  auto not_transformed = transform.apply(std::move(wrong_slice));
+  auto add2_failed = transform.add_slice(std::move(wrong_slice));
+  REQUIRE(!add2_failed); // FIXME: It is not required to fail, or succeed!
```
dominiklohmann, line 269: Please address the FIXME

libvast/test/transform.cpp
```diff
@@ -214,35 +248,46 @@ TEST(transform with multiple steps) {
...
-  CHECK_EQUAL((*not_transformed).at(0, 1), vast::data_view{"jklo"sv});
-  CHECK_EQUAL((*not_transformed).at(0, 2), vast::data{vast::integer{23}});
+    caf::get<vast::record_type>((*not_transformed)[0].layout()).field(2).name,
+    "index");
+  CHECK_EQUAL((*not_transformed)[0].at(0, 0),
+              vast::data_view{"asdf"sv}); // FIXME change
+  CHECK_EQUAL((*not_transformed)[0].at(0, 1),
+              vast::data_view{"jklo"sv}); // FIXME change
+  CHECK_EQUAL((*not_transformed)[0].at(0, 2),
+              vast::data{vast::integer{23}}); // FIXME change
```
dominiklohmann, line 290: Please address the three FIXMEs

libvast/test/transform.cpp
```diff
@@ -279,10 +329,15 @@ TEST(transformation engine - multiple matching transforms) {
...
+  auto transformed = engine.finish();
   REQUIRE_NOERROR(transformed);
-  CHECK_EQUAL(caf::get<vast::record_type>(transformed->layout()).num_fields(),
-              1ull);
+  REQUIRE_EQUAL(transformed->size(), 1ull);
+  CHECK_EQUAL(
+    caf::get<vast::record_type>((*transformed)[0].layout()).num_fields(), 1ull);
 }
 
+// FIXME: Test transformation_engine / transform with msgpack
```
dominiklohmann, line 341: I think this can be a TODO rather than a FIXME.

libvast/vast/transform_step.hpp
```diff
@@ -13,55 +13,33 @@
...
-  /// Takes a record batch with a corresponding vast layout and transforms it
-  /// into a new batch with a new layout.
-  //  TODO: When we have implemented a way to recover the `layout` from the
-  //  metadata stored in a record batch, we can drop the record type from
-  //  this function signature.
-  [[nodiscard]] virtual caf::expected<
-    std::pair<type, std::shared_ptr<arrow::RecordBatch>>>
-  operator()(type layout, std::shared_ptr<arrow::RecordBatch> batch) const = 0;
-};
+using transform_step_ptr = std::unique_ptr<transform_step>;
```
dominiklohmann, line 42: This is used in less than 10 places overall, so it probably makes sense to remove the alias and just use the `unique_ptr<transform_step>` directly to more clearly express the ownership to the reader of the code at the various call sites.

libvast/vast/transform_step.hpp
```diff
@@ -13,55 +13,33 @@
...
+  add(vast::id offset, type layout, std::shared_ptr<arrow::RecordBatch> batch)
+    = 0;
 
-// A transform step that operates on a generic table slice.
-// While the implementation is free to look at the encoding of
-// the slice and dispatch to specialized functions, this overload
-// requires the result of the step to be serialized back into
-// a finished table slice, while the variants below can pass around
-// intermediate state between the steps.
-class generic_transform_step : public virtual transform_step {
```
dominiklohmann, line 38: The removal of this warrants both a documentation page update and a changelog entry.

dominiklohmann (changes requested):
This is an absolutely massive change, I'm really glad to see it! 🚀

I've got a lot of feedback inline. Mostly minor stuff, some questions on the API.

Some comments I've repeated a lot:
- Please write log message in noun-verb-object order in active voice and present tense.
- For non-debug and non-trace logs, write log messages in a way that the user understands.
- For public API, provide at least a brief summary of every function and variable, and explain the pre- and post-conditions of functions.

libvast/src/transform.cpp
lava: I think it's a good idea to think about how to solve the issue, but it's orthogonal to this PR. The issue is not new, so it's not a regression either.

That said, I also don't really see how to solve this on our side. We'd either need a  `RecordBatch::Copy()` function that we can call in `as_record_batch()` (but that may have severe performance implications), or ideally arrow needs to ensure that record batches which share memory also always share the same destructor. (actually this might be an upstream bug, now that I think about it)

dominiklohmann: I agree that it shouldn't be a blocker for this PR, but we should definitely switch from (1) to (2) at some point in the future and we need a separate story for it so we actually do it. We should design the API to be as friendly to the plugin developer as possible.

libvast/vast/transform.hpp
lava: The transform is operating on its own queue here, but yes, this should be `private`.

libvast/vast/transform_step.hpp
lava: The issue is that you'd essentially have to solve the knapsack problem if you want to check this from outside. Let's say a transform step has input ranges `[0,10), [20,40)`. You get back table slices with sizes 8,9,2 and 10. Is this a valid transform or not?

libvast/src/system/partition_transformer.cpp
lava: We don't, but I'm not sure I see the connection?

libvast/src/system/sink.cpp
lava: Yeah, this was for an intermediate rewrite step. We caught all other instances, but this one seems to have been missing a FIXME.

libvast/src/transform.cpp
lava: No, this is an old comment that was just adopted with the new code.

libvast/vast/transform_step.hpp
dominiklohmann: There are three scenarios here:

- Import: We can move the ID assignment down into the INDEX stream stage and then we don't need to care about them in the transformations.
- Export: We don't need the IDs at all.
- Partition Transform: We can assign an entirely new ID range to the new partition at once.

libvast/src/transform.cpp
dominiklohmann: Alright, was just curious. Resolving this.

libvast/src/transform.cpp
lava: I've created https://app.shortcut.com/tenzir/story/31007/implement-proper-lifetime-extension-for-record-batches-created-from-arrow-table-slices

libvast/vast/transform.hpp
6yozo: The member variable `to_transform_` is the queue. So no need to pass it in as a parameter.

libvast/src/system/transformer.cpp
6yozo: The above is not an error case. Here this is an error case and the transformer skips the slice but continues the transformation with the rest of the slices.

dominiklohmann: If this is not an error, then why does this use `VAST_ERROR` rather than `VAST_WARN`?

6yozo: To be more precise:

```
      if (x.header == detail::stream_control_header::eof) {
        VAST_DEBUG("{} quits after receiving EOF control message in stream",
                   self->state.transformer_name);
        self->send_exit(self, caf::make_error(ec::end_of_input));
        return;
      }
      auto add_failed = self->state.transforms.add(std::move(x.body));
      if (add_failed) {
        VAST_ERROR("add failed {}", add_failed);
        return;
      }
```

The first `if` is not an error case. The second `if` is an error case, in this case the transformer skips the slice but continues the transformation with the rest of the slices. In this situation I believe that `VAST_WARN` can also be used. @lava  What is your opinion?

libvast/vast/transform_step.hpp
dominiklohmann: @lava and I discussed another option, which is to do the re-assigning in the outer context:
- Importer: We can safely assume that there are no gaps in the ID range, so we only need to make sure that the transform engine returned at most as many rows as it got.
- Partition Transformer: Needs to assign a new ID range anyways.
- Exporter: Can safely discard IDs, so no need to do anything.

libvast/src/transform_steps/select.cpp
6yozo: There is a separate story to handle this issue: [SC-31007](https://app.shortcut.com/tenzir/story/31007/implement-proper-lifetime-extension-for-record-batches-created-from-arrow-table-slices).

libvast/vast/transform.hpp
6yozo: One possible sequence of action: `new transform();` `add(slice1);` `add(slice2);` `finish();` `add(slice3);` `finish();` `delete transform`

The transform`add()` has to store information related to the added slices. When `finish()` is called, these resources must be released before starting the next add-finish run.

The `clear_slices()` function is not needed though, it is used only in one place, so I have removed it.

libvast/src/transform.cpp
6yozo: The `queue.clear();` aborts the transform when a `transform_step.add()` fails.

libvast/src/transform.cpp
6yozo: The `partition_transform` uses the `transform` directly, not via the `transformation_engine`, because of this `transform` has to provide a similar interface to `transformation_engine`.

libvast/src/transform.cpp
6yozo: This is the same problem as before:

One possible sequence of action: `new transform();` `add(slice1);` `add(slice2);` `finish();` `add(slice3);` `finish();` `delete transform`

The `transform.add()` has to store information related to the added slices. When `finish()` is called, these resources must be released before starting the next add-finish run.

lava (approved):
Great work, this was much bigger than anticipated.

The only thing missing now is the changelog entry, apart from this it looks good to me and we can already start building on it.

dominiklohmann (approved):
🚀

Please merge alongside the docs update.

PR #2022, opened 2022-01-06
Fix the arrow export when using experimental table slice impl

dispanser:
The `vast export arrow` unnecessarily recreates Arrow schema from the VAST schema, even though there's already a `RecordBatch` available that contains the required schema. When changing the representation for the `duration` data type in the new experimental table slice implementation, our existing arrow schema and the new arrow schema started to diverge, and this re-creation of the Arrow schema actually produces the wrong schema (the one for our regular arrow format). 

In this PR, we just use the existing Arrow schema attached to the `RecordBatch`, and avoid the problem of having to know the underlying table slice implementation alltogether.

### :memo: Checklist

Verify that the new integration test actually writes exerimental table slices and successfully exports them to arrow.

### DIFF
--- a/libvast/src/format/arrow.cpp
+++ b/libvast/src/format/arrow.cpp
@@ -19,6 +19,7 @@
 #include "vast/table_slice_builder.hpp"
 #include "vast/type.hpp"
 
+#include <arrow/api.h>
 #include <arrow/util/config.h>
 #include <caf/none.hpp>
 
@@ -47,10 +48,12 @@ writer::~writer() {
 caf::error writer::write(const table_slice& slice) {
   if (out_ == nullptr)
     return caf::make_error(ec::logic_error, "invalid arrow output stream");
-  if (!this->layout(slice.layout()))
-    return caf::make_error(ec::logic_error, "failed to update layout");
-  // Get the Record Batch and print it.
   auto batch = as_record_batch(slice);
+  if (const auto& layout = slice.layout(); current_layout_ != layout) {
+    if (!this->layout(batch->schema()))
+      return caf::make_error(ec::logic_error, "failed to update layout");
+    current_layout_ = layout;
+  }
   VAST_ASSERT(batch != nullptr);
   if (auto status = current_batch_writer_->WriteRecordBatch(*batch);
       !status.ok())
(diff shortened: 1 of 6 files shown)


libvast/src/format/arrow.cpp
```diff
@@ -19,6 +19,7 @@
 #include "vast/table_slice_builder.hpp"
 #include "vast/type.hpp"
 
+#include <arrow/api.h>
 #include <arrow/util/config.h>
```
dominiklohmann, line 23: Is this include still required with the removed #ifdef below?

vast/integration/vast_integration_suite.yaml
```diff
@@ -376,6 +376,19 @@ tests:
         input: data/suricata/eve.json
       - command: -N export arrow '#type == "suricata.http"'
         transformation: python @./misc/scripts/print-arrow.py
+  # arrow import / export roundtrip based on experimental table slices.
+  Arrow Experimental:
+    condition: version | jq -e '."Apache Arrow"'
+    tags: [export, arrow]
```
dominiklohmann, line 382:
```suggestion
    tags: [import, export, arrow]
```

libvast/test/format/arrow.cpp
```diff
@@ -56,7 +56,7 @@ TEST(arrow batch) {
   for (auto& slice : zeek_conn_log)
     writer.write(slice);
   // Cause the writer to close its current Arrow writer by switching the layout.
-  writer.layout(type{"stub", record_type{{"irrelevant", none_type{}}}});
+  REQUIRE(writer.layout(::arrow::schema({})));
```
dominiklohmann, line 59: Have empty schemas been allowed for a while now? They used to be the main reasons for us disallowing empty record types. Maybe we need to reconsider that restriction (cc @tobim).

dominiklohmann (approved):
The code change makes sense, and is 100% covered by CI. I left some minor comments inline.

libvast/src/format/arrow.cpp
dispanser: My mildly insufficient understanding of C++ makes me think that without that import, C++ only sees a forward declaration of the `RecordBatch`, which is fixed by importing this header. Maybe there's a better solution?

```
/home/pi/src/github/tenzir/vast/libvast/src/format/arrow.cpp:52:28: error: invalid use of incomplete type ‘using element_type = class arrow::RecordBatch’ {aka ‘class arrow::RecordBatch’}
```

libvast/test/format/arrow.cpp
dispanser: I actually saw complaints about zero columns a few days ago, so I was actually surprised this works. In this particular case, it's just some sentinel to trigger a desired behavior, which is maybe not what we want. Maybe we should do an explicit:

```c++
stream->Close();
```

instead?

libvast/src/format/arrow.cpp
dominiklohmann: The comment is about the `arrow/util/config.h`, not the `arrow/api.h` include. Your understanding for the latter one is correct, though.

libvast/test/format/arrow.cpp
dominiklohmann: Sounds good, might be worth a try.

libvast/src/format/arrow.cpp
dispanser: gotcha, fixed.

libvast/test/format/arrow.cpp
dispanser: done

PR #2024, opened 2022-01-07
Various bugfixes and API changes related to compaction

lava:
This is a set of changes to OSS VAST that are required for or were found while testing the compaction plugin.

### :memo: Checklist

Review commit-by-commit.


libvast/src/system/importer.cpp
```diff
@@ -270,9 +270,11 @@ importer(importer_actor::stateful_pointer<importer_state> self,
     if (self->state.stage) {
       self->state.stage->shutdown();
       self->state.stage->out().push(detail::framed<table_slice>::make_eof());
-      self->state.stage->out().force_emit_batches();
-      self->state.stage->out().close();
+      // We need to `fan_out_flush()` before we close, otherwise `close()`
+      // will delete all clean output paths and we might lose the eof.
```
dominiklohmann, line 274: I think this also affects other places in the code. Should we create a function that wraps these three calls in proper order so that we only have a single place for this?

libvast/src/system/meta_index.cpp
```diff
@@ -446,7 +455,11 @@ meta_index(meta_index_actor::stateful_pointer<meta_index_state> self,
                  metrics_metadata{{"query", id_str}});
       self->send(self->state.accountant, "meta-index.lookup.candidates",
                  result.size(), metrics_metadata{{"query", std::move(id_str)}});
-      return result;
+      // TODO: This reproduces the previous behavior, but it is a pessimization:
```
dominiklohmann, line 458: The reader of this code won't necessarily know what the previous behavior was. Please expand the comment.

libvast/src/system/meta_index.cpp
```diff
@@ -405,8 +405,17 @@ meta_index(meta_index_actor::stateful_pointer<meta_index_state> self,
       self->state.erase(old_partition);
       return atom::ok_v;
     },
+    [=](atom::candidates, vast::expression expr,
+        vast::ids ids) -> caf::result<meta_index_result> {
+      auto query = vast::query{};
+      query.expr = std::move(expr);
+      query.ids = std::move(ids);
+      query.id = vast::uuid::nil();
```
dominiklohmann, line 413: If I understand this handler correctly then all it does is to reply to an anonymous query, i.e., a query without an ID. But doesn't it mess up statistics and metrics if we allow that? Why can't we require all queries to have an ID, and simply assign that ID in the constructor of `query` at random?

libvast/src/system/source.cpp
```diff
@@ -172,9 +172,9 @@ source(caf::stateful_actor<source_state>* self, format::reader_ptr reader,
     if (self->state.mgr) {
       self->state.mgr->shutdown();
       self->state.mgr->out().push(detail::framed<table_slice>::make_eof());
-      self->state.mgr->out().force_emit_batches();
-      self->state.mgr->out().close();
       self->state.mgr->out().fan_out_flush();
+      self->state.mgr->out().close();
+      self->state.mgr->out().force_emit_batches();
```
dominiklohmann, line 177: This may need to be copied over to the `datagram_source`, I think.

libvast/vast/system/transformer.hpp
```diff
@@ -47,10 +52,18 @@ struct transformer_state {
...
 /// @param self The actor handle.
+/// @param transforms The set of transforms to be applied.
 transformer_actor::behavior_type
 transformer(transformer_actor::stateful_pointer<transformer_state> self,
             std::string name, std::vector<transform>&&);
 
+/// A transformer actor that is attached to a system component. This behaves
+/// identical to the regular `transformer` except for a slightly different
+/// logic during system shutdown.
+transformer_actor::behavior_type component_transformer(
```
dominiklohmann, line 63: Shouldn't this just be an argument to the `transformer` function above rather than wrapping the function? E.g., something like

```c++
enum class source_requires_shutdown {
  yes,
  no,
};
```

dominiklohmann (commented):
Just some initial thoughts from looking at the code. I will test this in a later review round.

libvast/src/system/meta_index.cpp
lava: I wasn't sure which way would mess up statistics collection more: if we assign a random id then we'd have lots of queries showing up *only* in the meta index but never receive any follow-up action, and it wouldn't be obvious how to filter them. This way they're all combined with a single id and can be ignored.

libvast/src/system/source.cpp
lava: The datagram source doesn't seem to manually `close()` it's connection at all, which seems to avoid all these issues. (it can get away with that because its not used in integration tests)

libvast/src/system/transformer.cpp
```diff
@@ -55,31 +55,68 @@ transformer_stream_stage_ptr attach_transform_stage(
...
+      // have to open a stream slot and `shutdown()` the stream again, or the
+      // importer will be stuck with a pending stream manager and not be
+      // properly cleaned up, also causing a deadlock on shutdown. On the other
+      // hand, some sources will shut down independently of the system shutdown,
+      // e.g. when importing from a file. In this case we may have the opposite
+      // race, where there is still data that the stream manager needs to push
+      // onto the stream, and we must not shutdown in this case since that would
+      // cause the removal of the path upstream. Since we cannot possibly tell
+      // from here which behavior is correct, we require the party that spawns
+      // the transformer to decide what they require.
```
lava, line 108: Note: I tried to make this less of a wall of text by adding some paragraphs, but `clang-format` doesn't like them :/

libvast/vast/system/transformer.hpp
lava: Maybe, I started that way but it seems that `spawn()` doesn't like default arguments, so we would have to have every call site understand the issue and provide the correct additional parameter. Since this is currently only relevant for the importer, this felt a bit like overkill.

dominiklohmann: I don't think specifying the argument at all call sites is overkill—it just makes the choice more explicit. I'd prefer having it visible at all call sites, but my opinion on this is also not too strong, so proceed as you like.

libvast/src/system/meta_index.cpp
tobim: The difference between regular queries and transforms is that the latter is designed to operate on whole partitions (or at least all events of a certain type therein). We initially decided to keep transforms apart from the `query` to reflect that discrepancy at the type level.

I'm not sure this was the correct decision, as there is definitely some value to having an ID to attach to transform metrics.
In any case, I would probably include the type of operation into the metrics data, irregardless of it being a query or not.
I essentially see 2 alternatives here: 
1. Add a `query::transform` as query command.
2. Add an explicit `uuid id` parameter to this and related handlers.
I slightly prefer (2) in this case.

Also some feedback about code structure:
I would move the logic from the `candidates` handler into a function or method on the actor state and pass the contents of a `query` in destructured form to avoid the ad-hoc query creation here.

libvast/src/system/importer.cpp
lava: Sounds reasonable, but I started implementing it and it quickly grew to a size where I think it makes more sense to make a topic branch with a separate review for this.

libvast/src/system/meta_index.cpp
lava: I refactored the code so that the handler now takes an `id` parameter und uses that to provide metrics.

libvast/vast/system/transformer.hpp
lava: I think I'll leave it as is then, the risk that someone passes the wrong argument to me feels greater than the benefit of being explicit here.

dominiklohmann (approved):
Thanks for addressing the feedback. I'm fine with all the changes made here, and they seem to pass all the tests. There should be no performance or behavior changes from anything in this PR so I did not give this an additional test run locally.

PR #2025, opened 2022-01-10
Allow  computing a VAST type from an Arrow schema

dispanser:
This provides the ability to partially recover a VAST type from a
provided Arrow schema.  In the current transformation from VAST to Arrow we lose certain information and cannot recover it. This requires changes in the way VAST layouts are represented in Arrow, and will be worked on in follow-up PRs.

At the moment, the supported functionality is:

- all basic types except enum
- lists
- address (every `FixedBinary(16)` is assumed to be an address)
- subnet (every `FixedBinary(17)` is assumed to be a subnet)
- record types (structs)
- nested records at the field level

Not supported:
- recursive top-level record types (because VAST -> conversion
  "flattens" the schema)
- pattern: cannot be distinguished from `string` type
- maps: cannot be distinguished from a list<structs<key, value>>
- enumeration: appears as `uint64` in Arrow, just like `count`

### :memo: Checklist

Please check that the unit tests cover the basic VAST datatypes. Are we missing anything here?

### DIFF
--- a/libvast/src/experimental_table_slice_builder.cpp
+++ b/libvast/src/experimental_table_slice_builder.cpp
@@ -526,12 +526,7 @@ table_slice experimental_table_slice_builder::finish() {
   auto layout_buffer = builder_.CreateVector(
     reinterpret_cast<const uint8_t*>(layout_bytes.data()), layout_bytes.size());
   // Pack schema.
-#if ARROW_VERSION_MAJOR >= 2
   auto flat_schema = arrow::ipc::SerializeSchema(*schema_).ValueOrDie();
-#else
-  auto flat_schema
-    = arrow::ipc::SerializeSchema(*schema_, nullptr).ValueOrDie();
-#endif
   auto schema_buffer
     = builder_.CreateVector(flat_schema->data(), flat_schema->size());
   // Pack record batch.
@@ -661,21 +656,20 @@ bool experimental_table_slice_builder::add_impl(data_view x) {
 
 std::shared_ptr<arrow::Schema> make_experimental_schema(const type& t) {
   VAST_ASSERT(caf::holds_alternative<record_type>(t));
-  const auto& rt = caf::get<record_type>(t);
+  const auto& rt = flatten(caf::get<record_type>(t));
   std::vector<std::shared_ptr<arrow::Field>> arrow_fields;
   arrow_fields.reserve(rt.num_leaves());
-  for (const auto& [field, index] : rt.leaves()) {
-    arrow_fields.emplace_back(
-      make_experimental_field(rt.key(index), field.type));
+  for (const auto& field : rt.fields()) {
+    arrow_fields.emplace_back(make_experimental_field(field));
   }
   auto metadata = arrow::key_value_metadata({{"name", std::string{t.name()}}});
   return std::make_shared<arrow::Schema>(arrow_fields, metadata);
 }
 
 std::shared_ptr<arrow::Field>
-make_experimental_field(const std::string& name, const type& t) {
-  const auto& arrow_type = make_experimental_type(t);
-  return arrow::field(name, arrow_type);
+make_experimental_field(const record_type::field_view& field) {
+  const auto& arrow_type = make_experimental_type(field.type);
+  return arrow::field(std::string{field.name}, arrow_type);
 }
 
 std::shared_ptr<arrow::DataType> make_experimental_type(const type& t) {
@@ -731,7 +725,8 @@ type make_vast_type(const std::shared_ptr<arrow::DataType>& arrow_type) {
     case arrow::Type::STRING:
       return type{string_type{}};
     case arrow::Type::TIMESTAMP: {
-      const auto& t = std::static_pointer_cast<arrow::TimestampType>(arrow_type);
+      const auto& t
+        = std::static_pointer_cast<arrow::TimestampType>(arrow_type);
       if (t->unit() != arrow::TimeUnit::NANO)
         die(fmt::format("unhandled Arrow type: Timestamp[{}]", t->unit()));
       return type{time_type{}};
@@ -745,9 +740,7 @@ type make_vast_type(const std::shared_ptr<arrow::DataType>& arrow_type) {
         case 17:
           return type{subnet_type{}};
         default:
-          VAST_ASSERT(
-            false,
-            fmt::format("unhandled Arrow type: FIXEDBINARY[{}]", width).c_str());
+          die(fmt::format("unhandled Arrow type: FIXEDBINARY[{}]", width));
       }
       return type{time_type{}};
     }
@@ -759,33 +752,25 @@ type make_vast_type(const std::shared_ptr<arrow::DataType>& arrow_type) {
     case arrow::Type::STRUCT: {
       std::vector<struct record_type::field> field_types;
       field_types.reserve(arrow_type->num_fields());
-      for (const auto& f : arrow_type->fields()) {
-        field_types.emplace_back(f->name(), make_vast_type(*f));
-      }
-
+      for (const auto& f : arrow_type->fields())
+        field_types.emplace_back(make_vast_type(*f));
       return type{record_type{field_types}};
     }
-
     default:
-      VAST_ASSERT(
-        false,
-        fmt::format("unhandled Arrow type: {}", arrow_type->ToString()).c_str());
-      return type{none_type{}};
+      die(fmt::format("unhandled Arrow type: {}", arrow_type->ToString()));
   }
 }
 
 // NOLINTNEXTLINE(misc-no-recursion)
-type make_vast_type(const arrow::Field& arrow_field) {
-  return make_vast_type(arrow_field.type());
+struct record_type::field make_vast_type(const arrow::Field& arrow_field) {
+  return {arrow_field.name(), make_vast_type(arrow_field.type())};
 }
 
 type make_vast_type(const arrow::Schema& arrow_schema) {
   std::vector<struct record_type::field> field_types;
   field_types.reserve(arrow_schema.num_fields());
-  for (const auto& f : arrow_schema.fields()) {
-    field_types.emplace_back(f->name(), make_vast_type(*f));
-  }
-
+  for (const auto& f : arrow_schema.fields())
+    field_types.emplace_back(make_vast_type(*f));
   return type{record_type{field_types}};
 }
 
--- a/libvast/src/experimental_table_slice_builder.cpp
+++ b/libvast/src/experimental_table_slice_builder.cpp
@@ -675,21 +675,21 @@ make_experimental_field(const record_type::field_view& field) {
 std::shared_ptr<arrow::DataType> make_experimental_type(const type& t) {
   using data_type_ptr = std::shared_ptr<arrow::DataType>;
   auto f = detail::overload{
-    [=](const auto& x) -> data_type_ptr {
+    [](const auto& x) -> data_type_ptr {
       using trait = column_builder_trait<std::decay_t<decltype(x)>>;
       return trait::make_arrow_type();
     },
-    [=](const list_type& x) -> data_type_ptr {
+    [](const list_type& x) -> data_type_ptr {
       return arrow::list(make_experimental_type(x.value_type()));
     },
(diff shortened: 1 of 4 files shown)


libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -665,14 +665,46 @@ std::shared_ptr<arrow::Schema> make_experimental_schema(const type& t) {
...
-    arrow_fields.emplace_back(std::move(field_ptr));
+    arrow_fields.emplace_back(
+      make_experimental_field(rt.key(index), field.type));
   }
   auto metadata = arrow::key_value_metadata({{"name", std::string{t.name()}}});
   return std::make_shared<arrow::Schema>(arrow_fields, metadata);
 }
 
+std::shared_ptr<arrow::Field>
+make_experimental_field(const std::string& name, const type& t) {
```
dominiklohmann, line 676: This should work on a `const record_type::field_view& field`  rather than the two listed parameters.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -665,14 +665,46 @@ std::shared_ptr<arrow::Schema> make_experimental_schema(const type& t) {
...
+  using data_type_ptr = std::shared_ptr<arrow::Field>;
+  auto f = detail::overload{
+    [=](const auto& x) -> data_type_ptr {
```
dominiklohmann, line 679: Why does this lambda take the name by value (`[=]`) rather than reference (`[&]`)?

dominiklohmann, line 679: I think the auto overload is quite dangerous. Maybe we should only have this for basic types, e.g.,

```
[&]<basic_type T>(const T& x) -> data_type_ptr { ... }
```

Then the compiler would error for new complex types that an overload is missing instead of silently using this one.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -704,4 +736,74 @@ std::shared_ptr<arrow::DataType> make_experimental_type(const type& t) {
   return caf::visit(f, t);
 }
 
+type make_vast_type(const std::shared_ptr<arrow::DataType>& arrow_type) {
```
dispanser, line 739: clangd complains: 

> Function 'make_vast_type' is within a recursive call chain

While that is correct, it's the nature of the type system to be recursive. This doesn't happen for e.g., `make_experimental_type`, which is also defined recursively.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -704,4 +736,74 @@ std::shared_ptr<arrow::DataType> make_experimental_type(const type& t) {
...
+        field_types.emplace_back(f->name(), make_vast_type(*f));
+      }
+
+      return type{record_type{field_types}};
+    }
+
+    default:
+      VAST_ASSERT(
+        false,
+        fmt::format("unhandled Arrow type: {}", arrow_type->ToString()).c_str());
```
dominiklohmann, line 790: If this always fails then maybe just call `die` instead in order to avoid running into undefined behavior in release builds.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -704,4 +736,74 @@ std::shared_ptr<arrow::DataType> make_experimental_type(const type& t) {
...
+      return type{none_type{}};
+  }
+}
+
+type make_vast_type(const arrow::Field& arrow_field) {
+  return make_vast_type(arrow_field.type());
+}
```
dominiklohmann, line 797:
```suggestion
struct record_type::field make_vast_type(const arrow::Field& arrow_field) {
  return {arrow_field.name(), make_vast_type(arrow_field.type())};
}
```

Otherwise we lose information from the field that we may need to later construct a `record_type` from it.

Note the difference between `record_type::field` and `record_type::field_view`: The former stores the name as a string, the latter only contains a string_view whose ownership is bound to the type. That's why you usually want to pass the latter, and return the former if you create a new one.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -704,4 +736,74 @@ std::shared_ptr<arrow::DataType> make_experimental_type(const type& t) {
...
+
+type make_vast_type(const arrow::Schema& arrow_schema) {
+  std::vector<struct record_type::field> field_types;
+  field_types.reserve(arrow_schema.num_fields());
+  for (const auto& f : arrow_schema.fields()) {
+    field_types.emplace_back(f->name(), make_vast_type(*f));
+  }
```
dominiklohmann, line 804: Coding style nit: We usually don't use braces for single-line scopes.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -704,4 +736,74 @@ std::shared_ptr<arrow::DataType> make_experimental_type(const type& t) {
...
+
```
dominiklohmann, line 805: Coding style nit: No whitespace within a function.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -704,4 +736,74 @@ std::shared_ptr<arrow::DataType> make_experimental_type(const type& t) {
...
+      return type{none_type{}};
+  }
+}
+
+type make_vast_type(const arrow::Field& arrow_field) {
+  return make_vast_type(arrow_field.type());
+}
+
+type make_vast_type(const arrow::Schema& arrow_schema) {
+  std::vector<struct record_type::field> field_types;
```
dominiklohmann, line 800: Do you _need_ to copy the names here, or can you come by with a `field_view`? To me it looks like the latter suffices because the ownership within this function is clear, so the implicit string copy within `emplace_back` 3 lines down is unnecessary.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -704,4 +736,74 @@ std::shared_ptr<arrow::DataType> make_experimental_type(const type& t) {
...
+      return type{integer_type{}};
+    case arrow::Type::UINT64:
+      return type{count_type{}};
+    case arrow::Type::DOUBLE:
+      return type{real_type{}};
+    case arrow::Type::DURATION:
+      return type{duration_type{}};
+    case arrow::Type::STRING:
+      return type{string_type{}};
+    case arrow::Type::TIMESTAMP:
```
dominiklohmann, line 755: If I understood correctly then the Arrow Timestamp data type can have different resolutions from nanoseconds. You probably need to check whether this is a nanosecond-resolution timestamp, and fail otherwise.

The same applies to duration.

dominiklohmann (commented):
Really good progress. I have some initial thoughts on the code only.

libvast/src/experimental_table_slice_builder.cpp
dominiklohmann: You can fix this by adding a `// NOLINTNEXTLINE(misc-no-recursion)` above the function.

I think the recursion itself is fine here. Types already have an enforced maximum depth, so you cannot run into infinite recursion or a stack overflow here realistically.

I don't know why it doesn't diagnose it for `make_experimental_type` as well.

libvast/src/experimental_table_slice_builder.cpp
dispanser: I think that does not currently work, because we're calling it like this:
```c++
  for (const auto& [field, index] : rt.leaves()) {
    arrow_fields.emplace_back(
      make_experimental_field(rt.key(index), field.type));
  }
```

IIUC, the `rt.key(index)` is necessary because we're building the fully qualified name to the leaf field, and `field.name` would not suffice.

I _think_ this should be possible as soon as we properly represent the VAST type in the Arrow schema in terms of the full tree, not a flattened view of leaves attached directly to the root.

libvast/src/experimental_table_slice_builder.cpp
dispanser: good catch, additional inspection of the time unit is necessary, even though for VAST-generated data this should already be the right time unit.

libvast/src/experimental_table_slice_builder.cpp
dominiklohmann: One thing you could do is to call `auto frt = flatten(rt)` once and then iterate over its `frt.fields()`. This would express the flattening more clearly as well.

libvast/src/experimental_table_slice_builder.cpp
dispanser: will do

libvast/src/experimental_table_slice_builder.cpp
dispanser: I'll give that a try. I like the cleaner duality between `make_vast_type` and `make_experimental_field`

libvast/src/experimental_table_slice_builder.cpp
dispanser: done, thanks

libvast/src/experimental_table_slice_builder.cpp
dispanser: resolved by pair programming, thanks!

libvast/vast/experimental_table_slice_builder.hpp
```diff
@@ -134,4 +135,20 @@ std::shared_ptr<arrow::Schema> make_experimental_schema(const type& t);
...
+
+/// Converts an Arrow `Schema` to a VAST `type`.
+/// @param arrow_schema The Arrow schema to convert.
+/// @returns A VAST type representation of `arrow_schema`.
+type make_vast_type(const arrow::Schema& arrow_schema);
+
+/// Converts an Arrow `DataType` to a VAST `type`
+/// @param arrow_type The arrow type to convert.
+/// @return A VAST type representation of `arrow_field`
+type make_vast_type(const std::shared_ptr<arrow::DataType>& arrow_type);
```
dominiklohmann, line 153: Nit: Add one empty line below.

Also, does this function need to get a `const std::shared_ptr<arrow::DataType>&`, or does a `const arrow::DataType&` also suffice?

libvast/test/experimental_table_slice.cpp
```diff
@@ -373,6 +373,75 @@ TEST(record batch roundtrip - adding column) {
...
+    {"a", string_type{}},
+    {"b", record_type{{"hits", count_type{}}, {"net", subnet_type{}}}}}});
+}
+
+auto schema_roundtrip(const type& t) {
+  const auto& arrow_schema = make_experimental_schema(t);
+  const auto& restored_t = make_vast_type(*arrow_schema);
+  if (t != restored_t) { // CHECK_EQUAL doesn't cut it
+    fmt::print(stderr, "`{}` != `{}`\n", t, restored_t);
+    fmt::print(stderr, "arrow schema: {}\n", arrow_schema->ToString(true));
```
dominiklohmann, line 418: Leftover debug prints that should be removed.

libvast/test/experimental_table_slice.cpp
```diff
@@ -373,6 +373,75 @@ TEST(record batch roundtrip - adding column) {
...
+    type{record_type{{"a", string_type{}}, {"b", address_type{}}}});
+  field_roundtrip(type{record_type{
+    {"a", string_type{}},
+    {"b", record_type{{"hits", count_type{}}, {"net", subnet_type{}}}}}});
+}
+
+auto schema_roundtrip(const type& t) {
+  const auto& arrow_schema = make_experimental_schema(t);
+  const auto& restored_t = make_vast_type(*arrow_schema);
+  if (t != restored_t) { // CHECK_EQUAL doesn't cut it
```
dominiklohmann, line 416: Note: For a better debugging experience, you can define this somewhere:

```
void inspect(caf::detail::stringification_inspector& f, const arrow::Schema& x) {
  auto str = x->ToString(true);
  f(str);
}
```

And similar for the other types. This is the code path that `CHECK_EQUAL` takes for printing things in case something goes wrong.

dominiklohmann (approved):
The changes you've made after the earlier review round look convincing to me. I did not do any additional local testing.

libvast/vast/experimental_table_slice_builder.hpp
dispanser: fixed, thanks

PR #2026, opened 2022-01-10
Create a dedicated function to shut down stream stages

lava:
Properly shutting down a caf stream stage is not an easy task, with many ordering details that one can get wrong. This commit introduces a utility function to simplify the issue.

Credits to @dominiklohmann for contributing the `is_specialization_of` template function.

### :memo: Checklist

Review as a whole.

### DIFF
--- a/libvast/src/system/importer.cpp
+++ b/libvast/src/system/importer.cpp
@@ -17,6 +17,7 @@
 #include "vast/concept/printable/vast/error.hpp"
 #include "vast/defaults.hpp"
 #include "vast/detail/fill_status_map.hpp"
+#include "vast/detail/shutdown_stream_stage.hpp"
 #include "vast/error.hpp"
 #include "vast/logger.hpp"
 #include "vast/plugin.hpp"
@@ -268,13 +269,8 @@ importer(importer_actor::stateful_pointer<importer_state> self,
   self->set_exit_handler([=](const caf::exit_msg& msg) {
     self->state.send_report();
     if (self->state.stage) {
-      self->state.stage->shutdown();
       self->state.stage->out().push(detail::framed<table_slice>::make_eof());
-      // We need to `fan_out_flush()` before we close, otherwise `close()`
-      // will delete all clean output paths and we might lose the eof.
-      self->state.stage->out().fan_out_flush();
-      self->state.stage->out().close();
-      self->state.stage->out().force_emit_batches();
+      detail::shutdown_stream_stage(self->state.stage);
       // Spawn a dummy transformer sink. See comment at `dummy_transformer_sink`
       // for reasoning.
       auto dummy = self->spawn(dummy_transformer_sink);
--- a/libvast/vast/detail/shutdown_stream_stage.hpp
+++ b/libvast/vast/detail/shutdown_stream_stage.hpp
@@ -1,26 +1,48 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2021 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
 #pragma once
+
+#include "vast/detail/specialization_of.hpp"
 
 #include <caf/stream_stage.hpp>
 
-#include <type_traits>
+namespace vast::detail {
 
-template <template <class...> class Template, class Instantiation>
-struct is_specialization_of : std::false_type {};
-
-template <template <class...> class Template, class... Args>
-struct is_specialization_of<Template, Template<Args...>> : std::true_type {};
-
-template <typename StreamStage>
-void shutdown_stream_stage(StreamStage& stage) requires(
-  is_specialization_of<caf::stream_stage, StreamStage>::value) {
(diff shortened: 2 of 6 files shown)


libvast/vast/concept/support/detail/is_specialization_of.hpp
```diff
@@ -0,0 +1,23 @@
...
+#pragma once
+
+#include <caf/detail/type_traits.hpp>
+
+#include <type_traits>
+
+namespace vast::detail {
+
+template <template <class...> class Template, class Instantiation>
+struct is_specialization_of : std::false_type {};
```
dominiklohmann, line 18: I think this needs some note that this only works for templates that do not have NTTPs.

Also, can you move this to `vast/detail/is_specialization_of.hpp`?

libvast/vast/concept/support/detail/is_specialization_of.hpp
```diff
@@ -0,0 +1,23 @@
...
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2021 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#pragma once
+
+#include <caf/detail/type_traits.hpp>
```
dominiklohmann, line 11: This include should not be necessary.

libvast/vast/detail/shutdown_stream_stage.hpp
```diff
@@ -1,26 +1,49 @@
...
-  // First we call `shutdown()` to notify all upstream
+/// Properly flushes and shuts down a `caf::stream_stage` connected to
+/// a `caf::broadcast_downstream_manager`.
+template <typename In, typename DownstreamManager>
+void shutdown_stream_stage(
+  caf::stream_stage_ptr<In, DownstreamManager>&
+    stage) requires(is_specialization_of<caf::broadcast_downstream_manager,
+                                         DownstreamManager>::value) {
+  if (!stage)
+    return;
```
dominiklohmann, line 25: Should this function instead take the stream stage by reference?

libvast/vast/detail/shutdown_stream_stage.hpp
```diff
@@ -1,26 +1,49 @@
...
-void shutdown_stream_stage(StreamStage& stage) requires(
-  is_specialization_of<caf::stream_stage, StreamStage>::value) {
-  // First we call `shutdown()` to notify all upstream
+/// Properly flushes and shuts down a `caf::stream_stage` connected to
+/// a `caf::broadcast_downstream_manager`.
+template <typename In, typename DownstreamManager>
+void shutdown_stream_stage(
+  caf::stream_stage_ptr<In, DownstreamManager>&
+    stage) requires(is_specialization_of<caf::broadcast_downstream_manager,
+                                         DownstreamManager>::value) {
```
dominiklohmann, line 23: clang-format does a better job if you put the requires clause after the template instead of here.

libvast/src/system/importer.cpp
```diff
@@ -268,13 +269,8 @@ importer(importer_actor::stateful_pointer<importer_state> self,
...
     self->state.send_report();
     if (self->state.stage) {
-      self->state.stage->shutdown();
       self->state.stage->out().push(detail::framed<table_slice>::make_eof());
-      // We need to `fan_out_flush()` before we close, otherwise `close()`
-      // will delete all clean output paths and we might lose the eof.
-      self->state.stage->out().fan_out_flush();
-      self->state.stage->out().close();
-      self->state.stage->out().force_emit_batches();
+      detail::shutdown_stream_stage(self->state.stage);
```
dominiklohmann, line 273: Does it significantly change the logic that we now effectively call shutdown after push rather than before it?

dominiklohmann (approved):
Perfect, thanks! I left some questions inline.

Also, please move the new trait header, its path looks like you couldn't decide between `concept`, `support`, and `detail` 😅

libvast/vast/detail/shutdown_stream_stage.hpp
lava: I'm not sure, it's always stored as a `stream_stage_ptr` so it seems more natural to just pass that instead of letting every call site check for null.

libvast/src/system/importer.cpp
lava: As far as I can tell it doesn't: `shutdown()` only sends upstream messages but does not touch any of the path buffers.

libvast/vast/detail/shutdown_stream_stage.hpp
dominiklohmann: Just thinking about the contract of the function.

```cpp
// As a developer I know this is a no-op if the stage is a nullptr.
shutdown_stream_stage(self->state.stage);
```

vs.

```cpp
// The precondition that the stage must still be valid is immediately
// visible at the call site.
if (self->state.stage)
  shutdown_stream_stage(*self->state.stage);
```

I don't have a strong preference, but if we have a former than there should be a comment above the function that it is a no-op if the stage is a nullptr.

libvast/src/system/importer.cpp
dominiklohmann: Thanks for double-checking. Fine by me then.

dominiklohmann (approved):

PR #2028, opened 2022-01-11
Simplify naming scheme in `type.fbs`

dominiklohmann:
This is a refactoring that @mavam and I talked about this morning in Slack. Figured I'd do it directly instead of writing the idea down since it was only a set of text replacements.

The general idea here is that instead of using a _vX suffix, we name the union members explicitly, adding a suffix for outdated versions only.

The version of FlatBuffers we depended on when writing most of our FlatBuffers table did not allow for explicitly naming union members, which is part of why we ended up with the existing naming scheme in the first place.

### :memo: Checklist

Review the `type.fbs` changes; everything else is mechanical.


mavam (approved):
A lot nicer to read! 🙏 

Since there were no functional changes, I didn't test this.

PR #2029, opened 2022-01-12
Use `arrow::MapType` to represent `vast::map` in arrow schema

dispanser:
To further align with the arrow ecosystem, we use `arrow::MapType` to more precisely express the semantics of `vast::map`.

### :memo: Checklist

This is hopefully obvious. Verify that a unit test exists that tests the desired table slice behavior (`single column - map`).

### DIFF
--- a/libvast/test/experimental_table_slice.cpp
+++ b/libvast/test/experimental_table_slice.cpp
@@ -407,7 +407,6 @@ TEST(arrow primitive type to field roundtrip) {
   // currently a value of type count, indistinguishable from a normal count
   // field_roundtrip(type{enumeration_type{{"first"}, {"third", 2}, {"fourth"}}});
   field_roundtrip(type{list_type{integer_type{}}});
-  // impossible to distinguish from list_type<stuct<key, value>>:
   field_roundtrip(type{map_type{integer_type{}, address_type{}}});
   field_roundtrip(
     type{record_type{{"key", integer_type{}}, {"value", address_type{}}}});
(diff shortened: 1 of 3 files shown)


libvast/test/experimental_table_slice.cpp
```diff
@@ -400,7 +400,7 @@ TEST(arrow primitive type to field roundtrip) {
   // field_roundtrip(type{enumeration_type{{"first"}, {"third", 2}, {"fourth"}}});
   field_roundtrip(type{list_type{integer_type{}}});
   // impossible to distinguish from list_type<stuct<key, value>>:
-  // field_roundtrip(type{map_type{integer_type{}, address_type{}}});
+  field_roundtrip(type{map_type{integer_type{}, address_type{}}});
```
dominiklohmann, line 403: The comment above can now be removed :)

dominiklohmann (commented):
Can you change the integration test that goes through print-arrow.py to make use of the map type?

dispanser:
> Can you change the integration test that goes through print-arrow.py to make use of the map type?

I briefly looked into it. I couldn't find a place where actually have a map type in use, so (IIUC) I'd have to define a new import type and ingest that before doing any assertions on this.

dominiklohmann:
> I briefly looked into it. I couldn't find a place where actually have a map type in use, so (IIUC) I'd have to define a new import type and ingest that before doing any assertions on this.

Let's pair on this tomorrow, I don't think you've touched that part of VAST yet. Probably easier if I show you quickly so you can get started.

dominiklohmann (approved):
Since `map_type` and `pattern_type` are not currently working in the active indexer we've decided to go ahead with this change as-is.

PR #2030, opened 2022-01-12
Add FlatBuffers for `value_index`

dominiklohmann:
This adds FlatBuffers tables for `value_index` and related classes, and adds unit tests for them. A follow-up PR will integrate this into the partition FlatBuffers table.

### :memo: Checklist

I recommend going over this diff in a call together, as it's the most complicated of the bunch.

If you want to take a peek I recommend looking at the FlatBuffers table first, `value_index.hpp` second, and then going file-by-file through the remaining files.

Note that early on in design we thought we could get by without serializing `value_index::opts_`. That turned out to be a false assumption, as the concrete value index implementation needs to be created before we can dispatch to its unpack implementation. Otherwise, all possible instantiations need to be reflected in the FlatBuffers tables, which is just not feasible.

This PR also fixes a few inefficiencies around coders, which are now non-copyable by design. This found some places where we actually copied coders. These changes are straightforward and can likely be skipped during review.

### DIFF
--- a/libvast/src/index/address_index.cpp
+++ b/libvast/src/index/address_index.cpp
@@ -8,8 +8,10 @@
 
 #include "vast/index/address_index.hpp"
 
+#include "vast/detail/assert.hpp"
 #include "vast/detail/legacy_deserialize.hpp"
 #include "vast/detail/overload.hpp"
+#include "vast/fbs/value_index.hpp"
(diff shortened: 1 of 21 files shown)


tobim (approved):
Went through this in a pair review. There is nothing I can criticise.

PR #2031, opened 2022-01-13
Dont warn for classes with defaulted virtual destructor

lava:
By default, clang-tidy warns if it encounters classes like

    struct A {
      virtual ~A() = default;
    };

requiring that we also define a copy constructor and assignment
operator. I don't see a good reason for this, and since there's
an option to change this behavior let's enable that option.

This also removes an earlier option for `readability-else-after-return`, since we completely disabled that check in an earlier commit.

### :memo: Checklist


6yozo (approved):
The new setting works as expected.

PR #2032, opened 2022-01-14
Start the telemetry loop of the index correctly

dominiklohmann:
If I understood the code before this change directly, then the index set the accountant actor properly upon spawning, and then expected to be sent the accountant actor again to start its telemetry loop—which never happens. This resulted in the telemetry loop never actually being started.

### :memo: Checklist

The change itself is trivial.

I requested @tobim for review because he worked on these metrics recently, and I know that they've been shown on our testbed. When did this break?

### DIFF
--- a/changelog/unreleased/bug-fixes/2032--index-telemetry.md
+++ b/changelog/unreleased/bug-fixes/2032--index-telemetry.md
@@ -1,0 +1,2 @@
+The index now emits the metrics `query.backlog.{low,normal}` and
+`query.workers.{idle,busy}` reliably.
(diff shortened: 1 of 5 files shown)


lava (approved):
Looks correct to me, if I remember correctly the accountant used to be passed as a message everywhere, so I assume this was forgotten during some refactoring.

dominiklohmann:
> if I remember correctly the accountant used to be passed as a message everywhere, so I assume this was forgotten during some refactoring.

This is now only the case for the importer and the legacy archive, all the remaining actors get the handle passed alongside the call to spawn. For the index it looks like this was only done partially.

PR #2034, opened 2022-01-17
Allow multiple independent calls to 'add_message_types()'

lava:
We used to check that all plugin type ids from all prior calls
to `add_message_types()` in the same process are disjunct.

However, this is too strict for unit tests, which will spawn
a fresh actor system for every test in a fixture.

Instead we now only check for overlaps in plugin id spaces
that are present within a single call to `add_message_types()`.

### :memo: Checklist


dominiklohmann (approved):
We paired on this in a call.

PR #2036, opened 2022-01-18
Use correct error codes in memory filesystem

lava:
Use the same error codes in the memory filesystem as for the
"real" filesystem actor, so that code can use the same logic
inside unit tests.

### :memo: Checklist


dominiklohmann (approved):
This is a no-brainer. Thanks for fixing!

PR #2038, opened 2022-01-20
Use Arrow extension type to represent enums in experimental table slices

dispanser:
As part of an aligment between VAST and Arrow data types, VAST enums are stored as an `arrow::DictionaryType` wrapped in a custom `arrow::ExtensionType`, which enables downstream consumers to more naturally consume enums inside Arrow (R, pandas, ...).

### :memo: Checklist

- [X] All user-facing changes have changelog entries.
- [X] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [X] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

Look at the experimental table slice builder first, followed by the associated table slice implementation.

PR #2039, opened 2022-01-20
Use an extension type to represent VAST address types in Arrow

dispanser:
Expose the VAST `address` type as an `ExtensionType` in Arrow (still wrapping the same fixed binary 16 representation under the hood).

### :memo: Checklist

Straight forward.

### DIFF
--- a/libvast/src/arrow_extension_types.cpp
+++ b/libvast/src/arrow_extension_types.cpp
@@ -79,6 +79,38 @@ enumeration_type enum_extension_type::get_enum_type() const {
   return this->enum_type_;
 }
 
+address_extension_type::address_extension_type()
+  : arrow::ExtensionType(arrow::fixed_size_binary(16)) {
+}
+
+std::string address_extension_type::extension_name() const {
+  return id;
+}
+
+bool address_extension_type::ExtensionEquals(const ExtensionType& other) const {
+  return other.extension_name() == this->extension_name();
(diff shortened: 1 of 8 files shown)


dispanser:
@dominiklohmann @tobim : this is ready for review now after the lessons from #2038 have been applied here as well.

tobim (approved):
There is nothing I can complain about in this change.

Just one "peripheral" nit: We could check the arrays `byte_width` when reading from the experimental table slice to match the type with and assertion.

dispanser:
> There is nothing I can complain about in this change.
> 
> Just one "peripheral" nit: We could check the arrays `byte_width` when reading from the experimental table slice to match the type with and assertion.

I'm not sure that is really necessary. If I understand the deserialization process right, Arrow should refuse to build the data type if the underlying physical storage type does not match in:

```c++
arrow::Result<std::shared_ptr<arrow::DataType>>
address_extension_type::Deserialize(std::shared_ptr<DataType> storage_type,
                                    const std::string& serialized) const {
  if (serialized != vast_id)
    return arrow::Status::Invalid("Type identifier did not match");
  if (!storage_type->Equals(this->storage_type_))
    return arrow::Status::Invalid("Storage type did not match "
                                  "fixed_size_binary(16)");
  return std::make_shared<address_extension_type>();
}
```

PR #2040, opened 2022-01-21
Fix record batch lifetime sharing

dominiklohmann:
This change fixes an issue that was rather hard to run into, and far from simple to debug once you did. Consider this example code:

```c++
table_slice A = get_table_slice();
auto B = as_record_batch(A);
auto C = B->Slice(...);
A = {};
B = {};
do_something_with(C); // crashes before this change
```

If C shares any memory with B, the above example crashed before this change. This happens because because only A and B shared ownership of A's underlying buffer, but the internal buffers of B did not. Any operation that created a new record batch C that sliced referenced B's internal buffers did thus break the structural lifetime-sharing with the table slice A.

This commit changes the lifetime-sharing setup and moves it from B to its internal buffers. This works seemlessly with the existing stream decoder when passing Arrow Buffers to it that structurally share the lifetime of the table slice's underlying chunk. To avoid a cyclic reference in the table slice causing a memory leak this additionally decouples the table slice implementation's lifetime from the table slice's lifetime.

### :memo: Checklist

Review carefully file-by-file.

I've verified that this works as expected by applying this change to #2020 locally and removing the `keep_alive` vector in `transformation_engine::finish()`.

Note that the `segment` unit test suite fails if there are any cyclic references in table slices because that will stop segments from being erased. These run through with this change, so we can be sure that the `decoupled_chunk` "hack" introduced by this PR works as designed.

### DIFF
--- a/libvast/src/arrow_table_slice.cpp
+++ b/libvast/src/arrow_table_slice.cpp
@@ -619,21 +619,16 @@ public:
     // nop
   }
 
-  template <class FlatSchema, class FlatRecordBatch>
   std::shared_ptr<arrow::RecordBatch>
-  decode(const FlatSchema& flat_schema,
-         const FlatRecordBatch& flat_record_batch) noexcept {
+  decode(const std::shared_ptr<arrow::Buffer>& flat_schema,
+         const std::shared_ptr<arrow::Buffer>& flat_record_batch) noexcept {
     VAST_ASSERT(!record_batch_);
-    if (auto status
-        = decoder_.Consume(flat_schema->data(), flat_schema->size());
-        !status.ok()) {
+    if (auto status = decoder_.Consume(flat_schema); !status.ok()) {
       VAST_ERROR("{} failed to decode Arrow Schema: {}", __func__,
                  status.ToString());
       return {};
     }
-    if (auto status = decoder_.Consume(flat_record_batch->data(),
-                                       flat_record_batch->size());
-        !status.ok()) {
+    if (auto status = decoder_.Consume(flat_record_batch); !status.ok()) {
       VAST_ERROR("{} failed to decode Arrow Record Batch: {}", __func__,
                  status.ToString());
       return {};
@@ -662,7 +657,9 @@ arrow_table_slice<FlatBuffer>::arrow_table_slice(
       die("failed to deserialize layout: " + render(err));
     state_.layout = type::from_legacy_type(intermediate);
     auto decoder = record_batch_decoder{};
-    state_.record_batch = decoder.decode(slice.schema(), slice.record_batch());
+    state_.record_batch = decoder.decode(
+      as_arrow_buffer(parent->slice(as_bytes(*slice.schema()))),
+      as_arrow_buffer(parent->slice(as_bytes(*slice.record_batch()))));
   } else {
     // We decouple the sliced type from the layout intentionally. This is an
     // absolute must because we store the state in the deletion step of the
@@ -672,7 +669,9 @@ arrow_table_slice<FlatBuffer>::arrow_table_slice(
     state_.layout = type{chunk::copy(as_bytes(*slice_.layout()))};
     VAST_ASSERT(caf::holds_alternative<record_type>(state_.layout));
     auto decoder = record_batch_decoder{};
-    state_.record_batch = decoder.decode(slice.schema(), slice.record_batch());
+    state_.record_batch = decoder.decode(
+      as_arrow_buffer(parent->slice(as_bytes(*slice.schema()))),
+      as_arrow_buffer(parent->slice(as_bytes(*slice.record_batch()))));
   }
 }
 
--- a/libvast/src/chunk.cpp
+++ b/libvast/src/chunk.cpp
@@ -154,11 +154,16 @@ std::shared_ptr<arrow::Buffer> as_arrow_buffer(chunk_ptr chunk) noexcept {
   if (!chunk)
     return nullptr;
   auto buffer = arrow::Buffer::Wrap(chunk->data(), chunk->size());
+  VAST_ASSERT(reinterpret_cast<const std::byte*>(buffer->data())
+              == chunk->data());
   auto* const buffer_data = buffer.get();
-  return {buffer_data, [chunk = std::move(chunk),
-                        buffer = std::move(buffer)](arrow::Buffer*) noexcept {
-            static_cast<void>(chunk);
-            static_cast<void>(buffer);
+  return {buffer_data, [chunk = std::move(chunk), buffer = std::move(buffer)](
+                         arrow::Buffer*) mutable noexcept {
+            // We manually call the destructors in proper order here, as the
+            // chunk must be destroyed last and the destruction order for lambda
+            // captures is undefined.
+            buffer = {};
+            chunk = {};
           }};
 }
 
--- a/libvast/src/chunk.cpp
+++ b/libvast/src/chunk.cpp
@@ -16,6 +16,7 @@
 #include "vast/io/save.hpp"
 #include "vast/logger.hpp"
 
+#include <arrow/buffer.h>
 #include <caf/deserializer.hpp>
 #include <caf/make_counted.hpp>
 #include <caf/serializer.hpp>
@@ -145,6 +146,20 @@ chunk_ptr chunk::slice(view_type view) const {
   return make(view, [this]() noexcept {
     this->deref();
   });
+}
+
+// -- free functions ----------------------------------------------------------
+
+std::shared_ptr<arrow::Buffer> as_arrow_buffer(chunk_ptr chunk) noexcept {
+  if (!chunk)
+    return nullptr;
+  auto buffer = arrow::Buffer::Wrap(chunk->data(), chunk->size());
+  auto* const buffer_data = buffer.get();
+  return {buffer_data, [chunk = std::move(chunk),
+                        buffer = std::move(buffer)](arrow::Buffer*) noexcept {
+            static_cast<void>(chunk);
+            static_cast<void>(buffer);
+          }};
 }
 
 // -- concepts -----------------------------------------------------------------
--- a/libvast/src/experimental_table_slice.cpp
+++ b/libvast/src/experimental_table_slice.cpp
@@ -649,13 +649,10 @@ public:
     // nop
   }
 
-  template <class FlatRecordBatch>
   std::shared_ptr<arrow::RecordBatch>
-  decode(const FlatRecordBatch& flat_record_batch) noexcept {
+  decode(const std::shared_ptr<arrow::Buffer>& flat_record_batch) noexcept {
     VAST_ASSERT(!record_batch_);
-    if (auto status = decoder_.Consume(flat_record_batch->data(),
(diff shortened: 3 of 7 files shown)


libvast/vast/chunk.hpp
```diff
@@ -180,10 +180,16 @@ class chunk final : public caf::ref_counted {
       };
       deleter_ = std::move(g);
     } else {
-      deleter_ = std::move(step);
+      deleter_ = std::forward<Step>(step);
```
dominiklohmann, line 183: Note to the reviewer: While not strictly related to the PR, I noticed that we wrongly used move rather than forward on a forwarding reference here.

libvast/src/chunk.cpp
```diff
@@ -147,6 +148,20 @@ chunk_ptr chunk::slice(view_type view) const {
   });
 }
 
+// -- free functions ----------------------------------------------------------
+
+std::shared_ptr<arrow::Buffer> as_arrow_buffer(chunk_ptr chunk) noexcept {
+  if (!chunk)
+    return nullptr;
+  auto buffer = arrow::Buffer::Wrap(chunk->data(), chunk->size());
```
lava, line 156: I'm a bit confused by the arrow docs, which state that

> The Buffer base class does not own its memory, but subclasses often do.

On the other hand, the API seems to be written in a way that implies that a `Buffer` may own its memory. Since arrow itself provides classes like `PyBuffer` and `NumPyBuffer`, I wonder if the intended usage would be to create a `VastChunkBuffer` subclass?

libvast/src/table_slice.cpp
```diff
@@ -153,12 +153,20 @@ table_slice::table_slice(chunk_ptr&& chunk, enum verify verify) noexcept
       },
       [&](const auto& encoded) noexcept {
         auto& state_ptr = state(encoded, state_);
+        // We create a second chunk with an intentionally decoupled reference
+        // count here that we bind to the lifetime of the original chunk. This
+        // avoids cyclic references between the table slice and its
+        // encoding-specific state.
+        auto decoupled_chunk = chunk::make(as_bytes(*chunk_), []() noexcept {});
```
lava, line 160: Does this warrant it's own `chunk::decouple()` function?

libvast/src/table_slice.cpp
```diff
@@ -153,12 +153,20 @@ table_slice::table_slice(chunk_ptr&& chunk, enum verify verify) noexcept
...
         state_ptr = state.get();
-        chunk_->add_deletion_step([state = std::move(state)]() noexcept {
-          --num_instances_;
-        });
+        chunk_->add_deletion_step(
+          [state = std::move(state), decoupled_chunk
+                                     = std::move(decoupled_chunk)]() noexcept {
+            --num_instances_;
+            static_cast<void>(decoupled_chunk);
+          });
```
lava, line 169: Couldn't it happen that `chunk`s reference count drops to zero and leaves `decoupled_chunk` dangling?

lava (commented):
The logic looks correct to me, although I'm not completely sure yet I understand the `decoupled_chunk` part.

In general I think it would make sense to rebase this on top of #2020 , so we can introduce this fix and remove the things that are fixed by it (ie. the keepalive vectors) in the same commit/PR.

libvast/src/table_slice.cpp
dominiklohmann: I don't think we want to expose that API, it's so easy to misuse. I'd rather not use it here and refactor the table slice dispatch logic to be easier to use instead.

libvast/src/chunk.cpp
dominiklohmann: `arrow::Buffer` is essentially what `vast::chunk` is: A view on a contiguous byte range with a deleter. The `std::shared_ptr` / `caf::intrusive_ptr` wrapping only serves refcounting purposes, and both classes are designed to be never be copied.

For `chunk`, we have the additional restrictions that (1) the memory is read-only, that (2) a sliced chunk does not remember its parents, and that (3) a chunk does type-erasure and we cannot retrieve the typed contiguous byte range it was created from.

The `Buffer` itself never owns memory, but you often work with sub classes of `Buffer` that do, and are then accessed via the base classes interface. One way to implement this function would be to, say, create this `ChunkBuffer` subclass that actually shares the refcount of the `Buffer`. This would serve the same purpose, but then we'd need to implement a bunch more functions from the `Buffer` interface.

libvast/src/table_slice.cpp
dominiklohmann: I'll sketch this out on a piece of paper to think about the possible paths we can take. We definitely need to be careful here.

libvast/src/chunk.cpp
lava: Sounds reasonable. From looking at `arrow/buffer.hpp`  doesn't seem to be doing virtual functions but rather let subclasses control its behavior by setting the protected member functions.

I don't really have a strong opinion on whether a subclass is a good idea or not.

libvast/src/table_slice.cpp
dominiklohmann: You are absolutely correct. I pushed a separate commit that fixes this issue.

libvast/src/table_slice.cpp
```diff
@@ -318,7 +318,7 @@ time table_slice::import_time() const noexcept {
 }
 
 void table_slice::import_time(time import_time) noexcept {
-  VAST_ASSERT(chunk_->unique());
+  VAST_ASSERT(chunk_->get_reference_count());
```
dominiklohmann, line 321: undo

libvast/src/arrow_table_slice.cpp
```diff
@@ -662,7 +657,9 @@ arrow_table_slice<FlatBuffer>::arrow_table_slice(
       die("failed to deserialize layout: " + render(err));
     state_.layout = type::from_legacy_type(intermediate);
     auto decoder = record_batch_decoder{};
-    state_.record_batch = decoder.decode(slice.schema(), slice.record_batch());
+    state_.record_batch = decoder.decode(
+      as_arrow_buffer(parent->slice(as_bytes(*slice.schema()))),
+      as_arrow_buffer(parent->slice(as_bytes(*slice.record_batch()))));
```
dominiklohmann, line 662: TODO: Now that #2020 is merged, rebase on top of recent master and remove the keep alive workaround in the transform engine.

libvast/src/transform.cpp
```diff
@@ -173,7 +173,6 @@ caf::expected<std::vector<table_slice>> transformation_engine::finish() {
   // is finished: If a copy was made, `batch` will hold the only reference
   // to its underlying table slice, but the RecordBatches created by the
   // transform step will most likely reference some of same underlying data.
```
dominiklohmann, line 175: The comment isn't needed anymore either.

lava: Do you still want to remove it in this PR?

dominiklohmann: Yes I should—totally forgot that I left this comment here as a reminder to self. Will fix in a minute.

libvast/src/experimental_table_slice.cpp
```diff
@@ -685,7 +682,8 @@ experimental_table_slice::experimental_table_slice(
   state_.layout = type{chunk::copy(as_bytes(*slice_.layout()))};
   VAST_ASSERT(caf::holds_alternative<record_type>(state_.layout));
   auto decoder = record_batch_decoder{};
-  state_.record_batch = decoder.decode(slice.arrow_ipc());
+  state_.record_batch = decoder.decode(
+    as_arrow_buffer(parent->slice(as_bytes(*slice.arrow_ipc()))));
```
lava, line 686: It looks like there's an overload ` Status Consume(const uint8_t *data, int64_t size)` as well, so I'm wondering if it's necessary to bother with creating a buffer at all? Does the decoded record batch still know about the buffer it was created from?

lava (approved):
My head hurts now, but the logic looks correct to me, with some smaller comments below.

I'm not sure how to test this in general, but I guess if we start seeing lots of random segfaults due to accessing unmapped memory we'll at least know where to start investigating.

libvast/src/experimental_table_slice.cpp
dominiklohmann: That overload calls `arrow::Buffer::Wrap` internally, and then causes the record batch to slice its internal buffer from the result of that, causing the lifetimes not to be shared correctly. We used this API before this PR.

libvast/src/chunk.cpp
```diff
@@ -147,6 +148,20 @@ chunk_ptr chunk::slice(view_type view) const {
...
+  auto* const buffer_data = buffer.get();
```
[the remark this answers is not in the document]
dominiklohmann: Might make sense just to verify assumptions. I'll add that.

libvast/src/chunk.cpp
```diff
@@ -147,6 +148,20 @@ chunk_ptr chunk::slice(view_type view) const {
...
+  return {buffer_data, [chunk = std::move(chunk),
+                        buffer = std::move(buffer)](arrow::Buffer*) noexcept {
+            static_cast<void>(chunk);
+            static_cast<void>(buffer);
```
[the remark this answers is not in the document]
dominiklohmann: Why would the order matter in which we decrement the shared refcounts?

libvast/src/table_slice.cpp
```diff
@@ -156,7 +156,15 @@ table_slice::table_slice(chunk_ptr&& chunk, enum verify verify) noexcept
...
           encoded, chunk_);
         state_ptr = state.get();
-        chunk_->add_deletion_step([state = std::move(state)]() noexcept {
+        const auto bytes = as_bytes(chunk_);
+        // We create a second chunk with an intentionally decoupled reference
+        // count here that we bind to the lifetime of the original chunk. This
+        // avoids cyclic references between the table slice and its
+        // encoding-specific state.
+        chunk_ = chunk::make(bytes, [state = std::move(state),
+                                     chunk = std::move(chunk_)]() noexcept {
```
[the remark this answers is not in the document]
dominiklohmann: Why would the order matter in which we decrement the shared refcounts?

libvast/src/chunk.cpp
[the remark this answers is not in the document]
lava: If this is the last reference to chunk and we clear it first, the buffer will point into empty memory.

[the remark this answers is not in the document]
dominiklohmann: Oh I see now. Will fix by manually destroying in the proper order for both this and the other instance.

libvast/src/table_slice.cpp
```diff
@@ -156,7 +156,15 @@ table_slice::table_slice(chunk_ptr&& chunk, enum verify verify) noexcept
         auto state = std::make_unique<std::decay_t<decltype(*state_ptr)>>(
           encoded, chunk_);
         state_ptr = state.get();
```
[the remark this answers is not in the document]
dominiklohmann: `state_ptr` is a reference into a specific union member for encoding-specific additional state, we need this.

PR #2042, opened 2022-01-24
Represent VAST subnet as arrow extension type

dispanser:
Use an extension type to represent VAST subnet types in Arrow

A subnet will be represented as a struct, the first element, `length`, is
a `uint8` specifying the prefix length. The second element, `address`, is
an existing extension type wrapping a 16-byte fixed width type.

### :memo: Checklist

1. `arrow_extension_types.{hpp,cpp}`
2. `experimental_table_slice_builder.cpp`
3. `experimental_table_slice.cpp`
4. integration test fixtures / unit test coverage

### DIFF
--- a/libvast/src/arrow_extension_types.cpp
+++ b/libvast/src/arrow_extension_types.cpp
@@ -79,6 +79,38 @@ enumeration_type enum_extension_type::get_enum_type() const {
   return this->enum_type_;
 }
 
+address_extension_type::address_extension_type()
+  : arrow::ExtensionType(arrow::fixed_size_binary(16)) {
+}
+
+std::string address_extension_type::extension_name() const {
+  return id;
+}
+
+bool address_extension_type::ExtensionEquals(const ExtensionType& other) const {
+  return other.extension_name() == this->extension_name();
+}
+
+std::shared_ptr<arrow::Array>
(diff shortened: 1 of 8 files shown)


libvast/src/arrow_extension_types.cpp
```diff
@@ -111,18 +111,72 @@ std::string address_extension_type::Serialize() const {
   return id;
 }
 
+const std::string address_extension_type::id = "vast.address";
+
+const std::shared_ptr<arrow::DataType> address_extension_type::arrow_type
+  = arrow::fixed_size_binary(16);
```
tobim, line 117: Why is this addition part of this commit?

dispanser: I learned that I want to unify the shape of these structs, e.g. to be able to make the write a generic round-trip test for all these. So I added this field when adding the second extension type to be able to write that code.

dispanser: @tobim: is there anything else stopping us from going forward with this?

tobim: No, apologies for the delay.

tobim (approved):
I trust the existing test in CI. The real test can only happen once we change the implementation into `arrow:v2`.

PR #2047, opened 2022-01-26
VAST type names and attributes to Arrow mapping

dispanser:
Stores the VAST type name and attribute sets inside the Arrow schema and fields, making it possible to restore the full VAST type information from an Arrow schema.

### :memo: Checklist

Commit by commit.

### DIFF
--- a/libvast/src/experimental_table_slice_builder.cpp
+++ b/libvast/src/experimental_table_slice_builder.cpp
@@ -889,28 +889,23 @@ type enrich_type_with_metadata(
   type t, const std::shared_ptr<const arrow::KeyValueMetadata>& metadata) {
   if (!metadata)
     return t;
-  int nesting_depth = -1;
   auto names_and_attrs = std::vector<
-    std::pair<std::string, struct std::vector<struct type::attribute>>>(1);
-  auto name_parser = "VAST:name:" >> parsers::i32 >> parsers::eoi;
-  auto attr_parser = "VAST:attributes:" >> parsers::i32 >> parsers::eoi;
+    std::pair<std::string, struct std::vector<struct type::attribute>>> {};
+  auto name_parser = "VAST:name:" >> parsers::u32 >> parsers::eoi;
+  auto attr_parser = "VAST:attributes:" >> parsers::u32 >> parsers::eoi;
   for (const auto& [key, value] :
        detail::zip(metadata->keys(), metadata->values())) {
     if (!key.starts_with("VAST:"))
       continue;
-    if (int32_t index{}; name_parser(key, index)) {
-      if (nesting_depth < index) {
-        nesting_depth = index;
+    if (uint32_t index{}; name_parser(key, index)) {
+      if (index >= names_and_attrs.size())
         names_and_attrs.resize(index + 1);
-      }
       names_and_attrs[index].first = value;
       continue;
     }
-    if (int32_t index{}; attr_parser(key, index)) {
-      if (nesting_depth < index) {
-        nesting_depth = index;
+    if (uint32_t index{}; attr_parser(key, index)) {
+      if (index >= names_and_attrs.size())
         names_and_attrs.resize(index + 1);
-      }
       names_and_attrs[index].second = deserialize_attributes(value);
       continue;
     }
--- a/libvast/src/experimental_table_slice_builder.cpp
+++ b/libvast/src/experimental_table_slice_builder.cpp
@@ -9,10 +9,13 @@
 #include "vast/experimental_table_slice_builder.hpp"
 
 #include "vast/arrow_extension_types.hpp"
+#include "vast/concept/parseable/core.hpp"
+#include "vast/concept/parseable/numeric.hpp"
 #include "vast/config.hpp"
 #include "vast/detail/byte_swap.hpp"
 #include "vast/detail/narrow.hpp"
 #include "vast/detail/overload.hpp"
+#include "vast/detail/zip_iterator.hpp"
 #include "vast/die.hpp"
 #include "vast/experimental_table_slice.hpp"
 #include "vast/fbs/table_slice.hpp"
@@ -24,6 +27,7 @@
 #include <arrow/io/api.h>
 #include <arrow/ipc/writer.h>
 
+#include <ranges>
 #include <simdjson.h>
 
 namespace vast {
@@ -701,8 +705,11 @@ deserialize_attributes(std::string_view serialized) {
   std::vector<struct type::attribute> attrs{};
   for (auto f : doc.get_object()) {
     auto value = std::string{f.value.get_string().value()};
-    // if (!f.value.is_string())
-    //   die("hard");
+    if (!f.value.is_string()) {
+      VAST_ERROR("Unexpected Arrow metadata content in json parser: {} -> {}",
+                 f.key, f.value);
+      continue;
+    }
     std::optional<std::string> value_opt
       = value.empty() ? std::optional<std::string>() : value;
     attrs.push_back({std::string{f.key}, value_opt});
@@ -712,8 +719,8 @@ deserialize_attributes(std::string_view serialized) {
 
 std::string
 serialize_attributes(const std::vector<type::attribute_view>& attrs) {
-  auto out = std::string{};
-  auto inserter = std::back_inserter(out);
+  auto result = std::string{};
+  auto inserter = std::back_inserter(result);
   fmt::format_to(inserter, "{{ ");
   bool first = true;
   for (const auto& [key, value] : attrs) {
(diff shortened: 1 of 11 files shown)


libvast/src/type.cpp
```diff
@@ -788,6 +788,53 @@ bool type::has_attributes() const noexcept {
...
+        co_return;
+      case fbs::type::Type::enriched_type: {
+        const auto* enriched_type = root->type_as_enriched_type();
+        std::vector<attribute_view> attrs{};
+        if (const auto* attributes = enriched_type->attributes()) {
+          for (const auto& attribute : *attributes) {
+            if (attribute->value() != nullptr
+                && attribute->value()->begin() != attribute->value()->end())
+              attrs.emplace_back(attribute->key()->string_view(),
+                                 attribute->value()->string_view());
```
dominiklohmann, line 820:
```suggestion
              attrs.push_back({attribute->key()->string_view(),
                               attribute->value()->string_view()});
```

dominiklohmann (commented):
💯 on the `type::names_and_attributes()` function, this is absolutely the correct way to do it.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -809,12 +873,50 @@ type make_vast_type(const arrow::DataType& arrow_type) {
...
+  if (auto ml = metadata->Get("VAST:max_level"); !ml.ok()) {
+    VAST_INFO("skipping non-VAST metadata for field '{}'. Metadata was: '{}'",
+              t.name(), metadata->ToString());
+    return t;
+  } else {
+    auto enriched_type = t;
+    int max_level = stoi(*ml);
+    // working from innermost to outermost nesting
+    for (int l = max_level; l >= 0; --l) {
+      auto name = metadata->Get(fmt::format("VAST:name:{}", l)).ValueOr("");
```
dominiklohmann, line 890: We call `metadata->Get(...)` a total of `1 + 2 * max_level` times. Do you think it's worth switching this to an approach that simply iterates over the metadata instead?

E.g., something like this:

```
auto max_level = metadata->Get(...);
if (max_level == 0)
  return t;
auto names_and_attributes = std::vector<std::pair<std::string_view, struct type::attribute>>{};
names_and_attributes.resize(max_level);
for (const auto& [key, value] : *metadata) {
  if (!value.starts_with("VAST:"))
    continue;
  // parse key and enter into names_and_attributes at parsed index
}
// iterate in reverse to reconstruct type
for (const auto* it = names_and_attributes.rbegin(); it != names_and_attributes.rend(); ++it) {
  t = type{it->first, t, it->second};
}
```

Not 100% sure if we should do this, but the current approach is very allocation heavy with all the formatting going on that we can mostly avoid.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -689,21 +691,80 @@ bool experimental_table_slice_builder::add_impl(data_view x) {
...
+      metadata->Append(fmt::format("VAST:name:{}", nesting_level),
+                       std::string{name});
+    if (!attrs.empty())
+      metadata->Append(fmt::format("VAST:attributes:{}", nesting_level),
+                       serialize_attributes(attrs));
+    ++nesting_level;
+  }
+  if (nesting_level == 0)
+    return nullptr;
+  metadata->Append("VAST:max_level", fmt::format("{}", nesting_level - 1));
```
dominiklohmann, line 745: I propose renaming this to `VAST:nesting-depth`, since nesting depth is both the term used for recursion limits inside both Arrow and VAST.

Not sure on the underscore vs. dashes situation, but we usually use dashes inside config keys.

dominiklohmann (commented):
I like it, this seems to work. I left some thoughts on the diff, and still want to play around with it a bit locally.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -809,12 +874,50 @@ type make_vast_type(const arrow::DataType& arrow_type) {
   return caf::visit(f, arrow_type);
 }
 
+type enrich_type_with_metadata(
+  const type& t,
```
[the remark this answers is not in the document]
dispanser: good point, I'll take care of that

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -746,8 +809,10 @@ std::shared_ptr<arrow::DataType> make_experimental_type(const type& t) {
   return caf::visit(f, t);
 }
 
+namespace {
+
 // NOLINTNEXTLINE(misc-no-recursion)
-type make_vast_type(const arrow::DataType& arrow_type) {
+type make_vast_type_int(const arrow::DataType& arrow_type) {
```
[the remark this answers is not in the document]
dispanser: I was initially facing compiler problems with just `make_vast_type`. Now I renamed it back and everything works, so I'm no longer able to justify the renaming. I'll revert the name change and add some docs as suggested.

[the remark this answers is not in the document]
dominiklohmann: I assume this was short for "internal" then? We usually use `_impl` suffix or a `do_` prefix for those. There's also a neat trick with recursive lambdas that you can use to define the recursive helper function inline, which would probably make this even easier to read.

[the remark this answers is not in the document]
dispanser: I've seen compiler errors (`clangd` complains about "no matching function"...). Renaming back (removing `_int` suffix) works for gcc, but I assume that if `clangd` doesn't accept it, it will probably fail in our CI as well.

I can see more wordy function names like `make_vast_type_from_arrow_datatype` but since this is just a local function, I'm not actually sure that's better.

I'll add comments as suggested.

libvast/src/experimental_table_slice_builder.cpp
dispanser: renamed to `VAST:nesting-depth` and the associated variable to `nesting_depth`.

dominiklohmann (approved):
I've given this a test run locally and looked at the generated metadata. It seems to work nicely on both schemas and fields.

Approving modulo the outstanding review comments.

libvast/src/experimental_table_slice_builder.cpp
dispanser: Good point. I'll look into this a bit. The `Get` is implemented as a linear scan, so this might indeed not be optimal, and iterating over the metadata entries instead is definitely an improvement.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -689,21 +691,81 @@ bool experimental_table_slice_builder::add_impl(data_view x) {
...
+deserialize_attributes(std::string_view serialized) {
+  simdjson::padded_string json{serialized};
+  simdjson::dom::parser parser;
+  auto doc = parser.parse(json);
+  // TODO: use field_view as soon as simdjson::ondemand hits (one less copy)
+  std::vector<struct type::attribute> attrs{};
+  for (auto f : doc.get_object()) {
+    auto value = std::string{f.value.get_string().value()};
+    // if (!f.value.is_string())
+    //   die("hard");
```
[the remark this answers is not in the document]
dispanser: Given that we are more lenient in the case that the `nesting-depth` can't be parsed to an int, I think it would be more consistent to also log an error here, and return the base type, assuming that we're seeing data from a newer VAST version that we don't understand, but at least we can work with the data itself.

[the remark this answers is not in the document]
dispanser: Correcting myself, since this happens inside a single key/value pair in the json, it's maybe even just a warning and we "lost" an attribute, but it's even less severe than returning the base type.

libvast/src/experimental_table_slice_builder.cpp
dispanser: We paired and came up with a more efficient solution along the lines of the code above, using our `parseable` machinery to reduce the allocations and iterating over the metadata only once to avoid the `O(n^2)` that result from the repeated calls to `metadata->Get()`.

dispanser: Revisiting this topic: now that we sequentially scan all metadata fields anyway, the need for `VAST:nesting-depth` is no longer there. 
I think we can get rid of that, which is also more closely what we initially thought of in our design document.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -24,6 +27,9 @@
 #include <arrow/io/api.h>
 #include <arrow/ipc/writer.h>
 
+#include <ranges>
```
dominiklohmann, line 30: This looks like an IWYU leftover.

I recommend disabling it for clangd using `-header-insertion=never`.

dominiklohmann (approved):

PR #2049, opened 2022-01-26
Correctly set release tag for Docker builds

dominiklohmann:
VAST v1.0.0-rc2 was wrongly tagged as v1.0.0-rc1 by this action, which should hopefully be resolved by this changed. We had the same issue before with other jobs in the workflow and this was usually the fix.

This also prepares the repository for an rc3 because we don't want the VAST v1.0.0 release to be wrongly tagged on Dockerhub.

### :memo: Checklist

I don't think this needs a changelog entry since the bug is not user-facing.

Review should be trivial. Sadly we cannot verify that CI does the correct thing until we've pushed a new release tag, hence the rc3 to be pushed once this is merged.


mavam (approved):
Good luck 🙂

PR #2057, opened 2022-01-29
Sum type for `arrow::DataType` and `arrow::Array`

dispanser:
This adds sum type information for `arrow::DataType`, `arrow::Array` and the commonly used `std::shared_ptr<...>` variations of both.

### :memo: Checklist

This is already in use in `libvast/src/experimental_table_slice_builder.cpp:make_vast_type`, so maybe start from there, and follow with the actual implementation of the various `sume_type_access<>` structs. Finally, please verify that the tests cover enough ground. Watch out for unnecessary code repetition, I don't have any experience writing this sort of (metaprogramming) code.


libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -522,6 +522,7 @@ experimental_table_slice_builder::column_builder::make(
       return std::make_unique<map_column_builder>(pool, std::move(key_builder),
                                                   std::move(value_builder));
     },
+    // TODO: this thing doesn't use a pool, all others do; why?
```
dominiklohmann, line 525: I think this should be unified. If we can pass the memory pool to Arrow we should probably do so.

We should also experiment with using `ARROW_DEFAULT_MEMORY_POOL=jemalloc` if we link against jemalloc (it's an optional dependency right now). Can you create a follow-up story for investigation?

libvast/test/arrow_extension_types.cpp
```diff
@@ -66,3 +70,103 @@ TEST(subnet type serde roundtrip) {
...
+template <class... T>
+auto is_type() {
+  return []<class... U>(const U&...) {
+    return (std::is_same_v<T, U> && ...);
+  };
+}
+
+TEST(arrow::DataType sum type) {
+  // Returns a visitor that checks whether the expected concrete types are the
+  // types resulting in the visitation.
```
dominiklohmann, line 83: The comment is out of place, probably should be above the `is_type` helper.

libvast/test/arrow_extension_types.cpp
```diff
@@ -66,3 +70,103 @@ TEST(subnet type serde roundtrip) {
 TEST(pattern type serde roundtrip) {
   serde_roundtrip<vast::pattern_extension_type>();
 }
+
+template <class... T>
+auto is_type() {
+  return []<class... U>(const U&...) {
+    return (std::is_same_v<T, U> && ...);
+  };
+}
```
dominiklohmann, line 79: This needs to be wrapped in an anonymous namespace to avoid ODR problems. Please also check if there are any other free functions that are local to this translation unit that are not yet wrapped in one.

libvast/vast/arrow_extension_types.hpp
```diff
@@ -158,24 +173,350 @@ class pattern_extension_type : public arrow::ExtensionType {
...
+// Arrow are based on the `types_list` for `arrow::Array`, from which the
+// `type_list` for `arrow::DataType` is derived. However, the actual mapping is
+// based on `arrow::DataType`: it's possible to access the underlying datatype
+// from an Array via its `::TypeClass`, but there's no way to go from DataType
+// back to the Array, so there's no way to define `apply` for `arrow::DataType`
+// in terms of indexing for `arrow::Array`.
+// For this machinery to work, we need to create a separate struct extending
+// `arrow::ExtensionArray` for every custom extension type and provide a
+// proper implementation for `::TypeClass`. See `pattern_array` etc.
+
```
dominiklohmann, line 231:
```suggestion
/// Sum type access definitions for `arrow::Array` and `arrow::DataType`
/// alongside their respective `std::shared_ptr` as typically occurring in Apache
/// Arrow are based on the `types_list` for `arrow::Array`, from which the
/// `type_list` for `arrow::DataType` is derived. However, the actual mapping is
/// based on `arrow::DataType`: it's possible to access the underlying datatype
/// from an Array via its `::TypeClass`, but there's no way to go from DataType
/// back to the Array, so there's no way to define `apply` for `arrow::DataType`
/// in terms of indexing for `arrow::Array`.
/// For this machinery to work, we need to create a separate struct extending
/// `arrow::ExtensionArray` for every custom extension type and provide a
/// proper implementation for `::TypeClass`. See `pattern_array` etc.
```

libvast/vast/arrow_extension_types.hpp
```diff
@@ -158,24 +173,350 @@ class pattern_extension_type : public arrow::ExtensionType {
...
+  template <class Result, class Visitor, class... Args>
+  static auto apply(const arrow::Array& x, Visitor&& v, Args&&... xs)
+    -> Result {
+    // A dispatch table that maps variant type index to dispatch function for
+    // the concrete type.
+    static constexpr auto table = []<class... Ts, int... Indices>(
+      detail::type_list<Ts...>,
+      std::integer_sequence<int, Indices...>) noexcept {
+      return std::array{
+        // decay to function pointer
```
dominiklohmann, line 304: This comment only applies to the unary + operator below, which makes this more confusing rather than helping the reader.

The array maps a variant index to a function pointer that does the actual visit dispatch.

libvast/vast/arrow_extension_types.hpp
```diff
@@ -158,24 +173,350 @@ class pattern_extension_type : public arrow::ExtensionType {
...
+  template <class Result, class Visitor, class... Args>
+  static auto apply(const arrow::DataType& x, Visitor&& v, Args&&... xs)
+    -> Result {
+    // A dispatch table that maps variant type index to dispatch function for
+    // the concrete type.
+    static constexpr auto table = []<class... Ts, int... Indices>(
+      detail::type_list<Ts...>,
+      std::integer_sequence<int, Indices...>) noexcept {
+      return std::array{
+        // decay to function pointer
```
dominiklohmann, line 372: Same as above, the comment adds confusion because it isn't clear that it applies to the unary + operator only.

libvast/vast/arrow_extension_types.hpp
```diff
@@ -158,24 +173,350 @@ class pattern_extension_type : public arrow::ExtensionType {
...
+  }
+};
+
+namespace detail {
+template <class Types>
+struct tl_map_shared_ptr;
+template <class... Ts>
+struct tl_map_shared_ptr<detail::type_list<Ts...>> {
+  using type = detail::type_list<std::shared_ptr<Ts>...>;
+};
```
dominiklohmann, line 395:
```suggestion
} // namespace caf

namespace vast::detail {

template <class Types>
struct tl_map_shared_ptr;

template <class... Ts>
struct tl_map_shared_ptr<detail::type_list<Ts...>> {
  using type = detail::type_list<std::shared_ptr<Ts>...>;
};

} // namespace vast::detail

namespace caf {
```

Either that, or put it into `vast/detail/type_traits.hpp` directly with a small comment above it. But don't put anything but a template specialization into another library's namespace.

libvast/vast/arrow_extension_types.hpp
```diff
@@ -158,24 +173,350 @@ class pattern_extension_type : public arrow::ExtensionType {
...
+  }
+
+  template <class T, int Index>
+  static T
+  get(const std::shared_ptr<arrow::Array>& x, sum_type_token<T, Index>) {
+    return std::static_pointer_cast<typename T::element_type>(x);
+  }
+
+  template <class T, int Index>
+  static std::optional<T> get_if(const std::shared_ptr<arrow::Array>* x,
```
dominiklohmann, line 427: I was wondering: Can't this just return a `T`, because `shared_ptr` already is nullable?

The same also applies for the `shared_ptr<DataType>` version below.

libvast/vast/arrow_extension_types.hpp
```diff
@@ -158,24 +173,350 @@ class pattern_extension_type : public arrow::ExtensionType {
...
+  template <class Result, class Visitor, class... Args>
+  static auto
+  apply(const std::shared_ptr<arrow::Array>& x, Visitor&& v, Args&&... xs)
+    -> Result {
+    // A dispatch table that maps variant type index to dispatch function for
+    // the concrete type.
+    static constexpr auto table = []<class... Ts, int... Indices>(
+      detail::type_list<Ts...>,
+      std::integer_sequence<int, Indices...>) noexcept {
+      return std::array{// decay to function pointer
```
dominiklohmann, line 443: Same thing about the comment, plus it really messes up the formatting here.

libvast/vast/arrow_extension_types.hpp
```diff
@@ -158,24 +173,350 @@ class pattern_extension_type : public arrow::ExtensionType {
...
+    if (x && is(*x, token))
+      return get(*x, token);
+    return {};
+  }
+
+  template <class Result, class Visitor, class... Args>
+  static auto
+  apply(const std::shared_ptr<arrow::Array>& x, Visitor&& v, Args&&... xs)
+    -> Result {
+    // A dispatch table that maps variant type index to dispatch function for
```
dominiklohmann, line 438:
```suggestion
    VAST_ASSERT(x);
    // A dispatch table that maps variant type index to dispatch function for
```

libvast/vast/arrow_extension_types.hpp
```diff
@@ -158,24 +173,350 @@ class pattern_extension_type : public arrow::ExtensionType {
...
+    if (x && is(*x, token))
+      return get(*x, token);
+    return {};
+  }
+
+  template <class Result, class Visitor, class... Args>
+  static auto
+  apply(const std::shared_ptr<arrow::DataType>& x, Visitor&& v, Args&&... xs)
+    -> Result {
+    // A dispatch table that maps variant type index to dispatch function for
```
dominiklohmann, line 499:
```suggestion
    VAST_ASSERT(x);
    // A dispatch table that maps variant type index to dispatch function for
```

libvast/vast/arrow_extension_types.hpp
```diff
@@ -158,24 +173,350 @@ class pattern_extension_type : public arrow::ExtensionType {
...
+    // the concrete type.
+    static constexpr auto table = []<class... Ts, int... Indices>(
+      detail::type_list<Ts...>,
+      std::integer_sequence<int, Indices...>) noexcept {
+      return std::array{// decay to function pointer
```
dominiklohmann, line 504: Same thing as above.

libvast/test/arrow_extension_types.cpp
```diff
@@ -66,3 +70,103 @@ TEST(subnet type serde roundtrip) {
...
+    },
+    [](const auto&) {
+      return -1;
+    },
+  };
+  CHECK_EQUAL(caf::visit(f, *str_arr), 101);
+  CHECK_EQUAL(caf::visit(f, pattern_arr), 100);
+  CHECK_EQUAL(caf::visit(f, *addr_arr), 99);
+  CHECK_EQUAL(caf::visit(f, *int_arr), -1);
+  caf::get_if<std::shared_ptr<vast::pattern_array>>(&str_arr);
```
dominiklohmann, line 167: This seems to be missing a check.

libvast/test/arrow_extension_types.cpp
```diff
@@ -66,3 +70,103 @@ TEST(subnet type serde roundtrip) {
...
+  CHECK(caf::visit(is_type<std::shared_ptr<arrow::StringArray>>(), str_arr));
+  CHECK(!caf::visit(is_type<std::shared_ptr<arrow::StringArray>>(), addr_arr));
+  CHECK(!caf::visit(is_type<std::shared_ptr<vast::pattern_array>>(), addr_arr));
+  CHECK(caf::visit(is_type<std::shared_ptr<vast::address_array>>(), addr_arr));
```
dominiklohmann, line 171: Can you also add basic tests for `caf::get_if` for the `shared_ptr` versions?

libvast/test/arrow_extension_types.cpp
```diff
@@ -66,3 +70,103 @@ TEST(subnet type serde roundtrip) {
...
+  CHECK(!caf::get_if<vast::pattern_array>(&*addr_arr));
+  CHECK(caf::get_if<vast::address_array>(&*addr_arr));
+  CHECK(!caf::get_if<vast::address_array>(&pattern_arr));
+  CHECK(caf::get_if<vast::pattern_array>(&pattern_arr));
+  CHECK(caf::get_if<vast::pattern_array>(&pattern_arr));
+
+  caf::visit(is_type<arrow::StringArray>(), *str_arr);
+  caf::visit(is_type<vast::pattern_array>(), pattern_arr);
+  caf::visit(is_type<vast::pattern_array>(), *str_arr);
+
```
dominiklohmann, line 148: Please remove whitespace within the test.

libvast/src/arrow_extension_types.cpp
```diff
@@ -1,10 +1,73 @@
...
+
 #include <simdjson.h>
 
+namespace caf {
+
+int sum_type_access<arrow::Array>::index_from_type(
+  const arrow::DataType& x) noexcept {
+  static constexpr int extension_id = -1;
+  static constexpr int unknown_id = -2;
+  static const auto table = []<class... Ts, int... Indices>(
```
dominiklohmann, line 24:
```suggestion
  // The first-stage O(1) lookup table from arrow::DataType id to the sum type
  // variant index defined by the type list. Returns unknown_id if the DataType
  // is not in the type list, and extension_id if the type is an extension type.
  static const auto table = []<class... Ts, int... Indices>(
```

Modulo formatting.

libvast/src/arrow_extension_types.cpp
```diff
@@ -1,10 +1,73 @@
...
+    tbl.fill(unknown_id);
+    (static_cast<void>(tbl[Ts::type_id] = arrow::is_extension_type<Ts>::value
+                                            ? extension_id
+                                            : Indices),
+     ...);
+    return tbl;
+  }
+  (sum_type_access<arrow::DataType>::types{},
+   std::make_integer_sequence<int, caf::detail::tl_size<data_types>::value>());
+  static const auto extension_table = []<class... Ts, int... Indices>(
```
dominiklohmann, line 37:
```suggestion
  // The second-stage O(n) lookup table for extension types that identifies the
  // types by their unique identifier string.
  static const auto extension_table = []<class... Ts, int... Indices>(
```

dominiklohmann (approved):
This is a great improvement to readability overall, I think. I left some comments inline, but they are all rather small.

I think this will make working with Arrow in our codebase much easier in the future.

libvast/src/experimental_table_slice_builder.cpp
dispanser: addressed (pool added for the builder, ticket created)

libvast/test/arrow_extension_types.cpp
dispanser: done for the entire unit

libvast/test/arrow_extension_types.cpp
dispanser: addressed

libvast/test/arrow_extension_types.cpp
dispanser: done

libvast/vast/arrow_extension_types.hpp
dispanser: We tried to do that but CAF disagrees, it insists on `get_if` to return a pointer to `shared_ptr`.

libvast/vast/arrow_extension_types.hpp
dispanser: moved over to `vast/detail/type_traits.hpp` as suggested.

PR #2063, opened 2022-02-02
Support wildcards in field suffix resolution

dominiklohmann:
NOTE: I felt like hacking on this for an hour this evening, and I got this to work really quickly thanks to our recent type system refactoring. I'll continue with the approach only if this is something that we actually want to consider using, or drop it if we don't. It'd be good to get some feedback on the feature itself and the implementation.

<hr />

This adds support for replacing any part of a field extractor with `*` to jump over a section. E.g., a query `suricata.*.flow.pkts_toserver > 0` will likely hit events of the types `suricata.alert` and `suricata.flow` types. The wildcard only ever jumps over a single section of a key, i.e., `suricata.*.pkts_toserver` and `suricata.*.*.pkts_toserver` will match different fields.

This is implemented generically in the `resolve_key_suffix` function in VAST's type system, which makes the same syntax available to all places that use suffix matching for field names. E.g., the `project` transform can now be configured to keep all fields in `suricata.flow.flow` by specifying `suricata.flow.flow.*`, rather than having to list all fields in it individually.

Note that it does not make sense to implement this for the `resolve_key` function as well, because that is required to return either zero or one results by design, because it returns an exact match.

### :memo: Checklist

`type::resolve_key_suffix` first as its the main piece of changed logic, then the rest. This also includes a simplification for the `#field` meta extractor logic which was able to simply use `type::resolve_key_suffix` now that it's using a generator-based approach.

I still need to write both unit and integration tests.

### DIFF
--- a/libvast/src/concept/parseable/vast/expression.cpp
+++ b/libvast/src/concept/parseable/vast/expression.cpp
@@ -117,6 +117,7 @@ static expression expand(data x) {
 static auto make_predicate_parser() {
   using parsers::alnum;
   using parsers::chr;
+  using parsers::str;
   using namespace parser_literals;
   // clang-format off
   // TODO: Align this with identifier_char.
@@ -124,7 +125,7 @@ static auto make_predicate_parser() {
   // A field cannot start with:
   //  - '-' to leave room for potential arithmetic expressions in operands
   //  - ':' so it won't be interpreted as a type extractor
-  auto field = !(':'_p | '-') >> (+field_char % '.');
+  auto field = !(':'_p | '-') >> ((+field_char | str{"*"}) % '.');
   auto operand
     = (parsers::data >> !(field_char | '.')) ->* to_data_operand
     | "#type"_p  ->* [] { return meta_extractor{meta_extractor::type}; }
--- a/libvast/src/system/meta_index.cpp
+++ b/libvast/src/system/meta_index.cpp
@@ -306,9 +306,18 @@ std::vector<uuid> meta_index_state::lookup_impl(const expression& expr) const {
                 // Compare the desired field name with each field in the
                 // partition.
                 auto matching = [&] {
-                  for (const auto& [fqf, _] : synopsis.second.field_synopses_) {
-                    if (fqf.name().ends_with(*s))
+                  for (const auto& [field, _] :
+                       synopsis.second.field_synopses_) {
+                    if (field.is_standalone_type()) {
+                      if (field.layout_name().ends_with(lhs.field))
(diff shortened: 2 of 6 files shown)


mavam:
This is going to be really helpful!

> The wildcard only ever jumps over a single section of a key, i.e., `suricata.*.pkts_toserver` and `suricata.*.*.pkts_toserver` will match different fields.

I'd like to get a better understanding how we are planning to use "wildcards". Are these actual [glob patterns](https://en.wikipedia.org/wiki/Glob_(programming))? My intuition is "yes" because the alternative would be a regular expression, which is at odds with `.` as field access operator. My hunch is that we need to step through more than one intermediate record and plow through `.` to deliver a good UX.

Related is the concept of making a field extractor unambiguous, i.e., by using single quotes. How does that related to globbing? Since quoting typically means literal interpretation, my expectation would that it turns off wildcarding.

dominiklohmann:
> I'd like to get a better understanding how we are planning to use "wildcards". Are these actual [glob patterns](https://en.wikipedia.org/wiki/Glob_(programming))? My intuition is "yes" because the alternative would be a regular expression, which is at odds with `.` as field access operator. My hunch is that we need to step through more than one intermediate record and plow through `.` to deliver a good UX.

They are not glob patterns. If you have a fully qualified field name with four sections, e.g., module.layout.a.b, you can replace only a full section with the wildcard. The suffix resolution algorithm here is written in a way that it either takes a full match for a section or a wildcard.

> Related is the concept of making a field extractor unambiguous, i.e., by using single quotes. How does that related to globbing?

With the current wildcard approach it is unambiguous. If a field is named `*` then it would also be picked up, it's just that we'd also pick up additional fields alongside it. This can be worked around by using a conjunction with a type extractor.

Also note that we never fully implemented single quotes in field extractors to allow arbitrary characters.

The good thing in general is that the only place that we need to touch to make this logic work closer to normal globbing is the `resolve_key_suffix` function. It is, however, really important to write this code to be as performant as possible, as this function is very much a hotspot for both import and export. Consider this a proof of concept that shows just how easy it is to implement this.

> Since quoting typically means literal interpretation, my expectation would that it turns off wildcarding.

Similarly, this could easily be implemented in `resolve_key_suffix` directly. When handling a section that is in quotes, the wildcard could simply be disabled.

dominiklohmann:
This PR can probably be considered a proof of concept for an upcoming change that simplifies and unifies how VAST handles extractors more generally. I'll open a new PR in the coming weeks that picks this work up again—stay tuned!

PR #2043, opened 2022-01-24
Add a coroutine-enabling wrapper for `caf::result<T>`

dominiklohmann:
This is best explained using a tony table:

<table>
<tr>
<td>BEFORE</td>
<td>AFTER</td>
</tr>
<tr>
<td>

```cpp
[self](T&...) -> caf::result<R> {
  A();
  self->request(d1, t, xs...).then(
    [](response_type& value) {
      B(value);
    },
    [](caf::error& err) {
      C(err);
    });
  D();
  return r;
}
```

</td>
<td>

```cpp
co_lift([self](T...) noexcept -> co_result<R> {
  A();
  auto task = co_request_then(*self, d1, t, xs...);
  D();
  auto tag = co_await co_deliver(r);
  if (auto value = co_await task)
    B(*value);
  else
    C(value.error());
  co_return tag;
})
```

</td>
</tr>
</table>

The main benefit is immediately obvious: We've got rid of callback hell.

There exist a few pitfalls with this that might be easy to run into:
- Using `self->request`, `self->make_response_promise`, and `self->delegate` within a `co_result` coroutine leads to undefined behavior and is inherently unsafe. Use `co_request_then(self, ...)`, `co_request_await(self, ...)`, `co_make_response_promise<T>(self)`, and `co_delegate(self, ...)` instead.
- After awaiting a response handle, `self->current_sender()` might return an unexpected value because the resumed coroutine behaves as if it was called from the response handle continuation.
- To deliver values early, i.e., on first suspend, use `co_deliver(...)`, which returns a tag type that later must be returned from the coroutine.
- CAF supports catching exceptions via `set_exception_handler`. Throwing exceptions from a `co_result<T>` coroutine will always bypass CAF's exception handler and call `std::terminate()`.

#### `co_request_then`

The analog to `self.request(dest, timeout, args...).then(...)` for `co_result<T>` coroutines.

When awaited suspends the coroutine, returns an expected value similar to `caf::function_view`, and continues the coroutine execution in the response handlers CAF runtime context.

#### `co_request_await`

The analog to `self.request(dest, timeout, args...).await(...)` for `co_result<T>` coroutines.

When awaited suspends the coroutine, returns an expected value similar to `caf::function_view`, and continues the coroutine execution in the response handlers CAF runtime context.

#### `co_request_receive`

The analog to `self.request(dest, timeout, args...).receive(...)` for `co_result<T>` coroutines.

[...]

PR #2064, opened 2022-02-03
VAST changes for spatial compaction

lava:
This contains 4 independent changes that are required by the compaction plugin.

### :memo: Checklist


libvast/vast/bloom_filter_synopsis.hpp
```diff
@@ -33,6 +33,14 @@ class bloom_filter_synopsis : public synopsis {
     // nop
   }
 
+  [[nodiscard]] synopsis_ptr clone() const override {
+    using self = bloom_filter_synopsis<T, HashFunction>;
+    // Cannot use `make_unique()` with a private constructor.
+    auto copy = std::unique_ptr<self>(new self(type()));
+    copy->bloom_filter_ = bloom_filter_;
+    return copy;
```
tobim, line 41: Isn't this equivalent?
```suggestion
    return std::make_unique<std::decay_t<decltype(*this)>>(type(),
                                                           bloom_filter_);

```

libvast/vast/bloom_filter_synopsis.hpp
```diff
@@ -94,6 +102,11 @@ class bloom_filter_synopsis : public synopsis {
 
 protected:
   bloom_filter<HashFunction> bloom_filter_;
+
+private:
+  /// Constructor that leaves `bloom_filter_` uninitialized.
+  bloom_filter_synopsis(vast::type x) : synopsis{std::move(x)} {
+  }
```
tobim, line 109: If the suggestion above is correct:
```suggestion
```

libvast/vast/partition_synopsis.hpp
```diff
@@ -12,13 +12,15 @@
...
 #include "vast/table_slice.hpp"
+#include "vast/uuid.hpp"
 
 namespace vast {
 
 /// Contains one synopsis per partition column.
 struct partition_synopsis final {
   partition_synopsis() = default;
   partition_synopsis(partition_synopsis&&) = default;
+  partition_synopsis(const partition_synopsis&);
```
tobim, line 23: I guess you should define copy assignment too.

libvast/vast/system/actors.hpp
```diff
@@ -271,7 +277,8 @@ using index_actor = typed_actor_fwd<
   // the nil uuid.
   // TODO: Add options to do an in-place transform keeping the old ids,
   // and to make a new partition preserving the old one(s).
-  caf::replies_to<atom::apply, transform_ptr, uuid>::with<vast::uuid>,
+  caf::replies_to<atom::apply, transform_ptr,
+                  uuid>::with<partition_synopsis_pair>,
```
tobim, line 281: You could avoid a bit of boilerplate with:
```suggestion
  caf::replies_to<atom::apply, transform_ptr,
                  uuid>::with<uuid, partition_synopsis_ptr>,
```
It's a matter of style, but in this case I think there is no ambiguity.

libvast/src/system/index.cpp
```diff
@@ -1203,30 +1218,38 @@ index(index_actor::stateful_pointer<index_state> self,
       self->send(lookup.worker, atom::supervise_v, query_id, lookup.query,
                  std::move(actors),
                  caf::actor_cast<receiver_actor<atom::done>>(sink));
-      auto rp = self->make_response_promise<vast::uuid>();
+      auto rp = self->make_response_promise<partition_synopsis_pair>();
+      // TODO: Implement some kind of monadic composition instead of these
+      // nested requests.
```
tobim, line 1223: :smile:

libvast/src/system/index.cpp
```diff
@@ -1203,30 +1218,38 @@ index(index_actor::stateful_pointer<index_state> self,
...
                   self->state.partition_synopsis_path(new_partition_id))
         .then(
           [self, rp, old_partition_id, new_partition_id](
             std::shared_ptr<partition_synopsis>& synopsis) mutable {
+            // Make a copy so the caller and the meta index don't need
+            // synchronization.
+            auto result = partition_synopsis_pair{
+              new_partition_id,
+              std::make_shared<partition_synopsis>(*synopsis)};
+
```
tobim, line 1236: Why do you go to such great lengths to avoid the shared ref?

tobim (changes requested):
My main concern is the deep copy of the partition synopsis, I don't understand why this is necessary.

libvast/src/system/index.cpp
lava: It's going to be used by different actors, and thus may be accessed concurrently. (actually we should even copy this once per receiver, now that I think about it) It may be possible to hand out just a `shared_ptr<const partition_synopsis>` here and trust that concurrent reads will be safe, but currently our classes are not really designed for this, so we'd probably at least put a very prominent warning in the docs that implementations of `synopsis` need to be thread-safe.

libvast/vast/system/actors.hpp
dominiklohmann: This won't work with CAF 0.18, which removes this automatic tuple wrapping and unwrapping. Please don't.

libvast/src/system/index.cpp
dominiklohmann: Are any synopses implementations using coders under the hood? Because that has a mutable cache, which I think would certainly be an issue. Other than that, the assumption in modern code should be that const equals thread safe access.

We could also make this a `caf::intrusive_cow_ptr`, which (1) does not need to be an unsafe message type, and (2) has exactly the interface you're looking for here. Accessing a non-const function would automatically clone if the synopsis is not unique.

libvast/src/system/index.cpp
dominiklohmann: #2043 will at least make this a lot easier to read, so maybe we'll just switch to that once we fixed the remaining issues for GCC. :)

libvast/src/system/index.cpp
lava: Sounds good, I'll test if that works.

libvast/src/system/index.cpp
```diff
@@ -570,23 +570,22 @@ void index_state::decomission_active_partition() {
   VAST_DEBUG("{} persists active partition to {}", *self, part_dir);
   self->request(actor, caf::infinite, atom::persist_v, part_dir, synopsis_dir)
     .then(
-      [=, this](std::shared_ptr<partition_synopsis>& ps) {
+      [=, this](partition_synopsis_ptr& ps) {
         VAST_DEBUG("{} successfully persisted partition {}", *self, id);
         // The meta index expects to own the partition synopsis it receives,
         // so we make a copy for the listeners.
-        auto copy = std::make_shared<partition_synopsis>(*ps);
+        // auto copy = std::make_shared<partition_synopsis>(*ps);
```
tobim, line 577: You can remove this and the comment above.

tobim (approved):
I approve of the high-level semantics that are implemented here.

dominiklohmann:
I think this just needs to be adapted to FlatBuffers naming changes on master :)

.clangd
```diff
@@ -0,0 +1,2 @@
+CompileFlags:                     # Tweak the parse settings
+  Add: [-D__cpp_lib_coroutine=201902L]             # treat all files as C++, enable more warnings
```
dominiklohmann, line 2: Why is this in here? This sounds like it'll interfere with a `$HOME/.clangd` file.

changelog/unreleased/features/2064--floating-point-config
```diff
@@ -0,0 +1,2 @@
+Floating point values in configuration files can now also be
+written as integers, without requiring a decimal dot.
```
dominiklohmann, line 2: This file is lacking the .md extension.

libvast/vast/system/actors.hpp
```diff
@@ -519,7 +527,8 @@ CAF_END_TYPE_ID_BLOCK(vast_actors)
 // never be sent over the network.
 #define vast_uuid_synopsis_map std::map<vast::uuid, vast::partition_synopsis>
 CAF_ALLOW_UNSAFE_MESSAGE_TYPE(std::shared_ptr<vast_uuid_synopsis_map>)
-CAF_ALLOW_UNSAFE_MESSAGE_TYPE(std::shared_ptr<vast::partition_synopsis>)
+CAF_ALLOW_UNSAFE_MESSAGE_TYPE(vast::partition_synopsis_ptr)
```
dominiklohmann, line 530: I don't think this needs to be unsafe anymore now that it is an intrusive pointer.

libvast/vast/system/actors.hpp
```diff
@@ -519,7 +527,8 @@ CAF_END_TYPE_ID_BLOCK(vast_actors)
...
+CAF_ALLOW_UNSAFE_MESSAGE_TYPE(vast::partition_synopsis_pair)
```
dominiklohmann, line 531: Similarly, this should not need to be unsafe anymore.

dominiklohmann (approved):
Found some minor stuff that should be fixed before merging.

.clangd
lava: It was committed accidentally, but you shouldn't have a `$HOME/.clangd` file anyways since that is reserved for project-specific settings: https://clangd.llvm.org/config#files

dominiklohmann: `s#$HOME/.clangd#$HOME#.config/clangd/config.yaml` then, but same deal. If anything this should go into a separate PR, but defining something that interferes with the standard library sounds dangerous.

libvast/vast/system/actors.hpp
lava: Yes. (and it makes sense to me, why should caf magically know how to serialize a type just because it's wrapped in an intrusive pointer)

```
/home/benno/src/vast/transforms/libvast/aux/caf/libcaf_core/caf/make_message.hpp:83:74: note: ‘caf::detail::tl_forall<caf::detail::type_list<caf::intrusive_cow_ptr<vast::partition_synopsis> >, caf::is_serializable_or_whitelisted>::value’ evaluates to false
/home/benno/src/vast/transforms/libvast/aux/caf/libcaf_core/caf/make_message.hpp:92:55: error: static assertion failed: at least one type has no type ID: please assign type IDs to all of your types via CAF_ADD_TYPE_ID (this check was enabled via CAF_ENABLE_TYPE_ID_CHECKS)
   92 |   static_assert(tl_forall<stored_types, has_type_id>::value,
```

dominiklohmann: Ah, so that is missing as well. I've written down a note to write an inspector for both partition synopsis and partition synopsis pair sometime in the future, we need this for the CAF 0.18 upgrade.

PR #2065, opened 2022-02-04
Allow fine-grained meta index configuration

lava:
### :memo: Checklist


libvast/src/partition_synopsis.cpp
```diff
@@ -142,7 +183,8 @@ caf::error unpack_(
     synopsis_ptr ptr;
     if (auto error = unpack(*synopsis, ptr))
       return error;
-    if (qf.is_standalone_type())
+    // We mark type-level synopses by using an empty string as name.
+    if (qf.name().empty())
```
dominiklohmann, line 187: This was changed with the type system refactoring; `is_standalone_type` supersedes that hack.

lava: I'm not sure that makes sense: why should the encoding of type vs. field synopses in the partition_synopsis flatbuffer be a concern of the record field? Handling this in pack/unpack seems clearer, and since we use an empty string as a marker in `pack()` it makes sense to me to do the same in `unpack()` for symmetry.

dominiklohmann: This was necessarily changed as part of the type system refactoring. The `name` function returns a `std::string`, which may allocate, while `is_standalone_type` is guaranteed not to. 

```cpp
std::string qualified_record_field::name() const noexcept {
  if (layout_name_.empty())
    return std::string{field_.name};
  if (field_.name.empty())
    return std::string{layout_name_};
  return fmt::format("{}.{}", layout_name_, field_.name);
}

bool qualified_record_field::is_standalone_type() const noexcept {
  return field_.name.empty();
}
```

This is also visible in the documentation of the function:

```cpp
/// @note This function exists for backwards compatibility. While internally a
/// qualified record field is modeled as a record type with an index to a
/// known field, the partition synopsis used this data structure for both
/// field synopses as intended, and for type synopses which may be considered
/// a bad design decision. As a consequence, we need to be able to distinguish
/// whether a qualified record field is actually just a type as used for type
/// synopses.
[[nodiscard]] bool is_standalone_type() const noexcept;
```

mavam: Do we have to consider that *all* types will soon have a non-empty `.name()`?

dominiklohmann: > Do we have to consider that _all_ types will soon have a non-empty `.name()`?

I think we dropped the idea of doing that because of the many issues it caused, including breaking this.

For the next-gen partition FlatBuffers table I'd like to get rid of this hack entirely and store type and field synopses separately, instead of abusing a qualified record field without a name as essentially a type.

libvast/src/system/spawn_index.cpp
```diff
@@ -41,6 +42,18 @@ spawn_index(node_actor::stateful_pointer<node_state> self,
     return caf::make_error(ec::lookup_error, "failed to find filesystem actor");
   const auto indexdir = args.dir / args.label;
   namespace sd = vast::defaults::system;
+  vast::index_config index_config;
+  const auto* settings = get_if(&args.inv.options, "vast.indexing");
```
dominiklohmann, line 46: Is that correct?

libvast/src/system/index.cpp
```diff
@@ -482,7 +482,12 @@ index_state::collect_query_actors(query_state& lookup,
 void index_state::add_flush_listener(flush_listener_actor listener) {
   VAST_DEBUG("{} adds a new 'flush' subscriber: {}", *self, listener);
   flush_listeners.emplace_back(std::move(listener));
-  detail::notify_listeners_if_clean(*this, *stage);
+  // We may need to call `notify_listeners_if_clean` if the subscription
+  // happens after the data has already completely passed the index, but
+  // we must not to call it before any data at all has arrived it would
+  // create a false positive.
+  if (active_partition.actor != nullptr)
+    detail::notify_listeners_if_clean(*this, *stage);
```
dominiklohmann, line 490: Does this need a separate bug fix changelog entry?

libvast/src/system/index.cpp
```diff
@@ -732,21 +736,22 @@ index(index_actor::stateful_pointer<index_state> self,
...
   self->state.index_opts["cardinality"] = partition_capacity;
+  self->state.synopsis_opts = std::move(index_config);
   // These options must be kept in sync with vast/address_synopsis.hpp and
   // vast/string_synopsis.hpp respectively.
-  self->state.synopsis_opts["max-partition-size"] = partition_capacity;
-  self->state.synopsis_opts["address-synopsis-fp-rate"] = synopsis_fp_rate;
-  self->state.synopsis_opts["string-synopsis-fp-rate"] = synopsis_fp_rate;
+  // self->state.synopsis_opts["max-partition-size"] = partition_capacity;
+  // self->state.synopsis_opts["address-synopsis-fp-rate"] = synopsis_fp_rate;
+  // self->state.synopsis_opts["string-synopsis-fp-rate"] = synopsis_fp_rate;
```
dominiklohmann, line 754: These lines can probably be removed, right?

libvast/src/system/spawn_index.cpp
dominiklohmann, line 46: This is missing from `vast.yaml.example`.

dominiklohmann (commented):
This whole PR needs a changelog entry and a docs update. We should also clarify in there how/whether this affects the `#skip` and `#index=hash` attributes.

I had some general questions inline, most importantly on the configuration section.

libvast/src/partition_synopsis.cpp
```diff
@@ -55,39 +93,45 @@ void partition_synopsis::add(const table_slice& slice,
...
+      // without a dedicated synopsis during lookup and .
+      field_synopses_.emplace(std::move(key), nullptr);
     }
+    // We need to prune the type's metadata here by converting it to a
+    // concrete type and back, because the type synopses are looked up
+    // independent from names and attributes.
+    auto prune = [&]<concrete_type T>(const T& x) {
+      return type{x};
+    };
+    auto cleaned_type = caf::visit(prune, leaf_it->field.type);
```
tobim, line 125: I don't see a fundamental reason that prevents us from supporting aliases in `get_type_fprate`, we would just have to call it here instead of before the loop. You decide if this should be part of this PR or another one :).

vast/integration/reference/basiccommand/step_00.ref
```diff
@@ -1 +1 @@
-381
+394
```
tobim, line 1: What test is this?

tobim (changes requested):
This looks alright to me, but I would like to see a unit test that verifies that synopses are constructed with the correct parameters.

libvast/src/system/index.cpp
lava: I think technically it's not really user-visible, because the whole notifying stream stage issues only are visible during integration tests.

libvast/src/system/spawn_index.cpp
lava: Do you mean the name of the config key? That's what we used in the design doc.

dominiklohmann: The config key. I think it's really confusing to use `indexing`, when there's also indexers and the index currently. Using `index` as a key makes much more sense to me given that we want to split what currently is the index into partition builders and the scheduler.

libvast/src/system/index.cpp
dominiklohmann: Fine by me.

libvast/src/partition_synopsis.cpp
dominiklohmann: Let's do this clean-slate with the next-gen partition FlatBuffers table. It needs to touch this anyways, so there's no need to risk this now.

libvast/src/system/spawn_index.cpp
lava: I don't necessarily disagree, but I'd like to have @mavam chime in before making any changes to the naming part since iirc we did already discuss this previously.

mavam: I'm not opposed to going with `index`. In fact, it was my primary suggestion. The reason why we went with `indexing` was ironically to avoid a naming conflict with the old index concept.

libvast/src/partition_synopsis.cpp
tobim: I don't think that extension would require a change to the table, but I'm fine with doing it at a later time too.

libvast/test/system/partition.cpp
```diff
@@ -87,4 +91,54 @@ TEST(load) {
...
+  auto index_opts = caf::settings{};
+  auto synopsis_opts = index_config{
+    .rules = {
+      {
+        .targets = {"suricata.http.http.url"s},
+        .fp_rate = 0.001,
+      },
+      {
+        .targets = {":address", ":string"},
+        .fp_rate = 0.01,
```
tobim, line 109: Let's use a value that differs from the default here.

libvast/test/system/partition.cpp
```diff
@@ -87,4 +91,54 @@ TEST(load) {
...
+      },
+    },
+  };
+  auto store_actor = vast::system::store_actor{};
+  auto store_name = "dummy_store";
+  auto store_header = nullptr;
+  auto active_partition
+    = self->spawn(::vast::system::active_partition, id, accountant, filesystem,
+                  index_opts, synopsis_opts, store_actor, store_name,
+                  store_header);
```
tobim, line 119: Is there a reason you chose to test this with a full partition instead of just the partition_synopsis in isolation?

lava: No particular reason, I started with the index to get an almost-end to end test, but the API was a bit too involved. I just use the partition synopsis now.

libvast/src/partition_synopsis.cpp
```diff
@@ -34,16 +36,54 @@ void partition_synopsis::shrink() {
...
+  return std::nullopt;
+}
+
+double get_type_fprate(const index_config& config, const type& type) {
+  for (const auto& [targets, fprate] : config.rules) {
+    for (const auto& name : targets) {
+      if (name == ":string" && type == string_type{})
+        return fprate;
+      else if (name == ":address" && type == address_type{})
+        return fprate;
```
dominiklohmann, line 58: Do we already have a follow-up story for generalizing this with the upcoming extractor change? Right now the logic is written per leaf in the schema, but it'd probably make more sense to have a single function that takes just an extractor and the schema in the medium term, and maps the offset of the fields to the fp-rate.

libvast/src/partition_synopsis.cpp
lava: I'm not sure I understand the comment, we already do the type synopsis lookup using the cleaned type so if that resolves aliases then everything should work automatically?

libvast/src/partition_synopsis.cpp
lava: What exactly are the relevant upcoming extractor changes you're thinking of?

There's currently no follow-up story for this, in general we've tried to keep this minimal since there was the prospect of a complete rewrite based on sketches in the near future anyways.

dominiklohmann: Internal ID is `sc-31993`. Rough summary of the relevant bits: We unify type and field extractors and allow the resolve function on types to resolve them both into a list of offsets. That allows us to use type extractors wherever we currently use field extractors.

lava: I've created `sc-32362` as a follow-up.

tobim (approved):
Thanks for adding the test!

PR #2072, opened 2022-02-08
Add .clangd file for VAST

lava:
Trying to use `clangd` on the VAST project often results in an error message `please use -fcoroutines to enable coroutines` for many files. This is caused by the combination of using clang with the libstdc++ header files, which assume that this macro will be enabled when using coroutines. (clang sets a different macro to signal its coroutine support)

### :memo: Checklist

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -1,0 +1,5 @@
+CompileFlags:
+  # We use coroutines in VAST, but the libstdc++ coroutine
+  # header will give an error unless this macro is defined,
+  # which is usually set by gcc's `-fcoroutines` flag.
+  Add: [-D__cpp_lib_coroutine=201902L] 


dominiklohmann:
I'm hesitant to approve this. The macro `__cpp_lib_coroutine` is defined by the C++ standard, and the reason why clang doesn't define it is because its coroutine support is incomplete. While this silences errors, it may also cause the wrong code path to be chosen and lead to hard-to-debug errors (or in this case, wrong results for clangd).

Did you test this with clang and libc++?

lava:
I'm not sure what specific error scenario you have in mind.

1. Regardless of the library macros, we require/assume anyways that clang has a level of coroutine support sufficient to compile VAST. So if our code causes miscompilation due to lacking coroutine support in VAST, we have to throw it out anyways.
2. For libc++, I think its safe to assume that clang can compile whatever is present in libc++.
3. For libstdc++, in the worst case clang might not be able to compile what's in there and clangd will not work for files that use coroutines. But that's exactly the situation we already have without this file when coroutines are incluede.

I didn't test with clang + libcpp, but a quick grep of the llvm codebase shows that this macro is only used in the libcxx unit tests.

PR #2074, opened 2022-02-09
Add new query language plugin

mavam:
This PR adds a new query language plugin that allows for specying a "frontend" language to VAST. The vision is that VAST can execute queries from multiple languages and uses a unified intermediate representation (such as [substrait](https://substrait.io]) internally. The VAST language then becomes one frontend of many others.

As an example frontend, we use [Sigma](https://github.com/SigmaHQ/sigma), for which VAST already has (syntactical) parser suppport.

### :memo: Checklist

- [x] Add new plugin type
- [ ] Allow for configuration of query language frontends in `vast.yaml`
- [x] Rewrite Sigma rule parser as query language frontend
- [ ] Move Sigma-specific logic into plugin
- [ ] Move unit tests into plugin
- [ ] Update [docs.tenzir.com/vast](https://docs.tenzir.com/vast)

### :dart: Review Instructions

File-by-file.

PR #2075, opened 2022-02-09
Add substrait query frontend

mavam:
This PR adds support for a new query language frontend using [substrait](https://substrait.io), a cross-language protobuf-spec for query plans. The idea of this query language plugin is that users can easily connect *any* query language that compiles into substrait, and VAST takes the query plan as binary substrait blob.

This code originated from our bi-monthly hackathon and was heavily inspired by the awesome work from the @duckdb folks over at https://github.com/duckdb/duckdb/pull/3034/. It's highly experimental in nature. 

Moving forward, we may consider pushing substrait deeper into VAST and adopting it as only internal representation for a query plan, similar to what Arrow aims at with Compute IR.  

### :memo: Checklist

- [x] Add substrait as subtree
- [x] Compile protobufs via CMake
- [x] Add plugin scaffold
- [ ] Implement translation logic
- [ ] Add unit tests
- [ ] Add integration tests
- [ ] Rebase onto master after merging #2074
- [ ] Add changelog entry
- [ ] Update docs

File-by-file.


dominiklohmann:
@mavam do you plan to proceed this work sometime in the future, or can we close this as it's unlikely to be picked up again?

mavam:
We can close it until we pick it up again. This was a 24h hackathon experiment. We got a good understanding how DuckDB integrates with Substrait and learned about DuckDB's internal query plan representation. We also got hands-on experience with Substrait and were able to extract WHERE expressions. So a success overall.

Before going deeper, we have to do some homework in VAST so that we can equip query engines with the needed partitions relevant to a query, as well as sqlizing VASTQL.

PR #2076, opened 2022-02-09
Implement a generic aggregation transform step

dominiklohmann:
This implements a generic aggregation transform step plugin. See the attached README file for more information.

### :memo: Checklist

File-by-file. Go over the algorithm in a call with me. Test locally alongside compaction.

### DIFF
--- a/libvast/src/transform.cpp
+++ b/libvast/src/transform.cpp
@@ -144,10 +144,8 @@ caf::error transformation_engine::validate(
       return transform.is_aggregate();
     });
   bool is_aggregate = first_aggregate != transforms_.end();
-  // FIXME: evaluate when aggregates are actually okay to use.
   auto is_aggregate_allowed
-    //= allow_aggregates == allow_aggregate_transforms::yes;
-    = true;
+    = allow_aggregates == allow_aggregate_transforms::yes;
   if (is_aggregate && !is_aggregate_allowed) {
     return caf::make_error(ec::invalid_configuration,
                            fmt::format("the transform {} is an aggregate",
--- a/libvast/src/transform.cpp
+++ b/libvast/src/transform.cpp
@@ -144,8 +144,10 @@ caf::error transformation_engine::validate(
       return transform.is_aggregate();
     });
   bool is_aggregate = first_aggregate != transforms_.end();
+  // FIXME: evaluate when aggregates are actually okay to use.
   auto is_aggregate_allowed
-    = allow_aggregates == allow_aggregate_transforms::yes;
+    //= allow_aggregates == allow_aggregate_transforms::yes;
+    = true;
   if (is_aggregate && !is_aggregate_allowed) {
     return caf::make_error(ec::invalid_configuration,
                            fmt::format("the transform {} is an aggregate",
--- a/plugins/aggregate/README.md
+++ b/plugins/aggregate/README.md
@@ -2,7 +2,10 @@
 
 The `aggregate` plugin for VAST adds a generic aggregation transformation step
 that allows for flexible configuration. This is best used in conjunction with
-the [`compaction` plugin][docs-compaction].
+the [`compaction` plugin][docs-compaction]. The aggreagtion operates on the
+scope of the transformation, which for import and export is the size of a single
+batch (configurable as `vast.import.batch-size`), and for compaction is the size
+of a partition (configurable as `vast.max-partition-size`).
 
 [docs-compaction]: https://docs.tenzir.com/vast/features/compaction
 
@@ -14,34 +17,20 @@ underlying mechanism that users already know from queries.
 
 ### `group-by`
 
-A list of columns to group by. VAST internally calculates the cross product for
+A list of columns to group by. VAST internally calculates the combined hash for
 all the configured columns for every row, and sorts the data into buckets for
 aggregation.
 
-### `round-temporal-multiple`
+### `time-resolution`
 
-An optional duration value that specifies the tolerancy when comparing time
+An optional duration value that specifies the tolerance when comparing time
 values in the `group-by` section.
 
-### `sum` / `min` / `max` / `mean` / `any` / `all`
+### `sum` / `min` / `max` / `any` / `all`
 
-A list of columns to perform the respective computation on within the grouped
-buckets. Fields that have no such compute operation and are not part of the
-`group-by` columns are dropped from the output.
-
-### `disable-eager-pass` / `disable-lazy-pass`
-
-By default, the `aggregate` transform step computes the aggregation in two
-passes:
-- An eager pass when a batch of data arrives at the transformation, and
-- a lazy pass when VAST requests the result of the transformation.
-
-Either of the passes may be disabled depending on the use case. We highly
-recommend leaving both passes enabled when using the `aggregate` transform with
-the `compaction` plugin.
-
-Note that if both passes are enabled, `mean` calculates the mean of the eager
-result during the lazy pass, which may introduce rounding errors.
+A list of columns to perform the respective aggregation function on within the
+grouped buckets. Fields that have no such function specified and are not part of
+the `group-by` columns are dropped from the output.
 
 ## Usage Example
 
@@ -68,7 +57,7 @@ vast:
             - suricata.flow.dest_port
             - suricata.flow.proto
             - suricata.flow.event_type
-          round-temporal-multiple: 1 minute
+          time-resolution: 1 minute
           sum:
             - suricata.flow.pcap_cnt
             - suricata.flow.flow.pkts_toserver
--- a/plugins/aggregate/README.md
+++ b/plugins/aggregate/README.md
@@ -97,7 +97,7 @@ time:
   rules:
     - name: compact-suricata-flow
       after: 7 days
-      transform: aggregate-suricata-flow
+      transform: suricata-flow-aggregate
       types:
         - suricata.flow
 ```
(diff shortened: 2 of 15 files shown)


mavam:
Awesome to see the Arrow Compute API being leveraged!

dominiklohmann:
[sc-29754]

plugins/aggregate/aggregate.cpp
```diff
@@ -0,0 +1,469 @@
...
+      VAST_ASSERT(batch_builder->num_fields()
+                  == detail::narrow_cast<int>(
+                    resolved_columns.buckets.begin()->second.size()));
+      // TODO: can we handle the error safely in some way?
+      VAST_ASSERT(make_batch_builder_result.ok());
+      // Iterate over all the buckets to perform the aggregations.
+      for (const auto& bucket : resolved_columns.buckets) {
+        for (int builder_column = 0;
+             auto action : resolved_columns.group_by_actions) {
+          switch (action) {
```
dominiklohmann, line 324: I'd like to experiment with additionally doing the aggregation eagerly inside `add(...)` already instead of doing it all at once here in `finish()`. The current way has quite a bit of overhead when applied to a whole partition at once.

plugins/aggregate/aggregate.cpp
```diff
@@ -0,0 +1,469 @@
...
+#include <arrow/table_builder.h>
+
+#include <unordered_map>
+
+namespace vast::plugins::aggregate {
+
+/// The configuration of an aggregate transform step.
+struct configuration {
+  /// Duration window for grouping time values.
+  std::optional<duration> time_window = {};
```
dominiklohmann, line 33: We should rename this option, `time-window` is really not a good name for it.

plugins/aggregate/aggregate.cpp
```diff
@@ -0,0 +1,469 @@
...
+    }
+    return hasher.finish();
+  }
+};
+
+} // namespace std
+
+namespace vast::plugins::aggregate {
+
+struct resolved_columns {
```
dominiklohmann, line 119: This struct has a really bad name. When @6yozo and I started prototyping this, it only held the resolved columns. Over time, we noticed that we have much more state that we want to cache per layout, and that is what this really is: our per-layout state.

plugins/aggregate/aggregate.cpp
```diff
@@ -0,0 +1,469 @@
...
+template <>
+struct std::hash<vast::plugins::aggregate::bucket_key> {
+  size_t
+  operator()(const vast::plugins::aggregate::bucket_key& x) const noexcept {
+    auto hasher = vast::xxh64{};
+    for (const auto& scalar : x) {
+      VAST_ASSERT(scalar);
+      vast::hash_append(hasher, scalar->hash());
+    }
+    return hasher.finish();
```
dominiklohmann, line 111: @mavam we don't happen to have something like `hash_combine` for combining multiple hash values from an external source, do we? What do you reckon would be best here to combine a series of hashes into a single hash?

plugins/aggregate/aggregate.cpp
```diff
@@ -0,0 +1,469 @@
...
+private:
+  /// The underlying configuration of the transformation.
+  configuration config_ = {};
+
+  /// Cache for mapping of layout to resolved columns.
+  std::unordered_map<type, resolved_columns> resolved_columns_cache_ = {};
+
+  /// Options for the `round_to_multiple` Arrow Compute function.
+  arrow::compute::RoundTemporalOptions round_temporal_options_{
+    1, arrow::compute::CalendarUnit::SECOND};
```
dominiklohmann, line 424: We should document clearly that we always use seconds as the base unit, because using nanoseconds here is broken in at least the Arrow 7.0.0 available via Homebrew on macOS.

mavam (commented):
```yaml
          time-window: 10 seconds
```

Can we rename this to something less confusing? E.g., `timestamp-granularity` and `duration-granularity`? Or `s/granularity/resolution/`?

plugins/aggregate/aggregate.cpp
dominiklohmann: Moving @mavam's comment from below here:

> ```yaml
> time-window: 10 seconds
> ```
> Can we rename this to something less confusing? E.g., `timestamp-granularity` and `duration-granularity`? Or `s/granularity/resolution/`?

plugins/aggregate/aggregate.cpp
dominiklohmann: I renamed this into `aggregation`.

plugins/aggregate/aggregate.cpp
dominiklohmann: milliseconds also works, and I think that is fine enough for rounding that we don't need to explain it.

libvast/src/transform.cpp
```diff
@@ -144,8 +144,10 @@ caf::error transformation_engine::validate(
       return transform.is_aggregate();
     });
   bool is_aggregate = first_aggregate != transforms_.end();
+  // FIXME: evaluate when aggregates are actually okay to use.
   auto is_aggregate_allowed
-    = allow_aggregates == allow_aggregate_transforms::yes;
+    //= allow_aggregates == allow_aggregate_transforms::yes;
+    = true;
```
dominiklohmann, line 150: This needs to be reverted before merging, it's a leftover from testing.

dispanser (approved):
Fine with me! Made a minor comment.

Digging through the arrow APIs, I found https://arrow.apache.org/docs/dev/cpp/compute.html#grouped-aggregations-group-by which may one day become an off-the-shelf way to do this without explicitly handling the grouping ourselves.

Great to see that generic and powerful aggregation plugin :+1:

plugins/aggregate/aggregate.cpp
```diff
@@ -0,0 +1,603 @@
...
+
+  /// Aggregates a vector of record batches into a set of record batches that
+  /// belong to the same buckets.
+  caf::error aggregate(const arrow::RecordBatchVector& batches) {
+    if (batches.empty())
+      return caf::none;
+    // If we don't have a builder yet for storing the intermediate results, this
+    // is where we create it.
+    if (!builder_) {
+      const auto initial_size = detail::narrow_cast<int>(buckets_.size());
```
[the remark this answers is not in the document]
dominiklohmann: I fixed this by making the buckets map a local variable to this function. Then I also don't need to clear it below.

dominiklohmann:
> Digging through the arrow APIs, I found https://arrow.apache.org/docs/dev/cpp/compute.html#grouped-aggregations-group-by which may one day become an off-the-shelf way to do this without explicitly handling the grouping ourselves.

@dispanser the main issue is that this only supports grouping on a single column, which is insufficient for our use cases.

plugins/aggregate/tests/aggregate.cpp
```diff
@@ -12,52 +12,62 @@
...
-  REQUIRE(client_sink_transforms);
+FIXTURE_SCOPE(aggregate_tests, fixture);
+
+TEST(aggregate Zeek conn log) {
+  auto opts = caf::settings{};
+  caf::put(opts, "group-by", std::vector<std::string>{"ts"});
+  caf::put(opts, "round-temporal-multiple", "1 day");
+  caf::put(opts, "sum", std::vector<std::string>{"duration", "resp_pkts"});
+  caf::put(opts, "min", std::vector<std::string>{"orig_ip_bytes"});
+  caf::put(opts, "max", std::vector<std::string>{"resp_ip_bytes"});
```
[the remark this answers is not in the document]
dominiklohmann: I'm only a user of the `transform_plugin` API, I didn't decide to have `transform_plugin::make_transform_step` take a `caf::settings`.

This is certainly not the PR to change it in. Please open a new story if you'd like to see that changed.

plugins/aggregate/tests/aggregate.cpp
```diff
@@ -12,52 +12,62 @@
...
+  auto aggregate_step = unbox(aggregate_plugin->make_transform_step(opts));
+  REQUIRE_EQUAL(rows(zeek_conn_log_full), 8462u);
+  for (const auto& slice : zeek_conn_log_full)
+    CHECK_EQUAL(aggregate_step->add(slice.layout(), to_record_batch(slice)),
+                caf::none);
+  const auto result = unbox(aggregate_step->finish());
+  REQUIRE_EQUAL(result.size(), 1u);
+  const auto aggregated_slice = table_slice{result[0].batch, result[0].layout};
+  // NOTE: I calculated this data ahead of time using jq, so it can safely be
+  // used for comparion here. -- DL
```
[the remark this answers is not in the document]
dominiklohmann: Does the expanded comment suffice for you @mavam?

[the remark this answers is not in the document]
mavam: More than I expected, yes! 🙂

lava (approved):
Looks great, thanks!

plugins/aggregate/README.md
```diff
@@ -0,0 +1,103 @@
...
+options that refer to fields support suffix matching, using the same syntax and
+underlying mechanism that users already know from queries.
+
+### `group-by`
+
+A list of columns to group by. VAST internally calculates the cross product for
+all the configured columns for every row, and sorts the data into buckets for
+aggregation.
+
+### `round-temporal-multiple`
```
dominiklohmann, line 21: We want to rename this to time-resolution

PR #2080, opened 2022-02-14
Return a hit count from the partition query handler

tobim:
This changes the query handlers in the partition and store actors to return a hit count instead of a done atom.

### :memo: Checklist

* Verify that the additional test checks are sufficient.


dominiklohmann (approved):
Small little improvement that I think will be valuable in the medium term. Let's ship it!

PR #2084, opened 2022-02-14
Employ `caf::visit` to reduce boilerplate when handling arrow data

dispanser:
This is purely a refactoring, leveraging our recently-created sum type
over `arrow::Array` to double-dispatch on the tuple `(VAST type, Arrow type)`.

### :memo: Checklist

It's a single commit / single file change.

Open question: The new implementation of `apply_index` could be slightly less efficient than the previous one, because we're now calling `value_at` for every element, instead of dispatching on the array type once and then have a tight loop. This could be fix by re-introducing some boilerplate in the now-defunct `index_applier`, but I'm not sure it's worth it -- I assume actually maintaining the index on `index.append(...)` is more expensive.


mavam:
Nice net-negative PR there. And double-dispatch is so much more readable...

dispanser:
I could do the same refactoring to the "old" (current arrow table slice) code in the `legacy` namespace. 

It is likely even possible to integrate handling old representations into `value_at`, adding a few lambdas (potentially amending our sum type with some of the legacy variants for address and subnet).

libvast/src/arrow_table_slice.cpp
```diff
@@ -986,196 +789,72 @@ auto record_at(const record_type& type, const arrow::StructArray& arr,
...
 }
 
-class index_applier {
-public:
-  using result_type = void;
-
-  index_applier(size_t offset, value_index& idx)
-    : offset_(detail::narrow_cast<int64_t>(offset)), idx_(idx) {
-    // nop
+auto apply_index(size_t offset, value_index& idx, const arrow::Array& arr,
```
dominiklohmann, line 850: I imagine that we don't need this function at all. Instead, let's add something like this:

```cpp
auto values(size_t column, const concrete_type auto& type, const auto& array)
    -> detail::generator<view<type_to_data_t<decltype(type)>>> {
  ...
}
```

And then a second version that is generic:

```cpp
auto values(size_t column, const type& type, const arrow::Array& array)
    -> detail::generator<data_view> {
  auto f = [column](const concrete_type auto& type, const auto& array) {
    return values(column, type, array);
  };
  for (auto&& result : caf::visit(f, type, array))
    co_yield result;
}
```

After this, we can simply use this in `apply_column_to_index`:

```cpp
if (const auto& batch = record_batch()) {
  const auto& layout = caf::get<record_type>(this->layout());
  auto type = layout.field(layout.resolve_flat_index(column)).type;
  const auto& array = state_.flat_columns[column];
  for (size_t row = 0; auto&& value : values(column, type, array)) {
    if (!caf::holds_alternative<view<caf::none_t>>(value))
      idx.append(value, offset + row);
    ++row;
  }
}

dominiklohmann (commented):
The inversion did slightly lower the indexing performance. I proposed a change inline that makes the API both even easier to use, and should fix that problem.

dispanser:
> The inversion did slightly lower the indexing performance. I proposed a change inline that makes the API both even easier to use, and should fix that problem.

I'd be interested to learn how you benchmark this. It seems tricky to isolate the indexing, avoiding a benchmark of the entire `import` process (which may be fine, too).

dominiklohmann:
> I'd be interested to learn how you benchmark this. It seems tricky to isolate the indexing, avoiding a benchmark of the entire `import` process (which may be fine, too).

`import csv` and `import netflow` are both super fast and easily get to 500k+ events/sec. The index will always be the slowest part of the import pipeline for these formats and slow them down via back pressure.

PR #2086, opened 2022-02-15
Improve name lookup in meta index to reduce FPs

dominiklohmann:
The candidate check of the meta index must use the proper suffix matching algorithm of the type system rather than plain suffix matching on a flat qualified record field's name in order to avoid false positive candidate partitions. E.g., the field extractor "ts" resolved to partitions that contained layouts with a field names "orig_pkts" rather than just those that only contained a field exactly named "ts". This change fixes that behavior.

### :memo: Checklist

@tobim please verify that this fixes the issue you've described. The check itself is rather simple.

This does not need a changelog entry since it should only be a speed improvement for some queries, but does not change any behavior in a user-facing way.

### DIFF
--- a/libvast/src/system/meta_index.cpp
+++ b/libvast/src/system/meta_index.cpp
@@ -307,8 +307,12 @@ std::vector<uuid> meta_index_state::lookup_impl(const expression& expr) const {
                 // Compare the desired field name with each field in the
                 // partition.
                 auto matching = [&] {
-                  for (const auto& [fqf, _] : synopsis.second.field_synopses_) {
-                    if (fqf.name().ends_with(*s))
+                  for (const auto& [field, _] :
+                       synopsis.second.field_synopses_) {
+                    VAST_ASSERT(!field.is_standalone_type());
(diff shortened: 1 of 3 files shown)


libvast/src/system/meta_index.cpp
```diff
@@ -329,7 +333,12 @@ std::vector<uuid> meta_index_state::lookup_impl(const expression& expr) const {
         },
         [&](const field_extractor& lhs, const data&) -> result_type {
           auto pred = [&](const auto& field) {
-            return field.name().ends_with(lhs.field);
+            VAST_ASSERT(!field.is_standalone_type());
+            auto rt = record_type{{field.field_name(), field.type()}};
+            for ([[maybe_unused]] const auto& offset :
+                 rt.resolve_key_suffix(lhs.field, field.layout_name()))
+              return true;
```
tobim, line 340: Can we do a type check here as well? That would help in case we have multiple fields with the same suffix name but different types.

tobim (approved):
I think the changelog entry contains a little bit too much low level detail but we can still make this more palatable when we compile the release notes.

PR #2095, opened 2022-02-17
Notify the type registry about transformed partitions

lava:
- Allow passing multiple transforms to apply handler
- Notify type registry about transformed partitions

### :memo: Checklist


dominiklohmann (approved):
Thanks for this rather quick fix.

There's a small issue I've noticed that is undefined behavior; it actually worked as expected when I ran this locally, but that line should definitely be moved up.

I'll wait with cutting the v1.1.0-rc1 until after this is merged.

PR #2097, opened 2022-02-18
Adjust index statistics for partition transforms

lava:
- Adjust index statistics for partition transforms
- Return partition_info from apply handler

### :memo: Checklist

Review commit-by-commit.


libvast/vast/fwd.hpp
```diff
@@ -121,6 +121,8 @@ class uuid;
 class value_index;
 
 struct attribute;
+struct augmented_partition_synopsis;
+struct layouts_diff;
```
tobim, line 125: leftover?

libvast/vast/partition_synopsis.hpp
```diff
@@ -84,6 +88,36 @@ struct partition_synopsis final : public caf::ref_counted {
...
+struct partition_info {
+  /// The partition id.
+  vast::uuid uuid = vast::uuid::nil();
+
+  /// Total number of events in the partition. The sum of all
+  /// values in `stats`.
+  size_t events = 0ull;
+
+  /// The newest import timestamp of the table slices in this partition.
+  time max_import_time;
```
tobim, line 101: These are also in the synopsis itself. Can you find a way to avoid the duplication? I'd accept push back if you convince me that it would be too much complexity.

libvast/src/system/index.cpp
```diff
@@ -1236,15 +1235,22 @@ index(index_actor::stateful_pointer<index_state> self,
...
+              .uuid = aps.uuid,
+              .events = aps.synopsis->events,
+              .max_import_time = aps.synopsis->max_import_time,
+              .stats = std::move(aps.stats),
+            };
+            // Update the index statistics. We only need to add the events of
+            // the new partition here, the subtraction of the old events is
+            // done in `erase`.
+            for (const auto& [name, stats] : result.stats.layouts)
+              self->state.stats.layouts[name].count += stats.count;
```
tobim, line 1249: I'd argue this becomes a problem now. If the `erase` message gets delayed because of a large backlog intermediate requests will produce incorrect results. Shouldn't the erase logic be factored into a function at this point?

tobim (changes requested):
I think the overall solution is fine but I hope we can find a better way to slice `partition_synopsis` and `partition_info`.

libvast/vast/partition_synopsis.hpp
lava: This struct does *not* contain the synopsis itself.

libvast/src/system/index.cpp
lava: Hm, it's an interesting question what the "correct" number would be, because the events will still exist on disk until the `erase` went through, even though they're not reachable through the meta index anymore.

The erase handler goes through a number of continuations itself, so it would be non-trivial to refactor it into a function. I think the best alternative would probably be to return two sets of `index_statistics` from the partition transformer, one for the incoming and one from the outgoing events, and then to make a second erase handler that does not update statistics or remove from the meta index.

libvast/vast/partition_synopsis.hpp
tobim: I'm saying that both the number of events and the max import time are also present in the synopsis.

libvast/src/system/index.cpp
tobim: I'm afraid we're thinking about this at the wrong level. The ultimate source of truth is the meta index because that controls which partitions are reachable. I think in the long run we should move the statistics for the passive database to the meta index and sum up it up with the active statistics in the node.

dominiklohmann: This should result in a follow-up story. Let's not make that change in this PR.

libvast/vast/partition_synopsis.hpp
dominiklohmann: That doesn't make any sense to me @tobim. Can you elaborate?

libvast/vast/index_statistics.hpp
```diff
@@ -0,0 +1,37 @@
...
+struct index_statistics {
+  /// The number of events for a given layout.
+  detail::heterogenous_string_hashmap<layout_statistics> layouts;
+
+  template <class Inspector>
+  friend auto inspect(Inspector& f, index_statistics& x) {
+    return f(caf::meta::type_name("index_statistics"), x.layouts);
+  }
+};
+} // namespace vast
```
dominiklohmann, line 37: Nit: missing whitespace above this line.

dominiklohmann (approved):
Approving, left some comments inline.

libvast/vast/partition_synopsis.hpp
tobim: I was confused by the names `partition_info` and `augmented_partition_synopsis` because I assumed that would exist for active partitions as well. For now we can leave things as they are to avoid refactoring the meta index with little benefit.

libvast/src/system/index.cpp
tobim: I agree this should happen in a different PR.

tobim (approved):

libvast/src/system/index.cpp
lava: I created https://app.shortcut.com/tenzir/story/31848/atomically-adjust-index-statistics-for-partition-transforms

PR #2099, opened 2022-02-21
Prefer CAF at `CAF_ROOT_DIR` over system CAF

dominiklohmann:
This is a small bug fix to our build scaffolding, which preferred a CAF installation in the default lookup paths over a manually specified installation with `CAF_ROOT_DIR`. The solution is to not consider the specified location to be a hint, but rather to be a path, and to hard-error if there is no CAF at the specified location via `NO_DEFAULT_PATH`.

### :memo: Checklist

@mavam please double-check that this fixes the build issue you've experienced.
@tobim please sanity-check the CMake changes; they are rather straightforward, I think.

I don't think this needs a changelog entry, it's a developer-facing bug-fix only.


mavam (commented):
Confirming that this fixes the issue for me. Leaving the review of the CMake voodoo to @tobim.

PR #2100, opened 2022-02-21
Fix race condition in CI for annotated tag pushes

dominiklohmann:
If a dependent job runs git-describe to download an artifact uploaded from a preceding job, we may run into race conditions because an annotated tag might've been pushed in the meantime. This avoids that issue by only calling git-describe once in the beginning, and then sharing that as a job output with all the other jobs.

### :memo: Checklist

Not user-facing. Reviewable by any YAML engineer; the fix is relatively straight forward. I recommend skimming the docs on [defining outputs for jobs](https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs).

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -30,6 +30,27 @@ jobs:
         with:
           access_token: ${{ github.token }}
 
+  determine-version:
+    name: Determine Version
+    runs-on: ubuntu-20.04
+    outputs:
+      build-version: ${{ steps.determine-version.outputs.build-version }}
+      release-version: ${{ steps.determine-version.outputs.release-version }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Fetch Tags
+        run: git fetch origin +refs/tags/*:refs/tags/*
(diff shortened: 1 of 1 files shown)


.github/workflows/vast.yaml
```diff
@@ -126,7 +149,7 @@ jobs:
       CXX: ${{ matrix.build.cxx }}
       CCACHE_ABSSTDERR: true
       CCACHE_COMPRESS: true
-      CCACHE_COMPRESSLEVEL: 9
+      CCACHE_COMPRESSLEVEL: 0
```
dispanser, line 152: This seems like an unrelated change; I'm fine with it, just making sure it's intended

dispanser (approved):
LGTM. I can't really test what it's doing, but the intention and implementation of the change is clear.

Minor remark / question wrt to the CCACHE compression level being changed.

.github/workflows/vast.yaml
dominiklohmann: It is intended. Just forgot to mention it in the description, but a recent release of ccache made level 0 much smarter with regards to the size limit, so it's better to use this now. I just remembered that I wanted to make this change when I worked on this PR.

PR #2102, opened 2022-02-21
Deploy VAST to AWS Fargate

rdettai:
Added a `cloud` directory at the root of the project with some Terraform scripts to deploy VAST on AWS

### :memo: Checklist

First take a look at the README in the new folder. Can then go through the different resources created by the Terraform script to get a better idea of what is going on under the hood.


cloud/Makefile
```diff
@@ -0,0 +1,36 @@
+-include default.env
+
+vpc_id ?= $(eval vpc_id := $(shell bash -c 'read -p "VPC ID (existing): " input; echo $$input'))$(vpc_id)
```
dominiklohmann, line 3: Mostly out of curiosity: Why pass things through `bash -c` everywhere?

cloud/Makefile
```diff
@@ -0,0 +1,36 @@
...
+		--propagate-tags TASK_DEFINITION \
+		--launch-type FARGATE  \
+		--network-configuration awsvpcConfiguration={subnets=[$(shell bash -c '${terraform} output ids_subnet_id')],securityGroups=[$(shell bash -c '${terraform} output vast_security_group')],assignPublicIp=ENABLED} \
+		--task-definition $(shell bash -c '${terraform} output vast_task_definition')
+
+list-tasks:
+	aws --region ${aws_region} ecs list-tasks --cluster $(shell bash -c 'terraform output fargate_cluster_name') | jq -r '.taskArns | map(.[64:]) | reduce .[] as $$item (""; . + $$item + " ")'
+
+stop-all-tasks:
+	for task in $(shell bash -c 'make list-tasks'); do \
```
dominiklohmann, line 33:
```suggestion
	for task in $(shell bash -c '$(MAKE) list-tasks'); do \
```

Make should always refer to itself via `$(MAKE)` as that always picks the correct version of Make when multiple are installed.

cloud/Makefile
```diff
@@ -0,0 +1,36 @@
...
+		aws --region ${aws_region} ecs stop-task --task $$task --cluster $(shell bash -c 'terraform output fargate_cluster_name'); \
+	done
+
```
dominiklohmann, line 36: Style nit: Empty line at the end of the file. We usually end files in a newline, but don't have empty lines at the end.

cloud/cluster.tf
```diff
@@ -0,0 +1,67 @@
...
+  tags = module.env.default_tags
+}
+
+resource "aws_iam_role_policy" "fargate_task_execution_policy" {
+  name = "${module.env.module_name}_task_execution_${module.env.stage}_${var.region_name}"
+  role = aws_iam_role.fargate_task_execution_role.id
+
+  policy = <<EOF
+{
+    "Version": "2012-10-17",
```
dominiklohmann, line 49: Style nit: Indentation depth for the heretic above is 2, but here it's 4 spaces. We usually indent with 2 spaces.

cloud/env/conf.tf
```diff
@@ -0,0 +1,33 @@
+locals {
+  envs = {
+    default = {
+      vast_image_version = "latest"
```
dominiklohmann, line 4: Should this use the latest release instead? `latest` here refers to current master. We might also want to consider changing that naming in our CI, it doesn't quite make sense to set it up that way.

dominiklohmann (commented):
Just skimmed the code, and I also approved you for being able to trigger our CI runs. I think this looks great—I'm excited to toy around with it tomorrow.

cloud/cluster.tf
rdettai: strange, this somehow escaped the ide auto escape

cloud/Makefile
rdettai: good to know!

cloud/Makefile
rdettai: just struggling with how Make handles variables 😄 Otherwise the command output is not properly written through. But I am definitively no Make expert, so if you have a more elegant way to write it, I am all for it 😉

cloud/Makefile
rdettai: added the setting to my ide, so shouldn't happen again 😉

rdettai:
I am not really happy with the feature set of this PR, because it does not make it convenient to check that the instance properly running (no way to connect to it).
https://github.com/tenzir/vast/pull/2108 adds the possibility to run a VAST client in Lambda, which will be a great way to interact with the instance.

dominiklohmann:
> We can either close this PR and focus on the bigger feature set of https://github.com/tenzir/vast/pull/2108, or merge this first, whatever works best for you.

I think we can close this one, then we can do the review later in one fell swoop. That's probably easiest.

PR #2108, opened 2022-02-23
Deploy VAST to AWS Lambda

rdettai:
This changes adds the possibility to deploy VAST on AWS Lambda and Fargate. This is a more interesting feature set as it will enable connecting to the VAST server on Fargate through client VAST instances on Lambda

### :memo: Checklist

This PR can be broken down into:
- the terraform scripts to deploy fargate task definition and the lambda function. It will create a subnet in a provided vpc where all this tooling will execute
- a wrapper image around VAST for the lambda function, containing the VAST CLI and the AWS Lambda runtime interface (Dockerfile + python handler)
- a makefile to simplify the interaction with the deployment script and test the deployment


dispanser (approved):
Great work, thank you so much.

I've tested it following the readme, and was able to deploy vast, import data, query data, ... so everything works as expected.

Two points:
1. I'm not sure the approach using environment variables / `default.env` is optimal. In my previous usages of terraform, I always relied on `tfvars` files to inject behavior (and some defaults for the variables). I can see how the approach taken here makes working with the Makefile easier, but it stops us from using `terraform` directly.
2. Now that you added the ability to directly execute commands on the ecs task using `make execute-command`, the existence of lambda functions is not strictly necessary any more. It's a nice showcase for how to integrate, and lambda is practically free, so it's not  really a problem.

My suggestion would be to move forward as is, and see how this actually gets used and adapt it from these learnings.

cloud/aws/Makefile
```diff
@@ -0,0 +1,75 @@
...
+init:
+	terraform init
+
+docker-login:
+	aws ecr get-login-password | \
+	docker login --username AWS --password-stdin \
+		"$(shell aws sts get-caller-identity --query 'Account' --output text).dkr.ecr.${aws_region}.amazonaws.com"
+
+docker-build:
+	docker build -t tenzir/vast-lambda:latest .
```
dominiklohmann, line 21: @dispanser / @rdettai Do we want to build this in CI and upload to Dockerhub? That should be trivial to add with a few adjustments in `.github/workflows/vast.yaml`.

cloud/aws/README.md
```diff
@@ -0,0 +1,17 @@
+# Tenzir cloud deployment
+
+Deploy Tenzir easily to AWS with this set of terraform scripts.
+
+You need Terraform>=0.12 and the AWS CLI V2 to be installed.
+
+Currently supported:
+- deploy VAST task definition to AWS Fargate with `make deploy`
+- start a VAST task using `make start-task`
```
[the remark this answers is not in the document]
rdettai: thanks for catching this

cloud/aws/README.md
```diff
@@ -0,0 +1,77 @@
...
+Terraform variables to provide:
+- `vpc_id`: the existing VPC to which you plan to attach your VAST stack
+- `subnet_cidr`: the ip range of the subnet where the VAST stack will be placed. A subnet with that range will be created. It should not overlap with an existing subnet in specified VPC and the VPCs peered to it.
+- `aws_region`: the region in which this VPC is defined
+
+## Features
+
+Currently supported:
+- deploy VAST both as a Fargate task definition and a Lambda function with `make deploy`
+- start a VAST server task using `make start-task`
```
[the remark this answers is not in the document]
rdettai: Haha, I think you pointed this out twice 😄. Thanks anyway!

rdettai:
@dispanser 
> I'm not sure the approach using environment variables / default.env is optimal. In my previous usages of terraform, I always relied on tfvars files to inject behavior (and some defaults for the variables). I can see how the approach taken here makes working with the Makefile easier, but it stops us from using terraform directly.

I don't love this `make` import either because it is kind of a hack, but it works well in practice. You can still use tfvars to provide the variables if you want to interact with terraform directly: `terraform apply -var-file="yourfile.tfvars"`

> Now that you added the ability to directly execute commands on the ecs task using make execute-command, the existence of lambda functions is not strictly necessary any more. It's a nice showcase for how to integrate, and lambda is practically free, so it's not really a problem.

Yeah, I think I over-engineered this a bit 😅. But I believe it can be very useful if you want to add extra tooling around `vast import` while keeping the server image to its bare minimum. Ideally, the vast server should be able to run on a much slimmer image (maybe scratch?), and in that case, going through lambda would be a very elegant way to keep the linux flavor you need/like to use the client CLI.

cloud/aws/Makefile
rdettai: Lambda requires the image to be in a local AWS repository (ECR), which means that we need to push it there at deploy time. Having the image available in DockerHub would not be a great addition to the workflow, the only benefit I see is that instead of build+push you could do pull+push.

Adding it to the CI for building only on the other hand could be nice as it could catch some bugs if the base image is changed in a way that breaks the extra layer we are adding here. I don't really know how likely it would be that the change is big enough to break the downstream image at build time 😄

dominiklohmann: Do you mind if I push a commit to this PR that adds it?

rdettai: did not add the docker push, as I don't think that image would be useful in docker hub, but I added the build for regression detection

dominiklohmann (approved):
Thanks for incorporating the CI changes as well.

rdettai:
@dominiklohmann done! you can re-activate auto-merge ;-)

PR #2058, opened 2022-01-31
Revamp partition sketching

mavam:
This PR is a workspace for the revamped partition sketching infrastructure.

### :memo: Checklist

- [ ] All user-facing changes have changelog entries.
- [ ] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [x] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

Look at the whole PR at once, excluding the commits that already were pre-reviewed independently. For @mavam , ignore the parts you wrote yourself :D

Ideally, build this locally and set `use_sketches` to true by default and play around if you see any unexpected segfaults etc.

PR #2110, opened 2022-02-23
Return kind of concrete types from `type::name()`

dominiklohmann:
As a VAST developer using the type system, I would like to have a consistent and meaningful interpretation of the type "name". There is currently a mismatch on the query side and the implementation internally. For example, :timestamp has the name "timestamp" whereas :time has no name. Given that there is one global type namespace, this is an inconsistency in the model.

This change fixes that inconsistency throughout the type system.

### :memo: Checklist

This change is absolutely not user-facing. I've tested that extensively. If as a reviewer, you find a place where this changes VAST's behavior, please let me know, because I would consider that a bug in this diff.

For the code itself, review starting at type.hpp and then looking at the individual changes.

### DIFF
--- a/libvast/src/experimental_table_slice_builder.cpp
+++ b/libvast/src/experimental_table_slice_builder.cpp
@@ -775,20 +775,31 @@ serialize_attributes(const std::vector<type::attribute_view>& attrs) {
 
 std::shared_ptr<const arrow::KeyValueMetadata>
 make_arrow_metadata(const type& t) {
-  auto metadata = std::make_shared<arrow::KeyValueMetadata>();
+  auto keys = std::vector<std::string>{};
+  auto values = std::vector<std::string>{};
   int nesting_depth = 0;
   for (const auto& [name, attrs] : t.names_and_attributes()) {
-    if (!name.empty())
-      metadata->Append(fmt::format("VAST:name:{}", nesting_depth),
-                       std::string{name});
-    if (!attrs.empty())
-      metadata->Append(fmt::format("VAST:attributes:{}", nesting_depth),
-                       serialize_attributes(attrs));
+    if (!name.empty()) {
+      keys.emplace_back(fmt::format("VAST:name:{}", nesting_depth));
+      values.emplace_back(std::string{name});
+    }
+    if (!attrs.empty()) {
+      keys.emplace_back(fmt::format("VAST:attributes:{}", nesting_depth));
+      values.emplace_back(serialize_attributes(attrs));
+    }
     ++nesting_depth;
   }
+  VAST_ASSERT(nesting_depth > 0);
+  if (nesting_depth == 1)
+    return nullptr;
   if (nesting_depth == 0)
     return nullptr;
-  return metadata;
+  // Strip the last layer of metadata again, as that is guaranteed to be the
+  // types kind.
+  VAST_ASSERT(keys.back() == fmt::format("VAST:name:{}", nesting_depth - 1));
+  keys.pop_back();
+  values.pop_back();
+  return arrow::KeyValueMetadata::Make(std::move(keys), std::move(values));
 }
 
 } // namespace
--- a/libvast/src/experimental_table_slice_builder.cpp
+++ b/libvast/src/experimental_table_slice_builder.cpp
@@ -795,7 +795,7 @@ make_arrow_metadata(const type& t) {
   if (nesting_depth == 0)
     return nullptr;
   // Strip the last layer of metadata again, as that is guaranteed to be the
-  // types kind.
+  // type's kind.
   VAST_ASSERT(keys.back() == fmt::format("VAST:name:{}", nesting_depth - 1));
   keys.pop_back();
   values.pop_back();
--- a/libvast/src/type.cpp
+++ b/libvast/src/type.cpp
@@ -519,7 +519,7 @@ legacy_type type::to_legacy_type() const noexcept {
     },
   };
   auto result = caf::visit(f, *this);
-  if (!name().empty())
+  if (has_name())
     result = legacy_alias_type{std::move(result)}.name(std::string{name()});
   for (const auto& attribute : attributes()) {
     if (attribute.value.empty())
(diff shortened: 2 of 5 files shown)


libvast/vast/type.hpp
```diff
@@ -328,6 +330,9 @@ class type final : public stateful_type_base {
   [[nodiscard]] std::optional<std::string_view>
   attribute(const char* key) && = delete;
 
+  /// Returns whether a type has any custom name.
+  [[nodiscard]] bool has_name() const noexcept;
```
mavam, line 334: There's a gap between the mental model we convey in the documentation and what `has_name` conveys. Users think in terms of *aliases*, only those have a custom name. So I'd rather use a function `is_alias()` that implies that the type name is not a default name.

I know it doesn't exist, but the mental model would call for something like `holds_alternative<alias_type>`.

libvast/src/type.cpp
```diff
@@ -686,20 +700,49 @@ detail::generator<std::string_view> type::names() const& noexcept {
...
+        co_yield "enum";
+        co_return;
       case fbs::type::Type::list_type:
+        co_yield "list";
+        co_return;
       case fbs::type::Type::map_type:
+        co_yield "map";
+        co_return;
       case fbs::type::Type::record_type:
+        co_yield "record";
```
mavam, line 745: The giant `switch` feels like it could be factored into a helper function that can be both called from `name()` and `names()`.

libvast/vast/type.hpp
mavam: The bigger problem is that every type has a name now, so `has_name` is confusing at best.

mavam (commented):
Looks good overall. I struggle with the function name `type::has_name` mostly.

libvast/vast/type.hpp
dominiklohmann: >  Users think in terms of aliases

Note that this API is for developers, not for users. The mental model here is that a type can be any of the concrete types with multiple metadata layers attached.

I agree that the name is confusing, but if we rename it, then we should also rename `has_attributes`, because I'd like to not break the symmetry between the two.

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -775,20 +775,31 @@ serialize_attributes(const std::vector<type::attribute_view>& attrs) {
...
     ++nesting_depth;
   }
+  VAST_ASSERT(nesting_depth > 0);
+  if (nesting_depth == 1)
+    return nullptr;
   if (nesting_depth == 0)
     return nullptr;
-  return metadata;
+  // Strip the last layer of metadata again, as that is guaranteed to be the
+  // types kind.
```
dominiklohmann, line 798:
```suggestion
  // type's kind.
```

libvast/vast/type.hpp
mavam: > Note that this API is for developers, not for users. 

I don't think it makes sense for the two to diverge, rather, I'd like to bring them closer together to avoid having to manage *two* mental models, and achieve a better DX. What's the problem with converging?

What's wrong with a free function to check whether a type is an alias, e.g., `is_alias(t)`? Then I know how to interpret `t.name()`.

tobim: I also think that `is_alias()` or the inverse `is_builtin()` are more common solution to this problem.

I was wondering why this function is even needed, currently we use it in `to_legacy_type()` and `assign_metadata()`.
Only the latter is interesting long term so maybe we can even localize the function?
I also just got the idea of representing the name as the `name` attribute instead of a separate field, but I'm not sure this would give us a net benefit.

mavam:
> I also just got the idea of representing the `name` as the name attribute instead of a separate field, but I'm not sure this would give us a net benefit.

This is an implementation detail in the flatbuffer.

I think it should be clear what a "type name" is and what an "alias name" is. In the end, they share the same namespace in schemas and modules. In fact, they *are* the same thing from a user's perspective. If I call `type::name()`, VAST should give me the builtin type name if it's not an alias, and the name of the alias otherwise. In that regard I think we are all on the same page.

What's missing is a mechanism to distinguish builtin vs. aliases. I have no strong opinion how, other than it should be completely independent of `type::name`. So far, `is_builtin(t)` strikes me as a good choice.

tobim:
> This is an implementation detail in the flatbuffer.

I mentioned that specifically in the other thread because `assign_metadata()` would be simplified to `assign_attributes` so that part have a simpler implementation.

libvast/vast/type.hpp
dominiklohmann: `is_builtin` is not the inverse of `is_alias`, and that is the main source of confusion here. What we need is a function that checks whether a type has a custom name, ignoring whether the type has any attributes.

Anyways, I came up with something I think we all can agree on, and implemented and pushed that as a separate commit. Here's the gist of it:
- A new private function `is_alias` that does exactly what `type::has_name` did before, that comes with a fat exclaimer that using it leads to bad code that is unlikely to be well-designed for the type system. It is only available within `type.cpp`.
- I restructured all uses of `type::has_name` outside of `type.cpp` not to rely on the function. These weren't all that many, so I think that's fine.

> I also just got the idea of representing the name as the name attribute instead of a separate field, but I'm not sure this would give us a net benefit.

Not implementable in a backwards-compatible manner, I think. We also call `type::name` and `type::names` in the hot path, and doing a binary search is just a waste of cycles here. The code paths that look at attributes are entirely separate from the code paths that look at names.

libvast/src/type.cpp
dominiklohmann: Actually this was not possible. I reduced it somewhat by naming the constants as `T::kind` for all `concrete_type T`.

libvast/vast/type.hpp
```diff
@@ -1096,8 +1101,8 @@ struct formatter<vast::type> {
   auto format(const vast::type& value, FormatContext& ctx)
     -> decltype(ctx.out()) {
     auto out = ctx.out();
-    if (const auto& name = value.name(); !name.empty())
-      out = format_to(out, "{}", name);
+    if (value.has_name())
+      out = format_to(out, "{}", value.name());
```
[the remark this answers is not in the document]
dominiklohmann: That's not true, because the formatter renders the signature instead of the kind.

dominiklohmann:
@mavam and I had a small chat about this PR and decided to scrap the whole idea. There's a better approach we can take for #2058, which is adding a generator that returns all actual aliases, i.e., metadata layers that have a custom name, either re-adding the attributes from metadata layers that were stripped away or stopping at the first metadata layer that has attributes. This would also allow for removing the `type::names()` function in its entirety.

A related bug fix for equality comparisons of types: the constructor for type that adds names and attributes is required to merge unnamed outermost metadata layers.

dominiklohmann:
Closing; will open a new PR soon with the replacement.

PR #2111, opened 2022-02-23
Move header files into include/ folder

lava:
To move closer to common C++ project layouts, move all header files into a `include/` subfolder. Medium-term, this will also make it easier to introduce distinct private and public header files when we start to introduce API or ABI versioning to libvast.

### :memo: Checklist

Review this pull request commit-by-commit


libvast/CMakeLists.txt
```diff
@@ -47,8 +47,8 @@ endif ()
 # -- flatbuffers ---------------------------------------------------------------
 
 file(GLOB flatbuffers_schemas CONFIGURE_DEPENDS
-     "${CMAKE_CURRENT_SOURCE_DIR}/vast/fbs/*.fbs"
-     "${CMAKE_CURRENT_SOURCE_DIR}/vast/fbs/legacy/*.fbs")
+     "${CMAKE_CURRENT_SOURCE_DIR}/include/vast/fbs/*.fbs"
+     "${CMAKE_CURRENT_SOURCE_DIR}/include/vast/fbs/legacy/*.fbs")
```
tobim, line 51: We should move those to a more logical place while we're at it
```suggestion
     "${CMAKE_CURRENT_SOURCE_DIR}/fbs/*.fbs"
     "${CMAKE_CURRENT_SOURCE_DIR}/fbs/legacy/*.fbs")
```
I think we should probably also install them?

mavam:
This is a breaking change, no?

dominiklohmann:
> This is a breaking change, no?

It shouldn't be. If done correctly, this only affects the internal structure. CMake still makes the same paths available.

dominiklohmann (approved):

PR #2112, opened 2022-02-24
Improve generic aggregate error handling

6yozo:
Improve generic aggregate error handling

### :memo: Checklist

### DIFF
--- a/plugins/aggregate/aggregate.cpp
+++ b/plugins/aggregate/aggregate.cpp
@@ -419,14 +419,11 @@ struct aggregation {
     auto cast_requirements
       = std::vector<std::optional<bool>>(builder_->num_fields(), std::nullopt);
     // Calculate whether a column needs to be casted only once.
-    auto needs_cast = [&](int column, std::shared_ptr<arrow::Scalar>& scalar) {
-      auto cast_requirement = cast_requirements[column];
-      if (!cast_requirement) {
-        if (!scalar)
-          return false;
-        cast_requirements[column] = std::optional<bool>(
-          !builder_->GetField(column)->type()->Equals(scalar->type));
-      };
+    auto needs_cast = [&](int column, const arrow::Scalar& scalar) {
+      auto& cast_requirement = cast_requirements[column];
+      if (!cast_requirement)
+        cast_requirement
+          = !builder_->GetField(column)->type()->Equals(scalar.type);
       return *cast_requirement;
     };
     builder_->SetInitialCapacity(detail::narrow_cast<int>(buckets_.size()));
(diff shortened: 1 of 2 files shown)


plugins/aggregate/aggregate.cpp
```diff
@@ -422,9 +421,12 @@ struct aggregation {
...
       result.resize(builder_->num_fields(), false);
       if (!buckets_.empty())
-        for (int column = 0; column < builder_->num_fields(); ++column)
-          result[column] = !builder_->GetField(column)->type()->Equals(
-            buckets_.begin()->second[column]->type);
+        for (int column = 0; column < builder_->num_fields(); ++column) {
+          auto scalar = buckets_.begin()->second[column];
+          result[column]
+            = scalar
+              && !builder_->GetField(column)->type()->Equals(scalar->type);
```
lava, line 428: With this change, can we remove the `&& scalar condition` in line 437 below?
```
if (column_needs_cast[column] && scalar) {
```

dominiklohmann, line 428: This change doesn't make any sense. You can only certainly say that you don't need to cast if the scalars for this column in _all_ buckets are null. This fix just exchanges one bug for another.

6yozo: When `column_needs_cast[column]` evaluates to `true` then it is still needed. On 425 the check is only done for the first field (`buckets_.begin()`), the check on 437 is done for each field.

6yozo: A better approach would be to change `column_needs_cast[column]` to three possible values: true, false, unknown. If it is unknown the algorithm can try to determine the value on each row until it can be determined.

6yozo: I have changed this behavior.

6yozo: I have changed this behavior accordingly.

plugins/aggregate/aggregate.cpp
```diff
@@ -417,22 +416,25 @@ struct aggregation {
...
+      return *cast_requirement;
+    };
     builder_->SetInitialCapacity(detail::narrow_cast<int>(buckets_.size()));
     for (auto&& bucket : std::exchange(buckets_, {})) {
       for (int column = 0; auto& scalar : bucket.second) {
-        auto* column_builder = builder_->GetField(column++);
+        auto* column_builder = builder_->GetField(column);
         VAST_ASSERT(column_builder);
-        if (column_needs_cast[column] && scalar) {
+        if (needs_cast(column++, scalar)) {
```
dominiklohmann, line 437: You introduced a subtle new bug here: `needs_cast` must always return false if `scalar` is a `nullptr`. Now it may return `true`, which causes you to call `scalar->CastTo(...)` on a nullptr.

I'd probably change it this way:

```cpp
if (scalar && needs_cast(column, *scalar)) {
  // cast
}
++column;
```

And then change the lambda like this:

```cpp
auto needs_cast = [&](int column, const arrow::Scalar& scalar) {
  auto& cast_requirement = cast_requirements[column];
  if (!cast_requirement)
    cast_requirement =
      !builder_->GetField(column)->type()->Equals(scalar.type));
  return *cast_requirement;
};
```

6yozo: I have changed the code accordingly.

dominiklohmann (approved):
I'm confident that this works as expected now.

PR #2117, opened 2022-02-25
Eploit synergies when evaluating many queries at the same time

tobim:
This set of changes introduces a new query scheduling algorithm that is designed to exploit synergies between queries candidate sets: When many queries are active simultaneously the sets of candidate partitions can be expected to overlap partially. To make use of this overlap we build a map from partition_id to the list query_ids interested in the partition. The partitions that ought to be materialized are selected based on the number of interested queries, and all queries get sent to the partition.

A measurement of the completion time of running 500 queries on a local database of 220 GB :
`master`: 155 seconds
this branch: 114,5 seconds

### :memo: Checklist


libvast/include/vast/system/index.hpp
```diff
@@ -277,25 +324,35 @@ struct index_state {
...
   /// The set of received but unprocessed queries.
   query_backlog backlog = {};
 
-  /// Maps query IDs to pending lookup state.
-  std::unordered_map<uuid, query_state> pending = {};
+  /// Maps query IDs to pending queries lookup state.
+  // std::unordered_map<uuid, query_state> pending_queries = {};
+
+  /// Maps candidate partition IDs to the queries .
+  // std::unordered_map<uuid, std::vector<uuid>> pending_partitions = {};
```
lava, line 336: Comments can probably be removed.

libvast/src/system/index.cpp
```diff
@@ -208,49 +208,152 @@ filesystem_actor& partition_factory::filesystem() {
...
   // Load partition from disk.
-  VAST_ASSERT(std::find(state_.persisted_partitions.begin(),
-                        state_.persisted_partitions.end(), id)
-              != state_.persisted_partitions.end());
+  if (std::find(state_.persisted_partitions.begin(),
+                state_.persisted_partitions.end(), id)
+      == state_.persisted_partitions.end()) {
+    VAST_WARN("{} can't load unknown partition {}", *state_.self, id);
+    // FIXME: returning a null actor crashes soon after.
+    // return {};
```
lava, line 216: This sounds like it still needs to be fixed

libvast/src/system/index.cpp
```diff
@@ -1056,144 +1214,63 @@ index(index_actor::stateful_pointer<index_state> self,
...
-      if (auto worker = self->state.next_worker()) {
-        VAST_VERBOSE("{} starts executing {} from a new request", *self, query);
-        return self->delegate(static_cast<index_actor>(self), atom::internal_v,
-                              std::move(query), std::move(*worker),
-                              sender->address());
-      }
-      VAST_VERBOSE("{} pushes query {} to the backlog", *self, query);
-      auto addr = self->current_sender()->address();
+      // `current_sender()` doesn't work after creating the response promise, so
+      // we must do it before.
```
lava, line 1215: We actually have it already stored in a variable `sender` here.

libvast/src/system/index.cpp
```diff
@@ -208,49 +208,152 @@ filesystem_actor& partition_factory::filesystem() {
...
+                      });
+    if (it != inactive_partitions.end()) {
+      VAST_INFO("found inactive partition");
+      it->queries.push_back(qid);
+      partitions.push_back(std::move(*it));
+      inactive_partitions.erase(it);
+      continue;
+    }
+    partitions.emplace_back(cand, std::vector{qid});
+    VAST_INFO("emplacing new {}", partitions.size());
```
lava: Leftover debug message

libvast/src/system/index.cpp
```diff
@@ -208,49 +208,149 @@ filesystem_actor& partition_factory::filesystem() {
...
+    inactive_partitions.push_back(std::move(inactive));
+  if (!active.queries.empty()) {
+    // TODO: Consider delaying the following loop until the next insert or
+    // activate.
+    for (const auto& qid : active.queries)
+      queries[qid].scheduled_partitions++;
+    return active;
   }
-  return std::nullopt;
+  return next();
```
lava, line 353: This is probably fine, but while thinking about it I found that clang has an `__attribute__((musttail))` that may be useful here.

lava (commented):
This was just from a first pass, I need to take a more in-depth look at the queue algorithm later.

In general I was wondering if it wouldn't be possible to keep the old worker system and add a feature toggle to decide what system the index uses? I understand that it would keep a lot of crufty code around, but on the other hand the risk in changing a core algorithm would be much smaller.

libvast/src/system/index.cpp
```diff
@@ -208,49 +208,149 @@ filesystem_actor& partition_factory::filesystem() {
...
+                               != p.queries.end();
+                      });
+  inactive_partitions = std::move(new_inactive);
+  std::sort(partitions.begin(), partitions.end());
+  return caf::none;
+}
+
+[[nodiscard]] caf::error pending_queue::remove_query(const uuid& qid) {
+  // If we are currently working on this query we don't remove it from the
+  // list yet.
```
lava, line 299: I'm not sure I understand the rationale here, when will we remove it instead? Does it imply that we will completely finish all queries as soon as at least one partition was handled?

libvast/include/vast/system/index.hpp
```diff
@@ -141,26 +141,84 @@ struct query_backlog {
...
+
+  [[nodiscard]] caf::error remove_query(const uuid& qid);
+
+  [[nodiscard]] size_t num_partitions() const;
+
+  [[nodiscard]] size_t num_queries() const;
+
+  [[nodiscard]] uuid create_query_id() const;
+
+  [[nodiscard]] std::optional<pq> next();
```
dominiklohmann, line 213: Please document. It's hard to understand what this is doing without going through all of the code.

vast/integration/reference/server-zeek-conn-log/step_07.ref
```diff
@@ -8,8 +8,6 @@
 {"path":["importer","ids","available"],"type":"number"}
 {"path":["importer","ids","block","end"],"type":"number"}
 {"path":["importer","ids","block","next"],"type":"number"}
-{"path":["index","backlog","num-low-priority"],"type":"number"}
-{"path":["index","backlog","num-normal-priority"],"type":"number"}
```
dominiklohmann, line 12: This removal needs a separate changelog entry.

libvast/include/vast/system/index.hpp
```diff
@@ -277,25 +324,35 @@ struct index_state {
...
+
+  /// A counter to track the number of partitions that are currently serving
+  /// lookups.
+  size_t running_partition_lookups = 0;
 
-  /// Caches idle/busy workers.
-  detail::stable_set<query_supervisor_actor> idle_workers = {};
-  detail::stable_set<query_supervisor_actor> busy_workers = {};
+  /// Keeps temporary statistics that are flushed with the metrics.
+  index_counters counters;
```
dominiklohmann, line 355: `= {}` because this otherwise contains garbage.

libvast/include/vast/system/index.hpp
```diff
@@ -315,6 +372,9 @@ struct index_state {
   /// Statistics about processed data.
   index_statistics stats = {};
 
+  /// Timekeeper for the scheduling algorithm.
+  struct measurement scheduler_measurement;
```
dominiklohmann, line 376: `= {}` as this may otherwise contain garbage.

libvast/include/vast/system/actors.hpp
```diff
@@ -263,19 +247,15 @@ using index_actor = typed_actor_fwd<
   caf::reacts_to<atom::subscribe, atom::create,
                  partition_creation_listener_actor, send_initial_dbstate>,
   // Evaluates a query, ie. sends matching events to the caller.
+  caf::replies_to<atom::evaluate, query, uint64_t>::with<query_response>,
```
dominiklohmann, line 250: Can you document what the additional parameter is for?

libvast/src/system/index.cpp
```diff
@@ -655,14 +668,21 @@ void index_state::add_partition_creation_listener(
...
-                      {"query.workers.busy", busy_workers.size()},
-                    }};
-  self->send(accountant, msg);
+  auto counters = std::exchange(this->counters, {});
+  auto msg = report{
+    .data = {
+      {"query.pending.partitions", pending_queries.num_partitions()},
+      {"query.pending.queries", pending_queries.num_queries()},
+      {"query.partition.materializations", counters.partition_materializations},
+      {"query.partition.lookups", counters.partition_lookups},
```
dominiklohmann, line 677: This is independent of the query, so it should probably be under the `index` category and without the metadata.

libvast/src/system/index.cpp
```diff
@@ -771,6 +788,107 @@ index_state::status(status_verbosity v) const {
...
+        }
+      };
+      st.self->request(partition_actor, caf::infinite, query_state.query)
+        .then(
+          [handle_completion, pid = next->partition, &st](uint64_t n) {
+            VAST_TRACE("{} received {} results from {}", *st.self, n, pid);
+            handle_completion();
+          },
+          [handle_completion, &st](const caf::error& err) {
+            VAST_WARN("{} failed in partition lookup: {}", *st.self, err);
```
dominiklohmann, line 883: Can we write this in a way that is meaningful to the user, e.g., somehow tell the user that for the query X we failed to get results from partition Y because of error Z?

libvast/src/system/index.cpp
```diff
@@ -1030,6 +1135,56 @@ index(index_actor::stateful_pointer<index_state> self,
...
+        .then([=, candidates = std::move(candidates)](
+                catalog_result& midx_result) mutable {
+          auto& midx_candidates = midx_result.partitions;
+          VAST_DEBUG("{} got initial candidates {} and from meta-index {}",
+                     *self, candidates, midx_candidates);
+          candidates.insert(candidates.end(), midx_candidates.begin(),
+                            midx_candidates.end());
+          std::sort(candidates.begin(), candidates.end());
+          candidates.erase(std::unique(candidates.begin(), candidates.end()),
+                           candidates.end());
```
dominiklohmann, line 1168: Can the meta index return duplicate candidate partitions?

libvast/src/system/index.cpp
```diff
@@ -1056,144 +1211,63 @@ index(index_actor::stateful_pointer<index_state> self,
...
```
dominiklohmann, line 1235: There seems to be a lot of duplication with the handler ~60 lines above. Can we remove that somehow?

libvast/src/system/index.cpp
```diff
@@ -1056,144 +1211,63 @@ index(index_actor::stateful_pointer<index_state> self,
...
-                      delta.count());
-      self->send(query_state.worker, atom::supervise_v, query_id,
-                 query_state.query, std::move(actors), client);
-      // Cleanup if we exhausted all candidates.
-      if (query_state.partitions.empty())
-        self->state.pending.erase(iter);
-      return {};
+    [self](atom::resolve,
+           vast::expression& expr) -> caf::result<catalog_result> {
+      auto lookup_id = vast::uuid::random();
```
dominiklohmann, line 1262: Please be consistent with the naming. Is a lookup ID a query ID?

libvast/src/system/index.cpp
```diff
@@ -1341,20 +1412,17 @@ index(index_actor::stateful_pointer<index_state> self,
...
+      query.id = self->state.pending_queries.create_query_id();
+      query.priority = 100;
+      auto input_size = detail::narrow_cast<uint32_t>(old_partition_ids.size());
+      auto err = self->state.pending_queries.insert(
+        query_state{.query = query,
+                    .client = caf::actor_cast<receiver_actor<atom::done>>(sink),
+                    .candidate_partitions = input_size,
+                    .requested_partitions = input_size},
+        std::vector{old_partition_ids});
+      VAST_ASSERT(err == caf::none);
```
dominiklohmann, line 1424: Why can't this fail?

dominiklohmann (commented):
A first glance at the code with a fair bit of questions. I did not try running this yet.

libvast/src/system/index.cpp
```diff
@@ -771,6 +788,107 @@ index_state::status(status_verbosity v) const {
...
+void schedule_lookups(index_state& st) {
+  if (st.pending_queries.partitions.empty())
+    return;
+  auto t = timer::start(st.scheduler_measurement);
+  auto num_scheduled = size_t{0};
+  auto on_return = caf::detail::make_scope_guard([&] {
+    t.stop(num_scheduled);
+  });
+  while (st.running_partition_lookups < st.max_concurrent_partition_lookups) {
+    // 1. Get the partition with the highest accumulated priority.
```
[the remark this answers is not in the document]
tobim: Before this PR we maintain a backlog with queries that weren't sent to the `catalog` yet. After this change we do the catalog lookup immediately. The purpose of this change is to avoid "unloading" partitions from memory when we know they will be needed for other queries as well. Maintaining completely separate queues doesn't make sense for this scheme, so we now use the priority as a scoring factor when sorting the queue.

libvast/src/system/index.cpp
```diff
@@ -208,49 +208,149 @@ filesystem_actor& partition_factory::filesystem() {
...
+  std::partition_copy(std::make_move_iterator(inactive_partitions.begin()),
+                      std::make_move_iterator(inactive_partitions.end()),
+                      std::back_inserter(partitions),
+                      std::back_inserter(new_inactive), [&](const auto& p) {
+                        return std::find(p.queries.begin(), p.queries.end(),
+                                         qid)
+                               != p.queries.end();
+                      });
+  inactive_partitions = std::move(new_inactive);
+  std::sort(partitions.begin(), partitions.end());
```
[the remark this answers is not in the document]
tobim: Unfortunately not, because we change priority values for existing items when an exporter requests another batch. Supporting that with a heap comes down to the same work but is more cumbersome to implement.

libvast/include/vast/system/index.hpp
```diff
@@ -141,26 +141,84 @@ struct query_backlog {
...
 
+struct index_counters {
+  size_t partition_materializations = 0;
+  size_t partition_lookups = 0;
+};
+
+struct pending_queue {
+  struct pq {
+    uuid partition;
+    std::vector<uuid> queries;
```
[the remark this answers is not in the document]
tobim: Good point. I'd like to use a helper template to abstract away the details, but we don't have that yet.

libvast/src/system/index.cpp
```diff
@@ -1030,6 +1135,56 @@ index(index_actor::stateful_pointer<index_state> self,
                       e);
           });
     },
+    [self](atom::evaluate, vast::query query,
+           uint64_t n) -> caf::result<query_response> {
```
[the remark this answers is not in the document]
tobim: Because the other case isn't implemented yet. I can actually remove this path because the `transform` handler doesn't even use it any more.

libvast/src/system/index.cpp
tobim: The logic in `schedule_lookups` is to send a `done` when the number of scheduled partitions has been handled. Now that you point it out I think it would be better to ignore it there and also remove running queries to avoid a potential leak.

libvast/include/vast/system/index.hpp
```diff
@@ -141,26 +141,84 @@ struct query_backlog {
...
   }
 };
 
+struct index_counters {
+  size_t partition_materializations = 0;
+  size_t partition_lookups = 0;
+};
+
+struct pending_queue {
+  struct pq {
```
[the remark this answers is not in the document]
tobim: My habit of abbreviating... was supposed to main "pending query". I think it makes sense to rename this to `entry` or `element`.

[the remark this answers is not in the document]
dominiklohmann: `value_type` would also make sense to be consistent with the standard library types.

libvast/src/system/index.cpp
tobim: We use the `create_query_id()` function that already deals with collisions. That is currently the only failure mode.

libvast/include/vast/query.hpp
```diff
@@ -135,8 +132,14 @@ struct query {
...
 
+  struct priority {
+    constexpr static uint8_t high = 80;
+    constexpr static uint8_t normal = 30;
+    constexpr static uint8_t low = 10;
+  };
+
   /// The query priority.
-  priority priority = priority::normal;
+  uint8_t priority = priority::normal;
```
dominiklohmann, line 142: This is not used anywhere.

libvast/src/system/index.cpp
```diff
@@ -771,6 +788,107 @@ index_state::status(status_verbosity v) const {
...
+                              == query_state.candidate_partitions
+                            ? query_response::complete
+                            : query_response::incomplete;
+          query_state.rp.deliver(response);
+        }
+        // 4. Remove queries with no more remaining partitions from
+        // pending_queries.
+        if (query_state.completed_partitions
+            == query_state.candidate_partitions)
+          st.pending_queries.queries.erase(qid);
```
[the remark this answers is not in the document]
tobim: I would expect that the query here is not reachable through the `partitions` and `inactive_partitions` structures any more.
I added an assertion to make this clear.

libvast/src/system/index.cpp
```diff
@@ -771,6 +788,107 @@ index_state::status(status_verbosity v) const {
...
+        // 5. recursively call schedule_lookups in the done handler. ...or
+        //    when all done? (6)
+        // 6. decrement st.running_partition_lookups when all queries that
+        //    were started are done. Keep track in the closure.
+        *cnt -= 1;
+        if (*cnt == 0) {
+          --st.running_partition_lookups;
+          schedule_lookups(st);
```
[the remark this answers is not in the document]
tobim: This gets called in a continuation to the request below. There is no real recursion here.

libvast/src/system/index.cpp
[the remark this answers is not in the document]
tobim: Removed.

libvast/src/system/index.cpp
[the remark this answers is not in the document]
tobim: TODO: We decide to make it possible to switch to the old behavior with a config option.

libvast/src/system/index.cpp
tobim: Not as far as I know, I simply inherited this part from the old implementation.

libvast/src/system/index.cpp
tobim: the other one is now gone.

libvast/include/vast/query.hpp
tobim: Fixed.

libvast/src/system/index.cpp
tobim: There is no `query` involved in this path, so it would be a bit misleading to name this variable `query_id`. This is also not an addition from this PR, it only shows up because of the merges.

libvast/include/vast/query_queue.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+  /// Creates an ID for a query and makes sure to avoid collisions with other
+  /// existing query IDs.
+  [[nodiscard]] uuid create_query_id() const;
+
+  /// Retrieves a handle to the contained queries.
+  [[nodiscard]] const std::unordered_map<uuid, query_state>& queries() const;
+
+  /// Retrieves a handle to the contained queries.
+  [[nodiscard]] std::unordered_map<uuid, query_state>& queries();
```
dominiklohmann, line 86: Afaict you only need the mutable overload in one place:

```c++
query_state.completed_partitions++;
```

I think the surrounding code section should be made a member function of the `query_queue`, then we can drop the unguarded access here.

libvast/include/vast/system/index.hpp
```diff
@@ -117,85 +118,19 @@ class partition_factory {
...
-  /// A response promise to signal batch completion to the client.
-  caf::typed_response_promise<query_response> rp = {};
-
-  template <class Inspector>
-  friend auto inspect(Inspector& f, query_state& x) {
-    return f(caf::meta::type_name("query_state"), x.query, x.client,
-             x.candidate_partitions, x.requested_partitions,
-             x.scheduled_partitions, x.completed_partitions);
-  }
+  index_state& state_;
```
dominiklohmann, line 121: This seems to break the usual separation of concerns. Why does the partition factory mutate the index state? Are the side effects of the partition factory documented?

libvast/include/vast/system/actors.hpp
tobim: This handler has been removed.

libvast/src/system/index.cpp
```diff
@@ -208,49 +208,149 @@ filesystem_actor& partition_factory::filesystem() {
...
+    std::back_inserter(active.queries), std::back_inserter(inactive.queries),
+    [&](const auto& qid) {
+      const auto& qs = queries[qid];
+      return qs.requested_partitions > qs.scheduled_partitions;
+    });
+  if (!inactive.queries.empty())
+    inactive_partitions.push_back(std::move(inactive));
+  if (!active.queries.empty()) {
+    // TODO: Consider delaying the following loop until the next insert or
+    // activate.
```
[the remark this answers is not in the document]
tobim: My local measurements indicate no bottleneck here currently. It doesn't even show up in the profiling.

libvast/src/system/index.cpp
```diff
@@ -831,38 +703,56 @@ void schedule_lookups(index_state& st) {
...
+          if (query_state.completed_partitions
+              == query_state.requested_partitions)
+            st.self->send(query_state.client, atom::done_v);
+          // 4. Remove queries with no more remaining partitions from
+          // pending_queries.
+          if (query_state.completed_partitions
+              == query_state.candidate_partitions) {
+            VAST_ASSERT(!st.pending_queries.reachable(qid));
+            st.pending_queries.queries().erase(qid);
+          }
```
dominiklohmann, line 754: This should go into a separate function that takes the query id and returns whether you should send the `atom::done_v` to the client.

See also the above comment on removing the mutable `queries` overload.

dominiklohmann (commented):
Some additional feedback. This is starting to look really good, I like the separation of concerns with the query queue. This is so much better than the old mechanism.

libvast/src/system/index.cpp
tobim: I'm not sure I understand, this report doesn't have any metadata attached? And we use the `query.` prefix for closely related metrics like the worker counts. I really don't like to add them to an `index` category, because I think the evaluation of the non-active parts of the database should be moved to a different component eventually anyways.

libvast/src/system/index.cpp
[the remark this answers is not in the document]
dominiklohmann: Good, resolving then.

libvast/src/system/index.cpp
dominiklohmann: I assumed that the new `query.*` metrics carry the same query ID as all the pre-existing metrics in the same category. Everything else would be unintuitive to the user. Maybe we can use the `scheduler.*` prefix here to anticipate the index -> scheduler + partition-builder split?

libvast/src/system/index.cpp
```diff
@@ -674,8 +676,9 @@ void schedule_lookups(index_state& st) {
     }
     VAST_TRACE("{} schedules partition {} for {}", *st.self, next->partition,
                next->queries);
-    // 2. Spin up the partition.
-    auto materialize = [&](const uuid& partition_id) -> partition_actor {
+    // 2. Aquire the actor for the selected partition, potentially materializing
+    //    it from its persisted state.
+    auto aquire = [&](const uuid& partition_id) -> partition_actor {
```
dominiklohmann, line 681: Typo in both documentation and lambda name: acquire contains a c.

libvast/src/system/index.cpp
```diff
@@ -945,21 +1059,12 @@ index(index_actor::stateful_pointer<index_state> self,
...
       for (const auto& id : ids) {
-        auto worker = self->state.pending[id].worker;
-        self->state.pending.erase(id);
-        // We might have already received the `atom::worker` from this worker,
-        // but discarded it because the query still had pending work. In this
-        // case, `atom::shutdown` should cause it to send another one.
-        if (worker)
-          self->send(worker, atom::shutdown_v, atom::sink_v);
+        if (auto err = self->state.pending_queries.remove_query(id))
+          VAST_DEBUG("{} {}", *self, err);
```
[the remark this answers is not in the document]
tobim: True, but it isn't critical. I turned it into a warning.

libvast/include/vast/system/index.hpp
tobim: It's done so we can keep a counter for the materializations. I'll add a note.

dominiklohmann: Can we move the materializations counter to the factory then?

libvast/include/vast/system/index.hpp
```diff
@@ -277,25 +324,35 @@ struct index_state {
...
   /// Maps exporter actor address to known query ID for monitoring
   /// purposes.
   std::unordered_map<caf::actor_addr, std::unordered_set<uuid>> monitored_queries
     = {};
 
-  /// The number of query supervisors.
-  size_t workers = 0;
+  /// The maximum number of partitions to serve queries at the same time.
+  /// Assigned from the `max-queries` config option.
+  /// TODO: Rename that option appropriately and deprecate the old name.
```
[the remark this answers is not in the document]
tobim: The corresponding config option is called `max-queries`. I think this is close enough that we don't need to add an alias.

libvast/src/system/index.cpp
tobim: Should we rename the others in that block as well then?

dominiklohmann: I think that makes sense as they're also not tied to a particular query.

libvast/src/system/index.cpp
tobim: I just tried this again and the issue has been resolved in the meantime. Instead of the commented out return I changed the warning message because it is more robust this way.

libvast/src/system/index.cpp
```diff
@@ -707,21 +707,9 @@ void schedule_lookups(index_state& st) {
...
-        if (query_state.completed_partitions
-            == query_state.requested_partitions)
-          st.self->send(query_state.client, atom::done_v);
-        if (query_state.completed_partitions
-            == query_state.candidate_partitions)
-          st.pending_queries.queries().erase(qid);
-      }
+      for (auto qid : next->queries)
+        if (auto client = st.pending_queries.handle_completion(qid))
+          st.self->send(*client, atom::done_v);
```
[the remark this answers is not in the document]
tobim: That would indeed be great, but the current index client protocol doesn't allow it. I already had a handler to do this but removed it again in e616c1326f4fa78a71322ef876c5f9968014edab because it wasn't necessary for this PR.

libvast/src/system/index.cpp
tobim: I just loopified the algorithm :shrug:.

libvast/src/query_queue.cpp
```diff
@@ -129,55 +129,56 @@ query_queue::activate(const uuid& qid, uint32_t num_partitions) {
...
-    std::make_move_iterator(result.queries.begin()),
-    std::make_move_iterator(result.queries.end()),
-    std::back_inserter(active.queries), std::back_inserter(inactive.queries),
-    [&](const auto& qid) {
-      auto it = queries_.find(qid);
-      if (it == queries_.end()) {
-        VAST_WARN("index tried to access non-existant query {}", qid);
-        // Consider it inactive.
-        return false;
+  while (!partitions.empty()) {
```
dominiklohmann, line 135: I think the if-clause above can be removed outright, it's already covered by the while-loop.

changelog/unreleased/changes/2117--scheduler-metrics.md
```diff
@@ -0,0 +1,23 @@
...
+                                            this metric was issued.
+* `scheduler.partition.schedulings` -> number of partitions that were used to
+*                                      evaluate queries since the last time
+                                       this metric was issued.
+* `scheduler.partition.lookups` -> number of queries that were sent to
+                                   partitions since the last time this metric
+                                   was issued.
+
+We also removed the keys `index.backlog.num-normal-priority` and
+`index.backlog.num-low-priority` from the output of `vast status`.
```
dominiklohmann, line 23: Two notes on this one:

- Active voice, present tense. 💣
- Currently, the changelog script joins all the lines within paragraphs, and then turns every paragraph into one changelog entry. This breaks your current formatting here, so I recommend merging the first two paragraphs and writing the middle one as text. If you build locally then the generated CHANGELOG.md will be in the build directory, which you can use to double-check whether the formatting works correctly.

libvast/include/vast/system/index.hpp
```diff
@@ -117,48 +118,21 @@ class partition_factory {
...
 
-  /// Cancels jobs associated with the given sender.
-  /// @returns The number of cancelled jobs.
-  uint64_t cancel(caf::actor_addr sender);
-
-  [[nodiscard]] std::optional<job> take_next();
-
-  std::deque<job> normal;
-  std::deque<job> low;
+  // Mutable reference so we can increment the materializations counter.
```
dominiklohmann, line 122: Would it be possible to move the counter here instead?

dominiklohmann (commented):
Continuing the stream of review comments.

vast/integration/reference/server-zeek-conn-log/step_07.ref
tobim: We decided to bring this back in the status output and the metrics.

changelog/unreleased/changes/2117--scheduler-metrics.md
tobim: Fixing...

libvast/include/vast/system/index.hpp
tobim: Yes that is also possible.

libvast/src/system/index.cpp
```diff
@@ -1341,20 +1412,17 @@ index(index_actor::stateful_pointer<index_state> self,
...
+      query.priority = 100;
+      auto input_size = detail::narrow_cast<uint32_t>(old_partition_ids.size());
+      auto err = self->state.pending_queries.insert(
+        query_state{.query = query,
+                    .client = caf::actor_cast<receiver_actor<atom::done>>(sink),
+                    .candidate_partitions = input_size,
+                    .requested_partitions = input_size},
+        std::vector{old_partition_ids});
+      VAST_ASSERT(err == caf::none);
+      schedule_lookups(self->state);
```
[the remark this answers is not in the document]
dominiklohmann: We discussed this in a call and decided that there's really no way to make this simpler with the current communication protocol with the clients.

libvast/src/system/index.cpp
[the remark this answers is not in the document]
tobim: This has been implemented in 8beee922e94d50fd41384575ad925bc9a34e2eeb.

dominiklohmann (approved):
Approving modulo docs update. Thanks so much for making this change, this is a big leap forward! 🚀

PR #2119, opened 2022-02-27
Log actor names together with the unique actor ID

tobim:
This changes the log output of `*self` from `name` to `name-id`.

For example `exporter` would now be formatted as `exporter-16` or `exporter-48`. This makes it easier to track the execution flow of specific instances of actors that exist multiple times in the same process.

### :memo: Checklist

### DIFF
--- a/changelog/unreleased/changes/2119--log-actors-with-id.md
+++ b/changelog/unreleased/changes/2119--log-actors-with-id.md
@@ -1,0 +1,2 @@
+Actor names in log messages now have an `-ID` suffix to make it easier to tell
+multiple instances of the same actor apart.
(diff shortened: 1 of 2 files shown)


dispanser (approved):

changelog/unreleased/changes/2119--log-actors-with-id.md
```diff
@@ -0,0 +1,2 @@
+Actor names in log messages now have an `-ID` suffix to make it easier to tell
+multiple instances of the same actor apart.
```
dominiklohmann, line 2: Should this explain that the ID is globally unique, i.e., that index-22 does not mean that at least 22 index actors have existed?

tobim: I see no reason to assume that from the given text.

PR #2120, opened 2022-02-27
Correctly terminate the explore command

tobim:
This commit fixes 2 issues that prevented the exporter for the initial query from exiting.

* When the explorer received the requested number of results for the initial query it did not communicate this to the exporter, causing the latter to wait for a follow-up request indefinitely.
* When the `max-events-query * max-events-context` was lower than `max-events` the sink would wait for additional results but the `explorer` would only wait to get terminated.

### :memo: Checklist

Run an explore command on a database with more than 5 partitions and an initial query with a low selectivity.

### DIFF
--- a/libvast/src/system/explore_command.cpp
+++ b/libvast/src/system/explore_command.cpp
@@ -8,6 +8,7 @@
 
 #include "vast/system/explore_command.hpp"
 
+#include "vast/concepts.hpp"
 #include "vast/defaults.hpp"
 #include "vast/error.hpp"
 #include "vast/logger.hpp"
@@ -33,12 +34,22 @@
 using namespace caf;
 using namespace std::chrono_literals;
 
+namespace vast::detail {
+
+template <concepts::unsigned_integral T>
+auto non_overflowing_multiply(T lhs, T rhs) {
+  if (lhs > (std::numeric_limits<T>::max() / rhs))
+    return std::numeric_limits<T>::max();
+  return lhs * rhs;
+}
+
+} // namespace vast::detail
+
 namespace vast::system {
 
 caf::message explore_command(const invocation& inv, caf::actor_system& sys) {
   using namespace std::string_literals;
   VAST_DEBUG("{}", inv);
-  const auto& options = inv.options;
   if (auto error = explorer_validate_args(inv.options))
     return make_message(error);
   // Read options and arguments.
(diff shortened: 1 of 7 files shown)


libvast/vast/system/explorer.hpp
```diff
@@ -26,6 +26,7 @@ struct explorer_state {
   struct event_limits {
     uint64_t total;
     uint64_t per_result;
+    uint64_t initial_query;
```
mavam, line 29:
```suggestion
    uint64_t taste_size;
```

The issue I have with "initial query" is that it doesn't reflect a quantity.

dominiklohmann: Our [style guide](https://github.com/tenzir/.github/blob/master/contributing.md#naming) recommends prefixing these kinda variables with `num_`:

> Prefix counter variables with `num_`.

libvast/src/system/explore_command.cpp
```diff
@@ -33,12 +34,22 @@
 using namespace caf;
 using namespace std::chrono_literals;
 
+namespace vast::detail {
+
+template <concepts::unsigned_integral T>
+auto non_overflowing_multiply(T lhs, T rhs) {
```
lava, line 40: Some nitpicking:

At least in rust this is called a `saturating_mul` (and I think I've heard the term elsewhere as well)

Also, you could save the division by using `__builtin_mul_overflow()` here.

Since it's already in `vast::detail` the intention probably was to pull it into a separate `detail/` header?

libvast/src/system/explore_command.cpp
```diff
@@ -47,11 +58,24 @@ caf::message explore_command(const invocation& inv, caf::actor_system& sys) {
...
+    = caf::get_or(inv.options, "vast.explore.max-events-context",
+                  defaults::explore::max_events_context);
+  uint64_t max_events = caf::get_or(inv.options, "vast.explore.max-events",
+                                    defaults::explore::max_events);
   // Get a local actor to interact with `sys`.
   caf::scoped_actor self{sys};
+  auto alternative_max_events
+    = detail::non_overflowing_multiply(max_events_query, max_events_context);
+  if (alternative_max_events < max_events) {
+    max_events = max_events_query * max_events_context;
```
lava, line 74: Is this just `max_events = alternative_max_events`?

lava (approved):
The fix makes sense to me.

libvast/vast/system/explorer.hpp
tobim: How did *that* get into the guide? The CONTRIBUTING.md that was removed from this repository didn't have that but the one that was added to the other repo had it from the beginning?

In this case the surrounding instance variable is called `limits`, which already indicates that the contained values are numeric.

PR #2121, opened 2022-02-27
Remove the get subcommand

tobim:
The `vast get` command was never useful except for debugging and it can't be made ergonomic with partition local ids, so it's better to remove it.

### :memo: Checklist

Search for leftovers?


dominiklohmann (approved):
The only real use case we had for this was the initial version of the Live Matching feature back in 2020. I don't see why we need to drag this feature around any longer. Let's get a step closer to having partition-local ID spaces!

PR #2123, opened 2022-02-28
Correctly handle partition transforms without output

lava:
The case where a partition transform removed all input data
and returned an empty output (which is important when deleting
partitions) was not handled correctly, leading to potential
segfaults.

This was not caught by test because we lacked an end-to-end test for whole-partition erasure, with existing tests all mocking the partition transformer.

### :memo: Checklist

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -1253,10 +1253,14 @@ index(index_actor::stateful_pointer<index_state> self,
         .then(
           [self, rp, old_partition_ids, new_partition_id,
            keep](augmented_partition_synopsis& aps) mutable {
+            // If the partition was completely deleted, `synopsis` may be null.
+            auto events = aps.synopsis ? aps.synopsis->events : 0ull;
+            auto time = aps.synopsis ? aps.synopsis->max_import_time
+                                     : vast::time::clock::time_point{};
             auto result = partition_info{
               .uuid = aps.uuid,
-              .events = aps.synopsis->events,
-              .max_import_time = aps.synopsis->max_import_time,
+              .events = events,
+              .max_import_time = time,
               .stats = std::move(aps.stats),
             };
             // Update the index statistics. We only need to add the events of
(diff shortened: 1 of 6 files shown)


libvast/test/system/partition_transformer.cpp
```diff
@@ -336,4 +336,86 @@ TEST(partition transform via the index) {
   self->send_exit(index, caf::exit_reason::user_shutdown);
 }
 
+TEST(deleting a partition using a transform) {
+  // Spawn index and fill with data
+  auto index_dir = std::filesystem::path{"/vast/index"};
+  auto archive = self->spawn(mock_archive);
```
tobim, line 342: Isn't it sufficient to use a default constructed `archive_actor` here?

libvast/src/system/index.cpp
```diff
@@ -1253,66 +1253,81 @@ index(index_actor::stateful_pointer<index_state> self,
...
+                        [](const caf::error& e) {
+                          VAST_WARN("index failed to erase {} from meta index",
+                                    e);
                         });
-                  },
-                  [rp](const caf::error& e) mutable {
-                    rp.deliver(e);
-                  });
-            }
-            for (size_t i = 1; i < old_partition_ids.size(); ++i) {
```
lava, line 1306: Btw, it looked to me like a separate bug that the `for`-loop was outside the scope of the `else`-branch so i moved it inside, but please verify @6yozo / @dominiklohmann .

libvast/test/system/partition_transformer.cpp
```diff
@@ -336,4 +336,86 @@ TEST(partition transform via the index) {
...
+      REQUIRE_EQUAL(partition_synopsis->partition_synopsis_type(),
+                    vast::fbs::partition_synopsis::PartitionSynopsis::legacy);
+      const auto* partition_synopsis_legacy
+        = partition_synopsis->partition_synopsis_as_legacy();
+      const auto* range = partition_synopsis_legacy->id_range();
+      events = range->end() - range->begin();
+    },
+    [](const caf::error& e) {
+      REQUIRE_SUCCESS(e);
+    });
```
tobim, line 398: It looks like `events` isn't used anywhere afterwards.

libvast/test/system/partition_transformer.cpp
```diff
@@ -336,4 +336,86 @@ TEST(partition transform via the index) {
   self->send_exit(index, caf::exit_reason::user_shutdown);
 }
 
+TEST(deleting a partition using a transform) {
```
tobim, line 339:
```suggestion
TEST(select transform with an empty result set) {
```

tobim (commented):
I only reviewed the unit test so far.

libvast/src/system/index.cpp
dominiklohmann: Yes, almost certainly a bug. Thanks for noticing/fixing. Hopefully I can find the time to finish the co_result<T> stuff on the sooner side such that we can avoid similar issues caused by the nesting depth making the code harder to follow in the future.

Feel free to resolve.

libvast/src/system/index.cpp
```diff
@@ -1253,66 +1253,81 @@ index(index_actor::stateful_pointer<index_state> self,
...
-                          rp.deliver(result);
-                        },
-                        [=](const caf::error& e) mutable {
-                          rp.deliver(e);
+              // other ones are "deleted" from the meta index. If the new
+              // partition is empty, all partitions are deleted.
+              auto start = size_t{0};
+              if (aps.synopsis) {
+                start = 1;
+                auto old_partition_id = old_partition_ids.at(0);
```
tobim, line 1294: You can avoid the explicit index management by choosing the last partition with `back()` and `pop_back()` instead.

libvast/src/system/index.cpp
```diff
@@ -1253,66 +1253,81 @@ index(index_actor::stateful_pointer<index_state> self,
...
+                      rp.deliver(e);
+                    });
+              } else {
+                rp.deliver(result);
+              }
+              for (size_t i = start; i < old_partition_ids.size(); ++i) {
+                auto partition_id = old_partition_ids[i];
+                self
+                  ->request(self->state.meta_index, caf::infinite,
+                            atom::erase_v, partition_id)
```
tobim, line 1324: It is not clear to me why we delegate the first erase to the index but the remaining ones to the meta index. Can you add an explanation?

tobim (approved):
I think the change fixes the issue but the code needs a bit of cleanup before it can be merged.

libvast/src/system/partition_transformer.cpp
```diff
@@ -78,18 +110,28 @@ void partition_transformer_state::fulfill(
   partition_transformer_actor::stateful_pointer<partition_transformer_state>
     self,
   stream_data&& stream_data, path_data&& path_data) const {
+  VAST_DEBUG("{} fulfilling promise", *self); // FIXME: remvoe
```
dominiklohmann, line 113: Outstanding fixme

libvast/src/system/index.cpp
lava: Good catch, I think that was an (old) bug, they should all go to the index. @dominiklohmann can you confirm?

dominiklohmann (approved):

libvast/src/system/index.cpp
dominiklohmann: We went over this, I think this is the correct change to make.

lava: We just agreed that this was indeed a bug in a video call.

PR #2125, opened 2022-03-01
Remove the concrete `none_type`

dominiklohmann:
The `none_type` is a blast from the past: It originated in the old type system that required an explicit variant member for signaling the empty type, which is not a requirement for the new type system. This will allow for removing `arrow::NullType` in VAST, which was—until now—required for symmetry in generic code like the custom sum type access for `arrow::Array` and `arrow::DataType`.

Note that type is still semi-regular—nothing changed about that. This change is purely internal.

### :memo: Checklist

I recommend going file-by-file. The changes in `project.hpp` are probably the most complicated, but other than that it should be easy to understand.

### DIFF
--- a/libvast/src/arrow_table_slice.cpp
+++ b/libvast/src/arrow_table_slice.cpp
@@ -858,10 +858,7 @@ auto values([[maybe_unused]] const Type& type, const Array& arr)
     if (arr.IsNull(i)) {
       co_yield {};
     } else {
-      if constexpr (std::is_same_v<Type, none_type>) {
-        VAST_ASSERT(false, "none_type not expected here");
-        __builtin_unreachable();
-      } else if constexpr (std::is_same_v<Type, bool_type>) {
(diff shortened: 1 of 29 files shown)


tobim:
Looks like the broker plugin broke your code.

dominiklohmann:
> Looks like the broker plugin broke your code.

That should be fixed now.

dispanser (approved):
I read the changes and verified that `none_type` doesn't exist in the code anymore.

I did see a few comments in `type.hpp` that reference `none_type`. I'm not sure if we somehow can reword those, given that they reference something that doesn't actually exist anymore.

dominiklohmann:
> I did see a few comments in `type.hpp` that reference `none_type`. I'm not sure if we somehow can reword those, given that they reference something that doesn't actually exist anymore.

I'll rewrite those from none_type to none type.

PR #2128, opened 2022-03-01
Rename `meta-index` to `catalog`

dominiklohmann:
We've long been searching for a new name for the meta index, as there exist already multiple things named index inside VAST. An internal vote resulted in the name `catalog` winning.

This change also affects a keys in the output of `vast status` and in certain metrics, and is thus a breaking change.

### :memo: Checklist

Verify that git-grep only shows old changelog entries for relevant queries.


mavam (commented):
These first substitutional changes are fine.

It feels that renaming the `index` should be done in here as well.

dominiklohmann:
> It feels that renaming the `index` should be done in here as well.

A quick vote in Slack said that this should be a separate PR.

mavam (approved):

PR #2133, opened 2022-03-03
Add a grand total event counter to the status output

6yozo:
Add a grand total event counter to the status output

### :memo: Checklist

### DIFF
--- a/changelog/unreleased/features/2133--status-grand-total.md
+++ b/changelog/unreleased/features/2133--status-grand-total.md
@@ -1,0 +1,2 @@
+The output of `vast status --detailed` now displays the total number of events
+stored in the DB.
--- a/changelog/unreleased/features/2133--status-grand-total.md
+++ b/changelog/unreleased/features/2133--status-grand-total.md
@@ -1,2 +1,2 @@
-The output of `vast status --detailed` now displays the total number of events
-stored in the DB.
+The output of `vast status` now displays the total number of events stored in
+the DB under the key `index.statistics.events.total`.
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -642,8 +642,8 @@ index_state::status(status_verbosity v) const {
       sum += layout_stats.count;
     }
     auto xs = record{};
-    xs["count"] = count{sum};
-    layout_object["total"] = xs;
+    xs["total"] = count{sum};
+    stats_object["all-layouts"] = xs;
     stats_object["layouts"] = std::move(layout_object);
     rs->content["statistics"] = std::move(stats_object);
     rs->content["catalog-bytes"] = catalog_bytes;
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -634,11 +634,16 @@ index_state::status(status_verbosity v) const {
   if (v >= status_verbosity::detailed) {
     auto stats_object = record{};
     auto layout_object = record{};
+    auto sum = uint64_t{0};
     for (const auto& [name, layout_stats] : stats.layouts) {
       auto xs = record{};
       xs["count"] = count{layout_stats.count};
       layout_object[name] = xs;
+      sum += layout_stats.count;
     }
+    auto xs = record{};
+    xs["count"] = count{sum};
+    layout_object["total"] = xs;
     stats_object["layouts"] = std::move(layout_object);
     rs->content["statistics"] = std::move(stats_object);
     rs->content["catalog-bytes"] = catalog_bytes;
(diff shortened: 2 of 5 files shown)


libvast/src/system/index.cpp
```diff
@@ -634,11 +634,16 @@ index_state::status(status_verbosity v) const {
...
+    auto sum = uint64_t{0};
     for (const auto& [name, layout_stats] : stats.layouts) {
       auto xs = record{};
       xs["count"] = count{layout_stats.count};
       layout_object[name] = xs;
+      sum += layout_stats.count;
     }
+    auto xs = record{};
+    xs["count"] = count{sum};
+    layout_object["total"] = xs;
```
dominiklohmann, line 646: Doesn't this override statistics if we ever have a schema named `total`?

dominiklohmann (commented):
Some minor remarks:
- The first two commits in this look unrelated to me.
- There's an integration test that needs an adjustment.
- Right now there's an ambiguity if there's a schema named `total`, which I think we can avoid by putting the sum in the outer statistics object.

changelog/unreleased/features/2133--status-grand-total.md
```diff
@@ -0,0 +1,2 @@
+The output of `vast status --detailed` now displays the total number of events
+stored in the DB.
```
dominiklohmann, line 2: As a user the first question I have when reading this is: Under what key?

vast/integration/reference/server-zeek-conn-log/step_07.ref
```diff
@@ -16,6 +16,7 @@
 {"path":["index","num-cached-partitions"],"type":"number"}
 {"path":["index","num-unpersisted-partitions"],"type":"number"}
 {"path":["index","partitions","active",0,"id"],"type":"string"}
+{"path":["index","statistics","all-layouts","total"],"type":"number"}
```
tobim, line 19: Imho the extra `all-layouts` key doesn't really add value, how about just:
```suggestion
{"path":["index","statistics","events"],"type":"number"}
```

changelog/unreleased/features/2133--status-grand-total.md
```diff
@@ -0,0 +1,2 @@
+The output of `vast status --detailed` now displays the total number of events
```
tobim, line 1: I think that information is important enough to be displayed on the `info` level.

tobim (changes requested):
I think we should emit this info for `vast status` without `--detailed` too.

vast/integration/reference/server-zeek-conn-log/step_07.ref
dominiklohmann: We could also switch the structure more generally, e.g.,

```json
"events": {
  "total": 123,
  "schema": {
    "suricata": {
      "alert": 1,
      "flow": 19
    },
    "zeek": {
      "conn": 100,
      "http": 3
    }
  }
}
```

6yozo: So the new output of `vast status --detailed` should be like:
```yaml
{
  ...
  "index": {
    ...
    "events": {
      "total": 25711
       schema": {
         "suricata": {
           "alert": 1,
           "flow": 19
       },      
       ...     
    },
    ...
  },
  ...
}
```

6yozo: Do we achieve that format by splitting the `name` in `stats.layouts` by dots?

https://github.com/tenzir/vast/blob/622a00dc8c5501c97bc8df8e5e9fdeb1deb6674f/libvast/src/system/index.cpp#L638

dominiklohmann: It's just an idea for an improved UX. Sounds like a good thing to discuss in Slack quickly based on a list of possible outputs. We need to break here anyways given that we want to `s/schema/module` and `s/layout/schema`, and while we're at it we should also think about the ergonomics.

dominiklohmann (approved):

libvast/src/system/index.cpp
```diff
@@ -631,19 +631,22 @@ index_state::status(status_verbosity v) const {
...
   auto rs = make_status_request_state<extra_state>(self);
+  auto stats_object = record{};
+  auto sum = uint64_t{0};
+  for (const auto& [_, layout_stats] : stats.layouts)
+    sum += layout_stats.count;
+  auto xs = record{};
+  xs["total"] = count{sum};
+  stats_object["events"] = xs;
+  if (v == status_verbosity::info)
+    rs->content["statistics"] = std::move(stats_object);
```
tobim, line 642: nit: We could also do this assignment unconditionally at the end (before the `fill_status_map` stuff). I don't have a strong opinion though so you are free to ignore this suggestion.

tobim (approved):

PR #2130, opened 2022-03-02
Tightly couple VAST's and Arrow's type systems

dominiklohmann:
This couples the VAST and Arrow type systems a bit more by making the conversion from and to Arrow part of VAST's type system directly, and adding type traits that allow for dispatching from VAST type to Arrow type directly. Additionally, this specializes the Arrow type traits for VAST's Arrow extension types.

### :memo: Checklist

- [x] All user-facing changes have changelog entries.
- [x] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [x] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

Go file-by-file. I recommend the following order:
1. Type system
2. Write path
3. Read path
4. Everything else

Open to pairing on this in a call. Sorry for this becoming one large commit—it's been a bit messy to figure out all the required bits and pieces here.

PR #2136, opened 2022-03-04
Enable efficient conversion from arrow to experimental table slices

dispanser:
Efficiently creates an `experimental` table slice from a `arrow::v1` table slice by re-using the existing `arrow::Array` structures and regrouping them into the proper (nested) structure.

Exceptions:
- `subnet` is fully rebuilt (from fixed width binary(17) to `struct(address, array)`)
- `enumeration` is derived via `arrow::compute::Cast` b/c we're using `Int16` instead of previously `UInt64`)

All other array types are used "as is"

### :memo: Checklist

The change basically is one big function (`convert_record_batch`).

Node that some final touches rely on #2130, so this is a WIP PR:
  - proper `ArrayBuilder` implementations 
  - a final, shared place for some enum dict array creation


dispanser:
@dominiklohmann I think it makes sense to merge your changes first, and finish this one when you're done, given that it needs the proper array builders to work sanely. 

I'm still looking for feedback on the implementation itself, maybe there are spots not marked with `TODO` that would benefit from changes in your PR #2130

libvast/src/arrow_table_slice.cpp
```diff
@@ -1277,4 +1278,155 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+      return std::make_shared<arrow::MapArray>(mt, la->length(),
+                                               la->value_offsets(),
+                                               la->values(), la->null_bitmap(),
+                                               la->null_count());
+    },
+    [&](const std::shared_ptr<arrow::Array>&,
+        const std::shared_ptr<arrow::StructType>& st)
+      -> std::shared_ptr<arrow::Array> {
+      // move back the iterator so it points to the first array in the struct
+      --array_iterator;
```
dominiklohmann, line 1394: This reads like a hack to me, but I'm not sure what a better way to implement this would be.

libvast/src/arrow_table_slice.cpp
```diff
@@ -1277,4 +1278,155 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+      VAST_ASSERT(ba->byte_width() == 17);
+      auto subnet_array_data = convert_subnet_array(ba);
+      // TODO: the type is not actually correct, we have
+      // left:  struct<address: fixed_size_binary[16], length: uint8>
+      // right: struct<address: extension<vast.address>, length: uint8>
+      // so we can't create the subnet_array either (arrow asserts storage
+      // type) this gets "fixed" by arrow when re-reading the record batch
+      // in, but the record batch under construction exposes the wrong type
+      // information. to fix this, we require proper AddressBuilder /
+      // SubnetBuilder IIUC
```
dominiklohmann, line 1367: A better way to write this would be to have this overload first split the 17-byte-width Array into a 16-byte-width Array and the UInt8Array, and then to recursively call this function again with he 16-byte-width Array to convert it to an address type. This would need to be done regardless of whether or not we have a custom builder for addresses and subnets.

Until we have said builder, I think you must also call `{subnet, address}_extension_type::MakeArray` on the underlying `arrow::ArrayData` of the extension arrays.

libvast/src/arrow_table_slice.cpp
```diff
@@ -1277,4 +1278,155 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+    }
+  }
+  if (auto result = subnet_builder->Finish(); result.ok())
+    return *result;
+  die("failed to finish Arrow subnet column builder");
+}
+
+// TODO: this is copied from
+// experimental_table_slice_builder/enum_column_builder, can this live somewhere
+// shared?
```
dominiklohmann, line 1318: This goes away with the new enum builder in #2130, so I think we can just leave it for now.

libvast/src/arrow_table_slice.cpp
```diff
@@ -1277,4 +1278,155 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+  }
+  die("failed to finish Arrow enum field array");
+}
+
+/// Consumes arrow arrays and builds up the appropriate array based on the
+/// provided data type.
+/// @param array_iterator An iterator to consume column arrays.
+/// @param arrow_type the data type we're building the array for.
+std::shared_ptr<arrow::Array>
+make_arrow_array(arrow::ArrayVector::const_iterator& array_iterator,
```
dominiklohmann, line 1335: How does this function handle `list<T>`? and `map<T, U>` when T or U are extension types? I think that's still missing, and we can probably do it recursively.

libvast/src/arrow_table_slice.cpp
```diff
@@ -1277,4 +1278,155 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+      field_names.reserve(st->num_fields());
+      for (const auto& field : st->fields()) {
+        children.push_back(make_arrow_array(array_iterator, field->type()));
+        field_names.push_back(field->name());
+      }
+      auto res = arrow::StructArray::Make(children, field_names);
+      return res.ValueOrDie();
+    },
+  };
+  return caf::visit(f, *array_iterator++, arrow_type);
```
dominiklohmann, line 1407: Does it maybe make more sense to only dispatch on the VAST type here, because we don't want to have to support the old Array's in the sum type access in the future anymore? We can then hardcode the static casts for the old Arrow encoding, because that is set in stone anyways.

libvast/vast/arrow_table_slice.hpp
```diff
@@ -150,4 +150,11 @@ extern template class arrow_table_slice<fbs::table_slice::arrow::v0>;
 extern template class arrow_table_slice<fbs::table_slice::arrow::v1>;
 extern template class arrow_table_slice<fbs::table_slice::arrow::experimental>;
 
+// -- utility functions --------------------------------------------------------
+
+/// Converts legacy (v1) table slice to the new representation, re-using
```
dominiklohmann, line 155: We can also use this on the v0—the record batches had the exact same structure.

dominiklohmann (commented):
Some comments on the code.

I think this is a great improvement, but the `make_arrow_array` probably needs to dispatch on VAST types rather than Arrow arrays because it needs to be fully recursive for nested Arrays.

libvast/src/arrow_table_slice.cpp
dispanser: You're right, those cases are not covered.

libvast/src/arrow_table_slice.cpp
dispanser: It is a hack, indeed. Initially I wanted to avoid `arrow_iterator++` in all branches. But when adding handlers for nests and list types, suddenly `--` is becoming more widespread as well, so I'll make the `++` explicit in according branches.

libvast/src/arrow_table_slice.cpp
dispanser: I'm not sure. I used the additional argument to the visitor only for some type safety and to avoid the explicit casting, and I have no strong opinion either way.

libvast/src/arrow_table_slice.cpp
dispanser: all fixed now: `list<T>`, `map<T, U>`, `list<record>`. Also refactored the conversion a bit to make a clearer separation between structural unrolling and "data mapping", i.e. `arrow::Array -> arrow::Array`

libvast/src/arrow_table_slice.cpp
dispanser: I would propose to wait for your changes and rebase / fix in this branch when the builders are available.

libvast/src/arrow_table_slice.cpp
dispanser: changed according to the suggestion. It's annoying to explicitely cast, but then we can handle types that are not part of `sum_type_access` for array, in particular `FixedSizeBinaryArray`

libvast/src/arrow_table_slice.cpp
```diff
@@ -702,7 +704,10 @@ value_at([[maybe_unused]] const Type& type,
         value_type at(size_type i) const override {
           VAST_ASSERT(!key_array->IsNull(value_offset + i));
           if (item_array->IsNull(value_offset + i))
-            return {};
+            return {
+              value_at(key_type, *key_array, value_offset + i),
+              {},
+            };
```
dispanser, line 710: @dominiklohmann the previous implementation created an empty record entry when the value was absent, instead of a record entry with a null value only. I came across this while writing a test for the conversion itself, and verified that this new behavior is identical to the legacy table slice behavior, so I assume it's "more correct".

libvast/src/arrow_table_slice.cpp
```diff
@@ -1141,4 +1146,171 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+  }
+  die("failed to finish Arrow enum field array");
+}
+
+/// Converts an array representing a single column from a previous arrow
+/// format to `arrow::experimental`.
+std::shared_ptr<arrow::Array>
+convert_column(const std::shared_ptr<arrow::Array>& arr, const type& t) {
+  auto f = detail::overload{
+    [&](const auto& t) -> std::shared_ptr<arrow::Array> {
```
dominiklohmann, line 1199: Can we restrain this to basic types?

```suggestion
    [&](const basic_type auto& t) -> std::shared_ptr<arrow::Array> {
```

libvast/src/arrow_table_slice.cpp
```diff
@@ -1141,4 +1146,171 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+    if (!dict_builder.Append(std::string{f.name}).ok())
+      die("failed to build Arrow enum field array");
+  if (auto array = dict_builder.Finish(); array.ok()) {
+    return *array;
+  }
+  die("failed to finish Arrow enum field array");
+}
+
+/// Converts an array representing a single column from a previous arrow
+/// format to `arrow::experimental`.
```
dominiklohmann, line 1195: I think the experimental mention can already be removed.

libvast/test/experimental_table_slice.cpp
```diff
@@ -43,6 +44,17 @@ auto make_slice(const record_type& layout, Ts&&... xs) {
...
+template <class T, class... Ts>
+auto make_legacy_slice(const record_type& layout, std::vector<T> x0,
+                       std::vector<Ts>... xs) {
+  auto builder = arrow_table_slice_builder::make(type{"rec", layout});
+  for (size_t i = 0; i < x0.size(); ++i) {
+    CHECK(builder->add(x0.at(i)));
+    CHECK(builder->add(xs.at(i)...));
+  }
+  return builder->finish();
+}
```
dominiklohmann, line 56: This will not work anymore son. Can you maybe create a record batch directly?

libvast/src/experimental_table_slice_builder.cpp
```diff
@@ -270,10 +270,11 @@ void verify_record_batch(const arrow::RecordBatch& record_batch) {
   auto check_col
     = [](auto&& check_col, const arrow::Array& column) noexcept -> void {
     auto f = detail::overload{
-      [&](const arrow::StructArray& sa) noexcept {
-        for (const auto& column : sa.fields())
-          check_col(check_col, *column);
-      },
+      // TODO: temporarily disabled, waiting for extensinon type array builders
```
dominiklohmann, line 273: These exist now.

libvast/src/arrow_table_slice.cpp
```diff
@@ -1141,4 +1146,171 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+    [&](const subnet_type&) -> std::shared_ptr<arrow::Array> {
+      auto ba = static_pointer_cast<arrow::FixedSizeBinaryArray>(arr);
+      VAST_ASSERT(ba->byte_width() == 17);
+      return convert_subnet_array(*ba);
+    },
+    [&](const enumeration_type& et) -> std::shared_ptr<arrow::Array> {
+      auto dict = make_dict_array(et);
+      auto indices = arrow::compute::Cast(*arr, arrow::uint8());
+      auto dict_array = std::make_shared<arrow::DictionaryArray>(
+        et.to_arrow_type()->storage_type(), indices.ValueUnsafe(), dict);
```
dominiklohmann, line 1238: The `ValueUnsafe()` should be preceded with an assertion.

libvast/src/arrow_table_slice.cpp
```diff
@@ -1141,4 +1146,171 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+      int index = 0;
+      for (const auto& f : rt.fields()) {
+        const auto& arr = sa->field(index++);
+        children.push_back(convert_column(arr, f.type));
+        field_names.emplace_back(f.name);
+      }
+      auto res = arrow::StructArray::Make(children, field_names);
+      if (!res.ok())
+        die("unable to construct nested struct array");
+      return *res;
```
dominiklohmann, line 1219: I think there's a `MoveValueUnsafe()` function on `arrow::Result` that should be suitable here. As-is, this won't be picked up by NRVO. Not a big deal, but more of a general C++ programming note.

libvast/src/arrow_table_slice.cpp
```diff
@@ -1141,4 +1146,171 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+    [&](const record_type& rt) -> std::shared_ptr<arrow::Array> {
+      // this case handles VAST type `list<record>`
+      const auto& sa = static_pointer_cast<arrow::StructArray>(arr);
+      arrow::ArrayVector children{};
+      std::vector<std::string> field_names{};
+      children.reserve(rt.num_fields());
+      field_names.reserve(rt.num_fields());
+      int index = 0;
+      for (const auto& f : rt.fields()) {
+        const auto& arr = sa->field(index++);
```
dominiklohmann, line 1212: I think `detail::zip` makes this easier to read.

libvast/src/arrow_table_slice.cpp
```diff
@@ -1141,4 +1146,171 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+      return std::make_shared<address_type::array_type>(
+        address_type::to_arrow_type(), arr);
+    },
+    [&](const subnet_type&) -> std::shared_ptr<arrow::Array> {
+      auto ba = static_pointer_cast<arrow::FixedSizeBinaryArray>(arr);
+      VAST_ASSERT(ba->byte_width() == 17);
+      return convert_subnet_array(*ba);
+    },
+    [&](const enumeration_type& et) -> std::shared_ptr<arrow::Array> {
+      auto dict = make_dict_array(et);
```
dominiklohmann, line 1235: You should not need this, as this is done in the constructor of the builder already. If you do `et.to_arrow_builder()` you get a custom builder that inherits from the `arrow::DictionaryBuilder` with all the fields pre-configured.

libvast/src/arrow_table_slice.cpp
```diff
@@ -1141,4 +1146,171 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+      auto indices = arrow::compute::Cast(*arr, arrow::uint8());
+      auto dict_array = std::make_shared<arrow::DictionaryArray>(
+        et.to_arrow_type()->storage_type(), indices.ValueUnsafe(), dict);
+      return std::make_shared<enumeration_type::array_type>(et.to_arrow_type(),
+                                                            dict_array);
+    },
+    [&](const list_type& lt) -> std::shared_ptr<arrow::Array> {
+      auto la = static_pointer_cast<arrow::ListArray>(arr);
```
dominiklohmann, line 1243: You never need this as a shared pointer, so why pointer cast over casting to a const-lvalue-ref of the underlying element type?

libvast/src/arrow_table_slice.cpp
```diff
@@ -1141,4 +1146,171 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+      return *res;
+    },
+    [&](const pattern_type&) -> std::shared_ptr<arrow::Array> {
+      return std::make_shared<pattern_type::array_type>(
+        pattern_type::to_arrow_type(), arr);
+    },
+    [&](const address_type&) -> std::shared_ptr<arrow::Array> {
+      return std::make_shared<address_type::array_type>(
+        address_type::to_arrow_type(), arr);
+    },
```
dominiklohmann, line 1228: Aren't these just `t.to_arrow_type()->MakeArray(arr->data())`?

libvast/src/arrow_table_slice.cpp
```diff
@@ -1141,4 +1146,171 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+  if (auto array = dict_builder.Finish(); array.ok()) {
+    return *array;
+  }
+  die("failed to finish Arrow enum field array");
+}
+
+/// Converts an array representing a single column from a previous arrow
+/// format to `arrow::experimental`.
+std::shared_ptr<arrow::Array>
+convert_column(const std::shared_ptr<arrow::Array>& arr, const type& t) {
```
dominiklohmann, line 1197: I think the naming of this function could use a bit of thought. From the name alone I would think this takes an arbitrary arrow Array and aligns it with our type system. But that isn't what this does.

libvast/src/arrow_table_slice.cpp
```diff
@@ -1141,4 +1146,171 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+  }
+  if (auto result = subnet_builder->Finish(); result.ok())
+    return *result;
+  die("failed to finish Arrow subnet column builder");
+}
+
+// TODO: this is copied from
+// experimental_table_slice_builder/enum_column_builder, can this live somewhere
+// shared?
+std::shared_ptr<arrow::Array> make_dict_array(const enumeration_type& et) {
```
dominiklohmann, line 1183: This function should not be needed after the other recommended changes.

libvast/src/arrow_table_slice.cpp
```diff
@@ -1141,4 +1146,171 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+
+std::shared_ptr<arrow::Array>
+convert_subnet_array(const arrow::FixedSizeBinaryArray& arr) {
+  auto* pool = arrow::default_memory_pool();
+  auto subnet_builder = subnet_type::make_arrow_builder(pool);
+  auto& address_builder = subnet_builder->address_builder();
+  auto& length_builder = subnet_builder->length_builder();
+  if (!address_builder.Resize(arr.length()).ok()
+      || !length_builder.Resize(arr.length()).ok()
+      || !subnet_builder->Resize(arr.length()).ok())
```
dominiklohmann, line 1161: Shouldn't this be `Reserve` rather than `Resize`, because we don't actually want to write zeroes?

libvast/src/arrow_table_slice.cpp
```diff
@@ -1141,4 +1146,171 @@ template class arrow_table_slice<fbs::table_slice::arrow::v0>;
...
+      auto span = std::span<const uint8_t, 17>{arr.GetValue(row), 17};
+      if (!(subnet_builder->Append().ok()
+            && address_builder.Append(arr.GetValue(row)).ok()
+            && length_builder.Append(span[16]).ok()))
+        die("rebuilding of subnet struct array failed");
+    }
+  }
+  if (auto result = subnet_builder->Finish(); result.ok())
+    return *result;
+  die("failed to finish Arrow subnet column builder");
```
dominiklohmann, line 1177: With how many things can go wrong in this function, should we consider making it return an `arrow::Result` or `caf::expected` instead?

dominiklohmann (commented):
This is really close to being done, I think. Some minor remarks inline.

libvast/src/arrow_table_slice.cpp
dominiklohmann: That's definitely a bug fix, thanks for catching this!

libvast/src/arrow_table_slice.cpp
dispanser: I don't think that works because one is a `vector`, the other a generator:

```
vast/libvast/src/arrow_table_slice.cpp:1211:72: error: no matching function for call to ‘zip(vast::detail::generator<vast::record_type::field_view>, const ArrayVector&)’
```

Do you have a suggestion on how to fix that?

dominiklohmann: That's probably just overly constrained, because I don't understand why `zip` wouldn't work with input iterators. Maybe just leave it as-is then, not worth spending time on now imo.

libvast/src/arrow_table_slice.cpp
dominiklohmann: Fair enough, that answers my question. Thanks for checking!

libvast/src/arrow_table_slice.cpp
dominiklohmann: If the slice append method just calls Append multiple times under the hood, we can also just call our `enumeration_type::arrow_builder::Append` function directly with `detail::narrow_cast<enumeration>(value)` in a loop. Could probably throw in a `Reserve` once in the beginning for good measure to avoid re-allocations. Open to pairing on this quickly if this is unclear.

libvast/src/arrow_table_slice.cpp
dispanser: from the resize docs:
```
  /// \brief Ensure that enough memory has been allocated to fit the indicated
  /// number of total elements in the builder, including any that have already
  /// been appended.
```

`Reserve` actually calls `Resize`, so I think that I'm using the right function, just the names are somewhat unconventional :)

libvast/src/arrow_table_slice.cpp
dispanser: from the resize docs:
```
  /// \brief Ensure that enough memory has been allocated to fit the indicated
  /// number of total elements in the builder, including any that have already
  /// been appended.
```

`Reserve` actually calls `Resize`, so I think that I'm using the right function, just the names are somewhat unconventional :)

libvast/src/arrow_table_slice.cpp
dispanser: I made it return an `arrow::Status` because that's what all the potential error paths from this function return anyway, so I could just forward them.

libvast/src/arrow_table_slice.cpp
dominiklohmann: That's really unfortunate naming then. Oh well.

libvast/src/arrow_table_slice.cpp
dispanser: Do you have a proposal? I see what you mean, but I don't have a good idea.
- `convert_array`
- `convert_array_to_latest_version`
- `transform_...` (recombine with above)

libvast/src/arrow_table_slice.cpp
dispanser: The current implementation doesn't have to loop at all, because we directly construct the `DictionaryArray` from its components. Since enum is one of the columns that actually occur with some frequency in the current schemas, I'd like to avoid all potential costly rebuilding. I prefer explicitely building the dictionary array to keep that performance over going through the builder.

I came up with
```c++
      auto b = et.make_arrow_builder(arrow::default_memory_pool());
      auto empty_result = b->Finish();
      if (!empty_result.ok()) {
        die("unable to finish empty enum dictionary builder");
      }
      const auto& dict
        = static_cast<const arrow::DictionaryArray&>(
            *static_cast<const arrow::ExtensionArray&>(**empty_result).storage())
            .dictionary();
```

to leverage the current builder implementation or dictionary creation while avoiding the full rebuild. This is a bit ugly, but maybe we can massage that to make it better?

dominiklohmann: I came up with this diff. Still kind of unergonomic, but I think this is at least somewhat better.

```diff
diff --git a/libvast/include/vast/type.hpp b/libvast/include/vast/type.hpp
index 90b095fe0..151354ea8 100644
--- a/libvast/include/vast/type.hpp
+++ b/libvast/include/vast/type.hpp
@@ -937,6 +937,11 @@ public:
 
     [[nodiscard]] std::shared_ptr<arrow::DictionaryArray> storage() const;
 
+    /// Create an array from a prepopulated indices Array.
+    static arrow::Result<std::shared_ptr<enumeration_type::array_type>>
+    make(const std::shared_ptr<enumeration_type::arrow_type>& type,
+         const std::shared_ptr<arrow::UInt8Array>& indices);
+
   private:
     using arrow::ExtensionArray::storage;
   };
@@ -997,6 +1002,7 @@ public:
 struct enumeration_type::arrow_type final : arrow::ExtensionType {
   friend class type;
   friend struct enumeration_type::builder_type;
+  friend struct enumeration_type::array_type;
 
   /// A unique identifier for this extension type.
   static constexpr auto name = "vast.enumeration";
diff --git a/libvast/src/type.cpp b/libvast/src/type.cpp
index f7db35af1..6d620b8c5 100644
--- a/libvast/src/type.cpp
+++ b/libvast/src/type.cpp
@@ -2135,6 +2135,24 @@ void enumeration_type::arrow_type::register_extension() noexcept {
   VAST_ASSERT(status.ok());
 }
 
+arrow::Result<std::shared_ptr<enumeration_type::array_type>>
+enumeration_type::array_type::make(
+  const std::shared_ptr<enumeration_type::arrow_type>& type,
+  const std::shared_ptr<arrow::UInt8Array>& indices) {
+  auto dict_builder
+    = string_type::make_arrow_builder(arrow::default_memory_pool());
+  for (const auto& [canonical, internal] : type->vast_type_.fields()) {
+    const auto append_status = dict_builder->Append(
+      arrow::util::string_view{canonical.data(), canonical.size()});
+    VAST_ASSERT(append_status.ok(), append_status.ToString().c_str());
+  }
+  ARROW_ASSIGN_OR_RAISE(auto dict, dict_builder->Finish());
+  ARROW_ASSIGN_OR_RAISE(auto storage,
+                        arrow::DictionaryArray::FromArrays(
+                          string_type::to_arrow_type(), indices, dict));
+  return std::make_shared<enumeration_type::array_type>(type, storage);
+}
+
 enumeration_type::builder_type::builder_type(std::shared_ptr<arrow_type> type,
                                              arrow::MemoryPool* pool)
   : arrow::StringDictionaryBuilder(string_type::to_arrow_type(), pool),
```

libvast/test/experimental_table_slice.cpp
dispanser: Hm, that's a bummer - of course you're right. On the other hand, the entire idea of the test is to construct a valid old table slice, and I don't really feel like re-implementing half of the "old" table slice builder logic to create proper outdated record batches.

An alternative would be to write the table slice into some sort of fixture, which would make the test considerably easier to implement. I _think_ that's fine because we already lost the ability to write old table slices by then, so the only source for them is some previously written partition anyway.

libvast/src/arrow_table_slice.cpp
dispanser: It's finally gone, thanks for helping with this.

libvast/src/arrow_table_slice.cpp
dispanser: this mostly works, and it's better (as in, more centralized).
Adopted, and thanks!

libvast/src/arrow_table_slice.cpp
mavam: `upgrade_array`?

dispanser: I'm willing to go with `upgrade_array_to_v2`. It's also in an anomymous namespace - I don't think this should block us.

dominiklohmann (approved):
Thanks! Let's merge this. 🚀

mavam:
Great to see this merged!

PR #2138, opened 2022-03-04
Merge unnamed metadata layers in nested types

dominiklohmann:
This is best explained by the unit test that this PR adds. A bug that is rather hard to run into, but really annoying when you actually run into it.

### :memo: Checklist

I discovered this while working on #2130, and figured I'd fix this in a separate PR because this should get reviewed separately.

Look at the unit test first, then at the implementation. Users were theoretically able to run into this, but that'd require such a weird setup that I don't think we need to add a changelog entry.

### DIFF
--- a/libvast/src/type.cpp
+++ b/libvast/src/type.cpp
@@ -232,7 +232,7 @@ type::type(chunk_ptr&& table) noexcept {
 }
 
 type::type(std::string_view name, const type& nested,
-           const std::vector<struct attribute>& attributes) noexcept {
+           const std::vector<attribute_view>& attributes) noexcept {
   if (name.empty() && attributes.empty()) {
     // This special case fbs::type::Type::exists for easier conversion of legacy
     // types, which did not require an legacy alias type wrapping to have a name.
@@ -274,10 +274,12 @@ type::type(std::string_view name, const type& nested,
           if (const auto* attributes = enriched_type->attributes()) {
             for (const auto* attribute : *attributes) {
               VAST_ASSERT(attribute->key());
-              merged_attributes.push_back({
-                attribute->key()->string_view(),
-                attribute->value() ? attribute->value()->string_view() : "",
-              });
+              if (attribute->value())
+                merged_attributes.push_back(
+                  {attribute->key()->string_view(),
+                   attribute->value()->string_view()});
+              else
+                merged_attributes.push_back({attribute->key()->string_view()});
             }
           }
           nested_bytes = as_bytes(*enriched_type->type());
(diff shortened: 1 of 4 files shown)


libvast/src/type.cpp
```diff
@@ -261,18 +309,32 @@ type::type(std::string_view name, const type& nested,
...
           builder, key_offset, value_offset));
       }
+      for (const auto& attribute : merged_attributes) {
+        // Skip over any attributes that were already in the new list of
+        // attributes.
+        if (std::any_of(attributes.begin(), attributes.end(),
+                        [&](const auto& new_attribute) noexcept {
+                          return new_attribute.key == attribute.key;
+                        }))
+          continue;
```
tobim, line 331: Have you considered changing the attributes parameter to those constructors to `std::vector<attribute_view>`? Since we're constructing a new flatbuffer here that shouldn't give us lifetime problems.

With that change you can start out with the provided `attributes` in the upper loop and apply the deduplication there, then you only need to deal with the merged attributes for the remainder of the function.

libvast/src/type.cpp
```diff
@@ -238,7 +238,55 @@ type::type(std::string_view name, const type& nested,
...
+          if (const auto* attributes = enriched_type->attributes()) {
+            for (const auto* attribute : *attributes) {
+              VAST_ASSERT(attribute->key());
+              merged_attributes.push_back({
+                attribute->key()->string_view(),
+                attribute->value() ? attribute->value()->string_view() : "",
+              });
+            }
+          }
+          nested_bytes = as_bytes(*enriched_type->type());
```
tobim, line 283: nit: This introduces an additional error into the `reserved_size` pre-calculation.

dominiklohmann: I don't see how. Can you elaborate? We only pre-calculate that if we do not have attributes in any form.

libvast/src/type.cpp
dominiklohmann: We'd probably need both overloads then, as we have `attribute` at many call sizes, and while we can make `attribute` implicitly convertible to `attribute_view` we cannot make that for containers of them. I do agree though that this would be a good API improvement.

libvast/src/type.cpp
dominiklohmann: We talked about this in a call. The code is correct here.

libvast/src/type.cpp
dominiklohmann: @tobim kindly provided me with a diff that implements most of this change. I'll incorporate that in a bit.

libvast/src/type.cpp
```diff
@@ -271,15 +270,25 @@ type::type(std::string_view name, const type& nested,
...
+              // Skip over any attributes that were already in the new list of
+              // attributes.
+              if (std::any_of(
+                    attributes.begin(), attributes.end(),
+                    [&](const auto& attribute) noexcept {
+                      return attribute.key
+                             == stripped_attribute->key()->string_view();
+                    }))
+                continue;
+              VAST_ASSERT(stripped_attribute->key());
```
tobim, line 284: You can move the assertion to the start of the current loop.

tobim (approved):

PR #2135, opened 2022-03-03
Tune AWS README

mavam:
This PR updates structure and content in the AWS README.

### :dart: Review Instructions

Read it.

PR #2139, opened 2022-03-04
Add option to use EFS as storage for the VAST server

rdettai:
This adds EFS as an alternative storage for the VAST server, which enables persistence across Fargate task executions.

### :memo: Checklist

The entrypoint is a new terraform variable called vast_server_storage_type which enables choosing between EFS and ATTACHED storage (main.tf). The rest is the basic config to mount EFS on Fargate.

It might we worth checking whether the access restrictions to the access point are tight enough, but this is likely not critical.

### DIFF
--- a/cloud/aws/fargate/ecs.tf
+++ b/cloud/aws/fargate/ecs.tf
@@ -23,52 +23,13 @@ resource "aws_security_group" "ecs_tasks" {
   }
 }
 
-resource "aws_security_group" "efs" {
-  name        = "${module.env.module_name}_${var.name}_efs_${module.env.stage}"
-  description = "allow inbound nfs access"
-  vpc_id      = var.vpc_id
-
-  ingress {
-    protocol        = "tcp"
-    from_port       = 2049
-    to_port         = 2049
-    security_groups = [aws_security_group.ecs_tasks.id]
-  }
-
-}
-
-resource "aws_efs_file_system" "efs_volume" {
-  count = var.storage_type == "EFS" ? 1 : 0
-  tags = {
-    Name = "${module.env.module_name}-${var.name}-${module.env.stage}"
-  }
-}
-
-resource "aws_efs_mount_target" "efs_target" {
-  count           = var.storage_type == "EFS" ? 1 : 0
-  file_system_id  = aws_efs_file_system.efs_volume[0].id
-  subnet_id       = var.subnet_id
-  security_groups = [aws_security_group.efs.id]
-}
-
-
-resource "aws_efs_access_point" "access_point_for_fargate" {
-  count          = var.storage_type == "EFS" ? 1 : 0
-  file_system_id = aws_efs_file_system.efs_volume[0].id
-
-  root_directory {
-    path = "/storage"
-    creation_info {
-      owner_gid   = 1000
-      owner_uid   = 1000
-      permissions = "755"
-    }
-  }
-
-  posix_user {
-    gid = 1000
-    uid = 1000
-  }
+module "efs" {
+  source                    = "./efs"
+  count                     = var.storage_type == "EFS" ? 1 : 0
+  name                      = var.name
+  vpc_id                    = var.vpc_id
+  subnet_id                 = var.subnet_id
+  ingress_security_group_id = aws_security_group.ecs_tasks.id
 }
 
 resource "aws_ecs_task_definition" "fargate_task_def" {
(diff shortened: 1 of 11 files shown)


rdettai:
Pending on https://github.com/tenzir/vast/pull/2135 before updating the README.

cloud/aws/modules.tf
```diff
@@ -13,6 +14,9 @@ module "vast_server" {
   task_cpu    = 2048
   task_memory = 4096
 
+  storage_type        = var.vast_server_storage_type
+  storage_mount_point = "/var/lib/vast"
```
mavam, line 18: Curious to hear @dominiklohmann's or @tobim's opinion on hardcoding this vs. making the DB dir configurable. I personally don't have a strong opinion.

dispanser: IMHO this is more of an implementation detail of the deployment.

cloud/aws/fargate/ecs.tf
```diff
@@ -23,6 +23,54 @@ resource "aws_security_group" "ecs_tasks" {
...
+    protocol        = "tcp"
+    from_port       = 2049
+    to_port         = 2049
+    security_groups = [aws_security_group.ecs_tasks.id]
+  }
+
+}
+
+resource "aws_efs_file_system" "efs_volume" {
+  count = var.storage_type == "EFS" ? 1 : 0
```
dispanser, line 41: Is it possible to create a module that contains all of the EFS-related logic? I see this ternary expression popping up several times, and moving the logic into a module might allow us to only make the EFS yes / no decision only once.

dispanser (commented):
Nice change, IMHO. Not approving because I didn't test yet.

I have a minor discussion point about the ability to move the EFS bits into some sort of module to keep the main terraform expression simpler.

cloud/aws/modules.tf
dominiklohmann: I agree with @dispanser here, this is an implementation detail of the deployment, and `/var/lib/vast` sounds like a good value for that.

rdettai: Also, the approach here is to avoid having too many different input variables to keep things as simple as possible for a newcomer.

cloud/aws/fargate/ecs.tf
rdettai: Good idea! I am used to do only one level of modules but here it would be very appropriate to have a second one 😄. Thanks!

cloud/aws/modules.tf
mavam: All good points. KISS.

dispanser (approved):
I tested the PR by:
- deploying
- importing
- re-deploying
- making sure the data was still there (via `vast count`)

This works as expected, which is great. Now we have an 8EB vast instance :-)

Things that didn't work, but are not strictly related to this PR:

```
make run-lambda CMD="wget -O - -o /dev/null https://raw.githubusercontent.com/tenzir/vast/master/vast/integration/data/suricata/eve.json  | vast import suricata"
sh: line 1: vast: command not found
```
Works for simple things, like `vast status`. I worked around this by using `make execute-command` instead and trigger the import from the VAST server instance

```
nix> make restart-vast-server
make: *** No rule to make target 'get-vast-tasks'.  Stop.
Started task c09ddb56f4024acdbb184b7cb16a1801
```
We end up with two VAST instances because the existing one isn't stopped, but the new one is started nevertheless.

Approving b/c the problems seem unrelated.

rdettai:
A few KPIs regarding ingestions rates in various setups:
- Region eu-central-1
- No tests using Lambda as client were performed.

```bash
wget -O - -o /dev/null https://cloudfuse-tenzir-test-data-eu-central-1.s3.eu-central-1.amazonaws.com/eve.json.gz | gzip -d | vast import suricata
```

Results:
| Type | Hardware | Event ingestion rate (k/s) |
| --- | --- | --- |
| LOCAL | Dell precision 8CPU+SSD+100Mb dl | 54 |
| EFS | Fargate 2 VCPU | 27 |
| ATTACHED | Fargate 2 VCPU | 27 |
| EFS | Fargate 4 VCPU | 45 |
| ATTACHED | Fargate 4 VCPU | 45 |

Note:
- performances are not always consistent, for instance on one of the runs for `ATTACHED / Fargate 2 VCPU` I got rates 20% higher
- the interesting result is that EFS and ATTACHED have the same performances for these instance sizes

mavam:
Really interesting to see the benchmark here. We're only taking a ~10% hit when compared to local, which I find remarkable at this early stage.

PR #2140, opened 2022-03-07
Pin VAST version in cloud development to release version

rdettai:
We want a more stable yet flexible way to manage the VAST version that is deployed to AWS. To achieve that:
- we pin the used vast version to the latest release by default
- we make the version configurable through a Terraform variable (or from `default.env` for make users)

### :memo: Checklist

- `Makefile`: To simplify how Terraform input variables are provided by make, I switched to environment variables (`TF_VAR_xxx`) instead of arguments (`--var=xxx`). 
- `conf.tf`: I removed the per workspace configuration which would not be usable as is anyway because we cannot easily provide variables per workspace without using Terraform Cloud.

The rest is a mechanical propagation of the newly defined variable.


rdettai:
Pending on https://github.com/tenzir/vast/pull/2135 before updating the README

dispanser (approved):
Verified that setting `vast_version` leads to different vast versions being deployed in fargate.

rdettai:
@dominiklohmann I think this is ready for merging

PR #2142, opened 2022-03-08
Make VAST tools run in a separate AWS VPC

rdettai:
It is more flexible to have the SOC in a separate VPC than the application that it monitors. This PR changes the AWS deployment scripts of VAST to comply with this good practice rather than spawning a new subnet in an existing VPC.

### :memo: Checklist

- ⚠️ input variable names changed, update your `default.env` file accordingly
- All the networking resources were moved to a separate module. This module splits the resources according to where they are (VAST/monitoring VPC or peered/monitored VPC).
- I added a new provider, aliased as `monitored_vpc`. This allows to later create settings where the monitored VPC and the VAST VPC are not in the same account/region.


rdettai:
I deployed an EC2 instance in the target VPC with a permissive security group and checked that I could ping it, it worked. The best would be to start setting up playground modules to have some better use cases to test this.

PR #2143, opened 2022-03-10
Rename the schema class to module

6yozo:
Rename the schema class to module

### :memo: Checklist


6yozo (commented):
When reviewing this PR, consider that it should contain only those renames that are not visible to the users.

Take into consideration that this PR may err on renaming too many things.

dominiklohmann (approved):
Thanks for being so careful with the renaming here. I think you actually caught all places that are not exposed to the user.

PR #2144, opened 2022-03-11
Add continuous logging in AWS Lambda

rdettai:
Currently, the lambda function is not logging the output of the executed command. We need this logs for longer running jobs where the CLI might timeout. Also, we would like these logs to be written to Logwatch continuously instead of the end of the command execution.

### :memo: Checklist

We are limited to specific "threading" libraries on AWS lambda. `multiprocessing` for instance usually doesn't work with the pool and queue interfaces. This is why we use the basic `threading` module with a small adapter to get the return values from the threaded function.

Also modified `logging.basicConfig(level=logging.INFO)` because it seems that it is not working properly on Lambda because of some pre-initialization of the logger in the Lambda python runtime.

### DIFF
--- a/cloud/aws/lambda-handler.py
+++ b/cloud/aws/lambda-handler.py
@@ -61,7 +61,7 @@ def handler(event, context):
         stderr = stderr_thread.join()
 
         # multiplex stdout and stderr into the result field
-        res = stdout if stdout != b"" else stderr
+        res = stdout if stdout != "" else stderr
         return {"result": res, "parsed_cmd": parsed_cmd}
 
     except Exception as err:
--- a/cloud/aws/lambda-handler.py
+++ b/cloud/aws/lambda-handler.py
@@ -12,7 +12,7 @@ class ReturningThread(Thread):
     return value when calling join()"""
 
     def __init__(self, target=None, args=()):
-        Thread.__init__(self, target, args)
+        Thread.__init__(self, target=target, args=args)
         self._return = None
 
     def run(self):
--- a/cloud/aws/lambda-handler.py
+++ b/cloud/aws/lambda-handler.py
@@ -2,16 +2,49 @@ import os
 import subprocess
 import logging
 import base64
+from threading import Thread
 
-logging.basicConfig(level=logging.INFO)
+logging.getLogger().setLevel(logging.INFO)
+
+
+class ReturningThread(Thread):
+    """A wrapper around the Thread class to actually return the threaded function
+    return value when calling join()"""
+
+    def __init__(
+        self, group=None, target=None, name=None, args=(), kwargs={}, Verbose=None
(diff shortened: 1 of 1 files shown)


cloud/aws/lambda-handler.py
```diff
@@ -2,27 +2,67 @@
...
-logging.basicConfig(level=logging.INFO)
+logging.getLogger().setLevel(logging.INFO)
+
+
+class ReturningThread(Thread):
+    """A wrapper around the Thread class to actually return the threaded function
+    return value when calling join()"""
+
+    def __init__(
+        self, group=None, target=None, name=None, args=(), kwargs={}, Verbose=None
```
dispanser, line 15: I'm probably just blind, but I can't spot where the `Verbose` argument is being used

cloud/aws/lambda-handler.py
```diff
@@ -2,27 +2,67 @@
...
+        stderr_thread = ReturningThread(
+            target=buff_and_print, args=(process.stderr, "stderr")
+        )
+        stderr_thread.start()
+        stdout = buff_and_print(process.stdout, "stdout")
+        stderr = stderr_thread.join()
 
-        return {"result": res.decode("utf-8"), "parsed_cmd": parsed_cmd}
+        # multiplex stdout and stderr into the result field
+        res = stdout if stdout != "" else stderr
```
dispanser, line 64: would it be desirable to have both stdout _and_ stderr in the output?

dispanser (approved):
I haven't actually tested tha the logs appear continuously, but the change looks good.

I added two comments you may consider, but I'm not sure.

cloud/aws/lambda-handler.py
rdettai: Yes, but I don't really know how to format it. One logical path would be to return both stdout and stderr in the result object. But then you need a bit more formatting on the client side. Do you have a suggestion?

cloud/aws/lambda-handler.py
rdettai: Copy pasted from Stackoverflow 😄. I am not very familiar with the python inheritance syntax and I assumed it was an override of the parent constructor, but I can see it's not!

cloud/aws/lambda-handler.py
dispanser: I'm not even sure if there's a sensible combination of stderr and stdout. Let's handle that "later" when we see the need.

rdettai:
@dominiklohmann not sure why the "Check Labels" workflow is hanging :-\

dominiklohmann:
> @dominiklohmann not sure why the "Check Labels" workflow is hanging :-\

The label on this PR pre-dated this new check. I just removed it and added it back.

PR #2145, opened 2022-03-11
Show the CMake build type in `vast version`

dominiklohmann:
A really quick fix that should make it easier to spot the obvious mistake of running the wrong version when talking about performance.

### :memo: Checklist

Don't think this needs a changelog entry. Diff is rather small, so review any way you want.


tobim (approved):
This change works, but we could even extend it to embed the `feature_summary` from CMake. At least `VAST_ENABLE_ASAN` and `VAST_ENABLE_ASSERTIONS` is also important information when you want to know what kind of performance to expect from a build.

CMakeLists.txt
```diff
@@ -367,15 +367,16 @@ unset(_VAST_ENABLE_ASSERTIONS_DEFAULT)
...
+if (CMAKE_BUILD_TYPE IN_LIST _vast_build_types_with_default_asan)
   set(_VAST_ENABLE_ASAN_DEFAULT ON)
 else ()
   set(_VAST_ENABLE_ASAN_DEFAULT OFF)
 endif ()
 
 option(VAST_ENABLE_ASAN
        "Insert pointer and reference checks into the generated binaries"
-       _VAST_ENABLE_ASAN_DEFAULT)
+       "${_VAST_ENABLE_ASAN_DEFAULT}")
```
dominiklohmann, line 379: This bug essentially makes us wait on tenzir/actor-framework#6 before merging this, otherwise the local development experience suffers on macOS.

dominiklohmann:
> This change works, but we could even extend it to embed the `feature_summary` from CMake. At least `VAST_ENABLE_ASAN` and `VAST_ENABLE_ASSERTIONS` is also important information when you want to know what kind of performance to expect from a build.

I've implemented this in a separate commit. Noticed that the default for `VAST_ENABLE_ASAN` was broken for a year and nobody noticed, so I fixed that as well.

tobim:
> Noticed that the default for VAST_ENABLE_ASAN was broken for a year and nobody noticed, so I fixed that as well.

Interesting. I always set it explicitly so I didn't notice.

libvast/src/system/version_command.cpp
```diff
@@ -40,6 +40,7 @@ record combine(const record& lhs, const record& rhs) {
 record retrieve_versions() {
   record result;
   result["VAST"] = version::version;
+  result["VAST Build Type"] = version::build_type;
```
mavam, line 43: Should we add the sanitizer configuration in here as well? We could have a `sanitizers` key.

dominiklohmann: That's already in this PR: It renders as `"VAST Build Type": "Debug with Assertions, Address Sanitizer"`.

mavam: Uh, that's not nice to parse. We're already using JSON, so while we're changing this, why not make it structured?

dominiklohmann: The build type is not even supposed to be parsed; it's entirely customizable. There is no standardization around it—it's just an arbitrary string.

mavam: And that's okay, I think. But let's not creep in more than needed into this opaque value. Why not make it easy for users to check whether a given sanitizer is on? Here's an example of what I mean:

```json
"Build Settings": {
  "Assertions": true,
  "Address Sanitizier": true,
  "UB Sanitizier": false,
}
```

dominiklohmann: The main reason I did not do it this way is because that's (1) more changes to `config.hpp.in` and (2) easily 3x the amount of CMake I need to write. But I can make that change. :-)

PR #2146, opened 2022-03-11
Load stores lazily

dominiklohmann:
This causes local segment stores of passive partitions to be loaded lazily, i.e., initialized only after the actor received its first message.

### :memo: Checklist

I was unable to reproduce the behavior from the removed comment locally. Please make sure this also holds true for other platforms.

It'd be great to measure this—if the impact is big, then we probably want to add a changelog entry.


dominiklohmann:
I added a changelog entry. After measuring locally, this turned out to have a visible effect for a small test database with 100 partitions. Quantifying the speed up does not make sense here, as it is highly dependent on the disk speeds and the operating system's scheduling of file reads / mmaps.

lava (approved):
Nice, always good to see problems go away automatically over time :D

PR #2148, opened 2022-03-11
Add support for exponents in the real parser

tobim:
### :memo: Checklist

* Start with `test/parseable.cpp`, check that coverage is sufficient.
* Read the implementation, note that I didn't use combinators to stay in line with the style of this file.

### DIFF
--- a/changelog/unreleased/bug-fixes/2148--real-scientific-notation.md
+++ b/changelog/unreleased/bug-fixes/2148--real-scientific-notation.md
@@ -1,0 +1,1 @@
+The real parser now supports input in scientific notation.
--- a/libvast/test/parseable.cpp
+++ b/libvast/test/parseable.cpp
@@ -77,7 +77,9 @@ TEST(choice - unused LHS) {
 TEST(choice triple) {
   using namespace parsers;
   auto fired = false;
-  auto p = chr{'x'} | i32 | eps->*[&] { fired = true; };
+  auto p = chr{'x'} | i32 | eps->*[&] {
+    fired = true;
+  };
   caf::variant<char, int32_t> x;
   CHECK(skip_to_eoi(p)("foobar", x));
   CHECK(fired);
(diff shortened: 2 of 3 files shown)


libvast/test/parseable.cpp
```diff
@@ -595,6 +597,50 @@ TEST(real) {
...
+    CHECK_EQUAL(f, l);
+  }
+  {
+    MESSAGE("explicit positive exponent");
+    auto str = ".456789e+43"s;
+    real d = 0;
+    auto f = str.begin();
+    auto l = str.end();
+    CHECK(p(f, l, d));
+    CHECK_EQUAL(d, 4.56789e42);
```
mavam, line 629: Why does the string end in 43 and the C++ literal in 42? (Also for other tests)

tobim: The dot moved one position to the right.

changelog/unreleased/bug-fixes/2148--real-scientific-notation.md
```diff
@@ -0,0 +1 @@
+The real parser now supports input in scientific notation.
```
mavam, line 1: Please provide an example for users.

mavam (approved):
This looks functional, nice. Agreeing to not make this combinator-style.

I'm approving modulo two changes:
1. An improved changelog entry.
2. Additional tests.

Regarding (2): could you add unit test for invalid scientific notation? For example, one where input ends in `e`. And one where the exponent is out of bounds. Also, I see only unit tests that start with `.NNN`. While I understand that other tests may cover the things in front of the dot, I do think we should have at least one "end-to-end" test, e.g., `4.56789e-323`.

tobim:
I added a test that generates a few reals and discovered that the implementation is very much incorrect.
I'm playing with the idea replacing that parser with https://github.com/fastfloat/fast_float.

mavam:
It looks like fast_float is already part of Arrow: https://github.com/apache/arrow/pull/8494.

You may be able to just call [this](https://github.com/apache/arrow/pull/8494/files#diff-4b09cf77402d40aa70602d9984998af2ebc2c29e91d9baab0a6e8b1dcd751a41R29):
```c++
  const auto res = ::arrow_vendored::fast_float::from_chars(s, s + length, *out);
```

tobim:
fast_float is not a public dependency in arrow so it isn't part of the installation. The only function that uses it is
```
bool StringToFloat(const char* s, size_t length, char decimal_point, double* out) {
  ::arrow_vendored::fast_float::parse_options options{
      ::arrow_vendored::fast_float::chars_format::general, decimal_point};
  const auto res =
      ::arrow_vendored::fast_float::from_chars_advanced(s, s + length, *out, options);
  return res.ec == std::errc() && res.ptr == s + length;
}
```
The check for the exact length match at the end makes it unusable for us.

And besides, we should probably integrate it properly anyways.

mavam:
> fast_float is not a public dependency in arrow so it isn't part of the installation.

Oh well...

Next try: what about simdjson? Can't we call into some of their real parsers? Or is the API also unusable from the outside?

tobim:
Closing because of newly identified bugs in the implementation. I will open a new PR that switches the internals to fast_float soon.

PR #2149, opened 2022-03-13
Expose nested type aliases

dominiklohmann:
The new function `type::aliases` returns all aliases of a type, from outermost to innermost. Essentially, this iterates over the internal metadata structure of a type and yields a slice of FlatBuffers table for every named metadata layer.

This is useful when wanting to write code that accepts both a type and any of its aliases, e.g., as a key of a map for a custom index configuration, or when tracking occurences of types in the catalog.

### :memo: Checklist

Depends on #2130.

The provided unit test should explain the functionality easily. Review that first, then the implementation. @mavam, please verify that this would work for your needs in #2058.

### DIFF
--- a/libvast/src/type.cpp
+++ b/libvast/src/type.cpp
@@ -1113,6 +1113,39 @@ detail::generator<type::attribute_view> type::attributes() const& noexcept {
   __builtin_unreachable();
 }
 
+detail::generator<type> type::aliases() const noexcept {
+  const auto* root = &table(transparent::no);
+  while (true) {
+    switch (root->type_type()) {
+      case fbs::type::Type::NONE:
+      case fbs::type::Type::bool_type:
+      case fbs::type::Type::integer_type:
+      case fbs::type::Type::count_type:
+      case fbs::type::Type::real_type:
+      case fbs::type::Type::duration_type:
+      case fbs::type::Type::time_type:
+      case fbs::type::Type::string_type:
+      case fbs::type::Type::pattern_type:
+      case fbs::type::Type::address_type:
+      case fbs::type::Type::subnet_type:
+      case fbs::type::Type::enumeration_type:
+      case fbs::type::Type::list_type:
+      case fbs::type::Type::map_type:
+      case fbs::type::Type::record_type:
+        co_return;
+      case fbs::type::Type::enriched_type: {
+        const auto* enriched_type = root->type_as_enriched_type();
+        if (enriched_type->name())
+          co_yield type{table_->slice(as_bytes(*enriched_type->type()))};
+        root = enriched_type->type_nested_root();
+        VAST_ASSERT(root);
+        break;
+      }
+    }
+  }
+  __builtin_unreachable();
+}
+
 bool is_container(const type& type) noexcept {
   const auto& root = type.table(type::transparent::yes);
   switch (root.type_type()) {
--- a/libvast/test/type.cpp
+++ b/libvast/test/type.cpp
@@ -721,6 +721,25 @@ TEST(enriched types) {
   CHECK_EQUAL(lat, at);
 }
 
+TEST(aliases) {
+  const auto t1 = bool_type{};
+  const auto t2 = type{"quux", t1};
+  const auto t3 = type{"qux", t2, {{"first"}}};
+  const auto t4 = type{"baz", t3};
+  const auto t5 = type{t4, {{"second"}}};
+  const auto t6 = type{"bar", t5, {{"third"}}};
+  const auto t7 = type{"foo", t6, {{"fourth"}}};
+  auto aliases = std::vector<type>{};
+  for (auto&& alias : t7.aliases())
+    aliases.push_back(std::move(alias));
+  REQUIRE_EQUAL(aliases.size(), 5u);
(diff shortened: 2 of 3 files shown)


libvast/src/type.cpp
```diff
@@ -1113,6 +1113,39 @@ detail::generator<type::attribute_view> type::attributes() const& noexcept {
...
+      case fbs::type::Type::address_type:
+      case fbs::type::Type::subnet_type:
+      case fbs::type::Type::enumeration_type:
+      case fbs::type::Type::list_type:
+      case fbs::type::Type::map_type:
+      case fbs::type::Type::record_type:
+        co_return;
+      case fbs::type::Type::enriched_type: {
+        const auto* enriched_type = root->type_as_enriched_type();
+        if (enriched_type->name())
```
mavam, line 1138: Is this check an internal "hack" to differentiate against anonymous types with attributes?

libvast/test/type.cpp
```diff
@@ -721,6 +721,25 @@ TEST(enriched types) {
...
+  const auto t4 = type{"baz", t3};
+  const auto t5 = type{t4, {{"second"}}};
+  const auto t6 = type{"bar", t5, {{"third"}}};
+  const auto t7 = type{"foo", t6, {{"fourth"}}};
+  auto aliases = std::vector<type>{};
+  for (auto&& alias : t7.aliases())
+    aliases.push_back(std::move(alias));
+  REQUIRE_EQUAL(aliases.size(), 5u);
+  CHECK_EQUAL(aliases[0], t6);
+  CHECK_EQUAL(aliases[1], t4);
```
mavam, line 737: Is `t5` skipped on purpose?

mavam (commented):
Thanks for this. Functionality-wise, this suits the bill. Interface-wise, I need the ability to iterate from the root type to the most specific leaf type, i.e., the generator should return `t1` first and `t6` last. Maybe this could be a template parameter the algorithm with an `order` enum tag with `forward` and `backward` values, allowing users to write `t.aliases<order::backward>()` as well as t.aliases<order::forward>()`.

Is it feasible to implement this efficiently? The generator would have to "pierce through" all intermediate types with some effort first, potentially allocate to keep breadcrum-like state somewhere to roll backwards from there. That said, it could be a stack vector or something as tree depth is not large in practice.

dominiklohmann:
> Is it feasible to implement this efficiently?

Just store the result in a vector and iterate over it backwards. I don't see any other way.

I question why you need the order the other way around, though. Wouldn't it make more sense to go from most specific to least specific? Can you describe what you want to achieve first?

libvast/test/type.cpp
dominiklohmann: Yes. The constructor of `type` folds it into the next named layer on construction (see #2138). This is important because it is not something a user can actually express in configuration.

libvast/src/type.cpp
dominiklohmann: `name()` here does not return the string view—instead, it returns a `flatbuffers::Indirect<T>*`, which holds a `nullptr` if the field is unset. So this is not a hack, but rather the proper way to check whether the field is set.

mavam:
> Just store the result in a vector and iterate over it backwards. I don't see any other way.

That's what I'm currently doing.

> I question why you need the order the other way around, though. Wouldn't it make more sense to go from most specific to least specific? Can you describe what you want to achieve first?

Let `t` be a `port` as an alias for `count`. Given `t`, I want to know "is there a sketch for `count`? If not, is there a sketch for `port`? And so on..." This question sequence happens when getting a new table slice at a partition builder. We bail out in this chain when the first type fails.

dominiklohmann:
> Let `t` be a `port` as an alias for `count`. Given `t`, I want to know "is there a sketch for `count`? If not, is there a sketch for `port`? And so on...".

Shouldn't that be the other way around? You iterate until you find the first suitable map entry.

```go
type bar = foo
type foo = port
type port = count
```

Assuming this type `bar`, you'd first check for `bar` directly, and then go over the aliases in order: `foo`, `port`, `count`.

You don't have a sketch for `bar`, you don't have one for `foo`, but you have one for `bar`. At this point you can stop iterating because you've seen that one before—and you know internally this implies the sketches for `count` as well.

Always open to pairing on this. But regardless of which way to iterate, if you want to do it the other way around you need to collect the yielded types in a vector. I don't see how we could iterate over a nested type structure backwards efficiently, considering it's a flat forward linked list internally.

mavam:
> Assuming this type `bar`, you'd first check for `bar` directly, and then go over the aliases in order: `foo`, `port`, `count`.

No, in this case not. Just because there's no sketch configuration for `bar` doesn't allow you to bail out yet. A `bar` is still a `count` and that ultimately matters.

That said, I think this works still because I can simply skip non-existent intermediate types, at least for my current use case. So I'll rebase onto this branch and will let you know how it goes.

libvast/test/type.cpp
mavam: Okay, not easy to see from this context, but fine with me.

mavam (approved):
This functionality does the trick, thanks!

PR #2152, opened 2022-03-14
Install example config files to docdir

dominiklohmann:
This changes plugin example configuration files to install to docdir rather than sharedir, which makes a lot more sense.

```
❯ tree path/to/install/share/doc
path/to/install/share/doc
└── vast
    ├── CHANGELOG.md
    ├── LICENSE
    ├── README.md
    ├── VAST.spdx
    ├── VERSIONING.md
    ├── examples
    │   ├── plugin
    │   │   └── example-analyzer
    │   │       └── example-analyzer.yaml
    │   └── vast.yaml
    └── plugin
        ├── aggregate
        │   ├── CHANGELOG.md
        │   └── README.md
        └── pcap
            ├── CHANGELOG.md
            └── README.md
```

### :memo: Checklist

I don't think this needs a changelog entry. Review file-by-file. I think the changes should be pretty much self-explanatory.

### DIFF
--- a/.cmake-format
+++ b/.cmake-format
@@ -41,7 +41,7 @@ with section("parse"):
         "vastinstallexampleconfiguration": {
             "spelling": "VASTInstallExampleConfiguration",
             "pargs": {
-                "nargs": 3,
+                "nargs": 4,
             },
         },
         "vastsystemizetarget": {
--- a/cmake/VASTRegisterPlugin.cmake
+++ b/cmake/VASTRegisterPlugin.cmake
@@ -40,6 +40,15 @@ macro (VASTNormalizeInstallDirs)
       )
     endif ()
   endforeach ()
+  # For the docdir especially, lowercase the project name.
+  if ("${CMAKE_INSTALL_DOCDIR}" STREQUAL
+      "${CMAKE_INSTALL_DATAROOTDIR}/doc/${PROJECT_NAME}")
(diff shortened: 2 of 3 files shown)


tobim:
Can we modify the _default_ docdir to `path/to/install/share/doc/vast`? The all-caps sticks out compared to the rest of the installed directory layout.

dominiklohmann:
> Can we modify the _default_ docdir to `path/to/install/share/doc/vast`? The all-caps sticks out compared to the rest of the installed directory layout.

I don't think we should. It defaults to the project name, which is `VAST` rather than `vast`. I'd rather stick with the recommendation here.

tobim:
> > Can we modify the _default_ docdir to `path/to/install/share/doc/vast`? The all-caps sticks out compared to the rest of the installed directory layout.
> 
> I don't think we should. It defaults to the project name, which is `VAST` rather than `vast`. I'd rather stick with the recommendation here.

Yes we should, an 3rd party tooling default is a poor justification for such an inconsistency.

dominiklohmann:
> Yes we should, an 3rd party tooling default is a poor justification for such an inconsistency.

I don't agree, but I also don't care as strongly about it as you do. Looking at my docdir, there seems to be a good mix of the two. I'll push the change in a bit.

tobim (approved):
Thanks!

PR #2154, opened 2022-03-14
Fix CMD execution from lambda and fargate

rdettai:
There were two bugs that are being fixed here:
- execute-command was not robust, some `CMD` arguments with pipes where failing
- run-lambda had a 60s timeout that was re-triggering the lambda client unexpectedly

### :memo: Checklist

The encoding to avoid escaping issues for execute-command is quite convoluted. Open to simplification suggestions.

### DIFF
--- a/cloud/aws/.gitignore
+++ b/cloud/aws/.gitignore
@@ -1,3 +1,3 @@
 .terraform
 default.env
-lambda-cmd.tmp
+cmd.tmp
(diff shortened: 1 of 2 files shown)


dispanser (approved):
Looks good.

I verified that I can now do `make execute-command CMD=...` now. We should probably add this to the repo, though.

PR #2156, opened 2022-03-15
Add a 'collect()' utility function

lava:
Mostly reduces the amount of typing required, but occasionally crucial for print statements where depending on the context it can be non-obvious to figure out the correct element type of the vector.

### :memo: Checklist

### DIFF
--- a/libvast/include/vast/detail/collect.hpp
+++ b/libvast/include/vast/detail/collect.hpp
@@ -24,8 +24,8 @@ Container collect(vast::detail::generator<T> g, size_t size_hint = 0) {
   Container result = {};
   if (size_hint)
     result.reserve(size_hint);
-  for (auto& x : g)
-    result.emplace(result.end(), x);
+  for (auto&& x : g)
+    result.emplace(result.end(), std::move(x));
   return result;
 }
 
--- a/libvast/include/vast/detail/generator.hpp
+++ b/libvast/include/vast/detail/generator.hpp
@@ -64,7 +64,7 @@ template <typename T>
 class generator_promise {
 public:
   using value_type = std::remove_reference_t<T>;
-  using reference_type = std::conditional_t<std::is_reference_v<T>, T, T&>;
+  using reference_type = T&&;
   using pointer_type = value_type*;
 
   generator_promise() = default;
@@ -78,12 +78,19 @@ public:
     return {};
   }
 
-  template <typename U = T,
-            std::enable_if_t<!std::is_rvalue_reference<U>::value, int> = 0>
-  stdcoro::suspend_always
-  yield_value(std::remove_reference_t<T>& value) noexcept {
-    m_value = std::addressof(value);
-    return {};
+  // This overload copies v to a temporary, then yields a pointer to that.
+  // This allows passing an lvalue to co_yield for a generator<NotReference>.
+  // Looks crazy, but taken from the reference implementation in P2529R0.
+  auto yield_value(const T& v) requires(
+    !std::is_reference_v<T> && std::copy_constructible<T>) {
+    struct Owner : stdcoro::suspend_always {
+      Owner(const T& val, pointer_type& out) : v(val) {
+        out = &v;
+      }
+      Owner(Owner&&) = delete;
+      T v;
+    };
+    return Owner(v, m_value);
   }
 
   stdcoro::suspend_always
@@ -177,10 +184,6 @@ public:
 
   reference operator*() const noexcept {
     return m_coroutine.promise().value();
-  }
-
-  pointer operator->() const noexcept {
-    return std::addressof(operator*());
   }
 
 private:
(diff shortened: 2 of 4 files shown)


libvast/include/vast/detail/collect.hpp
```diff
@@ -0,0 +1,27 @@
...
+#include <vast/detail/generator.hpp>
+
+namespace vast::detail {
+
+/// A utility function to collect all results produced by a
+/// `vast::detail::generator<T>` into a suitable standard container.
+/// Example:
+///     auto g = vast::detail::generator<int>{};
+///     auto v = vast::detail::collect<std::vector>(g);
+template <template <class, class> class Container, class Generated>
```
dominiklohmann, line 18: Why the limitation to two arguments, and why do you hardcode the allocator below instead of defaulting it to the template argument?

E.g., this could also be:

```
template <template <class...> class Container, class T, class Allocator = std::allocator<T>>
```

libvast/include/vast/detail/collect.hpp
```diff
@@ -0,0 +1,27 @@
...
+Container<Generated, std::allocator<Generated>>
+collect(vast::detail::generator<Generated> g) {
+  Container<Generated, std::allocator<Generated>> result;
+  for (auto& x : g)
+    result.push_back(x);
```
dominiklohmann, line 23: The more portable way to implement this such that it works for other containers as well is to use a back inserter, e.g.,

```
std::move(g.begin(), g.end(), std::back_inserter(result));
```

libvast/include/vast/detail/collect.hpp
```diff
@@ -0,0 +1,27 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2021 The VAST Contributors
```
dominiklohmann, line 6:
```suggestion
// SPDX-FileCopyrightText: (c) 2022 The VAST Contributors
```

libvast/include/vast/detail/collect.hpp
lava: Is it safe to move out of `g.begin()`? It returns a reference, so my hunch was probably not.

libvast/include/vast/detail/collect.hpp
```diff
@@ -0,0 +1,27 @@
...
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <vast/detail/generator.hpp>
+
+namespace vast::detail {
+
+/// A utility function to collect all results produced by a
+/// `vast::detail::generator<T>` into a suitable standard container.
+/// Example:
```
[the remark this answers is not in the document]
lava: As far as I can tell `std::ranges::to` did not make it into the standard:

```
/home/benno/src/vast/reviews/libvast/include/vast/detail/collect.hpp:31:23: error: ‘to’ is not a member of ‘std::ranges’
   31 |   return std::ranges::to<Container>(g);
```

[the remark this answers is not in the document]
dominiklohmann: `reserve` for random access iterators, and also more flexible allocator support. Does your proposed function work with `stack_vector`?

libvast/include/vast/detail/collect.hpp
dominiklohmann: It is safe. This being a reference is an artifact of our coroutine implementation being based on an early draft of the proposal for C++23. There's also sc-32185 to make the behavior a bit more intuitive.

libvast/include/vast/detail/collect.hpp
lava: That has the drawback that you have to write out the element type `T` if you want to change the allocator.

I've changed it now to just accept a container type, and a convenience overload for the common case of `std::vector<T>`.

libvast/include/vast/detail/collect.hpp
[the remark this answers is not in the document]
lava: I think the `insert`-based version should do both of these.

libvast/include/vast/detail/collect.hpp
lava: I've appended an implementation of `sc-32185` in a separate commit, to make the ownership semantics explicit. (the story is technically only a discussion story, but I guess we can discuss as part of the review in this PR; I expect the general thrust to be non-controversial)

libvast/include/vast/detail/collect.hpp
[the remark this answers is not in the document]
dominiklohmann: That's fair, yes—I meant with the reserve that you could do this for `to<...>` more generically, if it wasn't just for generators. Anyways, marking this as resolved.

libvast/include/vast/detail/collect.hpp
dominiklohmann: Thank you so much! I think we can archive that story then?

Also wrt sentinel support: I think that's just missing from the standard library implementation, the relaxed iterator specification was in C++20 already iirc.

Marking as resolved.

dominiklohmann (approved):
Thanks for the generator changes! 🚀

dominiklohmann:
[sc-32185]

PR #2158, opened 2022-03-18
Add CI job for terraform

rdettai:
We need a way to ensure that the Terraform scripts to deploy VAST on AWS keep working throughout the product lifecycle. This PR adds a Github Action workflow to test some basic commands of the deployment Makefile.

Note that:
- the CI run uses Terraform Cloud as a backend. This is safer than having the state locally in the Github action, because it might be hard to recover it if the CI job crashes.
- The run takes between 30 minutes and 1 hour to complete, mostly because tearing down some of the network resources is very slow on the AWS side.
- The CI run uses paying AWS resources. It tears down all the resources at the end of the job, so the costs of running it should be minimal (somewhere between $0.1 and $0.5). 
- It is worth noting that a failing job run might leave the resources running, which might incur higher costs. For this reason, this job is meant to be triggered manually and its outcome should be monitored.

### :memo: Checklist

- the required variables are configured through secrets, even though most of them are really just configuration variables. I am wondering if there is a better way to do this.
- the configuration instructions are in the Action configuration yaml file itself
- i used a random file `https://raw.githubusercontent.com/tenzir/vast/master/vast/integration/data/suricata/eve.json` for the ingestion test. We might want to use another one hosted in a more relevant place (Tenzir owned S3 bucket?)

### DIFF
--- a/.github/workflows/terraform.yaml
+++ b/.github/workflows/terraform.yaml
@@ -2,14 +2,14 @@ name: "Terraform"
 on: workflow_dispatch
 
 # This CI configuration has multiple dependencies:
-# - You need an AWS account with a user having sufficient IAM access to run 
-#   the Terraform scripts. Its crendientials should be set in the repository 
+# - You need an AWS account with a user having sufficient IAM access to run
+#   the Terraform scripts. Its crendientials should be set in the repository
 #   secrets as AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY.
 # - You need to choose an AWS region and a VPC within that region. These
 #   should be configured in the repository secrets as AWS_REGION and
 #   PEERED_VPC_ID.
 # - The VAST deployment creates a new VPC. Its CIDR should not overlap an
-#   existing one in your account. Configure it in the repository secrets as 
+#   existing one in your account. Configure it in the repository secrets as
 #   VAST_CIDR
 # - The CI run uses Terraform Cloud to store the state. Once you have an
 #   account setup, configure its organization name and the desired workspace
@@ -40,7 +40,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      
+
       - name: "Substitute backend with Terraform Cloud"
         run: |
           cat<<EOF > backend.tf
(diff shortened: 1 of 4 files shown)


rdettai:
I have the job setup in rdettai/vast and it is running as expected (CI equivalent of saying: works on my machine 😁). For instance, [this run](https://github.com/rdettai/vast/runs/5599588166) caught that `make execute-command` had some issue before https://github.com/tenzir/vast/pull/2154. We can expect it to pass once that PR is merged.

rdettai:
I think https://github.com/rdettai/vast/blob/198e8d34f95f742783bdf92f92eaa7648a192f8a/.github/workflows/vast.yaml#L819-L823 is not really required anymore.

rdettai:
The CI is now passing after #2154 was merged (see [this run](https://github.com/rdettai/vast/runs/5624868041?check_suite_focus=true)).

I encountered some issues though when I interrupted one of the runs. The interruption of the Terraform task was too brutal which lead to a widely inconsistent state and a lot of cleanup work. I think the fix is to switch back the run mode to Remote and not Local on Terraform Cloud
image

This way the interruption will be managed by Terraform Cloud and should be cleaner. 

To allow this I need to remove the push of the lambda image from the Terraform local-exec resource and do it from the Makefile instead.

rdettai:
After quite some effort, I did not manage to find a simple way to push the docker image for VAST on lambda into ECR:
- I tried to have Lambda point to a placeholder image when deployed by Terraform, then update the code with [`aws lambda update-funciton-conde`](https://docs.aws.amazon.com/cli/latest/reference/lambda/update-function-code.html), but lambda requires the image to be already present in ECR when being created, so the placeholder image itself would need to be uploaded (🐔 -> 🥚 or 🥚 -> 🐔 ?)
- I have looked to all sorts of way to copy images from repositories to others, ideally embedable into a simple lambda or task triggered when deploying the stack. For instance I looked at docker alternatives (skopeo, podman), but none was very inspiring 😄 

After having explored all this, I am now convinced the best approach is to have a multi-step deployment. A first Terraform stack deploys ECR. We add the image to the repository through make and then start a second stack deployment. This also has the benefit that we could tier the resources according to their lifecycle (this is valuable for long term storage resources like EFS and Lambda). This also introduces a bit of complexity as we have to manage dependencies between the stacks ourselves.

rdettai:
Closing this in favor of https://github.com/tenzir/vast/pull/2168

PR #2160, opened 2022-03-22
Backport bug fixes for a v1.1.1 release

dominiklohmann:
In large-scale deployments we observed the disk monitor erase request to return a timeout error, cancelling the ongoing request. This is a two-fold bug fix: First, we must not use a timeout for process-internal actor communication, especially not for such complicated nested loops that when partially executed and then never resumed leave the actor in an undefined state. Second, we must treat erase requests with a high priority because they should never be upheld by queued requests on the read or write path.

If this does not fix the bug, then a release with these changes will at the very least help us track down the actual source of the issue because it's no longer being shadowed from the request timeout error.

### :memo: Checklist

Run on our testbed.


lava (approved):
Changes look good; there are not unit tests but we verified manually that they work and improve disk monitor behavior on the testbed.

PR #2162, opened 2022-03-24
Support environment variables as alternate config mechanism

mavam:
This PR adds new functionality to use environment variables as alternate input to a YAML config file.

### :memo: Checklist

- [x] Add utility function to enumerate environment
- [x] Translate `VAST_` and `CAF_` env vars into settings
- [x] Add changelog entry
- [x] Add tests
- [x] Update docs

Commit-by-commit.

### DIFF
--- a/changelog/unreleased/features/2162--environment-variables-as-config.md
+++ b/changelog/unreleased/features/2162--environment-variables-as-config.md
@@ -1,6 +1,7 @@
 VAST has now complete support for passing environment variables as alternate
-path to configuration files. Environment variables have lower precedence than
-CLI arguments and higher precedence than config files. Non-empty variables
-of the form `VAST_FOO_BAR__BAZ` map to `vast.foo.bar_baz`, i.e.,
-double-escaping `_` yields a literal `_` in the config key. VAST only considers
-variables with the prefix `VAST_` and `CAF_`.
+path to configuration files. Environment variables have *lower* precedence than
+CLI arguments and *higher* precedence than config files. Variable names of the
+form `VAST_FOO__BAR_BAZ` map to `vast.foo.bar-baz`, i.e., `__` is a record
+separator and `_` translates to `-`. This does not apply to the prefix `VAST_`,
+which is considered the application identifier. Only variables with non-empty
+values are considered.
--- a/changelog/unreleased/features/2162--environment-variables-as-config.md
+++ b/changelog/unreleased/features/2162--environment-variables-as-config.md
@@ -1,0 +1,6 @@
+VAST has now complete support for passing environment variables as alternate
+path to configuration files. Environment variables have lower precedence than
+CLI arguments and higher precedence than config files. Non-empty variables
+of the form `VAST_FOO_BAR__BAZ` map to `vast.foo.bar_baz`, i.e.,
+double-escaping `_` yields a literal `_` in the config key. VAST only considers
+variables with the prefix `VAST_` and `CAF_`.
--- a/libvast/src/detail/env.cpp
+++ b/libvast/src/detail/env.cpp
@@ -29,18 +29,28 @@ auto env_mutex = std::mutex{};
 
 } // namespace
 
-std::optional<std::string> locked_getenv(const char* var) {
+std::optional<std::string_view> getenv(std::string_view var) {
   auto lock = std::scoped_lock{env_mutex};
   // NOLINTNEXTLINE(concurrency-mt-unsafe)
-  if (const char* result = ::getenv(var))
-    return result;
+  if (const char* result = ::getenv(var.data()))
+    return std::string_view{result};
   return {};
 }
 
-caf::expected<void> locked_unsetenv(const char* var) {
+caf::error setenv(std::string_view key, std::string_view value, int overwrite) {
   auto lock = std::scoped_lock{env_mutex};
   // NOLINTNEXTLINE(concurrency-mt-unsafe)
-  if (::unsetenv(var) == 0)
+  if (::setenv(key.data(), value.data(), overwrite) == 0)
+    return {};
+  return caf::make_error( //
+    ec::system_error,
+    fmt::format("failed in unsetenv(3): {}", ::strerror(errno)));
+}
+
+caf::error unsetenv(std::string_view var) {
+  auto lock = std::scoped_lock{env_mutex};
+  // NOLINTNEXTLINE(concurrency-mt-unsafe)
+  if (::unsetenv(var.data()) == 0)
     return {};
   return caf::make_error( //
     ec::system_error,
--- a/libvast/src/detail/env.cpp
+++ b/libvast/src/detail/env.cpp
@@ -8,7 +8,6 @@
 
 #include "vast/detail/env.hpp"
 
-#include "vast/detail/assert.hpp"
 #include "vast/error.hpp"
 
 #include <fmt/format.h>
@@ -62,10 +61,11 @@ generator<std::pair<std::string_view, std::string_view>> environment() {
   for (auto env = environ; *env != nullptr; ++env) {
     auto str = std::string_view{*env};
     auto i = str.find('=');
-    VAST_ASSERT(i != std::string::npos);
-    auto key = str.substr(0, i);
-    auto value = str.substr(i + 1);
-    co_yield std::pair{key, value};
+    if (i != std::string::npos) {
+      auto key = str.substr(0, i);
+      auto value = str.substr(i + 1);
+      co_yield std::pair{key, value};
+    }
   }
 }
 
--- a/libvast/src/detail/env.cpp
+++ b/libvast/src/detail/env.cpp
@@ -8,12 +8,16 @@
 
 #include "vast/detail/env.hpp"
 
+#include "vast/detail/assert.hpp"
 #include "vast/error.hpp"
 
 #include <fmt/format.h>
 
+#include <cstdio>
 #include <cstdlib>
 #include <mutex>
+
+extern char** environ;
 
 namespace vast::detail {
 
@@ -41,6 +45,18 @@ caf::expected<void> locked_unsetenv(const char* var) {
   return caf::make_error( //
     ec::system_error,
     fmt::format("failed in unsetenv(3): {}", ::strerror(errno)));
+}
+
+generator<std::pair<std::string_view, std::string_view>> environment() {
+  // Envrionment variables come as "key=value" pair strings.
+  for (auto env = environ; *env != nullptr; ++env) {
+    auto str = std::string_view{*env};
+    auto i = str.find('=');
+    VAST_ASSERT(i != std::string::npos);
+    auto key = str.substr(0, i);
+    auto value = str.substr(i + 1);
+    co_yield std::pair{key, value};
+  }
 }
 
 } // namespace vast::detail
--- a/libvast/src/plugin.cpp
+++ b/libvast/src/plugin.cpp
@@ -216,7 +216,7 @@ caf::error initialize(caf::actor_system_config& cfg) {
     }
     // Second, try to read the configuration from the plugin-specific
     // configuration files at <config-dir>/plugin/<plugin-name>.yaml.
-    for (auto&& config_dir : system::config_dirs(cfg)) {
+    for (auto&& config_dir : system::config_dirs()) {
       const auto yaml_path
         = config_dir / "plugin" / fmt::format("{}.yaml", plugin->name());
       const auto yml_path
--- a/libvast/src/plugin.cpp
+++ b/libvast/src/plugin.cpp
@@ -47,7 +47,7 @@ get_plugin_dirs(const caf::actor_system_config& cfg) {
         &cfg, "vast.plugin-dirs"))
     result.insert(dirs->begin(), dirs->end());
   if (!bare_mode)
-    if (auto home = detail::locked_getenv("HOME"))
+    if (auto home = detail::getenv("HOME"))
       result.insert(std::filesystem::path{*home} / ".local" / "lib" / "vast"
                     / "plugins");
   result.insert(detail::install_plugindir());
--- a/libvast/src/plugin.cpp
+++ b/libvast/src/plugin.cpp
@@ -216,7 +216,7 @@ caf::error initialize(caf::actor_system_config& cfg) {
     }
     // Second, try to read the configuration from the plugin-specific
     // configuration files at <config-dir>/plugin/<plugin-name>.yaml.
-    for (auto&& config_dir : system::config_dirs()) {
+    for (auto&& config_dir : system::config_dirs(cfg)) {
       const auto yaml_path
         = config_dir / "plugin" / fmt::format("{}.yaml", plugin->name());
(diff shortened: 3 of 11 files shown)


libvast/src/system/configuration.cpp
```diff
@@ -48,6 +48,17 @@ template <concrete_type T>
 struct has_extension_type
   : std::is_base_of<arrow::ExtensionType, typename T::arrow_type> {};
 
+/// Translates an environment variable to a config key.
+/// A '.' in the YAML config maps to '_' in the env key. A literal '_' in a
+/// key requires double-escaping, i.e., "__".
```
dominiklohmann, line 53: Can we re-use the mechanism from Dynaconf here, which I think is more widely used?

- `__` translates to a nesting layer: `.`
- `_` translates to a dash: `-`

Note that this does not apply to the application name prefix:

- `VAST_ENDPOINT` translates to `vast.endpoint`
- `VAST_IMPORT__BATCH_TIMEOUT` translates to `vast.import.batch-timeout`

libvast/src/system/configuration.cpp
```diff
@@ -144,18 +155,24 @@ caf::error configuration::parse(int argc, char** argv) {
...
   auto is_vast_opt = [](const auto& x) {
     return !x.starts_with("--caf.");
   };
   auto caf_opt = std::stable_partition(command_line.begin(), command_line.end(),
                                        is_vast_opt);
   std::vector<std::string> caf_args;
+  // Prepopulate CAF args with environment variables so that the CLI can
+  // override them.
+  for (const auto& [key, value] : detail::environment())
+    if (key.starts_with("CAF_") && !value.empty())
```
dominiklohmann, line 168: We should use `VAST_CAF__` as a prefix here to be consistent. VAST should not read environment variables that do not start with `VAST_`.

libvast/src/system/configuration.cpp
```diff
@@ -144,18 +155,24 @@ caf::error configuration::parse(int argc, char** argv) {
...
+      caf_args.emplace_back(
+        fmt::format("--{}={}", env_to_config_key(key), value));
```
dominiklohmann, line 170: How does this handle conflicts? E.g., what if I run VAST like this:

```bash
VAST_CAF__SCHEDULER__MAX_THROUGHPUT=10 vast --caf.scheduler.max-throughput=20 start
```

libvast/src/system/configuration.cpp
```diff
@@ -229,6 +246,11 @@ caf::error configuration::parse(int argc, char** argv) {
   }
   // Flatten everything for simplicity.
   merged_config = flatten(merged_config);
+  // Overwrite current config with VAST_ environment variables, potentially
+  // appending new config keys at the end.
+  for (const auto& [key, value] : detail::environment())
+    if (key.starts_with("VAST_") && !value.empty())
+      merged_config[env_to_config_key(key)] = std::string{value};
```
dominiklohmann, line 253: I don't think this handles nested keys correctly, does it?

libvast/src/system/configuration.cpp
```diff
@@ -229,6 +246,11 @@ caf::error configuration::parse(int argc, char** argv) {
   }
   // Flatten everything for simplicity.
   merged_config = flatten(merged_config);
+  // Overwrite current config with VAST_ environment variables, potentially
```
dominiklohmann, line 249: It should be noted somewhere that this cannot work for all options, because some are parsed _before_ we execute this piece of code.

libvast/src/system/configuration.cpp
dominiklohmann, line 253: Should the value be put through the `data` parser as an approximation here?

But even if you do this, I don't think the following will work:

```bash
# Works as expected, parses a list of strings:
vast start --commands='spawn source pcap'

# Does not work, parses as a single string
VAST_START__COMMANDS='spawn source pcap' vast start
```

dominiklohmann (commented):
Neat hack, but I think this needs a bit of work and most importantly tests.

libvast/src/system/configuration.cpp
mavam: `merged_config` is flattened with dots.

dominiklohmann: In that case it is irrelevant.

libvast/src/system/configuration.cpp
mavam: Agreed, this is the way the user would expect this to work.

libvast/src/system/configuration.cpp
mavam: Good point; addressed.

libvast/src/system/configuration.cpp
mavam: This adheres to the expectation: CLI arguments override environment variables.

libvast/src/system/configuration.cpp
mavam: I've fixed this up so that it applies to (almost) all options.

That said, I don't understand why setting `VAST_VERBOSITY` doesn't work. The logger initialization comes after the config parsing, what am I missing?

libvast/src/system/configuration.cpp
mavam: Not sure what you mean. Something like `VAST_EXPORT__TIMEOUT=1ns ./bin/vast -N export json 1.2.3.4` correctly parses the env variable as `duration`.

libvast/src/system/configuration.cpp
mavam: Note that setting `VAST_CONSOLE_VERBOSITY=debug` works as expected.

I found this in `logger.cpp`:
```c++
  // Allow `vast.verbosity` from the command-line to overwrite
  // the `vast.console-verbosity` setting from the config file.
  auto verbosity = caf::get_if<std::string>(&cfg_cmd, "vast.verbosity");
  if (verbosity) {
    if (loglevel_to_int(*verbosity, -1) < 0) {
      fmt::print(stderr,
                 "failed to start logger; vast.verbosity '{}' is invalid\n",
                 *verbosity);
      return false;
    }
    console_verbosity = *verbosity;
  }
```

Having seen that, I'm inclined to change the command-line pre-processing from

```c++
  const auto replacements = std::vector<std::pair<std::string, std::string>>{
    {"-qqq", "--verbosity=quiet"}, {"-qq", "--verbosity=error"},
    {"-q", "--verbosity=warning"}, {"-v", "--verbosity=verbose"},
    {"-vv", "--verbosity=debug"},  {"-vvv", "--verbosity=trace"},
  };
```
to
```c++
  const auto replacements = std::vector<std::pair<std::string, std::string>>{
    {"-qqq", "--console-verbosity=quiet"}, {"-qq", "--console-verbosity=error"},
    {"-q", "--console-verbosity=warning"}, {"-v", "--console-verbosity=verbose"},
    {"-vv", "--console-verbosity=debug"},  {"-vvv", "--console-verbosity=trace"},
  };
```

Objections?

mavam:
@tenzir/vast how would you test this feature?

libvast/src/system/configuration.cpp
dominiklohmann: But the env variable should be a string, not a duration, because we expect a string for that option.

mavam: Why does it work then?

dominiklohmann: Do we maybe parse as a data duration, and then when converting into a caf::settings we convert it back to a string, which we then convert back to a data string and then parse as a duration?

libvast/src/system/configuration.cpp
dominiklohmann: None from my side, although that is a breaking change that needs to be documented separately.

libvast/src/system/configuration.cpp
mavam: You're right, "transparent lists" don't work.

libvast/src/system/configuration.cpp
```diff
@@ -116,6 +130,8 @@ caf::error configuration::parse(int argc, char** argv) {
     for (const auto& [old, new_] : replacements)
       if (option == old)
         option = new_;
+  // FIXME(MV): why does this need to be here? The comment doesn't explain the
+  // need for this. Is there a hidden side-effect?
```
dominiklohmann, line 137: If you figure this out before merging I'm fine with this PR overall.

changelog/unreleased/features/2162--environment-variables-as-config.md
```diff
@@ -0,0 +1,7 @@
+VAST has now complete support for passing environment variables as alternate
+path to configuration files. Environment variables have *lower* precedence than
+CLI arguments and *higher* precedence than config files. Variable names of the
+form `VAST_FOO__BAR_BAZ` map to `vast.foo.bar-baz`, i.e., `__` is a record
+separator and `_` translates to `-`. This does not apply to the prefix `VAST_`,
+which is considered the application identifier. Only variables with non-empty
+values are considered.
```
dominiklohmann, line 7: Please create a corresponding docs PR before merging, and document that the list syntax differs from the list syntax on the command line.

libvast/src/system/configuration.cpp
```diff
@@ -148,8 +188,9 @@ caf::error configuration::parse(int argc, char** argv) {
   std::vector<std::string> caf_args;
   std::move(caf_opt, command_line.end(), std::back_inserter(caf_args));
   command_line.erase(caf_opt, command_line.end());
-  // Instead of the CAF-supplied `config_file_path`, we use our own
-  // `config_files` variable in order to support multiple configuration files.
+  // Collect config files. Instead of the CAF-supplied `config_file_path`, we
```
dominiklohmann, line 191: We still need to add a `VAST_CONFIG` variable to map onto the `vast --config=` option. This is handled manually below.

libvast/src/system/configuration.cpp
```diff
@@ -292,13 +341,6 @@ caf::error configuration::parse(int argc, char** argv) {
       put(content, key, value);
     }
   }
-  // If the user specifies a VAST_ENDPOINT, potentially use it. Precedence:
-  // 1. CLI argument
-  // 2. Environment variables
-  // 3. Config files
-  // 4. Defaults
-  if (auto vast_endpoint_env = detail::locked_getenv("VAST_ENDPOINT"))
-    caf::put(content, "vast.endpoint", *vast_endpoint_env);
```
dominiklohmann, line 301: Unrelated to this line, but I wanted to have a resolvable comment:

Can you double-check whether we still need to have the remaining `locked_getenv` calls in the codebase? Can we maybe remove that function altogether?

dominiklohmann (approved):
I left some final remarks inline. Other than that let's ship it!

libvast/src/system/configuration.cpp
mavam: Yep, I added a one-off...alas.

libvast/src/system/configuration.cpp
mavam: I clarified the comment.

libvast/src/plugin.cpp
```diff
@@ -216,7 +216,7 @@ caf::error initialize(caf::actor_system_config& cfg) {
     }
     // Second, try to read the configuration from the plugin-specific
     // configuration files at <config-dir>/plugin/<plugin-name>.yaml.
-    for (auto&& config_dir : system::config_dirs(cfg)) {
+    for (auto&& config_dir : system::config_dirs()) {
```
mavam, line 219: @dominiklohmann or @tobim: please note that I switched `config_dirs` into a singleton in this commit to begin breaking down the tight coupling into `actor_system` where possible.

mavam:
[sc-32484]

changelog/unreleased/features/2162--environment-variables-as-config.md
mavam: Turns out things work for lists. For example, `VAST_PLUGINS="sigma,pcap"` is equivalent to `--plugins=foo,bar`.

libvast/src/system/configuration.cpp
```diff
@@ -34,33 +34,184 @@
...
+    else
+      return caf::make_error(ec::invalid_configuration,
+                             fmt::format("cannot find configuration file {}",
+                                         config_file));
+  }
+  return result;
+}
+
+caf::expected<record>
+load_config_files(std::vector<std::filesystem::path> config_files) {
```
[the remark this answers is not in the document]
mavam: I'll refactor.

libvast/src/system/configuration.cpp
```diff
@@ -34,33 +34,184 @@
...
+/// @pre `!prefix.empty()`
+std::optional<std::string>
+env_to_config(std::string_view key, std::string_view prefix = "VAST") {
+  VAST_ASSERT(!prefix.empty());
+  // PREFIX_X is the shortest allowed key.
+  if (prefix.size() + 2 > key.size())
+    return std::nullopt;
+  if (!key.starts_with(prefix) || key[prefix.size()] != '_')
+    return std::nullopt;
+  auto suffix = key.substr(prefix.size() + 1);
```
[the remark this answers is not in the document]
mavam: Yes, I could, but I wanted the application prefix to not include the `_`.

libvast/src/system/configuration.cpp
```diff
@@ -34,33 +34,184 @@
...
+#include <cctype>
 #include <filesystem>
+#include <optional>
 #include <unordered_set>
 
 namespace vast::system {
 
 namespace {
 
+std::vector<std::filesystem::path> config_dirs_singleton = {};
```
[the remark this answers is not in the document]
mavam: No longer a thing after reverting this change in 1430b3088.

libvast/src/plugin.cpp
mavam: I've reverted this change and restored the original function signature of `config_dirs`. It turns we use this pattern of hauling things through the `actor_system_config` instance more pervasively than I initially thought.

libvast/src/detail/env.cpp
```diff
@@ -43,4 +47,16 @@ caf::expected<void> locked_unsetenv(const char* var) {
     fmt::format("failed in unsetenv(3): {}", ::strerror(errno)));
 }
 
+generator<std::pair<std::string_view, std::string_view>> environment() {
+  // Envrionment variables come as "key=value" pair strings.
+  for (auto env = environ; *env != nullptr; ++env) {
+    auto str = std::string_view{*env};
+    auto i = str.find('=');
+    VAST_ASSERT(i != std::string::npos);
```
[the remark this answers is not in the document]
lava: Yeah, fine with me. (returning an error would also be fine imho, but require a bigger refactoring)

lava (approved):
I investigated the last commit for a while, and it looks like the whole "convert settings based on the type in the configuration" is only done for options that come directly from caf. For the `vast.` options it looks like the default YAML/vast::data parser does (accidentally?) the correct thing already.

The logic on this part is pretty complicated so there may be some config parsing bugs left here, but even if so they would apply to the existing code and not be introduced by this PR so merging this should be fine.

PR #2165, opened 2022-03-28
Fix race condition with exporter timeouts

lava:
A query gets removed from the list of 'pending' queries
    in the index when the corresponding exporter actor goes
    DOWN.
    
    However, when the meta index lookup takes long enough the
    exporter can time out before the query is added to 'pending'.
    We did not check for this condition and thus would add
    an already-dead query to 'pending', which of course would
    never be able to progress afterwards.

### :memo: Checklist

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -1035,6 +1035,12 @@ index(index_actor::stateful_pointer<index_state> self,
         .then(
           [=, candidates
               = std::move(candidates)](meta_index_result& midx_result) mutable {
+            if (auto it = self->state.monitored_queries.find(sender->address());
+                it != self->state.monitored_queries.end()) {
+              VAST_VERBOSE("{} discards candidates for discarded query {}",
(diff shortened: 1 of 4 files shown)


libvast/src/system/index.cpp
```diff
@@ -1035,6 +1035,12 @@ index(index_actor::stateful_pointer<index_state> self,
         .then(
           [=, candidates
               = std::move(candidates)](meta_index_result& midx_result) mutable {
+            if (auto it = self->state.monitored_queries.find(sender->address());
```
lava, line 1038: After re-reading the commit I'm not so sure anymore that `sender` always refers to the correct actor here, at least for jobs that are taken from the backlog we probably need to pass an explicit `sender` argument to this actor.

lava: I've added another commit to make this more robust.

dispanser (approved):
@lava walked me through the changes, and the explanation makes sense to me.

PR #2170, opened 2022-03-29
Prepare the CLI scripts for the CI

rdettai:
This PR contains changes required for a CI running on Github Actions to work properly.

### :memo: Checklist

Here is what's happening:
- Use PTY only when necessary to avoid hanging with pipes
- Move files around for better organization, especially the `vast-server.yaml` config file needs to be at the same level as the Terraform files othewise it isn't uploaded to Terraform Cloud
- Improve UID and GID management in CLI to avoid conflicts if calling group already exists in the image. Also made the variable naming a bit more explicit.
- Lint terraform


dispanser (approved):
Looks good.

Haven't actually tried this b/c  I'm not a CI, though.

PR #2171, opened 2022-03-29
Add CI job for terraform (trial 2)

rdettai:
We need a way to ensure that the Terraform scripts to deploy VAST on AWS keep working throughout the product lifecycle. This PR adds a Github Action workflow to test some basic commands of the deployment CLI.

Note that:
- the CI run uses Terraform Cloud as a backend. This is safer than having the state locally in the Github action, because it might be hard to recover it if the CI job crashes.
- The run takes between 30 minutes and 1 hour to complete, mostly because tearing down some of the network resources is very slow on the AWS side.
- The CI run uses paying AWS resources. It tears down all the resources at the end of the job, so the costs of running it should be minimal (somewhere between $0.1 and $0.5). 
- It is worth noting that a failing job run might leave the resources running, which might incur higher costs. For this reason, this job is meant to be triggered manually and its outcome should be monitored.

### :memo: Checklist

- the required variables are configured through secrets, even though most of them are really just configuration variables. I am wondering if there is a better way to do this.
- the configuration instructions are in the Action configuration yaml file itself
- i used a random file `https://raw.githubusercontent.com/tenzir/vast/master/vast/integration/data/suricata/eve.json` for the ingestion test. We might want to use another one hosted in a more relevant place (Tenzir owned S3 bucket?)

### DIFF
--- a/.github/workflows/terraform.yaml
+++ b/.github/workflows/terraform.yaml
@@ -1,0 +1,156 @@
+name: "Terraform"
+on: workflow_dispatch
+
+# This CI configuration has multiple dependencies:
+# - You need an AWS account with a user having sufficient IAM access to run the
+#   Terraform scripts and other CLI commands.
+# - AWS credentials should be set in the Github repository secrets as
+#   AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY. They should have sufficient IAM
+#   priviledges to run Fargate tasks and Lambda functions.
+# - You need to choose an AWS region and a VPC within that region. These should
+#   be configured in the repository secrets as AWS_REGION and PEERED_VPC_ID.
+# - The VAST deployment creates a new VPC. Its ip range should not overlap an
+#   existing one in your account. Configure its CIDR in the repository secrets
+#   as VAST_CIDR (for example 172.30.0.0/16).
+# - The CI run uses Terraform Cloud to store the state. Within your Terraform
+#   Cloud account, you need to create two workspaces with the suffixes -step-1
+#   and -step-2. For each workspace, you also need to configure AWS credentials
+#   with sufficient priviledges to deploy the stack. Once this is done,
+#   configure the organization name and the common workspace prefix in the
+#   Github repository secrets as TF_ORGANIZATION and TF_WORKSPACE_PREFIX (the
+#   workspaces in Terraform Cloud should be $TF_WORKSPACE_PREFIX-step-1 and
+#   $TF_WORKSPACE_PREFIX-step-2).
+#   - You also need the API key from Terraform Cloud and set it in TF_API_TOKEN.
+
+# Notes:
+# - If this workflow is executed multiple times at in parallel, the Terraform
+#   state will be protected against inconsistencies by the lock provided by
+#   Terraform Cloud but the other tests will result in undefined behavior.
+# - If Terraform fails or is interrupted during deployment or destruction, the
+#   state might end up locked and subsequent runs will fail. In this case the
+#   state first needs to be unlocked, for instance from the Terraform Cloud UI.
+
+# Set these secrets to configure the CI run
+env:
+  VAST_PEERED_VPC_ID: "${{ secrets.VAST_PEERED_VPC_ID }}"
+  VAST_CIDR: "${{ secrets.VAST_CIDR }}"
+  VAST_AWS_REGION: "${{ secrets.VAST_AWS_REGION }}"
+  TF_ORGANIZATION: "${{ secrets.TF_ORGANIZATION }}"
+  TF_WORKSPACE_PREFIX: "${{ secrets.TF_WORKSPACE_PREFIX }}"
+  TF_API_TOKEN: ${{ secrets.TF_API_TOKEN }}
+  AWS_ACCESS_KEY_ID: "${{ secrets.AWS_ACCESS_KEY_ID }}"
+  AWS_SECRET_ACCESS_KEY: "${{ secrets.AWS_SECRET_ACCESS_KEY }}"
+  VASTCLOUD_NOTTY: 1
+
+jobs:
+  vast_on_aws:
+    name: "VAST ON AWS"
+    runs-on: ubuntu-20.04
+    defaults:
+      run:
+        working-directory: ./cloud/aws
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Terraform Format
+        run: terraform fmt -check -recursive -diff
+
+      - name: "Create config file"
+        if: ${{ !env.ACT }}
+        run: |
+          echo "vast_peered_vpc_id = $VAST_PEERED_VPC_ID" >> .env
+          echo "vast_cidr = $VAST_CIDR" >> .env
+          echo "vast_aws_region = $VAST_AWS_REGION" >> .env
+
+      - name: "Substitute backend with Terraform Cloud"
+        run: |
+          cat<<EOF > step-1/backend.tf
+          terraform {
+            cloud {
+              organization = "$TF_ORGANIZATION"
+              token        = "$TF_API_TOKEN"
+              workspaces {
+                name = "$TF_WORKSPACE_PREFIX-step-1"
+              }
+            }
+          }
+          EOF
+          cat<<EOF > step-2/backend.tf
+          terraform {
+            cloud {
+              organization = "$TF_ORGANIZATION"
+              token        = "$TF_API_TOKEN"
+              workspaces {
+                name = "$TF_WORKSPACE_PREFIX-step-2"
+              }
+            }
+          }
+          EOF
+
+      - name: Deploy
+        id: deploy
+        continue-on-error: true
+        run: ./vast-cloud deploy --auto-approve
+
+      - name: Retry deploy
+        if: steps.deploy.outcome=='failure'
+        run: |
+          echo "Deploy sometimes fails for an unexplained reason."
+          echo "Retrying after a few minutes usually works."
+          sleep 200
+          ./vast-cloud deploy --auto-approve
+
+      - name: Start and restart VAST server
+        run: |
+          echo "Run start-vast-server"
+          ./vast-cloud start-vast-server
+
+          echo "Run start-vast-server again"
+          ./vast-cloud start-vast-server 2> /dev/null \
+            && { echo "Starting server again should fail"; false; } \
+            || true
+          [[ $(./vast-cloud get-vast-server | wc -w) = "1" ]] \
+            || { echo "Only one task should be started"; false; }
+
+          echo "Run restart-vast-server"
+          ./vast-cloud restart-vast-server
+          [[ $(./vast-cloud get-vast-server | wc -w) = "1" ]] \
+            || { echo "Only one task should be started"; false; }
+
+          echo "The task needs a bit of time to boot, sleeping for a while..."
+          sleep 100
+
+      - name: Test db empty from Lambda
+        run: |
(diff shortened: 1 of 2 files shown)


rdettai:
@dominiklohmann 2 CI governance questions:

- should we move `terraform fmt -check` from here to the style-check.yaml workflow?
- should we remove the `docker build` for the lambda image in the vast.yaml workflow as it is covered here?

rdettai:
Waiting for https://github.com/tenzir/vast/pull/2170 to be merged...

rdettai:
Successful CI run on the fork: https://github.com/rdettai/vast/actions/runs/2060047672

.github/workflows/terraform.yaml
```diff
@@ -0,0 +1,156 @@
...
+        run: |
+          DATA_URL=https://raw.githubusercontent.com/tenzir/vast/master/vast/integration/data/suricata/eve.json
+          ./vast-cloud execute-command \
+            -c "wget -O - -o /dev/null $DATA_URL | vast import suricata"
+
+      - name: Test db not empty from Lambda
+        run: |
+          result=$(./vast-cloud run-lambda -c "vast count")
+          echo "Expected vast count 7, got $result"
+          [[ $result = "7" ]]
```
rdettai, line 142: @dispanser do you see any other test that would be valuable?

.github/workflows/terraform.yaml
```diff
@@ -0,0 +1,156 @@
...
+          [[ $(./vast-cloud get-vast-server | wc -w) = "1" ]] \
+            || { echo "Only one task should be started"; false; }
+
+          echo "Run restart-vast-server"
+          ./vast-cloud restart-vast-server
+          [[ $(./vast-cloud get-vast-server | wc -w) = "1" ]] \
+            || { echo "Only one task should be started"; false; }
+
+          echo "The task needs a bit of time to boot, sleeping for a while..."
+          sleep 100
```
dominiklohmann, line 124: These sleeps sound really brittle. Is there no way to idle until the server is available?

dominiklohmann:
> @dominiklohmann 2 CI governance questions:
> 
> * should we move terraform fmt -check from here to the style-check.yaml workflow?

Yes. Linting should not just run on workflow dispatch.

> * should we remove the docker build for the lambda image in the vast.yaml workflow as it is covered here?

No, as it's still valuable to verify that it actually builds correctly with every change to the Dockerfile.

.github/workflows/terraform.yaml
```diff
@@ -0,0 +1,156 @@
...
+        if: always()
+        run: ./vast-cloud destroy --auto-approve
+
+      - name: Retry destroy
+        if: steps.destroy.outcome=='failure'
+        run: |
+          echo "Destroy sometimes fails for an unexplained reason."
+          echo "Retrying after a few minutes usually works."
+          sleep 300
+          ./vast-cloud destroy --auto-approve
```
dominiklohmann, line 156: This can be simplified a fair bit by simply running in a loop in the run block with a configured amount of retries and sleep intervals. What are the consequences if the retries fail?

.github/workflows/terraform.yaml
```diff
@@ -0,0 +1,156 @@
...
+        continue-on-error: true
+        run: ./vast-cloud deploy --auto-approve
+
+      - name: Retry deploy
+        if: steps.deploy.outcome=='failure'
+        run: |
+          echo "Deploy sometimes fails for an unexplained reason."
+          echo "Retrying after a few minutes usually works."
+          sleep 200
+          ./vast-cloud deploy --auto-approve
```
dominiklohmann, line 104: Same deal here with the sleep. Let's figure out a way to reliably detect whether the deployment succeeded. There's gotta be a way to query that with the API. Ideally, the `vast-cloud deploy` command would only exit once VAST can be interacted with and abstract this away for us.

.github/workflows/terraform.yaml
```diff
@@ -0,0 +1,156 @@
...
+
+      - name: Start and restart VAST server
+        run: |
+          echo "Run start-vast-server"
+          ./vast-cloud start-vast-server
+
+          echo "Run start-vast-server again"
+          ./vast-cloud start-vast-server 2> /dev/null \
+            && { echo "Starting server again should fail"; false; } \
+            || true
```
dominiklohmann, line 114: Instead of this complicated Boolean logic, can we use an if condition here for readability? Same below this snippet.

.github/workflows/terraform.yaml
```diff
@@ -0,0 +1,156 @@
...
+
+      - name: Test db empty from Lambda
+        run: |
+          result=$(./vast-cloud run-lambda -c "vast count")
+          echo "Expected vast count 0, got $result"
+          [[ $result = "0" ]]
+
+      - name: Import data
+        run: |
+          DATA_URL=https://raw.githubusercontent.com/tenzir/vast/master/vast/integration/data/suricata/eve.json
```
dominiklohmann, line 134: This URL always points to master. Can you make it point to the current commit hash instead? If you're running in a `workflow_dispatch` context, this should work:

```
DATA_URL=https://raw.githubusercontent.com/tenzir/vast/blob/${{ github.sha }}/vast/integration/data/suricata/eve.json
```

.github/workflows/terraform.yaml
```diff
@@ -0,0 +1,156 @@
...
+  TF_ORGANIZATION: "${{ secrets.TF_ORGANIZATION }}"
+  TF_WORKSPACE_PREFIX: "${{ secrets.TF_WORKSPACE_PREFIX }}"
+  TF_API_TOKEN: ${{ secrets.TF_API_TOKEN }}
+  AWS_ACCESS_KEY_ID: "${{ secrets.AWS_ACCESS_KEY_ID }}"
+  AWS_SECRET_ACCESS_KEY: "${{ secrets.AWS_SECRET_ACCESS_KEY }}"
+  VASTCLOUD_NOTTY: 1
+
+jobs:
+  vast_on_aws:
+    name: "VAST ON AWS"
```
dominiklohmann, line 47: Correct title case would be `VAST on AWS`. I don't think the quotes are needed either, it's already a string in YAML.

dominiklohmann (changes requested):
Slightly belated review. I've got a few things here I'd like to see changed, especially wrt robustness.

.github/workflows/terraform.yaml
rdettai: I think that `https://raw.githubusercontent.com/tenzir/vast/blob/${{ github.sha }}` will have the drawback of not existing in a fork. If we go this way, I propose `https://raw.githubusercontent.com/${{ github.repository }}/blob/${{ github.sha }}`.

Will this not work if we are not in `workflow_dispatch` ? Isn't the commit always published when the workflow is executed, independently from the type of trigger?

dominiklohmann: Some triggers like `issue_comment` don't have a SHA attached.

+1 on embedding the repo as well.

.github/workflows/terraform.yaml
rdettai: I wanted to avoid making it too verbose. But you are right, this syntax is cryptic (in fact, I think the statement you highlighted is wrong, because it does not fail when it should). 

Isn't there a better way to assert in bash? Another solution would be to add a `vast-cloud integgration-test` command that runs all these. This brings a lots of benefits:
- give the ability to run the tests locally
- take advantage of the python syntax to do the asserts
- the github action code remains simpler

.github/workflows/terraform.yaml
dispanser: If the `destroy` doesn't finish successfully, there can be resources remaining that 
1. might cost money
2. might block a deploy, e.g. on the next iteration

.github/workflows/terraform.yaml
dispanser: > Another solution would be to add a `vast-cloud integgration-test` command that runs all these

That sounds like a great idea. Local reproducability is always helpful

dispanser:
Given that this workflow isn't running any longer than some of our other ones, I think it might be valuable to actually run this on every commit. The only thing I'm worried about is that this worflow must be unique wrt to the entire workflow, so multiple PRs cannot run this concurrently.

.github/workflows/terraform.yaml
rdettai: @dominiklohmann Do you mean a bash loop or is there a specific syntax with Github Action? With the current implementation I do like the fact that you can see from the execution summary what failed or not without looking at the logs. But I do agree that the current formulation is pretty verbose for a simple retry.

Not sure that having more than 1 retry would bring a lot of value, Terraform already implements a pretty comprehensive retry strategy, and it's quite rare that having an extra layer of retries is useful. Here it is necessary because we are using a resource with a specific configuration (Lambda in a VPC) that sometimes behaves unexpectedly and Terraform doesn't manage it well.

.github/workflows/terraform.yaml
rdettai: Managing retries is never that simple, because you have failures that you will want to retry and others that you don't. I think that improving this could be a follow up task that might be addressed later.

.github/workflows/terraform.yaml
rdettai: We could poll the AWS API or the run-command until it is ready. I think we can do it in a [subsequent task](https://github.com/tenzir/vast/issues/2182).

.github/workflows/terraform.yaml
dominiklohmann: I think this becomes obsolete if we can fix https://github.com/tenzir/vast/pull/2171#discussion_r841217213, so I don't mind either way until then.

dominiklohmann:
> Does it seam reasonable?

Yes. I'll leave the remaining review to @dispanser.

@dispanser We still need to add the secrets before we merge this. Can you send me what I need to add privately?

.github/workflows/terraform.yaml
rdettai: Actually, could we rather set this as an immutable external resource, for instance on S3 or a pinned commit? Depending on the Github environment complicates things if we want the tests to run from other places (like locally...)

rdettai: (also, the right URL seems to be `https://raw.githubusercontent.com/tenzir/vast/${{ github.sha }}` and not `https://raw.githubusercontent.com/tenzir/vast/blob/${{ github.sha }}` )

dominiklohmann: > Actually, could we rather set this as a hardcoded immutable external resource, for instance on S3 or a pinned commit? Depending on the Github environment complicates things if we want the tests to run from other places (like locally...)

Personally I'd prefer to keep all things related to the CI in the CI. You could also in-line the file contents in the workflow, given that it's only 7 lines.

rdettai: The issue here is that the data needs to be accessed from the cloud. Here we are asserting both that:
- we can access the public internet
- we can load then read the 7 lines of data

Inlining the content in the workflow does not solve that. This is why we were using `raw.githubusercontent.com` in the first place, because it is publicly accessible.
- The most simple solution I see for now is to pin to a specific version (say v1.0.0) that is immutable and will be kept around "indefinitely".
- We could create a cloud resource within the stack and upload the data for the duration of the script. But this would require quite some plumbing to set the access rights correctly. I think this should be considered if at some point we integrate such a resource to the stack for some other purpose.
- We could create a cloud resource independent from the stack that hosts this data, but that will not bring much value over using a pined version on `raw.githubusercontent.com`

dispanser (approved):
Thanks, Rémi! The code looks looks good enough to merge for now. Some larger items have been moved into separate issues to tacke independently.

PR #2176, opened 2022-03-31
Reduce the default log queue size for client commands

tobim:
The previous default of 1'000'000 lead to about 400MB of additional memory consumption. This is acceptable for the server but not for the often short-running client commands.

This PR also exposes the `--log-queue-size` option on the command line.

### :memo: Checklist


lava (approved):
Makes sense to me.

PR #2177, opened 2022-04-01
Add Terragrunt to the stack

rdettai:
Adding Terragrunt helps maintaining the configurations DRY and also
improves the dependency management between independently deployed
modules. This will be more and more important as we add optional
modules for the playground.

### :memo: Checklist

An interesting aspect of this change is the `./vast-cloud [deploy|destroy]` commands that now use  `terragrunt run-all` instead of chaining the modules themselves.

Another interesting catch is the usage of Terragrunt `after_hook` in the step-1 to deploy the lambda and a `run_cmd` variable resolution in the step-2. These both call vast-cloud from within the vast-cloud process. To make it work, some improvements had to be made to the `vast-cloud` script.


rdettai:
I rebased and after some struggle, I updated the CI to use Terragrunt. I found that Terragrunt is not always up to expectations with regards to the UX and documentation of the configuration functions (for instance the `remote_state` block does not support the `remote` backend (https://github.com/gruntwork-io/terragrunt/issues/779), and this is not documented anywhere)

dispanser (approved):
Tested on a local stack, works for me and looks nice.

PR #2178, opened 2022-04-03
Fix CLI verbosity shorthands

mavam:
This PR adds a patch to fix the mapping of shorthand verbosity options (e.g., `-v`, `-q`) to the proper underlying option `--console-verbosity`.

I realized that renaming `--verbosity` to `--console-verbosity` is actually a breaking change. We may want to revert it here, or in my opinion preferably, deprecate `--verbosity` in favor of `--console-verbosity`. The reason for preferring the latter is that it is the vocabulary used in the configuration files, as well as clearer what this does.

### :memo: Checklist

Look at the one-line change. :eyes:


dominiklohmann:
> I realized that renaming `--verbosity` to `--console-verbosity` is actually a breaking change.

This should get a separate changelog entry. I think the new name is better for the very reason you listed, but we should still document it.

dominiklohmann (approved):

PR #2179, opened 2022-04-03
Add VLAN tags to PCAP meta data

mavam:
This PR makes [802.1Q](https://en.wikipedia.org/wiki/IEEE_802.1Q) VLAN tag information queryable. VLAN parsing encompasses *single tagging* and *double tagging*. The first 802.1Q header after the MAC address is the *outer* VLAN tag and an optional, sub-sequent VLAN tag the *inner* VLAN tag:

TCPIP_802 1ad_DoubleTag svg

To expose the tags, the packet schema now includes a new sub-record `vlan` with two fields `outer` and `inner`. For example, to query all non-zero outer VLAN tags, you can know write `vlan.outer > 0`.

### :memo: Checklist

- [x] How do I add a changelog entry for plugins? IIRC we have just one `CHANGELOG.md` file in the top-level plugin directory. There's no `changelog` directory from where the master changelog gets generated, as we do for VAST itself.
- [x] Update [docs.tenzir.com/vast](https://docs.tenzir.com/vast).

### 💬 Discussion 

- [x] Should we remove `--disable-community-id` flag? We are unconditionally generating Community IDs now. The configuration flag only determines whether we add the extra string to the table slice.

Commit-by-commit.

Appreciate that the decapsulation logic has now moved into a separate `frame` abstraction that provides a factory function `frame::make`. Note that detailed error handling is not important because we're simply skipping invalid frames, which is why the function only returns a `std::optional<frame>`.

The last refactoring commit only shuffles code around but doesn't make any changes to the big-picture logic. The change adds several bounds checks to make the overall parsing safer as well.


dominiklohmann:
> How do I add a changelog entry for plugins? IIRC we have just one `CHANGELOG.md` file in the top-level plugin directory. There's no `changelog` directory from where the master changelog gets generated, as we do for VAST itself.

Just modify the plugin's changelog file, and bump its version. In this case it's a breaking change due to the removal of the `--disable-community-id` flag.

We probably need two separate changelog entries here.

> Update [docs.tenzir.com/vast](https://docs.tenzir.com/vast).

The graphic from this PR would probably also do nicely on there.

> Should we remove `--disable-community-id` flag?

I think so, yes.

mavam:
> Just modify the plugin's changelog file, and bump its version. In this case it's a breaking change due to the removal of the `--disable-community-id` flag.

Actually, the behavior of the flag just changed in that it no longer changes the schema. I decided to keep it.

mavam:
@dominiklohmann there's an issue with fmt 8.1 on macOS CI:

```c++
 VAST_DEBUG("skipping packet of type {:#0x}", frame->type);
```

This compiles fine locally, but the enum `frame->type` is apparently not auto-printable in CI. How should this be fixed?

dominiklohmann:
> This compiles fine locally, but the enum `frame->type` is apparently not auto-printable in CI. How should this be fixed?

It's an `enum class`, so you need to write a custom formatter for it. This has been the case since {fmt} 8.0, it's not a breaking change in 8.1.

mavam:
@dominiklohmann does this look reasonable?

```diff
diff --git a/plugins/pcap/main.cpp b/plugins/pcap/main.cpp
index f36e6a2b9..fe74c8e0f 100644
--- a/plugins/pcap/main.cpp
+++ b/plugins/pcap/main.cpp
@@ -143,6 +143,28 @@ enum class frame_type : char {
   vxlan = '\x14'
 };

+} // namespace vast::plugins::pcap
+
+namespace fmt {
+
+template <>
+struct formatter<vast::plugins::pcap::frame_type> {
+  template <typename ParseContext>
+  constexpr auto parse(ParseContext& ctx) {
+    return ctx.begin();
+  }
+
+  template <typename FormatContext>
+  auto format(vast::plugins::pcap::frame_type value, FormatContext& ctx) {
+    // TODO(C++23): std::to_underlying instead of static_cast.
+    return format_to(ctx.out(), "{:#0x}", static_cast<char>(value));
+  }
+};
+
+} // namespace fmt
+
+namespace vast::plugins::pcap {
+
 /// An 802.3 Ethernet frame.
 struct frame {
   static std::optional<frame>
@@ -505,7 +527,7 @@ protected:
       auto packet = packet::make(frame->payload, frame->type);
       if (!packet) {
         ++discard_count_;
-        VAST_DEBUG("skipping packet of type {:#0x}", frame->type);
+        VAST_DEBUG("skipping packet of type {}", frame->type);
         continue;
       }
       // Parse layer 4.
```

dominiklohmann:
Yes, except for the style nit `s/typename/class` in the templates. You can also inherit from the existing formatter for `char`, which makes everything even simpler and gains you support for `{:x}` for free.

plugins/pcap/CHANGELOG.md
```diff
@@ -2,6 +2,23 @@
...
+## v2.0.0
+
+### :gift: Features
+
+- VAST no longer discards parsed 802.1Q VLAN tags but instead writes them into
+  a new record `vlan`. This record contains two fields `outer` and `inner` that
+  correspond to the outer and inner VLAN identifier (VID), respectively.
+  [#2179](https://github.com/tenzir/vast/pull/2179)
+
+### :zap: Breaking Changes
```
dominiklohmann, line 14: I think this is just a change, not a breaking change.

```suggestion
### :warning: Changes
```

plugins/pcap/CHANGELOG.md
```diff
@@ -2,6 +2,23 @@
 
 This changelog documents all notable changes to the PCAP plugin for VAST.
 
+## v2.0.0
```
dominiklohmann, line 5:
```suggestion
## v1.1.0
```

Assuming you agree that this is not a breaking change.

dominiklohmann (approved):
I've verified that the new integration tests are correct using Wireshark.

plugins/pcap/CHANGELOG.md
mavam: It's breaking in that the packet layout is now different: the `community_field` is there even when you add `--disable-community-id`. The flag only change the value from the string to `null`.

dominiklohmann: That's not a breaking change, no? Adding a field should be backwards compatible in the type registry.

mavam:
@dominiklohmann can you explain to me the {fmt} error message?

dominiklohmann:
> @dominiklohmann can you explain to me the {fmt} error message?

I think your formatter shouldn't be a `const` member function. It can be `constexpr`, though, but the parent format function is definitely not marked `const`.

mavam:
> I think your formatter shouldn't be a `const` member function. It can be `constexpr`, though, but the parent format function is definitely not marked `const`.

Hm, I copied the example almost 1-to-1 from the {fmt} docs.

mavam:
Now you're trading macOS for Debian CI. 😆

PR #2186, opened 2022-04-04
Implement support for transforms that apply to every type and use compaction for aging

lava:
Aging used to work only partially; deleting events from the archive but not from the index. It was deprecated and scheduled for removal in VAST 2.0.

With this PR we change the internals of the eraser, so aging now works as a functionality-reduced form of temporal compaction.

Some supporting commits add support for transforms that apply to every type, and a new "invert" option for the "select" transform step.

### :memo: Checklist

### DIFF
--- a/libvast/include/vast/system/eraser.hpp
+++ b/libvast/include/vast/system/eraser.hpp
@@ -12,7 +12,10 @@
 
 #include "vast/ids.hpp"
 #include "vast/system/actors.hpp"
-#include "vast/system/query_processor.hpp"
+#include "vast/system/index.hpp"
+
+#include <caf/settings.hpp>
+#include <caf/typed_response_promise.hpp>
 
 #include <string>
 
@@ -20,34 +23,25 @@ namespace vast::system {
 
 /// Periodically queries the INDEX with a configurable expression and erases
 /// all hits from the ARCHIVE.
-class eraser_state : public query_processor {
+class eraser_state {
 public:
-  // -- member types -----------------------------------------------------------
-
-  using super = system::query_processor;
-
   // -- constants --------------------------------------------------------------
 
   static inline constexpr auto name = "eraser";
 
   // -- constructors, destructors, and assignment operators --------------------
 
-  eraser_state(caf::event_based_actor* self);
+  eraser_state() = default;
 
-  void init(caf::timespan interval, std::string query, index_actor index);
-
-protected:
   // -- implementation hooks ---------------------------------------------------
 
-  void transition_to(state_name x) override;
+  [[nodiscard]] record status(status_verbosity) const;
 
-  record status(status_verbosity) override;
-
-private:
   // -- member variables -------------------------------------------------------
+  index_actor index_ = {};
 
   /// Configures the time between two query executions.
-  caf::timespan interval_;
+  caf::timespan interval_ = {};
 
   /// The query expression that selects events scheduled for deletion. Note
   /// that we get the query as string on purpose. Taking an ::expression here
@@ -60,7 +54,7 @@ private:
 
   /// Keeps track whether we were triggered remotely and need to send a
   /// confirmation message and suppress the delayed message.
-  caf::response_promise promise_;
+  caf::typed_response_promise<atom::ok> promise_ = {};
 };
 
 /// Periodically queries `index` and erases all hits from `archive`.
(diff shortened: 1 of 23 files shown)


libvast/include/vast/system/eraser.hpp
```diff
@@ -12,67 +12,55 @@
...
 };
 
 /// Periodically queries `index` and erases all hits from `archive`.
 /// @param interval The time between two query executions.
 /// @param query The periodic query that selects events scheduled for deletion.
 ///              Note that we get the query as string on purpose. Taking an
-///              ::expression here instead would fix any query such as `#time <
-///              1 week ago` to the time of its parsing and not update
+///              ::expression here instead would fix any query such as
+///              `#time < 1 week ago` to the time of its parsing and not update
```
dominiklohmann, line 58: Nit: `#time < 1 week ago` is not a valid query. Same above in the documentation of the `query_` member variable of the state.

libvast/include/vast/system/eraser.hpp
```diff
@@ -12,67 +12,55 @@
...
 ///              properly.
 /// @param index A handle to the INDEX under investigation.
-caf::behavior
-eraser(caf::stateful_actor<eraser_state>* self, caf::timespan interval,
-       std::string query, index_actor index);
+/// @param transforms The transform config of the node.
```
dominiklohmann, line 61: This parameter was not added to the actual function signature afaict.

libvast/include/vast/system/make_transforms.hpp
```diff
@@ -38,7 +38,7 @@ make_transforms(transforms_location location, const caf::settings& settings);
 /// name.
 caf::expected<transform_ptr>
 make_transform(const std::string& name,
-               const std::vector<std::string>& event_types,
+               const std::optional<std::vector<std::string>>& event_types,
```
dominiklohmann, line 41: Isn't the optional part already covered by the empty vector?

libvast/include/vast/transform.hpp
```diff
@@ -33,18 +34,22 @@ class transform {
...
   /// @note The offsets of the slices may not be preserved.
   [[nodiscard]] caf::expected<std::vector<table_slice>> finish();
 
-  [[nodiscard]] const std::vector<std::string>& event_types() const;
-
   [[nodiscard]] const std::string& name() const;
 
 private:
+  [[nodiscard]] const std::optional<std::vector<std::string>>&
+  event_types() const;
```
dominiklohmann, line 51: Should probably be named "schema_names" or so.

libvast/include/vast/transform_steps/select.hpp
```diff
@@ -18,22 +18,26 @@ struct select_step_configuration {
   // The expression in the config file.
   std::string expression;
 
+  // Whether to select or to filter.
+  bool invert = false;
```
dominiklohmann, line 22: We have both `project` and `delete` steps, so why not introduce a `filter` step whose plugin implementation simply refers to the `filter` plugin?

libvast/src/system/eraser.cpp
```diff
@@ -14,80 +14,102 @@
...
+        = std::make_shared<vast::transform>("eraser_transform", std::nullopt);
+      auto select_config = select_step_configuration{
+        .expression = self->state.query_,
+        .invert = true,
+      };
+      transform->add_step(std::make_unique<select_step>(select_config));
+      auto rp = self->make_response_promise<atom::ok>();
+      self->request(self->state.index_, caf::infinite, atom::resolve_v, *expr)
+        .then(
+          [self, transform, rp = rp](catalog_result& result) mutable {
```
dominiklohmann, line 84: Why the `rp = rp` over just `rp` here and a few times below?

dominiklohmann (commented):
This is great, and thanks especially for turning the eraser into a typed actor.

I think this is almost ready to go. We need to undeprecate aging and also mark it as no longer experimental, and the rest should be covered by the remaining TODO comments and the few comments I left on the PR, which are mostly of stylistic nature.

libvast/include/vast/system/make_transforms.hpp
lava: Good question, I thought about it but felt that using an explicit optional is less surprising. I don't have a strong opinion though, so I'll change to empty vector.

dominiklohmann (approved):

libvast/src/system/eraser.cpp
lava: Good question, there used to be a mysterious compiler error if I didn't, but I can't reproduce.

dominiklohmann:
This is only missing a rebase, a changelog entry and a documentation update now, everything else is now good to go.

dispanser (approved):
I verified the last four commits done by @dominiklohmann . We talked about those earlier.

dominiklohmann:
[sc-32123]

PR #2189, opened 2022-04-05
Fix disk monitor partition blacklist

dominiklohmann:
This fixes a bug that would cause no partitions to be deleted by the disk monitor at as soon as the disk monitor had at least one blacklisted partition.

This does not need a changelog entry since the blacklist feature did not make it into a release yet.

### :memo: Checklist

Commit-by-commit.


lava (approved):
Good catch, thanks for the fix

PR #2192, opened 2022-04-05
Also look at type synopses when computing partition synopsis size

lava:
After the recent meta index configuration changes, the partition synopsis size would always show as `0` because it only contains type synopses by default.

### :memo: Checklist

### DIFF
--- a/libvast/include/vast/partition_synopsis.hpp
+++ b/libvast/include/vast/partition_synopsis.hpp
@@ -32,10 +32,10 @@ struct partition_synopsis final : public caf::ref_counted {
   partition_synopsis() = default;
   ~partition_synopsis() override = default;
   partition_synopsis(const partition_synopsis&) = delete;
-  partition_synopsis(partition_synopsis&&) = default;
+  partition_synopsis(partition_synopsis&& that) noexcept;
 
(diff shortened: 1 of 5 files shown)


libvast/include/vast/partition_synopsis.hpp
```diff
@@ -90,6 +90,9 @@ struct partition_synopsis final : public caf::ref_counted {
   friend partition_synopsis* ::caf::default_intrusive_cow_ptr_unshare<
     partition_synopsis>(partition_synopsis*& ptr);
   partition_synopsis* copy() const;
+
+  // Cached memory usage.
+  size_t memusage_ = 0ull;
```
dominiklohmann, line 95: Probably needs to be made `mutable` now, with a note that the function calculating the memusage is no longer thread safe despite being marked `const`. Or, alternatively, you no longer mark that function `const`.

dominiklohmann (approved):
Should be good to go.

libvast/include/vast/partition_synopsis.hpp
lava: I actually went for both by using an atomic here.

PR #2193, opened 2022-04-06
Minimize the threadpool for client commands

tobim:
Running many sub-commands in parallel can quickly result in very large numbers of threads being created. This can become a problem when these processes are in a restricted environment, and some start receiving "resource unavailable" errors. We can mitigate the problem by using only the minimum amount of workers for client commands, which don't need a these resources anyways.

### :memo: Checklist

### DIFF
--- a/changelog/unreleased/changes/2193--client-command-threads.md
+++ b/changelog/unreleased/changes/2193--client-command-threads.md
@@ -1,0 +1,2 @@
+Client commands such as `vast export` or `vast status` now create a less threads
+at runtime, lessening the risk of hitting system resource limits.
--- a/changelog/unreleased/changes/2193--client-command-threads.md
+++ b/changelog/unreleased/changes/2193--client-command-threads.md
@@ -1,2 +1,2 @@
-Client commands such as `vast export` or `vast status` now create a less threads
+Client commands such as `vast export` or `vast status` now create less threads
 at runtime, lessening the risk of hitting system resource limits.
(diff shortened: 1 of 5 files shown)


vast/vast.cpp
```diff
@@ -77,6 +77,11 @@ int main(int argc, char** argv) {
   // From here on, options from the command line can be used.
   detail::merge_settings(invocation->options, cfg.content,
                          policy::merge_lists::yes);
+  // Tweak CAF parameters in case we're running a client command.
+  bool is_server = invocation->full_name == "start"
+                   || caf::get_or(cfg.content, "vast.node", false);
+  if (!is_server)
+    cfg.set("scheduler.max-threads", 1);
```
dominiklohmann, line 84: Can you add a comment here to adjust this with a `caf.` prefix when updating to CAF 0.18+?

dominiklohmann, line 84: The changed default needs to also be updated in the `vast.yaml.example` file.

changelog/unreleased/changes/2193--client-command-threads.md
```diff
@@ -0,0 +1,2 @@
+Client commands such as `vast export` or `vast status` now create a less threads
```
dominiklohmann, line 1:
```suggestion
Client commands such as `vast export` or `vast status` now create less threads
```

vast/vast.cpp
dominiklohmann, line 84: Should we maybe only set this to 1 if the user did not specify the option, i.e., respect the usual precedence of command line > env > config > defaults? I think this could otherwise lead to very unexpected behavior.

dominiklohmann (commented):
Some initial thoughts. Will do some tests locally with this as well to see if any client commands run into issues now. There's always at least two actors running if we have a remote connection, so I'd like to test whether the new default should be 2 rather than 1.

dominiklohmann (approved):
This actually improved performance for client commands for me a noticeable bit. Thanks!

PR #2197, opened 2022-04-07
Experiment: Add support for inline export transforms using a pipe syntax

dominiklohmann:
This adds support for specifying export transforms in a convenient pipe syntax, e.g.,

```
vast export --pipe='aggregate(time-resolution: 1 day, group-by: ["timestamp"], sum: ["pkts_toclient", "pkts_toserver"])' json '#type == "suricata.flow" '|
```

Ultimately, the end goal would be to make this part of the query itself and to push it down to the query evaluation. The current state of this PR contains a hack-ish implementation that already has a huge benefit by making it easier to specify export transforms.

### :memo: Checklist

Don't, this is an experiment from a hackathon. Only keeping this here because we want to pick up the idea again at a later point in time, but please know that may be quite a ways out.


dominiklohmann:
I'm closing this one. We're in the process of redesigning pipelines, and that will supersede this experimental PR.

PR #2198, opened 2022-04-07
Untangle `exit_msg` relationship triangle in `sink_command`

dominiklohmann:
The sink_command must not force the exporter to exit before all events are sent out. Instead, it now no longer sends an exit message to the exporter, but rather as a normal prioriity message to the sink which the exporter monitors. This untangles the exit message relationship triangle between sink_command, sink, and exporter.

In effect, this fixes a rare race condition for fast queries with a large result set where not all events would successfully arrive at the client because the exporter caused the sink to shutdown before the sink relayed all events to the the sink_command. This bug also resulted in a non-zero exit code.

### :memo: Checklist

This fixes a race condition I found while reviewing #2193.

NOTE: The CI for this does not appear to pass yet, which I don't fully understand yet since it passes for me locally. I am investigating why this is; I may have introduced another race here.


dominiklohmann:
Superseded by the shutdown fix in #2215.

PR #2209, opened 2022-04-12
Add a convertible overload from data to string

dominiklohmann:
This fixes two issues downstream in the compaction plugin with the bytesize parser:
- Passing '90' would be parsed as a YAML scalar, which in turn becomes a VAST integer, which then failed to be converted to a string.
- As an immediate follow-up, I noticed that passing an integer to the bytesize parser is also not a good option. Because YAML scalars and CAF settings only support signed integers, we must prefer not rendering the unary plus in front of positive signed integers, as otherwise downstream parsers cannot handle the resulting strings from convertible.

### :memo: Checklist

I don't think this needs a changelog entry here, as this had no visible effect inside VAST directly. Rather, I'll add a changelog entry to the compaction plugin when updating it to incorporate this change.

### DIFF
--- a/libvast/include/vast/concept/convertible/data.hpp
+++ b/libvast/include/vast/concept/convertible/data.hpp
@@ -19,6 +19,7 @@
 #include "vast/error.hpp"
 #include "vast/logger.hpp"
 #include "vast/type.hpp"
+#include "vast/view.hpp"
 
 #include <caf/error.hpp>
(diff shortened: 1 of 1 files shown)


tobim (approved):
I feel a little uneasy about this conversion, but I will accept it because it doesn't conflict with any of the others.

PR #2159, opened 2022-03-21
Mark `experimental` encoding as `arrow.v2`

dominiklohmann:
This makes the previously experimental encoding the new default.

Outstanding tasks:

- [x] Make the experimental encoding the fault encoding
- [x] Remove the old Arrow table slice builder
- [x] Rename the new Arrow table slice builder
- [x] Add utilities for working with nested struct arrays in record batches
- [x] ~Update transform step: `identity`~ No changes required
- [x] Update transform step: `hash`
- [x] ~Update transform step: `count`~ No changes required
- [x] Update transform step: `delete`
- [x] ~Update transform step: `rename`~ No changes required
- [x] ~Update transform step: `select`~ No changes required
- [x] Update transform step: `project`
- [x] Update transform step: `replace`
- [x] Update transform step: `aggregate`

### :memo: Checklist

- [x] All user-facing changes have changelog entries.
- [x] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [x] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

Commit-by-commit.

PR #2210, opened 2022-04-12
Fix a segfault in the aggregate plugin

dominiklohmann:
This was just recently introduced in #2159. Just a minimal oversight, and since this has not made a release yet I don't think this deserves a separate changelog entry.

### :memo: Checklist

Run with any configuration that drops field that are before any group by fields to reproduce the crash.


6yozo (approved):
I have tested it on my local machine and the fix works as expected now.

PR #2212, opened 2022-04-12
Fix infinite loop in query removal

tobim:
This fixes a severe bug that can deadlock the index actor. The problem surfaced with one of the integration tests which should be more reliable after this is done.

### :memo: Checklist


mavam (approved):
The infinite loop is easy to see.

PR #2215, opened 2022-04-13
Stop accepting new queries after initiating shutdown

dominiklohmann:
Before this change, the index continued accepting queries after initiating shutdown, which could lead to an infinite hang because queries were coming in faster than the index was able to process them.

### :memo: Checklist

Try locally:

```
T1> vast [--use-legacy-query-scheduler] start
T2> while true; do { sleep 0.5; vast export null & }; done
T3> vast stop
```

### DIFF
--- a/changelog/unreleased/bug-fixes/2215--new-queries-after-stop.md
+++ b/changelog/unreleased/bug-fixes/2215--new-queries-after-stop.md
@@ -1,0 +1,3 @@
+VAST servers no longer accept queries after initiating shutdown. This fixes a
+potential infinite hang if new queries were coming in faster than we were able
+to process them.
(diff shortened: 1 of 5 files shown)


tobim (approved):
I got a live demo that the fix works.

dominiklohmann:
@tobim I've re-requested your review after the last two commits.

tobim (commented):
Before we accept this additional complexity, can we test whether bidirectional linkage between all 3 actors would work? I mean the sink command calls `self->link_to(snk)` and `self->link_to(exporter)`, and the exporter links itself to the sink? The rest of the lifetime management in the sink command can probably ripped out.

dominiklohmann:
> Before we accept this additional complexity, can we test whether bidirectional linkage between all 3 actors would work?

It does work, but we lose the ability to give a good exit code and error message if there's an unexpected shutdown. So I think listening for down messages is still preferred.

PR #2218, opened 2022-04-14
Fix heap-use-after-free in aggregate group-by keys

dominiklohmann:
It's not safe to store a `data_view` pointing into record batches whose underlying data may no longer be around after calls to `aggregation::add` when we want to access them in `aggregation::finish` later on. So now we store `data` instead, and to avoid having to materialize as often we instead use a robin map with support for heterogenous lookup from views.

This was a recently introduced bug only. Before the changes from #2159 we used `arrow::Scalar` for the bucket keys, which keeps the underlying data alive because it contains a strong reference to the buffer it is sliced from unlike `data_view`. Because this bug was never in a released version I think we do not need a changelog entry here.

### :memo: Checklist

@tobim try to reproduce the heap-use-after-free you've seen with the changes from this branch. For the actual code, it should be rather straightforward to review.

### DIFF
--- a/plugins/aggregate/aggregate.cpp
+++ b/plugins/aggregate/aggregate.cpp
@@ -25,8 +25,7 @@
 #include <arrow/config.h>
 #include <arrow/table.h>
 #include <arrow/table_builder.h>
-
-#include <unordered_map>
+#include <tsl/robin_map.h>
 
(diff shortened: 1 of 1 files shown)


tobim (approved):
I inspected the code and believe that this change solves the known heap-use-after free problem.

PR #2226, opened 2022-04-19
Fix a false assertion failure on client shutdown

dominiklohmann:
This fixes a recently introduced bug that only occured in builds with assertions enabled, and thus caused CI to fail occasionally. When zeroing out the exporter and sink actor handles we can no longer compare their addresses with the address from the monitored actor in the down handler of the sink command. In case both sink and exporter issued a down message before the sink command was able to shut down the pipeline we ran into an assertion failure because the second down message would be from an actor that the sink command no longer knew. This is now fixed by taking the addresses of the actors by value rather than keeping a reference to the actor handles, and doubly so by demoting the assertion to a warning message.

### :memo: Checklist

@tobim since you already reviewed #2215 which introduced this bug it probably makes sense for you to take a look at this as well.


lava (approved):
Fix looks correct to me.

PR #2228, opened 2022-04-20
Clean up transform steps (and native plugins generally)

dominiklohmann:
This follows up on an internal discussion to rename some transform step before the upcoming major release of VAST v2.0 to align their names with what we generally deemed most suitable.

Here's a summary of the changes:
- The `rename` step can now rename fields as well as schemas.
- The `where` step supersedes both the `filter` and `select` steps.
- The `delete` step is now called `drop`.
- The `project` step is now called `put`.
- The `replace` step now supports all complex types that can be inferred.
- The `aggregate` step is now called `summarize`.

And lastly, developer-facing only:
- Native plugins no longer need to be explicitly registered in the `CMakeLists.txt` file of `libvast`.
- It is now harder to misuse native plugins code by calling it directly from `libvast`; instead, calls must now go through the plugin API. This supersedes #2220.

### :memo: Checklist

Review the description. Then read the code commit-by-commit.

Before merging this still needs a documentation update.


libvast/src/transform_steps/drop.cpp
```diff
@@ -0,0 +1,121 @@
...
+      -> std::vector<
+        std::pair<struct record_type::field, std::shared_ptr<arrow::Array>>> {
+      return {};
+    };
+    auto transformations = std::vector<indexed_transformation>{};
+    for (const auto& field : config_.fields)
+      for (auto&& index : caf::get<record_type>(layout).resolve_key_suffix(
+             field, layout.name()))
+        transformations.push_back({std::move(index), transform_fn});
+    std::sort(transformations.begin(), transformations.end());
```
mavam: Not clear from the control flow why this needs sorting; maybe comment; not urgent though.

libvast/src/transform_steps/replace.cpp
```diff
@@ -17,73 +15,110 @@
...
-  VAST_TRACE("replace step adds batch");
-  // Get the target field if it exists.
-  const auto& layout_rt = caf::get<record_type>(layout);
-  auto column_index = layout_rt.resolve_key(config_.field);
-  if (!column_index) {
-    transformed_.emplace_back(layout, std::move(batch));
-    return caf::none;
+namespace {
+
+/// The configuration of a project transform step.
```
mavam: `project`?

mavam (approved):
Looks good! CI will sanity-check this.

I have some small cosmetic comments only.

libvast/src/transform_steps/drop.cpp
dominiklohmann: It's a precondition of the `transform_columns` function necessarily.

libvast/src/transform_steps/replace.cpp
dominiklohmann: Already fixed in a later commit.

PR #2231, opened 2022-04-24
Write new docs landing page

mavam:
This PR adds the first page of the new VAST docs.

Give it a read.


docs/docs/get-started/README.md
```diff
@@ -1,3 +1,106 @@
...
+1. [Setup VAST](/docs/setup-vast) describes how you can download, install, and
+   configure VAST in a variety of environments.
+   👉 *Start here if you want deploy VAST.*
+2. [Use VAST](/docs/use-vast) explains how to work with VAST, e.g., ingesting
+   data, running queries, matching threat intelligence, or integrating it with
+   other security tools.
+   👉 *Go here if you have a running VAST, and want to explore what you can do
+   with it.*
+3. [Understand VAST](/docs/understand-vast) describes the system design goals
+   and architecture, e.g., the actor model as concurrency and distribution
```
dominiklohmann, line 24: I think the actor model belongs into the _Develop VAST_ section. While we can mention that VAST uses the actor model and that the major system components comprise of at least one actor each, we should leave it at that level and only describe the major components here. Otherwise we risk losing readers early on in a section that actually is interesting to them.

When first mentioning the actor model, we can point to a page in the _Develop VAST_ section that goes in-depth, but is also appropriately technical.

mavam: I would disagree here. The actor model is a central part of the system architecture, and this is what this section is about. It's about learning VAST internals, what components exist, how the pieces hold together, etc.

CAF specifics go in *Develop VAST*, but not the high-level actor model design.

docs/docs/get-started/README.md
```diff
@@ -1,3 +1,106 @@
...
+:::tip Wanna give it a try?
+The [quick start guide](/docs/get-started/quick-start-guide) shows you how to
+get up and running and gives a quick tour of the key use cases.
+:::
+
+The documentation has the following key sections:
+
+1. [Setup VAST](/docs/setup-vast) describes how you can download, install, and
+   configure VAST in a variety of environments.
+   👉 *Start here if you want deploy VAST.*
```
dispanser, line 17: want _to_ deploy?

docs/docs/get-started/README.md
```diff
@@ -1,3 +1,106 @@
...
+
+We compare VAST to SIEM and data warehouses next, but skip a comparison with
+metrics services that we deem out of scope.
+
+### VAST vs. SIEM
+
+VAST *complements* a [SIEM][siem] nicely with the following use cases:
+
+- **Offloading**: route the high-volume telemetry to VAST that would otherwise
+  overload your SIEM or be cost-prohibitive or to ingest. By keeping the bulk of
```
dispanser, line 52: ~or~

docs/docs/get-started/README.md
```diff
@@ -1,3 +1,106 @@
...
+  the data in VAST, you remove bottlenecks and can selectively forward the
+  activity that matters to your SIEM.
+
+- **Compliance**: VAST has fine-grained retention span configuration to meet
+  GDPR and other regulatory requirements. When storage capacity needs careful
+  management, VAST's *compaction* feature allows for weighted ageing of your
+  data, so that you can specify relative importance of event types. Powerful
+  *transforms* allow you to anonymize, pseudonomyize, or encrypt specific
+  fields—either to sanitize PII data on import, or ad-hoc on export when data
```
dispanser, line 61: If we have links to siem and arrow, we might want to have one for the other acronyms as well, depending on the target audience PII may not be immediately obvoius.

dispanser (approved):
I think it reads great, minor spelling / grammar comments that are probably me being wrong, not actual mistakes ;-).

We might want to link to other acronyms as well (PII, COTS, ...) to cater to readers with a background not specificially security-related.

PR #2232, opened 2022-04-25
Migrate docs: Setup VAST

mavam:
This PR migrates and updates the installation instructions.

Read the new text entirely. I've made a lot of modifications.


docs/docs/setup-vast/install/linux.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Distribution Support
+
+:::tip Community contributions wanted!
+We are striving to bring VAST into the package managers of all major Linux
+distributions. Unfortunately we can do so only at a best-effort basis, but
+we much appreciate community contributions.
+:::
+
+### Debian
```
mavam, line 103: @satta I've lost track about the current issues we face regarding Debian packaging. Mind providing a brief update?

docs/docs/setup-vast/install/macos.md
```diff
@@ -0,0 +1,16 @@
+# macOS
```
mavam, line 1: @dominiklohmann should we add Homebrew instructions here as well?

docs/docs/setup-vast/install/linux.md
```diff
@@ -0,0 +1,116 @@
...
+you might need to change the location of the `vast` binary and configuration
+file.
+
+```config
+ExecStart=/path/to/vast --config=/path/to/vast.yaml start
+```
+
+In case you plan to run a PCAP source directly inside the VAST server process
+via `vast spawn source pcap`, you need to make sure that the VAST process gets
+the required privileges to listen on the network interfaces.
```
mavam, line 74: This feels a bit too specific, and since the CLI is changing in the future, we probably should rewrite this.

docs/docs/setup-vast/install/docker.md
```diff
@@ -0,0 +1,42 @@
...
+privileges. Database contents go into the volume exposed at `/var/lib/vast`.
+
+## Start the container
+
+Start VAST in a container and detach it to the background.
+
+```bash
+mkdir -p /var/lib/vast
+docker run -dt --name=vast --rm -p 42000:42000 -v /var/lib/vast:/var/lib/vast \
+  tenzir/vast:latest start
```
mavam, line 20: This feels a bit excessive. Isn't there a shorter way?

docs/docs/setup-vast/install/docker.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+## Build the main image
+
+Build the `tenzir/vast` image as follows:
+
+```bash
+docker build -t tenzir/vast:<TAG>
```
mavam, line 28: @dominiklohmann is `TAG` optional?

docs/docs/setup-vast/install/macos.md
dominiklohmann: That no longer exists. It's just building from source now. Installing via CMake also installs a launchd config file that can be enabled via `launchctl load -w ~/Library/LaunchAgents/com.tenzir.vast.plist`, but that's entirely optional.

docs/docs/setup-vast/install/docker.md
dominiklohmann: Not really, no.

- `-d` for detaching, i.e., running in background
- `-t` for terminal output
- `--name` to name the image
- `--rm` to clear older volumes
- `-p` to expose the port to the outer world
- `-v` to mount the database directory

docs/docs/setup-vast/install/docker.md
dominiklohmann: It defaults to `latest`. We upload this image though, not sure why we're providing build instructions here. The user should only be interested in _pulling_ the image.

docs/docs/setup-vast/install/linux.md
```diff
@@ -0,0 +1,116 @@
+# Linux
+
+Use our [pre-built packages](#download) below or [build VAST from
+source](/docs/setup-vast/build-from-source) to install VAST on any Linux
+distribution. To deploy VAST as system service, you can use our [systemd
+configuration](#systemd).
+
+To package VAST...
```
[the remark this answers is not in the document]
dominiklohmann: If `CPack` is available on the system you're building on you automatically get a `package` build target that packages VAST. Because this does not currently include systemd/launchd/rc.d configuration I don't think it's worth mentioning here (yet).

docs/docs/setup-vast/install/macos.md
mavam: Can you provide me snippet/diff that I can apply to this branch?

dominiklohmann: I'd just append one paragraph at the bottom:

> ## launchd
> 
> Installing VAST via CMake on macOS installs a [launchd agent](https://www.launchd.info) at `~/Library/LaunchAgents/com.tenzir.vast.plist`. To run VAST automatically on login, run this command:
> ```bash
> # NOTE: Use 'unload' rather than 'load' to unload the agent
> launchctl load -w ~/Library/LaunchAgents/com.tenzir.vast.plist
> ```

docs/docs/setup-vast/install/docker.md
mavam: Thanks, I added this explanation to the instructions.

docs/docs/setup-vast/install/linux.md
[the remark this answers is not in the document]
mavam: Okay.

docs/docs/setup-vast/install/linux.md
satta: Yes. From my point of view I am currently seeing the following roadblocks for providing up-to-date packages for VAST:

* Lack of packaging for VAST dependencies in the required versions, most importantly Apache Arrow. See this [Request For Package (ITP)](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=970021). A preliminary package [exists](https://salsa.debian.org/science-team/arrow) but needs stable maintainers who are willing and able to follow upstream's fast development process and ensure that new breaking releases are dealt with in a Debian policy-compliant way.
* (Minor issue) Lack of support for some (most?) non-Intel architectures that Debian supports. Not a dealbreaker but a nice-to-have.
* Missing ABI stability in VAST to provide a proper shared library package plus development headers package. Each ABI break would mean a new ABI version bump -- this essentially means that each ABI change would cause the new package to be treated like a completely new and unreviewed package that has to be manually cleared. Moreover one would need to ensure that the VAST version always matches the libvast version since IIRC there is no backward compatibility.

dominiklohmann: The third bullet would essentially require us to define a stable subset of libvast that we offer as public API with a stable ABI, but that adds impediments that no sane startup wants to deal with if they don't need to. I don't think there'll be an agreement between us and the Debian way of doing things in this regard for years to come.

We should probably just erase the section here from the docs page.

docs/docs/setup-vast/install/linux.md
dominiklohmann: Another way to set this up is to set `vast.start.commmands: [import pcap]`, which is a lot more intuitive and probably what we should be recommending instead.

Thinking of it further, it'd actually be nice to have some `vast remote-exec` command for simply executing a command in the context of a remote VAST instance. That could supersede `spawn source` a lot more generically, and would probably only take a few hours to gobble up. :thinking:

docs/docs/setup-vast/install/macos.md
```diff
@@ -0,0 +1,28 @@
...
+export CC="$(brew --prefix llvm)/bin/clang"
+export CXX="$(brew --prefix llvm)/bin/clang++"
+export LDFLAGS="-Wl,-rpath,$(brew --prefix llvm) ${LDFLAGS}"
+export CPPFLAGS="-isystem $(brew --prefix llvm)/include ${CPPFLAGS}"
+export CXXFLAGS="-isystem $(brew --prefix llvm)/include/c++/v1 ${CXXFLAGS}"
+```
+
+ ## launchd
+ 
+ Installing VAST via CMake on macOS copies a [launchd
```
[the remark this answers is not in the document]
mavam: I s/copies/configures/ for you. 🌻

docs/docs/setup-vast/install/linux.md
mavam: > Another way to set this up is to set `vast.start.commmands: [import pcap]`, which is a lot more intuitive and probably what we should be recommending instead.

Yes, that's also possible. My point was rather to not mention start commands at all, because my hunch is they will be much less relevant after the Event Router.

> Thinking of it further, it'd actually be nice to have some `vast remote-exec` command for simply executing a command in the context of a remote VAST instance.

That's probably the no-longer-existent `vast send` command, which basically allowed for sending messages to component actors from via the CLI.

docs/docs/setup-vast/install/docker.md
mavam: I've revamped the Docker docs. Mind giving it a read?

docs/docs/setup-vast/install/linux.md
mavam: Thanks for laying it out, @satta. I think (1) and (2) are possible to overcome.

(3) is complicated. ABI stability for libvast is not feasible for us, as @dominiklohmann pointed out. 

That said, our *plugin API* is a stable point of reference. I'm not sure if Debian requires true ABI stability or whether a subset suffices. In any case, our main dependency Arrow is also moving fast, without ABI guarantees, so the problem also applies to (1).

If ABI stability is not negotiable, we probably have to create our own repository where we simply serve DEBs for libvast, heeaders, and plugins. That said, before considering "our own", we probably would try to find out if [OBS](https://openbuildservice.org/) suits the bill.

docs/docs/setup-vast/install/linux.md
dominiklohmann: It's actually slightly different from the `vast send` command in that it does not send a message to an actor, but rather only communicates via commands with the node, which is much safer. I reckon this would be a nice hackathon project, maybe for our next one. :)

Anyways, we can also erase the section. I'd rather not document it than mention `vast spawn`, because that command has absolutely no test coverage.

docs/docs/get-started/README.md
```diff
@@ -10,7 +10,26 @@ The [quick start guide](/docs/get-started/quick-start-guide) shows you how to
...
+  use(Use):::action
+  learn(Learn):::action
+  develop(Develop):::action
+  %% Edges
+  setup --> use --> learn --> develop
+  %% Links
+  click setup "/vast/docs/setup-vast/" "Setup VAST"
+  click use "/vast/docs/use-vast"/ "Use VAST"
+  click learn "/vast/docs/learn-vast"/ "Learn VAST"
+  click develop "/vast/docs/develop-vast"/ "Develop VAST"
```
dominiklohmann, line 29:
```suggestion
  click use "/vast/docs/use-vast/" "Use VAST"
  click learn "/vast/docs/learn-vast/" "Learn VAST"
  click develop "/vast/docs/develop-vast/" "Develop VAST"
```

mavam: Jeez, your eyes!

docs/docs/setup-vast/tune.md
```diff
@@ -0,0 +1,220 @@
...
+The amount of memory that a VAST server process is allowed to use can currently
+not be configured directly as a configuration file option. Instead of such a
+direct tuning knob, the memory usage can be influenced through the configuration
+of the caching, catalog and disk monitor features.
+
+### Caching
+
+VAST groups table slices with the same schema in a *partition*. When building a
+partition, the parameter `max-partition-size` sets an upper bound on the number
+of records in a partition, across all table slices.
```
dominiklohmann, line 62: This should probably also mention `active-partition-timeout`, which can override this upper bound.

docs/docs/setup-vast/tune.md
```diff
@@ -0,0 +1,219 @@
...
+`vast.client-log-file` to enable logging. Note that relative paths are
+interpreted relative to the current working directory of the client process.
+
+Server log files rotate automatically after 10 MiB. The option
+`vast.disable-log-rotation` allows for disabling log rotation entirely, and the
+option `vast.log-rotation-threshold` sets the size limit when a log file should
+be rotated.
+
+VAST processes log messages in a dedicated thread, which buffers up to 1M
+messages by default. The option `vast.log-queue-size` controls this setting.
```
dominiklohmann, line 219: This default only applies to the server process. For client processes, it is 100.

docs/docs/setup-vast/tune.md
```diff
@@ -0,0 +1,219 @@
...
+the memory requirement. Increasing the `max-partition-size` from the default 1
+Mib to 8 Mib reduces the sketch factor to `0.016` for our sample data, if we
+reduce the `max-resident-partitions` to `2` for a fair comparison, we get the
+the following new composition:
+
+```
+11 * 1 Gib + 3 * 2.608 Gib + 0.016 * 2000 Gib
+11 Gib + 7.2 Gib + 32 Gib
+50.2 Gib
+```
```
dominiklohmann, line 179: Is there a way to render these calculations nicer?

Also, it probably should be a capital B for Bytes rather than b for bits throughout the calculations.

I also wonder whether these calculations are still correct given that we now have multiple active partitions (@tobim?).

docs/docs/setup-vast/tune.md
```diff
@@ -0,0 +1,219 @@
...
+        fp-rate: 0.1
+```
+
+This configuration includes two rules (= two sketches), where the first rule
+includes a field extractor and the second a type extractor. The first rule
+applies to a single field, `suricata.http.http.url`, and has false-positive rate
+of 0.5%. The second rule creates one sketch for all fields of type `addr` that
+has a false-positive rate of 10%.
+
+### Approximate memory requirements
```
dominiklohmann, line 127: I find that this section distracts from tuning. It's much too complicated for most readers, so maybe it should go into some separate section with advanced tutorials or so. It also appears to be out of date already.

docs/docs/setup-vast/tune.md
```diff
@@ -0,0 +1,219 @@
...
+
+A rough formula for estimating the memory requirements takes the configuration,
+the input data rate, and the amount of unique IP address and string values in
+the input data stream into account.
+
+```
+memory_est = (segments + 1) * max-segment-size + avg-partition-mem-size *
+(max-resident-partitions + 1) + sketch-factor * max-num-partitions
```
dominiklohmann, line 135: Likely outdated as this still refers to the archive segments.

docs/docs/setup-vast/download.md
```diff
@@ -0,0 +1,70 @@
...
+
+# Download
+
+You can obtain various artifacts of VAST along the CD pipeline.
+
+## Source Code
+
+Get the source code by cloning our git repository or downloading an archive.
+
+### git
```
dominiklohmann, line 13:
```suggestion
Get the source code by cloning our Git repository or downloading an archive.

### Git
```

The official spelling is with an uppercase G.

docs/docs/setup-vast/download.md
```diff
@@ -0,0 +1,70 @@
...
+
+We offer pre-built versions of VAST containing a statically linked binary, for the
+[latest release][latest-release] and the current development version.
+
+<div align="center" class="padding-bottom--md">
+  <a class="button button--md button--primary" href="https://github.com/tenzir/vast/releases/latest/download/vast-linux-static.tar.gz">Static Build (Release)</a>
+</div>
+
+For development versions or a specific git commit, you need to navigate through
+the github CI to find the build artifact:
```
dominiklohmann, line 48:
```suggestion
For development versions or a specific Git commit, you need to navigate through
the GitHub CI to find the build artifact:
```

docs/docs/setup-vast/download.md
```diff
@@ -0,0 +1,70 @@
...
+
+1. Browse to the [VAST static workflow][vast-workflow]
+2. Click on the latest run, e.g., `Merge pull request...`
+3. Scroll to the end of the page
+4. Click on the artifact filename, e.g.,
+   `vast-v1.0.0-101-g6e7a4ef1a4-linux-static.tar.gz`
+
+[vast-workflow]: https://github.com/tenzir/vast/actions?query=branch%3Amaster+workflow%3A%22VAST%22
```
dominiklohmann, line 56: I don't think this section adds much value, and would rather see it removed. It complicates an otherwise straightforward section.

docs/docs/setup-vast/download.md
```diff
@@ -0,0 +1,70 @@
...
+## Images
+
+Our CI builds Docker images for the latest release and the current development
+version.
+
+### Docker
+
+You can download pre-built Docker images from
+[DockerHub](https://hub.docker.com/repository/docker/tenzir/vast).
```
dominiklohmann, line 65:
```suggestion
[Docker Hub](https://hub.docker.com/repository/docker/tenzir/vast).
```

Official spelling uses two words.

docs/docs/setup-vast/install/linux.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Configure the unit
+
+Before you begin, find the line beginning with `ExecStart=` at the very bottom
+of the `[Service]` section in the unit file. Depending on your installation path
+you might need to change the location of the `vast` binary and configuration
+file.
+
+```config
+ExecStart=/path/to/vast --config=/path/to/vast.yaml start
```
dominiklohmann, line 46: Why not just recommend using the default directory, e.g., `$PREFIX/etc/vast/vast.yaml`? Then it wouldn't need to be specified here. We also don't specify it in the example in the repository.

dominiklohmann (approved):
Just some minor requests, this looks really good. Approving already assuming you agree with all my comments.

docs/docs/setup-vast/tune.md
mavam: I've removed it. We'll bring it back after it's fixed. (I created an internal story for it.)

docs/docs/setup-vast/download.md
mavam: Disagreeing on this one, I needed this more than once in the past.

dominiklohmann: But you're a developer, not a user.

docs/docs/setup-vast/install/linux.md
mavam: I don't know. @lava or @tobim might know.

(The content is unadapted from the existing docs.)

docs/docs/setup-vast/tune.md
mavam: Good call.

mavam: Added a sentence.

docs/docs/setup-vast/install/linux.md
dominiklohmann: Note that duplicating the config from `$PREFIX/etc/vast/vast.yaml` with `--config` is not side-effect free per sc-32665; we should definitely not be recommending doing so, and if you want to see the used config files that already is available in `vast status | jq .system`.

docs/docs/setup-vast/download.md
mavam: This is not about developers. This is about the user that wants to be running the bleeding-edge version.

mavam: It's _really_ handy to throw someone a static build, especially after we pushed a fix on master.

mavam: But we're not recommending it. We're giving the user the option. Nobody goes through these hoops unless they have a specific reason.

docs/docs/setup-vast/install/linux.md
mavam: I removed the contested `--config` option. If there is a strong opinion to keep it, I would create another internal story.

PR #2238, opened 2022-04-26
Adjust count when skipping partitions

lava:
Adjust count of outstanding queries when skipping some.

Also contains

### :memo: Checklist

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -599,7 +599,7 @@ void index_state::schedule_lookups() {
     for (auto qid : next->queries) {
       auto it = pending_queries.queries().find(qid);
       if (it == pending_queries.queries().end()) {
-        VAST_WARN("{} tried to access non-existant query {}", *self, qid);
+        VAST_WARN("{} tried to access non-existent query {}", *self, qid);
         *cnt -= 1;
         if (*cnt == 0)
           --running_partition_lookups;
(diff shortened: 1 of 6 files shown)


libvast/src/system/index.cpp
```diff
@@ -600,6 +600,9 @@ void index_state::schedule_lookups() {
       auto it = pending_queries.queries().find(qid);
       if (it == pending_queries.queries().end()) {
         VAST_WARN("{} tried to access non-existant query {}", *self, qid);
```
dominiklohmann, line 602: Spelling: It's existent with an e.

libvast/src/system/index.cpp
```diff
@@ -600,6 +600,9 @@ void index_state::schedule_lookups() {
...
+        *cnt -= 1;
+        if (*cnt == 0)
+          --running_partition_lookups;
```
dominiklohmann, line 605: This snippet is now duplicated between here and `handle_completion` directly below. Can we put it into a lambda to give it a name?

E.g.,

```cpp
auto finish_query = [&] {
  if (--*cnt > 0)
    return false;
  --running_partition_lookups;
  return true;
};
```

dominiklohmann (approved):
The logic of this checks looks correct to me from trying to follow the flow of the function.

libvast/src/system/index.cpp
tobim: You need to be careful with that. We call `handle_completion` when we get the response from the partition. If we create this helper closure outside of it the reference to `cnt` won't be valid later on. You would have to capture `cnt` by value and `finish_query` also by value in `handle_completion`.

PR #2191, opened 2022-04-05
Potential bugfixes for hanging index workers

lava:
Some VAST fixes for issues that have been identified while looking at an observed issue where all index workers had been hanging in "busy" state for days, waiting for some passive partition that for some reason never returned a result.

- Update log message to include actor
- Add timeout in query supervisor
- Clear in-progress map when aborting query

### :memo: Checklist

- [ ] All user-facing changes have changelog entries.
- [ ] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [ ] The PR description contains instructions for the reviewer, if necessary.

PR #2239, opened 2022-04-26
Handle indefinite hang when evaluating queries on irrelevant partitions

lava:
- Fix direct value return after response promise has been created
- Add timeout for scheduler lookups

### :memo: Checklist

### DIFF
--- a/libvast/include/vast/defaults.hpp
+++ b/libvast/include/vast/defaults.hpp
@@ -245,6 +245,10 @@ constexpr std::chrono::milliseconds signal_monitoring_interval
 constexpr std::chrono::milliseconds initial_request_timeout
   = std::chrono::seconds{10};
 
+/// Timeout for the scheduler to give up on a partition.
+constexpr std::chrono::milliseconds scheduler_timeout
+  = std::chrono::minutes{15};
+
 /// The period to wait until a shutdown sequence finishes cleanly. After the
 /// elapses, the shutdown procedure escalates into a "hard kill".
 /// @relates shutdown_kill_timeout
--- a/libvast/src/system/passive_partition.cpp
+++ b/libvast/src/system/passive_partition.cpp
@@ -348,8 +348,10 @@ partition_actor::behavior_type passive_partition(
       }
       auto start = std::chrono::steady_clock::now();
       auto triples = detail::evaluate(self->state, query.expr);
-      if (triples.empty())
-        return uint64_t{0};
+      if (triples.empty()) {
+        rp.deliver(uint64_t{0});
+        return rp;
+      }
       auto eval = self->spawn(evaluator, query.expr, triples);
       self->request(eval, caf::infinite, atom::run_v)
         .then(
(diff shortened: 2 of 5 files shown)


libvast/include/vast/defaults.hpp
```diff
@@ -245,6 +245,9 @@ constexpr std::chrono::milliseconds signal_monitoring_interval
 constexpr std::chrono::milliseconds initial_request_timeout
   = std::chrono::seconds{10};
 
+/// Timeout for the scheduler to give up on a partition.
+constexpr std::chrono::milliseconds scheduler_timeout = std::chrono::hours{1};
```
tobim, line 249: Isn't that extremely long for a single partition lookup? How about one minute?

libvast/src/system/passive_partition.cpp
```diff
@@ -348,8 +348,10 @@ partition_actor::behavior_type passive_partition(
       }
       auto start = std::chrono::steady_clock::now();
       auto triples = detail::evaluate(self->state, query.expr);
-      if (triples.empty())
-        return uint64_t{0};
+      if (triples.empty()) {
+        rp.deliver(0);
+        return rp;
+      }
```
tobim, line 354: Yes this looks sus. But I would really like to have a test for it to be sure.

tobim (approved):
The changes make sense, but consider my comments before merging.

libvast/include/vast/defaults.hpp
lava: I'm not sure what the correct value is, I went high because it was "infinite" before and I didn't want to change it too much. 1 minute feels too short, if we're running on a machine with a slow filesystem and high load I think it can be more than that. Maybe 15 minutes is a good starting point.

tobim: Also good, let's make it 51.

dominiklohmann: We need to make sure this does not cause shutdown to no longer function similar to what we had in https://github.com/tenzir/vast/pull/2191#discussion_r845153665

libvast/src/system/passive_partition.cpp
dominiklohmann: CAF 0.18 actually asserts in the destructor of a response promise that it was delivered exactly once, and the semantic of that didn't change. We were likely just in UB land here.

tobim: Oh and you'll need to call deliver with `deliver(uint64_t{0})`.

dominiklohmann: Shouldn't be necessary since it's a typed response promise, no?

lava: I added a test and verified that it times out without this change.

tobim:
I independently verified that this is a real bug by modifying the CAF promises.cpp example to return after creating a promise. This indeed leads to a deadlock.

PR #2240, opened 2022-04-26
Fix `vast.active-partition-timeout` option

dominiklohmann:
A recent refactoring caused us to no longer read this option correctly, always using the default. This simplifies and fixes the option.

### :memo: Checklist

Double-check that the option now works correctly.

This should not need a changelog entry since the bug was newly introduced.

### DIFF
--- a/libvast/include/vast/defaults.hpp
+++ b/libvast/include/vast/defaults.hpp
@@ -210,7 +210,7 @@ constexpr size_t disk_monitor_step_size = 1;
 constexpr size_t max_partition_size = 1'048'576; // 1_Mi
 
 /// Timeout after which an active partition is forcibly flushed.
-constexpr std::string_view active_partition_timeout = "1 hour";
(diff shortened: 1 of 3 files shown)


tobim (approved):
I don't understand why it didn't work before but at least this is a small simplification.

dominiklohmann:
> I don't understand why it didn't work before but at least this is a small simplification.

@mavam's little environment variable PR did a refactoring that caused this to break only when specifying it in the configuration file, because it was intermittently put into a `caf::config_option`, which has a `caf::timespan` alternative. Before that change, we consistently stored this as a `std::string`, because that is what we inferred from the YAML to `data` translation. The actual failed than happened at a later point in time where we called `caf::get_or` with a string for this option—which just silently returned the default value since the stored config option was of the wrong alternative.

tobim:
Thanks for the detailed explanation, it makes sense now.

PR #2241, opened 2022-04-27
Prepare repository for VAST v2.0.0-rc3

dominiklohmann:
All good things come in threes.


tobim (approved):
I confirmed that every relevant version string has been updated.

PR #2242, opened 2022-04-27
Add changelog to new docs page

dominiklohmann:
Small evening fun adventure checking whether we can easily add an always up-to-date changelog to the new docs website.

### :memo: Checklist

n/t


mavam:
Nice addition.

You could add it to the top-level navbar when [adding it as page](https://docusaurus.io/docs/creating-pages#add-a-markdown-page). I think this might be more exciting.

.github/workflows/docs.yaml
```diff
@@ -4,50 +4,71 @@ on:
...
+            lsb-release \
+            ninja-build \
+            pkg-config \
+            python3-dev \
+            python3-pip \
+            python3-venv \
+            wget \
+            yarnpkg
+          python3 -m pip install --upgrade pip
+          python3 -m pip install --upgrade cmake
```
mavam, line 45: This feels pretty heavy-weight...why?

.github/workflows/docs.yaml
```diff
@@ -4,50 +4,71 @@ on:
...
-        with:
-          node-version: 16.x
-          cache-dependency-path: docs/yarn.lock
-          cache: yarn
-      - name: Install dependencies
-        working-directory: docs
-        run: yarn install --frozen-lockfile
-      - name: Test build website
-        working-directory: docs
-        run: yarn build
```
mavam, line 46: I find this way more maintainable than going into CMake land. Is there no way around doing this with `yarn`?

dominiklohmann: Nope. We amalgamate the changelog using CMake as part of the build already, so unless we wanna exclude the files from the last release by using the copy in the source tree root that we create with every release we'll have to go through it.

I don't think maintainability needs to be a concern here, it's pretty straightforward if you ignore the yet to be simplified installation of dependencies above.

.github/workflows/docs.yaml
dominiklohmann: This is just copy-pasted from another workflow, we actually only need some subset of this but I wanted to get this going first. I'll simplify this before asking for a review.

dominiklohmann:
> You could add it to the top-level navbar when [adding it as page](https://docusaurus.io/docs/creating-pages#add-a-markdown-page). I think this might be more exciting.

You are right, this looks much better. Still needs some tweaks in order to unbreak the local build, though.

image

.github/workflows/docs.yaml
dominiklohmann: Take a look now, it's actually much simpler now that we re-use the output of the existing changelog job.

.github/workflows/vast.yaml
```diff
@@ -110,7 +110,13 @@ jobs:
       - name: Generate CHANGELOG.md
         run: |
           cmake --build build --target changelog
+      - name: Upload CHNAGELOG.md
```
mavam, line 113:
```suggestion
      - name: Upload CHANGELOG.md
```

docs/package.json
```diff
@@ -4,11 +4,11 @@
   "private": true,
   "scripts": {
     "docusaurus": "docusaurus",
-    "start": "docusaurus start",
-    "build": "docusaurus build",
+    "start": "cp -f ../CHANGELOG.md src/pages/changelog.md && docusaurus start",
```
mavam, line 7: Cleaner would be probably to integrate the existing CHANGELOG.md with some JSX. But this suits the bill just fine.

mavam (approved):
Really nice result!

docs/package.json
dominiklohmann: Not possible since it's out-of-tree. I tried this first. Second attempt was a repository-internal symlink, but that's broken in the webpack version that docusaurus currently requires.

mavam:
The top-level link to the unreleased section is broken and points: https://github.com/tenzir/vast/commits/master/changelog/unreleased.

PR #2246, opened 2022-04-28
Fix incorrect handling of duration in v1 -> v2 arrow conversion

dispanser:
Our table slice conversion code for `arrow::v1` -> `arrow::v2` conversion did not specifically handle duration data. While the underlying phsysical representation is actually the same (`int64`), the types don't align and an assertion is triggered in the process.

We add a separate handler for duration that doesn't fail, and add a test based on some hand-crafted arrow-v1 table slice.

### :memo: Checklist

Simple enough, one commit.


libvast/src/arrow_table_slice.cpp
```diff
@@ -1251,6 +1251,10 @@ upgrade_array_to_v2(const std::shared_ptr<arrow::Array>& arr, const type& t) {
       VAST_ASSERT(arr->type()->Equals(t.to_arrow_type()));
       return arr;
     },
+    [&](const duration_type&) -> std::shared_ptr<arrow::Array> {
+      // duration is backed by physical int64 array, but logical type differs
+      return arr;
```
dominiklohmann, line 1256: You're deep in undefined behavior land here with pointer aliasing rules. I think the correct way to handle this is to create a new Array from the existing Array's underlying ArrayData.

```suggestion
      // Duration is backed by physical Int64 array, but logical type differs.
      auto result = arr->View(duration_type::to_arrow_type());
      VAST_ASSERT(result.ok(), result.status().ToString().c_str());
      return result.MoveValueUnsafe();
```

libvast/test/arrow_table_slice.cpp
```diff
@@ -1050,6 +1050,18 @@ TEST(read_legacy_table_slice) {
   check_column(legacy_slice, 14, pattern_type{}, f3_pattern); // f11_2_2
 }
 
+TEST(convert_legacy_table_slice_all_types) {
+  auto bytes = unbox(vast::io::read(VAST_TEST_PATH "artifacts/table_slices/"
+                                                   "arrow-v1_all-types.bytes"));
+  auto legacy_slice
+    = table_slice{chunk::make(std::move(bytes)), table_slice::verify::yes};
+  fmt::print(stderr, "tp;was here {}, {}, {}\n", legacy_slice.columns(),
+             legacy_slice.rows(), bytes.size());
```
dominiklohmann, line 1059: Leftover debug message.
```suggestion
```

libvast/test/arrow_table_slice.cpp
```diff
@@ -1050,6 +1050,18 @@ TEST(read_legacy_table_slice) {
...
+  const auto& rb = to_record_batch(legacy_slice);
```
dominiklohmann, line 1060: This record batch appears to be unused.

dominiklohmann (approved):
This is a really nice catch, I'm glad you found this _before_ we released the v2.0. I have some minor comments on the PR.

Thanks especially for the extensive test, IIUC that now covers all types.

libvast/test/arrow_table_slice.cpp
dispanser: thanks, must be used to create a new table slice. added a commit.

PR #2253, opened 2022-05-02
Implement YAML Types

6yozo:
Implement parser and unit tests for YAML types and Record Algebra.

The YAML Type parser is an internal component, not directly visible to the user; a later PR will integrate it into VAST.

### :memo: Checklist

### DIFF
--- a/libvast/test/module.cpp
+++ b/libvast/test/module.cpp
@@ -261,12 +261,10 @@ caf::expected<type> to_type(const std::vector<type>& known_types,
       name_clash_specifier_cnt++;
       // right is the new record type
       merge_conflict_handling = record_type::merge_conflict::prefer_left;
-      VAST_TRACE(fmt::format("Prefer left: {}", merge_conflict_handling));
     }
     if (is_extend_found) {
       name_clash_specifier_cnt++;
       merge_conflict_handling = record_type::merge_conflict::prefer_right;
-      VAST_TRACE(fmt::format("Prefer right: {}", merge_conflict_handling));
     }
     if (name_clash_specifier_cnt >= 2)
       return caf::make_error(ec::parse_error,
--- a/libvast/test/module.cpp
+++ b/libvast/test/module.cpp
@@ -1,0 +1,710 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2022 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#define SUITE module
+
+#include "vast/module.hpp"
+
+#include "vast/aliases.hpp"
+#include "vast/concept/convertible/data.hpp"
+#include "vast/data.hpp"
+#include "vast/legacy_type.hpp"
+#include "vast/test/test.hpp"
+#include "vast/type.hpp"
+
+#include <caf/error.hpp>
+#include <caf/sum_type.hpp>
+#include <caf/test/dsl.hpp>
+#include <fmt/format.h>
+
+#include <optional>
+#include <string_view>
+
+using namespace vast;
+
+caf::expected<type>
+to_type(const std::vector<type>& known_types, const data& declaration,
+        std::string_view name = "");
+
+caf::expected<record_type>
+to_record(const std::vector<type>& known_types,
+          const std::vector<vast::data>& record_list) {
+  if (record_list.empty())
+    return caf::make_error(ec::parse_error, "parses an empty record");
+  auto record_fields = std::vector<record_type::field_view>{};
+  for (const auto& record_value : record_list) {
+    const auto* record_record_ptr = caf::get_if<record>(&record_value);
+    if (record_record_ptr == nullptr)
+      return caf::make_error(ec::parse_error, "parses an record with invalid "
+                                              "format");
+    const auto& record_record = *record_record_ptr;
+    if (record_record.size() != 1)
+      return caf::make_error(ec::parse_error, "parses an record field with an "
+                                              "invalid format");
+    auto type_or_error = to_type(known_types, record_record.begin()->second);
+    if (!type_or_error)
+      return type_or_error.error();
+    record_fields.emplace_back(record_record.begin()->first, *type_or_error);
+  }
+  return record_type{record_fields};
+}
+
+type get_known_type(const std::vector<type>& known_types,
+                    std::string_view name) {
+  for (const auto& known_type : known_types) {
+    if (name == known_type.name()) {
+      return known_type;
+    }
+  }
+  return {}; // none type
+}
+
+std::vector<std::string> reserved_names
+  = {"bool",   "integer", "count",   "real",   "duration", "time",
+     "string", "pattern", "address", "subnet", "type",     "attributes",
+     "enum",   "list",    "map",     "record", "base",     "implant",
+     "extend", "fields",  "key",     "value"};
+
+caf::expected<type> to_type(const std::vector<type>& known_types,
+                            const data& declaration, std::string_view name) {
+  const auto* known_type_name_ptr = caf::get_if<std::string>(&declaration);
+  if (std::any_of(reserved_names.begin(), reserved_names.end(),
+                  [&](const auto& reserved_name) {
+                    return name == reserved_name;
+                  }))
+    return caf::make_error(
+      ec::parse_error,
+      fmt::format("parses a new type with a reserved name: {}", name));
+  if (known_type_name_ptr != nullptr) {
+    const auto& known_type_name = *known_type_name_ptr;
+    // Check built-in types first
+    if (known_type_name == "bool")
+      return type{name, bool_type{}};
+    if (known_type_name == "integer")
+      return type{name, integer_type{}};
+    if (known_type_name == "count")
+      return type{name, count_type{}};
+    if (known_type_name == "real")
+      return type{name, real_type{}};
+    if (known_type_name == "duration")
+      return type{name, duration_type{}};
+    if (known_type_name == "time")
+      return type{name, time_type{}};
+    if (known_type_name == "string")
+      return type{name, string_type{}};
+    if (known_type_name == "pattern")
+      return type{name, pattern_type{}};
+    if (known_type_name == "address")
+      return type{name, address_type{}};
+    if (known_type_name == "subnet")
+      return type{name, subnet_type{}};
+    // Check type aliases aka. known types
+    const auto& known_type = get_known_type(known_types, known_type_name);
+    if (!known_type)
+      return caf::make_error(ec::parse_error, fmt::format("found unknown type: "
+                                                          "{}",
+                                                          known_type_name));
+    return type{name, known_type};
+  }
+  const auto* decl_ptr = caf::get_if<record>(&declaration);
+  if (decl_ptr == nullptr)
+    return caf::make_error(ec::parse_error, "parses type alias with invalid "
+                                            "format");
+  const auto& decl = *decl_ptr;
+  // Get the optional attributes
+  auto attributes = std::vector<type::attribute_view>{};
+  auto found_attributes = decl.find("attributes");
+  if (found_attributes != decl.end()) {
+    const auto* attribute_list = caf::get_if<list>(&found_attributes->second);
+    if (attribute_list == nullptr)
+      return caf::make_error(ec::parse_error, "parses an attribute list with "
+                                              "an invalid format");
+    for (const auto& attribute : *attribute_list) {
+      const auto* attribute_string_ptr = caf::get_if<std::string>(&attribute);
+      if (attribute_string_ptr != nullptr)
+        attributes.push_back({*attribute_string_ptr});
+      else {
+        const auto* attribute_record_ptr = caf::get_if<record>(&attribute);
+        if (attribute_record_ptr == nullptr)
+          return caf::make_error(ec::parse_error, "parses an attribute with an "
+                                                  "invalid format");
+        const auto& attribute_record = *attribute_record_ptr;
+        if (attribute_record.size() != 1)
+          return caf::make_error(ec::parse_error, "parses an attribute not "
+                                                  "having one and only one "
+                                                  "field");
+        const auto& attribute_key = attribute_record.begin()->first;
+        const auto* attribute_value_ptr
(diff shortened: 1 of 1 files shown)


libvast/test/module.cpp
```diff
@@ -0,0 +1,688 @@
...
+    }
+  }
+  return {}; // none type
+}
+
+std::vector<std::string> reserved_names
+  = {"bool",   "integer", "count",   "real",   "duration", "time",
+     "string", "pattern", "address", "subnet", "type",     "attributes",
+     "enum",   "list",    "map",     "record", "base",     "implant",
+     "extend", "fields",  "key",     "value"};
```
dominiklohmann, line 71: Why does everything from `base` to `value` need to be reserved?

libvast/test/module.cpp
```diff
@@ -0,0 +1,688 @@
...
+
+caf::expected<type> to_type(const std::vector<type>& known_types,
+                            const data& declaration, std::string_view name) {
```
dominiklohmann, line 74: The old parser checked whether the type name conformed to a given set of naming rules. Should we still enforce this or are we just more lenient now because the YAML grammar is more forgiving?

libvast/test/module.cpp
```diff
@@ -0,0 +1,688 @@
...
+  return {}; // none type
+}
+
+std::vector<std::string> reserved_names
+  = {"bool",   "integer", "count",   "real",   "duration", "time",
+     "string", "pattern", "address", "subnet", "type",     "attributes",
+     "enum",   "list",    "map",     "record", "base",     "implant",
+     "extend", "fields",  "key",     "value"};
+
+caf::expected<type> to_type(const std::vector<type>& known_types,
```
dominiklohmann, line 73: I was thinking about (1) where we want to put this and (2) how we should name this function.

I'm thinking that maybe `type.{hpp,cpp}` make sense for the location as free functions, or as a static member of type. Regarding the naming I personally like `caf::expected<type> type::from_declaration(...)`, and would like to hear @tobim's opinion on it.

dominiklohmann (commented):
Just took a first look. I think the functionality is almost complete. 

Can you add documentation to the code?

libvast/test/module.cpp
mavam: This is not so much about YAML. It's more so related to the query side, e.g., our extractor syntax.

dominiklohmann: On the extractor side we can accept everything if the user escapes it. We have no restrictions for it.

Note that this is about _type_ names, not about the names of _fields_ in record types. For fields we absolutely do not need to restrict anything, because they need to be able to match anything the user possibly wants to import.

mavam: Okay, makes sense. I think we can start as open as possible. I would perhaps avoid `.` and `:`, because these have special meanings in a full-qualified field and in YAML itself.

libvast/test/module.cpp
6yozo: Technically there is no need to restrict other names than the built-in types. So the list can be shortened to:
```
  = {"bool",    "integer", "count",  "real", "duration", "time", "string",
     "pattern", "address", "subnet", "enum", "list",     "map",  "record"};
```

libvast/test/module.cpp
6yozo: The `type suricata.component.common = record {...}` in the current schema file can be translated to the new YAML module syntax multiple ways:

- Option 1:
```yaml
module: suricata
types: 
  component:
    record:
    - common:
        record:
        - ...
```

- Option 2:
```yaml
module: suricata
types: 
  component:
    record:
    - common: common_record

  common_record:
    record:
    - ...
```

- Option 3:
```yaml
module: suricata
types: 
  component.common:
    record:
      ...
```
I think we are aiming for Option 1 or Option 2. Option 3 would require `.` in the type name.

The `:` may be allowed though if it gets through the YAML parser and gets into a type name, we may just use it:

```yaml
module: foo
types: 
  "app:proto":
    type: string
```


The new commit prevents both `.` and `:` in type names for now.

dominiklohmann: I still don't understand why either of `.` or `:` would cause an issue in type names and don't understand why we'd restrict them at all.

As far as my understanding goes handling escaping purely needs to be done on the extractor side because that receives a flat representation like `suricata.component.common.timestamp`, but the extraction logic already can disambiguate this because the type itself is not flattened in any way.

From the given examples I think we're aiming for option 3. Option 1 and 2 actually cannot work, because Suricata looks like this:

```json5
{
  "event_type": "flow",
  //...
}
```

Rather than like this, which would match the options 1 and 2:

```json5
{
  "common": {
    "event_type": "flow",
    // ...
  },
  // ...
}
```

libvast/test/module.cpp
dominiklohmann: This looks better now. Resolving this one!

libvast/test/module.cpp
6yozo: This clears the question. We must allow `.` in type names and there is no reason to restrict other characters.

I have left a comment in the code to reflect this and added a sentence to the YAML Modules doc.

The old parser used:

```
      = ("type" >> skp >> parsers::identifier >> skp >> '=' >> skp >> tp)
```

and `parsers::identifier` is:

```
constexpr inline auto identifier_char = (alnum | ch<'_'> | ch<'.'>);
constexpr inline auto identifier = +identifier_char;
```

The old schema parser allowed only alphanumeric characters the `_` and the `.`. The new parser will not impose these limitations.

libvast/test/module.cpp
```diff
@@ -0,0 +1,689 @@
...
+#include <fmt/format.h>
+
+#include <optional>
+#include <string_view>
+
+using namespace vast;
+
+caf::expected<type>
+to_type(const std::vector<type>& known_types, const data& declaration,
+        std::string_view name = "");
```
dominiklohmann, line 32: Please document this function.

dominiklohmann, line 32: How do we want to name this function?

In order to make this stand out against `type type::infer(const data&)` I suggest we name this `caf::expected<type> type::from_declaration(const data& declaration, ...)`.

libvast/test/module.cpp
```diff
@@ -0,0 +1,689 @@
...
+
+caf::expected<record_type>
+to_record(const std::vector<type>& known_types,
+          const std::vector<vast::data>& record_list) {
```
dominiklohmann, line 36:
```suggestion
caf::expected<record_type>
to_record(const std::vector<type>& known_types,
          const list& field_declarations) {
```

`list` is `std::vector<data>`, so please use that for clarity.

I think the name `field_declarations` makes sense for the argument if we're using "declaration" as the term for the relevant subsection of the YAML.

dominiklohmann: Please also do a general pass over the function and try to give appropriate variable names.

libvast/test/module.cpp
```diff
@@ -0,0 +1,689 @@
...
+  if (record_list.empty())
+    return caf::make_error(ec::parse_error, "parses an empty record");
+  auto record_fields = std::vector<record_type::field_view>{};
```
dominiklohmann, line 39:
```suggestion
  auto record_fields = std::vector<record_type::field_view>{};
  record_fields.reserve(field_declarations.size());
```

Always best practice to reserve memory if you already know the size it's gonna have.

libvast/test/module.cpp
```diff
@@ -0,0 +1,689 @@
...
+
+caf::expected<type>
+to_type(const std::vector<type>& known_types, const data& declaration,
+        std::string_view name = "");
+
+caf::expected<record_type>
+to_record(const std::vector<type>& known_types,
+          const std::vector<vast::data>& record_list) {
+  if (record_list.empty())
+    return caf::make_error(ec::parse_error, "parses an empty record");
```
dominiklohmann, line 38: In terms of error message, try to actually help the user:

> record types must have at least one field

Or similar.

libvast/test/module.cpp
```diff
@@ -0,0 +1,689 @@
...
+  auto record_fields = std::vector<record_type::field_view>{};
+  for (const auto& record_value : record_list) {
+    const auto* record_record_ptr = caf::get_if<record>(&record_value);
+    if (record_record_ptr == nullptr)
+      return caf::make_error(ec::parse_error, "parses an record with invalid "
+                                              "format");
```
dominiklohmann, line 44: > record type fields must be specified as a dictionary

libvast/test/module.cpp
```diff
@@ -0,0 +1,689 @@
...
+    const auto& record_record = *record_record_ptr;
+    if (record_record.size() != 1)
+      return caf::make_error(ec::parse_error, "parses an record field with an "
+                                              "invalid format");
```
dominiklohmann, line 48: > record type fields must be a dictionary with a single entry; got ...

libvast/test/module.cpp
```diff
@@ -0,0 +1,689 @@
...
+    auto type_or_error = to_type(known_types, record_record.begin()->second);
+    if (!type_or_error)
+      return type_or_error.error();
```
dominiklohmann, line 51:
```suggestion
      return caf::make_error(ec::parse_error,
        fmt::format("failed to parse record type field: {}", type_or_error.error()));
```

libvast/test/module.cpp
```diff
@@ -0,0 +1,689 @@
...
+    if (name == known_type.name()) {
+      return known_type;
+    }
+  }
+  return {}; // none type
+}
+
+std::vector<std::string> reserved_names
+  = {"bool",    "integer", "count",  "real", "duration", "time", "string",
+     "pattern", "address", "subnet", "enum", "list",     "map",  "record"};
```
dominiklohmann, line 69:
```suggestion
constexpr auto reserved_names
  = std::array{"bool",    "integer", "count",  "real", "duration", "time", "string",
     "pattern", "address", "subnet", "enum", "list",     "map",  "record"};
```

The list is known at compile time, so the size is fixed. Global variables should be immutable.

libvast/test/module.cpp
```diff
@@ -0,0 +1,689 @@
...
+                            const data& declaration, std::string_view name) {
+  const auto* known_type_name_ptr = caf::get_if<std::string>(&declaration);
+  // Prevent using reserved names as types names
+  if (std::any_of(reserved_names.begin(), reserved_names.end(),
+                  [&](const auto& reserved_name) {
+                    return name == reserved_name;
+                  }))
+    return caf::make_error(
+      ec::parse_error,
+      fmt::format("parses a new type with a reserved name: {}", name));
```
dominiklohmann, line 81: > type must not use reserved name: {}

libvast/test/module.cpp
```diff
@@ -0,0 +1,689 @@
...
+      return caf::make_error(ec::parse_error, fmt::format("found unknown type: "
+                                                          "{}",
+                                                          known_type_name));
+    return type{name, known_type};
+  }
+  const auto* decl_ptr = caf::get_if<record>(&declaration);
+  if (decl_ptr == nullptr)
+    return caf::make_error(ec::parse_error, "parses type alias with invalid "
+                                            "format");
+  const auto& decl = *decl_ptr;
```
dominiklohmann, line 119: Naming this `decl` with an argument named `declaration` is rather confusing.

Please use consistent names throughout the function so it makes more sense. Maybe it's `declaration_record` or so.

libvast/test/module.cpp
```diff
@@ -0,0 +1,689 @@
...
+  if (is_type_found)
+    type_selector_cnt++;
+  if (is_enum_found)
+    type_selector_cnt++;
+  if (is_list_found)
+    type_selector_cnt++;
+  if (is_map_found)
+    type_selector_cnt++;
+  if (is_record_found)
+    type_selector_cnt++;
```
dominiklohmann, line 175:
```suggestion
  int type_selector_cnt = is_type_found + is_enum_found + is_list_found + is_map_found + is_record_found;
```

Booleans convert to 0 or 1 in additions, which is a relict from C. Useful here, though, to check how many are true.

libvast/test/module.cpp
```diff
@@ -0,0 +1,689 @@
...
+  if (type_selector_cnt != 1)
+    return caf::make_error(ec::parse_error, "expects one of type, enum, list, "
+                                            "map, record");
+  // Type alias
+  if (is_type_found) {
+    auto type_expected = to_type(known_types, found_type->second);
+    if (!type_expected)
+      return type_expected.error();
```
dominiklohmann, line 183: This error message can be extended with local context: We're failing to parse an alias here with a nested error message.

libvast/test/module.cpp
```diff
@@ -0,0 +1,689 @@
...
+    auto enum_fields = std::vector<enumeration_type::field_view>{};
+    for (const auto& enum_value : enum_list) {
+      const auto* enum_string_ptr = caf::get_if<std::string>(&enum_value);
+      if (enum_string_ptr == nullptr)
+        return caf::make_error(ec::parse_error, "parses an enum value with an "
+                                                "invalid format");
+      enum_fields.push_back({*enum_string_ptr});
+    }
+    return type{name, enumeration_type{enum_fields}, std::move(attributes)};
+  }
```
dominiklohmann, line 204: Can you extract this to a separate helper function? Same goes for the other ones below.

Then please apply the above feedback on error messages and variables names as well and I think we're done.

libvast/test/module.cpp
dominiklohmann: This should probably come last; let's discuss this with @tobim.

dominiklohmann (commented):
Reviewed in-depth in a call. This is almost ready to go, just small bits now.

libvast/test/module.cpp
6yozo: I have created a [railroad diagram](https://media.app.shortcut.com/api/attachments/files/clubhouse-assets/59a979b5-168d-459a-ad4b-fe6d2b009b9e/6286118c-351e-4400-b00e-a909a44677e1/diagram.xhtml) that helps in reviewing variable names. Reviewing variable names makes more sense in the "yaml-modules" PR, though, where these names will become final.

dominiklohmann: That makes sense.

libvast/test/module.cpp
dominiklohmann: Let's do this as part of the YAML modules PR. Resolving for now.

dominiklohmann (approved):
Thanks for incorporating my review feedback. Let's merge this and then discuss YAML modules on top of this, and also where all these functions should go at the end. This is a big step towards YAML modules! 🚀

PR #2256, opened 2022-05-02
Deduplicate config files before loading

dominiklohmann:
Doesn't change anything about how things work, but effectively makes us not load files multiple times.

### :memo: Checklist

Don't think this needs a changelog entry since it's essentially only cleanup work.

### DIFF
--- a/vast/vast.cpp
+++ b/vast/vast.cpp
@@ -92,7 +92,7 @@ int main(int argc, char** argv) {
   // Print the configuration file(s) that were loaded.
   if (!cfg.config_file_path.empty())
     cfg.config_files.emplace_back(std::move(cfg.config_file_path));
-  for (auto& file : cfg.config_files)
+  for (const auto& file : system::loaded_config_files())
     VAST_INFO("loaded configuration file: {}", file);
   // Print the plugins that were loaded, and errors that occured during loading.
   for (const auto& file : *loaded_plugin_paths)
--- a/libvast/src/system/configuration.cpp
+++ b/libvast/src/system/configuration.cpp
@@ -20,6 +20,7 @@
 #include "vast/detail/installdirs.hpp"
 #include "vast/detail/load_contents.hpp"
 #include "vast/detail/overload.hpp"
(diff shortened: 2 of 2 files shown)


mavam:
> Don't think this needs a changelog entry since it's essentially only cleanup work.

Agreed.

lava (approved):
I didn't test but I applied the double-reversing algorithm using pen and paper, and it seems correct.

vast/vast.cpp
```diff
@@ -92,11 +92,13 @@ int main(int argc, char** argv) {
...
   if (!cfg.config_file_path.empty())
     cfg.config_files.emplace_back(std::move(cfg.config_file_path));
-  for (auto& file : cfg.config_files)
+  for (const auto& file : system::loaded_config_files())
     VAST_INFO("loaded configuration file: {}", file);
   // Print the plugins that were loaded, and errors that occured during loading.
   for (const auto& file : *loaded_plugin_paths)
     VAST_VERBOSE("loaded plugin: {}", file);
+  for (const auto& file : plugins::loaded_config_files())
+    VAST_INFO("loaded configuration file: {}", file);
```
mavam, line 101:
```suggestion
    VAST_INFO("loaded plugin configuration file: {}", file);
```

vast/vast.cpp
```diff
@@ -92,11 +92,13 @@ int main(int argc, char** argv) {
   // Print the configuration file(s) that were loaded.
   if (!cfg.config_file_path.empty())
     cfg.config_files.emplace_back(std::move(cfg.config_file_path));
-  for (auto& file : cfg.config_files)
+  for (const auto& file : system::loaded_config_files())
     VAST_INFO("loaded configuration file: {}", file);
```
mavam, line 96:
```suggestion
    VAST_INFO("loaded system configuration file: {}", file);
```

dominiklohmann: I've instead removed this; turns out that we printed this twice otherwise.

PR #2260, opened 2022-05-03
Add new debugging features to VAST tools

lava:
This PR adds some new debugging features to `lsvast` and `mdx-regenerate` (which has been renamed to `vast-regenerate`) that are helpful when working with index- and mdx-files.

### :memo: Checklist


tools/vast-regenerate/main.cpp
```diff
@@ -0,0 +1,306 @@
...
+#include <vast/system/configuration.hpp>
+#include <vast/system/index.hpp>
+#include <vast/uuid.hpp>
+#include <vast/value_index_factory.hpp>
+
+#include <fmt/core.h>
+
+#include <filesystem>
+
+int regenerate_mdx(const std::filesystem::path& dbdir) {
```
dominiklohmann, line 30: With how complex this functionality is, should we maybe turn it into a command plugin with its own unit- and integration-tests?

I think the same also applies for `lsvast`, which has integration tests defined in a very complicated way, because it predates the plugin API.

lava: Not sure if that makes sense with the current plugin API at least, since both of these tools are supposed to run without a VAST node, and afaik that's currently not possible for a plugin.

dominiklohmann: We can probably add an option to the `command_plugin` scaffolding that allows them to run exclusively, i.e., lock the PID file.

tools/vast-regenerate/main.cpp
```diff
@@ -0,0 +1,306 @@
+//    _   _____   __________
```
dominiklohmann, line 1: I cannot comment on file that were only renamed, but the `vast-regenerate` README still calls the tool `mdx-regenerate`.

tools/vast-regenerate/main.cpp
```diff
@@ -0,0 +1,306 @@
...
+    }
+    fmt::print(stderr, "successfully wrote {}\n", synopsis_path);
+  }
+  return 0;
+}
+
+caf::error write_index_bin(const std::vector<vast::uuid>& uuids,
+                           const vast::index_statistics& stats,
+                           const std::filesystem::path& index_file) {
+  flatbuffers::FlatBufferBuilder builder;
```
dominiklohmann, line 91: I'm worried about putting all this builder logic in here if it's essentially a duplicate from upstream libvast. We don't have any tests for this, so we will have to keep this in sync manually. I am not comfortable shipping this as part of a VAST installation without testing.

Is there some way we can aadd minimal tests for this, e.g., by verifying that vast-regenerate does not change databases that were created from the same VAST version?

dominiklohmann (commented):
Admittedly I didn't look at the code too much given that it's mostly debug tooling.

The `lsvast` changes I played with locally. They probably need to be adapted to include the changes from #2286 as well.

For `vast-regenerate` I'm worried about the complexity because we don't test whether it correctly uses the builders. Maybe we can (by default) cause it to write to a separate directory rather than replacing the files in-place?

tools/vast-regenerate/main.cpp
dominiklohmann: I'd also approve that, yes. We definitely shouldn't be shipping it with the static binary as-is.

nix/vast/default.nix
```diff
@@ -83,7 +83,7 @@ stdenv.mkDerivation rec {
     "-DVAST_ENABLE_BACKTRACE=ON"
     "-DVAST_ENABLE_JEMALLOC=ON"
     "-DVAST_ENABLE_LSVAST=ON"
-    "-DVAST_ENABLE_MDX_REGENERATE=ON"
+    "-DVAST_ENABLE_VAST_REGENERATE=OFF"
```
dominiklohmann, line 86: This variable name change is in conflict with #2309, which uses `VAST_ENABLE_MDX_REGENERATE` still. We need to be careful when merging either of the PRs.

PR #2263, opened 2022-05-04
Add a generic `compute` transform step for element-wise functions

dominiklohmann:
This follows up on an idea from @satta:

Example configuration:

```yaml
trim:
  fields:
    - hostname
  chars: '. '
```

At this stage of development it's more of an experiment rather than a new feature; I just wanted to showcase how easily such a thing is implemented. Before considering merging this we should discuss what options we want to support (ltrim, rtrim for example), what the configuration syntax should be, and we need to add tests.

### :memo: Checklist

Don't yet.


mavam:
Neat. This is good demo, but we need to find a scalable way to expose Arrow Compute functions. For example, there's [`FunctionRegistry::GetFunctionNames`](https://arrow.apache.org/docs/cpp/api/compute.html#_CPPv4NK5arrow7compute16FunctionRegistry16GetFunctionNamesEv) to dump all Arrow compute functions. The challenge is probably finding a unified representation to express the parameters (and check the validity at load time).

dominiklohmann:
> Neat. This is good demo, but we need to find a scalable way to expose Arrow Compute functions.

Yeah, as I suggested in Slack we can probably extend the "put" transform to take columns as input and make it possible to use all Scalar Compute functions on it. (We can theoretically also use the Vector Compute functions, but they're pipeline breakers, so that's for sometime in the future.)

libvast/native-plugins/compute.cpp
```diff
@@ -0,0 +1,152 @@
...
+      return caf::make_error(ec::unimplemented, "non-unary compute function");
+    auto function_options = std::make_unique<arrow::compute::TrimOptions>(". ");
+    auto fields = std::vector<std::string>{"hostname"};
+    auto inputs = std::vector<arrow::Datum>{};
+    return std::make_unique<compute_step>(std::move(function),
+                                          std::move(function_options),
+                                          std::move(fields), std::move(inputs));
+    return ec::unimplemented;
+  }
+};
```
dominiklohmann, line 148: @mavam I couldn't quite let this idea go: generally speaking, we "only" need to find some way to create `function_options` and `inputs` from the YAML configuration. Everything else is trivial.

This hardcoded version actually works nicely for the use case that was originally proposed by @satta as a trim transform step with the given example configuration.

libvast/native-plugins/compute.cpp
```diff
@@ -0,0 +1,152 @@
...
+    if (const auto* rec = caf::get_if<record>(&options))
+      if (rec->empty())
+        return caf::none;
+    return caf::make_error(ec::invalid_configuration, "expected empty "
+                                                      "configuration under "
+                                                      "vast.plugins.compute");
+  }
+
+  [[nodiscard]] const char* name() const override {
+    return "compute";
```
dominiklohmann, line 122: I find `compute` a bit too generic, so I'm open to ideas. Arrow calls the supported compute functions "element-wise" or "scalar" functions.

mavam: I like "compute" for now. I also like "mutate" from dplyr land, which puts emphasis on the fact that you're messing with the data.

But the best way to approach this is figuring out how it will gel with the pipeline syntax, then work backwards to YAML configuration.

libvast/native-plugins/compute.cpp
mavam: > generally speaking, we "only" need to find some way to create function_options and inputs from the YAML configuration. Everything else is trivial.

Yep, that was my understanding of the crux as well.

dominiklohmann: My current understanding is that we've just have to map them all by hand. It's not that many, though, and we gain the advantage of helpful error messages if we parse ourselves.

Here's a list of the possible options: https://arrow.apache.org/docs/cpp/api/compute.html#concrete-options-classes

dominiklohmann:
I'm closing this. We never found a way for this to make it past the prototype stage as there's no way to generically parse the arguments for Arrow's compute functions. We can pick this up again in the future.

PR #2264, opened 2022-05-04
Fix query pruning in the catalog

lava:
Updates the catalog to rewrite queries like `http.hostname == "foo" || dns.rrname == "foo"` to `:string == "foo"` (as opposed to rewriting them to `http.hostname == "foo"`, as was done previously).

Additionally, introduces a set of field names that have separate configured bloom filters (presumably with a higher precision than the default one), which should not be touched by the pruner.

### :memo: Checklist


dominiklohmann (approved):
The testbed metrics dashboard shows a significant reduction in catalog lookup time, so this is working nicely.

I've looked at the code and don't have any complaints either, but please add a changelog entry.

PR #2268, opened 2022-05-06
Compress in-memory slices with Zstd

dominiklohmann:
In a repeated test with 8'557'668 ingested Suricata events enabling the compression had a negligible effect on import speed (under 0.5%), but reduced the size of the archive and also the in-memory size of the loaded slices by roughly the same amount.

Here are the sizes of the database for three scenarios: Uncompressed, Zstd with the default compression level, and Zstd with the max compression level (50% slower on import, so irrelevant in the grand scheme of things).

```
❯ du -csh vast.db.uncompressed/{index,archive}
667M	vast.db.uncompressed/index
1.8G	vast.db.uncompressed/archive
2.5G	total

❯ du -csh vast.db.zstd.default/{index,archive}
668M	vast.db.zstd.default/index
426M	vast.db.zstd.default/archive
1.1G	total

❯ du -csh vast.db.zstd.max/{index,archive}
675M	vast.db.zstd.max/index
361M	vast.db.zstd.max/archive
1.0G	total
```

### :memo: Checklist

Discuss whether we want to enable this by default, or make this an option, or just not do it at all. After that comes testing and a changelog entry.

### DIFF
--- a/changelog/unreleased/features/2268--compress-table-slice.md
+++ b/changelog/unreleased/features/2268--compress-table-slice.md
@@ -1,0 +1,5 @@
+The `segment-store` store backend now compresses data before persisting it,
+resulting in 5x space savings for newly archived data, and an up to X% memory
+usage reduction. Depending on the speed of the disk this may improve the overall
+speed of VAST as well; we measured a negligible less than 1% performance penalty
+using the fastest disks we have (5GB/s).
--- a/changelog/unreleased/features/2268--compress-table-slice.md
+++ b/changelog/unreleased/features/2268--compress-table-slice.md
@@ -1,5 +1,7 @@
 The `segment-store` store backend now compresses data before persisting it,
-resulting in 5x space savings for newly archived data, and an up to X% memory
-usage reduction. Depending on the speed of the disk this may improve the overall
-speed of VAST as well; we measured a negligible less than 1% performance penalty
-using the fastest disks we have (5GB/s).
+resulting in over 2x space savings for newly written data with the default VAST
+configuration. This allowed us to increase the default partition size from
+1'048'576 to 4'194'304 events, and the default number of events in a single
+batch from 1024 to 65'536, yielding a significant performance increase at the
+cost of a ~20% memory increase at peak load. We think this is a better default;
+if you require less memory usage, reduce the value of `vast.max-partition-size`.
--- a/changelog/unreleased/features/2268--compress-table-slice.md
+++ b/changelog/unreleased/features/2268--compress-table-slice.md
@@ -2,7 +2,7 @@ VAST now compresses data with Zstd. When persisting data to the segment store,
 the default configuration achieves over 2x space savings. When transferring data
 between client and server processes, compression reduces the amount of
 transferred data by up to 5x. This allowed us to increase the default partition
-size from 1'048'576 to 4'194'304 events, and the default number of events in a
-single batch from 1,024 to 65'536. The superior performance increase comes at
+size from 1,048,576 to 4,194,304 events, and the default number of events in a
+single batch from 1,024 to 65,536. The superior performance increase comes at
 the cost of a ~20% memory footprint increase at peak load. Use the option
 `vast.max-partition-size` to tune this space-time tradeoff.
--- a/changelog/unreleased/features/2268--compress-table-slice.md
+++ b/changelog/unreleased/features/2268--compress-table-slice.md
@@ -1,9 +1,8 @@
-VAST now compresses data before persisting it or sending it between processes,
-resulting in over 2x space savings for newly written data using the
-`segment-store` store backend with the default VAST configuration, and resulting
-in an up to 5x reduction of transferred data between `vast import` processes and
-the VAST server. This allowed us to increase the default partition size from
-1'048'576 to 4'194'304 events, and the default number of events in a single
-batch from 1024 to 65'536, yielding a significant performance increase at the
-cost of a ~20% memory increase at peak load. We think this is a better default;
-if you require less memory usage, reduce the value of `vast.max-partition-size`.
+VAST now compresses data with Zstd. When persisting data to the segment store,
+the default configuration achieves over 2x space savings. When transferring data
+between client and server processes, compression reduces the amount of
+transferred data by up to 5x. This allowed us to increase the default partition
+size from 1'048'576 to 4'194'304 events, and the default number of events in a
+single batch from 1,024 to 65'536. The superior performance increase comes at
+the cost of a ~20% memory footprint increase at peak load. Use the option
+`vast.max-partition-size` to tune this space-time tradeoff.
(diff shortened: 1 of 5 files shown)


dominiklohmann:
> Out of curiosity, trying to understand the entropy: can you zstd-compress the index (on the filesystem) as well and report on the sizes?

667M to 229M when doing it per-file.

mavam:
> 667M to 229M when doing it per-file.

So down to ~34% of the original size.

dispanser:
I guess it would be an easy win if this could be made configurable. There seem to be people out there using a compressed file system with VAST, and for those scenarios the additional compression might become a performance degradation, as there would be two layers attempting to compress this.

tobim:
Very nice! I would like to measure the export performance and RSS as well.

> There seem to be people out there using a compressed file system with VAST, and for those scenarios the additional compression might become a performance degradation, as there would be two layers attempting to compress the data.

Probably not even necessary, at least btrfs opts out of compression very quickly if it doesn't manage to compress the first few kB itself. This should also be much more effective than compressing individual filesystem blocks.



 > >   Out of curiosity, trying to understand the entropy: can you zstd-compress the index (on the filesystem) as well and report on the sizes?
>
> 667M to 229M when doing it per-file.

That is promising, but we should probably do it a per-value-index basis instead of the whole file.

dominiklohmann:
> I guess it would be an easy win if this could be made configurable.

Yes, I think so. Even with Parquet stores coming up this is still valuable to reduce the memory usage.

> Very nice! I would like to measure the export performance and RSS as well.

Will do and report back!

> > > Out of curiosity, trying to understand the entropy: can you zstd-compress the index (on the filesystem) as well and report on the sizes?
> > 
> > 667M to 229M when doing it per-file.
> 
> That is promising, but we should probably do it a per-value-index basis instead of the whole file.

I think that's for a separate story, let's keep this focussed on the table slices only for now.

tobim:
> I think that's for a separate story, let's keep this focussed on the table slices only for now.

Definitely.

dominiklohmann:
> Very nice! I would like to measure the export performance

No noticeable difference for Zstd with default compression. Unpacking is probably slightly more expensive, but sending things over the wire is much cheaper. The higher compression level is especially bad when selecting subsets of table slices, because that requires re-compressing the results before sending them over with the current approach. But from this point of view alone I don't see why we'd not just always enable Zstd with default compression.

> and RSS as well.

For the Zstd default during export:

```
Maximum resident set size (kbytes): 2186784
Socket messages sent: 2818
Socket messages received: 584
```

For the uncompressed version:

```
Maximum resident set size (kbytes): 2406112
Socket messages sent: 19274
Socket messages received: 574
```

changelog/unreleased/features/2268--compress-table-slice.md
```diff
@@ -0,0 +1,5 @@
+The `segment-store` store backend now compresses data before persisting it,
+resulting in 5x space savings for newly archived data, and an up to X% memory
+usage reduction. Depending on the speed of the disk this may improve the overall
+speed of VAST as well; we measured a negligible less than 1% performance penalty
+using the fastest disks we have (5GB/s).
```
dominiklohmann, line 5: Note that X = 10% on my local M1 Max MBP 16' for a 100 GB database; I left this running earlier today.

We should roll this PR out on the testbed to get some numbers such that we can put them in here. I also measured with `vast.import.batch-size: 65536` rather than the default `1024` because I forgot to revert to the defaults, but quick local tests show that the numbers are almost as good with the measly default batch size.

dominiklohmann: > We should note that compression trades random-access performance.

That's not necessarily true. The compression is per column, and once you access it it actually stays decompressed in-memory in the record batch until the ref count drops to zero.

> In fact, if the archive is now smaller than the index, we should make dense indexes optional as well in the same release.

I disagree—that would be follow-up work. There's no reason to delay an improvement just because there's room for further improvement because of it. By that logic we would never be able to cut a release at all.

I also doubt that we can deduce from the relative size of store and dense indexes alone that the dense indexes are now irrelevant. While I agree that it is likely, we would need benchmarks confirming this before making that change. Please write a follow-up story if you want to investigate this further.

dominiklohmann:
I've extended this PR with the tuned defaults following my extensive measurements on our testbed server.

changelog/unreleased/features/2268--compress-table-slice.md
dominiklohmann: Note that I've rewritten the changelog entry since we started this discussion to also show the overall reduction in space and to also mention the tuned defaults.

libvast/include/vast/defaults.hpp
```diff
@@ -31,7 +31,7 @@ constexpr size_t max_recursion = 100;
 namespace import {
 
 /// Maximum size for sources that generate table slices.
-constexpr size_t table_slice_size = 1024;
+constexpr size_t table_slice_size = 1 << 16; // 64 ki
```
mavam, line 34: Why not use SI literals here? The header is cheap (no transitive includes) and this becomes *very* clear.

mavam (commented):
I think we can make these values accessible to those that haven't burnt powers of two in their heads.

changelog/unreleased/features/2268--compress-table-slice.md
```diff
@@ -0,0 +1,9 @@
+VAST now compresses data before persisting it or sending it between processes,
+resulting in over 2x space savings for newly written data using the
+`segment-store` store backend with the default VAST configuration, and resulting
+in an up to 5x reduction of transferred data between `vast import` processes and
+the VAST server. This allowed us to increase the default partition size from
+1'048'576 to 4'194'304 events, and the default number of events in a single
+batch from 1024 to 65'536, yielding a significant performance increase at the
+cost of a ~20% memory increase at peak load. We think this is a better default;
+if you require less memory usage, reduce the value of `vast.max-partition-size`.
```
mavam, line 9:
```suggestion
VAST now achieves 2-5x space savings on IPC operations by enabling Zstd column
compression. When persisting data to the segment store, the default
configuration achieves over 2x space savings. When transferring data between
client and server processes, compression reduces the amount of transferred data
by up to 5x. This allowed us to increase the default partition size from
1'048'576 to 4'194'304 events, and the default number of events in a single
batch from 1,024 to 65'536. The superior performance increase comes at the
cost of a ~20% memory footprint increase at peak load. Use the option
`vast.max-partition-size` to tune this space-time tradeoff.
```

libvast/include/vast/defaults.hpp
dominiklohmann: I don't want to add a `using namespace vast::si_literals` here, and I can't use the literals without that. I can also just spell the values out, but the literals would only be an option if I did it like this:

```cpp
constexpr auto table_slice_size = []() noexcept -> size_t {
  using namespace vast::si_literals;
  return 64_Ki;
}();
```

Which isn't all that readable and would cause the lambda to be instantiated once per translation unit that includes this file.

mavam: Uh, so annoying.

PR #2274, opened 2022-05-10
Report by schema metrics from the importer

tobim:
This replaces the metrics that are reported from each individual source with an accumulated value from the importer. In other words, we now report how many events of a type were received at the node as opposed to how many were produced from a source.

### :memo: Checklist

Commit by commit.


libvast/src/system/importer.cpp
```diff
@@ -226,9 +232,16 @@ void importer_state::send_report() {
   using namespace std::string_literals;
   auto elapsed = std::chrono::duration_cast<duration>(now - last_report);
   auto node_throughput = measurement{elapsed, measurement_.events};
+  auto num_schemas_seen = schema_counters.size();
+  std::vector<performance_sample> samples = {};
+  samples.reserve(num_schemas_seen + 2);
+  samples.emplace_back("importer"s, measurement_);
+  samples.emplace_back("node_throughput"s, node_throughput);
+  for (const auto& [name, count] : schema_counters)
+    samples.emplace_back(name, measurement{elapsed, count});
```
tobim, line 241: I think we should probably prefix these keys with `"schema"` or something similar.

After the key names have been decided the docs need to be updated.

libvast/include/vast/system/importer.hpp
```diff
@@ -96,8 +97,30 @@ struct importer_state {
...
+    }
+    bool operator()(std::string_view lhs, const std::string& rhs) const {
+      return lhs == rhs;
+    }
+    bool operator()(const std::string& lhs, const std::string& rhs) const {
+      return lhs == rhs;
+    }
+  };
+  tsl::robin_map<std::string, uint64_t, string_view_hash, string_view_equal>
+    schema_counters = {};
```
tobim, line 123: I just discovered that we already have `vast/detail/heterogenous_string_hash.hpp`.
Should use that here.

libvast/src/system/importer.cpp
mavam: > For the new metrics my preference would be just `throughput`, with the schema name as an additional metadata field so we can choose to either render accumulated values or to render the values individually.

It looks like we need to agree on the general structure of the metrics hierarchy first. Proposal: let's take a step back and agree on an intuitive scheme, then evaluate concrete suggestions.

tobim: > I'm against putting the schema name in the metric key itself, as that makes it rather hard to use the data from a dashboarding perspective.

I didn't consider this, but I guess you're right. You're run-of-the-mill dashboard won't be able to prefix-match on metrics keys.

I would still like to use the `ingest` identifier for the top-level. The key would then be `ingest.schemas` with each value tagged by the individual schema names. We can also add a value with the `total` tag that holds the same value as the current `node_throughput`.

tobim: Stacking the lines makes it really hard to see how the rates relative to each other and jitters propagate upwards. Maybe we should just keep `node_throughput` as a separate key for those who want the total counter.

dominiklohmann: Having the total in a separate key sounds good to me.

libvast/src/system/importer.cpp
```diff
@@ -226,9 +232,18 @@ void importer_state::send_report() {
   using namespace std::string_literals;
   auto elapsed = std::chrono::duration_cast<duration>(now - last_report);
   auto node_throughput = measurement{elapsed, measurement_.events};
+  auto num_schemas_seen = schema_counters.size();
+  std::vector<performance_sample> samples = {};
+  samples.reserve(num_schemas_seen + 2);
```
dominiklohmann, line 237: The +2 isn't necessary anymore.

dominiklohmann (approved):
Thanks for implementing this! Looking forward to plotting this on our dashboard. 📈

libvast/src/system/importer.cpp
```diff
@@ -226,9 +232,18 @@ void importer_state::send_report() {
...
+  samples.push_back(performance_sample{"importer"s, measurement_});
+  samples.push_back(performance_sample{"node_throughput"s, node_throughput});
+  for (const auto& [name, count] : schema_counters)
+    samples.push_back(performance_sample{
+      "ingest", measurement{elapsed, count}, {{"schema", name}}});
```
dominiklohmann, line 242: One additional note: Before merging, please add a changelog entry and update the docs section that lists all metrics.

PR #2277, opened 2022-05-13
wip: Write homogenous partitions from the partition transformer

lava:
NOTE: This PR contains unfinished work in progress.

Ensure that the partition transformer always produces uniform partitions, ie. partitions that only contain a single type.

### :memo: Checklist


libvast/src/system/spawn_index.cpp
```diff
@@ -58,20 +57,8 @@ spawn_index(node_actor::stateful_pointer<node_state> self,
...
-      opt("vast.active-partition-timeout", sd::active_partition_timeout),
-      opt("vast.max-resident-partitions", sd::max_in_mem_partitions),
-      opt("vast.max-taste-partitions", sd::taste_partitions),
-      opt("vast.max-queries", sd::num_query_supervisors),
-      std::filesystem::path{opt("vast.catalog-dir", indexdir.string())},
-      std::move(index_config));
-    VAST_VERBOSE("{} spawned the index", *self);
-    return caf::actor_cast<caf::actor>(handle);
+    VAST_WARN("The `vast.use-legacy-query-scheduler` is not supported anymore "
+              "and will be ignored.");
```
tobim, line 61: I would like to have this in a dedicated PR.
And I would prefer to treat this as an error.

libvast/src/system/partition_transformer.cpp
```diff
@@ -255,27 +296,40 @@ partition_transformer_actor::behavior_type partition_transformer(
...
-          = caf::make_error(ec::invalid_argument,
-                            "could not create store builder for backend {}",
-                            store_id);
-        store_or_fulfill(self, std::move(stream_data));
-        return;
-      }
-      self->state.data.store_header = builder_and_header->second;
-      self->state.store_builder = builder_and_header->first;
-      VAST_ASSERT(self->state.store_builder);
       self->state.stage = caf::attach_continuous_stream_stage(
```
tobim, line 299: We can use a (regular?) source instead. Maybe evaluate after the main change is functional.

libvast/src/system/partition_transformer.cpp
```diff
@@ -255,27 +296,40 @@ partition_transformer_actor::behavior_type partition_transformer(
...
+        [](caf::unit_t&, const caf::error&) { /* nop */ },
+        // For this stream stage we want a `broadcast_downstream_manager<T>`,
+        // where slices only go the store with the correct layout, so:
+        //
+        //   T:      vast::table_slice
+        //   Filter: vast::type
+        //   Select: vast::(anon)::partition_transformer_selector
+        caf::policy::arg<caf::broadcast_downstream_manager<
+          table_slice, vast::type, partition_transformer_selector>>{});
+      for (auto& [layout, partition_data] : self->state.data) {
```
tobim, line 314: `continue` here in case the partition has no data.

libvast/src/system/partition_transformer.cpp
```diff
@@ -291,41 +345,53 @@ partition_transformer_actor::behavior_type partition_transformer(
...
+          self->state.stage->out().push(slice);
+        }
       }
       self->state.finalize_data();
       detail::shutdown_stream_stage(self->state.stage);
       auto stream_data = partition_transformer_state::stream_data{};
       [&] {
-        { // Pack partition
+        stream_data.partition_chunks
+          = std::vector<std::tuple<vast::uuid, vast::type, vast::chunk_ptr>>{};
```
tobim, line 362: The vector should already be default initialized from line 359.

libvast/src/system/partition_transformer.cpp
```diff
@@ -95,91 +99,115 @@ void partition_transformer_state::add_slice(const table_slice& slice) {
...
+  for (auto& [layout, typed_indexers] : indexers)
+    for (auto& [qf, idx] : typed_indexers)
+      data[layout].indexer_chunks.emplace_back(qf.name(), chunkify(idx));
+  for (auto& [layout, partition_data] : data) {
+    auto& mutable_synopsis = partition_data.synopsis.unshared();
+    mutable_synopsis.shrink();
+    // TODO: It would probably make more sense if the partition
+    // synopsis keeps track of offset/events internally.
+    mutable_synopsis.offset = partition_data.offset;
+    mutable_synopsis.events = partition_data.events;
```
tobim, line 111: Can we find another way to group the logic. This can probably be inlined and fused together into the respective loops that come afterwards?

libvast/src/system/partition_transformer.cpp
```diff
@@ -95,91 +99,115 @@ void partition_transformer_state::add_slice(const table_slice& slice) {
...
     return;
   }
-  auto promise = path_data.promise;
   if (self->state.events == 0) {
-    promise.deliver(augmented_partition_synopsis{
-      .uuid = vast::uuid::nil(),
-      .stats = std::move(self->state.stats),
-      .synopsis = nullptr,
-    });
+    promise.deliver(std::vector<augmented_partition_synopsis>{});
```
tobim, line 126: We should probably store a transformation error in the state in case that went wrong.

tobim:
[sc32742]

tobim:
[sc-32742]

libvast/src/system/spawn_index.cpp
lava: I will split it out before the final review, but I disagree about making it an error because that would be a breaking change that would technically require a VAST 3.0 release.

tobim: I don't agree with overly strict interpretations of versioning guidelines, but I don't care strongly enough about this particular case, so fine by me.

For the actual change though, can you try to revert 8beee922e94d50fd41384575ad925bc9a34e2eeb instead of the manual deletion? I'm afraid that we'll leave dead code lying around.

libvast/src/system/partition_transformer.cpp
lava: After further discussion, we decided that using a source would be odd since caf assumes that sources are always generators, but we exlusively push to this one so we could do nothing but assert failure in `generate()`.

libvast/src/system/spawn_index.cpp
dominiklohmann: Can we move this deprecation warning to the other deprecation warnings in `main()` in `vast/vast.cpp`?

libvast/src/system/partition_transformer.cpp
lava: Actually it's not, the field is a `caf::expected<std::vector<...>>`, so the default is to initialize it as an error.

libvast/src/system/partition_transformer.cpp
```diff
@@ -291,41 +340,66 @@ partition_transformer_actor::behavior_type partition_transformer(
...
-        self->state.stage->out().push(slice);
+      for (auto& [_, data] : self->state.data) {
+        data.offset = offset;
+        for (auto& slice : self->state.slices) {
+          slice.offset(offset);
+          offset += slice.rows();
+          self->state.add_slice(slice);
+          self->state.stage->out().push(slice);
+        }
+      }
```
tobim, line 351: I don't understand this. Why do we push every slice for every unique layout?

libvast/src/system/partition_transformer.cpp
```diff
@@ -291,41 +340,66 @@ partition_transformer_actor::behavior_type partition_transformer(
...
+        for (auto& [qf, idx] : typed_indexers)
+          self->state.data[layout].indexer_chunks.emplace_back(qf.name(),
+                                                               chunkify(idx));
+      for (auto& [layout, partition_data] : self->state.data) {
+        auto& mutable_synopsis = partition_data.synopsis.unshared();
+        mutable_synopsis.shrink();
+        // TODO: It would probably make more sense if the partition
+        // synopsis keeps track of offset/events internally.
+        mutable_synopsis.offset = partition_data.offset;
+        mutable_synopsis.events = partition_data.events;
```
tobim, line 362: From here to Line 413 we do 3 passes over `self->state.data`. Is there a particular reason not to do a single loop that finishes up one partition on each iteration?

tobim (commented):
A few additional questions. I think this only needs a little polish now (and maybe a fix for the serialization data if I understand the code correctly).

libvast/src/system/partition_transformer.cpp
lava: I've combined the first two, the last one is still required to be separate because all indexers need to be chunkified already for creating the final flatbuffer.

libvast/src/system/partition_transformer.cpp
lava: You're right, it's a bug, we should only consider the slices with the correct layout here.

libvast/test/system/partition_transformer.cpp
```diff
@@ -229,10 +243,101 @@ TEST(delete transform / persist before done) {
...
     });
 }
 
+TEST(partition with multiple types) {
+  // Spawn partition transformer
+  auto store_id = "segment-store"s;
+  auto synopsis_opts = vast::index_config{};
+  auto index_opts = caf::settings{};
+  auto transform = std::make_shared<vast::transform>(
+    "partition_transform"s, std::vector<std::string>{"zeek.conn"});
```
tobim, line 256: The outcome should be the same in this case, but the following should be more correct:
```suggestion
    "partition_transform"s, std::vector<std::string>{});
```

tobim (approved):
It's great that this is fixed now. Only the test needs to use the VAST_FMT_RUNTIME macro fix.

PR #2284, opened 2022-05-17
Parquet store plugin

dispanser:
Preliminary write / read path for the parquet store plugin.

### :memo: Checklist

Please take a look at the general approach, mostly on the read path with respect to transforming the table


mavam (commented):
Can't wait to see this landing in master!

Quick comment regarding the plugin naming: could you rename it to `parquet`? All other plugins don't have the plugin type in their name, because (1) it's possible to implement multiple plugins, and (2) there is just one global namespace for names that the current plugins map to in a one-to-one fashion.

dispanser:
We probably need some CI scaffolding to to build and test the parquet plugin.

dominiklohmann:
> We probably need some CI scaffolding to to build and test the parquet plugin.

You need to add it in three places:
- The Dockerfile
- The Nix build script for the static binary
- The configuration matrix of the VAST GitHub Actions workflow file

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,564 @@
...
+restore_enum_chunk_array(const vast::type& t,
+                         std::shared_ptr<arrow::ChunkedArray> array) {
+  auto f = detail::overload{
+    [&](const enumeration_type& et)
+      -> std::optional<std::shared_ptr<arrow::ChunkedArray>> {
+      return map_chunked_array(et, array, fix_enum_array);
+    },
+    [&](
+      const list_type&) -> std::optional<std::shared_ptr<arrow::ChunkedArray>> {
+      return map_chunked_array(t, array, map_array);
```
dominiklohmann, line 213: I don't understand why you're passing in `t` rather than the already resolved `list_type` here. If that is passed to `map_array` wouldn't you just need to visit in there again?

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,564 @@
...
+    = parse_arrow_schema_from_metadata(parquet_reader->metadata());
+  std::unique_ptr<parquet::arrow::FileReader> file_reader{};
+  if (!parquet::arrow::OpenFile(bufr, arrow::default_memory_pool(),
+                                &file_reader)
+         .ok())
+    return caf::exit_reason::unhandled_exception;
+  std::shared_ptr<arrow::Table> table{};
+  if (!file_reader->ReadTable(&table).ok())
+    return caf::exit_reason::unhandled_exception;
+  return map_enum_dictionary(arrow_schema, table);
```
dominiklohmann, line 338: This function definitely needs a better name, but I think the general idea works.

It'd be nice to see how much overhead this has. You can easily instrument by using a steady clock, e.g.,

```c++
const auto start = std::chrono::steady_clock::now();
// do stuff
const auto delta = std::chrono::steady_clock::now() - start;
VAST_DEBUG("someone did stuff in {}", data{std::chrono::duration_cast<duration>(delta)});
```

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,564 @@
...
+    return {};
+  return *arrow_schema;
+}
+
+caf::expected<std::shared_ptr<arrow::Table>>
+read_parquet_buffer(const chunk_ptr& chunk) {
+  const auto* ptr
+    = static_cast<const uint8_t*>(static_cast<const void*>(chunk->data()));
+  auto bufr = std::make_shared<arrow::io::BufferReader>(
+    ptr, detail::narrow_cast<int64_t>(chunk->size()));
```
dominiklohmann, line 326: Is there an overload of the BufferReader constructor that takes an `arrow::Buffer`? If so, we should use that instead and convert the chunk into a buffer. If I recall correctly the buffer reader is then able to signal that it can be used without copying data. Not sure if Parquet makes use of that, but in any case the code would also be easier to read because you don't need all the pointer casting.

(Also note that the double cast in the line above should just be a single `reinterpret_cast`.)

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,564 @@
...
+  self->state.self_ = self;
+  self->state.id_ = id;
+  self->state.accountant_ = accountant;
+  self->state.fs_ = fs;
+  self->state.path_ = store_path_for_partition(id);
+  self->request(self->state.fs_, caf::infinite, atom::mmap_v, self->state.path_)
+    .then(
+      [self](const chunk_ptr& chunk) {
+        if (auto table = read_parquet_buffer(chunk); table) {
+          self->state.table_ = *table;
```
dominiklohmann, line 354: Can we assert here or inside the `read_parquet_buffer` function that the adjusted schema of the table matches the schema in the parquet metadata exactly?

dominiklohmann (commented):
This looks really good so far. As we talked about I mostly took a look at the post read table fixup that we sadly have to do. I think the way we're doing it now works nicely. I also left a few remarks inline for things I've noticed.

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,564 @@
...
+    if (std::optional<std::shared_ptr<arrow::Array>> c = m(t, chunk))
+      chunks.push_back(*c);
+    else
+      return arr;
+  }
+  auto result = arrow::ChunkedArray::Make(chunks);
+  return result.ValueOrDie();
+}
+
+std::optional<std::shared_ptr<arrow::Array>>
```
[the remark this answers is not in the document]
dominiklohmann: I think it's okay. Do we really need the second level of nullability, though? Doesn't a shared_ptr already suffice as you could just return nullptr instead of nullopt?

plugins/parquet-store/parquet_store.cpp
dispanser: That's actually difficult because we reconstruct the table expicitly with the desired schema - I don't see what we could compare against here.

Two things:
1. `arrow_table_slice_builder::create(arrow::RecordBatch)` runs `Verify` which during development threw errors when the table schema and array type did not match
2. while traversing the schema tree, I added:
```c++
    [&](const auto&) -> std::optional<std::shared_ptr<arrow::ChunkedArray>> {
      VAST_ASSERT(t.to_arrow_type()->Equals(array->type()));
      return {};
    },
```
To make sure that the array type matches the schema type on every node and every level. Similar code has been added to the regular `arrow::Array` handlers.

I hope that addresses your concern.

dominiklohmann: Yes, that works perfectly for me. Thanks!

plugins/parquet-store/parquet_store.cpp
dispanser: renamed function to `align_table_to_schema`, which hopefully better captures the intend here.

I added the time tracking directly to the function.

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,564 @@
...
+  arrow::io::BufferReader input(schema_buf);
+  auto arrow_schema = arrow::ipc::ReadSchema(&input, &dict_memo);
+  if (!arrow_schema.ok())
+    return {};
+  return *arrow_schema;
+}
+
+caf::expected<std::shared_ptr<arrow::Table>>
+read_parquet_buffer(const chunk_ptr& chunk) {
+  const auto* ptr
```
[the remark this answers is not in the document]
dispanser: I've added this, but want to point out that the existing (segment) store doesn't seem to do this, either. I'm not sure about the contract of the file system layer, but I'd assume to receive some form of error if the data couldn't be read instead of a null ptr.

plugins/parquet-store/parquet_store.cpp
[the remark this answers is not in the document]
dispanser: That's a great point, didn't think of that. I stripped out all those `std::optionals` and things just keep working, it's magical.

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,564 @@
...
+                    rb.status());
+        }
+      }
+    },
+    [&](const query::extract& extract) {
+      if (extract.policy == query::extract::preserve_ids) {
+        return self->send(extract.sink, table_slice{});
+      }
+      for (const auto& rb : record_batch_reader) {
+        if (rb.ok()) {
```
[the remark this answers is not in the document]
dispanser: Improved error handling to leverage `caf::expected` as it's already promised in the outer function as return type.

dominiklohmann (commented):
This is looking really good! 🚀

I played around with it for an hour or so and did some benchmarks, and I took a second look at the code, for which I left some comments, although none of them are major blockers in any way.

Regarding the performance, There was no noticeable difference compared to current master in the import performance, and the store files are ~40% smaller than with the segment store backend.

For the export performance I noticed a dip that was larger than I expected (~20%). When running many queries in parallel, the conversion from `std::shared_ptr<arrow::Table>` to `std::vector<table_slice>` dominates benchmarks. I was wondering if it would be a good idea to cache the result of that conversion and to only do it once the first query arrives.

The next step after fixing that would be to take this PR and the `rebuild` command from #2321 and to try a large-scale conversion on our test server to evaluate just how well this performs for larger setups than I can run locally.

.github/workflows/vast.yaml
```diff
@@ -621,6 +621,9 @@ jobs:
           - name: Sigma
             target: sigma
             path: plugins/sigma
+          - name: Parquet
+            target: parquet-store
```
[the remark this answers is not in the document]
mavam: Yes please!

libvast/src/type.cpp
```diff
@@ -3305,7 +3305,12 @@ int sum_type_access<arrow::DataType>::index_from_type(
                               caf::detail::tl_size<extension_types>::value>());
   // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index)
   auto result = table[x.id()];
-  VAST_ASSERT(result != unknown_id, "unexpected Arrow type id");
+  VAST_ASSERT(result != unknown_id, fmt::format("unexpected Arrow type id '{}' "
```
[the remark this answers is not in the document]
dispanser: since `x` is of type `arrow::DataType&` I'm just doing `x.ToString()` three lines down. What additional information could we print here based on visiting the id?

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,617 @@
...
+    if (auto new_arr
+        = restore_enum_chunk_array(rt.field(i).type, table->column(i)))
+      arrays.push_back(new_arr);
+    else
+      arrays.push_back(table->column(i));
+  }
+  auto new_table = arrow::Table::Make(target_schema, arrays, table->num_rows());
+  const auto delta = std::chrono::steady_clock::now() - start;
+  VAST_DEBUG("table schema aligned in {}[ns]",
+             data{std::chrono::duration_cast<duration>(delta)});
```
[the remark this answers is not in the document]
dispanser: is this to say that we can remove the measurement and debug statement, or do we keep it anyway?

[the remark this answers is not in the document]
dominiklohmann: It's still an interesting metric and doesn't hurt to keep around, I think.

plugins/parquet-store/parquet_store.hpp
```diff
@@ -0,0 +1,117 @@
...
+  /// Number of events in this store.
+  size_t num_rows_ = {};
+};
+
+struct store_state {
+  static constexpr const char* name = "passive-parquet-store";
+  uuid id_ = {};
+  system::store_actor::pointer self_ = {};
+
+  std::shared_ptr<arrow::Table> table_ = {};
```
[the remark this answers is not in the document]
dispanser: I'm trying this. I used `std::optional<std::vector<table_slice>>` to be able to exppress that the data has not been loaded yet, which is mostly the same approach as the segment store plugin, which uses `caf::optional<vast::segment>`.

Does `caf::optional` provide any benefits here over its counterpart from `std`?

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,617 @@
...
+caf::expected<uint64_t>
+handle_lookup(Actor& self, const vast::query& query,
+              const std::shared_ptr<arrow::Table>& table) {
+  auto table_slices = std::vector<table_slice>{};
+  for (const auto& rb : arrow::TableBatchReader(*table)) {
+    if (!rb.ok())
+      return caf::make_error(ec::format_error,
+                             fmt::format("unable to read record batch: {} ",
+                                         rb.status().ToString()));
+    table_slices.push_back(create_table_slice(*rb));
```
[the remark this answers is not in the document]
dispanser: As I moved to transform into table slices immediately and don't store the table at all, the second `handle_lookup` is gone completely so I assume this comment doesn't apply any more.

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,617 @@
...
+      duration runtime = std::chrono::steady_clock::now() - start;
+      auto id_str = fmt::to_string(query.id);
+      self->send(self->state.accountant_, "parquet-store.lookup.runtime",
+                 runtime,
+                 vast::system::metrics_metadata{{"query", id_str},
+                                                {"store-type", "passive"}});
+      self->send(self->state.accountant_, "parquet-store.lookup.hits",
+                 *num_hits,
+                 vast::system::metrics_metadata{{"query", id_str},
+                                                {"store-type", "passive"}});
```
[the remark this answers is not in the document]
dispanser: The metrics are the same as used for the segment store:
```
$store.lookup.runtime
$store.lookup.hits
```

I assume the documentation would go into the plugin README?

plugins/parquet-store/parquet_store.hpp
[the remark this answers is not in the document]
dominiklohmann: `caf::optional` is deprecated, please don't use it anymore. It doesn't exist on CAF master anymore.

dispanser:
@dominiklohmann : addressed all comments in the previous commit, so it's probably easiest to just review that one. Renaming of the plugin will be done last to make the diff more digestible.

plugins/parquet-store/parquet_store.cpp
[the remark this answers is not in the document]
dominiklohmann: Metric docs go here until they're migrated to vast.io: https://docs.tenzir.com/vast/features/metrics

changelog/unreleased/features/2284--parquet-store-plugin.md
```diff
@@ -0,0 +1,6 @@
+A new parquet store plugin allows VAST to store its data as parquet files,
+increasing storage efficiency at the expense of higher deserialization costs.
+Storage requirement for the archive (excluding indices) is reduced by 50%
```
dominiklohmann, line 3: I think the "excluding indices" part is a bit too complicated for users, which would be interested the overall numbers instead.

changelog/unreleased/features/2284--parquet-store-plugin.md
```diff
@@ -0,0 +1,6 @@
...
+compared to the existing segment store with zstd compression enabled.
```
dominiklohmann, line 4: Typo: s/zstd/Zstd

changelog/unreleased/features/2284--parquet-store-plugin.md
```diff
@@ -0,0 +1,6 @@
...
+CPU usage for suricata import is up ~ 10%,  mostly related to the more
+expensive serialization.
```
dominiklohmann, line 6: Did you benchmark this again after switching to storing the list of table slices over the Arrow table?

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,722 @@
...
+  auto arrow_schema
+    = parse_arrow_schema_from_metadata(parquet_reader->metadata());
+  std::unique_ptr<parquet::arrow::FileReader> file_reader{};
+  if (!parquet::arrow::OpenFile(bufr, arrow::default_memory_pool(),
+                                &file_reader)
+         .ok())
+    return caf::exit_reason::unhandled_exception;
+  std::shared_ptr<arrow::Table> table{};
+  if (!file_reader->ReadTable(&table).ok())
+    return caf::exit_reason::unhandled_exception;
```
dominiklohmann, line 485: This looks like the wrong error code to me in both cases. Can we use something fitting from `vast::ec`, and add the stringified status to the error?

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,722 @@
...
+    }
+    VAST_TRACE("{} received batch of {} table slices for partition {}", *self,
+               batch.size(), self->state.id_);
+  };
+}
+
+std::shared_ptr<parquet::WriterProperties>
+writer_properties(const configuration& config) {
+  auto builder = parquet::WriterProperties::Builder{};
+  builder.created_by("VAST telemetry engine")
```
dominiklohmann, line 514: @mavam what do we want to use here?

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,722 @@
...
+    ->max_row_group_length(detail::narrow_cast<int64_t>(config.row_group_size))
+    ->enable_dictionary()
+    ->compression(parquet::Compression::ZSTD)
+    ->compression_level(9)
```
dominiklohmann, line 518: Why 9 instead of the default level? Should we make this configurable instead?

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,722 @@
...
+    ->version(parquet::ParquetVersion::PARQUET_2_6);
+  return builder.build();
+}
+
+// TODO: make this static or something - no need to recompute every time
```
dominiklohmann, line 523: Should we still do this?

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,722 @@
...
+  auto sink = arrow::io::BufferOutputStream::Create().ValueOrDie();
+  auto batches = arrow::RecordBatchVector{};
+  for (const auto& slice : slices)
+    batches.push_back(create_record_batch(slice));
+  auto table = arrow::Table::FromRecordBatches(batches).ValueOrDie();
+  auto writer_props = writer_properties(config);
+  auto arrow_writer_props = arrow_writer_properties();
+  auto status
+    = parquet::arrow::WriteTable(*table, arrow::default_memory_pool(), sink,
+                                 1 << 24, writer_props, arrow_writer_props);
```
dominiklohmann, line 567: Is the 2^24 here the row group size?

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,722 @@
...
+  VAST_ASSERT(status.ok(), status.ToString().c_str());
+  return sink->Finish().ValueOrDie();
+}
+
+auto finish_parquet(
+  system::store_builder_actor::stateful_pointer<store_builder_state> self) {
+  return [self](caf::unit_t&, const caf::error&) {
+    auto buffer
+      = write_parquet_buffer(self->state.table_slices_, self->state.config_);
+    VAST_TRACE("{} writing partition {} with {} events in {} table slices, ",
```
dominiklohmann, line 577:
```suggestion
    VAST_TRACE("{} writes partition {} with {} events in {} table slices, ",
```

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,722 @@
...
+               *self, self->state.id_, self->state.num_rows_,
+               self->state.table_slices_.size());
+    auto c = chunk::make(buffer);
+    auto bytes = std::span<const std::byte>{c->data(), c->size()};
```
dominiklohmann, line 581: This appears to be unused.

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,722 @@
...
+                 *num_hits,
+                 vast::system::metrics_metadata{{"query", id_str},
+                                                {"store-type", "passive"}});
+      return num_hits;
+    },
+    [self](atom::erase, const ids& xs) -> caf::result<uint64_t> {
+      auto num_rows = rank(xs);
+      VAST_ASSERT(num_rows == 0
+                  || num_rows
+                       == static_cast<unsigned long>(self->state.num_rows_));
```
dominiklohmann, line 666: We have a `detail::narrow_cast` that wraps a `static_cast` and in debug mode checks if narrowing occurred.

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,722 @@
...
+  caf::error initialize(data options) override {
+    if (caf::holds_alternative<caf::none_t>(options))
+      return caf::none;
+    if (const auto* rec = caf::get_if<record>(&options))
+      if (rec->empty())
+        return caf::none;
+    return convert(options, config_);
+    // return caf::make_error(ec::invalid_configuration, //
+    //                        "expected empty configuration under "
+    //                        "vast.plugins.parquet-store");
```
dominiklohmann, line 137:
```suggestion
    return convert(options, config_);
```

This should do the trick already, I think.

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,722 @@
...
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2022 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <vast/arrow_table_slice_builder.hpp>
+#include <vast/chunk.hpp>
+#include <vast/concept/convertible/data.hpp>
+#include <vast/concept/convertible/to.hpp>
```
dominiklohmann, line 12: You're not using this and some other headers, e.g., base64, zip_iterator, convertible/to, io/write. Can you do a pass to remove unnecessary includes?

plugins/parquet-store/parquet_store.cpp
```diff
@@ -0,0 +1,722 @@
...
+#include <parquet/arrow/writer.h>
+
+#include <memory>
+#include <utility>
+
+namespace vast::plugins::parquet_store {
+
+/// Configuration for the Parquet plugin.
+struct configuration {
+  uint64_t row_group_size{parquet::DEFAULT_MAX_ROW_GROUP_LENGTH};
```
dominiklohmann, line 43: Can you add a test that this works?

dominiklohmann (approved):
Approving modulo outstanding comments. This is running fine, and the remaining things I've noticed are all minor. Looks great, and I think we can ship it as part of v2.1 actually. Really great work!

plugins/parquet-store/parquet_store.cpp
mavam: I would just use `VAST`.

plugins/parquet-store/parquet_store.cpp
dispanser: The arrow documentation is rather sparse on the possible error conditions. I chose `vast::ec::parse_error`, which hopefully catches the typical stuf happening when opening and reading parquet files (the bytes are already clear as the're coming from the file system actor).

changelog/unreleased/features/2284--parquet-store-plugin.md
dispanser: this measurement is only for the import, so the change shouldn't have an import. But thanks for the reminder; I intended to add some measurements for the read path as well, just didn't get to it yet.

plugins/parquet-store/parquet_store.cpp
dispanser: I used this [blog post](https://thedatafreak.medium.com/parquet-zstd-vs-gzip-740a83571ecd) as a baseline, but I'm adding a config option and just default to 9 instead.

plugins/parquet-store/README.md
```diff
@@ -0,0 +1,50 @@
+# Parquet Plugin for VAST
```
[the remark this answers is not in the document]
dispanser: @mavam , please review the readme

plugins/parquet-store/parquet_store.cpp
mavam: Can you add a comment that describes this? Otherwise someone else will have no clue how the magic numbers have been chosen.

dispanser:
now part of feather PR #2413

PR #2285, opened 2022-05-17
Support globs in CMake variable `VAST_PLUGINS`

dominiklohmann:
This allows for using non-recursive CMake globbing expression in the `VAST_PLUGINS` CMake cache variable. E.g., configuring a VAST build using `cmake -B build -S . -D VAST_PLUGINS=plugins/*` will build all bundled open-source plugins alongside VAST.

### :memo: Checklist

Developer-facing only; not sure if we need to put this somewhere into the docs.

### DIFF
--- a/vast/CMakeLists.txt
+++ b/vast/CMakeLists.txt
@@ -125,17 +125,30 @@ install(FILES "${PROJECT_SOURCE_DIR}/vast.yaml.example"
 
 set(VAST_PLUGINS
     ""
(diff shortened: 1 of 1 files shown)


dominiklohmann:
> Slightly tangential: is it worthwhile to offer an option that (1) enables all plugins in `plugins/` and (2) `example/plugins/`?

I'd consider that overkill now that you can just say `VAST_PLUGINS=plugins/*`.

PR #2292, opened 2022-05-21
Show GitHub and Slack icon instead text in docs

mavam:
This PR prettifies the navbar by changing the GitHub text to an icon.

I also upgraded from CSS to Sass and enabled the corresponding plugin. This promotes DRY CSS moving forward. E.g., we can now easily add more icons there if need be, e.g., one for Slack.

EDIT: I've also added the Slack logo.

EDIT2: There's now also an announcement bar to make it clearer that this site is still WIP.

Review by commit.

And try it locally, please.

Also, feedback would be appreciated on the colored Slack logo. Should it be monochrome instead?


docs/docusaurus.config.js
```diff
@@ -54,13 +58,26 @@ const config = {
             label: 'Changelog',
             position: 'left',
           },
+          {
+            href: 'http://slack.tenzir.com',
+            'aria-label': 'Slack',
```
dominiklohmann, line 63: :+1: to making this monochrome

dominiklohmann (approved):
Works nicely for me locally.

> Also, feedback would be appreciated on the colored Slack logo. Should it be monochrome instead?

I think that'd look better, as long as you can get that to work with dark mode correctly.

docs/docusaurus.config.js
mavam: Works now also in dark mode.

PR #2295, opened 2022-05-23
Fix crash in query evaluation for new partitions

dominiklohmann:
This fixes a rather annoying timing bug: We create new active partitions on demand, and immediately feed them events. If, however, a query arrives it may be scheduled before any of the events were added to the active partition, leading to it not having any indexers, which in turn leads to it not having a combined layout. The query evaluation wrongly assumed that every partition always had a combined layout, which caused a crash.

### :memo: Checklist

Follow the changed logic and verify that this was indeed an issue.


tobim:
I'll test this by adding a `return` after creating the first active partition.

PR #2299, opened 2022-05-23
Deduplicate explore hits by hash instead of ID

tobim:
The explore command can fire of multiple sub-queries after initial hits were produced for the given expression. When these sub-queries produce overlapping result sets we need to make sure that we don't present any events multiple times to the user. We achieve this by maintaining a set of already forwarded events in the explorer actor.

Before this change that set consisted of event ids, relying on the fact that those are globally unique. With this change the set to contain event hashes instead. That means we have to produce a hash of every result candidate, which is less efficient compared to the ids approach, but it shouldn't increase the delay noticeably for the case where we only produce a small number of intermediate results.

Changing the overall approach to wait for the initial query to complete and generating a single big expression from the intermediate results should be more efficient in general, but the change at hand is less invasive.

The last commit also removes the internal `preserve-ids` option from the export command. It wasn't used any more and keeping support for it would just cost developer time for no reason.

### :memo: Checklist

A style and `#include` review should suffice.

### DIFF
--- a/libvast/native-plugins/segment_store.cpp
+++ b/libvast/native-plugins/segment_store.cpp
@@ -147,24 +147,10 @@ caf::expected<uint64_t> handle_lookup(Actor& self, const vast::query& query,
       for (size_t i = 0; i < slices.size(); ++i) {
         const auto& slice = slices[i];
         const auto& checker = checkers[i];
-        if (extract.policy == query::extract::preserve_ids) {
-          for (auto& sub_slice : select(slice, ids)) {
-            if (query.expr == expression{}) {
-              self->send(extract.sink, sub_slice);
-              num_hits += sub_slice.rows();
-            } else {
-              auto hits = evaluate(checker, sub_slice);
-              num_hits += rank(hits);
-              for (auto& final_slice : select(sub_slice, hits))
-                self->send(extract.sink, final_slice);
-            }
-          }
-        } else {
-          auto final_slice = filter(slice, checker, ids);
-          if (final_slice) {
-            num_hits += final_slice->rows();
-            self->send(extract.sink, *final_slice);
-          }
+        auto final_slice = filter(slice, checker, ids);
+        if (final_slice) {
+          num_hits += final_slice->rows();
+          self->send(extract.sink, *final_slice);
         }
       }
     },
--- a/libvast/src/system/explorer.cpp
+++ b/libvast/src/system/explorer.cpp
@@ -219,7 +219,6 @@ explorer(caf::stateful_actor<explorer_state>* self, node_actor node,
         auto query = to_string(*expr);
         VAST_TRACE("{} spawns new exporter with query {}", *self, query);
         auto exporter_invocation = invocation{{}, "spawn exporter", {query}};
-        caf::put(exporter_invocation.options, "vast.export.preserve-ids", true);
         caf::put(exporter_invocation.options, "vast.export.max-events",
                  st.limits.per_result);
         ++self->state.running_exporters;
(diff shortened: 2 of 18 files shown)


libvast/native-plugins/segment_store.cpp
```diff
@@ -147,24 +147,10 @@ caf::expected<uint64_t> handle_lookup(Actor& self, const vast::query& query,
...
         const auto& checker = checkers[i];
-        if (extract.policy == query::extract::preserve_ids) {
-          for (auto& sub_slice : select(slice, ids)) {
-            if (query.expr == expression{}) {
-              self->send(extract.sink, sub_slice);
-              num_hits += sub_slice.rows();
-            } else {
-              auto hits = evaluate(checker, sub_slice);
-              num_hits += rank(hits);
-              for (auto& final_slice : select(sub_slice, hits))
```
dominiklohmann, line 158: Is `select` now dead code?

libvast/src/system/explorer.cpp
```diff
@@ -41,43 +41,36 @@ explorer_state::explorer_state(caf::event_based_actor*) {
...
-    }
+    default_hash h;
+    for (size_t j = 0; j < slice.columns(); ++j)
+      hash_append(h, slice.at(i, j));
+    auto digest = h.finish();
+    auto [_, new_] = returned_ids.insert(digest);
+    unseen.append_bits(new_, 1);
+    if (num_sent + rank(unseen) == limits.total)
+      break;
   }
```
dominiklohmann, line 61: Note that the summarize plugin already does the same thing for a subset of columns (those that we group by) but in a columnar fashion. Should we do that here as well?

tobim: Do you mean creating a "column" of hashers and moving them over the slice left to right?

dominiklohmann: Yes, and directly on the record batch as well. Pretty sure this isn't hard to write, the question is rather whether we want to spend the hour here.

libvast/native-plugins/segment_store.cpp
tobim: Almost, we still use it in `truncate` which is used in the sink. `truncate` could be changed easily though.

dominiklohmann: truncate should just use `arrow::RecordBatch::Slice` instead.

libvast/src/system/explorer.cpp
tobim: Imho we don't.

libvast/native-plugins/segment_store.cpp
dominiklohmann: Same deal with `split`. So just `to_record_batch`, and then re-create a table slice from the same layout and the sliced record batch (we have a constructor for that already). Shouldn't be hard too write, so I think it'd fit in the scope of this PR.

dominiklohmann (approved):
I verified locally that this deduplicates correctly. I don't have anything to add to the two comments I left earlier already, so I'm already leaving my approval.

Before merging I'd like to
1. see `select` go,
2. a follow-up story for removal of the ID assignment in the importer created, and
3. a TODO comment for the columnar iteration for the explore deduplication added.

libvast/native-plugins/segment_store.cpp
dominiklohmann: Ah damnit, so maybe an additional follow-up story for that then.

tobim:
> 1. see `select` go,

Sorry, can't do: https://github.com/tenzir/vast/pull/2299#discussion_r879554032

PR #2302, opened 2022-05-23
Add index metric for created active partitions

lava:
Add a new index metric for created active partitions, so it becomes possible to monitor how many new partitions are created over time.

### :memo: Checklist

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -505,6 +505,19 @@ void index_state::decomission_active_partition(const type& layout) {
       [=, this](partition_synopsis_ptr& ps) {
         VAST_DEBUG("{} successfully persisted partition {} {}", *self, layout,
                    id);
+        // Send a metric to the accountant.
+        if (accountant) {
+          auto report = vast::system::report {
+            .data = {
+              {"index.partitions-persisted", uint64_t{1}},
+              {"index.partitions-persisted-events", ps->events},
+            },
+            .metadata = {
+              {"layout", std::string{layout.name()}},
+            },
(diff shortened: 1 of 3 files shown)


libvast/src/system/index.cpp
```diff
@@ -664,6 +665,7 @@ void index_state::send_report() {
   auto query_counters = get_query_counters(pending_queries);
   auto msg = report{
     .data = {
+      {"index.active_partitions_created", counters.active_partitions_created},
```
dominiklohmann, line 668: I don't think this metric quite fulfills the needs I had when I created this internal feature request. Sorry for not making this clearer.

From a dashboarding perspective, I'd like to see metric emitted at the timepoint when a partition was rotated whose value is the number of events in the partition. The metadata should contain the schema name, so I can reliably observe whether partitions are rotated despite being underful, e.g., because the active partition timeout triggered.

libvast/src/system/index.cpp
```diff
@@ -672,6 +686,7 @@ void index_state::send_report() {
   auto query_counters = get_query_counters(pending_queries);
   auto msg = report{
     .data = {
+      {"index.partitions-created", counters.active_partitions_created},
```
dominiklohmann, line 689: What value does this metric add for the operator? It's just the number of active partitions created since the start of VAST, right?

dominiklohmann (approved):
This is nice. I question the value of one metric (personally I'd just drop it), and the other two require documentation on https://docs.tenzir.com/vast/features/metrics/

Thanks for this! 📈

PR #2303, opened 2022-05-23
Rework docs entry

mavam:
This PR streamlines the docs landing page under /docs. It is typical to begin with a "What is X?" section and then "Why X?" to help navigate the reader.

This PR also moves the quick start guide under Setup.

Build the page locally and give the first two sections a read.


web/docs/about-vast/README.md
```diff
@@ -0,0 +1,228 @@
...
+search engine uses (inverted) indexes and ranking methods to return the most
+relevant results for a given combination of search terms.
+
+:::note Recommendation
+However, most of the security telemetry arrives as structured log/event data, as
+opposed to unstructured textual data. If your primary use case involves working
+with text, VAST might not be a good fit. That said, it's still possible to
+leverage information retrieval techniques for security analytics. For example,
+Elastic is a popular SIEM that works well for medium-sized workloads but
+exhibits poor vertical scaling.
```
dispanser, line 146: It would really be great if we could cite benchmarks.

I found

https://elasticsearch-benchmarks.elastic.co/index.html#tracks/security/nightly/default/30d

Which is specifically aimed at security (last menu point on the left). Since they cite < 30k indexed documents on a 3-node-system, it would probably be easy to come up with a higher number (but of course we'd need a comparable data set).

dispanser (commented):
Made a few comments, specifically for the "Why not VAS" section.

web/docs/about-vast/README.md
dominiklohmann: Is the raw data available somewhere? This would make for a really good comparison docs page.

web/src/pages/index.tsx
```diff
@@ -24,7 +24,7 @@ function HomepageHeader() {
         <div className={styles.buttons}>
           <Link
             className="button button--secondary button--lg"
-            to="/docs/get-started">
+            to="/docs/about-vast">
             Get Started
```
dominiklohmann, line 28: I find it a bit weird to have a docs page labelled "Get Started" if this button with the same name doesn't link to it. Can we rename the "Get Started" docs page?

web/docs/about-vast/README.md
```diff
@@ -0,0 +1,228 @@
...
+VAST is an embeddable telemetry engine for structured event data, purpose-built
+for use cases in security operations. Consider VAST if you want to:
+
+- Build a federated detection and response architecture
+- Store, aggregate, and manage massive amounts of telemetry
+- Operationalize threat intelligence
+- Empower threat hunters
+- Engineer and deploy detections
+
+If you're unsure whether VAST is the right tool for your use case, keep reading.
```
dominiklohmann, line 20: Only the first sentence explains what VAST is. The remainder explains why you'd want to use it. That doesn't match the section title.

web/docs/about-vast/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Why not VAST?
+
+VAST is the right choice when you need to analyze large amounts of security
+telemetry for specific use cases in security operations. There exist a lot of
+database systems out there, and you may ask yourself the question: why not use
+that other system instead?
+
+To help answer this question, we offer a brief guidance when other systems might
+be a better fit.
```
tobim, line 109: The following sections don't actually match this description. Each subsection lays out how the discussed category of databases is not a good fit for the security use-case. All of those should be captioned `VAST vs. ...` and moved to the section above.

What belongs in here is:
* You require SQL
* You require RBAC
* You need a relevance score as part of your query results
* ??

Each section would then point the reader to the db-types that might be a better fit.

web/docs/about-vast/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+Timeseries databases share a lot in common with [OLAP
+engines](#vast-vs-data-warehouses), but put center data organization around
+time.
+
+:::note Recommendation
+If you plan to access your event data through time domain and need to model the
+majority of data as series, a timeseries DBs may suit the bill. If you access
+data through other (spatial) attributes, like IP addresses or domains, a
+timeseries DB might not be good fit—especially for high-cardinality
```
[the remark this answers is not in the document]
mavam: Agreed. IOx isn't really a timeseries DB anymore, I would argue. Maybe simply add "traditional" as qualifier? Or would that not be enough?

web/docs/about-vast/README.md
mavam: I might remove the relative comparison until we have benchmarks, otherwise we have no leg to stand on.

web/docs/about-vast/README.md
mavam: Good point. I took your approach and reworked the structure a bit. Essentially the "why not VAST" section is now gone.

web/README.md
```diff
@@ -1,10 +1,28 @@
-# VAST Docs
+# VAST Website
 
-This directory the [official VAST documentation][site], along with the
-[Docusaurus](https://docusaurus.io/) scaffold for local building, viewing, and
-deploying.
+This directory contains code and content of the official VAST website at
+https://vast.io. We use [Docusaurus](https://docusaurus.io/) as driving
+framework and an [enhanced flavor of Markdown](TODO) for primary content.
```
dominiklohmann, line 5: https://docs.github.com/en/get-started/writing-on-github is a good resource on GHFM.

web/README.md
```diff
@@ -26,8 +44,8 @@ yarn build
 
 The resulting `build` directory can be served using any static hosting service.
 
-We use a [github actions workflow][workflow] to deploy this site with every push
-to master.
+For the main site, we use a [github actions workflow][workflow] to deploy with
+every push to master.
 
 [site]: https://vast.io/
 [workflow]: https://github.com/tenzir/vast/blob/master/.github/workflows/docs.yaml
```
dominiklohmann, line 51: This link is outdated; we build as part of the VAST workflow instead, and deploy in a separate deployment environment: https://github.com/tenzir/vast/actions/workflows/pages/pages-build-deployment

web/README.md
```diff
@@ -26,8 +44,8 @@ yarn build
 
 The resulting `build` directory can be served using any static hosting service.
 
-We use a [github actions workflow][workflow] to deploy this site with every push
-to master.
+For the main site, we use a [github actions workflow][workflow] to deploy with
```
dominiklohmann, line 47: `s/github actions/GitHub Actions` for proper capitalization

web/docs/develop-vast/write-documentation.md
```diff
@@ -0,0 +1,72 @@
...
+We use [Docusaurus](https://docusaurus.io/) as site framework.
+
+### Build and View Locally
+
+To view the entire site (including the documentation) locally,
+change to the [`/web`](https://github.com/tenzir/vast/tree/master/web/)
+directory and invoke [yarn](https://yarnpkg.com/):
+
+```bash
+yarn
```
dominiklohmann, line 15: We should recommend using `yarn install --frozen-lockfile` over `yarn` so users don't auto-upgrade the lockfile depending on their globally installed packages.

web/README.md
```diff
@@ -1,10 +1,28 @@
-# VAST Docs
+# VAST Website
 
-This directory the [official VAST documentation][site], along with the
-[Docusaurus](https://docusaurus.io/) scaffold for local building, viewing, and
-deploying.
+This directory contains code and content of the official VAST website at
+https://vast.io. We use [Docusaurus](https://docusaurus.io/) as driving
```
dominiklohmann, line 4: Should we generally write links as `[vast.io](https://vast.io)` instead as that reads a bit nicer?

dominiklohmann (approved):
Some minor comments inline, otherwise I think we can merge this.

@dispanser's and @tobim's unresolved feedback should be addressed before merging or at the very least put into follow-up stories so we don't lose track of it.

web/docs/develop-vast/write-documentation.md
mavam: Does `yarn` without any arguments really do an auto-upgrade? Every time newer versions were out, I always a get notice in the shell, but the upgrade doesn't happen by default.

web/README.md
mavam: Yeah, but I'd rather get folks to read the Docusaurus docs, as there's more than GFM to consider.

web/README.md
mavam: I think it depends on the context. Here, the URI scheme helps understand where you land.

web/docs/about-vast/README.md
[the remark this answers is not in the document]
mavam: I tweaked it slightly, lemme know if this ain't cutting it.

web/docs/develop-vast/write-documentation.md
mavam: Per our discussion, I can see now why this is relevant.

PR #2305, opened 2022-05-24
Fall back to string when parsing config options from environment

dispanser:
Fix regression where setting `VAST_ENDPOINT=127.0.0.1:42000` wouldn't work because the parsing would fail.

### :memo: Checklist

It's a one-line change.

### DIFF
--- a/libvast/src/system/configuration.cpp
+++ b/libvast/src/system/configuration.cpp
@@ -261,10 +261,7 @@ caf::error merge_environment(record& config) {
                                            "variable {}={} to VAST value: {}",
                                            key, value, x.error()));
         } else {
-          return caf::make_error(ec::parse_error,
-                                 fmt::format("could not parse environment "
-                                             "variable {} value '{}': {}",
-                                             key, value, config_value.error()));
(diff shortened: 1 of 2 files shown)


libvast/src/system/configuration.cpp
```diff
@@ -261,10 +261,7 @@ caf::error merge_environment(record& config) {
                                            "variable {}={} to VAST value: {}",
                                            key, value, x.error()));
         } else {
-          return caf::make_error(ec::parse_error,
-                                 fmt::format("could not parse environment "
-                                             "variable {} value '{}': {}",
-                                             key, value, config_value.error()));
+          config[*config_key] = std::string{value};
```
dominiklohmann, line 264: I'm fine with the fix, but it needs a bug fix changelog entry.

dominiklohmann (approved):
Thanks for fixing this so promptly!

libvast/src/system/configuration.cpp
dispanser: made one. If you're fine with that one or have a wording proposal please take over the merge.

PR #2306, opened 2022-05-24
Unify extractor semantics

dominiklohmann:
As VAST developer and user, the notion of an "extractor" is not easy to grok. This PR is all about simplifying it.

### Rationale

Currently, there are multiple places in which we resolve a field name suffix, or a field name prefix, or a type name into a list of offsets. Just to name a few: almost every transform step plugin, candidate check, catalog lookup, matcher plugin, inventory plugin, table slice projection. Essentially all throughout our code base. Some of these support prefix matching, some suffix matching, some type resolution. Some support concept resolution. It's a mess, and the user never really knows what subset of functionality they can expect to work, when ideally we'd just support all of it everywhere.

### :memo: Checklist

As of opening this draft PR, I already have most of the required changes locally, but have yet to clean them up and sort them into nicely reviewable commits. Once that's done I'll ask for reviews. Until then it's probably not worth reviewing.


libvast/include/vast/type.hpp
```diff
@@ -346,6 +346,14 @@ class type final : public stateful_type_base {
   /// Returns all aliases of this type, excluding this type itself.
   [[nodiscard]] detail::generator<type> aliases() const noexcept;
 
+  /// Resolves an extractor on this type.
+  [[nodiscard]] detail::generator<offset>
+  resolve(std::string_view extractor) const noexcept;
+
+  /// Resolves an extractor on this type.
+  [[nodiscard]] detail::generator<offset>
+  resolve(std::vector<std::string_view> extractor) const noexcept;
```
mavam, line 355: Just curious from DX perspective because it's not (yet) documented: what's the difference in semantics over the singleton overload? Is the same as calling N times `resolve`, just faster?

dominiklohmann: Yes. I identified at some point during development that for transform steps especially we often resolved a bunch of extractors, sorted the results and then de-duplicated them. We also get this for free as the algorithm works on a vector of string views anyways.

This will become obvious once I've got the remaining bits and pieces in. I'm just extracting what I've got piece-by-piece while making finishing touches like writing documentation.

web/docs/understand-vast/data-model/extractors.md
```diff
@@ -0,0 +1,76 @@
...
+
+*Extractors* describe fields within a schema. The extractor resolution converts
+an extractor into an ordered list of *offsets*, which each uniquely describe a
+single (possibly nested) field in a given schema. The extractor resolution takes
+place once per schema. This enables a consistent way to specify which fields ot
+operate on throughout VAST across multiple schemas, e.g., in the [query
+language](/docs/understand-vast/query-language/) or the [index
+configuration](/docs/setup-vast/tune#tune-sketch-parameters).
+
+Let's look at how extractors work based on a subset of the `zeek.conn` schema:
```
mavam, line 11:
```suggestion
An *extractor* describes a set of record fields within a schema. Extractors
leverage the full power of VAST's type system and unify representation of fields
and types. They occur in the [query
language](/docs/understand-vast/query-language/) and the [configuration
file](/docs/setup-vast/tune#tune-sketch-parameters).

Internally, VAST resolves an extractor to an ordered list of *offsets*, each of
which uniquely describes a single field in the schema. Let's look at how
extractors work based on a subset of the `zeek.conn` schema:
```

web/docs/understand-vast/data-model/extractors.md
```diff
@@ -0,0 +1,76 @@
...
+
+```yaml
+# builtins.yaml
```
mavam, line 14: You could just use the module name here instead. I'm not sure if we agreed on a "global" namespace yet?

web/docs/understand-vast/data-model/extractors.md
```diff
@@ -0,0 +1,76 @@
...
+      - orig_pkts: count
+      # ... and a whole bunch of other fields
+```
+
+## Field Extractors
+
+Field extractors have the form `baz` or `foo.bar.baz`, where `foo`, `bar`, and
+`baz` describe field names. Field extractors match leaf fields by their
+(potentially nested) field name, and may optionally be prefixed with a type
+name to reduce redundancy.
```
mavam, line 42:
```suggestion
A *field extractor* has the form `z` or `x.y.z`, where `x`, `y`, and `z`
describe field names. Field extractors match leaf fields by their
(potentially nested) field name, and may optionally be prefixed with a type
name to disambiguate matches across multiple schemas.

```

mavam (commented):
Left some feedback on the docs only.

web/docs/understand-vast/data-model/extractors.md
dominiklohmann: It needs to be global if we want backwards compatibility, so in the interest of developer time we should have a global namespace rather than a `vast` namespace for the builtin types.

mavam: I can see the argument, but I don't think we've agreed to prioritize backwards compatibility yet.

dominiklohmann: I have yet to see a convincing argument _for_ breaking. The global namespace is consistent with our current naming. There's three ways to proceed here:

1. We use the global namespace as I suggested, and do not break anything.
2. We don't use a global namespace for built-in types and add a workaround for backwards compatibility. I see three options here currently:
   - We lazily add the `vast.` prefix when materializing table slices from the store, and when merging partition synopses in the catalog on startup.
   - We modify type equality comparison to no longer be just a call to `memcmp`, and instead do a deep equality comparison that accounts for this. Type hashing gets an additional branch to check for builtin types to keep the hashes consistent. Additionally, we would need to change the extractor resolution from this PR to account for the change.
   - We make a hard break and write a custom upgrade tool for the database.
3. We don't use a global namespace and make a hard break.

Of all these options only (1) sounds sensible to me given that we've managed to come by with a global namespace so far without any issues.

I think this discussion is very much relevant to this PR, and for anything other than option (1) I would want to see a convincing argument. Not just because of backwards compatibility, but also because of the development effort required for (2), given that we cannot afford to do (3) realistically.

dominiklohmann:
This is on hold now until I find some time to refactor how the catalog stores the partitions. I've added a `blocked` label.

PR #2308, opened 2022-05-25
Add a metric that counts false positive partitions

tobim:
### :memo: Checklist

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -616,6 +616,10 @@ void index_state::schedule_lookups() {
           [this, handle_completion, qid, pid = next->partition](uint64_t n) {
             VAST_DEBUG("{} received {} results for query {} from partition {}",
                        *self, n, qid, pid);
+            // No results indicate this partition was a false positive for the
+            // query.
+            if (n == 0)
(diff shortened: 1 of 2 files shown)


libvast/src/system/index.cpp
```diff
@@ -664,6 +668,7 @@ void index_state::send_report() {
   auto query_counters = get_query_counters(pending_queries);
   auto msg = report{
     .data = {
+      {"catalog.lookup.false-positives", counters.partition_false_positives},
```
dominiklohmann, line 671: Can you add the query metadata to this metric?

dominiklohmann (approved):
Fine by me, modulo yet missing documentation. Should we roll this out to the testbed to see how we can plot this?

libvast/src/system/index.cpp
tobim: I want to do that. But it requires us to issue it from the query completion function instead. I hope I can get to this on the sooner side.

dominiklohmann:
Closing this as we've identified this approach not to work with the current architecture. The planned catalog centralization will make this easier, and then we'll get back to this.

PR #2309, opened 2022-05-26
Add CMake presets for Nix based development

tobim:
Add some CMake presets i've been using to the repo.

Once you are in a `nix-shell` or `nix develop`  based environment you can use `cmake --preset nix-{clang,gcc}-{debug,redeb,release}` to configure build trees and `cmake --build --preset nix-{clang,gcc}-{debug,redeb,release}`

### :memo: Checklist


CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+        "CMAKE_CXX_COMPILER_LAUNCHER": "ccache",
+        "CMAKE_FIND_FRAMEWORK": "LAST",
+        "CMAKE_FIND_PACKAGE_PREFER_CONFIG": "ON",
+        "CMAKE_FIND_USE_SYSTEM_PACKAGE_REGISTRY": "OFF",
+        "CMAKE_FIND_USE_PACKAGE_REGISTRY": "OFF",
+        "CMAKE_OSX_SYSROOT": "",
+        "VAST_ENABLE_DEVELOPER_MODE": "ON",
+        "VAST_ENABLE_LSVAST": "ON",
+        "VAST_ENABLE_MDX_REGENERATE": "ON",
+        "VAST_PLUGINS": "${sourceDir}/plugins/broker;${sourceDir}/plugins/pcap;${sourceDir}/plugins/sigma;${sourceDir}/plugins/summarize"
```
mavam, line 23: Is there a way to generate this dynamically based on the source tree, so that it doesn't have to be updated manually?

(Applies to all occurrences of `VAST_PLUGINS`.)

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+      "inherits": ["nix-base", "gcc-base"],
+      "displayName": "GCC Release Config",
+      "binaryDir": "${sourceDir}/build/gcc/release",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Release",
+        "CMAKE_INSTALL_PREFIX": "${sourceDir}/install/gcc/release",
+        "VAST_ENABLE_BUNDLED_CAF": "OFF"
+      },
+      "environment": {
+        "LDFLAGS": "-fuse-ld=gold"
```
mavam, line 121: Perhaps `gold` too opinionated as a default?

nix/README.md
```diff
@@ -0,0 +1,50 @@
+# Nix build support facilities
```
mavam, line 1: Can you add the meat of this to https://vast.io/docs/setup-vast/build#nix? In other places, we also cross-reference the main docs page from the repo.

nix/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Development Environment
+
+The dependencies for a dynamic built can be fetched with
```
mavam, line 5: As a reader, it would be nice to get a little intro paragraph on what this is about and why I want it. Especially because nix is so mysterious for most people.

nix/README.md
```diff
@@ -0,0 +1,50 @@
...
+```sh
+nix develop
+```
+
+The `direnv` tool has good support for nix and can be used with an `.envrc` with
+the content:
```
mavam, line 11: Why would I do that? It would be nice to shed a bit of insight for the use case here, e.g., "to do X, use Y."

mavam (commented):
@dispanser is probably the best for the functional tests. From my side:

1. It would be nice to add the to the website
2. Help the user a bit more with the "whys" behind the actions

CMakePresets.json
dominiklohmann:
```suggestion
        "VAST_PLUGINS": "${sourceDir}/plugins/*"
```

This already works.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+      "cacheVariables": {
+        "CMAKE_CXX_COMPILER_LAUNCHER": "ccache",
+        "CMAKE_FIND_FRAMEWORK": "LAST",
+        "CMAKE_FIND_PACKAGE_PREFER_CONFIG": "ON",
+        "CMAKE_FIND_USE_SYSTEM_PACKAGE_REGISTRY": "OFF",
+        "CMAKE_FIND_USE_PACKAGE_REGISTRY": "OFF",
+        "CMAKE_OSX_SYSROOT": "",
+        "VAST_ENABLE_DEVELOPER_MODE": "ON",
+        "VAST_ENABLE_LSVAST": "ON",
+        "VAST_ENABLE_MDX_REGENERATE": "ON",
```
dominiklohmann, line 22: It's `s/MDX_REGENERATE/VAST_REGENERATE`. I also don't think that tool should be enabled by default. We don't have any tests for it and it is highly dangerous to use, so we should never be shipping it with any VAST packages.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "CMAKE_INSTALL_PREFIX": "${sourceDir}/install/clang/debug",
+        "VAST_ENABLE_ASAN": "ON",
+        "VAST_ENABLE_BUNDLED_CAF": "ON",
+        "VAST_PLUGINS": "${sourceDir}/plugins/pcap;${sourceDir}/plugins/sigma;${sourceDir}/plugins/summarize"
+      }
+    },
+    {
+      "name": "nix-clang-redeb",
```
dominiklohmann, line 51: Why `redeb` rather than `RelWithDebInfo` or the lowercase `relwithdebinfo`?

I think it's better to stick close to the defaults here.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+    },
+    {
+      "name": "nix-clang-debug",
+      "inherits": ["nix-base", "clang-base"],
+      "displayName": "Clang Debug Config",
+      "binaryDir": "${sourceDir}/build/clang/debug",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "CMAKE_INSTALL_PREFIX": "${sourceDir}/install/clang/debug",
+        "VAST_ENABLE_ASAN": "ON",
```
dominiklohmann, line 45: That's already the default for Debug builds. No need to explicitly specify it here.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+        "VAST_ENABLE_BUNDLED_CAF": "ON",
+        "VAST_PLUGINS": "${sourceDir}/plugins/pcap;${sourceDir}/plugins/sigma;${sourceDir}/plugins/summarize"
+      }
+    },
+    {
+      "name": "nix-clang-redeb",
+      "inherits": ["nix-base", "clang-base"],
+      "displayName": "Clang Redeb Config",
```
dominiklohmann, line 53:
```suggestion
      "displayName": "Clang RelWithDebInfo Config",
```

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+        "CMAKE_CXX_COMPILER": "clang++"
+      },
+      "environment": {
+        "LDFLAGS": "-fuse-ld=lld"
+      }
+    },
+    {
+      "name": "nix-clang-debug",
+      "inherits": ["nix-base", "clang-base"],
+      "displayName": "Clang Debug Config",
```
dominiklohmann, line 40:
```suggestion
      "displayName": "Nix Clang Debug Config",
```

Since it inherits from the nix-base it should also specify that it's for Nix in the display name. The same applies to all the other ones below.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+      "binaryDir": "${sourceDir}/build/clang/debug",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "CMAKE_INSTALL_PREFIX": "${sourceDir}/install/clang/debug",
```
dominiklohmann, line 44: The modern CMake way of specifying the install prefix is at install time via `--prefix path/to/install` rather than specifying the prefix at build time.

If we put this in our presets then we should at the very least add it to our gitignore, although I think it'd be much better to just not specify this at all.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+    },
+    {
+      "name": "nix-clang-redeb",
+      "inherits": ["nix-base", "clang-base"],
+      "displayName": "Clang Redeb Config",
+      "binaryDir": "${sourceDir}/build/clang/redeb",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "CMAKE_INSTALL_PREFIX": "${sourceDir}/install/clang/redeb",
+        "VAST_ENABLE_BUNDLED_CAF": "OFF"
```
dominiklohmann, line 58: Why do we disable that here? If we enable it for Debug builds because of the added debug information, then we should also enable it for RelWithDebInfo builds for for the added debug information.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+      "name": "nix-gcc-debug",
+      "inherits": ["nix-base", "gcc-base"],
+      "displayName": "GCC Debug Config",
+      "binaryDir": "${sourceDir}/build/gcc/debug",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "CMAKE_INSTALL_PREFIX": "${sourceDir}/install/gcc/debug",
+        "VAST_ENABLE_ASAN": "ON",
+        "VAST_ENABLE_BUNDLED_CAF": "ON",
+        "VAST_PLUGINS": "${sourceDir}/plugins/pcap;${sourceDir}/plugins/sigma;${sourceDir}/plugins/summarize"
```
dominiklohmann, line 90: These are duplicated between gcc-base and clang-base; should they be merged into a hidden debug-base? The same applies to the other build configurations.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+      "cacheVariables": {
+        "BUILD_SHARED_LIBS": "OFF",
+        "BUILD_TESTING": "OFF",
+        "CMAKE_C_COMPILER": "x86_64-unknown-linux-musl-gcc",
+        "CMAKE_CXX_COMPILER": "x86_64-unknown-linux-musl-g++",
+        "CMAKE_SYSTEM_NAME": "Linux",
+        "CMAKE_SYSTEM_PROCESSOR": "x86_64",
+        "CPACK_GENERATOR": "DEB",
+        "CPACK_PACKAGING_INSTALL_PREFIX": "/opt",
+        "CPACK_PACKAGE_CONTACT": "engineering@tenzir.com",
```
dominiklohmann, line 137: This does not belong in here. I already commented on this above, but this should really be in the top-level CMakeLists.txt file directly behind an option that we can then enable here.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+      "name": "nix-static",
+      "inherits": ["nix-base"],
+      "hidden": true,
+      "cacheVariables": {
+        "BUILD_SHARED_LIBS": "OFF",
+        "BUILD_TESTING": "OFF",
+        "CMAKE_C_COMPILER": "x86_64-unknown-linux-musl-gcc",
+        "CMAKE_CXX_COMPILER": "x86_64-unknown-linux-musl-g++",
+        "CMAKE_SYSTEM_NAME": "Linux",
+        "CMAKE_SYSTEM_PROCESSOR": "x86_64",
```
dominiklohmann, line 134: Why do we need to set these two manually?

CMakePresets.json
tobim: [Until mold is wrapped to be used with nixpkgs](https://github.com/NixOS/nixpkgs/pull/172452) we must choose between bfd and gold for linking LTO-enabled object files from GCC.
bfd is still the default, but I don't see a reason not to use gold on Linux.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+        "CPACK_GENERATOR": "DEB",
+        "CPACK_PACKAGING_INSTALL_PREFIX": "/opt",
+        "CPACK_PACKAGE_CONTACT": "engineering@tenzir.com",
+        "VAST_ENABLE_BACKTRACE": "ON",
```
dominiklohmann, line 138: This is already the default.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+        "VAST_ENABLE_BUNDLED_CAF": "OFF",
+        "VAST_ENABLE_JEMALLOC": "ON",
+        "VAST_ENABLE_RELOCATABLE_INSTALLATIONS": "ON",
```
dominiklohmann, line 141: Similarly, this also already is the default.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+    },
+    {
+      "name": "nix-gcc-static-release",
+      "inherits": ["nix-static"],
+      "displayName": "GCC Static Release Config",
+      "binaryDir": "${sourceDir}/build/gcc-static/release",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Release",
+        "CMAKE_INSTALL_PREFIX": "${sourceDir}/install/gcc-static/release",
+        "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "ON"
```
dominiklohmann, line 174: This is already the default for Release builds.

dominiklohmann (changes requested):
I have lots of small inline notes on the CMake cache variables you're setting. I would like to avoid duplication as much as possible; business logic should stay inside the CMakeLists.txt files rather than in the presets and the presets should not explicitly specify what's already the default.

CMakePresets.json
tobim: Not that I know of. At least on in this file. We could add an `VAST_ENABLE_BUNDLED_PLUGINS` option to CMakeLists.txt. @dominiklohmann will probably have an opinion on that.

CMakePresets.json
dominiklohmann: This build is not restricted to Linux, so that'd be one reason not to specify it.

CMakePresets.json
dominiklohmann: What I specified already works, we expand the list of plugins via globbing.

nix/README.md
tobim: I don't think we should do this in our own docs. Do we explain why the user should use apt or brew? no.

nix/README.md
tobim: Can do.

dominiklohmann (commented):
Also just wanna say: Thanks for pushing this. I've been wanting this for a long time (basically since the feature came out), and it's great to see this happening! ❤️

CMakePresets.json
tobim: We get an ICE during optimization without it. Unfortunately I can't add a comment about that to this file.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+      "name": "nix-base",
+      "hidden": true,
+      "generator": "Ninja",
+      "cacheVariables": {
+        "CMAKE_CXX_COMPILER_LAUNCHER": "ccache",
+        "CMAKE_FIND_FRAMEWORK": "LAST",
+        "CMAKE_FIND_PACKAGE_PREFER_CONFIG": "ON",
+        "CMAKE_FIND_USE_SYSTEM_PACKAGE_REGISTRY": "OFF",
+        "CMAKE_FIND_USE_PACKAGE_REGISTRY": "OFF",
+        "CMAKE_OSX_SYSROOT": "",
```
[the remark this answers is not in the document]
tobim: These settings improve reproducibility, they are sensible defaults for using CMake with Nix and are taken from the [nixpkgs stdenv hook for CMake](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/build-managers/cmake/setup-hook.sh).

CMakePresets.json
tobim: Static binaries only make sense for Linux, but you're right we probably shouldn't hard-code `x86_64`.

dominiklohmann: We should probably add something like this to our CMake then.

```cmake
if (VAST_ENABLE_STATIC_BINARY AND CMAKE_SYSTEM_NAME NOT EQUAL "Linux")
  message(FATAL_ERROR ...)
endif ()
```

CMakePresets.json
tobim: These presets are for development, not distribution.

CMakePresets.json
```diff
@@ -0,0 +1,216 @@
...
+    "minor": 21,
+    "patch": 0
+  },
+  "configurePresets": [
+    {
+      "name": "nix-base",
+      "hidden": true,
+      "generator": "Ninja",
+      "cacheVariables": {
+        "CMAKE_CXX_COMPILER_LAUNCHER": "ccache",
```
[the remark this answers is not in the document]
tobim: ccache is a part of the `nix develop` closure. Enabling it doesn't have any drawbacks in my eyes. I do see a problem when the user has something else in the environment variable though.

CMakePresets.json
tobim: It is `MDX_REGENERATE` btw.

dominiklohmann: https://github.com/tenzir/vast/pull/2260 changes that, so please at least leave a comment in there to change the preset accordingly.

I don't see how this would only be for development and not for distribution. If I build the preset `nix-static-release` I get this, and that's exactly what we distribute.

CMakePresets.json
dominiklohmann: What do you think about just making it the default, i.e., not adding a `-redeb/-relwithdebinfo` suffix at all?

CMakePresets.json
tobim: 1. Having a CAF with debug info doesn't add that much value in redeb mode, doing that only for full Debug builds is enough.
2. Switching to Bundled CAF would mean disabling the broker plugin, and then you have to build in Release mode again when you want to check whether the plugin still works.

dominiklohmann: I forgot about that aspect of the Broker plugin. Should we even use the bundled CAF for Debug builds then? It probably makes more sense to just never use it so all presets support every plugin.

CMakePresets.json
tobim: The only drawback of gold is that it can only produce ELF binaries, but the list of systems that support ELF is quite long:
https://en.wikipedia.org/wiki/Executable_and_Linkable_Format#Applications

gold is a very sensible default.

CMakePresets.json
tobim: I don't like that idea because it breaks the scheme. How would we name the subdirectoy for the build tree? Having `build/clang` for this but then also `build/clang/debug` for the debug build is very messy.

CMakePresets.json
tobim: > I don't see how this would only be for development and not for distribution. If I build the preset nix-static-release I get this, and that's exactly what we distribute.

The package we distribute is not built with these presets but via https://github.com/tenzir/vast/blob/master/nix/vast/default.nix. If that file starts making use of the preset we can still add overrides with the `cmakeFlags` variable.

CMakePresets.json
tobim: Hm, I don't know. Being able to jump to definition of CAF function is valuable to me. That doesn't work with prebuilt versions.

nix/README.md
mavam: I'm not saying we should write documentation for other tools.

My point is: simply tell the user why I should care. I've never heard of `direnv`. Why do I care and need it here?

CMakePresets.json
tobim: I don't want to change my workflow just because Kitware added another runmode to CMake. Why should I be forced to use this and type in the prefix path every time?
And besides, according to the docs you can still use `--prefix` if you want even if `CMAKE_INSTALL_PREFIX` was used during configuration.`

Adding `/install` to gitignore is a good idea though.

dominiklohmann: Fine by me if you add it to the gitignore file.

CMakePresets.json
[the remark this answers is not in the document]
dominiklohmann: Really wondering about the sysroot being disabled as that breaks the build for me when not using Nix, but I assume that's okay then. Feel free to resolve.

CMakePresets.json
tobim: That is why the debug presets exclude the broker plugin.
Maybe we can use a different method to detect the issue. It should work of the CAF used by broker is the same version as the bundled one.

dominiklohmann: In `plugins/broker/CMakeLists.txt`, change this:

```cmake
if (VAST_ENABLE_BUNDLED_CAF)
  message(FATAL_ERROR "Cannot use bundled CAF with Broker plugin")
endif ()
```

to

```cmake
if (VAST_ENABLE_BUNDLED_CAF)
  message(WARNING "Cannot use bundled CAF with Broker plugin")
  return()
endif ()
```

And move it between the `find_package(VAST)` and `VASTRegisterPlugin(...)` calls. That should already do it.

web/docs/setup-vast/download.md
```diff
@@ -60,6 +60,16 @@ command:
...
 
+### Nix
+
+Our repository contains a `flake.nix` that provides a VAST as an app, you can
+use `vast = "github:tenzir/vast/stable"` as an input in your own flake or just
+try it out with:
+
+```bash
+nix run github:tenzir/vast/stable
+```
```
tobim, line 71: Caveat for the reviewer: This won't work until the stable branch includes these commits. Remove the `/stable` and it should work.

CMakePresets.json
dominiklohmann: Still not convinced by this one, so maybe it's one for democracy, i.e., ask people in Slack to comment on this thread. To me, `redeb` is inconsistent and just adds mental overhead since I already know `RelWithDebInfo`.

dominiklohmann (approved):
Approving the CMake bits modulo the yet unresolved comments. 🚀

Please also get an additional review from someone that uses Nix.

CMakePresets.json
lava: Unresolving this because the other PR is now merged, so we have to `s/MDX_REGENERATE/VAST_REGENERATE`

CMakePresets.json
tobim: `RelWithDebInfo` came out the winner in an internal vote.

dispanser (approved):
I tried this out and it's really a nice experience.

For the first time, I compiled VAST using clang instead, which appears to be significantly faster on my machine :100: 

I tried `nix-{clang,gcc}-{release,debug}` combinations.

PR #2313, opened 2022-05-27
Render CMake build summary on STATUS log level

dominiklohmann:
Our Debian CI did not show the build summary, unlike the macOS CI and local builds. This happens because CMake itself on Debian appears to hide log output below STATUS level by default for whatever reason.

### :memo: Checklist

Check that it now renders in CI.


tobim (approved):
I was curious how the FeatureSummary module managed to write it's content without the dashes. Turns out it builds a big string with all its lines and prints a single message at the end.
I would say the fix here is good enough.

tobim:
I should have known you would change it. You missed the compiler messages. Sorry for the late notice.

dominiklohmann:
Damnit. Will change.

PR #2314, opened 2022-05-27
Mark erased partitions as such in the query queue

tobim:
We don't remove the partition from the queue directly because the query API requires clients to keep track of the number of candidate partitions. Removing the partition from the queue would require us to update the partition counters in the query states but the client would go out of sync. That would require the index to deal with a few complicated corner cases.
A partition that failed to load but was not erased can now be logged with a proper warning thanks to this addition.

### :memo: Checklist

This isn't really testable in the unit test framework because the only observable effect is a log message. Just read the code changes and see if you can find flaws.

### DIFF
--- a/libvast/include/vast/query_queue.hpp
+++ b/libvast/include/vast/query_queue.hpp
@@ -53,6 +53,7 @@ public:
     uuid partition;
     uint64_t priority = 0;
     std::vector<uuid> queries;
+    bool erased = false;
 
     friend auto operator<=>(const entry& lhs, const entry& rhs) {
       return lhs.priority <=> rhs.priority;
@@ -93,6 +94,9 @@ public:
 
   /// Removes a query from the queue entirely.
   [[nodiscard]] caf::error remove_query(const uuid& qid);
+
+  /// Removes a partition from the queue.
+  bool mark_partition_erased(const uuid& pid);
 
   /// Retrieves the next partition to be scheduled and the related queries and
   /// increments the scheduled counters for the latter.
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -547,6 +547,16 @@ void index_state::schedule_lookups() {
       VAST_DEBUG("{} did not find a partition to query", *self);
       return;
     }
+    auto immediate_completion = [&](const query_queue::entry& x) {
+      for (auto qid : x.queries)
+        if (auto client = pending_queries.handle_completion(qid))
+          self->send(*client, atom::done_v);
+    };
+    if (next->erased) {
+      VAST_DEBUG("{} skips erased partition {}", *self, next->partition);
+      immediate_completion(*next);
+      continue;
+    }
     VAST_DEBUG("{} schedules partition {} for {}", *self, next->partition,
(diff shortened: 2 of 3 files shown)


libvast/src/system/index.cpp
```diff
@@ -569,9 +569,9 @@ void index_state::schedule_lookups() {
                  it != persisted_partitions.end())
           part = inmem_partitions.get_or_load(partition_id);
       }
-      if (!part)
-        VAST_VERBOSE("{} failed to load partition {} that was part of a query",
-                     *self, partition_id);
+      if (!part && !next->erased)
```
dominiklohmann, line 572: Can we skip the loading attempt directly above if the partition is marked erased?

libvast/include/vast/query_queue.hpp
```diff
@@ -53,6 +53,7 @@ class query_queue {
     uuid partition;
     uint64_t priority = 0;
```
dominiklohmann, line 54: When calculating the priority of a query we should take into account whether a partition is erased or not. An erased partition should likely have a priority of 0.

dominiklohmann (approved):
This is great, thanks! I agree with the approach to tombstone rather than actually erase the partitions.

libvast/include/vast/query_queue.hpp
tobim: I'm not sure adding such an effect is a good idea. That would mean erasing a partition prolongs related queries while we already know there won't be any more results. I'd rather not touch the priority here.

dominiklohmann: I can see both sides. Fine by me. Can you add a comment on why we don't touch the priority in the function that marks a partition erased for future reference?

tobim: I don't think it makes sense to comment about why something is not done unless the reader would totally expect that thing. In this case a comment would be warranted if it would adjust the priority.

Also, a changing the priority here would require us to sort as well, otherwise the effect would be very unpredictable.

lava: If we would update priority, wouldn't it make more sense anyways to set it to "infinite" rather than zero? Since it costs nothing to evaluate an erased partition, we should do so as soon as possible.

tobim: i thought about that as well, but let's weigh cost vs benefit.

The benefit is that a backlogged query that gets a candidate erased would be able to finish a bit faster _iff the erased partition is the one with the lowest prio of that query_.

The cost is that implementing a "set it to infinite" is not straight forward, because the current priority type is `uint64_t` which does not have a notion of "infinite", and wraps around. It becomes a game of tradeoffs, the options are:
1. Check for `erased` before updating the priority in all places - brittle
2. Use `double` instead - IEEE 754 is a mess but probably the sanest option
3. Choose a reasonably high value that is still higher than what you can reasonably expect from normal operation - also brittle, what if we need to scale the range at some point?
4. Implement a custom integral wrapper that prevents the wraparound - high effort and future maintenance

I'm going to close this discussion now because we've already spent too much time on this minor detail.

PR #2317, opened 2022-05-30
Enable conversions from data and type to Arrow Scalar

dominiklohmann:
This adds support for converting a matching type and data pair into an Arrow Scalar. The conversion is only implemented in one direction for now until we have a use case for the other direction as well.

This solves a need from @dispanser for #2284 and from @mavam for #2058.

### :memo: Checklist

Developer-facing only. This should be covered by the additional tests.

### DIFF
--- a/libvast/src/view.cpp
+++ b/libvast/src/view.cpp
@@ -8,6 +8,7 @@
 
 #include "vast/view.hpp"
 
+#include "vast/arrow_table_slice_builder.hpp"
 #include "vast/detail/narrow.hpp"
 #include "vast/detail/overload.hpp"
 #include "vast/operator.hpp"
@@ -248,7 +249,7 @@ bool type_check(const type& x, const data_view& y) {
         return false;
       for (size_t i = 0; const auto& [k, v] : u) {
         const auto field = t.field(i++);
-        if (field.name != k || type_check(field.type, v))
+        if (field.name != k || !type_check(field.type, v))
           return false;
       }
       return true;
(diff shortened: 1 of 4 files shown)


libvast/src/view.cpp
```diff
@@ -388,4 +389,99 @@ data_view to_internal(const type& t, const data_view& x) {
...
+  VAST_ASSERT(type_check(type, view));
+  auto f = detail::overload{
+    [&]<concrete_type Type>(
+      [[maybe_unused]] const Type& t) -> std::shared_ptr<arrow::Scalar> {
+      if constexpr (detail::is_any_v<Type, map_type, enumeration_type,
+                                     subnet_type>) {
+        // TODO: This is slow because we create a single-element Array and then
+        // use that API to get a scalar for the first element, but I cannot
+        // figure out how to create a scalar for map, enumeraion, or subnet
+        // types directly. -- DL
```
dispanser, line 404: since a subnet is mostly just a struct with `address` and `length` fields, we might be able to use the same approach used for `record_type` . For the others, I have no clue :shrug:

dispanser (approved):
Looks good!

I'm fine with merging this, but there's no immediate need that it solves, and I'm a bit worried about  the performance implications for the array-builder based code path.

dominiklohmann:
I'm closing this. There are too many performance implications with the approach chosen, and we lost the one use case we had because we solved it differently in the meantime. Should we ever need this again we can always pick it up, but as-is this is adding dead code.

PR #2318, opened 2022-05-30
Don't exceed max partition size in partition transformer

lava:
NOTE: The code compiles and looks kinda good, but hasn't been tested at all so far.

### :memo: Checklist


tobim (commented):
I think this solution is satisfactory, but it we should adapt one of the unit tests to make sure it works as expected.

lava:
I actually moved these changes over into #2277, since I retroactively found some issues that were already present there and I didn't want to separate them back out and rebase both PRs.

PR #2321, opened 2022-06-01
Add a `rebuild` command plugin

dominiklohmann:
The new plugin rebuilds the whole database by running an `identity` transformation on the database, causing all partitions to be rebuilt entirely. This conveniently merges partitions up to the configured `max-partition-size`, upgrades to the most recent table slice encoding, and migrates all data to the currently configured store backend.

### :memo: Checklist

Test it out locally. The testing of the individual pieces is already covered, and integration testing is impossible until we have a `flush` command (next Hackathon?).

For the code I recommend just reading the whole diff at once.

For stress-testing, I recommend running VAST with a low `vast.active-partition-timeout`, ingesting a bunch of data, and then looking at the results of rebuilding.


libvast/fbs/partition_synopsis.fbs
```diff
@@ -13,6 +13,9 @@ table LegacyPartitionSynopsis {
 
   /// The import time range of this partition.
   import_time_range: vast.fbs.interval;
+
+  /// The version number of this partition.
```
lava, line 17: Can we expand a bit here? This is likely to be encountered in the future by someone without the context of the `rebuild` command, who will wonder why we have a separate version number in addition to the flatbuffer versioning itself.

libvast/include/vast/system/actors.hpp
```diff
@@ -275,6 +279,10 @@ using index_actor = typed_actor_fwd<
   // preserving the old one(s).
   caf::replies_to<atom::apply, transform_ptr, std::vector<uuid>,
                   keep_original_partition>::with<std::vector<partition_info>>,
+  // A shorthand to the above handler that runs an identity transformation on
+  // the provided partitions without keeping the originals.
```
lava, line 283: Imho the main reason for having a separate handler is that it allows a *client command* to trigger the transform, just having a shorthand would probably not justify a separate handler.

libvast/src/system/catalog.cpp
```diff
@@ -405,8 +403,17 @@ catalog(catalog_actor::stateful_pointer<catalog_state> self,
     },
     [=](atom::candidates, vast::uuid lookup_id,
         const vast::expression& expr) -> caf::result<catalog_result> {
+      return self->delegate(static_cast<catalog_actor>(self),
+                            atom::candidates_v, lookup_id, expr,
+                            defaults::latest_partition_version);
```
lava, line 408: That may actually give missing results if you run VAST on a db created by a newer version. Not sure if we want to support that, but we can avoid the issue by passing an explicit `optional<uint64_>` as max version.

lava (approved):
Looks great to me in general, and we already tested it on the testbed so we also know that it actually works.

libvast/src/system/catalog.cpp
dominiklohmann: I added a comment for this. I don't think it's an actual issue in practice: If a user downgrades from a version with a changed latest partition version we're unlikely to be able to read the newer partitions anyways.

dominiklohmann: Actually, the fix is rather easy: We can just pass the max value.

libvast/fbs/partition_synopsis.fbs
dominiklohmann: I added a bit of meat here, explaining why we need this.

web/docs/setup-vast/tune.md
```diff
@@ -170,3 +170,38 @@ be rotated.
...
+in order to minimize resource usage. Setting the step size to a larger value
+allows for merging undersized partitions.
+
+Terminating the rebuild command will cause the VAST server to continue
+rebuilding all currently in-flight partitions.
+
+An optional expression allows for filtering partitions with an expression. The
+partition-level lookup may have false-positives. For example, to rebuild all
+outdated partitions containing `suricata.flow` events older than 2 weeks, run
+`vast rebuild '#type == "suricata.flow" && #import_time < 2 weeks ago'`.
```
dominiklohmann, line 207: @mavam please review this new docs section. The rest was already reviewed and approved.

web/docs/setup-vast/tune.md
```diff
@@ -170,3 +170,38 @@ be rotated.
...
+recent internal batch encoding and indexes.
+
+This is how you run it:
+
+```bash
+vast rebuild [--all] [--step-size=<number>] [--parallel=<number] [<expression>]
+```
+-
+The `--all` flag makes the rebuild command consider _all_ partitions rather than
+only outdated ones. This is useful when you change configuration options and
```
mavam, line 192: Please defined "outdated". Is this time-based? Is there a version?

web/docs/setup-vast/tune.md
```diff
@@ -170,3 +170,38 @@ be rotated.
...
+want to regenerate all partitions.
+
+The `--step-size` and `--parallel` options are performance tuning knobs. The
+step size controls the number of partitions to transform at once, and the
+parallelism level controls how many runs to start in parallel. Both default to 1
+in order to minimize resource usage. Setting the step size to a larger value
+allows for merging undersized partitions.
```
mavam, line 199: I'm struggling with the UX: shouldn't the system figure out the step size? Otherwise we have a correctness issue. The use can accidentally provide step size that's too small and not get the desired effect. For example, if the partition size is currently `S`, and the new size `N` where `N >> S`, then step size `p` must be chosen such that `p * S >= N`. How can the user know this?

web/docs/setup-vast/tune.md
```diff
@@ -170,3 +170,38 @@ be rotated.
...
+
+Terminating the rebuild command will cause the VAST server to continue
+rebuilding all currently in-flight partitions.
```
mavam, line 202: Please clarify what "in-flight" partition means here.

web/docs/setup-vast/tune.md
mavam, line 207:
```suggestion
An optional expression allows for restricting the set of partitions to rebuild.
VAST performs a catalog lookup with expression to identify the set of candidate
partitions. This process may yield false positives, as with regular queries, and
in this case cause potentially unaffected partitions to undergo a rebuild process.
For example, to rebuild all outdated partitions containing `suricata.flow` events
older than 2 weeks, run the following command:

```bash
vast rebuild '#type == "suricata.flow" && #import_time < 2 weeks ago'
```
```

**Review by mavam — COMMENTED** [review]
It's good to see now docs being updated inline with the corresponding feature!

Some general, well-intended writing feedback: you can improve your documentation game by answering the "why would I want this?" question first. When presenting a feature, tell the user what problems or why they need it. Thereafter delve into how the feature works.

File: libvast/vendored/indicators/indicators.hpp [hunk 8]
```diff
@@ -0,0 +1,4648 @@
+
```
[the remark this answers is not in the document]
tobim: See https://github.com/tenzir/vast/pull/2327.

web/docs/setup-vast/tune.md
dominiklohmann: I discussed this with @tobim and we came up with a few improvements.

First, we changed the catalog lookup to return partitions sorted by their import time range, with newest partitions coming first.

Second, we want to change the rebuild algorithm to look at partitions sorted by their type. This requires a slight change to the code I'll make tomorrow.

Third, we changed the default for the step size to 4. We 4x'd the max partition size for VAST v2.1, so we're hoping this is good enough with the other two changes in mind. I explained what the tradeoff is.

mavam: > First, we changed the catalog lookup to return partitions sorted by their import time range, with newest partitions coming first.

Woah, I wanted that for a while, but for reasons of deterministic behavior. Assuming there's just one query, the scheduler will not rip apart the order of execution, right? Can I think of this now as plowing through results backwards in time?

> Second, we want to change the rebuild algorithm to look at partitions sorted by their type.

I'm missing context to understand what the effect of this would be.

> Third, we changed the default for the step size to 4. We 4x'd the max partition size for VAST v2.1, so we're hoping this is good enough with the other two changes in mind.

Isn't step size orthogonal to partition size? But yeah, I agree this is probably good enough in practice.

dominiklohmann: > > First, we changed the catalog lookup to return partitions sorted by their import time range, with newest partitions coming first.
> 
> Woah, I wanted that for a while, but for reasons of deterministic behavior. Assuming there's just one query, the scheduler will not rip apart the order of execution, right? Can I think of this now as plowing through results backwards in time?

Not quite currently; this is the relevant snippet from the index:

```cpp
      std::vector<uuid> candidates;
      for (const auto& [_, active_partition] : self->state.active_partitions)
        candidates.push_back(active_partition.id);
      for (const auto& [id, _] : self->state.unpersisted)
        candidates.push_back(id);
      self
        ->request(self->state.catalog, caf::infinite, atom::candidates_v, query)
        .then([=, candidates = std::move(candidates)](
                catalog_result& midx_result) mutable {
          auto& midx_candidates = midx_result.partitions;
          VAST_DEBUG("{} got initial candidates {} and from meta-index {}",
                     *self, candidates, midx_candidates);
          candidates.insert(candidates.end(), midx_candidates.begin(),
                            midx_candidates.end());
          std::sort(candidates.begin(), candidates.end());
          candidates.erase(std::unique(candidates.begin(), candidates.end()),
                           candidates.end());
```

Note that we still sort to be able to deduplicate with `sort`, although there's technically no need to deduplciate that way given how few unpersisted partitions we usually have. The results would also not quite be in most-recently imported order, because the partitions are still scanned from front to back rather than back to front, but I assume this would already be a big improvement.

> > Second, we want to change the rebuild algorithm to look at partitions sorted by their type.
> 
> I'm missing context to understand what the effect of this would be.

This is so the step size of 4 is good enough most of the time. If we take any four partitions we are unlikely to merge anything, so instead we want to take four partitions of the same type that are close in import time range.

mavam: > Not quite currently; this is the relevant snippet from the index:

I'm not too eager in understanding how to implement this, as I'm lacking context, but I'd like to understand the high-level invariants to get clarity on the mental model. Mind using plain words?

libvast/src/system/catalog.cpp
```diff
@@ -455,16 +492,31 @@ catalog(catalog_actor::stateful_pointer<catalog_state> self,
...
-      else
-        result_candidates = std::move(ids_candidates);
+      } else {
+        result_candidates.reserve(ids_candidates.size());
+        for (const auto& ids_candidate : ids_candidates) {
+          const auto syn = self->state.synopses.find(ids_candidate);
+          if (syn == self->state.synopses.end())
+            VAST_WARN("{} ignores unknown candidate partition hint {} for "
+                      "query {}",
+                      *self, ids_candidate, query.id);
```
tobim, line 510: Can you add a comment about why this check is needed?

tobim (commented):
I only reviewed  Return a partition info from the catalog result" so far.

libvast/include/vast/partition_synopsis.hpp
```diff
@@ -108,18 +114,35 @@ struct partition_info {
...
+
+  friend bool
+  operator==(const partition_info& lhs, const partition_info& rhs) noexcept {
+    return lhs.uuid == rhs.uuid;
+  }
+
+  friend bool
+  operator==(const partition_info& lhs, const class uuid& rhs) noexcept {
+    return lhs.uuid == rhs;
+  }
```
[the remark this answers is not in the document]
dominiklohmann: Because clang-13 doesn't appear to have that implemented yet.

libvast/include/vast/partition_synopsis.hpp
```diff
@@ -108,18 +114,35 @@ struct partition_info {
...
+
+  friend std::strong_ordering
+  operator<=>(const partition_info& lhs, const partition_info& rhs) noexcept {
+    return lhs.uuid <=> rhs.uuid;
+  }
+
+  friend std::strong_ordering
+  operator<=>(const partition_info& lhs, const class uuid& rhs) noexcept {
+    return lhs.uuid <=> rhs;
+  }
```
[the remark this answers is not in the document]
dominiklohmann: Since C++20 `operator<=>` and `operator==` are automatically generated for the other direction.

libvast/src/system/catalog.cpp
```diff
@@ -278,25 +309,30 @@ std::vector<uuid> catalog_state::lookup_impl(const expression& expr) const {
...
+                  for (const auto& [field, _] : part_syn->field_synopses_) {
                     VAST_ASSERT(!field.is_standalone_type());
                     auto rt = record_type{{field.field_name(), field.type()}};
                     for ([[maybe_unused]] const auto& offset :
                          rt.resolve_key_suffix(*s, field.layout_name()))
                       return true;
                   }
                   return false;
-                }();
+                }(part_syn);
```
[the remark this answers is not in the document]
dominiklohmann: Because I cannot capture part of a structured binding in a lambda until C++23.

libvast/fbs/partition_synopsis.fbs
```diff
@@ -18,6 +19,10 @@ table LegacyPartitionSynopsis {
   /// the `rebuild` command to determine whether a partitions needs to be
   /// upgraded to a more recent version.
   version: ulong;
+
+  /// The schema of this partition. Note that this field was not present for
+  /// partition synopses with a version number of 0.
```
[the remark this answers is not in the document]
dominiklohmann: It's in `defaults.hpp` currently. I'll make that more prominent and move it to config.hpp or so.

web/docs/setup-vast/tune.md
dominiklohmann: We went over this in a call in Slack, but to give a short summary of changes since we started this discussion:

The catalog lookup now returns a vector of partition infos, which contains the type (iff the partition is homogenous) and the number of events. This in turn allowed me to get rid of the step size entirely.

The index now utilizes this to work on partitions by their recency, assuming no partition scheduling is happening. The order is active partitions followed by unpersisted partitions followed by persisted partitions reverse-sorted by their max import time.

libvast/src/system/catalog.cpp
dominiklohmann: Turns out it wasn't. I added a comment why and added an assertion instead.

libvast/native-plugins/rebuild.cpp
```diff
@@ -0,0 +1,407 @@
...
+              VAST_ASSERT(result.size() > 1);
+              if (result.back().events <= self->state.max_partition_size / 2) {
+                self->state.remaining_partitions.push_back(
+                  std::move(result.back()));
+                self->state.num_transformed -= 1;
+                self->state.num_results -= 1;
+                self->state.num_total += 1;
+              }
+            }
+            if (is_heterogeneous || is_oversized)
```
dominiklohmann, line 263: Introduce an extra flag that we set to true if we actually pushed back.

tobim (approved):
We did the final review on the rebuilding algorithm in a call. Feedback is already integrated.

PR #2323, opened 2022-06-02
Enable `to<...>` to accept additional arguments

dispanser:
During our hackathon project, we noticed that
`to<arrow::compute::Exception>(vast_expr, vast_layout)` wouldn't work,
  which is addressed in this small PR.

### :memo: Checklist

Very small. Doesn't require any doc updates b/c it's not user-facing, only internal API.


mavam (approved):
We paired this today.

PR #2328, opened 2022-06-05
Fix static binary build of the parquet-store plugin

tobim:
The parquet installation links to thrift, but fails to call `find_dependency` for it. Thrift itself tries to find libevent but doesn't install its own FindLibevent module.


dominiklohmann (approved):
I approve of the CMake changes in here, this is the correct way to fix this. Probably should be merged by @dispanser since it targets his PR.

dispanser:
thanks a lot for this.

PR #2329, opened 2022-06-07
Skip push when lambda image did not change

rdettai:
Currently, each time we run the deploy script a new version of the lambda image is tagged on ECR and the lambda is updated to use that new image. This adds a check on the image digest to avoid this unnecessary update. It speeds up greatly subsequent deploy commands.

### :memo: Checklist


rdettai:
I am closing this after all because it is not very convenient to do it separately from the cloudtrial PR

PR #2336, opened 2022-06-09
Improve usability of CSV format

dominiklohmann:
Three changes:
- Add a `vast import csv --separator=...` option that defaults to `,`, and tests it for space and tab separators.
- Render enums in their canonical string representation instead of their internal numerical representation.
- Import additional columns in the csv import as strings instead of crashing.

### :memo: Checklist

@rdettai can you give this a try since you requested these?

For the actual code reviewer, take a look at it commit-by-commit.


libvast/src/format/csv.cpp
```diff
@@ -167,8 +168,15 @@ reader::reader(const caf::settings& options, std::unique_ptr<std::istream> in)
   if (in != nullptr)
     reset(std::move(in));
   using defaults = vast::defaults::import::csv;
-  opt_.separator
+  opt_.separator = defaults::separator[0];
+  auto seperator_option
     = get_or(options, "vast.import.csv.separator", defaults::separator);
+  if (seperator_option.size() != 1)
+    VAST_WARN("{} encountered invalid vast.import.csv.separator '{}'; must be "
+              "a single character",
```
mavam, line 176: Is this really a restriction?

dominiklohmann: Yes, unless we want to rewrite large parts of the parser. We exclude the separator character from the allowed character list in many places, which isn't trivial to exchange with an arbitrary-length string.

mavam: Okay, so nothing in principle that would hold us back, but too much scope-creep for this PR.

rdettai (changes requested):
I get the following error when I use `vast import --type=aws.flowlogs csv --separator=" "`:
- `vast.format.csv.reader encountered invalid vast.import.csv.separator ''; must be a single character`

But the import seems to have worked fine...
Any Idea what might be happening?

dominiklohmann:
@rdettai and I went over this in a call and discovered another issue with the CSV parser that is independent of the issues solved by this PR. I plan to take a further look tomorrow.

rdettai:
right, it didn't work, it actually ignored the option altogether and inserted everything into a single column 😄. This triggered a follow up:
- my source file looks like this:
```
> cat source.csv
version account-id interface-id srcaddr dstaddr srcport dstport protocol packets bytes start end action log-status
2 462855639117 eni-0a0c4004d2115a27f - - - - - - - 1654735756 1654735786 - NODATA
```
- the schema I have matches these columns
- when I import this as CSV (with "," as separator), 
  - the importer sees this file as one column (that is expected)
  - the importer accepts the event in this file and puts its content into the a single column. That shouldn't happen because the name of that mega column (the concatenation `version account-id interface-id ... action log-status`) does not match any field in the schema
```
> vast export csv
aws.flowlogs,"2 462855639117 eni-0a0c4004d2115a27f - - - - - - - 1654735756 1654735786 - NODATA"
```

PR #2341, opened 2022-06-13
Add web page on vision

mavam:
This PR adds a page on the vision of VAST.

Build locally, read, and take a look at the images.


web/docs/about-vast/vision.md
```diff
@@ -0,0 +1,157 @@
...
+:::
+
+[intel-soc]: https://www.intel.com.au/content/www/au/en/it-management/intel-it-best-practices/modern-scalable-cyber-intelligence-platform-kafka.html
+
+How do we get there? To answer this question, we first have to understand where
+we are today.
+
+## The SOC Architecture Maze
+
+Today's SOC architecture is product-centric: SIEM harbors all data, SOAR
```
dominiklohmann, line 60: Needs disclaimer on the abbreviations and target audience for this section.

dominiklohmann (approved):
The new content is nice, it essentially fills the void I was looking at recently. I think we need to iterate on this a few more times, but let's do that step-by-step. This is a really good starting point for this comparison! 🚀

dominiklohmann:
> Can we do anything about https://github.com/actions/cache/issues/820?

Don't think we can—just rerun it.

web/docs/about-vast/vision.md
mavam: I created a follow-up story for this.

PR #2342, opened 2022-06-13
Fix warnings for builds without assertions

dominiklohmann:
This fixes a rather obvious oversight in the recent PR #2330 that we did not catch in CI because that always builds with assertions.

### :memo: Checklist

n/t


mavam (approved):
We paired this.

PR #2344, opened 2022-06-14
Add content to Use Cases

mavam:
This PR adds content to the Use Cases section. There's also one clarification on OpenDXL in the Vision section.

Give it a read.


web/docs/about-vast/use-cases.md
```diff
@@ -4,6 +4,163 @@ sidebar_position: 3
...
+The core idea is to keep the bulk of the data decentralized at their respective
+origin, and disseminate security content via a [security data
+fabric](vision#security-data-fabric) to the VAST nodes—effectively shipping
+compute to the data. The SOC will always need central management and
+collaboration, and selective backhauling of data. The architecture does not
+exclude this, but the focus is on offloading as much processing to the edge as
+possible, including contextualization and pre-processing.
+
+This approach differs from traditional architectures that centralizes *all*
+security event data, with SIEMs of as work horse. Specifically, when building a
```
dispanser, line 29: "SIEMs of as work horse" reads a bit bulky. Is there an "of" too much?

web/docs/about-vast/vision.md
```diff
@@ -33,21 +33,34 @@ moves away from point-to-point integrations:
...
 ![Security Data Fabric](/img/security-data-fabric.dark.png#gh-dark-mode-only)
 
+The network of VAST nodes forms the security data fabric. Communication takes
+place over a pluggable messaging *backbone*, such as Kafka, RabbitMQ, or MQTT.
+In this architecture, VAST assumes a mediator function, with a backbone-facing
+and local-facing side. On the backbone-facing side, VAST implements the security
+domain knowledge and analytical processing power to enable composable use cases,
+such as routing security content to detection engines, or executing security on
+top of the available telemetry. On the local-facing side, VAST integrates
+security tools by infusing them with fabric-provided data and, vice versa, and
```
dispanser, line 43: "and, vice versa, and" doesn't read smooth IMHO. A single "and" might be all we need here.

dispanser (approved):
Reads great, thanks for making it!

I made two minor comments inline. 

One thing I noticed is that there's a significant overlap between the "General Issue" section in the "vision" and some of the "central security data storage" discussion in the "Use Cases" section. I think it's fine to have some level of redundancy, but we'd have to make sure that they don't diverge / start contradicting each other if one or the other is updated down the road.

PR #2346, opened 2022-06-15
Compress serialized indexers

dominiklohmann:
This works around a limitation of FlatBuffers: The maximum size of a FlatBuffers table is 2 GiB. Now while us running into this can be solved by moving indexers to separate files, we can get quite a ways by simply compressing the individual indexers.

In my initial measurements, I saw reductions between 2-10x in size of the partition FlatBuffers tables, with partitions containing mostly string indexes compressing best.

In limited testing, I was able to bump the max-partition-size from 4 Mi events to 64 Mi events without running into any size limitations, so with this change we should be good for a while, and can even consider bumping the size again.

### :memo: Checklist

Test with old and newly created databases. Rebuild an older one. Review the code file-by-file.

### DIFF
--- a/changelog/unreleased/features/2346--indexer-compression.md
+++ b/changelog/unreleased/features/2346--indexer-compression.md
@@ -1,5 +1,5 @@
 VAST now compresses on-disk indexes with Zstd, resulting in a 50-80% size
 reduction depending on the type of indexes used, and reducing the overall index
 size to below the raw data size. This improves retention spans significantly.
-For example, using the default configuration, partitions of `suricata.ftp`
+For example, using the default configuration, the indexes for `suricata.ftp`
 events now use 75% less disk space, and `suricata.flow` 30% less.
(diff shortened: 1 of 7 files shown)


changelog/unreleased/features/2346--indexer-compression.md
```diff
@@ -0,0 +1,5 @@
+VAST now compresses on-disk indexes with Zstd, resulting in a 50-80% size
+reduction depending on the type of indexes used, and reducing the overall index
+size to below the raw data size. This improves retention spans significantly.
+For example, using the default configuration, partitions of `suricata.ftp`
```
tobim, line 4: A user could easily interpret this as a 75% improvement overall.
```suggestion
For example, using the default configuration, the indexes for `suricata.ftp`
```

tobim (approved):
This looks great!

PR #2347, opened 2022-06-15
Fix a crash when loading older partition synopses

dominiklohmann:
This fixes a potential crash when loading older partition synopses that I introduced recently. 

### :memo: Checklist

Please try this out @6yozo / @tobim. I don't think this needs a dedicated changelog entry, since this was a recent introduction.


6yozo (approved):
I have tested the change on my local machine with an old partition. Everything works as expected.

PR #2353, opened 2022-06-16
Avoid clashes between rebuild, compaction, and aging

dominiklohmann:
Before this change, running at least two instances of rebuild, compaction, or aging caused compaction to break its internal state, rebuild to abort or even crash, and aging to crash the VAST server. This change makes it so we're a bit more lenient with overlapping partition transforms.

### :memo: Checklist

Don't think this needs a changelog entry, as this was not an issue before v2.1.

Review file-by-file, and definitely test locally.

### DIFF
--- a/libvast/native-plugins/rebuild.cpp
+++ b/libvast/native-plugins/rebuild.cpp
@@ -51,9 +51,15 @@ struct rebuilder_state {
   size_t num_transformed = {};
   size_t num_results = {};
   size_t num_heterogeneous = {};
+  bool inaccurate_statistics = false;
 
   /// Creates the indicator bar that shows the current progress.
   void create() {
+    const auto console_verbosity_level = loglevel_to_int(
+      caf::get_or(self->config(), "vast.console-verbosity",
+                  std::string{defaults::logger::console_verbosity}));
+    if (console_verbosity_level < VAST_LOG_LEVEL_INFO)
+      return;
     VAST_ASSERT(!indicator);
     indicators::show_console_cursor(false);
     self->attach_functor([] {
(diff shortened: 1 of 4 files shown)


libvast/native-plugins/rebuild.cpp
```diff
@@ -239,38 +239,42 @@ rebuilder(rebuilder_actor::stateful_pointer<rebuilder_state> self,
...
-            if (is_oversized) {
-              VAST_ASSERT(result.size() > 1);
-              if (result.back().events <= self->state.max_partition_size / 2) {
-                self->state.remaining_partitions.push_back(
-                  std::move(result.back()));
-                needs_second_stage = true;
-                self->state.num_transformed -= 1;
-                self->state.num_results -= 1;
-                self->state.num_total += 1;
+              if (is_oversized) {
```
tobim, line 258: I think we need to derive this from the `result` now.

libvast/native-plugins/rebuild.cpp
```diff
@@ -239,38 +239,42 @@ rebuilder(rebuilder_actor::stateful_pointer<rebuilder_state> self,
...
-                std::copy(result.begin(), result.end(),
-                          std::back_inserter(self->state.remaining_partitions));
-                needs_second_stage = true;
+            self->state.num_transformed += num_partitions;
+            self->state.num_results += result.size();
+            if (!result.empty()) {
+              if (is_heterogeneous) {
+                self->state.num_transformed -= num_partitions + 1;
+                self->state.num_results -= result.size();
+                self->state.num_heterogeneous -= 1;
```
tobim, line 249: First unconditionally incrementing and then doing additional corrections in the branches is heavily confusing. Please rewrite this so that every branch does a single self-contained modification of these variables.

tobim (changes requested):
Just a few observations. I haven't thought about the `eraser` yet, but it might also need some adaptations.

libvast/src/system/index.cpp
```diff
@@ -1329,9 +1321,45 @@ index(index_actor::stateful_pointer<index_state> self,
...
+              = current_sender->address();
+          } else if (match->second == current_sender->address()) {
+            // Getting overlapping partitions from the same sender is a hard
+            // error. This indicates a logic error upstream.
+            return caf::make_error(
+              ec::logic_error,
+              fmt::format("{} refuses to apply transformation '{}' to "
+                          "partition {} because it is currently being "
+                          "transformed by the same sender",
+                          *self, transform->name(), partition));
```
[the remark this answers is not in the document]
dominiklohmann: This has helped me catch a bug during development of the `rebuild` command. I think it's generally nice to let the user know they're likely doing something wrong.

I'm not opposed to removing it, but I think it's not an expensive check that may help avoid some surprises.

libvast/native-plugins/rebuild.cpp
```diff
@@ -234,34 +250,63 @@ rebuilder(rebuilder_actor::stateful_pointer<rebuilder_state> self,
...
+            // match the number of events in the partitions that went in we ran
+            // into a conflict with other partition transformations on an
+            // overlapping set.
+            const auto detected_conflict
+              = current_run_events
+                != std::transform_reduce(result.begin(), result.end(), size_t{},
+                                         std::plus<>{},
+                                         [](const partition_info& partition) {
+                                           return partition.events;
+                                         });
```
tobim, line 267: To be honest I don't really get the point of this. My understanding says we could just iterate over the result partitions and put them into the second stage if they are under the size threshold. Then we can remove `is_oversized`. Then again, I don't understand this algorithm as well as @dominiklohmann, and this version works. So I'm approving this.

tobim (approved):
I want to get this into rc3 for end-to-end testing, we can still adjust the algorithm a bit if it becomes necessary.

PR #2355, opened 2022-06-17
AWS Concepts

rdettai:
Add concepts to AWS types

### :memo: Checklist

Concepts need to be on the server side, so we replace the default vast image with a custom one that contains the concept files. This has also the benefit of moving the image closer to fargate, hence improving server startup time.


rdettai:
@mavam I have only added two very basic concepts that will be there in most AWS events: the region and the account id. The goal here is mostly to show that it works. Obviously this can be built upon and extended, I think there is a lot more modelling that could be done.

rdettai:
Integration test [OK](https://github.com/rdettai/vast/runs/6932222105)

PR #2357, opened 2022-06-17
Tune web content a bit

mavam:
This PR tunes some web content.

Ideally read it top-to-bottom again.


tobim (approved):
From a high-level perspective I think this is a good improvement.

web/blog/apache-arrow-as-platform-for-security-data-engineering/index.md
```diff
@@ -15,70 +14,89 @@ Influx's [IOx][iox], DataDog's [Husky][husky], Anyscale's [Ray][ray],
...
+platform is support for the *security* domain. This is where Arrow's [extension
+types][extension-types] come into play. They add *semantics* to otherwise
+generic types, e.g., by telling the user "this is a transport-layer port" and
+not just a 16-bit unsigned integer, or "this is a connection 4-tuple to
+represent a network flow" instead of "this is a record with 4 fields of type
+string and unsigned integer". Extension types are composable and allow for
+creating a rich typing layer with meaningful domain objects on top of a
+standardized data representation. Since they are embedded in the data, they do
+not have to be made available out-of-band when crossing the boundaries of
+different tools. Now we have self-describing security data.
```
tobim, line 85: This is only half-true, since extension types must be registered so they can be used, and the extra operations can't be pushed down into the arrow library itself. @dominiklohmann correct me if there is some way to extend the API.

mavam: It sounds like you're talking about "code" whereas I'm talking about "data". Client-libraries with type munching will always be a nicer DX than pure-play Arrow extension types, although you get quite for without it.

PR #2363, opened 2022-06-20
Emit event and partition metrics from the catalog

dominiklohmann:
This adds `catalog.num-{events,partitions}` metrics to the output of the catalog, both containing the schema name and the partition version in the metrics metadata under `metadata_schema` and `metadata_partition-version` respectively.

### :memo: Checklist

@tobim we talked about having such a metric this morning, and I happened to need this for debugging another thing I was looking at, so I figured I'd just create a quick PR for this.

I think the review should be straightforward. I tested this locally by looking at the `metrics.log` file and verifying the numbers match what's in the status output.


tobim (approved):
Looks promising. Looking forward to seeing this on a dashboard.

PR #2365, opened 2022-06-20
Add a few additional checks to the query queue

tobim:
These checks are motivated by some spurious warnings about failed partition loads during testing. The added assertions are intended to verify that the query queue works correctly.

### :memo: Checklist

A "static" review should suffice. The commits are orthogonal.

### DIFF
--- a/libvast/include/vast/detail/algorithms.hpp
+++ b/libvast/include/vast/detail/algorithms.hpp
@@ -18,7 +18,7 @@ namespace vast::detail {
 template <class T>
 concept has_contains = requires(T& t, typename T::value_type& x) {
   requires concepts::container<T>;
-  requires t.contains(x);
+  { t.contains(x) } -> std::convertible_to<bool>;
 };
 
 template <concepts::range T, class U>
--- a/libvast/include/vast/detail/algorithms.hpp
+++ b/libvast/include/vast/detail/algorithms.hpp
@@ -25,7 +25,8 @@ template <concepts::range T, class U>
 bool contains(const T& t, const U& x) {
   if constexpr (has_contains<T>)
     return t.contains(x);
-  return std::find(t.begin(), t.end(), x) != t.end();
+  else
+    return std::find(t.begin(), t.end(), x) != t.end();
 }
 
 template <class Collection>
--- a/libvast/include/vast/detail/algorithms.hpp
+++ b/libvast/include/vast/detail/algorithms.hpp
@@ -8,10 +8,25 @@
 
 #pragma once
 
+#include "vast/concepts.hpp"
+
 #include <algorithm>
 #include <vector>
 
 namespace vast::detail {
+
+template <class T>
+concept has_contains = requires(T& t, typename T::value_type& x) {
+  requires concepts::container<T>;
+  requires t.contains(x);
+};
+
+template <concepts::range T, class U>
+bool contains(const T& t, const U& x) {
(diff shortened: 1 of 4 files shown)


libvast/src/query_queue.cpp
```diff
@@ -62,18 +77,17 @@ query_queue::insert(query_state&& query_state, std::vector<uuid>&& candidates) {
...
-    auto it = std::find_if(partitions.begin(), partitions.end(), [&](auto& x) {
-      return x.partition == cand;
-    });
+    auto it = std::find(partitions.begin(), partitions.end(), cand);
     if (it != partitions.end()) {
       it->priority += query_state_it->second.query.priority;
       it->queries.push_back(qid);
+      VAST_ASSERT_CHEAP(!detail::contains(inactive_partitions, cand),
+                        "A partition must not be active and inactive at the "
+                        "same time");
```
dominiklohmann, line 86: Isn't this check being executed `candidates * inactive-partitions` times? I wouldn't call this cheap.

libvast/include/vast/query_queue.hpp
```diff
@@ -55,9 +55,9 @@ class query_queue {
     std::vector<uuid> queries;
     bool erased = false;
 
-    friend auto operator<=>(const entry& lhs, const entry& rhs) {
-      return lhs.priority <=> rhs.priority;
-    }
+    friend bool operator<(const entry& lhs, const entry& rhs) noexcept;
+    friend bool operator==(const entry& lhs, const uuid& rhs) noexcept;
+    friend bool operator==(const uuid& lhs, const entry& rhs) noexcept;
```
dominiklohmann, line 60: Only one of these should be needed in C++20, as the [primary comparison operators can be reversed](https://brevzin.github.io/c++/2019/07/28/comparisons-cpp20/).

libvast/include/vast/detail/algorithms.hpp
```diff
@@ -8,11 +8,26 @@
...
+
 #include <algorithm>
 #include <vector>
 
 namespace vast::detail {
 
+template <class T>
+concept has_contains = requires(T& t, typename T::value_type& x) {
+  requires concepts::container<T>;
+  requires t.contains(x);
```
dominiklohmann, line 21:
```suggestion
  { t.contains(x) } -> std::convertible_to<bool>;
```

I don't think you're checking what you want to check otherwise.

libvast/include/vast/detail/algorithms.hpp
```diff
@@ -8,11 +8,26 @@
...
+};
+
+template <concepts::range T, class U>
+bool contains(const T& t, const U& x) {
+  if constexpr (has_contains<T>)
+    return t.contains(x);
+  return std::find(t.begin(), t.end(), x) != t.end();
```
dominiklohmann, line 28:
```suggestion
  else
    return std::find(t.begin(), t.end(), x) != t.end();
```

Without this else, the function fails to compile if `t.contains(x)` is valid, but `*t.begin() == x` is not, which would be the case for maps, where dereferencing the iterator would yield the key-value pairs, while contains just takes the key.

dominiklohmann (approved):
Approving this already, but I've left a few remarks inline that I'd like to see addressed. I think this is a good way of debugging this, we can probably deploy this and the other changes since the v2.1.0-rc2 as a v2.1.0-rc3 tomorrow.

PR #2366, opened 2022-06-20
Optimize catalog field extractor matching

tobim:
An attempt to improve the performance by doing a cheaper suffix match and moving the type check back.

### :memo: Checklist

### DIFF
--- a/libvast/src/system/catalog.cpp
+++ b/libvast/src/system/catalog.cpp
@@ -344,14 +344,31 @@ catalog_state::lookup_impl(const expression& expr) const {
         },
         [&](const field_extractor& lhs, const data& d) -> result_type {
           auto pred = [&](const auto& field) {
-            if (!compatible(field.type(), x.op, d))
+            auto match_name = [&] {
+              auto field_name = field.field_name();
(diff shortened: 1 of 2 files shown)


libvast/src/system/catalog.cpp
```diff
@@ -344,14 +344,31 @@ catalog_state::lookup_impl(const expression& expr) const {
...
-            if (!compatible(field.type(), x.op, d))
+            auto match_name = [&] {
+              auto field_name = field.field_name();
+              auto key = std::string_view{lhs.field};
+              if (field_name.length() >= key.length()) {
+                auto pos = field_name.length() - key.length();
+                auto sub = field_name.substr(pos);
+                return sub == key && (pos == 0 || field_name[pos - 1] == '.');
+              }
+              auto layout_name = field.layout_name();
```
dispanser, line 355: there is no code path in the unit tests that exercise the case `key.length() > field_name.length()`, so it's a bit hard to see correctness.

dispanser (approved):
I added a few assertions in one of our existing test cases to convince myself that the code is doing what it seems to be doing, and pushed those as discussed.

PR #2368, opened 2022-06-22
Fix missing bucket when datasource not used

rdettai:
Currently, if the datasource bucket is not defined in `.env`, a default dummy value is used and hence the datasource that is in charge of checking that the bucket is valid fails. We propose using a provisioner instead.

### :memo: Checklist


rdettai:
https://github.com/rdettai/vast/runs/6999918998

dispanser (approved):
I have not tested this specifically but it looks OK

PR #2370, opened 2022-06-22
Improve error management in Lambda

rdettai:
Currently:
- the commands that fail are not logged as errors in lambda
- the way stdout and stderr are multiplexed is hard to use in practice

We propose to raise an exception is the command returns a non 0 code and to return an object with both stdout and stderr from the lambda otherwise.

### :memo: Checklist

- Note that in the lambda, we log the stderr in real time to allow better monitoring of long running lambdas. To avoid duplication of those logs when an exception is raised, we silence it in a "hacky way" (temp redirect of the python sys.stdout)
- the web documentation wast not updated with the new output format of the lambda as this would be an excessive level of detail. The CLI help was updated instead.

### DIFF
--- a/cloud/aws/cli/core.py
+++ b/cloud/aws/cli/core.py
@@ -274,7 +274,16 @@ def run_lambda(c, cmd):
     )
     resp_payload = lambda_res["Payload"].read().decode()
     if "FunctionError" in lambda_res:
-        mess = f'{lambda_res["FunctionError"]}: {resp_payload}'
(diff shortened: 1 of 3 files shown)


rdettai:
https://github.com/rdettai/vast/actions/runs/2541220246

dispanser (approved):
Approving this; tested the initial version, but I think our following discussion has been integrated into the final result.

PR #2371, opened 2022-06-22
Fix schema names in the catalog metrics

tobim:
### :memo: Checklist

Test if the metrics are correct now. Ideally with a database containing both old and new partitions.


tobim:
This turned out to be unnecessary, the code is already correct as-is.

PR #2374, opened 2022-06-23
Prevent slices coming in when a shutdown is in progress

6yozo:
This fix is intended to prevent the occasional hang when VAST is shut down.

The tests suggest that the fix is effective, but more tests might be needed. 

### :memo: Checklist

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -885,14 +885,15 @@ index(index_actor::stateful_pointer<index_state> self,
     [](caf::unit_t&) {
       // nop
     },
-    [self](caf::unit_t&, caf::downstream<table_slice>& out, table_slice x) {
+    [self, stage = self->state.stage.get()](
+      caf::unit_t&, caf::downstream<table_slice>& out, table_slice x) {
       VAST_ASSERT(x.encoding() != table_slice_encoding::none);
+      if (stage->running())
+        return;
       auto&& layout = x.layout();
       // TODO: Consider switching layouts to a robin map to take advantage of
       // transparent key lookup with string views, avoding the copy of the name
       // here.
-      if (self->state.is_shutting_down)
-        return;
       self->state.stats.layouts[std::string{layout.name()}].count += x.rows();
       auto& active = self->state.active_partitions[layout];
       if (!active.actor) {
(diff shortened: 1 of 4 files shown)


libvast/src/system/index.cpp
```diff
@@ -891,6 +891,8 @@ index(index_actor::stateful_pointer<index_state> self,
       // TODO: Consider switching layouts to a robin map to take advantage of
       // transparent key lookup with string views, avoding the copy of the name
       // here.
+      if (self->state.is_shutting_down)
```
tobim, line 894: Maybe we can reuse the flag from CAF to avoid the extra state?
```suggestion
      if (stage->is_shutting_down)
```
Should be possible if we capture the stage pointer (make sure it isn't an owning type though).

tobim (approved):
This looks ok, but I would like you to switch to my suggestion, or add a comment why we can't use that if it doesn't work.

libvast/src/system/index.cpp
tobim: A solution that actually works should look something like:
```c++
if ((stage->flags() & (caf::stream_manager::is_stopped_flag | caf::stream_manager::is_shutting_down_flag)) != 0)
```

6yozo: Is it the stage which is set up a couple of lines above? It does not have ->flags().

```cpp
   self->state.stage = detail::attach_notifying_stream_stage(
    self,
    /* continuous = */ true,
    [](caf::unit_t&) {
      // nop
    },
    [self](caf::unit_t&, caf::downstream<table_slice>& out, table_slice x) {
      VAST_ASSERT(x.encoding() != table_slice_encoding::none);
      if ((self->state.stage->flags()
           & (caf::stream_manager::is_stopped_flag
              | caf::stream_manager::is_shutting_down_flag))
          != 0)
```

libvast/src/system/index.cpp
```diff
@@ -887,6 +887,8 @@ index(index_actor::stateful_pointer<index_state> self,
     },
     [self](caf::unit_t&, caf::downstream<table_slice>& out, table_slice x) {
       VAST_ASSERT(x.encoding() != table_slice_encoding::none);
+      if (self->state.stage->running())
```
tobim, line 890:
```suggestion
      if (!self->state.stage->running())
```

libvast/src/system/index.cpp
lava: For future reference, there's a `getf()` function in the stage so you don't have to work with the raw `flags_` variable.

PR #2375, opened 2022-06-23
Don't export the indicators dependency

tobim:
Building plugins against an installed VAST was broken. This is the fix.

### :memo: Checklist

Try building a plugin against an installed VAST, see that it fails without this commit.

### DIFF
--- a/libvast/CMakeLists.txt
+++ b/libvast/CMakeLists.txt
@@ -533,7 +533,16 @@ add_feature_info("VAST_ENABLE_BUNDLED_INDICATORS"
 if (NOT VAST_ENABLE_BUNDLED_INDICATORS)
   find_package(indicators 2.2.0 QUIET)
   if (indicators_FOUND)
-    target_link_libraries(libvast_native_plugins PRIVATE indicators::indicators)
+    # CMake exports private interface dependencies because they might depend on
+    # other non-interface libraries. CMake could scan the depency tree to figure
+    # out when this is actually needed, but it doesn't.
+    # We know that indicators itself doesn't have any dependency, so we work
+    # around this behavior by not linking to the target.
+    # https://gitlab.kitware.com/cmake/cmake/-/issues/15415 for more discussion.
+    get_target_property(indicators_include_directory indicators::indicators
+                        INTERFACE_INCLUDE_DIRECTORIES)
+    target_include_directories(
+      libvast_native_plugins PRIVATE "${indicators_include_directory}/include")
     dependency_summary("indicators" indicators::indicators "Dependencies")
   endif ()
 endif ()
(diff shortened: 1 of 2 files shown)


libvast/CMakeLists.txt
```diff
@@ -533,7 +533,16 @@ add_feature_info("VAST_ENABLE_BUNDLED_INDICATORS"
...
+    # CMake exports private interface dependencies because they might depend on
+    # other non-interface libraries. CMake could scan the depency tree to figure
+    # out when this is actually needed, but it doesn't.
+    # We know that indicators itself doesn't have any dependency, so we work
+    # around this behavior by not linking to the target.
+    # https://gitlab.kitware.com/cmake/cmake/-/issues/15415 for more discussion.
+    get_target_property(indicators_include_directory indicators::indicators
+                        INTERFACE_INCLUDE_DIRECTORIES)
+    target_include_directories(
+      libvast_native_plugins PRIVATE "${indicators_include_directory}/include")
```
6yozo, line 545: I cannot tell exactly what this change is doing, but it builds fine, so it must be ok.

tools/vast-regenerate/main.cpp
```diff
@@ -93,10 +93,8 @@ caf::error write_index_bin(const std::vector<vast::uuid>& uuids,
   for (const auto& uuid : uuids) {
     if (auto uuid_fb = pack(builder, uuid))
       partition_offsets.push_back(*uuid_fb);
-    else {
-      fmt::print(stderr, "failed to pack uuid {}: {}\n", uuid, uuid_fb.error());
-      return 1;
-    }
+    else
+      return uuid_fb.error();
```
6yozo, line 97: We are losing the context here("failed to pack uuid"), but the proper error message is returned to the caller; this is ok as it is or maybe:

```
      return caf::make_error(vast::ec::convert_error,
                             fmt::format("failed to pack uuid {}", uuid),
                             uuid_fb.error());
```

tools/vast-regenerate/main.cpp
```diff
@@ -284,7 +282,7 @@ with an incorrect all-zero event count.
       path = arg;
     }
   }
-  if (!path) {
+  if (path.empty()) {
```
6yozo, line 285: Nice! Now it can detect when the path is not given in CLI.

6yozo (approved):
The PR is good to go, as long as my understanding of CMAKE goes.

tools/vast-regenerate/main.cpp
tobim: All `pack` overloads share `caf::error` as return type. The one for `uuid` can't fail though, and I don't see how that would change in the future.

PR #2376, opened 2022-06-23
The index shall not quit on write errors

tobim:
The index used to quit if partitions fail to persist their data, this is not really necessary as VAST can still execute queries just fine when the disk is full.

### :memo: Checklist

This should be trivial enough to accept without manual testing.


dispanser (approved):
That's straight-forward.

I tested this by injecting a write error (via `chown / chmod`) and observed VAST keeping running.

PR #2378, opened 2022-06-24
Yet another docs update

mavam:
Another round of docs updates.

Go commit-by-commit.


dispanser (approved):
I didn't read it all, because as stated, there's no new content so this is only about changing the structure, which I find good :+1:

PR #2384, opened 2022-06-27
Increase task shutdown timeout

rdettai:
When a Fargate task is stopped, the process inside receives a SIGTERM command with a grace period that defaults to 30 sec before receiving a SIGKILL. We propose here to increase that period to 120 (Fargate's maximum) to leave more time for VAST to flush its content.

### :memo: Checklist


lava (approved):
Makes sense to me

PR #2387, opened 2022-06-28
Implement more Sigma rule modifiers

mavam:
This PR adds supports for the Sigma rule modifiers `lt`, `lte`, `gt`, `gte`, and `cidr`. These are part of the new pySigma effort and not available in "legacy" sigmac.

In addition, this PR implements the `base64` and `base64offset` modifier in a composable fashion, i.e., chaining via `base64|base64` is also possible.

### :memo: Checklist

- [x] The changes are reflected in the docs.

It suffices to read the unit tests and updated documentation.


dominiklohmann (approved):
Just a minor request from my side. I've added the blocked label to indicate that this PR should only be merged after the release.

KaanSK (commented):
I was able to understand the change, code and overall functionality just by going over sigma.md and code changes. LGTM.

Only field I would like to make a comment on is "Focus on Endpoint" section in sigma.md . The conformity table under "Compatibility" already provides enough insights and users can make their judgement on use case.

mavam:
> Only field I would like to make a comment on is "Focus on Endpoint" section in sigma.md

Feel free to share the comment here. 🙂

KaanSK:
> > Only field I would like to make a comment on is "Focus on Endpoint" section in sigma.md
> 
> Feel free to share the comment here. 🙂

The section may give false impression to users. I personally think that this section should be removed. The conformity table under "Compatibility" provides detailed information for user to be able to assess the requirements and abilities of VAST in sigma regard.

plugins/sigma/src/parse.cpp
```diff
@@ -275,47 +270,144 @@ caf::expected<expression> parse_search_id(const data& yaml) {
...
+            if (!str)
+              return caf::make_error(ec::type_clash, //
+                                     "utf16le/wide only works with strings");
+            // Hand-roll conversion.
+            std::string result;
+            result.reserve(str->size() * 2);
+            for (auto c : *str) {
+              result.push_back(c);
+              result.push_back(0x00);
+            }
```
mavam, line 314: @dominiklohmann here's a naive attempt to hand-roll support for UTF16 strings. Can this work?

(It's currently dead code, but we have a story for making this work.)

dominiklohmann: That's most certainly not correct—you cannot map multi-byte (or rather, multi-codepoint) UTF8 strings onto UTF16 codepoints like this. I propose throwing this out right now and asking users to do this conversion externally. Since most of the world only supports UTF8 nowadays I don't think we need to implement this at all, ever, unless someone is paying us enough money to go down this rabbit hole or std::text exists in standard C++.

mavam: > I propose throwing this out right now and asking users to do this conversion externally. 

This is not an option. It looks like my question was not very clear. Sigma has a few UTF modifiers (`utf16le`, `utf16be`, and `utf8`) that take ASCII strings and produce the corresponding UTF output. These need to be supported along other modifiers.

Since our raw data is just bytes, we _can_ perform this translation easily for UTF16, since it's just byte padding. UTF8 is a different story.

> I don't think we need to implement this at all

This is not correct. In order to support the Sigma rules out there, we need to implement these modifiers.

dominiklohmann: You don't validate here that your input is ASCII. YAML supports UTF-8, and that's what's likely to be your input here. If you want to do the naïve conversion, then you must verify that assumption first.

Regarding the use case I'm still not convinced. Yes, we need it to claim "full Sigma support." But does anyone _actually use_ it?

mavam: > Regarding the use case I'm still not convinced. Yes, we need it to claim "full Sigma support." But does anyone actually use it?

Fair question. I git-grepped the existing rules in the repo:

```
❯ git grep '|wide' | wc -l
       1   <-- unit test only
❯ git grep '|utf16be' | wc -l
       0
❯ git grep '|utf16le' | wc -l
       0
❯ git grep '|utf8' | wc -l
       0
```

In comparison (also low)

```
❯ git grep '|base64' | wc -l
       9
❯ git grep '|base64offset' | wc -l
       8
```

So your hunch was correct, there are not many (at least official) rules that use the codepoint conversion features. That's helpful wrt. prioritization of this feature.

PR #2392, opened 2022-07-01
Replace LB with service DNS

rdettai:
Currently we use a network load balancer to identify and route to the VAST service. This creates an unnecessary intermediary that introduces an extra fix cost. Using service discovery is actually cheaper (we only pay per DNS resolution and a tiny $0.5 monthly fee for the zone) and allows direct connection between the services.

This PR prepares the addition of new services to create a fully featured playground.

### :memo: Checklist

- We create the private service DNS outisde the fargate module to allow re-using it for new services (this was not the case with the existing load balancer)


rdettai:
https://github.com/rdettai/vast/actions/runs/2616461356

dispanser (approved):
Great change. Not actually run, we'll let the CI do that.

PR #2393, opened 2022-07-01
Render reals with at least one decimal place

patszt:
Json export will always render reals with at least one decimal place

### :memo: Checklist

### DIFF
--- a/libvast/include/vast/concept/printable/vast/json.hpp
+++ b/libvast/include/vast/concept/printable/vast/json.hpp
@@ -49,7 +49,7 @@ struct json_printer
 
   template <class Iterator>
   struct print_visitor {
-    print_visitor(Iterator& out) : out_{out} {
+    explicit print_visitor(Iterator& out) : out_{out} {
     }
 
     bool operator()(const caf::none_t&) {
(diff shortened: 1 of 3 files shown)


libvast/include/vast/concept/printable/vast/json.hpp
```diff
@@ -85,13 +85,11 @@ struct json_printer
...
           return printers::str.print(out_, "null");
         auto str = std::to_string(x);
-        real i;
         if constexpr (std::is_floating_point_v<T>) {
-          if (std::modf(x, &i) == 0.0)
-            // Do not show 0 as 0.0.
-            str.erase(str.find('.'), std::string::npos);
+          // Avoid trailing zeros.
+          if (real i; std::modf(x, &i) == 0.0)
+            str.erase(str.find('.') + 2, std::string::npos);
```
lava, line 91: Is it guaranteed that the index is always valid? If not we need to check, since `erase()` will throw `std::out_of_bounds`.

lava (approved):
Looks great, thanks for adding new unit tests as well!

dominiklohmann:
I think we should merge this one only after the v2.1 release. I'm adding a `blocked` label temporarily.

libvast/include/vast/concept/printable/vast/json.hpp
patszt: Based on https://en.cppreference.com/w/cpp/string/basic_string/to_string it should be.

PR #2395, opened 2022-07-01
Port PyVAST and Threat Bus docs

mavam:
Port PyVAST and Threat Bus docs, plus a few fixes.

Commit-by-commit. You can skip the Threat Bus commit, which just integrates the existing content from https://docs.tenzir.com/threatbus/ under *Use VAST → Integrate → Threat Bus*.


dominiklohmann (approved):
This reads nicely. Digging the tab-style Python/CLI way to display examples. I mostly glanced over the Threat Bus stuff, since that's almost a carbon copy of the old docs.

PR #2399, opened 2022-07-04
Show default index config in vast.yaml.example

dominiklohmann:
The old values did not reflect the default, but the vast.yaml.example file advertises that all shown values are the default exactly. This changes the default fp-rate for string and address types to actually be the default.

### :memo: Checklist

Simple text review. Should not need a changelog entry.


mavam (approved):
Paired this with @dominiklohmann.

PR #2405, opened 2022-07-04
Refactor EFS module for better re-usability

rdettai:
Currently the EFS module in included in the Fargate module, so each task will have its own filesystem. If we want the ability to re-use the file system, it's better if it is at the same level as the fargate module.

### :memo: Checklist

- moving the security groups into `securitygroups.tf` in the parent module to give a better view on network restrictions between services
- otherwise, mostly mechanical changes imposed by the refactoring

### DIFF
--- a/cloud/aws/terraform/common/lambda/inputs.tf
+++ b/cloud/aws/terraform/common/lambda/inputs.tf
@@ -34,3 +34,5 @@ variable "vpc_id" {
 variable "subnets" {
   default = []
 }
+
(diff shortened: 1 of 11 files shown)


rdettai:
https://github.com/rdettai/vast/actions/runs/2616579779

dispanser (approved):
I'll leave the testing to the CI, but this looks reasonable.

:+1: also for adapting the image representing the overall architecture.

PR #2391, opened 2022-06-30
Add a `union` action to the `summarize` plugin

dominiklohmann:
The new ~`gather`~ `union` action creates a list of all unique values in the selected columns in the grouped events.

### :memo: Checklist

- [x] All user-facing changes have changelog entries.
- [x] The changes are reflected on [docs.tenzir.com/vast](https://docs.tenzir.com/vast), if necessary.
- [x] The PR description contains instructions for the reviewer, if necessary.

### :dart: Review Instructions

Should be covered by the unit test, but feel free to try this out locally as well.

I think the docs for the `summarize` actions could use a lot of work, but that's for a separate PR.

For the code review itself, I recommend going file-by-file.

PR #2417, opened 2022-07-09
Summarize operator with pluggable aggregation functions

dominiklohmann:
This rewrites the `summarize` pipeline operator to have pluggable aggregation functions through the new `aggregation_function_plugin` plugin type, which each implement a simple virtual `aggregation_function` interface for incremental aggregation.

In addition, this converts both the summarize plugin and the existing aggregation functions to native plugins, making them always available when using VAST.

The configuration for the aggregation also changed, allowing users to specify output field names explicitly and aggregating multiple columns into one. This is the new configuration that I primarily used for testing with `suricata.flow` events:

```yaml
summarize:
  group-by:
    - timestamp
    - proto
    - event_type
  time-resolution: 1 hour
  aggregate:
    timestamp_min:
      min: timestamp
    timestamp_max:
      max: timestamp
    pkts_toserver: sum
    pkts_toclient: sum
    bytes_toserver: sum
    bytes_toclient: sum
    start: min
    end: max
    alerted: any
    ips:
      distinct:
        - src_ip
        - dest_ip
```

### :memo: Checklist

  - [x] Document summarize changes.
  - [x] Document `aggregation_function_plugin` plugin type.

I recommend going file-by-file, and testing locally with compaction. The following order is ideal:
1. Review the API of `aggregation_function_plugin`.
2. Review the API of `aggregation_function`
3. Review the `summarize` pipeline operator implementation.
4. Review the `summarize` configuration, and how it's bound to a schema to form an `aggregation`.
5. Review the individual aggregation function plugins.

This PR also includes the work from #2391, which I've closed accordingly.


mavam:
Just wondering: should `sample` take an argument `n` that specifies the sample size? Right now, we have `n = 1` hard-coded. To me, a `sample` function suggests that it's configurable in its desired output length.

dominiklohmann:
> Just wondering: should `sample` take an argument `n` that specifies the sample size? Right now, we have `n = 1` hard-coded. To me, a `sample` function suggests that it's configurable in its desired output length.

I don't think so to be honest, at least not for the foreseeable future while we still use the YAML based configuration. It's hard to fit additional parametrization for the aggregations functions in it, and I'd want to wait with doing so until we have an actual use case or request for it.

changelog/unreleased/changes/2417--aggregation-function-plugins.md
```diff
@@ -0,0 +1,6 @@
+Aggregation functions in the `summarize` operator are now plugins, which makes
+them easily extensible. The syntax of `summarize` now supports specification of
+output field names, similar to SQL's `AS` in `SELECT f(x) AS name`.
```
dominiklohmann, line 3: @mavam we should also mention that `summarize` is no longer a plugin, but rather a built-in operator. As such, it is no longer versioned separately and does not need to be enabled explicitly in the configuration anymore.

mavam: Ah, right. Feel free to adjust the wording.

libvast/native-plugins/sum.cpp
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  [[nodiscard]] const char* name() const override {
+    return "sum";
+  };
+
+  [[nodiscard]] caf::expected<std::unique_ptr<aggregation_function>>
+  make_aggregation_function(const type& input_type) const override {
+    auto f = detail::overload{
+      [&]<basic_type Type>(
```
dispanser, line 59: handle `sum(string)` as error

libvast/native-plugins/summarize.cpp
```diff
@@ -0,0 +1,772 @@
...
+      auto instance
+        = aggregation_function->make_aggregation_function(input_type);
+      if (!instance)
+        return caf::make_error(ec::invalid_configuration,
+                               fmt::format("aggregation function {} failed "
+                                           "to instantiate for input type "
+                                           "{}: {}",
+                                           aggregation.function_name,
+                                           input_type, instance.error()));
+      auto output_type = (*instance)->output_type();
```
dispanser, line 341: Should not be able to fail, maybe?

dispanser (approved):
Looked over this together, minor comments inline

changelog/unreleased/features/2417--new-aggregation-functions.md
```diff
@@ -0,0 +1,3 @@
+The `summarize` operator supports three new aggregation functions: `sample`
+takes the first value in every group, `distinct` filters out duplicate values,
```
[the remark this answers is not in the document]
dominiklohmann: We had this discussion already: Because the order is not deterministic it's essentially random.

libvast/native-plugins/summarize.cpp
```diff
@@ -232,7 +232,7 @@ struct group_by_column {
            schema_rt.resolve_key_suffix(extractor, schema.name())) {
         auto field = schema_rt.field(offset);
         auto& column = result.emplace_back();
-        column.input = std::move(offset);
+        column.input = offset;
```
mavam, line 235: NB: we really should have use-after-move clangd check as part of our dev environment, as provided by clang-tidy.

dominiklohmann: There's one for clang-tidy, but it's not working apparently. At least for me it did not show up here.

https://releases.llvm.org/5.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/misc-use-after-move.html#misc-use-after-move

PR #2422, opened 2022-07-11
Specify Arrow export in helptext

rdettai:
The arrow export is building on top of the Arrow IPC format and adds the possibility to stream data with multiple schemas by concatenating the IPC streams. This should be explicited in the cmd helptext.

### :memo: Checklist

NA

### DIFF
--- a/libvast/src/system/application.cpp
+++ b/libvast/src/system/application.cpp
@@ -124,7 +124,10 @@ auto make_export_command() {
   export_->add_subcommand("null",
                           "exports query without printing them (debug option)",
                           opts("?vast.export.null"));
-  export_->add_subcommand("arrow", "exports query results in Arrow format",
+  export_->add_subcommand("arrow",
+                          "exports query results in Arrow format with separate "
+                          "IPC streams for each schema, all concatenated "
+                          "together",
                           opts("?vast.export.arrow"));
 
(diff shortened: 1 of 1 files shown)


dominiklohmann (approved):

libvast/src/system/application.cpp
```diff
@@ -124,7 +124,10 @@ auto make_export_command() {
   export_->add_subcommand("null",
                           "exports query without printing them (debug option)",
                           opts("?vast.export.null"));
-  export_->add_subcommand("arrow", "exports query results in Arrow format",
+  export_->add_subcommand("arrow",
+                          "exports query results in Arrow format with separate "
+                          "IPC streams for each schema, all concatenated "
+                          "together",
```
mavam, line 130: What does "all concatenated together" add in terms of understanding? Isn't that already implied by "IPC streams"?

dominiklohmann: It's just a single output stream instead of one output stream per schema as you'd usually expect. I found this pretty clear and an improvement over the current helptext, but maybe you have an idea to improve it further.

mavam: I'm fine with the wording, just want to adjust my mental model to better understand IPC streams.

PR #2426, opened 2022-07-13
Use a partition-local ID space for new partitions

dominiklohmann:
This brings the query performance for Parquet and Feather stores up to the same level as the as the segment store, because it allows them to use the selection bitmap returned from the dense index lookup to reduce the number of rows required to evaluate.

This is a bit of a prototype: All integration tests pass with all store backends, which is what I've used for testing locally, but I did not adapt the unit tests yet to the changes and they yet fail to build.

~In addition, this also unveiled a shutdown issue with the new passive stores that I do not understand yet, but that we need to look into: Shutdown takes a long while even after queries completed. This is most noticeable with `vast --node export --max-events=1 json`, which yields its one result almost instantly but then takes seconds to shut down entirely. A similar behavior exists for `vast rebuild`, which is terribly slow with the new stores.~ Unrelated—this is because of the chunk / row-group sizes.

### :memo: Checklist

Note that this branches off of existing work in #2413.

~I'm on vacation for the next week, so either someone takes over this PR to finish it for v2.2 or we target this for v2.3. I wouldn't mind either way.~ We're targeting v2.2 for this.

~If you're that someone wanting to take over this PR, the~ The following yet needs to be done
- [x] Adapt the unit tests to the changes
- [x] ~Investigate why shutting down passive stores takes so long, and fix the issue~ Moved to a separate story
- [x] Double-check that all the changes made are backwards-compatible

Review-wise it's just a single commit so far that hopefully should be self-explanatory.


dominiklohmann:
[sc-34146]

libvast/src/system/catalog.cpp
```diff
@@ -491,62 +487,25 @@ catalog(catalog_actor::stateful_pointer<catalog_state> self,
...
       VAST_TRACE_SCOPE("{} {}", *self, VAST_ARG(query_context));
-      catalog_result expression_candidates;
-      std::vector<vast::uuid> ids_candidates;
       bool has_expression = query_context.expr != vast::expression{};
       bool has_ids = !query_context.ids.empty();
-      if (!has_expression && !has_ids)
+      if (has_ids)
+        return caf::make_error(ec::invalid_argument, "catalog expects queries "
+                                                     "not to have ids");
+      if (!has_expression)
```
tobim, line 495: You can replace this with an assertion now.

plugins/parquet/parquet.cpp
```diff
@@ -399,21 +401,26 @@ class passive_parquet_store final : public passive_store {
...
+
+  [[nodiscard]] size_t num_events() const override {
+    return num_events_;
   }
 
 private:
   std::vector<table_slice> slices_ = {};
   configuration parquet_config_ = {};
-  size_t num_rows_ = {};
+  size_t num_events_ = {};
```
tobim, line 423: In the context of parquet `num_rows_` is more appropriate for this data member.

plugins/parquet/parquet.cpp
```diff
@@ -426,8 +433,13 @@ class active_parquet_store final : public active_store {
   /// @returns An error on failure.
   [[nodiscard]] caf::error add(std::vector<table_slice> new_slices) override {
     slices_.reserve(new_slices.size() + slices_.size());
-    slices_.insert(slices_.end(), std::make_move_iterator(new_slices.begin()),
-                   std::make_move_iterator(new_slices.end()));
+    for (auto& slice : new_slices) {
+      if (slice.offset() == invalid_id)
+        slice.offset(num_events_);
```
tobim, line 438: When would this be the case? Shouldn't we assert instead?

libvast_test/src/events.cpp
```diff
@@ -209,24 +209,19 @@ events::events() {
   REQUIRE_EQUAL(rows(suricata_stats_log), 1u);
   REQUIRE_EQUAL(suricata_stats_log[0].columns(), 1u);
   // Assign IDs.
-  auto i = id{0};
   auto assign_ids = [&](auto& slices) {
+    auto i = id{0};
```
tobim, line 213: You can actually get rid of `assign_ids` entirely.

libvast/native-plugins/feather.cpp
```diff
@@ -74,37 +75,51 @@ auto wrap_record_batch(const table_slice& slice)
...
 };
 
 class active_feather_store final : public active_store {
   [[nodiscard]] caf::error add(std::vector<table_slice> new_slices) override {
     slices_.reserve(new_slices.size() + slices_.size());
-    slices_.insert(slices_.end(), std::make_move_iterator(new_slices.begin()),
-                   std::make_move_iterator(new_slices.end()));
+    for (auto& slice : new_slices) {
+      if (slice.offset() == invalid_id)
+        slice.offset(num_events_);
```
tobim, line 118: Same as for parquet: I think we should just reject `invalid_id` with an assert.

tobim (approved):
This is great work! I only have a few minor improvements.

libvast/src/system/active_partition.cpp
```diff
@@ -219,6 +218,7 @@ active_partition_actor::behavior_type active_partition(
     [=](caf::unit_t&, caf::downstream<table_slice_column>& out, table_slice x) {
       VAST_TRACE_SCOPE("partition {} got table slice {} {}",
                        self->state.data.id, VAST_ARG(out), VAST_ARG(x));
+      x.offset(self->state.data.events);
```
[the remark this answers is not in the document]
dominiklohmann: I've added comments to this as discussed in Slack.

libvast/src/system/catalog.cpp
dominiklohmann: Discussed in Slack; comment is not applicable

libvast/src/segment_builder.cpp
```diff
@@ -30,7 +30,10 @@ segment_builder::segment_builder(size_t initial_buffer_size,
   reset(id);
 }
 
-caf::error segment_builder::add(const table_slice& x) {
+caf::error segment_builder::add(table_slice x) {
+  if (x.offset() == invalid_id)
+    x.offset(num_events_);
+  VAST_ASSERT(x.offset() == num_events_);
```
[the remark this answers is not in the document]
dominiklohmann: We increment `num_events` by `x.rows()` only after this line. Reasoning I added in a comment.

PR #2427, opened 2022-07-13
Fix matcher doc typo

rdettai:
Fixing typo and discussing documentation

### :memo: Checklist

NA


mavam (approved):

rdettai:
https://github.com/tenzir/vast/blob/0226c38a946c34c326eb2cb9d41e2b46ee6737f0/web/docs/use-vast/detect/match-threat-intel.md#L342-L343

@mavam why couldn't we update threat intel from files through the client?

rdettai:
Ohter questions:
- why is the schema + indicator concept provided for Pulsedive in the matcher plugin and not the Feodo tracker?
- why are the matchers created client side not persisted like server side ones?

rdettai:
https://github.com/tenzir/vast/blob/0226c38a946c34c326eb2cb9d41e2b46ee6737f0/web/docs/use-vast/detect/match-threat-intel.md#L290-L293

This section should not be called Context Usablility (this is a copy/paste from the section above. Also the trailing `:::` is a typo

mavam:
> @mavam why couldn't we update threat intel from files through the client? The [Replace state at the server](https://vast.io/docs/use-vast/detect/match-threat-intel#replace-state-at-the-server) feature looks like an anti-pattern to me.

The idea behind state replacement is that it allows for shipping matcher "storage" easily. For example, one matcher backend is DCSO's [bloom](https://github.com/DCSO/bloom). Anyone can generate Bloom filters with `bloom` and directly load them into a matcher.

This use case of shipping intel "packages" in binary form is not yet that common, but the more we go into the edge, the more relevance such a feature has.

mavam:
@rdettai and I discussed this out-of-band.

> @mavam why couldn't we update threat intel from files through the client? The [Replace state at the server](https://vast.io/docs/use-vast/detect/match-threat-intel#replace-state-at-the-server) feature looks like an anti-pattern to me.

The documentation does a bad job at showing a good use case for this feature, which was intended for large (> GBs) updates that would be expensive to push through the client, and typically arrive on the node out-of-band through the filesystem. I'll rephrase this in the docs.

I looked at the code again and wanted to see if the following warning is correct:

> You must ensure manually that VAST does not write to the matcher state file while exchanging the underlying state file.

At a first glance, I actually think the statement is wrong and it *is* safe, given that a matcher actor operates entirely sequentially.

@dominiklohmann if we go from `self->request` to `self->await` in the implementation, would you agree that this is safe?

> Also the trailing ::: is a typo

I fixed this in a work-in-progress PR.

> This section should not be called Context Usablility

Also fixed.

> why is the schema + indicator concept provided for Pulsedive in the matcher plugin and not the Feodo tracker?

This is an oversight. For a complete example, we should do both.

> why are the matchers created client side not persisted like server side ones?

Because that would entail rewriting the server-side configuration. In my opinion, the ephemeral matchers are the ones that would be expendable. I see them as an interactive feature. What's missing is perhaps the ability to dump a matcher configuration so that it's easy to create a persistent matcher from an ephemeral one. But so far we haven't gotten a lot of feedback on this feature.

dominiklohmann:
> > You must ensure manually that VAST does not write to the matcher state file while exchanging the underlying state file.
> 
> At a first glance, I actually think the statement is wrong and it _is_ safe, given that a matcher actor operates entirely sequentially.
> 
> @dominiklohmann if we go from `self->request` to `self->await` in the implementation, would you agree that this is safe?

We already do `self->request(filesystem, ...).await(...)` when matchers reload their state for exactly this reason.

mavam:
> We already do `self->request(filesystem, ...).await(...)` when matchers reload their state for exactly this reason.

Ah right, so then should I update the docs? I think they make it sound more dangerous than it actually is.

PR #2429, opened 2022-07-14
Rename transform to pipeline

dispanser:
This PR attempts to rename everything related to pipelines which previously went by the name `transform`.

### :memo: Checklist

- [X] The changes are reflected on [vast.io](https://vast.io), if necessary.

It's difficult to see everything at once; maybe a good place to start is the doc diffs and go from there. Check that all classes and their members (header files)  properly reflect the desired changes. There's probably a bunch of  local variables that are not renamed properly , as it's difficult to track down al


mavam:
The website build fails, but it's not trivial to resolve. Some links point to a draft blog posts, which the local build via `yarn start` can handle well, but not when packaging things up with `yarn build`, which omits the blog post.

The annoying way to fix this is by simply writing down what links need to be changed in the last PR prior to the next release.

mavam:
@dispanser maybe replace the last merge commit with a rebase?

lava (approved):
Looks good to me; although the PR is huge it is ultimately only renaming things so the potential for errors should be very low.

As far as testing the backwards compatibility goes, I verified that the integration test still passes with an unchanged config file after my changes.

PR #2430, opened 2022-07-14
Implement optional dense indexes

patszt:
Creation and usage of optional dense index can be optional for a given type or field name.
It can be configured in vast.yaml under index in each rule.
By default the index is always created.

### :memo: Checklist

Starting with docs might be useful to grasp what the change is supposed to bring


changelog/unreleased/features/2430--optional-dense-indexes.md
```diff
@@ -0,0 +1,4 @@
+VAST can be configured to not create dense indexes for specific fields or
+types. To disable the creation of dense indexes the user should modify
+the index section in .yaml configuration file. For more informations see
+[here](https://vast.io/docs/setup-vast/tune#skip-dense-index-creation).
```
dominiklohmann, line 4: Please reword this to use active voice and present tense.

web/docs/setup-vast/tune.md
```diff
@@ -184,6 +183,34 @@ applies to a single field, `suricata.http.http.url`, and has false-positive rate
...
+        # don't create dense indexes for suricata.http.http.url fields
+      - targets:
+          - suricata.http.http.url
+        create-dense-index: false
+        # don't create dense indexes for addr type columns
+      - targets:
+          - :addr
+        create-dense-index: false
+```
+
```
dominiklohmann, line 213: Please reword this to use active voice and present tense.

Lead with what problem this solves for the user so that it's clear why they'd want to do this in the first place, i.e., the disk and memory usage vs. query performance trade-off.

dominiklohmann (commented):
I only looked at the docs so far; I'll leave the actual code review to the rest of the team.

web/docs/setup-vast/tune.md
patszt: I am not sure if memory usage is higher with the indexers. It can be, but it also may be not noticeable to the user.
Should i test it separately or just ignore the memory usage information?

vast.yaml.example
```diff
@@ -156,12 +156,17 @@ vast:
...
     # rules:
-    #   Every rule adjusts the false-positive rate of a set of targets.
+    #   Every rule adjusts the behaviour of VAST for a set of targets.
     #   VAST creates one synopsis per target. Targets can be either types
-    #   or field names. All rules for types other than string and address
-    #   are currently ignored.
+    #   or field names.
+    #
+    #   fp-rate - false positive rate. Only string and address type targets
+    #             are supported
```
lava, line 164: I'd probably say will only have an effect for string an address type targets.

libvast/include/vast/system/evaluator.hpp
```diff
@@ -76,6 +83,7 @@ struct evaluator_state {
 /// @pre `!eval.empty()`
 evaluator_actor::behavior_type
 evaluator(evaluator_actor::stateful_pointer<evaluator_state> self,
-          expression expr, std::vector<evaluation_triple> eval);
+          expression expr, std::vector<evaluation_triple> eval,
+          ids all_possible_ids_for_partition);
```
lava, line 87: Maybe this should be `optional<ids>` since you sometimes also send an empty `ids{}`

libvast/src/detail/partition_common.cpp
```diff
@@ -0,0 +1,42 @@
...
+  });
+}
+} // namespace
+
+ids get_ids_for_evaluation(
+  const std::unordered_map<std::string, ids>& type_ids,
+  const std::vector<system::evaluation_triple>& evaluation_triples) {
+  if (needs_evaluation_without_indexer(evaluation_triples)) {
+    auto all_ids = vast::ids{};
+    for (const auto& [_, ids] : type_ids) {
```
lava, line 32: Current style is to not use braces when the body is just a single line of code.

libvast/src/system/active_partition.cpp
```diff
@@ -62,6 +62,129 @@
...
+  fbs::PartitionSynopsisBuilder ps_builder(synopsis_builder);
+  ps_builder.add_partition_synopsis_type(
+    fbs::partition_synopsis::PartitionSynopsis::legacy);
+  ps_builder.add_partition_synopsis(ps->Union());
+  auto ps_offset = ps_builder.Finish();
+  fbs::FinishPartitionSynopsisBuffer(synopsis_builder, ps_offset);
+  return fbs::release(synopsis_builder);
+}
+
+void serialize(
```
lava, line 92: I think this needs a small comment to explain the effect it has on the state.

libvast/src/system/active_partition.cpp
```diff
@@ -412,114 +539,12 @@ active_partition_actor::behavior_type active_partition(
                            *self, self->state.persisted_indexers, valid_count);
                 return;
```
lava, line 540: The old code was not pretty here, nowadays we do have `detail::fanout_counter` to handle this pattern. It may make sense to use it here and then call `serialize()` from its completion handler.

lava (approved):
Looks great, and very good test coverage!

libvast/include/vast/system/evaluator.hpp
patszt: I have changed variable name so it no longer "lies"

libvast/src/system/active_partition.cpp
```diff
@@ -62,6 +62,129 @@
...
 namespace vast::system {
 
+namespace {
+
+bool should_skip_index_creation(const type& type,
+                                const qualified_record_field& qf,
+                                const std::vector<index_config::rule>& rules) {
+  if (type.attribute("skip").has_value()) {
+    return true;
+  }
```
lava, line 72: I just noticed we're also fixing sc-35515 with this PR.

Can you add one more integration test that imports anything with the `#skip` attribute, flushes the db, and then exports that event again to ensure this works, and add a separate changelog entry (ie. a new paragraph in the `.md` file) for the bugfix?

patszt: Will be added soon

web/docs/setup-vast/tune.md
```diff
@@ -184,6 +184,32 @@ applies to a single field, `suricata.http.http.url`, and has false-positive rate
...
+  creation rule should be applied.
+- `partition-index` Doesn't create partition indexes for given targets if set to
+  false. By default the value is set to true.
+#### Example
+
+```yaml
+vast:
+  index:
+    rules:
+        # don't create partition indexes for suricata.http.http.url fields
```
dominiklohmann, line 203:
```suggestion
        # Don't create partition indexes the suricata.http.http.url field.
```

dominiklohmann (approved):
Thanks for updating the docs. Just some minor requests for improvements.

web/docs/setup-vast/tune.md
dominiklohmann: I think we can omit it for now, the disk usage is much more important.

PR #2431, opened 2022-07-14
Make partition deletion resilient against oversize

tobim:
We now reingest data from oversized partitions on startup.

### :memo: Checklist

Testing instructions: You can increase the size of a partition files manually beyond the fbs size limit, then start VAST:
```
vast -N import zeek < libvast_test/artifacts/logs/zeek/conn.log
dd if=/dev/zero bs=1G seek=3 count=0 of=vast.db/index/<the_uuid>
vast start &
vast flush
```
After that sequence the database should be fully repaired.


tobim:
This still needs a fix for the index statistics.

nix/overlay.nix
```diff
@@ -117,4 +117,15 @@ in
...
+    version  = "0.1.3";
+    vendorSha256 = "sha256-Mw1cRIwmDS2Canljkuw96q2+e+z14MUcU5EtupUcTDQ=";
+    src = final.fetchFromGitHub {
+      rev = "v${version}";
+      owner = "satta";
+      repo = pname;
+      hash = "sha256-75QrtuOduUNT9g2RJRWUow8ESBqsDDXCMGVNQKFc+SE=";
+    };
+  };
 }
```
lava, line 131: I guess this commit is in the wrong PR?

libvast/include/vast/detail/generator.hpp
```diff
@@ -208,23 +208,16 @@ class [[nodiscard]] generator {
...
 
-  generator& operator=(generator&& other) noexcept {
-    if (m_coroutine)
-      m_coroutine.destroy();
-    m_coroutine = other.m_coroutine;
-    other.m_coroutine = nullptr;
-    return *this;
-  }
-
   iterator begin() {
```
lava, line 222: The commit introducing the move assignment was fixing a rule-of-five violation, which you seem to reintroduce by removing it. A better fix would probably to define the other version as `operator=(const generator&)`, which is how the assignment operator is supposed to look anyways.

libvast/include/vast/segment.hpp
```diff
@@ -31,6 +34,38 @@ class segment {
...
+
+    [[nodiscard]] bool equals(iterator other) const;
+
+    [[nodiscard]] segment::iterator::difference_type
+    distance_to(iterator other) const;
+
+  private:
+    flat_slice_iterator nested_;
+    interval_iterator intervals_;
+    chunk_ptr chunk_;
```
lava, line 66: Why do we need the chunk? If we do let's at least add a comment that this is intentional, by default an iterator doesn't keep it's container alive.

libvast/src/system/index.cpp
```diff
@@ -283,9 +287,14 @@ caf::error index_state::load_from_disk() {
...
       continue;
     auto ext = entry.path().extension();
-    if (ext.empty())
-      partitions.push_back(partition_uuid);
-    else if (ext == std::filesystem::path{".mdx"})
+    if (ext.empty()) {
+      if (entry.file_size() >= FLATBUFFERS_MAX_BUFFER_SIZE) {
+        auto store_path = dir / ".." / store_path_for_partition(partition_uuid);
+        if (std::filesystem::exists(store_path, err))
+          oversized_partitions.push_back(partition_uuid);
```
lava, line 294: What about the case where no store file exists? (ie. because VAST crashed while saving it) There will still be an oversized partition on disk, so we should probably at least issue a warning.

libvast/src/system/index.cpp
```diff
@@ -342,9 +351,47 @@ caf::error index_state::load_from_disk() {
...
+  // ingest and create partitions with large import time ranges, thereby messing
+  // up compaction. The latter has the inverse properties.
+  // While I think a dedicated transform is cleaner, the re-ingest approach is a
+  // lot simpler to implement, so we go with that.
+  detail::spawn_container_source(
+    self->system(),
+    [](std::vector<uuid> xs,
+       std::filesystem::path dir) -> detail::generator<table_slice> {
+      for (const auto& id : xs) {
+        VAST_INFO("index recovers {}", id);
```
lava, line 367: `index recovers partiton {}`

libvast/src/system/index.cpp
```diff
@@ -342,9 +351,47 @@ caf::error index_state::load_from_disk() {
...
+        auto part_path = dir / to_string(id);
+        auto chk = chunk::mmap(store_path);
+        if (!chk) {
+          VAST_WARN("index failed to recover data from {}: {}\n"
+                    "You can try to manually recover the data with\n"
+                    "$ lsvast {} | vast import json",
+                    store_path, store_path, chk.error());
+          continue;
+        }
+        auto seg = segment::make(std::move(*chk));
```
lava, line 378: Need to check the return value of  `segment::make()`

libvast/src/system/index.cpp
```diff
@@ -1237,32 +1284,81 @@ index(index_actor::stateful_pointer<index_state> self,
...
+                    VAST_WARN("failed to load partition for deletion at {} "
+                              "because its size of {} exceeds the maximum "
+                              "allowed size of {}. The index statistics will "
+                              "be incorrect until the database has been "
+                              "rebuilt and restarted",
+                              path, chunk->size(), FLATBUFFERS_MAX_BUFFER_SIZE);
+                    // !!! Fallback path. We try to erase the corresponding
+                    // segment store file if it exists.
+                    // TODO: Also attempt to erase parquet stores once that
+                    // implementation is merged.
```
lava, line 1297: It's merged now, so probably this needs to become a discussion story about extending the store plugin API to allow for lookup by partition id without having a store header available.

libvast/src/system/index.cpp
```diff
@@ -1237,32 +1284,81 @@ index(index_actor::stateful_pointer<index_state> self,
...
+                    auto store_path = store_path_for_partition(partition_id);
+                    self
+                      ->request<caf::message_priority::high>(
+                        self->state.filesystem, caf::infinite, atom::erase_v,
+                        path)
+                      .then(
```
lava, line 1303: Would it make sense to move this logic into a separate "fallback erase" handler?

libvast/src/system/index.cpp
```diff
@@ -969,6 +1009,13 @@ index(index_actor::stateful_pointer<index_state> self,
     },
     caf::policy::arg<caf::broadcast_downstream_manager<
       table_slice, vast::type, i_partition_selector>>{});
+  // Read persistent state.
```
lava, line 1012: Should we add a note that this needs to happen after initializing the stream stage because it may start importing new data?

libvast/src/system/index.cpp
```diff
@@ -1237,32 +1284,81 @@ index(index_actor::stateful_pointer<index_state> self,
...
+                        store_path)
+                      .then(
+                        [self, partition_id, store_path](atom::done) {
+                          VAST_DEBUG("{} erased partition store {} from "
+                                     "filesystem",
+                                     *self, partition_id);
+                        },
+                        [self, partition_id,
+                         store_path](const caf::error& err) {
+                          VAST_DEBUG("{} failed to erase store {} at {}, "
```
lava, line 1324: We're in a very special case recovery handler, so these can probably be safely higher than `DEBUG()`

lava (approved):
Looks good to me from a high-level perspective: When we attempt to erase partitions that are corrupted we try to guess the store file and erase that; and when we attempt to start VAST and encounter an oversized partition we try to fix the DB state by reimporting the data.

It should actually be possible to create an unit test for this by creating a regular partition, adjusting the seek position to >= 2G, and verifying that the data is queryable after the index has restarted.

libvast/include/vast/segment.hpp
tobim: We could also keep a pointer to the segment itself and get the chunk from it when we construct the table slice.

libvast/src/system/index.cpp
tobim: Oh dear, I hoped I could avoid this for the scope of this PR. Given that oversized partitions should be much rarer from this point forward I think we should keep this PR on segments only.

nix/overlay.nix
tobim: Moved to #2437.

libvast/src/system/index.cpp
tobim: In that case the data could also be stored in the archive. That is actually more likely, because the segment store should usually be quite a bit quicker with packing.

lava: Hm, fair enough. But if we just ignore it here we have this 2G file that we just silently ignore. Should we maybe issue a message and delete the partition?

libvast/src/system/index.cpp
lava: I guess the handshake would be blocked until the actor is actually spawned, so probably you're right.

vast/integration/vast_integration_suite.yaml
```diff
@@ -592,3 +592,14 @@ tests:
         expected_result: ignore
       - command: export --timeout=800ms json
         expected_result: ignore
+
+  Self Repair:
+    steps:
+      - command: -N import zeek
+        input: data/zeek/conn.log.gz
+        transformation: "@./misc/scripts/break-sizelimit.sh"
```
lava, line 600: Isn't `transformation` just piping it to this command? Is it guaranteed that the files are already written to disk when the script runs? If not, we can insert an additional `vast flush` step to attach the transformation to.

libvast/src/system/index.cpp
lava: Ideally "much rarer" will be "never" :D I agree on keeping this PR on segments, but let's at least reword the TODO

vast/integration/vast_integration_suite.yaml
tobim: We're using `vast -N` here, so the partition is guaranteed to exist.

libvast/src/system/index.cpp
tobim: A message maybe, but I'm against automatic deletion in this case, because then it will be impossible to recover the data even manually.

vast/integration/vast_integration_suite.yaml
lava: I don't see why? I thought this would expand to the equivalent of

```
vast -N import zeek | ./misc/scripts/break-sizelimit.sh
```

vast/integration/integration.py
```diff
@@ -259,13 +260,16 @@ def run_step(
             out = None
             if step.transformation:
                 LOGGER.debug(f"transforming output with `{step.transformation}`")
+                env = os.environ.copy()
+                env["INTEGRATION_DB_DIRECTORY"] = db_dir
```
dominiklohmann, line 264: This variable should be prefixed with `VAST_`.

vast/integration/vast_integration_suite.yaml
tobim: You are right! I totally missed that. I'll fix.

tobim: Oh, I see you already added another step. Nevermind then.

libvast/include/vast/detail/generator.hpp
```diff
@@ -208,14 +209,12 @@ class [[nodiscard]] generator {
...
-    if (m_coroutine)
+    if (m_coroutine) {
       m_coroutine.destroy();
+    }
   }
 
-  generator& operator=(generator other) noexcept {
-    swap(other);
-    return *this;
-  }
```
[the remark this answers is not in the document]
lava: It can result in an overload resolution failure:

```
$ cat assignment-operator.cpp 
#include <utility>

class A {
public:
	A() {}
	A(A&&) {}

	A& operator=(A other) { return *this; }
	A& operator=(A&& other) { return *this; }
};


int main() {
  A a1, a2;
  a1 = std::move(a2);
}
benno@bourbaki:~/code/cpp$ g++ assignment-operator.cpp 
assignment-operator.cpp: In function ‘int main()’:
assignment-operator.cpp:15:20: error: ambiguous overload for ‘operator=’ (operand types are ‘A’ and ‘std::remove_reference<A&>::type’ {aka ‘A’})
   15 |   a1 = std::move(a2);
      |                    ^
assignment-operator.cpp:8:12: note: candidate: ‘A& A::operator=(A)’
    8 |         A& operator=(A other) { return *this; }
      |            ^~~~~~~~
assignment-operator.cpp:9:12: note: candidate: ‘A& A::operator=(A&&)’
    9 |         A& operator=(A&& other) { return *this; }
      |            ^~~~~~~~
```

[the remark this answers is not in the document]
dominiklohmann: Thanks, that makes sense.

PR #2439, opened 2022-07-19
Fix `count_matching` when the calling store impl does not provide offset

dispanser:
This came up during feather testing: `count` yielded zero everywhere because vast failed to construct a valid `ids{}` set for invalid `offset`.

### :memo: Checklist

### DIFF
--- a/libvast/src/table_slice.cpp
+++ b/libvast/src/table_slice.cpp
@@ -769,7 +769,7 @@ std::optional<table_slice> filter(const table_slice& slice, const ids& hints) {
 uint64_t count_matching(const table_slice& slice, const expression& expr,
                         const ids& hints) {
   VAST_ASSERT(slice.encoding() != table_slice_encoding::none);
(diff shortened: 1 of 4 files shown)


libvast/src/table_slice.cpp
```diff
@@ -769,7 +769,7 @@ std::optional<table_slice> filter(const table_slice& slice, const ids& hints) {
 uint64_t count_matching(const table_slice& slice, const expression& expr,
                         const ids& hints) {
   VAST_ASSERT(slice.encoding() != table_slice_encoding::none);
-  const auto offset = slice.offset();
+  const auto offset = slice.offset() == invalid_id ? 0 : slice.offset();
```
dominiklohmann, line 772: Can you add a unit test for this?

dominiklohmann (approved):
Obvious fix and `filter` does the same already. Thanks!

PR #2442, opened 2022-07-20
Fix any and all on empty bitmaps

dominiklohmann:
These used to return the wrong values. Conjunctions on an empty set should always return true, and disjunctions should always return false. This did not have any impact on query results in any way luckily, because in practice we never ran these functions with empty bitmaps.

### :memo: Checklist

Double-check the logic.

Probably not worth a changelog entry since this had no user-facing impact, at least not from what I can tell.


patszt (approved):

dominiklohmann:
> How did you go after determining this reasoning?

This is the standard behavior for C++ in general, for example in pack expansion: `(xs && ...)` for an empty pack `xs` yields `true`, and `(xs || ...)` yields `false`.

PR #2445, opened 2022-07-20
Reset entire active partition state on decommission

dominiklohmann:
This change fixes a bug for repeated calls to `flush` or calls to flush with a very low `vast.active-partition-timeout` that was caused by us not resetting the entire state the index held for the active partition. Flushing twice with no additional events added in between should no longer cause errors, but rather return (almost) immediately and do nothing.

### :memo: Checklist

This should not need a changelog entry, as this was a newly introduced bug with `vast flush`, which was not part of any release yet.

Review-wise: Try to reproduce the bug locally before and after, and read the changes in the code.


patszt (approved):
I can't verify if the changes will fix the bug, but the code changes in the PR seem proper to me

PR #2446, opened 2022-07-20
Run commands from scripts

rdettai:
Currently, you can run commands on the VAST lambda or server only from a string passed as option to a CLI commands. This makes it hard to provide more complex commands, among other because of escaping issues. It would be simpler if we could execute commands from a script store in a file.

### :memo: Checklist

- docker is now binding the entire file system to make it possible to specify files that are outside the repository tree (@dispanser, this should also fix the git worktree issues) -> see `cloud/vast/vast-cloud`
- `--cmd` option of `run-lambda` and `execute-command` now gets the
- `run-lambda` and `execute-command` also take an `--env` list option to pass environment variables to the provided script. 
- also included an unrelated quick fix in `cloud/aws/terraform/common/s3notif/inputs.tf` because the dns resolution sometimes mysteriously fails with the default s3 domain


dispanser (approved):
tested with `run-lambda file:///...`

rdettai:
I will add a few tests before merging

rdettai:
https://github.com/tenzir/vast/actions/runs/2719686571

rdettai:
https://github.com/tenzir/vast/runs/7539753040

PR #2447, opened 2022-07-20
Fix `vast.export.json.omit-nulls` for nested records

dominiklohmann:
The `--omit-nulls` option of the JSON export overeagerly removed entire nested records whose first fields was null, instead of just removing the nested field.

### :memo: Checklist

The new unit and integration tests cover this extensively. The code change is rather simple to follow.

Let's increment the number of bugs caused by table slice column access using flat indices rather than offsets and move that up a bit on the roadmap 🙈


tobim (approved):
I carefully read the new unit test and it is enough to convince me that the function now behaves as expected.

PR #2449, opened 2022-07-20
Implement a flatbuffer container class to hold excess table slices in segments

lava:
This implements a new `SegmentedFile` flatbuffer as well as
    a new `flatbuffer_container` utility class that uses this
    flatbuffer to provide access to multiple flatbuffers stored
    in the same file.
    
This is similar in spirit to the size-prefixed flatbuffers
    provided by upstream, but allows random access into the
    contained flatbuffers which will become more relevant when
    we use the same approach for storing dense indices in partitions.
    
The segment is updated to optionally allow an implementation based
    on flatbuffer containers, which should allow us to store segments
    that hold more than 2GiB worth of table slice data.

### :memo: Checklist

There are three natural parts building on each other, the segmented file, the flatbuffer container and the segment. Review bottom-up or top-down as you wish.


libvast/include/vast/fbs/flatbuffer_container.hpp
```diff
@@ -0,0 +1,88 @@
...
+public:
+  // The `chunk` must begin with a flatbuffer that has
+  // `fbs::SegmentedFileHeader` as root type.
+  explicit flatbuffer_container(vast::chunk_ptr chunk);
+
+  /// Get the chunk at position `idx`.
+  chunk_ptr get_raw(size_t idx) const;
+
+  /// Return the chunk at position `idx` interpreted as a `T`.
+  template <typename T>
```
dominiklohmann, line 35:
```suggestion
  template <typename T>
    requires (std::is_trivial_v<T>)
```

Otherwise the below reinterpret_cast would be UB, as you'd need `std::start_lifetime_as` which afaict is not yet implemented in all compilers we support.

changelog/unreleased/bug-fixes/2449--oversized-segments.md
```diff
@@ -0,0 +1 @@
+VAST can now store data in segments bigger than 2GiB in size each.
```
dominiklohmann, line 1: This needs to be extended to cover partitions as well.

dominiklohmann (commented):
Just some minor comments on the code. I just started another test round with a database that would be corrupt without this change and will let that run over night.

dominiklohmann (approved):
Approving after an extensive pair-review session. We want to roll this out as VAST v2.3.0-rc2 for further testing.

lava:
Running this overnight seemed to be stable, and I could query the ingested data, so merging this now.

PR #2454, opened 2022-07-22
Detach the catalog actor from CAF's thread pool

dominiklohmann:
Running the catalog as a detached actor improved export performance under high load for me by ~5% on my local machine when spawning 500 queries in parallel with a ~100 GB database. The exact improvement depends on the workload, so YMMV.

The catalog is usually our busiest component, so giving it its own thread seemed like a good fit. We should measure the effects of this change in a large-scale deployment to find out whether we want to make it the default.

### :memo: Checklist

Measure!

### DIFF
--- a/libvast/src/system/catalog.cpp
+++ b/libvast/src/system/catalog.cpp
@@ -30,6 +30,7 @@
 #include "vast/time.hpp"
 
 #include <caf/binary_serializer.hpp>
+#include <caf/detail/set_thread_name.hpp>
 
 #include <type_traits>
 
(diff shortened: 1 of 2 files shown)


tobim:
Have you seen 6b4c333d373ed5003b2884fb2aa031953ccd9bb8?

dominiklohmann:
> Have you seen [6b4c333](https://github.com/tenzir/vast/commit/6b4c333d373ed5003b2884fb2aa031953ccd9bb8)?

Hah, I did not! You never opened a PR for that, did you?

tobim:
No, I did it for the catalog performance investigation. I think the explicit thread name is valuable.
I implemented that with the extra handler because I didn't know whether the actor function would be run in the dedicated thread context or not, and I didn't want to spend time figuring it out, but maybe we should do that now.

dominiklohmann:
I agree that the name is valuable; I'll figure out whether setting it immediately in the behavior-returning function works as part of this story.

dominiklohmann:
@tobim I checked in the code and we can just set it immediately, as the initialization code runs in the actor's context. I verified this by printing thread ids as well.

I updated the code to set thread names for the catalog and the filesystem.

image

tobim (approved):

PR #2455, opened 2022-07-22
Defer table slice IPC backing until serialization

dominiklohmann:
This reverts commit 6b4791e59e6e1a7df167fa5854f2d405492c8427 and implements fixes on top of it.

VAST's table slices currently have an IPC backing that we always create, even when we already have an in-memory record batch. This is very cheap for the Segment store, but unnecessarily expensive for the Feather and Parquet stores.

### :memo: Checklist

Follow the code changes file-by-file.

This should be covered by the tests.

Performance-wise, this makes a difference only when using Parquet/Feather stores or when using transforms extensively. I did not add a changelog entry, as we can much rather add the overall performance delta to the release notes.

### DIFF
--- a/libvast/src/arrow_table_slice.cpp
+++ b/libvast/src/arrow_table_slice.cpp
@@ -797,6 +797,14 @@ arrow_table_slice<FlatBuffer>::arrow_table_slice(
     // chunk at all, but rather create it on the fly only.
     if (batch) {
       state_.record_batch = batch;
+      // Technically we could infer an outer buffer here as Arrow Buffer
+      // instances remember which parent buffer they were sliced from, so we if
+      // know that the schema, the dictionary, and then all columns in order
+      // concatenated are exactly the parent-most buffer we could get back to
+      // it. This is in practice not a bottleneck, as we only create from a
+      // record batch directly if we do not have the IPC backing already, so we
+      // chose not to implement it and always treat the IPC backing as not yet
+      // created.
       state_.is_serialized = false;
     } else {
       auto decoder = record_batch_decoder{};
--- a/libvast/src/arrow_table_slice.cpp
+++ b/libvast/src/arrow_table_slice.cpp
@@ -754,9 +754,12 @@ index_column_arrays(const std::shared_ptr<arrow::RecordBatch>& record_batch) {
 
 template <class FlatBuffer>
 arrow_table_slice<FlatBuffer>::arrow_table_slice(
-  const FlatBuffer& slice, [[maybe_unused]] const chunk_ptr& parent) noexcept
+  const FlatBuffer& slice, [[maybe_unused]] const chunk_ptr& parent,
+  const std::shared_ptr<arrow::RecordBatch>& batch) noexcept
   : slice_{slice}, state_{} {
   if constexpr (std::is_same_v<FlatBuffer, fbs::table_slice::arrow::v0>) {
+    VAST_ASSERT(!batch, "pre-existing record batches are only supported for "
+                        "the most recent encoding");
     VAST_DIAGNOSTIC_PUSH
     VAST_DIAGNOSTIC_IGNORE_DEPRECATED
     // This legacy type has to stay; it is deserialized from disk.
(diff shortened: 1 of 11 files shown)


libvast/src/segment_builder.cpp
```diff
@@ -41,6 +41,13 @@ caf::error segment_builder::add(table_slice x) {
   VAST_ASSERT(x.offset() == num_events_);
   if (x.offset() < min_table_slice_offset_)
     return caf::make_error(ec::unspecified, "slice offsets not increasing");
+  if (!x.is_serialized()) {
+    auto serialized_x = table_slice{to_record_batch(x), true};
+    serialized_x.import_time(x.import_time());
+    serialized_x.offset(x.offset());
+    x = std::move(serialized_x);
+  }
+  VAST_ASSERT(x.is_serialized());
```
tobim, line 50: That's a bit much right after the logic to serialize.
```suggestion
```

libvast/src/table_slice.cpp
```diff
@@ -229,21 +231,10 @@ table_slice table_slice::unshare() const noexcept {
...
-  // Check whether the slices contain different data.
-  auto flat_layout = flatten(caf::get<record_type>(lhs.layout()));
-  for (size_t row = 0; row < lhs.rows(); ++row)
-    for (size_t col = 0; col < flat_layout.num_fields(); ++col)
-      if (lhs.at(row, col, flat_layout.field(col).type)
-          != rhs.at(row, col, flat_layout.field(col).type))
-        return false;
-  return true;
+  constexpr auto check_metadata = true;
+  return to_record_batch(lhs)->Equals(*to_record_batch(rhs), check_metadata);
```
tobim, line 237: We should probably still use the old logic for `msgpack`?

libvast/src/arrow_table_slice.cpp
```diff
@@ -790,9 +795,15 @@ arrow_table_slice<FlatBuffer>::arrow_table_slice(
     // table slice's chunk, and storing a sliced chunk in there would cause a
     // cyclic reference. In the future, we should just not store the sliced
     // chunk at all, but rather create it on the fly only.
-    auto decoder = record_batch_decoder{};
-    state_.record_batch = decoder.decode(
-      as_arrow_buffer(parent->slice(as_bytes(*slice.arrow_ipc()))));
+    if (batch) {
+      state_.record_batch = batch;
+      state_.is_serialized = false;
```
tobim, line 800: I guess Arrow has no function to figure this out? A comment would be helpful.

tobim (approved):
Code LGTM. A unit test to check correct behavior of `is_serialized` would be good though.

libvast/src/table_slice.cpp
dominiklohmann: I did consider that, although there's only very little of that remaining. We switched the default so long ago that I think we should not maintain additional code for it, and we are actively warning on startup if partitions with a partition version of 0 exist.

dominiklohmann: I added a comment explaining my reasoning.

tobim: That reasoning doesn't fully convince me. We know of old deployments that still have data from more than a year ago. Just because we don't use this function in prod right now doesn't mean we won't in the future, and in theory there could be plugins using it too.

libvast/test/arrow_table_slice.cpp
```diff
@@ -859,6 +859,12 @@ TEST(full_table_slice) {
   check_column(slice, 12, count_type{}, f2_count);     // f11_1_2
   check_column(slice, 13, address_type{}, f4_address); // f11_2_1
   check_column(slice, 14, pattern_type{}, f3_pattern); // f11_2_2
+  MESSAGE("test is_serilaized");
+  CHECK(slice.is_serialized());
+  auto slice2 = table_slice{to_record_batch(slice)};
+  CHECK(!slice2.is_serialized());
```
tobim, line 865: That should be the result after `filter`/`select`/`split` as well, right?

dominiklohmann: Yes; I'll add an additional check.

libvast/src/table_slice.cpp
dominiklohmann: I added it back.

PR #2457, opened 2022-07-22
Fix use-after-move for the active-partition-timeout

dominiklohmann:
We recently changed `decommission_active_partition` to invalidate iterators, but forgot to adapt the `vast.active-partition-timeout` trigger for it accordingly, so that operatored on schemas that were already moved-from.

### :memo: Checklist

Change is straightforward, just a slight re-ordering to account for the possible iterator invalidation.

I don't think this needs a changelog entry, I introduced this myself just recently in #2445.


lava (approved):
Looks correct to me

PR #2459, opened 2022-07-23
(WIP) Make sure terminated queries are removed from the backlog

tobim:
When the index receives a new query it monitors the client so it
can clean up related state in case said client quits early.
Unfortunately this mechanism doesn't work in case the client
terminates before the index starts monitoring. In that case the
index-side state never gets cleaned and the queries remain in a
pending state indefinitely.

This commit introduces an extra roundtrip from the index to the
client. In case that new request times out we clean out the
internal state, and effectively cover the blind window.

ATTENTION:
It seems that request responses from scoped/blocking actors don't
get routed back to the requester correctly, which is a problem for
the index unit test.

TODO:
The query-processor / counter don't work yet.

### :memo: Checklist

### DIFF
--- a/libvast/include/vast/query_queue.hpp
+++ b/libvast/include/vast/query_queue.hpp
@@ -25,6 +25,9 @@ struct query_state {
   /// The query client.
   system::receiver_actor<atom::done> client = {};
 
+  /// The response promise to deliver upon completion.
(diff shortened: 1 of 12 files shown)


dominiklohmann:
I'm closing this one, the issue at hand was resolved already.

PR #2460, opened 2022-07-25
Fix crash for aggregation functions returning lists

dominiklohmann:
This fixes a crash for aggregation functions with different input and output types, which just happened to work most of the time because the memory access into the storage of an Arrow ListArray works transparently, so we passed the unit tests. The additional assertions ensure that this works as expected. I double-checked that just adding the assertions causes the unit tests to fail.

### :memo: Checklist

Newly introduced bug with v2.2, so no changelog entry. I tested with this configuration locally:

```yaml
vast:
  pipelines:
    aggregate-flows:
      - summarize:
          group-by:
            - timestamp
            - src_ip
            - dest_ip
            - dest_port
            - proto
          time-resolution: 1 hour
          aggregate:
            timestamp_min:
              min: timestamp
            timestamp_max:
              max: timestamp
            flow.pkts_toserver: sum
            flow.pkts_toclient: sum
            flow.bytes_toserver: sum
            flow.bytes_toclient: sum
            flow.start: min
            flow.end: max
            flow.alerted: any
            src_port: distinct
            community_id: distinct
      - extend:
          fields:
            event_type: flow_agg
      - rename:
          schemas:
            - from: suricata.flow
              to: suricata.flow_agg

plugins:
  compaction:
    time:
      interval: 30 min
      step-size: 1
      rules:
        - after: 30 mins
          name: flow-reduction
          pipeline: aggregate-flows
          preserve-input: true
          types:
            - suricata.flow
```

And then run the following:

```
vast start
vast import --blocking suricata < path/to/eve.json
vast flush
vast compaction run --age=0s flow-reduction
```

This crashed (in release builds) before this change.

### DIFF
--- a/libvast/include/vast/arrow_table_slice.hpp
+++ b/libvast/include/vast/arrow_table_slice.hpp
@@ -289,6 +289,7 @@ template <concrete_type Type>
 view<type_to_data_t<Type>>
 value_at(const Type& type, const std::same_as<arrow::Array> auto& arr,
          int64_t row) noexcept {
+  VAST_ASSERT(type.to_arrow_type()->id() == arr.type_id());
   VAST_ASSERT(!arr.IsNull(row));
   if constexpr (arrow::is_extension_type<type_to_arrow_type_t<Type>>::value)
     return value_at(type, *caf::get<type_to_arrow_array_t<Type>>(arr).storage(),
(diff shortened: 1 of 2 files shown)


lava (approved):
Looks good; we already have a test for this (which didn't trigger because the previous behavior was UB)

dominiklohmann:
[sc-35875]

PR #2464, opened 2022-07-25
Fix the slow path of the count aggregation function

dominiklohmann:
This used to count *nil* values rather than values, i.e., do the opposite of what it was supposed to do. The fast-path for buckets with >= 2 contiguous rows worked correctly.

### :memo: Checklist

New code, so no changelog entry necessary.


patszt (approved):
Looks good!

PR #2467, opened 2022-07-26
Align default row-group-size with other stores

dominiklohmann:
This aligns the default Parquet store row-group size with the other stores' batch sizes. This speeds up selective queries in the Parquet store, and generally causes our system to run a bit smoother, as CAF isn't well-suited to handle single messages whose serialized size is a few GB.

### :memo: Checklist

New feature, so no changelog entry. The change should be simple to review, as it's only adjusting a default value.


mavam (approved):
Makes sense.

PR #2468, opened 2022-07-26
Fix bad documentation predicate

rdettai:
### :memo: Checklist

### DIFF
--- a/web/docs/use-vast/detect/match-threat-intel.md
+++ b/web/docs/use-vast/detect/match-threat-intel.md
@@ -271,7 +271,7 @@ feed_url='https://pulsedive.com/premium/?key=&header=true&fields=id,type,risk,th
 # Ingest the feed into the matcher 'ips', but skip all retired indicators.
 curl -sSL "${feed_url}" |
   vast matcher import -t pulsedive csv ips \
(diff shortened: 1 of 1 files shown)


rdettai:
@mavam from the docs, `ips` has `match-types: addr`, but the pulsedive `indicator.valuse` is `pulsedive.ioc` which is a `string` in `pulsedive.schema`, so we are loading a bunch of `string`s into a matcher that matches `addr`. Is that supposed to work?

web/docs/use-vast/detect/match-threat-intel.md
```diff
@@ -271,7 +271,7 @@ feed_url='https://pulsedive.com/premium/?key=&header=true&fields=id,type,risk,th
 # Ingest the feed into the matcher 'ips', but skip all retired indicators.
 curl -sSL "${feed_url}" |
   vast matcher import -t pulsedive csv ips \
-    'risk !~ ":retired" || (type != "ip" && type != "ipv6")'
+    'risk !~ /:retired/ && (type == "ip" || type == "ipv6")'
```
mavam, line 274:
```suggestion
    'risk !~ /:retired/ && type ~ /ip.*/
```

Would this be more idiomatic?

PR #2469, opened 2022-07-26
Fix serialization of empty chunks

dominiklohmann:
Empty chunks did not survive a CAF serialization roundtrip before this.

### :memo: Checklist

Build the matcher plugin against this branch, and create an empty exact filter matcher and try to save it to disk. This used to crash, because an empty chunk did not properly arrive at the client process. Alternatively, trust the added unit test.

This was not user-facing in VAST itself, so I don't think this should get a changelog entry.

### DIFF
--- a/libvast/src/chunk.cpp
+++ b/libvast/src/chunk.cpp
@@ -417,10 +417,10 @@ caf::error read(const std::filesystem::path& filename, chunk_ptr& x) {
 caf::error inspect(caf::serializer& sink, const chunk_ptr& x) {
   using vast::detail::narrow;
   if (x == nullptr)
-    return sink(uint32_t{0});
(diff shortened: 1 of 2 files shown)


libvast/src/chunk.cpp
```diff
@@ -417,24 +417,28 @@ caf::error read(const std::filesystem::path& filename, chunk_ptr& x) {
 caf::error inspect(caf::serializer& sink, const chunk_ptr& x) {
   using vast::detail::narrow;
   if (x == nullptr)
-    return sink(uint32_t{0});
+    return sink(int64_t{-1});
```
patszt, line 420: Naming this constant would be nice for the reader

patszt (approved):
Seems good after 1 to 1 discussion

PR #2470, opened 2022-07-27
Fix missing options sometimes not causing an error

dominiklohmann:
Missing arguments for the `--plugins`, `--plugin-dirs`, and `--schema-dirs` options on the command line caused an assertion failure in debug mode and ran into undefined behavior in release mode. In practice, this was unlikely to cause a crash.

### :memo: Checklist

Run `vast --plugin-dirs= start` and observe the error.


mavam (approved):
I didn't test this, but the replacement of the assertion with proper error handling is easy to grok.

PR #2471, opened 2022-07-28
Add memory-usage to index and catalog telemetry reports

patszt:
As suggested by @tobim it would be beneficial to have some memory usage being reported in the telemetry.
The measurement is not precise but it can still help when looking for a memory leak

### :memo: Checklist

### DIFF
--- a/libvast/src/query_queue.cpp
+++ b/libvast/src/query_queue.cpp
@@ -236,4 +236,15 @@ query_queue::handle_completion(const uuid& qid) {
   return result;
 }
 
+size_t query_queue::memusage() const {
+  auto usage = size_t{0u};
+  for (const auto& [uid, query_state] : queries_) {
+    usage += sizeof(uid) + sizeof(query_state);
+  }
+  usage += partitions.size() * sizeof(decltype(partitions)::value_type);
+  usage += inactive_partitions.size()
+           * sizeof(decltype(inactive_partitions)::value_type);
+  return usage;
+}
+
 } // namespace vast
--- a/libvast/src/query_queue.cpp
+++ b/libvast/src/query_queue.cpp
@@ -12,6 +12,15 @@
 
 namespace vast {
 
+namespace {
+std::size_t memusage(const std::vector<query_queue::entry>& entries) {
+  return std::accumulate(cbegin(entries), cend(entries), std::size_t{0u},
+                         [](const auto& accumulated, const auto& current) {
+                           return accumulated + current.memusage();
+                         });
+}
+} // namespace
+
 bool operator<(const query_queue::entry& lhs,
                const query_queue::entry& rhs) noexcept {
   return lhs.priority < rhs.priority;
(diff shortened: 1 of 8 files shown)


changelog/unreleased/features/2471--memory-usage-metrics-for-catalog-and-index.md
```diff
@@ -0,0 +1,3 @@
+The telemetry reports now contain memory usage for index and catalog components
+. These reports only count some part of used memory in order to help with
+debugging memory leaks
```
tobim, line 3:
```suggestion
VAST now emits per-component memory usage metrics under the `memory-usage` key.
As of now the index and catalog components calculate this value.
```

libvast/src/query_queue.cpp
```diff
@@ -235,4 +235,15 @@ query_queue::handle_completion(const uuid& qid) {
   return result;
 }
 
+size_t query_queue::memusage() const {
+  auto usage = size_t{0u};
+  for (const auto& [uid, query_state] : queries_) {
+    usage += sizeof(uid) + sizeof(query_state);
```
tobim, line 241: The `query_state` contains a `query_context` which in turn has multiple heap-stored members, I'm afraid we need a proper `memusage()` function for those.

libvast/src/system/index.cpp
```diff
@@ -793,12 +793,40 @@ void index_state::send_report() {
...
+std::size_t index_state::memusage() const {
+  auto calculate_usage = []<class T>(const T& collection) -> std::size_t {
+    return collection.size() * sizeof(typename T::value_type);
+  };
+  auto usage = std::size_t{0u};
+  for (const auto& [type, partition_info] : active_partitions) {
+    usage += as_bytes(type).size() + sizeof(partition_info);
+  }
+  usage += persisted_partitions.size()
+           * sizeof(decltype(persisted_partitions)::key_type);
```
tobim, line 819:
```suggestion
           * sizeof(decltype(persisted_partitions)::value_type);
```
I didn't know `unordered_set` had a `key_type`, but even then, `value_type` seems still more appropriate.

libvast/src/system/index.cpp
```diff
@@ -793,12 +793,40 @@ void index_state::send_report() {
...
+  usage += pending_queries.memusage();
+  for (const auto& [addr, uuids] : monitored_queries) {
+    usage += sizeof(addr) + calculate_usage(uuids);
+  }
+  usage += calculate_usage(flush_listeners);
+  usage += calculate_usage(partition_creation_listeners);
+  usage += calculate_usage(partitions_in_transformation);
+  return usage;
```
tobim, line 827: I believe this is missing a `sizeof(*this)`.

tobim (changes requested):
Almost good to go, just a few types need a real `memusage` instead of a bare `sizeof`.

tobim:
This also needs an update to `web/docs/setup-vast/monitor.md`.

libvast/src/query_queue.cpp
patszt: Do we wan't a memusage for expression? It seems that i must either visit it recursively or i need to have some counter that is updated and persisted on creation.

tobim: I agree, it isn't worth the effort.

tobim (approved):
Code wise this looks good to me.

PR #2475, opened 2022-07-31
Embed .github content in docs

mavam:
This PR embeds the content from our [.github](https://github.com/tenzir/.github) repository in the docs. 

Review the last commit and try it locally.


web/Makefile
```diff
@@ -9,20 +9,52 @@ NOTEBOOK_OUTPUT_DIR := docs/try-vast
...
 
+notebooks: $(NOTEBOOK_OUTPUTS)
+
 $(NOTEBOOK_OUTPUTS): $(NOTEBOOK_INPUTS)
 	@$(MAKE) -C $(NOTEBOOK_INPUT_DIR) md
 	cp -r $(NOTEBOOK_INPUT_DIR)/output/md/* $(NOTEBOOK_OUTPUT_DIR)
+
+.github:
+	git clone https://github.com/tenzir/.github.git
+	cd .github && git pull
```
lava, line 29: The second command is redundant; right after the clone you already have the latest state, and later on the recipe will not be executed again because the file with that name already exists.

If you want do execute a pull on every make invocation, the way to do it would be something like this:

```
.github:
    git clone https://github.com/tenzir/.github.git

update-dot-github: .github
    cd .github && git pull

$(DOT_GITHUB_OUTPUT_DIR)/code-of-conduct.md: update-dot-github
	@echo "---\nsidebar_position: 0\n---\n" > $@
	@cat .github/$(@F) >> $@
```

web/Makefile
```diff
@@ -9,20 +9,52 @@ NOTEBOOK_OUTPUT_DIR := docs/try-vast
...
+	@cat .github/$(@F) >> $@
+
+clean:
+	rm -f $(CHANGELOG_OUTPUT)
+	$(MAKE) -C $(NOTEBOOK_INPUT_DIR) clean
+	git clean -f -d -x $(NOTEBOOK_OUTPUT_DIR)
+	rm $(DOT_GITHUB_OUTPUT_DIR)/code-of-conduct.md
+	rm $(DOT_GITHUB_OUTPUT_DIR)/coding-style.md
+	rm $(DOT_GITHUB_OUTPUT_DIR)/security.md
+	rm $(DOT_GITHUB_OUTPUT_DIR)/workflow.md
```
lava, line 60: Why not just use `git clean` as well?
```
git clean -fd $(DOT_GITHUB_OUTPUT_DIR) 
```

If there's a good reason against, it would probably make sense to define a variable so you don't have to keep them in sync:
```
DOT_GITHUB_OUTPUT_FILES=  $(DOT_GITHUB_OUTPUT_DIR)/code-of-conduct.md \
  $(DOT_GITHUB_OUTPUT_DIR)/coding-style.md \
  $(DOT_GITHUB_OUTPUT_DIR)/security.md \
  $(DOT_GITHUB_OUTPUT_DIR)/workflow.md
```

Also, I think this also needs to remove the `.github` directory.

lava (approved):
Not much to review here; I assume you already tested locally that it looks good.

Feel free to just ignore the `Makefile` comments if you don't have time to implement them, it's just style nitpicks.

web/Makefile
mavam: The "good reason against" was mostly to not wipe somebody's working tree where they have something uncommitted. The `git clean` approach deletes that without further ado.

What you deem that as "good reason" or being overly cautious?

web/Makefile
mavam: Good catch. I removed the redundancy.

I thought about doing something that pulls on every call initially, but then decided it's not worth the complexity. It's so rare that we change `.github` that it felt weird accounting for it in the common case.

web/Makefile
lava: Yeah, sounds reasonable to me.

PR #2476, opened 2022-07-31
Add install.sh convenience script

mavam:
This PR is an attempt at providing and `install.sh` script that makes it easy to
install VAST, using the following idiom:

```bash
/bin/bash -c "$(curl -fsSL https://vast.io/install.sh)"
```

The idea is that it does The Right Thing based on the given platform. From a
user perspective, you can run `vast` in your shell afterwards.

### :memo: Checklist

The code is heavily inspired by Homebrew's installer. I've copied it over and then stripped it down.

It's not yet tested, so probably won't work. I just wanted to see where we land and start the conversation.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,0 +1,600 @@
+#!/bin/bash
+#
+# This script installs VAST in /opt/vast.
+#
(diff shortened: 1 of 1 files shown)


mavam:
Closing until it comes back up again.

PR #2478, opened 2022-08-01
First stab at new Python bindings

mavam:
This PR includes the first infrastructure work for a revamped version of Threat Bus in the form of VAST Python layer, sparked by two recent hackathons.

### :memo: Checklist

- [x] Simple `asyncio` MVP
- [x] In-memory fabric backbone
- [x] MISP 0mq data acqusition
- [x] MISP REST API initilization
- [x] MISP-to-STIX: complete event conversion
- [x] VAST-to-STIX: convert query results to STIX
- [x] Package via Poetry

To review this, make sure you have a VAST instance running locally. The tooling is not yet capable of spawning one, but that will come next.

The primary objective for the review would actually be the scaffolding, not so much the functions (which are still alpha). Please take a look at the CI workflow, the project setup (`pyproject.toml`), and the Poetry-based workflows. The README should make you get started so that you can enter environment to invoke an example.


mavam:
[sc-35523]

python/vast/vast.py
```diff
@@ -0,0 +1,164 @@
...
+                stderr=asyncio.subprocess.PIPE,
+            )
+
+        logger.debug(f"> vast {' '.join(self.args)}")
+        if not stdin:
+            return await run(*self.args, stdin=False)
+        proc = await run(*self.args, stdin=True)
+        proc.stdin.write(str(stdin).encode())
+        await proc.stdin.drain()
+        proc.stdin.close()
```
rdettai, line 50: I guess you took this out of pyvast, but I think this might lead to trouble if your command streams large inputs into large outputs as you are waiting for all of stdin to be buffered in before starting to read the output

python/vast/utils/arrow.py
```diff
@@ -0,0 +1,26 @@
+import ipaddress
+
+import pyarrow as pa
+
+
+class IPAddressType(pa.PyExtensionType):
```
dominiklohmann, line 6: In VAST this type is called `vast.address`, and its name in code is `address_type`. I think we should name it `AddressType` to match developer expectations here.

python/vast/utils/arrow.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self):
+        pa.PyExtensionType.__init__(self, pa.binary(16))
```
dominiklohmann, line 8: Not sure about the API, but don't you need to put the extension type identifier `vast.address` somewhere?

python/vast/utils/asyncio.py
```diff
@@ -0,0 +1,44 @@
...
+        logger.info(f"received exit signal {signal.name}")
+    tasks = [t for t in asyncio.all_tasks() if t is not asyncio.current_task()]
+    [task.cancel() for task in tasks]
+    logger.debug(f"cancelling {len(tasks)} outstanding tasks")
+    await asyncio.gather(*tasks, return_exceptions=True)
+    loop.stop()
+
+
+def handle_exception(loop, context):
+    # context["message"] will always be there, but context["exception"] may not
```
dominiklohmann, line 21: Why? Where is the context object here coming from?

python/vast/utils/config.py
```diff
@@ -0,0 +1,22 @@
...
+import dynaconf
+
+CONFIG_FILES = ["config.yaml", "config.yml"]
+
+
+def create(config_files=CONFIG_FILES):
+    return dynaconf.Dynaconf(
+        settings_files=config_files,
+        load_dotenv=True,
+        envvar_prefix="VAST",
```
dominiklohmann, line 11: I think this may lead to problems with VAST also running. Should we consider a separate prefix, e.g., `VAST_PYTHON`?

python/vast/vast.py
```diff
@@ -0,0 +1,164 @@
...
+            while True:
+                reader = pa.ipc.RecordBatchStreamReader(istream)
+                try:
+                    while True:
+                        batch = reader.read_next_batch()
+                        name = vast.utils.arrow.name(batch.schema)
+                        logger.debug(f"got batch of {name}")
+                        num_batches += 1
+                        num_rows += batch.num_rows
+                        batches[name].append(batch)
```
dominiklohmann, line 157: This is broken if the you have batches with the same schema name and a different physical layout. I recommend switching to another structure for `batches`. It should still be tables, but ideally a list of them rather than a dict.

dominiklohmann (commented):
🟢 Greenlight for the CI side of things and the Python code. For the Python things I had some minor remarks, but I think this is a good start.

This needs a lot of comments in code still. Reviewing this probably took me 5x longer than it should've because I needed to follow the code given that it's not yet ready to run locally, so for future reviews we should avoid that overhead by documenting from the get go.

python/vast/apps/db.py
```diff
@@ -0,0 +1,97 @@
...
+    indicator_to_vast_matcher_ioc,
+    matcher_result_to_sighting,
+)
+
+logger = logging.getLogger("db-app")
+
+
+class DB:
+    def __init__(self, fabric: Fabric):
+        self.db = pyvast.VAST(container={"runtime": "docker", "name": "vast-pro"})
```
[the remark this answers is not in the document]
mavam: The DB app was an experiment; it'll get absorbed by the VAST node in the future.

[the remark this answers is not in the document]
dominiklohmann: Please put this in a comment in the code. It's not clear otherwise, and that thought may get lost.

python/vast/utils/arrow.py
mavam: Since it's called `IPAddress` in the native Python library, I find this name more pythonic.

python/vast/utils/arrow.py
mavam: Good point, adding that.

python/vast/utils/asyncio.py
mavam: This is passed in here:
```python
    loop.set_exception_handler(handle_exception)
```

The `context` is part of the expected signature.

python/vast/utils/arrow.py
mavam: I removed the type.

python/vast/utils/config.py
mavam: I would like to see the two be really one in the future, which is why I think `VAST` is better suited. If this causes clashes, let's think about it.

python/vast/vast.py
mavam: I fixed this, but it required a workaround of going through the string representation of the schema:

```python
# TODO: this might be slow, but schemas are not
# hashable. The string representation is the next best
# thing to determine uniqueness.
batches[batch.schema.to_string()].append(batch)
```

python/vast/vast.py
mavam: I've added a comment to look out for this. Does that work for now?

dominiklohmann (approved):

PR #2479, opened 2022-08-01
Switch order of keyword expansion and native plugin removal

lava:
During startup, VAST silently removes all native plugins
from the list of plugins to be loaded (they're part of
libvast and don't need to be loaded) and then expands
the special keywords "bundled" and "all" into a list of plugins.

However, this breaks if we have native plugins called "bundled"
or "all". Which we do indeed have. As a workaround, we switch
around the order of the two steps so expansion happens first.

### :memo: Checklist


patszt (approved):
looks good

PR #2481, opened 2022-08-02
Use atoms for all CAF message handlers

patszt:
All CAF message handlers will have atom as a first argument.
This should help with code navigation

### :memo: Checklist


libvast/include/vast/system/actors.hpp
```diff
@@ -143,7 +143,7 @@ using evaluator_actor = typed_actor_fwd<
 /// The INDEXER actor interface.
```
lava, line 143: GitHub won't let me comment outside of the current diff, but the `receiver_actor` further up in this file still introduces a message handler without an atom:

```
template <class T, class... Ts>
using receiver_actor = typename typed_actor_fwd<
  // Add a new source.
typename caf::reacts_to<T, Ts...>>::unwrap;
```

lava (approved):
Great to finally see this completed :tada:

libvast/include/vast/system/actors.hpp
patszt: Hmm should it be enforced for the receiver_actor to have an atom? Or the users for receiver_actor should pass the atom?
Any suggestion for atom example for our use cases?

lava: I'm not sure I understand the distinction between the two cases, I was thinking of something like this:

```
template <class T, class... Ts>
using receiver_actor = typename typed_actor_fwd<
  // Add a new source.
typename caf::reacts_to<atom::receive, T, Ts...>>::unwrap;
```

patszt: Perfect, i was thinking about the same :)

PR #2486, opened 2022-08-04
Prepare repository for VAST v2.2.0

lava:
### :memo: Checklist


mavam:
In [b41475d](https://github.com/tenzir/vast/pull/2486/commits/b41475dfe532edfb219eb6c7df119850a36a1966), I accidentally wordmithed the wrong release 🙈. Feel free to drop this commit.

web/blog/vast-v2.2/index.md
```diff
@@ -96,3 +96,19 @@ summarize:
...
+
+## Regular Expression Predicates
+
+The light at the end of the tunnel for regular expression support is here. VAST
+can now execute query expressions with predicates of the form `x ~ /f.o.*bar/`.
+
+We introduced this functionality "accidentally" during a refactoring that
+transposed predicate evaluation from being row-wise to column-wise. That said,
+the performance is not anywhere near where we would deem it production-grade. We
+have planned a series of optimizations for the next coming releases. Stay tuned!
```
dominiklohmann, line 108: This still causes warnings in the evaluator (fixed by #2487), but that's not going to be part of this release. I would not want to announce this now, especially since we're currently instantiating the regex object once per event instead of once per batch or ideally once per query.

web/blog/vast-v2.2/index.md
```diff
@@ -96,3 +96,19 @@ summarize:
...
+
+## Catalog Lookup Tune-Up
+
+As part of our continuous effort to reduce the latency of VAST's catalog, the
+central component that evaluates queries to yield a candidate set of partitions,
+we improved the per-partition lookup time by XXX.
```
dominiklohmann, line 114: We should mention that the effect of this is most visible for users spawning many queries in parallel, as the rate at which the catalog does its lookup majorly impacts how well the scheduling algorithm works.

changelog/v2.2.0/changes/2417.md
```diff
@@ -4,5 +4,4 @@ operator are now plugins, which makes them easily extensible. The syntax of
 `summarize` now supports specification of output field names, similar to SQL's
 `AS` in `SELECT f(x) AS name`.
 
-The `count` pipeline operator no longer exists. If you relied on the
-undocumented operator's behavior, please let us know.
+The `count` pipeline operator no longer exists.
```
dominiklohmann, line 7: I would want to add the "undocumented" back here.

web/blog/vast-v2.2/index.md
mavam: Fair enough, I'm in fact also for _not_ announcing it. But then I thought there might be value for those folks that have just small amounts of testing data.

Feel free to scratch the block.

web/blog/vast-v2.2/index.md
mavam: Can you take a stab at tweaking the text?

changelog/v2.2.0/changes/2417.md
mavam: The sentence didn't provide any value in my eyes. I think this could be fixed by adding a single word: "the *undocumented* `count` pipeline operator ..."

web/blog/vast-v2.2/index.md
netantho: For what it's worth, by reading
`(dport <= 1024 || :addr in 10.0.0.0/8) && ! (#type == /zeek.*/)` in https://vast.io/docs/understand-vast/query-language/expressions, I thought it was already supported and was a bit confused about the warning I got on the server.

As alternative, could be announced as a alpha/beta feature with disclaimer about performance and warnings :-). It would give some lead time for users to integrate with it.

dominiklohmann: We'll do a pass over the docs page for expressions soon. For now, I've removed the section from the release blog post.

web/blog/vast-v2.2/index.md
dominiklohmann: We discussed in Slack that we'll be removing the section entirely, as the effect is minimal for most scenarios, and explaining the scenario under which this has a visible effect is too much for this blog post.

dominiklohmann (approved):
Good to go from my side; @lava you should probably look over all the changes again before cutting the release.

PR #2488, opened 2022-08-05
Fix broken caching in CI

dominiklohmann:
This is an attempt at repairing the broken caching in CI.

### :memo: Checklist

Review the slugs; other than that there isn't much we can do except for merging and seeing whether it works.


lava (approved):
Do I understand correctly that only the changes to `build-nix` are required for the bugfix and the rest is for improving caching behavior?

Either way, looks good to me.

dominiklohmann:
> Do I understand correctly that only the changes to `build-nix` are required for the bugfix and the rest is for improving caching behavior?

That actually just snuck in because I still had the change locally; it just avoids a merge conflict after the release PR is merged essentially.

PR #2492, opened 2022-08-11
Demote logging message that doesn't indicate error

mavam:
The issued warning is actually a non-issue and just confuses users (including myself).

n/t


dispanser (approved):
Makes a lot of sense

PR #2496, opened 2022-08-15
Fix artifact name for out-of-tree static binary builds

dominiklohmann:
This fixes the artifact name for out-of-tree static binary builds like we do in our proprietary plugin CI for the `nix/static-binary.sh` script, which now runs `git describe` with an explicit git directory set.

### :memo: Reviewer Checklist

### DIFF
--- a/nix/static-binary.sh
+++ b/nix/static-binary.sh
@@ -16,10 +16,12 @@ usage() {
   echo
 }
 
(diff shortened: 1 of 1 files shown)


mavam (approved):
`-C` seems to do the right thing here, per the `git` man page:

```
-C <path>
           Run as if git was started in <path> instead of the current working directory.
           When multiple -C options are given, each subsequent non-absolute -C <path> is
           interpreted relative to the preceding -C <path>. If <path> is present but
           empty, e.g.  -C "", then the current working directory is left unchanged.
```

PR #2503, opened 2022-08-17
Respect `--connection-timeout` in more places

dominiklohmann:
This expands on the work from #2499 and adds support for the new option to some additional places.

### :memo: Reviewer Checklist


libvast/src/system/node.cpp
```diff
@@ -268,13 +268,7 @@ caf::message dump_command(const invocation& inv, caf::actor_system&) {
                                              "type-registry"));
   caf::error request_error = caf::none;
   auto rp = self->make_response_promise();
-  // The overload for 'request(...)' taking a 'std::chrono::duration' does not
-  // respect the specified message priority, so we convert to 'caf::duration'
-  // by hand.
-  const auto timeout = caf::duration{defaults::system::initial_request_timeout};
-  self
-    ->request<caf::message_priority::high>(type_registry, timeout, atom::get_v,
```
lava, line 276: Why are we not using a timeout anymore here? (same question applies to the other places below where it was removed)

libvast/src/system/node_control.cpp
```diff
@@ -8,22 +8,47 @@
...
+  auto timeout = caf::duration{defaults::system::node_connection_timeout};
+  if (auto connection_timeout_arg
+      = caf::get_if<std::string>(&options, "vast.connection-timeout")) {
+    if (auto batch_timeout = to<duration>(*connection_timeout_arg))
+      timeout = caf::duration{*batch_timeout};
+    else
+      VAST_WARN("client cannot set vast.connection-timeout to {} as it "
+                "is not a valid duration",
+                *connection_timeout_arg);
+  }
```
lava, line 32: Should we create a story for making a helper function for the "get as string and parse afterwards" pattern? (we could also fix it right now, but this sounds like a good small starter story that depends on nothing else)

libvast/src/system/node.cpp
```diff
@@ -331,7 +325,10 @@ caf::message dump_command(const invocation& inv, caf::actor_system&) {
         }
       },
       [=](caf::error& err) mutable {
-        request_error = std::move(err);
+        request_error = caf::make_error(ec::unspecified,
+                                        "'dump' failed to get taxonomies from "
+                                        "type-registry: {}",
+                                        std::move(err));
```
lava, line 331: This seems to be missing a call to `fmt::format()`

Also, not related to this PR, but rust's anyhow has this cool concept of adding additional context to an error. Since we use this pattern a lot I wonder if we could implement a helper that works in spririt like this:

```
  request_error = err.context("while trying to get taxonomies from type-registry in 'dump'");
```

lava (approved):
Looks good to me overall.

libvast/src/system/node.cpp
dominiklohmann: Because this is always running with the same process, so using a timeout just doesn't make sense here. I did not understand why we were using one in the first place.

libvast/src/system/node_control.cpp
dominiklohmann: My main worry is about the warning, but I suppose we could have that function return an `expected<duration>` for that purpose. I'm not sure it's worth it, though, considering this is so far only in two places in the code base.

libvast/src/system/node.cpp
dominiklohmann: We actually have an epic internally for error contexts and rendering errors more nicely, but we've just never gotten around to doing it.

The call to format is definitely missing, nice catch.

I was thinking about poisoning `caf::make_error` and introducing our own `vast::make_error` that always goes through {fmt}. That would avoid this rather common error.

libvast/src/system/node.cpp
mavam: I strongly disagree. Timeouts are not just for IPC. Every request should have a timeout, ideally as low as possible. If a request times out, then ideally we try again, but keeping requests timeouts low dramatically increases the responsiveness of a system.

There's a great talk from the WhatsApp team that doubled down on this point in their Erlang stack. I can't find it right now though.

dominiklohmann: I understand this generally and I've actually seen that talk—it's a great one—, but CAF is making this really hard in practice. Specifically, I am thinking of two issues:

1. Streaming and regular messages have different queues, with streaming always being preferred. This can lead to requests not going through when using timeouts when an actor is busy while streaming. 

3. CAF's response promises cannot by design know when they're cancelled, so the work of the request will continue even throughout multiple stages without us ever being able to know whether the source of the request is still alive. E.g., if you have a request from actor A going to actor B that delegates to actor C, then B will always delegate to C regardless of whether the request from A to B timed out. This is a design flaw in CAF that would cause repeated requests to do repeated work regardless of whether or not they were cancelled.

mavam: Good points.

In my view, these issues make it harder to find the right *tight* lower bound on the timeout, but it doesn't invalidate the low-timeout approach in principle.

I can see the CAF Streaming throws a wrench in the timeout computation, as high load can cause infinite delays. Is there a fix in CAF to use the same queue? In theory, that shouldn't be hard.

The response cancellation seems less of an issue in this case. It does the right thing by entering the error case when no *response* has been received in the provided time frame. You're asking for "exactly-once" semantics of the work elicited by the request. I think that requirement is not needed for the common case, and needs a different primitive when it's required.

dominiklohmann: I would like to decouple the timeout discussion from this PR. If we're doing this then we need to discuss it more generally, but having a few timeouts of seemingly random durations throughout VAST hurts developers in debugging because they lead to very unexpected failures if VAST is really busy, especially with streaming and cascading requests.

This will hopefully get better with CAF 0.19, but we have yet to see the new streaming implementation on top of flows, so who really knows. For now, this is not something that we can or should do, and this also makes it so we should not re-open this discussion before we upgraded to CAF 0.19.

libvast/src/system/node_control.cpp
dominiklohmann: Done and implemented.

libvast/src/system/node.cpp
mavam: Agreeing here, half-heartedly. I'll leave it to @lava to resolve this conversation, assuming you are okay with the approach taken.

libvast/src/system/node_control.cpp
lava: So far it's only used in two places, but I think this is actually the intended behavior in 99% of places where currently `caf::get_if()` is used.

PR #2507, opened 2022-08-17
Display store load failures to the user

dominiklohmann:
Before this change, when a store failed to load users would get an opaque error message like this:

> [11:11:19.496] index-14 failed to evaluate query ac71d436-6b82-415d-8048-a0ba0ef0426e for partition e6522411-fa5c-4cf4-85ca-ff60fc1852ee: !! request_receiver_down

This can be easily reproduced by calling `self->quit(...)` in the `self->request(filesystem, ...).await(...)` handler that loads the store from a chunk in `vast/store.cpp`.

Instead, the partitions now display an error message like this and shut themselves down:

> [17:17:35.725] partition-5645eed0-d37e-457d-8b15-ea9d58498353-66 cannot reach its feather store and shuts down: <error from store>

This helps both the user understand what went wrong and developers in debugging.

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/src/system/passive_partition.cpp
+++ b/libvast/src/system/passive_partition.cpp
@@ -196,9 +196,20 @@ partition_actor::behavior_type passive_partition(
   self->state.filesystem = std::move(filesystem);
   self->state.name = "partition-" + id_string;
   VAST_TRACEPOINT(passive_partition_spawned, id_string.c_str());
+  self->set_down_handler([=](const caf::down_msg& msg) {
+    if (msg.source != self->state.store.address()) {
+      VAST_WARN("{} ignores DOWN from unexpected sender: {}", *self,
+                msg.reason);
+      return;
+    }
+    VAST_ERROR("{} cannot reach its {} store and shuts down: {}", *self,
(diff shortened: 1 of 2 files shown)


libvast/src/system/passive_partition.cpp
```diff
@@ -196,9 +196,20 @@ partition_actor::behavior_type passive_partition(
...
   self->state.name = "partition-" + id_string;
   VAST_TRACEPOINT(passive_partition_spawned, id_string.c_str());
+  self->set_down_handler([=](const caf::down_msg& msg) {
+    if (msg.source != self->state.store.address()) {
+      VAST_WARN("{} ignores DOWN from unexpected sender: {}", *self,
+                msg.reason);
+      return;
+    }
+    VAST_ERROR("{} cannot reach its {} store and shuts down: {}", *self,
+               self->state.store_id, msg.reason);
```
lava, line 206: The grammar sounds a bit off, I suggest `{} shuts down after DOWN from store {}: {}"`.

lava (approved):
Looks good; still missing a changelog entry.

PR #2509, opened 2022-08-18
Make partition transformer comply to the dense index configuration

dispanser:
Fixes a bug in which `vast rebuild --all` produced fully indexed partitions, because it was ignoring the index configuration.

### :memo: Reviewer Checklist

We don't need a changelog entry because the feature (optional dense indexes) hasn't been there before so there's no actual change for the previous rebuild behavior.

### DIFF
--- a/libvast/include/vast/system/active_partition.hpp
+++ b/libvast/include/vast/system/active_partition.hpp
@@ -37,6 +37,11 @@
 
 namespace vast::system {
 
+/// Determines whether the index creation should be skipped for a given field.
+bool should_skip_index_creation(const type& type,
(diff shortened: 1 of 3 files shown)


dominiklohmann:
I can confirm that this is working in practice; @dispanser and I paired on this quickly.

lava (approved):
The change looks innocent enough, and it's already tested :)

Do we have a follow-up story to also care about custom false-positive rates in the partition transformer?

dominiklohmann:
> Do we have a follow-up story to also care about custom false-positive rates in the partition transformer?

That should already be working, the index_config is passed to the index factory correctly.

PR #2513, opened 2022-08-19
Build a Debian package around the static binary

tobim:
It is now possible to build a `package` target to create Debian package, for example `vast_2.3.0-1_amd64.deb`.

### Todo:

- [x] Change Nix to do a package only build to avoid creating the binaries twice
- [x] Upload the package to GCS & Extend installation instructions.

### :memo: Reviewer Checklist


mavam (commented):
Maybe we can move the `debian/` dir to `scripts/debian/`?

debian/postinst
```diff
@@ -0,0 +1,55 @@
...
+        dpkg-statoverride \
+            --quiet --update \
+            --add vast vast 0750 /var/log/vast
+    fi
+fi
+
+# Automatically added by dh_installsystemd/13.3.4
+if [ "$1" = "configure" ] || [ "$1" = "abort-upgrade" ] || [ "$1" = "abort-deconfigure" ] || [ "$1" = "abort-remove" ] ; then
+	# This will only remove masks created by d-s-h on package removal.
+	deb-systemd-helper unmask 'vast.service' >/dev/null || true
```
dominiklohmann, line 28: If the postinstall script references our systemd service then we should actually install it like we do for rc.d and launchd.

cmake/Package.cmake
```diff
@@ -0,0 +1,24 @@
+# -- packaging -----------------------------------------------------------------
+
+set(CPACK_PACKAGING_INSTALL_PREFIX "/usr")
```
dominiklohmann, line 3: Shouldn't this respect the CMAKE_INSTALL_PREFIX?

nix/vast/default.nix
```diff
@@ -123,6 +126,24 @@ stdenv.mkDerivation rec {
...
+      -UCMAKE_INSTALL_INCLUDEDIR \
+      -UCMAKE_INSTALL_OLDINCLUDEDIR \
+      -UCMAKE_INSTALL_MANDIR \
+      -UCMAKE_INSTALL_INFODIR \
+      -UCMAKE_INSTALL_DOCDIR \
+      -UCMAKE_INSTALL_LIBDIR \
+      -UCMAKE_INSTALL_LIBEXECDIR \
+      -UCMAKE_INSTALL_LOCALEDIR \
+      -DCMAKE_INSTALL_PREFIX="/usr" ..
+    cmake --build . --target package -j $NIX_BUILD_CORES
```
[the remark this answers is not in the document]
tobim: 1. Afaik resetting these variables on an existing build tree has no effect.
2. Nix build anything incrementally.

nix/static-binary.sh
```diff
@@ -105,7 +105,7 @@ EOF
 fi
 
 echo running "nix-build --no-out-link -E \'${exp}\'"
-result=$(nix-build --no-out-link -E "${exp}")
+result=$(nix-build --keep-failed --no-out-link -E "${exp}")
```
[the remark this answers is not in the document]
tobim: I can change that.

cmake/Package.cmake
```diff
@@ -0,0 +1,24 @@
...
+set(CPACK_PACKAGE_VENDOR "Tenzir")
+set(CPACK_PACKAGE_CONTACT "engineering@tenzir.com")
+set(CPACK_VERBATIM_VARIABLES ON)
+
+set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE")
+set(CPACK_RESOURCE_FILE_README "${CMAKE_CURRENT_SOURCE_DIR}/README.md")
+set(CPACK_INSTALLED_DIRECTORIES "/var/lib/vast" "/var/log/vast")
```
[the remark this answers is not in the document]
tobim: No, `/var` is always `/var`. At least that is how I understand the FHS.

cmake/Package.cmake
tobim: Yeah I think we can just remove that.

CMakeLists.txt
```diff
@@ -736,6 +742,11 @@ install(
         "${CMAKE_CURRENT_SOURCE_DIR}/README.md"
   DESTINATION "${CMAKE_INSTALL_DOCDIR}")
 
+
+# -- packaging -----------------------------------------------------------------
+
+include(Package)
```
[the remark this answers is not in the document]
tobim: The current custom tar.gz thing is broken because the nix build doesn't do a regular install anymore. The way to go is to use CPack to create that as well. Once we have component based installation the output should be equivalent.

CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ project(
   VERSION
     ${VAST_VERSION_MAJOR}.${VAST_VERSION_MINOR}.${VAST_VERSION_PATCH}.${VAST_VERSION_TWEAK}
   DESCRIPTION "Visibility Across Space and Time"
-  HOMEPAGE_URL "https://github.com/tenzir/vast"
+  HOMEPAGE_URL "https://vast.io"
```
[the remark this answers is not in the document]
tobim: Ah, I didn't know that is used somewhere. I just thought `vast.io` is more appropriate.

CMakeLists.txt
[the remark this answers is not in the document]
tobim: I changed things so we now get a `.tar.zst` from CPack.

nix/vast/default.nix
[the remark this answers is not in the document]
tobim: The version you saw was already out of date when you commented. These `-U`s are now done as part of the configuration step. 
The function `stdenv.mkDerivation` doesn't have an explicit packaging phase, so we create a *package* variant instead now.

scripts/dbsize.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+df --output=pcent "$1" | tr -dc ‘0-9’
```
dominiklohmann, line 3: This requires GNU coreutils' version of `df`, so in order to at least also make this work on macOS (which I would much appreciate for testing actually), please change it like so:

```bash
PATH="$(brew --prefix coreutils)/libexec/gnubin:${PATH}" df --output=pcent . | tr -dc '0-9'
```

(Or make it a bit smarter to not always modify $PATH, but instead to only do so if `brew --prefix coreutils` doesn't fail.)

CMakeLists.txt
```diff
@@ -724,6 +724,14 @@ if (VAST_ENABLE_CORELIGHT_CAT)
           DESTINATION ${CMAKE_INSTALL_BINDIR})
 endif ()
 
+option(VAST_ENABLE_DBSIZE_SH "Install dbsize.sh with VAST" ON)
```
dominiklohmann, line 727: We should probably rename that script to `vast-du-pcent.sh` or so, and add an explanatory comment to inside the script given that we now install it.

CMakeLists.txt
```diff
@@ -736,6 +744,11 @@ install(
         "${CMAKE_CURRENT_SOURCE_DIR}/README.md"
   DESTINATION "${CMAKE_INSTALL_DOCDIR}")
 
+
+# -- packaging -----------------------------------------------------------------
+
+include(Package)
```
dominiklohmann, line 750: Can you rename the file from `Package.cmake` to `VASTPackage.cmake`? This has become common practice to be able to disambiguate which project a CMake include came from.

cmake/Package.cmake
```diff
@@ -0,0 +1,23 @@
...
+set(CPACK_DEBIAN_COMPRESSION_TYPE gzip)
+set(CPACK_DEBIAN_FILE_NAME DEB-DEFAULT)
+set(CPACK_DEBIAN_PACKAGE_NAME vast)
+set(CPACK_DEBIAN_PACKAGE_RELEASE 1)
+set(CPACK_DEBIAN_PACKAGE_SECTION "contrib/database")
+set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA
+    "${CMAKE_CURRENT_SOURCE_DIR}/debian/postinst"
+    "${CMAKE_CURRENT_SOURCE_DIR}/debian/postrm"
+    "${CMAKE_CURRENT_SOURCE_DIR}/debian/prerm")
+set(CPACK_DEBIAN_PACKAGE_DEBUG ON)
```
dominiklohmann, line 21: Should this be removed before merging?

cmake/Package.cmake
```diff
@@ -0,0 +1,23 @@
+# -- packaging -----------------------------------------------------------------
+
+set(CPACK_PACKAGE_VENDOR "Tenzir")
```
dominiklohmann, line 3: We might wanna put Tenzir GmbH here, not sure exactly.

CMakeLists.txt
```diff
@@ -724,6 +724,14 @@ if (VAST_ENABLE_CORELIGHT_CAT)
           DESTINATION ${CMAKE_INSTALL_BINDIR})
 endif ()
 
+option(VAST_ENABLE_DBSIZE_SH "Install dbsize.sh with VAST" ON)
+add_feature_info("VAST_ENABLE_DBSIZE_SH" VAST_ENABLE_DBSIZE_SH
+                 "install dbsize.sh with VAST.")
+if (VAST_ENABLE_DBSIZE_SH)
+  install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/scripts/dbsize.sh"
+          DESTINATION ${CMAKE_INSTALL_LIBEXECDIR})
```
dominiklohmann, line 732:
```suggestion
          DESTINATION "${CMAKE_INSTALL_LIBEXECDIR}")
```

cmake/Package.cmake
```diff
@@ -0,0 +1,23 @@
...
+set(CPACK_PACKAGE_CONTACT "engineering@tenzir.com")
+set(CPACK_VERBATIM_VARIABLES ON)
+
+set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE")
+set(CPACK_RESOURCE_FILE_README "${CMAKE_CURRENT_SOURCE_DIR}/README.md")
+set(CPACK_INSTALLED_DIRECTORIES "/var/lib/vast" "/var/log/vast")
+
+set(CPACK_GENERATOR TZST;DEB)
```
dominiklohmann, line 11:
```suggestion
set(CPACK_GENERATOR "TZST")

if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
  list(APPEND CPACK_GENERATOR "DEB")
endif ()
```

cmake/Package.cmake
```diff
@@ -0,0 +1,23 @@
...
+set(CPACK_DEBIAN_COMPRESSION_TYPE gzip)
+set(CPACK_DEBIAN_FILE_NAME DEB-DEFAULT)
+set(CPACK_DEBIAN_PACKAGE_NAME vast)
```
dominiklohmann, line 14:
```suggestion
set(CPACK_DEBIAN_COMPRESSION_TYPE "gzip")
set(CPACK_DEBIAN_FILE_NAME "DEB-DEFAULT")
set(CPACK_DEBIAN_PACKAGE_NAME "vast")
```

For consistency with the rest of our CMake.

dominiklohmann (commented):
This is looking great. I had some minor comments inline, and modulo manual testing this is ready to go from my side. Obviously still needs the Runtime/Development component split to be perfect, but I'll do that in a separate PR once this is merged to avoid the conflicts.

Thanks so much for working on this! 🎉

cmake/Package.cmake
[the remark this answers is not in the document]
dominiklohmann: I stand corrected. Thanks for double-checking.

nix/vast/default.nix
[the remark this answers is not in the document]
dominiklohmann: Yeah this is looking much better now. Thanks!

CMakeLists.txt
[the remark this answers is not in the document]
dominiklohmann: `git grep \.tar\.gz` still yields quite a few links that will be broken after the next release with this merged. Are you sure we don't just wanna stick with the more widely available `.tar.gz`?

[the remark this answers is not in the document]
tobim: Reverted to tar.gz for link stability. 
Although `zst` is commonly accepted in most places nowadays.

cmake/Package.cmake
tobim: I would keep it. The extra output isn't overly verbose:
```
CPackDeb:Debug: CPACK_TOPLEVEL_DIRECTORY          = '/home/tobim/t/vast/deb/build/gcc-static/redeb/_CPack_Packages/Linux/DEB'
CPackDeb:Debug: CPACK_TOPLEVEL_TAG                = 'Linux'
CPackDeb:Debug: CPACK_TEMPORARY_DIRECTORY         = '/home/tobim/t/vast/deb/build/gcc-static/redeb/_CPack_Packages/Linux/DEB/VAST-2.3.0-Linux'
CPackDeb:Debug: CPACK_OUTPUT_FILE_NAME            = 'vast_2.3.0-1_amd64.deb'
CPackDeb:Debug: CPACK_OUTPUT_FILE_PATH            = '/home/tobim/t/vast/deb/build/gcc-static/redeb/vast_2.3.0-1_amd64.deb'
CPackDeb:Debug: CPACK_PACKAGE_FILE_NAME           = 'VAST-2.3.0-Linux'
CPackDeb:Debug: CPACK_PACKAGE_INSTALL_DIRECTORY   = 'VAST 2.3.0'
CPackDeb:Debug: CPACK_TEMPORARY_PACKAGE_FILE_NAME = '/home/tobim/t/vast/deb/build/gcc-static/redeb/_CPack_Packages/Linux/DEB/vast_2.3.0-1_amd64.deb'
CPackDeb:Debug: CPACK_DEBIAN_PACKAGE_CONTROL_STRICT_PERMISSION = ''
CPackDeb:Debug: CPACK_DEBIAN_PACKAGE_SOURCE       = ''
```

cmake/Package.cmake
tobim: "Tenzir" should be enough.

cmake/Package.cmake
tobim: The Generator selection is now done from the command line.

CMakeLists.txt
tobim: I renamed it to `vast-df-percent.sh` and modified it for portability.

scripts/debian/postinst
```diff
@@ -22,6 +22,10 @@ if [ "$1" = "configure" ]; then
     fi
 fi
 
+if [ "$1" = "configure" ] ; then
+	ln -sf /opt/vast/lib/systemd/system/vast.service /lib/systemd/system/
+fi
```
dominiklohmann, line 27: Is there documentation for this that you can add as a comment?

tobim:
This is now already complete, I already tested `dpkg -i`, `apt remove`, and `apt purge`.

The last things I want to add are
- [ ] the `-rc` suffix is missing from the deb filename.
- [ ] The deb package from the debian ci build should be uploaded as artifact
- [ ] The docs need a slight adaptation

I think I can check these off this evening.

CMakeLists.txt
```diff
@@ -100,6 +99,7 @@ else ()
 endif ()
 unset(_VAST_PARENT_DIRECTORY)
 
+
```
dominiklohmann, line 102: Unnecessary newline that you introduced here.

nix/overlay.nix
```diff
@@ -95,6 +95,9 @@ in
       (inputs.nix-filter.lib.inDirectory ../tools)
       (inputs.nix-filter.lib.inDirectory ../plugins)
       (inputs.nix-filter.lib.inDirectory ../schema)
+      (inputs.nix-filter.lib.inDirectory ../scripts)
+      (inputs.nix-filter.lib.inDirectory ../systemd)
+      (inputs.nix-filter.lib.inDirectory ../debian)
```
dominiklohmann, line 100: Both the systemd and the debian directory moved.

.github/workflows/vast.yaml
```diff
@@ -299,7 +299,7 @@ jobs:
           cmake -B "$BUILD_DIR" \
             -DCMAKE_INSTALL_PREFIX:STRING="${PWD}/opt/vast" \
             -DCPACK_PACKAGE_FILE_NAME:STRING="$PACKAGE_NAME" \
-            -DCPACK_GENERATOR:STRING=TGZ \
+            -DCPACK_GENERATOR:STRING=TGZ;DEB \
```
dominiklohmann, line 302: The semicolon requires you to add quotes, otherwise bash does the wrong thing here.

vast/CMakeLists.txt
```diff
@@ -57,6 +57,16 @@ elseif (${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
...
+elseif (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+  # Not all Linux distros use systemd, but those we care about at the moment do.
+  # TODO: We don't consider CPACK_PACKAGE_INSTALL_PREFIX here, this could
+  # potentially be solved with a more involved CPack setup. See also the
+  # comment at the beginning of cmake/Package.cmake.
+  configure_file(
+    "${CMAKE_CURRENT_SOURCE_DIR}/services/systemd/vast.service.in"
+    "${CMAKE_CURRENT_BINARY_DIR}/vast.service" @ONLY)
+  install(FILES "${CMAKE_CURRENT_BINARY_DIR}/vast.service"
+          DESTINATION "${CMAKE_INSTALL_LIBDIR}/systemd/system/")
```
dominiklohmann, line 69: Now that we have this logic we should remove the `systemd` directory from the repository root and move the remaining contents over to `vast/services/systemd` as well.

This also requires changing one line in the Dockerfile in the repository root, which references `systemd/vast.yaml`.

dominiklohmann, line 69: This needs to be covered in the documentation under `web/docs/setup-vast/install/linux.md`, which now links to the wrong files and contains outdated information.

.github/workflows/vast.yaml
```diff
@@ -942,20 +949,34 @@ jobs:
...
+      - name: Upload Debian Release Package
+        if: github.event_name == 'release'
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ github.event.release.upload_url }}
+          asset_path: "${{ steps.build_static_package.outputs.store_path }}/${{ steps.get_artifact_names.outputs.deb }}"
+          # The asset name is constant so we can permanently link to
+          # https://github.com/tenzir/vast/releases/latest/download/vast-linux-static.deb
```
dominiklohmann, line 975: We should update the docs on `web/docs/setup-vast/download.md` to reference the Deb in addition to the tar.gz.

.github/workflows/vast.yaml
dominiklohmann, line 302: The additional Deb should also be uploaded for to GitHub as an artifact and for releases for the `build-debian`  job.

dominiklohmann (commented):
Some final minor remarks. This works really well in practice, so once this round is through I'd want to just have one additional docs review.

This also needs changelog entries still, I think:
- With the next release we'll start providing .deb packages.
- The newly bundled `vast-df-percent.sh` script needs to be explained.
- VAST now installs a systemd unit on Linux by default.

vast/CMakeLists.txt
tobim: Why are we copying the service file into the docker container?

tobim:
I didn't get around to updating the docs yet, and the CI copy changes are probably not correct yet either.

mavam:
Given the complexity of this PR and new developer tooling, we probably should add some details on how to build the package in https://vast.io/docs/develop-vast/contributing/build-environment.

vast/CMakeLists.txt
dominiklohmann: We are not, we're just re-using the same `vast.yaml` file that we reference by default in the service file.

lava:
Can we reduce the scope of this PR by not actually providing *public* debian packages yet but only producing them as an additional CI artifact that can be used in our downstream workflows? Having them available for users is a nice extra, but it feels like spending all the time on getting the documentation right is delaying the important part.

Speaking of artifacts, the current version only produces a single file called `artifact` which doesn't look right, although it already has the correct files:
```
$ unzip artifact.zip 
Archive:  artifact.zip
  inflating: vast-v2.3.0-rc2-29-ge0e580cf1a-Linux.tar.gz  
  inflating: vast_v2.3.0-rc2-29-ge0e580cf1a_amd64.deb  
```

dominiklohmann:
> Can we reduce the scope of this PR by not actually providing _public_ debian packages yet but only producing them as an additional CI artifact that can be used in our downstream workflows? Having them available for users is a nice extra, but it feels like spending all the time on getting the documentation right is delaying the important part.

So far we haven't spent any time on that, and I personally think that getting that right should be part of this PR so we don't ship something that's undocumented.

Whether we only attach to releases or only produce for downstream workflows is mostly irrelevant, as the former is already done and implemented.



> Speaking of artifacts, the current version only produces a single file called `artifact` which doesn't look right, although it already has the correct files:
> 
> ```
> $ unzip artifact.zip 
> Archive:  artifact.zip
>   inflating: vast-v2.3.0-rc2-29-ge0e580cf1a-Linux.tar.gz  
>   inflating: vast_v2.3.0-rc2-29-ge0e580cf1a_amd64.deb  
> ```

This requires calling the upload-artifact action twice instead of once with both files iirc, should be easy to find. The per-plugin CI will fail with this broken also, because it attempts to download the tar.gz artifact from the preceding runs.

vast/CMakeLists.txt
dominiklohmann: We're doing this in a follow-up PR.

.github/workflows/vast.yaml
dominiklohmann: We've decided to instead explicitly disable DEB + non-static builds.

dominiklohmann (approved):
Changelog entry still missing, other than that LGTM.

web/docs/setup-vast/install/linux.md
```diff
@@ -70,6 +70,18 @@ systemctl start vast
...
+### Debian
+
+We provide [pre-built packages](/docs/setup-vast/download#debian) for Debian
+and Debian derived distributions. After downloading, you can install VAST with
+
+```bash
+dpkg -i vast-${version}_amd64.deb
+```
+
+The debian package will automatically create a `vast` system user and install
```
dominiklohmann, line 82:
```suggestion
The Debian package automatically creates a `vast` system user and installs
```

web/docs/setup-vast/install/linux.md
```diff
@@ -70,6 +70,18 @@ systemctl start vast
 
 ## Distribution Support
 
+### Debian
+
+We provide [pre-built packages](/docs/setup-vast/download#debian) for Debian
+and Debian derived distributions. After downloading, you can install VAST with
```
dominiklohmann, line 76:
```suggestion
and Debian-derived distributions. After downloading, install VAST using `dpkg`:
```

changelog/unreleased/features/2513--debian-package.md
```diff
@@ -0,0 +1,7 @@
+VAST is now distributed as a Debian Package. The package for this release can be
+obtained from
+[https://storage.googleapis.com/tenzir-public-data/vast-static-builds/vast_v2.4.0_amd64.deb].
+This package automatically installs a systemd service and creates a `vast` user
+for the database process, greatly simplifying the setup process. Refer to [the
+docs](https://vast.io/docs/setup-vast/install/linux#debian) for detailed
+instructions.
```
dominiklohmann, line 7: Please rewrite using active voice, and link to the GitHub releases page rather than our public GCS bucket.

.github/workflows/vast.yaml
```diff
@@ -307,6 +307,7 @@ jobs:
           cmake -B "$BUILD_DIR" \
             -DCMAKE_INSTALL_PREFIX:STRING="${PWD}/opt/vast" \
             -DCPACK_PACKAGE_FILE_NAME:STRING="$PACKAGE_NAME" \
+            -DCPACK_PACKAGING_INSTALL_PREFIX:STRING="" \
```
dominiklohmann, line 310: This triggers a warning in CI, so I think we should add a comment on why we're adding this.

dominiklohmann (commented):
Some additional minor things I've come across when working on #2544.

PR #2515, opened 2022-08-19
Don't abort startup if individual partitions fail to load

tobim:
If something is wrong with a partition we now skip it instead of preventing VAST from starting up.

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -371,9 +371,10 @@ caf::error index_state::load_from_disk() {
         = std::filesystem::directory_iterator(markersdir, err);
       if (err)
         return caf::make_error(ec::filesystem_error,
-                               fmt::format("{} failed to list directory contents "
-                                           "of {}: {}", *self,
-                                           dir, err.message()));
+                               fmt::format("{} failed to list directory "
+                                           "contents "
+                                           "of {}: {}",
+                                           *self, dir, err.message()));
       for (auto const& entry : transforms_dir_iter) {
         if (entry.path().extension() != ".marker")
           continue;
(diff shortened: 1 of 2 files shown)


libvast/src/system/index.cpp
```diff
@@ -345,76 +345,86 @@ caf::error index_state::load_from_disk() {
...
-              [uuid](atom::done) {
-                VAST_DEBUG("index erased partition {} during startup", uuid);
-              },
-              [uuid](const caf::error& e) {
-                VAST_WARN("index failed to erase partition {} during startup: "
-                          "{}",
-                          uuid, e);
-              });
+                               fmt::format("failed to list directory contents "
+                                           "of {}: {}",
```
dominiklohmann, line 354:
```suggestion
                               fmt::format("{} failed to list directory contents "
                                           "of {}: {}", *self,
```

For consistency with the other log messages.

libvast/src/system/index.cpp
```diff
@@ -345,76 +345,86 @@ caf::error index_state::load_from_disk() {
...
+                                         static_cast<store_actor>(global_store),
+                                         filesystem, path);
+            self->request(partition, caf::infinite, atom::erase_v)
+              .then(
+                [uuid](atom::done) {
+                  VAST_DEBUG("index erased partition {} during startup", uuid);
+                },
+                [uuid](const caf::error& e) {
+                  VAST_WARN("index failed to erase partition {} during "
+                            "startup: "
```
dominiklohmann, line 399:
```suggestion
                  VAST_WARN("{} failed to erase partition {} during "
                            "startup: ", *self,
```

For consistency with the other log messages.

libvast/src/system/index.cpp
```diff
@@ -345,76 +345,86 @@ caf::error index_state::load_from_disk() {
...
-    std::filesystem::remove_all(markersdir);
+      // TODO: This does not handle store files, which may already have been
+      // written. Since a store file may also be written before the partition
+      // itself, there does not currently seem to be a bulletproof way of
+      // handling this.
+      std::filesystem::remove_all(markersdir);
+      return caf::none;
+    }();
+    if (error)
+      VAST_ERROR("{} failed to finish leftover transforms: {}", *self, error);
```
dominiklohmann, line 427: I think this should be a warning and not an error: After all, this will not stop VAST from functioning.

dominiklohmann (approved):
I tried this out by making select files unreadable to the user of the VAST process. This works as expected.

I added a `blocked` label so we can merge this only after the v2.3 release.

libvast/src/system/index.cpp
tobim: Done.

PR #2517, opened 2022-08-20
Switch to poetry for notebook scaffolding

mavam:
This PR switches to a poetry-based workflow for managing the build of example notebooks.

### :memo: Reviewer Checklist

- [ ] Executing `make` in `examples/notebooks/` works
- [ ] Executing `yarn && yarn start` in `web/` executes the notebook workflow


mavam:
@dispanser since you struggled with building the docs locally, does this branch work for you?

examples/notebooks/shell.nix
```diff
@@ -0,0 +1,13 @@
+# brings poetry into scope and makes sure that poetry-installed C-lib basedpython
```
dispanser, line 1: @tobim : maybe you want to have a brief look at this rudimentary `shell.nix`? I know it's not pretty, but at least allows me to run the docs locally.

examples/notebooks/shell.nix
```diff
@@ -0,0 +1,13 @@
...
+# python dependencies can "see" libstdc++. Ref https://nixos.wiki/wiki/Packaging/Quirks_and_Caveats#ImportError:_libstdc.2B.2B.so.6:_cannot_open_shared_object_file:_No_such_file
+{ pkgs ? (import <nixpkgs> {}).pkgs }:
+with pkgs;
+mkShell {
+  buildInputs = [
+    poetry
```
dispanser, line 7: we could add yarn here

dispanser (approved):
I tested it and got it to run with my NixOS-based setup after some fiddling.

I can now run the docs again locally, which is awesome❗

examples/notebooks/shell.nix
tobim: The essential thing is that it does the job. We can worry about "beauty" some other time.

tobim (approved):

examples/notebooks/shell.nix
mavam: That's technically in `web/` and not `examples/notebooks/`, no?

(Resolving, as we are agreeing to ship this.)

PR #2524, opened 2022-08-23
Use proper config key to disable partition indexes

dominiklohmann:
The integration tests and the example configuration file did not use the proper configuration key for disabling partition indexes, which we renamed from `create-dense-index` to `partition-index` late in the development process of that feature. This fixes that oversight.

### :memo: Reviewer Checklist

### DIFF
--- a/vast.yaml.example
+++ b/vast.yaml.example
@@ -172,10 +172,10 @@ vast:
     #   fp-rate - false positive rate. Has effect on string and address type
     #             targets
     #
-    #   create-dense-index - VAST will not create dense index when set to false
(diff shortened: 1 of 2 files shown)


dispanser (approved):
the change looks good. What I don't understand is why the test didn't fail, if it was using the wrong key. I assume it does not actually assert the absence of indexes. Something we should fix eventually (not here).

dominiklohmann:
> I assume it does not actually assert the absence of indexes. Something we should fix eventually (not here).

You're correct. The test only checks whether the output is still correct with partition indexes disabled.

PR #2526, opened 2022-08-24
Improve PID file handling after reboot

dominiklohmann:
After a reboot (e.g., when restarting a container), PIDs may be re-assigned to another process. If the PID in a PID file that lingered from a VAST crash is assigned to a non-VAST process we should ignore the PID file instead of preventing startup.

### :memo: Reviewer Checklist

- [x] Try reproducing on this branch again.

### DIFF
--- a/libvast/src/detail/pid_file.cpp
+++ b/libvast/src/detail/pid_file.cpp
@@ -27,6 +27,29 @@
 
 namespace vast::detail {
 
+#if VAST_LINUX
+
+namespace {
+
+bool pid_belongs_to_vast(pid_t pid) {
+  const auto proc_pid_path = fmt::format("/proc/%d/status", pid);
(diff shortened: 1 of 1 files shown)


rdettai (commented):
maybe I did something wrong when building your branch, but I still get `!! filesystem_error: PID file found: /var/lib/vast/pid.lock, terminate process 1`

Have you tried the steps below running VAST into 2 different containers sharing the same volume for `/var/lib/vast` ?

dominiklohmann:
@rdetta I forgot to push one last fix I already had locally. That was dumb, sorry about the churn. 

I tried your reproduction steps and managed to reproduce the error on master, and on this branch I now see an additional debug log message that we ignore the PID lock file. Can you try again (with `vast --console-verbosity=debug start` so you see the new log message)?

rdettai (approved):
Run my test again and a warning is now displayed instead of terminating

PR #2528, opened 2022-08-25
Improve the responsiveness and performance of the default store impl

dispanser:
Introduces more fine-grained control over processing of a single
partition by allowing to process at the finer granularity provided by
individual table slices. For this, we introduced a new actor interface
which has these additional continuation message types defined, as
we weren't able to expose the smaller store actor interface while 
internally using the wider one with additional messages.

It also removes unnecessary dupliacted computation of the VAST schema
from the Arrow schema, which is now only computed once per partition.
This helps both parquet and feather stores, and a few other callers can now 
utilize that new constructor to inject an already-known VAST schema as well.

### Instructions

1. check `store.hpp` / `store.cpp` / `actor.hpp` for the newly introduced messages  for the handling of "continuations"
2. See `feather.cpp` / `parquet.cpp` for the application of the new `table_slice` constructor which accepts a VAST schema
3. remaining files all either implement the new `table_slice` constructor (lots of places forwarding the arg), or use it with an already existing schema and avoid some recomputations.

### :memo: Reviewer Checklist


libvast/include/vast/arrow_table_slice.hpp
```diff
@@ -71,8 +71,10 @@ class arrow_table_slice final {
   /// @param slice The encoding-specific FlatBuffers table.
   /// @param parent The surrounding chunk.
   /// @param batch A pre-existing record batch.
+  /// @param schema VAST schema matching the record batch schema.
```
mavam, line 74: Is the idea that the new parameter coerces the provided batch into a desired schema?

dispanser: No, this is purely an optimization - we often have many record batches from the same schema (single feather / parquet / segment store), and don't want to recompute the VAST schema in the constructor every time.

dominiklohmann, line 74: The documentation should explain that this argument is optional and exists purely for optimization purposes.

libvast/include/vast/arrow_table_slice_builder.hpp
```diff
@@ -44,7 +44,7 @@ class arrow_table_slice_builder final : public table_slice_builder {
 
   /// @pre `record_batch->schema()->Equals(make_experimental_schema(layout))``
   [[nodiscard]] table_slice static create(
-    const std::shared_ptr<arrow::RecordBatch>& record_batch,
+    const std::shared_ptr<arrow::RecordBatch>& record_batch, type schema = {},
```
dominiklohmann, line 47: Argument is missing from the docs.

libvast/include/vast/msgpack_table_slice.hpp
```diff
@@ -46,8 +46,10 @@ class msgpack_table_slice final {
   /// @param slice The encoding-specific FlatBuffers table.
   /// @param parent The surrounding chunk.
   /// @param batch A pre-existing record batch. Must always be nullptr.
+  /// @param schema A VAST schema matching the record batch. Must always be none.
```
dominiklohmann, line 49: Same deal as above, explain that this is optional.

libvast/include/vast/store.hpp
```diff
@@ -36,20 +33,36 @@ class passive_store {
...
-  lookup(system::store_actor::pointer self,
-         const query_context& query_context) const;
+  /// Retrieve the schema associated with the data in the store.
+  /// @returns The VAST schema of the stored data.
+  [[nodiscard]] virtual type schema() const;
+
+  /// Execute a count query against the store.
+  /// @param expr The expression to filter events.
+  /// @param selection Pre-filtered ids to consider.
+  /// @return
```
dominiklohmann, line 43: The @returns here is kinda important to fill out, as we expect users to implement this.

We should also mention that this has a default implementation.

libvast/include/vast/store.hpp
```diff
@@ -36,20 +33,36 @@ class passive_store {
...
+  [[nodiscard]] virtual detail::generator<uint64_t>
+  count(const expression& expr, const ids& selection) const;
+
+  /// Execute an extract query against the store.
+  /// @param expr The expression to filter events.
+  /// @param selection Pre-filtered ids to consider.
+  [[nodiscard]] virtual detail::generator<table_slice>
```
dominiklohmann, line 50: Same as for count(...), explain the desired return value for implementations.

libvast/include/vast/type.hpp
```diff
@@ -128,6 +128,13 @@ class type final : public stateful_type_base {
     no,  ///< Include internal types. Use with caution.
   };
 
+  /// indicates whether we want to recursively process the type's children
```
dominiklohmann, line 131:
```suggestion
  /// Indicates whether we want to recursively process the type's aliases
```

libvast/native-plugins/feather.cpp
```diff
@@ -83,31 +83,50 @@ class passive_feather_store final : public passive_store {
...
-      slice.offset(num_events_);
-      num_events_ += slice.rows();
-      slice.import_time(derive_import_time(time_col));
-      slices_.push_back(std::move(slice));
+
+    table_ = std::move(table);
+    auto arrow_field = table_->schema()->GetFieldByName("event");
+    if (!arrow_field) {
+      return caf::make_error(ec::format_error, "schema does not have mandatory "
+                                               "`event` column");
```
dominiklohmann, line 91:
```suggestion
                                               "'event' column");
```

libvast/native-plugins/feather.cpp
```diff
@@ -83,31 +83,50 @@ class passive_feather_store final : public passive_store {
...
   [[nodiscard]] detail::generator<table_slice> slices() const override {
-    for (const auto& slice : slices_)
-      co_yield slice;
+    auto offset = id{};
+    for (auto rb : arrow::TableBatchReader(*table_)) {
+      VAST_ASSERT(rb.ok(), rb.status().ToString().c_str());
+      auto time_col = rb.ValueUnsafe()->GetColumnByName("import_time");
+      auto unwrapped_rb = unwrap_record_batch(rb.MoveValueUnsafe());
+      auto slice = table_slice{unwrapped_rb, schema_};
+      // auto slice = table_slice{unwrapped_rb, schema};
```
dominiklohmann, line 109:
```suggestion
```

dominiklohmann (commented):
Some initial notes from glancing over what we did yesterday on this.

libvast/src/arrow_table_slice_builder.cpp
```diff
@@ -143,16 +144,19 @@ table_slice arrow_table_slice_builder::finish() {
...
-  return create_table_slice(record_batch, this->builder_,
+  return create_table_slice(record_batch, this->builder_, layout(),
                             table_slice::serialize::yes);
 }
 
 table_slice arrow_table_slice_builder::create(
-  const std::shared_ptr<arrow::RecordBatch>& record_batch,
+  const std::shared_ptr<arrow::RecordBatch>& record_batch, type schema,
   table_slice::serialize serialize, size_t initial_buffer_size) {
+#if VAST_ENABLE_ASSERTIONS
```
patszt, line 154: Can we just do VAST_ASSERT(verify_record_batch(*record_batch)) ?

libvast/src/msgpack_table_slice.cpp
```diff
@@ -253,10 +253,11 @@ msgpack_map_view::value_type msgpack_map_view::at(size_type i) const {
 template <class FlatBuffer>
 msgpack_table_slice<FlatBuffer>::msgpack_table_slice(
   const FlatBuffer& slice, [[maybe_unused]] const chunk_ptr& parent,
-  const std::shared_ptr<arrow::RecordBatch>& batch) noexcept
+  const std::shared_ptr<arrow::RecordBatch>& batch, type schema) noexcept
```
patszt, line 256: Is this param needed here? It seems unused, but i assume this is required for some generic code that passes all these parameters into different functions

libvast/src/type.cpp
```diff
@@ -459,6 +459,11 @@ type::type(std::string_view name, const type& nested,
         attributes_offsets.emplace_back(fbs::type::detail::CreateAttribute(
           builder, key_offset, value_offset));
       };
+      std::sort(attributes.begin(), attributes.end(),
```
patszt, line 462: I have heard that this file is very optimized. I am not sure what implications it has. Can someone more experienced confirm that this is ok? :D

libvast/test/type.cpp
```diff
@@ -714,8 +714,7 @@ TEST(enriched types) {
   CHECK_EQUAL(aat.attribute("second"), "");
   CHECK_EQUAL(aat.attribute("third"), "nestingworks");
   CHECK_EQUAL(aat.attribute("fourth"), std::nullopt);
-  CHECK_EQUAL(fmt::format("{}", aat), "l2 #third=nestingworks #first=value "
-                                      "#second");
+  CHECK_EQUAL(fmt::format("{}", aat), "l2 #third=nestingworks");
```
patszt, line 717: I would expect the nested at part to be displayed as well. But this is just my naive intuition. Can you elaborate a little bit why this is the expected behaviour?

libvast/src/msgpack_table_slice.cpp
dominiklohmann: We no longer generate anything but Arrow v2 table slices, so this is essentially dead code. The parameter was just added to conform to a shared interface for construction as you assumed.

libvast/src/type.cpp
dominiklohmann: Yes, this is actually a required bug fix. The size of this vector is 0 or 1 most of the time, so it's not critical.

libvast/test/type.cpp
dominiklohmann: It was only just changed in this PR to align how we print and parse types. The nested attributes are part of the nested type already, so printing them again just sounds confusing. What we could do is add an option for the formatter to print the whole type structure instead of just the first layer.

libvast/src/store.cpp
```diff
@@ -177,30 +314,30 @@ system::store_builder_actor::behavior_type default_active_store(
   self->state.store = std::move(store);
   self->state.path = std::move(path);
   self->state.store_type = std::move(store_type);
+  self->set_down_handler([self](const caf::down_msg& down_msg) {
+    for (const auto& [query_id, state] : self->state.running_extractions) {
```
patszt, line 318: Seems to be duplicated twice. How about a free function in anon namespace.

[self](const caf::down_msg& down_msg)
{
    handle_down_msg(self, down_msg);
}

patszt (commented):
Looks nice overall to my inexperienced eye :)

patszt (approved):

libvast/test/type.cpp
patszt: Alright, if we find it useful to have whole thing printed then we can implement it

libvast/include/vast/msgpack_table_slice.hpp
dispanser: According to the method signature, it's not optional. I added that it _must be none_, as for the batch.

libvast/include/vast/arrow_table_slice.hpp
dispanser: It's currently not optional, should I change that?

libvast/src/arrow_table_slice_builder.cpp
dispanser: yeah, nice idea. had to change the return type of `verify_record_batch` to `bool`, but that's fine I guess.

libvast/src/store.cpp
dispanser: Nice idea, thanks for that! Applied.

libvast/include/vast/arrow_table_slice.hpp
dominiklohmann: The parameter may not be optional, but the argument semantically is, as you can pass in a default-constrcuted type and the constructor infers it from the batch for you.

libvast/native-plugins/segment_store.cpp
```diff
@@ -212,6 +213,12 @@ system::store_actor::behavior_type passive_local_store(
           rp.delegate(static_cast<system::store_actor>(self), atom::erase_v,
                       std::move(ids));
         }
+        auto startup_duration = std::chrono::steady_clock::now() - start;
+        self->send(self->state.accountant, "passive-store.startup.runtime",
```
dominiklohmann, line 217: Can you add this to monitor.md in the docs?

IIRC we were also unsure about the naming, probably something for @mavam.

mavam: I think I'm too far away to have a useful opinion on this. 😉

dominiklohmann: It's a metric that measures the duration from spawning a store to it being able to answer the first query. You're usually good at naming things, that's why I thought of you, and Thomas and I were pairing on this and neither of us came up with a good name, so `startup.runtime` is what we've got now. 😅

dispanser: good catch.

mavam: To be frank, `startup.runtime` sounds like an oxymoron. If the semantics are "time until things are ready", then here are some suggestions:

- `startup.time`
- `startup.latency`
- `startup.duration`
- `startup.boot-time`
- `startup.load-time`
- `startup.warmup`

HTH

dispanser: What about `load.duration`? I find `startup` to be somewhat imprecise as it's not entirely clear what "startup" for a store needs.

mavam: @dispanser and I agreed in a huddle that `init.runtime` is conveying the semantics best.

dispanser: we settled for `init.runtime` as we're initializing the store, and that word clearly conveys that this is a one-time operation (per store).

PR #2529, opened 2022-08-25
Don't create chunks for nonexisting indices

lava:
- Don't create chunks for nonexisting indexes
- Cleanup flatbuffer container leftovers

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/src/system/partition_transformer.cpp
+++ b/libvast/src/system/partition_transformer.cpp
@@ -375,10 +375,16 @@ partition_transformer_actor::behavior_type partition_transformer(
         mutable_synopsis.shrink();
         mutable_synopsis.offset = 0;
         mutable_synopsis.events = data.events;
-        // Create the value indices.
         for (auto& [qf, idx] :
-             self->state.partition_buildup.at(data.id).indexers)
-          data.indexer_chunks.emplace_back(qf.name(), chunkify(idx));
+             self->state.partition_buildup.at(data.id).indexers) {
+          // Special case to preserve null values, since `chunkify(nullptr)`
+          // returns an empty chunk. We defensively treat every value index
+          // of size 0 as nonexistent, even if the input was not null.
+          auto chunk = chunkify(idx);
+          if (chunk->size() == 0)
+            chunk = nullptr;
+          data.indexer_chunks.emplace_back(qf.name(), chunk);
+        }
       }
       detail::shutdown_stream_stage(self->state.stage);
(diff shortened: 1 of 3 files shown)


libvast/src/system/partition_transformer.cpp
```diff
@@ -378,7 +378,8 @@ partition_transformer_actor::behavior_type partition_transformer(
         // Create the value indices.
         for (auto& [qf, idx] :
              self->state.partition_buildup.at(data.id).indexers)
-          data.indexer_chunks.emplace_back(qf.name(), chunkify(idx));
+          data.indexer_chunks.emplace_back(qf.name(),
+                                           idx ? chunkify(idx) : nullptr);
       }
```
dominiklohmann, line 383: If we encounter an indexer chunk that is empty, should we also add special handling to just treat it as if it were a nullptr just in case users ran into the bug?

As-is, this would only fix the issue for newly rebuilt partitions. Were we already able to run into this with v2.2?

dominiklohmann (approved):
This is a great find, thanks for this. I double-checked with a new partition using `lsvast`, and the disabled indexers are indeed zero-sized now after rebuilding, while they weren't before.

As a test for this, should we maybe run `lsvast` on a single-database partition before and after rebuilding and compare the output? We could extend the `lsvast` integration test suite for this.

libvast/src/system/partition_transformer.cpp
lava: It only became possible to trigger the bug with #2509 so v2.2 should be fine, but I think it makes sense anyways as a defensive coding measure.

dominiklohmann: Okay, but then at least we don't need a changelog entry.

dominiklohmann: NB: You pushed an empty changelog file that should probably be removed again. Other than that this is ready to go imo.

dominiklohmann: @lava I unresolved to get you to see this comment

> NB: You pushed an empty changelog file that should probably be removed again. Other than that this is ready to go imo.

lava: I'll resolve again to avoid re-running CI, but I'll fix it in the rc3 PR.

dominiklohmann: Okay that's fine by me, just wanted to make sure you noticed.

PR #2538, opened 2022-08-29
Update incorrect skipping of non-existing value indexes

lava:
- Better lsvast output for segmented partitions
- Add formatter for chunk_ptr
- Add unit test
- Fix inverted threshold settings
- Don't chunkify null indices at all

### :memo: Reviewer Checklist


libvast/src/system/partition_transformer.cpp
```diff
@@ -377,11 +377,12 @@ partition_transformer_actor::behavior_type partition_transformer(
         mutable_synopsis.events = data.events;
         for (auto& [qf, idx] :
              self->state.partition_buildup.at(data.id).indexers) {
-          // Special case to preserve null values, since `chunkify(nullptr)`
-          // returns an empty chunk. We defensively treat every value index
-          // of size 0 as nonexistent, even if the input was not null.
-          auto chunk = chunkify(idx);
-          if (chunk->size() == 0)
+          auto chunk = chunk_ptr{};
+          // Note that `chunkify(nullptr)` return a chunk of size > 0.
```
tobim, line 381: Is that not a bug in `chunkify()`?

lava: Arguably yes, but that goes into the caf serialization code and would need more follow-up investigation to determine if and how we can fix it.

tobim: Is there any reason not to add
```
  if (!idx)
    return nullptr;
```
to the beginning of chunkify?
I fear that the solution here will potentially be forgotten at other call sites.

dominiklohmann: > Arguably yes, but that goes into the caf serialization code and would need more follow-up investigation to determine if and how we can fix it.

The CAF serialization handles nullptr and empty chunk roundtrips correctly. We even have unit tests for that:

https://github.com/tenzir/vast/blob/c0cd3d0d1fe4a2418ed01f8ae701c6ea7a4a0dcd/libvast/test/chunk.cpp#L86-L103

tobim: After thinking about this some more I'd like to retract my earlier suggestion :).
Changing `chunkify` in such a way would break the roundtrip of the `value_index_ptr`.

I now believe that the solution in this commit is actually not wrong, but lines 385 and 386 are questionable, since chunk size is a poor proxy for data validity.

libvast/src/value_index.cpp
```diff
@@ -248,6 +248,8 @@ bool inspect(detail::legacy_deserializer& source, value_index_ptr& x) {
 }
 
 vast::chunk_ptr chunkify(const value_index_ptr& idx) {
+  if (!idx)
+    return nullptr;
```
tobim, line 252: Doing this makes it impossible to serialize `value_index_ptr`s that are `nullptr`. I think I was wrong about this, sorry.

lava: Taken out of the review again.

libvast/src/system/partition_transformer.cpp
lava: The idea of lines 385/386 is to guard against potential buggy implementations of `chunkify()` that return empty chunks instead of nullptr. Since an empty chunk cannot possibly contain any useful indexing information, this transformation should be valid. Of course it's not perfect since we could also have a non-empty invalid chunk (as we saw here), but I don't think there's any way to reliably detect that.

tobim: > ... cannot possibly contain any useful indexing information ...

Why not? Don't you make assumptions about he encoding mechanism with that?

lava: > Don't you make assumptions about he encoding mechanism with that?

Keep in mind we're talking about a 0-byte sized index, it cannot hold information no matter the encoding.

tobim (approved):
The changes make sense to me, and I'm confident the new test will prevent the same problem to lip though in the future.

PR #2548, opened 2022-08-31
Cloud MISP

rdettai:
This PR adds a MISP plugin to the cloud deployment
- after intensive experimentation with  Ngrok and Cloudflare tunnel, these solutions where discarded:
   - too much per-requisits for the user (have an account, for cloudflare have a domain, for ngrok activate the paying tier)
   - hard to secure, especially TCP tunnels
- we went for a more traditional ssh tunnel with a keypair stored in the terraform state for now. This makes it harder to share the demo

### :memo: Reviewer Checklist

Also note that:
- You should rebuild the CLI (`VASTCLOUD_REBUILD=1 ./vast-cloud`)
- Changes to core were necessary to allow using EFS from the plugin
- We added `sudo` capabilities to the user in the CLI image to enable binding low ports
- In `cloud/aws/resources/configs/misp/` configs where copied over from our internal repo. These should probably be shared with the docker-compose configs once they are ported to this repository


rdettai:
I disabled the mount of persistent storage on the path where MISP is storing its file because I still had some write issues. Files attached to events will be lost accross MISP restarts, but apart from that it should work

cloud/aws/terraform/misp/definition.tf
```diff
@@ -140,20 +142,40 @@ locals {
...
+      name      = "ssh-server"
+      essential = true
+      portMappings = [{
+        containerPort = 2222
+        hostPort      = 2222
+        protocol      = "tcp"
+      }]
+      environment = [{
+        name  = "PUBLIC_KEY"
+        value = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAINBtjQQ3SxIzW5IhviByGdr3Y0x5vroJRf6Rw9DM7jD7"
```
tobim: Would it be possible to create the key-pair ad-hoc before the instance is created?

rdettai: Is it possible that the version of the code that you commented isn't the latest one? Because the current behavior is that a keypair is created when the stack is deployed.

rdettai:
[Terraform](https://github.com/tenzir/vast/actions/workflows/terraform.yaml)

web/docs/setup-vast/deploy/aws.md
```diff
@@ -225,3 +225,18 @@ datasource resources only, use:
...
 :::
+
+### Run MISP
+
+[MISP](https://www.misp-project.org/) is a popular community driven open source
+thread intelligence platform. It integrates naturally with VAST. To deployed a
+preconfigured instance:
+- add `misp` to the list of plugins in `VAST_CLOUD_PLUGINS`
+- deploy (or re-deploy) the VAST stack including MISP with `./vast-cloud deploy`
+- start the MISP server with `./vast-cloud misp.start`, wait a few minutes, MISP
```
KaanSK, line 236: Dot notation in CLI was a bit unexpected. Would it be better to have something like `--misp-start`, `--misp-stop` ? Also in the future adding new commands would cause overload of commands that could be confusing. Maybe we can align on a template for commands for services and have only a template for them in documentation. Ex:

```
For services in [misp, ex1, ex2] following commands can be used:
{service}.start
{service}.stop
{service}.status
...
```

I am aware that CLI libraries do not allow many customizations but at least would like to discuss about a template.

KaanSK:
After deployment, it is really unclear what is going on in the environment. Would it be possible to give a summary text on
ID/IP of MISP and Vast Server with available ports? It could be my lack of experience with these a bit but I needed to manually check stuff on both AWS and terraform code.

Additionally, what could be the reason for the output below? I expected the output to align with the test case. The screenshot is after the successful deployment

image

KaanSK (changes requested):

cloud/aws/cli/plugins/misp.py
```diff
@@ -0,0 +1,86 @@
...
+    eni_id = next(
+        x["value"] for x in eni_attach["details"] if x["name"] == "networkInterfaceId"
+    )
+
+    eni_desc = aws("ec2").describe_network_interfaces(NetworkInterfaceIds=[eni_id])
+    return eni_desc["NetworkInterfaces"][0]["Association"]["PublicIp"]
+
+
+@task
+def tunnel(c, ui_port="8080", zmq_port="50000"):
```
KaanSK, line 62: Tunnel seems to be not successful after deploy. No error messages.

image

cloud/aws/cli/plugins/misp.py
```diff
@@ -0,0 +1,86 @@
+import os
```
KaanSK, line 1: Would it make sense to include a logging mechanism to console and/or a file for implicit errors? In the issue with `misp.tunnel` I unfortunately do not directly get an error message.

KaanSK (changes requested):

web/docs/setup-vast/deploy/aws.md
rdettai: I am also not fond of this dot notation. To me whitespace would be the most logical. But I couldn't find an easy way to achieve this with the PyInvoke library yet.

> Also in the future adding new commands would cause overload of commands that could be confusing

not sure what you mean

> Maybe we can align on a template for commands for services and have only a template for them in documentation.

Interesting point. Right now the VAST server service does not comply to the `{service}.xxxx` pattern, but I think that it would be better if it did :-)

cloud/aws/cli/plugins/misp.py
rdettai: I am wondering why no error message is bubbling up here. The issue is that you didn't start MISP

cloud/aws/cli/plugins/misp.py
rdettai: what to do mean by "logging mechanism"? Usually a CLI just prints everything that can be useful to the user to the console. Are you looking for something like a `--verbose` option? That could be an interesting addition. 

There is a way to display stack traces that is not documented in the high level docs as it is intended for development of the CLI itself, you have to set VASTCLOUD_TRACE=1 in the environment.

rdettai:
> After deployment, it is really unclear what is going on in the environment. Would it be possible to give a summary text on
ID/IP of MISP and Vast Server with available ports? It could be my lack of experience with these a bit but I needed to manually check stuff on both AWS and terraform code.

Not sure how an ID/IP would be useful as the services are not exposed to the outside world. The ports are displayed when you start the tunneling, but maybe clearer listing would be nice. 

> Additionally, what could be the reason for the output below? I expected the output to align with the test case. The screenshot is after the successful deployment. Instead of "No Task", a bit more verbose output could be shown.

Did you start the service with `misp.start` as pointed out in the docs? I agree the message should be made more explicit.

web/docs/setup-vast/deploy/aws.md
rdettai: created a [story](https://app.shortcut.com/tenzir/story/37554/make-all-cloud-service-commands-obey-the-same-pattern) to track it to avoid snowballing in this PR

Regarding the dot notation, I don't have a good alternative solution and I think it's good enough for now.

cloud/aws/cli/plugins/misp.py
rdettai: I made the message a bit more explicit (`No task running`)

rdettai:
I made service status messages more explicit and fixed the integration test. Could you take a new look at it @KaanSK ?

cloud/aws/cli/plugins/misp.py
KaanSK: I just checked the behavior and satisfied with overall verbosity on status/errors.
image

Putting this to resolved.

cloud/aws/resources/configs/misp/misp-init.py
```diff
@@ -0,0 +1,93 @@
...
+
+# When MISP runs for the first time, an admin user doesn't exist yet. Browsing
+# to the frontend triggers creation of an admin user lazily. But want eager
+# behavior on startup, so we must trigger 'cake UserInit' to perform the
+# intilization. The output includes an ephemeral API key that we can use
+# subsequently to connect via PyMISP.
+result = subprocess.run([CAKE, "UserInit"], capture_output=True)
+api_key = result.stdout.decode("utf-8").splitlines()[-1]
+if "MISP instance already initialised" in api_key:
+    print("skipping initialization of already existing admin user")
```
KaanSK, line 40: This comment concerns all `print` statements.

Wouldn't it be better to integrate a logging library with proper configuration that defaults to logging to console? Proper logging has following advantages:

1. Ability to provide where (which line) code is broken
2. Log can be redirected to console, file, socket easily
3. Based on severity, logs can be suppressed
4. Maybe the most important thing, format can be changed in a central manner. (encapsulating log message in JSON for example)

KaanSK (changes requested):

KaanSK:
I followed the instructions and I was able to get VAST Server, MISP instance up and ready. Approach seems more robust and status/error messages seems more verbose.

image

I went through the code both infra and logic related, to the extend of my knowledge. Put comment(s) on discussion points.

cloud/aws/resources/configs/misp/misp-init.py
rdettai: I agree with you for larger systems, but this is just a tiny independent startup script that works around the fact that these settings are not accessible through the MISP config files directly.

PR #2549, opened 2022-08-31
Add a health check to the tenzir/vast Docker image

dominiklohmann:
The health check is primarily useful container orchestration, where it allows for detecting (1) when a VAST server can first be connected to and (2) when a VAST server is unresponsive.

For example, in a Docker Compose file services running `vast import` can depend on a `vast start` service being healthy instead of just running to avoid failing to connect because the service started, but did not listen for incoming connections yet.

Outstanding tasks:
- [ ] Test whether this fixes the described race condition in our test setup
- [ ] Add a changelog entry

@lava I've requested you for review since we talked about this in Slack earlier today; let's see if this fixes the flakiness we've seen.

### :memo: Reviewer Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -154,5 +154,14 @@ EXPOSE 42000/tcp
 WORKDIR /var/lib/vast
 VOLUME ["/var/lib/vast"]
 
+# Every 15 seconds, run the least expensive remote request there is: the status
+# request limited to the getting the remote version. We only start this after 2
+# seconds to give the VAST server some time to start. The health check is
+# primarily useful container orchestration, where it allows for detecting (1)
+# when a VAST server can first be connected to and (2) when a VAST server is
+# unresponsive.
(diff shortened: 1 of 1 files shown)


lava:
Is there a good way to test this outside of the internal CI, or is the best approach here to just verify this doesn't break the Dockerfile and then to just merge it and observe what happens?

dominiklohmann:
> Is there a good way to test this outside of the internal CI, or is the best approach here to just verify this doesn't break the Dockerfile and then to just merge it and observe what happens?

Build the image locally:

```
docker build . -t tenzir/vast
```

Then run and inspect the health:

```
docker run --rm --name vast -p 42000:42000 tenzir/vast start
docker inspect --format='{{json .State.Health}}' vast
```

dominiklohmann:
This is made obsolete by #2574, which adds a more natural check for this place.

PR #2552, opened 2022-09-01
Replace CAF 18.6 deprecated methods in favour of supported free function

patszt:
CAF 18.6 removed the make_sink, make_source and make_continuous_source methods.
The free functions i have replaced them with are supported both in CAF 17.6 and 18.6
### :memo: Reviewer Checklist

### DIFF
--- a/libvast/native-plugins/segment_store.cpp
+++ b/libvast/native-plugins/segment_store.cpp
@@ -191,7 +191,11 @@ system::store_actor::behavior_type passive_local_store(
         if (!seg) {
           VAST_ERROR("{} couldn't create segment from chunk: {}", *self,
                      seg.error());
-          self->send_exit(self, caf::exit_reason::unknown);
+          self->send_exit(self,
+                          caf::make_error(vast::ec::format_error,
+                                          fmt::format("{} failed to create "
+                                                      "segment from chunk: {}",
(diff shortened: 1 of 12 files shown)


libvast/native-plugins/segment_store.cpp
```diff
@@ -190,7 +191,7 @@ system::store_actor::behavior_type passive_local_store(
         if (!seg) {
           VAST_ERROR("{} couldn't create segment from chunk: {}", *self,
                      seg.error());
-          self->send_exit(self, caf::exit_reason::unhandled_exception);
```
patszt, line 193: in CAF 18.6 this enum value is removed

dominiklohmann: Note that this doesn't have to be a `caf::exit_reason`, you can actually provide any error code here. I think it'd make a lot of sense to just use `caf::make_error(vast::ec::something, fmt::format("{} failed to create segment from chunk: {}", *self, seg.error()))` here.

patszt: @dominiklohmann 
Good tip, i will change it as proposed

dominiklohmann (approved):
This is a no-brainer refactoring, thanks!

PR #2555, opened 2022-09-02
Fix artifact uploads to GCS

tobim:
### :memo: Reviewer Checklist

- [x] check that gsutil allows specifying a full path to the destination

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -379,7 +379,7 @@ jobs:
       - name: Upload Artifact to GCS
         if: ${{ github.event_name == 'push' || github.event_name == 'release' }}
         run: |
-          gsutil -m cp "${{ env.BUILD_DIR }}/package/${{ env.PACKAGE_NAME }}.tar.gz" "gs://${{ secrets.GCS_BUCKET }}/$PACKAGE_NAME.tar.gz"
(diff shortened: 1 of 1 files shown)


.github/workflows/vast.yaml
```diff
@@ -379,7 +379,7 @@ jobs:
       - name: Upload Artifact to GCS
         if: ${{ github.event_name == 'push' || github.event_name == 'release' }}
         run: |
-          gsutil -m cp "$BUILD_DIR/$PACKAGE_NAME.tar.gz" "gs://${{ secrets.GCS_BUCKET }}"
+          gsutil -m cp "${{ env.BUILD_DIR }}/package/${{ env.PACKAGE_NAME }}.tar.gz" "gs://${{ secrets.GCS_BUCKET }}/$PACKAGE_NAME.tar.gz"
```
tobim, line 382: I'm not sure whether we have any downstream jobs that depend on that name, so maybe we can even keep the filename from cpack?

dominiklohmann: The proprietary plugin CI depends on it, but you can easily adjust that to use another name.

dominiklohmann, line 382:
```suggestion
          gsutil -m cp "${{ env.BUILD_DIR }}/package/${{ env.PACKAGE_NAME }}.tar.gz" "gs://${{ secrets.GCS_BUCKET }}/${{ env.PACKAGE_NAME }}.tar.gz"
```

Let's just make this consistent.

dominiklohmann (approved):

.github/workflows/vast.yaml
tobim: Let's do that some other time.

PR #2557, opened 2022-09-02
Adjust code to CAF 18.6 config/settings API restrictions

patszt:
CAF 18.6 does compile time type checks for get_if, get, holds_alternative. It requires the passed type to be part of caf::config_value::types (none_t, integer, boolean, real, timespan, uri,
                                  string, list, dictionary)

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/src/defaults.cpp
+++ b/libvast/src/defaults.cpp
@@ -18,7 +18,8 @@
 namespace vast::defaults::import {
 
 size_t test::seed(const caf::settings& options) {
-  if (auto val = caf::get_if<size_t>(&options, "vast.import.test.seed"))
+  constexpr auto key = std::string_view{"vast.import.test.seed"};
+  if (auto val = caf::get_if<caf::config_value::integer>(&options, key))
     return *val;
   std::random_device rd;
   return rd();
--- a/vast/vast.cpp
+++ b/vast/vast.cpp
@@ -155,7 +155,9 @@ int main(int argc, char** argv) {
   std::string_view max_threads_key = CAF_VERSION < 1800
                                        ? "scheduler.max-threads"
                                        : "caf.scheduler.max-threads";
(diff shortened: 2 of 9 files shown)


vast/vast.cpp
```diff
@@ -220,9 +222,9 @@ int main(int argc, char** argv) {
   if (auto result = run(*invocation, sys, root_factory); !result)
     run_error = std::move(result.error());
   else
-    result->apply({[&](caf::error& err) {
```
patszt, line 223: apply gets removed in CAF 18.6

It constructed caf::message_handler from passed callable and called it with *this. So this is the exact same behaviour as previously

libvast/src/defaults.cpp
```diff
@@ -18,7 +18,9 @@
 namespace vast::defaults::import {
 
 size_t test::seed(const caf::settings& options) {
-  if (auto val = caf::get_if<size_t>(&options, "vast.import.test.seed"))
+  if (auto val = caf::get_if<caf::config_value::integer>(&options, "vast."
+                                                                   "import."
+                                                                   "test.seed"))
```
dominiklohmann, line 23: If you add a `//` after `&options,`, you force clang-format to break the line and not break the string. This is quite handy for config options, which we never want to break so this continues to shop up for `git grep vast.import.test.seed`.

The same applies to other instances in the same PR, sadly we have to force clang-format to do the right thing here. :|

dominiklohmann (approved):
If CI doesn't complain about it I'm also happy with it. Just one minor suggestion inline.

vast/vast.cpp
dominiklohmann: That makes sense. I believe there was also a typed message view that you can create from the message, but this is probably safer since it doesn't assert that the typed message view's signature actually matches.

PR #2558, opened 2022-09-04
Blog: A Git Retrospective

mavam:
This blog post is a Quarto document that demonstrates how we can mix and match Python, R, and shell in a single Markdown notebook. 

The `.qmd` file is the complete input to the generated GFM output ending in `.md`.

I've currently added the generated outputs as well, because generating deterministic output from R is not as easy. (While there exists support for renv, I've only played a little bit with it and have a lot of open questions, e.g., how to add the generated `renv` to a git repo?) This almost duplicates the `.qmd` file, which is suboptimal.

Having written now an entire blog post once, I feel that this format is superior to the jupytext Markdown we are using for our Try VAST tutorials. We will migrate them to Quarto.

I had to write a Lua extension and add it to the Quarto project base (= /web) to get code-folding support for `<details><summary>...` blocks.

~Open issue: there is currently no support for dark mode 😱! Given the convenience of just writing a notebook and being done, this may be acceptable. Since we can generate SVGs from R, we could also write a Quarto Pandoc filter in Lua that patches the SVG and injects `@media` CSS.~ EDIT: dark mode is now there: https://github.com/tenzir/vast/pull/2558/commits/9f99fbecbd01f382164c4164ffe8c2c60fd64449

### :memo: Reviewer Checklist

- [x] The blog post builds locally and looks nice.
- [x] We discussed if we want to adopt Quarto as default notebook format.
- [x] We discussed if we should include the generated output in the repo.


mavam:
I'm not making much progress with CI and asked for help here: https://github.com/quarto-dev/quarto-actions/issues/43

mavam:
@dominiklohmann could you take a look at this? I still struggle with the optimal way to do the local package installation, but in CI the GH Actions take care of that. The current setup assumes that you have one R package installed: `pak`.

Then this should work in `/web`:

```
make quarto
```

I haven't figured out yet how to run this conditionally, only when the generated Quarto inputs have changed. This requires a bit more Make futzing.

dominiklohmann:
> I haven't figured out yet how to run this conditionally, only when the generated Quarto inputs have changed. This requires a bit more Make futzing.

I pushed a good approximation. Won't catch everything, but it's somewhat nice at least.

.github/workflows/vast.yaml
```diff
@@ -164,10 +164,29 @@ jobs:
...
+      # build, so must clean up prior to doing so. The Right Fix is a one-line
+      # PR against r-lib/actions to remove the .github dir at action exit.
+      - name: Clean up after setup-r-dependencies
+        run: rm -rf web/.github
+      - name: Setup Quarto
+        uses: quarto-dev/quarto-actions/setup@v2
+      - name: Render Quarto Project
+        uses: quarto-dev/quarto-actions/render@v2
+        with:
+          path: web
```
mavam, line 185: Let's replace this with `make quarto`

dominiklohmann, line 185: This shouldn't be needed anymore, it's already included in `make`, which `yarn start` invokes.

dominiklohmann (approved):

web/blog/a-git-retrospective/index.qmd
```diff
@@ -0,0 +1,201 @@
+---
+title: A Git Retrospective
+authors: mavam
+draft: true
```
dominiklohmann, line 4: This should go before merging, I think.

mavam: Already removed, simply not yet pushed.

PR #2560, opened 2022-09-05
Fix Parquet plugin build on macOS

dominiklohmann:
size_t is not an alias for uint64_t on macOS. The only safe assumption to make is that they have the same size, but they are not required to be the same type by the standard.

Should be trivial to review. This is a regression we recently introduced in #2528.

### :memo: Reviewer Checklist

### DIFF
--- a/plugins/parquet/parquet.cpp
+++ b/plugins/parquet/parquet.cpp
@@ -465,7 +465,7 @@ public:
       co_yield slice;
   }
 
-  [[nodiscard]] size_t num_events() const override {
+  [[nodiscard]] uint64_t num_events() const override {
     return num_rows_;
   }
 


mavam:
Shouldn't CI test this?

dominiklohmann:
> Shouldn't CI test this?

We only test the plugins for the Debian build; at the time we made this decision this was mostly because of the total runtime. I'm not sure if that still applies, but in any case I don't want to make that change as part of this PR.

dispanser:
> Shouldn't CI test this?

I found the very same problem in feather (see https://github.com/tenzir/vast/pull/2528/commits/1b34442574114c87dd28ed6268cb742b6fd2e12d) due to CI (I don't run MacOS builds locally), but I'm honestly surprised that it doesn't find it for parquet - the plugin is built after all, isn't it?

dispanser (approved):

PR #2562, opened 2022-09-06
RFC-002: Fault Tolerance

tobim:
This is a proposal to add fault tolerance mechanisms to VAST. 
It's main contribution is a discussion about remodeling the `catalog` component on top of a distributed and fault tolerant "metastore".

[Rendered](https://github.com/tenzir/vast/blob/story/sc-31750/fault-tolerance-rfc/rfc/002-fault-tolerance/README.md)


rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,433 @@
...
+- **Status**: WIP
+- **Created**: Sep 5, 2022
+- **ETA**: Oct 5, 2022
+- **Authors**:
+  - [Tobias Mayer](https://github.com/tobim)
+  - [Thomas Peiselt](https://github.com/mavam)
+  - [Matthias Vallentin](https://github.com/mavam)
+- **Contributors**:
+  - [Dominik Lohmann](https://github.com/dominiklohmann)
+- **Discussion**: [PR #2511](https://github.com/tenzir/vast/pull/2511)
```
mavam, line 12:
```suggestion
- **Discussion**: [PR #2562](https://github.com/tenzir/vast/pull/2562)
```

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,433 @@
...
+  conflict-free state, it does not help with each VAST node "seeing the full
+  picture", i.e., being aware of the transactions that are triggered from within
+  other VAST nodes.
+
+* There doesn’t seem to be a full CDC (change data capture) functionality, only
+  watching single keys - there could be a key for a "transaction_id" (an integer
+  increment for every transaction), and from there we could work around the
+  problem by requesting all recent transactions whenever this key is modified.
+
+We need to read further on how to XXX.
```
dominiklohmann, line 365: There are a few XXX, TODO, TBD blocks in the document. Is it already worth doing a pass, or should I wait until this is filled in?

mavam: @tobim owns this primarily, so I would let him drive the discussion. For this particular, I'm looking on feedback on what to write.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,433 @@
...
+The final commit of the data takes place in the metastore. This is the action
+that requires an ACK, which needs to travel all the way back to the source of
+the input stream so that the event stream can be advanced. While there could be
+more fine-grained forms of ACK'ing, such as at the source or already at the
+loader, these are out of scope for this proposal because VAST partitions can be
+made arbitrarily small in theory. Specifically, VAST considers a partition as
+complete when either a maximum number of events have accumulated or when a
+timeout fires. The implication is that the source may have to keep track of the
+table slice sent to the partition builder, and acknowledge those upstream to the
+loader recursively.
```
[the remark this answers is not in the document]
dominiklohmann: I agree. 

NB: The deleted section references the loader component, which is something [we have planned](https://github.com/tenzir/public-roadmap/issues/19) but not yet implemented. We should either make such things explicit in the document, or avoid doing so.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,433 @@
+# Fault Tolerance
+
+- **Status**: WIP
+- **Created**: Sep 5, 2022
+- **ETA**: Oct 5, 2022
+- **Authors**:
+  - [Tobias Mayer](https://github.com/tobim)
+  - [Thomas Peiselt](https://github.com/mavam)
```
dispanser, line 8:
```suggestion
  - [Thomas Peiselt](https://github.com/dispanser)
```

Can't let that stand :-)

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,433 @@
...
+service. We propose to use a battle-tested implementation for this purpose, as
+opposed to hand-rolling it ourselves.
+
+Once a resilient metastore exists, we can deploy an arbitrary number of catalog
+instances to achieve the desired level of redundancy.
+
+### Case Studies
+
+We now turn to deployment case studies to validate the architectural approach of
+a fault-tolerant catalog. Furthermore, we assume that VAST has an resilient
```
dispanser, line 152:
```suggestion
a fault-tolerant catalog. Furthermore, we assume that VAST has a resilient
```
?

dispanser (commented):
some initial comments, great start but we have some open TODOs

rfc/002-fault-tolerance/README.md
tobim: oof

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,433 @@
...
+cost-benefit spectrum, e.g., power, networking, or disk. We treat failure of any
+of the core components as complete system failure. For example, if the network
+interface dies, the machine is unreachable, which we treat as an entire system
+failure.
+
+![Single Deployment](single-deployment.png)
+
+No fault tolerance mechanism can work around an entire system failure. The only
+requirement is that VAST crashes safely, without corrupting its existing state.
+The resilient read/write path and metastore achieve this goal.
```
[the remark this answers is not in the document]
tobim: We discussed this topic in a call and came to the conclusion that embedded or edge deployments usually don't have the same fault tolerance requirements as central nodes, and even in that case one could still connect to a carrier with delivery guarantees.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,433 @@
...
+
+We are assuming that we can begin with reading from a reliable medium that
+supports an indexed stream of events, such as Apache Kafka. If no such data
+acquisition mechanism is available, we may consider a local disk buffer to
+emulate a reliable medium. For this proposal, working around an unreliable data
+transport channel is out of scope.
+
+The final commit of the data takes place in the metastore. This is the action
+that requires an ACK, which needs to travel all the way back to the source of
+the input stream so that the event stream can be advanced. While there could be
```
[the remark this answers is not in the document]
tobim: @dispanser convinced me in a call that the described "ack"ing scheme is not sound. A change proposal for this paragraph will come later.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,433 @@
...
+
+TBD
+
+## Annex
+
+### A1: Considerations regarding the persistence layer as a shared component
+
+As partitions are identified by UUID we consider the write end mostly
+conflict-free. In case a name clash happens anyways the filesystem abstraction
+can just reject one of the writes, this problem can be solved client side by
```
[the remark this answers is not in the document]
tobim: I wrote this paragraph as an intro to the rest of these considerations. Removing it would make the rest harder to follow, even though the issue here is rather theoretical.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,433 @@
...
+
+Ideally, a partition shall be removed from persistent storage when it is not
+reachable from any node any more.
+
+Deletions are executed with a delay after the transaction was ACK'ed. The
+deletion would be triggered from the metastore that submitted the transaction,
+on whichever node it lives.
+
+This approach introduces a new tuning knob, namely the duration of the delay
+before the deletion takes place on disk.
```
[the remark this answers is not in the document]
tobim: We discussed rollback and time-travel, but can't identify a use case for them.
As described in your comment, this could be added later if the need arises.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,426 @@
...
+
+- Watches are possible on single keys and prefixes in the key-space, which would
+  allow us to model our use-case naturally.
+
+- Etcd comes with a gRPC API, making it easy to embed for us. We're already
+  depending on gRPC transitively via Apache Arrow (cf. Flight).
+
+##### Consul
+
+TBD
```
tobim, line 377:
```suggestion
Consul is a service discovery and distributed configuration system that contains a distributed key-value store (consul KV).

* multi-Key transactions are not advertised.
* HTTP API
* no suitable C++ client library available.
```

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,426 @@
...
+
+- Multi write transactions are shaped like if-then-else statements, they always
+  have a comparison, a list of writes on success and a list of writes on failure
+
+- Watches are possible on single keys and prefixes in the key-space, which would
+  allow us to model our use-case naturally.
+
+- Etcd comes with a gRPC API, making it easy to embed for us. We're already
+  depending on gRPC transitively via Apache Arrow (cf. Flight).
+
```
tobim, line 374:
```suggestion

- An official C++ client library exists:https://github.com/etcd-cpp-apiv3/etcd-cpp-apiv3
```

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,426 @@
...
+VAST already had a
+[working-albeit-minimal](https://github.com/tenzir/vast/pull/926) implementation
+in the past that we removed, given that we didn't need the functionality at the
+time. There are some [open questions on composability of log and state
+machine](https://groups.google.com/g/raft-dev/c/O80OSWrQieo) that we should
+revisit.
+
+## Alternatives
+
+TBD
```
tobim, line 400:
```suggestion
### Autonomous Neighbors

The Catalog component could be extended with a bespoke replication mechanism to track the state update of all other Catalogs in the same cluster. A recovering node would receive a delta of the updates that it missed during downtime to synchronize with its peers.

The advantage of doing this in the application layer is that it could naturally extend from the metadata to the data plane, freeing VAST from the dependency on a distributed storage layer.

This idea has two major drawbacks: It would complicate the API surface of the catalog component and require far-reaching refactorings in many client components. Additionally, the difficulty in designing and implementing a correct resynchronization protocol is not to be underestimated.
```

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,426 @@
...
+
+We are assuming that we can begin with reading from a reliable medium that
+supports an indexed stream of events, such as Apache Kafka. If no such data
+acquisition mechanism is available, we may consider a local disk buffer to
+emulate a reliable medium. For this proposal, working around an unreliable data
+transport channel is out of scope.
+
+The final commit of the data takes place in the metastore. This is the action
+that requires an ACK, which needs to travel all the way back to the source of
+the input stream so that the event stream can be advanced.
```
dispanser, line 250: A transaction that writes one or more partitions needs to store additional metadata, containing the offsets this partition is based on. This atomic commit of both the transaction and its (kafka) source metadata allows the source to pick up exactly at the point of failure, without risk of re-ingesting twice or missing any events. It thus provides exactly-once semantics for our write path.

dispanser, line 250:
```suggestion
A transaction that writes one or more partitions needs to store additional
metadata, containing the offsets this partition is based on. This atomic commit
of both the transaction and its (kafka) source metadata allows the source to
pick up exactly at the point of failure, without risk of re-ingesting twice or
missing any events. It thus provides exactly-once semantics for our write path.
```

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,426 @@
...
+updates, such that the catalog can learn about new partitions from other catalog
+instances.
+
+Once a catalog learns about a new partition, it will load its sketches so that
+they become available for querying.
+
+Given these requirements, we briefly looked at the following systems with regard
+to their suitability.
+
+#### Candidates
```
netantho, line 319: I understand that you're still researching, once you have a clear idea, it would be good to clearly say which candidate you recommend. I don't feel I have enough information to fully review the proposal until the trade-offs are clearer.

netantho (commented):
So far looks good for my use-cases:
- Current use-case: Single node with using local filesystem. The ACKs on the write-paths will be great.
- Future potential use-case: Multi-cloud deployments with sometimes unreliable networking. The ability of the implementation to deal with brain-split will probably be key. The future expansion allowing cloud-based storage will also be great (S3 API-compatible).

rfc/002-fault-tolerance/README.md
tobim: I don't think we will fully commit to one specific solution in the scope of this RFC, although I think we can narrow it down to "native RAFT" or etcd. I would still keep the preliminary evaluation of the other alternatives to show that those were also considered.
As etcd is already battle tested I think it makes sense to give it the recommendation.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,443 @@
...
+
+Etcd is a distributed key-value store that uses Raft for quorum-based consensus. From a CAP Theorem perspective, this brings us into the CP realm, because a master must be elected to perform writes. Stale reads may be supported, though, which we can tolerate.
+
+Here are some incomplete notes of taking a first look:
+
+- Multi write transactions are shaped like if-then-else statements, they always
+  have a comparison, a list of writes on success and a list of writes on failure
+
+- Watches are possible on single keys and prefixes in the key-space, which would
+  allow us to model our use-case naturally.
```
tobim, line 370:
```suggestion
  allow us to model our use-case naturally. Importantly, key updates are streamed to the client,
  allowing a gap-free observation of changes in the key-space.
```

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,443 @@
...
+share a single metastore. We would need at least 3 Raft instances to obtain a
+quorum majority when 1 node fails, 5 Raft nodes to tolerate 2 node failures.
+
+VAST already had a
+[working-albeit-minimal](https://github.com/tenzir/vast/pull/926) implementation
+in the past that we removed, given that we didn't need the functionality at the
+time. There are some [open questions on composability of log and state
+machine](https://groups.google.com/g/raft-dev/c/O80OSWrQieo) that we should
+revisit.
+
```
tobim, line 408:
```suggestion

#### Recommendation

As of our current understanding etcd and the home grown Raft implementation match the requirements most closely. Of those two etc is the mature product which has been hardened and improved over many years of heavy use. For that reason **we recommend etcd for the backend** of the fault tolerant catalog.
```

rfc/002-fault-tolerance/README.md
tobim: I incorporated that into the proposal with [db37533](https://github.com/tenzir/vast/pull/2562/commits/db37533243683055f01b4eb34da35c99a10da563).

tobim:
I would like to bump the ETA to October 19. @mavam is that OK?

rfc/002-fault-tolerance/README.md
dominiklohmann: That's great. I probably won't get to reviewing it in depth this week, but I've created a reminder for me to go through early next week.

rfc/002-fault-tolerance/README.md
netantho: Thanks, that makes sense and helps me :-) (feel free to close this thread)

netantho (approved):
I think the proposal matches the requirements I would have if I had to run VAST with high availability. The availability guarantees seem good for while still keeping a relativelz low operational overhead. And thanks for considering and documenting various options, that helps understanding the constraints and trade-offs.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+The catalog requires the following interface from the metastore:
+
+1. `get: [uuid] → [partition]`
+2. `add: [partition] → ACK`
+3. `delete: [uuid] → ACK`
+4. `replace: [uuid], [partition] → ACK`
+
+The semantics of the mentioned types are as follows:
+
+- `partition`: metadata about a partition, such as its URL and its schema.
```
dominiklohmann, line 295: This looks like it should be a superset of what is currently called the "partition info", i.e., the state that the catalog holds per partition except for the sparse index. Should we re-use that name here, as it already is established in the code base?

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+2. `add`: registers a list of new partitions.
+
+3. `delete`: removes a list of partition from the metastore (doesn't delete the
+   underlying data)
+
+4. `replace`: performs an atomic deletion and addition of partition data, by
+   deleting the given UUIDs and adding new partition metadata.
+
+In theory, the functionality of `replace` is a superset of `add` and `delete`.
```
dominiklohmann, line 312:
```suggestion
In praxis, the functionality of `replace` is a superset of `add` and `delete`.
```

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+For ease of exposition, we kept them as separate API functions.
+
+#### Subscription API
+
+The metastore must also expose a subscription mechanism to hook into the feed of
+updates, such that the catalog can learn about new partitions from other catalog
+instances.
```
dominiklohmann, line 319: How do we handle failure for this mechanism? Does committing wait for a response from all subscribers?

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+Once a catalog learns about a new partition, it will load its sketches so that
+they become available for querying.
```
dominiklohmann, line 322: IIUC this implies that the catalog responds to the request to add a new partition only after it has loaded its sketches. Can we make this explicit?

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+  forbidden, even if they overlap - to enable our use case, we’d have to verify
+  the existence of all the partitions we intend to delete, followed by doing all
+  the (SET, CLEAR) operations OR rollback / cancel the transaction.
+
+* While this strategy allows for serializing the transactions into a
+  conflict-free state, it does not help with each VAST node "seeing the full
+  picture", i.e., being aware of the transactions that are triggered from within
+  other VAST nodes.
+
+* There doesn’t seem to be a full CDC (change data capture) functionality, only
```
dominiklohmann, line 361: Style confusion: Why use * here and - everywhere else for the list? (Not that it really matters...)

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+  watching single keys - there could be a key for a "transaction_id" (an integer
+  increment for every transaction), and from there we could work around the
+  problem by requesting all recent transactions whenever this key is modified.
+
+##### etcd
+
+Etcd is a distributed key-value store that uses Raft for quorum-based consensus.
+From a CAP Theorem perspective, this brings us into the CP realm, because a
```
dominiklohmann, line 369:
```suggestion
From a CAP theorem perspective, this brings us into the CP realm, because a
```

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+master must be elected to perform writes. Stale reads may be supported, though,
+which we can tolerate.
+
+Here are some incomplete notes of taking a first look:
+
+- Multi write transactions are shaped like if-then-else statements, they always
```
dominiklohmann, line 375:
```suggestion
- Multi-write transactions are shaped like if-then-else statements, they always
```

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+- A seemingly maintained third party client library with the relevant features
+- exists: https://github.com/oliora/ppconsul. The relevant transaction API is
+- supported.
+
+- While it is possible to add run a blocking query to a single key or a prefix,
+- such a query has to be re-registered after every return. If multiple changes
+- have been applied before the new query is registered the client misses
+- updates. We can not observe changes in the key space directly, only the
+- current values. That means modeling a change feed as desired requires a custom
+- protocol.
```
dominiklohmann, line 417: There are lots of dashes here that throw off the rendering.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+Consul is a service discovery and distributed configuration system that contains
+a distributed key-value store (consul KV). Apart from the bare key-value store
+it also provides service mesh, traffic management, and service to service
+encryption. For the use case at hand we identified the following relevant
+properties:
+
+- The only way to interact with a Consul service is through an HTTP API, meaning
+- there would be some overhead for each transaction. This also means that we
+- would have to use HTTP long polling to observe changes on a specific key.
```
dominiklohmann, line 398: And that is bad because...?

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+As of our current understanding etcd and the home grown Raft implementation
+match the requirements most closely. Of those two etc is the mature product
+which has been hardened and improved over many years of heavy use. For that
+reason **we recommend etcd for the backend** of the fault tolerant catalog.
+
+## Alternatives
+
+### Autonomous Neighbors
+
+The Catalog component could be extended with a bespoke replication mechanism to
```
dominiklohmann, line 449:
```suggestion
The catalog component could be extended with a bespoke replication mechanism to
```

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+track the state update of all other Catalogs in the same cluster. A recovering
```
dominiklohmann, line 450:
```suggestion
track the state update of all other catalogs in the same cluster. A recovering
```

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+a distributed storage layer.
+
+This idea has two major drawbacks: It would complicate the API surface of the
+catalog component and require far-reaching refactorings in many client
+components. Additionally, the difficulty in designing and implementing a correct
+resynchronization protocol is not to be underestimated.
+
+## Annex
+
+### A1: Considerations regarding the persistence layer as a shared component
```
dominiklohmann, line 465:
```suggestion
### A1: Considerations Regarding the Persistence Layer as a Shared Component
```

dominiklohmann (approved):
I gave this a very thorough review. I'm fine with the direction of this generally speaking. I'm approving this already under the assumption that all of my open questions will be clarified.

Thanks so much for your work on this, this was a really big (but necessary) piece of work for the evolution of work. I'm excited to see how we'll implement this going forward. 🚀

rfc/002-fault-tolerance/README.md
dominiklohmann: I finally got to this; it's approved modulo 32 outstanding comments. Resolving this thread.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+A transaction that writes one or more partitions needs to store additional
+metadata, containing the offsets this partition is based on. This atomic commit
+of both the transaction and its source position allows the source to pick up
+exactly at the point of failure, without risk of re-ingesting twice or missing
+any events. It thus provides exactly-once semantics for our write path.
+
+In technical terms, we can extend the CAF Streaming message type to include a
+handle to the source and source specific position, and then trigger an ACK once
+the data has been persisted. VAST already implements a similar mechanism today
```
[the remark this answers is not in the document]
rdettai: I don't know what the relationship should be, but I also believe it should be explicit because it helps understanding the sequencing of the transaction commit

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+that acknowledges when events arrived at an active partition, which (almost)
+guarantees that they are queryable.
+
+When an exising source reconnects to an index component after a fault, it
+needs to request the most recently acknowlegded source position from the
+catalog and advance to the returned position to prevent double ingestion.
```
rdettai, line 262: Actually I see the skipping on the client side of events that were already ingested as an optimization. The really crucial mechanism is for the transactional component to discard the events that were already ingested.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+In the above diagram, both VAST nodes serve as a valid entry point for data, as
+well as queries. If one location fails, the other remains fully operational.
+This architecture scales beyond two locations, with the metastore remaining the
+bottleneck. Therefore, finding a high-performance and scalable metastore is of
+essence.
+
+#### Example: Load-based scaling
+
+When VAST is deployed on top of an auto scaling engine, like Kubernetes or
```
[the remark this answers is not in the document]
tobim: This is usually written as either one word "autoscaling" or two, but rarely with a hyphen.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+A transaction that writes one or more partitions needs to store additional
+metadata, containing the offsets this partition is based on. This atomic commit
+of both the transaction and its source position allows the source to pick up
+exactly at the point of failure, without risk of re-ingesting twice or missing
+any events. It thus provides exactly-once semantics for our write path.
+
+In technical terms, we can extend the CAF Streaming message type to include a
+handle to the source and source specific position, and then trigger an ACK once
+the data has been persisted. VAST already implements a similar mechanism today
+that acknowledges when events arrived at an active partition, which (almost)
```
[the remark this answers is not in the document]
tobim: The notification logic is not pushed to the indexers, it ends at the partition. And even it it were, for the current context we treat the indexers as an implementation detail of the partition.

rfc/002-fault-tolerance/README.md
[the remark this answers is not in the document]
tobim: It does not make any assumption about the relationship between sources and partitions, note that the text says "offsets", not "offset".

@rdettai unfortunately we can't make a hard decision about this here because we want to retain some tuning / optimization opportunities in this area for now.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+units of partitions. Assuming that we can build a fault-tolerant catalog, the
+problem of making the write path fault-tolerant reduces to ensuring exactly-once
+processing semantics on the ingest stream.
+
+If the circumstances require high-fidelity data acquisition, we can resort to an
+external data plane with exactly-once processing stream semantics, such as
+Apache Kafka. VAST must not acknowledge events from the data plane until having
+received a confirmation that the corresponding partition is durable. In the
+presence of a crash, VAST must restart from the last acknowledged position in
+the stream.
```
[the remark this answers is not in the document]
tobim: Why do you think it implies that?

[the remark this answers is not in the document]
dominiklohmann: How do you roll back otherwise if you store this on a partition level? Say you have two sources that ingest events of the same schema. Now one source crashes. Can you still ack the other source?

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+Because security telemetry data is already lossy upstream (e.g., due to packet
+loss or lacking instrumentation), operators are often willing to accept looser
+guarantees. Therefore, we deem it sufficient to offer a mechanism that
+implements interaction with a durable message bus.
```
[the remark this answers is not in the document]
tobim: As @rdettai already wrote [in another comment](https://github.com/tenzir/vast/pull/2562/files#r1003186045), this isn't actually the correct way to establish exactly once. I'll change this paragraph accordingly.

I also don't want to go into the various options in detail here, because there are tradeoffs to be explained for each possible level of granularity and the additional text would only distract from the center of focus.

In the end, the entire section here is about dividing the problem into sub-problems. In the end, this RFC mainly concerns itself with the ground truth of metadata storage. If we need further architectural changes to harden the write path they should best be discussed in a separate RFC.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+cost.
+
+## Solution Proposal
+
+The most important architectural effort concerns making the catalog a
+fault-tolerant component. To this end, we propose to factor the logic that
+operates on partition metadata into a dedicated **metastore** component. As a
+first point of contact for queries, the catalog also keeps per-partition index
+structures in memory. This catalog functionality is out of scope, as it
+represents derived, immutable state from the metastore.
```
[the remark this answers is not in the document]
tobim: I reserve myself the freedom use the word informally here. These questions can only be answered with a lot more context, meaning further down the document.

For you: Whether the metastore is encapsulated in the catalog or exposed as a standalone service actor depends on whether we want to make it available to other clients besides the catalog. It might also depend on external processes services depending on the backing implementation. How this is all organized is an exercise in code organization.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+Adjacent fault-tolerance efforts outside the catalog scope involve replication
+of components. For example, the failure domain of the read path is a single
+query, which we can deploy in a dedicated operating system process (that extends
+to a single cloud function or container execution). Likewise, we can isolate
+failure domains on the write path by spawning each data source (or pipeline) in
+its own process. This flexibility is a by-product of the actor model
+architecture that treats internal components as microservices. The spectrum of
+fate sharing is the OS process boundary. The choice highly depends on the
+specifics of the deployment environment.
```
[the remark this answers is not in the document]
tobim: Isolation, should have written isolation...

rfc/002-fault-tolerance/README.md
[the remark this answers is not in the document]
dominiklohmann: The term "component" is already established in VAST, and everyone knowing that will read this sentence differently (as did I). That's why I'm asking you to make this more explicit.

rfc/002-fault-tolerance/README.md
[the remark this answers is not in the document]
dominiklohmann: So s/replication/isolation?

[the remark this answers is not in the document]
tobim: Yes. i replaced it now and explained that this mechanism is free once we have the meta store.

rfc/002-fault-tolerance/README.md
tobim: It is neither "good" nor "bad", but important to know nonetheless. The last point in this list explains the limitations that cone with a plain HTTP API.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+metastore-specific integration.
+
+#### Request-Response API
+
+The catalog requires the following interface from the metastore:
+
+1. `get: [uuid] → [partition]`
+2. `add: [partition] → ACK`
+3. `delete: [uuid] → ACK`
+4. `replace: [uuid], [partition] → ACK`
```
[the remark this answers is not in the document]
tobim: You're right. This is function is missing.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+When an exising source reconnects to an index component after a fault, it
+needs to request the most recently acknowlegded source position from the
+catalog and advance to the returned position to prevent double ingestion.
+
+### Read Path
+
+For historical queries, we can achieve fault-tolerance on the read part by
+moving query execution into a dedicated process. This requires that the catalog
+is the sole owner of partitions, and that the catalog returns the set of
+candidate partitions relevant to answer a query.
```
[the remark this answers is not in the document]
tobim: In a running process the catalog is the owner. 

Here we talk about fault tolerant query execution that comes after the catalog lookup.

If you want fault-tolerance end-to-end you need the distributed consensus style catalog as proposed below.

Now that I think of it this clause is only confusing. It cryptically hints at the current index playing a part in partition ownership at startup before they are transferred to the catalog. This is simply a historical artifact of when the catalog was a sub component of the index.

In any case, this is too nitty-gritty and I'll just take it out.

rfc/002-fault-tolerance/README.md
tobim: I don't think this is true. Sketches can still be loaded lazily when a query that hits the corresponding table (type) arrives. We only need to make sure that the list of UUIDs per table is consistent.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+Live queries bypass the catalog and would therefore need to implement ACK'ing
+independently, at the granularity of table slices. We do not cover this process
+in detail, as it constitutes a subset of the ACK'ing for the write path outlined
+above.
```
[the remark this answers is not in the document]
tobim: In a sense you're right, brushing this aside as a mere simple case of the write path is a bit dismissive.

I believe the acking has to be done by the final sink of the data stream, which probably means multiple tools have to agree on the protocol here. Definitely out of scope.

rfc/002-fault-tolerance/README.md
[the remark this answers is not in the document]
tobim: Yes, data that is not commited is not queryable. I think this actually makes sense, as it resolves a notorious unpredictability problem around concurrent reads and writes.

I imagine the latency issue will be resolved by actually ousting the active partition and commiting each incoming slice as an unindexed partition directly to the catalog. The indexing will then only be done as part of the rollup.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+## Implementation
+
+We briefly touch on the implementation of write and read path, before dedicating
+the majority of attention to the metastore
+
+### Write Path
+
+To achieve fault-tolerance on the write path, we need an acknowledgement
+mechanism to confirm whether events at the source have been successfully
+processed. The current architecture looks as follows:
```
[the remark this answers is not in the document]
tobim: I don't think we should burden ourselves with legacy-this or legacy-that. We only want to discuss the relevant *delta* to make it fault tolerant. If we discard the changes we already agreed upon but did not implement yet we're going to multiply the length of this document while blurring the message at the same time.

I could change it to "The currently planned architecture", but isn't this only irritating?

Would you agree to "starting architecture" or "base architecture"?

rfc/002-fault-tolerance/README.md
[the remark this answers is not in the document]
tobim: "component" is still the most suitable term and it is used intentionally here. We should not call it an actor because several of the potential backends would run in a separate process that would only be wrapped by an actor, others would consist of multiple actors.
I will also refrain from specifying who shall create / own the metastore here. Artificial constraints only add complexity to the overall design.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+We briefly touch on how the mechanism for making the write and read path
+resilient, and then focus the discussion on fault tolerance within the VAST
+server node, specifically the catalog.
+
+### Resilient Write Path
+
+Data sources provide a stream of events. VAST parses them and bundles them into
+units of partitions. Assuming that we can build a fault-tolerant catalog, the
+problem of making the write path fault-tolerant reduces to ensuring exactly-once
+processing semantics on the ingest stream.
```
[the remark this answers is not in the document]
tobim: >     * Does this intentionally ignore batching?

yes

>     * Are exactly-once semantics wanted on the event or the partition level?
> 
>     * How do we handle partitions that contain events from multiple sources?

Both these questions are now answered in the implementation section for the write path.

rfc/002-fault-tolerance/README.md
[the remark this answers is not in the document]
tobim: As a result of @rdettai I revised the section on the write path implementation. Incoming data from sources must indeed now be in their own partition.

The reason for this isn't crashing sources alone (that could be solved by storing dedicated offsets  for every source that contributed to a partition), but sources that loose their state while a database node experiences a fault. The originally proposed approach would break in this case so I changed it so the catalog would discard duplicate partitions, and that only works if a partition comes from a single source.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+### Resilient Read Path
+
+The desired fault isolation domain of the read path is the individual query that
+may crash or become a straggler. To date, VAST does not perform distributed
+query execution, so we can consider a single query an atomic execution unit.
+When a query fails, the remainder of the system remains unaffected if the query
+ran in its own execution context. For example, a crashing cloud function will
+not affect other running queries. But when VAST executes multiple queries in the
+same server node, all queries exhibit fate sharing. We leave it to the operator
+to decide what model works best for the scenario at hand.
```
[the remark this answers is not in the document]
tobim: A "query crashes" means that code execution involved in evaluating a query leads to process termination. If the entire database lives in a single process with out replication this means that the entire service goes down.
This is why this sections talks about process isolation.
Additionally, even if the process doesn't terminate, a faulty query execution might still negatively affect other queries, either blocking them or by making them produce wrong results, which is arguably even worse.

rfc/002-fault-tolerance/README.md
tobim: Comment that this limits us in an annoying way.

rfc/002-fault-tolerance/README.md
tobim: TODO: What I commented is actually more correct than what is in the text, so update.

rfc/002-fault-tolerance/README.md
tobim: To address the original comment: This is only about the API of the metastore and not the API of the catalog. The latter is not covered in this doc, so I'm resolving this thread.

rfc/002-fault-tolerance/README.md
tobim: TODO: Look up if this is the correct phrasing.

rfc/002-fault-tolerance/README.md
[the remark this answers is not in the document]
tobim: As a side-effect of the change to make atomic inserts to the catalog to main mechanism for dedup this was actually added.

rfc/002-fault-tolerance/README.md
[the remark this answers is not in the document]
tobim: Acked by @dominiklohmann .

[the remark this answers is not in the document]
tobim: TODO: Add a preamble that we only talk about the delta to the previous architecture, even if that is not fully implemented yet.

rfc/002-fault-tolerance/README.md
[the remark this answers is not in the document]
tobim: covered in another thread.

rfc/002-fault-tolerance/README.md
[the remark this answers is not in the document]
tobim: TODO: clarify that this partially depends on the distributed catalog.

rfc/002-fault-tolerance/README.md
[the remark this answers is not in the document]
tobim: TODO: Replace the last sentence with::
```
This may require a unique identifier for every data source, possibly requiring the user to provide one explicitly.
```

rfc/002-fault-tolerance/README.md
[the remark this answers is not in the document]
tobim: We talked about this and found a compromise: Just take the work "component" out, the sentence still works the same.

rfc/002-fault-tolerance/README.md
tobim: I actually took out the last sentence and edited the last bullet instead.

rfc/002-fault-tolerance/README.md
tobim: I was not able to find any source that claims "in praxis" is proper usage. I'll use "in practice" for now until you are able to convince me otherwise :wink:.

rfc/002-fault-tolerance/README.md
```diff
@@ -234,62 +240,58 @@ adds the requirement for a quick resynchronization ability to the metastore.
...
+server (and thereby the catalog) crashes during partition submission. In this
+case the ACK of the transaction may not arrive at the client. The client would
+then have to re-submit the same partition after a timeout. This will not
+introduce duplicates if the client initiates the submission with a uniquely
+chosen transaction ID. Since the server keeps a durable log of all transaction
+IDs, it will not accept an already completed transaction again. Alternatively,
+the server can generate a unique transaction ID and require the client to
+provide that ID during the final submission. Both approaches have the same
+effect, and we leave the implementation detail to the specific metastore at
+hand.
```
mavam, line 287: @tobim this is the only content delta of my commit. The rest is wordsmithing.

mavam (approved):
Reads nicely to me now! Ready to ship if @tobim agrees on the section on duplicate detection.

rfc/002-fault-tolerance/README.md
tobim: I think server-issued transaction ids are an interesting idea, but they seem pretty under specified right now. Naively, if the server crashes the state that supports the transaction id generation may get lost, and we're back to square one.

mavam: But when the client commits a transaction, with either their own or server-generated ID, and doesn't get an ACK, the client would retry regardless with the same ID.

tobim: The latest addition clears my concerns.

tobim:
I think this is finally ready. Any objections against merging today already? If not I would do that later.

mavam:
It's ready to go from my side.

There is one outstanding comment from @dominiklohmann. When I did a full pass yesterday, I didn't find it worth going into detail, but if you feel strongly about it, @dominiklohmann, it should be addressed prior to merging.

rfc/002-fault-tolerance/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+We consider two approaches:
+
+1. **Active-passive:** (or hot-spare) a redundant machine replaces the primary system that failed.
+2. **Load sharing**: there exist multiple system instances that operate on the same metastore.
+
+Option (1) falls into the high-availability (HA) category that requires
+substantial involvement from operations, as it requires non-trivial
+re-allocation of resources in realtime. It often provides a first-order solution
+when the performance requirements are met.
```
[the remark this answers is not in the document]
dominiklohmann: Not important enough to bikeshed, so closing this.

rfc/002-fault-tolerance/README.md
```diff
@@ -1,8 +1,8 @@
 # Fault Tolerance
 
-- **Status**: In-Review
+- **Status**: Accepted
 - **Created**: Sep 5, 2022
-- **ETA**: Nov 30, 2022
+- **Accepted**: Nov :8, 2022
```
mavam, line 5: Why is there a colon in there?

tobim: I have no explanation for this. Fix incoming.

tobim: #2741

PR #2567, opened 2022-09-07
Add initial REST API server for VAST

lava:
### :memo: Reviewer Checklist


dominiklohmann:
[sc-34483]

plugins/rest/src/plugin.cpp
```diff
@@ -0,0 +1,100 @@
...
+        .add<std::string>("key-path", "path to TLS private key")
+        .add<std::string>("bind", "listen address of server")
+        .add<uint16_t>("port", "listen port"));
+    rest_command->add_subcommand("generate-token", "generate auth token",
+                                 command::opts("?rest.token"));
+    rest_command->add_subcommand("specification", "print openAPI spec",
+                                 command::opts("?rest.spec"));
+    auto token_command
+      = [](const vast::invocation&, caf::actor_system&) -> caf::message {
+      auto& server = server_singleton();
```
lava, line 61: Make this a `component_plugin` as well, so that the server state can be implemented in the component to have shared access and can be persisted in the server.

plugins/rest/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+
+find_package(VAST REQUIRED)
+VASTRegisterPlugin(
+  TARGET rest
+  ENTRYPOINT src/plugin.cpp
+  SOURCES ${rest_sources}
+  TEST_SOURCES ${rest_tests}
+  INCLUDE_DIRECTORIES include)
+
+find_package(fmt REQUIRED)
```
lava, line 26: @dominiklohmann I'm not sure why this line is required

plugins/rest/www/status.html
```diff
@@ -0,0 +1,70 @@
...
+        rgba(186, 12, 248, 1) 80%,
+        rgba(251, 7, 217, 1) 90%,
+        rgba(255, 0, 0, 1) 100%
+    );
+}
+
+.blink_me {
+  font-family: "Comic Sans", Arial, serif;
+  animation: blinker 1s linear infinite;
+}
```
lava, line 43: Remove rainbow effect and add a boring status page instead

libvast/native-plugins/rest_api/status.cpp
```diff
@@ -0,0 +1,130 @@
...
+            auto to = context.find_last_of('}');
+            bool escape = false;
+            for (auto& c : context)
+              if (c == '\\') {
+                escape = true;
+                c = ' ';
+              } else if (escape && c == 'n')
+                c = ' ';
+              else
+                escape = false;
```
lava, line 65: This needs to be fixed so we ideally get a new handler in the node that directly returns a string, but alternatively we can just use this pattern:

```
    caf::message_handler{[&](const std::string& str) {
      // use string
    }}(e.context());
```

plugins/rest/src/server_command.cpp
```diff
@@ -0,0 +1,277 @@
...
+      // }
+      vast::record params;
+      if (endpoint.params) {
+        for (auto const& leaf : endpoint.params->leaves()) {
+          auto name = leaf.field.name;
+          auto restinio_name
+            = restinio::string_view_t{name.data(), name.size()};
+          VAST_INFO("checking request parameter {}", name);
+          if (string_params.has(restinio_name)) {
+            auto value = std::string{string_params[restinio_name]};
```
lava, line 115: @dominiklohmann Here we still need to cast value to the type expected to the function.

libvast/native-plugins/rest_api/export.cpp
```diff
@@ -196,82 +192,120 @@ class plugin final : public virtual rest_endpoint_plugin {
...
+    auto result = from_yaml(spec);
+    VAST_ASSERT(result);
+    return *result;
   }
 
+  // TODO: This is a bit silly, probably it makes more sense to just
+  // return a `map` from `plugin::endpoints()`?
+  constexpr static uint64_t EXPORT_ENDPOINT = 0ull;
+  constexpr static uint64_t QUERY_NEW_ENDPOINT = 0ull;
+  constexpr static uint64_t QUERY_NEXT_ENDPOINT = 0ull;
```
mavam, line 275: Why do we need these IDs? I couldn't infer the use case.

Dockerfile
```diff
@@ -19,11 +19,13 @@ RUN apt-get update && \
       git-core \
       gnupg2 \
       jq \
+      libasio-dev \
       libcaf-dev \
       libbroker-dev \
       libflatbuffers-dev \
       libfmt-dev \
       libpcap-dev tcpdump \
+      librestinio-dev \
```
dominiklohmann, line 28: These need to be added to the instructions for building from source, and need to be installed in CI (where this plugin isn't being built yet).

plugins/rest/src/server_command.cpp
```diff
@@ -0,0 +1,312 @@
...
+    return req->create_response(restinio::status_not_found()).done();
+  });
+  router->http_get(
+    "/", [](auto request, auto) -> restinio::request_handling_status_t {
+      return request->create_response()
+        .append_header(restinio::http_field::server, "VAST")
+        .append_header_date_field()
+        .append_header(restinio::http_field::content_type, "text/html; "
+                                                           "charset=utf-8")
+        .set_body(restinio::sendfile("../plugins/rest/www/status.html"))
```
dominiklohmann, line 260: This will only work when running VAST from the build tree. We should install the www directory somewhere into DATAROOTDIR, I think.

plugins/rest/src/server_command.cpp
```diff
@@ -0,0 +1,312 @@
...
+          },
+          [self, response](const caf::error& err) mutable {
+            VAST_WARN("{} received error while authenticating request: {}",
+                      *self, err);
+            response->abort(500, "internal server error");
+          });
+    },
+    [self](atom::internal, atom::request, restinio_response_ptr& response,
+           const api_endpoint& endpoint, system::rest_handler_actor handler) {
+      // TODO: Also consider params in the request body here.
```
dominiklohmann, line 112: Do we need to implement this before merging, or is this for a follow-up story?

plugins/rest/src/generate_token_command.cpp
```diff
@@ -0,0 +1,42 @@
...
+  if (auto* err = std::get_if<caf::error>(&node_opt))
+    return caf::make_message(std::move(*err));
+  const auto& node
+    = std::holds_alternative<system::node_actor>(node_opt)
+        ? std::get<system::node_actor>(node_opt)
+        : std::get<scope_linked<system::node_actor>>(node_opt).get();
+  // The typed `get_node_components()` only works for actors whose type id is
+  // defined in the main vast namespace, so we have to work around manually.
+  // TODO: Introduce a `component_plugin::component_name()` function so we
+  // can name the component "authenticator" instead of "rest".
```
dominiklohmann, line 25: That name is also used for the top-level entry in `vast status`, so the question becomes whether users find `vast status rest` or `vast status authenticator` easier.

plugins/rest/src/authenticator.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include "rest/authenticator.hpp"
+
+#include <vast/detail/base64.hpp>
+
+#include <caf/scoped_actor.hpp>
+
+#include <array>
+
+// TODO: Investigate if this is available on mac
+#include <sys/random.h>
```
dominiklohmann, line 18: Is there a reason not to use the C++ standard library tools here?

plugins/rest/include/rest/restinio_response.hpp
```diff
@@ -0,0 +1,51 @@
...
+  ~restinio_response() override;
+
+  restinio_response(restinio_response&&) = default;
+  restinio_response(const restinio_response&) = delete;
+  restinio_response& operator=(const restinio_response&) = delete;
+  restinio_response& operator=(restinio_response&&) = delete;
+
+  void append(std::string body) override;
+
+  // TODO: Add `&&` qualifier?
```
dominiklohmann, line 39: Probably makes sense given that it's rather likely for the object to be in an unusable state after calling abort.

libvast/test/rest_api.cpp
```diff
@@ -0,0 +1,108 @@
...
+  auto request = vast::http_request{
+    .params = {},
+    .response = response,
+  };
+  self->send(handler, vast::atom::http_request_v, status_endpoint.endpoint_id,
+             std::move(request));
+  run();
+  CHECK_EQUAL(response->error_, caf::error{});
+  CHECK(!response->body_.empty());
+  // TODO: Verify that the response contains valid YAML
```
dominiklohmann, line 99: That can probably be done with `from_yaml`, which parses YAML strings into a `data`.

libvast/include/vast/type.hpp
```diff
@@ -1384,6 +1384,13 @@ class record_type final : public stateful_type_base {
 
   /// Returns a new, flattened record type.
   friend record_type flatten(const record_type& type) noexcept;
+
+  /// Enables integration with CAF's type inspection.
+  template <class Inspector>
+  friend auto inspect(Inspector& f, record_type& x) ->
+    typename Inspector::result_type {
+    return f(caf::meta::type_name("vast.record_type"), x.table_);
+  }
```
dominiklohmann, line 1393: Adding this for concrete types looks rather dangerous to me. CAF's type inspection only works well for semi-regular types, and record type isn't semi-regular, so this will never be usable unless you wrap it into an optional. That means that we have no way to test this inspect handler.

Could you change the code that is using this to use a `type` rather than a `optional<record_type>`?

libvast/include/vast/plugin.hpp
```diff
@@ -333,6 +340,32 @@ class query_language_plugin : public virtual plugin {
...
+
+  /// List of API endpoints provided by this plugin.
+  [[nodiscard]] virtual const std::vector<api_endpoint>&
+  api_endpoints() const = 0;
+
+  /// Actor that will handle this endpoint.
+  //  TODO: This should get some integration with component_plugin so that
+  //  the component can be used to answer requests directly.
+  [[nodiscard]] virtual system::rest_handler_actor
+  handler(caf::actor_system& system, system::node_actor node) const = 0;
```
dominiklohmann, line 366: How much effort do you think it would be to refactor this to not require plugin developers to write their own actor?

libvast/include/vast/plugin.hpp
```diff
@@ -333,6 +340,32 @@ class query_language_plugin : public virtual plugin {
   parse(std::string_view query) const = 0;
 };
 
+// -- rest endpoint plugin -----------------------------------------------------
+
+class rest_endpoint_plugin : public virtual plugin {
```
dominiklohmann, line 345: Naming question: it's the REST endpoint plugin, the REST handler actor, but a set of API endpoint. Can you add a comment on how they relate, ideally somewhere in the plugin API?

libvast/native-plugins/rest_api/export.cpp
```diff
@@ -0,0 +1,237 @@
...
+#include <vast/system/node_control.hpp>
+#include <vast/system/query_cursor.hpp>
+#include <vast/table_slice.hpp>
+
+#include <caf/stateful_actor.hpp>
+#include <caf/typed_event_based_actor.hpp>
+
+namespace vast::plugins::rest_api::export_ {
+
+// clang-format off
```
dominiklohmann, line 29: If you comment the handlers similar to how we do it in actors.hpp you won't need this.

plugins/rest/include/rest/restinio_response.hpp
```diff
@@ -0,0 +1,51 @@
...
+#include <vast/plugin.hpp>
+
+#include <restinio/compiler_features.hpp>
+#include <restinio/message_builders.hpp>
+#include <restinio/request_handler.hpp>
+
+namespace vast::plugins::rest {
+
+// TODO: Does it make sense to make use of the "extra data" to store a
+// caf::intrusive_ptr ?
```
dominiklohmann, line 21: What for?

plugins/rest/include/rest/restinio_response.hpp
```diff
@@ -0,0 +1,51 @@
...
+using request_handle_t
+  = restinio::generic_request_handle_t<restinio::no_extra_data_factory_t::data_t>;
+using response_t
+  = restinio::response_builder_t<restinio::user_controlled_output_t>;
+
+class restinio_response final : public vast::http_response {
```
dominiklohmann, line 27: Since users are supposed to work with this I think it makes sense to comment this well, especially the override functions.

plugins/rest/include/rest/restinio_response.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+  // TODO: Add `&&` qualifier?
+  void abort(uint16_t error_code, std::string message) override;
+
+  // Get a handle to the original request.
+  [[nodiscard]] const request_handle_t& request() const;
+
+private:
+  request_handle_t request_;
+  response_t response_; // it looks like we need to keep this alive artificially
```
dominiklohmann, line 47: Do we understand why that is?

plugins/rest/src/authenticator.cpp
```diff
@@ -0,0 +1,185 @@
...
+}
+
+auto authenticator_state::generate() -> token_t {
+  auto random_bytes = std::array<char, 16>{};
+  // Use `getrandom()` for cryptographically secure random bytes.
+  while (auto error
+         = ::getrandom(random_bytes.data(), random_bytes.size(), 0)) {
+    // We should never get EFAULT or EINVAL unless our arguments
+    // are incorrect, and we already checked that the kernel
+    // supports `getrandom()`. (TODO on the last part)
```
dominiklohmann, line 86: Open TODO.

plugins/rest/src/authenticator.cpp
```diff
@@ -0,0 +1,185 @@
...
+    VAST_ASSERT_CHEAP(error == EAGAIN || error == EINTR);
+  }
+  auto token = detail::base64::encode(
+    std::string_view{random_bytes.data(), random_bytes.size()});
+  // TODO: Allow the caller to pass the desired name and expiry date.
```
dominiklohmann, line 91: Open TODO.

plugins/rest/src/plugin.cpp
```diff
@@ -0,0 +1,86 @@
...
+namespace vast::plugins::rest {
+
+/// The API plugin.
+class plugin final : public virtual command_plugin,
+                     public virtual component_plugin {
+  /// Initializes a plugin with its respective entries from the YAML config
+  /// file, i.e., `plugin.<NAME>`.
+  /// @param config The relevant subsection of the configuration.
+  [[nodiscard]] caf::error initialize(data data) override {
+    // FIXME
```
dominiklohmann, line 37: Open FIXME

plugins/rest/src/plugin.cpp
```diff
@@ -0,0 +1,86 @@
...
+    auto config = to<configuration>(data);
+    return {};
+  }
+
+  /// Returns the unique name of the plugin.
+  [[nodiscard]] const char* name() const override {
+    return "rest";
```
dominiklohmann, line 44: In terms of naming, do you think it makes sense to call this `web` instead, since it's more powerful than just a REST API?

plugins/rest/src/plugin.cpp
```diff
@@ -0,0 +1,86 @@
...
+        .add<std::string>("mode", "Server mode. One of "
+                                  "dev,server,upstream,mtls.")
+        .add<std::string>("certificate-path", "path to TLS cert")
+        .add<std::string>("key-path", "path to TLS private key")
+        .add<std::string>("bind", "listen address of server")
+        .add<uint16_t>("port", "listen port"));
+    rest_command->add_subcommand("generate-token", "generate auth token",
+                                 command::opts("?rest.token"));
+    rest_command->add_subcommand("specification", "print openAPI spec",
+                                 command::opts("?rest.spec"));
```
dominiklohmann, line 75: These all need documentation, ideally both in a README.md and on vast.io.

plugins/rest/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+cmake_minimum_required(VERSION 3.18...3.23 FATAL_ERROR)
+
+project(
+  rest
+  VERSION 1.0.0
```
dominiklohmann, line 5: This plugin is yet lacking a CHANGELOG.md file that matches this version.

plugins/rest/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+cmake_minimum_required(VERSION 3.18...3.23 FATAL_ERROR)
```
dominiklohmann, line 1: I think upstream is at 3.24 now for the policy opt-in version.

plugins/rest/include/rest/configuration.hpp
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  enum class server_mode {
+    debug,
+    server,
+    upstream,
+    mtls,
+  };
+
+  std::string mode = "debug"; // FIXME: switch default to 'server'
```
dominiklohmann, line 42: Unaddressed FIXME.

libvast/native-plugins/rest_api/export.cpp
```diff
@@ -0,0 +1,237 @@
+//    _   _____   __________
```
dominiklohmann, line 1: Would it maybe make more sense to bundle these native plugins as part of the REST plugin itself? I think that can be implemented in our CMake with some shuffling around.

The URL should probably also match plugin name, i.e., `rest_endpoints` rather than `rest_api`.

dominiklohmann (commented):
I looked at the code more in depth and left a bunch of comments on things that I think should be looked at.

There are no major issues with the code. It'd be nice if the REST endpoints didn't have to implement a whole actor interface, but I don't know how feasible that actually is.

The biggest remaining piece of work is probably documentation, both for the usage of the REST plugin and the plugin API of the REST endpoints.

I also gave this a quick spin locally, and it looks like it's working fine. Once we addressed the remaining comments I think we can quickly approve and merge this. I'm also really interesting in seeing how this integrates with #2574.

plugins/rest/src/server_command.cpp
dominiklohmann: This is fixed for at least basic types, but there's still an outstanding TODO for lists. Did you want to address that still, or is that waiting for the story that is about implementing "parse this string as a data with this type information" more generically?

plugins/rest/CMakeLists.txt
dominiklohmann: I cannot reproduce this locally, but looking at the snippet where we link against it there are some other cleanups we can do:

https://github.com/tenzir/vast/blob/e42a9652e50833c5ba53bbd34d6ff12892a1cfa4/libvast/CMakeLists.txt#L432-L443

We actually require {fmt} >= 7 afaict, so maybe it's got something to do with that?

plugins/rest/src/server_command.cpp
lava: It's a nice-to-have, but all parameters are currently correctly documented as query-parameters so it's certainly not a blocker.

libvast/test/rest_api.cpp
lava: Sadly, the comment is incorrect and it's JSON data instead of YAML.

plugins/rest/src/authenticator.cpp
dominiklohmann: Fair point. I think we already link against OpenSSL via CAF, so maybe we can use that for it?

libvast/include/vast/type.hpp
lava: Hm, `optional<record_type>` is the semantically correct type so if we use `type` then we'd have to downcast and assert that it's a record on the call site. Could we instead allow empty `record_type` to make it semi-regular? Or alternatively, we could just create an `inspect(Inspector&, caf::optional<record_type>&)` overload to make sure it only works if the record type is correctly wrapped.

dominiklohmann: We cannot make concrete types semi-regular, that would break quite a few assumptions that we can currently make about record types.

I'm fine with adding an inspect overload like this:

```cpp
template <class Inspector, concrete_type Type>
auto inspect(Inspector& f, std::optional<Type>& x)
    -> typename Inspector::result_type {
  return f(...);
}
```

(Actually not sure if that works generically with concrete type because of the usual ADL wonkiness, but if it doesn't then maybe just for `record_type` as a friend.)

libvast/include/vast/plugin.hpp
lava: From the plugin server PoV it would be easy to have the plugin provide a function like 

```
class rest_endpoint_plugin {
public:
  using handler = std::function<void(http_request&&)>;

  virtual handler get_handler() const = 0;
};
```

and to call that when handling the request.

The problem is that on the plugin developer side it's pretty difficult to provide any useful functionality unless you can interact with the actor system.

plugins/rest/include/rest/restinio_response.hpp
lava: Yes, that comment is technically outdated; restinio has some optimizations where doesn't make a copy when given a string_view so that it becomes cheap to return static string literals, but that breaks when passing a regular string as `string_view`.

plugins/rest/src/plugin.cpp
lava: Yes, makes sense to me

plugins/rest/include/rest/restinio_response.hpp
lava: The user-facing comments are in `vast::http_response`, the `restinio_response` is an implementation detail of the rest plugin.

libvast/native-plugins/rest_api/export.cpp
lava: Either way makes sense to me; I mainly split them out of the rest plugin so I don't accidentally rely on internals of the REST plugin in the implementation, which would lead to future problems for closed-source endpoints.

plugins/rest/src/authenticator.cpp
lava: I think this one is fine leave for now until we also add some ways to list and display the generated tokens.

plugins/rest/src/server_command.cpp
lava: I'll probably create a separate story for supporting lists, so we can implement it when we actually want to add the first endpoint with a list parameter.

plugins/rest/src/authenticator.cpp
lava: Good point, recent version of OpenSSL use `getentropy()` on Linux anyways to implement `RANDOM_bytes()`

libvast/include/vast/plugin.hpp
dominiklohmann: I see. Maybe this is something that we can abstract away in the future by making it easier to interact with the actor system. Feel free to resolve this for now.

plugins/rest/src/plugin.cpp
mavam: Documentation on vast.io is a separate story. The README will just point there.

(IMO this can be resolved.)

plugins/rest/src/generate_token_command.cpp
lava: Ideally we'd have a separate component called `rest` that displays the actually interesting state (ie. listen address and mode), but I'm not sure that's implementable with the current plugin framework.

libvast/include/vast/system/actors.hpp
```diff
@@ -419,6 +419,9 @@ using exporter_actor = typed_actor_fwd<
   // Conform to the protocol of the STATUS CLIENT actor.
   ::extend_with<status_client_actor>::unwrap;
 
+using rest_handler_actor = system::typed_actor_fwd<
+  caf::reacts_to<atom::http_request, uint64_t, http_request>>::unwrap;
```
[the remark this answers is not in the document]
lava: I agree it makes sense to create strong typedef's for all the various id's that we're sending as messages, but currently we use uint64/uint32 for everything else as well so that's probably a separate refactoring.

plugins/rest/src/server_command.cpp
lava: I've removed the html part for now, we can add it back in a later commit.

libvast/native-plugins/rest_api/export.cpp
lava: I did the renaming, I didn't do the reshuffling for now since at least then we have all native plugins in one place.

libvast/native-plugins/rest_api/export.cpp
lava: I know, but I still don't like the idea of adding more work to fool a tool into behaving as if it was turned off, instead of just turning it off :D

plugins/web/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# Changelog
+
+This changelog documents all notable changes to the PCAP plugin for VAST.
```
netantho, line 3: Typo: **web** plugin?

lava: Of course, thanks :D

dominiklohmann (approved):
We peer-reviewed this. Approving this as-is modulo the remaining comments. We have a few things left to do in follow-ups.

PR #2568, opened 2022-09-08
Remove the shutdown grace period

dominiklohmann:
This removes the `vast.shutdown-grace-period` functionality and its underlying mechanism. This has two reasons:

1. The mechanism was broken, and in case of yet unpersisted data taking longer than the grace period to persist on shutdown caused both data loss and an infinite hang.

2. It is the purpose of a service manager to issue hard kills to hanging process, not of the hanging process itself.

Ultimately we decided that the internal hard kill causes more issues than it helps solves, so now the new grace period is infinite: VAST simply waits for as long as it takes to shut down properly.

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/src/system/terminator.cpp
+++ b/libvast/src/system/terminator.cpp
@@ -22,11 +22,8 @@
 namespace vast::system {
 
 template <class Policy>
-caf::behavior terminator(caf::stateful_actor<terminator_state>* self,
-                         std::chrono::milliseconds grace_period,
-                         std::chrono::milliseconds kill_timeout) {
-  VAST_TRACE_SCOPE("terminator {} {} {}", VAST_ARG(self->id()),
-                   VAST_ARG(grace_period), VAST_ARG(kill_timeout));
+caf::behavior terminator(caf::stateful_actor<terminator_state>* self) {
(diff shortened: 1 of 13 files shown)


libvast/src/system/terminator.cpp
```diff
@@ -22,11 +22,8 @@
 namespace vast::system {
 
 template <class Policy>
-caf::behavior terminator(caf::stateful_actor<terminator_state>* self,
-                         std::chrono::milliseconds grace_period,
-                         std::chrono::milliseconds kill_timeout) {
-  VAST_TRACE_SCOPE("terminator {} {} {}", VAST_ARG(self->id()),
-                   VAST_ARG(grace_period), VAST_ARG(kill_timeout));
+caf::behavior terminator(caf::stateful_actor<terminator_state>* self) {
+  VAST_TRACE_SCOPE("terminator {}", VAST_ARG(self->id()));
```
mavam, line 26: Is this helpful in practice?

dominiklohmann: Yes.

mavam (approved):
I inspected the diff. Overall, the changes seem reasonable. This is hard to test, but since it's only removal of functionality and potential side effects, VAST will be easier to test moving forward.

PR #2518, opened 2022-08-22
Python API: Add status and count command, as well as a utility to format IPs in dataframe

netantho:
This modifies and expands the Python API:
- Rename query methods to export to stick with the CLI command names  and make them public
- Add the `status` and `count` commands
- Add a utility to format IPs in dataframes, taken from https://github.com/tenzir/vast/blob/9acb47c27a69710bc9c3cabdc61e956dafff65d5/examples/jupyter/vast-connections.ipynb

### :memo: Reviewer Checklist

Review this pull request by ensuring the following items:

- [ ] All user-facing changes have changelog entries
- [ ] User-facing changes are reflected on [vast.io](https://github.com/tenzir/vast/tree/master/web)

PR #2570, opened 2022-09-09
More Python experiments

mavam:
This PR rotates around quite a bit of Python code, as this is still quite in progress.

The objective of this PR is to stabilize the VAST part of the Python bindings, and to factor out the fabric work into a separate piece. The goal is to integrate #2518 right afterwards, along with some Arrow extension type wrappers.

From a packaging perspective, I'd like to remove misp-stix as mandatory dependency, as this is the only thing that holds up publishing the package.

This PR also adds several tests for core infrastructure/tooling. As the fabric API matures, the tests will extend.

For VAST itself, we also need unit tests. The best way doing so is probably using the dockerized VAST version (cf. PyVAST), but it warrants a small discussion.

### :memo: Reviewer Checklist

- [x] Refactor directory layout to create a split of `vast.py` and `fabric` inside the `vast` package directory
- [x] Configure poetry to make `fabric` an extra
- [x] The `VAST` class is a reasonable interface to interact with VAST
- [x] Clean up unit tests
- [x] misp-stix is an optional dependency
- [x] Publish package on Test PyPI


mavam:
Publishing to Test PyPI worked:

```
❯ poetry build
Building vast (0.1.0)
  - Building sdist
  - Built vast-0.1.0.tar.gz
  - Building wheel
  - Built vast-0.1.0-py3-none-any.whl

❯ poetry publish -r test-pypi

Publishing vast (0.1.0) to test-pypi
 - Uploading vast-0.1.0-py3-none-any.whl 100%
 - Uploading vast-0.1.0.tar.gz 100%
```

dominiklohmann (approved):
We went over this in a call. Want to merge now rather than later.

PR #2571, opened 2022-09-09
Validate indexer spawn arguments before spawning

dominiklohmann:
This changes the `{active,passive}_indexer` actors to require their arguments to be valid instead of shutting down if they are spawned with invalid arguments.

We suspect this to be the cause for some error messages we've seen in our testbed in an edge case.

### :memo: Reviewer Checklist


libvast/src/system/indexer.cpp
```diff
@@ -30,15 +30,11 @@ namespace vast::system {
 
 active_indexer_actor::behavior_type
 active_indexer(active_indexer_actor::stateful_pointer<indexer_state> self,
-               const std::string& name, value_index_ptr value_index) {
+               const std::string& name, value_index_ptr index) {
+  VAST_ASSERT_CHEAP(!name.empty());
+  VAST_ASSERT_CHEAP(index);
```
lava, line 35: I think it would be good to ensure this in the type system somehow, in an earlier comment you mentioned the idea of a `active_indexer::verified_spawn_args` class, or if that is too much effort for one PR we could go with a `nonnull<value_index_ptr>`.

libvast/src/system/indexer.cpp
```diff
@@ -108,16 +104,11 @@ active_indexer(active_indexer_actor::stateful_pointer<indexer_state> self,
...
-  if (!idx) {
-    VAST_ERROR("{} got invalid value index pointer", *self);
-    self->quit(caf::make_error(ec::end_of_input, "invalid value index "
-                                                 "pointer"));
-    return indexer_actor::behavior_type::make_empty_behavior();
-  }
-  self->state.name = fmt::format("indexer-{}", idx->type());
+                uuid partition_id, value_index_ptr index) {
+  VAST_ASSERT_CHEAP(index);
+  self->state.name = fmt::format("indexer-{}", index->type());
```
lava, line 109: If we're touching this anyways, does it make sense to proactively switch to a static name to reduce the scope of the CAF 0.18 refactoring? Or would it just create additional merge conflicts? (cc/ @patszt )

lava (commented):
Just a first glance

libvast/src/system/indexer.cpp
dominiklohmann: I did not do this because of the extra code it requires, but I probably should because it's a cleaner abstraction. I'll push something in a bit.

libvast/src/system/indexer.cpp
patszt: I think either change won;t impact refactoring

libvast/src/system/indexer.cpp
dominiklohmann: After not having the time to get to this for four days, I propose that we postpone this refactoring for a later time. What we've got here solves the problem at hand. I just don't have the bandwidth currently to work on refactorings like this.

libvast/src/system/indexer.cpp
dominiklohmann: I'll leave this as-is then.

libvast/src/system/indexer.cpp
dominiklohmann: @lava ping

lava (approved):
As discussed in the previous review comments, this should solve the immediate issue we see. We can revisit a general `checked_spawn_args` refactoring later when there's a good opportunity.

PR #2576, opened 2022-09-14
Publicly accessible cloud demo HTTP services

rdettai:
Add a cloudflare plugin that sets up a tunnel. Only MISP is configured for now, but this should scale to any number of apps.

TODO: 
- [x] check if we cannot setup the Access Application protection from terraform

### :memo: Reviewer Checklist

Additional notes:
- This uses the Cloudflare Terraform provider, which is quite poorly documented but almost feature complete. Unfortunately, Cloudflare does some default resource creations behind the scenes that don't play very well with Terraform (or any API programming).
- I had to add a security group to MISP to allow access to its web UI port.


rdettai:
[Terraform](https://github.com/tenzir/vast/actions/workflows/terraform.yaml?query=branch%3Astory%2Fsc-37134%2Fcloudflare)

dispanser (approved):
I was able to deploy this and see a MISP instance via some cloud flare URI, so that's great.

One minor remark inline. I had some difficulties to configure the cloudflare domain / user etc (we had to pair). I'm not knowledgable in cloudflare, so I might not be a good benchmark, but maybe a few extra lines of walk-through would be handy. Given the non-intuitive way the web site is written, maybe even a screenshot or two

PR #2511, opened 2022-08-19
RFC-001: Composable Pipelines

mavam:
This PR adds the first RFC. This is an experiment. We want to gather public
feedback for ideas and want to try a PR-based discussion in a Markdown document.

PR #2577, opened 2022-09-14
RFC-003: Remote Pipelines

mavam:
This PR augments [RFC-001](https://github.com/tenzir/vast/pull/2511) with remote pipelines.

### DIFF
--- a/rfc/003-remote-pipelines/README.md
+++ b/rfc/003-remote-pipelines/README.md
@@ -183,16 +183,6 @@ follows:
 vast mutate lake1 'where 10.0.0.0/8 | put orig.h resp.h'
 ```
 
-
-
-
-
----
-
-TODO: here be dragons; the following is verbatim copied from RFC-001 and the
-content still needs to be adapted.
-
-
 ## Implementation
 
 ### Case study: remote matcher
--- a/rfc/003-remote-pipelines/README.md
+++ b/rfc/003-remote-pipelines/README.md
@@ -1,8 +1,8 @@
 # Remote Pipelines
 
-- **Status**: WIP
+- **Status**: Obsolete
 - **Created**: Sep 14, 2022
-- **ETA**: Oct 14, 2022
+- **Dismissed**: Feb 08, 2023
 - **Authors**:
   - [Matthias Vallentin](https://github.com/mavam)
 - **Contributors**:
--- a/rfc/003-remote-pipelines/README.md
+++ b/rfc/003-remote-pipelines/README.md
@@ -9,6 +9,7 @@
   - [Dominik Lohmann](https://github.com/dominiklohmann)
   - [Rémi Dettai](https://github.com/rdettai)
   - [Tobias Mayer](https://github.com/tobim)
+  - [Anthony Verez](https://github.com/netantho)
 - **Discussion**: [PR #2577](https://github.com/tenzir/vast/pull/2577)
 
 ## Overview
--- a/rfc/003-remote-pipelines/README.md
+++ b/rfc/003-remote-pipelines/README.md
@@ -5,24 +5,95 @@
 - **ETA**: Oct 14, 2022
 - **Authors**:
   - [Matthias Vallentin](https://github.com/mavam)
+- **Contributors**:
+  - [Dominik Lohmann](https://github.com/dominiklohmann)
+  - [Rémi Dettai](https://github.com/rdettai)
+  - [Tobias Mayer](https://github.com/tobim)
 - **Discussion**: [PR #2577](https://github.com/tenzir/vast/pull/2577)
 
 ## Overview
 
-This proposal enhances the local pipeline execution with a remote component,
-making it possible to connect pipeline across process boundaries.
+This proposal enhances [local pipeline execution][rfc-001] with a remote
+component, making it possible to connect pipeline across process boundaries.
+
+[rfc-001]: https://github.com/tenzir/vast/pull/2511
 
 ## Problem Statement
 
-As of RFC-001, VAST pipelines can execute locally. But a one-shot pipeline
-execution is not sufficient. After this building block, natural questions arise:
+As of [RFC-001][rfc-001], VAST pipelines can execute locally. This offers
+control over the execution at the location of a specific VAST instance. The
+pipeline executes in situ, with the `load` and `store` operators offering
+customization points for data source and sink.
 
-- How can I connect a local and remote pipeline?
-- How can I manage (add/update/remove) a set of continuously running pipelines?
+While local pipelines work for ad-hoc data crunching or investigative queries,
+they do not support composition of continuously running pipelines. Consider the
+example of a live pipeline that connects to a data source (e.g., Suricata,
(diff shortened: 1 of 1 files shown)


mavam:
[sc-37588]

rfc/003-remote-pipelines/README.md
```diff
@@ -0,0 +1,206 @@
+# Remote Pipelines
+
+- **Status**: WIP
+- **Created**: Sep 14, 2022
+- **ETA**: Oct 14, 2022
```
dominiklohmann, line 5: I think it doesn't make much sense to start this discussion window before we at least have the basics of RFC-001 implemented. This will be so much easier to discuss once we've got the implementation experience.

mavam: I'll update the ETA once RFC-001 is merged.

rfc/003-remote-pipelines/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+```bash
+vast spawn pipe1 'from - | read zeek | where orig_bytes > 1 GiB'
+vast expose pipe1 tcp://1.2.3.4:5555 
+```
+
+Attach to `pipe1` to read from it locally:
+
+```bash
+vast exec 'pull pipe1 | where resp_bytes > 1 KiB'
```
tobim, line 60: This treats named pipelines as a second class construct.
How about getting rid of the `pull` and making name assignment part of the language instead of adding an extra command:

```
vast exec '
  let
    # local assignment 
    pipe1 = from - | read zeek | where orig_bytes > 1 GiB
  in 
    # exposing to the global namespace
    publish pipe1'

vast exec 'pipe1 | where resp_bytes > 1 KiB | write to lakeX as potential_exfiltration'

vast exec 'from /tmp/conn.log | pipe1'
```

PS: the `read zeek` should not be a thing, let's get rid of that.

tobim (changes requested):
The proposed design conflates name assignment with remoting.

As an analyst I don't want to have to know details about the topology of the VAST cluster.
I would like to propose that we keep the representations of "logical" and "physical" query plans separate for now, because an IR that can cover the needs of both seems to add undesired complexity.

rfc/003-remote-pipelines/README.md
```diff
@@ -0,0 +1,206 @@
...
+execution is not sufficient. After this building block, natural questions arise:
+
+- How can I connect a local and remote pipeline?
+- How can I manage (add/update/remove) a set of continuously running pipelines?
+- How do pipelines interact with VAST server nodes? Specifically, how can I send
+  data into a VAST node, and how can I query data from a VAST node?
+
+This proposal seeks an answer to these questions.
+
+## Solution Proposal
```
rdettai, line 27: I think we should create 2 distinct named entities:
- "named pipelines": a chain of operators that will execute provided that data will flow into its first operator
- "named dataflows" (not sure about the term, but we can use it here as it isn't too overloaded yet): a specific point in the graph of data flows to which you can attach a pipelines input or ouput

In the plumbing analogy that we, Mario loving data engineers, love: "pipelines" are pipes, "dataflows" are logical sources/sinks

This decoupling has many benefits:
- you can combine the output of M pipelines and subscribe to that with N other pipelines
- these named dataflows are very similar to the "topic" concept in the Fabric/Threatbus, it will provide good abstraction points for domain specific abstract "streams of interest". For instance, you could have a default `ocsf_finding` "named dataflow". Every app wanting to contribute to it can create a pipeline that pushes to it and every app wanting to subscribe to all alerts will now that this is the best place to start.

The example bellow would look something like this:
```bash
vast spawn pipe1 'from - | read zeek | where orig_bytes > 1 GiB | push zeek_large'
vast exec 'pull zeek_large | where resp_bytes > 1 KiB'
```

rfc/003-remote-pipelines/README.md
```diff
@@ -0,0 +1,206 @@
...
+
+Assuming an existing mechanism to bind a pipeline to name, the new operators
+allow for interconnecting pipelines as follows.
+
+Create a named pipeline `pipe1` at a VAST node and expose in the global control
+plane:
+
+```bash
+vast spawn pipe1 'from - | read zeek | where orig_bytes > 1 GiB'
+vast expose pipe1 tcp://1.2.3.4:5555
```
rdettai, line 54: Is this a leftover from a previous design? I thought from RFC-001 that `from`/`read` where now relegated to the realm of the unexposed physical operators.

Also I am not sure what `vast expose` does here. Is it attaching to the input or the ouput? 

Finally, I believe that the address (`tcp://...`) should not be an argument to the `expose` command but rather something that provided in the response

rfc/003-remote-pipelines/README.md
rdettai: Interesting point. From a parser perspective `vast exec 'pull pipe1 | ...` seems easier to maintain than `vast exec 'pipe1 | ...`, because in the second scenario, your pipeline/dataflow (see comment above) names can conflict with operator names.

rfc/003-remote-pipelines/README.md
mavam: I like the additional concept of "dataflow" (or just "flow") as abstraction for pipeline composition.

You could argue that once you have a non-void operator, you can turn that into a flow by giving it a name, and optionally binding it to remote endpoint. 

For example:

```
load zeek://1.2.3.4 | where #type == "zeek.notice"
```

This pipeline has type `<Void, Arrow>`. In step one, you would give it a name:

```
vast spawn zeek-alerts 'load zeek://1.2.3.4 | where #type == "zeek.notice"'
```

If you want to use it outside of your current VAST node, you would have to expose it in some fashion, e.g., to a service registry:

```
vast publish zeek-alerts --endpoint 1.2.3.4:59232
```

mavam:
@rdettai and I had another discussion on this. We tried to avoid `push` and `pull` as explicit operators (at least from the user perspective) and rather thought about exposing pipelines as a whole through a control plane approach, which is basically naming à la DNS for us right now.

image

We thought the term *flow* is a good one for a pipeline that's exposed to the control plane.
The primary reason for not having `push` and `pull` would be the flexibility to reconfigure flows on the fly. We understand the implementation may actually have such operators, but that will come second. For now, we are mostly concerned with high-level UX and composability.

I'll update the RFC once we converged a little further.

dominiklohmann:
@mavam I think we can close this as not accepted. As we're inching closer to having RFC-001 implemented it has become clear that remoteness of pipelines is an implementation detail that the user should not need to set. Instead, all transfer between processes required within a pipeline is _between_ operators as needed, and operators only say which node they want to run on. A pipeline can then either be started locally or remotely, but that doesn't warrant an RFC, as it's not really a complicated matter.

mavam:
Agreed. I've pushed a commit to mark the proposal as obsolete.

Should we just resolve all outstanding comments and then merge?

dominiklohmann:
> Should we just resolve all outstanding comments and then merge?

Yes, that makes sense.

dominiklohmann (approved):
Approving as rejected.

PR #2582, opened 2022-09-17
Move the version string into a central JSON file

tobim:
CMake is not the only tool that is interested in the version string, so we move it into a format that is widely supported.
Note that the value of `vast-version-fallback` is not prefixed with 'v', because it is easier to add where needed than to remove when undesired.

In order to allow this change, the minimum required CMake version is bumped from 3.18 to 3.19.

### :memo: Reviewer Checklist

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -65,10 +65,14 @@ jobs:
     steps:
       - name: Install Dependencies
         run: |
+          echo 'deb http://deb.debian.org/debian bullseye-backports main' \
+            > /etc/apt/sources.list.d/backports.list
           apt-get update
           apt-get -y install \
             build-essential \
             ca-certificates \
+            cmake/bullseye-backports \
+            cmake-data/bullseye-backports \
             flatbuffers-compiler-dev \
             g++-10 \
             gcc-10 \
@@ -91,8 +95,6 @@ jobs:
             python3-pip \
             python3-venv \
             wget
-          python3 -m pip install --upgrade pip
-          python3 -m pip install --upgrade cmake
       - name: Checkout
         uses: actions/checkout@v3
         with:
@@ -234,12 +236,16 @@ jobs:
     steps:
       - name: Install Dependencies
         run: |
+          echo 'deb http://deb.debian.org/debian bullseye-backports main' \
+            > /etc/apt/sources.list.d/backports.list
           apt-get update
           apt-get -y install \
             apt-transport-https \
             build-essential \
             ca-certificates \
             ccache \
+            cmake/bullseye-backports \
+            cmake-data/bullseye-backports \
             curl \
             flatbuffers-compiler-dev \
             g++-10 \
@@ -271,10 +277,7 @@ jobs:
           apt-get -y install ./apache-arrow-apt-source-latest-$(lsb_release --codename --short).deb && \
           apt-get update
           apt-get -y install libarrow-dev libparquet-dev
-
-          # install CMake from pip -- we need at least 3.17 in CI for CCache
           python3 -m pip install --upgrade pip
-          python3 -m pip install --upgrade cmake
 
           cmake --version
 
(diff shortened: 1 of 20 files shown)


dominiklohmann:
@tobim I'm fine with the idea generally. Can you re-request the review once all the CI workflows ran through, or should I be taking a look at this now already?

dominiklohmann:
@tobim I'm fine with the idea generally. Can you re-request the review once all the CI workflows ran through, or should I be taking a look at this now already?

tobim:
CI is green now.

scripts/prepare-release
```diff
@@ -47,10 +45,9 @@ fi
 git fetch origin master
 git switch -C "topic/release-${new_version}" origin/master
 
-# Change the fallback version.
-perl -i -pe "s/${last_rc_version}/${new_version}/g" \
-  "${source_dir}/cmake/VASTVersionFallback.cmake" \
-  "${source_dir}/nix/vast/default.nix"
+# Update the fallback version.
+version_json="$(jq --arg v "${new_version#v}" '."vast-version-fallback" = $v' version.json)"
+echo -E "${version_json}" > version.json)
```
dominiklohmann, line 50: Is there a reason for not just piping the output of jq to version.json here?

version.json
```diff
@@ -0,0 +1,17 @@
...
+    "If you modify it by hand you should also run the `update-changelog`",
+    "target and add the changes to CHANGELOG.md and the changelog directory",
+    "to the same commit."
+  ],
+  "vast-version-fallback": "2.3.0",
+  "vast-partition-version_COMMENT": [
+    "The partition version. This number must be bumped alongside the release",
+    "version for releases that contain major format changes to the on-disk",
+    "layout of VAST's partitions."
+  ],
```
dominiklohmann, line 15: This is very ugly compared to what we had previously. Do you know if we can make this either YAML or JSON5 so we can actually use comments?

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.18...3.24 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.19...3.24 FATAL_ERROR)
```
dominiklohmann, line 1: This needs to be documented
- in the docs: https://vast.io/docs/setup-vast/build
- in a changelog entry

After that, feel free to check the reviewer checklist.

.github/workflows/vast.yaml
```diff
@@ -271,10 +277,7 @@ jobs:
           apt-get -y install ./apache-arrow-apt-source-latest-$(lsb_release --codename --short).deb && \
           apt-get update
           apt-get -y install libarrow-dev libparquet-dev
-
-          # install CMake from pip -- we need at least 3.17 in CI for CCache
           python3 -m pip install --upgrade pip
```
dominiklohmann, line 280: Is this still required now?

dominiklohmann (approved):
Liking this in general, just some smaller questions.

scripts/prepare-release
tobim: I assume you mean redirecting the output with `>`. The problem is that this operator clears the contents of the file it writes to, and that can happen before `jq` gets around to read the input. That is why we write it to an intermediate variable first.

version.json
tobim: :cry: I was actually quite proud of this solution when I came up with it. This file should really be JSON as that is much more widely supported than any of the alternatives.

.github/workflows/vast.yaml
tobim: I'm not sure if it is strictly required, but we use pip to install dependencies for the integration tests.

dominiklohmann: That is always already upgrading pip (and also running inside a venv), so I think this can go.

scripts/prepare-release
dominiklohmann: Instead of passing version.json as a positional argument to jq you can also pass it via stdin, i.e., `jq ... < version.json > version.json`, which should guarantee the ordering we want to have.

tobim: Can you provide a source for that?

dominiklohmann: Apparently that's a Zsh extension I was using for so long, I thought of it being standard.

You could still use `sponge` from moreutils, but I don't know how widely that is available.

Feel free to resolve.

tobim: I'd rather avoid depending on moreutils, it is usually not part of a base install.

PR #2584, opened 2022-09-19
Common helper class implementation for new inspector api and common function for extracting std::vector from configs 

patszt:
CAF 0.18.6 introduces new inspector API (https://actor-framework.readthedocs.io/en/stable/TypeInspection.html)
The common_inspection file introduces classes that allow the VAST inspectors to abide to this API.
The real example of the new inspector API can be seen in CAF repo (https://github.com/actor-framework/actor-framework/blob/cf0af7de5c4bbd93fc5e25e220a03cacc4c98953/libcaf_core/caf/uri.hpp#L242)

In the 0.18.6 we cannot also use caf::get_if<std::vector<std::string>>. Every caf::get_if/get_as etc has a static_check.
The passed T into get_if<T> must be a config value_type. The only std::vector that is a config value type is:
using list = std::vector<config_value>;
This Is why i have created a helper function for extracting std::vector<T>

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/include/vast/detail/inspection_common.hpp
+++ b/libvast/include/vast/detail/inspection_common.hpp
@@ -14,28 +14,24 @@
 namespace vast::detail {
 
 template <class Inspector, class AfterInspectionCallback>
-class inspection_object_with_after_inspection_callback {
-public:
-  inspection_object_with_after_inspection_callback(
-    Inspector& inspector, AfterInspectionCallback callback)
-    : inspector_{inspector}, callback_{std::move(callback)} {
-  }
+class inspection_object_with_after_inspection_callback;
 
-  template <class... Fields>
-  constexpr bool fields(Fields&&... fields) {
-    return (fields(inspector_) && ...) && callback_();
-  }
-
-  constexpr inspection_object_with_after_inspection_callback&
-  pretty_name(std::string_view) noexcept {
-    return *this;
-  }
-
-private:
-  Inspector& inspector_;
-  AfterInspectionCallback callback_;
-};
-
+/// Common class used by VAST inspectors as a return type for object method.
+/// The CAF inspector API requires an inspector to have object method.
+/// This method should return an object which guides the inspection procedure.
+/// Example of usage:
+/// bool inspect(Inspector& f, some_type& x)
+///{
+///   auto cb = []{};
+///   return f.object(x).
+///     on_load(cb).
+///     fields(f.field("name", x.name), f.field("value", x.value));
+///}
+/// Such code should result in inspection of x.name by the inspector. If the
+/// inspection succeeds then it should proceed to inspection of x.value. If all
+/// provided fields succeed then callback (cb variable) should be called. The
+/// provided string literals are optionally used by human readable CAF
+/// inspectors. Current VAST inspectors have no usage for them.
 template <class Inspector>
 class inspection_object {
 public:
@@ -67,6 +63,31 @@ public:
 
 private:
   Inspector& inspector_;
+};
+
+/// Enhanced inspection_object that is responsible for calling callback after
+/// inspection.
+template <class Inspector, class AfterInspectionCallback>
+class inspection_object_with_after_inspection_callback {
+public:
+  inspection_object_with_after_inspection_callback(
+    Inspector& inspector, AfterInspectionCallback callback)
+    : inspector_{inspector}, callback_{std::move(callback)} {
+  }
+
+  template <class... Fields>
+  constexpr bool fields(Fields&&... fields) {
+    return (fields(inspector_) && ...) && callback_();
+  }
+
+  constexpr inspection_object_with_after_inspection_callback&
+  pretty_name(std::string_view) noexcept {
+    return *this;
+  }
+
+private:
+  Inspector& inspector_;
+  AfterInspectionCallback callback_;
 };
 
 template <class T>
--- a/libvast/include/vast/detail/inspection_common.hpp
+++ b/libvast/include/vast/detail/inspection_common.hpp
@@ -1,0 +1,88 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2022 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#pragma once
+
+#include <string_view>
+#include <utility>
+
+namespace vast::detail {
+
+template <class Inspector, class AfterInspectionCallback>
+class inspection_object_with_after_inspection_callback {
+public:
+  inspection_object_with_after_inspection_callback(
+    Inspector& inspector, AfterInspectionCallback callback)
+    : inspector_{inspector}, callback_{std::move(callback)} {
+  }
+
+  template <class... Fields>
+  constexpr bool fields(Fields&&... fields) {
+    return (fields(inspector_) && ...) && callback_();
+  }
+
+  constexpr inspection_object_with_after_inspection_callback&
+  pretty_name(std::string_view) noexcept {
+    return *this;
+  }
+
+private:
+  Inspector& inspector_;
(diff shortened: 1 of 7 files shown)


libvast/include/vast/detail/settings.hpp
```diff
@@ -27,4 +31,58 @@ void merge_settings(const caf::settings& src, caf::settings& dst,
...
+/// @tparam T expected type of the list values.
+/// @param cfg_value value which should be a list type of Ts.
+/// @param error_message_on_list_extraction_failure error context of the error
+/// in case the cfg_value is not a list type.
+/// @return all list values extracted as Ts or error in 2 cases: the cfg_value
+/// is not of a list type, any value in the list is not of type T.
+template <class T>
+caf::expected<std::vector<T>> unpack_config_list_to_vector(
+  const caf::config_value& cfg_value,
+  std::string_view error_message_on_list_extraction_failure) {
```
tobim, line 45: This API looks rather awkward to me. Why do we need to pass an error context in instead of composing it when it bubbles up the stack?

libvast/include/vast/detail/settings.hpp
```diff
@@ -27,4 +31,58 @@ void merge_settings(const caf::settings& src, caf::settings& dst,
...
+    return caf::make_error(ec::invalid_configuration,
+                           error_message_on_list_extraction_failure.data());
+
+  std::vector<T> ret;
+  ret.reserve(list->size());
+  for (const auto& e : *list) {
+    const auto* val = caf::get_if<T>(&e);
+    if (!val)
+      return caf::make_error(ec::invalid_configuration,
+                             "Unexpected type in unpack_config_list_to_vector");
```
tobim, line 57: This should rather say: "Type mismatch while unpacking config list: expected {caf::pretty_type_name<T>}, got {e}".

libvast/include/vast/detail/inspection_common.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+#pragma once
+
+#include <string_view>
+#include <utility>
+
+namespace vast::detail {
+
+template <class Inspector, class AfterInspectionCallback>
+class inspection_object_with_after_inspection_callback {
```
tobim, line 17: The types defined in this file need to be documented. As it is I don't understand when I'm supposed to use them.

libvast/include/vast/detail/settings.hpp
patszt: Indeed it looks akward

libvast/include/vast/detail/inspection_common.hpp
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+private:
+  Inspector& inspector_;
+};
+
+/// Enhanced inspection_object that is responsible for calling callback after
+/// inspection.
+template <class Inspector, class AfterInspectionCallback>
+class inspection_object_with_after_inspection_callback {
```
tobim, line 71: This class name is quite the mouth-full. Can we unify the two and default initialize `AfterInspectionCallback` to a noop?

tobim (changes requested):
This is already a lot more understandable now. Just a few more things regarding the `inpection_object`.

libvast/src/system/make_pipelines.cpp
```diff
@@ -133,10 +134,12 @@ make_pipelines(pipelines_location location, const caf::settings& settings) {
...
       return caf::make_error(ec::invalid_configuration, "pipeline name must "
                                                         "be a string");
-    auto events = caf::get_if<std::vector<std::string>>(&(*pipeline)["events"]);
-    if (!events)
-      return caf::make_error(ec::invalid_configuration,
-                             "pipeline event types must be a list of strings");
+    auto events = detail::unpack_config_list_to_vector<std::string>(
+      (*pipeline)["events"]);
+    if (!events) {
+      VAST_ERROR("Unable to extract events from pipeline config");
```
dominiklohmann, line 140: Please stick to the usual noun-verb-object order for log messages here and start in lowercase.

libvast/include/vast/detail/settings.hpp
```diff
@@ -27,4 +32,58 @@ void merge_settings(const caf::settings& src, caf::settings& dst,
...
+/// @param error_message_on_list_extraction_failure error context of the error
+/// in case the cfg_value is not a list type.
+/// @return all list values extracted as Ts or error in 2 cases: the cfg_value
+/// is not of a list type, any value in the list is not of type T.
+template <class T>
+caf::expected<std::vector<T>>
+unpack_config_list_to_vector(const caf::config_value& cfg_value) {
+  const auto* list = caf::get_if<caf::config_value::list>(&cfg_value);
+  if (!list)
+    return caf::make_error(ec::invalid_configuration, "Failed to extract "
```
dominiklohmann, line 48:
```suggestion
    return caf::make_error(ec::invalid_configuration, "failed to extract "
```

libvast/include/vast/detail/settings.hpp
```diff
@@ -27,4 +32,58 @@ void merge_settings(const caf::settings& src, caf::settings& dst,
...
+                                                      "config value as list");
+
+  std::vector<T> ret;
+  ret.reserve(list->size());
+  for (const auto& e : *list) {
+    const auto* val = caf::get_if<T>(&e);
+    if (!val)
+      return caf::make_error(
+        ec::invalid_configuration,
+        fmt::format("Type mismatch while unpacking config list: expected {}, "
```
dominiklohmann, line 58:
```suggestion
        fmt::format("type mismatch while unpacking config list: expected {}, "
```

libvast/include/vast/detail/settings.hpp
```diff
@@ -27,4 +32,58 @@ void merge_settings(const caf::settings& src, caf::settings& dst,
...
+template <class T>
+caf::expected<std::vector<T>>
+unpack_config_list_to_vector(const caf::actor_system_config& cfg,
+                             std::string_view cfg_list_key) {
+  const auto& content = caf::content(cfg);
+  const auto it = content.find(cfg_list_key);
+  if (it == content.cend())
+    return caf::make_error(
+      ec::invalid_configuration,
+      fmt::format("No key: {} found in actor system config", cfg_list_key));
```
dominiklohmann, line 85:
```suggestion
      fmt::format("failed to find key '{}' in configuration", cfg_list_key));
```

The user doesn't care about the actor system, and then there's the usual ordering.

dominiklohmann (commented):
I only looked at the log messages. I'll leave the remaining review to Tobias.

libvast/include/vast/detail/inspection_common.hpp
patszt: It started with being unified. I just thought that maybe a performance overhead for extra unused dummy member might be unwanted. In the end the compiler can optimize it and my worry is premature optimization. I could just maybe rename it to inspection_object_with_callback ?

tobim: > In the end the compiler might optimize it and my worry is premature optimization.

I would assume that. As long as the no-op's definition is visible it should be extremely easy for the compiler to skip the call.

patszt: Done

libvast/include/vast/detail/inspection_common.hpp
```diff
@@ -0,0 +1,96 @@
...
+///{
+///   auto cb = []{};
+///   return f.object(x).
+///     on_load(cb).
+///     fields(f.field("name", x.name), f.field("value", x.value));
+///}
+/// Such code should result in inspection of x.name by the inspector. If the
+/// inspection succeeds then it should proceed to inspect x.value. If all
+/// provided fields succeed then the callback `cb` should be called.
+template <class Inspector, class AfterInspectionCallback = NoOpCallback>
```
tobim, line 37:
```suggestion
template <class Inspector, class AfterInspectionCallback = decltype([] {
                             return true;
                           })>
```

You might argue that this is less readable and I agree, but returning `true` from a "NoOp" is pretty random and it makes it less likely to be useful anywhere else.

tobim (approved):
I'm happy with how this turned out!

PR #2586, opened 2022-09-19
Support building VAST with clang-15

dominiklohmann:
### :memo: Reviewer Checklist

### DIFF
--- a/libvast/CMakeLists.txt
+++ b/libvast/CMakeLists.txt
@@ -217,14 +217,16 @@ else ()
     add_library(CAF::core ALIAS libcaf_core_${_linkage_suffix})
     target_compile_features(libcaf_core_${_linkage_suffix} INTERFACE cxx_std_17)
     target_compile_options(
-      libcaf_core_${_linkage_suffix} PRIVATE -Wno-maybe-uninitialized
-                                             -Wno-unknown-warning-option)
+      libcaf_core_${_linkage_suffix}
+      PRIVATE -Wno-maybe-uninitialized -Wno-unqualified-std-cast-call
+              -Wno-unknown-warning-option)
     VASTSystemizeTarget(libcaf_io_${_linkage_suffix})
     set_target_properties(libcaf_io_${_linkage_suffix} PROPERTIES EXPORT_NAME
                                                                   io)
     target_compile_options(
-      libcaf_io_${_linkage_suffix} PRIVATE -Wno-maybe-uninitialized
-                                           -Wno-unknown-warning-option)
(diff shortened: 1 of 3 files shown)


libvast/src/format/test.cpp
```diff
@@ -176,7 +176,8 @@ struct randomizer {
   void operator()(const string_type&, std::string& str) {
     lcg gen{static_cast<lcg::result_type>(sample())};
     std::uniform_int_distribution<size_t> unif_size{0, 256};
-    std::uniform_int_distribution<char> unif_char{32, 126}; // Printable ASCII
+    std::uniform_int_distribution<signed char> unif_char{
```
dominiklohmann, line 179: Note for the reviewer: `char` is not a valid integer type, and older clang and gcc versions were less compliant to the standard here and just allowed it. `signed char`, however, is a valid integer type, so that just works as a portable replacement.

libvast/src/arrow_table_slice_builder.cpp
```diff
@@ -108,7 +108,8 @@ create_table_slice(const std::shared_ptr<arrow::RecordBatch>& record_batch,
   return result;
 }
 
-bool verify_record_batch(const arrow::RecordBatch& record_batch) {
+[[maybe_unused]] bool
+verify_record_batch(const arrow::RecordBatch& record_batch) {
```
dominiklohmann, line 112: Note to the reviewer: We only used this in an assertion below, and clang-15 now warns in builds without assertions that this function is unused.

libvast/CMakeLists.txt
```diff
@@ -217,14 +217,16 @@ else ()
...
+      PRIVATE -Wno-maybe-uninitialized -Wno-unqualified-std-cast-call
+              -Wno-unknown-warning-option)
     VASTSystemizeTarget(libcaf_io_${_linkage_suffix})
     set_target_properties(libcaf_io_${_linkage_suffix} PROPERTIES EXPORT_NAME
                                                                   io)
     target_compile_options(
-      libcaf_io_${_linkage_suffix} PRIVATE -Wno-maybe-uninitialized
-                                           -Wno-unknown-warning-option)
+      libcaf_io_${_linkage_suffix}
+      PRIVATE -Wno-maybe-uninitialized -Wno-unqualified-std-cast-call
```
dominiklohmann, line 228: Note to the reviewer: CAF has some unqualified calls to `std::move` via ADL, and that now triggers a warning (and rightfully so!), but for the bundled CAF it makes a lot more sense to just silence the warning.

tobim (approved):
Thanks for fixing these.

PR #2587, opened 2022-09-20
Make `feather` the default store-backend

dominiklohmann:
This changes the default `vast.store-backend` to `feather`.

~I filed this PR a bit earlier because I wanted to check whether there are any additional issues that CI finds that I don't see locally, so please don't review before this is out of draft mode.~ Reviewable now. I recommend going commit-by-commit.

### :memo: Reviewer Checklist


mavam:
Meta review comment: you may want to fix the typo in your last commit.

vast/vast.cpp
```diff
@@ -67,10 +67,15 @@ try_handle_deprecations(vast::system::default_configuration& cfg) {
...
     VAST_WARN("the 'vast.store-backend' option 'archive' is deprecated; "
-              "automatically using 'segment-store' instead");
-    caf::put(cfg.content, "vast.store-backend", "segment-store");
+              "automatically using 'feather' instead");
+    caf::put(cfg.content, "vast.store-backend", "feather");
+  } else if (caf::get_or(cfg, "vast.store-backend", "feather")
+             == "segment-store") {
+    VAST_WARN("the 'vast.store-backend' option 'segment-store' is deprecated; "
+              "automatically using 'feather' instead");
+    caf::put(cfg.content, "vast.store-backend", "feather");
```
lava, line 78: This may or may not need to be split off in a separate PR, depending on the results of ongoing discussion.

libvast/native-plugins/feather.cpp
```diff
@@ -172,8 +172,11 @@ class active_feather_store final : public active_store {
   }
 
   [[nodiscard]] detail::generator<table_slice> slices() const override {
-    for (const auto& slice : slices_)
-      co_yield slice;
+    // We need to make a copy of the slices here because the slices_ vector may
+    // get invalidated while we iterate over it.
```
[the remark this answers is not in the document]
dominiklohmann: This is possible because we iterate over the generator across multiple actor handlers because we want queries to be interruptible after every handled slice. There's no need for a mutex because it's only concurrency and not parallelism, but we cannot iterate over it while also adding to it.

vast/vast.cpp
dominiklohmann: Yeah, we'll likely vote on it tomorrow. In case of an undecided result I'll create a follow-up story for discussion.

libvast/native-plugins/feather.cpp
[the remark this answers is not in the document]
dominiklohmann: I double-checked this with @tobim, who agreed that the change is sound.

vast/vast.cpp
dominiklohmann: We discussed this between @tobim @mavam and myself today, and came to the agreement to drop the segment-store as-is in this PR alongside bumping the partition version with the next minor release. We consider the advantage of being able to drop the segment store with VAST v3.0 as greater than the potential for error. That resulted in a vote of 3-1 for making the removal.

PR #2592, opened 2022-09-21
RFC-001 minor amendments

rdettai:
As I was reading through RFC-001 to ensure the continuity with RFC-003, I noticed some minor glitches.

### :memo: Reviewer Checklist

### DIFF
--- a/rfc/001-composable-pipelines/README.md
+++ b/rfc/001-composable-pipelines/README.md
@@ -23,7 +23,7 @@ that *everything is a pipeline* and build a composable UX around this concept.
 Pipelines offer flexible data reshaping at [predefined customization
 points](https://vast.io/docs/use-vast/transform). For example, they can be
 triggered for specific events on import/export, or invoked by compaction.
-Howerver, it is impossible to run pipelines in an ad-hoc manner, for example as
+However, it is impossible to run pipelines in an ad-hoc manner, for example as
 a post-processing step after a query or for an ad-hoc ingest of a data set.
 
 This RFC proposes dynamic pipelines that do not require static definition in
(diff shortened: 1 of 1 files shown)


rfc/001-composable-pipelines/README.md
```diff
@@ -97,12 +97,6 @@ store<Arrow, Void>
 
 This results in a `<Void, Void>` pipeline that does not "leak".
 
-Assuming that there won't be other types at the logical level besides `Void` and
-`Arrow`, we can simplify the typing discussion to a binary model: either data
-flows (= `Arrow`) or not (= `Void`). This would be desirable as it reduces
-complexity and obviates the need of thinking about typing at the logical level.
-The first implementation will reveal whether this is feasible.
-
```
rdettai, line 105: The `Matcher` type is mentioned in this document, so we have more than 2 logical types

rfc/001-composable-pipelines/README.md
```diff
@@ -118,6 +112,11 @@ data.
 | `load`   | `Void`     | `Arrow`     | Read data from a provided location
 | `store`  | `Arrow`    | `Void`      | Write data into a provided location
 
+These operators take two arguments:
+
+1. **Carrier**: an URL like `s3://aws` or `-` for stdin
+2. **Format**: `json`, `csv`, `feather`, `parquet`, etc.
+
```
rdettai, line 119: This was specified further in the document, but is already necessary at this point in the reading.

mavam: First, I think we don't need to duplicate this information. If we move it up, I would delete it down below and reference it.

Second, I think it doesn't really matter here because we're not providing examples. We're not going into the specification of the operators, so I think it's not too important here. (Not a strong opinion, though.)

rfc/001-composable-pipelines/README.md
mavam: Duly noted.

rfc/001-composable-pipelines/README.md
rdettai: there is an example right below for which this introduction is useful:

> Another to look at this is that export FORMAT EXPR | op1 | op2 creates two pipelines:
> - `where EXPR | op1 | op2`
> - `store - FORMAT`

I wasn't sure about repeating it, I think you are right.

mavam (approved):

PR #2593, opened 2022-09-22
Restinio nix package integration

dispanser:
Scaffolding to enable building the web plugin with nix

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -25,6 +25,35 @@ in
   prev.xxHash.overrideAttrs (old: {
     patches = [ ./xxHash/static.patch ];
   });
+  http-parser = if !isStatic then prev.http-parser else
+    prev.http-parser.overrideAttrs (_ : {
+      postPatch = let
+        cMakeLists = prev.writeTextFile {
+          name = "http-parser-cmake";
+          text = ''
+            cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
+            project(http_parser)
+            include(GNUInstallDirs)
+            add_library(http_parser http_parser.c)
+            target_compile_options(http_parser PRIVATE -Wall -Wextra)
+            target_include_directories(http_parser PUBLIC .)
+            set_target_properties(http_parser PROPERTIES PUBLIC_HEADER http_parser.h)
+            install(
+              TARGETS http_parser
+              ARCHIVE DESTINATION "''${CMAKE_INSTALL_LIBDIR}"
+              LIBRARY DESTINATION "''${CMAKE_INSTALL_LIBDIR}"
+              RUNTIME DESTINATION "''${CMAKE_INSTALL_BINDIR}"
+              PUBLIC_HEADER DESTINATION "''${CMAKE_INSTALL_INCLUDEDIR}")
+          '';
+        };
+      in ''
+        cp ${cMakeLists} CMakeLists.txt
+      '';
+      nativeBuildInputs = [ prev.buildPackages.cmake ];
+      makeFlags = [];
+      buildFlags = [];
+      doCheck = false;
+  });
   restinio = final.callPackage ./restinio { };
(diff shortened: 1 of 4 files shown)


nix/restinio/default.nix
```diff
@@ -0,0 +1,47 @@
...
+, stdenv
+, fetchFromGitHub
+, cmake
+, http-parser
+, fmt
+}:
+
+let
+  pname = "restinio";
+  version = "v.0.6.14";
```
tobim, line 11: I think there is a newer version available by now.

nix/restinio/default.nix
```diff
@@ -0,0 +1,47 @@
...
+    "-DRESTINIO_TEST=OFF"
+    "-DRESTINIO_SAMPLE=OFF"
+    "-DRESTINIO_INSTALL_SAMPLES=OFF"
+    "-DRESTINIO_BENCH=OFF"
+    "-DRESTINIO_INSTALL_BENCHES=OFF"
+    "-DRESTINIO_FIND_DEPS=ON"
+    "-DRESTINIO_USE_EXTERNAL_HTTP_PARSER=ON"
+    "-DRESTINIO_ALLOW_SOBJECTIZER=OFF"
+  ];
+  buildInputs = [
```
tobim, line 35: Since this is an interface library the dependencies are always public.
```suggestion
  propagatedBuildInputs = [
```

nix/restinio/default.nix
```diff
@@ -0,0 +1,47 @@
+{ lib
+, stdenv
+, fetchFromGitHub
+, cmake
+, http-parser
+, fmt
```
tobim, line 6: That is what `spdlog` uses, mixing two versions would probably fail.
```suggestion
, fmt_8
```

nix/vast/default.nix
```diff
@@ -75,6 +78,9 @@ stdenv.mkDerivation (rec {
     simdjson
     spdlog
     zeek-broker
+    http-parser
+    asio
```
tobim, line 82: Add those to the `propagatedBuildInputs` of `restinio` instead.

nix/overlay.nix
```diff
@@ -25,6 +25,7 @@ in
   prev.xxHash.overrideAttrs (old: {
     patches = [ ./xxHash/static.patch ];
   });
+  restinio = final.callPackage ./restinio { };
```
tobim, line 28: Let's just hack this:
```suggestion
  http-parser = if !isStatic then prev.http-parser else
    prev.http-parser.overrideAttrs (_ : {
      postPatch = let
        cMakeLists = prev.writeTextFile {
          name = "http-parser-cmake";
          text = ''
            cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
            project(http_parser)
            include(GNUInstallDirs)

            add_library(http_parser http_parser.c)
            target_compile_options(http_parser PRIVATE -Wall -Wextra)
            target_include_directories(http_parser PUBLIC .)

            set_target_properties(http_parser PROPERTIES PUBLIC_HEADER http_parser.h)
            install(
              TARGETS http_parser
              ARCHIVE DESTINATION "''${CMAKE_INSTALL_LIBDIR}"
              LIBRARY DESTINATION "''${CMAKE_INSTALL_LIBDIR}"
              RUNTIME DESTINATION "''${CMAKE_INSTALL_BINDIR}"
              PUBLIC_HEADER DESTINATION "''${CMAKE_INSTALL_INCLUDEDIR}")
          '';
        };
      in ''
        cp ${cMakeLists} CMakeLists.txt
      '';
      nativeBuildInputs = [ prev.buildPackages.cmake ];
      makeFlags = [];
      buildFlags = [];
      doCheck = false;
  });
  restinio = final.callPackage ./restinio { };
```

tobim (approved):
Thanks for driving this forward!

PR #2594, opened 2022-09-23
Restructure docs links

mavam:
Aside from a few minor fixups, this PR performs two major changes:

1. Strip the `-vast` suffix from the docs links, e.g., `/docs/about-vast`
2. Hoist `/docs/develop/contribuging` to `/docs/contribute`

I git-grepped my way through the repository to catch all locations.

I double-checked that the redirects are in place when doing `yarn build`.

### :memo: Reviewer Checklist

- [x] Perform some spot checking or trust CI.


mavam:
@dominiklohmann any ideas why the CHANGELOG generation fails? I adapted the links both in the underlying sources and the generated output. In my mental model, there should be no further side effects.

dominiklohmann:
> @dominiklohmann any ideas why the CHANGELOG generation fails? I adapted the links both in the underlying sources and the generated output. In my mental model, there should be no further side effects.

You're not allowed to modify the generated CHANGELOG.md file when you're not also cutting a release. You can modify the individual `changelog/**/*.md` entries, though. The CHANGELOG.md file even has a disclaimer for this at the top:

```md
<!--
This file is generated automatically. Add individual changelog entries to the 'changelog/' subdirectory instead.
-->

```

dominiklohmann (approved):
Renders for me locally.

PR #2596, opened 2022-09-23
Implement VAST extension types in Python

mavam:
This PR integrates all features from #2518.

- [x] Add pattern extension type
- [x] Add IP address extension type
- [x] Add subnet extension type
- [x] Add enum extension type
- [x] Make extension type registration idempotent
- [x] Add unit tests for (de)serialization of types
- [x] Find solution or workaround for [ARROW-17839](https://issues.apache.org/jira/browse/ARROW-17839)
- [x] Create follow-up story for [supporting native conversion to pandas](https://arrow.apache.org/docs/python/extending_types.html#conversion-to-pandas)

### :memo: Reviewer Checklist

All functionality is covered by unit tests. Please check them.

### DIFF
--- a/python/tests/test_utils_arrow.py
+++ b/python/tests/test_utils_arrow.py
@@ -126,9 +126,6 @@ def test_ipc():
         "bar": 2,
     }
     enum_type = vua.EnumType(fields)
-    # Note: if we do not register the type prior to performing IPC, the
-    # roundtrip fails.
-    pa.register_extension_type(enum_type)
     enum_storage = pa.array([2, 1], pa.uint8())
     enum_array = pa.ExtensionArray.from_storage(enum_type, enum_storage)
     # Assemble a record batch.
--- a/python/tests/test_utils_arrow.py
+++ b/python/tests/test_utils_arrow.py
@@ -11,7 +11,7 @@ def test_unpack_ip():
 
 
 def test_ip_address_extension_type():
-    ty = vua.IPAddressType()
+    ty = vua.AddressType()
     bytes = b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\n\x01\x15\xa5"
     storage = pa.array([bytes], pa.binary(16))
     arr = pa.ExtensionArray.from_storage(ty, storage)
--- a/python/tests/test_utils_arrow.py
+++ b/python/tests/test_utils_arrow.py
@@ -5,9 +5,15 @@ import pytest
 import vast.utils.arrow as vua
 
 
-def test_unpack_ip():
-    bytes = b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\n\x01\x15\xa5"
-    assert vua.unpack_ip(bytes) == ipaddress.IPv4Address("10.1.21.165")
+def test_pattern_extension_type():
+    ty = vua.PatternType()
+    storage = pa.array(["/foo*bar/", "/ba.qux/"], pa.string())
+    arr = pa.ExtensionArray.from_storage(ty, storage)
+    arr.validate()
+    assert arr.type is ty
+    assert arr.storage.equals(storage)
+    assert arr[0].as_py() == "/foo*bar/"
+    assert arr[1].as_py() == "/ba.qux/"
 
 
 def test_ip_address_extension_type():
@@ -76,3 +82,8 @@ def test_schema_alias_extraction():
     assert names[1] == "bar"
     # The first name is the top-level type name.
     assert vua.name(schema) == "foo"
+
+
+def test_unpack_ip():
+    bytes = b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\n\x01\x15\xa5"
+    assert vua.unpack_ip(bytes) == ipaddress.IPv4Address("10.1.21.165")
--- a/python/tests/test_utils_arrow.py
+++ b/python/tests/test_utils_arrow.py
@@ -55,7 +55,7 @@ def test_enum_extension_type():
     ty = vua.EnumType(fields)
     assert ty.__arrow_ext_serialize__().decode() == json.dumps(fields)
     pa.register_extension_type(ty)
-    storage = pa.array([1, 2, 3, 2, 1], pa.uint32())
+    storage = pa.array([1, 2, 3, 2, 1], pa.uint8())
     arr = pa.ExtensionArray.from_storage(ty, storage)
     arr.validate()
     assert arr.type is ty
--- a/python/tests/test_utils_arrow.py
+++ b/python/tests/test_utils_arrow.py
@@ -1,6 +1,7 @@
 import json
 import ipaddress
 import pyarrow as pa
+import pytest
 
 import vast.utils.arrow as vua
 
@@ -78,6 +79,7 @@ def ipc_read_batch(buf):
     return reader.read_next_batch()
 
 
+@pytest.mark.skip(reason="blocked by https://github.com/apache/arrow/pull/14106")
 def test_ipc():
     # Create sample patterns.
     patterns = ["/foo/", "/bar/"]
--- a/python/tests/test_utils_arrow.py
+++ b/python/tests/test_utils_arrow.py
@@ -5,6 +5,22 @@ import pytest
 import vast.utils.arrow as vua
 
 
+def test_unpack_ip():
+    bytes = b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\n\x01\x15\xa5"
+    assert vua.unpack_ip(bytes) == ipaddress.IPv4Address("10.1.21.165")
+
+
+def test_ip_address_extension_type():
+    ty = vua.IPAddressType()
+    bytes = b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\n\x01\x15\xa5"
+    storage = pa.array([bytes], pa.binary(16))
+    arr = pa.ExtensionArray.from_storage(ty, storage)
+    arr.validate()
+    assert arr.type is ty
+    assert arr.storage.equals(storage)
+    assert arr[0].as_py() == ipaddress.IPv4Address("10.1.21.165")
+
+
 def test_schema_name_extraction():
     # Since Arrow cannot attach names to schemas, we do this via metadata.
     schema = pa.schema(
@@ -25,8 +41,3 @@ def test_schema_alias_extraction():
     assert names[1] == "bar"
     # The first name is the top-level type name.
     assert vua.name(schema) == "foo"
-
-
-def test_unpack_ip():
-    bytes = b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\n" b"\x01\x15\xa5"
-    assert vua.unpack_ip(bytes) == ipaddress.IPv4Address("10.1.21.165")
--- a/python/tests/test_utils_arrow.py
+++ b/python/tests/test_utils_arrow.py
@@ -32,6 +32,22 @@ def test_subnet_extension_type():
     assert arr[0].as_py() == ipaddress.IPv4Network("10.1.21.0/24")
 
 
+def test_enum_extension_type():
+    fields = {
+        1: "foo",
+        2: "bar",
+        3: "baz",
+    }
+    ty = vua.EnumType(fields)
+    pa.register_extension_type(ty)
+    storage = pa.array([1, 2, 3, 2, 1], pa.uint32())
+    arr = pa.ExtensionArray.from_storage(ty, storage)
+    arr.validate()
+    assert arr.type is ty
+    assert arr.storage.equals(storage)
+    assert arr.to_pylist() == ["foo", "bar", "baz", "bar", "foo"]
+
+
 def test_schema_name_extraction():
     # Since Arrow cannot attach names to schemas, we do this via metadata.
     schema = pa.schema(
--- a/python/tests/test_utils_arrow.py
+++ b/python/tests/test_utils_arrow.py
@@ -131,11 +131,15 @@ def test_ipc():
     assert isinstance(p, pa.ExtensionArray)
     assert p.type == pattern_type
     assert p.storage.to_pylist() == patterns
+    assert p.to_pylist() == patterns
     # Validate addresses.
     a = batch.column("a")
     assert isinstance(a, pa.ExtensionArray)
     assert a.type == address_type
     assert a.storage.to_pylist() == addresses
+    assert a.to_pylist() == [
+        ipaddress.IPv4Address(x) for x in ["10.1.21.165", "10.1.21.166"]
+    ]
     # Validate subnets.
     s = batch.column("s")
     assert isinstance(s, pa.ExtensionArray)
@@ -147,6 +151,7 @@ def test_ipc():
     assert isinstance(e, pa.ExtensionArray)
     assert e.type == enum_type
     assert e.storage.equals(enum_storage)
+    assert e.to_pylist() == ["bar", "foo"]
 
 
 def test_schema_name_extraction():
--- a/python/tests/test_utils_arrow.py
+++ b/python/tests/test_utils_arrow.py
@@ -95,8 +95,9 @@ def test_ipc():
     # Create sample subnets.
     network = b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\n\x01\x15\x00"
     networks = [network, network]
+    prefixes = [24, 25]
     address_array = pa.array(networks, pa.binary(16))
-    length_array = pa.array([24, 25], pa.uint8())
+    length_array = pa.array(prefixes, pa.uint8())
     subnet_type = vua.SubnetType()
     subnet_storage = pa.StructArray.from_arrays(
         [
--- a/python/tests/test_utils_arrow.py
+++ b/python/tests/test_utils_arrow.py
@@ -19,7 +19,7 @@ def test_pattern_extension_type():
 
 def test_ip_address_extension_type():
     ty = vua.AddressType()
-    bytes = b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\n\x01\x15\xa5"
+    bytes = vua.pack_ip("10.1.21.165")
     storage = pa.array([bytes], pa.binary(16))
     arr = pa.ExtensionArray.from_storage(ty, storage)
     arr.validate()
@@ -30,7 +30,7 @@ def test_ip_address_extension_type():
 
 def test_subnet_extension_type():
     ty = vua.SubnetType()
-    bytes = b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\n\x01\x15\x00"
+    bytes = vua.pack_ip("10.1.21.0")
     address_array = pa.array([bytes, bytes], pa.binary(16))
     length_array = pa.array([24, 25], pa.uint8())
     storage = pa.StructArray.from_arrays(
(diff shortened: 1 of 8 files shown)


mavam:
I didn't copy over from #2518 this functionality:

```python
def fmt_ip(df: pd.DataFrame, ip_columns: list[str]) -> pd.DataFrame:
mavam marked this conversation as resolved.
    """Formats IP of a dataframe Pass columns to format as arguments."""
    columns = []
    for col_name in df.keys():
        if col_name in ip_columns:
            columns.append(df[col_name].apply(unpack_ip))
        else:
            columns.append(df[col_name])
    return pd.concat(columns, axis=1)
```

and the corresponding test:

```python
def test_fmt_ip():
    df = pd.DataFrame(
        [b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\n\x01\x15\xa5"],
        columns=["ip"],
    )
    assert vua.fmt_ip(df, ip_columns=["ip"]).ip[0] == ipaddress.IPv4Address(
        "10.1.21.165"
    )
```

Everything else should be covered. That said, I think there might be an automatic conversion possible when calling `to_pandas()` on a `Table` instance, per the [Arrow documentation](https://arrow.apache.org/docs/python/extending_types.html#conversion-to-pandas).

mavam:
Some inspiration for the pandas extension type: https://github.com/ContinuumIO/cyberpandas/blob/master/cyberpandas/ip_array.py

mavam:
Here's an example that shows the registered extension types when printing a Zeek HTTP log:

```
pyarrow.Table
ts: timestamp[ns]
uid: string
id.orig_h: extension<vast.address<IPAddressType>>
id.orig_p: uint64
id.resp_h: extension<vast.address<IPAddressType>>
id.resp_p: uint64
trans_depth: uint64
method: string
host: string
uri: string
referrer: string
user_agent: string
request_body_len: uint64
response_body_len: uint64
status_code: uint64
status_msg: string
info_code: uint64
info_msg: string
filename: string
tags: list<item: string>
  child 0, item: string
username: string
password: string
proxied: list<item: string>
  child 0, item: string
mime_type: string
md5: string
extraction_file: string
```

Note `extension<vast.address<IPAddressType>>`.

python/vast/utils/arrow.py
```diff
@@ -1,7 +1,72 @@
...
+    def as_py(self) -> ip.IPv4Network | ip.IPv6Network:
+        buffer = self.value.as_py()
+        network = unpack_ip(buffer[0:16])
+        length = buffer[16]
+        return ip.ip_network((network, length), strict=False)
+
+
+class SubnetType(pa.ExtensionType):
+    def __init__(self):
+        pa.ExtensionType.__init__(self, pa.binary(17), "vast.subnet")
```
dominiklohmann, line 50: This should match its equivalent in libvast:

https://github.com/tenzir/vast/blob/1c3eec37e96db254ad73a405164718b4d435d779/libvast/src/type.cpp#L1995-L2000

python/vast/utils/arrow.py
```diff
@@ -1,7 +1,72 @@
...
+class IPAddressType(pa.ExtensionType):
+    def __init__(self):
+        pa.ExtensionType.__init__(self, pa.binary(16), "vast.address")
+
+    def __arrow_ext_serialize__(self):
+        return b""
+
+    @classmethod
+    def __arrow_ext_deserialize__(self, storage_type, serialized):
+        return IPAddressType()
```
dominiklohmann, line 31: These should match their equivalents in libvast:

https://github.com/tenzir/vast/blob/1c3eec37e96db254ad73a405164718b4d435d779/libvast/src/type.cpp#L1913-L1926

python/vast/utils/arrow.py
```diff
@@ -1,7 +1,72 @@
...
+class SubnetType(pa.ExtensionType):
+    def __init__(self):
+        pa.ExtensionType.__init__(self, pa.binary(17), "vast.subnet")
+
+    def __arrow_ext_serialize__(self):
+        return b""
+
+    @classmethod
+    def __arrow_ext_deserialize__(self, storage_type, serialized):
+        return SubnetType()
```
dominiklohmann, line 57: These should match their equivalents in libvast:

https://github.com/tenzir/vast/blob/1c3eec37e96db254ad73a405164718b4d435d779/libvast/src/type.cpp#L2016-L2029

python/vast/utils/arrow.py
```diff
@@ -1,7 +1,72 @@
...
+        num = num - (65535 << 32)
+    return ip.ip_address(num)
+
+
+class IPAddressScalar(pa.ExtensionScalar):
+    def as_py(self) -> ip.IPv4Address | ip.IPv6Address:
+        return unpack_ip(self.value.as_py())
+
+
+class IPAddressType(pa.ExtensionType):
```
dominiklohmann, line 22: In VAST, we so far have called the type just the "address" type, not "IP address."

I understand that IP address is a clearer name, and we'd probably choose that if we were to make a breaking change in VAST, but for now I think it'd be better to align the names for consistency.

```suggestion
class AddressType(pa.ExtensionType):
```

dominiklohmann (commented):
Just took an early peek at this. I left some pointers to where you can find the extension type implementations in libvast, which should make this a bit easier.

python/vast/utils/arrow.py
mavam: I did this structurally. Unfortunately extension types don't compose in Python:

```python
        # This gives us a pyarrow.lib.ArrowNotImplementedError :-/
        # arrow_type = pa.struct([("address", AddressType()),
        #                       ("length", pa.uint8())])
        arrow_type = pa.struct([("address", pa.binary(16)), ("length", pa.uint8())])
```

Fortunately it doesn't matter in this circumstance.

Please resolve if you think this is adequately addressed. (Pun intended.)

dominiklohmann: It doesn't matter when reading, but if we ever want to start generating arrays for our extension types inside Python (e.g., when modifying a column) then this will be a problem, because VAST won't be able to read them.

We should probably adapt the subnet extension type in VAST not to rely upon the address extension type then. We're not really using it in data, so it's only a theoretical and not a practical breaking change.

I'm fine with the change itself. Should we create a follow-up story for making the adjustment in VAST?

mavam:
@dominiklohmann can you check whether the names in bbdcf65 are correct?

python/vast/utils/arrow.py
mavam: Ah yes, good call. The interop is of course broken between C++ and Python if they don't use the *exact* same layout.

> I'm fine with the change itself. Should we create a follow-up story for making the adjustment in VAST?

I think we have to. 😒

python/vast/utils/arrow.py
```diff
@@ -1,7 +1,125 @@
...
+        return SubnetScalar
+
+
+class EnumScalar(pa.ExtensionScalar):
+    def as_py(self) -> str:
+        return self.type.field(self.value.as_py())
+
+
+class EnumType(pa.ExtensionType):
+    ext_name = "vast.enumeration_type"
```
dominiklohmann, line 88:
```suggestion
    ext_name = "vast.enumeration"
```

https://github.com/tenzir/vast/blob/1c3eec37e96db254ad73a405164718b4d435d779/libvast/include/vast/type.hpp#L1015-L1016

python/vast/utils/arrow.py
```diff
@@ -1,7 +1,125 @@
...
+
+class SubnetScalar(pa.ExtensionScalar):
+    def as_py(self) -> ip.IPv4Network | ip.IPv6Network:
+        address = unpack_ip(self.value[0].as_py())
+        length = self.value[1].as_py()
+        return ip.ip_network((address, length), strict=False)
+
+
+class SubnetType(pa.ExtensionType):
+    ext_name = "vast.subnet_type"
```
dominiklohmann, line 55:
```suggestion
    ext_name = "vast.subnet"
```

https://github.com/tenzir/vast/blob/1c3eec37e96db254ad73a405164718b4d435d779/libvast/include/vast/type.hpp#L848-L849

python/vast/utils/arrow.py
```diff
@@ -1,7 +1,125 @@
...
+    return ip.ip_address(num)
+
+
+class AddressScalar(pa.ExtensionScalar):
+    def as_py(self) -> ip.IPv4Address | ip.IPv6Address:
+        return unpack_ip(self.value.as_py())
+
+
+class AddressType(pa.ExtensionType):
+    ext_name = "vast.address_type"
```
dominiklohmann, line 24:
```suggestion
    ext_name = "vast.address"
```

https://github.com/tenzir/vast/blob/1c3eec37e96db254ad73a405164718b4d435d779/libvast/include/vast/type.hpp#L756-L757

python/vast/utils/arrow.py
```diff
@@ -1,7 +1,125 @@
...
+    def __reduce__(self):
+        return EnumScalar, ()
+
+    def __arrow_ext_scalar_class__(self):
+        return EnumScalar
+
+
+# TODO: move to appropriate location
+pa.register_extension_type(AddressType())
+pa.register_extension_type(SubnetType())
```
dominiklohmann, line 121:
```suggestion
pa.register_extension_type(AddressType())
pa.register_extension_type(SubnetType())
pa.register_extension_type(EnumType())
```

python/vast/utils/arrow.py
```diff
@@ -1,7 +1,125 @@
...
+
+
+class EnumScalar(pa.ExtensionScalar):
+    def as_py(self) -> str:
+        return self.type.field(self.value.as_py())
+
+
+class EnumType(pa.ExtensionType):
+    ext_name = "vast.enumeration_type"
+    ext_type = pa.uint32()
```
dominiklohmann, line 89:
```suggestion
    ext_type = pa.uint8()
```

https://github.com/tenzir/vast/blob/1c3eec37e96db254ad73a405164718b4d435d779/libvast/src/type.cpp#L2204-L2211

Note that this is an uint8 array for backwards compatibility with VAST's `enumeration` data variant, which has historically been an alias of `uint8_t`.

python/vast/utils/arrow.py
```diff
@@ -1,7 +1,125 @@
...
+        return EnumScalar, ()
+
+    def __arrow_ext_scalar_class__(self):
+        return EnumScalar
+
+
+# TODO: move to appropriate location
+pa.register_extension_type(AddressType())
+pa.register_extension_type(SubnetType())
+
```
dominiklohmann, line 122: Can you also add the `PatternType` while you're at it? it's just a wrapper around a string/utf8 type in Arrow, with the extension name `vast.pattern`.

dominiklohmann (approved):
Kinda excited already to play around with this. 🚀

python/vast/utils/arrow.py
```diff
@@ -1,7 +1,125 @@
...
+    def field(self, key):
+        return self._fields[key]
+
+    def __arrow_ext_serialize__(self) -> bytes:
+        return json.dumps(self._fields).encode()
+
+    @classmethod
+    def __arrow_ext_deserialize__(self, storage_type, serialized: bytes):
+        fields = json.loads(serialized.decode())
```
dominiklohmann, line 107: I think you need to match VAST's serialization format exactly here:

Serialization:

https://github.com/tenzir/vast/blob/1c3eec37e96db254ad73a405164718b4d435d779/libvast/src/type.cpp#L2252-L2265

Deserialization:

https://github.com/tenzir/vast/blob/1c3eec37e96db254ad73a405164718b4d435d779/libvast/src/type.cpp#L2228-L2250

python/vast/utils/arrow.py
mavam: We can't register an `EnumType` without fields.

Since we don't know all possible enums in advance, I wonder how this chicken-and-egg issue will work with IPC.

python/vast/utils/arrow.py
mavam: I could only find this in the unit tests:

```cpp
  CHECK_EQUAL(fmt::format("{}", et), "enum {first: 0, third: 2, fourth: 3}");
```

It doesn't look like the output, given that it's invalid JSON. Do we have a valid JSON somewhere so that can make sure to get the *exact* representation?

mavam: In 9cabcaa I reversed the field order during serialization, so that it's compatible with VAST.

I test that this generates valid JSON:

```python
    fields = {
        "foo": 1,
        "bar": 2,
        "baz": 3,
    }
    ty = vua.EnumType(fields)
    assert ty.__arrow_ext_serialize__().decode() == json.dumps(fields)
```

But we probably need a more thorough check by actually reading VAST data in the future.

python/vast/utils/arrow.py
mavam: Fixed in 364ba22fb.

mavam:
@dominiklohmann Do we have an example record batch with all extension types somewhere? This would be lovely for testing so that we can be assured that the two implementations align.

mavam:
I'm inclined to move the pandas conversion into a separate PR, given that this one is all about extension types. Thoughts?

python/vast/utils/arrow.py
mavam: Now that I added the tests for IPC, I have a data point: Arrow needs the concrete extension type to be registered, otherwise the roundtrip fails. How do we get around this in C++ land?

python/vast/utils/arrow.py
mavam: > I think we have to. 😒

Maybe not. I got a bit further in [96e3dfe](https://github.com/tenzir/vast/pull/2596/commits/96e3dfebf907974cfabc3f69d302effba7c1770a). Fortunately [`StructArray` instancess can have extension types](https://github.com/apache/arrow/blob/356e7f836c145966ebbeb65c3b65d82348e4234e/python/pyarrow/tests/test_extension_type.py#L825)!

There's just an issue when putting them into a `RecordBatch`. The error I'm getting now is:

```
Unsupported cast to extension<vast.address<AddressType>> from fixed_size_binary[16]
```

The folks at Huggingface have hand-rolled a [custom cast](https://github.com/huggingface/datasets/pull/3575). We may have to do the same here.

mavam: I also created [ARROW-17839](https://issues.apache.org/jira/browse/ARROW-17839) for this.

mavam: I've escalated this to [ARROW-17839](https://issues.apache.org/jira/browse/ARROW-17839).

python/vast/utils/arrow.py
dominiklohmann: You just need to register one instance of the extension type:

https://github.com/tenzir/vast/blob/1c3eec37e96db254ad73a405164718b4d435d779/libvast/src/type.cpp#L2138-L2144

python/vast/utils/arrow.py
dominiklohmann: We have an `all_types.schema` file somewhere in the repository. If you create sample data of that, export it as Arrow, and then store that in some .bin file in the repository I think you have a good unit test setup.

dominiklohmann:
> @dominiklohmann Do we have an example record batch with all extension types somewhere? This would be lovely for testing so that we can be assured that the two implementations align.

I commented this in some thread. We don't right now (except for the old Arrow v1 FlatBuffers table), but it should be easy enough to create and check in as a binary file for use in unit tests.

python/vast/utils/arrow.py
dominiklohmann: > I could only find this in the unit tests:
> 
> ```cpp
> CHECK_EQUAL(fmt::format("{}", et), "enum {first: 0, third: 2, fourth: 3}");
> ```
> 
> It doesn't look like the output, given that it's invalid JSON. Do we have a valid JSON somewhere so that can make sure to get the exact representation?

This is the printer of the enumeration type, it's not testing the Arrow extension type's serialization format. It's not related to it at all.

dispanser: we only test the ser-de cycle (write schema, read it back) and AFAICT do not have tests that actually look at the json.

There's an integration test looking at the output of `print_arrow.py` containing `ARROW:extension:metadata: '{ "A": 0, "B": 1, "C": 2 }'` over at `vast/integration/reference/arrow-full-data-model/step_01.ref`

dispanser: arion contains an IPC file dump that contains the `all-types.schema` mentioned above in `web/arion/tests/data/all-types_arrow-v2.ipc`. If we can read this from python, we should be good.

rdettai:
A small side note that slipped through the previous PR: the README should instruct to run 
```bash
poetry install --all-extras
```

Otherwise neither the examples nore the tests work

python/vast/utils/arrow.py
mavam: Fortunately there is a fix underway: https://github.com/apache/arrow/pull/14106.

python/vast/utils/arrow.py
mavam: Got it. I added the same stub enum type registration.

python/vast/utils/arrow.py
mavam: I kept disabled the failing unit test because it will likely take a a while until we can consume the fix here.

Since subnet types are quite rare in VAST data, I'm proposing that we proceed with this PR. The majority of Python analysis still works, and users will only be impacted when exporting subnet data. @dominiklohmann if this sounds acceptable, I would create a separate story for re-enabling the disabled test.

python/vast/utils/arrow.py
mavam: Good call. I created a follow-up story to do this.

Resolving this thread because complete IPC support is currently still blocked anyway.

python/vast/utils/arrow.py
```diff
@@ -11,3 +150,21 @@ def names(schema: pa.Schema):
...
+    if (num >> 32) == 65535:
+        num = num - (65535 << 32)
+    return ip.ip_address(num)
+
+
+# Modules are intialized exactly once, so we can perform the registration here.
+pa.register_extension_type(PatternType())
+pa.register_extension_type(AddressType())
+pa.register_extension_type(SubnetType())
+pa.register_extension_type(EnumType({"stub": 0}))
```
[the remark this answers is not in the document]
mavam: @dominiklohmann suggested this because we're doing this in C++.

> Isn't registration required for each specific enum definition anyway?

I'm actually skeptical that our approach works at all, because registration of an extension type happens by name. We currently register this particular instance under the type name `vast.enumeration`. Registering *another* enum fails.

python/tests/test_utils_arrow.py
```diff
@@ -1,9 +1,172 @@
...
+@pytest.mark.skip(reason="blocked by https://github.com/apache/arrow/pull/14106")
+def test_ipc():
+    # Create sample patterns.
+    patterns = ["/foo/", "/bar/"]
+    pattern_type = vua.PatternType()
+    pattern_storage = pa.array(patterns, pa.string())
+    pattern_array = pa.ExtensionArray.from_storage(pattern_type, pattern_storage)
+    # Create sample IP addresses.
+    addresses = [
+        b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff\xff\n\x01\x15\xa5",
```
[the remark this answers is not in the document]
mavam: Really? 😅 

This is actually an IPv4 address embedded in the RFC-compliant IPv6 container. I think dotted decimal would be more intuitive.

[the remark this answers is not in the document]
mavam: I added a new function `pack_ip` that does the opposite of `unpack_ip`. If that suits the bill for you, please resolve.

python/vast/utils/arrow.py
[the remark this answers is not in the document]
rdettai: What about adding the hash of the `_fields` to the name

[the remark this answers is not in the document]
dominiklohmann: This is required to mirror what we do on the C++ side of things for compatibility.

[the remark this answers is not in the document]
mavam: But how does it even work in C++ land?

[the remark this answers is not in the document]
dominiklohmann: The registration only ever calls member functions of the extension type that are required to have a constant output regardless of the types instantiation. Essentially, you don't register an instance of the extension type, but rather the extension type itself. Registering only takes an instance of the type because it's a badly designed API.

python/vast/utils/arrow.py
dominiklohmann: That sounds like the right approach to me.

PR #2597, opened 2022-09-28
Fix shutdown hang in the filesystem actor

dominiklohmann:
This fixes a regression I introduced myself last week when adding metrics to the filesystem actor: A cycle between the accountant and the filesystem that caused an infinite hang on shutdown when metrics are enabled. We didn't catch this in integration tests because those, well, run with metrics disabled.

Because I didn't want to add a `actor_cast<T>(some_weak_actor_ptr)` to our code base in a seemingly random place, I added a typed abstraction for weak actor handles to our code base, which feels like a thing that we should contribute upstream as `typed_actor<...>::weak_handle` or similar.

This shouldn't need a changelog entry since this regression was introduced only after the last releaase.

For testing, it's easy enough to see it in action by running this command, which shuts down as of this PR, but was hanging indefinitely on master:

```
vast --bare-mode --enable-metrics start --commands=stop
```

### :memo: Reviewer Checklist

### DIFF
--- a/.github/workflows/style-check.yaml
+++ b/.github/workflows/style-check.yaml
@@ -18,8 +18,9 @@ jobs:
 
       - name: Clang Format
         run: |
-          brew install clang-format
-          PATH="$(brew --prefix clang-format)/bin:${PATH}"
+          pip install --upgrade clang-format
+          which clang-format
+          clang-format --version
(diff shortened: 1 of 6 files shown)


dispanser (approved):
@dominiklohmann walked me through this, and showed me that it now works. 

An integration test would be nice.

PR #2598, opened 2022-09-29
Make style check CI clang-format installation more robust

dominiklohmann:
This started failing recently. We used Linuxbrew to install clang-format historically in the Style Check CI workflow because it was (1) pre-installed and (2) allowed for installing a more recent clang-format version. (1) looks to no longer be the case on newer runners, and (2) is also not true anymore, so let's just install it from PyPI.

### :memo: Reviewer Checklist

### DIFF
--- a/.github/workflows/style-check.yaml
+++ b/.github/workflows/style-check.yaml
@@ -18,8 +18,9 @@ jobs:
 
       - name: Clang Format
         run: |
-          brew install clang-format
-          PATH="$(brew --prefix clang-format)/bin:${PATH}"
+          pip install --upgrade clang-format
+          which clang-format
+          clang-format --version
           git diff -U0 --no-color $(git merge-base origin/master HEAD) -- ':(exclude)*.js' |
             scripts/clang-format-diff.py -p1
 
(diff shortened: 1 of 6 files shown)


.github/workflows/style-check.yaml
```diff
@@ -18,8 +18,9 @@ jobs:
 
       - name: Clang Format
         run: |
-          brew install clang-format
-          PATH="$(brew --prefix clang-format)/bin:${PATH}"
+          pip install --upgrade clang-format
+          which clang-format
```
mavam, line 22: Why do we need this?

.github/workflows/style-check.yaml
```diff
@@ -18,8 +18,9 @@ jobs:
...
+          clang-format --version
```
mavam, line 23: Is this just diagnostics?

mavam (approved):

.github/workflows/style-check.yaml
dominiklohmann: If this job fails this and the clang-format version are the first things you're looking for if you cannot reproduce the failure locally.

.github/workflows/style-check.yaml
dominiklohmann: See https://github.com/tenzir/vast/pull/2598#discussion_r983815089

PR #2605, opened 2022-09-30
Make it less likely for rebuild to require rebatching

dominiklohmann:
Scenario: We have the partitions A and B, and automatic rebuilding merges them into the partition AB. All except for the last batch in AB are perfectly sized.

Now comes partition C with undersized batches. If the rebuilder operates in the order C -> AB it is more likely to have to rebatch than if it were to operate in the order AB -> C, because the beginning of AB is already guaranteed not to require rebatching.

This change makes it so we always rebuild the selected partitions for a given rebuild run in time ascending order, while keeping the selection order for rebuilding to be time descending.

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/native-plugins/rebuild.cpp
+++ b/libvast/native-plugins/rebuild.cpp
@@ -599,11 +599,10 @@ struct rebuilder_state {
         detail::narrow_cast<int64_t>(desired_batch_size)));
     }
     emit_telemetry();
-    // Newer partitions are more likely to contain undersized batches, so in the
-    // interest of letting rebatching for the rebuild of the current set of
-    // partitions not have to rebatch anything for as long as possible we sort
-    // the partitions for the current run by time ascending before sending them
-    // off.
+    // We sort the selected partitions from old to new so the rebuild transform
+    // sees the batches (and events) in the order they arrived. This prevents
+    // the rebatching from shuffling events, and rebatching of already correctly
+    // sized batches just for the right alignment.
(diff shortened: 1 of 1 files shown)


libvast/native-plugins/rebuild.cpp
```diff
@@ -599,6 +599,15 @@ struct rebuilder_state {
         detail::narrow_cast<int64_t>(desired_batch_size)));
     }
     emit_telemetry();
+    // Newer partitions are more likely to contain undersized batches, so in the
+    // interest of letting rebatching for the rebuild of the current set of
+    // partitions not have to rebatch anything for as long as possible we sort
+    // the partitions for the current run by time ascending before sending them
+    // off.
```
tobim, line 606: I had to read that comment multiple times before I started to understand what it is trying to say.
How about rephrasing to:
```suggestion
    // We sort the selected partitions from old to new so the rebuild transform
    // sees the batches (and events) in the order they arrived. This prevents
    // the rebatching from shuffling events, and rebatching of already correctly
    // sized batches just for the right alignment.
```

tobim (approved):
Makes total sense from a functional pov, but the explanation needs some work.

libvast/native-plugins/rebuild.cpp
dominiklohmann: Thanks for improving the wording, yours is much clearer.

PR #2606, opened 2022-10-01
Make enum extension types work

mavam:
This PR fixes the failing `EnumType` unit tests.

Turns out that `pa.array()` with `type=T` where `T` is a dictionary type is not sufficient to use the proper type for the dictionary indices. We had `uint8` in our type, but unless we explicitly go through `pa.DictionaryArray.from_arrays`, we will get a `int8`. This type clash was the issue in the past.

### DIFF
--- a/python/tests/test_utils_arrow.py
+++ b/python/tests/test_utils_arrow.py
@@ -48,15 +48,6 @@ def test_subnet_extension_type():
     assert arr[1].as_py() == ipaddress.IPv4Network("10.1.21.0/25")
 
 
-# Currently fails with:
-#
-# TypeError: Incompatible storage type
-#   dictionary<values=string, indices=int8, ordered=0> for extension type
-#   extension<vast.enumeration<EnumType>>
-#
-# The discussion in https://issues.apache.org/jira/browse/ARROW-17839 ideally
-# tells us exactly what we are doing wrong here.
-@pytest.mark.xfail(reason="https://issues.apache.org/jira/browse/ARROW-17839")
 def test_enum_extension_type():
     fields = {
         "foo": 1,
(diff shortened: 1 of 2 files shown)


mavam:
Reading a feather file now correctly identifies the enum type:

```
event: struct<timestamp: timestamp[ns], flow_id: uint64, pcap_cnt: uint64, vlan: list<item: uint64>, in_iface: string, src_ip: extension<vast.address<AddressType>>, src_port: uint64, dest_ip: extension<vast.address<AddressType>>, dest_port: uint64, proto: string, event_type: string, community_id: string, dns: struct<version: uint64, type: extension<vast.enumeration<EnumType>>, id: uint64, flags: string, qr: bool, rd: bool, ra: bool, aa: bool, tc: bool, rrname: string, rrtype: string, rcode: string, ttl: uint64, tx_id: uint64, grouped: struct<A: list<item: extension<vast.address<AddressType>>>, NS: list<item: string>, AAAA: list<item: extension<vast.address<AddressType>>>, CNAME: list<item: string>, TXT: list<item: string>, MX: list<item: string>, PTR: list<item: string>, SOA: list<item: struct<mname: string, rname: string, serial: uint64, refresh: uint64, retry: uint64, expire: uint64, minimum: uint64>>, SSHFP: list<item: struct<fingerprint: string, algo: uint64, type: uint64>>, SRV: list<item: struct<priority: uint64, weight: uint64, port: uint64, name: string>>>, answers: list<item: struct<rrname: string, rrtype: string, ttl: uint64, rdata: string, soa: struct<mname: string, rname: string, serial: uint64, refresh: uint64, retry: uint64, expire: uint64, minimum: uint64>, sshfp: struct<fingerprint: string, algo: uint64, type: uint64>, srv: struct<priority: uint64, weight: uint64, port: uint64, name: string>>>>>
```

Note the field `event.dns.type` with type `extension<vast.enumeration<EnumType>>`

mavam:
@dominiklohmann maybe we should also do the memo table fix in here?

dominiklohmann:
> @dominiklohmann maybe we should also do the memo table fix in here?

I think this will only become relevant once we add builders for the extension types on the Python side, so not now.

dominiklohmann (approved):
3/4—we're getting there! 🚀

PR #2607, opened 2022-10-03
Deduce version of Python bindings automatically

mavam:
This PR leverages https://github.com/mtkennerly/poetry-dynamic-versioning to deduce the version of the Python bindings dynamically.

The goal is to use the exact same version as VAST, so that both VAST and the bindings are always in sync.

### :memo: Reviewer Checklist

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,8 +1,6 @@
 [tool.poetry]
 name = "vast"
-# TODO: discuss how to align this with the VAST version, e.g., adopt the VAST
-# version from the top-level directory.
-version = "0.1.0"
+version = "0.0.0"
 description = "A security telemetry engine for detection and response"
 authors = ["Tenzir <engineering@tenzir.com>"]
 maintainers = ["Tenzir <engineering@tenzir.com>"]
@@ -41,6 +39,9 @@ pytest-asyncio = "^0.19.0"
 [tool.poetry.extras]
 fabric = ["misp-stix", "pyzmq"]
 
+[tool.poetry-dynamic-versioning]
+enable = true
+
 [build-system]
-requires = ["poetry>=0.12"]
-build-backend = "poetry.masonry.api"
+requires = ["poetry-core>=1.0.0", "poetry-dynamic-versioning"]
+build-backend = "poetry_dynamic_versioning.backend"


mavam:
I played with this a bit and think that we are not far away from the desired outcome. When I run `poetry build` in `/python` I get:

```
Building vast (2.3.0.post415.dev0+0a1f7c555)
  - Building sdist
  - Built vast-2.3.0.post415.dev0+0a1f7c555.tar.gz
  - Building wheel
  - Built vast-2.3.0.post415.dev0+0a1f7c555-py3-none-any.whl
```

mavam:
I think our version format is actually not standard, causing some issues with the validation: I tried this in `pyproject.toml`:

```toml
[tool.poetry-dynamic-versioning]
enable = true
style = "semver"
# FIXME: none of the allowed style options ("semver", "pvp", "pep440") work with
# this format.
format = "v{base}-{distance}-{commit}"
```

I think I got `format` right, but the error I'm getting when setting any of the allowed `style` variables is:

```
Version 'v2.3.0-415-0a1f7c555' does not conform to the PVP style
Version 'v2.3.0-415-0a1f7c555' does not conform to the Semantic Versioning style
Invalid PEP 440 version: 'v2.3.0-415-0a1f7c555'
```

tobim:
I found a blog using `poetry version` from 2020: https://www.nicholasnadeau.com/post/2020/08/one-version-to-rule-them-all-keeping-your-python-package-version-number-in-sync-with-git-and-poetry/

But I suppose that would only work for published artifacts but not for local development.

mavam:
@tobim I switched back to manual versioning via our release script that uses exactly that.

.github/workflows/vast.yaml
```diff
@@ -1096,3 +1096,43 @@ jobs:
...
+          python-version: "3.10"
+      - name: Setup Poetry image
+        uses: abatilo/actions-poetry@v2.1.6
+        with:
+          poetry-version: 1.2.1
+      - name: Configure Test PyPI
+        working-directory: python
+        run: |
+          poetry config repositories.test-pypi https://test.pypi.org/legacy/
+          poetry config pypi-token.test-pypi "${{ secrets.TEST_PYPI_TOKEN }}"
```
dominiklohmann, line 1128: I noticed that this is a new secret. For new secrets, please manage them as [Organization Secrets](https://github.com/organizations/tenzir/settings/secrets/actions), and enable them for the tenzir/vast repository only. This has the advantage that we can see all secrets in one place, and enable them for individual repositories selectively.

.github/workflows/vast.yaml
```diff
@@ -1096,3 +1096,43 @@ jobs:
...
+      - name: Configure PyPI
+        working-directory: python
+        run: |
+          poetry config pypi-token.pypi "${{ secrets.PYPI_TOKEN }}"
+      - name: Publish ${{ needs.determine-version.outputs.release-version }} to Test PyPI
+        working-directory: python
+        run: poetry publish --build -r test-pypi
+      - name: Publish to PyPI
+        working-directory: python
+        run: poetry publish --build
```
dominiklohmann, line 1138: Style request for consistency:

```suggestion
        run: |
          poetry publish --build -r test-pypi
      - name: Publish to PyPI
        working-directory: python
        run: |
          poetry publish --build
```

.github/workflows/vast.yaml
```diff
@@ -1096,3 +1096,43 @@ jobs:
...
+      - determine-version
+    if: github.event_name == 'release'
+    name: Python Package
+    runs-on: ubuntu-latest
+    env:
+      DEBIAN_FRONTEND: noninteractive
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.ref }}
```
dominiklohmann, line 1115: That's already the default.

```suggestion
```

python/README.md
```diff
@@ -44,7 +44,21 @@ poetry run pytest
 The following instructions concern maintainers who want to publish the Python
 package to PyPI.
 
-### Setup Test PyPI
+> **Note**
+> Our releasing scripts and CI od these steps automatically. You do not need to
+> intervene anywhere. The instructions below merely document the steps taken.
```
dominiklohmann, line 49:
```suggestion
> Our releasing scripts and CI run these steps automatically. You do not need to
> intervene anywhere. The instructions below merely document the steps taken.
```

python/pyproject.toml
```diff
@@ -1,8 +1,6 @@
 [tool.poetry]
 name = "vast"
-# TODO: discuss how to align this with the VAST version, e.g., adopt the VAST
-# version from the top-level directory.
-version = "0.1.0"
+version = "2.3.0-rc2"
```
dominiklohmann, line 3: This should be changed to be exactly 2.3.0, which is the last release.

.github/workflows/vast.yaml
```diff
@@ -1096,3 +1096,43 @@ jobs:
       - name: Run unit tests
         working-directory: python
         run: poetry run pytest
+      - name: Build package
+        working-directory: python
+        run: poetry build
```
dominiklohmann, line 1101: Same as below: Please turn this into a block with | for consistency; this makes it easier to spot run blocks when scrolling through the file.

.github/workflows/vast.yaml
```diff
@@ -1096,3 +1096,43 @@ jobs:
...
+    env:
+      DEBIAN_FRONTEND: noninteractive
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.ref }}
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
```
dominiklohmann, line 1119: This is hard-coded, but above we're using a matrix. Does poetry upload packages per Python version, or do we always just use the latest version here?

dominiklohmann (approved):
Some minor remarks and a question, other than that I think we're good to go for testing this with v2.4.0-rc1 soon.

.github/workflows/vast.yaml
mavam: Yes, this is on purpose. We want to publish the package exactly once, not once per Python version.

mavam: (Please resolve, if you do not disagree.)

PR #2613, opened 2022-10-07
Use CAF 0.18 approach for argument parsing in unit tests main function

patszt:
caf::message_builder::extract_opts method was remove in CAF 0.18.
We can apply the same approach as in configuration.cpp to use caf::config_option_set in order to parse and process the arguments passed into vast-test.

### :memo: Reviewer Checklist

### DIFF
--- a/libvast_test/src/main.cpp
+++ b/libvast_test/src/main.cpp
@@ -79,6 +79,8 @@ extern std::set<std::string> config;
 } // namespace vast::test
 
 namespace {
+
+// Retrieves arguments after the '--' delimiter.
 std::vector<std::string> get_test_args(int argc, const char* const* argv) {
   // Parse everything after after '--'.
   constexpr std::string_view delimiter = "--";
@@ -89,6 +91,7 @@ std::vector<std::string> get_test_args(int argc, const char* const* argv) {
     return {};
   return {args_start + 1, end};
 }
+
 } // namespace
 
 int main(int argc, char** argv) {
@@ -111,8 +114,10 @@ int main(int argc, char** argv) {
       std::cout << options.help_text() << std::endl;
       return 0;
     }
-    vast::test::config = {make_move_iterator(begin(test_args)),
-                          make_move_iterator(end(test_args))};
+    vast::test::config = {
+      std::make_move_iterator(std::begin(test_args)),
+      std::make_move_iterator(std::end(test_args)),
+    };
   }
   for (auto& plugin : vast::plugins::get_mutable()) {
     if (auto err = plugin->initialize({})) {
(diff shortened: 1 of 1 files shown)


libvast_test/src/main.cpp
```diff
@@ -82,32 +78,41 @@ extern std::set<std::string> config;
...
     }
-    if (res.opts.count("help") > 0) {
-      std::cout << res.helptext << std::endl;
+    if (caf::get_or(cfg, "help", false)) {
+      std::cout << options.help_text() << std::endl;
       return 0;
     }
-    vast::test::config = std::move(res.opts);
+    vast::test::config = {make_move_iterator(begin(test_args)),
+                          make_move_iterator(end(test_args))};
```
dominiklohmann, line 115: Please qualify calls into namespace std instead of ADL-using `make_move_iterator`, `begin`, and `end` here.

libvast_test/src/main.cpp
```diff
@@ -82,32 +78,41 @@ extern std::set<std::string> config;
 
 } // namespace vast::test
 
-int main(int argc, char** argv) {
+namespace {
+std::vector<std::string> get_test_args(int argc, const char* const* argv) {
```
dominiklohmann, line 82:
```suggestion

// Retrieves arguments after the '--' delimiter.
std::vector<std::string> get_test_args(int argc, const char* const* argv) {
```

Assuming my understanding is correct here.

libvast_test/src/main.cpp
```diff
@@ -82,32 +78,41 @@ extern std::set<std::string> config;
...
-  }
+  constexpr std::string_view delimiter = "--";
+  auto start = argv + 1;
+  auto end = argv + argc;
+  auto args_start = std::find(start, end, delimiter);
+  if (args_start == end)
+    return {};
+  return {args_start + 1, end};
+}
+} // namespace
```
dominiklohmann, line 92:
```suggestion

} // namespace
```

libvast_test/src/main.cpp
dominiklohmann, line 115: With brace-init, you usually get a lot better formatting from clang-format if you add a trailing comma. We don't have a fixed style guideline for this, so I'll leave that decision up to you.

dominiklohmann (approved):
Looks great to me. I just have some minor suggestions, but approving already.

libvast_test/src/main.cpp
patszt: My motivation here is to have a code that doesn't need any change if we introduce a 3rd party container instead of STL one.
It could provide it's own begin/end free function.

Although my approach has one pitfall that i have just recently read.
Let's say the 3rd party container has .begin() and .end() method.

Then std::begin(x) would return x-.begin(), whereas begin(x) will lead to compilation error.

The solution would be to use using std::begin, but this seems like too much for a code that won't probably use non STL container.

I will qualify into std :D

PR #2614, opened 2022-10-07
Attempt to fix the static binary build for the web plugin

dit7ya:
PR to check if the regular CI build works with this fix.

### :memo: Reviewer Checklist


lava (approved):
I think @tobim still had some open questions about whether it makes sense to always use our own CMakeLists for restinio so we can avoid the custom http-parser workarounds that the static build uses, but in the meantime this solution seems to work well enough.

plugins/web/CMakeLists.txt
```diff
@@ -26,11 +26,16 @@ VASTRegisterPlugin(
 
 find_package(OpenSSL REQUIRED)
 
-# Setup dependency code from submodules.
-list(PREPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
-add_subdirectory(aux)
-target_link_libraries(web PRIVATE restinio)
-list(POP_FRONT CMAKE_MODULE_PATH)
+if (VAST_ENABLE_SYSTEM_RESTINIO)
```
dominiklohmann, line 29: Should this use the same logic as other optionally bundled packages, e.g., similar to `VAST_ENABLE_BUNDLED_CAF` or `VAST_ENABLE_BUNDLED_ROBIN_MAP`?

The option here is essentially undocumented, and the logic differs from what we do in `libvast/CMakeLists.txt` for the others. The naming also differs, which I find rather confusing.

tobim: Makes sense to align with the rest.
I would put the option into the plugin CMakelists.txt and enable it by default.

dominiklohmann: Yeah, I'm pushing a fix to this directly, we just talked about this quickly in Slack.

dominiklohmann: @tobim mind taking a look at that last commit I just pushed and resolving this if you're fine with the change?

plugins/web/CMakeLists.txt
```diff
@@ -26,10 +26,18 @@ VASTRegisterPlugin(
...
   find_package(fmt REQUIRED)
-  find_package(restinio REQUIRED)
-else ()
+  find_package(restinio)
+endif ()
+if (NOT restinio_FOUND)
+  if (NOT VAST_ENABLE_BUNDLED_RESTINIO)
+    message(
+      STATUS "Cannot find installed restinio; falling back to bundled version")
+  endif ()
```
tobim, line 40: If the user opts out explicitly it should rather fail, should it not?

dominiklohmann: We talked about this in the Huddle; I pushed a change in the most recent commit that makes this a bit better.

tobim (approved):

PR #2616, opened 2022-10-07
Fix terminator lifetime issue

patszt:
Terminator and it's response promise could die before the response was received. This caused actors to not shutdown properly.
The call to terminate will now capture terminator into response handlers so it can never die before it responds.

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/include/vast/system/actors.hpp
+++ b/libvast/include/vast/system/actors.hpp
@@ -507,6 +507,9 @@ using node_actor = typed_actor_fwd<
   // TODO: Make this a signal_monitor_client_actor
   caf::reacts_to<atom::signal, int>>::unwrap;
 
+using terminator_actor = typed_actor_fwd<caf::replies_to<
+  atom::shutdown, std::vector<caf::actor>>::with<atom::done>>::unwrap;
+
 } // namespace vast::system
 
 // -- type announcements -------------------------------------------------------
--- a/libvast/include/vast/system/actors.hpp
+++ b/libvast/include/vast/system/actors.hpp
@@ -507,8 +507,10 @@ using node_actor = typed_actor_fwd<
   // TODO: Make this a signal_monitor_client_actor
   caf::reacts_to<atom::signal, int>>::unwrap;
 
-using terminator_actor = typed_actor_fwd<caf::replies_to<
-  atom::shutdown, std::vector<caf::actor>>::with<atom::done>>::unwrap;
+using terminator_actor = typed_actor_fwd<
+  // Shut down the given actors.
+  caf::replies_to<atom::shutdown,
+                  std::vector<caf::actor>>::with<atom::done>>::unwrap;
 
 } // namespace vast::system
 
--- a/libvast/include/vast/system/terminate.hpp
+++ b/libvast/include/vast/system/terminate.hpp
@@ -34,9 +34,9 @@ namespace vast::system {
 
 /// Wrapper which extends terminator's lifetime to the response handlers.
 template <class ResponseHandle>
-class termination_result {
+class terminate_result {
 public:
-  termination_result(terminator_actor terminator, ResponseHandle response)
+  terminate_result(terminator_actor terminator, ResponseHandle response)
     : terminator_{std::move(terminator)}, response_{std::move(response)} {
   }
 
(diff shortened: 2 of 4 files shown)


libvast/include/vast/system/terminate.hpp
```diff
@@ -31,6 +32,45 @@ struct parallel;
...
+/// Wrapper which extends terminator's lifetime to the response handlers.
+template <class ResponseHandle>
+class termination_result {
+public:
+  termination_result(terminator_actor terminator, ResponseHandle response)
+    : terminator_{std::move(terminator)}, response_{std::move(response)} {
+  }
+
+  template <class ResponseHandler, class ErrorHandler>
+  decltype(auto)
```
dominiklohmann, line 44: Is there any distinct advantage to returning `decltype(auto)` over `auto` here?

libvast/include/vast/system/terminate.hpp
```diff
@@ -31,6 +32,45 @@ struct parallel;
...
+  then(ResponseHandler responseHandler, ErrorHandler errorHandler) {
+    return response_.then(
+      [f = std::move(responseHandler), t = terminator_](atom::done) mutable {
+        f(atom::done_v);
```
dominiklohmann, line 48:
```suggestion
        std::move(f)(atom::done_v);
```

Otherwise this will not work with mutable lambdas that capture move-only variables.

libvast/include/vast/system/terminate.hpp
```diff
@@ -31,6 +32,45 @@ struct parallel;
 
 namespace vast::system {
 
+/// Wrapper which extends terminator's lifetime to the response handlers.
+template <class ResponseHandle>
+class termination_result {
```
dominiklohmann, line 37: Naming-wise I think it'd be clearer to name this `terminate_result`, because it really is the output of the `terminate` function and not the output of the termination itself. This would be in line with standard library naming as well, e.g., `std::from_chars_result` for `std::from_chars`.

libvast/include/vast/system/actors.hpp
```diff
@@ -507,6 +507,9 @@ using node_actor = typed_actor_fwd<
   // TODO: Make this a signal_monitor_client_actor
   caf::reacts_to<atom::signal, int>>::unwrap;
 
+using terminator_actor = typed_actor_fwd<caf::replies_to<
+  atom::shutdown, std::vector<caf::actor>>::with<atom::done>>::unwrap;
```
dominiklohmann, line 511: Throughout the file we force line breaks in the typed actor definitions like this by commenting every handler. This makes it easy to understand what a given handler does from looking at this one file only.

```suggestion
using terminator_actor = typed_actor_fwd<
  // Shut down the given actors.
  caf::replies_to<atom::shutdown, std::vector<caf::actor>>::with<
    atom::done>>::unwrap;
```

dominiklohmann (approved):
Thanks for converting one more actor to typed interfaces as well while you were at it. I left some minor remarks inline, but I think this is a great way to fix the issue.

libvast/include/vast/system/terminate.hpp
patszt: Yeah, the name is quite bad. I forgot to change it once i have changed the approach. What do you think about simply terminate_response_handle?

dominiklohmann: I'm fine with either `terminate_result` or `terminate_result_handle`, I think both work fine.

libvast/include/vast/system/terminate.hpp
patszt: Can you provide some example or some resource? I have tried to reproduce this with online compiler and it seems to work without the move.

`auto k = std::make_unique<int>(5);   
    auto f = [b = std::move(k)]()mutable{
        b.reset();
    };

    auto f2 = [k= std::move(f)]()mutable{k();};
    f2();`

dominiklohmann: Consider an invocable f that can only be called as an rvalue, e.g.,

```cpp
struct F {
  auto operator() && {}
};

auto g = [f = F{}]() mutable {
  // this works:
  std::move(f)(); 
  // this fails:
  // f();
};
```

The same can occur when your function f tries to mutate its own capture, but then it's implicit rather than explicit:

```c++
auto ptr = std::make_unique<int>();

auto f = [ptr = std::move(ptr)]() mutable {
  ptr.release();
};

auto g = [f = std::move(f)]() mutable {
  // this works:
  std::move(f)(); 
  // this fails:
  // f();
};
```

PR #2617, opened 2022-10-07
Add Parquet/Feather blog 2/3

mavam:
This PR adds the second blog post on Parquet and Feather.

### :memo: Reviewer Checklist

- [x] The post builds and looks good locally


web/blog/parquet-and-feather-writing-security-telemetry/index.qmd
```diff
@@ -0,0 +1,868 @@
...
+To better understand this spectrum, we instrumented the write path of VAST,
+which consists roughly of the following steps:
+
+1. Parse the input
+2. Convert it into Arrow record batches
+3. Ship Arrow record batches to a VAST server
+4. Write Arrow out into a Parquet or Feather store
+
+Since steps (1–3) are the same for both stores, we ignore them in the following
+analysis and solely zoom in on (4).
```
dominiklohmann, line 50:
```suggestion
4. Write Arrow record batches out into a Parquet or Feather store
5. Create an index from Arrow record batches

Since steps (1–3) and (5) are the same for both stores, we ignore them in the following
analysis and solely zoom in on (4).
```

web/blog/parquet-and-feather-writing-security-telemetry/index.qmd
```diff
@@ -0,0 +1,868 @@
...
+       record_type{
+         {"outer", count_type{}},
+         {"inner", count_type{}},
+       }},
+      {"community_id", type{string_type{}, {{"index", "hash"}}}},
+      {"payload", type{string_type{}, {{"skip"}}}},
+    },
+  };
+}
+```
```
dominiklohmann, line 149: Instead of showing C++ code I would recommend just writing it out in our schema language:

```go
type pcap.packet = record {
  time: timestamp,
  src: addr,
  dst: addr,
  sport: port,
  dport: port,
  vlan: record {
    outer: count,
    inner: count,
  },
  community_id: string #index=hash,
  payload: string #skip,
}
```

web/blog/parquet-and-feather-writing-security-telemetry/index.qmd
```diff
@@ -0,0 +1,868 @@
...
+{"timestamp":"2021-11-17T14:55:32.127935+0100","flow_id":987097466129838,"pcap_cnt":241,"event_type":"fileinfo","src_ip":"198.71.247.91","src_port":80,"dest_ip":"172.70.135.112","dest_port":37220,"proto":"TCP","http":{"hostname":"lifeisnetwork.com","url":"/","http_user_agent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36","xff":"137.135.117.126","http_content_type":"text/html","http_method":"GET","protocol":"HTTP/1.1","status":200,"length":51},"app_proto":"http","fileinfo":{"filename":"/","sid":[],"gaps":false,"state":"CLOSED","stored":false,"size":51,"tx_id":4}}
+```
+
+Note that Zeek's tab-separated value (TSV) format is already a structured table,
+whereas Suricata data needs to be demultiplexed first through the `event_type`
+field.
```
dominiklohmann, line 123: How is that relevant to the analysis? Didn't you just say above that this only focuses on step (4), whereas this takes place during step (1-2)?

web/blog/parquet-and-feather-writing-security-telemetry/index.qmd
```diff
@@ -0,0 +1,868 @@
...
+- **Store**: the type of store plugin used in the measurement, i.e., `parquet`
+  or `feather`.
+- **Construction time**: the time it takes to convert Arrow record batches into
+  Parquet or Feather. We fenced the corresponding code blocks and computed the
+  difference in nanoseconds.
+- **Input size**: the number of bytes that the to-be-converted record batches
+  consume.
+- **Output size**: the number of bytes that the store file takes up.
+- **Number of events**: the total number of events in all input record batches
+- **Number of record batches**
```
dominiklohmann, line 178: Should this have an explanation after it as well?

web/blog/parquet-and-feather-writing-security-telemetry/index.qmd
```diff
@@ -0,0 +1,868 @@
...
+- **Schema**: the name of the schema; there exists one store file per schema
+- **Zstd compression level**: the applied Zstd compression level
+
+Every row corresponds to a single store file where we varied some of these
+parameters. We used [hyperfine](https://github.com/sharkdp/hyperfine) as
+benchmark driver tool, configured with 8 runs. Let's take a look at the data.
```
dominiklohmann, line 184: How did you invoke Hyperfine? The tool has a lot of configuration options. Did you configure warm-up runs? Did you clean the database after every run? This is all valuable information that I think should be included for reproducibility.

web/blog/parquet-and-feather-writing-security-telemetry/index.qmd
```diff
@@ -0,0 +1,868 @@
...
+    "reply_received": "yes"
+  }
+}
+```
+
+This matches our mental model. A few hundred bytes per event with some outliers.
+
+### Batching
+
+On the inside, a store is just a concatenation of Arrow record batches. In
```
dominiklohmann, line 368:
```suggestion
On the inside, a store is just a concatenation of homogeneous Arrow record batches. In
```

Or suffix it with "of the same schema" or so, but this is rather important.

web/blog/parquet-and-feather-writing-security-telemetry/index.qmd
```diff
@@ -0,0 +1,868 @@
...
+Feather they are literally concatenated sequentially, whereas Parquet
+concatenates them column-wise. Most stores only consist of a few record
+batches. PCAP is the only difference.
```
dominiklohmann, line 371: This is not true on either front.

Feather is essentially the wire format of record batches, but that is not a literal concatenation. Instead, schemas and dictionaries are only included when they change, so for our stores just once in the beginning. In order to access a given row in a Feather file, you always need to start at the beginning, iterating over batch by batch until you arrive at the desired batch, which you then materialize before you can access the desired row in the batch via random access.

Parquet has a notion of row groups, which are much like a record batch, except that they are created on write, so their size is set by Parquet rather than the incoming data. Parquet offers random access over both the row groups and within an individual batch that is materialized from a row group. The on-disk layout of Parquet is still row-group by row-group, and in that column by column, so there's no big difference between Parquet and Feather in that regard. Parquet encodes columns using different encoding techniques than Arrow's IPC format.

dominiklohmann (approved):
This is a really good read. I left some comments inline, mostly in the first part of the post. The analysis especially is really thorough, and I enjoyed reading it.

One thing I did not comment, because it would require a larger evaluation, is that you should provide the average batch size used per schema. I think it's fine without it, but given how critical that parameter is for performance we should've included it as well.

web/blog/parquet-and-feather-enabling-open-investigations/index.md
```diff
@@ -4,6 +4,7 @@ authors:
   - mavam
   - dispanser
 date: 2022-10-07
+last_updated: 2022-10-22
```
dominiklohmann, line 7: The generated `index.md` file is not ignored by Git, can you add a `.gitignore` file that just contains `/index.md` into this directory?

mavam: Note that this is the *previous* blog post that is not compiled from a `qmd` file.

I guess it's good and bad that it's in theory indistinguishable from a Quarto notebook. 😅

dominiklohmann: Ah damnit, I just left the comment on the wrong file. I meant to add it to the current blog post.

web/blog/parquet-and-feather-writing-security-telemetry/index.qmd
mavam: It's relevant in the sense of showing the reader how that data looks like in its raw form.

web/blog/parquet-and-feather-writing-security-telemetry/index.qmd
mavam: See the adjacent `benchmark.fish` file in this PR.

web/blog/parquet-and-feather-enabling-open-investigations/index.md
mavam: I'll transpose this comment accordingly.

web/blog/parquet-and-feather-writing-security-telemetry/index.qmd
mavam: Thanks for the detailed explanation. I added that into the section in [4939950](https://github.com/tenzir/vast/pull/2617/commits/49399507cc73a07a82ab793c4a5255e79214f7e5).

mavam: If this reads well, feel free to resolve.

dominiklohmann: Yes, this is great!

web/blog/parquet-and-feather-writing-security-telemetry/index.qmd
dominiklohmann: Fine by me.

web/blog/parquet-and-feather-writing-security-telemetry/index.qmd
dominiklohmann: Can you link to it here?

mavam: It's linked in the previous paragraph. I think it follows naturally when reading this in a flow, which is admittedly different when pausing in review mode.

From my perspective, it's taken care of. If you agree, please resolve.

dominiklohmann: I'd prefer the link in this paragraph directly, but ultimately I have no strong opinion on this. Feel free to resolve.

PR #2618, opened 2022-10-08
Fix race in plugin tests for CTest with -j2 or higher

dominiklohmann:
I just noticed locally when running `ctest --test-dir build -j8` that there was a race for building libvast, which happens because we have test fixtures for vast-test and the individual plugin unit test binaries that ensure that we build the plugins. However, there was a race between the fixture setup the individual plugin unit test binaries in case libvast was not yet built. This change fixes that by making the dependency more explicit, and ensuring that the fixture for the vast-test binary has priority when running CTest.

### :memo: Reviewer Checklist

### DIFF
--- a/cmake/VASTRegisterPlugin.cmake
+++ b/cmake/VASTRegisterPlugin.cmake
@@ -652,11 +652,14 @@ function (VASTRegisterPlugin)
     add_test(NAME build-${PLUGIN_TARGET}-test
              COMMAND "${CMAKE_COMMAND}" --build "${CMAKE_BINARY_DIR}" --config
                      "$<CONFIG>" --target ${PLUGIN_TARGET}-test)
-    set_tests_properties(build-${PLUGIN_TARGET}-test
-                         PROPERTIES FIXTURES_SETUP vast_unit_test_fixture)
+    set_tests_properties(
+      build-${PLUGIN_TARGET}-test
+      PROPERTIES FIXTURES_SETUP vast_${PLUGIN_TARGET}_unit_test_fixture
+                 FIXTURES_REQUIRED vast_unit_test_fixture)
     add_test(NAME ${PLUGIN_TARGET} COMMAND ${PLUGIN_TARGET}-test -v 4 -r 60)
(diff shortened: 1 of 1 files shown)


cmake/VASTRegisterPlugin.cmake
```diff
@@ -652,11 +652,14 @@ function (VASTRegisterPlugin)
     add_test(NAME build-${PLUGIN_TARGET}-test
              COMMAND "${CMAKE_COMMAND}" --build "${CMAKE_BINARY_DIR}" --config
                      "$<CONFIG>" --target ${PLUGIN_TARGET}-test)
-    set_tests_properties(build-${PLUGIN_TARGET}-test
-                         PROPERTIES FIXTURES_SETUP vast_unit_test_fixture)
+    set_tests_properties(
+      build-${PLUGIN_TARGET}-test
+      PROPERTIES FIXTURES_SETUP vast_${PLUGIN_TARGET}_unit_test_fixture
+                 FIXTURES_REQUIRED vast_unit_test_fixture)
```
tobim, line 658: Does that mean that as soon as any test with `FIXTURES_SETUP vast_unit_test_fixture` completed the ones with the corresponding `FIXTURES_REQUIRED` would be unblocked?

In any case I believe a dedicated name for each plugin is much better, but why are do we need `FIXTURES_REQUIRED vast_unit_test_fixture`?

dominiklohmann: Both the `build-${PLUGIN_TARGET}-test` and the `build-vast-test` target attempt to build `libvast`, which introduces a race between the two. This is solved by making `build-${PLUGIN_TARGET}-test` depend on `build-vast-test` with this intermediate per-plugin fixture.

tobim: Ah, because the `build-X` tests aren't modeled in CMakes own dependency graph.

dominiklohmann: They cannot be as they aren't targets, but rather tests, for which you can only declare dependencies via fixtures.

tobim (approved):
The change makes sense.
I didn't test this locally / I wouldn't do anything that isn't already done by CI.

PR #2620, opened 2022-10-10
Fix VPC flowlogs and cloudrail plugins

rdettai:
Story details: https://app.shortcut.com/tenzir/story/38249

# Instructions to reviewer

- the fix is trivial, but I added some CI to actually deploy the flowlogs/cloudtrail datasources to prevent this regression from happening again
- I struggled quite a bit with Github actions to have it work:
  - I split the workflow into multiple jobs to make the different checks more independent (the workflow is now a very pretty graph 😉)
  - this created the need to cache the CLI docker image to avoid having to rebuild it in each job

### DIFF
--- a/.github/workflows/terraform.yaml
+++ b/.github/workflows/terraform.yaml
@@ -136,9 +136,11 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Init
+        run: ./vast-cloud init
+
       - name: Run test playbook
-        run: |
-          ./vast-cloud tests.run
+        run: ./vast-cloud tests.run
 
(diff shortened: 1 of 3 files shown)


rdettai:
[Terraform](https://github.com/tenzir/vast/actions/workflows/terraform.yaml?query=branch%3Astory%2Fsc-38249%2Ffix-datasources)

dispanser (approved):
I can confirm that the changes to the path to the `bucket-notif.sh` fix vpc / cloudtrail.

I can't review the changes for the github action, but I assume a run will tell if it works or not.

PR #2623, opened 2022-10-11
Enable configuration of the zstd compression level for feather store

dispanser:
This came up during work on the parquet / feather benchmarks, where we want to measure the effect of various compression levels for both parquet and feather.

This change follows closely what we do for the parquet plugin to make the zstd compression level configurable.

### :memo: Reviewer Checklist


dispanser:
I guess I still need to add a changelog entry, extend `vast.yaml.example`, and find a place in our documentation to add the parameter.

NB, there is no documentation for parquet plugin configuration options. Where's a good place for that?

dominiklohmann:
> NB, there is no documentation for parquet plugin configuration options. Where's a good place for that?

I'd say under tuning: https://vast.io/docs/setup/tune

libvast/native-plugins/feather.cpp
```diff
@@ -24,6 +25,23 @@
...
+  int64_t zstd_compression_level{9};
+
+  template <class Inspector>
+  friend auto inspect(Inspector& f, configuration& x) {
+    return f(x.zstd_compression_level);
+  }
+
+  static const record_type& layout() noexcept {
+    static auto result = record_type{
+      {"zstd-compression-level", integer_type{}},
```
dominiklohmann, line 39: This is `vast.plugins.feather.zstd-compression-level`, right? I think this should be `vast.compression-level` instead and apply to both plugins at the same time. That'd be a lot easier to communicate in documentation.

dominiklohmann (changes requested):

libvast/native-plugins/feather.cpp
```diff
@@ -24,6 +25,23 @@
 
 namespace vast::plugins::feather {
 
+/// Configuration for the Feather plugin.
+struct configuration {
+  int64_t zstd_compression_level{9};
```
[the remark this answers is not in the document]
dispanser: fair, will change.

libvast/native-plugins/feather.cpp
dispanser: That makes sense. I'm not aware of a way to access the overall system config from within a plugin, though.

dominiklohmann: For any given actor that runs in the server process you can call `content(self->system().config())` to get the configuration as a `caf::settings` object. This may need to be passed in from the `store_actor_plugin` to the `store_plugin`.

libvast/native-plugins/feather.cpp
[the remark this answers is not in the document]
dispanser: Given that the current default compression level is "1", with a single, global setting I would have to make a change for the parquet plugin defaults, assuming it also inherits from this default. I think that would be fine (given a changelog entry).

libvast/native-plugins/feather.cpp
dispanser: Would it make sense to also adapt `chunk::compress` and `arrow_table_slice_builder::create_table_slice` to use this new system-wide default? I'm a bit hesitant b/c tradeoffs for storage vs other data representations may be different / separate concerns.

libvast/native-plugins/feather.cpp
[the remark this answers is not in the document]
dominiklohmann: I think changing it for Parquet has far less impact than changing it for Feather; not sure what we want to do here. @mavam any thoughts?

libvast/native-plugins/feather.cpp
dominiklohmann: I think those are implementation details and should not be touched in the context of this PR.

libvast/native-plugins/feather.cpp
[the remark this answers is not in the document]
mavam: I have no intuition. I'd like to measure this first so that I can make data-driven assumptions.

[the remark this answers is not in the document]
dominiklohmann: Me neither. This is why we currently use what Arrow gives us a default level.

I propose changing Parquet to also make use of Arrow's default level for Zstd, and then creating a follow-up for investigating what level we should actually be using as a default.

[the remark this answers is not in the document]
mavam: Then we are going to compare apples to oranges: the compression level must be equal to understand the delta. I would argue that we use one `zstd-compression-level` that is used for both.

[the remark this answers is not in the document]
dominiklohmann: That's exactly what I suggested.

libvast/native-plugins/feather.cpp
mavam: I think it actually makes sense to have a `zstd-compression-level`. A compression level without a specific algorithm is meaningless. The range is different, some go from 1 to 3, some to 15, etc.

dominiklohmann: Agree on naming it `vast.zstd-compression-level` rather than just `vast.compression-level`, but I still think it shouldn't be `vast.plugins.{parquet,feather}.zstd-compression-level`, as that's just very hidden.

mavam: Agreed as well.

libvast/include/vast/plugin.hpp
```diff
@@ -313,7 +313,8 @@ class store_plugin : public virtual store_actor_plugin {
 
   /// Create a store for active partitions.
   [[nodiscard]] virtual caf::expected<std::unique_ptr<active_store>>
-  make_active_store() const = 0;
+  make_active_store(
+    const caf::dictionary<caf::config_value>& vast_config) const = 0;
```
dominiklohmann, line 317: 1. `caf::dictionary<caf::config_value>` is usually called `caf::settings` (which is an alias for the former, but much more readable)
2. Do we really want to pass in _all_ settings here instead of just the one that is relevant? One of the goals of the `store_plugin` was to make it simple to use over the `store_actor_plugin`.

dispanser: The main argument for passing in all settings is that we don't know which settings might be relevant for a plugin. E.g., if a plugin doesn't use zstd, but a different one, we'd have to extend the interface, and touch all plugin implementations. On the other hand, parquet could possibly use the global `table_slice_size` for producing properly sized chunks when reading parquet, but currently uses its own setting for this.

mavam: I've see the passing-in-all-settings pattern across languages quite frequently. I think it works, actually, given that the configuration is often considered public global state.

changelog/unreleased/2623.md--compression-level-config.md
```diff
@@ -0,0 +1,6 @@
+There is a new setting, `vast.zstd-compression-level`, to control the
+compression level of the zstd algorithm used in both the feather and parquet
+store backend. The default value is `1`, which already was the default for the
+feather store backend, but is a change for parquet, which previously had its
+own default of `9`. We find `1` to be a very good trade-off between performance
+and storage size.
```
dominiklohmann, line 6: The filename for this if off; this will break the changelog generation.

dominiklohmann, line 6: I'm not sure we should mention that the default is 1, because that isn't even what we set it to in the code. The default that we use is the default that Apache Arrow has (or recommends, in a way). It could change in future Arrow versions.

vast.yaml.example
```diff
@@ -128,6 +128,9 @@ vast:
   # store plugin.
   store-backend: feather
 
+  # zstd compression level applied to both feather and parquet store backends.
+  zstd-compression-level: 1
```
dominiklohmann, line 132:
```suggestion
  # Zstd compression level applied to both Feather and Parquet store backends.
  # zstd-compression-level: <default>
```

web/docs/setup/tune.md
```diff
@@ -101,6 +101,23 @@ VAST supports [rebuilding the entire database](#rebuild-partitions) in case you
 want to switch to a different store format. However, VAST works perfectly fine
 with a mixed-storage configuration, so a full rebuild is not required.
 
+### Compression
+
+Partitions in feather and parquet format are compressed using zstd. To fine-tune
+the space-time trade-off, VAST allows you to control the compression level used:
```
dominiklohmann, line 107: Please use active-voice, and start with the problem: As a user, why would I want to change the compression level?

Spelling-wise, Zstd, Feather, and Parquet are usually all capitalized.

libvast/src/plugin.cpp
```diff
@@ -353,7 +353,9 @@ caf::expected<store_actor_plugin::builder_and_header>
 store_plugin::make_store_builder(system::accountant_actor accountant,
                                  system::filesystem_actor fs,
                                  const vast::uuid& id) const {
-  auto store = make_active_store();
+  const auto& vast_config = caf::get<caf::settings>(
+    content(fs->home_system().config()).find("vast")->second);
```
dominiklohmann, line 357: Does this still work if I provide no config at all, or do you de-reference a nullptr then? `find(...)->...` looks rather unsafe to me.

Similarly, what happens if it's a string and not a dictionary? Doesn't `caf::get<caf::settings>` just run into undefined behavior (or an assertion) then?

Since the configuration is user input we need to access it with rather defensive mechanisms.

libvast/native-plugins/feather.cpp
```diff
@@ -159,8 +184,7 @@ class active_feather_store final : public active_store {
     // TODO: Set write_properties.chunksize to the expected batch size
     write_properties.compression = arrow::Compression::ZSTD;
     write_properties.compression_level
-      = arrow::util::Codec::DefaultCompressionLevel(arrow::Compression::ZSTD)
-          .ValueOrDie();
+      = detail::narrow<int>(feather_config_.zstd_compression_level);
```
dominiklohmann, line 187: Do we need to validate here that the level is valid for the codec? Same for Parquet.

libvast/include/vast/plugin.hpp
```diff
@@ -313,7 +313,7 @@ class store_plugin : public virtual store_actor_plugin {
 
   /// Create a store for active partitions.
   [[nodiscard]] virtual caf::expected<std::unique_ptr<active_store>>
-  make_active_store() const = 0;
+  make_active_store(const caf::settings& vast_config) const = 0;
```
dominiklohmann, line 316: Please document the new parameter.

changelog/unreleased/2623.md--compression-level-config.md
dominiklohmann, line 6: Please use active voice and present tense for changelog entries.

dominiklohmann (commented):
I think this is _almost_ there. Happy to see this lands, especially in light of the data analysis from the recent blog post.

libvast/native-plugins/feather.cpp
dispanser: I tested it with out-of-bounds numbers, and it seems to work / accept the nearest valid value.

However, verifying the config is a good idea, but at this particular stage, it's too late IMHO - we should produce a message on startup.

changelog/unreleased/2623.md--compression-level-config.md
dispanser: That's a valid point. However, if we chose to do so then we are shipping a software with a software knob we don't know the default value for and / or it depends on the specific arrow version, which is not exactly user friendly.

dominiklohmann: There are quite a few default values for which we do not know a concrete value. It's just "the default," or "automatically chosen." This applies to quite a few CAF-related tuning options, for example. I think we should be accurate here, and saying "the default level for Apache Arrow" is in my opinion.

libvast/src/plugin.cpp
dispanser:
```
  const auto& vast_config
    = caf::get_or(content(fs->home_system().config()), "vast", caf::settings{});
```

Is probably better, which yields a default / empty settings in case there's nothing called `vast` in there. `caf::get_or` also does the right thing if the config option is not itself a `caf::settings`, so I hope that is sufficiently safe.

dominiklohmann: You could also use `caf::get_if`, which is able to correctly distinguish between "not provided" and "not provided."

I think for this particular one, there's actually an overload of `caf::get_*` that you can use to step through the nesting directly:

```
caf::get_or(settings, "vast.zstd-compression-level", default_zstd_compression_level)
```

dispanser: Indeed, `caf::get_or` also just usees `get_if` under the hood.

The question here is what parts of the config we want to expose to the module. At the moment, the construct above confines the config that is exposed to plugins to the `vast` namespace, which I think is the correct thing to do, as I believe a plugin shouldn't directly deal with `caf` config options.

dominiklohmann: I find this to be very problematic from a DX perspective, because you then can no longer grep for the fully qualified config options. If you pass in a large settings object like this you might as well just pass in everything.

dispanser: Modified accordingly. We now pass in the entire config, and access `vast.zstd-compression-level` in the plugin.

libvast/native-plugins/feather.cpp
dispanser: The last unaddressed point - is there a place where the basic config is verified? I know there's a place where we "translate" legacy config bits...

dominiklohmann: This is probably the correct spot for it.

https://github.com/tenzir/vast/blob/2edd66e81d3e60ea5ced22f2fd30271f54f031ad/vast/vast.cpp#L187-L204

dispanser: done in latest [commit](https://github.com/tenzir/vast/pull/2623/commits/dda026d3e15ee9a07cc4d5b5810b32f9051b16b6). It's a bit convoluted, and also the range according to `arrow::util::Codec::M{in,ax}imumCompressionLevel(arrow::Compression::ZSTD)` is `[-131072, 22]`, so the lower bound is not reliable.

vast/vast.cpp
```diff
@@ -187,18 +187,42 @@ int main(int argc, char** argv) {
...
+      return EXIT_FAILURE;
+    }
+    if (!max_level.ok()) {
+      VAST_ERROR("unable to configure Zstd codec for Apache Arrow: {}",
+                 max_level.status().ToString());
+      return EXIT_FAILURE;
+    }
+    if (compression_level < min_level.ValueUnsafe()
+        || compression_level > max_level.ValueUnsafe()) {
+      VAST_ERROR("zstd compression level '{}' outside of valid range [{}, {}]",
```
dominiklohmann, line 216:
```suggestion
      VAST_ERROR("Zstd compression level '{}' outside of valid range [{}, {}]",
```

changelog/unreleased/2623.md
```diff
@@ -0,0 +1,4 @@
+VAST has a new configuration setting, `vast.zstd-compression-level`, to control
```
dominiklohmann, line 1:
```suggestion
VAST has a new configuration setting `vast.zstd-compression-level` to control
```

changelog/unreleased/2623.md
```diff
@@ -0,0 +1,4 @@
...
+the compression level of the Zstd algorithm used in both the Feather and
+Parquet store backend. The default value is taken from Apache Arrow, which
```
dominiklohmann, line 3:
```suggestion
Parquet store backends. The default value is taken from Apache Arrow, which
```

changelog/unreleased/2623.md
```diff
@@ -0,0 +1,4 @@
...
+changes the default for Parquet, which previously had its own default of `9`.
```
dominiklohmann, line 4: Suggestion for the last sentence to avoid the double "which":

> The default level is set by the Apache Arrow library, and for Parquet is no longer explicitly defaulted to 9.

dominiklohmann (approved):
Just wordsmithing now. I think this is ready to go! 🚀

PR #2627, opened 2022-10-12
Allow cross-origin requests in dev mode

lava:
A minimal change to the web plugin to allow development against separate API and frontend servers. The mechanism may be reused to insert other relevant headers in the future.

For production usage we probably need to at least make the allowed origins user-configurable, but it's not clear yet when this will be needed since we also expect to be serving the frontend and the API from the same origin eventually.

### :memo: Reviewer Checklist

### DIFF
--- a/plugins/web/src/server_command.cpp
+++ b/plugins/web/src/server_command.cpp
@@ -164,7 +164,7 @@ request_dispatcher_actor::behavior_type request_dispatcher(
 
 void setup_route(caf::scoped_actor& self, std::unique_ptr<router_t>& router,
                  std::string_view prefix, request_dispatcher_actor dispatcher,
-                 vast::rest_endpoint endpoint,
+                 const server_config& config, vast::rest_endpoint endpoint,
                  system::rest_handler_actor handler) {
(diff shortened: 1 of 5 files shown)


plugins/web/src/server_command.cpp
```diff
@@ -164,7 +164,7 @@ request_dispatcher_actor::behavior_type request_dispatcher(
 
 void setup_route(caf::scoped_actor& self, std::unique_ptr<router_t>& router,
                  std::string_view prefix, request_dispatcher_actor dispatcher,
-                 vast::rest_endpoint endpoint,
+                 const server_config& config, vast::rest_endpoint endpoint,
```
mavam, line 167: Could you adapt the example config with a commented illustration how to use this?

lava: Right now there's nothing exposed to the user, we just always allow arbitrary origins in dev mode.

dit7ya (approved):
I can confirm building from this branch fixes the CORS issues in the frontend - Chrome and Firefox both works fine.

PR #2631, opened 2022-10-13
Properly recover corrupted partition data on startup

lava:
This contains two fixes:

  1) Fix the check for the flatbuffer identifier at the
     beginning of a partition flatbuffer to look at the
     correct bytes.

  2) Instead of running the recovered data through import,
     use a partition transformer to recreate the partitions.
     This avoids creating partitions with huge import time
     ranges by mixing the data with regular imports, which
     would confuse compaction and rebuilding.

Testing: We manually "corrupted" partitions by using `truncate -s 3G` on some partitions that were created by a static build of VAST 2.2, and verified that the data from these was present and queryable after a `vast start`. I'm not sure how we would do automated testing of this.

### :memo: Reviewer Checklist


libvast/src/system/index.cpp
```diff
@@ -555,47 +556,75 @@ caf::error index_state::load_from_disk() {
...
+      }
+      std::error_code err{};
+      if (!std::filesystem::remove(store_path, err))
+        VAST_WARN("{} failed to remove store file {} after recovery: {}", index,
+                  store_path, err);
+      if (!std::filesystem::remove(part_path, err))
+        VAST_WARN("{} failed to remove partition file {} after recovery: {}",
+                  index, part_path, err);
+      for (auto slice : *seg)
+        self->send(transformer, std::move(slice));
```
tobim, line 606: Sending all slices of all oversized partitions into one transformer might materialize a lot of data and risking large memory consumption.

tobim: We agreed to run one transformation per partition in a call.

dominiklohmann:
[sc-38413]

tobim (approved):
We paired most of the time for this fix and the unit test.

PR #2635, opened 2022-10-14
Add a timeout to the UDS metric sink

tobim:
We now wait for up to one second while trying to write metrics, but only if the previous line was not dropped.
Additionally, the file descriptor used for sending is put into non-blocking mode to avoid endless blocking in case the listening socket hangs up but isn't cleaned up fully.

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/include/vast/detail/posix.hpp
+++ b/libvast/include/vast/detail/posix.hpp
@@ -38,6 +38,10 @@ struct uds_datagram_sender {
   static caf::expected<uds_datagram_sender> make(const std::string& path);
 
   /// Send the content of `data to `dst`.
+  /// @param data The data that should be written.
+  /// @param timeout_usec A duration to wait for readiness of the receiver.
+  /// @returns `true` if the data was sent or `false` if it was dropped because
+  ///          the timeout was reached
   caf::expected<bool> send(std::span<char> data, int timeout_usec = 0);
 
   /// The file descriptor for the "client" socket.
--- a/libvast/include/vast/detail/posix.hpp
+++ b/libvast/include/vast/detail/posix.hpp
@@ -40,9 +40,9 @@ struct uds_datagram_sender {
   /// Send the content of `data to `dst`.
   /// @param data The data that should be written.
   /// @param timeout_usec A duration to wait for readiness of the receiver.
-  /// @returns `true` if the data was sent or `false` if it was dropped because
-  ///          the timeout was reached
-  caf::expected<bool> send(std::span<char> data, int timeout_usec = 0);
+  /// @returns `no_error` if the data was sent or `timeout` if it was dropped
+  ///          because the timeout was reached
+  caf::error send(std::span<char> data, int timeout_usec = 0);
 
   /// The file descriptor for the "client" socket.
   int src_fd = -1;
--- a/libvast/include/vast/detail/posix.hpp
+++ b/libvast/include/vast/detail/posix.hpp
@@ -38,7 +38,7 @@ struct uds_datagram_sender {
   static caf::expected<uds_datagram_sender> make(const std::string& path);
 
   /// Send the content of `data to `dst`.
-  caf::error send(std::span<char> data);
+  caf::expected<bool> send(std::span<char> data, int timeout_usec = 0);
 
   /// The file descriptor for the "client" socket.
   int src_fd = -1;
@@ -100,8 +100,8 @@ struct [[nodiscard]] unix_domain_socket {
   /// @param path The filesystem path identifying the server socket.
   /// @param type The socket type.
   /// @returns A UNIX domain socket handle.
-  static unix_domain_socket connect(const std::string& path,
-                                    socket_type type = socket_type::stream);
+  static unix_domain_socket
+  connect(const std::string& path, socket_type type = socket_type::stream);
 
   /// Checks whether the UNIX domain socket is in working state.
   /// @returns `true` if the UNIX domain socket is open and operable.
--- a/libvast/include/vast/detail/posix.hpp
+++ b/libvast/include/vast/detail/posix.hpp
@@ -134,8 +134,14 @@ struct [[nodiscard]] unix_domain_socket {
 /// Polls a file descriptor for ready read events via `select(2)`.
 /// @param fd The file descriptor to poll
 /// @param usec The number of microseconds to wait.
-/// @returns `caf::none` if *fd* has ready events for reading.
-[[nodiscard]] caf::error poll(int fd, int usec = 100000);
+/// @returns `true` if read operations on *fd* would not block.
+[[nodiscard]] caf::expected<bool> rpoll(int fd, int usec);
+
+/// Polls a file descriptor for write readiness via `select(2)`.
+/// @param fd The file descriptor to poll
+/// @param usec The number of microseconds to wait.
+/// @returns `true` if write operations on *fd* would not block.
+[[nodiscard]] caf::expected<bool> wpoll(int fd, int usec);
 
 /// Wraps `close(2)`.
 /// @param fd The file descriptor to close.
--- a/libvast/src/detail/posix.cpp
+++ b/libvast/src/detail/posix.cpp
@@ -11,7 +11,6 @@
 #include "vast/config.hpp"
 #include "vast/detail/assert.hpp"
 #include "vast/detail/raise_error.hpp"
-#include "vast/die.hpp"
 #include "vast/error.hpp"
 #include "vast/logger.hpp"
 
@@ -327,26 +326,28 @@ caf::error make_blocking(int fd) {
   return make_nonblocking(fd, false);
 }
 
-caf::error poll(int fd, int usec) {
+caf::expected<bool> rpoll(int fd, int usec) {
   fd_set rdset;
   FD_ZERO(&rdset);
   FD_SET(fd, &rdset);
   timeval timeout{0, usec};
   auto rc = ::select(fd + 1, &rdset, nullptr, nullptr, &timeout);
-  if (rc < 0) {
-    switch (rc) {
-      default:
-        vast::die("unhandled select(2) error");
-      case EINTR:
-      case ENOMEM:
-        return caf::make_error(ec::filesystem_error,
-                               "failed in select(2):", std::strerror(errno));
-    }
-  }
-  if (!FD_ISSET(fd, &rdset))
+  if (rc < 0)
     return caf::make_error(ec::filesystem_error,
-                           "failed in fd_isset(3):", std::strerror(errno));
-  return caf::none;
+                           "failed in select(2):", std::strerror(errno));
+  return !!FD_ISSET(fd, &rdset);
+}
+
+caf::expected<bool> wpoll(int fd, int usec) {
+  fd_set write_set;
+  FD_ZERO(&write_set);
+  FD_SET(fd, &write_set);
+  timeval timeout{0, usec};
+  auto rc = ::select(fd + 1, nullptr, &write_set, nullptr, &timeout);
+  if (rc < 0)
+    return caf::make_error(ec::filesystem_error,
+                           "failed in select(2):", std::strerror(errno));
+  return !!FD_ISSET(fd, &write_set);
 }
 
 caf::error close(int fd) {
--- a/libvast/src/detail/posix.cpp
+++ b/libvast/src/detail/posix.cpp
@@ -132,22 +132,21 @@ uds_datagram_sender::make(const std::string& path) {
   return std::move(result);
 }
 
-caf::expected<bool>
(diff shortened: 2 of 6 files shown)


libvast/src/detail/posix.cpp
```diff
@@ -318,26 +330,28 @@ caf::error make_blocking(int fd) {
   return make_nonblocking(fd, false);
 }
 
-caf::error poll(int fd, int usec) {
+caf::expected<bool> rpoll(int fd, int usec) {
   fd_set rdset;
```
mavam, line 334: Can you call this `read_set` so that the implementation is symmetric to the `write_set` you introduced below?

libvast/src/system/accountant.cpp
```diff
@@ -93,6 +93,9 @@ struct accountant_state_impl {
   /// Handle to the uds output channel.
   std::unique_ptr<detail::uds_datagram_sender> uds_datagram_sink = nullptr;
 
+  /// Handle to the uds output channel is currently dropping it's input.
```
mavam, line 96:
```suggestion
  /// Handle to the UDS output channel is currently dropping its input.
```

mavam (commented):
This probably works, but I don't understand the reason behind polling the socket conditionally. Could you elaborate?

Why not poll unconditionally? It seems we're now at least dropping one line with higher probability.

tobim:
> Why not poll unconditionally? It seems we're now at least dropping one line with higher probability.

If we wait every time and run into timeouts constantly the message inbox of the accountant will grow as soon as the average time between messages is lesser than the timeout. Since the inbox is an unbounded buffer it would run out of memory eventually, which is exactly the situation that this change is supposed to fix.

libvast/src/system/accountant.cpp
```diff
@@ -216,7 +219,20 @@ struct accountant_state_impl {
...
                           detail::uds_datagram_sender& dest) {
     auto buf = to_json_line(actor_id, ts, key, x, meta1, meta2);
-    dest.send(std::span<char>{reinterpret_cast<char*>(buf.data()), buf.size()});
+    // Only poll the socket for write readiness if it did not drop the previous
+    // line.
+    auto timeout_usec = uds_datagram_sink_dropping ? 0 : 1'000'000;
+    auto delivered = dest.send(
+      std::span<char>{reinterpret_cast<char*>(buf.data()), buf.size()},
+      timeout_usec);
+    if (!delivered) {
```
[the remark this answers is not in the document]
tobim: I'm not 100% sure, but I think calling `.error()` on an `expected` that doesn't contain one is UB.
Besides, we use `operator!` for this everywhere else.

libvast/src/system/accountant.cpp
```diff
@@ -216,7 +219,20 @@ struct accountant_state_impl {
                           const metrics_metadata& meta2, time ts,
                           detail::uds_datagram_sender& dest) {
     auto buf = to_json_line(actor_id, ts, key, x, meta1, meta2);
-    dest.send(std::span<char>{reinterpret_cast<char*>(buf.data()), buf.size()});
+    // Only poll the socket for write readiness if it did not drop the previous
+    // line.
+    auto timeout_usec = uds_datagram_sink_dropping ? 0 : 1'000'000;
```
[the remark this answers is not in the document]
tobim: I extended the comment, hopefully to your satisfaction.

I opted to skip the `std::chrono` with the assumption that future readers will understand that `1'000'00` microseconds are the same as one second.

libvast/src/system/accountant.cpp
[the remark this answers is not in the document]
mavam: Yeah, that would make sense actually.

libvast/include/vast/detail/posix.hpp
```diff
@@ -38,7 +38,11 @@ struct uds_datagram_sender {
   static caf::expected<uds_datagram_sender> make(const std::string& path);
 
   /// Send the content of `data to `dst`.
-  caf::error send(std::span<char> data);
+  /// @param data The data that should be written.
+  /// @param timeout_usec A duration to wait for readiness of the receiver.
+  /// @returns `true` if the data was sent or `false` if it was dropped because
+  ///          the timeout was reached
+  caf::expected<bool> send(std::span<char> data, int timeout_usec = 0);
```
mavam, line 45: I'm still unsure that `expected<bool>` is the right thing here. It feels like an anti-pattern, for two reasons:

1. There's also `caf::no_error`, which aliases to an error code with 0. The semantics are basically "not an error".
2. There is probably an error code for a timeout that makes sense to return.

Keeping it at `caf::error` and returning a timeout error code feels more appropriate. At least from the comment of the return suggests that.

libvast/src/system/accountant.cpp
```diff
@@ -216,7 +219,26 @@ struct accountant_state_impl {
                           const metrics_metadata& meta2, time ts,
                           detail::uds_datagram_sender& dest) {
     auto buf = to_json_line(actor_id, ts, key, x, meta1, meta2);
-    dest.send(std::span<char>{reinterpret_cast<char*>(buf.data()), buf.size()});
+    // Only poll the socket for write readiness if it did not drop the previous
+    // line. Not doing this could increase the average message processing time
+    // of the accountant by timeout value, and just as with any actor, receiving
```
mavam, line 224:
```suggestion
    // of the accountant by the timeout value, and just as with any actor, receiving
```

libvast/include/vast/detail/posix.hpp
mavam: I don't have strong opinions on this though. Feel free to resolve if you think it aint't worth it to go through another round.

libvast/src/detail/posix.cpp
```diff
@@ -131,12 +132,23 @@ uds_datagram_sender::make(const std::string& path) {
   return std::move(result);
 }
 
-caf::error uds_datagram_sender::send(std::span<char> data) {
+caf::expected<bool>
+uds_datagram_sender::send(std::span<char> data, int timeout_usec) {
+  if (timeout_usec) {
```
mavam, line 137:
```suggestion
  if (timeout_usec > 0) {
```

libvast/src/system/accountant.cpp
```diff
@@ -216,7 +219,26 @@ struct accountant_state_impl {
...
+    // any re-initialization of the listening side would not incur data loss
+    // without being excessive.
+    auto timeout_usec = uds_datagram_sink_dropping ? 0 : 1'000'000;
+    auto delivered = dest.send(
+      std::span<char>{reinterpret_cast<char*>(buf.data()), buf.size()},
+      timeout_usec);
+    if (!delivered) {
+      VAST_WARN("{} failed to write metrics to UDS sink: {}", *self,
+                delivered.error());
+      VAST_WARN("{} disables the UDS metrics sink", *self);
```
mavam, line 237:
```suggestion
      VAST_WARN("{} resets the UDS metrics sink", *self);
```

mavam:
> If we wait every time and run into timeouts constantly the message inbox of the accountant will grow as soon as the average time between messages is lesser than the timeout. Since the inbox is an unbounded buffer it would run out of memory eventually, which is exactly the situation that this change is supposed to fix.

My mental model is this:

0. The common case is that VAST produces less that than the other side can consume
1. We want to be able to tolerate outages on the other end, for brief amounts of time
2. If the socket becomes unavailable, we accept a small amount of wait time
3. In the common case, `wpoll` is a no-op because the socket is always ready

By this logic, it doesn't hurt to poll unconditionally. It seems this scenario of intermittent outages is not reflected, rather, we're punishing directly by dropping a message on the floor immediately.

So only for the non-common case of *continuous* outage, we need to drop messages. Does that make sense?

libvast/include/vast/detail/posix.hpp
tobim: To me the it makes most sense to return a 3-state type. Because I want to handle timeouts separately from errors at the call site. No strong opinion either though.

libvast/src/system/accountant.cpp
tobim: Isn't "disable" clearer here? If it says "reset" I would ask to what state?

libvast/include/vast/detail/posix.hpp
dominiklohmann: It's a four-state type right now, though:

- true
- false
- caf::no_error
- caf::error

I think for legibility of the code it makes sense to just have this return an enum class or a variant.

tobim:
> > If we wait every time and run into timeouts constantly the message inbox of the accountant will grow as soon as the average time between messages is lesser than the timeout. Since the inbox is an unbounded buffer it would run out of memory eventually, which is exactly the situation that this change is supposed to fix.
> 
> My mental model is this:
> 
>     0. The common case is that VAST produces less that than the other side can consume
> 
>     1. We want to be able to tolerate outages on the other end, for brief amounts of time
> 
>     2. If the socket becomes unavailable, we accept a small amount of wait time

That is what we do.

>     3. In the common case, `wpoll` is a no-op because the socket is always ready

That doesn't make it a no-op. In fac t, I'm going to change it to try to send first, and only `wpoll` if we get `EAGAIN`. No need to introduce an unforced pessimization here.

> By this logic, it doesn't hurt to poll unconditionally. It seems this scenario of intermittent outages is not reflected, rather, we're punishing directly by dropping a message on the floor immediately.

Intermittent outages *are* covered. "Intermittent" in the sense that we can't send every so often, and it recovers relatively quickly. The logic to wait only in the good state has no effect in this scenario. What it does cover is the case of being unable to send for a longer time.

> So only for the non-common case of _continuous_ outage, we need to drop messages. Does that make sense?

Well, yes, that is exactly what this code does.

libvast/src/system/accountant.cpp
mavam: I am thinking of "reset" as something that makes starts again from the beginning. "Disabling" to me sounds like it's no longer available.

mavam (approved):
Unless @dominiklohmann has some mechanical feedback on this PR, I'm giving this greenlight modulo pending discussion threads.

dominiklohmann:
> Unless @dominiklohmann has some mechanical feedback on this PR, I'm giving this greenlight modulo pending discussion threads.

I don't know that much about this mechanism as you do, my only concern is that `expected<bool>` is a really bad return type that we should never use anywhere, but I already commented that.

mavam:
> I don't know that much about this mechanism as you do, my only concern is that `expected<bool>` is a really bad return type that we should never use anywhere, but I already commented that.

Yes, I also think we should revert this to `caf::error` with timeout being a specific error code.

dominiklohmann:
Personal opinion: `caf::expected<void>` is a strictly better `caf::error` as a function return type, because it has bool-esque semantics, so I think that'd be even better.

tobim:
Here is what the call sites look like when using `caf::error` or `caf::expected<void>` respectively:
* `error`
```c++
    if (auto err = dest.send(
          std::span<char>{reinterpret_cast<char*>(buf.data()), buf.size()},
          timeout_usec)) {
      if (err == ec::timeout)
        uds_datagram_sink_dropping = true;
      else {
        VAST_WARN("{} failed to write metrics to UDS sink: {}", *self,
                  success.error());
        VAST_WARN("{} disables the UDS metrics sink", *self);
        uds_datagram_sink.reset();
        return;
      }
    }
```
* `expected<void>`
```c++
    if (auto success = dest.send(
          std::span<char>{reinterpret_cast<char*>(buf.data()), buf.size()},
          timeout_usec);
        !success) {
      if (success.error() == ec::timeout)
        uds_datagram_sink_dropping = true;
      else {
        VAST_WARN("{} failed to write metrics to UDS sink: {}", *self,
                  success.error());
        VAST_WARN("{} disables the UDS metrics sink", *self);
        uds_datagram_sink.reset();
        return;
      }
    }
```
Both seem to be downgrades over `caf::expected<bool>`. The main issue I have with it is that I consider timeout handling _for a function that takes a timeout_ as regular control flow, not error handling.

mavam:
> The main issue I have with it is that I consider timeout handling _for a function that takes a timeout_ as regular control flow, not error handling.

Why not switch-case on `caf::error` as return value? That's idiomatic and doesn't require shoehorning the logic into a binary if-else.

tobim:
> Why not switch-case on caf::error as return value? That's idiomatic and doesn't require shoehorning the logic into a binary if-else.

I implemented that now so we can move on.
I'll still push another updated to try to send immediately so we only have to go through `wpoll` if that didn't work.

mavam (approved):

libvast/src/system/accountant.cpp
tobim: It does turn off. If we get a hard error there is no point in trying again.

mavam: Okay, then please disregard my suggestion.

PR #2636, opened 2022-10-16
Install Python bindings along with VAST

dominiklohmann:
This changes our CMake to install the Python bindings if Poetry is available at configure-time.


Dockerfile
```diff
@@ -48,6 +50,7 @@ RUN echo 'deb http://deb.debian.org/debian bullseye-backports main' \
       ./apache-arrow-apt-source-latest-$(lsb_release --codename --short).deb && \
     apt-get update && \
     apt-get -y --no-install-recommends install libarrow-dev=9.0.0-1 libprotobuf-dev libparquet-dev=9.0.0-1 && \
+    python3 -m pip install --upgrade poetry && \
```
rdettai, line 53: we should validate that there is no conflict with the Poetry installed in https://github.com/tenzir/vast/pull/2625

dominiklohmann:
I've marked this as blocked; we agreed earlier in the team to only make this change _after_ the 2.4 release.

Dockerfile
dominiklohmann: I verified this locally.

tobim:
If I run this locally I get the python stuff installed in `lib` and but `libvast.so` & co in `lib64`.
Also, adding the installation path to the PYTHONPATH is not enough to get `import vast` to work:
```
$ PYTHONPATH=$PW
D/lib/python3.10/site-packages:$PYTHONPATH python
Python 3.10.8 (main, Oct 11 2022, 11:35:05) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import stix2
>>> import vast
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'vast'
```
I believe that is because the `vast` library is actually not installed:
```
$ ls -1 lib/python3.10/site-packages/
...
stix2/
stix2-3.0.1.dist-info/
stix2patterns/
stix2_patterns-2.0.0.dist-info/
urllib3/
urllib3-1.26.13.dist-info/
vast-2.4.1.dist-info/
```

dominiklohmann:
> If I run this locally I get the python stuff installed in `lib` and but `libvast.so` & co in `lib64`.

I fixed a change that should align this. Can you run your test again?

CMakeLists.txt
```diff
@@ -891,6 +891,63 @@ install(
...
+    CODE "
+    file(REMOVE_RECURSE \"\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR}/vast/python\")
+    file(MAKE_DIRECTORY \"\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR}/vast/python\")
+    file(
+      COPY \"${CMAKE_CURRENT_SOURCE_DIR}/python/\"
+      DESTINATION \"\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR}/vast/python\")
+    execute_process(
+      COMMAND \"${POETRY_PATH}\" build --no-interaction --format=wheel
+      WORKING_DIRECTORY \"\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_DATADIR}/vast/python\"
+      COMMAND_ERROR_IS_FATAL ANY)
```
tobim, line 937: I think we should put this into the build tree / do it as part of the build.
It can also be shortened to
```
      COMMAND \"${POETRY_PATH}\" build --no-interaction --format=wheel --output=\"${CMAKE_CURRENT_BUILD_DIR}/python\"
```
once https://github.com/python-poetry/poetry/pull/7208/files is merged an released.

tobim:
The `lib` ,location is actually correct, see https://serverfault.com/a/88739.
The other thing seems still wrong, `import vast` doesn't work.

tobim (approved):
I can confirm that all dependencies are installed in the correct place.

PR #2641, opened 2022-10-19
Add a --web-root option to the web plugin

lava:
### :memo: Reviewer Checklist

### DIFF
--- a/plugins/web/src/server_command.cpp
+++ b/plugins/web/src/server_command.cpp
@@ -280,22 +280,26 @@ auto server_command(const vast::invocation& inv, caf::actor_system& system)
       return request->create_response(restinio::status_temporary_redirect())
         .append_header(restinio::http_field::server, "VAST")
         .append_header_date_field()
-        .append_header(restinio::http_field::location, "/status.html")
+        .append_header(restinio::http_field::location, "/index")
         .done();
     });
-  VAST_VERBOSE("using {} as ", server_config->webroot);
+  VAST_VERBOSE("using {} as document root", server_config->webroot);
   router->http_get(
     "/:path(.*)", restinio::path2regex::options_t{}.strict(true),
     [webroot = server_config->webroot](auto req, auto /*params*/) {
       auto ec = std::error_code{};
       auto http_path = req->header().path();
-      VAST_DEBUG("serving static file {}", http_path);
       auto path = std::filesystem::path{std::string{http_path}};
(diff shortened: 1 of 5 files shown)


dit7ya (commented):
~~What do you think about calling it `--web-ui-dir` or `--web-ui-root` instead of `--web--root`?~~

dit7ya:
Also, that CORS error should go away with this patch

```diff
diff --git a/src/routes/status/+page.svelte b/src/routes/status/+page.svelte
index 9bbed5b..07c081d 100644
--- a/src/routes/status/+page.svelte
+++ b/src/routes/status/+page.svelte
@@ -14,7 +14,7 @@
 
 	const getStatus = async () => {
 		try {
-			const API_BASE = 'http://localhost:42001/api/v0';
+			const API_BASE = 'http://127.0.0.1:42001/api/v0';
 			const url = `${API_BASE}/status?verbosity=detailed`;
 			const response = await fetch(url);
```

plugins/web/src/server_command.cpp
```diff
@@ -277,15 +278,43 @@ auto server_command(const vast::invocation& inv, caf::actor_system& system)
       return request->create_response(restinio::status_temporary_redirect())
         .append_header(restinio::http_field::server, "VAST")
         .append_header_date_field()
-        .append_header(restinio::http_field::location, "/api/v0/status")
+        .append_header(restinio::http_field::location, "/status.html")
```
dit7ya, line 281: I think this is not what we want. Routing is being fully managed in the frontend. So this is creating a `/status.html` route instead of the better-looking `status` (and is also messing with the frontend router). The backend should just serve the webroot directory at some endpoint, everything after that is frontend's responsibility.

mavam:
Historically, the "web root" is the place where a web server places its file, so `--web-root` to me feels like a well-known thing. I have no strong opinions on calling it differently though.

dit7ya:
My mistake. I thought `web/docs`, `web/blog` and `plugins/web/ui` are at the same level, and are also served from the CLI - so I wanted to disambiguate. But since that is not the case, `--web-root` is perfectly fine.

lava:
I changed it to just `--root` for now, since that's what nginx also uses in their configuration and there doesn't seem to be consensus between "document root" and "web root" between different web servers.

tobim (approved):

lava:
I "merged" this manually by rebasing and pushing onto the base branch, to keep the commit graph a bit simpler by avoiding multiple merge bases.

PR #2642, opened 2022-10-19
Clear failed partitions from the cache

lava:
### :memo: Reviewer Checklist

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -986,6 +986,12 @@ void index_state::schedule_lookups() {
            pid = next->partition](const caf::error& err) {
             VAST_WARN("{} failed to evaluate query {} for partition {}: {}",
                       *self, qid, pid, err);
+            // We don't know if this was a transient error or if the
+            // partition/store is corrupted. However, the partition actor has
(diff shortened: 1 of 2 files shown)


dominiklohmann (approved):
Thanks for fixing this, I did not think this would be as simple.

I do not know a good way to test this, but we should see during the v2.4 rc-phase whether it still occurs on our testbed.

libvast/src/system/index.cpp
```diff
@@ -986,6 +986,12 @@ void index_state::schedule_lookups() {
            pid = next->partition](const caf::error& err) {
             VAST_WARN("{} failed to evaluate query {} for partition {}: {}",
                       *self, qid, pid, err);
+            // We don't know if this was a transient error or if the
+            // partition/store is corrupted. However, the partition actor has
+            // possibly already exited so at least we have to clear it from the
+            // cache so that subsequent queries get a chance to respawn it
+            // cleanly instead of trying to talk to the dead.
+            inmem_partitions.drop(pid);
```
dominiklohmann, line 994: Can you add a changelog entry for this? This existed since 2.0 IIUC.

PR #2647, opened 2022-10-24
Make the expression parser composable

dominiklohmann:
Before this change to the expression parser the newly added test failed because the expression parser required that it ended at the end of parseable input.

Should not need a changelog entry, as this was not user-facing so far. It's relevant for being able to implement pipelines, though.

### :memo: Reviewer Checklist


dispanser (approved):
This change looks easy enough and the test convincing enough that I take the liberty to approve it.

FWIW I tried adding a random string at the end of an otherwise valid expression, which did not work much to my surprise ;-)

dominiklohmann:
@dispanser that works because when we parse it we put in a range rather than a pair of iterators, and that appends the eoi parser by default.

PR #2648, opened 2022-10-24
Provide a formatting overload of make_error

tobim:
Error messages produced from `caf::make_error` aren't formatted with fmt automatically, that means you have to call `fmt::format` to produce a dynamic message as the second argument to `make_error`. Sometimes this is forgotten, leading to strange looking error messages that consist of the verbatim format string followed by the list of arguments.

This annoyed me several times already, and this time was one time to many. So instead of fixing another local occurence I added a function that does the right thing.

I only switched over the buggy call sites for now, but I think we should switch over the codebase eventually for consistency.

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/include/vast/error.hpp
+++ b/libvast/include/vast/error.hpp
@@ -12,6 +12,7 @@
 
 #include <caf/error.hpp>
 #include <caf/make_message.hpp>
+#include <fmt/core.h>
 
 namespace vast {
 
@@ -95,6 +96,12 @@ const char* to_string(ec x);
 /// A formatting function that converts an error into a human-readable string.
 /// @relates ec
 std::string render(caf::error err);
+
+template <class... T>
+caf::error make_error(ec code, fmt::format_string<T...> fmt, T&&... args) {
+  return caf::make_error(code, fmt::format(std::move(fmt),
+                                           std::forward<T&&>(args)...));
+}
 
 } // namespace vast
 
--- a/libvast/include/vast/error.hpp
+++ b/libvast/include/vast/error.hpp
@@ -97,11 +97,20 @@ const char* to_string(ec x);
 /// @relates ec
 std::string render(caf::error err);
 
-template <class... T>
-caf::error make_error(ec code, fmt::format_string<T...> fmt, T&&... args) {
+/// An error factory that delegates message creation to fmt::format.
+/// @relates ec
+#if FMT_VERSION >= 80000
+template <class... Ts>
+caf::error make_error(ec code, fmt::format_string<Ts...> fmt, Ts&&... args) {
   return caf::make_error(code, fmt::format(std::move(fmt),
-                                           std::forward<T&&>(args)...));
+                                           std::forward<Ts&&>(args)...));
 }
+#else
+template <class S, class... Ts>
+caf::error make_error(ec code, const S& fmt, Ts&&... args) {
+  return caf::make_error(code, fmt::format(fmt, std::forward<Ts&&>(args)...));
+}
+#endif
 
 } // namespace vast
 
--- a/libvast/src/system/node_control.cpp
+++ b/libvast/src/system/node_control.cpp
@@ -45,9 +45,8 @@ spawn_at_node(caf::scoped_actor& self, const node_actor& node, invocation inv) {
         result = std::move(actor);
       },
       [&](caf::error& err) {
-        result
-          = caf::make_error(ec::unspecified, "failed to spawn '{}' at node: {}",
-                            inv.full_name, err);
+        result = vast::make_error(ec::unspecified, "failed to {} at node: {}",
+                                  inv.full_name, err);
       });
   return result;
 }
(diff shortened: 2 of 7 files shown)


libvast/include/vast/error.hpp
```diff
@@ -96,6 +97,21 @@ const char* to_string(ec x);
...
+caf::error make_error(ec code, fmt::format_string<Ts...> fmt, Ts&&... args) {
+  return caf::make_error(code, fmt::format(std::move(fmt),
+                                           std::forward<Ts&&>(args)...));
+}
+#else
+template <class S, class... Ts>
+caf::error make_error(ec code, const S& fmt, Ts&&... args) {
+  return caf::make_error(code, fmt::format(fmt, std::forward<Ts&&>(args)...));
+}
+#endif
```
dominiklohmann, line 113: Would you mind going one step further and actually poisoning `caf::make_error` when called with `vast::ec` as a first argument type?

```c++
namespace caf {

/// Poison `caf::make_error` for `vast::ec` to ensure users go through the
/// properly formatted version `vast::make_error`.
template <class... Args>
error make_error(::vast::ec, Args&&...) = delete;

} // namespace caf
```

I really like the idea of this, but I think it'd make sense to just do this once and for all so we don't end up with a mixed style in the code base.

libvast/src/system/node_control.cpp
```diff
@@ -45,9 +45,8 @@ spawn_at_node(caf::scoped_actor& self, const node_actor& node, invocation inv) {
         result = std::move(actor);
       },
       [&](caf::error& err) {
-        result
-          = caf::make_error(ec::unspecified, "failed to spawn '{}' at node: {}",
-                            inv.full_name, err);
+        result = vast::make_error(ec::unspecified, "failed to {} at node: {}",
```
dominiklohmann, line 48:
```suggestion
        result = make_error(ec::unspecified, "failed to {} at node: {}",
```

Same goes for everywhere else—not sure why we're explicitly specifying `vast::` from within the VAST code base now. Just feels very noisy to me.

(Obviously a style thing, so YMMV, and my opinion is not too strong on this. I just like consistency.)

dominiklohmann (commented):
This is a really great effort. I have some questions on this.

libvast/src/system/node_control.cpp
tobim: I'd like to drop the qualifier, but unfortunately it seems to be required. Thanks to ADL the compiler finds the overload from the `caf` namespace and prefers that because it doesn't require a conversion for the second argument.

If we change our own overload to
```c++
template <class... Args>
caf::error(ec code, Args&&... args) {
  if constexpr (sizeof...(Args) == 0)
    return caf::error{code};
  else
    return caf::error{code, fmt::format(std::forward<Args&&>(args)...)};
}
```
`fmt` complains about the format string isn't constant (which is wrong). I haven't looked if that can be worked around yet. Maybe you know?

dominiklohmann: If you do the poisoning I suggested you get the correct one via ADL, always.

I think you just need a separate overload for the zero-argument version, and then have one that takes 1+ arguments, with the one being the {fmt} format string. I believe there is a `fmt::vformat*` family of functions you can use for inspiration on the format string argument.

tobim: > If you do the poisoning I suggested you get the correct one via ADL, always.

No you don't, it will take the specialized one. The `= delete` has no effect on the overload set.

tobim: `fmt::vformat` should work, but that isn't compile-time checked.

dominiklohmann: vformat itself is not, but you want to check only once anyways, and that you do via the template arguments, by making sure you take a `fmt::format_string<Ts...>` and then matching `Ts...`.

Regarding ADL, shouldn't it always take the overload in namespace VAST because of `vast::ec` anyways? I don't understand why it wouldn't work without the full qualification.

libvast/include/vast/error.hpp
dominiklohmann: Thinking about this some more, I think we definitely must do this before we can merge this PR. It's a very bad DX to have two make_error functions with such different behavior.

libvast/src/system/node_control.cpp
lava: > Regarding ADL, shouldn't it always take the overload in namespace VAST because of vast::ec anyways?

ADL only adds to the overload set, for `vast::ec` this doesn't make a difference because we're in namespace VAST anyways. But it doesn't have an effect on overload resolution, if `caf::make_error` is visible and a better match it will be selected.

dominiklohmann: It shouldn't be a better match here, though, if the first argument to the function is in the vast namespace.

tobim: Best idea I have been able to come up with is to name our version of the function differently. (`gen_error`?)

dominiklohmann: We could also consider just putting our overload into the namespace `caf`.

tobim: That won't change the fact that our overload needs a conversion so the other one will still be preferred.

tobim:
Closing because this turned out to be more complicated that I thought.

PR #2654, opened 2022-10-26
Fix UDS datagram metrics reporting

tobim:
* Fix the success condition for `sendto`
* Instead of shutting down on anything but timeouts we now just keep on trying to write to the socket.

This can be tested with the following `vast.yaml`:
```
vast:
  enable-metrics: true
  metrics:
    uds-sink:
      enable: true
      path: /tmp/vast-metrics.sock
      type: datagram
```
Start a vast node with this and then start and stop socat with this command:
```bash
socat  'UNIX-RECV:/tmp/vast-metrics.sock' -
```

`socat` should print data to stdout every time it is active.

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/src/detail/posix.cpp
+++ b/libvast/src/detail/posix.cpp
@@ -137,7 +137,7 @@ caf::error uds_datagram_sender::send(std::span<char> data, int timeout_usec) {
   // the happy path.
   if (::sendto(src_fd, data.data(), data.size(), 0,
                reinterpret_cast<sockaddr*>(&dst), sizeof(struct sockaddr_un))
-      == 0)
+      >= 0)
     return caf::none;
   if (errno != EAGAIN && errno != EWOULDBLOCK)
     return caf::make_error(ec::system_error, "::sendto: ", ::strerror(errno));
@@ -152,7 +152,7 @@ caf::error uds_datagram_sender::send(std::span<char> data, int timeout_usec) {
   // The next send would go to the correct destination in that case.
   if (::sendto(src_fd, data.data(), data.size(), 0,
                reinterpret_cast<sockaddr*>(&dst), sizeof(struct sockaddr_un))
-      == 0)
+      >= 0)
     return caf::none;
   if (errno != EAGAIN && errno != EWOULDBLOCK)
     return caf::make_error(ec::system_error, "::sendto: ", ::strerror(errno));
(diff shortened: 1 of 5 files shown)


mavam:
How did you test that this fixes the issue?

libvast/src/system/accountant.cpp
```diff
@@ -230,18 +235,16 @@ struct accountant_state_impl {
...
-          break;
-        }
-        default: {
-          VAST_WARN("{} failed to write metrics to UDS sink: {}", *self, err);
-          VAST_WARN("{} turns off metrics reporting via UDS", *self);
-          uds_datagram_sink.reset();
-          return;
-        }
+      const auto code = ec{err.code()};
+      if (code == ec::timeout)
```
dominiklohmann, line 239: You should be able to just say `if (err == ec::timeout)`, regardless of whether or not the error has a context in addition to the code.

libvast/src/detail/posix.cpp
```diff
@@ -137,7 +137,7 @@ caf::error uds_datagram_sender::send(std::span<char> data, int timeout_usec) {
   // the happy path.
   if (::sendto(src_fd, data.data(), data.size(), 0,
                reinterpret_cast<sockaddr*>(&dst), sizeof(struct sockaddr_un))
-      == 0)
+      >= 0)
```
dominiklohmann, line 140: If the return value is the number of bytes sent, shouldn't we actually check whether it's exactly `data.size()` here and below?

dominiklohmann (approved):
I verified that this works by testing it locally with the provided instructions for socat.

libvast/src/system/accountant.cpp
tobim: `ec{err.code()}` is needed a bit further down anyways so why not give it a name and avoid the indirection through an extra `operator!=`? I don't think it makes a difference in readability.

libvast/src/detail/posix.cpp
tobim: Fair point. I'm afraid we have to come up with a new error code for that though.

libvast/src/system/accountant.cpp
dominiklohmann: Ah, didn't see that use of it. Feel free to resolve.

lava (approved):
I deployed this on our testbed and it emprically fixes the UDS datagram metrics reporting.

tobim:
[sc-37951]

PR #2667, opened 2022-10-27
Add /query endpoint to REST API

lava:
- Make initial taste size configurable
- Send 'atom::done' to the sink instead of the sender
- Implement /query endpoint

### :memo: Reviewer Checklist


web/openapi/openapi.yaml
```diff
@@ -16,6 +16,73 @@ components:
       in: header
       name: X-VAST-Token
 paths:
+  /query:
+    post:
```
netantho, line 20: Have you considered making it a GET?

Pro: The request doesn't change the state of VAST [ref](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods)
Pro: Easier to refresh and share.  Alternatively, the svelte frontend could have the query in GET parameters while still querying the backend API through a POST.
Cons: Would limit the size of the queries (but queries are small in VAST?). For example on the Splunk setup I know, the search is passed as a GET parameter, at least on the frontend (not sure if different for the backend).

web/openapi/openapi.yaml
```diff
@@ -16,6 +16,73 @@ components:
       in: header
       name: X-VAST-Token
 paths:
+  /query:
```
netantho, line 19: For symmetry with the `/query/<id>` endpoint, should `/query` maybe also take `n` and return `position`? I imagine that it will make things easier as a user (of the backend directly and for the frontend implementation).

web/openapi/openapi.yaml
mavam: Are there downsides when supporting both GET and POST? Then we can let the user choose their preferred method.

web/openapi/openapi.yaml
mavam: IIRC our discussions, the idea is that `/query` serves as "constructor" returning the ID, so it's the endpoint to use when you don't have a query yet. So `n` and `position` are not defined at this point.

web/openapi/openapi.yaml
dominiklohmann: GET would require random access iterator semantics if I understand correctly, while POST just requires input iterator semantics. This ultimately depends on whether we want to keep data sets in memory, or apply some deterministic ordering to them (currently, there is no order between events from different partitions, just within a single partition).

netantho: Ah I didn't consider or know about the random order, I think it would be an argument for the POST instead of GET.
In terms of web expectations I think you'd use GET if the output is always the same for the same data (so I think you would want to enforce a default ordering over different partitions).
For the limit of the URL it's 2040 characters [ref](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers), but again Splunk doesn't seem to run into the issue so maybe VAST wouldn't either.

In practical use cases I would find it convenient to be able to just save a search by copy-pasting the URL, which wouldn't be possible if neither the backend or frontend support a GET for /query, or would require short URL-generating code (ala Google flight, booking.com, etc.) which would probably something in the backend and a database to keep the short URL mappings.

netantho: I think I've now understood the workflow proposed: The user would make a `/query` request to start the search and then `/query/<id>` to see the results?
In that case I think both use cases with POST and GET would make sense. Some users would want to always see the same data as a result, some others (like what I described) would want to save the search but not freeze the result data.

mavam: Yeah, the workflow is:
1. Come with an expression to `/query` to create a query, get back an ID
2. Plow through `/query/{id}` in a forward-iterator fashion, i.e., a one-time sweep through the result set, where you can provide the chunking (n).

For (1), POST or GET don't matter that much. But for (2), we don't want GET right now because there is no idempotence: a refresh will show the *next* batch of results for the provided `n`.

To put this into perspective: we plan to have a fully deterministic "dataset" abstraction in the future. This will then support GET so that you can share the URL and get all the idempotent properties you'd expect when refreshing a page.

netantho: Thanks, I think it's what was confusing me. I think I get (😉) it now and it makes sense, thanks all for your patience and the great work.

web/openapi/openapi.yaml
lava: I just noticed that I actually forget to commit the latest changes before opening the PR, locally I had already renamed these to `/query/new` and `/query/:id:/next` because I also felt the endpoint names were not very intuitive.

Since they also contain a few bug fixes I'm going to push them, and if you like the original names better feel free to open another comment to revert the renaming @mavam

web/openapi/openapi.yaml
```diff
@@ -16,6 +16,73 @@ components:
...
+      responses:
+        200:
+          description: Success.
+          content: application/json
+          schema:
+            type: object
+            properties:
+              position:
+                type: integer
+                description: The number of events that had already been returned before this call.
```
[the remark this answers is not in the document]
lava: Yes, although I'm not convinced that this description is easier to understand ;)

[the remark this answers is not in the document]
mavam: (I honestly just wanted to try the math mode in GitHub.)

libvast/include/vast/query_context.hpp
```diff
@@ -123,6 +123,9 @@ struct query_context {
   /// The query expression.
   expression expr = {};
 
+  /// The initial taste size.
+  std::optional<uint32_t> taste = std::nullopt;
```
lava, line 127: Consider using just an uint32_t with a default value

lava: Nvm, the default value is not known at compile time because we have the `vast.max-taste-partitions` config setting

libvast/builtins/endpoints/query.cpp
```diff
@@ -0,0 +1,391 @@
...
+              auto id_string = fmt::format("{:l}", cursor.id);
+              self->state.live_queries_[id_string] = handler;
+              self->send(handler, atom::provision_v, cursor);
+              response->append(fmt::format("{{\"id\": \"{}\"}}\n", id_string));
+            },
+            [self, response = rq.response](const caf::error& e) {
+              response->abort(500, fmt::format("received error response from "
+                                               "index: {}\n",
+                                               e));
+            });
```
tobim, line 309: The `self` capture is unused and breaks CI.

libvast/builtins/endpoints/query.cpp
```diff
@@ -0,0 +1,390 @@
...
+  request_multiplexer_state() = default;
+
+  static constexpr auto name = "request_multiplexer";
+
+  system::index_actor index_ = {};
+  std::unordered_map<std::string, query_manager_actor> live_queries_;
+  struct request_queue {
+    query_manager_actor handler;
+    std::vector<atom::http_request> pending;
+  };
```
lava, line 181: Remove this class

libvast/builtins/endpoints/query.cpp
```diff
@@ -0,0 +1,390 @@
...
+  std::optional<system::query_cursor> cursor = std::nullopt;
+};
+
+struct request_multiplexer_state {
+  request_multiplexer_state() = default;
+
+  static constexpr auto name = "request_multiplexer";
+
+  system::index_actor index_ = {};
+  std::unordered_map<std::string, query_manager_actor> live_queries_;
```
lava, line 177: Add initializer

tobim (approved):
I consider this sufficient for the initial functionality. We may discover shortcomings when we start implementing the data explorer but such risks are always present and they should not prevent the basic functionality from landing.

lava (commented):
Comments from a pair-reviewing session.

PR #2669, opened 2022-10-27
Downgrade flatbuffers to 1.12.0

tobim:
This fixes a segfault in the static binary, where Arrow brings in a copy of flatbuffers at exactly version 1.12.0. When VAST was built with the newer 22.9.29 ODR violations would go undetected by the linker, but eventually cause faults in the array builders.

### :memo: Reviewer Checklist

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -39,13 +39,14 @@ in
     ];
   });
   flatbuffers = prev.flatbuffers.overrideAttrs(_: rec {
-    version = "22.9.29";
+    version = "1.12.0";
     src = prev.fetchFromGitHub {
       owner = "google";
       repo = "flatbuffers";
       rev = "v${version}";
-      sha256 = "sha256-Wc+DLihAN0EvaGkyuCreoRvzsNS3tIN7e2qIZhGldkw=";
+      hash = "sha256-L1B5Y/c897Jg9fGwT2J3+vaXsZ+lfXnskp8Gto1p/Tg=";
     };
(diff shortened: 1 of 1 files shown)


dit7ya:
What is going on with flatbuffer versions? https://github.com/google/flatbuffers/releases has two different kind of versioning!

Do we know which version we had before commit a05e2a02efde8eaa4dd2955c200b828805b85fc2 ?

If it came from nixpkgs directly then probably we had `2.0.0` - via https://github.com/NixOS/nixpkgs/blob/c4998c7bf6c84e3f5d7abe63a6ea8d110c2dbd95/pkgs/development/libraries/flatbuffers/default.nix#L10

(That is the rev for nixpkgs before the bump)

So maybe we should down/sidegrade to `2.0.0`? I am absolutely not sure about anything, just thinking out loud.

tobim:
> What is going on with flatbuffer versions? https://github.com/google/flatbuffers/releases has two different kind of versioning!

flatbuffers switched to calver this summer.

> So maybe we should down/sidegrade to 2.0.0? I am absolutely not sure about anything, just thinking out loud.

The best option is to use the exact version that Arrow uses to eliminate the risk of ODR problems.

dit7ya:
> The best option is to use the exact version that Arrow uses to eliminate the risk of ODR problems.
Understood. Thanks!

dispanser (approved):
Tested locally. Compared with master, this produces a binary that doesn't crash when ingesting suricata data.

tobim:
I also ran the integration tests on the resulting artifact locally fwiw.

PR #2679, opened 2022-11-01
Add SBOM generate script for nodejs and python

dit7ya:
This adds a script for generating SPDX SBOMS for python and nodejs dependencies.
The approach is a bit convoluted (generates CycloneDX format first, and then converts to SPDX) since I could not find a package that can do online lookup of license fields of nodejs dependencies.

This script can be used to append to the exisitng (and manually written) VAST.spdx file.
  
### :memo: Reviewer Checklist


dit7ya:
Sorry for the base branch change - the `plugins/web/ui` directory does not exist on master.

scripts/sbom-python-node.sh
```diff
@@ -0,0 +1,33 @@
...
+gum spin --spinner dot --title "Generating CycloneDX SBOM for the Python packages..." \
+  -- cdxgen python -o boms/python_cyclonedx.json
+
+# Convert to SPDX formats
+touch boms/output.spdx
+
+syft convert boms/ui_cyclonedx.json -o spdx-tag-value >> boms/output.spdx
+syft convert boms/python_cyclonedx.json -o spdx-tag-value >> boms/output.spdx
+
+cat boms/output.spdx
```
lava, line 30: `cat` can actually take multiple files as arguments. I also think it makes sense to apply some basic filtering like this, so the format looks similar to our current `VAST.spdx` file:

```
syft convert boms/ui_cyclonedx.json -o spdx-tag-value > boms/ui.spdx
syft convert boms/python_cyclonedx.json -o spdx-tag-value > boms/python.spdx

cat boms/ui.spdx boms/python.spdx | \
    grep -e "SPDXID:" -e PackageName: -e PackageVersion: -e PackageLicenseConcluded: -e PackageHomepage: -e ExternalRef: -e'^$'
```

lava (approved):
I tested it locally and it works great, thanks for taking the initiative here! Just a few minor style comments below.

In the long term we also need to consider if we want to have one long `.spdx` file or three separate ones, but we don't need to solve that question in this PR.

scripts/sbom-python-node.sh
```diff
@@ -0,0 +1,33 @@
+# Dependencies - cdxgen and syft
```
mavam, line 1:
```suggestion
#!/bin/sh
# Dependencies - cdxgen and syft
```

scripts/sbom-python-node.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# make a temporary directory
+BOMDIR=$(mktemp -d)
+
+# Generate the frontend ui SBOM
+# the main reason for using cdxgen (and then converting to spdx format)
+# is that it can look up nodeJS license fields online
+# FETCH_LICENSE=true gum spin --spinner dot --title \
+#   "Generating CycloneDX SBOM for the Frontend UI packages. Fetching the licenses from online may take a while..." \
+#   -- cdxgen plugins/web/ui -o $BOMDIR/ui_cyclonedx.json
```
lava, line 20: Is this supposed to be commented out?

lava:
Whoops, looks like I forgot about this PR for a while, but since we're doing an rc2 anyways I'm using the opportunity to sneak this in.

PR #2680, opened 2022-11-01
Add SBOM for VAST python package

lava:
This was generated by running the SBOM-generation script from #2679 and manually looking up the licenses for the packages that came out as `NONE` on PyPI.

### :memo: Reviewer Checklist


mavam:
What's the strategy for fusion the output of the script with your manual adjustments?

lava:
> What's the strategy for fusion the output of the script with your manual adjustments?

So far we don't have an automated solution for this; ideally we (or upstream) will figure out why so many packages get reported as `NONE` and we can use the script output directly. Alternatively, we'd probably have to write some program with a set of license overrides to cleanup the raw output. But for the 2.4 release, I think we don't get around just doing this manually.

python/vast-python.spdx
```diff
@@ -0,0 +1,631 @@
+PackageName: antlr4-python3-runtime
```
dominiklohmann, line 1: Can you append this file to VAST.spdx instead of it being its own package (with some delimiters making it clear where Python begins and ends), or change the installation logic in CMake to also install this SBOM file?

dominiklohmann (approved):

CMakeLists.txt
```diff
@@ -790,6 +790,7 @@ install(
   FILES "${CMAKE_CURRENT_BINARY_DIR}/CHANGELOG.md"
         "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE"
         "${CMAKE_CURRENT_SOURCE_DIR}/VAST.spdx"
+        "${CMAKE_CURRENT_SOURCE_DIR}/python/vast-python.spdx"
```
mavam, line 793: How about we call these files `sbom.spdx` and `sbom-python.spdx`? Perhaps with a `vast-` prefix to make it even clearer.

dominiklohmann, line 793: This will also have to be changed to it's attached to releases. Please grep for all occurrences of VAST.spdx.

lava: I just appended the entries to the existing `VAST.spdx` for now to minimize the amount of changes.

lava:
Adding the blocked label until we actually merged https://github.com/tenzir/vast/pull/2636 to install the python bindings by default.

dominiklohmann:
I have rebased this on top of #2636

dominiklohmann:
Unblocking and merging this now that I've changed the target to be the packaging PR.

PR #2682, opened 2022-11-04
Add "-total" metric keys for schema-dependent metrics

Dakostu:
This PR adds three metric keys `catalog.num-partitions-total`, `catalog.num-events-total` and `ingest-total` that sum up all schema-based metrics by their respective schema-based metric counterparts.

### :memo: Reviewer Checklist


libvast/src/system/catalog.cpp
```diff
@@ -516,27 +516,41 @@ catalog(catalog_actor::stateful_pointer<catalog_state> self,
         num_partitions += 1;
         num_events += synopsis->events;
       }
+      auto total_num_partitions = count{0};
+      auto total_num_events = count{0};
       auto r = report{};
       r.data.reserve(num_partitions_and_events_per_schema_and_version.size());
       for (const auto& [schema_and_version, num_partitions_and_events] :
            num_partitions_and_events_per_schema_and_version) {
+        auto schema_num_partitions = num_partitions_and_events.first;
```
patszt, line 525: Could be changed to one liner:
auto [num_partitions, num_events] = num_partitions_and_events;

But tbh it's not a big deal. Pick the one you like more :)

libvast/src/system/importer.cpp
```diff
@@ -153,9 +153,14 @@ void importer_state::send_report() {
...
   samples.push_back(performance_sample{"node_throughput"s, node_throughput});
-  for (const auto& [name, count] : schema_counters)
+  auto total_count = count{0};
+  for (const auto& [name, count] : schema_counters) {
+    total_count += count;
     samples.push_back(performance_sample{
       "ingest", measurement{elapsed, count}, {{"schema", name}}});
+  }
+  auto total_measurement = measurement{elapsed, total_count};
+  samples.push_back(performance_sample{"ingest-total"s, total_measurement});
```
patszt, line 163: In the story it was suggested to name the key ingest.rate-total. I think we should stick with it

patszt (changes requested):
Congrats on first PR. Please verify the naming of the ingest-total.
Other than that the changes look good

libvast/src/system/catalog.cpp
Dakostu: Structured bindings are "more modern" and already used in the rest of the file - I applied your suggestion.

libvast/src/system/importer.cpp
Dakostu: The log will produce the following values for the `ingest-total` `performance_sample`: 
Screenshot_20221104_124521

So we would have `ingest-total.rate` amongst other things.

It looks like those three suffixes are added by the `accountant_actor`: 
Screenshot_20221104_124632

Do you think adding extra logic for this edge case (the `performance_sample` is total) would be worth it?

patszt: Okay, then it is not worth it

patszt (approved):

PR #2683, opened 2022-11-04
Refactor synopsis inspection API for easier maintenance

patszt:
This is the cleanup of synopsis hierarchy that i've done for CAF 0.18.6 update.
Instead of creating virtual function for each inspector that has the same implementation we can create one using "double dispatch"

deserialize(vast::detail::legacy_deserializer& will be gone with the main PR as all inspect function will return bool 
The inspect(legacy_deserializer, synopsis_ptr) will also be gone as the common template inspect function will suffice 

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/include/vast/min_max_synopsis.hpp
+++ b/libvast/include/vast/min_max_synopsis.hpp
@@ -80,12 +80,11 @@ public:
     return sizeof(min_max_synopsis);
   }
 
-  caf::error serialize(caf::serializer& sink) const override {
-    return sink(min_, max_);
-  }
-
-  caf::error deserialize(caf::deserializer& source) override {
-    return source(min_, max_);
+  caf::error inspect(supported_inspectors& inspector) override {
+    return std::visit(detail::overload{[this](auto inspector) {
+                        return inspector(min_, max_);
+                      }},
+                      inspector);
   }
 
   bool deserialize(vast::detail::legacy_deserializer& source) override {
--- a/libvast/include/vast/synopsis.hpp
+++ b/libvast/include/vast/synopsis.hpp
@@ -13,6 +13,7 @@
 #include "vast/aliases.hpp"
 #include "vast/detail/legacy_deserialize.hpp"
 #include "vast/fbs/synopsis.hpp"
+#include "vast/legacy_type.hpp"
 #include "vast/operator.hpp"
 #include "vast/type.hpp"
 #include "vast/view.hpp"
@@ -21,6 +22,7 @@
 
 #include <memory>
 #include <optional>
+#include <variant>
 
 namespace vast {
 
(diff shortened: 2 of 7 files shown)


libvast/include/vast/synopsis.hpp
```diff
@@ -115,4 +108,66 @@ pack(flatbuffers::FlatBufferBuilder& builder, const synopsis_ptr&,
...
+      return sink(ptr->type().to_legacy_type());
+    },
+    [&] {
+      synopsis::supported_inspectors i{std::ref(sink)};
+      return ptr->inspect(i);
+    });
+}
+
+template <class Inspector>
+  requires(std::is_same_v<typename Inspector::result_type, caf::error>)
```
patszt, line 164: This will be gone in main CAF PR. It is needed as stringification inspector triggers compilation error here.

Dakostu (approved):
This approach seems good to me and we already discussed performance concerns (result: there should be no serious concerns) so I'll approve.

libvast/include/vast/min_max_synopsis.hpp
```diff
@@ -80,12 +80,11 @@ class min_max_synopsis : public synopsis {
...
-    return sink(min_, max_);
-  }
-
-  caf::error deserialize(caf::deserializer& source) override {
-    return source(min_, max_);
+  caf::error inspect(supported_inspectors& inspector) override {
+    return std::visit(detail::overload{[this](auto inspector) {
+                        return inspector(min_, max_);
+                      }},
+                      inspector);
```
dominiklohmann, line 87:
```c++
const auto f = [this](auto& inspector) {
  return inspector(min_, max_);
};
return std::visit(f, inspector);
```

Three things in this:
- Take `inspector` by reference rather than making a copy.
- Remove the overload—it's not necessary if there's just a single lambda.
- Put the visitor into a separate variable to avoid the additional indentation for readability.

libvast/include/vast/synopsis.hpp
```diff
@@ -73,11 +78,7 @@ class synopsis {
 
   // -- serialization ----------------------------------------------------------
 
-  /// Saves the contents (excluding the layout!) of this slice to `sink`.
-  virtual caf::error serialize(caf::serializer& sink) const = 0;
-
-  /// Loads the contents for this slice from `source`.
-  virtual caf::error deserialize(caf::deserializer& source) = 0;
+  virtual caf::error inspect(supported_inspectors& inspector) = 0;
```
dominiklohmann, line 81: I think it's rather dangerous to name this function `inspect`, as that usually means it's being called directly by VAST.

dominiklohmann: Maybe something like `inspect_impl` or so is better here, just to make it obvious that it is being called from the actual `inspect` overload.

patszt: I will rename it in my next PR

PR #2684, opened 2022-11-04
Run integration tests on the Nix Static binary

dit7ya:
This PR adds support for running the integration tests on the nix-built static binary.

There is already an `installCheckPhase` test for the same. However, that fails in the static build as `systemdMinimal` is a dependency (which is broken for pkgsStatic platform on nixpkgs right now).

The approach in this PR is to just extract the binary and run the tests in a nix shell with the python etc. dependencies. This makes the integration tests decoupled from the build process, and as we need to just test the resulting binary, IMO, this approach is fine.

**Note to reviewers**: Please check the Upload Logs step in the CI, I have not tested that. Also, `mach-nix` is certainly not required for the Python shell, but it is pretty handy. Let me know if you want me to do it the regular way.

### :memo: Reviewer Checklist

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -489,6 +489,19 @@ jobs:
           DEB=$(ls "${{ steps.build_static_packages.outputs.store_path }}" | grep "${{ matrix.nix.target }}.*.deb")
           echo "tar_gz=${TAR_GZ}" >> $GITHUB_OUTPUT
           echo "deb=${DEB}" >> $GITHUB_OUTPUT
+      - name: Run Integration Tests
+        run: |
+          mkdir -p vast-integration-test
+          tar xf ./result/${{ steps.get_artifact_names.outputs.tar_gz }} --directory vast-integration-test
+          nix shell --impure mach-nix#gen.python.coloredlogs.jsondiff.pyarrow.pyyaml.schema --command python vast/integration/integration.py --app \
+            vast-integration-test/opt/vast/bin/vast
+      - name: Upload Integration Test Logs on Failure
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: "vast-integration-test-nix-static-${{ matrix.nix.target }}"
+          path: "vast-integration-test"
+          if-no-files-found: error
       - name: Upload Tarball to Github
         uses: actions/upload-artifact@v3
         with:


dit7ya:
As you can see, running the integration tests from our flake's devShell takes way longer than using mach-nix for Python packages (just because there are a bunch of extra dependencies for the entire devShell). So I would recommend using the previous commit only. Let me know what you think.

.github/workflows/vast.yaml
```diff
@@ -990,6 +990,19 @@ jobs:
...
+        run: |
+          tar xf ./result/${{ steps.get_artifact_names.outputs.tar_gz }}
+          nix shell --command python vast/integration/integration.py --app opt/vast/bin/vast
+
+      - name: Upload Integration Test Logs on Failure
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: "vast-integration-test-nix-static-${{ matrix.build.compiler }}-${{ matrix.configure.tag }}"
+          path: "${{ env.BUILD_DIR }}/vast/vast-integration-test"
```
dominiklohmann, line 1003: I don't see $BUILD_DIR set in the environment. What you want here is the directory that CMake does the build in, i.e., the directory that is passed to CMake via `-B path/to/build`, and then inside that `/vast/vast-integration/test`.

.github/workflows/vast.yaml
```diff
@@ -990,6 +990,19 @@ jobs:
...
+      - name: Run Integration Tests
+        run: |
+          tar xf ./result/${{ steps.get_artifact_names.outputs.tar_gz }}
+          nix shell --command python vast/integration/integration.py --app opt/vast/bin/vast
+
+      - name: Upload Integration Test Logs on Failure
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: "vast-integration-test-nix-static-${{ matrix.build.compiler }}-${{ matrix.configure.tag }}"
```
dominiklohmann, line 1002: There is no `matrix.build` and `matrix.configure` for the Nix jobs. For disambiguation in the matrix, I propose just using this:

```suggestion
          name: "vast-integration-test-nix-static-${{ matrix.nix.target }}"
```

dominiklohmann (approved):
This is great, thanks for taking the initiative on this. I just have some minor requests for this.

dit7ya:
Due to the CI changes, I manually ran the workflow here - https://github.com/tenzir/vast/actions/runs/3434450744/jobs/5725773552

dominiklohmann (approved):
I'm fairly certain that this works! Thanks a lot for making this change

PR #2689, opened 2022-11-08
Load "all" plugins by default & allow "empty" values

Dakostu:
Load "all" plugins if no "plugins" configuration key is defined in the configuration.

### :memo: Reviewer Checklist


web/docs/setup/configure.md
```diff
@@ -174,9 +174,11 @@ Static plugins do not require installation since they are compiled into VAST.
...
 
 The onfiguration key `vast.plugins` specifies the list of plugins that should
-load at startup. The `bundled` and `all` plugin names are reserved. When
-`bundled` is specified VAST loads all plugins built along with it. When `all` is
-specified VAST loads all available plugins in the configured plugin directories.
+load at startup. The `all` plugin name is reserved. When `all` is specified
+VAST loads all available plugins in the configured plugin directories. If no
+`vast.plugins` key is specified, VAST will load `all` plugins by default.
+To load no plugins at all, specify a `vast.plugins` key with no
+values in the configuration file or use the launch parameter `--plugins=[]`.
```
dominiklohmann, line 181: When testing this, please make sure it works for all these scenarios (and combinations of them)
1. `vast --plugins= start`
2. `vast --plugins=[] start`
3. `VAST_PLUGINS= vast start`
4. `VAST_PLUGINS=[] vast start`
5. `vast start` with `vast: { plugins: [] }` in your config file.

The empty value should be supported as well, since the documentation states that we take a comma-separated list of plugins.

dominiklohmann (commented):
Just one minor comment. Thanks for tackling this one; I'll leave the remaining review to the team.

mavam:
This is makes the OOTB experience so much better!

web/docs/setup/configure.md
Dakostu: 1 ends up in a `parser_error`
3 and 4 end up in `all` being included despite an empty value being specified.

Yep, this is probably the specified "fix merging of configuration options with a known list type to allow for distinguishing between not specified and empty list (`plugins = []`)" task I'm working on.

libvast/src/system/configuration.cpp
```diff
@@ -444,12 +444,24 @@ caf::error configuration::parse(int argc, char** argv) {
         .add<std::vector<std::string>>("?vast", "schema-dirs", "")
         .add<std::vector<std::string>>("?vast", "plugin-dirs", "")
         .add<std::vector<std::string>>("?vast", "plugins", "");
-  auto [ec, it] = plugin_opts.parse(content, plugin_args);
-  if (ec != caf::pec::success) {
-    VAST_ASSERT(it != plugin_args.end());
-    return caf::make_error(ec, fmt::format("failed to parse option '{}'", *it));
+
+  // edge case for "plugins" option: no argument specified (e.g. "--plugins=")
+  // this case should then be resolved to "load no plugins" (empty list)
```
dominiklohmann, line 449: Is there a way that we can fix this generally instead of adding a one-off for a single option?

Dakostu: 1. Generally for all plugin-only options: Yes, I can remove the `starts_with("--plugins=")` check.
2. Generally for all options: Well, `CAF` does the parsing for us in `config_option_set.cpp`, I probably would have to implement workarounds like this one...

Dakostu: ... or maybe a helper function/method.

Dakostu: ... or I could try to iterate through all command arguments instead of only `plugin_args`.

dominiklohmann: I think either works, but ultimately the option handling should be consistent for options of the same type. In this particular case, I'd expect all options that contain a list of strings to behave the same. If I can specify them as a comma-separated list then it'd only be natural for the empty string to be parsed as an empty list.

tobim:
[sc-37984]

libvast/src/system/configuration.cpp
Dakostu: Wow, implementing this took a bit of time.. the problem is we have multiple `parse` ings in the `vast` application setup and also the setup workflow steps can't be just moved around (plugin arguments need to be parsed before loading them and then parsing arguments again).

But I think I managed to implement an approach.

web/docs/setup/configure.md
Dakostu: As of now, my current implementation has minimal `configuration.cpp` changes. @patszt feel free to check if conflicts would arise (in the future).

dominiklohmann:
@tobim can I ask you to take the review from here on out?

libvast/include/vast/command.hpp
```diff
@@ -192,8 +192,7 @@ struct invocation {
 /// @returns an error for malformed input, `none` otherwise.
 /// @relates command
 caf::expected<invocation>
-parse(const command& root, command::argument_iterator first,
-      command::argument_iterator last);
+parse(const command& root, std::vector<std::string>& arguments);
```
tobim, line 195: I think `arguments` should really be taken by value. Alternatively you could also revert to using the iterators and just forcing the copy. This code is basically as far away from performance critical as it gets.

tobim (approved):
I tested this change locally and all my calls to vast start and other sub commands behaved as expected.
Ready to be merged once the code change request is addressed.

changelog/unreleased/changes/2689--load-all-plugins-by-default.md
```diff
@@ -0,0 +1 @@
+If no "plugins" configuration key is defined in the configuration file or launch parameter, VAST will by default load all plugins now. This can be circumvented by specifying a "plugins" configuration key with no values.
```
dominiklohmann, line 1: Three small requests for this:
- Can you reflow this text to 80 columns to align with the other Markdown files in the repository?
- Please refer to configuration options with their full name and use Markdown code blocks, i.e., `vast.plugins` rather than "plugins."
- Use active voice and present tense in documentation and changelog entry.

libvast/test/system/configuration.cpp
```diff
@@ -41,6 +41,8 @@ struct fixture {
     auto [root, _] = system::make_application(argv[0]);
     REQUIRE(root);
     // Parse the CLI.
+    cfg.command_line = ::sanitize_arguments(*root, cfg.command_line.begin(),
+                                            cfg.command_line.end());
     auto invocation
       = ::parse(*root, cfg.command_line.begin(), cfg.command_line.end());
```
dominiklohmann, line 47: Is there a reason why sanitizing the arguments is not part of the call to parse?

dominiklohmann (approved):
This is great, just some smaller things left to do. I also added a `blocked` label to prevent this from being merged accidentally in the feature freeze for v2.4; whether this gets in regardless is up to @lava.

changelog/unreleased/changes/2689--load-all-plugins-by-default.md
Dakostu: I improved the changelog formatting in https://github.com/tenzir/vast/pull/2689/commits/ed9a3a0a8ccdae0c13e7103a9b32e77f2da056d6

libvast/test/system/configuration.cpp
Dakostu: Well, making the sanitation use the iterator as parameters made me also think about separation of concerns and whether the sanitation should be explicitily called.. but after a night's sleep I figured making `sanitize_arguments` part of `parse` again would make application setup slightly easier again, so I applied that change in https://github.com/tenzir/vast/pull/2689/commits/2444e979117d7e4b7678bfe2dca902e10681af63

dominiklohmann: I think in this case it's fine to have one call the other. In general I agree with aiming for separation of concerns, but our application setup already takes quite a few steps and I think this can too easily be forgotten otherwise.

changelog/unreleased/changes/2689--load-all-plugins-by-default.md
dominiklohmann: It's still passive voice. I recommend the following:

> VAST now loads all plugins by default. To revert to the old behavior, explicitly set the `vast.plugins` option to have no value.

Dakostu: I applied your suggestion in https://github.com/tenzir/vast/pull/2689/commits/b8f4424f45dc1ea20698e8de9add1bb3f95f4761

dominiklohmann:
Since the first rc will be tagged rc2 anyways we decided to squeeze this in in today's standup.

PR #2690, opened 2022-11-08
Use one virtual method for inspection of value_index hierarchy

patszt:
WIth CAF 0.18.6 update the number of supported inspectors grows to 4. With the old architecture the code was a bit messy

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/include/vast/system/transformer.hpp
+++ b/libvast/include/vast/system/transformer.hpp
@@ -15,6 +15,7 @@
 #include "vast/system/sink.hpp"
 #include "vast/table_slice.hpp"
 
+#include <caf/broadcast_downstream_manager.hpp>
 #include <caf/settings.hpp>
 #include <caf/stream_stage.hpp>
 #include <caf/typed_event_based_actor.hpp>
--- a/libvast/include/vast/value_index.hpp
+++ b/libvast/include/vast/value_index.hpp
@@ -14,6 +14,7 @@
 #include "vast/error.hpp"
 #include "vast/ewah_bitmap.hpp"
 #include "vast/ids.hpp"
+#include "vast/legacy_type.hpp"
 #include "vast/type.hpp"
 #include "vast/view.hpp"
 
@@ -23,6 +24,7 @@
 #include <caf/settings.hpp>
 
 #include <memory>
+#include <variant>
 
 namespace vast {
 
@@ -34,9 +36,13 @@ using value_index_ptr = std::unique_ptr<value_index>;
 /// and an explit query for nil, e.g., `x != 42 || x == nil`.
 class value_index {
 public:
+  using supported_inspectors
+    = std::variant<std::reference_wrapper<caf::serializer>,
+                   std::reference_wrapper<caf::deserializer>>;
+
   value_index(vast::type x, caf::settings opts);
 
-  virtual ~value_index();
+  virtual ~value_index() noexcept = default;
 
   using size_type = typename ids::size_type;
 
@@ -79,9 +85,7 @@ public:
 
   // -- persistence -----------------------------------------------------------
 
-  virtual caf::error serialize(caf::serializer& sink) const;
-
-  virtual caf::error deserialize(caf::deserializer& source);
+  virtual caf::error inspect_impl(supported_inspectors& inspector);
 
   virtual bool deserialize(detail::legacy_deserializer& source);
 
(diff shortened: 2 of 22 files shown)


libvast/include/vast/system/transformer.hpp
```diff
@@ -15,6 +15,7 @@
 #include "vast/system/sink.hpp"
 #include "vast/table_slice.hpp"
 
+#include <caf/broadcast_downstream_manager.hpp>
```
patszt, line 18: This include is needed. The file uses broadcast_downstream_manager but in CAF 0.18.6 lack of it leads to compilation error.
It is included by some other caf dependency here i believe

tobim: Since this seems unrelated to the changes of the rest of the commit: It would be great if you could put such fixes in a separate commit next time.

tobim (commented):
After reflecting on this for a while I'm not quite sure what the end goal is here.

We already have a flatbuffer implementation of `value_index`. But that seems to be unused, I'm not sure why. However, if we plan to switch to that as the serialization format, then we could actually drop support for `caf::binary_deserializer` and `caf::binary_serializer`, leaving only the `legacy_deserializer` to support existing data.

patszt:
> 

Will legacy_deserializer be removed after we drop support for some old partition version?

patszt:
> After reflecting on this for a while I'm not quite sure what the end goal is here.
> 
> We already have a flatbuffer implementation of `value_index`. But that seems to be unused, I'm not sure why. However, if we plan to switch to that as the serialization format, then we could actually drop support for `caf::binary_deserializer` and `caf::binary_serializer`, leaving only the `legacy_deserializer` to support existing data.

> 

Will legacy_deserializer be removed after we drop support for some old partition version?
The end goal really is to have less repetition in code. The repetition was bigger with the caf upgrade

tobim:
> Will legacy_deserializer be removed after we drop support for some old partition version? The end goal really is to have less repetition in code. The repetition was bigger with the caf upgrade and this PR will reduce the size of main CAF upgrade PR

I think `legacy_deserializer` is the only one we want to keep.

I also think that you actually have to switch over to the flatbuffers persistence method before upgtading to 0.18, because the output of the `caf::serializer` changed, and our `legacy_deserializer` won't be able to parse data written with the new serializer.

mavam:
Is the 2 GiB file limit also a concern here?

libvast/include/vast/value_index.hpp
```diff
@@ -34,9 +36,13 @@ using value_index_ptr = std::unique_ptr<value_index>;
...
 class value_index {
 public:
+  using supported_inspectors
+    = std::variant<std::reference_wrapper<caf::serializer>,
+                   std::reference_wrapper<caf::deserializer>>;
+
   value_index(vast::type x, caf::settings opts);
 
-  virtual ~value_index();
+  virtual ~value_index() noexcept = default;
```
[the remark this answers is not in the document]
patszt: Is this a thing? If the effect is not noticable then i'd stay with = default for slightly better/easier readability.

tobim:
> Is the 2 GiB file limit also a concern here?

Good point. @lava as far as I understand the `external_indices` approach could be translated to `external_bitmaps` if we change the persist return type of `active_indexer` to `std::vector<chunk_ptr>`, right?

tobim:
> The integration tests works so i'd assume that persistance and read path are working.

This breaks my mental model. I assumed that the serializer changes in CAF 0.18 were breaking.

dominiklohmann:
> Is the 2 GiB file limit also a concern here?

This is handled correctly already.

> Good point. @lava as far as I understand the `external_indices` approach could be translated to `external_bitmaps` if we change the persist return type of `active_indexer` to `std::vector<chunk_ptr>`, right?

Yes, that is correct.

> This breaks my mental model. I assumed that the serializer changes in CAF 0.18 were breaking.

They are, and this has to be handled separately in the PR that upgrades to CAF 0.18. This is just one of the pieces. It's not relevant to this PR all too much.

tobim:
> They are, and this has to be handled separately in the PR that upgrades to CAF 0.18. This is just one of the pieces. It's not relevant to this PR all too much.

It seems relevant to me. If we remove support for the binary serializer, then what are the "4 supported inspectors" from the PR description?

patszt:
> > They are, and this has to be handled separately in the PR that upgrades to CAF 0.18. This is just one of the pieces. It's not relevant to this PR all too much.
> 
> It seems relevant to me. If we remove support for the binary serializer, then what are the "4 supported inspectors" from the PR description?

One of supported inspector is caf::strinigifcation_inspector. The CAF unit test macros now use strinigification inspector to print values in a failed check. I think it is simpler to allow our current code to be inspectable by it. The workaround would be
to create inspect(stringification_inspector, value_index_ptr) that does nothing in unit tests. Allowing stringification inspector might help in debugging though. It only costs adding it to supported inspectors after refactor.

CAF 0.18 also requires all types used in actor's communication are binary serializable. There is some kind of static_assertion.

Most of the 0.17 outputs are the same as 0.18
I have found that the output is different for caf::config_value due to "ABI" changes.
The underlying variant changed:
0.18 is type_list<none_t, integer, boolean, real, timespan, uri, string, list, dictionary>
and 0.17 is type_list<integer, boolean, real, atom, timespan, uri, string, list, dictionary>

We save caf::config_value as a part of value_index so we need legacy deserializer for old variant.

One way would be to create a serializer that is CAF independant. This can work, but also can break in the future as we depend on caf::config_value.
Anyway this would lead to the 4th serializer. Either caf::binary_deserializer to support the newly written caf::config_values with caf::binary_serializer. Or use our own serializer + legacy_deserializer. This will also lead to 4 inspectors as we have to support binary_serializer due to static_assertion.

This design allows us to add support for new inspectors quickly and reduces code duplication. In the worst case i still think this makes the code more maintable.

PR #2691, opened 2022-11-09
Reporting frontend skeleton

dit7ya:
This PR adds a /reporting route to the frontend which includes a basic skeleton for creating a report - a query block and a markdown block.

Though functionality and design-wise this is not much, we want to merge this so that we can work in smaller PRs.

To review, run in the frontend folder `yarn dev` and visit http://localhost:5174/report


tobim:
I think it makes sense to write the documentation after this feature has evolved a bit. Right now it is very much experimental and should not be advertised, but in order to make progress it would still be nice to have the scaffold and basic functionality in `master`.

dit7ya (commented):
Remove the unneeded LayerCake stuff with a commit.

plugins/web/ui/src/lib/components/IconButton.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
```
dit7ya, line 1: This is not showing anywhere, whereas it needs to be on blocks' Edit and Delete buttons.

plugins/web/ui/src/lib/components/Menu.svelte
```diff
@@ -0,0 +1,63 @@
+<script lang="ts">
+  import { Menu, MenuButton, MenuItems, MenuItem, Transition } from '@rgossiaux/svelte-headlessui';
+
+  function classNames(...classes: (string | false | null | undefined)[]) {
```
dit7ya, line 4: Change to arrow functions.

plugins/web/ui/src/lib/components/Table.svelte
```diff
@@ -19,32 +23,44 @@
...
-                <th {...attrs}>
+              <Subscribe attrs={cell.attrs()} let:attrs props={cell.props()} let:props>
+                <th
+                  {...attrs}
+                  on:click={props.sort.toggle}
+                  class={showHeaderBorder ? 'border border-slate-300' : ''}
+                >
                   <Render of={cell.render()} />
+                  {#if props.sort.order === 'asc'}
+                    ⬇️
```
dit7ya, line 45: Change the emoji to an icon.

plugins/web/ui/src/lib/components/TextField.svelte
```diff
@@ -0,0 +1,24 @@
+<script lang="ts">
```
dit7ya, line 1: Remove this leftover.

plugins/web/ui/src/lib/util/api.ts
```diff
@@ -0,0 +1,18 @@
+export const getQuery = async (expression: string, limit = 100) => {
+  try {
+    const API_BASE = import.meta.env.VITE_VAST_API_ENDPOINT ?? `${import.meta.env.BASE_URL}api/v0`;
+
+    // const url = `${API_BASE}/export?expression=${encodeURIComponent('#type == "zeek.conn" && id.orig_h in 192.168.0.0/16')}`;
```
dit7ya, line 5: Remove comment.

plugins/web/ui/src/lib/util/misc.ts
```diff
@@ -0,0 +1,29 @@
+type ConnMinimal = {
```
dit7ya, line 1: Remove me (chart related).

plugins/web/ui/src/routes/report/Editor.svelte
```diff
@@ -0,0 +1,120 @@
+<script lang="ts">
```
dit7ya, line 1: Rename me to something better.

dit7ya: Report Editor.

plugins/web/ui/src/routes/report/Editor.svelte
```diff
@@ -0,0 +1,120 @@
...
+  const handlePdfExport = () => {
+    window.print();
+  };
+  // const handleImport = () => {
+  //
+  // }
+
+  const handleJsonExport = () => {
+    // Create a blob of the data
+    let fileName = 'vast-report.json'; // TODO use the title of the report
```
dit7ya, line 18: FIXME.

plugins/web/ui/src/routes/report/Editor.svelte
```diff
@@ -0,0 +1,120 @@
...
+                title: '',
+                content: ''
+              }
+            }
+          ]
+        };
+      });
+    }
+  };
+  $: console.log($report);
```
dit7ya, line 60: REMOVE.

plugins/web/ui/src/routes/report/Importer.svelte
```diff
@@ -0,0 +1,42 @@
...
+  import Button from '$lib/components/Button.svelte';
+
+  // a reference to the component, used to call FilePond methods
+  let pond;
+
+  // the name to use for the internal file input
+  let name = 'filepond';
+
+  const handleAddFile = async (err, fileItem) => {
+    // console.log('A file has been added', fileItem);
```
dit7ya, line 15: Cleanup.

plugins/web/ui/src/routes/report/MarkdownBlock.svelte
```diff
@@ -0,0 +1,36 @@
...
+  import Viewer from '$lib/components/markdown/Viewer.svelte';
+  import gfm from '@bytemd/plugin-gfm';
+  import highlight from '@bytemd/plugin-highlight';
+  import mermaid from '@bytemd/plugin-mermaid';
+  import BlockHeader from './BlockHeader.svelte';
+
+  let showEditor = true;
+
+  export let parameters = {
+    title: 'TODO make me editable',
```
dit7ya, line 13: Here.

plugins/web/ui/src/routes/report/QueryBlock.svelte
```diff
@@ -0,0 +1,52 @@
+<script lang="ts">
```
dit7ya, line 1: Remove ME.

mavam (changes requested):
This is great start!

Here's what I noticed playing with it:

### Report name bug

When I try to set a report name, remove all characters, I cannot get the box back:

image

### Icon weirdness

Per the above screenshot, the icons look a bit weird to me, with grey borders and some degree of pixelation. IIRC we wanted initially a very minimalistic black-and-white style. Maybe apply that here too?

### Moving cells

I couldn't find a way to move sections around via click-and-drag. I would expect this sort of functionality next to the edit and delete icon.

### Cell titles

When writing reports, I don't care too much about the cell title. Consider this:
image

### Heavy-weight horizontal separators

Per the above screenshot, I find the separators between the cells too much in the way. Can we downtone them? We could also make them vertical, so that the content reads cleanly but hoving over a cell demarcates the scope with a vertical bar on the right side, next to the other cell controls.

mavam (dismissed):
I found a few more things.

### Export overlay bug

This says it all:

image

### PDF export includes export menu

Should also be clear from this screenshot:

image

### Global mode switch

When I've opened a few text boxes and I click the report button, all of these expand. Then I have to click save again on all of them to go into viewing mode. Could we have a global toggle?

lava (approved):
The code looks good to me on a high level, I think my Svelte still isn't strong enough to really comment on the details :D

As already discussed, I agree that it makes sense to merge this now and create separate stories for any outstanding bugs noticed by @mavam in his review.

Since we still have the whole frontend marked as experimental I'm not too worried about adding half-finished features, for the future it would probably be nice to introduce some feature toggle so we can develop this on master with "frontend-reporting=off" set by default and then we could flip the default to "on" once everything is done.

tobim (approved):
The functionality is in the state that we agreed on. I took only a cursory look at the code and didn't see any obvious issues.

mavam (approved):
Looking at the last 4 commits (code only), I would say this is in good shape now to ship.

PR #2692, opened 2022-11-09
Fix the scaffolding for docker-based VAST python apps

dispanser:
Enable `import vast` to work, which wasn't the case previously because it didn't actually properly run `poetry install`.

### :memo: Reviewer Checklist


docker/misp/docker-compose.app.yaml
```diff
@@ -3,10 +3,13 @@ services:
   misp-app:
     image: ${VAST_CONTAINER_REGISTRY:-docker.io}/tenzir/vast-poetry:${VAST_CONTAINER_REF:-latest}
     volumes:
-      - ./app/:/app
-      - ../../python/:/vast
-    working_dir: /app
+      - bindings-cache:/home/vast/.cache
+      - ./app/:/opt/misp/app
+      - ../../python/:/python
```
rdettai, line 8:
```suggestion
      - ./app/:/opt/misp/app  # relative paths must match on host and in the container
      - ../../python/:/python
```

python/pyproject.toml
```diff
@@ -29,16 +29,16 @@ coloredlogs = "^15.0"
 pandas = "^1.5"
 pyarrow = "^10.0"
 # fabric
-misp-stix = {version = "^2.4", optional = true}
+misp-stix = {version = "^2.4"}
```
rdettai, line 32: did you identify why this optional dependency was leaking into the core? It would really be better if we could keep this optional

rdettai (approved):

python/pyproject.toml
dispanser: If this is optional, `stix2` is missing when doing `import vast`.

We could probably have an explicit dependency for stix2 instead, let me try that.

dispanser: that seems to work, resolving. You have time until the build passes to raise additional concerns ;-)

PR #2706, opened 2022-11-15
Rename "native plugins" to "builtins"

dominiklohmann:
This is a code base cleanup related to #2689.

Historically speaking, "native plugins" were functionality that was part of libvast but used the plugin API. This allowed us to have defaults that we ship as built-in functionality. The only reason why we did not name them builtins from the get-go was because the term builtin was too close to the term bundled, which we already used for plugins that were built alongside VAST in the same build configuration (via `-DVAST_PLUGINS=...`). Now that bundled plugins are a purely internal term we can instead start referring to native plugins as builtins.

In addition, this changes the show command to be a built-in instead without making any functional changes to it.

### :memo: Reviewer Checklist


Dakostu (approved):
I finished a build with this and launched a client with `vast show` - works. So a careful approval from me.

I did find two remainders of `native plugin` syntax though.

PR #2707, opened 2022-11-15
Add `vast import arrow` enabling ingestion of arrow IPC format

dispanser:
This is a dual to `vast export arrow`, so you can now `vast export arrow | vast -e ... import arrow` to duplicate data to another node.

Note that this assumes VAST schema compabilitity, i.e. we assume that the IPC data has been created by VAST itself, in particular we only support the subset of Arrow that's supported in VAST, and expect the metadata to carry the necessary information to reconstruct the full VAST type.

### :memo: Reviewer Checklist


dominiklohmann:
Nice! If I understand this correctly, this will allow me to do the following:

```bash
vast -d path/to/db1 -e :42001 start
vast -d path/to/db2 -e :42002 start
# make db2 a mirror of db1
vast -e :42001 export -u arrow | vast -e :42002 import arrow
```

libvast/src/format/arrow.cpp
```diff
@@ -74,4 +75,109 @@ bool writer::layout(const std::shared_ptr<::arrow::Schema>& schema) {
...
+  // through existing table slices / record batches from the producer system.
+  VAST_ASSERT(max_slice_size > 0);
+  auto reader = ::arrow::ipc::RecordBatchStreamReader::Open(input_.get());
+  if (!reader.ok()) {
+    return caf::make_error(ec::logic_error, fmt::format("error creating stream "
+                                                        "reader: '{}'",
+                                                        reader.status()));
+  }
+  size_t produced = 0;
+  while (produced < max_events) {
```
dominiklohmann, line 135: As far as I can tell this loop is missing the timeout handling the other readers have. That's important in case the input stream stalls in some way (`vast.import.read-timeout`), often leading to the import process not shutting down properly, or in case the user sets a low batch timeout (`vast.import.batch-timeout`).

Speaking of the the batch timeout and batch size options, I suppose it is intentional that this reader ignores them? Should we warn if the user did set custom options for them? Because of re-batching that mostly shouldn't matter, but if you intend to implement this then there's a re-batching function in the rebuild command that you can re-use.

libvast/src/format/arrow.cpp
```diff
@@ -74,4 +75,109 @@ bool writer::layout(const std::shared_ptr<::arrow::Schema>& schema) {
   return false;
 }
 
+arrow_istream_wrapper::arrow_istream_wrapper(std::shared_ptr<std::istream> input)
+  : input_(std::move(input)), pos_(0) {
+  set_mode(::arrow::io::FileMode::READ);
+}
+
+::arrow::Status arrow_istream_wrapper::Close() {
+  // TODO: is there a way to actually close an istream?
```
dominiklohmann, line 84: Since `input_` is just a shared pointer here you may as well check whether it's a nullptr and use that for signaling whether the stream is closed.

libvast/src/format/arrow.cpp
```diff
@@ -74,4 +75,109 @@ bool writer::layout(const std::shared_ptr<::arrow::Schema>& schema) {
...
+}
+
+void reader::reset(std::unique_ptr<std::istream> in) {
+  input_ = std::make_unique<arrow_istream_wrapper>(std::move(in));
+}
+
+caf::error reader::module(vast::module x) {
+  module_ = x;
+  return caf::none;
+}
```
dominiklohmann, line 173: We don't make use of this at all because we always use the schema attached to the incoming batches. Should this return an error instead, similar to the NetFlow/IPFIX reader?

```c++
caf::error reader::module([[maybe_unused]] class module module) {
  // The VAST types are automatically generated and cannot be changed.
  return caf::make_error(ec::no_error,
                         "schema cannot be changed as it is generated "
                         "dynamically");
}
```

libvast/src/system/application.cpp
```diff
@@ -343,6 +343,7 @@ auto make_command_factory() {
     {"import test", import_command},
     {"import zeek", import_command},
     {"import zeek-json", import_command},
+    {"import arrow", import_command},
```
dominiklohmann, line 346: This is lacking the remote dual to import, namely `spawn source arrow`. This has to be added both in this file and in `node.cpp`.

changelog/unreleased/features/2707--arrow-import.md
```diff
@@ -0,0 +1,2 @@
+VAST can now import Arrow IPC data, which is the same format it already
+supports for export.
```
dominiklohmann, line 2: Passive voice :bomb:

libvast/include/vast/format/arrow.hpp
```diff
@@ -55,4 +56,54 @@ class writer : public format::writer {
...
+  /// @param in new input stream.
+  void reset(std::unique_ptr<std::istream> in) override;
+
+  caf::error module(vast::module x) override;
+
+  [[nodiscard]] vast::module module() const override;
+
+  [[nodiscard]] const char* name() const override;
+
+protected:
```
dominiklohmann, line 100: The `reader` class is marked `final`, so there cannot be a difference between `protected` and `private` visibility here. I would recommend just using the latter to make it a bit less confusing.

libvast/src/format/arrow.cpp
```diff
@@ -74,4 +75,109 @@ bool writer::layout(const std::shared_ptr<::arrow::Schema>& schema) {
...
+  return input_->eof();
+}
+
+::arrow::Result<int64_t> arrow_istream_wrapper::Tell() const {
+  return pos_;
+}
+
+::arrow::Result<int64_t>
+arrow_istream_wrapper::Read(int64_t nbytes, void* out) {
+  input_->read(reinterpret_cast<char*>(out), nbytes);
```
dominiklohmann, line 98: A static cast should do here if I'm not mistaken.

dominiklohmann (commented):
Just a peek 👀 Super exciting stuff! 🚀

libvast/src/format/arrow.cpp
dispanser: Thanks, I didn't see that! I was wondering, but couldn't escape the API design.

I'll add a similar error, even though `ec::no_error` doesn't strike me as very descriptive here.

libvast/src/format/arrow.cpp
dispanser: I changed the method to set `input_ = nullptr` on `Close()` and check for `input_` on `closed()`. Hope that does the trick.

libvast/src/format/arrow.cpp
dispanser: With this change, `vast spawn source arrow` now prints:

```
[14:14:34.079] arrow-reader source failed to set schema: !! logic_error: schema is derived from the Arrow input and can't be changed
```

I don't understand the differences between `import` and `spawn source` enough to understand why it doesn't happen for `vast import`. I wonder if a debug-level log message may be the better approach to prevent this error from popping up. FWIW, the data _is_ actually imported with `vast spawn source arrow`.

libvast/src/format/arrow.cpp
dispanser: The initial plan was to make a relatively simple import, irgnoring the batch sizes alltogether and stick with the granularity of the upstream VAST node.

As you mention, rebuild should take care of mismatches in table slice size with the target node config.

I implemented the timeout checks, but find it difficult to test.

dominiklohmann: I agree that we should not take the batch timeout and the batch size into consideration here.

I don't think your implementation works correctly for the read timeout, though. The call to get the next batch from the reader appears to be blocking, so it by definition does not respect the timeout.

libvast/src/format/arrow.cpp
tobim: I think you actually have to use `ec::no_error` here. That is the only alternative that makes the `caf::error` false-y.

libvast/src/format/arrow.cpp
dispanser: I see your point - what I've written there doesn't make much sense. I revisited the csv reader as my go-to reference implementation, and I see two different concepts wrt timing:

1. `read_timeout_` (20ms default); this seems to be the "real deal" in terms of stalled input. 
First of all, I don't see a way to integrate such a concept into the arrow record batch reader without introducing a separate thread / actor, because it's regular blocking I/O. 
Second, I'm in doubt whether this is actually what we want here. The use case of keeping one VAST node "in sync" with another one by doing 
`vast export --continuous arrow <query> | vast import arrow`. In such a case, stalling on read seems to be normal behavior, and not an exception.
2. `batch_timeout_`: we may not want to be involved in this for now, just taking the input batch granularity, relying on rebuild to optimize the data async. However, I wonder why the csv implementation appears to stop processing when
```c++
last_batch_sent_ + batch_timeout_ < reader_clock::now()
```
Shouldn't it just `finis()` and prepare for the next batch?

dominiklohmann: Also @tobim @lava, your input would be highly appreciated on this since we spent so much time back then to implement the read timeouts correctly.

dominiklohmann: After looking at this earlier today in a call together we quickly agreed to write a follow-up story for now and not to respect the read timeout for now.

dispanser: We agreed to delay the correct implementation of read timeouts to a later stage, probably as soon as we face the consequences of skipping this here. It's a bit tricky to implement, and the feature can already be useful.

libvast/src/format/arrow.cpp
dispanser: Switched to `ec::no_error` and the log message disappears.

changelog/v2.4.0-rc1/features/2707--arrow-import.md
```diff
@@ -0,0 +1,2 @@
+VAST now imports Arrow IPC data, which is the same format it already
+supports for export.
```
dominiklohmann, line 2: This file is in the wrong directory, it should be in `changelog/unreleased`.

dominiklohmann (approved):

cmake/VASTPackage.cmake
```diff
@@ -37,6 +37,9 @@ set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE")
 set(CPACK_RESOURCE_FILE_README "${CMAKE_CURRENT_SOURCE_DIR}/README.md")
 set(CPACK_INSTALLED_DIRECTORIES "/var/lib/vast" "/var/log/vast")
 
+set(CPACK_DEBIAN_PACKAGE_EPOCH "2")
+set(CPACK_DEBIAN_PACKAGE_RELEASE "2")
```
dominiklohmann, line 41: I believe this is in the wrong PR here.

changelog/v2.4.0-rc1/features/2707--arrow-import.md
dominiklohmann, line 2: Please also add documentation before merging here: https://vast.io/docs/use/ingest

dispanser: Added some docs, also for `export arrow`. A quick pass over that is appreciated. (maybe @mavam instead)?

dispanser: resolving as @mavam added a commit dubbed "fine-tune arrow docs", so I assume it's fine.

PR #2652, opened 2022-10-25
TheHive and Cortex Compose scripts

rdettai:
This is a complete revamp of the https://github.com/tenzir/vast/pull/2532 with a more stable init script.

### :memo: Reviewer Checklist

Review this pull request by ensuring the following items:

- [x] All user-facing changes have changelog entries
- [ ] User-facing changes are reflected on [vast.io](https://github.com/tenzir/vast/tree/master/web)

Todo:
- [x] Add TheHive orgadmin
- [x] Configure analyzer
- [ ] ~replace pyvast with the new bindings~ To be done separately to close this off

PR #2710, opened 2022-11-17
Fix TheHive compose tests

dispanser:
This is addressing multiple issues that bubbled up after https://github.com/tenzir/vast/pull/2652
- There was a `$` missing before `exit_code`
- Lint issues
- A transient failure in the Cortex init script due to concurrent access to a mounted volume


rdettai:
There is still a [transient issue](https://app.shortcut.com/tenzir/story/39576/transient-error-on-thehive-init), but I think we can merge this already and collect more data on cases where it fails

dominiklohmann (approved):
I agree with wanting to merge this but making the check optional. Thanks for this!

dominiklohmann:
Merging this in preparation for the next release candidate.

PR #2714, opened 2022-11-17
Enable clangd IWYU integration

dominiklohmann:
When pairing with @Dakostu earlier today I noticed that his Qt Creator showed inline hints for include-what-you-use via clangd. Turns out this is an option in clangd that we can enable, and Qt Creator does so by default. I think this makes for a better dev experience for everyone, so let's enable it by default instead of relying on some editor's custom flags for clangd.

I've also done a small pass over some headers (chosen randomly) to test how well this works. IWYU has come a long way since we first tried using it a few years ago.

### :memo: Reviewer Checklist

### DIFF
--- a/libvast/include/vast/concept/parseable/core.hpp
+++ b/libvast/include/vast/concept/parseable/core.hpp
@@ -8,26 +8,26 @@
 
 #pragma once
 
-#include "vast/concept/parseable/core/action.hpp"
-#include "vast/concept/parseable/core/and.hpp"
-#include "vast/concept/parseable/core/as.hpp"
-#include "vast/concept/parseable/core/choice.hpp"
-#include "vast/concept/parseable/core/difference.hpp"
-#include "vast/concept/parseable/core/end_of_input.hpp"
-#include "vast/concept/parseable/core/epsilon.hpp"
-#include "vast/concept/parseable/core/guard.hpp"
-#include "vast/concept/parseable/core/ignore.hpp"
-#include "vast/concept/parseable/core/kleene.hpp"
-#include "vast/concept/parseable/core/list.hpp"
-#include "vast/concept/parseable/core/literal.hpp"
-#include "vast/concept/parseable/core/maybe.hpp"
-#include "vast/concept/parseable/core/not.hpp"
-#include "vast/concept/parseable/core/operators.hpp"
-#include "vast/concept/parseable/core/optional.hpp"
-#include "vast/concept/parseable/core/plus.hpp"
-#include "vast/concept/parseable/core/repeat.hpp"
-#include "vast/concept/parseable/core/rule.hpp"
-#include "vast/concept/parseable/core/sequence.hpp"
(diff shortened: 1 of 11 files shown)


libvast/include/vast/concept/parseable/core.hpp
```diff
@@ -8,26 +8,26 @@
...
+#include "vast/concept/parseable/core/not.hpp"          // IWYU pragma: export
+#include "vast/concept/parseable/core/operators.hpp"    // IWYU pragma: export
+#include "vast/concept/parseable/core/optional.hpp"     // IWYU pragma: export
+#include "vast/concept/parseable/core/plus.hpp"         // IWYU pragma: export
+#include "vast/concept/parseable/core/repeat.hpp"       // IWYU pragma: export
+#include "vast/concept/parseable/core/rule.hpp"         // IWYU pragma: export
+#include "vast/concept/parseable/core/sequence.hpp"     // IWYU pragma: export
+#include "vast/concept/parseable/core/sequence_choice.hpp" // IWYU pragma: export
+#include "vast/concept/parseable/core/skip.hpp" // IWYU pragma: export
+#include "vast/concept/parseable/core/when.hpp" // IWYU pragma: export
```
dispanser, line 33: That comments are a bit irritating to me - tooling information should not leak into the code base itself, IMHO.

dispanser (approved):
Tooling improvements are always welcome. Especially helpful for people less familiar with the code base and the which-header-is-it puzzle.

libvast/include/vast/concept/parseable/core.hpp
dominiklohmann: These aren't the only ones. We now have the following magic comments in the code line:

- `NOLINT` to disable all clang-tidy rules on the current line
- `NOLINT(rule)` to disable clang-tidy rule `rule` on the current line
- `NOLINTNEXTLINE` and `NOLINTNEXTLINE(rule)` to do the same for the next line
- `clang-format off` and `clang-format on` to disable clang-format for the block of code inbetween
- `IWYU pragma: keep` for an unexplained unused include
- `IWYU pragma: export` for an include that is re-exported

lava: A bit surprising that IWYU doesn't infer this automatically, this kind of pattern is really common e.g. in boost.

Anyways, this place sounds like the perfect use-case for `// IWYU pragma: begin_exports` and `// IWYU pragma: end_exports` ;)

dominiklohmann: clangd's Include Cleaner doesn't support these comments from IWYU, so that's not an option.

Still, I don't think magic comments for the linter are much of an issue. It's better to have the warnings explicitly disabled in some cases than not to have the warnings at all.

lava: Makes sense, no objection from my side then

PR #2716, opened 2022-11-18
Catch exception when encountering invalid query parameters

lava:
This was included in #2667, but since it's not clear if that will be merged in time for 2.4 it deserves to have its own PR.

- Don't throw an exception when encountering invalid query parameters
- Add test with invalid query paramter

### :memo: Reviewer Checklist

### DIFF
--- a/plugins/web/src/server_command.cpp
+++ b/plugins/web/src/server_command.cpp
@@ -113,16 +113,22 @@ request_dispatcher_actor::behavior_type request_dispatcher(
     [self](atom::internal, atom::request, restinio_response_ptr& response,
            const rest_endpoint& endpoint, system::rest_handler_actor handler) {
       // TODO: Also consider params in the request body here.
-      restinio::query_string_params_t string_params
-        = restinio::parse_query(response->request()->header().query());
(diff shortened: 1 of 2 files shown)


plugins/web/src/server_command.cpp
```diff
@@ -113,16 +113,22 @@ request_dispatcher_actor::behavior_type request_dispatcher(
...
       // TODO: Also consider params in the request body here.
-      restinio::query_string_params_t string_params
-        = restinio::parse_query(response->request()->header().query());
+      auto query_params = std::optional<restinio::query_string_params_t>{};
+      try {
+        query_params
+          = restinio::parse_query(response->request()->header().query());
+      } catch (restinio::exception_t& e) {
+        return response->abort(
+          400, fmt::format("failed to parse query parameters: {}\n", e.what()));
```
tobim, line 122: Why do we need a newline at the end?

lava: It makes for a more pleasant user experience when using `curl` (or any command-line tool) instead of a browser.

tobim (approved):

PR #2719, opened 2022-11-18
Add 'pseudonymize' pipeline operator

Dakostu:
Add a `pseudonymize` operator that pseudonymizes IP addresses in records.


web/docs/understand/query-language/operators/pseudonymize.md
```diff
@@ -0,0 +1,18 @@
...
+## Parameters
+
+- `seed: [string]`: The 64-byte seed that describes a hexadecimal value. When the seed is shorter than 64 bytes, the operator will append zeros to match the size; when it is longer, it will truncate the seed.
+- `fields: <list>`: The list of fields to apply the pseudonymization to. During pseudonymization, the operator will ignore specified fields that do not contain IP addresses.
+
+## Example
+
+```yaml
+pseudonymize:
+  seed: "abcdef1234567890987654321fedcbaabcdef1234567890987654321fedcbaab"
```
dominiklohmann, line 14:
```suggestion
  method: crypto-pan
  seed: "abcdef1234567890987654321fedcbaabcdef1234567890987654321fedcbaab"
```

Can you add a method key here for future extensibility, and rewrite the documentation above accordingly?

libvast/test/address.cpp
```diff
@@ -146,3 +148,230 @@ TEST(parseable) {
...
+  for (const auto& [original, pseudonymized] : addresses) {
+    auto original_address = *to<address>(original);
+    auto pseudonymized_address_expectation = *to<address>(pseudonymized);
+
+    auto pseudonymized_adress_actual
+      = address::pseudonymized(original_address, seed);
+
+    REQUIRE_EQUAL(pseudonymized_adress_actual,
+                  pseudonymized_address_expectation);
+  }
```
dominiklohmann, line 376: This exact same code occurs a few times in this file. Should we move it into a free function in an anonymous namespace?

Also, as per the style guide, please remove the whitespace between the lines within functions.

libvast/test/address.cpp
```diff
@@ -146,3 +148,230 @@ TEST(parseable) {
   CHECK(a.is_v6());
   CHECK(to_string(a) == str);
 }
+
+TEST(IPv4 pseudonymization - seed 1) {
+  std::array<vast::address::byte_type, 32> seed
+    = {21,  34,  23,  141, 51,  164, 207, 128, 19, 10, 91, 22, 73, 144, 125, 16,
+       216, 152, 143, 131, 121, 121, 101, 39,  98, 87, 76, 45, 42, 132, 34,  2};
+  // test data from
+  // https://github.com/noinkling/node-cryptopan/blob/main/src/test/test_data.ts
```
dominiklohmann, line 157: No request for changes, but thank you so much for writing extensive tests. This is really nice to see!

libvast/src/address.cpp
```diff
@@ -16,13 +16,103 @@
...
 #include <utility>
 
 namespace vast {
 
+namespace {
+inline uint32_t bitmask32(size_t bottom_bits) {
+  return bottom_bits >= 32 ? 0xffffffff : ((uint32_t{1} << bottom_bits) - 1);
+}
+
+class AddressEncryptor {
```
dominiklohmann, line 35: Please use snake_case in the code base for consistency.

libvast/src/address.cpp
```diff
@@ -16,13 +16,103 @@
...
+#include <openssl/ossl_typ.h>
 #include <sys/socket.h>
 
 #include <cstdlib>
 #include <utility>
 
 namespace vast {
 
+namespace {
+inline uint32_t bitmask32(size_t bottom_bits) {
```
dominiklohmann, line 31:
```suggestion

inline uint32_t bitmask32(size_t bottom_bits) {
```

libvast/src/address.cpp
```diff
@@ -16,13 +16,103 @@
...
+  }
+
+  AddressEncryptor(const AddressEncryptor&) = delete;
+
+  ~AddressEncryptor() {
+    EVP_CIPHER_CTX_cleanup(ctx_);
+    EVP_CIPHER_CTX_free(ctx_);
+  }
+
+  auto encrypt(address::byte_array bytes, size_t byte_offset) {
```
dominiklohmann, line 58: Would it make sense to make this the call operator, so you don't `encryptor.encrypt(...)`, but rather `encryptor(...)`? Just reads a lot nicer to me.

(If you disagree just resolve, I have no particularly strong opinion on this.)

libvast/builtins/operators/pseudonymize.cpp
```diff
@@ -0,0 +1,182 @@
...
+          if (!caf::holds_alternative<address_type>(field.type)) {
+            VAST_ASSERT(false
+                        && "record batch field to be pseudonymized but does "
+                           "not "
+                           "have address type");
+            VAST_WARN("Field {} is to be pseudonymized but does not contain "
+                      "IP "
+                      "address values; skipping pseudonymization",
+                      field);
+            return {{field, array}};
```
dominiklohmann, line 72: Since this is a very local precondition violation (you should not create this transformation if `caf::holds_alternative<address_type>(caf::get<record_type>(layout).field(index).type)` is `false`), you can also just call `die` here instead of having the `if`, e.g., check before even creating the transformation whether the field you're pointing to is of the correct type.

NB: There's a two-argument overload to `VAST_ASSERT`, so you don't need to do `condition && "message"`, but rather can do `condition, "message"`.

libvast/builtins/operators/pseudonymize.cpp
```diff
@@ -0,0 +1,182 @@
...
+          }
+          auto new_array = builder->Finish().ValueOrDie();
+          return {
+            {field, new_array},
+          };
+        };
+        transformations.push_back({index, std::move(transformation)});
+      }
+    }
+    std::sort(transformations.begin(), transformations.end());
```
dominiklohmann, line 99: You also need to remove duplicate transformations after the sorting, which is relevant if the users adds multiple extractors that resolve to the same offsets.

libvast/builtins/operators/pseudonymize.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+  /// Applies the transformation to an Arrow Record Batch with a corresponding
+  /// VAST layout.
+  [[nodiscard]] caf::error
+  add(type layout, std::shared_ptr<arrow::RecordBatch> batch) override {
+    std::vector<indexed_transformation> transformations;
+    for (const auto& field : config_.fields) {
+      for (const auto& index : caf::get<record_type>(layout).resolve_key_suffix(
+             field, layout.name())) {
+        auto transformation = [&](struct record_type::field field,
```
dominiklohmann, line 59: Can you move this lambda out of the two for loops? I think the code would be much easier to read then.

libvast/include/vast/address.hpp
```diff
@@ -86,6 +88,14 @@ class address : detail::totally_ordered<address>, detail::bitwise<address> {
     return result;
   }
 
+  /// Construct a pseudonymized address using the Crypto-PAn algorithm.
+  /// @param original The address to be pseudonymized.
+  /// @param seed 256-bit seed for the cipher and padding.
+  /// @returns A copy of the `original` address with pseudonymized bytes.
+  static address pseudonymized(
+    const address& original,
+    const std::array<byte_type, pseudonymization_seed_array_size>& seed);
```
dominiklohmann, line 97: Most of the functions in our code base are verbs, while classes are usually nouns. In this case, I think it makes sense to name the function `pseudonymize`, and to make it a friend of `address` instead of a static member function.

changelog/unreleased/features/2719--add-pseudonymize-pipeline-operator.md
```diff
@@ -0,0 +1 @@
+VAST has a new `pseudonymize` pipeline operator that pseudonymizes IP addresses in user-specified fields.
```
dominiklohmann, line 1: Please rewrite in active voice (and limit to 80 cols.)

libvast/test/address.cpp
Dakostu: :+1:

web/docs/setup/build.md
```diff
@@ -28,6 +28,7 @@ dependencies and versions.
 |✓|C++ Compiler|C++20 required|VAST is tested to compile with GCC >= 10.0 and Clang >= 13.0.|
 |✓|[CMake](https://cmake.org)|>= 3.19|Cross-platform tool for building, testing and packaging software.|
 |✓|[CAF](https://github.com/actor-framework/actor-framework)|>= 0.17.6|Implementation of the actor model in C++. (Bundled as submodule.)|
+|✓|[OpenSSL](https://www.openssl.org)||Utilities for secure networking and cryptography.|
```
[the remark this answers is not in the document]
Dakostu: Added in 45f64fd2f

web/docs/understand/query-language/operators/pseudonymize.md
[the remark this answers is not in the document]
Dakostu: Fixed in c36af173b

Dakostu: Added a `method` key that currently does nothing in f880b6e08

libvast/src/address.cpp
Dakostu: Removed bad habit in 31020cabe

libvast/src/address.cpp
Dakostu: Good idea. Implemented in 8f1d2581d

libvast/builtins/operators/pseudonymize.cpp
Dakostu: Should be fixed in 061a2d7c8

libvast/builtins/operators/pseudonymize.cpp
Dakostu: Fixed in 163cfe3e8
This makes me wonder why `transformations` are supposed to be in a `std:vector` after all...

libvast/builtins/operators/pseudonymize.cpp
Dakostu: Fixed in 7b202c871

libvast/include/vast/address.hpp
Dakostu: I oriented myself according to the static `v4()` and `v6()` methods (is `v6` a verb?). I renamed the method to `pseudonymize` in

changelog/unreleased/features/2719--add-pseudonymize-pipeline-operator.md
Dakostu: Fixed in d80c8d03c

libvast/builtins/operators/pseudonymize.cpp
Dakostu: Actually, I moved this check outside of the lambda in https://github.com/tenzir/vast/pull/2719/commits/d3d0d4aa4e3efd6a7a71c4d305d7c44c708bc9a6

libvast/include/vast/address.hpp
dominiklohmann: The friend just reduces the function overload set, this is a small win for compilation time.

libvast/builtins/operators/pseudonymize.cpp
dominiklohmann: I don't think there's a particular reason for this, so if you want to make this change please go ahead. Such refactorings are always welcome as an initiative (but maybe in a different branch).

libvast/src/address.cpp
```diff
@@ -16,13 +16,104 @@
...
 
+namespace {
+
+inline uint32_t bitmask32(size_t bottom_bits) {
+  return bottom_bits >= 32 ? 0xffffffff : ((uint32_t{1} << bottom_bits) - 1);
+}
+
+class address_encryptor {
+public:
+  address_encryptor(const std::array<address::byte_type, 32>& key) {
```
dominiklohmann, line 38: Single-argument constructors should be marked explicit. If you want to make a constructor implicit intentionally, use `explicit(false)`.

libvast/src/address.cpp
```diff
@@ -16,13 +16,104 @@
...
+    EVP_CipherUpdate(ctx_, pad_.data(), &pad_out_len, key.data() + block_size_,
+                     block_size_);
+  }
+
+  address_encryptor(const address_encryptor&) = delete;
+
+  ~address_encryptor() {
+    EVP_CIPHER_CTX_cleanup(ctx_);
+    EVP_CIPHER_CTX_free(ctx_);
+  }
```
dominiklohmann, line 57: There are two design bugs in the highlighted code:
- Destructors, move constructors, and move-assignment operators should always be marked noexcept (see [here](https://quuxplusone.github.io/blog/2022/08/26/vector-pessimization/))
- Per the [rule of five](https://en.cppreference.com/w/cpp/language/rule_of_three), if you define any of the five then you must also define the remaining.

You can avoid all of this by applying the rule of zero, which is much cleaner design anyways: Make `ctx_` a `std::unique_ptr` with a custom deleter that calls the two functions, and then simply remove the explicitly defaulted copy constructor and the custom destructor.

libvast/src/address.cpp
```diff
@@ -16,13 +16,104 @@
...
+    }
+    return bytes;
+  }
+
+private:
+  static constexpr inline auto msb_of_byte_mask = 0b10000000;
+  EVP_CIPHER_CTX* ctx_;
+  const EVP_CIPHER* cipher_;
+  int block_size_;
+  std::vector<address::byte_type> pad_;
```
dominiklohmann, line 73:
```suggestion
  EVP_CIPHER_CTX* ctx_ = {};
  const EVP_CIPHER* cipher_ {};
  int block_size_ = {};
  std::vector<address::byte_type> pad_ = {};
```

Even though it's not strictly needed for the vector, it is very much a good habit to get into for all member variables. The first three are trivially constructible and may as such contain garbage values if you don't explicitly initialize them.

libvast/builtins/operators/pseudonymize.cpp
```diff
@@ -0,0 +1,191 @@
...
+        if (!caf::holds_alternative<address_type>(
+              caf::get<record_type>(layout).field(index).type)) {
+          VAST_ASSERT(false, "record batch field to be pseudonymized but does "
+                             "not "
+                             "have address type");
+          VAST_WARN(fmt::format("Field {} is to be pseudonymized but does not "
+                                "contain "
+                                "IP "
+                                "address values; skipping pseudonymization",
+                                field_name));
```
dominiklohmann, line 96: I think this can just be a debug message; it's not worth of a warning, and it's certainly not something that should trigger an assertion.

The `VAST_*` family of macros also already do the formatting for you, so there's no need to call `fmt::format` youself. 

For log messages, please adhere to noun-verb-object order, e.g., `VAST_DEBUG("pseudonymize operator skips field '{}' of unsupported type '{}'", ...)`.

dominiklohmann (commented):
Some additional small things I've found.

libvast/builtins/operators/pseudonymize.cpp
Dakostu: Fixed in https://github.com/tenzir/vast/pull/2719/commits/3dae63715e3a1f2e5444ff71519e7736cd3605bf

libvast/src/address.cpp
Dakostu: Good idea with the `unique_ptr`, it certainly avoids me having to `delete` the rest of the constructors.
Implemented in https://github.com/tenzir/vast/pull/2719/commits/01eaddb2b09fbaeaf0cc8ead313af0af0ef9bd46

libvast/builtins/operators/pseudonymize.cpp
Dakostu: I'll put the refactoring off to a later date. Maybe there is a valid reason but I can't think of any right now.

dominiklohmann (approved):
Let's ship it! (once we're out of feature freeze)

PR #2720, opened 2022-11-19
Remove the generated changelog from the repository

dominiklohmann:
The changelog is being created on build dynamically, and an always up-to-date version is available at vast.io/changelog. Having a duplicate from the last release in the repository needlessly adds overhead to developer workflows.


.github/workflows/vast.yaml
```diff
@@ -278,13 +260,14 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
-      - name: Delete existing CHANGELOG.md
-        run:
-          rm CHANGELOG.md
-      - name: Download new CHANGELOG.md
+      - name: Delete existing changelog
+        run: |
+          rm: web/src/pages/changelog.md
```
mavam, line 265: Is this step still necessary? I would imagine it suffices to download (= overwrite) the existing one.

web/src/pages/changelog.md
```diff
@@ -0,0 +1,5 @@
+# Changelog
+
+This is a stub file. To generate the changelog, build VAST and move the
+generated `CHANGELOG.md` file in the build directory to
```
mavam, line 4:
```suggestion
generated `CHANGELOG.md` file from the build directory to
```

web/src/pages/changelog.md
```diff
@@ -0,0 +1,5 @@
...
+`src/pages/changelog.md`.
```
mavam, line 5: This should be in .gitignore then.

web/Makefile
```diff
@@ -21,9 +18,6 @@ all: changelog dot-github rfc quarto
 
 changelog: $(CHANGELOG_OUTPUT)
```
mavam, line 19: Why keep this target (and the reference above)?

mavam (approved):
Thanks for taking the initiative on this.

.github/workflows/vast.yaml
dominiklohmann: The download doesn't overwrite but rather error if the file exists. We do want to have the stub file, though, otherwise local builds do not work.

web/src/pages/changelog.md
dominiklohmann: That is not correct. The stub file has to be in the repo, otherwise local docs build will not work without copying the changelog.

web/Makefile
dominiklohmann: No reason to, I overlooked this one.

web/src/pages/changelog.md
mavam: Yes, my point is actually different. I'm saying keep the stub file _and_ add it to .gitignore, so generating it (which overwrites the stub) doesn't cause the repo to get dirty.

.github/workflows/vast.yaml
mavam: Okay, if the download fails otherwise, there's no way around it.

web/src/pages/changelog.md
dominiklohmann: We agreed over lunch to do this only if it turns out to be necessary, i.e., not now as part of this PR.

PR #2722, opened 2022-11-21
Temporarily disable publishing on PyPI in CI

dominiklohmann:
Our new Python bindings are currently waiting for a PEP 541 request to resolve a name-squatting issue on PyPI. As such, the CI will not be able to publish the package. Now with the bundling from #2636 being delayed until after VAST v2.4 this is not a big issue, as that leaves us some more time to resolve the remaining issues until the next release. We do, however, need to disable the publishing for v2.4 as that's already in feature freeze.


lava (approved):
Makes sense

PR #2725, opened 2022-11-22
Fix exit code check in Docker Compose tests

dominiklohmann:
These tests did not properly check for the exit code of all spawned containers since we started explicitly setting a container name. This fixes that oversight.

### DIFF
--- a/docker/test
+++ b/docker/test
@@ -32,18 +32,12 @@ else
   export VAST_CONTAINER_REGISTRY="${VAST_CONTAINER_REGISTRY:-ghcr.io}"
   export VAST_CONTAINER_REF="${VAST_CONTAINER_REF:-$(git rev-parse --abbrev-ref HEAD | sed -E 's/[^a-zA-Z0-9._-]+/-/g;s/^-*//')}"
   if [[ "${VAST_CONTAINER_REF}" == "master" ]]; then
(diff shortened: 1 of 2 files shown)


docker/test
```diff
@@ -61,28 +61,28 @@ sleep 3
...
 
 # Print logs for easier debugging in case something went wrong.
-docker compose ${compose_files} ps
-docker compose ${compose_files} logs
+docker compose ps
+docker compose logs
 
 # Assert that all containers that exited have exited cleanly.
 exited_all="$(docker ps --all --filter "name=${COMPOSE_PROJECT_NAME}" --filter 'status=exited' -q | sort -u)"
 exited_cleanly="$(docker ps --all --filter "name=${COMPOSE_PROJECT_NAME}" --filter 'exited=0' -q | sort -u)"
```
rdettai, line 83: Isn't `docker compose ps --all` an alternative to `docker ps --all --filter "name=${COMPOSE_PROJECT_NAME}"`. That would spare us the dance around the container name.

dominiklohmann: TIL that `docker compose ps` exists. That does make a lot of things a lot easier, yes! I will refactor the script accordingly.

dominiklohmann: I implemented this accordingly. This is so much better!

mavam (approved):
Looks like the feedback is addressed.

PR #2728, opened 2022-11-23
Add ability to pass POST request parameters as JSON

lava:
After this commit, the web plugin will look at the top-level keys of a JSON object passed in the request body of a POST request in order to find matches for the parameters declared by an endpoint. For example, using the query endpoint a request would look like

    curl -XPOST -H "Content-Type: application/json" \
        -d'{"expression": "129.168.0.1"}' \
        http://127.0.0.1:42001/api/v0/query/new


libvast/builtins/endpoints/export.cpp
```diff
@@ -74,6 +74,53 @@ static auto const* SPEC_V0 = R"_(
...
+    description: Export data from VAST according to a query. The query must be a valid expression in the VAST Query Language. (see https://vast.io/docs/understand/query-language)
+    parameters:
+      - in: query
+        name: expression
+        schema:
+          type: string
+        required: true
+        default: A query matching every event.
+        description: The query expression to execute.
+        example: ":ip in 10.42.0.0/16"
```
dominiklohmann, line 89:
```suggestion
        example: ":addr in 10.42.0.0/16"
```

PR #2730, opened 2022-11-24
Make data predicate evaluation column-major

dominiklohmann:
This is a major performance optimization for expression evaluation that essentially fills in one of the last TODOs from tenzir/vast#2440: Making data predicate evaluation itself column-major, i.e., moving as much as possible of the evaluation out of the hot loop.

Here's how the new expression algorithm works, roughly, with (3b) and (4) being the new and improved steps:
1. Normalize the selection bitmap from the dense index result to the length of the batch + offset.
2. Determine whether the expression is empty, a connective of some sort, or a predicate. For connectives, resolve them recursively and combine the resulting bitmaps accordingly.
3. Evaluate predicates: a) If it's a meta extractor, operate on the batch metadata. In case of a match, the selection bitmap is the result directly. b) If it's a data predicate, access the desired array, and lift the resolved types for both sides of the predicate into a compile time context for the column evaluator.
4. The column evaluator has specialization based on the three-tuple of lhs type, relational operator, and rhs view. The generic fall back case iterates over all fields per the selection bitmap to do the evaluation using the cell evaluator, which can be specialized per relational operator.

We've seen some pretty great results. When using a query on a 6M event database that's doing multiple substring searches (for which we have neither sparse nor dense index support), the full scan of the database is now over 2x quicker than before.

Concretely, we measured the following numbers using Hyperfine with three warmup runs and ten measured runs for this commit compared to its merge-base with master `de86af3fca`.

New:

Time (mean ± σ):     11.038 s ±  0.360 s    [User: 28.600 s, System: 4.430 s]
Range (min … max):   10.808 s … 11.569 s    10 runs

Old:

Time (mean ± σ):     22.805 s ±  0.616 s    [User: 45.060 s, System: 4.826 s]
Range (min … max):   22.042 s … 23.568 s    10 runs

This was a Hackathon group effort with @patszt, @dispanser, and @dominiklohmann.

Co-authored-by: Patryk Sztyglic <patryk.sztyglic@gmail.com>
Co-authored-by: Thomas Peiselt <pi@kulturguerilla.org>
Co-authored-by: Dominik Lohmann <mail@dominiklohmann.de>


libvast/src/evaluate.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+// A utility function for evaluating meta extractors in predicates. This is
+// always a yes or no question per batch, so the function does not have to deal
+// with bitmaps at all.
+bool evaluate_meta_extractor(const table_slice& slice,
+                             const meta_extractor& lhs, relational_operator op,
+                             const data& rhs) {
+  switch (lhs.kind) {
+    case meta_extractor::kind::type:
+      return evaluate(materialize(slice.layout().name()), op, rhs);
```
mavam, line 37: Is materialization needed here?

libvast/src/evaluate.cpp
```diff
@@ -0,0 +1,465 @@
...
+    case meta_extractor::kind::field: {
+      const auto* s = caf::get_if<std::string>(&rhs);
+      if (!s) {
+        VAST_WARN("#field can only compare with string");
```
mavam, line 41: Why not pattern?

libvast/src/evaluate.cpp
```diff
@@ -0,0 +1,465 @@
...
+      return neg != result;
+    }
+    case meta_extractor::kind::import_time:
+      return evaluate(data{slice.import_time()}, op, rhs);
+  }
+  __builtin_unreachable();
+}
+
+template <relational_operator Op>
+struct cell_evaluator;
```
mavam, line 69: NB: I really like the way you simplified the problem from triple dispatch to double dispatch!

libvast/src/evaluate.cpp
```diff
@@ -0,0 +1,465 @@
...
+};
+
+template <>
+struct cell_evaluator<relational_operator::match> {
+  static bool evaluate(auto, auto) noexcept {
+    return false;
+  }
+
+  static bool evaluate(view<std::string> lhs, view<pattern> rhs) noexcept {
+    return rhs.match(lhs);
```
mavam, line 168: You also dispatch to `cell_evaluator<equals>` to have only one place in the code where you dispatch to the concrete implementation.

libvast/src/evaluate.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+// Expression evaluation takes place in multiple resolution steps:
+// 1. Normalize the selection bitmap from the dense index result to the length
+//    of the batch + offset.
+// 2. Determine whether the expression is empty, a connective of some sort, or
+//    a predicate. For connectives, resolve them recursively and combine the
+//    resulting bitmaps accordingly.
+// 3. Evaluate predicates:
+//    a) If it's a meta extractor, operate on the batch metadata. In case of a
+//       match, the selection bitmap is the result directly.
```
mavam, line 342:
```suggestion
//       match, the selection bitmap is the very result.
```

libvast/src/evaluate.cpp
```diff
@@ -0,0 +1,465 @@
...
+//    b) If it's a data predicate, access the desired array, and lift the
+//       resolved types for both sides of the predicate into a compile time
```
mavam, line 344:
```suggestion
//       resolved types for both sides of the predicate into a compile-time
```

libvast/src/evaluate.cpp
```diff
@@ -0,0 +1,465 @@
...
+    },
+  };
+  const auto evaluate_expression
+    = [&](const auto& self, const expression& expr, ids selection) -> ids {
+    const auto evaluate_expression_impl = detail::overload{
+      [&](const caf::none_t&, const ids&) {
+        return ids{offset + num_rows, false};
+      },
+      [&](const negation& negation, const ids& selection) {
+        return selection ^ self(self, negation.expr(), selection);
```
mavam, line 415: Can you document why XOR does The Right Thing?

mavam (approved):
Looks good at a high level. I only left some minor questions/comments. If CI passes, this would work for me.

libvast/src/evaluate.cpp
dominiklohmann: I can, but note that this code was only moved, we did not touch it at all.

libvast/src/evaluate.cpp
dominiklohmann: We did not touch this code; we could probably delegate to the cell_evaluator here instead, then we wouldn't need to materialize at all.

libvast/src/evaluate.cpp
dominiklohmann: Thanks!

libvast/src/evaluate.cpp
mavam: @tobim IIRC you wrote this. Can't we just dispatch to the new `evaluate` here?

mavam: That's actually quite implicit. Shouldn't `#field == /.*name/` do a suffix match on a fully qualified name?

(We'll create a story for this once we agree.)

dominiklohmann: That's a totally different semantic, because `zeek.conn.id.orig_h` will be matched by `#field == "X"` for any of `orig_h`, `id.orig_h`, `zeek.conn.orig_h`, but not by `_h`.

dominiklohmann: In my opinion, this operator is rather useless: `#field == "X"` is just the same as `X != nil`. We don't use it anywhere, really, so I'd rather just remove it.

libvast/src/evaluate.cpp
dominiklohmann: I did end up pushing a commit that implements this in terms of the cell evaluator. It's a lot more verbose, but hey, it's one step further towards not needing `evaluate(data, op, data) -> bool` at all anymore.

libvast/src/evaluate.cpp
dominiklohmann: I added a comment for this, please check back. :)

libvast/src/evaluate.cpp
mavam: Let's pencil the removal for a separate story. I do see some value in it, but the discussion goes beyond the scope of this PR. It should probably also something we discuss in the community Slack.

libvast/src/evaluate.cpp
mavam: I'm not quite getting it. This lacks context for me:
> For bits that are zero because of the offset the result bitmap is also always guaranteed to be zero [..]

libvast/src/evaluate.cpp
```diff
@@ -0,0 +1,518 @@
...
+bool evaluate_meta_extractor(const table_slice& slice,
+                             const meta_extractor& lhs, relational_operator op,
+                             const data& rhs) {
+  switch (lhs.kind) {
+    case meta_extractor::kind::type: {
+      switch (op) {
+#define VAST_EVAL_DISPATCH(op)                                                 \
+  case relational_operator::op: {                                              \
+    auto f = [&](const auto& rhs) noexcept {                                   \
+      return cell_evaluator<relational_operator::op>::evaluate(                \
```
mavam, line 305: This is where the name `cell_evaluator` breaks. We can use it a lot more broadly, it seems.

libvast/src/evaluate.cpp
dominiklohmann: Is it clearer to you if I write that for `selection & self(self, negation.expr(), selection)` is the same as `self(self, negation.expr(), selection)`?

libvast/src/evaluate.cpp
```diff
@@ -0,0 +1,518 @@
...
+// A utility function for evaluating meta extractors in predicates. This is
+// always a yes or no question per batch, so the function does not have to deal
+// with bitmaps at all.
+bool evaluate_meta_extractor(const table_slice& slice,
+                             const meta_extractor& lhs, relational_operator op,
+                             const data& rhs) {
+  switch (lhs.kind) {
+    case meta_extractor::kind::type: {
+      switch (op) {
+#define VAST_EVAL_DISPATCH(op)                                                 \
```
mavam, line 302: Couldn't this be a templated function instead of a macro?

dominiklohmann: Not without reflection or a lot of __FUNCTION_NAME__ magic à la [magic_enum](https://github.com/Neargye/magic_enum/).

libvast/src/evaluate.cpp
mavam: I just thought that the notion of cell is no longer applicable in this context.

libvast/src/evaluate.cpp
mavam: No. 🙂

It should be the mental model that's conveyed, i.e., why the logical operation on bitmaps make sense.

But we don't have to go crazy here. This suits the bill. Someone who goes deeper can draw it out on paper.

libvast/src/evaluate.cpp
dominiklohmann: It does somewhat, but then this whole cell evaluator here is in an anonymous namespace and only available in this one translation unit, so I must say that unless someone comes up with a convincingly better name I don't see much reason to change that.

PR #2601, opened 2022-09-30
Disable ASan in macOS CI job

dominiklohmann:
### :memo: Reviewer Checklist

Review this pull request by ensuring the following items:

- [ ] All user-facing changes have changelog entries
- [ ] User-facing changes are reflected on [vast.io](https://github.com/tenzir/vast/tree/master/web)

PR #2733, opened 2022-11-24
Re-enable ASan in macOS CI

dominiklohmann:
Now that GitHub Actions always uses Arrow 10 in macOS, it's time to see whether the issue from tenzir/vast#2601 is fixed.

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -606,11 +606,7 @@ jobs:
             cc: clang
             cxx: clang++
             dependencies-script-path: scripts/macOS/install-dev-dependencies.sh
-            # Note that we disable address sanitizer because of a hit inside Arrow
-            # when roundtripping table slices that started occurring with the
-            # 9.0.0_3 release of Arrow on Homebrew. It's likely a false positive,
-            # but we haven't had the time to investigate it further yet.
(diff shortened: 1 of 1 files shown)


.github/workflows/vast.yaml
```diff
@@ -606,11 +606,7 @@ jobs:
             cc: clang
             cxx: clang++
             dependencies-script-path: scripts/macOS/install-dev-dependencies.sh
-            # Note that we disable address sanitizer because of a hit inside Arrow
-            # when roundtripping table slices that started occurring with the
-            # 9.0.0_3 release of Arrow on Homebrew. It's likely a false positive,
-            # but we haven't had the time to investigate it further yet.
-            cmake-extra-flags: -DVAST_ENABLE_ASAN:BOOL=OFF -DVAST_ENABLE_BUNDLED_CAF:BOOL=ON
+            cmake-extra-flags: -DVAST_ENABLE_BUNDLED_CAF:BOOL=ON
```
mavam, line 609: Why do you also disable bundling CAF here?

mavam (approved):
CI will tell.

.github/workflows/vast.yaml
dominiklohmann: I enable it, not disable it. That is because on macOS, CAF 0.17.6 is not available on Homebrew, so we need to use the bundled version.

PR #2734, opened 2022-11-24
Allow read access to user home dir in the systemd unit

tobim:
In case the XDG config directory of the vast user is not in `/var/lib/vast` the sandboxing configuration of the unit would deny access to the configuration files in that place, causing the main process to exit.

In theory the config directory may be outside of the home directory, but systemd has no mechanism to specify that, so we make the home dir read only instead.


dominiklohmann (approved):
Ready to go modulo changelog entry! Thanks for fixing this so quickly

changelog/unreleased/bug-fixes/2734--relax-systemd-service-sandbox.md
```diff
@@ -0,0 +1,2 @@
+The systemd service for vast no longer fails if the home directory of the vast
```
mavam, line 1:
```suggestion
The systemd service no longer fails if the home directory of the VAST
```

tobim: The user name is not capitalized, so I only took the other edit.

PR #2736, opened 2022-11-24
Upload Docker images tagged with commit SHA on release

dominiklohmann:
This was an oversight we recently introduced, and it caused the Docker Compose CI job to fail early in the VAST v2.4 release candidate cycle. Essentially, we always have to upload images with tagged with their SHA so downstream jobs in CI can refer to them consistently.


mavam (approved):
Looks like it will work.

PR #2738, opened 2022-11-25
Avoid old CalVer-based Debian packages overriding new SemVer-based packages

dominiklohmann:
VAST switched it's versioning scheme from CalVer to SemVer, so we have to set the package epoch so an older package with calendar-based versioning is not considered newer than this one. This for whatever reason implies that the package revision must be set, which is just zero for us.

Additionally, this fixes the links in our documentation for downloading the prebuilt packages by aligning them with reality.


lava (approved):
Logically this looks good to me, assuming that `CPACK_DEBIAN_PACKAGE_EPOCH` does what one would expect and that `steps.get_artifact_names.outputs.deb` holds the correct filename.

tobim:
I'm going to reproduce this error locally and try to fix it with quicker eval.

cmake/VASTPackage.cmake
```diff
@@ -37,6 +37,13 @@ set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE")
 set(CPACK_RESOURCE_FILE_README "${CMAKE_CURRENT_SOURCE_DIR}/README.md")
 set(CPACK_INSTALLED_DIRECTORIES "/var/lib/vast" "/var/log/vast")
 
+# VAST switched it's versioning scheme from CalVer to SemVer, so we have to set
+# the package epoch so an older package with calendar-based versioning is not
+# considered newer than this one. This for whatever reason implies that the
+# package revision must be set, which is just zero for us.
+set(CPACK_DEBIAN_PACKAGE_EPOCH "1")
+set(CPACK_DEBIAN_PACKAGE_RELEASE "0")
```
tobim, line 45:
```suggestion
set(CPACK_DEBIAN_PACKAGE_RELEASE "1")
```
Looks like the cpack module doesn't use `if(DEFINED ...)` to check for that variable. Let's just set it to truthy value as a workaround.

tobim (approved):
With the suggested change the filename comes out as `vast_1:2.4.0-rc2-16-ga7debff9ee-1_amd64.deb` for example.

dominiklohmann:
> With the suggested change the filename comes out as `vast_1:2.4.0-rc2-16-ga7debff9ee-1_amd64.deb` for example.

That's fine. For releases we set the name manually and in all other cases the CI greps for *.deb files so it always uploads the correct one.

I also rebased on recent master to get the Parquet fixes in.

PR #2740, opened 2022-11-25
Remove all references to "--plugins=all" in dev section of docs

Dakostu:


dominiklohmann:
@mavam I've requested you for review here. This follows up on making `vast.plugins=all` the default. There were simply some places in the docs that still referenced having to set that explicitly.

mavam (approved):

PR #2742, opened 2022-11-28
Make it easy to create docker images with Nix

tobim:
This can be used with
```
nix run .#stream-image | docker load
```
It reuses the regular static binary output so it won't rebuild VAST if the static binary is already available.

According to `docker image ls` the resulting image weighs 117 Mb,
significantly lighter than the Debian based image which is ~330 Mb.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -43,6 +43,14 @@
       };
       apps.vast = flake-utils.lib.mkApp { drv = packages.vast; };
       apps.vast-static = flake-utils.lib.mkApp { drv = packages.vast-static; };
+      apps.stream-image = {
+        type = "app";
+        program = "${pkgs.dockerTools.streamLayeredImage {
+          name = "tenzir/vast";
+          tag = "latest";
+          config.Cmd = [ "${self.packages.${system}."vast-static"}/bin/vast" ];
+        }}";
+      };
       apps.default = apps.vast;
       devShell = import ./shell.nix { inherit pkgs; };
       hydraJobs = { inherit packages; } // (
--- a/flake.nix
+++ b/flake.nix
@@ -43,7 +43,7 @@
       };
       apps.vast = flake-utils.lib.mkApp { drv = packages.vast; };
       apps.vast-static = flake-utils.lib.mkApp { drv = packages.vast-static; };
-      apps.stream-image = {
+      apps.stream-static-image = {
         type = "app";
         program = "${pkgs.dockerTools.streamLayeredImage {
           name = "tenzir/vast-slim";
@@ -52,6 +52,7 @@
             vast-dir = "/var/lib/vast";
           in {
             Entrypoint = [ "${self.packages.${system}."vast-static"}/bin/vast" ];
+            CMD = [ "--help" ];
             Env = [
               # When changing these, make sure to also update the entries in the
               # Dockerfile.
--- a/flake.nix
+++ b/flake.nix
@@ -48,7 +48,23 @@
         program = "${pkgs.dockerTools.streamLayeredImage {
           name = "tenzir/vast";
           tag = "latest";
-          config.Cmd = [ "${self.packages.${system}."vast-static"}/bin/vast" ];
+          config = let
+            vast-dir = "/var/lib/vast";
+          in {
+            Entrypoint = [ "${self.packages.${system}."vast-static"}/bin/vast" ];
+            Env = [
+              "VAST_ENDPOINT=0.0.0.0:42000"
+              "VAST_DB_DIRECTORY=${vast-dir}"
+              "VAST_LOG_FILE=/var/log/vast/server.log"
+            ];
+            ExposedPorts = {
+              "42000/tcp" = {};
+            };
+            WorkingDir = "${vast-dir}";
+            Volumes = {
+              "${vast-dir}" = {};
+            };
+          };
         }}";
       };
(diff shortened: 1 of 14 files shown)


dominiklohmann:
This is not a drop-in replacement for the existing tenzir/vast image, right?

dit7ya (commented):
Tried testing it with

`➜ nix run github:tenzir/vast/topic/flake-docker-image#stream-image | docker load`

but

```fish

➜ docker run -dt --name=vast --rm tenzir/vast start
f2d5f2f1ece9b652cd6b5c8d52c31319b2dcc764deaec28a6e3fd89525bb548f
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "start": executable file not found in $PATH: unknown.
```

Probably due to https://github.com/NixOS/nixpkgs/issues/99760

flake.nix
```diff
@@ -45,6 +45,14 @@
       };
       apps.vast = flake-utils.lib.mkApp { drv = packages.vast; };
       apps.vast-static = flake-utils.lib.mkApp { drv = packages.vast-static; };
+      apps.stream-image = {
+        type = "app";
+        program = "${pkgs.dockerTools.streamLayeredImage {
+          name = "tenzir/vast";
```
dit7ya, line 51: Maybe we should name it differently to distinguish from the primary Docker image?

flake.nix
```diff
@@ -45,6 +45,14 @@
...
+          tag = "latest";
+          config.Cmd = [ "${self.packages.${system}."vast-static"}/bin/vast" ];
```
dit7ya, line 53:
```suggestion
          config.entryPoint = [ "${self.packages.${system}."vast-static"}/bin/vast" ];
```

dit7ya: This change fixes the error above.

dit7ya (changes requested):

flake.nix
tobim: I'm not sure, I think we should rather use the tag to make the distinction.

tobim:
> This is not a drop-in replacement for the existing tenzir/vast image, right?

With the config changes I just added there are only 2 remaining differences:
1. The container doesn't have any applications besides VAST, so no bash, coreutils, etc. This is not a problem because you should use `nsenter` for inspection / debugging anyways.
2. We use the static binary to build this container, so injecting additional plugins via bind-mount is not possible. (I don't think anybody does this yet).
3. When built remotely aka `nix run github:tenzir/vast#stream-image` Nix will use the archived version of the sources, meaning it will only have the fallback version number. We can add support for our versioning theme with `nix run git+https://github.com/tenzir/vast#stream-image`, but not the former.

flake.nix
dominiklohmann: I think using the same name for an image with a different set of functionality is a really bad UX because of the FROM statement in Dockerfiles. If I accidentally use this in FROM instead of our regular image because this defaulted to the same name and I had it built locally then I'd get some really hard-to-debug problems.

dominiklohmann: What if we named this `tenzir/vast-slim` and then actually uploaded it in the CI as well?

GTrunSec:
if you want to drop the existing tenzir/vast image, I would recommend using the `[nix2container](https://talks.nixcon.org/nixcon-2022/talk/AV3Q8C/)` as an image builder instead.  smaller granularity -> [example](https://github.com/GTrunSec/vast2nix/blob/main/nix/vast/oci-images/mkOCI.nix)？

dit7ya:
I tested `nix2container` for this PR locally, the size was not smaller, but the build time was faster though.

dominiklohmann:
I don't think we want to drop the regular image, as that supports mounting in additional plugins. This one doesn't. I'm still all for calling this vast-slim.

dominiklohmann (approved):
From my point of view this is all good to go. @tobim, please take a look through the changes I've made before merging.

libvast/src/plugin.cpp
```diff
@@ -65,16 +65,19 @@ get_plugin_dirs(const caf::actor_system_config& cfg) {
...
-                      / fmt::format("libvast-plugin-{}.{}", name,
-                                    VAST_MACOS ? "dylib" : "so");
+    auto maybe_path = dir / plugin_file_name;
     auto ec = std::error_code{};
     if (std::filesystem::is_regular_file(maybe_path, ec))
       return maybe_path;
   }
   return caf::make_error(ec::invalid_configuration,
-                         fmt::format("failed to find the {} plugin", name));
+                         fmt::format("failed to find the {} plugin as {} in\n  "
```
dominiklohmann, line 77:
```suggestion
                         fmt::format("failed to find the {} plugin as {} in "
```

nix/vast/default.nix
```diff
@@ -52,11 +51,19 @@ let
   versionFallback = (builtins.fromJSON (builtins.readFile ./../../version.json)).vast-version-fallback;
   version = if (versionOverride != null) then versionOverride' else versionFallback;
   versionShort = if (versionShortOverride != null) then versionShortOverride' else version;
+
+  plugins = [
+    "plugins/parquet"
+    "plugins/pcap"
+    "plugins/sigma"
+    "plugins/web"
```
dominiklohmann, line 59: This is missing CEF

PR #2743, opened 2022-11-28
Skip null indexers when creating status response

lava:
When creating the status response for the active partition, some indexers may be null. These return an "invalid argument" error when attempting to fetch the status from them.

Discovered by @tobim during our recent hackathon.


libvast/src/system/active_partition.cpp
```diff
@@ -652,6 +652,8 @@ active_partition_actor::behavior_type active_partition(
       const auto timeout = defaults::system::status_request_timeout / 5 * 3;
       indexer_states.reserve(self->state.indexers.size());
       for (auto& i : self->state.indexers) {
+        if (!i.second)
+          continue;
```
dominiklohmann, line 656: This should probably get a changelog entry.

dominiklohmann (approved):
I trust that you tested this. The change itself makes a lot of sense.

PR #2744, opened 2022-11-28
Use std::filesystem::exists(path, err) call instead of std::filesystem::exists(path) call

Dakostu:
I noticed that my compiler showed a `unused no-discard return value` warning, The culprit is a`std::filesystem::exists(path)` call whose return value seems to get ignored. I propose replacing it with an `std::filesystem::exists(path, err)` instead.


lava (approved):
D'uh

PR #2745, opened 2022-11-28
Merge the type-registry into the catalog

Dakostu:
This merged the functionality of the type-registry into the catalog, in preparation for a future refactoring where type information is solely maintained based on the available partitions.

PR #2751, opened 2022-11-29
Add a `show schemas` command

dominiklohmann:
This adds a new `vast show schemas` command to show the schemas of all data currently held by VAST. Additionally, this adds an optional positional argument to the `show` command for filtering by module or schema name.


mavam:
@tobim can I ask you to review the code-related pieces of this PR? I'm happy to take the documentation side.

libvast/builtins/commands/show.cpp
```diff
@@ -40,71 +144,69 @@ caf::message show_command(const invocation& inv, caf::actor_system& sys) {
...
-                 {"name", std::move(name)},
-                 {"description", std::move(concept_.description)},
-                 {"fields", std::move(fields)},
-                 {"concepts", std::move(concepts)},
-               }},
-            };
-            result.push_back(std::move(entry));
+  auto command_result = caf::expected<list>{list{}};
+  if (show_concepts || show_models) {
+    self->request(catalog, caf::infinite, atom::get_v, atom::taxonomies_v)
```
tobim, line 149: Please add an appropriate timeout while you're touching this code.

libvast/builtins/commands/show.cpp
```diff
@@ -40,71 +144,69 @@ caf::message show_command(const invocation& inv, caf::actor_system& sys) {
...
+      meta_extractor{meta_extractor::type},
+      relational_operator::not_equal,
+      data{""},
+    }};
+    auto query_context
+      = query_context::make_extract("show", self, std::move(catch_all_query));
+    query_context.id = uuid::random();
+    self
+      ->request(catalog, caf::infinite, atom::candidates_v,
+                std::move(query_context))
```
tobim, line 185: Why exactly do we have to send a query when there is a handler to retrieve all known types?
This puts unnecessary stress on the catalog.

vast/integration/vast_integration_suite.yaml
```diff
@@ -402,6 +405,7 @@ tests:
       - command: show concepts
         transformation: jq '.[] | select(.concept.name == "net.app") | .concept.fields | map(split(".")[0]) | unique'
       - command: show --yaml models
+      - command: show --yaml schemas
```
tobim, line 408: Maybe some test where we actually have data instead?

libvast/src/type.cpp
```diff
@@ -789,6 +789,83 @@ data type::construct() const noexcept {
...
+    return record{
+      {std::string{name}, std::move(definition)},
+    };
+  }
+  // At this point we've gone through all named aliases, but the last innermost
+  // type may still have attributes.
+  VAST_ASSERT(name().empty());
+  auto make_type_definition = [&]() noexcept -> data {
+    if (!*this)
+      return {};
```
tobim, line 827: Afaics this path is untested.

vast/integration/reference/server-zeek-multiple-imports/step_02.ref
```diff
@@ -0,0 +1,43 @@
...
+        "ts": {
+          "timestamp": "time"
+        },
+        "uid": {
+          "type": "string",
+          "attributes": [
+            {
+              "index": "hash"
+            }
+          ]
```
tobim, line 14: Why is attributes a list of objects instead of just an object? Afaik this is inconsistent with the modules proposal.

libvast/src/type.cpp
```diff
@@ -789,6 +789,83 @@ data type::construct() const noexcept {
...
+  // Check if there is an alias, and if there is then visit then one first.
+  for (const auto& alias : aliases()) {
+    auto definition = attributes_enriched_definition(alias.to_definition());
+    const auto name = this->name();
+    if (name.empty())
+      return definition;
+    return record{
+      {std::string{name}, std::move(definition)},
+    };
+  }
```
tobim, line 821: Doing this inline means that all use sites of aliases will be replaced by their definition. Wouldn't it be a bit nicer to do this only for unnamed aliases and collecting the named ones in a set to print there definitions separately?

tobim (approved):
Good feature, but a few high-level comments that I'd like to see addressed.

vast/integration/reference/server-zeek-multiple-imports/step_02.ref
dominiklohmann: It is consistent with the proposal, but I agree that a record makes a lot more sense here.

libvast/builtins/commands/show.cpp
dominiklohmann: This handler with this query is really quite cheap. The ask was to print all schemas for which we have data, and there is no other handler that can get me this information.

vast/integration/vast_integration_suite.yaml
dominiklohmann: I'll move the `--yaml` over to one of the other two uses.

libvast/src/type.cpp
dominiklohmann: That could be a future extension, but I'd much rather like to see that as an operation on multiple type definitions to form a whole module.

libvast/src/type.cpp
dominiklohmann: I added a test.

dominiklohmann: We just decided in a call a few moments later that we will just add an assertion to `to_definition` that the type is in a valid state.

libvast/builtins/commands/show.cpp
dominiklohmann: I re-used the existing option we have for this, the node connection timeout.

dominiklohmann:
Note that the CI fails in this not because of my changes but rather because I based this on top of #2745, which still contains a bug in a rewritten unit test. I did this to avoid the merge conflict.

libvast/builtins/commands/show.cpp
tobim: I see that we don't make a full pass for negations, but this is not one. You should be able to turn the O(n) into O(1) by using `negation{expression{}}` instead.

dominiklohmann: That's a neat idea. Never thought of that one before.

vast/integration/reference/server-zeek-multiple-imports/step_02.ref
dominiklohmann: We talked about this and agreed to make the change.

dominiklohmann: Done.

mavam:
@dominiklohmann missing baseline update?

```diff
+++ vast-integration-test/server-zeek-multiple-imports/step_02.out
@@ -7,11 +7,9 @@
         },
         "uid": {
           "type": "string",
-          "attributes": [
-            {
-              "index": "hash"
-            }
-          ]
+          "attributes": {
+            "index": "hash"
+          }
         },
         "id.orig_h": "addr",
         "id.orig_p": {
```

dominiklohmann:
> @dominiklohmann missing baseline update?

Yes after the most recent change for attributes. I pushed another commit to fix this.

libvast/builtins/commands/show.cpp
```diff
@@ -20,8 +22,111 @@ namespace vast::plugins::show {
 
 namespace {
 
+caf::error print_definition(const data& definition, bool as_yaml) {
```
lava, line 25: This should probably be exposed publicly, with some API changes (ie. return a string and make the second parameter a `enum format {json, yaml}`.

Or alternatively maybe we can just add `{:y}` and `{:j}` format modifiers to the printer for `data`?

dominiklohmann: I'm not sure how that'd work given that formatters cannot error and that both `to_yaml` and `to_json` return a `caf::expected<std::string>`.

lava: Do we expect instances of `data` that cannot be rendered as json or yaml?

web/docs/use/introspect/README.md
```diff
@@ -1,5 +1,233 @@
...
+            pkts_toserver: count
+            pkts_toclient: count
+            bytes_toserver: count
+            bytes_toclient: count
+            start: time
+            end: time
+            age: count
+            state: string
+            reason: string
+            alerted: bool
```
mavam, line 218: @dominiklohmann Looking at the output, I realized that this doesn't conform to the spec: record fields are ordered, so the `record` object must map to an array. From our [YAML Types](https://docs.google.com/document/d/1thYBUQ2mbobUyLbLF8Rc0LkexKK7UrnVuACiuCoYXXU/edit):

```yaml
connection:
  record:
  - src_ip: addr
  - dst_ip: addr
  - src_port: port
  - dst_port: port
  - proto: string
```

dominiklohmann: I fixed this in the most recent commit. I did not touch your docs, but rather only the code and the test baselines.

mavam: I'll adapt the docs. Now that I can just get the Docker container from GHCR, updating will be a breeze.

libvast/builtins/commands/show.cpp
tobim: Why do those function return an `expected` though? `vast::data` to JSON/YAML should be total and never fail.

mavam: In theory, yes. But in practice? What about maximum string lengths? Special chars of an implementation that cause weird corner cases? Would you put your hand down for the theory?

dominiklohmann: The "corner case" I'm worried about primarily is a stack overflow. At least the YAML printer returns an error instead of printing if the input is nested too deeply.

In any case, I don't think this is relevant for this PR at all. I'm writing a follow-up story so we don't forget about this, but am resolving here.

PR #2752, opened 2022-11-29
Prepare repository for VAST v2.4-rc3

lava:


version.json
```diff
@@ -4,7 +4,7 @@
     "annotated git tag without the leading 'v'.",
     "This value gets updated automatically by `scripts/prepare-release`."
   ],
-  "vast-version-fallback": "2.4.0-rc2",
+  "vast-version-fallback": "2.4-rc3",
```
dominiklohmann, line 7:
```suggestion
  "vast-version-fallback": "2.4.0-rc3",
```

I think you need to re-run the script with the missing patch version, not sure in what ways CMake will fail otherwise.

lava: Right, I've opened a fixed PR

PR #2753, opened 2022-11-29
Refactor compose files into separate directory

rdettai:
We are currently having problems with relative paths in Compose because of [the way the base path is defined with overlays](https://docs.docker.com/compose/reference/#specifying-multiple-compose-files). This PR colocates every Compose file and their configuration examples into a single directory to solve this.

It also cleans up the MISP integration by removing the WIP app. We keep it in a separate draft branch for now (https://github.com/tenzir/vast/pull/2727)


docker/compose/README.md
```diff
@@ -0,0 +1,35 @@
...
+    contains the configurations to interact with the `vast` service (e.g
+    networking)
+  - otherwise a keyword indicating what is being enabled in the `integration` on
+    top of the base configuration (e.g `vast.volume.yaml` configures volumes for
+    the `vast` service to provide persistence)
+
+## The environment files
+
+If a `.env` file is present at the base of the project directory (by default
+this directory), Docker Compose automatically loads environment from it. For
```
dominiklohmann, line 23:
```suggestion
this directory), Docker Compose automatically loads environment variables from it. For
```

docker/compose/README.md
```diff
@@ -0,0 +1,35 @@
...
+each integration, we provide a `{integration}-env.example` file with the
+supported configurations (values are defaulted in the Docker Compose files as
```
dominiklohmann, line 25:
```suggestion
supported configuration options (values are defaulted in the Docker Compose files as
```

docker/compose/README.md
```diff
@@ -0,0 +1,35 @@
...
+well). You can optionnally create a `.env` file in the project directory and
+cherry pick the variables you want to customize from the example files.
```
dominiklohmann, line 27:
```suggestion
cherry-pick the variables you want to customize from the example files.
```

dominiklohmann: This last sentence is passive voice, might want to rephrase in terms of the action the user wants to take:

> To customize from the example files, create a `.env` file in the project directory and cherry-pick environment variables as needed.

docker/compose/README.md
```diff
@@ -0,0 +1,35 @@
...
+    top of the base configuration (e.g `vast.volume.yaml` configures volumes for
+    the `vast` service to provide persistence)
+
+## The environment files
+
+If a `.env` file is present at the base of the project directory (by default
+this directory), Docker Compose automatically loads environment from it. For
+each integration, we provide a `{integration}-env.example` file with the
+supported configurations (values are defaulted in the Docker Compose files as
+well). You can optionnally create a `.env` file in the project directory and
```
dominiklohmann, line 26:
```suggestion
well). You can optionally create a `.env` file in the project directory and
```

dominiklohmann (approved):
This is great, just some minor requests for the documentation. This solves so many of the weird issues.

web/docs/setup/deploy/docker-compose.md
```diff
@@ -60,22 +65,24 @@ Docker Compose:
 # Load both VAST and Zeek, and the import that sits between the two.
```
dominiklohmann, line 65: Unrelated to this line, but if you look at https://vast.io/docs/setup/deploy/docker-compose then you can see that the :::info block above is not properly closed out and extends into this text. Can you fix this as well?
Unrelated to this line, but if you look at https://vast.io/docs/setup/deploy/docker-compose then you can see that the :::info block above is not properly closed out and extends into this text. Can you fix this as well?

image

docker/compose/README.md
rdettai: I feel it doesn't convey the fact that it's optional very well, does it?

web/docs/setup/deploy/docker-compose.md
rdettai: On it. I am also changing the sentence 
```
...try using `docker-compose` instead—although we don't test it explicitly, most commands
should work that way as well.
```
to 
```
...try using `docker-compose` instead. We don't test our script explicitly with this older
version, so although most commands should work, we recommand upgrading to a
recent version of Docker and Compose.
```

docker/compose/README.md
rdettai: I propose

> Optionally, create a `.env` file in the project directory and cherry-pick the variables you want to customize from the example files.

Good for you?

PR #2762, opened 2022-12-02
Remove dscat

patszt:
Dscat requires some code alignment with CAF 0.18.6 bump. It was decided that it is no longer used and can be removed


scripts/dscat.py
```diff
@@ -1,110 +0,0 @@
-#!/usr/bin/env python
```
tobim, line 1: This is technically a  separate implementation in python 2. You wouldn't have to remove it for the CAF upgrade, but I'll accept it anyways because python 2 is EOLd and i don't think I've ever seen anybody use this.

tobim (approved):
Out with the unsupported cruft :hankey:

PR #2763, opened 2022-12-05
Fix transient errors with TheHive

rdettai:
TheHive tests transiently fail (e.g in https://github.com/tenzir/vast/actions/runs/3619876264/jobs/6101834415). This PR tries to identify the reason and fix it.

### DIFF
--- a/docker/compose/thehive.app.yaml
+++ b/docker/compose/thehive.app.yaml
@@ -20,6 +20,7 @@ services:
       - thehive
     networks:
       - vast
+      - thehive
   thehive:
     networks:
       - vast
--- a/docker/thehive/test
+++ b/docker/thehive/test
@@ -28,10 +28,10 @@ docker compose run --no-TTY vast import --blocking suricata \
 
 # add the app to the stack
 COMPOSE_FILE="$COMPOSE_FILE:compose/thehive.app.yaml"
-docker compose up --build -d thehive-app
+docker compose up --build --no-recreate -d thehive-app
 
 # check that the historical import took place
-docker compose run thehive-app count-alerts 2>&1 | \
+docker compose run --no-deps thehive-app count-alerts 2>&1 | \
     tee /dev/stderr | \
     grep -q "alert_count=1"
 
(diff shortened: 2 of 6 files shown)


docker/thehive/test
```diff
@@ -28,10 +28,10 @@ docker compose run --no-TTY vast import --blocking suricata \
 
 # add the app to the stack
 COMPOSE_FILE="$COMPOSE_FILE:compose/thehive.app.yaml"
-docker compose up --build -d thehive-app
+docker compose up --build --no-recreate -d thehive-app
 
 # check that the historical import took place
-docker compose run thehive-app count-alerts 2>&1 | \
+docker compose run --no-deps thehive-app count-alerts 2>&1 | \
```
rdettai, line 34: The key here are these `--no-recreate` and `--no-deps` which avoid that TheHive init gets interrupted in the middle, leaving it in an undefined state

docker/compose/thehive.app.yaml
```diff
@@ -20,6 +20,7 @@ services:
       - thehive
     networks:
       - vast
+      - thehive
```
rdettai, line 23: when `--no-deps` is used, the network needs to be explicitly set (otherwise  `depends_on` is enough to setup the network)

docker/thehive/thehive/entrypoint.py
```diff
@@ -36,6 +40,29 @@
...
 
 
+def is_initialized() -> bool:
+    if not path.isfile(CHECKPOINT_LOCATION):
+        logging.info("TheHive and Cortex not initialized yet")
+        return False
+    with open(CHECKPOINT_LOCATION, "r") as checkpoint_file:
+        status = checkpoint_file.read()
+        if status != "COMPLETED":
+            raise Exception(INCONSITENT_STATE_ERR)
```
rdettai, line 50: we also make the entrypoint script easier to debug by explicitly failing if the init is only partially completed

dispanser (approved):
Thanks, it _seems_ reasonable. I have no good way to test this, so I trust you on these changes.

PR #2765, opened 2022-12-05
Short-circuit earlier when filtering batches

dominiklohmann:
The short-circuiting in the function evaluating a filter expression on a batch of data short-circuited later than it could've, which made queries that return all or no events unnecessarily expensive. This moves the checks to their proper places.

Noticed by @tobim when benchmarking a bulk-export for VAST v2.4.

### DIFF
--- a/libvast/src/table_slice.cpp
+++ b/libvast/src/table_slice.cpp
@@ -740,24 +740,25 @@ filter(const table_slice& slice, expression expr, const ids& hints) {
   if (!hints.empty())
     selection &= hints;
   // Do no rows qualify?
-  auto selection_rank = rank(selection);
-  if (selection_rank == 0)
+  if (!any(selection))
     return std::nullopt;
-  const auto has_expr = expr != expression{};
-  if (!has_expr) {
-    // Do all rows qualify?
-    if (rank(slice_ids) == selection_rank)
-      return slice;
-  } else {
+  // Evaluate the filter expression.
+  if (!caf::holds_alternative<caf::none_t>(expr)) {
     // Tailor the expression to the type; this is required for using the
     // evaluate function, which expects field and type extractors to be resolved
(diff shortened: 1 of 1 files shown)


libvast/src/table_slice.cpp
```diff
@@ -792,10 +790,8 @@ filter(const table_slice& slice, expression expr, const ids& hints) {
       VAST_ASSERT(ret);
     }
   }
-  if (builder->rows() == 0)
-    return std::nullopt;
-  if (builder->rows() == slice.rows())
-    return slice;
+  VAST_ASSERT(builder->rows() != 0);
+  VAST_ASSERT(builder->rows() != slice.rows());
```
tobim, line 794:
```suggestion
  VAST_ASSERT_CHEAP(builder->rows() != 0);
  VAST_ASSERT_CHEAP(builder->rows() != slice.rows());
```

libvast/src/table_slice.cpp
```diff
@@ -740,24 +740,25 @@ filter(const table_slice& slice, expression expr, const ids& hints) {
   if (!hints.empty())
     selection &= hints;
   // Do no rows qualify?
-  auto selection_rank = rank(selection);
-  if (selection_rank == 0)
+  if (!any(selection))
```
tobim, line 743: Both `rank` and `any` iterate over the bit ranges. If you switch back to `rank` here and at line 755 you can save the pass at line 759.

dominiklohmann: I cannot, as `selection` may be updated by the call to `evaluate` in L753.

Doesn't `any` return as soon as it finds the first one bit, rather than iterating over the entire range?

tobim: Ah, yes, of course it does.

PR #2766, opened 2022-12-06
Eliminate shutdown lag from the signal monitor

tobim:
The simple `sleep_for` used to introduce a shutdown delay of up to 750ms to every client command invocation. This replaces the naive sleep with wakeup on event logic, which is reused to signal the monitoring thread to stop.

Instructions for the reviewer:
- [ ] Run `count`/`export`/etc. on an empty database and see for yourself that the command returns in less than 750ms.
- [ ] Verify that regular interrupts still work with commands that run a little longer.

### DIFF
--- a/libvast/src/system/signal_monitor.cpp
+++ b/libvast/src/system/signal_monitor.cpp
@@ -21,7 +21,6 @@
 #include <cstdlib>
 #include <cstring>
 #include <iostream>
-#include <thread>
 
 using namespace caf;
 
@@ -30,6 +29,7 @@ namespace {
 // Keeps track of all signals by their value from 1 to 31. The flag at index 0
 // is used to tell whether a signal has been raised or not.
 std::atomic<bool> signals[32];
+std::condition_variable* cvp;
 
 extern "C" void signal_monitor_handler(int sig) {
   // Catch termination signals only once to allow forced termination by the OS
@@ -41,6 +41,7 @@ extern "C" void signal_monitor_handler(int sig) {
   }
   signals[0] = true;
   signals[sig] = true;
+  cvp->notify_one();
 }
 
 } // namespace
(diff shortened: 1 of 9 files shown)


libvast/src/system/signal_monitor.cpp
```diff
@@ -41,24 +41,30 @@ extern "C" void signal_monitor_handler(int sig) {
   }
   signals[0] = true;
   signals[sig] = true;
+  cvp->notify_one();
```
mavam, line 44: The standard says:

> The behavior of any function other than a [plain old function] used as a signal handler in a C++ program is implementation-defined.

The function `notify_one` is not from a "common subset", so the code is not well defined.

dominiklohmann (approved):
This makes the integration tests run around 90 seconds faster on my machine in a release build. Wow. Crazy to think we had not noticed this since Feb 2019 where this was introduced in 92fe59046ce1b11e63575ed8a06a6e70107e1141.

libvast/src/system/signal_monitor.cpp
dominiklohmann: TIL! I dismissed my approval.

I found this great blog post on an approach that is not UB from what I can tell. This works by blocking signals in all threads except for the signal monitor one, and then having the signal monitor thread use `sigwait` rather than a condition variable.

https://thomastrapp.com/posts/signal-handlers-for-multithreaded-c++/#no-need-to-call-me-ill-sigwait-for-you

tobim: I half expected that there was a catch. Well at least @dominiklohmann already found a blueprint for a more sound approach.

tobim:
As explained in https://github.com/tenzir/vast/pull/2766#discussion_r1041864630 this is technically not allowed.

PR #2768, opened 2022-12-07
Infer non-config types in extend and replace operators

dominiklohmann:
Previously, using an address value in the extend or replace operators did not work. This is now fixed, and will produce the proper type as long as it can be inferred from the data. This bug applied to all types that were not also used as types in configuration, i.e., address, subnet, pattern, and map.

### DIFF
--- a/changelog/unreleased/bug-fixes/2768--infer-non-config-types.md
+++ b/changelog/unreleased/bug-fixes/2768--infer-non-config-types.md
@@ -1,0 +1,3 @@
+The `replace` and `extend` pipeline operators wrongly inferred address, subnet,
+pattern, and map values as strings. They are now inferred correctly. To force a
+value to be inferred as a string, wrap it inside double quotes.
--- a/libvast/builtins/operators/replace.cpp
+++ b/libvast/builtins/operators/replace.cpp
@@ -8,6 +8,8 @@
 
 #include <vast/arrow_table_slice.hpp>
 #include <vast/arrow_table_slice_builder.hpp>
+#include <vast/concept/parseable/to.hpp>
+#include <vast/concept/parseable/vast/data.hpp>
 #include <vast/detail/narrow.hpp>
 #include <vast/error.hpp>
 #include <vast/pipeline.hpp>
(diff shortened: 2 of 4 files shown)


libvast/builtins/operators/replace.cpp
```diff
@@ -51,10 +53,24 @@ struct bound_configuration {
...
+      // that we need to re-parse the data if it's a string.
+      auto reparsed_value = [](auto value) {
+        const auto* str = caf::get_if<std::string>(&value);
+        if (!str)
+          return value;
+        auto result = to<data>(*str);
+        if (!result)
+          return value;
+        return std::move(*result);
+      }(value);
```
dominiklohmann, line 68: Should this be a friend function of `data` more generally, maybe along the lines of `parse_if_string`?

mavam: Why friend function and not just a normal free function? I would argue that the interface of `data` should be minimal and that parsing/printing is an orthogonal concern.

dominiklohmann: Two, yes:
- Make it easy for the function to be discovered (either by looking at the header or at tooling), and
- Minimally improve compile times.

It's just generally a good practice. Free functions are less discoverable, so they tend to be overlooked if you don't know they exist. As the team grows we need to make sure that we have functions in sane places, as we cannot expect everyone to always read through all changes.

I know there are two camps for this discussion and that you specifically are very much in the other one here. :)

dominiklohmann: Ultimately, I'd also be happy with just leaving the lambda here as-is for now.

changelog/unreleased/bug-fixes/2768--infer-non-config-types.md
```diff
@@ -0,0 +1,3 @@
+The `replace` and `extend` pipeline operators wrongly inferred address, subnet,
+pattern, and map values as strings. They are now inferred correctly. To force a
+value to be inferred as a string, wrap it inside double quotes.
```
dispanser, line 3: I tested with my recent react-eu example, and indeed now the replace value is of type address. However, "wrapping it in double quote" doesn't change that, so it's not showing up as string in my tests. Pipeline snippet:

```
      - replace:
          fields:
            src_ip: "1.2.3.4"
```

dispanser (approved):
Works as advertised, except for encorcing the inferred type to be a string via wrapping in doublequotes does not seem to work as described.

changelog/unreleased/bug-fixes/2768--infer-non-config-types.md
dominiklohmann: You will have to escape those quotes, the one's you added are eaten by the YAML parser.

dispanser: that indeed works.

libvast/builtins/operators/replace.cpp
dominiklohmann: Given that we need to (slightly) rewrite these two operators anyways with the executor refactoring, we can use that opportunity for cleanup. I'll just rebase this now and leave the lambdas as they are.

PR #2770, opened 2022-12-08
Reduce regex compilations to speed up pattern search

dominiklohmann:
This change makes it so we only instantiate regex objects once per batch instead of once per row, causing a massive speedup even for simple pattern searches.

No changelog entry as this wasn't possible at all prior to #2769, which this is based upon.


mavam (approved):
Nice and clean. Does this require extra tests?

dominiklohmann:
> Does this require extra tests?

No, the functionality is already covered by tests. It just takes a different path now.

PR #2771, opened 2022-12-08
Move taxonomy resolution to the catalog

Dakostu:
This removes the eager taxonomy resolution we've previously done in the spawn command and moves it directly inside the catalog, which now does the lookup per set of partitions that have the same schema. To enable this, the catalog now stores its partitions sorted by schema.
In addition, this PR disallows using `vast -N status`, which as a command does not make a lot of sense and became racy with this change.


Dakostu:
Converting to draft until I get a proper rebase going...

libvast/src/system/node.cpp
```diff
@@ -283,6 +282,13 @@ spawn_accountant(node_actor::stateful_pointer<node_state> self) {
 } // namespace
 
 caf::message status_command(const invocation& inv, caf::actor_system&) {
+  if (caf::get_or(inv.options, "vast.node", false)) {
```
Dakostu, line 285: The reason `status` for spawned nodes has been disabled is that the new [catalog loading handler call in the index](https://github.com/tenzir/vast/pull/2771/files#diff-d59f67c3b5fcc2153473271c6457c878430b53982892f4988d4116e6c19a0f8fR1266) would cause a race condition -> timeout in spawned nodes - and also I've been told that spawned nodes being able to launch `status` would not make sense semantically

libvast/include/vast/system/actors.hpp
```diff
@@ -298,7 +297,7 @@ using index_actor = typed_actor_fwd<
   // erased, returns the nil uuid. When keep_original_partition is no: does an
   // in-place pipeline keeping the old ids, and makes a new partition
   // preserving the old one(s).
-  caf::replies_to<atom::apply, pipeline_ptr, std::vector<uuid>,
+  caf::replies_to<atom::apply, pipeline_ptr, std::unordered_map<uuid, type>,
```
tobim, line 300: It would be great if you could update the description, it looks like it has gone out of sync with earlier changes already.
It is not clear why `unordered_map<uuid, type>` is used here. Are we actually accessing specific values by id or would a `vector<pair<uuid, type>>` be more appropriate?

libvast/include/vast/system/actors.hpp
```diff
@@ -228,22 +227,21 @@ using catalog_actor = typed_actor_fwd<
...
-  caf::reacts_to<atom::put, taxonomies>,
   // Registers a given layout.
   caf::reacts_to<atom::put, vast::type>,
   // Retrieves the known taxonomies.
   caf::replies_to<atom::get, atom::taxonomies>::with< //
     taxonomies>,
   // Loads the taxonomies on disk.
   caf::replies_to<atom::load>::with< //
-    atom::ok>,
+    taxonomies>,
```
tobim, line 244: I'm not sure about this. At the minimum it needs to be a 2 argument handler that takes `atom::taxonomies` as well.

tobim (changes requested):
I only reviewed a small part so far, but I have to leave now and don't want to risk loosing it.
I think it would be great to have a high level summary of these changes so I identify the individual pieces.

libvast/include/vast/system/actors.hpp
Dakostu: I updated the description in https://github.com/tenzir/vast/pull/2771/commits/564ee29d906a6c51ab0000db4625809b3edc7f2f

Your question made me realize that `std::vector<partition_info>` already does the trick. I removed those `std::unordered_map`s in https://github.com/tenzir/vast/pull/2771/commits/2bc28382f19ba1a396dcb7463c2e41bf370264d7

libvast/src/system/index.cpp
```diff
@@ -1764,32 +1812,35 @@ index(index_actor::stateful_pointer<index_state> self,
       return rp;
     },
     [self](atom::apply, pipeline_ptr pipeline,
-           std::vector<vast::uuid> old_partition_ids,
+           std::unordered_map<uuid, type> old_partitions_per_type,
```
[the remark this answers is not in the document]
Dakostu: Renamed variables in https://github.com/tenzir/vast/pull/2771/commits/b5f80154e806f6da761935d304ab58f16d86f2e4

libvast/include/vast/system/actors.hpp
```diff
@@ -282,10 +280,11 @@ using index_actor = typed_actor_fwd<
                  partition_creation_listener_actor, send_initial_dbstate>,
   // Evaluates a query, ie. sends matching events to the caller.
   caf::replies_to<atom::evaluate, query_context>::with<query_cursor>,
-  // Resolves a query to its candidate partitions.
+  // Resolves a query to its candidate partitions per type.
   // TODO: Expose the catalog as a system component so this
   // handler can go directly to the catalog.
-  caf::replies_to<atom::resolve, expression>::with<catalog_result>,
+  caf::replies_to<atom::resolve,
+                  expression>::with<std::unordered_map<type, catalog_result>>,
```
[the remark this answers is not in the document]
Dakostu: Renamed `catalog_result` in https://github.com/tenzir/vast/pull/2771/commits/b87bf1e67a95ea56ec93556b6dd6078d94fbf4e1

libvast/include/vast/system/actors.hpp
Dakostu: Why exactly? Is the `atom::taxonomies` input then supposed to be an output parameter?

Dakostu:
@tobim 
> I think it would be great to have a high level summary of these changes so I identify the individual pieces.

There is an added summary to the top post of this PR. 
Also maybe some additional information:

| Component  | Before this PR  | After this PR |
| ------------- | ------------- |  ------------- |
| `node`  | Calls the `catalog` to handle taxonomies and call `spawn_actually`  | Only calls `spawn_component` to only spawn and register a component without taxonomy resolution |
| `catalog`  | Stores partition `uuid`s for `lookup`s  and `taxonomies` from the former type-registry | Stores partitions separated by `schema` and passes `taxonomies` to the `index` |
| `index` | Stores partition `uuid`s | Stores partitions separated by type and caches `taxonomies` from the `catalog` to pass on to newly spawned `active_partition`s for `resolve`ing |
| `active_partition` | Does not store `taxonomies` | Stores `taxonomies` and handles them when calling `resolve()`  in the `atom::query` handler |

libvast/include/vast/system/actors.hpp
tobim: So it looks like you actually want to call the `get, taxonomies` handler, maybe the side-effect is not necessary.

libvast/include/vast/system/index.hpp
```diff
@@ -255,15 +255,15 @@ struct index_state {
...
+  std::unordered_map<uuid, std::pair<type, partition_actor>> unpersisted = {};
 
   /// The set of passive (read-only) partitions currently loaded into memory.
   /// Uses the `partition_factory` to load new partitions as needed, and evicts
   /// old entries when the size exceeds `max_inmem_partitions`.
   detail::lru_cache<uuid, partition_actor, partition_factory> inmem_partitions;
 
   /// The set of partitions that exist on disk.
-  std::unordered_set<uuid> persisted_partitions = {};
+  std::unordered_map<uuid, type> persisted_partitions = {};
```
tobim, line 266: We paired to implement a different approach that doesn't need to add more use to this map. The index shouldn't need to know about the full set of partitions in the database in the long run.

libvast/src/system/node.cpp
tobim: I don't understand what race to be honest. The status output of the catalog may be empty because the partitions haven't been merged yet, but what does it have to do with loading the taxonomies?

libvast/src/system/index.cpp
```diff
@@ -1242,6 +1262,18 @@ index(index_actor::stateful_pointer<index_state> self,
   self->state.accountant = std::move(accountant);
   self->state.filesystem = std::move(filesystem);
   self->state.catalog = std::move(catalog);
+  self->request(self->state.catalog, caf::infinite, atom::load_v)
+    .await(
```
tobim, line 1266: You could (probably should) use the `get_v, taxonomies_v` handler here and reinstate the loading in `spawn_catalog`.

libvast/src/system/spawn_catalog.cpp
```diff
@@ -27,14 +27,6 @@ spawn_catalog(node_actor::stateful_pointer<node_state> self,
...
         ? self->spawn<caf::detached>(catalog, accountant, args.dir / args.label)
         : self->spawn(catalog, accountant, args.dir / args.label);
-  self->request(handle, caf::infinite, atom::load_v)
-    .await([](atom::ok) {},
-           [](caf::error& err) {
-             VAST_WARN("catalog failed to load taxonomy "
-                       "definitions: {}",
-                       std::move(err));
-             // TODO: Shutdown when failing?
-           });
```
tobim, line 37: I think it is better to keep this here.

libvast/src/system/node.cpp
dominiklohmann: 1. The index starts up, loads partitions
2. Query arrives at the index, is deferred until partitions are available
3. Index is done loading partitions, sends them to catalog
4. Catalog loads taxonomies
5. Query is at the catalog, resolves taxonomies
6. Resolves query and catalog lookup result go back to index

The race is between (4) and (5), which can happen in any order. This was an issue with the type-registry as well, but that just happened to load a lot faster.

Once we remove the type-registry state we shouldn't have this race anymore.

Regardless, `vast -N status` doesn't make much sense to support, which is why I suggested disabling it so we can move forward faster with this PR.

libvast/include/vast/system/actors.hpp
dominiklohmann: Considering this `load` handler will be superseded soon when we get rid of the type-registry state file and just use the available partition synopses instead I don't think it makes sense to invest a lot of time here.

libvast/src/system/spawn_catalog.cpp
dominiklohmann: Why?

tobim: Because why should the index tell the catalog to load taxonomies from disk? Either it should be done automatically or here.

libvast/include/vast/system/index.hpp
Dakostu: I applied the finishing touches to this in https://github.com/tenzir/vast/pull/2771/commits/949861c2451c45c50c33bb68018a7922bdf286f5

libvast/src/system/spawn_catalog.cpp
dominiklohmann: IMO: Yes, as it doesn't make sense to use.

tobim: The loading logic was moved to the actor spawning function now. This is the right place.

libvast/test/system/catalog.cpp
```diff
@@ -44,7 +44,7 @@ namespace {
 
 constexpr size_t num_partitions = 4;
 constexpr size_t num_events_per_parttion = 25;
-constexpr uint32_t taste_count = 4;
+constexpr size_t taste_count = 4;
 constexpr size_t num_query_supervisors = 1;
```
tobim, line 49: These 2 variables should not be necessary any more.

Dakostu: Removed them in https://github.com/tenzir/vast/pull/2771/commits/d056ae26271d60fcc387da2b2b8fb661470d716d

tobim (approved):
This looks right to me now.

libvast/include/vast/system/actors.hpp
Dakostu: The `atom::load` handler has been replaced by a method.

libvast/src/system/node.cpp
Dakostu: `vast -N status` will remain forbidden.

PR #2772, opened 2022-12-08
Fix broken promises in index.cpp

patszt:
One of the promise is never delivered when persist request to partition transformer return an error.

The other is a race condition that got visible during CAF 0.18.6 bump.
The only thing keeping partition_transfomer alive in index is the index's request to partition_transfomer and the pending queries.
It can happen that partition_transfomer lifetime is over before it can fulfill it's own internal promise that is needed to fulfill index promise. This leads to the broken promise error.

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -1861,10 +1861,14 @@ index(index_actor::stateful_pointer<index_state> self,
           };
       // TODO: Implement some kind of monadic composition instead of these
       // nested requests.
+      // TODO: With CAF 0.19 it will no longer be needed to keep
+      // partition_transformer alive in the lambda as the promise kept in the
+      // state will keep the actor alive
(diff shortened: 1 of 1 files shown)


libvast/src/system/index.cpp
```diff
@@ -1968,10 +1972,11 @@ index(index_actor::stateful_pointer<index_state> self,
                               });
                         }
                       },
-                      [self](const caf::error& e) {
+                      [self, rp](caf::error e) mutable {
```
dominiklohmann, line 1975:
```suggestion
                      [self, rp](caf::error& e) mutable {
```

This actually took the lvalue-ref intentionally, see https://github.com/actor-framework/actor-framework/discussions/1220 for why.

dominiklohmann (approved):
Thanks for backporting this!

libvast/src/system/index.cpp
patszt: Thanks for the clarification

PR #2776, opened 2022-12-09
Remove the `#field` meta extractor

dominiklohmann:
The predicate `#field == "X"` matched all events for which the field extractor `X` had data, i.e., it was semantically equivalent to `X != nil`.

Fundamentally, the `#field` meta extractor broke our model of extractors, since it did not result in a single value or a list of values to compare against, but rather created an implicit disjunction with changed relational operator semantics compared to all other predicates.

This removes the operator without a deprecation period as it received very little usage.

### DIFF
--- a/web/docs/understand/query-language/expressions.md
+++ b/web/docs/understand/query-language/expressions.md
@@ -146,9 +146,8 @@ types because an alias is a strict refinement of an existing type.
 
 #### Meta Extractor
 
-Meta extractors have the forms `#E` where `E` is from the fixed set of tokens
-`type`,  and `import_time`. They work on the event metadata (e.g., their
-schema) instead of the value domain.
+Meta extractors have the form `#extractor`. They work on the event metadata
+(e.g., their schema) instead of the value domain.
 
 - `#type`: on the event name in a schema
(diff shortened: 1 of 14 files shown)


dominiklohmann:
[sc-40052]

web/docs/understand/query-language/expressions.md
```diff
@@ -149,18 +149,16 @@ types because an alias is a strict refinement of an existing type.
 #### Meta Extractor
 
 Meta extractors have the forms `#E` where `E` is from the fixed set of tokens
-`type`, `field`, and `import_time`. They work on the event metadata (e.g., their
+`type`,  and `import_time`. They work on the event metadata (e.g., their
```
mavam, line 152:
```suggestion
type and import timestamp. They work on the event metadata (e.g., their
```

mavam (approved):
Looks fully covered.

web/docs/understand/query-language/expressions.md
dominiklohmann: That doesn't make sense in the sentence. I' rather shorten it to this:

> Meta extractors have the form `#extractor`.

And remove the whole second half of the sentence. It doesn't add anything with the list below.

PR #2779, opened 2022-12-09
Use a faster runner when the 'fast-ci' label is specified

lava:

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -592,7 +592,7 @@ jobs:
     strategy:
       matrix:
         vast:
-          - os: ${{ contains(github.event.pull_request.labels.*.name, 'fast-ci') && 'ubuntu-20.04' || 'ubuntu-20.04-64core' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/vast.yaml
```diff
@@ -592,22 +592,26 @@ jobs:
     strategy:
       matrix:
         vast:
-          - os: ubuntu-20.04
+          - os: ${{ contains(github.event.pull_request.labels.*.name, 'fast-ci') && 'ubuntu-20.04' || 'fast-ci' }}
```
dominiklohmann, line 595:
```suggestion
          - os: ${{ contains(github.event.pull_request.labels.*.name, 'fast-ci') && 'fast-ci' || 'ubuntu-20.04' }}
```

dominiklohmann:
Looking at the runtime it wasn't actually all that much faster when there's a cache hit, so I'm not sure this is worth it @lava.

dominiklohmann:
Closing as this was not as effective as we had hoped.

PR #2782, opened 2022-12-10
Migrate vite to version 4

dit7ya:
Vite new major version 4 was announced yesterday, it uses Rollup 3 now - https://vitejs.dev/blog/announcing-vite4.html. 

Also migrating early is better as the code base itself is smaller.
If there is no issues with CI building, I think this is good to go.


dominiklohmann (approved):
Looks good to me. We can update the SBOM files with the next release, I don't think it makes sense to always keep them up-to-date on master. That's just too much work.

PR #2786, opened 2022-12-12
Upgrade GCP setup action

rdettai:
Using `google-github-actions/setup-gcloud` with `service_account_key` is deprecated, replacing it with `google-github-actions/auth@v1` and Workload Identity


dominiklohmann (approved):
Thanks for digging through the documentation here!

rdettai:
@dominiklohmann the CI skipped all the parts where GCP could have executed. Let's merge this anyway and we can revert if something goes wrong

PR #2795, opened 2022-12-14
Add suricata.quic type

dispanser:
This adds schema information and some test data for `suricata.quic`, which is part of the next suricata version 7.

The associated pieces of the schema in suricata itself, this may be useful to verify against the VAST schema definition in case I missed something - we have no actual samples that contain the `cyu` and `extensions` fields, so there's no test covering those.

```json
        "quic": {
            "type": "object",
            "optional": true,
            "properties": {
                "cyu": {
                    "type": "array",
                    "items": {
                        "type": "object",
                        "properties": {
                            "hash": {
                                "type": "string"
                            },
                            "string": {
                                "type": "string"
                            }
                        },
                        "additionalProperties": false
                    }
                },
                "extensions": {
                    "type": "array",
                    "items": {
                        "type": "object",
                        "properties": {
                            "name": {
                                "type": "string"
                            },
                            "type": {
                                "type": "integer"
                            },
                            "values": {
                                "type": "array",
                                "items": {
                                    "type": "string"
                                }
                            }
                        },
                        "additionalProperties": false
                    }
                },
                "ja3": {
                    "type": "object",
                    "optional": true,
                    "properties": {
                        "hash": {
                            "type": "string"
                        },
                        "string": {
                            "type": "string"
                        }
                    },
                    "additionalProperties": false
                },
                "ja3s": {
                    "type": "object",
                    "optional": true,
                    "properties": {
                        "hash": {
                            "type": "string"
                        },
                        "string": {
                            "type": "string"
                        }
                    },
                    "additionalProperties": false
                },
                "sni": {
                    "type": "string"
                },
                "ua": {
                    "type": "string"
                },
                "version": {
                    "type": "string"
                }
            },
            "additionalProperties": false
        },
```


schema/types/suricata.schema
```diff
@@ -30,6 +30,29 @@ type suricata.component.app_proto = record {
   app_proto: string
 }
 
+type suricata.component.ja3 = record {
+      hash: string #index=hash,
+      string: string,
```
tobim, line 35: Please align the indentation.
```suggestion
  hash: string #index=hash,
  string: string,
```

tobim (approved):
I would prefer to merge this today instead of waiting for a sample that contains all fields.

schema/types/suricata.schema
dispanser: thanks, that slipped by :)

tobim:
The `import-time` integration test seems to consistently fail on macOS. Any ideas?

PR #2796, opened 2022-12-14
Remove broker plugin

patszt:
We've removed the Broker plugin—it was written for Zeek 3.x and never worked with newer versions. 
It's dependencies were also troublesome to get them working with CAF 0.18.
The real-time Zeek import will return in a future plugin by connecting to Zeek directly via a websocket


dominiklohmann (approved):
Thanks for splitting this one out!

PR #2712, opened 2022-11-17
Introduce regression tests

dominiklohmann:
This adds automatic regression tests to our CI, comparing the current build against all previously tagged releases.

PR #2797, opened 2022-12-14
Remove the archive

dominiklohmann:
Following the introduction of a minimum partition version of 1 with tenzir/vast#2778, we can now—finally—remove the archive and everything that belongs to it.


libvast/builtins/stores/segment_store.cpp
```diff
@@ -468,16 +285,11 @@ class plugin final : public virtual store_actor_plugin {
...
-    auto header = chunk::make(std::move(path_str));
-    auto builder
-      = fs->home_system().spawn(active_local_store, accountant, fs, id);
-    return builder_and_header{
-      static_cast<system::store_builder_actor&&>(builder), std::move(header)};
+  make_store_builder([[maybe_unused]] system::accountant_actor accountant,
+                     [[maybe_unused]] system::filesystem_actor fs,
+                     [[maybe_unused]] const vast::uuid& id) const override {
+    return caf::make_error(ec::logic_error, "segment-store plugin is read-only "
+                                            "since VAST v2.4");
```
tobim, line 292: Please restrict this PR to the archive only. The segment store should be done in a separate PR.

tobim (dismissed):
A major milestone.

I think ripping out the segment store in the same PR is a bit overeager. As it stands the read part is still present and reachable, but not tested.

libvast/builtins/stores/segment_store.cpp
dominiklohmann: I disagree on this. Just removing the archive and not removing the segment_store and the write path of the segment-store store backend is a lot more work. Since users already cannot call into this code since the last release it doesn't make sense to keep it around.

dominiklohmann: Would you be fine with a middle ground that we use a small database created with VAST v2.3 and check whether we can still read it? Because that's what #2712 would do; I'm stuck with it currently because of some Docker in CI woes, but I think that we will always have things that we can still read but no longer write and we don't really want to keep too much of the surrounding code, I think.

tobim: > Just removing the archive and not removing the segment_store and the write path of the segment-store store backend is a lot more work.

How can that be? You don't have to touch the segment store at all.

Throwing out everything without consideration is just sloppy.

dominiklohmann: It's more work because we then cannot remove the `segment_builder` and `segment_store`, which are the last pieces of VAST that actually require table slices themselves to have an offset. The offset is a major annoyance when writing functions that work with table slices.

I also don't really see the point of maintaining a test for something that we no longer support. The read path of the segment store will be tested again as of #2712, and there is no need to keep the write path around if it's just dead code.

What does this unit test verify that isn't also covered by importing data with VAST v2.3 and exporting it with VAST v3.0?

mavam: I'm with Dominik on this. Why not throw it out if it'll be gone with the next release?

dominiklohmann: Just to make that clear @mavam, the read path of the partition-local segment-store will not be "gone with the next release." That would require bumping the minimum required partition version to two instead of one.

I did remove the write path of the partition-local segment-store, and Tobias complained that because of this I also had to remove the unit test for its read path, since a unit test that cannot write also cannot read.

My reasoning for why this is entirely fine is that a test that ingests data from an older version of VAST (>=v2.2, <v2.4) and then tries to read it with the most recent version of VAST is much better suited for this, because it means that we do not have to keep code around that is _only_ used in unit tests. Since that will be achieved by #2712 I think we can safely already remove the code and old unit tests here.

tobim: We came to the agreement that we'll add a small segment file to `libvast_test` and use it to preserve the unit test for the segment store read path. Then the rest can go.

dominiklohmann: I added such a test. Mind taking another look and approving if you're fine with it?

dominiklohmann: The commit to look at for this is https://github.com/tenzir/vast/pull/2797/commits/37dc01599030659e481aaed8398a32730c85dccd. I enabled auto-merge, so all that's missing is an approval. :)

tobim: The newly added test was all that was needed.

tobim: The newly added test was all that was needed.

tobim (approved):
:partying_face:

PR #2798, opened 2022-12-14
Remove outdated table slice encodings

dominiklohmann:
Following the introduction of a minimum partition version of 1 with tenzir/vast#2778, we're no longer able to encounter any table slices that are MessagePack-encoded or use an older version of the mapping of our data model onto Apache Arrow.

There's a future refactoring that is now unblocked, which is a total replacement of the legacy baggage that is the `table_slice` API with an easier-to-use wrapper around the same underlying data that makes more assumptions about the underlying column-major memory layout.

This is based on tenzir/vast#2797 to avoid merge conflicts.


mavam (approved):
The changes look good. Can you convince the reviewer that they are complete?

dominiklohmann:
> @mavam
> Can you convince the reviewer that they are complete?

The process was the following:
- Remove the four outdated entries in the FlatBuffers table without changing the index of the current one; I manually tested this for backwards compatibility.
- Remove code that tried to access the now removed entries in the code that flatc generated.
- Remove all code related to MessagePack.
- Remove all deprecated functions related to Arrow.
- Remove everything relating to the upgrade path from older Arrow encodings to the most recent one.

The one thing `table_slice` did right was hiding the underlying FlatBuffers table, which was only ever included in a single file, namely `table_slice.cpp`. This made it rather trivial to do remove the old encodings.

PR #2800, opened 2022-12-15
Use Yarn v3 with proper Nix support

dit7ya:
We want to use Yarn v3, which offers much better+faster+smoother DX compared to Yarn v1. However, for our current build via Nix, we are relying upon https://github.com/serokell/nix-npm-buildpackage which supports Yarn v1 only.

This PR uses https://github.com/stephank/yarn-plugin-nixify to generate `yarn-project.nix` file, which can be used for building the entire thing via Nix, while using Yarn v3 for local development.

TODO in the next commits: actually integrate to our top level nix and (if required) CMake build files.


dit7ya:
I want to wait a little bit for the the yarn-plugin-nixify to mature - that's why I want this PR to remain in the draft state for a while.

dit7ya:
Closing as not really relevant after #2922.

PR #2803, opened 2022-12-16
Optimize `select` and `filter` on table slices

dominiklohmann:
This makes a significant change to how `select` on table slices operates: Instead of re-creating the selected data, it now slices the incoming table slices without copying anything. `filter` is now just `concatenate(select(...))`, where `concatenate` is a new operation that we kind of already had for the internal `rebatch` operator of the `rebuild` command.

Additionally, this fixes `truncate` and `split` to properly handle import timestamp and offset, renames `truncate` to `head`, introduces a `tail` operation, and changes `split` to just be a call to `head` and `tail` respectively.

This should be tested for the performance delta before merging. In terms of review, `join` is the only thing here that may look really scary, but remember that it's almost a carbon copy of what we had already reviewed for the `rebuild` command.

### DIFF
--- a/libvast/builtins/commands/rebuild.cpp
+++ b/libvast/builtins/commands/rebuild.cpp
@@ -71,7 +71,7 @@ public:
       const auto remainder = desired_batch_size_ - buffered_rows;
       auto [head, tail] = split(slice, slice.rows() - remainder);
       buffer_.push_back(head);
-      auto result = join(std::exchange(buffer_, {}));
+      auto result = concatenate(std::exchange(buffer_, {}));
       results_.emplace_back(result.layout(), to_record_batch(result));
       buffer_.push_back(tail);
     }
@@ -80,7 +80,7 @@ public:
 
   caf::expected<std::vector<pipeline_batch>> finish() override {
     if (rows(buffer_) > 0) {
-      auto result = join(std::exchange(buffer_, {}));
+      auto result = concatenate(std::exchange(buffer_, {}));
       results_.emplace_back(result.layout(), to_record_batch(result));
     }
     return std::exchange(results_, {});
--- a/libvast/builtins/commands/rebuild.cpp
+++ b/libvast/builtins/commands/rebuild.cpp
@@ -56,178 +56,41 @@ public:
   /// @param schema The schema of the input and output table slices.
   /// @param desired_batch_size The desired output batch size; guaranteed to be
   /// exactly matched for all but the last output batch.
-  rebatch_operator(type schema, int64_t desired_batch_size)
-    : schema_{std::move(schema)},
-      desired_batch_size_{desired_batch_size},
-      arrow_schema_{schema_.to_arrow_schema()},
-      builder_{caf::get<record_type>(schema_).make_arrow_builder(
-        arrow::default_memory_pool())} {
-    const auto resize_result = builder_->Resize(desired_batch_size_);
-    VAST_ASSERT(resize_result.ok(), resize_result.ToString().c_str());
+  rebatch_operator(type schema, size_t desired_batch_size)
+    : schema_{std::move(schema)}, desired_batch_size_{desired_batch_size} {
+    // nop
   }
 
   caf::error
   add(type schema, std::shared_ptr<arrow::RecordBatch> batch) override {
-    VAST_ASSERT(schema == schema_);
-    VAST_ASSERT(builder_ != nullptr);
-    const auto batch_size = batch->num_rows();
-    if (builder_->length() == 0 && batch_size == desired_batch_size_) {
-      result_.push_back(batch);
-      return caf::none;
+    auto slice = table_slice{batch, std::move(schema)};
+    const auto buffered_rows = rows(buffer_) + slice.rows();
+    if (buffered_rows < desired_batch_size_) {
+      buffer_.push_back(std::move(slice));
+    } else {
+      const auto remainder = desired_batch_size_ - buffered_rows;
+      auto [head, tail] = split(slice, slice.rows() - remainder);
+      buffer_.push_back(head);
+      auto result = join(std::exchange(buffer_, {}));
+      results_.emplace_back(result.layout(), to_record_batch(result));
+      buffer_.push_back(tail);
     }
-    for (auto offset = int64_t{}; offset < batch->num_rows();) {
-      const auto length = std::min(desired_batch_size_ - builder_->length(),
-                                   batch_size - offset);
-      append_slice_to_builder(batch, offset, length);
-      offset += length;
-      VAST_ASSERT(builder_->length() <= desired_batch_size_);
-      if (builder_->length() == desired_batch_size_)
-        finish_builder();
-    }
-    return caf::none;
+    return {};
   }
 
   caf::expected<std::vector<pipeline_batch>> finish() override {
-    if (builder_->length() > 0)
-      finish_builder();
-    auto result = std::vector<pipeline_batch>{};
-    result.reserve(result_.size());
-    for (auto&& batch : std::exchange(result_, {}))
-      result.emplace_back(schema_, std::move(batch));
-    return result;
+    if (rows(buffer_) > 0) {
+      auto result = join(std::exchange(buffer_, {}));
+      results_.emplace_back(result.layout(), to_record_batch(result));
+    }
+    return std::exchange(results_, {});
   }
 
 private:
-  /// Appends a slice of a batch to the currently ongoing builder.
-  /// @param batch The record batch to partially rebuild.
-  /// @param offset The starting row inside the batch.
-  /// @param length THe length of the slice for our batch.
-  ///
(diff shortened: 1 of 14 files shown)


libvast/include/vast/id_range.hpp
```diff
@@ -0,0 +1,27 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2018 The VAST Contributors
```
mavam, line 6:
```suggestion
// SPDX-FileCopyrightText: (c) 2022 The VAST Contributors
```

libvast/include/vast/table_slice.hpp
```diff
@@ -314,14 +283,39 @@ class table_slice final {
 
 // -- operations ---------------------------------------------------------------
 
+/// Joins all slices in the given range.
+/// @param slices The input table slices.
+table_slice join(std::vector<table_slice> slices);
```
mavam, line 288: The name is confusing. When I read "join", I am thinking "what's the key you're joining on?". If you're literally concatenating them, I have some suggestions:

1. `concat(enate)`
2. `bind_rows` (dplyr style, which also has `bind_cols`)
3. `stack`

But the vertical data fusion must come up, which is definitely not the case with `join`.

libvast/builtins/commands/rebuild.cpp
```diff
@@ -56,178 +56,41 @@ class rebatch_operator final : public pipeline_operator {
...
-      VAST_ASSERT(builder_->length() <= desired_batch_size_);
-      if (builder_->length() == desired_batch_size_)
-        finish_builder();
+    auto slice = table_slice{batch, std::move(schema)};
+    const auto buffered_rows = rows(buffer_) + slice.rows();
+    if (buffered_rows < desired_batch_size_) {
+      buffer_.push_back(std::move(slice));
+    } else {
+      const auto remainder = desired_batch_size_ - buffered_rows;
+      auto [head, tail] = split(slice, slice.rows() - remainder);
```
mavam, line 72: This split should be purely logical (e.g., based on bitmaps or tracking row offsets). I would not expect this function to perform expensive work other than adding data to a builder.

libvast/include/vast/bitmap_algorithms.hpp
```diff
@@ -492,6 +508,29 @@ auto select(const IDs& ids) {
   return select_range<Bit, range_type>(bit_range(ids));
 }
 
+/// Yields all contiguous ranges of *Bit* in *bitmap*.
+/// @param bitmap The bitmap.
+/// @returns A generator id ranges.
+template <bool Bit = true, class Bitmap>
+auto select_runs(const Bitmap& bitmap) -> detail::generator<id_range> {
```
mavam, line 515: I wish I had generators when I wrote the first version of all these algorithms. So much complexity now...

libvast/src/table_slice.cpp
```diff
@@ -399,101 +393,204 @@ std::span<const std::byte> as_bytes(const table_slice& slice) noexcept {
...
+  if (slices.empty())
+    return {};
+  if (slices.size() == 1)
+    return std::move(slices[0]);
+  auto schema = slices[0].layout();
+  VAST_ASSERT(std::all_of(slices.begin(), slices.end(),
+                          [&](const auto& slice) {
+                            return slice.layout() == schema;
+                          }),
+              "join requires slices to be homogeneous");
```
mavam, line 412: Should this assertion come earlier?

libvast/src/table_slice.cpp
```diff
@@ -399,101 +393,204 @@ std::span<const std::byte> as_bytes(const table_slice& slice) noexcept {
...
-}
-
-std::vector<table_slice>
-select(const table_slice& slice, const ids& selection) {
-  std::vector<table_slice> result;
-  select(result, slice, selection);
-  return result;
 }
 
 table_slice truncate(table_slice slice, size_t num_rows) {
```
mavam, line 564: NB: Calling this function `head` (and the dual `tail`) would make for more intuitive semantics.

mavam (approved):
I didn't review the implementation in `table_slice.cpp`. If you think that deserves a critical look, I would ask @tobim.

There are a few high-level things that we should discuss soon. For example, the bitmap representation is now almost a decade old. But you are aware of the need to take another look at the assumptions back then.

libvast/include/vast/id_range.hpp
dominiklohmann: This was just copied out of another header to resolve a circular dependency, so I wasn't sure whether to keep the original date or use a new one. I didn't actually change any of the code in this.

libvast/include/vast/table_slice.hpp
dominiklohmann: I like `concat`, and definitely agree on `join` being confusing since we're operating on two-dimensional data here. Thanks for the suggestions!

libvast/builtins/commands/rebuild.cpp
dominiklohmann: Yeah, `split` basically just calls `batch->Slice(...)` twice now, so this has O(1). In most cases we should be in the `if` case instead of the `else` here, as after a single rebuild data should have the desired batch size already.

NB: I didn't change the rebuild algorithm at all, it now just calls the functions on table slices rather than doing everything on the underlying record batches, as that now maps to the same set of operations.

libvast/include/vast/bitmap_algorithms.hpp
dominiklohmann: I thought the same when I wrote this—this was so easy to write. I would love to just scrap the bitmap stuff and to reinvent it with generators. There are so many opportunities for optimization here.

libvast/src/table_slice.cpp
dominiklohmann: Hah, I thought about renaming it to `head` yesterday when I touched it. I'll do that!

libvast/src/table_slice.cpp
dominiklohmann: I don't think it can come any earlier. In order to check this we need to (1) have dropped all empty slices from the list and (2) have gotten the schema from the first entry for comparison with the others, unless we want to implement this using `adjacent_find`, but I find that a lot harder to read.

libvast/src/store.cpp
```diff
@@ -222,8 +222,8 @@ base_store::count(expression expr, ids selection) const {
 detail::generator<table_slice>
 base_store::extract(expression expr, ids selection) const {
   for (const auto& slice : slices()) {
-    if (auto filtered_slice = filter(slice, expr, selection))
-      co_yield std::move(*filtered_slice);
+    for (auto selected : select(slice, expr, selection))
+      co_yield std::move(selected);
```
dominiklohmann, line 226: This is the one change I would like to get @tobim's input on. Everything else is mostly trivial, shuffling code around, or already reviewed by @mavam.

I'd also be happy to move this into a follow-up PR as this likely has performance implications.

libvast/include/vast/table_slice.hpp
dominiklohmann: I renamed it to `concatenate`.

libvast/src/table_slice.cpp
dominiklohmann: Renamed it to `head`, and also introduced `tail` and then implemented `split` in terms of `head` and `tail`.

libvast/include/vast/table_slice.hpp
```diff
@@ -314,23 +283,49 @@ class table_slice final {
...
+/// that match the given expression. Does not preserve ids; use `select`
+/// instead if the id mapping must be maintained.
+/// @param slice The input table slice.
+/// @param expr The expression to evaluate.
+/// @param hints An ID set for pruning the events that need to be considered.
+/// @returns a new table slice consisting only of events matching the given
+///          expression.
+/// @pre `slice.encoding() != table_slice_encoding::none`
+std::optional<table_slice>
+filter(const table_slice& slice, expression expr, const ids& hints);
```
dominiklohmann, line 309: Meta-comment: If this is now much faster than before than we should write a changelog entry.

mavam: I would argue that performance-related changes should rather go in a release blog post.

dominiklohmann: Makes sense to me.

libvast/src/store.cpp
tobim: I benchmarked this on my laptop and discovered that `filter` performs a lot better for queries that hit lots of small runs in the data. `select` is actually up to 25x times slower. Let's revert this change, the rest is good to go!

dominiklohmann: I rebased onto master to resolve the merge conflict and dropped the one commit that made this change (it changed nothing else). Thanks for testing this extensively!

PR #2805, opened 2022-12-17
Make feather stores read incrementally

dominiklohmann:
This changes `feather` stores not to use the official Apache Feather reader anymore, but rather use its lower-level building blocks to do reads incrementally. This means that loading a feather store should not take any time anymore; the actual reading only happens on demand as batches are accessed. This is implemented in a way that the store caches already read batches so they're not read again for concurrent queries.

NB: This technically drops support for reading Apache Feather V1 files, but VAST never wrote them to begin with and we never officially said we support that.

In terms of review: We need to test this for performance changes, and if they are noteworthy write a changelog entry for this.

### DIFF
--- a/libvast/builtins/stores/feather.cpp
+++ b/libvast/builtins/stores/feather.cpp
@@ -10,6 +10,8 @@
 #include <vast/chunk.hpp>
 #include <vast/concept/convertible/data.hpp>
 #include <vast/data.hpp>
+#include <vast/detail/collect.hpp>
+#include <vast/detail/generator.hpp>
 #include <vast/detail/narrow.hpp>
 #include <vast/error.hpp>
 #include <vast/fwd.hpp>
@@ -20,10 +22,14 @@
 #include <arrow/io/file.h>
 #include <arrow/io/memory.h>
 #include <arrow/ipc/feather.h>
+#include <arrow/ipc/reader.h>
 #include <arrow/table.h>
+#include <arrow/util/iterator.h>
 #include <arrow/util/key_value_metadata.h>
 
 namespace vast::plugins::feather {
+
+namespace {
 
 /// Configuration for the Feather plugin.
 struct configuration {
@@ -43,8 +49,6 @@ struct configuration {
     return result;
   }
 };
-
-namespace {
 
 auto derive_import_time(const std::shared_ptr<arrow::Array>& time_col) {
   return value_at(time_type{}, *time_col, time_col->length() - 1);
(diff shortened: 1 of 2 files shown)


libvast/builtins/stores/feather.cpp
```diff
@@ -92,60 +96,89 @@ auto wrap_record_batch(const table_slice& slice)
...
+                                     cached_slices_[0].layout()};
+        slice.offset(offset);
+        slice.import_time(derive_import_time(import_time_column));
+        cached_slices_.push_back(std::move(slice));
+      } else {
+        co_return;
+      }
+      co_yield cached_slices_[i];
+      offset += cached_slices_[i].rows();
+      ++i;
```
dominiklohmann, line 158: Meta-comment: If this is now much faster than before than we should write a changelog entry.

dominiklohmann: When monitoring the `passive-store.init.runtime` metric, it is now consistently <10ms per store when previously during peak load it averaged ~500ms per store for me when testing on a database with 300M events in a test that rebuilds the whole database (i.e., a stress test).

I then ran `vast export -n 1 json` 10 times for a cold start of VAST with both master and this PR 20 times each on the same database, and the averaged results positively surprised me:

||master|this PR|
|-:|-:|-:|
|stores loaded|10|5|
|per store init time|55.1ms|4.2ms|
|total store init time|551.2ms|21.0ms|
|time to first result|19.8ms|14.5ms|

I would say that these results are significant. It doesn't make VAST much faster to the user because we still have the signal monitor bug in VAST that Tobias wanted to fix, but regardless of that it makes queries that are done quickly or cancelled now take significantly less CPU time in VAST, and in general makes loading a store much more friendly to the actor scheduler, which _should_ make the whole system feel a bit more responsive under high load.

dominiklohmann: Some further testing using the command `vast export null`, again on a cold start of VAST averaged over 20 runs:

||master|this PR|
|-:|-:|-:|
|time to first result|69.2ms|25.4ms|
|total time|39.38s|33.30s|
|per store init time|386.5ms|7.3ms|

Yeah, this definitely needs a changelog entry.

libvast/builtins/stores/feather.cpp
```diff
@@ -92,60 +96,89 @@ auto wrap_record_batch(const table_slice& slice)
...
+    auto next = generator();
+    if (!next.is_finished())
+      next.Wait();
+    VAST_ASSERT(next.is_finished());
+    auto result = next.MoveResult().ValueOrDie();
+    if (arrow::IsIterationEnd(result))
+      co_return;
+    co_yield std::move(result);
+  }
+}
```
mavam, line 116: So readable, thanks to generators!

libvast/builtins/stores/feather.cpp
mavam: > To summarize: Yeah, this definitely needs a changelog entry.

I would put this comment directly in the next release blog post, even within this PR.

dominiklohmann: I added a changelog entry, a blog post for v2.4.1, and bumped versions.

All this needs is approval and it needs to be cherry-picked onto a v2.4.x branch, and then we can cut a new release tomorrow that includes just this over v2.4.0.

web/blog/vast-v2.4.1/index.md
```diff
@@ -0,0 +1,61 @@
...
+
+[github-vast-release]: https://github.com/tenzir/vast/releases/tag/v2.4.1
+
+<!--truncate-->
+
+## Reading Feather Files Incrementally
+
+VAST's Feather store naïvely used the [Feather reader][feather-reader] from the
+Apache Arrow C++ library in its initial implementation. However, its API is
+rather limited: It does not support reading incrementally. We've swapped this
```
mavam, line 21:
```suggestion
rather limited: It does not support reading record batches incrementally. We've swapped this
```

I was stuck with the "what are we reading" when going over this.

web/blog/vast-v2.4.1/index.md
```diff
@@ -0,0 +1,61 @@
...
+
+[feather-reader]: https://github.com/apache/arrow/blob/apache-arrow-10.0.1/cpp/src/arrow/ipc/feather.h#L57-L108
+
+This is best explained visually:
+
+![Incremental Reads](incremental-reads-light.png#gh-light-mode-only)
+![Incremental Reads](incremental-reads-dark.png#gh-dark-mode-only)
+
+Within the scope of a single store, queries take the same amount of time
+overall, but there exist two distinct advantages of this approach:
```
mavam, line 32: Is it really the same performance? The drawing suggests that I/O and compute are now interleaved, which I would imagine is faster due to pipelining.

web/blog/vast-v2.4.1/index.md
```diff
@@ -0,0 +1,61 @@
...
+1. The first result arrives much faster at the client.
+2. Stores do less work for cancelled queries.
+
+One additional benefit that is not immediately obvious only comes into play a
```
mavam, line 36:
```suggestion
One additional benefit that is not immediately obvious comes into play when a
```

web/blog/vast-v2.4.1/index.md
```diff
@@ -0,0 +1,61 @@
...
+query arrives at multiple stores in parallel. Because the dksk reads are more
```
mavam, line 37:
```suggestion
query arrives at multiple stores in parallel: disk reads are more
```

web/blog/vast-v2.4.1/index.md
```diff
@@ -0,0 +1,61 @@
...
+evenly spread out now, the disk reads are less likely to overlap between stores
```
mavam, line 38:
```suggestion
evenly spread out now, making them less likely to overlap between stores.
```

web/blog/vast-v2.4.1/index.md
```diff
@@ -0,0 +1,61 @@
...
+queries in parallel. For disk-read performance-bound deployments this can lead
```
mavam, line 39:
```suggestion
For deployments wither slower I/O paths this can lead
```

web/blog/vast-v2.4.1/index.md
```diff
@@ -0,0 +1,61 @@
...
+to a significant query performance improvement.
+
+To verify and test this, I've created a VAST database with 300M Zeek events from
```
mavam, line 42:
```suggestion
To verify and test this, we've created a VAST database with 300M Zeek events.
```

Events alone is not sufficient to express the juiciness of a dataset. You need to add the actual size in GiB.

web/blog/vast-v2.4.1/index.md
```diff
@@ -0,0 +1,61 @@
...
+a Corelight sensor. All tests were performed on a cold start of VAST, i.e., I
```
mavam, line 43:
```suggestion
All tests were performed on a cold start of VAST, i.e., we
```

web/blog/vast-v2.4.1/index.md
```diff
@@ -0,0 +1,61 @@
...
+stopped and started VAST after every repetition of each test.
+
+I performed three tests:
```
mavam, line 46:
```suggestion
We performed three tests:

```

web/blog/vast-v2.4.1/index.md
```diff
@@ -0,0 +1,61 @@
...
+1. Export a single event (20 repetitions)
+2. Export all events (20 repetitions)
+3. Rebuild the entire database (3 repetitions)
```
mavam, line 49:
```suggestion
1. Export a single event (20 times)
2. Export all events (20 times)
3. Rebuild the entire database (3 times)
```

mavam: Can you add a linke to rebuild?

web/blog/vast-v2.4.1/index.md
```diff
@@ -0,0 +1,61 @@
+---
+title: VAST v2.4.1
+description: Faster Initial Results
```
mavam, line 3:
```suggestion
description: Faster Query Taste
```

or "Faster Time to Taste"

mavam (approved):
Nicely written blog post. I have only minor remarks.

I was left wondering why this "issue" only concern Feather. Can you add a sentence about the relationship to Parquet? Most readers will be more familiar with this format.

web/blog/vast-v2.4.1/index.md
```diff
@@ -0,0 +1,61 @@
...
+out with a more efficient implementation that does.
+
+[feather-reader]: https://github.com/apache/arrow/blob/apache-arrow-10.0.1/cpp/src/arrow/ipc/feather.h#L57-L108
+
+This is best explained visually:
+
+![Incremental Reads](incremental-reads-light.png#gh-light-mode-only)
+![Incremental Reads](incremental-reads-dark.png#gh-dark-mode-only)
+
+Within the scope of a single store, queries take the same amount of time
```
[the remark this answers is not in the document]
dominiklohmann: I reworded that to say "Feather store file," which picks up what we had in the paragraph before.

web/blog/vast-v2.4.1/index.md
dominiklohmann: I made it more explicit that this is about a single query in a single store.

web/blog/vast-v2.4.1/index.md
dominiklohmann: I've added this in parantheses.

dominiklohmann:
> I was left wondering why this "issue" only concern Feather. Can you add a sentence about the relationship to Parquet? Most readers will be more familiar with this format.

I added a note at the bottom that we plan to make the same improvement for our Parquet stores.

PR #2806, opened 2022-12-17
Fix build with unbundled CAF

tobim:
The recent CAF upgrade broke libvast_test with a prebuilt copy of CAF. Moving the dependency handling into the main CMakeLists.txt file resolves the issue.


cmake/VASTRegisterPlugin.cmake
```diff
@@ -412,6 +412,12 @@ function (VASTRegisterPlugin)
     VASTNormalizeInstallDirs()
   endif ()
 
+  # Enable compile commands for external plugins.
+  # Must be done before the targets are created for CMake >= 3.20.
+  if (NOT "${CMAKE_PROJECT_NAME}" STREQUAL "VAST")
+    set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+  endif ()
```
dominiklohmann, line 419: This should already be enabled via the `VASTExportCompileCommands` function. Doesn't that work for you?

CMakeLists.txt
```diff
@@ -679,6 +679,119 @@ if (VAST_ENABLE_MANPAGES)
...
+  )
+else ()
+  # Use bundled CAF.
+  if (NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/libvast/aux/caf/CMakeLists.txt")
+    message(
+      FATAL_ERROR
+        "CAF not found, either use -DCAF_ROOT_DIR=... or initialize the libvast/aux/caf submodule"
+    )
+  else ()
+    set(VAST_ENABLE_BUNDLED_CAF ON)
```
dominiklohmann, line 731: If I understand correctly then this is the only line that is actually problematic and would need to be changed or, as this is used in `VASTConfig.cmake.in`.

It makes a lot more sense to keep the CAF include in `libvast/CMakeLists.txt` and then to fix the actual underlying issue instead of moving things up because of directory-scope variables.

cmake/VASTRegisterPlugin.cmake
tobim: Did you read the comment? The function must be called with a target, but the option must be set before the targets are created.

CMakeLists.txt
tobim: It is neither. `libvast_test` doesn't see `CAF::test` when `find_package` is called in a different branch of the directory tree.
I think it is only logical to move the dependency setup one level up since both libraries need it.

cmake/VASTRegisterPlugin.cmake
dominiklohmann: The function does not set the option, but rather the target property which is initialized from the option (which is why the option would need to be set earlier).

CMakeLists.txt
dominiklohmann: Okay, that convinces me. Can you put that into a comment above where we load this? I believe the `find_package(OpenSSL)` can also move back to `libvast`.

cmake/VASTRegisterPlugin.cmake
tobim: The function does not set the target property. It even has a comment:
```
  # TODO: Use the EXPORT_COMPILE_COMMANDS property when using CMake >= 3.20.
```
The approach taken here works for all currently supported CMake versions.

CMakeLists.txt
tobim: The `find_package(OpenSSL)` is needed here because it's missing from `CAFConfig.cmake`.

dominiklohmann: Isn't it rather missing from VASTConfig.cmake? We added the OpenSSL dependency because of the pseudonymize operator, not because of CAF.

cmake/VASTRegisterPlugin.cmake
dominiklohmann: Not my brightest review 🙈

Should we just bump the minimum to 3.20 then? Otherwise we make it so targets defined before `VASTRegisterPlugin(...)` that link against the plugin target are not picked up by this.

tobim: Bumping to 3.20 would not help in this case. A plugin that defines another library beforehand is a rather weird setup and I suspect it wouldn't work with the current scaffold anyways.
I'm very much in favor of just merging this working fix and moving the focus back to user problems.

CMakeLists.txt
tobim: CAF::openssl depends on openssl but CAF itself does not pull it in. Before the CAF bump this was handled by the find module, but I'd rather not keep on maintaining that.
VASTConfig.cmake might technically need to pull in OpenSSL for static libvast installations, but nobody is using such a configuration at the moment. Feel free to push a fix for that to this branch, but leave the `find_package(OpenSSL)` here.

dominiklohmann: Makes sense. I might take a stab at cleaning this up a bit sometime next year. May also be a good idea to get a few fixes in upstream while CAF 0.19 is still in the rc-phase.

cmake/VASTRegisterPlugin.cmake
dominiklohmann: Why wouldn't it, if you set the target property on libvast, which we publicly link plugins against? That should propagate all the way up.

I care deeply about the maintainability of our CMake scaffolding because I invested a lot of time to get it to be mostly target-based, which is why I'm questioning every new piece of directory-based logic. If we didn't have that then our plugin API wouldn't be possible in the first place.

> I'm very much in favor of just merging this working fix and moving the focus back to user problems.

By that logic this PR shouldn't have existed in the first place since it's only relevant for developers using an oddly specific build setup that we do not test in CI. I think this argument is moot.

If you make such changes that are at first sight against best practices, then at least provide comments inline for why things are moved to what appears to be the wrong place so the next person touching it doesn't move it back, i.e., anticipate questions in review comments.

Feel free to resolve once such comments are in-place.

dominiklohmann (approved):
Approving modulo the remaining review comments.

cmake/VASTRegisterPlugin.cmake
tobim: In added a comment with directions towards a better fix that can be implemented after we assume CMake 3.20.

PR #2807, opened 2022-12-18
Eliminate signal monitoring shutdown lag

tobim:
This replaces the signal monitor with an adapter approach that translates POSIX signals into CAF messages.
The feature is intentionally limited to SIGINT and SIGTERM and should always lead to termination. Other signals are not supported in order to avert the temptation of using signals for poor mans IPC.

To the reviewer:
- [x] verify that client commands exit normally upon completion.
- [x] verify that client commands can be interrupted with a single `CTRL-C`.
- [x] verify that commands that start a node attempt a graceful shutdown with `CTRL-C` but can be terminated immediately with a second `CTRL-C`.
- [x] verify that client commands with the `-N` option exit regularly upon completion.


libvast/src/system/signal_reflector.cpp
```diff
@@ -0,0 +1,74 @@
...
+  sigaddset(&sigset, SIGTERM);
+  return sigset;
+}
+
+signal_reflector_actor::behavior_type signal_reflector(
+  signal_reflector_actor::stateful_pointer<signal_reflector_state> self,
+  sigset_t sigset) {
+  self->set_exit_handler([=](const caf::exit_msg&) {
+    self->state.stop = true;
+    kill(getpid(), SIGTERM);
```
dominiklohmann, line 42: I'm not sure this is correct. On this branch, if I press C-c once, `vast start` exits with exit code 143. That should be zero instead.

libvast/src/system/signal_reflector.cpp
```diff
@@ -0,0 +1,74 @@
...
+      if (!self->state.handler) {
+        if (raise(signum) != 0) {
+          // We don't want a backtrace when we get here.
+          std::signal(SIGABRT, SIG_DFL); // NOLINT
+          std::terminate();
+        }
+        __builtin_unreachable();
+      }
+      std::cerr << "\rinitiating graceful shutdown... (repeat request to "
+                   "terminate immediately)\n";
```
dominiklohmann, line 64: This doesn't work at all for me. I always get an ungraceful immediate exit, for both `vast start` and `vast -N ...`.

libvast/src/system/signal_reflector.cpp
```diff
@@ -0,0 +1,74 @@
...
+  sigemptyset(&sigset);
+  sigaddset(&sigset, SIGINT);
+  sigaddset(&sigset, SIGTERM);
+  return sigset;
+}
+
+signal_reflector_actor::behavior_type signal_reflector(
+  signal_reflector_actor::stateful_pointer<signal_reflector_state> self,
+  sigset_t sigset) {
+  self->set_exit_handler([=](const caf::exit_msg&) {
```
dominiklohmann, line 40: Do we care about the exit reason here?

libvast/src/system/signal_reflector.cpp
```diff
@@ -0,0 +1,74 @@
...
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2022 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "vast/system/signal_reflector.hpp"
+
+#include <caf/blocking_actor.hpp>
+
+#include <csignal>
```
dominiklohmann, line 13:
```suggestion
#include <csignal>
#include <unistd.h>
```

Otherwise this doesn't build on macOS.

dominiklohmann (changes requested):
The current approach does not exit gracefully for me.

libvast/src/system/pivot_command.cpp
```diff
@@ -56,14 +51,17 @@ caf::message pivot_command(const invocation& inv, caf::actor_system& sys) {
...
-  // Start signal monitor.
-  std::thread sig_mon_thread;
-  auto guard = system::signal_monitor::run_guarded(
-    sig_mon_thread, sys, defaults::system::signal_monitoring_interval, self);
+  if (local_node) {
+    // Register as the termination handler.
+    auto signal_reflector
+      = sys.registry().get<signal_reflector_actor>("signal-reflector");
+    self->send(signal_reflector, atom::subscribe_v);
+  }
```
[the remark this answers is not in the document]
tobim: None of these commands used the signal monitor before, so I did not "forget" about them. 
I guess it should be fine to do this in `spawn_node`.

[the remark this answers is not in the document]
dominiklohmann: Fair enough, I didn't actually grep for that but for `spawn_or_connect_to_node` instead. Still makes sense to tie this together, and also to change `flush` to use `connect_to_node` instead.

libvast/include/vast/system/signal_reflector.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include <caf/typed_event_based_actor.hpp>
+
+namespace vast::system {
+
+sigset_t termsigset();
+
+struct signal_reflector_state {
+  constexpr static inline auto name = "signal-reflector";
+
+  size_t dummy = 0;
```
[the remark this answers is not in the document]
tobim: This was needed with CAF 0.17 because that version tries to be a little too smart when initializing state structs. It's gone now.

libvast/src/system/signal_reflector.cpp
tobim: No. It should always be `user_shutdown`, and even if we get something else it should not affect the behavior.

libvast/src/system/pivot_command.cpp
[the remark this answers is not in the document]
tobim: I moved the `subscribe` part to `spawn_node` and added handling to `show`.
I think `rebuild stop` should also just use `connect_to_node`. `rebuild start` should work with `-N`, but it looks like it currently doesn't.
`web server` is a bit more involved because the main loop is managed by restinio.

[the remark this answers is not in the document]
dominiklohmann: `rebuild -N start` has the problem that the catalog returns an empty set before it received the partition synopses from the index. I hope that we can soon move the loading logic into the catalog to resolve this properly. I think it's fine to leave this with `spawn_or_connect_to_node`.

libvast/src/system/signal_reflector.cpp
dominiklohmann: This now works as expected.

libvast/src/system/signal_reflector.cpp
dominiklohmann: This now works as expected.

tobim:
macOS CI fails because it doesn't have `jthread`, but homebrew libc++ should have it. I'm not sure what's wrong here.
Could it be that homebrew clang uses apples libc++ by default?

libvast/src/system/signal_reflector.cpp
```diff
@@ -0,0 +1,74 @@
...
+        __builtin_unreachable();
+      }
+      std::cerr << "\rinitiating graceful shutdown... (repeat request to "
+                   "terminate immediately)\n";
+      self->send(self->state.handler, atom::signal_v, signum);
+    },
+    [self](atom::subscribe) {
+      self->state.handler
+        = caf::actor_cast<termination_handler_actor>(self->current_sender());
+    },
```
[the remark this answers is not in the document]
tobim: Since this is now wrapped in `connect_to_node` I would rather avoid the extra parameter.

dominiklohmann:
> macOS CI fails because it doesn't have `jthread`, but homebrew libc++ should have it. I'm not sure what's wrong here.
> Could it be that homebrew clang uses apples libc++ by default?

Homebrew libc++ doesn't have it, so you'll have to use `std::thread` instead and join manually.

libvast/src/system/signal_reflector.cpp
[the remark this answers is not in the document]
dominiklohmann: Which means that it won't be immediately visible when calling `spawn_node` or `spawn_or_connect_to_node`; not sure if that's a good idea, because it's pretty likely that the next developer calling one of these forgets about the requirement to add a handler that takes `(atom::signal, int)`.

I don't want to fight this strongly, but if you really don't agree then please put _very visible_ comments above these two functions.

vast/vast.cpp
```diff
@@ -231,13 +235,33 @@ int main(int argc, char** argv) {
...
   // thread-safe.
   auto sys = caf::actor_system{cfg};
+  // The reflector scope variable cleans up the reflector on destruction.
+  scope_linked<system::signal_reflector_actor> reflector{
+    sys.spawn<caf::detached>(system::signal_reflector)};
+  std::atomic<bool> stop = false;
+  auto signal_monitoring_thread = std::thread([&] {
+    int signum = 0;
+    sigwait(&sigset, &signum);
+    VAST_WARN("received signal {}", signum);
```
dominiklohmann, line 245:
```suggestion
    VAST_DEBUG("received signal {}", signum);
```

libvast/src/system/index.cpp
```diff
@@ -868,20 +868,25 @@ void index_state::decommission_active_partition(
...
+              self->send_exit(actor, caf::exit_reason::normal);
               if (completion)
                 completion(caf::none);
             },
             [=, this](const caf::error& err) {
               VAST_DEBUG("{} received error for request to persist partition "
                          "{} {}: {}",
                          *self, schema, id, err);
+              unpersisted.erase(id);
+              self->send_exit(actor, err);
```
dominiklohmann, line 880: I don't think this is correct: If we fail to persist a partition it is still unpersisted. It makes sense to do this in the success case, but not so much in the error case.

libvast/src/system/index.cpp
```diff
@@ -868,20 +868,25 @@ void index_state::decommission_active_partition(
...
               if (completion)
                 completion(err);
             });
       },
       [=, this](caf::error& err) {
         VAST_ERROR("{} failed to persist partition {} {} with error: {}", *self,
                    schema, id, err);
+        unpersisted.erase(id);
+        self->send_exit(actor, err);
```
dominiklohmann, line 889: I don't think this is correct: If we fail to persist a partition it is still unpersisted. It makes sense to do this in the success case, but not so much in the error case.

dominiklohmann (commented):
I still sometimes see the 30s delay in the integration tests:

```
2022-12-21 10:38:28 DEBUG    stopping server fixture: [PosixPath('/Users/dominiklohmann/Desktop/vast/master/build/bin/vast'), '--bare-mode', '--config=/Users/dominiklohmann/Desktop/vast/master/plugins/web/integration/vast.yaml', '-e', ':42024', 'stop']
2022-12-21 10:38:59 INFO     ran all 5 steps successfully
```

For the remainder, I left some questions, but I think this is mostly ready to go. Works nicely locally.

libvast/src/system/index.cpp
tobim: Doing nothing is not viable here. That would mean the partitions would stay in memory indefinitely.

dominiklohmann: Should the debug message above then be on the error level and also say that it drops the partition? (Same for the other comments below)

dominiklohmann:
> I still sometimes see the 30s delay in the integration tests:
> 
> ```
> 2022-12-21 10:38:28 DEBUG    stopping server fixture: [PosixPath('/Users/dominiklohmann/Desktop/vast/master/build/bin/vast'), '--bare-mode', '--config=/Users/dominiklohmann/Desktop/vast/master/plugins/web/integration/vast.yaml', '-e', ':42024', 'stop']
> 2022-12-21 10:38:59 INFO     ran all 5 steps successfully
> ```

We identified this to be a separate bug specific to the `web` plugin, where the `web server` command does not shut down properly. This must not necessarily be fixed by this PR.

dominiklohmann (approved):
Approving modulo the ongoing open discussions.

libvast/src/system/signal_reflector.cpp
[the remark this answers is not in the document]
tobim: I added comments for `spawn_node` and `spawn_or_connect_to_node`.

libvast/src/system/index.cpp
tobim: The one below was already an error. I clarified both messages.

dominiklohmann:
I enabled auto-merge here since this was really annoying over the past few weeks and shouldn't be blocking further work while most of everyone is on vacation.

PR #2808, opened 2022-12-18
Require VAST's schema metadata to exist in Arrow import

dominiklohmann:
This fixes a crash when importing arbitrary record batches and instead prints a proper error message. The fix for this is a bit more involved, but this is something better in the meantime.

### DIFF
--- a/libvast/src/format/arrow.cpp
+++ b/libvast/src/format/arrow.cpp
@@ -139,7 +139,15 @@ reader::read_impl(size_t max_events, size_t max_slice_size, consumer& f) {
       // record batch with status `OK` and `nullptr` as data and re-initialize
       // the reader.
       if (batch->batch != nullptr) {
-        const auto slice = table_slice{batch->batch};
+        auto schema = type::from_arrow(*batch->batch->schema());
+        if (!schema) {
+          return caf::make_error(ec::format_error,
+                                 fmt::format("failed to read next record "
+                                             "batch: schema metadata is "
+                                             "incompatible with VAST"));
+          continue;
+        }
+        const auto slice = table_slice{batch->batch, schema};
         f(slice);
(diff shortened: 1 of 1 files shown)


dominiklohmann:
Actually this turned out not to be the right approach. Even `type::from_arrow` assumes that the metadata is set, and I would say rightfully so. We will have to implement type inference from Arrow to get this to work properly.

mavam:
> Even type::from_arrow assumes that the metadata is set, and I would say rightfully so. We will have to implement type inference from Arrow to get this to work properly.

Is `from_arrow` only called on Arrow data that we already "made fit" into VAST?

dominiklohmann:
> Is `from_arrow` only called on Arrow data that we already "made fit" into VAST?

Yes. We just need to do this in the reader directly as downstream of it we want to be able to make the assumption that the data fits into our type system already. I don't want to deal with the mess that is Arrow's type visitor pattern in more than one place in the code base.

libvast/src/format/arrow.cpp
```diff
@@ -139,7 +139,15 @@ reader::read_impl(size_t max_events, size_t max_slice_size, consumer& f) {
...
       // the reader.
       if (batch->batch != nullptr) {
-        const auto slice = table_slice{batch->batch};
+        auto schema = type::from_arrow(*batch->batch->schema());
+        if (!schema) {
+          return caf::make_error(ec::format_error,
+                                 fmt::format("failed to read next record "
+                                             "batch: schema metadata is "
+                                             "incompatible with VAST"));
+          continue;
```
patszt, line 148: not needed continue after return

dominiklohmann:
I'm closing this as the fix requires casting, which we'll just have to implement before releasing the next version of VAST.

PR #2809, opened 2022-12-19
Fix parsing of CAF CLI options

patszt:
One function in configuration.cpp was still removing the caf. prefix when parsing CLI arguments. This logic gets removed as CAF now needs all it's options to be preceded by "caf."


dominiklohmann (approved):
Thanks for providing a test for this one, makes the approval straightforward. I trust CI on this one.

PR #2814, opened 2022-12-19
Fix UB in `order(...)` for floating point numbers

dominiklohmann:
The `coder` unit tests using the `order(...)` function on floating point numbers started failing for release builds on Debian after the CAF 0.18 bump, but the unit test suite wasn't touched since 2016, and the code itself not since 2018, so it was pretty like to be undefined behavior that we ran into. That turned out be rather easy to spot looking at the function.


mavam (approved):
`bit_cast` is indeed the right fix. But do you need to launder the pointer?

dominiklohmann:
> But do you need to launder the pointer?

Laundering suffices if you are no longer using `x` afterwards. We do use it again in the unit tests, though. Bitcasting is defined in terms of memcpy, which then all compilers in practice elide, so that will never have this issue.

Btw, I confirmed the issue to work using our Dockerfile, slightly modified to build and run `vast-test` rather than `vast`. I reproduced the issue on msater, and was unable to with this fix.

PR #2819, opened 2022-12-21
Make static binary CI integration test more reliable

tobim:
Using the `github` flake input scheme such as with `mach-nix` often hits GitHub's API rate limit.
This can be avoided by exposing an environment with the dependencies for the integration tests as a dedicated `shell` output.

- [x] Verify that CI passes.

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -496,7 +496,7 @@ jobs:
           VAST_BUILD_VERSION: ${{ needs.configure.outputs.build-version }}
           VAST_BUILD_VERSION_SHORT: ${{ needs.configure.outputs.build-version-short }}
         run: |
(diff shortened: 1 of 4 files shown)


dominiklohmann (approved):
The CI passes for this one, and the diff doesn't look like anything unexpected to me. If this gets rid of the rate limitation then I'd be happy to see this merged.

tobim:
Added a small improvement so we don't have to build our own version of pyarrow. But then I realized that we're using arrow 10. Which hasn't landed on `nixpkgs` master yet. The improvement is still worthwhile, so let's wait for the rebuild.

dit7ya (approved):
Can't comment on the Arrow stuff but the mach-nix removal looks good and is a nice improvement.

dominiklohmann:
I enabled auto-merge here since this was really annoying today and shouldn't be blocking further work while most of everyone is on vacation.

PR #2820, opened 2022-12-21
Remove the `augmented_partition_synopsis`

dominiklohmann:
Following the introduction of a minimum partition version of 1 with tenzir/vast#2778, the augmented partition synopsis is now irrelevant as the schema is always contained in the partition synopsis already. This has no user-facing implications, it's purely an internal cleanup.

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -610,7 +610,8 @@ caf::error index_state::load_from_disk() {
     caf::scoped_actor blocking{self->system()};
     blocking->request(transformer, caf::infinite, atom::persist_v)
       .receive(
-        [&synopses, &stats, this](std::vector<partition_synopsis_pair> result) {
+        [&synopses, &stats,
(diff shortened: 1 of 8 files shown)


libvast/src/system/index.cpp
```diff
@@ -610,8 +610,7 @@ caf::error index_state::load_from_disk() {
     caf::scoped_actor blocking{self->system()};
     blocking->request(transformer, caf::infinite, atom::persist_v)
       .receive(
-        [&synopses, &stats,
-         this](std::vector<augmented_partition_synopsis> result) {
+        [&synopses, &stats, this](std::vector<partition_synopsis_pair> result) {
```
patszt, line 613: Could this actually take lvalue ref?

patszt (approved):

libvast/src/system/index.cpp
dominiklohmann: Yes. I pushed a change that implements this to avoid the copy. Nice spot!

PR #2821, opened 2022-12-21
Remove rebuild support for heterogeneous partitions

dominiklohmann:
Following the introduction of a minimum partition version of 1 with tenzir/vast#2778, all partitions are now guaranteed to be homogeneous. The rebuilder as such does not need to take care of heterogeneous partitions any longer.

### DIFF
--- a/libvast/builtins/commands/rebuild.cpp
+++ b/libvast/builtins/commands/rebuild.cpp
@@ -129,7 +129,6 @@ struct statistics {
   size_t num_rebuilding = {};
   size_t num_completed = {};
   size_t num_results = {};
-  size_t num_heterogeneous = {};
 };
 
 /// The state of an in-progress rebuild.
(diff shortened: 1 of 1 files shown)


libvast/builtins/commands/rebuild.cpp
```diff
@@ -405,30 +395,7 @@ struct rebuilder_state {
...
-        run->remaining_partitions.erase(heterogenenous_partition);
-      } else {
-        // Wait until we have all heterogeneous partitions transformed into
-        // homogenous partitions before starting to work on the homogenous
-        // parittions. In practice, this leads to less undeful partitions.
-        auto rp = self->make_response_promise<void>();
-        return run->delayed_homogeneous_rebuilds.emplace_back(std::move(rp));
-      }
-    } else {
       // Take the first homogenous partition and collect as many of the same
```
tobim, line 399: Please reformat. The indentation is one level too deep.
This comment also doesn't have to mention "homogenous" any more.

tobim (approved):
Looks correct apart from the missing unindent.

libvast/builtins/commands/rebuild.cpp
dominiklohmann: Not sure why our `git format-branch` didn't pick this up, but once I dedented the commented it formatted the chunk correctly. I pushed the change, and also removed the unnecessary "homogeneous."

PR #2822, opened 2022-12-21
Remove rewriting of partition synopses without id range

dominiklohmann:
Following the introduction of a minimum partition version of 1 with tenzir/vast#2778, all partitions are now guaranteed to have an id range specified in their partition synopsis. Rewriting old partition synopses on startup is no longer necessary.


tobim (approved):
Good to see this go.

PR #2825, opened 2022-12-23
Add new API for partition_info by uuid extraction in Catalog

patszt:
The compaction plugin allows the user to compact a given partition by hand. It is required for the user to pass the UUID of a partition as an input. Since the index API changed from uuid to partition_info it was needed for compaction plugin to somehow map one to the other one. I find the extension of catalog API to be the cleanest way to do it.

### DIFF
--- a/libvast/include/vast/partition_synopsis.hpp
+++ b/libvast/include/vast/partition_synopsis.hpp
@@ -114,6 +114,12 @@ struct partition_info {
     // nop
   }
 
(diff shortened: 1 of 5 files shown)


tobim (approved):
I think this change is fine for the time being.

In the medium term we should follow this up with a change to the `partition_creation_listener` interface: It should simply forward the `replace` messages with `partition_info`s instead of `partition_synopsis_pair`s directly from the catalog. The listeners could then store the `partition_info`s as needed.

PR #2827, opened 2022-12-23
Debugging ASAN issues on CI - token policy

patszt:
This PR is used to reproduce the known ASAN issue on CI. The reproduction on local machine was giving no results.


dominiklohmann:
We fixed the ASan issue, so this PR is no longer relevant.

PR #2829, opened 2022-12-30
Create a custom type for handling CLI options

patszt:
The CAF 0.18.6 upgrade introduced a changes to caf::config_value_set parsing. The most used syntax for list config values was a simple CSV, e.g. --plugins=web,pcap

This doesn't work out of box and a free function was introduced to convert the incoming list arguments into a format that caf::config_value_set could parse. This was a ticking bomb, because you had to remember about it and "sanitize" the list value arguments before parsing. Instead of using caf::config_value_set and sanitizing the list value arguments the new type was created to do it automatically.

### DIFF
--- a/libvast/include/vast/config_options.hpp
+++ b/libvast/include/vast/config_options.hpp
@@ -1,0 +1,102 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#pragma once
+
+#include "vast/detail/string.hpp"
+
+#include <caf/config_option_set.hpp>
+#include <caf/string_view.hpp>
+#include <fmt/format.h>
+
+#include <unordered_set>
+
+namespace vast {
+
+/// A collection of configuration options that parses CLI arguments into a
+/// caf::settings object. Each configuration option consists of the option type
+/// (e.g. string, int, list), the name (e.g. console-verbosity) a category (e.g.
+/// vast or global if not specified) and a description.
+class config_options {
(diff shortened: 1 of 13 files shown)


libvast/src/config_options.cpp
```diff
@@ -0,0 +1,68 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2022 The VAST Contributors
```
dominiklohmann, line 6:
```suggestion
// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
```

libvast/include/vast/config_options.hpp
```diff
@@ -0,0 +1,102 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2022 The VAST Contributors
```
dominiklohmann, line 6:
```suggestion
// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
```

changelog/unreleased/breaking-changes/2829--change-reader-and-writer-plugin-api.md
```diff
@@ -0,0 +1,4 @@
+The reader_options method for a reader plugin and writer_options method for a
+writer plugin must return 'vast::config_options'. The new type allows for a
+customization in the CLI arguments parsing logic that may be needed in the
+future without breaking the API as it is owned by VAST.
```
dominiklohmann, line 4: I don't think this is user-facing, but rather developer-facing. No changelog entry should be necessary.

dominiklohmann (approved):
Thanks for this fix!

PR #2830, opened 2023-01-02
Add Plausible analytics to website

mavam:
This PR adds [Plausible](https://plausible.io/) analytics to our website.

### DIFF
--- a/web/docusaurus.config.js
+++ b/web/docusaurus.config.js
@@ -260,6 +260,8 @@ const config = {
         additionalLanguages: ['r'],
       },
     }),
+
+  scripts: [{src: 'https://plausible.io/js/script.js', defer: true, 'data-domain': 'vast.io'}],
 };
 
 module.exports = config;
--- a/web/docusaurus.config.js
+++ b/web/docusaurus.config.js
@@ -261,7 +261,7 @@ const config = {
       },
     }),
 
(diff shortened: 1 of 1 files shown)


mavam:
Per the [Docusaurus docs](https://docusaurus.io/docs/next/api/docusaurus-config#scripts), there's also an `async` option. Not sure if we want that.

web/docusaurus.config.js
```diff
@@ -260,6 +260,8 @@ const config = {
         additionalLanguages: ['r'],
       },
     }),
+
+  scripts: [{src: 'https://plausible.io/js/script.js', defer: true, 'data-domain': 'vast.io'}],
```
dit7ya, line 264:
```suggestion
  scripts: [{src: 'https://plausible.io/js/script.js', async:true, defer: true, 'data-domain': 'vast.io'}],
```

https://docusaurus.io/docs/api/docusaurus-config#scripts

> Note that <script> added here are render-blocking, so you might want to add async: true/defer: true to the objects.

This is how plausible themselves do it as well - https://github.com/plausible/docs/blob/08d804dea840f10c49e4118b5ddd5fe0bc722fd8/docusaurus.config.js#L8.

dit7ya (approved):

web/docusaurus.config.js
mavam: Added `async`, thanks.

PR #2832, opened 2023-01-03
Introduce a potpourri of smaller improvements

dominiklohmann:
This is a set of improvements seemingly unrelated to one another that amassed in one branch locally whenever I was traveling this vacation season.

These are the notable changes:
1. Plugin names are now case-insensitive
2. VASTQL is now a query-language plugin (builtin) so it no longer requires special-casing; this will allow for making the query language choice explicit in the future
3. Removed PCAP-specific logic from the export command
4. Removed handling of old deprecated options since the next release is going to be a major release
5. VAST now shuts down instantly when metrics are enabled instead of being held alive for up to the duration of the telemetry interval (10s)

(1), (4), and (5) probably need changelog entries.

Not urgent to review, I won't be back in action before next week anyways, and it might very well be the case that I push another few things onto this. :)


examples/plugins/analyzer/example_analyzer.cpp
```diff
@@ -150,7 +150,7 @@ class example_plugin final : public virtual analyzer_plugin,
   }
 
   /// Returns the unique name of the plugin.
-  const char* name() const override {
+  std::string_view name() const override {
```
mavam, line 153: Shouldn't these simply return a `std::string`?

dominiklohmann: Considering these are supposed to be string literals I don't think it makes much of a difference in practice; it probably is the safer option, though.

mavam: Yeah, and it makes it easier to use as a value. It's highly unlikely to be in the critical path and a solid standard library implementation uses SSO for 99% of these short names.

libvast/builtins/query-languages/vastql.cpp
```diff
@@ -0,0 +1,36 @@
...
+    if (parsers::expr(query, result))
+      return result;
+    return caf::make_error(ec::invalid_query,
+                           fmt::format("not a valid query: '{}'", query));
+  }
+};
+
+} // namespace vast::plugins::vastql
+
+VAST_REGISTER_PLUGIN(vast::plugins::vastql::plugin)
```
mavam, line 36: Nice. Please also add a docs page similar in nature to the Sigma frontend. We actually link to `VASTQL` several times in the docs. It might be worthwhile git-grepping through these places and decide whether you want to link to the plugin docs or the main language docs.

NB: Should QL really be PL soon? I could imagine that our pipeline language will be more than just queries, as it will also connect security tools.

libvast/include/vast/plugin.hpp
```diff
@@ -471,19 +480,24 @@ class plugin_ptr final {
...
-  auto it
+  const auto found
     = std::find_if(plugins.begin(), plugins.end(), [&](const plugin_ptr& p) {
-        return name == p->name();
+        const auto plugin_name = p->name();
+        return std::equal(name.begin(), name.end(), plugin_name.begin(),
+                          plugin_name.end(),
+                          [](unsigned char lhs, unsigned char rhs) noexcept {
+                            return std::tolower(lhs) == std::tolower(rhs);
+                          });
```
mavam, line 493: Nit: these seem so generally useful that I would almost put them as helper functions into `detail/strings.hpp`. Having something like `to_upper` and `to_lower` that can be invoked on a string(-view) would make code a lot more readable.

libvast/src/plugin.cpp
```diff
@@ -53,7 +53,7 @@ get_plugin_dirs(const caf::actor_system_config& cfg) {
         cfg, "vast.plugin-dirs"))
     result.insert(dirs->begin(), dirs->end());
   else
-    VAST_WARN("Unable to extract vast plugin dirs. {}", dirs.error());
+    VAST_WARN("failed to to extract vast plugin dirs: {}", dirs.error());
```
mavam, line 56:
```suggestion
    VAST_WARN("failed to to extract plugin dirs: {}", dirs.error());
```

Otherwise all-caps?

libvast/src/plugin.cpp
```diff
@@ -245,7 +242,7 @@ load(const std::vector<std::string>& bundled_plugins,
   // Sort loaded plugins by name.
   std::sort(get_mutable().begin(), get_mutable().end(),
             [](const auto& lhs, const auto& rhs) {
-              return std::strcmp(lhs->name(), rhs->name()) < 0;
+              return lhs->name() < rhs->name();
```
mavam, line 245: Isn't this [`lexicographical_compare`](https://en.cppreference.com/w/cpp/algorithm/lexicographical_compare)?

libvast/src/system/spawn_arguments.cpp
```diff
@@ -36,20 +37,21 @@ normalized_and_validated(std::vector<std::string>::const_iterator begin,
...
-      else
-        VAST_DEBUG("failed to parse query as {} language: {}", language->name(),
-                   expr.error());
-    }
+  auto query_languages = collect(plugins::get<query_language_plugin>());
+  if (const auto* vastql = plugins::find<query_language_plugin>("VASTQL")) {
+    const auto it
+      = std::find(query_languages.begin(), query_languages.end(), vastql);
+    VAST_ASSERT(it != query_languages.end());
+    std::rotate(query_languages.begin(), it, it + 1);
```
mavam, line 47: A comment along the lines of "trying VASTQL first before all others..." would make the C++ algorithm magic more digestable.

plugins/pcap/main.cpp
```diff
@@ -661,6 +661,15 @@ class writer : public format::writer {
...
   caf::error write(const table_slice& slice) override {
+    auto&& layout = slice.layout();
+    // TODO: relax this check. We really only need the (1) flow, and (2) PCAP
+    // payload. Everything else is optional.
+    if (!congruent(layout, make_packet_type())) {
+      return caf::make_error(
+        ec::format_error, fmt::format("pcap-writer is unable to write batch "
+                                      "of schema '{}': expected 'pcap.packet'",
+                                      slice.layout()));
+    }
```
mavam, line 672: I would do this not via congruency, now that the writer gets everything. Two approaches come to mind:

1. Just look at the schema name and bail out if it's not `pcap.packet`
2. Ensure that the blob payload has the right attribute value, e.g., `libpcap-packet`, to ensure that you can stuff it into the libpcap function without producing garbage

mavam (approved):
Neat stuff.

I know the PR is still an draft mode, but it was very reviewable already. 🙂

libvast/src/system/spawn_arguments.cpp
dominiklohmann: There is a comment two lines above saying 

> ```cpp
> // Get all query languages, but make sure that VAST is at the front.
> // TODO: let the user choose exactly one language instead.
> ```

plugins/pcap/main.cpp
dominiklohmann: I did not change this logic, I just moved it up a bit. I would prefer not to touch this for now as it's not currently something I want to invest time in.

libvast/src/plugin.cpp
dominiklohmann: Same deal as `std::string_view::operator<=>(std::string_view)`, just less readable.

plugins/pcap/main.cpp
mavam: But you fundamentally changed what data goes into reader, so you must adapt the logic accordingly. You cannot see this in isolation, the effect of your change is across multiple components that now must be re-harmonized.

mavam: In other words: this wasn't an issue in the past, but now it is.

dominiklohmann: I chose to restore the type-name restriction

libvast/include/vast/plugin.hpp
dominiklohmann: I chose to make plugin_ptr comparable with a strong ordering instead; this removes the need for the custom compare object in three places.

libvast/builtins/query-languages/vastql.cpp
dominiklohmann: I went through all the other things, but this is definitely not something I'm doing on the side. I'd prefer to do this in a follow-up PR. If you agree, then please write a story for it and resolve this; other than that the PR is now on auto-merge.

libvast/builtins/commands/rebuild.cpp
```diff
@@ -143,13 +143,13 @@ struct run {
 /// The interface of the REBUILDER actor.
 using rebuilder_actor = system::typed_actor_fwd<
   // Start a rebuild.
-  caf::reacts_to<atom::start, start_options>,
+  auto(atom::start, start_options)->caf::result<void>,
```
mavam, line 146: Why `auto` + `->` instead of putting `result<T>` to the front?

libvast/src/plugin.cpp
```diff
@@ -362,7 +356,7 @@ store_plugin::make_store_builder(system::accountant_actor accountant,
     = std::filesystem::path{"archive"} / fmt::format("{}.{}", id, name());
   auto store_builder = fs->home_system().spawn<caf::lazy_init>(
     default_active_store, std::move(*store), fs, std::move(accountant),
-    std::move(path), name());
+    std::move(path), std::string{name()});
```
mavam, line 359: Is this still needed after switching to `std::string`?

libvast/src/plugin.cpp
```diff
@@ -380,10 +374,9 @@ store_plugin::make_store(system::accountant_actor accountant,
   const auto id = uuid{header.subspan<0, uuid::num_bytes>()};
   auto path
     = std::filesystem::path{"archive"} / fmt::format("{}.{}", id, name());
-  return fs->home_system().spawn<caf::lazy_init>(default_passive_store,
-                                                 std::move(*store), fs,
-                                                 std::move(accountant),
-                                                 std::move(path), name());
+  return fs->home_system().spawn<caf::lazy_init>(
+    default_passive_store, std::move(*store), fs, std::move(accountant),
+    std::move(path), std::string{name()});
```
mavam, line 379: Same here.

mavam (approved):
Nice touch ups. 🌴

libvast/builtins/commands/rebuild.cpp
dominiklohmann: Practical reason: I did this with three calls to sed, and I couldn't figure out how to do both arguments and return value in one go.

I must say I do also find this much more readable as it aligns better with the lambdas in the behavior set.

mavam: Sed sets the coding standards, great. 😅

If you think it's more readable, I'm fine with it. Personally, I think we could reduce the number of characters. I would argue that it should look like a `std::function<...>` declaration where people put the return value first.

dominiklohmann: Honestly, I'd prefer using trailing return types everywhere. It's enforceable with clangd and you can even automatically transform to it using clangd's suggested actions. The scoping for trailing return types is much more natural than for leading return types. May actually be something to just enforce globally in the future to stay up to date with the times.

dominiklohmann:
[sc-40310]

PR #2834, opened 2023-01-03
Only enable consteval for compilers with support

tobim:
Our docs claim that VAST can be built with clang 13 and up, but `consteval` support was added in 15.

### DIFF
--- a/libvast/include/vast/config.hpp.in
+++ b/libvast/include/vast/config.hpp.in
@@ -154,6 +154,12 @@ extern bool has_undefined_behavior_sanitizer;
 #  define VAST_POSIX 0
 #endif
 
+#if (defined __cpp_consteval) && __cpp_consteval >= 201811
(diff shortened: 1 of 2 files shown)


mavam:
Should we consider updating to Clang 15 instead?

tobim:
NixOS doesn't provide clang 15 yet. Bumping would leave me in the dust.

mavam:
How hard is it to update your compiler? Does that take a lot of effort?

tobim:
I have to wait for llvm 15 to get merged into nixpkgs. But even if I could use it already, I don't think this fix is complex enough to justify dropping support for 2 compiler generations just so we can avoid it.

tobim:
Is there any real reason to reject this?

dominiklohmann:
> Is there any real reason to reject this?

No.

dominiklohmann (approved):

PR #2835, opened 2023-01-04
Implement a retry mechanism for VAST clients failing to connect to the server

patszt:
1. Added a new actor that allows the caller to connect with a remote node in async/sync manner with a retry mechanism.
2. Adjusted current connect_to_node to use the new actor as a implementation detail.

I have done a simple test to check the retry mechanism.
Run export -> wait a 2 - 3 seconds and run vast start.
Previously the client exits with error:

> ! system_error: failed to connect to VAST node at 127.0.0.1:42000: !! caf::sec::cannot_connect_to_node: ip_connect failed 127.0.0.1 message("ip_connect failed", "127.0.0.1", 42000)

Now it manages to export the data.

### DIFF
--- a/changelog/unreleased/features/2835--retry-mechanism-for-vast-clients.md
+++ b/changelog/unreleased/features/2835--retry-mechanism-for-vast-clients.md
@@ -1,0 +1,7 @@
+We changed VAST client processes to attempt connecting to a VAST server
+multiple times until the configured connection timeout (vast.connection-timeout
+, defaults to 5 minutes) runs out. A fixed delay between connection attempts
+(vast.connection-retry-delay, defaults to 3 seconds) ensures that clients to
+not stress the server too much. Set the connection timeout to zero to let VAST
+client attempt connecting indefinitely, and the delay to zero to disable the
+retry mechanism.
--- a/libvast/include/vast/defaults.hpp
+++ b/libvast/include/vast/defaults.hpp
@@ -265,7 +265,7 @@ inline constexpr std::chrono::milliseconds status_request_timeout
 
 /// Timeout for initial connections to the node.
 inline constexpr std::chrono::milliseconds node_connection_timeout
-  = std::chrono::seconds{10};
+  = std::chrono::minutes{5};
 
 /// Timeout for the scheduler to give up on a partition.
 inline constexpr std::chrono::milliseconds scheduler_timeout
--- a/libvast/include/vast/defaults.hpp
+++ b/libvast/include/vast/defaults.hpp
@@ -291,6 +291,10 @@ inline constexpr bool create_partition_index = true;
 /// Whether to spawn central components in separate threads.
 inline constexpr bool detach_components = true;
 
+/// Time to wait before trying to make another connection attempt to a remote
+/// VAST node.
+inline constexpr auto node_connection_retry_delay = std::chrono::seconds{3u};
+
 } // namespace system
 
 } // namespace vast::defaults
--- a/libvast/include/vast/system/connector.hpp
+++ b/libvast/include/vast/system/connector.hpp
@@ -1,0 +1,39 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#pragma once
+
+#include "vast/system/actors.hpp"
+
+#include <caf/io/middleman_actor.hpp>
+#include <caf/timespan.hpp>
+#include <caf/typed_event_based_actor.hpp>
+
+#include <chrono>
+#include <optional>
+
+namespace vast::system {
+
+struct connector_state {
+  // Actor responsible for TCP connection with a remote node.
+  caf::io::middleman_actor middleman;
+};
+
+/// @brief Creates an actor that establishes the connection to a remote VAST
+/// node.
+/// @param retry_delay Delay between two connection attempts. Don't retry if not
+/// set.
+/// @param deadline Time point after which the connector can no longer connect
+/// to a remote VAST node. Try connecting until success if not set.
+/// @return actor handle that can be used to connect with a remote VAST node.
+connector_actor::behavior_type
+connector(connector_actor::stateful_pointer<connector_state> self,
+          std::optional<caf::timespan> retry_delay,
+          std::optional<std::chrono::steady_clock::time_point> deadline);
+
+} // namespace vast::system
--- a/libvast/src/system/application.cpp
+++ b/libvast/src/system/application.cpp
@@ -426,7 +426,10 @@ auto make_root_command(std::string_view path) {
         .add<std::string>("store-backend", "store plugin to use for imported "
                                            "data")
         .add<std::string>("connection-timeout", "the timeout for connecting to "
-                                                "a VAST server (default: 10s)");
+                                                "a VAST server (default: 5m)")
+        .add<std::string>("connection-retry-delay",
+                          "the delay between vast node connection attempts"
+                          "a VAST server (default: 3s)");
   ob = add_index_opts(std::move(ob));
   auto root = std::make_unique<command>(path, "", std::move(ob));
   root->add_subcommand(make_count_command());
--- a/libvast/src/system/connect_to_node.cpp
+++ b/libvast/src/system/connect_to_node.cpp
@@ -11,28 +11,58 @@
 #include "vast/fwd.hpp"
 
 #include "vast/concept/parseable/vast/endpoint.hpp"
-#include "vast/concept/printable/to_string.hpp"
-#include "vast/concept/printable/vast/port.hpp"
-#include "vast/data.hpp"
 #include "vast/defaults.hpp"
 #include "vast/endpoint.hpp"
 #include "vast/error.hpp"
 #include "vast/logger.hpp"
+#include "vast/system/connect_request.hpp"
+#include "vast/system/connector.hpp"
 #include "vast/system/node_control.hpp"
 #include "vast/system/version_command.hpp"
 
-#include <caf/io/middleman.hpp>
-#include <caf/openssl/all.hpp>
 #include <caf/scoped_actor.hpp>
 #include <caf/settings.hpp>
 
 namespace vast::system {
+namespace {
 
-caf::expected<node_actor>
-connect_to_node(caf::scoped_actor& self, const caf::settings& opts) {
-  // Fetch values from config.
-  auto id = get_or(opts, "vast.node-id", defaults::system::node_id.data());
-  auto timeout = node_connection_timeout(opts);
+void assert_data_completness(const record& remote_version,
+                             const record& local_version) {
+  if (!local_version.contains("VAST"))
+    die("no VAST key found in a local version");
+  if (!remote_version.contains("VAST"))
+    die("no VAST key found in a remote version");
+  if (!local_version.contains("plugins"))
+    die("no plugins key found in a local version");
+  if (!remote_version.contains("plugins"))
+    die("no plugins key found in a remote version");
+}
+
+bool check_version(const record& remote_version) {
+  const auto local_version = retrieve_versions();
+  assert_data_completness(remote_version, local_version);
+  if (local_version.at("VAST") != remote_version.at("VAST")) {
+    VAST_WARN("client version {} does not match remote version {}; "
+              "this may cause unexpected behavior",
+              local_version.at("VAST"), remote_version.at("VAST"));
+    return false;
+  }
+  VAST_DEBUG("client verified that local VAST version matches remote "
+             "VAST version {}",
+             local_version.at("VAST"));
+  if (local_version.at("plugins") != remote_version.at("plugins")) {
+    VAST_WARN("client plugins {} do not match remote plugins {}; "
+              "this may cause unexpected behavior",
+              local_version.at("plugins"), remote_version.at("plugins"));
+    return false;
+  }
+  VAST_DEBUG("client verified that local VAST plugins match remote "
+             "VAST plugins {}",
+             local_version.at("plugins"));
+  return true;
+}
+
+caf::expected<endpoint> get_node_endpoint(const caf::settings& opts) {
   endpoint node_endpoint;
   auto endpoint_str
     = get_or(opts, "vast.endpoint", defaults::system::endpoint.data());
@@ -47,48 +77,56 @@ connect_to_node(caf::scoped_actor& self, const caf::settings& opts) {
   if (node_endpoint.port->type() != port_type::tcp)
     return caf::make_error(ec::invalid_configuration, "invalid protocol",
                            *node_endpoint.port);
-  VAST_DEBUG("client connects to remote node with id {}", id);
-  auto host = node_endpoint.host;
   if (node_endpoint.host.empty())
     node_endpoint.host = defaults::system::endpoint_host;
-  VAST_INFO("client connects to VAST node at {}", endpoint_str);
-  auto result = [&]() -> caf::expected<node_actor> {
-    if (self->system().has_openssl_manager()) {
-      return caf::openssl::remote_actor<node_actor>(
-        self->system(), node_endpoint.host, node_endpoint.port->number());
-    }
-    auto& mm = self->system().middleman();
-    return mm.remote_actor<node_actor>(node_endpoint.host,
-                                       node_endpoint.port->number());
-  }();
+  return node_endpoint;
+}
+
+std::optional<std::chrono::steady_clock::time_point>
+get_deadline(caf::timespan timeout) {
+  if (caf::is_infinite(timeout))
+    return {};
+  return {std::chrono::steady_clock::now() + timeout};
+}
+
+std::optional<caf::timespan> get_retry_delay(const caf::settings& settings) {
+  auto retry_delay
+    = caf::get_or<caf::timespan>(settings, "vast.connection-retry-delay",
+                                 defaults::system::node_connection_retry_delay);
+  if (retry_delay == caf::timespan::zero())
+    return {};
+  return retry_delay;
+}
+
+} // namespace
+
+caf::expected<node_actor>
+connect_to_node(caf::scoped_actor& self, const caf::settings& opts) {
+  // Fetch values from config.
+  auto node_endpoint = get_node_endpoint(opts);
+  if (!node_endpoint)
+    return std::move(node_endpoint.error());
+  auto timeout = node_connection_timeout(opts);
+  auto connector_actor
+    = self->spawn(connector, get_retry_delay(opts), get_deadline(timeout));
+  auto result = caf::expected<node_actor>{caf::error{}};
+  self
+    ->request(connector_actor, caf::infinite, atom::connect_v,
+              connect_request{node_endpoint->port->number(),
+                              node_endpoint->host})
+    .receive(
+      [&](node_actor& res) {
+        result = std::move(res);
+      },
+      [&](caf::error& err) {
+        result = std::move(err);
+      });
   if (!result)
-    return caf::make_error(
-      ec::system_error,
-      fmt::format("failed to connect to VAST node at {}:{}: {}",
-                  node_endpoint.host, node_endpoint.port->number(),
-                  std::move(result.error())));
-  VAST_DEBUG("client connected to VAST node at {}:{}", node_endpoint.host,
-             to_string(*node_endpoint.port));
+    return result;
   self->request(*result, timeout, atom::get_v, atom::version_v)
     .receive(
       [&](record& remote_version) {
-        auto local_version = retrieve_versions();
-        if (local_version["VAST"] != remote_version["VAST"])
-          VAST_WARN("client version {} does not match remote version {}; "
-                    "this may cause unexpected behavior",
-                    local_version["VAST"], remote_version["VAST"]);
-        else
-          VAST_DEBUG("client verified that local VAST version matches remote "
-                     "VAST version {}",
-                     local_version["VAST"]);
-        if (local_version["plugins"] != remote_version["plugins"])
-          VAST_WARN("client plugins {} do not match remote plugins {}; "
-                    "this may cause unexpected behavior",
-                    local_version["plugins"], remote_version["plugins"]);
-        else
-          VAST_DEBUG("client verified that local VAST plugins match remote "
-                     "VAST plugins {}",
-                     local_version["plugins"]);
+        check_version(remote_version);
       },
       [&](caf::error& error) {
         result = caf::make_error(ec::version_error,
--- a/libvast/src/system/connector.cpp
+++ b/libvast/src/system/connector.cpp
@@ -1,0 +1,232 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "vast/system/connector.hpp"
+
+#include "vast/data.hpp"
+#include "vast/detail/weak_run_delayed.hpp"
+#include "vast/logger.hpp"
+#include "vast/system/connect_request.hpp"
+
+#include <caf/io/middleman.hpp>
+#include <caf/openssl/all.hpp>
+#include <fmt/format.h>
+
+#include <optional>
+
+namespace vast::system {
+
+namespace {
+
+bool is_recoverable_error_enum(caf::sec err_enum) {
+  switch (err_enum) {
+    case caf::sec::none:
+    case caf::sec::unexpected_message:
+    case caf::sec::unexpected_response:
+    case caf::sec::request_receiver_down:
+    case caf::sec::no_such_group_module:
+    case caf::sec::no_actor_published_at_port:
+    case caf::sec::unexpected_actor_messaging_interface:
+    case caf::sec::state_not_serializable:
+    case caf::sec::unsupported_sys_key:
+    case caf::sec::unsupported_sys_message:
+    case caf::sec::disconnect_during_handshake:
+    case caf::sec::cannot_forward_to_invalid_actor:
+    case caf::sec::no_route_to_receiving_node:
+    case caf::sec::failed_to_assign_scribe_from_handle:
+    case caf::sec::failed_to_assign_doorman_from_handle:
+    case caf::sec::cannot_close_invalid_port:
+    case caf::sec::cannot_connect_to_node:
+    case caf::sec::cannot_open_port:
+    case caf::sec::network_syscall_failed:
+    case caf::sec::invalid_argument:
+    case caf::sec::invalid_protocol_family:
+    case caf::sec::cannot_publish_invalid_actor:
+    case caf::sec::cannot_spawn_actor_from_arguments:
+    case caf::sec::end_of_stream:
+    case caf::sec::no_context:
+    case caf::sec::unknown_type:
+    case caf::sec::no_proxy_registry:
+    case caf::sec::runtime_error:
+    case caf::sec::remote_linking_failed:
+    case caf::sec::cannot_add_upstream:
+    case caf::sec::upstream_already_exists:
+    case caf::sec::invalid_upstream:
+    case caf::sec::cannot_add_downstream:
+    case caf::sec::downstream_already_exists:
+    case caf::sec::invalid_downstream:
+    case caf::sec::no_downstream_stages_defined:
+    case caf::sec::stream_init_failed:
+    case caf::sec::invalid_stream_state:
+    case caf::sec::unhandled_stream_error:
+    case caf::sec::bad_function_call:
+    case caf::sec::feature_disabled:
+    case caf::sec::cannot_open_file:
+    case caf::sec::socket_invalid:
+    case caf::sec::socket_disconnected:
+    case caf::sec::socket_operation_failed:
+    case caf::sec::unavailable_or_would_block:
+    case caf::sec::malformed_basp_message:
+    case caf::sec::serializing_basp_payload_failed:
+    case caf::sec::redundant_connection:
+    case caf::sec::remote_lookup_failed:
+    case caf::sec::no_tracing_context:
+    case caf::sec::all_requests_failed:
+    case caf::sec::field_invariant_check_failed:
+    case caf::sec::field_value_synchronization_failed:
+    case caf::sec::invalid_field_type:
+    case caf::sec::unsafe_type:
+    case caf::sec::save_callback_failed:
+    case caf::sec::load_callback_failed:
+    case caf::sec::conversion_failed:
+    case caf::sec::connection_closed:
+    case caf::sec::type_clash:
+    case caf::sec::unsupported_operation:
+    case caf::sec::no_such_key:
+    case caf::sec::broken_promise:
+    case caf::sec::connection_timeout:
+    case caf::sec::action_reschedule_failed:
+      return true;
+    case caf::sec::incompatible_versions:
+    case caf::sec::incompatible_application_ids:
+    case caf::sec::request_timeout:
+      return false;
+  }
+  return false;
+}
+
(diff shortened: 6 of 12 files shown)


libvast/include/vast/system/connector.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+struct connector_state {
+  /// The amount of connection attempts left before failing.
+  std::size_t remaining_connection_attempts{0u};
+  /// Delay between two connection attempts.
+  caf::timespan connection_retry_delay{0u};
+  /// The port of a remote node server.
+  class port port;
+  /// The hostname or IP address of a remote node server.
+  std::string host;
```
mavam, line 32: These are technically not needed, as you copy them into the actor via the function signature.

libvast/include/vast/system/connector.hpp
```diff
@@ -0,0 +1,55 @@
...
+  /// Middleman request timeout.
+  caf::timespan node_connection_timeout;
```
mavam, line 34: Why does this need to be part of the actor state?

libvast/src/system/connect_to_node.cpp
```diff
@@ -11,28 +11,41 @@
...
 
 namespace vast::system {
+namespace {
 
-caf::expected<node_actor>
-connect_to_node(caf::scoped_actor& self, const caf::settings& opts) {
-  // Fetch values from config.
-  auto id = get_or(opts, "vast.node-id", defaults::system::node_id.data());
-  auto timeout = node_connection_timeout(opts);
+void check_version(record& remote_version) {
```
mavam, line 28: Should this really take a *non-const* lvalue?

mavam: I would also argue that the function should return a `bool`.

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,168 @@
...
+    case caf::sec::save_callback_failed:
+    case caf::sec::load_callback_failed:
+    case caf::sec::conversion_failed:
+    case caf::sec::connection_closed:
+    case caf::sec::type_clash:
+    case caf::sec::unsupported_operation:
+    case caf::sec::no_such_key:
+    case caf::sec::broken_promise:
+    case caf::sec::connection_timeout:
+    case caf::sec::action_reschedule_failed:
```
mavam, line 92: Did anything speak against using `default:` here? Seems like you already select those that matter with `return true` below.

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,168 @@
...
+           max_connection_attempts](atom::connect) -> caf::result<node_actor> {
+            if (self->state.promise.pending())
+              return self->state.promise;
+            self->state.remaining_connection_attempts = max_connection_attempts;
+            self->state.promise = self->make_response_promise<node_actor>();
+            VAST_INFO("client connects to VAST node at {}", self->state.host);
+            self->send(self, atom::internal_v, atom::start_v);
+            return self->state.promise;
+          },
+          [self](atom::internal, atom::start) {
```
mavam, line 135: Do you really need this second handler? It looks like you can do everything within a single handler, with the following logic:

1. Try connecting immediately
2. On success, return the node actor
3. On failure, create a response promise
4. Send yourself the same message again.

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,168 @@
...
+                               ? self->system().openssl_manager().actor_handle()
+                               : self->system().middleman().actor_handle();
+            self
+              ->request(middleman, self->state.node_connection_timeout,
+                        caf::connect_atom_v, self->state.host,
+                        self->state.port.number())
+              .then(
+                [self](const caf::node_id& node_id, caf::strong_actor_ptr& node,
+                       const std::set<std::string>&) {
+                  VAST_DEBUG("client connected to VAST node: {} at {}:{}",
```
mavam, line 146:
```suggestion
                  VAST_VERBOSE("client connected to VAST node: {} at {}:{}",
```

mavam: If not even INFO

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,168 @@
...
+    return false;
+  const auto err_code = std::underlying_type_t<caf::sec>{err.code()};
+  auto err_enum = caf::sec{caf::sec::none};
+  if (!caf::from_integer(err_code, err_enum)) {
+    VAST_WARN("unable to retrieve error code for a remote node connection "
+              "error:{}",
+              err);
+    return false;
+  }
+  return is_non_recoverable_error_enum(err_enum);
```
mavam, line 112: My hunch is that a simple switch here with a default case would suffice.

mavam (changes requested):
Everything is there, nice. I think we can gut out the code a bit to make it simpler.

libvast/src/system/connect_to_node.cpp
patszt: It has to because of the [] operator.
I was thinking about replacing it with .at(), but then i need to also make sure that the key is there and handle the situation where the key is not there. Since this is just a helper function in anonymous namespace i decided to keep it as it was previously.

Also returning boolean would require the caller to handle it. I didn't want to change the behaviour of connect_to_node in this PR.

libvast/src/system/connector.cpp
patszt: I like the idea of having compilation error when the new caf::sec enum values are introduced. Such compilation error will force us to check if the newly introduced error code should also be considered as a non recoverable one.

mavam: Good argument, I'm convinced.

libvast/src/system/connect_to_node.cpp
mavam: > Also returning boolean would require the caller to handle it. I didn't want to change the behaviour of connect_to_node in this PR.

But you're refactoring the API already, so why stop half way?

> It has to because of the [] operator.

Does anything speak against transforming that into a call to `find`, other than that it's more verbose?

The primary issue is that an API with a signature of `void check_version(record& remote_version)` if not very nice. There should be no need to take non-const reference to express the logging side-effect. The name of the function also suggests that it yields a response. If the version check fails, the you'd want to abort at the call site. (The warnings are no reason to abort, though.) If you wan to keep `operator[]`, you could take a full copy of the input `record`.

More generally, the version check should not be part of the `connector`. Once you have a node actor at the call site, then it's time to the version check there.

patszt: > The primary issue is that an API with a signature of void check_version(record& remote_version) if not very nice. There should be no need to take non-const reference to express the logging side-effect. The name of the function also suggests that it yields a response. If the version check fails, the you'd want to abort at the call site. (The warnings are no reason to abort, though.) If you wan to keep operator[], you could take a full copy of the input record.

It's not nice but since it's not public (implementation detail in cpp file) i found it to be okayish. I can go with const&, but then what should i do when there is no key available? The current master doesn't care about such cases. Should i return error from connect_to_node? 

> The name of the function also suggests that it yields a response. If the version check fails, the you'd want to abort at the call site. (The warnings are no reason to abort, though.)

If we don't have a good action in case this function fails then i don't see a point of having a return value that is never used. Please notice it's not a public function, it's purpose is to to group logic into smaller more readable block
Maybe i could change the name of the function to be print_versions_mismatch() ?

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,168 @@
...
+                      ec::system_error,
+                      fmt::format("failed to connect to VAST node at {}:{}: {}",
+                                  self->state.host, self->state.port.number(),
+                                  std::move(err))));
+                  } else {
+                    self->delayed_send(self, self->state.connection_retry_delay,
+                                       atom::internal_v, atom::start_v);
+                  }
+                });
+          }};
```
dominiklohmann, line 165:
```suggestion
          },};
```

And then reformat to get rid of a few levels of unnecessary indentation.

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,168 @@
...
+          caf::timespan connection_retry_delay, vast::port port,
+          std::string host, caf::timespan connection_timeout) {
+  self->state.connection_retry_delay = connection_retry_delay;
+  self->state.port = std::move(port);
+  self->state.host = std::move(host);
+  self->state.node_connection_timeout = connection_timeout;
+  return {[self,
+           max_connection_attempts](atom::connect) -> caf::result<node_actor> {
+            if (self->state.promise.pending())
+              return self->state.promise;
```
dominiklohmann, line 128: This introduces a potential indefinite hang. Isn't it just a logic error if you ask to connect twice?

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,168 @@
...
+            self->state.remaining_connection_attempts = max_connection_attempts;
+            self->state.promise = self->make_response_promise<node_actor>();
+            VAST_INFO("client connects to VAST node at {}", self->state.host);
+            self->send(self, atom::internal_v, atom::start_v);
+            return self->state.promise;
```
dominiklohmann, line 133: Should this just `self->delegate` instead of using a response promise, which makes the whole logic a lot easier to read?

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,168 @@
...
+                [self](caf::error& err) {
+                  if (is_non_recoverable_error(err)
+                      || --self->state.remaining_connection_attempts == 0) {
+                    self->state.promise.deliver(caf::make_error(
+                      ec::system_error,
+                      fmt::format("failed to connect to VAST node at {}:{}: {}",
+                                  self->state.host, self->state.port.number(),
+                                  std::move(err))));
+                  } else {
+                    self->delayed_send(self, self->state.connection_retry_delay,
```
dominiklohmann, line 161: `self->delayed_send(self, delay, ...)` introduces a hang for the duration of the delay, so this shouldn't be used here. If you wait for #2832 to be merged you can replace this with

```cpp
detail::weak_run_delayed(self, delay, [self] {
  self->send(self, atom::internal_v, atom::start_v);
});
```

Or just replace the internal handler altogether, I don't think it should be needed here at all if you model this with delegations rather a response promise.

patszt: Is there something i could delegate if i implement it in a single handler?
I want to return node_actor here, but the request to middleman has different API. I can't delegate this message AFAIK.
The success case must convert the output and the error case must trigger retry.

mavam: That's how I see it as well. The promise is the piece that precisely decouples the two operations. That's what we want. Otherwise retrying is not possible.

dominiklohmann: I'm not arguing against using response promises, I'm arguing against using response promises shared between multiple handlers, i.e., stored in the actor state. These are very hard to reason about correctly and as such make the code less maintainable. We've had a fair amount of issues in the past because of them.

Here's how you can do it without keeping a promise in the state:

```c++
[self](atom::connect) {
  setup()
  // delegate to internal handler
  return self->delegate(self, atom::internal_v, atom::connect_v);
},
[self](atom::internal, atom::connect) {
  auto rp = self->make_response_promise<node_actor>();
  connect().then(
    [self, rp](node_actor& node) mutable {
      rp.deliver(std::move(node));
    },
    [self, rp](caf::error& err) mutable {
      if (!should_retry(err)) {
        rp.deliver(std::move(err));
        return;
      }
      detail::run_weak_delayed(
        self, self->state.retry_delay,
        [self, rp]() mutable {
          rp.delegate(self, atom::internal_v, atom::connect_v);
        });
    },
  return rp;
},
```

This has the big advantage that you no longer need to make sure that the connect handler is only called once, and could actually make the connect handler take the endpoint as an argument now.

The `self` passed to `delegate` might need to be `static_cast<connector_actor>(self)` instead, not sure if that deficiency of CAF was fixed already.

libvast/src/system/connector.cpp
patszt: I have tried implementing it.
I found this version less readable, but maybe it's because i didnt implement it properly.

What i don't like is:
More complicated logic on entry to atom::connect because it is now called on each retry. And the actor state must be initialized only if no "run" is in progress.

I have to use caf::delayed_anon_send to ignore the returned response promise in atom::connect. This seems like a workaround.

I think the logic is simpler with two handlers. I don't think splitting logic into multiple handlers in this case is a bad thing.
We have one handler that setups the run, and the other as implementation detail for retry.

Similar to public and private methods with difference being that the "implementation" handler can be also called by mistake :D

I have pushed a commit to show my implementation of single handler

libvast/src/system/connector.cpp
patszt: Could You elaborate on the infinite hang here? I imagine that requests/responses can be delayed by some time when the system is under bigger load. I guess some messages can be prioritized over this one.

We can also configure the timeout here. Is it possible that the timeout is ignored and messages are further delayed?

I just thought that we can support concurrent requests very easily here by this simple if statement. I have no use case currently that will take advantage of this.

libvast/src/system/connector.cpp
patszt: If we decide to use two handlers then i think delegate should be used (commented in previous comment as well :P)

libvast/src/system/connector.cpp
dominiklohmann: When you create a response promise, it stores the current message ID in the response promise so it knows where to route the eventual reply. In this particular case you are returning a response promise that has a message ID that is not the current one—which means that whoever request that second `(atom::connect)`  will never get a reply. This can cause an infinite hang if the requester is a blocking actor and does not use a timeout.

The rule in general is to only return response promises in the same function that created them.

patszt: Thanks for clarification. Then i would need a collection of response promises if i were to accept multiple requests in parallel i guess.

libvast/src/system/connector.cpp
mavam: @dominiklohmann This looks like a lot of complexity. I wouldn't be able to write such code, nor reason about it. Why not try it with a single handler?

```cpp
[self](atom::connect) {
  // Initialize promise exactly once.
  auto return_promise = false;
  if (!self->state.promise) {
    self->state.promise = self->make_response_promise<node_actor>();
    return_promise = true;
  }
  // Try to connect.
  connect_via_middleman(...).then(
    [=](node_actor& node) {
      // We're done.
      self->state.promise.deliver(std::move(node));
    },
    [=](caf::error& err) mutable {
      // Bail out if we can't loop.
      if (!should_retry(err)) {
        rp.deliver(std::move(err));
        return;
      // Back off on failure. If this is an issue with hanging, maybe run_weak_delayed here?
      self->delayed_send(self, delay, atom::connect_v);
    },
  );
  // Return the promise to the user, exactly once.
  if (return_promise)
    return self->state.promise;
  else
    return {};
},
```

This is a simple loop, a single message handler, and easy to reason about.

libvast/src/system/connect_to_node.cpp
mavam: > It's not nice but since it's not public (implementation detail in cpp file) i found it to be okayish. I can go with const&, but then what should i do when there is no key available? 

If there's no key found, that would be a logic error, i.e., assertion, because it violates an invariant. But crashing is not nice, so I'd terminate cleanly. (This is the `false` case that I mention below.)

> The current master doesn't care about such cases. Printing a warning could be fine with the function returning void. Otherwise return false/error and abort the connection procedure?

Yes, return `true` if the check succeeds, and `false` if we cannot continue. 

> Maybe i could change the name of the function to be print_versions_mismatch() ?

No, it's actually a version check. We're just not acting responsibly on it (yet).

mavam:
@patszt can you please update the user-facing docs as well at the section where it would make sense?

libvast/src/system/connector.cpp
mavam: Maybe I've underestimated the complexity, but my naive thinking [in the comment down below](https://github.com/tenzir/vast/pull/2835#discussion_r1061841615) looks rather simple.

libvast/src/system/connector.cpp
mavam: @dominiklohmann I thought about the gymnastics with our `run_weak_delayed` helper a bit more. Shouldn't this something we upstream to CAF? That is, shouldn't `self->delayed_send(self, ...)` work exactly the way you implemented it? I find it a lot more readable when going with CAF messaging defaults, easier to teach, and easier to maintain.

dominiklohmann: > This is a simple loop and easy to reason about.

Your code snippet won't work, as it breaks the promise for concurrent requests. Storing the promise in the state is not a good idea as now you'd need to track the request alongside it. This is a category of errors that you can avoid if you keep the promise to the handler that created it, as I explained [here](https://github.com/tenzir/vast/pull/2835#discussion_r1061512572).

Regarding readability, I think everyone that works with CAF must understand messaging basics, including delegations, so this must not be an argument. We should aim to write maintainable and correct code that is easy to reason about: If a handler can be designed such that it does not need to modify state then we should aim for that. This is a local complexity increase that reduces the complexity of the system.

You can implement the snippet I wrote with a single handler as well, it all comes down to whether you have some setup in the state that you only want to do once. If you don't need that you also don't need a second handler.

I would actually go one further with the approach that I wrote and factor out the business logic from the requests, which also leaves you with just a single handler and even less things that need to be kept in the state:

```cpp
class connection_request {
  // Get the connection request, or an error.
  caf::expected<connection_request> retry(const caf::error& err) const;

  // ...

private:
  std::string endpoint_;
  size_t remaining_retries_;
  caf::timespan retry_delay_;
};

// ...

[self](atom::connect, connection_request req) {
  auto rp = self->make_response_promise<node_actor>();
  connect_via_middleman(...).then(
    [self, rp](const caf::actor& node) mutable {
      rp.deliver(caf::actor_cast<node_actor>(node));
    },
    [self, rp, req = std::move(req)](caf::error& err) mutable {
      auto retry_req = req.retry(err);
      if (!retry_req) {
        rp.deliver(std::move(retry_req.error()));
        return;
      }
      detail::weak_run_delayed(
        self, self->state.retry_delay,
        [self, rp, req = std::move(*retry_req)]() mutable {
          rp.delegate(self, atom::connect_v, std::move(req));
        });
    });
  return rp;
},
```

<hr />

> That is, shouldn't self->delayed_send(self, ...) work exactly the way you implemented it?

Not exactly. The following two do the same thing:

```cpp
// <= CAF 0.17
self->delayed_send(handle, delay, ...)

// >= CAF 0.18
self->run_delayed(delay, [self] { self->send(handle, ...); });
```

But they both keep `self` alive for the duration of the delay because they store a strong reference to `self` on the actor clock.

The `weak_run_delayed` function I wrote stores a weak actor pointer on the actor clock instead and only calls the delayed action if the actor is still alive when the delay expired.

NB: The advantage of the more generic `self->run_delayed(...)` introduced with CAF 0.18 is that it fills the gap of the nonexistent but desirable `self->delayed_delegate(...)`, `rp.delayed_delegate(...)`, and `self->delayed_request(...)`. I consider `self->delayed_send(...)` to be a function that should be deprecated to avoid API bloat.

mavam: Lot's of details, much appreciated for you taking the time to elaborate.

Overall, I'm in agreement with your solution. It's indeed nicer to basically move the state around, rather than to manually initialize and teardown (same issues as with non-RAII code).

That said, really the only *mutable* state is the counter in this case, everything else can simply be copied. I would say that just wrapping the counter doesn't require shipping a around a "fat" struct that also needs its own type ID. But that's just a minor implementation trade-off. The overall approach is nice.

I would still argue that CAF should provide that "weak" version of `run_delay` that you've implemented.

dominiklohmann: I opened a PR upstream for this, but I don't wanna go through the hoops of backporting it to 0.18.x and ask for a new release for such a minor function. We can live with our own version until then.

> I would say that just wrapping the counter doesn't require shipping a around a "fat" struct that also needs its own type ID. But that's just a minor implementation trade-off.

I think it's worth it for legibility at the call site alone. Sending a `size_t` is much less expressive than sending a `connection_request`. It's a write-once read-many-times effort. Plus I think we actually do want to include the endpoint and the initial retry delay, there's no reason to pass them to the actor on construction instead.

mavam: I have no strong opinion on the final part of implementation. I'm glad we're agreeing on the *design pattern*. That's what creates long-term sustainability. Whether it's a `size_t` or small struct is just the icing on the cake :wink:.

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,178 @@
...
+                  std::size_t remaining_connection_attempts) {
+  return remaining_connection_attempts > 0 && is_recoverable_error(err);
+}
+
+} // namespace
+
+connector_actor::behavior_type
+connector(connector_actor::pointer self, std::uint64_t max_connection_attempts,
+          caf::timespan connection_retry_delay, vast::port port,
+          std::string host, caf::timespan connection_timeout) {
```
dominiklohmann, line 126: Can we wrap these 5 arguments into a struct and pass them to the `atom::connect` handler instead? I don't think any of them are needed at spawn time of the actor, and if you pass the handler then the actor actually supports connecting to multiple different endpoints out of the box.

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,178 @@
...
+                     : self->system().middleman().actor_handle();
+  return {
+    [self, max_connection_attempts,
+     host](atom::connect) -> caf::result<node_actor> {
+      VAST_INFO("client connects to VAST node at {}", host);
+      return self->delegate(static_cast<connector_actor>(self),
+                            atom::internal_v, atom::connect_v,
+                            max_connection_attempts);
+    },
+    [self, port, host, middleman, connection_timeout, connection_retry_delay](
```
dominiklohmann, line 138: The middleman should be captured in the state instead. Storing strong actor handles in a behavior can introduce cyclic references. It didn't in this case, but let's stick to best practices with CAF.

web/docs/use/run/README.md
```diff
@@ -57,3 +57,11 @@ To select a specific VAST server to connect to,
 `--endpoint=host:port` on the command line, exporting the environment variable
 `VAST_ENDPOINT=host:port`, or setting the configuration option
 `vast.endpoint: host:port` in your `vast.yaml`.
+
+## Client connection failure
+
+In the event of a connection failure, the clients will attempt to connect up to
+100 times with a 3-second delay between each attempt.
```
dominiklohmann, line 64: This is still limited to `vast.connection-timeout`, right? So by default it will be at most four attempts: at 0s, 3s, 6s, and 9s, and then the fourth attempt will be cancelled after 1s.

Do the defaults make sense with this in mind?

libvast/include/vast/defaults.hpp
```diff
@@ -291,6 +291,14 @@ inline constexpr bool create_partition_index = true;
 /// Whether to spawn central components in separate threads.
 inline constexpr bool detach_components = true;
 
+/// Maximum retries attempts when connecting to a remote vast node.
+inline constexpr std::size_t node_connection_retries = 100u;
+
+/// Time to wait before trying to make another connection attempt to a remote
+/// vast node.
+inline constexpr caf::timespan node_connection_retry_delay
+  = std::chrono::seconds{3u};
```
dominiklohmann, line 300: Does it make sense to use an exponential backoff approach here?

E.g., first reconnect takes 20ms, second 40ms, third 80ms, and so on, but it's capped at a configurable maximum value?

We do the same for stalling inputs in readers and there it's been working really well, because for small system hickups you continue almost instantly but for bigger ones you cut the system some slack so you don't overload it.

Also cc @mavam—what do you think?

For the readers we do the following: 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1s, 1s, 1s, ...

libvast/include/vast/defaults.hpp
```diff
@@ -291,6 +291,14 @@ inline constexpr bool create_partition_index = true;
 /// Whether to spawn central components in separate threads.
 inline constexpr bool detach_components = true;
 
+/// Maximum retries attempts when connecting to a remote vast node.
+inline constexpr std::size_t node_connection_retries = 100u;
+
+/// Time to wait before trying to make another connection attempt to a remote
+/// vast node.
```
dominiklohmann, line 298:
```suggestion
/// VAST node.
```

libvast/include/vast/system/connector.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include <caf/timespan.hpp>
+#include <caf/typed_event_based_actor.hpp>
+#include <caf/typed_response_promise.hpp>
+
+#include <string>
+
+namespace vast::system {
+
+/// @brief Creates an actor that establishes the connection to a remote vast
+/// node
```
dominiklohmann, line 25:
```suggestion
/// @brief Creates an actor that establishes the connection to a remote VAST
/// node.
```

dominiklohmann (commented):
Some minor requests, and then I think we soon got it. 🚀

web/docs/use/run/README.md
mavam: @patszt and I discussed that we want the initial time period to cover 5mins / 300secs. This is where this 100 * 3 split comes from.

libvast/include/vast/defaults.hpp
mavam: I don't think so. You'd use exponential backoff mostly when you have multiple parties competing for a resource on a shared medium. Here, we can model this with independent failure event, so a uniform distribution makes the most sense.

I could see—as a follow-up in the future—a dual mode where you try every 1s for 10 times, and then switch to a lower polling frequency. But let's add that only when we need to. Sometimes firewalls are also in the way and temporarily block aggressive connection attempts.

web/docs/use/run/README.md
dominiklohmann: How is this supposed to interact with `vast.connection-timeout`? For example, we advise users to set `VAST_CONNECTION_TIMEOUT=0s` when they want clients to use an infinite timeout. Shouldn't we be re-using that configuration option such that users can still set the total?

patszt: vast.connection-timeout is used as a timeout between middleman and the connector actor. At least in a blocking connect_to_node.cpp.

This means that a 3 second delay is started only after connection failure. The duration of connection process is not taken into account.
Maybe the usage of vast.connection-timeout as middleman timeout request is bad in this case.

What do you think about having another parameter that models this timeout?
For now it could be a hardcoded value just like the retry delay and retry amount.

I would change vast.connection-timeout to vast.node-request-timeout
Use vast.connection-timeout as a middleman request timeout

changelog/unreleased/features/2835--retry-mechanism-for-vast-clients.md
```diff
@@ -0,0 +1,7 @@
+VAST clients now try to connect multiple times to a server. The option
+"vast.connection-timeout" dictates how long the connection process lasts. Value
+0 is used as infinite timeout. The option "vast.connection-retry-delay" is used
+as a delay between two connection attempts. The vast.connection-retry-delay can
+also be specified from CLI by using --connection-retry option. The default
+value of vast.connection-timeout is now 5 minutes and the value of
+vast.connection-retry-delay is 3 seconds.
```
dominiklohmann, line 7: I slightly rewrote it to me more along the usual style:

> We changed VAST client processes to attempt connecting to a VAST server multiple times until the configured connection timeout (`vast.connection-timeout`, defaults to 5 minutes) runs out. A fixed delay between connection attempts (`vast.connection-retry-delay`, defaults to 3 seconds) ensures that clients to not stress the server too much. Set the connection timeout to zero to let VAST client attempt connecting indefinitely, and the delay to zero to disable the retry mechanism.

libvast/src/system/application.cpp
```diff
@@ -426,7 +426,10 @@ auto make_root_command(std::string_view path) {
         .add<std::string>("store-backend", "store plugin to use for imported "
                                            "data")
         .add<std::string>("connection-timeout", "the timeout for connecting to "
-                                                "a VAST server (default: 10s)");
+                                                "a VAST server (default: 10s)")
```
dominiklohmann, line 429:
```suggestion
                                                "a VAST server (default: 5m)")
```

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,256 @@
...
+bool should_retry(const caf::error& err,
+                  const std::optional<caf::timespan>& remaining_time,
+                  caf::timespan delay) {
+  return remaining_time && *remaining_time > delay && is_recoverable_error(err);
+}
+
+std::string format_time(caf::timespan timespan) {
+  if (caf::is_infinite(timespan))
+    return "infinite";
+  std::vector<std::string> measurements;
```
dominiklohmann, line 136: We have a similar formatter for `duration` (which is an alias for `caf::timespan`) when it is wrapped inside a `data`, so this function can also be written like this:

```cpp
std::string format_time(caf::timespan timespan) {
   if (caf::is_infinite(timespan))
     return "infinite";
  return fmt::to_string(data{timespan});
}
```

Definitely :+1: for making sure that the timespans are user-readable in log messages, that's really nice to see.

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,256 @@
...
+    timespan -= std::chrono::duration_cast<caf::timespan>(ms);
+  }
+
+  return fmt::format("{}", fmt::join(measurements, " "));
+}
+
+void log_connection_failed(caf::timespan remaining_time,
+                           caf::timespan retry_delay) {
+  VAST_INFO("remote node connection failed. Retrying to "
+            "connect in {} (remaining time: {})",
```
dominiklohmann, line 164:
```suggestion
  VAST_INFO("remote node connection failed; attempting to "
            "reconnect in {} (remaining time: {})",
```

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,256 @@
...
+                                           *self));
+      auto rp = self->make_response_promise<node_actor>();
+      self
+        ->request(self->state.middleman, *remaining_time, caf::connect_atom_v,
+                  request.host, request.port)
+        .then(
+          [rp, request](const caf::node_id& node_id,
+                        caf::strong_actor_ptr& node,
+                        const std::set<std::string>&) mutable {
+            VAST_INFO("client connected to VAST node: {} at {}:{}",
```
dominiklohmann, line 188:
```suggestion
            VAST_INFO("client connected to VAST node at {}:{}",
```

I think the `node_id` is too much of an internal to be shown in a log message on INFO level. Maybe you can add a separate DEBUG message that shows it.

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,256 @@
...
+                      to_string(node_id), request.host, request.port);
```
dominiklohmann, line 189: Generally you can avoid the `to_string(...)` on the `node_id` here by adding a specialization for `use_deep_to_string_formatter<caf::node_id>` in `logger_formatters.hpp`.

libvast/src/system/connector.cpp
```diff
@@ -0,0 +1,256 @@
...
+                                           *self));
+      auto rp = self->make_response_promise<node_actor>();
+      self
+        ->request(self->state.middleman, *remaining_time, caf::connect_atom_v,
+                  request.host, request.port)
+        .then(
+          [rp, req = request](const caf::node_id& node_id,
+                              caf::strong_actor_ptr& node,
+                              const std::set<std::string>&) mutable {
+            VAST_INFO("client connected to VAST node: {} at {}:{}",
```
dominiklohmann, line 232: Same as above.

dominiklohmann (approved):
This works really nicely in practice. Just some really minor stuff remaining. Great work, Patryk! 🙏

libvast/src/system/connector.cpp
patszt: Thanks for the tip

libvast/src/system/connector.cpp
patszt: Seems that real_printer provide wrong output. For example 1.9998 will be printer as 1.1
It multiplies the 0.998 * 100 (MaxDigits * 10) resulting in 99.8. Then it uses std::round which results in 100. The 100 is then translated to 0.1 part. Should i try to fix it as a part of this PR? I have found a nice C++ 17 function that can simplify the code here (std::to_chars).

dominiklohmann: I don't think that it should go into this, but please create an issue so we can track this. I don't think `std::to_chars` is available on all platforms we support.

PR #2836, opened 2023-01-04
Remove fabric related code from Python bindings

rdettai:
This PR cleans up legacy code and fixes the logger:
- remove argparse wich was not appropriate as this is a library
- set `Logger.propagate=False` to avoid inheriting configurations from root logger

## TODO

- [x] Better document the asyncio utils
- [x] Add some docs about the configs


python/vast/utils/asyncio.py
```diff
@@ -3,7 +3,7 @@
 
 import vast.utils.logging
 
-logger = vast.utils.logging.get()
+logger = vast.utils.logging.get(__name__)
```
mavam, line 6: Have you checked that this is basically `vast`?

python/vast/vast/vast.py
```diff
@@ -13,7 +13,7 @@
 
 from .cli import CLI
 
-logger = vast.utils.logging.get("vast.vast")
+logger = vast.utils.logging.get(__name__)
```
mavam, line 16: Same as above :-)

mavam (approved):
Glad to see this go (for now).

You may want to look at the CI and packaging scripts that still have code that installs all extras via poetry.

Note to self: I'll still document the asyncio utils.

python/vast/vast/vast.py
rdettai: `__name__` is vast.vast.vast

python/vast/utils/config.py
```diff
@@ -1,29 +1,19 @@
...
     settings = Dynaconf(
-        settings_files=config_files,
+        settings_files=CONFIG_FILES,
         load_dotenv=True,
         envvar_prefix="VAST",
     )
     settings.validators.register(
-        Validator("fabric.logging.console_verbosity", default="debug"),
-        Validator("fabric.logging.file_verbosity", default="quiet"),
-        Validator("fabric.logging.filename", default="vast.log"),
```
rdettai, line 16: @mavam I am wondering about the config namespace we should use here. 
- Do we want the config to be the same as for VAST? That would mean that setting VAST_CONSOLE_VERBOSITY=debug sets the debug logging for both the bindings and the binary.
- [VAST config](https://github.com/tenzir/vast/blob/master/vast.yaml.example#L68) uses `-` and we use `_` here, should we unify this?

python/vast/vast/vast.py
mavam: Ooof, I don't think that's something the user would like.

python/vast/utils/asyncio.py
mavam: That looks reasonable, from a user's perspective.

python/vast/vast/vast.py
mavam: No, I don't know anything better. I'm just surprised that it's repeated *three times*.

python/vast/utils/config.py
mavam: Yes, let's unify where we can.

PR #2838, opened 2023-01-06
Add TUI

mavam:
This PR adds an experimental TUI for VAST.

Screenshot:
image


mavam:
@dominiklohmann what do you think about introducing some this functionality (data rendering) as part of a pipeline sink operator? Basically sift through a pipeline result interactively. The scope would be much reduced this way. And it's highly useful for CLI demos in a light-weight fashion.

dominiklohmann:
I'd be very happy about that because it'd be less effort to maintain than this would be in its current state.

mavam:
Are there any docs on how to write an operator?

I understand that I could just look at any other one, but it seems such an FAQ-like thing that it deserves a page here: https://vast.io/docs/develop/guides

dominiklohmann:
You'll have to wait until the new API is ready, and yes, that will come with a guide for writing operators. You cannot use the current one as this operator would have the output element type void.

mavam:
Closing this for now. It will come back in two phases:
1. As sink pipeline operator, to get a small-scale version into master
2. As `vast tui` command plugin and terminal-based node frontend

PR #2839, opened 2023-01-06
Fix multiple carousel bugs

dit7ya:
These are the bugs fixed.

 -    Its z-index is higher than the top bar's z-index
 -    Blog post images don't show
 -    It shows multiple "latest releases"
 -   On Safari it is a lot taller than it's likely supposed to be (also on mobile Safari)

Make sure to use the following to test this.

```fish
cd web
yarn install
yarn start
```

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -28,8 +28,8 @@
     "raw-loader": "^4.0.2",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "react-markdown": "^8.0.4",
     "redocusaurus": "^1.3.0",
+    "remark": "^14.0.2",
     "sass": "^1.52.1",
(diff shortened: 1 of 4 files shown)


mavam (commented):
I'm not sure what I'm doing wrong, but I can't get the figure to show:

image

I've used `yarn start` with the Docker-based build env. I also removed all draft blogs to make sure I get the same setup as the live version.

At least I don't encounter any height discrepancies in Safari.

dit7ya:
@mavam Are you sure you are running the latest commit from this branch? There should be only 4 slides in the carousel in the latest commit ...

mavam (approved):
My bad. I was still on the old branch. 🙈 

Looks good to me now!

PR #2842, opened 2023-01-09
Remove extra string quotes from POST parameters

lava:
As an artifact of using the `simdjson::minify()` function, string arguments passed via JSON body of a POST request would gain an additional pair of surrounding string quotes, turning queries like `ip in 0.0.0.0/32` internally into queries like `:string == "ip in 0.0.0.0/32"`.


dominiklohmann (approved):
Thanks for providing a test for this as well. 🙏

PR #2846, opened 2023-01-09
Devirtualize the table slice builder

dominiklohmann:
It's been almost a year since we've had more than one implementation of the table slice builder. We cannot really justify having every call to `table_slice_builder::add(...)` for every individual cell of data added to VAST go through virtual dispatch. This simply removes the pure virtual base class, renames `arrow_table_slice_builder` to `table_slice_builder`, and removes the `factory<table_slice_builder>` specialization.

In very much non-scientific tests this has increased the JSON reader throughput by 2-3% on my local machine. I didn't bother to benchmark any further, as we have quite a few performance-related changes for the next release that we can benchmark togethet to save on time. Ultimately, that's gonna be the numbers that's actually interesting to users.


libvast/include/vast/table_slice_builder.hpp
```diff
@@ -3,52 +3,42 @@
...
-  /// Forbid copy-assignment.
-  table_slice_builder& operator=(const table_slice_builder&) = delete;
-
-  /// Forbid move-construction.
-  table_slice_builder(table_slice_builder&&) = delete;
+  /// Constructs an Arrow table slice builder instance.
+  /// @param layout The layout of the slice.
+  /// @param initial_buffer_size The buffer size the builder starts with.
+  /// @returns A table_slice_builder instance.
+  explicit table_slice_builder(type layout, size_t initial_buffer_size
```
mavam, line 37:
```suggestion
  /// @param schema The schema of the slice.
  /// @param initial_buffer_size The buffer size the builder starts with.
  /// @returns A table_slice_builder instance.
  explicit table_slice_builder(type schema, size_t initial_buffer_size
```

libvast/src/format/single_layout_reader.cpp
```diff
@@ -42,8 +41,7 @@ caf::error single_layout_reader::finish(consumer& f, caf::error result) {
 }
 
 bool single_layout_reader::reset_builder(type layout) {
-  VAST_TRACE_SCOPE("{} {}", VAST_ARG(table_slice_type_), VAST_ARG(layout));
-  builder_ = factory<table_slice_builder>::make(table_slice_type_, layout);
+  builder_ = std::make_shared<table_slice_builder>(std::move(layout));
```
mavam, line 44: `s/layout/schema/` would also help here, might be worthwhile to do that globally where possible.

libvast/src/table_slice_builder.cpp
```diff
@@ -33,6 +52,136 @@ table_slice_builder::~table_slice_builder() noexcept {
...
+  auto fbs_ipc_buffer = flatbuffers::Offset<flatbuffers::Vector<uint8_t>>{};
+  if (serialize == table_slice::serialize::yes) {
+    auto ipc_ostream = arrow::io::BufferOutputStream::Create().ValueOrDie();
+    auto opts = arrow::ipc::IpcWriteOptions::Defaults();
+    opts.codec
+      = arrow::util::Codec::Create(
+          arrow::Compression::ZSTD,
+          arrow::util::Codec::DefaultCompressionLevel(arrow::Compression::ZSTD)
+            .ValueOrDie())
+          .ValueOrDie();
```
mavam, line 87: Maybe add a comment to make this configurable.

mavam (approved):
High-level review only. It looks like a few functions have new arguments. I didn't double-check the migrated business logic.

There might be an opportunity to switch the language from "layout" to "schema".

libvast/src/format/single_layout_reader.cpp
dominiklohmann: I'll try to make some adjustments because this has been bothering me as well, but please note that most of this PR was created semi-automatically, so this will probably be more effort than this PR was so far.

libvast/src/table_slice_builder.cpp
dominiklohmann: I think we can actually remove it. This was only necessary for the segment store, Feather and Parquet already have compression configurable on their own.

PR #2847, opened 2023-01-10
Fix table slice state on non unique chunk when setting import time

patszt:
Fix of the latest ASAN issues on the CI. I have retried unit tests jobs multiple times and the issue doesn't seem to reproduce
Thanks to @dominiklohmann for figuring this out.


tobim (approved):
I say this can go in without a changelog entry, since we've never seen it outside of CI runs.

PR #2848, opened 2023-01-10
Fix missing dense indexes and blocking import hangs

tobim:
Table columns for which no dense indexes are created cause that the last columns in the same table would also not be indexed.
For example for a schema with fields: `<foo, bar #skip, baz, buz>` only `foo` and `baz` would have a dense index.
If that resulted in no dense index creation at all the `import --blocking` mechanism also failed, causing the import process to hang forever.

The solution is two fold:
For the index creation, we now avoid the stage filter and `table_slice_column` altogether and let the indexer retrieve the correct column from the slice. This should make the code easier to understand. Together with that we don't bypass the column counter increment for skipped fields any more, so indexers are spawned with the correct column index.
For the blocking import we now wait for batches to be handled by the store instead of any of the connected indexers.


dominiklohmann:
What bug does this fix and why is this the correct solution?

libvast/include/vast/system/index.hpp
```diff
@@ -89,17 +89,6 @@ struct active_partition_info {
   /// The slot ID that identifies the partition in the stream.
   caf::stream_slot stream_slot = {};
 
-  /// The store actor that holds the segments for this partition.
-  // NOTE: Logically this should belong inside the active partition, but the way
-  // the CAF streaming api works makes it really annoying to have the partition
-  // stream both whole table_slices to the store and table_slice_columns to
-  // the indexers. So barring a major refactoring, we just have the partition
-  // do the streaming.
-  store_builder_actor store = {};
```
dominiklohmann, line 98: I am so happy to finally see this moved.

libvast/src/system/active_partition.cpp
```diff
@@ -329,7 +323,7 @@ active_partition_actor::behavior_type active_partition(
   active_partition_actor::stateful_pointer<active_partition_state> self,
   uuid id, accountant_actor accountant, filesystem_actor filesystem,
   caf::settings index_opts, const index_config& synopsis_opts,
-  store_actor store, std::string store_id, chunk_ptr header,
+  const store_actor_plugin* store_plugin,
```
dominiklohmann, line 326: I think in terms of API it'd be nicer to just pass in the name of the store plugin here rather than doing the lookup in the index actor, as then the index actor doesn't need to know about the store at all.

libvast/test/system/active_partition.cpp
```diff
@@ -111,18 +126,18 @@ TEST(No dense indexes serialization when create dense index in config is false)
...
-  const std::string input_store_id = "some-id";
-  auto header_data = 3523532ull;
-  auto partition_header
-    = vast::chunk::make(&header_data, sizeof(header_data), []() noexcept {});
-  auto sut
-    = sys.spawn(vast::system::active_partition, partition_id,
-                vast::system::accountant_actor{}, filesystem, caf::settings{},
-                index_config_, vast::system::store_actor{}, input_store_id,
-                partition_header, std::make_shared<vast::taxonomies>());
+  // FIXME: We should implement a mock store and use that for this test.
```
dominiklohmann, line 129: Wouldn't it already suffice to use the default store with the memory filesystem here?

Regardless, do you still want to fix this within this PR? Otherwise `s/FIXME/TODO`.

libvast/src/system/active_partition.cpp
```diff
@@ -341,64 +335,63 @@ active_partition_actor::behavior_type active_partition(
...
+        ids.append_bits(true, last - first);
+        self->state.data.events += x.rows();
+        self->state.data.synopsis.unshared().add(
+          x, self->state.partition_capacity, self->state.synopsis_index_config);
+        size_t column_idx = -1;
+        for (const auto& [field, offset] :
+             caf::get<record_type>(schema).leaves()) {
+          column_idx++;
+          // TODO: The qualified record field is a leftover from heterogeneous
+          // partitions, the indexers can be indexed by the column instead.
```
dominiklohmann, line 387: I think we should index them by offset instead of flat index, as we may want to create an index for a field in a record in a list, for example.

dominiklohmann (approved):
This is fantastic. As we discussed earlier, I pushed a fix for the broken unit test and the merge conflicts with master directly to the branch. There's only minor feedback on the PR itself.

I ran the integration tests that failed frequently 100 times each with master and with this PR in a release build:

```bash
sh build/vast/integration-Release.sh -t "(Arrow Full Data Model|Heterogeneous JSONL Import)"
```

For master, I failed 4/100 times.

With this change, it did not fail a single time.

libvast/src/system/active_partition.cpp
tobim: I'd agree if we wanted to stick with the index as it is, but we're going to replace it with a "partition builder" in the medium term, and that can definitely do the plugin lookup on behalf of the partitions it creates.

dominiklohmann: I still think that'd be cleaner code, but I don't want to argue about this—as you said, this is bound to be replaced soon regardless.

PR #2850, opened 2023-01-10
Move eager expression parsing from spawn arguments to target components

Dakostu:
As a preparation for the new pipeline operator parsing, expression parsing needs to be moved into the target component creation workflow instead of before target component creation.


libvast/include/vast/system/exporter.hpp
```diff
@@ -84,7 +84,9 @@ struct exporter_state {
 /// @param options The query options.
 /// @param pipelines The applied pipelines.
 exporter_actor::behavior_type
-exporter(exporter_actor::stateful_pointer<exporter_state> self, expression expr,
-         query_options options, std::vector<pipeline>&& pipelines);
+exporter(exporter_actor::stateful_pointer<exporter_state> self,
+         std::optional<expression> expr,
+         std::vector<std::string> spawn_arguments, query_options options,
```
dominiklohmann, line 89: Passing in both doesn't make much sense to me. The conversion from `spawn_arguments` to `expression` should likely take place in the `spawn_exporter` and `spawn_counter` functions respectively, such that we only need to normalize and validate the query that we got passed in.

libvast/src/system/exporter.cpp
```diff
@@ -199,11 +200,21 @@ void handle_batch(exporter_actor::stateful_pointer<exporter_state> self,
...
+         std::vector<std::string> spawn_arguments, query_options options,
+         std::vector<pipeline>&& pipelines) {
+  // Parse given expression.
+  if (!expr) {
+    auto normalized_expr = system::get_expression(spawn_arguments);
+    if (!normalized_expr) {
+      self->quit(normalized_expr.error());
+    }
+    expr.emplace(*normalized_expr);
+  }
```
dominiklohmann, line 214: I believe this should only normalize and validate (see the above review comment).

After you quit you need to `return exporter_actor::behavior_type::make_empty_behavior()`, otherwise you immediately afterwards dereference an invalid expression.

I would also recommend enriching the error with some additional context, e.g.,

```c++
self->quit(caf::make_error(
  ec::format_error,
  fmt::format("exporter failed to normalize and validate expression: {}",
    normalized_expr.error())));
```

libvast/include/vast/system/spawn_arguments.hpp
```diff
@@ -67,7 +67,7 @@ caf::expected<expression>
 normalized_and_validated(std::vector<std::string>::const_iterator begin,
                          std::vector<std::string>::const_iterator end);
 
-caf::expected<expression> get_expression(const spawn_arguments& args);
+caf::expected<expression> get_expression(const std::vector<std::string>& args);
```
dominiklohmann, line 70: The spawn arguments still include an optional expression. That shouldn't exist anymore after this change.

dominiklohmann (changes requested):
Some initial feedback. Thanks for tackling this one! 🙏

libvast/include/vast/system/exporter.hpp
dominiklohmann: We're trying to get it out of `spawn_arguments`, which is populated in `spawn_command`. The more specific `spawn_{exporter,counter}` do argument checking/parsing and then spawn the actual component. To summarize:
- Use `spawn_{exporter,counter}` to parse the expression
- Validate the expression in the exporter/counter directly

This ensures that all requirements are directly encoded in the types of the function arguments.

Dakostu: Okay, I got it now. I addressed all your concerns in https://github.com/tenzir/vast/pull/2850/commits/3f38bb3f70f35d2f442084f444575aee3f903942.

libvast/src/system/pivoter.cpp
```diff
@@ -78,6 +78,15 @@ pivoter_state::pivoter_state(caf::event_based_actor*) {
 
 caf::behavior pivoter(caf::stateful_actor<pivoter_state>* self, node_actor node,
                       std::string target, expression expr) {
+  auto normalized_expr = normalize_and_validate(std::move(expr));
+  if (!normalized_expr) {
+    self->quit(caf::make_error(ec::format_error,
+                               fmt::format("pivoter failed to normalize and "
```
dominiklohmann, line 84: Generally you can replace the actor name with `{}` and then provide `*self` to the formatter to avoid the duplication. This also applies to the other two instances of this.

dominiklohmann (approved):
This is great, thank you! 🙏

Already looking forward to the next PR that is about parsing an expression and a pipeline for the export command!

libvast/src/system/pivoter.cpp
Dakostu: That's handy. Implemented that in https://github.com/tenzir/vast/pull/2850/commits/acd9d338f87347b0acbaf74216b25a8a3d070460

PR #2851, opened 2023-01-10
Flush every batch when exporting arrow

tobim:
This is necessary for continuous export with Apache Arrow as output format. The `StdoutStream` uses `cout` internally, which buffers written data by default. This buffering means that the most recent batch does not arrive in its entirety at the destination until more results are produced.


dominiklohmann (approved):
Fine by me. We can fix this properly with the `to [carrier] arrow` operator in the medium-term.

PR #2854, opened 2023-01-12
TheHive: rename deprecated configuration key

1wilkens:
As stated in the migration docs from 4.x to TheHive 5.y [link], the configuration key to enable modules has been changed. This commit silences a warning about usage of the deprecated key.

[link]: https://docs.strangebee.com/thehive/setup/installation/upgrade-from-4.x/#prepare-for-the-new-installation

### DIFF
--- a/docker/thehive/thehive/application.conf.template
+++ b/docker/thehive/thehive/application.conf.template
@@ -16,7 +16,7 @@ db.janusgraph {
   }  
 }
 
-play.modules.enabled += org.thp.thehive.connector.cortex.CortexModule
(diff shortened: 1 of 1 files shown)


dominiklohmann (approved):
Thanks for upstreaming the fix! The docs are pretty clear on this one. 🙏

dominiklohmann:
Note that the CI will partially fail for this one as we currently have an open bug in it that prohibits it from running correctly for PRs from forks, so I'll manually override it in this particular case.

PR #2855, opened 2023-01-12
Run docker, docker-compose, vast-io and python job when /python changes

patszt:
The python bindings are used in docker, docker-compose and vast-io. It is safer to run these jobs when there are changes in the /python directory.


dominiklohmann (approved):
Looks correct to me, @rdettai should be able to quickly verify this working with his next PR.

PR #2857, opened 2023-01-12
Fix bug causing Arrow imports to skip buffered events

dominiklohmann:
This is a two-fold commit: First, the Arrow reader threw away stream readers when the number of requested events were exceeded. In case the reader had more batches buffered this led to data being thrown away. This did not completely fix the race in the integration tests, because there appears to be a race condition when running the Arrow import with a local server process by setting the `vast.node` option.

The bug became much rarer after the first change, so in order to make the tests more reliable I also changed the tests not to use a local server process. Afterwards, I was unable to reproduce the issue in over a thousand repetitions of the Arrow Import test in release mode, when without any of the two changes it occurred every ~10th time and with the first fix every ~50th time on my local machine with a release build.

No changelog entry required as this feature was introduced after the last release.


tobim (approved):
Since I can't reproduce the test failure locally I only comment on the logic changes:
Keeping the internal batch reader between invocations of `read_impl` definitely fixes _a_ bug. I expect that this will make the integration tests more reliable.

PR #2858, opened 2023-01-12
Skip record batches with incompatible metadata on import

dominiklohmann:
This causes the Arrow import to skip Arrow data that was not produced by VAST, as that cannot yet be handled by VAST. A proper fix for this is in sight, but for now this will have to do. It's obviously better to reject gracefully than to crash here.

No changelog entry as the Arrow import was only introduced after the latest release.

### DIFF
--- a/libvast/src/format/arrow.cpp
+++ b/libvast/src/format/arrow.cpp
@@ -158,6 +158,19 @@ reader::read_impl(size_t max_events, size_t max_slice_size, consumer& f) {
                                        close_result.ToString()));
       continue;
     }
+    // Skip record batches with incompatible metadata. This check is not
+    // complete because it doesn't visit every (potentially nested) array in the
+    // batch, but it's a good enough heuristic to prevent users from running
+    // into crashes. A proper fix for this requires casting to VAST-compatible
+    // record batch.
+    if (read_result->batch->schema()->metadata()->FindKey("VAST:name:0")
+        == -1) {
+      VAST_ERROR("{} skips record batch with {} rows: metadata is "
+                 "incomaptible with VAST",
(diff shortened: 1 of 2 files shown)


libvast/src/format/arrow.cpp
```diff
@@ -158,6 +158,19 @@ reader::read_impl(size_t max_events, size_t max_slice_size, consumer& f) {
...
       continue;
     }
+    // Skip record batches with incompatible metadata. This check is not
+    // complete because it doesn't visit every (potentially nested) array in the
+    // batch, but it's a good enough heuristic to prevent users from running
+    // into crashes. A proper fix for this requires casting to VAST-compatible
+    // record batch.
+    if (read_result->batch->schema()->metadata()->FindKey("VAST:name:0")
+        == -1) {
+      VAST_ERROR("{} skips record batch with {} rows: metadata is "
```
mavam, line 168:
```suggestion
      VAST_WARN("{} skips record batch with {} rows: metadata is "
```

The expected semantics for ERROR are that we fail.

web/docs/understand/formats/arrow.md
```diff
@@ -20,12 +20,25 @@ representations.
 
 ## Input
 
-The `import arrow` command imports [`Arrow IPC`][arrow-ipc] data. Since this
-format carries the schema alongside the data, `import arrow` is self-contained
-and does not require an additional schema.
+The `import arrow` command imports [`Arrow IPC`][arrow-ipc] data. This allows
+for transferring data between VAST nodes flexibly:
```
mavam, line 24:
```suggestion
for efficiently transferring data between VAST nodes:
```

mavam (approved):

PR #2860, opened 2023-01-13
Fix some shutdown issues in the web plugin

lava:
This PR fixes 2/3 shutdown issues I found in the web plugin; in particular the web plugin now shuts down correctly when run as a separate command when receiving a signal and when detecting a node shutdown.

Note that it does not address the situation where `web server` is passed as a start command to `vast start`.

### DIFF
--- a/plugins/web/src/server_command.cpp
+++ b/plugins/web/src/server_command.cpp
@@ -12,11 +12,13 @@
 #include "web/configuration.hpp"
 #include "web/mime.hpp"
 #include "web/restinio_response.hpp"
+#include "web/restinio_server.hpp"
 
 #include <vast/concept/convertible/data.hpp>
 #include <vast/concept/parseable/to.hpp>
 #include <vast/concept/parseable/vast/data.hpp>
 #include <vast/concept/parseable/vast/expression.hpp>
+#include <vast/detail/exception_ptr.hpp>
 #include <vast/format/json.hpp>
 #include <vast/logger.hpp>
 #include <vast/plugin.hpp>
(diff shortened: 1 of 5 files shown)


libvast/include/vast/detail/exception_ptr.hpp
```diff
@@ -0,0 +1,32 @@
...
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <exception>
+#include <string>
+
+namespace vast::detail {
+
+// Taken from https://stackoverflow.com/a/37222762/92560
+// with minor adjustments and style changes.
+inline std::string what(const std::exception_ptr& eptr) {
```
tobim, line 16: Where is this used?

lava: Ah, right, it was required by an earlier version that used an async asio call to start the server; I'll remove it again from the PR.

plugins/web/src/server_command.cpp
```diff
@@ -391,56 +390,60 @@ auto server_command(const vast::invocation& inv, caf::actor_system& system)
...
+  // Post initial action to asio event loop. Note that the action
+  // must have been posted *before* calling `io_context.run()`.
+  asio::post(io_context, [&] {
+    std::visit(detail::overload{[](auto& server) {
+                 server->open_sync();
+               }},
+               server);
+  });
+  // Launch the thread on which the server will work.
+  std::thread server_thread{[&] {
```
lava, line 405: A word of warning; I had a bit of a headache concerning the lifetimes of variables when capturing local variables by reference here but ultimatately I thought that since we're blocking below until the thread is joined again this *should* be safe.

Dakostu (approved):
My original `SIGINT` sending issue has been fixed.

Dakostu:
@lava 
Does this warrant a changelog entry?

PR #2861, opened 2023-01-13
Fix linux PID file check on startup

lava:
This check seems to have at some point been converted from `sprintf` to `fmt::format` without adjusting the syntax. (although its a bit surprising that `fmt` doesn't throw a compile error here)


changelog/unreleased/bug-fixes/2861--pid-file-startup.md
```diff
@@ -0,0 +1 @@
+VAST no longer ignore existing PID lock files on linux.
```
mavam, line 1:
```suggestion
VAST no longer ignores existing PID lock files on Linux.
```

mavam (approved):
Wowza... I also would have expected a compile-time error here.

PR #2863, opened 2023-01-13
Ignore num-partitions in integration tests

tobim:
Only the number of events is really important, so we ignore partitions.
This should improve test reliability a bit more.

### DIFF
--- a/vast/integration/vast_integration_suite.yaml
+++ b/vast/integration/vast_integration_suite.yaml
@@ -237,7 +237,7 @@ tests:
         input: data/zeek/conn.log.gz
       - command: flush
       - command: status --detailed
-        transformation: jq '.catalog.schemas | del(."vast.metrics") | del(."zeek.conn"."import-time")'
+        transformation: jq '.catalog.schemas | with_entries(.value = .value."num-events")'
       - command: status --detailed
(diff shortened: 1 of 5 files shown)


vast/integration/vast_integration_suite.yaml
```diff
@@ -237,7 +237,7 @@ tests:
         input: data/zeek/conn.log.gz
       - command: flush
       - command: status --detailed
-        transformation: jq '.catalog.schemas | del(."vast.metrics") | del(."zeek.conn"."import-time")'
+        transformation: jq '.catalog.schemas | del(."vast.metrics") | del(."zeek.conn"."import-time") | del(."zeek.conn"."num-partitions")'
```
dominiklohmann, line 240: Can you apply the same change to the web plugin integration tests?

dominiklohmann (approved):
Good idea.

This is probably a better transformation:

```
❯ vast -qq status catalog | jq '.catalog.schemas | with_entries(.value = .value."num-events")'
{
  "suricata.alert": 15,
  "suricata.dns": 15,
  "suricata.fileinfo": 15,
  "suricata.flow": 15,
  "suricata.http": 15,
  "suricata.netflow": 15,
  "suricata.quic": 8,
  "suricata.stats": 15,
  "zeek.conn": 59234
}
```

PR #2864, opened 2023-01-13
Rename count, int, real, and addr to uint64, int64, double, and ip respectively

dominiklohmann:
This changes the basic type names of the `count`, `int`, `real`, and `addr` types to `uint64`, `int64`, `double`, and `ip`. It also changes the type extractors accordingly.

The change is written in a way that is backwards-compatible: the old type tokens are still recognized and parsed by VAST, and a warning is omitted once per deprecated type token on startup if it is parsed.

This PR does intentionally not change the identifier of the Arrow extension type for the `ip` type from `vast.address` to `vast.ip`, as that is a breaking change that while possible to implement requires a lot of effort. If we want to change that we can always do so in the future, but I intentionally spared myself from it for this PR.

Most of this renaming was created semi-automatically through the help of clangd and `:lua vim.lsp.buf.rename()`, and some clever git-grepping in combination with quickfix lists and `:cdo s/old/new/g`. I don't expect this to be thoroughly reviewed, as I don't think this is possible given the number of changes. Please do look at the documentation changes.


libvast/include/vast/arrow_extension_types.hpp
```diff
@@ -69,14 +69,19 @@ class enum_extension_type : public arrow::ExtensionType {
...
 /// Address representation as an Arrow extension type.
 /// Internal (physical) representation is a 16-byte fixed binary.
-class address_extension_type : public arrow::ExtensionType {
+class ip_extension_type : public arrow::ExtensionType {
 public:
+  // NOTE: The identifier for the extension type of VAST's ip type has not
+  // changed when the type was renamed from address to ip because that would be
+  // a breaking change. This is fixable by registering two separate extension
+  // types with the same functionality but different ids, but that's a lot of
+  // effort for something users don't usually see.
```
mavam, line 78: We're making breaking changes with 3.0, so isn't the time now to do this change?

libvast/include/vast/detail/string_literal.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+#pragma once
+
+#include <algorithm>
+#include <cstdint>
+#include <string_view>
+
+namespace vast::detail {
+
+/// String literal wrapper for NTTPs.
```
mavam, line 17: Googling doesn't reveal anything meaningful for me here. Can this be made less obscure? Maybe add a URL?

libvast/src/concept/parseable/vast/legacy_type.cpp
```diff
@@ -14,13 +14,20 @@
...
 #include "vast/concept/parseable/string/quoted_string.hpp"
 #include "vast/concept/parseable/vast/identifier.hpp"
+#include "vast/detail/string_literal.hpp"
 
 #include <caf/optional.hpp>
 
 namespace vast {
 
-template <class T>
+template <class T, detail::string_literal DeprecationNotice = "">
```
mavam, line 23: This looks like a `detail::`'ish thing potentially useful in other places as well.

libvast/src/evaluate.cpp
```diff
@@ -27,6 +27,12 @@ namespace vast {
 
 namespace {
 
+template <class LhsView, class RhsView>
+inline constexpr auto requires_stdcmp
+  = std::is_integral_v<LhsView> && std::is_integral_v<RhsView>
+    && !std::is_same_v<LhsView, RhsView>
+    && !detail::is_any_v<bool, LhsView, RhsView>;
+
```
mavam, line 35: Any reason not to use `std::cmp_*` unconditionally?

libvast/src/index/ip_index.cpp
```diff
@@ -128,25 +128,25 @@ flatbuffers::Offset<fbs::ValueIndex> address_index::pack_impl(
...
-  VAST_ASSERT(from_address);
-  if (from_address->byte_indexes()->size() != bytes_.size())
+caf::error ip_index::unpack_impl(const fbs::ValueIndex& from) {
+  const auto* from_ip = from.value_index_as_ip();
+  VAST_ASSERT(from_ip);
+  if (from_ip->byte_indexes()->size() != bytes_.size())
     return caf::make_error(ec::format_error,
                            fmt::format("unexpected number of byte indexes in "
-                                       "address index: expected {}, got {}",
+                                       "ip index: expected {}, got {}",
```
mavam, line 143:
```suggestion
                                       "IP index: expected {}, got {}",
```

python/vast/utils/arrow.py
```diff
@@ -35,12 +35,17 @@ def __arrow_ext_scalar_class__(self):
...
         return None if self.value is None else unpack_ip(self.value.as_py())
 
 
-class AddressType(pa.ExtensionType):
+class IPType(pa.ExtensionType):
+    # NOTE: The identifier for the extension type of VAST's ip type has not
+    # changed when the type was renamed from address to ip because that would be
+    # a breaking change. This is fixable by registering two separate extension
+    # types with the same functionality but different ids, but that's a lot of
+    # effort for something users don't usually see.
```
mavam, line 48: (Anchor comment for the existing discussion on this topic.)

web/blog/vast-v2.3/index.md
```diff
@@ -96,7 +96,7 @@ vast:
         partition-index: false
         # Don't create partition indexes for fields of type addr.
```
mavam, line 97:
```suggestion
        # Don't create partition indexes for IP address fields.
```

web/openapi/openapi.yaml
```diff
@@ -28,11 +28,11 @@ paths:
             default: A query matching every event.
           required: true
           description: The query expression to execute.
-          example: :addr in 10.42.0.0/16
+          example: :ip in 10.42.0.0/16
```
mavam, line 31: Finally this works the Benno always wanted to write it. 😉

mavam (approved):
Nice 🐒 work.

I think this PR is the right place to also add the corresponding blog post pieces that describe the change.

libvast/include/vast/arrow_extension_types.hpp
dominiklohmann: I wouldn't consider updating the internal representation of the extension type worth the requireed effort. If you think it is then please create do this in a follow-up PR. This will only ever be visible to devs.

libvast/src/concept/parseable/vast/legacy_type.cpp
dominiklohmann: Yes, but not for this PR.

libvast/include/vast/detail/string_literal.hpp
dominiklohmann: It's literally what the comment says: a class making string literals usable as non-type template parameters. A similar technique exists in {fmt}, just to list one example. It's a reallx common meta-progrmaming technique.

NB: I actually wrote this last year when I first dabbled with `co_result<T>`, and this is a carbon-copy from that PR.

libvast/src/evaluate.cpp
dominiklohmann: It only works for integral but non-bool types. With the removal of `vast.:integer` I got `-Wsign-compare` warnings from clang here, so I had to fix it given that our CI is using `-Werror`.

web/blog/vast-v2.3/index.md
dominiklohmann: I don't think we should update older blog posts.

libvast/include/vast/arrow_extension_types.hpp
dominiklohmann: I wrote a follow-up story for this.

libvast/include/vast/detail/string_literal.hpp
dominiklohmann: I expanded the comment above this class to make this a bit clearer.

python/vast/utils/arrow.py
dominiklohmann: As I said above, let's consider doing this in a follow-up story.

dominiklohmann:
> I think this PR is the right place to also add the corresponding blog post pieces that describe the change.

See https://github.com/tenzir/vast/pull/2865.

PR #2866, opened 2023-01-16
Add tests for the fanout counter

lava:
These are moved from their previous location in the closed-source compaction plugin.


dominiklohmann (approved):
I'm assuming that this is a carbon-copy (except for the comment at the very top)  of the already reviewed tests, so have some blind approval.

PR #2868, opened 2023-01-16
Run start commands asynchronously

lava:
Run start commands asynchronously
    
    We used to run start commands sequentially, but this has the
    disadvantage the if any start command blocks indefinitely
    (eg. `web server`) the subsequent commands are never executed
    and the start command itself never gets to the CTRL-C signal
    handling code below.

### DIFF
--- a/libvast/src/system/start_command.cpp
+++ b/libvast/src/system/start_command.cpp
@@ -45,11 +45,12 @@ using command_runner_actor = system::typed_actor_fwd<
 command_runner_actor::behavior_type
 command_runner(command_runner_actor::pointer self) {
   return {
-    [self](atom::run, vast::invocation& hook_invocation) -> caf::result<void> {
+    [self](atom::run, vast::invocation& invocation) -> caf::result<void> {
       auto [root, root_factory] = make_application("vast");
-      auto result = run(hook_invocation, self->home_system(), root_factory);
+      auto result = run(invocation, self->home_system(), root_factory);
       if (!result)
-        return result.error();
+        VAST_ERROR("failed to run start command {}: {}", invocation,
+                   result.error());
       return {};
     },
   };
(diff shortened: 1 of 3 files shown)


libvast/src/system/start_command.cpp
```diff
@@ -118,9 +143,9 @@ caf::message start_command(const invocation& inv, caf::actor_system& sys) {
         return caf::make_message(hook_invocation.error());
       detail::merge_settings(inv.options, hook_invocation->options,
                              policy::merge_lists::yes);
-      auto result = run(*hook_invocation, sys, root_factory);
-      if (!result)
-        return caf::make_message(result.error());
+      auto runner = self->spawn(command_runner);
+      command_runners.push_back(runner);
+      self->send(runner, atom::run_v, *hook_invocation);
```
dominiklohmann, line 148: Instead of maintaining the runners in a separate vector, would it make sense to use `request.then()` and to capture the runner in the response handling lambdas? That way you don't keep the actors alive for longer than you actually need to.

We should be using request here anyways for logging, I think, in case an error is returned.

dominiklohmann (approved):
Works for me locally, and the code looks good as well except for one minor suggestion. This also needs a bug fix changelog entry.

dominiklohmann (commented):
Just one additional question that just came to my mind.

libvast/src/system/start_command.cpp
lava: We can't use `then()` here because `self` is a blocking actor.

For now I've just logged the error in the handler itself, although I think the "proper" solution would be to introduce a `struct command_result { optional<caf::error> e; }` that we can receive in the `do_receive` loop below.

dominiklohmann: Fine by me.

PR #2869, opened 2023-01-16
Python library ergonomics improvements

tobim:
This changes the API to go through an object of the `VAST` type
instead of working directly on the type. This way it is possible
to specify certain option parameters like the connection endpoint
for a group of commands without requiring filesystem or process
level state.

This makes it also possible to run the integration tests of the
bindings without docker.

### DIFF
--- a/python/tests/test_vast.py
+++ b/python/tests/test_vast.py
@@ -17,6 +17,8 @@ if "VAST_PYTHON_INTEGRATION" not in os.environ:
     )
 
 TEST_DB_DIR = "/tmp/test-vast-db"
+if os.path.isdir(TEST_DB_DIR):
+    shutil.rmtree(TEST_DB_DIR)
 
 
 @pytest.fixture()
@@ -128,9 +130,7 @@ async def test_export_continuous_rows(endpoint):
     vast = VAST(endpoint)
 
     async def run_export():
-        result = vast.export(
-            '#type == "suricata.alert"', ExportMode.CONTINUOUS, limit=1
-        )
+        result = vast.export('#type == "suricata.alert"', ExportMode.CONTINUOUS)
         return await anext(to_json_rows(result))
 
     task = asyncio.create_task(run_export())
--- a/python/tests/test_vast.py
+++ b/python/tests/test_vast.py
@@ -19,30 +19,43 @@ if "VAST_PYTHON_INTEGRATION" not in os.environ:
 TEST_DB_DIR = "/tmp/test-vast-db"
 
 
-@pytest.fixture(autouse=True)
-async def vast_server():
+@pytest.fixture()
+async def endpoint():
     proc = await asyncio.create_subprocess_exec(
-        "vast", "-d", TEST_DB_DIR, "start", stderr=asyncio.subprocess.PIPE
+        "vast",
+        "-e",
+        ":0",
+        "-d",
+        TEST_DB_DIR,
+        "start",
+        "--print-endpoint",
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.PIPE,
     )
-    await asyncio.sleep(3)
-    yield
+    endpoint = await proc.stdout.readline()
+    endpoint = endpoint.decode("utf-8").strip()
+    logger.debug(f"{endpoint = }")
+    yield endpoint
     proc.terminate()
+    # TODO: kill if in case of timeout.
     await asyncio.wait_for(proc.wait(), 5)
     await asyncio.to_thread(shutil.rmtree, TEST_DB_DIR)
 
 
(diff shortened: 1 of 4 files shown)


python/vast/vast/vast.py
```diff
@@ -76,8 +76,9 @@ def _next_batch():
 class VAST:
     """An instance of a VAST node."""
 
-    def __init__(self):
-        pass
+    def __init__(self, endpoint="127.0.0.1:42000"):
```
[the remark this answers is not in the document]
rdettai: agreed, if we assume that this is the equivalent to setting `--endpoint`, it should fallback to the binary configuration.  Should this maybe be a config object instead of one argument by parameter? I guess that it would make sense to add the other configs also (client binary verbosity, connection timeout...).

python/tests/test_vast.py
```diff
@@ -20,59 +20,79 @@
...
-    await asyncio.sleep(3)
-    yield
+    endpoint = await proc.stdout.readline()
+    endpoint = endpoint.decode("utf-8").strip()
+    logger.debug(f"{endpoint = }")
+    yield endpoint
     proc.terminate()
+    # TODO: kill if in case of timeout.
     await asyncio.wait_for(proc.wait(), 5)
     await asyncio.to_thread(shutil.rmtree, TEST_DB_DIR)
```
rdettai, line 43: if we run outside a container, should we also call this cleanup before starting the server in case the states gets dirty for some reason?

tobim: Sounds reasonable.

python/tests/test_vast.py
```diff
@@ -102,14 +125,19 @@ async def test_export_historical_rows():
...
 
 @pytest.mark.asyncio
-async def test_export_continuous_rows():
+async def test_export_continuous_rows(endpoint):
+    vast = VAST(endpoint)
+
     async def run_export():
-        result = VAST.export('#type == "suricata.alert"', ExportMode.CONTINUOUS)
+        result = vast.export(
+            '#type == "suricata.alert"', ExportMode.CONTINUOUS, limit=1
```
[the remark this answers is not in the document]
tobim: :man_facepalming: I tried with a build that didn't have the fix.

tobim:
@rdettai maybe you have an idea why the thehive test fails with this?

python/vast/vast/vast.py
```diff
@@ -76,8 +76,9 @@ def _next_batch():
 class VAST:
     """An instance of a VAST node."""
 
-    def __init__(self):
-        pass
+    def __init__(self, endpoint=None):
+        if endpoint is not None:
+            self.cli = CLI(endpoint=endpoint)
```
rdettai, line 81: Didn't try it, but this seems to imply that you leave `self.cli` as undefined when `endpoint` is `None`. That would then fail when you call `self.cli.xxx()`, no? Ideally, this should somehow be covered in the tests.

python/vast/vast/cli.py
```diff
@@ -65,17 +66,21 @@ def run(*args, stdin) -> subprocess.Popen:
...
 
         def command(*args, **kwargs):
             if kwargs:
-                self.args.extend(CLI.arguments(**kwargs))
+                new.args.extend(CLI.arguments(**kwargs))
             if args:
-                self.args.extend(args)
-            return self
+                new.args.extend(args)
+            return new
```
rdettai, line 84: what problem does this solve exactly?

python/vast/vast/vast.py
rdettai: > @rdettai maybe you have an idea why the thehive test fails with this?

This might actually be the reason

python/vast/vast/cli.py
tobim: Every call into this function creates a new copy of `Cli` instead of mutating the object in place. Without this a `Cli` object is not reusable as a second invocation would just append commands an options to the command line that was set in the previous call.
Example:
```
>>> mycli = Cli(endpoint="127.0.0.1:42000")
>>> mycli.import().zeek().exec()
vast -e 127.0.0.1:42000 import zeek
>>> mycli.status().exec()
vast -e 127.0.0.1:42000 import zeek status
```

python/vast/vast/vast.py
tobim: Good catch!

python/vast/vast/cli.py
mavam: I asked myself the same question looking at this. It would be good to document.

python/vast/vast/vast.py
tobim: Looks like that was not the problem.

rdettai (approved):
I'll append the update of thehive script according to these changes before merging

PR #2873, opened 2023-01-16
Remove the `pattern` type

dominiklohmann:
This follows up on the removal of the match operator and introduction of pattern search using predicates of the form `lhs == /rhs/` by removing the pattern type without replacement.

In theory this is a hard and incompatible breaking change. However, we know of no single user that has relied upon the pattern type being available, and as such it _should_ be safe to remove it without a transparent upgrade path.

To make clear just how unlikely it is that anyone ran into this: users would've had to have a field of type `pattern #index=skip` in their schema, as attempting to index a pattern field crashed because we never actually implemented support for it, which was (1) never documented and (2) nobody ever complained about it.


libvast/fbs/type.fbs
```diff
@@ -89,6 +89,7 @@ union Type {
   duration_type: DurationType,
   time_type: TimeType,
   string_type: StringType,
+  // NOTE: The PatternType was removed with VAST v3.0.
```
mavam, line 92: Can we mark the field as `(deprecated)` then?

libvast/src/concept/parseable/vast/legacy_type.cpp
```diff
@@ -64,7 +64,8 @@ bool legacy_type_parser::parse(Iterator& f, const Iterator& l,
     | "duration"_p  ->* type_factory<legacy_duration_type>
     | "time"_p      ->* type_factory<legacy_time_type>
     | "string"_p    ->* type_factory<legacy_string_type>
-    | "pattern"_p   ->* type_factory<legacy_pattern_type>
+    // We removed support for pattern types with VAST v3.0.
+    // | "pattern"_p   ->* type_factory<legacy_pattern_type>
```
mavam, line 68: Why not ditch it entirely?

mavam (approved):
Wow, that was fast. 🙂

libvast/fbs/type.fbs
dominiklohmann: That doesn't work on union members, and setting a manual `(id: 9)` for the field after it to skip one index only works for a FlatBuffers version that is too recent for Debian.

libvast/src/concept/parseable/vast/legacy_type.cpp
dominiklohmann: I cannot remove the `legacy_pattern_type` as that also changes how some other legacy types are serialized, so I figured it's best to just leave it in and comment it out here. Once we upgrade the minimum partition version to 3 (I believe) we can get rid of it. Patryk removed the last remaining serialized uses of it with the CAF upgrade.

I would really love to get rid of the legacy type system, but we're still not quite there yet.

PR #2874, opened 2023-01-17
Implement a first basic pipeline string parser for the currently existing pipeline operators

Dakostu:
TBD:
- [ ] cleanup/optimization
- [ ] description


libvast/builtins/operators/drop.cpp
```diff
@@ -129,6 +131,19 @@ class plugin final : public virtual pipeline_operator_plugin {
...
+  parse_pipeline_string(std::string_view str) const override {
+    record options;
+    auto parse_result = system::parse_pipeline(str);
+    if (parse_result.parse_error) {
+      return {parse_result.new_str_it, parse_result.parse_error};
+    }
+    options["fields"] = parse_result.extractors;
+    options["schemas"] = parse_result.extractors;
+    return {parse_result.new_str_it, make_pipeline_operator(options)};
+  }
```
dominiklohmann, line 146: Let's take a step back to the drawing board for how we write these parsers such that they are maintainable. I think we should re-use the basics from parseable in this:

```cpp
// Turn the remaining pipeline into an iterator pair.
auto f = str.begin();
const auto l = str.end();

// Assemble our parser.
using parsers::space, parsers::alnum, parsers::chr, parsers::eoi;
const auto ws = ignore(*space);
const auto extractor_char = alnum | chr{'_'} | chr{'-'} | chr{':'};
const auto extractor = !(':'_p | '-') >> (+extractor_char % '.');
const auto extractor_list = (extractor % (',' >> ws));
const auto p = extractor_list >> ws >> (eoi | '|');

// Parse!
auto extractors = std::vector<std::string>{};
if (!p(f, l, extractors))
  return {f, caf::make_error(...)};
if (extractors.empty())
  return {f, caf::make_error(...)};
return {f, std::make_unique<drop_operator>(std::move(extractors))};
```

Do you want to pair on this?

libvast/src/detail/string.cpp
```diff
@@ -153,6 +153,15 @@ std::vector<std::string_view> split(std::string_view str, std::string_view sep,
   return pos;
 }
 
+std::string_view trim(std::string_view str) {
+  auto not_space = [](char c) {
```
dominiklohmann, line 157:
```suggestion
  auto not_space = [](unsigned char c) {
```

Actually UB otherwise, see https://en.cppreference.com/w/cpp/string/byte/isspace

> The behavior is undefined if the value of ch is not representable as `unsigned char` and is not equal to [EOF](http://en.cppreference.com/w/cpp/io/c).

dominiklohmann (commented):
Just took a first look, and I think we can re-use a lot more from parseable to improve maintainability. I commented on one operator as an example.

Dakostu:
This will be replaced by  a PR based on the Hackathon pipeline operator branch.

PR #2875, opened 2023-01-18
Add a new component_plugin::component_name() function

lava:
This makes it possible for plugins to use a component name that is more fitting than the plugin name (ie. 'matcher' vs 'matcher_supvervisor'), or keep the old component name when renaming the plugin.


dominiklohmann (approved):
Exactly as we just talked about. Thanks!

PR #2876, opened 2023-01-19
Publish REST API blog post

mavam:
This PR pushes the REST API blog post over the finish line.

### DIFF
--- a/web/blog/the-new-rest-api/index.md
+++ b/web/blog/the-new-rest-api/index.md
@@ -195,6 +195,12 @@ curl "https://vast.example.org:42001/api/v0/status?verbosity=detailed" \
 }
 ```
 
+:::caution Status changes in v3.0
+In the upcoming v3.0 release, the statistics under the key `.index.statistics`
+will move to `.catalog`. This change is already merged into the master branch.
+Consult the [status key reference](/docs/setup/monitor#reference) for details.
+:::
+
 ### Perform a HTTP health check
 
 The `/status` endpoint can also be used as a HTTP health check in
--- a/web/blog/the-new-rest-api/index.md
+++ b/web/blog/the-new-rest-api/index.md
@@ -212,7 +212,7 @@ services:
   web:
     image: tenzir/vast
     environment:
-      - "VAST_START__COMMANDS=['web server --mode=dev']"
+      - "VAST_START__COMMANDS=web server --mode=dev"
     ports:
       - "42001:42001"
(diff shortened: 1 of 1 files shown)


web/blog/the-new-rest-api/index.md
```diff
@@ -157,16 +170,191 @@ vast web server --mode=mtls
...
+The `/status` endpoint can also be used as a HTTP health check in
+`docker-compose`:
+
+```yaml
+version: '3.4'
+services:
+  web:
+    image: tenzir/vast
+    environment:
+      - "VAST_START__COMMANDS=['web server --mode=dev']"
```
dominiklohmann, line 209: This syntax is going away with VAST v3.0, where it's just comma-separated. As that already worked with VAST v2.4, I would recommend using that here.

```suggestion
      - "VAST_START__COMMANDS=web server --mode=dev"
```

web/blog/the-new-rest-api/index.md
```diff
@@ -157,16 +170,191 @@ vast web server --mode=mtls
...
-end-to-end examples. We built the REST API for two reasons::
+end-to-end examples.
+
+### See what's inside VAST
+
+One straightforward example is checking the number of records in VAST:
+
+```bash
+curl "https://vast.example.org:42001/api/v0/status?verbosity=detailed" \
+  | jq .index.statistics
```
dominiklohmann, line 181: Can we add a note here that this will change with VAST v3.0?

web/blog/the-new-rest-api/index.md
```diff
@@ -157,16 +170,191 @@ vast web server --mode=mtls
...
-1. Enable third parties to integrate with VAST
-2. Develop our own web frontend
+```bash
+curl "http://127.0.0.1:42001/api/v0/export?limit=2&expression=net.src.ip%20in%20192.168.0.0%2f16"
+```
+
+```json
+{
+  "version": "v2.4.0-457-gb35c25d88a",
+  "num_events": 2,
```
dominiklohmann, line 233: Should this key be using a dash rather than an underscore in the future? We use dashes in the status output keys, for example.

dominiklohmann (approved):
This is a good read. I'd say we ship it. I just have some minor notes.

web/blog/the-new-rest-api/index.md
mavam: Created an issue for that.

PR #2878, opened 2023-01-19
Remove unnecessary job dependencies in CI

dominiklohmann:


mavam (approved):

mavam:
It looks like the website CI run only ran now and revealed that my previous PR contained broken links. 😒

dominiklohmann:
Afaict that's a timing issue with a command not properly blocking for long enough. This we can only fix by reworking how we do blocking imports. I'd just re-run for now.

mavam:
Okay, I fixed the broken link (needed a force-push after rebasing). Let's see where this gets us.

dominiklohmann:
Yes, that's a bug in TheHive starting up. It's why the job is not required for branch protections to pass.

PR #2879, opened 2023-01-22
Use SVGs instead of PNGs on vast.io

mavam:
The effect of this PR is that we can now finally use SVGs for our diagram in the docs, as opposed to the tedious process of double-exporting PNGs in light and dark mode. This makes it a lot easier for docs maintainers to produce visual content.


mavam:
While this works nicely for SVGs that are in the same directory as the Markdown file, path resolution fails when referencing SVGs relative to `/img/`.

I asked for ideas in https://github.com/alvinometric/remark-inline-svg/issues/1.

mavam:
Does that actually do SVG inlining for you. For me I get an `img` node:

web/blog/apache-arrow-as-platform-for-security-data-engineering/index.md
```diff
@@ -59,8 +59,7 @@ we started using Arrow as optional dependency as additional column-oriented
 encoding. We abstracted a batch of data encoding-independent behind a "table
```
dominiklohmann, line 59: I can't comment on a submodule bump, so here's a comment on an unrelated line: This PR should not be changing the CAF submodule from 0.18.7 back down to 0.18.6.

mavam: Thanks for the 🦅 👀 !

dominiklohmann (approved):
This works great for me locally. Thanks for also moving the images.

web/src/pages/index.module.css
```diff
@@ -105,3 +105,7 @@
   max-height: 200px;
   max-width: 500px;
 }
+
+[data-theme="dark"] .svgWrapper svg {
+  filter: invert(100%) hue-rotate(180deg);
+}
```
mavam, line 111: We have already a definition in our global CSS file. Should we perhaps just reuse that class definition? Or is there a benefit of doing that here separately?

mavam: (Your fixups work like a charm otherwise, by the way. I've enabled auto-merge to to signal we're done minus this thread.)

dit7ya: Okay, I have reused the `markdown-inline-svg` class.

PR #2884, opened 2023-01-25
Relax integration test strictness

tobim:
This is a band aid over integration test failures in CI that block progress of unrelated PRs. The real fixes will take a little bit of time to implement correctly, so we this is proposed as a temporary workaround.

### DIFF
--- a/vast/integration/integration.py
+++ b/vast/integration/integration.py
@@ -364,16 +364,19 @@ class Server:
         LOGGER.debug(f"stopping server fixture: {command}")
         stop_out = open(self.cwd / "stop.out", "w")
         stop_err = open(self.cwd / "stop.err", "w")
-        stop = 0
         try:
             stop = spawn(
                 command,
                 cwd=self.cwd,
                 stdout=stop_out,
                 stderr=stop_err,
-            ).wait(STEP_TIMEOUT)
+            )
+            stop.wait(STEP_TIMEOUT)
         except:
-            stop.kill()
+            pass
+        else:
+            if isinstance(stop, subprocess.Popen):
+                stop.kill()
         try:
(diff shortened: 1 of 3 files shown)


vast/integration/integration.py
```diff
@@ -610,7 +610,7 @@ def match(x, names):
                 test_result = Result.FAILURE
                 for i in range(0, args.repetitions):
                     test_result = tester.run(name, definition)
-                    if test_result is Result.TIMEOUT:
+                    if test_result is not Result.SUCCESS:
```
dominiklohmann, line 613: Shouldn't this be `if not a is b`, rather than `if a is not b`? Or does Python just handle this the same way (the precedence is confusing).

dominiklohmann (approved):
Pretty much what we discussed earlier, so let's see if that works. Ideally by the time we switch to bats (i.e., #2859) we can just drop this again, or rather not implement the same workaround in there.

I did not run locally, so before merging, please carefully take a look at the CI runs to see if this works as expected.

vast/integration/integration.py
tobim: https://peps.python.org/pep-0008/#programming-recommendations

tobim: Sorry for the brevity, I was a bit preoccupied.

dominiklohmann: No worries, I think your reply actually had the perfect level of information density.

dominiklohmann (approved):
Approving under the assumption that this is temporary. Please also create an issue for re-enabling the newly disabled test.

tobim:
https://github.com/orgs/tenzir/projects/11/views/4?filterQuery=&pane=issue&itemId=18955379

PR #2887, opened 2023-01-26
Align output of the Zeek TSV reader with schemas

dominiklohmann:
This changes the Zeek TSV reader to respect available schemas from schema files in case the parsed output can be casted to a schema with an exactly matching name.

In case the casting is impossible, the Zeek reader warns exactly once when switching to a new file and ignores the user-provided schema.

To achieve this, I have introduced a new set of functions `vast::cast` and `vast::can_cast`, with the former operating on arrays and the latter operating on types. For now, this only supports reshuffling columns in record types (and corresponding struct arrays), but the mechanism is written in a way that is easily extensible via template specializations.

```[tasklist]
### Tasks
- [x] Implement no-op cast operations
- [x] Implement cast operations on record types to reshuffle columns
- [x] Utilize cast operation in Zeek TSV reader
- [x] Implement casting for lists
- [x] Write unit tests for cast/can_cast
- [x] Write changelog entry
- [x] ~Update docs~ Docs already had unflattened fields
```

Fixes tenzir/issues#36.

To whoever is reviewing this: I'm sorry for the ginormous diff. We happen to use the Zeek TSV reader in a lot of integration tests.


mavam:
I can see the value of the casting, but I don't understand the big picture. I would imagine it suffices to:

1. Read the Zeek fields
2. Unflatten records by splitting on `.`
3. Set the type in the builder

Then we're done. Am I missing something here?

dominiklohmann:
> Then we're done. Am I missing something here?

Yes. With what you've suggested you get data with a different schemas for the Zeek JSON and the Zeek TSV readers, which is a really bad UX when you're working with both.

> Unflatten records by splitting on `.`

Before this change, the Zeek reader already looked at the schema files we had loaded, and iff it found a matching schema attempted to adjust the schema to that. However, that was incomplete—the end result was a schema that was neither that of the data nor the one we found in the loaded schema files.

Also, there is at least one edge case where unflattening based on dots does not work at all: Consider Zeek data that has the following field names (in order): `id.orig_h`, `ts`, `id.orig_p`. What you suggested breaks in this case as you would need to re-order leaves.

mavam:
> Yes. With what you've suggested you get data with a different schemas for the Zeek JSON and the Zeek TSV readers, which is a really bad UX when you're working with both.

I don't see how that would happen.

> [..] the Zeek reader already looked at the schema files we had loaded

Is that really the desired UX? The format is self-describing. I would expect that the first layer is just onboarding the data. Transforming it according to known schemas should happen down the line (in the processing model we are currently towards to).

> Also, there is at least one edge case where unflattening based on dots does not work at all: Consider Zeek data that has the following field names (in order): `id.orig_h`, `ts`, `id.orig_p`. What you suggested breaks in this case as you would need to re-order leaves.

That is an artificial example. Zeek doesn't create such data.

dominiklohmann:
> > [..] the Zeek reader already looked at the schema files we had loaded
> 
> Is that really the desired UX? [...]

Yes, I would say so. If the user describes the schema of their data, then we should adhere to that. The Zeek reader still works in isolation—but as a user I would not want two different schemas to show up when I run `vast show schemas zeek.conn`.

Besides, how else would the Zeek reader know about the `port` and `timestamp` aliases, or about `#index=hash` attribute? Just unflattening on a dot is not good enough.

> > [...] different schemas for the Zeek JSON and the Zeek TSV readers [...]
> 
> I don't see how that would happen.

Look at the diff of the integration test reference files. For example, you can see the field `local_resp` missing before this change in the ingested `zeek.conn` events from Zeek TSV files, which was (1) described in the schema file and (2) available in the Zeek JSON data.

> That is an artificial example. Zeek doesn't create such data.

Then think of CSV instead of Zeek, for which you cannot rely on the educated guess that nobody will ever write such data. It has all the same problems.

Remember the CSV reader problems that @rdettai encountered? This is able to fix them all the same, agnostic to the readers themselves. It's a building block that while for now only applied to Zeek TSV will in the near future be useful in all the many variants of the `from` and `read` operators.

libvast/include/vast/cast.hpp
```diff
@@ -0,0 +1,266 @@
...
+}
+
+/// Casts a table slice to another schema.
+/// @param from_slice The table slice to cast.
+/// @param to_schema The schema to cast towards.
+/// @pre can_cast(from_slice.schema(), to_schema)
+/// @post result.schema() == to_schema
+/// @returns A slice that exactly matches *to_schema*.
+auto cast(table_slice from_slice, const type& to_schema) noexcept
+  -> table_slice;
```
mavam, line 264:
```suggestion
/// Casts a table slice to another schema.
/// @param from The table slice to cast.
/// @param to The schema to cast towards.
/// @pre can_cast(from_slice.schema(), to_schema)
/// @post result.schema() == to_schema
/// @returns A slice that exactly matches *to_schema*.
auto cast(table_slice from, const type& to) noexcept -> table_slice;
```

To me, "cast" already implies what's going on. It doesn't need to be duplicated into the prefix of the argument identifiers.

(Bikeshedding comment, because everything else looks great.)

libvast/include/vast/format/zeek.hpp
```diff
@@ -251,6 +251,7 @@ class reader final : public single_schema_reader {
   std::string unset_field_;
   vast::module module_;
   type schema_;
+  type output_schema_;
```
mavam, line 254: Maybe add a comment that this is only here temporarily until the `cast` operator is available?

vast/integration/reference/arrow-export/step_01.ref
```diff
@@ -1,30 +1,37 @@
...
   VAST:name:0: 'timestamp'
+  VAST:name:0: 'zeek.conn_id'
   child 0, item: string
+  child 0, orig_h: fixed_size_binary[16]
+  child 1, orig_p: uint64
+  child 2, resp_h: fixed_size_binary[16]
+  child 3, resp_p: uint64
 -- schema metadata --
 VAST:name:0: 'zeek.conn'
+_path: string
```
mavam, line 24: Note to self: I think it's okay to effectively duplicate the type name here, as it preserves the of the input. In fact, it's probably a highly desirable property in case we need to convince someone that we really don't throw away anything, i.e., the original output can be reconstructed if need be.

plugins/pcap/integration/reference/pivot-with-pcap/step_08.ref
```diff
@@ -1,4 +1,4 @@
...
-{"ts": "2018-10-30T09:43:22.519540", "uid": "C6OJP11NPTa6teXh6b", "id.orig_h": "192.168.168.100", "id.orig_p": 61427, "id.resp_h": "72.247.178.26", "id.resp_p": 80, "proto": "tcp", "service": "http", "duration": "1.92m", "orig_bytes": 302, "resp_bytes": 384, "conn_state": "SF", "local_orig": null, "local_resp": null, "missed_bytes": 0, "history": "ShADadFf", "orig_pkts": 17, "orig_ip_bytes": 1066, "resp_pkts": 15, "resp_ip_bytes": 1172, "tunnel_parents": null, "community_id": "1:9a4Fl8C3ce0tZxYivPMIMH6RcWc="}
+{"_path": null, "_write_ts": null, "ts": "2018-10-30T08:59:47.089835", "uid": "CPWqXM3DsYjHkrX69j", "id": {"orig_h": "192.168.168.100", "orig_p": 61257, "resp_h": "17.253.55.211", "resp_p": 80}, "proto": "tcp", "service": "http", "duration": "38.7ms", "orig_bytes": 288, "resp_bytes": 4080, "conn_state": "SF", "local_orig": null, "local_resp": null, "missed_bytes": 0, "history": "ShADadFf", "orig_pkts": 7, "orig_ip_bytes": 664, "resp_pkts": 6, "resp_ip_bytes": 4400, "tunnel_parents": null, "community_id": "1:UY9rhotoYM0Z0HYmZjBZwc/GhcA="}
```
mavam, line 1: `_path` and `_write_ts` should probably be not `null`, right?

vast/integration/reference/pivot/step_06.ref
```diff
@@ -1,12 +1,12 @@
...
-1540889849.325226	CcAj0Y2U8EU6iPHeW4	192.168.168.100	61209	72.247.178.18	80	tcp	http	115.720806	445	868	SF	-	-	0	ShADadFf	17	1209	15	1656	-	1:6vVq2GDCfUtz+oh2oh6y5sM8Q5M=
-1540889849.346318	CpOjhd2G4wG4Ae8J6i	192.168.168.100	61210	72.247.178.18	80	tcp	-	5.0340250	0	0	SF	-	-	0	ShAFaf	4	220	3	164	-	1:XW8RnI0g5aEs7vhnOuehNkGD8H4=
+-	-	1540889849.325075	C4xvGs2g6UWECwk2Rb	192.168.168.100	61206	72.247.178.18	80	tcp	http	115.721079	890	1736	SF	-	-	0	ShADadFf	19	1758	16	2576	-	1:aiB9ixqgqVUlxDucVgiCzPCos7A=
+-	-	1540889849.325117	CUtqdo25al9U3RGVr3	192.168.168.100	61207	72.247.178.18	80	tcp	http	115.721076	445	868	SF	-	-	0	ShADadFf	17	1209	15	1656	-	1:JvzWHQdEF7h98wpzA1qPdP0/cAU=
+-	-	1540889849.325145	CnDxlq40ITKKeEYTW3	192.168.168.100	61208	72.247.178.18	80	tcp	http	115.721008	445	868	SF	-	-	0	ShADadFf	17	1209	15	1656	-	1:yZL4x1ra1wearOeXlyhuOH+A1qE=
+-	-	1540889849.325226	CcAj0Y2U8EU6iPHeW4	192.168.168.100	61209	72.247.178.18	80	tcp	http	115.720806	445	868	SF	-	-	0	ShADadFf	17	1209	15	1656	-	1:6vVq2GDCfUtz+oh2oh6y5sM8Q5M=
+-	-	1540889849.346318	CpOjhd2G4wG4Ae8J6i	192.168.168.100	61210	72.247.178.18	80	tcp	-	5.0340250	0	0	SF	-	-	0	ShAFaf	4	220	3	164	-	1:XW8RnI0g5aEs7vhnOuehNkGD8H4=
```
mavam, line 12: This is what I mean: many folks still use awk to go through their Zeek logs. I know it's brittle in the first place, but this one breaks everything, and not just some custom columns appended at the end.

vast/integration/reference/pivot/step_06.ref
```diff
@@ -1,12 +1,12 @@
 #empty_field	(empty)
-#fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	proto	service	duration	orig_bytes	resp_bytes	conn_state	local_orig	local_resp	missed_bytes	history	orig_pkts	orig_ip_bytes	resp_pkts	resp_ip_bytes	tunnel_parents	community_id
+#fields	_path	_write_ts	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	proto	service	duration	orig_bytes	resp_bytes	conn_state	local_orig	local_resp	missed_bytes	history	orig_pkts	orig_ip_bytes	resp_pkts	resp_ip_bytes	tunnel_parents	community_id
```
mavam, line 2: This output is not correct. The Zeek TSV writer has the metadata embedded in the header:

- `_path` is `#path`
- `_write_ts` is not of concern and can be dropped.

No tool that relies on Zeek TSV data can use this output.

plugins/pcap/integration/reference/pivot-with-pcap/step_08.ref
```diff
@@ -1,4 +1,4 @@
...
+{"_path": null, "_write_ts": null, "ts": "2018-10-30T09:09:06.022106", "uid": "ClKkVB2KudjVql07B3", "id": {"orig_h": "192.168.168.100", "orig_p": 61290, "resp_h": "17.253.55.204", "resp_p": 80}, "proto": "tcp", "service": "http", "duration": "41.73ms", "orig_bytes": 291, "resp_bytes": 4102, "conn_state": "SF", "local_orig": null, "local_resp": null, "missed_bytes": 0, "history": "ShADadFf", "orig_pkts": 7, "orig_ip_bytes": 667, "resp_pkts": 6, "resp_ip_bytes": 4422, "tunnel_parents": null, "community_id": "1:+PG4ulcu1wtMyigJpD3ohP1Jj7E="}
```
mavam, line 2: NB: this is *not* how "Zeek JSON" users would expect their data. The unflattening is arguably a strict improvement, but it could break assumptions for folks that process Zeek data. I would document this somewhere for now.

As a follow-up, I realized that we probably want `flatten` and `unflatten` operators, at least a function we can use in `put`. That might be just an option to `cast`, e.g., `cast --unflatten`. We should write that down somewhere.

plugins/web/integration/reference/export-endpoint/step_02.ref
```diff
@@ -1,442 +1,562 @@
...
+    "local_resp": null,
     "missed_bytes": 0,
     "history": "D",
     "orig_pkts": 7,
     "orig_ip_bytes": 546,
     "resp_pkts": 0,
     "resp_ip_bytes": 0,
-    "tunnel_parents": []
+    "tunnel_parents": [],
+    "community_id": null
```
mavam, line 84: I'm pondering whether I like that we now have all custom fields present, like `community_id` and `local_resp`. The fact that we have the `--omit-*` functions makes me think it's okay, though.

mavam (changes requested):
Overall, solid work here on the core side.

Change requests:
- There seems to be a bug in the output, `_path` and `_write_ts` are always `null`
- The Zeek TSV writer is too far away from TSV output. I propose to (1) move `_path` to `#path` and drop `_write_ts` for now.
- We need documentation changes that we "upgraded" the Zeek JSON logs to be structured. This breaks tools that rely on the flattened version, e.g., RITA, ZAT, etc. While there is probably `jq` hotfix for some use cases where the user can impose, we can't assume that. We should probably also mention that an upcoming `cast` operator (or whatever we choose) will make it possible transition explicitly.

plugins/pcap/integration/reference/pivot-with-pcap/step_08.ref
dominiklohmann: The fields are not present in the Zeek TSV we ingest, but are present in the matching schema because we also use that for Zeek JSON. That's why the value is always null.

vast/integration/reference/pivot/step_06.ref
dominiklohmann: That would be a bug in the Zeek writer then, which was not changed by this PR at all. What is it supposed to do except for taking the data it gets and rendering it exactly as it is?

plugins/pcap/integration/reference/pivot-with-pcap/step_08.ref
dominiklohmann: > As a follow-up, I realized that we probably want flatten and unflatten operators, at least a function we can use in put. That might be just an option to cast, e.g., cast --unflatten. We should write that down somewhere.

There's no need for those options, as which parts of the output are flattened and which aren't is defined in the resulting schema. As such, if you want the output to be flattened you can just change your Zeek schemas to contain four fields `id.*` instead of a nested record `ìd`, which as of this PR equally works for Zeek JSON and Zeek TSV.

> NB: this is not how "Zeek JSON" users would expect their data. The unflattening is arguably a strict improvement, but it could break assumptions for folks that process Zeek data. I would document this somewhere for now.

The Zeek JSON import was not changed by this PR at all, it behaves the exact has behaved since we added it, so I still don't understand what you think changed with it.

The JSON export also already supports a `--flatten` option to flatten on rendering.

vast/integration/reference/pivot/step_06.ref
dominiklohmann: This is not a Zeek log. This an export of a `zeek.conn` from VAST in the Zeek TSV format. Why would you expect that not to contain all fields present in the original data?

(Besides, what's the use case for exporting as Zeek TSV, anyways?)

libvast/include/vast/format/zeek.hpp
dominiklohmann: The readers need to change so much for the from/read refactoring that I think we are likely to clone them anyways. I don't think we can plug the current reader API into the upcoming framing, so we'll pick out what we need then.

libvast/include/vast/cast.hpp
dominiklohmann: I'm against this change. The implementation of this function is much easier to read with the very explicit variable names, and the variable names in the function definition should match those in its definition.

dominiklohmann:
@mavam To address the meta comment from your review, as it's not in a review comment:

> * There seems to be a bug in the output, `_path` and `_write_ts` are always `null`

They are not present in Zeek TSV, so these fields are only set when you import Zeek JSON. This is working as expected.

> * The Zeek TSV writer is too far away from TSV output. I propose to (1) move `_path` to `#path` and drop `_write_ts` for now.

I don't think the Zeek TSV writer should make implicit changes to data based on the type name. It is not restricted to writing Zeek logs.

> * We need documentation changes that we "upgraded" the Zeek JSON logs to be structured. This breaks tools that rely on the flattened version, e.g., RITA, ZAT, etc. While there is probably `jq` hotfix for some use cases where the user can impose, we can't assume that. We should probably also mention that an upcoming `cast` operator (or whatever we choose) will make it possible transition explicitly.

- `vast export json --flatten` has existed for a long time now to flatten on export.
- The imported data is now exactly aligned with the schema, which we have always done for Zeek JSON, but until now not for Zeek TSV. This is a bug fix, not a feature.

vast/integration/reference/pivot/step_06.ref
mavam: > This is not a Zeek log. 

Huh? It is a Zeek conn.log as far as I can tell. But that shouldn't matter here. Zeek TSV output and logs are treated equivalently from the consumer perspective.

> Why would you expect that not to contain all fields present in the original data?

Zeek TSV has metadata fields, like `#path`. These contain the type (or log) name. It doesn't make sense to add a `_path` field on top of that. That's belt and suspenders. No tool that processes Zeek TSV logs (or query output) relies on a `_path` field.

> (Besides, what's the use case for exporting as Zeek TSV, anyways?)

To re-hydrate Zeek logs for tools that rely on them. RITA comes to mind.

vast/integration/reference/pivot/step_06.ref
mavam: > That would be a bug in the Zeek writer then, which was not changed by this PR at all.

That might be the developer view, but that doesn't matter from a user perspective. The output is now vastly different and that's what matters. 

We had this recently: making a change that has global effects and then arguing "I haven't touched the other code" is not a valid point. We must look at this from the user perspective, not (just) the developer viewpoint.

> What is it supposed to do except for taking the data it gets and rendering it exactly as it is?

It should strip Zeek-related meta data, if the input is Zeek data. My previous comments has two bullets that describe what the user expectation is. At a high level, the expectation is that meta data gets transposed into the places where the user expects it.

That said, I can see where you coming from and I agree that this breaks a symmetry. My suggestion to appease: make this transposition a toggle with an option, e.g., `--embed-metadata`.

plugins/pcap/integration/reference/pivot-with-pcap/step_08.ref
mavam: I think this discussion is subsumed by the one below, where I proposed `--embed-metadata` to resolve this. This would strip these fields.

vast/integration/reference/pivot/step_06.ref
dominiklohmann: I don't think logic about types based on their name should sneak into the writers for specific formats. That is an approach that is (1) hidden from the user and (2) does not scale. It is bound to bite us sometime in the future.

A writer for a given format should just write data. That is its single responsibility.

Downstream tooling that works on Zeek logs should access the fields by field name, not by column index. Which downstream tool do you know of that just uses the column index? Wouldn't you always go through `zeek-cut` first?

What we can do here is the following:
- Remove `_path` from the schema. It is only needed for Zeek JSON reader to determine which schema to use (the selector field of the JSON reader), so there isn't much use in actually putting it into the data. If users really want that, they are already able to put the field back using pipeline operators.
- Move `_write_ts` to the end of all schemas. We simply cannot remove it, as it is data and not metadata.

dominiklohmann: I pushed a commit that makes this change. If you're fine with that then all we need is another changelog entry.

mavam (approved):
This works for now. I think the Zeek writer is troublesome in many ways and we've reached a working solution.

The fact that we have to add `_write_ts` to all schemas feels wrong, but there's no alternative at the moment.

dominiklohmann:
This took a while to get to this point, but I think it's in a way better spot than what we had before this PR now. Thanks for all the discussion, @mavam. 🙏

PR #2888, opened 2023-01-29
Be lenient about whitespace in pipeline parser

dominiklohmann:
This makes it possible to also include whitespace characters like non-breaking spaces and newlines in pipelines, effectively allowing the user to write complex pipelines with formatting however they prefer. For example, if you'd like to get really funky with your formatting, this would now be possible:

```bash
vast export json <<EOF
identity
| drop ts |
select orig_h,
  orig_p
 ,resp_h
  | identity
EOF
```

@Dakostu I requested you for review since we had already talked about this one—and I have mistaken the char classes the last time around.

### DIFF
--- a/libvast/builtins/query-languages/vastql.cpp
+++ b/libvast/builtins/query-languages/vastql.cpp
@@ -37,11 +37,11 @@ class plugin final : public virtual query_language_plugin {
         std::optional<pipeline>{},
       };
     }
-    using parsers::blank, parsers::expr, parsers::eoi;
+    using parsers::space, parsers::expr, parsers::eoi;
     auto f = query.begin();
     const auto l = query.end();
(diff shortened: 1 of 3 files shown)


Dakostu (approved):
So `parsers::space` does in fact do consider more complex whitespace such as tabs and line breaks?
If that is the case, then OK.

dominiklohmann:
> So `parsers::space` does in fact do consider more complex whitespace such as tabs and line breaks?
> If that is the case, then OK.

Yes, I was mistaken about the two char classes last week in my review. See [isspace](https://en.cppreference.com/w/cpp/string/byte/isspace) and [isblank](https://en.cppreference.com/w/cpp/string/byte/isblank). Sorry for the confusion.

PR #2890, opened 2023-01-29
Pass table slices through pipeline operators

dominiklohmann:
This lowers the boundary to writing additional operators, as falling down to the lower-level Arrow APIs is now opt-in rather than being required. It also makes it so that pipeline operators that do not create new data (e.g., summarize) can keep the import timestamps unchanged, which fixes a long-standing issue that caused compaction and rebuilding to shift import timestamps.

Most importantly, this will make the transition to the upcoming rewritten executor and its pipeline operator API changes a lot easier.

This is based on #2889.

### DIFF
--- a/libvast/src/pipeline.cpp
+++ b/libvast/src/pipeline.cpp
@@ -145,8 +145,8 @@ caf::error pipeline::add_slice(table_slice slice) {
 }
 
 caf::error
-pipeline::process_queue(const std::unique_ptr<pipeline_operator>& op,
-                        std::vector<table_slice>& result, bool check_schema) {
+pipeline::process_queue(pipeline_operator& op, std::vector<table_slice>& result,
+                        bool check_schema) {
   caf::error failed{};
   const auto size = to_transform_.size();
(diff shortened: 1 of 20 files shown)


libvast/src/pipeline.cpp
```diff
@@ -120,42 +119,48 @@ caf::expected<std::vector<table_slice>> pipeline::finish() {
...
-  to_transform_.emplace_back(std::move(schema), std::move(batch));
+  to_transform_.emplace_back(std::move(slice));
   return caf::none;
 }
 
-caf::error pipeline::process_queue(const std::unique_ptr<pipeline_operator>& op,
-                                   std::vector<pipeline_batch>& result,
-                                   bool check_schema) {
+caf::error
+pipeline::process_queue(const std::unique_ptr<pipeline_operator>& op,
```
patszt, line 148: Not related to the changes. Feel free to drop. 
Unique_ptr part seems not used here. I think the parameter should just be pipeline_operator&

patszt (approved):

PR #2891, opened 2023-01-29
Implement `head` and `taste` operators

dominiklohmann:
The `head [n]` operator limits the result to the first `n` events. The limit defaults to 1.

The `taste [n]` operator behaves similarly, except that the limit is applied per unique schema. E.g., running `vast export json 'taste'` will return exactly one event of schema in the database.

```[tasklist]
### TODO
- [ ] https://github.com/tenzir/vast/pull/2890
- [ ] Unit and integration tests
- [ ] Documentation
```

### DIFF
--- a/libvast/builtins/operators/head.cpp
+++ b/libvast/builtins/operators/head.cpp
@@ -1,0 +1,98 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <vast/concept/parseable/numeric/integral.hpp>
+#include <vast/concept/parseable/vast/pipeline.hpp>
+#include <vast/error.hpp>
+#include <vast/logger.hpp>
+#include <vast/pipeline.hpp>
+#include <vast/plugin.hpp>
+#include <vast/table_slice.hpp>
+
+#include <arrow/type.h>
+
+namespace vast::plugins::head {
+
+namespace {
+
+class head_operator : public pipeline_operator {
+public:
+  explicit head_operator(uint64_t remaining) noexcept : remaining_{remaining} {
+    // nop
+  }
+
+  caf::error add(table_slice slice) override {
+    // When switching to the new operator API, we ideally want to implement an
+    // early exit for head here; for now this operator ends up being a lot of
+    // no-ops once we reached the limit.
+    if (remaining_ == 0)
+      return {};
+    slice = vast::head(slice, remaining_);
+    remaining_ -= slice.rows();
+    buffer_.push_back(std::move(slice));
+    return {};
+  }
+
+  caf::expected<std::vector<table_slice>> finish() override {
+    return std::exchange(buffer_, {});
+  }
+
+private:
+  std::vector<table_slice> buffer_ = {};
+  uint64_t remaining_ = {};
+};
+
+class plugin final : public virtual pipeline_operator_plugin {
+public:
+  // plugin API
+  caf::error initialize(data) override {
+    return {};
+  }
+
+  [[nodiscard]] std::string name() const override {
+    return "head";
+  };
+
+  // transform plugin API
+  [[nodiscard]] caf::expected<std::unique_ptr<pipeline_operator>>
+  make_pipeline_operator(const record&) const override {
+    return caf::make_error(ec::unimplemented,
+                           "the head operator is not available in the YAML "
+                           "operator syntax");
+  }
+
+  [[nodiscard]] std::pair<std::string_view,
+                          caf::expected<std::unique_ptr<pipeline_operator>>>
+  make_pipeline_operator(std::string_view pipeline) const override {
+    using parsers::optional_ws, parsers::required_ws,
+      parsers::end_of_pipeline_operator, parsers::u64;
+    const auto* f = pipeline.begin();
+    const auto* const l = pipeline.end();
+    const auto p
+      = -(required_ws >> u64) >> optional_ws >> end_of_pipeline_operator;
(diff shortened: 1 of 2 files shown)


libvast/builtins/operators/head.cpp
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  caf::error add(table_slice slice) override {
+    // When switching to the new operator API, we ideally want to implement an
+    // early exit for head here; for now this operator ends up being a lot of
+    // no-ops once we reached the limit.
+    if (remaining_ == 0)
+      return {};
+    slice = vast::head(slice, remaining_);
+    remaining_ -= slice.rows();
```
mavam, line 36: Maybe add an assert that `remaining_` doesn't underflow here?

libvast/builtins/operators/taste.cpp
```diff
@@ -0,0 +1,103 @@
...
+    auto remaining_it = remaining_.find(slice.schema());
+    if (remaining_it == remaining_.end()) {
+      remaining_it
+        = remaining_.emplace_hint(remaining_.end(), slice.schema(), limit_);
+    }
+    auto& remaining = remaining_it->second;
+    if (remaining == 0)
+      return {};
+    slice = vast::head(slice, remaining);
+    remaining -= slice.rows();
```
mavam, line 40: As above: underflow-safe?

libvast/builtins/operators/taste.cpp
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  caf::error add(table_slice slice) override {
+    // Determine the number of remaining events for this schema.
+    auto remaining_it = remaining_.find(slice.schema());
+    if (remaining_it == remaining_.end()) {
+      remaining_it
+        = remaining_.emplace_hint(remaining_.end(), slice.schema(), limit_);
+    }
+    auto& remaining = remaining_it->second;
```
mavam, line 36: Why not simply `remaining_[slice.schema()]`?

libvast/builtins/operators/head.cpp
dominiklohmann: It's safe and checked for, but I suppose an extra assertion for the reader of this code doesn't hurt.

libvast/builtins/operators/taste.cpp
dominiklohmann: That would default-construct the value to `0` if the schema hasn't been seen yet, while this is initializing the value to `limit_`.

libvast/builtins/operators/taste.cpp
dominiklohmann: See https://github.com/tenzir/vast/pull/2891#discussion_r1090263470; I've added an assertion.

libvast/builtins/operators/taste.cpp
dominiklohmann: I've expanded the comment directly above to make this clearer.

web/docs/understand/query-language/operators/taste.md
```diff
@@ -0,0 +1,18 @@
...
+Get the first result of each unique schema.
+
+```
+taste
+```
+
+Get the first five results of each unique schema.
+
+```
+head 5
```
Dakostu, line 17:
```suggestion
taste 5
```

libvast/test/pipeline.cpp
```diff
@@ -163,6 +162,68 @@ concatenate(caf::expected<std::vector<vast::table_slice>> slices) {
 
 FIXTURE_SCOPE(pipeline_tests, pipelines_fixture)
 
+TEST(head 1) {
```
Dakostu, line 165: There should be unit tests that cover using `head` and `taste` with numbers <= 0. If the pipeline parsing should fail at that point (or in the case of 0, show nothing) , it should be explicitly proven in these unit tests.

web/docs/understand/query-language/operators/head.md
```diff
@@ -0,0 +1,17 @@
+# head
+
+Limits the input to the first N results. The limit defaults to 1.
```
mavam, line 3:
```suggestion
Limits the input to the first N results. The limit defaults to 10.
```

This is in line with splunk and most other tools and what users expect.

mavam (changes requested):
We need `N = 10` as default to match the users' expectation from other tools.

Dakostu (approved):
I looked into the code and played around with the operators a bit and I don't see any (big) problems. So it's an approval from me. I did leave some comments though. Good luck with the demo!

web/docs/understand/query-language/operators/head.md
dominiklohmann: I changed the default to 10 for both, and also updated the documentation accordingly.

web/docs/understand/query-language/operators/taste.md
dominiklohmann: I fixed this, thanks!

web/docs/understand/query-language/operators/head.md
mavam: Great, thanks!

libvast/test/pipeline.cpp
dominiklohmann: The parser will fail with negative numbers, I can add a test for this. Is `head 0` or `taste 0` really an error, though? The current behavior is that it just always yields nothing.

Dakostu: `head 0 ` is not an error to me, it just shows nothing (which is logically correct behavior), ie. `results.empty()`.
It just helps to have such cases specified in unit tests.

dominiklohmann: Okay, will add them.

dominiklohmann: I believe I covered everything in https://github.com/tenzir/vast/pull/2891/commits/916060881347d0e5a0978c185e4c3f068312bc85 – mind taking another look?

Dakostu: Yes, it's all good now. Thank you.

PR #2892, opened 2023-01-29
Remove leftover mentions of old names of basic types

tobim:
We still had some uses of the old type names `count`, `real`, and `addr` in the docs. These aren't really explained any more so they can be confusing.

### DIFF
--- a/web/docs/understand/formats/zeek.md
+++ b/web/docs/understand/formats/zeek.md
@@ -63,10 +63,9 @@ vast import zeek < conn.log
 ```
 
 :::info type mapping
-The `import zeek` command implicitly
-[aliases](/docs/understand/data-model/type-system.md) the types `count`, `real`,
-and `addr` from the Zeek TSV `#types` header line to VAST's basic types
-`uint64`, `double`, and `ip`.
+The `import zeek` command maps the types `count`, `real`, and `addr` from the
+Zeek TSV `#types` header line to VAST's basic types `uint64`, `double`, and
+`ip`.
 :::
 
 When Zeek [rotates logs][zeek-log-rotation], it produces compressed batches of
--- a/web/docs/understand/formats/zeek.md
+++ b/web/docs/understand/formats/zeek.md
@@ -62,8 +62,15 @@ You can import this log as follows:
 vast import zeek < conn.log
 ```
 
+:::info type mapping
+The `import zeek` command implicitly
+[aliases](/docs/understand/data-model/type-system.md) the types `count`, `real`,
+and `addr` from the Zeek TSV `#types` header line to VAST's basic types
+`uint64`, `double`, and `ip`.
+:::
+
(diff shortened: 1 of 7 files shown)


web/docs/understand/formats/zeek.md
```diff
@@ -62,8 +62,15 @@ You can import this log as follows:
 vast import zeek < conn.log
 ```
 
+:::info type mapping
+The `import zeek` command implicitly
+[aliases](/docs/understand/data-model/type-system.md) the types `count`, `real`,
+and `addr` from the Zeek TSV `#types` header line to VAST's basic types
+`uint64`, `double`, and `ip`.
```
tobim, line 69: Note to the reviewers: This is actually not the case right now, but I think it should be.
What happens instead is that the types are just rewritten to the corresponding basic types instead of aliases.

mavam: Are you saying that the Zeek reader should create `count`, `real`, and `addr` aliases? I think that's actually a good idea. This way, Zeek users could writes queries with `:addr`, which is both idiomatic and implicitly restricted to Zeek data.

mavam (approved):

web/docs/understand/formats/zeek.md
dominiklohmann: This is impossible as long as we're still supporting the old type names for backwards compatibility.

This comment in the docs should be removed for now, I think.

web/docs/understand/formats/zeek.md
```diff
@@ -62,8 +62,15 @@ You can import this log as follows:
...
+:::
+
 When Zeek [rotates logs][zeek-log-rotation], it produces compressed batches of
-`*.tar.gz` regularly. If log freshness is not a priority, you could trigger an
+`*.log.gz` regularly. If log freshness is not a priority, you could trigger an
```
dominiklohmann, line 73:
```suggestion
`*.tar.gz` regularly. If log freshness is not a priority, you could trigger an
```

I believe this was already correct.

tobim: Zeek creates `gz` not `tar.gz`.

dominiklohmann: Nevermind then, I must've misremembered.

web/docs/understand/formats/zeek.md
tobim: I still think it would be good to be explicit about the relationship between Zeek and VAST types to avoid uncertainty for the reader.

tobim: I think we should still change the zeek reader after the deprecation period, but I applied the change for now.

dominiklohmann: I agree! But right now it's too much of a conflict, and I wouldn't want to bring out another Threat Bus release just for this.

Issue #2600, opened 2022-09-30
Implement RFC-001: Composable Pipelines


dominiklohmann:
### Description

This issue is about implementing [RFC-011: Composable Pipelines](https://github.com/tenzir/vast/tree/master/rfc/001-composable-pipelines). In the spirit of continuing the discussion out in the open, we want to lay out our plan for the implementation after having done a feasibility analysis on some designs we considered. Here's what we ended up with:

#### Element Type

The element type defines how operators can be combined. For starters, we want to support the element types *Void* and *Arrow*, but have the code written with extensibility in mind. Element types define whether they require a schema.

#### Logical and Physical Operators

A logical operator has a known input and output element type, and is able to instantiate a physical operator with an interface depending on the input and output type for a given input schema. The input schema may be null if the input type does not require a schema.

A physical operator is instantiated from a logical operator and an input schema, and maps an input element generator to an output element generator.

#### Pipelines and Plans

A pipeline is an ordered sequence of logical operators with matching input and output types along the sequence. The pipeline's input type is the input type of the first operator in the sequence, and the pipeline's output type is the output type of the last operator in the sequence. A pipeline is by definition a logical operator.

A plan is a pipeline whose input and output types are *Void*.

#### Logical Operator Plugin

A logical operator plugin creates a logical operator from its textual representation.

When parsing a pipeline `op1 [args...] | op2 [args...] | ...` VAST looks for the `op1` and `op2` logical operator plugins and passes them their respective argument string to instantiate a logical operator. For example, `where '#type == "suricata.flow"'` could instantiate the logical `where` operator with the expression `#type == "suricata.flow"`. 

The exact syntax of the individual logical operators is considered an implementation detail that we will discuss further in review.

#### Executor

An executor takes ownership of a plan and executes it in its (potentially asynchronous) execution model. The executor must execute a logical operator on only one thread at a time, but may concurrently access physical operator instantiations from one logical operator. The executor is responsible for transferring ownership of elements between logical operators by using thread-safe queues, e.g., an actor mailbox when using the actor runtime as an execution model.

In the future, we may consider relaxing the requirement not to execute physical operators instantiations from one logical operator in parallel if the logical operator allows for it.

### Definition of Done

- [ ] We have implemented and tested the proposed pipeline executor.
- [ ] All existing pipeline operators are converted to the new plugin API.
[...]

PR #2896, opened 2023-01-31
Remove the transformer actor

Dakostu:
https://github.com/tenzir/vast/issues/2600 makes the `transformer` actor obsolete, so here is a quick removal (with the `table_slice` transformation responsibility moved to the components that formerly used the `transformer`).

Fixes tenzir/issues#11


libvast/src/system/importer.cpp
```diff
@@ -40,23 +40,33 @@ namespace vast::system {
...
-  void process(caf::downstream<detail::framed<table_slice>>& out,
+  void process(caf::downstream<table_slice>& out,
                std::vector<table_slice>& slices) override {
     VAST_TRACE_SCOPE("{}", VAST_ARG(slices));
     uint64_t events = 0;
     auto t = timer::start(state.measurement_);
     for (auto&& slice : std::exchange(slices, {})) {
+      auto import_ok = state.executor.add(std::move(slice));
+      if (import_ok) {
+        VAST_WARN("importer can not add slice to pipeline executor - data "
```
patszt, line 60: Could we also print the import_ok error here?

libvast/src/system/importer.cpp
```diff
@@ -40,23 +40,33 @@ namespace vast::system {
...
+                  "loss");
+      }
+    }
+    auto transformed_slices = state.executor.finish();
+    if (!transformed_slices) {
```
patszt, line 65: Should we continue in such case? Later on the transformed_slices are dereferenced. I think this would crash when bool operator fails for caf::expected.

libvast/src/system/source.cpp
```diff
@@ -139,15 +139,31 @@ void source_state::filter_and_push(
...
                  reader->name(), filtered_slice->rows(), unfiltered_rows,
                  slice.schema());
-      push_to_out(std::move(*filtered_slice));
+      auto import_ok = executor.add(std::move(*filtered_slice));
+      if (import_ok) {
+        VAST_WARN("source can not add slice to pipeline executor - data "
+                  "loss");
+      }
+      auto transformed_slices = executor.finish();
+      for (auto&& slice : *transformed_slices) {
```
patszt, line 148: I think we should also check if the expected is set.
If not - print a warning and return.

changelog/unreleased/bug-fixes/2896--remove-the-transformer-actor.md
```diff
@@ -0,0 +1,2 @@
+The transformer actor does not exist anymore. This resolves potential shutdown
+issues when using pipelines that contain aggregations.
```
patszt, line 2: Hmm i wonder if we really need a changelog entry for this. Could someone more advanced comment on this? I think this issue might have been invisible to the users so we might as well just not bring it up.

libvast/src/system/importer.cpp
Dakostu: I added an `import_ok` print in https://github.com/tenzir/vast/pull/2896/commits/e9bfda4fb7575c59b37e284cf8b05d763d403c7e

libvast/src/system/source.cpp
Dakostu: I applied the same `transformed_slices` checks to the `source` in https://github.com/tenzir/vast/pull/2896/commits/75e61108bb2d5a492bcb87bc207a2b603874d1b9

libvast/src/system/importer.cpp
tobim: `import_ok` implies the opposite meaning of what it actually carries. As a reader I'm very confused by line 59 which I can only parse as "is the import ok? then print a warning".
How about changing this to
```
if (auto err = state.executor.add(std::move(slice))
  VAST_ERROR(...);
```?

Comment (Dakostu) [reply libvast/src/system/importer.cpp]: That's true - now I wonder where that name originally came from... probably from when I thought `executor::add()` would return a `caf::expected` instead of a `caf::error`. I renamed the variable in https://github.com/tenzir/vast/pull/2896/commits/1cf34b766a041ca1349e717b5121fc397061c2f3

File: changelog/unreleased/bug-fixes/2896--remove-the-transformer-actor.md (hunk 4 above)
Comment (tobim) [reply changelog/unreleased/bug-fixes/2896--remove-the-transformer-actor.md:2]: If the mentioned shutdown issues are potential in the sense of assumed, then I would indeed not bring it up in the changelog. I would also refrain from mentioning particular actors in these messages, as they are not elements that the reader understands. As a reader I would also like to see something akin to
```
Pipelines containing aggregations do not prevent VAST processes from terminating any more.
```
But again, only if the bug was actually confirmed.

Comment (dominiklohmann) [reply changelog/unreleased/bug-fixes/2896--remove-the-transformer-actor.md:2]: Instead of aggregations it is actually pipelines reducing the number of total events, but only when `vast.export.max-events` was set.

Comment (tobim) [reply changelog/unreleased/bug-fixes/2896--remove-the-transformer-actor.md:2]: So a better message would be:
```
Pipelines that reduce the number of events do not prevent `vast export` processes that have a `max-events` limit from terminating any more.
```

File: libvast/src/system/importer.cpp (hunk 2 above)
Comment (patszt) [reply libvast/src/system/importer.cpp:65]: I did mean the keyword continue :D
The else achieves the same, maybe it is a bit less readable for me but i am fine with both.

**Review by patszt — APPROVED** [review]
Code looks good. Just requires the agreement on the changelog side and i think it is good to go

File: libvast/src/system/source.cpp [hunk 5]
```diff
@@ -168,10 +190,10 @@ source(caf::stateful_actor<source_state>* self, format::reader_ptr reader,
...
   self->state.done = false;
-  self->state.transformer
-    = self->spawn(transformer, "source-transformer", std::move(pipelines));
-  if (!self->state.transformer) {
-    VAST_ERROR("{} failed to spawn transformer", *self);
+  self->state.executor = pipeline_executor{std::move(pipelines)};
+  if (auto err = self->state.executor.validate(
+        pipeline_executor::allow_aggregate_pipelines::yes)) {
+    VAST_ERROR("source contains invalid pipeline {}", err);
     self->quit();
```
dominiklohmann, line 197: One minor design note on this while I'm seeing it (even though you didn't this logic): In general we always want to print errors as late as possible because that allows us to either handle the error, or to further enrich it with context before printing it.

The pattern here will lead to a VAST client that exits with a zero exit code, indicating a successful exit, because the error isn't "returned" from the actor. To achieve that, you usually want to do the following instead:

```c++
self->quit(caf::make_error(...));
return my_actor::behavior_type::make_empty_behavior(); // or just {} for untyped actors
```

Actors that monitor the source (i.e., the import command's scoped actor) will then receive a down message from the source with exactly this error, and can render it before shutting down.

changelog/unreleased/bug-fixes/2896--remove-the-transformer-actor.md
Dakostu: I changed the changelogmessage to what you suggested in https://github.com/tenzir/vast/pull/2896/commits/b8a033309998307e58eea090684c888c146ec244.

Regarding the question whether this should be included at all: I don't know whether users in our target audiences have reported this bug, but I am also some kind of user and I discovered this bug quite easily.

libvast/src/system/source.cpp
dominiklohmann: Perfect, thanks!

Dakostu:
If there are no objections from you @dominiklohmann @tobim I will merge this by 15:00 CET. I will merge earlier if you approve the PR.

dominiklohmann (approved):

dominiklohmann:
Go ahead, I think this is ready to go!

PR #2897, opened 2023-01-31
Fix package generation with bundled CAF

dominiklohmann:
Building VAST with a bundled CAF failed to correctly install the CAF libraries, causing CPack-created packages to require CAF to be installed separately.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -792,6 +792,7 @@ else ()
            "\nmark_as_advanced(caf_build_header_path)")
     set(CAF_FOUND true)
   endif ()
+  # Make bundled CAF available for component-based CPack installations.
   install(TARGETS libcaf_core libcaf_openssl libcaf_io COMPONENT Runtime)
   install(TARGETS libcaf_test COMPONENT Development)
 endif ()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -792,6 +792,8 @@ else ()
            "\nmark_as_advanced(caf_build_header_path)")
     set(CAF_FOUND true)
   endif ()
+  install(TARGETS libcaf_core libcaf_openssl libcaf_io COMPONENT Runtime)
+  install(TARGETS libcaf_test COMPONENT Development)
```
tobim, line 796: I was confused by this initially until I realized this is for adding the target related build artifacts to the install components that are used for our packaging. Can you add a comment for the next reader? You can just take my wording for it.

I was also wondering what happens if a target is part of multiple install components?

tobim (approved):

CMakeLists.txt
dominiklohmann:
```suggestion
  # Make bundled CAF available for component-based CPack installations.
  install(TARGETS libcaf_core libcaf_openssl libcaf_io COMPONENT Runtime)
  install(TARGETS libcaf_test COMPONENT Development)
```

dominiklohmann: > I was also wondering what happens if a target is part of multiple install components?

This is gracefully handled by CMake, the deduplication happens based on the target path afaik.

PR #2898, opened 2023-01-31
Use CAF streaming in export command

patszt:
The previous send and receive approach resulted in the export command receiving errors from caf::middleman after the CAF 0.18.6 upgrade.

This issue doesn't occur with the streaming approach.
The shift to the streaming complicates the implementations of explore and pivot commands. It was agreed to remove them and reintroduce them in the future as pipeline operators.

The export max events management is already done by the sink so it was removed from the exporter responsibilities.

TODO:
- [x] Squash commits
- [x] changelog
- [x] check docs


tobim:
> The export max events management is already done by the sink so it was removed from the exporter responsibilities.

I would like you to bring that back. Consider the case of a catch-all export or query like `#type == "zeek.conn"`. The default number of initial (taste) partitions is `5`, that means you would quickly load the data of 5 partitions onto the middleman, creating a lot of work just to throw the data away in the other process.

patszt:
> > The export max events management is already done by the sink so it was removed from the exporter responsibilities.
> 
> I would like you to bring that back. Consider the case of a catch-all export or query like `#type == "zeek.conn"`. The default number of initial (taste) partitions is `5`, that means you would quickly load the data of 5 partitions onto the middleman, creating a lot of work just to throw the data away in the other process.

Good point, i would still want to elaborate if the additional work is a big deal.
The exporter receives table slices one by one. Once we are at the event limit we will still receive them, but they will be ignored. We don't send any incoming table_slices to the sink if it is past the limit.

In the PR we will send the slice that is past the limit and the sink should should quit after it processes it.
The down message from sink will also quit the exporter. The additional work can be done in between sending the last over limit slice and receiving quit message.
So we may serialize/deserialize more slices than on the current master, but will it be much more in the practice? I would like to believe it to not be relevant, but benchmarks are the only answer.

tobim:
> The down message from sink will also quit the exporter. The additional work can be done in between sending the last over limit slice and receiving quit message.

Yes, I would assume that can be a lot. In my mental model the input buffer at the middleman would be full, and the one at the sink likely too (depending on the speed of the formatter).

dominiklohmann:
We cannot bring that limit in the exporter back. It is simply incompatible with pipelines, as the only position at which you can enforce such a limit is at the very end of the pipeline.

mavam:
Also, isn't the limit simply the `head` operator?

dominiklohmann:
In the future, yes, but for now not yet. This is because operators cannot stop the stream early currently, which a properly implemented head operator should do. Once we have that we can remove the max events option in its entirety.

tobim:
> We cannot bring that limit in the exporter back. It is simply incompatible with pipelines, as the only position at which you can enforce such a limit is at the very end of the pipeline (which is the sink).

We know the full pipeline beforehand. In the common case there won't be any more client side transformations, so I don't get how limiting cutting the stream from the exporter should not be possible.

dominiklohmann:
You are ignoring the fact that pipelines can also be specified in configuration currently, which you cannot know about in the importer, thus causing a hang.

I think you are introducing a worry about a potential overhead here that you have not proven to actually matter, and that in any case will be entirely irrelevant in a few months time because the entire export will just be part of one pipeline instead of us having multiple pipelines executed at server or sink side as part of an export command. Let's focus on getting there instead of optimizing this intermediate version.

dominiklohmann (approved):

PR #2899, opened 2023-01-31
Update response format of the /export endpoint

lava:
* Renamed the 'num_events' field to 'num-events', in accordance with all other VAST field names.
* Remove the 'version' field. The version information is still accessible from the /status endpoint.

### DIFF
--- a/libvast/builtins/endpoints/export.cpp
+++ b/libvast/builtins/endpoints/export.cpp
@@ -92,7 +92,7 @@ static auto const* SPEC_V0 = R"_(
             schema:
                 type: object
                 properties:
-                  num_events:
+                  num-events:
                     type: int64
                   version:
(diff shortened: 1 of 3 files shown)


tobim (approved):
`num-events` is a trivial fix.
I fully agree that the `version` field was misplaced in this response.

libvast/builtins/endpoints/export.cpp
```diff
@@ -370,7 +366,6 @@ std::string format_result_flat(const std::vector<table_slice>& slices,
 // clang-format off
 //
 // {
-//   "version": "v2.4-rc2",
 //   "num_events": 3,
```
tobim, line 369:
```suggestion
//   "num-events": 3,
```

PR #2901, opened 2023-01-31
Do not warn about missing indexes

dominiklohmann:
Partition index creation can selectively be disabled in configuration, and doing so should not lead to warnings when attempting to access data. This PR achieves that by simply not attempting to deserialize non-existing indexes.

Additionally, this fixes the error message in case the deserialization of an existing index fails and changes it from an error to a warning, as VAST already handles this gracefully.

Before:

```
[17:48:32.267] user.scheduled-actor-23 failed to deserialize value index at 5
```

After:

```
[17:51:37.517] passive-partition (834e31a1-8e25-49be-8a8d-fd1e7eda27a9) failed to deserialize value index for field suricata.ftp.dest_ip
```

I don't think this is worth a changelog entry, as it's a really minor thing.

### DIFF
--- a/libvast/src/system/passive_partition.cpp
+++ b/libvast/src/system/passive_partition.cpp
@@ -166,7 +166,9 @@ indexer_actor passive_partition_state::indexer_at(size_t position) const {
     indexer = self->spawn(passive_indexer, id, std::move(value_index));
     return indexer;
   }
-  VAST_ERROR("{} failed to deserialize value index at {}", *self, position);
+  VAST_WARN("passive-partition ({}) failed to deserialize value index for "
+            "field {}",
+            id, qualified_index->field_name()->string_view());
   return {};
 }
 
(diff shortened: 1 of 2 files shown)


libvast/src/system/passive_partition.cpp
```diff
@@ -161,12 +161,16 @@ indexer_actor passive_partition_state::indexer_at(size_t position) const {
...
+    return {};
   if (auto value_index
       = unpack_value_index(*qualified_index->index(), *container)) {
     indexer = self->spawn(passive_indexer, id, std::move(value_index));
     return indexer;
   }
-  VAST_ERROR("{} failed to deserialize value index at {}", *self, position);
+  VAST_WARN("passive-partition ({}) failed to deserialize value index for "
+            "field {}",
+            id, qualified_index->field_name()->string_view());
```
tobim, line 173: `unpack_value_index` itself prints errors in case the binary or legacy deserializer fails, does it really make sense to print a second warning in that case?

dominiklohmann: I would rather consider that warning to be misplaced, as it's impossible to investigate if you don't know which partition the failure originated from. Should we remove that one?

dominiklohmann: Done. Mind approving the PR if that is fine with you?

tobim (approved):

PR #2903, opened 2023-02-01
Process table slices in stream handler for source

patszt:
This change is introduced to fix the quitting import commands that is visible on the testbed.
These changes were running for about a week on the testbed and the issue never reproduced. It also made the import throughput graph more stable.

The idea is to respect what the CAF wants. The stream handler is called with 10000 events -> we push the 10000 events.
Previously the source actor delegated providing these events in it's own internal handler which increased the backpressure.


tobim (approved):
In addition to fixing a bug this also simplifies to logic quite a bit. Glad to see this resolved!

PR #2906, opened 2023-02-02
Fix rounding issue when printing duration types

patszt:
The values with trailing '9's were sometimes rounded incorrectly. E.g 1.999 was displayed as 1.1 instead of 2.0.

The changes in integration test references were checked manually against the input data. From my analysis all the overridden values are correctly displayed now.


libvast/include/vast/concept/printable/std/chrono.hpp
```diff
@@ -72,27 +74,43 @@ struct duration_printer : printer_base<duration_printer<Rep, Period, Policy>> {
...
+  print_adaptive(auto& out, double duration, std::string_view suffix) {
+    auto decimal = double{0.0};
+    const auto fractional = std::modf(duration, &decimal);
+    if (fractional == 0.0) {
+      fmt::format_to(out, "{:.1f}{}", duration, suffix);
+      return true;
+    }
+    auto double_str = fmt::format("{:.2f}", duration);
+    if (double_str.back() == '0')
+      double_str.pop_back();
```
dominiklohmann, line 88: If I understand correctly then this works because the `:.2f` format specifier guarantees that there are exactly two trailing zeroes, so we must never remove more than one.

Do we then need the `fractional == 0.0` section above? If I understand correctly then L80-85 are not necessary in this function.

changelog/unreleased/bug-fixes/2906--proper-display-of-duration-types.md
```diff
@@ -0,0 +1,6 @@
+Duration types with trailing '9' digits are now displayed correctly. The
+duration of 1.999 seconds was displayed as "1.1" and now it is shown as "2.0".
+VAST at most displays two digits after decimal point for duration types and
+this causes the values to be rounded respectively before being displayed.
+Previously VAST rounded down digits from 0 - 4 and now it does that from the
+0 - 5 range.
```
dominiklohmann, line 6: Here's a suggestion for a rephrasing.

> We fixed incorrect printing of human-readable durations in some edge cases. E.g., the value 1.999s was rendered as 1.1s instead of the expected 2.0s. This bug affected the JSON and CSV export formats, and all durations printed in log messages or the status command.

I wouldn't describe how VAST renders and rounds the data. For the user-facing changelog what's relevant is what this affected and that it is now fixed.

dominiklohmann (approved):
Thanks for finding and fixing this one. Just one smaller question on the code.

libvast/include/vast/concept/printable/std/chrono.hpp
patszt: It is not needed. Thanks for pointing this out.

PR #2908, opened 2023-02-02
Add a workaround to fix CAF OpenSSL options

tobim:
In the `actor_system_config`, some settings are not like the others. While most modules read their options directly from the `content` object, the SSL configuration is stored in member variables of the main config class. The CAF command line and config file parsers handle this correctly, but settings from `vast.yaml` files or those that are passed in as environment variables are injected through a different code path.

This commit adds assignments to set the aforementioned class members explicitly.

I considered the alternative solution of translating non CAF-native option formats to native ones and let the `parse` function handle it, but that approach would be a large amount of work rather little benefit.

### DIFF
--- a/changelog/unreleased/bug-fixes/2908--fix-openssl-configuration.md
+++ b/changelog/unreleased/bug-fixes/2908--fix-openssl-configuration.md
@@ -1,0 +1,2 @@
+Options passed in the `caf.openssl` section in the configuration file or as
+`VAST_CAF__OPENSSL__*` environment variables are no longer ignored.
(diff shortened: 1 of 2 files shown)


dominiklohmann (approved):
The fix makes sense to me, and I agree it's okay to fix this up afterwards instead of swapping the order. Should we ask upstream why the SSL options are treated this way rather than being read from the `content` settings object?

lava (approved):

tobim:
> Should we ask upstream why the SSL options are treated this way rather than being read from the content settings object?

I asked about it in the chat.

PR #2910, opened 2023-02-03
Enable pipeline export with max events

patszt:
After removal of max events handling in the exporter it is now possible to use export pipelines with max number of events.


libvast/test/system/exporter.cpp
```diff
@@ -87,9 +87,8 @@ struct fixture : fixture_base {
   }
 
   void spawn_exporter(query_options opts) {
-    exporter
-      = self->spawn(system::exporter, expr, opts, defaults::export_::max_events,
-                    std::vector<vast::pipeline>{}, index);
+    exporter = self->spawn(system::exporter, expr, opts,
+                           std::vector<vast::pipeline>{}, index);
```
dominiklohmann, line 91: (Not really related to this line, just wanted a comment that can be checked off.)

Can you add an integration that that runs `vast export -n 2 json 'head 1'` with a database that contains at least 3 events? Because that would never have quit before this change.

dominiklohmann (dismissed):
Perfect, thank you! Just one small request for a test.

libvast/test/system/exporter.cpp
patszt: Good idea, i was testing it with where operator to reduce the amount of events.
Seems that this hangs with the suggested command :(

In the end exporter waits for sink to finish. Exporter has all data from index so it waits for sink. Sink received only 1 event instead so it still waits for more. For the historical query the exporter knows if it will receive more table slices or not. It is possible for exporter to send some end of input message to the sink. Not sure if it is worth implementing though. We are about to rewrite this logic anyways.

dominiklohmann: Okay, then this requires further investigation. I've dismissed my approval for now. Let's take a look at this together next week, but let's also be cautious about the time we spend on this bug.

dominiklohmann: This is all that's needed:

```diff
diff --git a/libvast/src/system/exporter.cpp b/libvast/src/system/exporter.cpp
index 380580777..489614b3d 100644
--- a/libvast/src/system/exporter.cpp
+++ b/libvast/src/system/exporter.cpp
@@ -97,6 +97,8 @@ void ship_results(exporter_actor::stateful_pointer<exporter_state> self,
                transformed.error());
     return;
   }
+  if (transformed->empty())
+    return;
   if (!self->state.source) [[unlikely]]
     attach_stream(self);
   for (auto& t : *transformed)
@@ -374,6 +376,8 @@ exporter(exporter_actor::stateful_pointer<exporter_state> self, expression expr,
               {"query", fmt::to_string(self->state.query_context.id)}});
         if (!self->state.source)
           self->send_exit(self->state.sink, caf::exit_reason::user_shutdown);
+        else
+          shutdown_stream(self->state.source);
       }
     },
   };
diff --git a/vast/integration/reference/export-shutdown-behavior/step_00.ref b/vast/integration/reference/export-shutdown-behavior/step_00.ref
new file mode 100644
index 000000000..e69de29bb
diff --git a/vast/integration/reference/export-shutdown-behavior/step_01.ref b/vast/integration/reference/export-shutdown-behavior/step_01.ref
new file mode 100644
index 000000000..e69de29bb
diff --git a/vast/integration/reference/export-shutdown-behavior/step_02.ref b/vast/integration/reference/export-shutdown-behavior/step_02.ref
new file mode 100644
index 000000000..d00491fd7
--- /dev/null
+++ b/vast/integration/reference/export-shutdown-behavior/step_02.ref
@@ -0,0 +1 @@
+1
diff --git a/vast/integration/reference/export-shutdown-behavior/step_03.ref b/vast/integration/reference/export-shutdown-behavior/step_03.ref
new file mode 100644
index 000000000..d00491fd7
--- /dev/null
+++ b/vast/integration/reference/export-shutdown-behavior/step_03.ref
@@ -0,0 +1 @@
+1
diff --git a/vast/integration/reference/export-shutdown-behavior/step_04.ref b/vast/integration/reference/export-shutdown-behavior/step_04.ref
new file mode 100644
index 000000000..e69de29bb
diff --git a/vast/integration/reference/export-shutdown-behavior/step_05.ref b/vast/integration/reference/export-shutdown-behavior/step_05.ref
new file mode 100644
index 000000000..e69de29bb
diff --git a/vast/integration/vast_integration_suite.yaml b/vast/integration/vast_integration_suite.yaml
index cf10ed35b..7066b9969 100644
--- a/vast/integration/vast_integration_suite.yaml
+++ b/vast/integration/vast_integration_suite.yaml
@@ -747,3 +747,17 @@ tests:
       - command: export json
       - command: export arrow
         transformation: python @./misc/scripts/print-arrow-batch-size.py
+
+  Export shutdown behavior:
+    fixture: ServerTester
+    tags: [import, export]
+    steps:
+      - command: export json
+      - command: import --blocking suricata
+        input: data/suricata/eve.json
+      - command: export --max-events=2 json 'head 1'
+        transformation: jq -ers 'length'
+      - command: export json 'head 1'
+        transformation: jq -ers 'length'
+      - command: export --max-events=1 json 'head 0'
+      - command: export json 'head 0'
```

patszt: Thanks for the implementation. I wonder why i didn't see that "end of input message to the sink" is the same as closing the stream :D

dominiklohmann: No worries, I just looked at this again yesterday and thought "this cannot be so complicated."

I'll approve again under the assumption that you'll incorporate my diff or some variant of it. :)

dominiklohmann: No worries, I just looked at this again yesterday and thought "this cannot be so complicated."

I'll approve again under the assumption that you'll incorporate my diff or some variant of it. :)

dominiklohmann (approved):

libvast/test/system/exporter.cpp
patszt: Shutting down stream on the last atom::done message can close the stream when it didn't pull all the events yet.
Adding the if (results.empty()) before shutdown_stream in your diff seemed to work. One of the simpler integration test failed so i started looking a bit more into this issue. I removed all the changes and i don't understand why the client is stuck on 
vast export -n 2 json 'head 1'

The stream was started, it pushes all events. The stream says it should shut down on the third handler, but it doesn't shut down.

libvast/src/system/exporter.cpp
```diff
@@ -73,9 +73,12 @@ void attach_stream(exporter_actor::stateful_pointer<exporter_state> self) {
...
           if (continuous)
             return false;
-          return state.self_ptr->state.query_status.received
-                   == state.self_ptr->state.query_status.expected
-                 && state.self_ptr->state.results.empty();
+          auto should_end = state.self_ptr->state.query_status.received
+                              == state.self_ptr->state.query_status.expected
+                            && state.self_ptr->state.results.empty();
+          if (should_end)
+            shutdown_stream(state.self_ptr->state.source);
```
patszt, line 80: Based on the CAF header files documentation this shouldn't cause any weird issues.

Should we wait on neverlord's comment before merging it? @dominiklohmann

dominiklohmann: I think we can merge this as-is. The reply will help regardless in improving our understanding.

PR #2911, opened 2023-02-03
Set default values to new CAF 0.18.6 streaming options

patszt:
After short experimenting on the testbed these values suggested by @dominiklohmann seem to work.
Increasing the batch size to 2 seemed to occasionally cause a spike in the throughput. Sometimes it takes longer for the stream to buffer two tokens and send them together.

Changing the buffer size didn't seem to affect the import throughput.
A value of 64 corresponds to the maximum number of events that a one partition can hold.

### DIFF
--- a/libvast/src/system/default_configuration.cpp
+++ b/libvast/src/system/default_configuration.cpp
@@ -50,6 +50,9 @@ default_configuration::default_configuration() {
   // We do a combination of (1) and (2).
   set("caf.stream.max-batch-delay", caf::timespan{15ms});
   set("caf.scheduler.max-throughput", 500);
+  set("caf.stream.credit-policy", "token-based");
+  set("caf.stream.token-based-policy.batch-size", 1);
+  set("caf.stream.token-based-policy.buffer-size", 64);
 }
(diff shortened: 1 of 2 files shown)


dominiklohmann (approved):
The graphs on the testbed look stable enough that I have high confidence in these values.

libvast/src/system/default_configuration.cpp
```diff
@@ -50,6 +50,9 @@ default_configuration::default_configuration() {
   // We do a combination of (1) and (2).
   set("caf.stream.max-batch-delay", caf::timespan{15ms});
   set("caf.scheduler.max-throughput", 500);
+  set("caf.stream.credit-policy", "token-based");
+  set("caf.stream.token-based-policy.batch-size", 1);
+  set("caf.stream.token-based-policy.buffer-size", 64);
```
dominiklohmann, line 55: Can you add these modified defaults also to the vast.yaml.example file?

patszt: Yup, good idea

PR #2913, opened 2023-02-03
Remove default scheduler.max-threads value for client commands

Dakostu:
TBD


Dakostu:
Closing this as this simple removal approach turns out to be too naive.

PR #2915, opened 2023-02-05
Upgrade docusaurus to ^2.3.1 and update all yarn packages

dit7ya:
All yarn packages were upgraded via the following commands

```
yarn upgrade
syncyarnlock
yarn install
```

No version have been upgraded with breaking changes (major semver is preserved).

https://github.com/vasilevich/sync-yarnlock-into-packagejson

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -118,7 +118,7 @@ RUN apt-get update && \
     apt-get -y --no-install-recommends install \
       ./apache-arrow-apt-source-latest-$(lsb_release --codename --short).deb && \
     apt-get update && \
-    apt-get -y --no-install-recommends install libarrow1000 libparquet1000 && \
+    apt-get -y --no-install-recommends install libarrow1100 libparquet1100 && \
     rm -rf /var/lib/apt/lists/*
 
 USER vast:vast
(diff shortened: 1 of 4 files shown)


mavam (approved):
LGTM.

@dominiklohmann may have an idea how to fix the regression test failures in CI.

dominiklohmann:
> @dominiklohmann may have an idea how to fix the regression test failures in CI.

I'll forward that question to @rdettai, the failure makes little sense to me.

rdettai:
when running 
```
docker run --pull=always --rm --name vast-regression -e VAST_CONSOLE_VERBOSITY=verbose -v vast-regression:/var/lib/vast/ ghcr.io/tenzir/vast:topic-docusaurus-upgrade start
```

I get
```
vast: error while loading shared libraries: libarrow.so.1100: cannot open shared object file: No such file or directory
```

It seems that the docker image build itself is broken, and as they are no tests in the `docker-vast`  job itself and the `docker-compose` job is skipped, it bubbles up only here.

dominiklohmann:
Ah, I figured it out. @dit7ya do you mind if I push the fix directly onto this?

PR #2916, opened 2023-02-06
Set up the Community Edition

tobim:
The Community Edition bundles OSS VAST with proprietary plugins. These are developed in a private repository that needs to be cloned over SSH, meaning only Tenzir staff and CI will be able to build `vast-ce` and `vast-ce-static`.

Implementation wise we use `builtins.fetchGit` which acts during evaluation time so Nix can clone the repo wit the current users credentials.

This also adds a `withPlugins` function to the vast package definition so that we can build VAST with different plugin lists easily.
Usage example:

```nix
vast.withPlugins (ps: [ compaction matcher ])
```

All plugins that are developed in the open source VAST repo are always enabled, there is no way to opt out yet.

This PR also adds version calculation logic to `flake.nix`, so that pure builds report their correct version:

```bash
$ nix run 'git+https://github.com/tenzir/vast?ref=topic/flake-version' -- -q version
{
  "VAST": "v2.4.1-936-gf5b87c2136",
  "Build Configuration": {
    "Type": "Release",
    "Tree Hash": "9e809707b1744cf198c0c1246fb6b2ac",
    "Assertions": false,
    "Address Sanitizer": false,
    "Undefined Behavior Sanitizer": false
  },
  "CAF": "0.18.6",
  "Apache Arrow": "10.0.0",
  "jemalloc": "5.3.0-0-g54eaed1d8b56b1aa528be3bdd1877e59c56fa90c",
  "plugins": {
    "cef": "v0.1.0",
    "parquet": "v1.0.0",
    "pcap": "v1.1.0",
    "sigma": "v1.1.0",
    "web": "v1.0.0"
  }
}
```

This comes with the caveat that it only works with `git+https://github.com/tenzir/vast`, but not when the repo is addressed as `github:tenzir/vast`.

TODO:
- [x] Add the repo that holds the proprietary plugins as a regular sub module
- [x] Add a CI Job to build and upload the community edition
- [x] Onboard the remaining plugins that should be included in the CE
- [x] Avoid the need to repeat the listing VAST dependencies in the plugin package 
- [x] Make it possible to run `nix/update.sh` without a local Nix installation


dominiklohmann:
Question: Wouldn't it be better to keep this logic outside of Nix such that it works for the regular build and for our Docker builds as well?

tobim:
> Question: Wouldn't it be better to keep this logic outside of Nix such that it works for the regular build and for our Docker builds as well?

That is what the first TODO is about.

dominiklohmann:
> That is what the first TODO is about.

Makes sense then. Thanks! So this implies the dependency inversion between our own plugins and VAST?

tobim:
> So this implies the dependency inversion between our own plugins and VAST?
 
Yes.

tobim:
resolved tenzir/issues#102

cmake/VASTRegisterPlugin.cmake
```diff
@@ -272,11 +272,7 @@ macro (VASTInstallExampleConfiguration target source prefix destination)
 
   # Set a temporary variable for the example dir location. Because we're in a
   # macro we're unsetting the variable again later on.
-  if (VAST_CMAKE_INSTALL_DOCDIR)
-    set(_example_dir "${VAST_CMAKE_INSTALL_DOCDIR}/examples")
-  else ()
-    set(_example_dir "${CMAKE_INSTALL_DOCDIR}/examples")
-  endif ()
```
dominiklohmann, line 279: Why do we need this change? Can we also remove VAST_CMAKE_INSTALL_DOCDIR from VASTConfig.cmake.in?

nix/update.sh
```diff
@@ -51,3 +56,21 @@ update-source() {
...
+vast_plugins_json="$(jq --arg rev "${vast_plugins_rev#rev}" \
+  '."rev" = $rev' "${dir}/vast/plugins/source.json")"
+echo -E "${vast_plugins_json}" > "${dir}/vast/plugins/source.json"
+
+echo "Extracting plugin versions..."
+{
+  printf "{\n"
+  for cm in contrib/vast-plugins/*/CMakeLists.txt; do
+    plugin="$(basename "$(dirname "$cm")")"
+    version="$(sed -n 's|[^(]VERSION\s\+\([0-9A-Za-z\.-_]\+\)$|\1|p' "$cm" | tr -d '[:space:]')"
```
dominiklohmann, line 71: Please add a comment on what this does.

nix/vast/plugins/versions.nix
```diff
@@ -0,0 +1,6 @@
+{
+  compaction = "1.4.0";
+  inventory = "1.0.0";
+  matcher = "3.0.0";
+  netflow = "1.0.0";
+}
```
dominiklohmann, line 6: Didn't we wanna get rid of these?

nix/update.sh
```diff
@@ -51,3 +56,21 @@ update-source() {
 
 update-source caf "libvast/aux/caf"
 update-source fast_float "libvast/aux/fast_float"
+
+echo "Updating contrib/vast-plugins"
+vast_plugins_rev="$(get-submodule-rev "contrib/vast-plugins")"
+vast_plugins_json="$(jq --arg rev "${vast_plugins_rev#rev}" \
+  '."rev" = $rev' "${dir}/vast/plugins/source.json")"
+echo -E "${vast_plugins_json}" > "${dir}/vast/plugins/source.json"
```
dominiklohmann, line 64: Ideally the workflow for updating this repo is just bumping the submodule. This looks like there are additional requirements. Can you document the workflow for updating the plugins?

dominiklohmann (commented):
A first look at everything. I'm excited for this change!

.gitmodules
```diff
@@ -10,3 +10,6 @@
 [submodule "libvast/aux/caf"]
 	path = libvast/aux/caf
 	url = https://github.com/actor-framework/actor-framework
+[submodule "contrib/vast-plugins"]
+	path = contrib/vast-plugins
+	url = git@github.com:tenzir/vast-plugins
```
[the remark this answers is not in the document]
tobim: Hm, `git@` is my default, but we might not need the ssh-agent-action in some places if we switch. I'm gonna try that, but only after the other issues are resolved.

.github/workflows/vast.yaml
```diff
@@ -624,35 +661,12 @@ jobs:
...
           upload_url: ${{ github.event.release.upload_url }}
-          asset_path: "${{ steps.build_static_packages.outputs.store_path }}/${{ steps.get_artifact_names.outputs.deb }}"
+          asset_path: "${{ steps.build_static_packages.outputs.deb_github }}"
           # The asset name is constant so we can permanently link to
           # https://github.com/tenzir/vast/releases/latest/download/vast-linux-static.deb
           # for a build of the latest release.
           asset_name: "vast-linux-static.deb"
           asset_content_type: application/vnd.debian.binary-package
-
-  docker-vast-slim:
```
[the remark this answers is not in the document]
tobim: It should no longer build vast twice.

cmake/VASTRegisterPlugin.cmake
tobim: It is explained in the commit message:

> Before this change example configs and READMEs of out of tree
plugins would always installed into the docs location of the main
installation. If this is desired we can still achieve it by setting
`CMAKE_INSTALL_DOCDIR` explicitly when builiding the plugin.
>
> This change exists for two reasons:
> 1. It doesn't really make sense to install a plugin into one place
>    but the docs for it somewhere else.
> 2. When building out of tree plugins with Nix the main installation
>    tree of VAST is read-only, so the build would fail.

nix/vast/plugins/versions.nix
tobim: That also needs changes to the C++ code for the version command. I hoped to do that in a follow up PR.

nix/update.sh
tobim: How about we fail the "VAST ... (Nix)" jobs with a nice message if it goes out of sync?

dominiklohmann: Then the script should run without requiring Nix to be installed, the workflow should be documented, and the script should probably be in `/scripts` rather than `/nix`.

I don't understand why we would want to track versions for plugins in multiple places. This seems like an anti-pattern to me if we already have the submodule revision as a source of truth.

.gitmodules
dominiklohmann, line 15: To check in CI if this is in the correct branch:

```
git -C contrib/vast-plugins fetch origin main
git -C contrib/vast-plugins merge-base --is-ancestor $(git -C contrib/vast-plugins rev-parse HEAD) $(git -C contrib/vast-plugins rev-parse origin/main)
```

The second command returns a non-zero exit code if the commit the submodule is pinned to is not in origin/main.

IIRC we wanted to add this as a separate job to the workflow that is required to pass the branch protections. A separate job so it prohibits you from merging, but doesn't stop the other jobs from running.

nix/update.sh
tobim: As of this PR these numbers are meaningless. I only left them here to avoid touching the C++ code as well. As explained above it would be good to remove them in a separate PR.

.github/workflows/vast.yaml
```diff
@@ -668,11 +682,11 @@ jobs:
       - name: Build Docker Image
         if: github.event_name != 'workflow_dispatch'
         run: |
-          nix run .#stream-static-image | docker load
+          nix run .#stream-${{ matrix.vast-nix.package }}-image | docker load
       - name: Upload Docker Image
         if: github.event_name != 'workflow_dispatch'
         run: |
-          images=('tenzir/vast-slim')
+          image=$(echo "tenzir/${{ matrix.vast-nix.package }}" | sed -i 's|static|slim|')
```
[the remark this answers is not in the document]
tobim: I used only `vast-slim` and `vast-ce-slim` because the nix apps for creating them are now also called `*-slim` instead of `*-static`.

.github/workflows/vast.yaml
```diff
@@ -500,8 +525,15 @@ jobs:
     needs:
       - configure
     if: ${{ needs.configure.outputs.run-vast-nix == 'true' }}
-    name: VAST (Nix)
+    name: VAST Static (${{ matrix.vast-nix.name }})
```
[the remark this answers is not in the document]
tobim: I guess we should also change the Docker jobs accordingly?

nix/update.sh
dominiklohmann: Okay if that is something you essentially want to do near-term, not so much if follow-up means medium- to long-term.

.github/workflows/vast.yaml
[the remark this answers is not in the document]
dominiklohmann: Makes sense to me 👍

nix/update.sh
tobim: I will do the follow up shortly after this one was merged. We need to come up with a solution for versioning 3rd party plugins though.

.gitmodules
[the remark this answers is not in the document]
tobim: I gave up after fighting with the checkout-action for almost an hour.

nix/update.sh
```diff
@@ -1,53 +1,38 @@
...
-    fi
+toplevel="$(git -C "${dir}" rev-parse --show-toplevel)"
+mainroot="$(git -C "${dir}" rev-parse --path-format=absolute --show-toplevel)"
+gitdir="$(git -C "${dir}" rev-parse --path-format=absolute --git-common-dir)"
+echo "toplevel = ${toplevel}"
+echo "mainroot = ${mainroot}"
+echo "gitdir = ${gitdir}"
+
+command -v docker > /dev/null && {
+  echo "Updating with with a nix docker container..."
```
dominiklohmann, line 23: When I run this, it replaces the plugins/version.json file with this for me:

```diff
M nix/vast/plugins/versions.nix
@@ -1,6 +1,6 @@
 {
-  compaction = "1.4.0";
-  inventory = "1.0.0";
-  matcher = "3.0.0";
-  netflow = "1.0.0";
+  compaction = "";
+  inventory = "";
+  matcher = "";
+  netflow = "";
 }
```

.github/workflows/vast.yaml
```diff
@@ -67,10 +67,10 @@ jobs:
       cmake-version-args: ${{ steps.configure.outputs.cmake-version-args }}
       docker-build-args: ${{ steps.configure.outputs.docker-build-args }}
```
dominiklohmann, line 68: These need to be updated as well for the CE versions because they show the wrong version string currently, unless the goal is to also fix this in the follow-up by removing this logic entirely.

```
❯ docker run --platform=linux/amd64 ghcr.io/tenzir/vast-ce:topic-flake-version version
<jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
<jemalloc>: (This is the expected behaviour if you are running under QEMU)
[13:41:18.133] loaded plugin: /opt/tenzir/vast/lib/vast/plugins/libvast-plugin-netflow.so
[13:41:18.133] loaded plugin: /opt/tenzir/vast/lib/vast/plugins/libvast-plugin-sigma.so
[13:41:18.133] loaded plugin: /opt/tenzir/vast/lib/vast/plugins/libvast-plugin-pcap.so
[13:41:18.133] loaded plugin: /opt/tenzir/vast/lib/vast/plugins/libvast-plugin-web.so
[13:41:18.133] loaded plugin: /opt/tenzir/vast/lib/vast/plugins/libvast-plugin-matcher.so
[13:41:18.133] loaded plugin: /opt/tenzir/vast/lib/vast/plugins/libvast-plugin-parquet.so
{
  "VAST": "v2.4.0-1081-g25588c556c",
  "Build Configuration": {
    "Type": "Release",
    "Tree Hash": "9bbefa8033590ed7dc15be3806f8caf0",
    "Assertions": false,
    "Address Sanitizer": false,
    "Undefined Behavior Sanitizer": false
  },
  "CAF": "0.18.7",
  "Apache Arrow": "11.0.0",
  "jemalloc": null,
  "plugins": {
    "matcher": "v3.0.0-g",
    "netflow": "v1.0.0-g",
    "parquet": "v1.0.0-g601159c6b4",
    "pcap": "v1.1.0-ge4e37142e6",
    "sigma": "v1.1.0-ge4e37142e6",
    "web": "v1.0.0-g25588c556c"
  }
}
```

.github/workflows/vast.yaml
```diff
@@ -579,18 +663,19 @@ jobs:
...
-          gsutil cp "gs://${{ env.PUBLIC_GCS_BUCKET }}/${{ env.STATIC_BINARY_FOLDER }}/${{ steps.get_artifact_names.outputs.tar_gz }}" "gs://${{ env.PUBLIC_GCS_BUCKET }}/${{ env.STATIC_BINARY_FOLDER }}/vast-static-latest.tar.gz"
-          gsutil cp "${{ steps.build_static_packages.outputs.store_path }}/${{ steps.get_artifact_names.outputs.deb }}" "gs://${{ env.PUBLIC_GCS_BUCKET }}/${{ env.STATIC_BINARY_FOLDER }}/vast-static-latest.deb"
+          TAR_GZ_BASENAME="$(basename "${{ steps.build_static_packages.outputs.tar_gz }}")"
+          gsutil cp "${{ steps.build_static_packages.outputs.tar_gz }}" "gs://${{ env.PUBLIC_GCS_BUCKET }}/${{ env.STATIC_BINARY_FOLDER }}/$(basename "${TAR_GZ_BASENAME}")"
+          gsutil cp "gs://${{ env.PUBLIC_GCS_BUCKET }}/${{ env.STATIC_BINARY_FOLDER }}/${TAR_GZ_BASENAME}" "gs://${{ env.PUBLIC_GCS_BUCKET }}/${{ env.STATIC_BINARY_FOLDER }}/vast-static-latest.tar.gz"
+          gsutil cp "${{ steps.build_static_packages.outputs.deb }}" "gs://${{ env.PUBLIC_GCS_BUCKET }}/${{ env.STATIC_BINARY_FOLDER }}/vast-static-latest.deb"
```
dominiklohmann, line 669: These need to be either vast or vast-ce depending on the matrix.

.github/workflows/vast.yaml
```diff
@@ -579,18 +663,19 @@ jobs:
...
         env:
           PUBLIC_GCS_BUCKET: tenzir-public-data
           STATIC_BINARY_FOLDER: vast-static-builds
         run: |
-          VERSION=$(echo "${{ steps.get_artifact_names.outputs.tar_gz }}" | cut -d"-" -f2)
-          gsutil cp "${{ steps.build_static_packages.outputs.store_path }}/${{ steps.get_artifact_names.outputs.tar_gz }}" "gs://${{ env.PUBLIC_GCS_BUCKET }}/${{ env.STATIC_BINARY_FOLDER }}/vast-${VERSION}-static-latest.tar.gz"
-          gsutil cp "${{ steps.build_static_packages.outputs.store_path }}/${{ steps.get_artifact_names.outputs.deb }}" "gs://${{ env.PUBLIC_GCS_BUCKET }}/${{ env.STATIC_BINARY_FOLDER }}/vast-${VERSION}-static-latest.deb"
+          VERSION="${{ steps.configure.outputs.build-version-short }}"
+          gsutil cp "${{ steps.build_static_packages.outputs.tar_gz }}" "gs://${{ env.PUBLIC_GCS_BUCKET }}/${{ env.STATIC_BINARY_FOLDER }}/vast-${VERSION}-static-latest.tar.gz"
+          gsutil cp "${{ steps.build_static_packages.outputs.deb }}" "gs://${{ env.PUBLIC_GCS_BUCKET }}/${{ env.STATIC_BINARY_FOLDER }}/vast-${VERSION}-static-latest.deb"
```
dominiklohmann, line 678: Same here, this also needs to be either vast or vast-ce

.github/workflows/vast.yaml
```diff
@@ -513,32 +593,36 @@ jobs:
...
+          echo "TARBALL is at: $TAR_GZ"
           echo "tar_gz=${TAR_GZ}" >> $GITHUB_OUTPUT
+          DEB="$(find "$STORE_PATH" -type f -name '*.deb')"
+          echo "DEB is at: $DEB"
           echo "deb=${DEB}" >> $GITHUB_OUTPUT
+          # The debian package filename contains a colon (:), but GitHub does
+          # not allow this character in attached assets, so we create an extra
+          # copy.
+          DEB_GITHUB="$PWD/$(basename "$DEB" | tr ':' '_')"
+          echo "DEB_GITHUB is at: $DEB_GITHUB"
```
dominiklohmann, line 619: The package name must be set so it's either beginning with vast or vast-ce; currently it's always the former, so there's a race in CI for the upload.

dominiklohmann (commented):
There are some naming and versioning problems still, but I believe other than that we've got it now.

nix/update.sh
tobim: Was the submodule initialized when you tried that?

dominiklohmann: Yes, although it was not a fresh checkout. Let me try again after recreating the worktree, because I had to meddle with the submodule in my first review and I wanna make sure that I didn't mess this up because of that.

.github/workflows/vast.yaml
tobim: I had a quick look. Ugh, I'd rather not add the submodule checkout to the configure script only to remove it next week again.

dominiklohmann: Yeah, then just do it as part of the plugin version removal.

nix/update.sh
dominiklohmann: I traced it down with `set -x`, this is always returning an empty string for me:

```
version="$(sed -n 's|[^(]VERSION\s\+\([0-9A-Za-z\.-_]\+\)$|\1|p' "$cm" | tr -d '[:space:]')"
```

I'll get you a patch in a minute that fixes this.

dominiklohmann: Actually, considering this file should not exist as of the follow-up PR, maybe we should just leave that out. Ideally, that non-portable sed-call should not be needed.

dominiklohmann:
```diff
diff --git a/nix/update-plugins.sh b/nix/update-plugins.sh
index e36fbee19..0860f62b3 100755
--- a/nix/update-plugins.sh
+++ b/nix/update-plugins.sh
@@ -24,7 +24,12 @@ echo "Extracting plugin versions..."
     # Extract the plugin version from CMakeLists.txt.
     # TODO: Remove this and only store the plugin names when we no longer print
     # plugin versions in `vast version`.
-    version="$(sed -n 's|[^(]VERSION\s\+\([0-9A-Za-z\.-_]\+\)$|\1|p' "$cm" | tr -d '[:space:]')"
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+      SED=gsed
+    else
+      SED=sed
+    fi
+    version="$("${SED}" -n 's|[^(]VERSION\s\+\([0-9A-Za-z\.-_]\+\)$|\1|p' "$cm" | tr -d '[:space:]')"
     echo "plugin = $plugin, version = $version" >&2
     echo "  $plugin = \"$version\";"
   done
```

Here's a hack-ish workaround

dominiklohmann (approved):
As discussed in Slack, approving modulo the open remaining tasks so you can merge this yourself after carefully verifying that the artifacts are uploaded correctly in CI.

tobim:
> As discussed in Slack, approving modulo the open remaining tasks so you can merge this yourself after carefully verifying that the artifacts are uploaded correctly in CI.

The PR artifacts work now, but we need to wait for the RC to verify the release artifacts.

mavam:
Big step forward, nice work! 🚀

PR #2922, opened 2023-02-08
Remove the frontend prototype bundled with the web plugin

dominiklohmann:
This removes the existing frontend prototype from the web plugin that was bundled with the plugin installation and served as the default document root. Some parts of the frontend that we have in development are designed to be closed-source, and it is easier to develop at the current development stage in a single repository that is not bound to the release process of VAST itself.

An open-source version of the frontend may return in the future.


plugins/web/include/web/configuration.hpp
```diff
@@ -21,7 +21,7 @@ struct configuration {
     static auto result = vast::record_type{
       {"bind", vast::string_type{}},    {"port", vast::int64_type{}},
       {"mode", vast::string_type{}},    {"certfile", vast::string_type{}},
-      {"keyfile", vast::string_type{}}, {"web-root", vast::string_type{}},
```
lava, line 24: Please don't remove the option to have a webroot at all, that's independent of the UI itself.

I think its realistic to expect this to still be useful for local development, to run against an artifact created from the closed-source CI, or longer-term for a developer-centric UI bundled with VAST itself.

scripts/sbom-python-node.sh
```diff
@@ -15,9 +15,9 @@ BOMDIR=$(mktemp -d)
 # Generate the frontend ui SBOM
 # The primary reason for using cdxgen (and then converting to spdx format via syft)
 # is that it can look up nodeJS license fields online
-FETCH_LICENSE=true gum spin --spinner dot --title \
-  "Generating CycloneDX SBOM for the Frontend UI packages. Fetching the licenses from online may take a while..." \
-  -- cdxgen plugins/web/ui -o $BOMDIR/ui_cyclonedx.json
+# FETCH_LICENSE=true gum spin --spinner dot --title \
+#   "Generating CycloneDX SBOM for the Frontend UI packages. Fetching the licenses from online may take a while..." \
+#   -- cdxgen plugins/web/ui -o $BOMDIR/ui_cyclonedx.json
```
lava, line 20: Might as well remove the commented-out lines. (and probably rename the script)

lava (approved):
The removal of the UI looks complete to me; the webroot removal seems a bit overzealous though.

plugins/web/include/web/configuration.hpp
dominiklohmann: Fair point. I added this back.

scripts/sbom-python-node.sh
dominiklohmann: Done.

PR #2924, opened 2023-02-08
Prevent query_processor from hanging when there are no candidate partitions

Dakostu:
Call `process_done()` to prevent `query_processor` and -derived classes from hanging when there are no candidate partitions to process.

Fixes tenzir/issues#68.


tobim (approved):
The problem is obvious in hindsight. I think this solution is complete.

dominiklohmann:
Can you change the title of this PR so it makes sense to everyone?

PR #2927, opened 2023-02-09
Remove leftover traces of vast-ui

tobim:


nix/overlay.nix
```diff
@@ -193,16 +193,4 @@ in
     '';
     subPackages = [ "cmd/speeve" ];
   };
-
-  vast-ui = nnbp.buildYarnPackage {
```
dit7ya, line 197: We should remove `nix-npm-buildpackage` from the flack and update the lockfile as well. I can take care of that if you want.

dit7ya:
Didn't want to push here and opened #2928 instead.

tobim:
> Didn't want to push here and opened #2928 instead.

Thanks. I went ahead and merged it into this branch.

dit7ya (approved):

PR #2930, opened 2023-02-09
Apply overlooked wordsmithing feedback

dominiklohmann:


web/blog/from-slack-to-discord/index.md
```diff
@@ -29,7 +29,7 @@ that abandons Slack. [Numerous][meilisearch] [open-source][appwrite]
 You can join our Discord community chat via <https://vast.io/discord>.
 :::
 
-Here are our top reasons why we switched:
+Here are the four top reasons why we are switching:
```
dit7ya, line 32:
```suggestion
Here are the top four reasons why we are switching:
```

IMHO, "top four" sounds much better than "four top" :neutral_face: - that is what I suggested originally as well.

mavam: Agreed. I was going for this integration of feedback a bit too hastily.

mavam (approved):

PR #2932, opened 2023-02-09
Use `fan_out_request` for request with `void` result

dominiklohmann:
Now that we depend on CAF 0.18.7 and `fan_out_request` is properly working for requests with the result type `void` we can simplify this code a bit. The same could also be done in the partition transformer, but that's a bit more involved and likely to be replaced with some form of pipeline soon anyways, so I didn't bother.


patszt (approved):
Thanks for the cleanup.

PR #2933, opened 2023-02-09
Start versioning docs with the next release

dominiklohmann:
This is two changes in one:
1. Add a dropdown for version selection to the docs on vast.io.
2. With the next release, automatically create a saved version of the docs when running the `prepare-release` script.

If you want to see what this looks like locally, run `yarn docusaurus docs:version v3.0.0` from the web directory, and then preview with `yarn docusaurus start` as usual.


mavam:
Nice. I always shied away from this changed because I thought it's a massive change. This looks like a clean surgical operation.

mavam (approved):
Nice, this works really well locally for me.

Before merging, we need to agree on the naming. In our changelog, we talk about "unreleased" and here we say "next". This needs to be streamlined. I like the former better. Can we rename "next"?

dominiklohmann:
> Can we rename "next"?

I'm not sure. I'll make sure to align it, but if it's a tough change to make then I'll unbureaucratically align unreleased to next in the changelog.

There's one other thing we need to change, but that's also mostly mechanical: Images in /static are unversioned (by design), so we need to move those referenced in the documentation over into the /docs directory.

mavam:
> There's one other thing we need to change, but that's also mostly mechanical: Images in /static are unversioned (by design), so we need to move those referenced in the documentation over into the /docs directory.

@dit7ya will love this, as it breaks the carousel fix for good.

dominiklohmann:
> @dit7ya will love this, as it breaks the carousel fix for good.

The carousel images are in blog posts, so they should be in /static anyways.

scripts/prepare-release
```diff
@@ -50,9 +50,13 @@ version_json="$(jq --arg v "${new_version#v}" '."vast-version-fallback" = $v' ve
 echo -E "${version_json}" > version.json
 
 # If the new version is not a release candidate, change mentions of the version
-# in README.md as well.
+# in README.md as well, and create a new docs version.
 if [[ "${new_version}" != *"-rc"* ]]; then
   perl -i -pe "s/${last_version}/${new_version}/g" "${source_dir}/README.md"
+  pushd "${source_dir}/web"
+  yarn install --frozen-lockfile
+  yarn docusaurus docs:version "${new_version}"
```
mavam, line 58: After calling this, I get new files in the repo:

```
	web/versioned_docs/
	web/versioned_sidebars/
	web/versions.json
```

Where are they added to the repo? Does that need to happen as well?

mavam:
From https://docusaurus.io/docs/2.1.0/versioning:

> By default, the current docs version is labeled as `Next` and hosted under `/docs/next/*`, but it is entirely configurable to fit your project's release lifecycle.

scripts/prepare-release
dominiklohmann: The prepare-release script adds them to the repo, commits them, and creates a PR. That's already part of the script.

dominiklohmann:
I want to make the following changes still:

- Move blog assets to /static
- Move docs assets to /docs
- Rename changelog unreleased to next (or vice versa if possible, but appears unlikely)
- Use relative links to `.md` files when cross-referencing docs
- Avoid creating a new docs version with patch releases, they should instead update the matching minor release docs (e.g., cutting v3.0.1 should update the docs for v3.0)

Still some work to do, but it's mostly mechanical and can partially be automated. I think I may let this sit for a while as it's not high up on my priority list, but this is something that I can easily take care of whenever I feel like working on something that requires very few brain cells being available.

mavam:
> Use relative links to .md files when cross-referencing docs

Aren't we doing this already?

dominiklohmann:
> Aren't we doing this already?

It's mostly `/docs/foo/bar` that we use, but that has to be `@site/foo/bar.md` instead. This will also cause Docusaurus to properly error on wrong links.

mavam:
> It's mostly `/docs/foo/bar` that we use, but that has to be `@site/foo/bar.md` instead. This will also cause Docusaurus to properly error on wrong links.

That feels very unergonomic. When you are in `foo/`, you should be able to just link to `bar.md`.

dominiklohmann:
> That feels very unergonomic. When you are in `foo/`, you should be able to just link to `bar.md`.

That also works, you're just not allowed to start your links with a leading slash.

mavam:
@dominiklohmann if you wait for #2879, I can do the movement into /docs as part of the image re-generation.

mavam:
FWIW, this is unblocked now and only requires a rebase onto latest master.

dominiklohmann:
We decided to add the relative links to #2971 instead, so I'm going ahead with merging this one.

PR #2935, opened 2023-02-10
Implement STDIN loader plugin

Dakostu:
Implement the basic foundation for the `stdin` loader plugin that creates a `generator<chunk_ptr>` for the `stdin` input string. Currently unreachable from the user POV.

Fixes tenzir/issues#73.


libvast/src/plugin.cpp
```diff
@@ -380,6 +380,38 @@ store_plugin::make_store(system::accountant_actor accountant,
...
+    auto input = std::string{};
+    auto c = char{};
+    while (std::cin.get(c)) {
+      input += c;
+    }
+    if (not input.empty()) {
+      auto chunk = chunk::make(std::move(input));
+      co_yield std::move(chunk);
+    }
+    co_return;
```
dominiklohmann, line 397: I think there are some problems here that we need to take a look at with some detail:
- When do we yield? We don't want to read everything and then yield just a single chunk, as that destroys the pipelining we're trying to achieve
- How do we handle a stalling input? `std::istream::get(char&)` may block indefinitely.
- Can we read more than one character at a time? Maybe BUFSIZ is a good starting point, which you could use through `rdbuf`? (Although there is a [blog post](https://www.zverovich.net/2020/08/04/optimal-file-buffer-size.html) from a well-known author saying that the value should be larger than that, at least for writing.)
- Do we need to go through the iostreams API for this?

libvast/src/plugin.cpp
```diff
@@ -380,6 +380,38 @@ store_plugin::make_store(system::accountant_actor accountant,
...
 }
 
+// -- loader plugin -------------------------------------------------------------
+
+auto stdin_loader_plugin::make_loader(options, const bool_operator*) const
+  -> caf::expected<input_loader> {
+  return []() -> detail::generator<chunk_ptr> {
+    auto input = std::string{};
+    auto c = char{};
+    while (std::cin.get(c)) {
```
mavam, line 390: This should most likely do unbuffered reads via the underlying streambufs.

We have some utilities (IIRC `fdinbuf`) where you can also set the buffer size.

libvast/src/plugin.cpp
mavam: For now, I'd go with a reasonably large buffer size like 16k. At some later point, I'd then do some benchmarks. Our `fdinbuf` has half the size as default: https://github.com/tenzir/vast/blob/master/libvast/include/vast/detail/fdinbuf.hpp

dominiklohmann: Looking at fdinbuf we can probably get rather far with it. Any logic for incremental backoff should likely be implemented generically upstream of the source in a pipeline, so in case of a timeout it should suffice to emit an empty chunk.

mavam: > Looking at fdinbuf we can probably get rather far with it.

Yeah, we use it for quite a while now when reading from STDIN, so it's battle-proven.

> Any logic for incremental backoff should likely be implemented generically upstream of the source in a pipeline, so in case of a timeout it should suffice to emit an empty chunk.

What's the benefit of emitting empty chunks? Why simply not emit anything when hitting a timeout?

dominiklohmann: You need to yield control to outside of the generator coroutine, for which you must yield something. If you have nothing but are not at the end of your input, then you yield nothing, which in this case is an empty chunk. The executor does not forward empty batches of elements (empty chunks or empty table slices) to the next operator.

Dakostu: `fdinbuf` is what we are looking for. I think we can even omit the char-by-char insertion from the buffer into a string but I might need to implement some kind of buffer slice span method for `fdinbuf`.

dominiklohmann: There are two ways to go about this after reading something:
1. You can swap out the rdbuf and "move" the ownership of the old rdbuf into the chunk
2. You can copy from the rdbuf into a newly created chunk

For (1) you'll likely need a way to access an rvalue reference to `fdinbuf::buffer_`, and for (2) you'll likely need an `as_bytes(const fdinbuf&)` overload. Both should work; I'm not sure without measuring how the extra copy in (2) compares against the fdinbuf construction and rdbuf swap in (1). I wouldn't recommend spending time measuring that, but please add a comment to the code explaining that we have these two options and did not benchmark.

Dakostu: Hmmm, I think I've been too naive with `fdinbuf::buffer_`... I noticed that `fdinbuf` uses a putback area in its `buffer` with a `putback_area_size` of 10 - e.g. when beginning to read data, `fdinbuf`  only starts writing into `buffer_` from its 10th index. Then, in the next chunk, the putback area will be then filled with what looks like `gptr() - putback_area_size`. So basically I can not simply use all the contents of the `buffer_` and an `{eback(), egptr()}` slice of the `buffer_`  will always return a chunk with a maximum size of `~16k - putback_area_size`.

The main directions I can go to from here are:
1. Continue the `{eback(), egptr()}` approach.
2. Check whether `fdinbuf` actually needs a putback area in our case.
3. Disregard direct-from-buffer approach and append a (size-reserved) string char-by-char.

dominiklohmann: As a variation (3b), you could take some inspiration from `detail::absorb_line`.

Dakostu: I decided to opt for the 3-3b implementation.

libvast/include/vast/plugin.hpp
```diff
@@ -398,6 +398,44 @@ class rest_endpoint_plugin : public virtual plugin {
...
+
+  [[nodiscard]] auto make_default_parser(options, const bool_operator*) const
+    -> caf::expected<input_parser> override;
+
+  [[nodiscard]] caf::error initialize(data config) override;
+
+  [[nodiscard]] std::string name() const override;
+
+  static constexpr inline auto max_chunk_size = size_t{16384};
+  static constexpr inline auto read_timeout = std::chrono::milliseconds{5000};
```
Dakostu, line 436: How long should the stdin loader timeout be? I arbitrarily chose 5 seconds, but I don't know what value would be truly sensible for our users' use cases.

dominiklohmann: For the current readers this is configurable as `vast.import.read-timeout`, which I believe defaults to 20ms. We probably need to talk about options handling, I don't think we've considered that in the loader plugin design so far.

libvast/src/plugin.cpp
```diff
@@ -380,6 +382,55 @@ store_plugin::make_store(system::accountant_actor accountant,
...
 }
 
+// -- loader plugin -------------------------------------------------------------
+
+auto stdin_loader_plugin::make_loader(options, const bool_operator*) const
+  -> caf::expected<input_loader> {
+  return []() -> generator<chunk_ptr> {
+    auto in_buf = detail::fdinbuf(STDIN_FILENO, max_chunk_size);
+    in_buf.read_timeout() = read_timeout;
+    auto chunk_str = std::string{};
```
dominiklohmann, line 392: I think this should not be a `std::string` to avoid assumptions about encoding, and to avoid having to also always ensure that a null-terminator exists (which is mandated by the standard). Generally, use vectors over strings for growable buffers, and use `std::byte` or `unsigned char` to avoid making assumptions about the encoding.

libvast/src/plugin.cpp
```diff
@@ -380,6 +382,55 @@ store_plugin::make_store(system::accountant_actor accountant,
...
+      if (current_char != detail::fdinbuf::traits_type::eof()) {
+        chunk_str += static_cast<char>(current_char);
+      } else if (in_buf.timed_out() && chunk_str.empty()) {
+        co_yield chunk::make_empty();
+        continue;
+      } else if (!in_buf.timed_out()) {
+        if (chunk_str.empty()) {
+          break;
+        }
+      }
```
dominiklohmann, line 406: I think it's not ideal to check for the timeout for every single byte read, this seems like a lot of overhead to me. Can we do this in the outer loop?

libvast/src/plugin.cpp
```diff
@@ -380,6 +382,55 @@ store_plugin::make_store(system::accountant_actor accountant,
                                                  std::move(path), name());
 }
 
+// -- loader plugin -------------------------------------------------------------
+
+auto stdin_loader_plugin::make_loader(options, const bool_operator*) const
+  -> caf::expected<input_loader> {
+  return []() -> generator<chunk_ptr> {
+    auto in_buf = detail::fdinbuf(STDIN_FILENO, max_chunk_size);
```
dominiklohmann, line 390: IIUC code using an rdbuf directly must be guarded with a `std::istream::sentry`, but I suppose that's irrelevant here because we don't actually use an istream but rather just the rdbuf? Is it a deliberate decision not to go through an istream for reading?

libvast/src/plugin.cpp
```diff
@@ -380,6 +382,55 @@ store_plugin::make_store(system::accountant_actor accountant,
...
+  -> caf::expected<input_parser> {
+  return caf::make_error(ec::unimplemented, "parser currently not implemented");
+}
+
+caf::error stdin_loader_plugin::initialize(data) {
+  return caf::none;
+}
+
+std::string stdin_loader_plugin::name() const {
+  return "stdin_loader_plugin";
```
dominiklohmann, line 431: `from std_loader_plugin read json` reads a bit weird to me, so I think this should just be `stdin` instead.

libvast/src/plugin.cpp
```diff
@@ -380,6 +382,55 @@ store_plugin::make_store(system::accountant_actor accountant,
                                                  std::move(path), name());
 }
 
+// -- loader plugin -------------------------------------------------------------
+
+auto stdin_loader_plugin::make_loader(options, const bool_operator*) const
```
dominiklohmann, line 387: Let's talk about the second argument: What do we generally need in loaders? Is it a pointer to the operator control plane?

dominiklohmann (commented):
Nice, this is starting to look a lot better. Some further questions on the input handling.

libvast/src/plugin.cpp
mavam: The only reason why you'd go through an istream is when you want formatted input. Since we're doing block-level access of the file descriptor, that strikes me as the right (implicit?) design choice.

libvast/include/vast/plugin.hpp
mavam: It would be good to first write down exactly what the timeout does. Someone who sees `read_timeout` needs more context to make a decision of what they want. Ideally it doesn't require going to the call sites.

libvast/src/plugin.cpp
Dakostu: `istream` has the ability to format and interpret in/output, which seems unnecessary in this case.

libvast/include/vast/plugin.hpp
Dakostu: The `read_timeout` is passed on to the `fdinbuf` timeout value. If the `fdinbuf` file descriptor polling does not pass within that timeframe, there is a timeout. In that case it would make sense to add some coherence by unifying this setting with `vast.import.read-timeout`.

I wonder whether the operator control plane or the pipeline executor would be able to pass on option values somehow.

Dakostu: If `plugin` already contains a subset of the options, maybe it can also be expanded to also include global VAST options for easy access.

libvast/src/plugin.cpp
Dakostu: I fixed the name in https://github.com/tenzir/vast/pull/2935/commits/ad2821c61f68d9ab92d2cca3bca422f19c580039

libvast/src/plugin.cpp
Dakostu: I specialized the timeout check to only trigger when we reach `eof()` since `fdinbuf` returns an `eof()` upon reaching a timeout. https://github.com/tenzir/vast/pull/2935/commits/ea29892b310fee9a14ce51876008f8093b946235

libvast/src/plugin.cpp
Dakostu: Yeah, I forgot to switch out the `std::string` because it was easier to debug. I switched to `std::vector<std::byte>` in https://github.com/tenzir/vast/pull/2935/commits/ea29892b310fee9a14ce51876008f8093b946235

libvast/include/vast/plugin.hpp
dominiklohmann: I agree generally, but that should probably not be in this PR but rather in a separate one. 

Right now, there's a member function of the `plugin` base class that gets just the relevant subsection:

```cpp
/// Initializes a plugin with its respective entries from the YAML config
/// file, i.e., `plugin.<NAME>`.
/// @param config The relevant subsection of the configuration.
[[nodiscard]] virtual caf::error initialize(data config) = 0;
```

We could consider extending this to get all options, and require the plugins themselves to access the subsection relevant to them. Since this is a bigger change that affects everyone, I would recommend starting a separate discussion in Slack and if agreed upon creating a separate PR.

The alternative would be to use a workaround like `caf::get_or(content(ctrl.self()->config()), "vast.read-timeout", 20ms)`, but that mostly works by coincidence and not be design.

libvast/src/plugin.cpp
Dakostu: What should the loaders generally be able to do aside from loading? Maybe some form of up-/downstream communication. The only kind of upstream communication that the `stdin` loader is doing right now are `chunk_ptr`s to the parser. I can however imagine loaders having to `warn` when a timeout is reached or something strange happens to the carrier that prevents it from sending further data. But then again IDK if that should be the responsibility of the loader plugin or the `from`/`read` umbrella operators.

libvast/src/plugin.cpp
dominiklohmann: There are probably no error scenarios for this specific loader, but I can imagine for example a TCP loader that would stop with a meaningful error when the handshake fails. Maybe we can just postpone this discussion until the next loader, and for now just blindly take a `operator_control_plane& ctrl`.

libvast/src/plugin.cpp
Dakostu: I replaced the `bool_operator`s with our new `operator_control_plane` in https://github.com/tenzir/vast/pull/2935/commits/1c1f320cf0cc59a67a998ad9a2b54a49987c80eb

libvast/include/vast/plugin.hpp
Dakostu: I created a draft PR that implements the plugin option extension: #2949

libvast/include/vast/plugin.hpp
```diff
@@ -402,6 +403,47 @@ class rest_endpoint_plugin : public virtual plugin {
...
+    -> caf::expected<input_parser> override;
+
+  [[nodiscard]] caf::error
+  initialize(const record& plugin_config, const record& global_config) override;
+
+  [[nodiscard]] std::string name() const override;
+
+private:
+  std::chrono::milliseconds read_timeout{vast::defaults::import::read_timeout};
+};
```
dominiklohmann, line 445: This should be a builtin rather than directly in the libvast, otherwise this defeats the purpose of the plugin API.

libvast/include/vast/plugin.hpp
```diff
@@ -402,6 +403,47 @@ class rest_endpoint_plugin : public virtual plugin {
     = 0;
 };
 
+// -- loader plugin -----------------------------------------------------
+
+using input_loader = std::function<auto()->generator<chunk_ptr>>;
+using input_parser = std::function<auto(chunk_ptr)->generator<table_slice>>;
```
dominiklohmann, line 409: I think the design doc had these as just `loader` and `parser`, respectively. The word `input` seems rather confusing to me here. The signature for the `parser` also changed to `(generator<chunk_ptr>) -> generator<table_slice>`.

libvast/include/vast/plugin.hpp
```diff
@@ -402,6 +403,47 @@ class rest_endpoint_plugin : public virtual plugin {
...
+using options = record;
```
dominiklohmann, line 410: This was marked "exposition only" in the design doc as we weren't sure on the shape of the options yet; please just use `const record&` directly below and remove this alias.

dominiklohmann (changes requested):

libvast/include/vast/plugin.hpp
Dakostu: Unfortunately using `using parser = ... `  creates a conflict with `concept parser = ...` in `parser.hpp:122` which causes a compilation error. 
I can circumvent this error by defining `loader`/`parser` in the `vast::plugins` namespace instead of `vast`.

libvast/include/vast/plugin.hpp
Dakostu: I removed the alias in https://github.com/tenzir/vast/pull/2935/commits/405db6875bd7a26ce4bb440f9010f8335dc84413

libvast/include/vast/plugin.hpp
Dakostu: I moved `loader` and `parser` into `vast::plugins` in https://github.com/tenzir/vast/pull/2935/commits/405db6875bd7a26ce4bb440f9010f8335dc84413

dominiklohmann: It doesn't make much sense to me to keep them under `vast::plugins::{loader,parser}`; that is not the appropriate place for them. Can we perhaps either rename the conflicting concept, or move `loader` and `parser` to `vast::loader_plugin::{loader,parser}`?

Please also make sure that at the very least plugin interfaces are commented well.

Dakostu: It's easier to move them into `vast::loader_plugin` -> https://github.com/tenzir/vast/pull/2935/commits/ea6253b3f94ce8661ae92c76e3f5fe7eae0e38b7

libvast/builtins/loaders/stdin_loader.cpp
```diff
@@ -0,0 +1,102 @@
+//    _   _____   __________
```
dominiklohmann, line 1: Unrelated to this line: Can you rename this file to be stdin.cpp so it's in line with the other builtins where the filename exactly matches the plugin name?

libvast/builtins/loaders/stdin_loader.cpp
```diff
@@ -0,0 +1,102 @@
...
+      in_buf.read_timeout() = read_timeout;
+      auto current_data = std::vector<std::byte>{};
+      current_data.reserve(max_chunk_size);
+      auto eof_reached = false;
+      while (not eof_reached) {
+        auto current_char = in_buf.sbumpc();
+        if (current_char != detail::fdinbuf::traits_type::eof()) {
+          current_data.emplace_back(static_cast<std::byte>(current_char));
+        } else {
+          eof_reached = (not in_buf.timed_out());
```
dominiklohmann, line 37: Can you write an issue to track that this behavior is currently untested? I don't expect a test for it within this PR, as that's much easier to add as an integration test later on.

dominiklohmann (commented):
Just some minor things now, I think functionality-wise this has arrived at the desired point.

libvast/builtins/loaders/stdin_loader.cpp
Dakostu: Done.

libvast/test/loader_plugin.cpp
```diff
@@ -0,0 +1,176 @@
...
+  auto loader_plugin = vast::plugins::find<vast::loader_plugin>("stdin");
+  REQUIRE(loader_plugin);
+  current_loader = loader_plugin->make_loader({}, control_plane);
+  REQUIRE(current_loader);
+  auto loaded_chunk_generator = (*current_loader)();
+  auto str_chunk = chunk::copy(str);
+  for (const auto& chunk : loaded_chunk_generator) {
+    REQUIRE(std::equal(chunk->begin(), chunk->end(), str_chunk->begin(),
+                       str_chunk->end()));
+  }
```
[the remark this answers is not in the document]
Dakostu: Does `collect` work in this way? I tested your lines and there seems to be a compilation error:
image

`collect` takes a `generator` by value, which seems to call a deleted `generator(const generator& other)` copy-constructor. 

Strangely, I did get it to work with: 

```c++
  auto loader = loader_plugin->make_loader({}, control_plane);
  auto chunks = collect((*loader)());
```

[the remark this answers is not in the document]
Dakostu: And also with:
`auto chunks = collect((unbox(current_loader))());`

So probably just a bracket mishap.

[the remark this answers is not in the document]
dominiklohmann: Two problems with this: Generators are non-copyable, and I had one pair of parentheses wrong above.

```
auto loader = unbox(loader_plugin->make_loader({}, control_plane));
auto chunks = collect(loader());
```

libvast/builtins/loaders/stdin_loader.cpp
```diff
@@ -0,0 +1,102 @@
...
+namespace vast::plugins::stdin_loader {
+
+// -- loader plugin -----------------------------------------------------
+class plugin : public virtual loader_plugin {
+public:
+  static constexpr inline auto max_chunk_size = size_t{16384};
+
+  [[nodiscard]] auto make_loader(const record&, operator_control_plane&) const
+    -> caf::expected<loader> override {
+    return [this]() -> generator<chunk_ptr> {
```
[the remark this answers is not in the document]
Dakostu: The only thing the lambda needs from the plugin instance (as of now) is the `read_timeout` field. It can capture a `read_timeout` copy. -> https://github.com/tenzir/vast/pull/2935/commits/e435401783a132b238d243db154e8e448183d62c

libvast/src/plugin.cpp
```diff
@@ -33,6 +36,7 @@
 #include <dlfcn.h>
 #include <memory>
 #include <tuple>
+#include <unistd.h>
```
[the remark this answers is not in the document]
Dakostu: I thought I needed those since the Mac CI run had a build error due to missing includes.. but now it works again so I removed them in https://github.com/tenzir/vast/pull/2935/commits/eaa93a43c9c711744a29093ad06818fbebe59520

libvast/builtins/loaders/stdin_loader.cpp
Dakostu: I renamed the file to `stdin.cpp`. https://github.com/tenzir/vast/pull/2935/commits/709a347f8a2c22abba3180e8896f5769de03fbc5

libvast/test/loader_plugin.cpp
```diff
@@ -0,0 +1,176 @@
...
+} // namespace
+
+FIXTURE_SCOPE(loader_plugin_tests, fixture)
+
+TEST(stdin loader - process simple input) {
+  input_file
+    = freopen(VAST_TEST_PATH "artifacts/inputs/simple.txt", "r", stdin);
+  auto str = std::string{"foobarbaz\n"};
+  auto loader_plugin = vast::plugins::find<vast::loader_plugin>("stdin");
+  REQUIRE(loader_plugin);
```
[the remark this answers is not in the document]
Dakostu: I moved the `loader_plugin` loading to the `fixture` constructor. https://github.com/tenzir/vast/pull/2935/commits/c8c91a6b3187476e454e75ed259c4015d7581bef

libvast/test/loader_plugin.cpp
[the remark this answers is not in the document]
Dakostu: I implemented the `unbox`/`collect` combination in https://github.com/tenzir/vast/pull/2935/commits/c8c91a6b3187476e454e75ed259c4015d7581bef

libvast/test/loader_plugin.cpp
```diff
@@ -0,0 +1,176 @@
...
+  for (const auto& chunk : loaded_chunk_generator) {
+    REQUIRE(std::equal(chunk->begin(), chunk->end(), str_chunk->begin(),
+                       str_chunk->end()));
+  }
+}
+
+TEST(stdin loader - chunking longer input) {
+  constexpr auto max_chunk_size = size_t{16384};
+  input_file
+    = freopen(VAST_TEST_PATH "artifacts/inputs/longer_input.txt", "r", stdin);
```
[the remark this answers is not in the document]
Dakostu: Damn, this is really beautiful... this cleans up the code nicely. I used this approach in https://github.com/tenzir/vast/pull/2935/commits/c8c91a6b3187476e454e75ed259c4015d7581bef

libvast/test/loader_plugin.cpp
```diff
@@ -0,0 +1,176 @@
...
+  fixture() {
+  }
+
+  ~fixture() {
+    if (input_file) {
+      fclose(input_file);
+    }
+  }
+
+  int old_stdin_fd;
```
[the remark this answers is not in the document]
Dakostu: Oh yes, this was an old remnant - I actually realized the current `freopen` approach did not work properly. I implemented a better, `freopen`-less approach in https://github.com/tenzir/vast/pull/2935/commits/6cc101f9d23a0fa725e2b6fd0cc6c38ae0c71d9a

dominiklohmann (approved):
🚀

PR #2937, opened 2023-02-10
Support pipelines in export and query endpoints

patszt:
Fixed a bug which made the /query/next return the same results infinitely. Now it returns 0 events if all events were shipped.
The query and export endpoints now accept an optional pipeline string in the query parameter.

The changelog entry for /query endpoints is already in unreleased state


lava (approved):
Looks good to me, I didn't test locally but the new unit and integration tests make me confident that this should work as expected.

PR #2944, opened 2023-02-13
Add CORS preflight request handling

lava:
For "non-simple" requests to external domains, the browser sends a preflight CORS request with method OPTIONS. The server needs to return a success response (with a matching origin, method, and headers) in order for the browser to go ahead with the request itself.

Fixes https://github.com/tenzir/issues/issues/86

### DIFF
--- a/plugins/web/include/web/configuration.hpp
+++ b/plugins/web/include/web/configuration.hpp
@@ -19,9 +19,13 @@ namespace vast::plugins::web {
 struct configuration {
   static const record_type& schema() noexcept {
     static auto result = vast::record_type{
-      {"bind", vast::string_type{}},    {"port", vast::int64_type{}},
-      {"mode", vast::string_type{}},    {"certfile", vast::string_type{}},
-      {"keyfile", vast::string_type{}}, {"web-root", vast::string_type{}},
+      {"bind", vast::string_type{}},
+      {"port", vast::int64_type{}},
+      {"mode", vast::string_type{}},
+      {"certfile", vast::string_type{}},
+      {"keyfile", vast::string_type{}},
+      {"web-root", vast::string_type{}},
+      {"cors-allowed-origin", vast::string_type{}},
     };
     return result;
   }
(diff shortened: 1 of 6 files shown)


plugins/web/src/server_command.cpp
```diff
@@ -330,9 +360,13 @@ auto server_command(const vast::invocation& inv, caf::actor_system& system)
                   *server_config, std::move(endpoint), handler);
     }
   }
+  // Set up implicit CORS preflight handlers for all endpoints if desired
+  if (server_config->cors_allowed_origin)
+    setup_cors_preflight_handlers(router, *server_config->cors_allowed_origin);
   // Set up non-API routes.
   router->non_matched_request_handler([](auto req) {
-    VAST_VERBOSE("404 not found: {}", req->header().path());
+    VAST_VERBOSE("404 not found: {} {}", req->header().method(),
```
dominiklohmann, line 368: The method here is not formattable.

plugins/web/include/web/configuration.hpp
```diff
@@ -19,9 +19,13 @@ namespace vast::plugins::web {
...
-      {"bind", vast::string_type{}},    {"port", vast::int64_type{}},
-      {"mode", vast::string_type{}},    {"certfile", vast::string_type{}},
-      {"keyfile", vast::string_type{}}, {"web-root", vast::string_type{}},
+      {"bind", vast::string_type{}},
+      {"port", vast::int64_type{}},
+      {"mode", vast::string_type{}},
+      {"certfile", vast::string_type{}},
+      {"keyfile", vast::string_type{}},
+      {"web-root", vast::string_type{}},
+      {"cors-allowed-origin", vast::string_type{}},
```
dominiklohmann, line 28: This new argument must be documented; it also requires a changelog entry announcing its existence.

dominiklohmann (approved):
This works well for me locally, just two smaller requests.

dit7ya (commented):
Cannot comment on the C++ implementation but I tested this branch with our Vercel deployments and it is working as expected.

PR #2948, opened 2023-02-15
Introduce simdjson submodule

patszt:
Simdjson introduces ondemand API which is very useful in JSON parser implementation.
The package which includes this API seems to be available from the ubuntu 22.04 release.
Also compiling it locally from the installed package seems to yield compilation failures on my machine when working with doubles.

Having simdjson as submodule enables creation of workarounds for these issues.


libvast/CMakeLists.txt
```diff
@@ -255,20 +255,20 @@ string(APPEND VAST_FIND_DEPENDENCY_LIST "\nfind_package(spdlog 1.5.0 REQUIRED)")
...
-  REQUIRED
-  # simdjson 0.8.0 exports it package configuration file under the wrong name,
-  # so we look for "simdjson-targets.cmake" as well.
-  CONFIGS
-  "simdjson-config.cmake"
-  "simdjson-targets.cmake")
-string(APPEND VAST_FIND_DEPENDENCY_LIST
-       "\nfind_package(simdjson REQUIRED CONFIGS "
-       "\"simdjson-config.cmake\" \"simdjson-targets.cmake\")")
+# Link against simdjson
```
tobim, line 258: We should have an option `VAST_ENABLE_BUNDLED_SIMDJSON` that defaults to off.
Use the old code the option is disabled and the new one otherwise.

tobim:
This needs a rebase for the Nix based build.
The plugins also fail, I don't know the reason for that yet.

tobim:
Actually I foresee a few problems with vendored public dependencies. We should remove this library from our API and link it statically into `libvast` instead.

There are 2 ways we can go about it:
1. Create wrappers around `::simdjson::dom::parser` and `::simdjson::dom::object`.
2. Move `vast/format/json.hpp` and the selectors headers into the `src` directory so that they aren't installed any more.

I strongly prefer (2) because our API surface is unnecessarily large anyways despite us explicitly not committing to any versioning or stability in that regard. And, equally important, moving a few files is less work.

dominiklohmann:
> Actually I foresee a few problems with vendored public dependencies.

Can you elaborate on them?

tobim:
Conflicts can occur in different places depending on the exact configuration that the user tries to install.
* If the prefix is `/usr` and simdjson is already installed the installation of VAST will fail
* If the prefix is `/opt/vast` the installation will succeed, but building downstream projects becomes very brittle in case a different version of simdjson is installed in one of the default locations.
* Running vast may fail for the same reason depending on the lookup order of the dynamic loader.

I'm aware that this is not a problem for the Docker image and it is also masked by the static binary configuration, but I'd still prefer us not setting a trap for our future selves.

dominiklohmann:
Right, but we have these problems for the bundled CAF as well. It is not in scope of this PR to handle these problems. If you want to avoid them, set `VAST_ENABLE_BUNDLED_SIMDJSON=OFF`. I would argue that it is too much of a distraction currently to start handling this correctly.

tobim:
> I would argue that it is too much of a distraction currently to start handling this correctly.

Also fine, we just need to be aware that it'll come up when we start supporting shared binary releases.

dominiklohmann (approved):
I think we've now got everything for this, assuming CI runs through.

dominiklohmann:
This is blocked by https://github.com/simdjson/simdjson/pull/1961; we could point to a fork instead, but let's wait a bit and see if the fix gets merged quickly.

PR #2949, opened 2023-02-15
Pass global options to plugins during initialization

Dakostu:
This PR implements a second `record` which contains all VAST-related options. These options will be passed on to `plugins::initialize()` together with a `record` for plugin-relevant options. Currently, none of the plugins have the need for reading global option values, but the currently WIP `loader` plugin will.

Fixes tenzir/issues#103.


libvast/include/vast/plugin.hpp
```diff
@@ -112,8 +112,12 @@ class plugin {
 
   /// Initializes a plugin with its respective entries from the YAML config
   /// file, i.e., `plugin.<NAME>`.
-  /// @param config The relevant subsection of the configuration.
-  [[nodiscard]] virtual caf::error initialize(data config) = 0;
+  /// @param plugin_config The relevant subsection of the configuration.
+  /// @param global_config The entire VAST configuration for potential access to
+  /// global options.
+  [[nodiscard]] virtual caf::error
+  initialize(data plugin_config, data global_config)
```
dominiklohmann, line 119: Does it make sense to pass both as a const-ref so we don't copy a large data object for every single plugin initialization, and the plugin subsection twice at that?

dominiklohmann: Additionally, should they both be `const record&` rather than `data`? At least for the global config, we always already have them in a record, so stuffing that back into a data just causes plugins to needlessly check whether the data contains a record. The same goes for the plugin-specific config, where if not present we can pass in an empty record instead, avoiding the need to check in each plugin.

Dakostu: Yes, they should; I was initially doubting whether the lifetime of the `record`s would suffice for plugin initialization, but yes, they would because each plugin processes the original `record` by converting it into a `configuration` anyway. I changed the signature in https://github.com/tenzir/vast/pull/2949/commits/b51d8b9b88beab81ae3cf5be88ec8a55195f6d71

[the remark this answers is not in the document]
Dakostu: Updated documentation in https://github.com/tenzir/vast/pull/2949/commits/ae1b5d0b95a5f71dcc1ff7a69abeafd75965e445

libvast/src/plugin.cpp
```diff
@@ -246,17 +246,25 @@ load(const std::vector<std::string>& bundled_plugins,
...
-      if (auto opts_data = to<record>(*opts))
-        merged_config = std::move(*opts_data);
+      if (auto plugin_opts_data = to<record>(*plugin_opts))
+        merged_config = std::move(*plugin_opts_data);
       else
         VAST_DEBUG("unable to read plugin options from VAST configuration at "
                    "plugins.{}: {}",
-                   opts_data.error(), plugin->name());
+                   plugin_opts_data.error(), plugin->name());
     }
```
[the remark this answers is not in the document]
Dakostu: I optimized the workflow in [40c5525](https://github.com/tenzir/vast/pull/2949/commits/40c5525f910ccb14a9e54a30a1c0dd7bc52cb7fb)

changelog/unreleased/breaking-changes/2949--pass-global-options-to-plugins-during-initialization.md
```diff
@@ -0,0 +1,5 @@
+There is a small addition to the plugin interface - the plugin initialization
+in `initialize()` now uses two `record` parameters. The former contains all
+plugin-relevant, the latter contains absolutely all configuration options. The
+plugins can now use both parameters for read-only access to all options in
+VAST.
```
dominiklohmann, line 5: I don't think this is user-facing, so this doesn't warrant a changelog entry. We don't do dev-facing changelog entries currently.

libvast/src/plugin.cpp
```diff
@@ -246,17 +246,28 @@ load(const std::vector<std::string>& bundled_plugins,
...
-          &cfg, fmt::format("plugins.{}", plugin->name()))) {
-      if (auto opts_data = to<record>(*opts))
-        merged_config = std::move(*opts_data);
-      else
-        VAST_DEBUG("unable to read plugin options from VAST configuration at "
-                   "plugins.{}: {}",
-                   opts_data.error(), plugin->name());
+    // First, try to read the configurations from the merged VAST configuration.
+    if (plugins_record.contains(plugin->name())) {
+      merged_config = caf::get<record>(plugins_record.at(plugin->name()));
```
dominiklohmann, line 270: This crashes for a config file like this:

```
plugins:
  myplugin: foo
```

Because the code here assumes that `plugins.myplugin` holds a record, when in this example it holds a string. The solution would be to use `caf::get_if` here instead to verify that it holds a record. It's probably a hard error if it doesn't.

dominiklohmann (approved):
This is great, just two minor notes. Please also coordinate with Tobias on merging between this and #2916 as there are some merge conflicts between these PRs. I started a thread in Slack to discuss this.

libvast/src/plugin.cpp
Dakostu: I added a validity check in https://github.com/tenzir/vast/pull/2949/commits/0a7f27d4225433ada4bb26c9b2ddab89abd5d609

nix/vast/plugins/source.json
```diff
@@ -1,7 +1,7 @@
 {
   "url": "git@github.com:tenzir/vast-plugins",
   "ref": "main",
-  "rev": "7aece543105143f21fb2f65f49c1f684d04626ed",
+  "rev": "20234dc85667a5aa55a307e712c15eb9d846ab40",
   "submodules": true,
   "shallow": true
```
tobim, line 6: You should be able to kickstart CI with:
```suggestion
  "shallow": true,
  "allRefs": true
```

Dakostu: I've used your script from https://github.com/tenzir/vast/pull/2965 and the Nix CI runs work now.

PR #2950, opened 2023-02-16
Perform some docs smithing

mavam:
This PR reworks the docs in preparation to the upcoming v3.0 release, documenting pipelines and adjusting terminology.


web/docs/use/import/README.md
```diff
@@ -25,21 +25,21 @@ For example, to import a file in JSON, use the `json` format:
...
-For example, you might want to import [Suricata EVE
-JSON](/docs/understand/formats/suricata), but skip over all `suricata.stats`
-events:
+For example, you could reduce a Suricata EVE JSON feed to just flow records:
 
 ```bash
-vast import suricata '#type != "suricata.stats"' < path/to/eve.json
-```
+vast import suricata < eve.json 'src_ip in 10.0.0.0/8 
+  | select timestamp, flow_id, src_ip, dest_ip, src_port'`
```
dominiklohmann, line 38: This is just a syntax error, but you can use one of the following instead:

```suggestion
vast import -r eve.json suricata 'src_ip in 10.0.0.0/8 
  | select timestamp, flow_id, src_ip, dest_ip, src_port'`
```

```suggestion
< eve.json vast import suricata 'src_ip in 10.0.0.0/8 
  | select timestamp, flow_id, src_ip, dest_ip, src_port'`
```

```suggestion
vast import suricata 'src_ip in 10.0.0.0/8 
  | select timestamp, flow_id, src_ip, dest_ip, src_port'` \
  < eve.json
```

web/docs/understand/language/pipelines.md
```diff
@@ -0,0 +1,129 @@
...
+Operators can produce, transform, or consume data. Think of it as UNIX pipes or
+Powershell commands where output from one command is input to the next:
+
+![Pipeline Chaining](pipeline-chaining.excalidraw.svg)
+
+VAST distinguishes three types of operators:
+
+1. **Source**: generates new data
+2. **Transformation**: modifies data
+3. **Sink**: consumes data
```
dominiklohmann, line 19: Should this be s/data/events to match the terminology used in other places? That's the element type that the operators are working on.

web/docs/understand/language/pipelines.md
```diff
@@ -0,0 +1,129 @@
...
+blocks to build nested pipelines:
+
+![Nested Pipeline](pipeline-nested.excalidraw.svg)
+
+Pipelines have a *logical* and *physical* representation. Users only specify the
+logical form, whereas the VAST defines the mapping to a physical representation
+that potentially looks differently. This distinction allows VAST to
+transparently optimize the execution. For example, filter predicates may be
+"pushed down" to reduce the amount of data that flows through a pipeline. Such
+optimizations are common practice in declartive query languages.
```
dominiklohmann, line 39:
```suggestion
optimizations are common practice in declarative query languages.
```

web/docs/understand/language/pipelines.md
```diff
@@ -0,0 +1,129 @@
...
+Here is how you write this pipeline in the VAST language:
+
+```cpp
+from vast
+| where #type == "zeek.weird" && note == "SSL::Invalid_Server_Cert"
+| summarize count(num) by id.resp_h
+| sort
+| head 20
+| write json to stdout
+```
```
dominiklohmann, line 81: I would refrain from putting this here at this point without an exclaimer immediately above it. We do not have `vast exec` as part of v3.0, and we also do not have all these operators.

web/docs/understand/language/pipelines.md
mavam: Not in this context. We've introduced this as dataflow, also in the diagram. At this high level, I don't want to add a term that needs to be defined.

Rather, I see this in a different part where talk about the element types.

web/docs/understand/language/pipelines.md
mavam: Have you read the text below? This is not even about `vast exec`. It's just an example for a source and sink, both of which are implicit in `vast export <args>`.

web/docs/use/transform/README.md
```diff
@@ -2,8 +2,36 @@
 
 VAST has powerful features for transforming [data in
 motion](#modify-data-in-motion) and [data at rest](#modify-data-at-rest). Both
-aspects rely on [pipelines](/docs/understand/query-language/pipelines) as
-building block.
+aspects rely on [pipelines](/docs/understand/language/pipelines) as building
+block.
+
+## Define a static pipeline
```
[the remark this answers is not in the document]
mavam: I removed "static.

The newly introduced terminology is actually *open pipeline*, so I think `vast.pipelines` remains valid.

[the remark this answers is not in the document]
dominiklohmann: Those are different things to me. Any pipeline definition can be used as a logical operator, independent of whether they're open or closed. Whether they're closed or opened only changes the context in which they can be instantiated.

web/docs/understand/language/pipelines.md
dominiklohmann: Fine then, I have no strong opinion on this. I would've preferred it, though. :)

mavam: Of course, you're the person building this. 😉

What's key here is a gentle introduction and leaving implementation aspects to later sections where the distinction matters.

web/docs/use/transform/README.md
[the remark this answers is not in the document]
mavam: Right now, we don't have `vast.operators` but only `vast.pipelines` in the config file. So this would probably at most be a change in a future version.

But you basically just said that using "pipelines" is not wrong:

> Any pipeline definition can be used as a logical operator

So then what's wrong with `vast.pipelines`?

dominiklohmann (approved):
Approving modulo the two open discussions. Maybe we can have a quick chat for finding a way to resolve them.

web/docs/understand/language/pipelines.md
dominiklohmann: In a quick call, we agreed to put one sentence below this saying explicitly that we do not support running closed pipelines yet, but that this will be coming soon.

web/docs/use/transform/README.md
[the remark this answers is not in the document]
dominiklohmann: In a quick call we agreed to slightly rephrase this: We want to call the building blocks operators, so users define operators here in their config (and as such we should rename this to `vast.operators` sometime in the future), because then the term pipeline unambiguously refers to something instantiated.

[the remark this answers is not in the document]
mavam: Right. I'm leaving this for now as is, because we should do one global rename from `vast.pipelines` to `vast.operators`.

PR #2951, opened 2023-02-16
Support case insensitivity in patterns

Dakostu:
This PR adds support for case-insensitive patterns via an optional `i` suffix in the pattern string.


libvast/include/vast/concept/parseable/vast/pattern.hpp
```diff
@@ -29,7 +29,12 @@ struct access::parser_base<pattern>
 
   template <class Iterator>
   bool parse(Iterator& f, const Iterator& l, pattern& a) const {
-    return pattern_parser{}(f, l, a.str_);
+    if (!pattern_parser{}(f, l, a.str_)) {
+      return false;
+    }
+    auto case_insens_parser = parsers::chr{pattern::case_insensitive_flag};
+    a.case_insensitive_ = case_insens_parser(f, l, unused);
+    return true;
```
dominiklohmann, line 37: The way this is implemented, `pattern_parser{}(...)` and `to<pattern>(...)` behave differently, because the former does not support the optional `i` flag. Please unify this.

libvast/include/vast/view.hpp
```diff
@@ -78,9 +78,9 @@ class pattern_view : detail::totally_ordered<pattern_view> {
 public:
   static pattern glob(std::string_view x);
 
-  explicit pattern_view(const pattern& x);
+  explicit pattern_view(const pattern& x, bool case_insensitive = false);
```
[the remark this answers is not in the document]
Dakostu: True, it would be a simple removal of this tautology. I implemented a simple `case_insensitive` getter method in https://github.com/tenzir/vast/pull/2951/commits/9f8007bf38c4b350f75fd061673f81b6cd53652c

dominiklohmann (commented):
I found a small parser inconsistency. Can you also add an integration test showing to work end-to-end in queries, like `vast export json '/GoOgLe.CoM/i'`?

libvast/include/vast/concept/parseable/vast/pattern.hpp
Dakostu: Can you elaborate on where the divergence lies, because I tried to unit test `to<pattern>` and all checks pass:

```cpp
TEST(to pattern) {
  auto p1 = to<pattern>("/test/");
  CHECK(p1);
  CHECK_EQUAL(p1->string(), "test");
  CHECK(!p1->case_insensitive());
  auto p2 = to<pattern>("/test/i");
  CHECK(p2);
  CHECK_EQUAL(p2->string(), "test");
  CHECK(p2->case_insensitive());
}
```

dominiklohmann: Here's an example that fails:

```cpp
auto str = std::string{"/foo/i"};
auto p = pattern_parser{};
auto result = pattern{};
CHECK(p(str, result)); // this fails
```

This is because the `pattern_parser` itself does not know about the flags, which is bad when you want to combine it. Now this isn't used anywhere in the code base currently, but it's still an easy trap to fall into. I see two possible fixes:
- Rename `pattern_parser` to `slash_delimited_string`, because that's what the parser is doing.
- Change the `pattern_parser` to also include the optional `/i` suffix and have it parse into a pattern.

I believe the former is more correct.

Dakostu: Ahh, the problem is much simpler than I thought then. I renamed the `pattern` parsers in https://github.com/tenzir/vast/pull/2951/commits/0a05c366639d88bfa94d71648e95a2bdd95361c2 and added an integration test in https://github.com/tenzir/vast/pull/2951/commits/7d6238fd08206cfe587317772ebbe46811ef005c

dominiklohmann (approved):
This works nicely for me. I have just some minor cosmetic requests now; feel free to resolve them yourself when you're done with them.

libvast/include/vast/view.hpp
```diff
@@ -80,11 +80,13 @@ class pattern_view : detail::totally_ordered<pattern_view> {
...
   explicit pattern_view(const pattern& x);
 
-  explicit pattern_view(std::string_view str);
+  explicit pattern_view(std::string_view str, bool case_insensitive = false);
 
   [[nodiscard]] bool match(std::string_view x) const;
   [[nodiscard]] bool search(std::string_view x) const;
   [[nodiscard]] std::string_view string() const;
+  [[nodiscard]] bool case_insensitive() const;
+  [[nodiscard]] std::regex generate_regex() const;
```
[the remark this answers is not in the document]
Dakostu: Removed functionality duplication in https://github.com/tenzir/vast/pull/2951/commits/d1f9a4e9c52b3c0c44153f073ee9cfa377c1c377

libvast/include/vast/pattern.hpp
```diff
@@ -53,8 +57,14 @@ class pattern : detail::totally_ordered<pattern>,
   /// @returns `true` if the pattern matches inside *str*.
   [[nodiscard]] bool search(std::string_view str) const;
 
+  /// Generates a regular expression for searching and matching.
+  /// @returns an `std::regex` object or an `std::regex_error`.
+  [[nodiscard]] std::regex generate_regex() const;
```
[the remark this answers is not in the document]
Dakostu: I turned `generate_regex()` into a `caf::error`-based `make_regex()` in https://github.com/tenzir/vast/pull/2951/commits/a1a450be750e46a017828c376ca194fc5f8f80be. Unfortunately this requires the `<regex>` header for now (the expensive include may have probably been the reason why `std::regex` generation has been scattered among the evaluation methods, but IMO there should be one place to control regex generation, especially now with flags)

PR #2952, opened 2023-02-16
Add schema information to the `/query` endpoint

dominiklohmann:
This PR adds schema information to the `/query` endpoint. Use it like this:

```bash
# Create a new query
❯ curl -XPOST -H "Content-Type: application/json" -d "{\"query\": \"taste 10 | drop timestamp\"}" http://localhost:42001/api/v0/query/new
{"id": "4b4f23d7-0807-4f3b-9994-6bf6bfb610d0"}

# Fetch next 10 results for that query incrementally
❯ curl -XGET -H "Content-Type: application/json" http://localhost:42001/api/v0/query/4b4f23d7-0807-4f3b-9994-6bf6bfb610d0/next?n=10
```

The output contains two parts: An `events` section containing the raw, row-based data as JSON and also a unique identifier for the schema of the row. The same identifier for the schema also occurs in the `schemas` section, which renders the full type definition as also available via `vast show schemas --expand`.

Fixes tenzir/issues#98

### DIFF
--- a/libvast/builtins/endpoints/query.cpp
+++ b/libvast/builtins/endpoints/query.cpp
@@ -6,14 +6,15 @@
 // SPDX-FileCopyrightText: (c) 2022 The VAST Contributors
 // SPDX-License-Identifier: BSD-3-Clause
 
-#include "vast/format/json.hpp"
-#include "vast/pipeline.hpp"
-
+#include <vast/arrow_table_slice.hpp>
 #include <vast/command.hpp>
 #include <vast/concept/convertible/to.hpp>
 #include <vast/concept/parseable/numeric.hpp>
 #include <vast/concept/parseable/to.hpp>
 #include <vast/concept/parseable/vast/expression.hpp>
+#include <vast/concept/printable/vast/json.hpp>
+#include <vast/format/json.hpp>
+#include <vast/pipeline.hpp>
 #include <vast/plugin.hpp>
 #include <vast/query_context.hpp>
 #include <vast/system/actors.hpp>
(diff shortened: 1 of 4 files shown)


libvast/builtins/endpoints/query.cpp
```diff
@@ -181,13 +150,68 @@ struct query_manager_state {
...
+        out_iter = fmt::format_to(out_iter, "}},{{");
+      first = false;
+      out_iter = fmt::format_to(
+        out_iter, "\"schema-ref\":\"{:x}\",\"definition\":", hash(schema));
+      const auto ok = printer.print(out_iter, schema.to_definition(true));
+      VAST_ASSERT_CHEAP(ok);
+    }
+    out_iter = fmt::format_to(out_iter, "}}]}}\n");
+    shippable_events_count -= written;
+    shipped_events_count += written;
```
patszt, line 259: I see that you have removed the "position" from the response.
With that being gone we can remove shipped_events_count completely as it is now not used elsewhere

dominiklohmann: Good point; I removed the variable in https://github.com/tenzir/vast/pull/2952/commits/4750258412fa7a4dfeedc30ea798f61b2a88982f

libvast/builtins/endpoints/query.cpp
```diff
@@ -181,13 +150,68 @@ struct query_manager_state {
...
-    shipped_events_count += drained_events;
-    auto events = detail::split(events_as_json, "\n");
-    return fmt::format("{{\"position\": {}, \"events\": [\n{}]}}\n",
-                       shipped_events_count, fmt::join(events, ",\n"));
+    auto printer = json_printer{{.oneline = true}};
+    auto result = std::string{};
+    auto out_iter = std::back_inserter(result);
+    auto seen_schemas = std::unordered_set<type>{};
+    auto written = size_t{0};
+    out_iter = fmt::format_to(out_iter, "{{\"events\":[");
```
patszt, line 203: I think we can just initialize result with "{"events":[" instead of initializing result + calling fmt.
Not a big deal, one line less to read though :D

patszt (approved):
The code looks to be working as intended.
I think it could be shorter/more readable if we split the json formatting into multiple pieces.
E.g:

1. Format slices
2. Format schemas
3. return fmt::format("events: {}, schenas: {}", events_json, schemas_json)

This will allocate more and will in general be slower than the current approach.
I believe the performance was selected here.
In the end i am not sure if this performance is needed or if it is significant in the grand scheme of things, but currently it is still readable so i don't mind it the way it is.

libvast/builtins/endpoints/query.cpp
dominiklohmann: I made this change.

lava:
Fixes https://github.com/tenzir/issues/issues/122

PR #2953, opened 2023-02-17
Allow whitespace before expressions in queries

dominiklohmann:


libvast/builtins/query-languages/vastql.cpp
```diff
@@ -43,7 +43,8 @@ class plugin final : public virtual query_language_plugin {
     auto parsed_expr = expression{};
     const auto optional_ws = ignore(*space);
     bool has_expr = true;
-    if (!expr(f, l, parsed_expr)) {
+    const auto expr_parser = optional_ws >> expr;
```
mavam, line 46:
```suggestion
    const auto expr_parser = optional_ws >> expr >> optional_ws;
```

Why not that?

mavam (approved):

libvast/builtins/query-languages/vastql.cpp
dominiklohmann: Because there is further code below this because there may be a pipeline after this, and that already handles the whitespace / eoi at the end. There's no need to handle this twice.

mavam: Isn't a pipeline optional?

mavam: What happens if a pipeline is not there?

dominiklohmann: Yes, but the expression as well, and then there's an `optional_ws >> eoi` at the end that we always require.

PR #2954, opened 2023-02-17
Add the --taste-size option to vast export

tobim:
The new option allows the user to set the maximum number of partitions that are part of the initial evaluation batch. This functionality has been present internally already, but was not exposed on the command line until now.

This is particularly useful for queries that contain pipeline-breaking operators, as those can yield unexpected results with the default taste size.

### DIFF
--- a/libvast/include/vast/defaults.hpp
+++ b/libvast/include/vast/defaults.hpp
@@ -83,6 +83,9 @@ inline constexpr std::string_view read = "-";
 /// Maximum number of results.
 inline constexpr size_t max_events = 0;
 
+/// Number of paritions to include in the first eval batch.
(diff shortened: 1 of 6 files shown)


dominiklohmann:
I'm not sure we actually want to add this as an option given that I don't know whether we want to support this in the medium-term, and that #2955 fixes the issue that prompted this properly.

mavam:
I think "taste size" is confusing because there is also a `taste` operator with very different semantics.

I'm feeling uneasy about exposing something that should be an implementation, and as @dominiklohmann said, is not going to be relevant soon. We also have #2955 to address our need.

tobim:
The problem that prompted me to open this is solved in a better way in #2955.

PR #2955, opened 2023-02-17
Make summarize in export block properly

dominiklohmann:
Note that this fix is for `export` only, this will stay slightly incorrect for both partition transformations and `import` until we have the new executor in place.

### DIFF
--- a/libvast/include/vast/pipeline.hpp
+++ b/libvast/include/vast/pipeline.hpp
@@ -38,7 +38,7 @@ public:
 
   void add_operator(std::unique_ptr<pipeline_operator> op);
 
-  /// Returns true if any of the pipeline operators is aggregate.
+  /// Returns true if any of the pipeline operators is blocking.
   [[nodiscard]] bool is_blocking() const;
 
   /// Tests whether the transform applies to events of the given type.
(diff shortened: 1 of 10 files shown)


libvast/src/system/exporter.cpp
```diff
@@ -232,15 +237,15 @@ exporter(exporter_actor::stateful_pointer<exporter_state> self, expression expr,
         : query_context::priority::normal;
   VAST_DEBUG("spawned exporter with {} pipelines", pipelines.size());
   self->state.pipeline = pipeline_executor{std::move(pipelines)};
```
tobim, line 239: Doesn't make sense to get results in batches in this case.
```suggestion
  self->state.pipeline = pipeline_executor{std::move(pipelines)};
  if (self->state.pipeline.is_blocking())
    self->state.query_context.taste = std::numeric_limits<uint32_t>::max();
```

libvast/include/vast/pipeline.hpp
```diff
@@ -39,7 +39,7 @@ class pipeline {
   void add_operator(std::unique_ptr<pipeline_operator> op);
 
   /// Returns true if any of the pipeline operators is aggregate.
-  [[nodiscard]] bool is_aggregate() const;
+  [[nodiscard]] bool is_blocking() const;
```
tobim, line 42: The comment needs to be adjusted.

tobim (approved):
Thanks for the fix!

libvast/src/system/exporter.cpp
dominiklohmann: Nice catch!

PR #2958, opened 2023-02-20
Fix debian package updates

tobim:
The package conflicts directive caused unintended conflicts when trying to upgrade or side-grade from the open source edition to the community edition or back.

We now only encode the edition name into the package output file name but leave the package name itself as "vast".


lava (approved):
I didn't test personally, but it looks like the way to go for now. (In the future, we probably also want to include the edition name in `vast version` )

PR #2962, opened 2023-02-21
Prune unwanted store paths from *-slim images

tobim:
The static binary contains references to some store paths that are not needed at runtime, namely `tzinfo` and `openssl.etc`.


mavam (approved):
Not sure if @dit7ya should take a deeper look here for the Nix stuff.

PR #2963, opened 2023-02-22
Ignore dense indexes in the active partition lookup path

tobim:
This PR encompasses 2 changes:
* Fix a crash in the query eval logic of the active partition (a82557cd)
* Fix a logic error causing wrong query results, also in the active partition (0e6f24eb0)

### DIFF
--- a/libvast/test/system/active_partition.cpp
+++ b/libvast/test/system/active_partition.cpp
@@ -205,8 +205,7 @@ TEST(No dense indexes serialization when create dense index in config is false) 
   CHECK_EQUAL(unbox(result), make_ids({0, 0}));
 }
 
-TEST(delegate query to store with all possible ids in partition when query is to
-       be done without dense indexer) {
+TEST(delegate query to the store) {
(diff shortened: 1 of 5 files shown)


libvast/test/system/active_partition.cpp
```diff
@@ -253,9 +253,6 @@ TEST(delegate query to store with all possible ids in partition when query is to
       FAIL(err);
     });
   REQUIRE_EQUAL(last_query_contexts.size(), 1u);
-  vast::ids expected_ids;
```
patszt, line 256: I think it is worth to also change the name of the test. I can't find how to add a suggestion to not changed code, but my proposal for the new name is:

**delegate query to the store**

dominiklohmann:
> @dominiklohmann what do You think about removing usage of the indexers for the lookup path?

I'm totally fine with it. With the slices being in memory already my intuition says that that if anything the performance loss is minor here. I would expect this to be a speedup in situations where the indexes did not help even.

Given that we want to make the active partition timeout a lot smaller and stop querying active partitions at all this is totally fine to me.

PR #2966, opened 2023-02-22
Make plugin versioning optional

tobim:
... and drop versions form bundled and Tenzir maintained plugins.


mavam (approved):
I didn't get the CMake magic, but the rest looks correct.

PR #2969, opened 2023-02-27
Bump nixpkgs to 9499706a6

tobim:
This brings in a few version updates for the static binary and vast*-slim docker images:

* GCC 12
* Arrow 11.0
* simdjson 3.1.2
* robin-map 1.2.1
* flatbuffers 22.11.23
* clang 15

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "flake-compat": {
       "flake": false,
       "locked": {
-        "lastModified": 1641205782,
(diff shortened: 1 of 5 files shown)


dit7ya (approved):
LGTM in general, and I guess waiting for all CI checks to pass is enough to test the LLVM and GCC upgrades?

tobim:
> LGTM in general, and I guess waiting for all CI checks to pass is enough to test the LLVM and GCC upgrades?

LLVM is only for dev, libraries from the `pkgsStatic` overlay will have to be rebuilt in CI now (arrow, CAF, and so on). They get uploaded to our cachix instance, so only the first CI run should take more time.

PR #2972, opened 2023-02-28
Revert accidental submodule bump revert

dominiklohmann:
This reverts a submodule bump reversion accidentally introduced in tenzir/vast#2969.


tobim:
This sets the submodule to a newer merge than we had before, so `nix/update-plugins.sh` also needs to be run.

tobim (approved):

PR #2974, opened 2023-02-28
Support pattern case insensitivity in Sigma plugin

Dakostu:
With our `pattern`s now supporting case insensitivity, this PR adds the same support to the Sigma value transpilation. Namely, all strings that contain Sigma wildcards are now turned into case-insensitive patterns, and `startswith` and `endswith` substrings are turned into case-sensitive patterns with  respective `^` and `$` metacharacters.


plugins/sigma/src/parse.cpp
```diff
@@ -369,12 +366,10 @@ caf::expected<expression> parse_search_id(const data& yaml) {
...
       auto make_predicate_expr = [&](const data& value) -> expression {
-        // Convert strings to patterns if wildcarding is present.
+        // Convert strings to case-insensitive patterns.
         if (auto str = caf::get_if<std::string>(&value))
-          if (!str->empty())
-            if (auto pat = make_pattern(*str); pat && *pat)
-              return predicate{extractor, relational_operator::equal,
-                               data{std::move(**pat)}};
+          if (auto pat = make_pattern(*str))
+            return predicate{extractor, op, data{std::move(*pat)}};
```
Dakostu, line 372: The previous version was transpiling all wildcard-strings to a predicate with the `equals` operator, but here we're using the previously defined `op` operator so e.g. our `startswith` will create a predicate with `ni <pattern>` instead of `== <pattern>`. I'm not 100% sure if that should be the correct behavior though, more like 80% sure.

plugins/sigma/tests/tests.cpp
```diff
@@ -458,9 +461,9 @@ TEST(real example) {
   auto selection1 = R"__(CommandLine ni "7z.exe a -v500m -mx9 -r0 -p")__"s;
   auto selection2a = R"__(ParentCommandLine ni "wscript.exe" && ParentCommandLine ni ".vbs")__"s;
   auto selection2b = R"__(CommandLine ni "rundll32.exe" && CommandLine ni "C:\Windows" && CommandLine ni ".dll,Tk_")__"s;
-  auto selection3 = R"__(ParentImage ni "\rundll32.exe" && ParentCommandLine ni "C:\Windows" && CommandLine ni "cmd.exe /C ")__"s;
+  auto selection3 = R"__(ParentImage == /\rundll32.exe$/ && ParentCommandLine ni "C:\Windows" && CommandLine ni "cmd.exe /C ")__"s;
   auto selection4 = R"__(CommandLine ni "rundll32 c:\windows\\" && CommandLine ni ".dll ")__"s;
-  auto specific1 = R"__(ParentImage ni "\rundll32.exe" && Image ni "\dllhost.exe")__"s;
+  auto specific1 = R"__(ParentImage == /\rundll32.exe$/ && Image == /\dllhost.exe$/)__"s;
```
mavam, line 466: Out of curiosity: is the `\` now triggering an escape sequence, i.e., is the first character of `/\rundll32.exe$/` a `\r` or `r`? I'd suggest to add a unit test that confirms the expected behavior.

plugins/sigma/tests/tests.cpp
```diff
@@ -158,7 +158,25 @@ TEST(modifier - re) {
...
   )__";
   auto search_id = to_search_id(yaml);
-  auto expected = to_expr("foo == /^.*$/");
+  auto expected = to_expr("foo == /^.*$/i");
+  CHECK_EQUAL(search_id, expected);
+}
+
+TEST(modifier - re 2) {
+  auto yaml = R"__(
+    foo|re: '.*foobar.*'
```
mavam, line 167: I would argue that this should instead map to `/foobar/ in foo` (minus operand flipping).

Why? Because the regex can be optimized more easily.

plugins/sigma/tests/tests.cpp
```diff
@@ -158,7 +158,25 @@ TEST(modifier - re) {
...
+  )__";
+  auto search_id = to_search_id(yaml);
+  auto expected = to_expr("foo == /.*foobar.*/i");
```
mavam, line 170: Sanity check: is the `|re` modifier also implying case insensitivity in the spec?

mavam: I took another look at the spec and am bit confused: it seems there is no way to produce a case-insensitive regex.

I've asked around and got the answer that "you can start your regex with `(?i)` if you want case insensitivity". I would argue that we don't do that until a user asks for it, though, as it's not spec-compliant.

plugins/sigma/src/parse.cpp
```diff
@@ -316,33 +323,51 @@ caf::expected<expression> parse_search_id(const data& yaml) {
...
                                  "utf16 not yet implemented");
-        } else if (*i == "endswith" || *i == "startswith" || *i == "re") {
+        } else if (*i == "startswith") {
+          auto to_re = [](const data& d) -> caf::expected<data> {
+            auto f = detail::overload{[](const auto& x) -> caf::expected<data> {
+              auto str = fmt::format("^{}.*", to_string(x));
+              auto result = pattern::make(str);
+              if (!result)
+                return std::move(result.error());
+              return std::move(*result);
```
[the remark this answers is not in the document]
dominiklohmann: The early return in case of an error is much more common throughout the code base, actually, at least in code added in the last year or so.

An NB on your NB: We most definitely should be moving the result of the error function on an expected as that returns by lvalue ref and errors are pretty expensive to copy. So much so, that a missing move in a hot loop once caused the summarize operators to be 20% slower than it could've been in a code pattern just like this.

tobim:
> and startswith and endswith substrings are turned into case-sensitive patterns with respective ^ and $ metacharacters.

According to https://socprime.com/blog/sigma-rules-the-beginners-guide/ it should still be insensitive in case it is in the detection component.

mavam:
@tobim there are either regex patterns (`|re`) or strings. Only for the latter there's case insensitivity. For the former, Sigma relies on the respective regex engines to have inline pattern flags for things like case insensitivity.

Dakostu:
Moving this comment to the top-level:

>We had a huddle regarding this and the results are:
>* for `re`, do not set case insensitivity; the RE engine will parse case-insensitivity in the inline string.
>* string values should always be transpiled into case-insensitive patterns, regardless of whether they contain wildcards or not.

libvast/include/vast/detail/string.hpp
```diff
@@ -79,6 +79,12 @@ std::string byte_escape_all(std::string_view str);
 /// @relates byte_escape bytes_escape_all
 std::string byte_unescape(std::string_view str);
 
+/// Escapes a string by splitting all singular control characters into two
+/// chars, e.g. the character \n becomes a two-character string "\n".
```
mavam, line 83:
```suggestion
/// chars, e.g., the character '\n' becomes a two-character string "\n".
```

plugins/sigma/src/parse.cpp
```diff
@@ -321,18 +321,47 @@ caf::expected<expression> parse_search_id(const data& yaml) {
...
+            }};
+            return caf::visit(f, d);
+          };
+          transforms.emplace_back(to_re);
+        } else if (*i == "endswith") {
+          op = relational_operator::equal;
+          auto to_re = [](const data& d) -> caf::expected<data> {
+            auto f = detail::overload{[](const auto& x) -> caf::expected<data> {
+              auto str = vast::detail::control_char_escape(to_string(x));
+              str = fmt::format(".*{}$", str);
```
mavam, line 342: Logically, wouldn't it make sense to wrap the translated pattern into `.*{}$` _after_ you've converted it into a pattern? Otherwise you will treat it as string input, no?

plugins/sigma/tests/tests.cpp
```diff
@@ -155,26 +163,62 @@ TEST(modifier - re) {
...
 
+TEST(modifier - startswith - proper backslash escaping) {
+  auto yaml = R"__(
+    foo|startswith: "C:\rundll32"
+  )__";
+  auto search_id = to_search_id(yaml);
+  auto expected = to_expr(R"(foo == /^C:\rundll32.*/i)");
+  CHECK_EQUAL(search_id, expected);
+  auto str = to_string(expected);
+  CHECK_EQUAL(str, "foo == /^C:\\rundll32.*/i");
```
mavam, line 201: Can you explain the escaping difference? I'm not understanding why only the second `rundll` example has a double `\\`.

mavam (approved):
I think it's there! Just some smaller comments before this can ship.

plugins/sigma/tests/tests.cpp
```diff
@@ -62,7 +70,7 @@ TEST(maps - single value) {
 TEST(maps - empty value) {
   auto yaml = "foo: ''";
   auto search_id = to_search_id(yaml);
-  auto expected = to_expr("foo == \"\"");
+  auto expected = to_expr("foo == //i");
```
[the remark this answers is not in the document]
dominiklohmann: Yeah, but not in this PR. We can do this in the future. @dakostu, can you write an issue for this?

mavam (approved):
Looks like the last 🐛 got fixed. 🚢 it.

PR #2975, opened 2023-02-28
Add a vast-manager docker image

lava:
This PR builds up with a refactoring before adding the titular addition and closes with some minor cleanups.

1. The first two commits rework the `configure` job, specifically the logic to decide which checks to run.
2. The third commit introduces the vast-manager Docker image. At the same time we push a few policies up to the build matrix in the `docker-vast` job.
3. The sixth commit applies some fixes for the workflow dispatch trigger.

Notes for the reviewer:
In order to test whether the configuration of the pull request trigger works as desired I added the following diff in between:

```diff
commit d9be5cfda89c59d43966cd8b7cfb88a1674ac0c6
Author: Tobias Mayer <tobim@fastmail.fm>
Date:   Fri Mar 3 14:32:30 2023 +0100

    test commit, check triggers

diff --git a/.github/workflows/vast.yaml b/.github/workflows/vast.yaml
index 114f66d6f..a1d9d2938 100644
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -219,10 +219,10 @@ jobs:
           echo "::notice sourcing configure helpers"
           source ./.github/workflows/configure_helpers.bash
           # Run all checks if this file changed.
-          if is_changed .github/workflows/vast.yaml; then
-            run_all_checks
-            exit 0
-          fi
+          #if is_changed .github/workflows/vast.yaml; then
+          #  run_all_checks
+          #  exit 0
+          #fi
           run_if_changed python "python/"
           run_if_changed vast-io "web/" "python/"
           run_if_changed changelog "CHANGELOG.md" "changelog/" "web/" "python/"
diff --git a/web/README.md b/web/README.md
index cf4d9f523..78cf11b27 100644
--- a/web/README.md
+++ b/web/README.md
@@ -1,3 +1,5 @@
+test
+
 # VAST Website
 
 This directory contains code and content of the official VAST website at
```

Which produced the expected build graph (https://github.com/tenzir/vast/actions/runs/4327352101):
image

TODO:
- [ ] Onboard the vast-manager-slim image
- [ ] Make it possible to choose an alternative vast-plugins rev for the workflow dispatch
- [ ] Make it possible to select editions for the workflow dispatch
- [x] Try to get rid of deprecation warnings from the aws-actions (use unreleased versions?)


.github/workflows/vast.yaml
```diff
@@ -140,110 +142,107 @@ jobs:
...
-            run_vast_nix=true
-          elif [[ $GITHUB_EVENT_NAME == "workflow_dispatch" ]]; then
-            run_vast_nix=true
-          fi
-          echo "run-vast-nix=${run_vast_nix}" >> $GITHUB_OUTPUT
+          run_docker_compose="$(any ${run_vast} ${run_vast_plugins} ${run_python})"
+          run_if_changed_default docker-compose $run_docker_compose "docker/" "Dockerfile" ".dockerignore"
+          run_docker_vast="$(any ${run_docker_compose} ${run_vast_io} ${run_python})"
+          run_if_changed_default docker-vast $run_docker_vast
+          run_if_changed_default regression-tests $run_docker_vast
```
tobim, line 238: I realized these invocations with out passing paths aren't tested yet.
Just leaving that as a comment for myself.

.github/workflows/configure_helpers.bash
```diff
@@ -0,0 +1,33 @@
...
+  local default="$2"
+  shift
+  shift
+  job_shellvar="run_$(echo "$job" | tr '-' '_')"
+  if is_changed "${@}"; then
+    declare $job_shellvar="true"
+    echo "run-${job}=true" >> $GITHUB_OUTPUT
+  else
+    declare $job_shellvar="false"
+  fi
```
tobim, line 16: This needs to use the default.

.github/workflows/vast.yaml
```diff
@@ -151,93 +158,97 @@ jobs:
...
-          run_vast_nix=${run_vast}
-          if ${run_vast_plugins}; then
-            run_vast_nix=true
-          elif ! is_unchanged nix/ flake.lock *.nix; then
-            run_vast_nix=true
-          elif [[ $GITHUB_EVENT_NAME == "workflow_dispatch" ]]; then
-            run_vast_nix=true
-          fi
-          echo "run-vast-nix=${run_vast_nix}" >> $GITHUB_OUTPUT
+          run_docker_compose="$(any ${run_vast} ${run_vast_plugins} ${run_python})"
```
tobim, line 234: Verify that this dependency is really needed.

.github/workflows/vast.yaml
```diff
@@ -151,93 +158,97 @@ jobs:
...
+          run_if_changed_default docker-compose $run_docker_compose "docker/" "Dockerfile" ".dockerignore"
+          run_docker_vast="$(any ${run_docker_compose} ${run_vast_io} ${run_python})"
```
tobim, line 236: These dependencies should be inverted.
If `docker` is run, then all 3 others should run too. If for example `docker-compose` is run, then it should fall back on the image from the merge base in case `docker` is not run.

.github/workflows/vast.yaml
```diff
@@ -461,24 +496,36 @@ jobs:
...
+        with:
+          role-to-assume: ${{ secrets.AWS_INFRASTRUCTURE_GITHUB_ACTIONS_ROLE }}
+          aws-region: ${{ secrets.AWS_INFRASTRUCTURE_DEFAULT_REGION }}
+      - name: Login to AWS ECR
+        # if: contains(matrix.docker-vast.registries, '622024652768.dkr.ecr.eu-west-1.amazonaws.com')
+        if: matrix.docker-vast.name == 'vast-manager'
+        uses: aws-actions/amazon-ecr-login@v1
       - name: Publish Docker Images
-        if: ${{ github.event_name == 'push' || github.event_name == 'release' }}
+        if: github.event_name == 'push' || github.event_name == 'release' || (matrix.docker-vast.name == 'vast-manager' && inputs.upload-docker-image-to-ecr)
```
tobim, line 509: Make these conditionals more precise. Make the registry selection more fine-grained in the matrix.

Dockerfile
```diff
@@ -54,6 +54,8 @@ RUN cmake -B build -G Ninja \
       ${VAST_BUILD_OPTIONS} \
```
tobim, line 54: Move this build args to the end so we can override defaults.

.github/workflows/vast.yaml
```diff
@@ -235,13 +235,14 @@ jobs:
...
           run_if_changed_default docker-vast $run_docker_vast
+          run_if_changed_default regression-tests $run_docker_vast
           run_if_changed vast-nix "$VAST_SOURCES[@]" plugins/ contrib/vast-plugins/ flake.nix flake.lock nix/
 
   policy-enforcement:
     name: Policy Enforcement
     needs:
       - configure
-    if: ${{ needs.configure.outputs.run-policy-enforcement == 'true' }}
+    if: github.event_name != 'workflow_dispatch'
```
tobim, line 245: This should be moved back to `run-policy-enforcement` and set from the configuration job.

.github/workflows/vast.yaml
```diff
@@ -461,24 +496,36 @@ jobs:
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USER }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - name: Configure AWS Credentials
+        # if: contains(matrix.docker-vast.registries, '622024652768.dkr.ecr.eu-west-1.amazonaws.com')
+        if: matrix.docker-vast.name == 'vast-manager'
+        uses: aws-actions/configure-aws-credentials@v1-node16
+        with:
+          role-to-assume: ${{ secrets.AWS_INFRASTRUCTURE_GITHUB_ACTIONS_ROLE }}
+          aws-region: ${{ secrets.AWS_INFRASTRUCTURE_DEFAULT_REGION }}
```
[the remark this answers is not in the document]
rdettai: done

.github/workflows/vast.yaml
```diff
@@ -434,44 +468,56 @@ jobs:
...
+      - name: Login to Docker Hub
+        if: github.event_name == 'push' || github.event_name == 'release'
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USER }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - name: Configure AWS Credentials
+        # if: contains(matrix.docker-vast.registries, '622024652768.dkr.ecr.eu-west-1.amazonaws.com')
+        if: matrix.docker-vast.name == 'vast-manager'
+        uses: aws-actions/configure-aws-credentials@v1-node16
```
rdettai, line 500: any reason why we are not using `aws-actions/configure-aws-credentials@v2` here?

tobim: That version did not exist at the time I wrote this.

.github/workflows/vast.yaml
tobim: Will do that at a later time.

.github/workflows/vast.yaml
tobim: Will switch to a reusable workflow that supports this, but that has to land on master first it seems.

Dockerfile.extra
```diff
@@ -0,0 +1,58 @@
...
+      cmake --build build-netflow --parallel && \
+      cmake --install build-netflow --strip && \
+      rm -rf build-netflow
+
+# -- vast-ce -------------------------------------------------------------------
+
+FROM production AS vast-ce
+
+COPY --from=plugins --chown=vast:vast $PREFIX/lib/vast/plugins/libvast-plugin-matcher.so $PREFIX/lib/vast/plugins/
+COPY --from=plugins --chown=vast:vast $PREFIX/share/doc/matcher $PREFIX/share/doc/matcher
```
dominiklohmann, line 42: I find this to be incredibly brittle. Can we instead of listing all installed files manually here install the proprietary plugins into a separate prefix and then merge them into VAST's install prefix afterwards?

Dockerfile
```diff
@@ -51,16 +41,18 @@ ENV PREFIX="/opt/tenzir/vast" \
...
+      -D VAST_ENABLE_AVX2_INSTRUCTIONS:BOOL="OFF" \
       -D VAST_ENABLE_UNIT_TESTS:BOOL="OFF" \
       -D VAST_ENABLE_DEVELOPER_MODE:BOOL="OFF" \
       -D VAST_ENABLE_BUNDLED_CAF:BOOL="ON" \
       -D VAST_ENABLE_BUNDLED_SIMDJSON:BOOL="ON" \
       -D VAST_ENABLE_MANPAGES:BOOL="OFF" \
       -D VAST_ENABLE_PYTHON_BINDINGS_DEPENDENCIES:BOOL="ON" \
-      -D VAST_PLUGINS:STRING="plugins/*" && \
+      -D VAST_PLUGINS:STRING="plugins/*" \
+      ${VAST_BUILD_OPTIONS} && \
```
dominiklohmann, line 55: Why was this moved to the end?

Dockerfile
```diff
@@ -51,16 +41,18 @@ ENV PREFIX="/opt/tenzir/vast" \
...
+      -D VAST_ENABLE_AVX_INSTRUCTIONS:BOOL="OFF" \
+      -D VAST_ENABLE_AVX2_INSTRUCTIONS:BOOL="OFF" \
       -D VAST_ENABLE_UNIT_TESTS:BOOL="OFF" \
       -D VAST_ENABLE_DEVELOPER_MODE:BOOL="OFF" \
       -D VAST_ENABLE_BUNDLED_CAF:BOOL="ON" \
       -D VAST_ENABLE_BUNDLED_SIMDJSON:BOOL="ON" \
       -D VAST_ENABLE_MANPAGES:BOOL="OFF" \
       -D VAST_ENABLE_PYTHON_BINDINGS_DEPENDENCIES:BOOL="ON" \
-      -D VAST_PLUGINS:STRING="plugins/*" && \
+      -D VAST_PLUGINS:STRING="plugins/*" \
```
dominiklohmann, line 54: Should this be removed in favor of the new default that includes the plugins in `contrib/`? If not, then we probably shouldn't copy `contrib/` into the Dockerfile in the first place.

Dockerfile.extra
```diff
@@ -0,0 +1,58 @@
...
+COPY --from=plugins --chown=vast:vast $PREFIX/lib/vast/plugins/libvast-plugin-netflow.so $PREFIX/lib/vast/plugins/
+COPY --from=plugins --chown=vast:vast $PREFIX/share/doc/netflow $PREFIX/share/doc/netflow
+
+# -- vast-ee -------------------------------------------------------------------
+
+FROM vast-ce AS vast-ee
```
dominiklohmann, line 48: These stages are lacking documentation at https://vast.io/docs/setup/deploy/docker#build-your-own-vast-image

Dockerfile.extra
```diff
@@ -0,0 +1,58 @@
+# -- plugins -------------------------------------------------------------------
```
dominiklohmann, line 1: IIUC this file isn't currently used anywhere; I'd like to see how you plan to call into this from CI, otherwise it's not really possible to review this PR.

.github/workflows/vast.yaml
```diff
@@ -140,110 +142,107 @@ jobs:
...
+          # Run all checks if this file changed.
+          if is_changed .github/workflows/vast.yaml; then
+            run_all_checks
+            exit 0
           fi
-          echo "run-vast-plugins=${run_vast_plugins}" >> $GITHUB_OUTPUT
-          # Build vast if no artifact is present in google store
+          run_if_changed python "python/"
+          run_if_changed vast-io "web/" "python/"
+          run_if_changed changelog "CHANGELOG.md" "changelog/" "web/" "python/"
```
dominiklohmann, line 219:
```suggestion
          run_if_changed changelog "changelog/" "web/" "python/"
```

That file doesnt exist anymore, actually.

dominiklohmann (commented):
Just a first look through the PR; I think this is still pretty early, but I'm excited to see how you're planning to integrate the Dockerfile.extra into CI and also for local use.

Dockerfile
```diff
@@ -54,6 +54,8 @@ RUN cmake -B build -G Ninja \
       ${VAST_BUILD_OPTIONS} \
       -D CMAKE_INSTALL_PREFIX:STRING="$PREFIX" \
       -D CMAKE_BUILD_TYPE:STRING="Release" \
+      -D VAST_ENABLE_AVX_INSTRUCTIONS:BOOL="OFF" \
+      -D VAST_ENABLE_AVX2_INSTRUCTIONS:BOOL="OFF" \
```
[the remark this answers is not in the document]
tobim: Why should images produced in CI differ from those that you build locally? If you want to build an image with AVX instructions you can now override it with `VAST_BUILD_OPTIONS`.

Dockerfile.extra
tobim: They are not usable for anyone without access to private repositories. Adding public docs for them would just give people the wrong impression.

Dockerfile.extra
tobim: It's documented in the commit message that introduces the file, and you can also see it in the `docker` CI workflow.

Dockerfile.extra
tobim: Are we certain that none of the plugins makes use of the install prefix in any way? Maybe not now, but maybe in the future?
I thought about 2 ways to do this cleanly:
1. use the install manifest to generate a tarball that can be copied and extracted in the new image.
2. Use a dedicated build context per plugin and move the existing stuff in the prefix out of the way before installing.
3. Use the install manifest to copy all installed files of a plugin into a dedicated tree and merge it back with the `COPY` directive.

I think I like (3) best, because it has the least overhead while not installing a nasty trap for our future selves.

tobim: I forgot about `DESTDIR`, which should make this trivial.

Dockerfile.extra
dominiklohmann: I think you don't need the extra Dockerfile at all if you just add a stage at the very end that says `FROM production AS vast`.

lava:
Closing this since its for an outdated fleet architecture.

PR #2977, opened 2023-02-28
Fix infinite recursion in the record algebra parser

tobim:
Fixes parsing of
```
type broken = record {
  test: : string
}
```
as reported by @satta.


satta:
That was fast, thanks :)

dominiklohmann (approved):
Thanks for providing a test for this as well.

PR #2981, opened 2023-03-01
Add a new /openapi.json endpoint to the web plugin

lava:
This allows easier integration with various OpenAPI renderers.

### DIFF
--- a/plugins/web/src/server_command.cpp
+++ b/plugins/web/src/server_command.cpp
@@ -386,7 +386,7 @@ auto server_command(const vast::invocation& inv, caf::actor_system& system)
       return request->create_response(restinio::status_temporary_redirect())
         .append_header(restinio::http_field::server, "VAST")
         .append_header_date_field()
-        .append_header(restinio::http_field::location, "/index")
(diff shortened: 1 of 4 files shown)


plugins/web/src/server_command.cpp
```diff
@@ -385,7 +386,16 @@ auto server_command(const vast::invocation& inv, caf::actor_system& system)
       return request->create_response(restinio::status_temporary_redirect())
         .append_header(restinio::http_field::server, "VAST")
         .append_header_date_field()
-        .append_header(restinio::http_field::location, "/index")
+        .append_header(restinio::http_field::location, "/api/v0/status")
```
dominiklohmann, line 389: Sneaky, but I think this is a good change.

dominiklohmann (approved):
Works for me locally! Thanks 🙏

image

plugins/web/src/server_command.cpp
lava: We do have `api_version::latest` that we could use here, but printing it here requires writing a new enum formatter so I'd probably wait with this until we introduce `v1`.

PR #2982, opened 2023-03-01
Open the metrics file sink in append mode

tobim:
This should make the metrics sink more compatible with logrotate. THe copytruncate mode should no longer cause NULL bytes to appear.


dominiklohmann (approved):
Straightforward change, this makes sense to me.

PR #2983, opened 2023-03-01
Add internal REST proxy to the node

lava:
Allow transporting REST endpoint requests via the CAF protocol, indepentend of the web plugin.

There are alternatives options in how deep we want to integrate this into the node:

* We could funnel all requests (ie. also those coming from the `web` plugin) through this handler in the node, and get rid of the need to spawn handlers ad-hoc at the price of one additional routing indirection.
* We could also decouple this a bit more by running this internal proxy as a separate component in the node.

I went somewhere in the middle with this PR, but I don't really have good arguments either way.


libvast/include/vast/system/actors.hpp
```diff
@@ -395,6 +395,13 @@ using datagram_source_actor = typed_actor_fwd<
 using node_actor = typed_actor_fwd<
   // Run an invocation in the node.
   auto(atom::run, invocation)->caf::result<caf::message>,
+  // Execute a REST endpoint on this node.
+  // This allows other nodes to access REST endpoints of this node over
+  // the CAF protocol, even if this node is not running a web server.
+  // Note that nodes connected via CAF trust each other completely,
+  // so this skips all authorization and access control mechanisms
+  // that come with HTTP(s).
+  auto(atom::proxy, http_request_description)->caf::result<std::string>,
```
dominiklohmann, line 404: IIUC the web plugin already is a component plugin; wouldn't it make more sense to add this handler to that component so we do not pollute the interface of the node actor itself?

lava: The web plugin has an `authorizer` component that runs in the node, but the request dispatching itself is currently handled inside the plugin. (and I'm not sure it's a good idea to keep this, the main reason is so it can persist its state)

But we can of course introduce a new `rest_proxy` component, that's the second option listed in the PR description. It probably shouldn't depend on the `web` plugin, since all the endpoints are also native plugins. The main downside would be that it introduces one additional routing hop between the web plugin and the endpoint, so increased latency.

libvast/include/vast/system/builtin_rest_endpoints.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+namespace vast::system {
+
+// All endpoint ids of the built-in VAST endpoints.
+
+enum class query_endpoints : uint64_t { new_, next };
+
+enum class status_endpoints : uint64_t {
+  status,
+};
```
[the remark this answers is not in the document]
lava: Ideally this would be in a header file belonging to the plugin itself, but I'm not sure we can do that with our current scaffolding for builtin plugins?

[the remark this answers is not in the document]
dominiklohmann: It's not possible, but can we proxy without needing these IDs instead? This looks rather brittle. When I want to proxy "/query/new", shouldn't it suffice to say that I want to proxy "/query/new" instead of saying that I want to proxy "/query" with the endpoint ID `query_endpoints::new_`?

libvast/include/vast/system/actors.hpp
dominiklohmann: I think I would prefer the second option from a code splitting perspective. Can you explain the extra hop? That doesn't quite make sense to me yet.

libvast/include/vast/system/builtin_rest_endpoints.hpp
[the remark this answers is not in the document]
lava: The underlying idea is that we don't want to duplicate the routing logic in the server to decide which path corresponds to which endpoint. And this is not trivial, since you can have the same path for different endpoints (ie. GET, POST), and can have endpoints with non-static paths (`/query/6c7bda29-67bc-48b5-9167-cc34f5b4c640/next`).

One approach to avoid this could be to add a method to build a unique canonical path (that looks like `GET /query/:id/next` from an endpoint and then string-match based on that.

Dockerfile
```diff
@@ -26,16 +26,6 @@ COPY vast ./vast
...
 
-# Resolve repository-internal symlinks.
-# TODO: We should try to get rid of these long-term, as Docker does not work
-# well with repository-internal symlinks. The the integration test symlinks we
-# can get rid of by copying the integration test directory to the build
-# directory when building VAST.
-RUN for f in contrib/vast-plugins/*/CMakeLists.txt; do \
-      name="$(basename "$(dirname "$f")")"; \
-      ln -sf "../contrib/vast-plugins/$name" "plugins/$name"; \
-    done
```
dominiklohmann, line 37: This loop is not unnecessary. It allows `docker build .` to build an image with all proprietary plugins as you've checked them out locally.

Docker's COPY command does not copy symlinks.

The comment above, however, should probably be reworded.

lava: Experimentally it does work with symlinks, at least on my machine (docker 23.0.1)

That said, both versions are not ideal as they require you to modify your working tree in order to not build some of the plugins (preventing interactive rebases in the meantime). I think ideally we should have a script like `scripts/setup-closed-source-plugins.sh` that checks out the submodule and then symlinks (or copies) the plugins as needed.

lava: I used a different approach for development now, we should eventually address this in a separate PR.

lava:
Obsoleted by recent architecture changes; successor PR is #3073

PR #2984, opened 2023-03-02
Add extractor predicates

jachris:
Extractors such as `id.orig_h` and `:uint64` are now allowed where a predicate is expected. They expand to `id.orig_h != nil` and `:uint64 != nil`, respectively.

Closes tenzir/issues#50.


libvast/src/concept/parseable/vast/expression.cpp
```diff
@@ -114,23 +115,43 @@ static expression expand(data x) {
...
+  return predicate{
+    std::move(operand),
+    relational_operator::not_equal,
+    predicate::operand{data{}},
+  };
+}
+
+auto make_field_char_parser() {
   using parsers::chr;
+  return parsers::alnum | chr{'_'} | chr{'-'} | chr{':'};
```
tobim, line 128: I find it strange that this is out of sync with the `field_name` from the `legacy_type` parser:
```
  auto field_name = parsers::identifier | parsers::qqstr;
```
The identifier parser is defined standalone as:
```
constexpr inline auto identifier_char = (alnum | ch<'_'> | ch<'.'>);
constexpr inline auto identifier = +identifier_char;
```
It would be great to turn this into a hierarchy that builds on shared blocks. Alternatively the relationship needs to be documented in both places.

jachris: We agreed to resolve this issue later, as part of tenzir/issues#30.

libvast/src/concept/parseable/vast/expression.cpp
```diff
@@ -114,23 +115,44 @@ static expression expand(data x) {
...
+
+auto make_field_char_parser() {
   using parsers::chr;
-  using namespace parser_literals;
-  // clang-format off
   // TODO: Align this with identifier_char.
-  auto field_char = alnum | chr{'_'} | chr{'-'} | chr{':'};
+  return parsers::alnum | chr{'_'} | chr{'-'} | chr{':'};
+}
+
```
tobim, line 131:
```suggestion

// Creates a parser that returns an operand containing either a type or field extractor.
```

web/docs/understand/language/expressions.md
```diff
@@ -185,6 +185,14 @@ Value predicates of type `subnet` expand more broadly. Given a subnet
 This makes it easier to search for IP addresses belonging to a specific subnet.
 :::
 
+### Extractor Predicates
+
+Similarly, predicates that test for the existance of a field or type can be
+written tersely as well. That is, every predicate `E != nil`, where `E` is a
+field or type extractor, can also be equivalently written as `E`. For example,
+if the field extractor `id.orig_h` is used as a predicate, it expands to
+`id.orig_h != nil`. The type extractor `:uint64` expands to `:uint64 = nil`.
```
tobim, line 194: To be honest I find "Extractor Predicate" to be quite impenetrable.
It would make more sense to me if we explain this feature as a "syntactic sugar" or "short form" of type and field extractors respectively.

mavam: It's an advanced concept anyway, so I think the "weight" of the terminology is not that important in this case.

libvast/test/expression_parseable.cpp
```diff
@@ -165,11 +165,46 @@ TEST(parseable - value predicate) {
...
+  CHECK_EQUAL(pred->rhs, predicate::operand{data{}});
+}
+
+TEST(parseable - custom type extractor predicate) {
+  expression expr;
+  CHECK(parsers::expr(":foo.bar"s, expr));
+  auto pred = caf::get_if<predicate>(&expr);
+  REQUIRE(pred);
+  auto extractor = caf::get_if<type_extractor>(&pred->lhs);
+  REQUIRE(extractor);
```
tobim, line 207: One more check please.
```suggestion
  REQUIRE(extractor);
  auto expected = type{"foo.bar", type{}};
  CHECK_EQUAL(extractor->type, expected);
```

tobim (approved):
Thank you for adding this nice opportunity for query golfing.

PR #2985, opened 2023-03-02
Use `re2::RE2` over `std::regex`

dominiklohmann:
This makes regex evaluation _a lot_ faster on my machine; around 25% for something trivial like `uid == /C.*/`, and around 75% faster for more complex regular expressions like `ProcessGuid == /\{[0-9a-f]{8}-[0-9a-f]{4}-5ec2-7.15-[0-9a-f]{12}\}/`


cmake/Findre2.cmake
```diff
@@ -0,0 +1,103 @@
...
+endif ()
+if (re2_FIND_QUIETLY)
+  list(APPEND find_package_args QUIET)
+endif ()
+set(_vast_cmake_module_path "${CMAKE_MODULE_PATH}")
+unset(CMAKE_MODULE_PATH)
+find_package(re2 ${find_package_args})
+set(CMAKE_MODULE_PATH "${_vast_cmake_module_path}")
+if (re2_FOUND)
+  set(re2_FOUND TRUE)
```
tobim, line 34: I don't get it.

cmake/Findre2.cmake
```diff
@@ -0,0 +1,103 @@
...
+set(find_package_args)
+if (re2_FIND_VERSION)
+  list(APPEND find_package_args ${re2_FIND_VERSION})
+endif ()
+if (re2_FIND_QUIETLY)
+  list(APPEND find_package_args QUIET)
+endif ()
+set(_vast_cmake_module_path "${CMAKE_MODULE_PATH}")
+unset(CMAKE_MODULE_PATH)
+find_package(re2 ${find_package_args})
```
tobim, line 31: Can't you just add `CONFIG` instead of the module path dance?

cmake/Findre2.cmake
tobim: Ah, I see this is the module from Arrow. I think you can remove this line.

tobim (approved):
Please add this so the non-static nix package continues to build:
```diff
diff --git a/nix/vast/default.nix b/nix/vast/default.nix
index a51acccf5..c83afe6e2 100644
--- a/nix/vast/default.nix
+++ b/nix/vast/default.nix
@@ -24,6 +24,7 @@
     jemalloc,
     libunwind,
     xxHash,
+    re2,
     python3,
     jq,
     tcpdump,
@@ -107,6 +108,7 @@
           libpcap
           libunwind
           libyamlcpp
+          re2
           restinio
         ];
         propagatedBuildInputs = [
```

cmake/Findre2.cmake
dominiklohmann: Yeah, definitely. Not sure what they were thinking with this one.

cmake/Findre2.cmake
dominiklohmann: Good point, that is probably even more correct.

PR #2988, opened 2023-03-03
Prepare repository for VAST v3.0.0-rc1

dominiklohmann:
After a longer than usual time between releases, we are finally ready to release the first candidate for VAST v3.0. This prepares the repository for that release.


scripts/prepare-release
```diff
@@ -44,9 +44,8 @@ fi
 # Create a new branch that bases off of current origin/master.
 git fetch origin master
 git switch -C "topic/release-${new_version}" origin/master
-
 # Update the fallback version and its ancestor count.
-rev_count="$(($(git rev-list --count) + 1))"
```
dominiklohmann, line 49: @tobim please ack this change: What we had before did not work.

dominiklohmann: Talked about this in a huddle; rev-count is a very confusing name, as this is actually ancestor-count + 1. I fixed the script and updated the comment to reflect that.

tobim: We resolved it by passing `HEAD` to the command.

PR #2990, opened 2023-03-06
Remove the /export endpoint

Dakostu:
This PR removes all `/export` REST endpoints, with formatting functionality ported over to the `/query` endpoint.


libvast/include/vast/defaults.hpp
```diff
@@ -166,8 +166,8 @@ inline constexpr const size_t rotate_files = 3;
 // -- constants for the various builtin REST endpoints -------------------------
 namespace rest {
 
-/// Settings related to the `/export` endpoint.
-namespace export_ {
+/// Settings related to the `/query` endpoint.
+namespace query_ {
```
lava, line 170: The underscore was just a relict of `export` being a keyword ;) But it also doesn't hurt to leave it in.

lava (approved):
Looks good to me

libvast/include/vast/defaults.hpp
Dakostu: I'll remove it - no fancy underscores necessary.

PR #2995, opened 2023-03-07
Set CAF middleman connection timeout to 120 seconds

patszt:
The VAST server seems to sometimes lose connection with the import clients under heavy load. Changing the default timeout from 30 seconds to 120 seconds may prevent that.


dominiklohmann (approved):
Thanks!

PR #3000, opened 2023-03-08
Remove a test only covers std::chrono

tobim:


dominiklohmann (approved):
There goes the dream of having #3000 be exactly the v3.0 release PR. 🥹

PR #3001, opened 2023-03-08
Switch main and fallback code paths for erasing partitions

tobim:
Going through the `store_header` in the partition file requires file system reads. This affects query performance, and inversely, a high query load may cause the disk monitor's throughput to drop below the ingest rate.

Now, we check whether the store file is located at one of a few probable paths before falling back to the logic described above.

### DIFF
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -1643,8 +1643,9 @@ index(index_actor::stateful_pointer<index_state> self,
                 self->state.filesystem, caf::infinite, atom::mmap_v, path)
               .then(
                 [=](const chunk_ptr& chunk) mutable {
-                  VAST_DEBUG("{} erased partition {} from filesystem", *self,
-                             partition_id);
+                  VAST_DEBUG("{} mmapped partition {} to extract store path "
+                             "for erasure",
+                             *self, partition_id);
                   if (!chunk) {
                     erase_dense_index_file();
                     rp.deliver(caf::make_error( //
@@ -1679,9 +1680,8 @@ index(index_actor::stateful_pointer<index_state> self,
                   rp.delegate(partition_actor, atom::erase_v);
                 },
                 [=](caf::error& err) mutable {
-                  VAST_WARN("{} failed to erase partition {} from "
-                            "filesystem: "
-                            "{}",
+                  VAST_WARN("{} failed to load partition {} for erase fallback "
+                            "path: {}",
                             *self, partition_id, err);
                   erase_dense_index_file();
                   rp.deliver(std::move(err));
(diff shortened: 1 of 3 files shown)


dominiklohmann (approved):
I'm still a bit reluctant to switch the fallback and main paths, but as you said, let's switch this back once the URIs are available in the catalog.

libvast/src/system/index.cpp
```diff
@@ -1584,17 +1591,67 @@ index(index_actor::stateful_pointer<index_state> self,
...
+                  VAST_WARN("{} failed to erase partition "
+                            "synopsis {} at {}: {}",
+                            *self, partition_id, synopsis_path, err);
+                });
+            // A helper function to erase the dense index file with some
+            // logging.
+            auto erase_dense_index_file = [=] {
+              self
+                ->request<caf::message_priority::high>(
+                  self->state.filesystem, caf::infinite, atom::erase_v, path)
```
lava, line 1617: Shouldn't this be the `store_path` that's computed below?

lava: Ah, nevermind

libvast/src/system/index.cpp
```diff
@@ -1607,77 +1664,26 @@ index(index_actor::stateful_pointer<index_state> self,
...
-                       synopsis_path](const caf::error& err) {
-                        VAST_WARN("{} failed to erase partition "
-                                  "synopsis {} at {}: {}",
-                                  *self, partition_id, synopsis_path, err);
-                      });
+                  // TODO: We could send `all_ids` as the second
+                  // argument here, which doesn't really make sense
+                  // from an interface perspective but would save the
+                  // partition from recomputing the same bitmap.
+                  auto partition_actor
```
lava, line 1678: You probably still want to explicitly drop this in case we don't get to the fallback path, otherwise there's a deleted partition hanging around in the cache.

lava (approved):
Looks correct to me, although long-term we probably need a more robust system to handle store paths.

libvast/src/system/index.cpp
tobim: I don't think that would be a problem.

libvast/src/system/index.cpp
dominiklohmann: That partition should be rotated out once we attempt to access it, as that would fail—which (I think) we handle gracefully already.

libvast/src/system/index.cpp
dominiklohmann: Tobias and I talked about this and agree that this is not a problem; the partition will be rotated out of the cache.

PR #3003, opened 2023-03-09
Fix post-deploy state of the ansible role

tobim:


lava (approved):
Looks like the role is duplicated when comparing to https://github.com/tenzir/appliances/pull/153 ? Anyways, looks good.

tobim:
> Looks like the role is duplicated when comparing to https://github.com/tenzir/appliances/pull/153 ? Anyways, looks good.

The internal repo has copy to simplify operations. They should be kept in sync bar some experimental changes.

PR #3006, opened 2023-03-10
Trigger new compaction runs immediately on error

dominiklohmann:
Failing in a compaction run should not cause VAST to wait for the configured interval until it starts a new run, as this may lead to the disk running full. This bumps the compaction plugin to include the changes done by @lava to fix this behavior.


tobim:
I changed the base to the related improvement branch so I can deploy it on the internal testbed.

tobim:
@dominiklohmann can I push here to fix the nix submodule rev?

dominiklohmann:
> @dominiklohmann can I push here to fix the nix submodule rev?

Yes, certainly.

tobim (approved):

PR #3007, opened 2023-03-10
Implement remaining operators using new API

jachris:


libvast/builtins/operators/replace.cpp
```diff
@@ -164,7 +164,39 @@ class replace_operator : public pipeline_operator {
...
+    return [transformations = std::move(bound_config->transformations)](
+             generator<table_slice> input) -> generator<table_slice> {
+      for (auto&& slice : input) {
+        co_yield transform_columns(slice, transformations);
+      }
+    };
+  }
+
+  [[nodiscard]] auto to_string() const noexcept -> std::string override {
+    return fmt::format("replace");
```
dominiklohmann, line 190: 👆

dominiklohmann (approved):
Approving modulo a few minor comments, to be merged after its base PR.

libvast/builtins/operators/summarize.cpp
```diff
@@ -729,6 +731,54 @@ class summarize_operator : public pipeline_operator {
...
+
+  [[nodiscard]] auto
+  make_physical_operator(const type& input_schema,
+                         operator_control_plane* ctrl) noexcept
+    -> caf::expected<physical_operator<events, events>> override {
+    if (auto aggregation = aggregation::make(input_schema, config_)) {
+      return [aggregation = std::move(*aggregation), ctrl](
+               generator<table_slice> input) mutable -> generator<table_slice> {
+        for (auto&& slice : input) {
+          aggregation.add(to_record_batch(slice));
```
dominiklohmann, line 752:
```suggestion
          aggregation.add(to_record_batch(slice));
          co_yield {};
```

jachris: What are your thoughts on this alternative?
```cpp
if (slice.rows() == 0) {
  co_yield {};
} else {
  aggregation.add(to_record_batch(slice));
}
```
I'm mainly asking because I'm not sure whether only yielding if the input actually stalled should be allowed. We might want to prohibit this, and/or document the requirements more clearly.

dominiklohmann: They have slightly different behavior in terms of interruptibility; I think we should yield once per batch.

PR #3008, opened 2023-03-10
Deprecate `vast.pipeline-triggers`

dominiklohmann:
These will be superseded by node ingress/egress pipelines soon, so let's be kind to our users and deprecate them now already.

### DIFF
--- a/vast/vast.cpp
+++ b/vast/vast.cpp
@@ -151,6 +151,16 @@ int main(int argc, char** argv) {
                export_transforms.error());
     return EXIT_FAILURE;
   }
+  // Warn if vast.pipeline-triggers are defined, as the functionality will be
+  // superseded by the new pipeline executor in a future release. The model
+  // doesn't make much sense when switching from a client-server to a multi-node
+  // architecture, so we plan to replace them with node ingress/egress pipelines
+  // in the near future.
+  if (caf::get_if<caf::settings>(&cfg, "vast.pipeline-triggers")) {
+    VAST_WARN("the 'vast.pipeline-triggers' option is deprecated and will be "
+              "removed in the next release; use inline import and export "
+              "pipelines instead");
+  }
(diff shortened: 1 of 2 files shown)


vast/vast.cpp
```diff
@@ -151,6 +151,17 @@ int main(int argc, char** argv) {
...
+  // superseded by the new pipeline executor in a future release. The model
+  // doesn't make much sense when switching from a client-server to a multi-node
+  // architecture, so we plan to replace them with node ingress/egress pipelines
+  // in the near future.
+  if (const auto* pipeline_triggers
+      = caf::get_if<caf::settings>(&cfg, "vast.pipeline-triggers")) {
+    VAST_WARN("the 'vast.pipeline-triggers' option is deprecated and will be "
+              "removed in the next release; use inline import and export "
+              "pipelines instead");
+  }
```
lava, line 164: This probably needs approval from product; so far we've always marketed pipelines as a way to enforce policy via server configuration; as far as I can tell inline pipelines aren't a replacement for that since they must be supplied by the user.

dominiklohmann: It does have approval from product, our editions & features doc lists this as node ingress and egress pipelines.

From an engineering perspective, we would need to write a lot of unnecessary code in the migration from the client/server import/export pipelines to node ingress/egress pipelines if we want to switch to the new executor. That is what's motivating this PR.

dominiklohmann: Also cc @jachris and @mavam on this discussion; @jachris is working on that migration, and @mavam would need to sign off on this plan, or explicitly agree to us doing additional work to keep the feature in the meantime.

mavam: Approving here from a product perspective: we'll be without for a rather short time to speed up engineering. That's a sensible trade-off for where we are.

mavam (approved):

PR #3009, opened 2023-03-10
Implement the `json` printer

Dakostu:
This PR adds a new `printer` plugin type with the first `json` implementation that transforms table slices into a stream of [NDJSON](https://github.com/ndjson/ndjson-spec) data.

Currently unreachable from the user POV.

### DIFF
--- a/libvast/builtins/printers/json.cpp
+++ b/libvast/builtins/printers/json.cpp
@@ -6,17 +6,41 @@
 // SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
 // SPDX-License-Identifier: BSD-3-Clause
 
+#include <vast/arrow_table_slice.hpp>
+#include <vast/concept/printable/vast/json.hpp>
 #include <vast/plugin.hpp>
+
+#include <arrow/record_batch.h>
 
 namespace vast::plugins::json_printer {
 
 class plugin : public virtual printer_plugin {
 public:
   [[nodiscard]] auto
-  make_printer(const record&, [[maybe_unused]] type input_schema,
+  make_printer(const record& options, [[maybe_unused]] type input_schema,
                const operator_control_plane&) const
     -> caf::expected<printer_plugin::printer> override {
-    return [](generator<table_slice>) -> generator<chunk_ptr> {
+    auto input_type = caf::get<record_type>(input_schema);
+    return [input_type, plugin_options = options](
+             generator<table_slice> slices) -> generator<chunk_ptr> {
+      // JSON printer should output NDJSON, see:
+      // https://github.com/ndjson/ndjson-spec
+      auto printer = vast::json_printer{{.oneline = true}};
+      for (const auto& slice : slices) {
+        auto result = std::string{};
+        auto resolved_slice = resolve_enumerations(slice);
+        auto array
+          = to_record_batch(resolved_slice)->ToStructArray().ValueOrDie();
+        auto out_iter = std::back_inserter(result);
+        for (const auto& row : values(input_type, *array)) {
+          VAST_ASSERT_CHEAP(row);
+          const auto ok = printer.print(out_iter, *row);
+          VAST_ASSERT_CHEAP(ok);
+          out_iter = fmt::format_to(out_iter, "\n");
+          auto chunk = chunk::make(std::exchange(result, {}));
+          co_yield std::move(chunk);
+        }
+      }
       co_return;
     };
   }
--- a/libvast/builtins/printers/json.cpp
+++ b/libvast/builtins/printers/json.cpp
@@ -1,0 +1,49 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <vast/plugin.hpp>
+
+namespace vast::plugins::json_printer {
+
+class plugin : public virtual printer_plugin {
+public:
+  [[nodiscard]] auto
+  make_printer(const record&, [[maybe_unused]] type input_schema,
+               const operator_control_plane&) const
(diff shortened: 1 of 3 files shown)


dominiklohmann:
@Dakostu I changed the reviewer from the `vast` team to the `security-pipelines-squad` team, that way only people that can actually review this get a notification.

libvast/builtins/printers/json.cpp
```diff
@@ -0,0 +1,73 @@
...
+  [[nodiscard]] auto
+  make_printer(const record& options, [[maybe_unused]] type input_schema,
+               const operator_control_plane&) const
+    -> caf::expected<printer_plugin::printer> override {
+    auto input_type = caf::get<record_type>(input_schema);
+    return [input_type, plugin_options = options](
+             generator<table_slice> slices) -> generator<chunk_ptr> {
+      // JSON printer should output NDJSON, see:
+      // https://github.com/ndjson/ndjson-spec
+      auto printer = vast::json_printer{{.oneline = true}};
```
dominiklohmann, line 28: Can you add a TODO here that since this printer is per-schema we can write an optimized version of it that gets the schema ahead of time and only expects data corresponding to exactly that schema?

libvast/builtins/printers/json.cpp
```diff
@@ -0,0 +1,73 @@
...
+          = to_record_batch(resolved_slice)->ToStructArray().ValueOrDie();
+        auto out_iter = std::back_inserter(result);
+        for (const auto& row : values(input_type, *array)) {
+          VAST_ASSERT_CHEAP(row);
+          const auto ok = printer.print(out_iter, *row);
+          VAST_ASSERT_CHEAP(ok);
+          out_iter = fmt::format_to(out_iter, "\n");
+          auto chunk = chunk::make(std::exchange(result, {}));
+          co_yield std::move(chunk);
+        }
```
dominiklohmann, line 42:
```suggestion
        for (const auto& row : values(input_type, *array)) {
          VAST_ASSERT_CHEAP(row);
          const auto ok = printer.print(out_iter, *row);
          VAST_ASSERT_CHEAP(ok);
          out_iter = fmt::format_to(out_iter, "\n");
        }
        co_yield chunk::make(std::exchange(buffer, {}));
```

Let's not yield once per row—I imagine that to be pretty slow—but rather once per table slice instead.

libvast/builtins/printers/json.cpp
```diff
@@ -0,0 +1,73 @@
...
+               const operator_control_plane&) const
+    -> caf::expected<printer_plugin::printer> override {
+    auto input_type = caf::get<record_type>(input_schema);
+    return [input_type, plugin_options = options](
+             generator<table_slice> slices) -> generator<chunk_ptr> {
+      // JSON printer should output NDJSON, see:
+      // https://github.com/ndjson/ndjson-spec
+      auto printer = vast::json_printer{{.oneline = true}};
+      for (const auto& slice : slices) {
+        auto result = std::string{};
```
dominiklohmann, line 30:
```suggestion
      auto buffer = std::vector<char>{};
      for (const auto& slice : slices) {
```

This has to be a `vector<char>` so you don't have trailing null bytes in the chunk you yield. It can also be outside of the loop as you clear it out with `std::exchange` right before yielding anyways.

dominiklohmann (approved):
Just minor things, but approving already. I think this is pretty nice!

libvast/builtins/printers/json.cpp
Dakostu: Ah, yes, the classic debugging `string`. I replaced it with a loop-global `std::vector<char>` in https://github.com/tenzir/vast/pull/3009/commits/71d99569909b08a70f61bfa4f302e3189bf94a2c

libvast/builtins/printers/json.cpp
Dakostu: Oh, yeah, I was tinkering around with the row formatting but that is in fact not needed anymore.
I moved the yielding outside of the per-row loop in https://github.com/tenzir/vast/pull/3009/commits/71d99569909b08a70f61bfa4f302e3189bf94a2c

libvast/test/printer_plugin.cpp
```diff
@@ -0,0 +1,336 @@
...
+      FAIL("no mock implementation available");
+    }
+
+    [[nodiscard]] virtual auto concepts() const noexcept
+      -> const concepts_map& {
+      FAIL("no mock implementation available");
+    }
+  };
+
+  fixture() {
```
dominiklohmann, line 76: Possible improvement: There's a `fixtures::events` fixture that contains a bunch of table slices that you could use for testing.

Dakostu: Wow, `fixtures::events` is really handy. I added two unit tests using data from `fixtures::events` in https://github.com/tenzir/vast/pull/3009/commits/561a6e810c7481799fe338f5b7b4555e306d51a8

libvast/builtins/printers/json.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+#include <arrow/record_batch.h>
+
+namespace vast::plugins::json_printer {
+
+class plugin : public virtual printer_plugin {
+public:
+  [[nodiscard]] auto
+  make_printer(const record& options, [[maybe_unused]] type input_schema,
+               const operator_control_plane&) const
```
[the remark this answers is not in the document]
jachris: We agreed to use `const operator_control_plane&` everywhere.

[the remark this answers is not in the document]
Dakostu: So I assume the const-ref is a go? I do wonder whether dangling references are a possibility here - and whether with pointers we would have dangling pointers instead.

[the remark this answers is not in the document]
jachris: The executor must ensure that references/pointers to the control plane remain valid throughout the entire pipeline execution. But after taking a second look at this, I think we actually want `operator_control_plane&` (without `const`). At least that is what we will use for `logical_pipeline`.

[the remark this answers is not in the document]
Dakostu: I'll change the parameter to `logical_pipeline&` and enable auto-merge.

PR #3010, opened 2023-03-11
Implement the `vast exec` command

dominiklohmann:
This command takes a single positional argument that must parse as a valid logical pipeline, and then immediately executes it.

The current implementation is pretty naïve, but it suffices for initial testing once we have the from and to operators in place to actually be able to write a closed pipeline.

I've written a follow-up issue to implement signal handling in the executor; I've intentionally not implemented that for now as the API for the exeuctor will change to an asynchronous rather than just concurrent API soon after this initial version, which makes it easier to implement signal handling correctly.

Fixes tenzir/issues#202

### DIFF
--- a/libvast/builtins/commands/exec.cpp
+++ b/libvast/builtins/commands/exec.cpp
@@ -1,0 +1,74 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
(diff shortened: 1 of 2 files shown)


libvast/builtins/commands/exec.cpp
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  [[nodiscard]] std::string name() const override {
+    return "exec";
+  }
+
+  [[nodiscard]] std::pair<std::unique_ptr<command>, command::factory>
+  make_command() const override {
+    auto exec = std::make_unique<command>("exec", "execute a pipeline locally",
+                                          command::opts("?vast.exec"));
```
dominiklohmann, line 61: Reminder to self: write a changelog entry.

PR #3013, opened 2023-03-13
Prototype the `pipeline_executor_actor` interface

dominiklohmann:
Putting a logical pipeline into an actor interface requires making logical pipelines (and as such logical operators) copyable. This tests that to the point of simply making it possible to compile VAST with such an actor interface with tests yet to be written.

### DIFF
--- a/libvast/src/logical_operator.cpp
+++ b/libvast/src/logical_operator.cpp
@@ -20,11 +20,15 @@ auto runtime_logical_operator::done() const noexcept -> bool {
   return false;
 }
 
-auto runtime_logical_operator::copy() const noexcept -> logical_operator_ptr {
+auto runtime_logical_operator::copy() const noexcept
+  -> caf::expected<logical_operator_ptr> {
   const auto repr = to_string();
   auto result = logical_pipeline::parse(repr);
-  if (not result)
-    VAST_WARN("failed to copy logical pipeline '{}': {}", repr, result.error());
+  if (not result) {
+    return caf::make_error(
+      ec::logic_error, fmt::format("failed to copy logical pipeline '{}': {}",
+                                   repr, result.error()));
+  }
   return std::make_unique<logical_pipeline>(std::move(*result));
 }
 
--- a/libvast/src/logical_operator.cpp
+++ b/libvast/src/logical_operator.cpp
@@ -1,0 +1,31 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "vast/logical_operator.hpp"
+
+#include "vast/logical_pipeline.hpp"
+
+namespace vast {
+
+auto runtime_logical_operator::detached() const noexcept -> bool {
+  return false;
+}
+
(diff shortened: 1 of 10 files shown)


libvast/src/logical_pipeline.cpp
```diff
@@ -281,6 +281,34 @@ auto make_local_producer(std::span<logical_operator_ptr> ops,
...
+
+logical_pipeline& logical_pipeline::operator=(const logical_pipeline& rhs) {
+  if (this != &rhs) {
+    ops_.clear();
+    ops_.reserve(rhs.ops_.size());
+    for (const auto& op : ops_)
+      ops_.push_back(op ? op->copy() : nullptr);
+  }
+  return *this;
+}
```
jachris, line 301: This violates our invariant that a `logical_pipeline` does not contain another `logical_pipeline`. Furthermore, we currently prohibit `nullptr` operators in `make`, and I would personally prefer to maintain this invariant everywhere. We can perhaps throw an exception if `copy()` returns a `nullptr`.

libvast/src/logical_operator.cpp
```diff
@@ -0,0 +1,31 @@
...
+auto runtime_logical_operator::done() const noexcept -> bool {
+  return false;
+}
+
+auto runtime_logical_operator::copy() const noexcept -> logical_operator_ptr {
+  const auto repr = to_string();
+  auto result = logical_pipeline::parse(repr);
+  if (not result)
+    VAST_WARN("failed to copy logical pipeline '{}': {}", repr, result.error());
+  return std::make_unique<logical_pipeline>(std::move(*result));
```
jachris, line 28: This probably needs a `return nullptr` after `VAST_WARN`.

libvast/src/logical_operator.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include "vast/logical_operator.hpp"
+
+#include "vast/logical_pipeline.hpp"
+
+namespace vast {
+
+auto runtime_logical_operator::detached() const noexcept -> bool {
+  return false;
+}
```
jachris, line 17: This implementation does not match the documentation: `Returns whether any of the logical pipeline's operators are detached.`

libvast/src/logical_pipeline.cpp
```diff
@@ -367,6 +395,13 @@ auto logical_pipeline::output_element_type() const noexcept
   return ops_.back()->output_element_type();
 }
 
+auto logical_pipeline::detached() const noexcept
+  -> bool {
+  return std::any_of(ops_.begin(), ops_.end(), [](const auto& op) noexcept {
+    return op ? op->detached() : false;
+  });
```
jachris, line 402: This can also be changed if we know that `op != nullptr`.

jachris (approved):

libvast/src/logical_pipeline.cpp
dominiklohmann: Good catch, I'll simplify the logic and flatten here.

libvast/src/logical_operator.cpp
jachris: Oh, and I even commented on it afterwards... Nevermind.

libvast/src/logical_operator.cpp
dominiklohmann: I made this return an expected so we can choose to print the error at the call site instead of warning.

dominiklohmann:
We discussed a different approach to doing `inspect` overloads internally. I will open a follow-up PR that takes care of this sometime this week, ideally.

PR #3014, opened 2023-03-14
Implement the `stdout` dumper

Dakostu:
This PR adds a new `dumper` plugin type with the first `stdout` implementation that sends out chunks to `stdout`.

Currently unreachable from the user POV.

### DIFF
--- a/libvast/include/vast/plugin.hpp
+++ b/libvast/include/vast/plugin.hpp
@@ -438,7 +438,6 @@ public:
   // Alias for the byte chunk generation function.
   using printer
     = std::function<auto(generator<table_slice>)->generator<chunk_ptr>>;
-
   // Alias for the byte chunk -> table_slice transformation function.
   using dumper
     = std::function<auto(generator<chunk_ptr>)->generator<std::monostate>>;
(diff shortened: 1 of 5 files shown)


libvast/include/vast/plugin.hpp
```diff
@@ -460,6 +460,37 @@ class printer_plugin : public virtual plugin {
...
+  using printer
+    = std::function<auto(generator<table_slice>)->generator<chunk_ptr>>;
+
+  // Alias for the byte chunk dumping function.
+  using dumper
+    = std::function<auto(generator<chunk_ptr>)->generator<std::monostate>>;
+
+  /// Returns the dumper.
+  [[nodiscard]] virtual auto
+  make_dumper(const record&, type input_schema, operator_control_plane&) const
```
patszt, line 479: Could we change it to const operator_control_plane& for now? I think it might be good for const correctness.
Later on we might forget that it is non const while we only use const methods anyway.

libvast/test/dumper_plugin.cpp
```diff
@@ -0,0 +1,139 @@
...
+  fixture() {
+    // TODO: Move this into a separate fixture when we are starting to test more
+    // than one dumper type.
+    dumper_plugin = vast::plugins::find<vast::dumper_plugin>("stdout");
+    REQUIRE(dumper_plugin);
+    current_dumper = unbox(dumper_plugin->make_dumper({}, {}, control_plane));
+  }
+
+  ~fixture() {
+  }
```
patszt, line 65: I think this can be removed

libvast/include/vast/plugin.hpp
Dakostu: We had a similar discussion [here](https://github.com/tenzir/vast/pull/3009#pullrequestreview-1334983726).

patszt (approved):

libvast/test/dumper_plugin.cpp
Dakostu: True, actually. Removed the destructors in https://github.com/tenzir/vast/pull/3014/commits/42d15545115f201df6df061e86d30553fc2588b1

PR #3016, opened 2023-03-14
Implement `from` and `read` operators

jachris:


libvast/builtins/operators/from_read.cpp
```diff
@@ -0,0 +1,257 @@
...
+#include <vast/plugin.hpp>
+
+#include <arrow/type.h>
+#include <caf/error.hpp>
+
+namespace vast::plugins::from_read {
+
+namespace {
+
+class from_operator final : public crtp_operator<from_operator> {
```
patszt, line 24: Since this already uses polymorphism and one can be constructed with either loader or parser i would see it as two classes instead. One responsible from handling the loader and the other for handling the parser.

libvast/builtins/operators/from_read.cpp
```diff
@@ -0,0 +1,257 @@
...
+        return loader.error();
+      }
+      // TODO: Reconsider this.
+      return std::invoke(
+        [](loader_plugin::loader loader) -> generator<chunk_ptr> {
+          for (auto&& x : loader()) {
+            co_yield x;
+          }
+        },
+        std::move(*loader));
```
Dakostu, line 48:
```suggestion
      return loader_();
```

Does turning the `loader` into a field `loader_` and using it like in the suggestion work?

libvast/builtins/operators/from_read.cpp
```diff
@@ -0,0 +1,257 @@
...
+  };
+
+  auto make_operator(std::string_view pipeline) const
+    -> std::pair<std::string_view, caf::expected<operator_ptr>> override {
+    using parsers::optional_ws_or_comment, parsers::end_of_pipeline_operator,
+      parsers::identifier, parsers::required_ws_or_comment;
+    const auto* f = pipeline.begin();
+    const auto* const l = pipeline.end();
+    const auto p = optional_ws_or_comment >> identifier
+                   >> -(required_ws_or_comment >> string_parser{"read"}
```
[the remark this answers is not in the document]
jachris: The operator name is already parsed at this point (this is because we have to dispatch to the `operator_plugin` with the same name). Hence we check whether we have (`from`) `X | ...` or (`from`) `X read Y`.

libvast/builtins/operators/from_read.cpp
Dakostu: I did something similar in my `write`/`to` branch and it seems to compile at least :thinking:

libvast/builtins/operators/from_read.cpp
jachris: Ideally `from_operator` would not care about the parser, and `read_operator` would not care about the loader. However, the `make_default_parser` and `make_default_loader` API expects an `operator_control_plane&`, which we only get when instantiating the already-built pipeline. Perhaps we can change the API to improve the situation.

jachris: API was changed to improve this situation.

libvast/builtins/operators/from_read.cpp
```diff
@@ -0,0 +1,185 @@
...
+  explicit read_operator(const plugin& parser, record config)
+    : parser_plugin_{parser}, config_{std::move(config)} {
+  }
+
+  auto operator()(generator<chunk_ptr>, operator_control_plane&) const
+    -> caf::expected<generator<table_slice>> {
+    // TODO: Implement this.
+    return caf::make_error(ec::unimplemented,
+                           "parser_plugin does not exist yet");
+  }
```
dominiklohmann, line 56: Please write a follow-up issue for doing this once we have the JSON parser so we can track this somewhere.

dominiklohmann (approved):

PR #3018, opened 2023-03-16
Avoid crashing when reading a pre-2.0 partition

dominiklohmann:
This fixes a crash when starting VAST v3.0 with a partition from before v2.0, which was not correctly ignored.


tobim (approved):
Tested with a vast database built with v1.1. Works as expected.

PR #3020, opened 2023-03-16
Include outdated but not undersized partitions in automatic rebuilds

dominiklohmann:
This fixes a bug where automatic rebuilds picked up undersized and outdated partitions, but not outdated partitions that were appropriately sized.


libvast/builtins/commands/rebuild.cpp
```diff
@@ -277,28 +277,19 @@ struct rebuilder_state {
...
-                });
-            }
+            std::erase_if(
+              result.partition_infos, [&](const partition_info& partition) {
+                const auto erase_if_undersized
+                  = run->options.undersized
+                    && static_cast<bool>(partition.schema)
+                    && partition.events > detail::narrow_cast<size_t>(
+                         detail::narrow_cast<double>(max_partition_size)
+                         * undersized_threshold);
```
tobim, line 287: I believe we need to do the inverse: similar to "erase_if_**not**_outdated", we want to remove items from the list if they are **not** undersized.

Additionally, the predicate `static_cast<bool>(partition.schema)` is not clear to me, what does it achieve?

tobim (changes requested):
This solution seems to have an oversight in the logic. Can you check it again?

libvast/builtins/commands/rebuild.cpp
tobim: Ah, the check is actually correct, but the variable name should still reflect that.

dominiklohmann: > `static_cast<bool>(partition.schema)`

I think we can actually remove that now, as VAST no longer supports reading partitions that have a heterogeneous schema. This was the check that indicated whether a partition was homogeneous.

tobim (approved):
I tested that this works as expected now with a vast database created with v2.2.

PR #3021, opened 2023-03-16
Align default branch refs to main

tobim:
This change reflects the switch to `main` as the default branch name within the code base.
References in the documentation, scripts and CI workflows are updated to the new name.


jachris (approved):
Looks complete to me.

PR #3024, opened 2023-03-17
Implement `to` and `write` operators

Dakostu:
This PR adds plugins for formatting and outputting events in our new pipeline ecosystem, with the `json` printer and `stdout` dumper as first implementations.

Also, some changes to the ~~`element_type` and~~ `dumper`/`writer` plugin APIs.


libvast/builtins/operators/to.cpp
```diff
@@ -0,0 +1,119 @@
...
+                                          "allow joining but dumper '{}' "
+                                          "requires joining",
+                                          printer->name(), dumper->name()))};
+    }
+    auto print_op = std::make_unique<print_operator>(std::move(*printer));
+    auto dump_op = std::make_unique<dump_operator>(std::move(*dumper));
+    auto ops = std::vector<logical_operator_ptr>{};
+    ops.emplace_back(std::move(print_op));
+    ops.emplace_back(std::move(dump_op));
+    auto sub_pipeline = logical_pipeline::make(std::move(ops));
```
dominiklohmann, line 107: IIUC this does not correctly handle the two cases where the dumper does not require its inputs to be joined. We don't currently have such a dumper, but at least for now we should probably error in these cases or implement it correctly.

libvast/builtins/operators/to.cpp
```diff
@@ -0,0 +1,119 @@
...
+      if (!printer) {
+        return {std::string_view{f, l},
+                caf::make_error(ec::syntax_error,
+                                fmt::format("failed to parse "
+                                            "to operator: no '{}' printer "
+                                            "found",
+                                            printer_name))};
+      }
+    } else {
+      printer = dumper->make_default_printer();
```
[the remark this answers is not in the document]
Dakostu: If either no `printer` or `dumper` is available, then the `to` and `write` operators can not fulfill their purpose of printing+dumping. A `printer` without `dumper` has bytes, but nowhere to dump them; a `dumper` without `printer` would not get any bytes to dump. That would then be an error, in my eyes.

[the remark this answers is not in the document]
dominiklohmann: Currently you either run into an assertion inside `make_default_printer()` (which you cannot assume to be there for all printers) or you just dereference a nullptr blindly below this when you check whether the printer allows for merging its output.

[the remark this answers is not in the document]
Dakostu: `nullptr` == there is no default dumper/printer available.
With that being a possibility, I can swap out the assertion from the above discussion for a nullptr check that returns an error.

libvast/builtins/operators/to.cpp
dominiklohmann: I would say yes. We probably need to instantiate the dump operator once per schema in this case.

libvast/builtins/operators/to.cpp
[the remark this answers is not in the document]
Dakostu: I replaced the assertion with a nullptr check in https://github.com/tenzir/vast/pull/3024/commits/2e1e92395262be05e46bc5be1b03765bda2e225b

libvast/builtins/operators/to.cpp
Dakostu: I implemented a `print_dump_operator` in https://github.com/tenzir/vast/pull/3024/commits/a8cbd8ca23a07c77105893e00786e7db40e7a05f

libvast/include/vast/dump_operator.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include "vast/logical_operator.hpp"
+
+#include <vast/plugin.hpp>
+
+namespace vast {
+
+/// The logical operator for dumping data that will have to be joined later
+/// during pipeline execution.
+class dump_operator : public logical_operator<chunk_ptr, void> {
```
dominiklohmann, line 17: @jachris found a nice approach in #3016 by creating a single file `libvast/builtins/from_read.cpp` that calls `VAST_REGISTER_PLUGIN` twice, avoiding the need for the logic shared between the from and read operators to leak into libvast itself. I would recommend you follow the same approach here for the write and to operators.

libvast/include/vast/dump_operator.hpp
```diff
@@ -0,0 +1,46 @@
...
+public:
+  explicit dump_operator(const dumper_plugin& dumper) noexcept
+    : dumper_plugin_{dumper} {
+  }
+
+  [[nodiscard]] auto
+  make_physical_operator(const type& input_schema,
```
dominiklohmann, line 24: The signature of this function changed with #3039; I would like to take a second look at this PR after you've rebased it on current main, as I think most of the diff will magically disappear then. The logic for the operator does look good to me, though.

dominiklohmann (commented):
This is looking great, but I would love to take another look after a rebase so I can also view it with the now revised API.

libvast/include/vast/dump_operator.hpp
Dakostu: I moved everything into `libvast/builtins/operators/write_to.cpp` in https://github.com/tenzir/vast/pull/3024/commits/da051bbe03ddfcb5be49854d9fabdff6ffa857c2

libvast/include/vast/dump_operator.hpp
Dakostu: Rebased & realigned logic in https://github.com/tenzir/vast/pull/3024/commits/5459167226ee2f0dcf1c633db42818a96f43988d

libvast/builtins/operators/write_to.cpp
```diff
@@ -0,0 +1,327 @@
...
+/// The operator for printing data that will have to be joined later
+/// during pipeline execution.
+class print_operator final : public crtp_operator<print_operator> {
+public:
+  explicit print_operator(const printer_plugin& printer) noexcept
+    : printer_plugin_{&printer} {
+  }
+
+  auto operator()(generator<table_slice> input, const type& input_schema,
+                  operator_control_plane& ctrl) const
```
jachris, line 38: The new API does not take a schema anymore, you can remove this parameter.

Btw, if `make_printer` should stay schema specific, you can probably use `schematic_operator<print_operator, printer_plugin::printer, chunk_ptr>` here.

Dakostu: We realigned the API.

libvast/builtins/operators/write_to.cpp
```diff
@@ -0,0 +1,350 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2021 The VAST Contributors
```
dominiklohmann, line 6:
```suggestion
// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
```

dominiklohmann (approved):
I think this can be merged. As we discussed in Slack, there are a few things we'll have to change regarding serialization and deserialization, but since we don't have that

PR #3026, opened 2023-03-20
Fix usage of moved from variable in rebuild metrics

patszt:
The count of currently rebuilt partitions was calculated based on size of a moved from vector. 
This resulted in wrong rebuild metrics


dominiklohmann (approved):
I verified locally that the metrics now display as expected, and as a follow-up pushed another fix that makes rebuilding also stop correctly when assertions are enabled.

Thanks for the quick reaction to the bug report, @patszt! 🚀

PR #3030, opened 2023-03-21
Migrate Debian based builds to bookworm

tobim:
Since bookworm is now in hard freeze this upgrade should be safe. This will allow us to bump the minimum version of GCC to 12 should the need arise.

### DIFF
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -1090,6 +1090,7 @@ jobs:
         with:
           submodules: recursive
       - name: Setup Python
+        if: ${{ matrix.setup.os == 'macos-latest' }}
         uses: actions/setup-python@v4
         with:
           python-version: "3.9"
@@ -1109,7 +1110,6 @@ jobs:
             apt-transport-https \
             curl
           ./scripts/debian/install-dev-dependencies.sh
-          python3 -m pip install --upgrade pip
       - name: Determine VAST Package Name
         id: configure
         run: |
--- a/.github/workflows/vast.yaml
+++ b/.github/workflows/vast.yaml
@@ -981,6 +981,7 @@ jobs:
     runs-on: ${{ matrix.setup.os }}
     container: ${{ matrix.setup.container }}
     strategy:
+      fail-fast: false
       matrix:
         setup:
           - os: ubuntu-20.04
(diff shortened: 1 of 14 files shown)


tobim:
CI should run now after the Apache JFrog issues are resolved.
The only remaining question is whether we should bump the minimum requirement for GCC already, since 10 will not be tested anymore once this PR is merged.

.github/workflows/vast.yaml
```diff
@@ -775,11 +775,11 @@ jobs:
       matrix:
         vast:
           - os: ubuntu-20.04
-            container: python:3.10-slim-bullseye
+            container: debian:bookworm-slim
             name: Debian
             compiler: GCC
-            cc: gcc-10
-            cxx: g++-10
+            cc: gcc-12
```
dominiklohmann, line 781: I think we should update the table in the docs in build.md already to gcc12+ and clang17+, as those are now the only versions we actively test with.

dominiklohmann (approved):
<3

Dockerfile
```diff
@@ -95,18 +95,18 @@ RUN apt-get update && \
...
       ca-certificates \
       gnupg2 \
-      libasan5 \
+      libasan6 \
       libc++1 \
       libc++abi1 \
-      libflatbuffers1 \
-      libfmt7 \
+      libflatbuffers2 \
+      libfmt9 \
```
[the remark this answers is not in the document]
tobim: It is still at 8 in nixpkgs because nobody bothered to bump it for spdlog yet.

.github/workflows/vast.yaml
patszt: Nice, i am actually struggling with my code not working on gcc 10 due to a bug. I am trying to update to gcc 11 myself. Can't wait for it to get merged :D

cmake/VASTRegisterPlugin.cmake
```diff
@@ -160,7 +160,15 @@ function (VASTCompileFlatBuffers)
   add_library(${FBS_TARGET} INTERFACE)
 
   # Link our target against FlatBuffers.
-  find_package(Flatbuffers REQUIRED CONFIG)
+  find_package(Flatbuffers CONFIG)
+  # Ugly workaround for Flatbuffers 2.0.8, which briefly changed the
+  # project name to "FlatBuffers".
```
dominiklohmann, line 165: The same workaround needs to be appended to `VAST_FIND_DEPENDENCY_LIST` right below.

PR #3031, opened 2023-03-24
Improve tuning docs

mavam:
This PR improves the tuning documentation.


web/docs/setup/tune.md
```diff
@@ -230,45 +173,75 @@ vast:
...
+```yaml
+vast:
+  zstd-compression-level: 1
+```
 
-VAST processes log messages in a dedicated thread, which by default buffers up
-to 1M messages for servers, and 100 for clients. The option
-`vast.log-queue-size` controls this setting.
+Currently, the default value is taken from Apache Arrow itself, using
+`arrow::util::Codec::DefaultCompressionLevel(arrow::Compression::ZSTD)`
```
dominiklohmann, line 235: I think this is too much of an implementation detail; if you want to keep it, I would rather link to the documentation of the function than just giving its signature, which is rather meaningless.


Link to docs: https://arrow.apache.org/docs/cpp/api/utilities.html#_CPPv4N5arrow4util5Codec26UseDefaultCompressionLevelEv

web/docs/setup/multi-tier-indexing.excalidraw.svg
dominiklohmann, line 1: We previously had a vote on whether we want to use "sparse and dense indexes" or "catalog and partition indexes," and the latter won. I think it's easier to just always refer to the sketch data structures as catalog indexes here.

This also applies to the text below this image.

web/docs/setup/tune.md
```diff
@@ -54,129 +52,69 @@ The `vast.import.read-timeout` option determines how long a call to read data
...
-1. **Feather**: writes Apache Feather V2 files
-2. **Parquet**: writes [Apache Parquet](https://parquet.apache.org/) files
-3. **Segment**: writes Apache Arrow IPC with a thin wrapper (deprecated,
-   read-only)
+## Storage Engine
 
-VAST defaults to the `feather` store. Enable the Parquet store by loading the
-plugin and adjusting `vast.store-backend`:
+The central component of VAST's storage engine is the *catalog*. It owns the
+partitions, keeps meta data about them, and maintains a set of sparse secondary
```
dominiklohmann, line 58:
```suggestion
partitions, keeps metadata about them, and maintains a set of sparse secondary
```

web/docs/setup/tune.md
```diff
@@ -54,129 +52,69 @@ The `vast.import.read-timeout` option determines how long a call to read data
...
+A false positive during a sparse index lookup can have a noticeable impact on the
+query latency by materializing irrelevant partitions. Based on the cost of I/O,
+this penalty may be substantial. Conversely, reducing the false positive rate
+increases the memory consumption, leading to a higher resident set size and
+larger RAM requirements.
 
 You can control this space-time trade-off in the configuration section
-`vast.index` by specifying index *rules*. Each rule corresponds to one sketch
-and consists of the following components:
+`vast.index` by specifying index *rules*. Each rule configures one sketch and
```
dominiklohmann, line 116: Each entry in that list configures potentially multiple catalog or partition indexes, as a rule can refer to multiple targets. Essentially, it allows for:
- Enabling field-specific catalog indexes per extractor (off by default)
- Changing the false-positive rate of catalog indexes per extractor or type for catalog indexes that support a false-positive rate
- Disabling partition indexes for a given extractor

I think there are quite a few holes in that design, but let's document what we got as that is not the current focus.

web/docs/setup/tune.md
```diff
@@ -230,45 +173,75 @@ vast:
...
+  store-backend: parquet
+```
+
+There's an inherent space-time tradeoff between the stores that affects CPU,
+memory, and storage characteristics. Compared to the Feather store, Parquet
+differs as follows:
+
+1. Parquet files occupy ~40% less space, which also reduces I/O pressure during
+   querying.
+2. Parquet utilizes more CPU cycles during ingest (~10%) and querying.
```
dominiklohmann, line 201: Note that the Feather store behaves much better with queries that end early, because it allows for queries to be stopped early within a store, while our Parquet store only allows for stopping a query early after every store that has been evaluated. This is the primary reason why Parquet is not the default. Now while that is entirely fixable for Parquet, we simply haven't had the time to do so. We should caution about this, because for many parallel queries this makes a really large performance difference and improves latencies and responsiveness throughout the system.

web/docs/setup/tune.md
```diff
@@ -333,3 +306,41 @@ vast rebuild start '#type == "suricata.flow" && #import_time < 2 weeks ago'
...
+The `stop` command gracefully brings down a VAST server that has been started
+with the `start` command.
+
+It is also possible to send a signal `SIGINT(2)` to the `vast` process instead
+of using `vast stop`, but in only works on the same machine that runs the
+server process. We recommend using `vast stop`, as it also works over the wire.
+
+The `stop` command blocks until the server process has terminated, and returns
+a zero exit code upon success, making it suitable for use in launch system
+scripts.
```
dominiklohmann, line 321: This is not true: The `stop` command blocks until the node actor shut down in the node, which is by definition before the process shut down. Using it in scripting via `vast stop && vast start` is not possible because of that, as there would still be a VAST process running.

(One way to fix this from our side would be to remove the PID lock as part of the node actor shut down, but that's a fix for another day.)

web/docs/setup/tune.md
```diff
@@ -333,3 +306,41 @@ vast rebuild start '#type == "suricata.flow" && #import_time < 2 weeks ago'
...
+
+:::note
+The server waits for ongoing import processes to terminate before shutting down
+itself. In case an import process is hanging, you can always terminate the
+hanging process manually to shutdown the server.
+:::
```
dominiklohmann, line 327: This is more of a "known bug" that we simply haven't been able to fix, but hopefully will go away implicitly with the pipeline-first efforts. I think we should rephrase it or remove it.

web/docs/setup/tune.md
```diff
@@ -54,129 +52,69 @@ The `vast.import.read-timeout` option determines how long a call to read data
...
+upper bound on the number of records in a partition, across all table slices.
+The parameter `vast.active-partition-timeout` provides a time-based upper bound:
+once reached, VAST considers the partition as complete, regardless of the number
+of records.
 
-VAST compresses partitions using Zstd for partitions at rest. To fine-tune
-the space-time trade-off, VAST offers a setting, `vast.zstd-compression-level`
-to allow fine-tuning the compression level:
+The two parameters are decoupled to allow for independent control of throughput
+and freshness.
```
dominiklohmann, line 85: This should also mention that VAST automatically merges undersized partitions over time, which makes the amortized performance penalty for choosing a lower active partition timeout zero.

web/docs/setup/tune.md
```diff
@@ -54,129 +52,69 @@ The `vast.import.read-timeout` option determines how long a call to read data
...
 
-VAST supports [rebuilding the entire database](#rebuild-partitions) in case you
-want to switch to a different store format. However, VAST works perfectly fine
-with a mixed-storage configuration, so a full rebuild is not required.
+VAST groups table slices with the same schema in a partition. There exist
+mutable *active partitions* that VAST writes to during ingestion, and immutable
+*passive partitions* that VAST reads from during query execution.
 
-### Compression
+When constructing a partition, the parameter `vast.max-partition-size` sets an
```
dominiklohmann, line 78:
```suggestion
When constructing a partition, the parameter `vast.max-partition-size` (defaults to 4 Mi) sets an
```

web/docs/setup/tune.md
```diff
@@ -54,129 +52,69 @@ The `vast.import.read-timeout` option determines how long a call to read data
...
+upper bound on the number of records in a partition, across all table slices.
+The parameter `vast.active-partition-timeout` provides a time-based upper bound:
```
dominiklohmann, line 80:
```suggestion
The parameter `vast.active-partition-timeout` (defaults to 10 s) provides a time-based upper bound:
```

web/docs/setup/tune.md
```diff
@@ -54,129 +52,69 @@ The `vast.import.read-timeout` option determines how long a call to read data
...
-partition, the parameter `vast.max-partition-size` sets an upper bound on the
-number of records in a partition, across all table slices. The parameter
-`vast.active-partition-timeout` provides a time-based upper bound: once reached,
-VAST considers the partition as complete, regardless of the number of records.
-
-A LRU cache of partitions accelerates queries to recently used partitions. The
-parameter `vast.max-resident-partitions` controls the number of partitions in
-the LRU cache.
+VAST maintains a LRU cache of partitions to accelerate queries involving recent
+partitions. The parameter `vast.max-resident-partitions` controls the number of
```
dominiklohmann, line 90:
```suggestion
partitions. The parameter `vast.max-resident-partitions` (defalts to 10) controls the number of
```

web/docs/setup/tune.md
```diff
@@ -54,129 +52,69 @@ The `vast.import.read-timeout` option determines how long a call to read data
...
-maintained in active memory to provide high responsiveness.
+### Tune number of sketches and their sizes
 
-As a consequence, the overall amount of data in a VAST instance and the
-`max-partition-size` determine the memory requirements of the catalog. The
-option `max-partition-size` is inversely linked to the number of sketches in the
+The catalog's memory usage is a function of the stored data. The configuration
+option `max-partition-size` determines an upper bound of the number of records
+per partition, which is inversely linked to the number of sketches in the
 catalog. That means increasing the `max-partition-size` is an effective method
```
dominiklohmann, line 104:
```suggestion
option `vast.max-partition-size` determines an upper bound of the number of records
per partition, which is inversely linked to the number of sketches in the
catalog. That means increasing the `vast.max-partition-size` is an effective method
```

web/docs/setup/tune.md
```diff
@@ -7,15 +7,13 @@ sidebar_position: 5
...
+## Batching
 
 VAST processes events in batches. Because the structured data has the shape of a
 table, we call these batches *table slices*. The following options control their
 shape and behavior.
 
-:::note Implementation Note
-Table slices are implemented as *Record Batches* in Apache Arrow.
-:::
+Internally, VAST implements a table slice as an Arrow *record batch*.
```
dominiklohmann, line 16: I think this is a meaningless implementation detail for tuning. Let's just remove this sentence.

dominiklohmann (commented):
I did a first read through this and left some comments; this is quite the improvement over the old docs. I'd like to give it another read after the feedback has been incorporated.

web/docs/setup/multi-tier-indexing.excalidraw.svg
mavam: I adjusted the text. Now the index types don't have the name "sparse" and "dense"; these adjectives are rather used in the text to characterize them.

web/docs/setup/tune.md
mavam: I rewrote the indexing section to reflect this. Please resolve if this works.

web/docs/setup/tune.md
mavam: Added a note.

web/docs/setup/tune.md
mavam: Can you suggest rewording? I only copied this text to a different location and your comment doesn't is too low-level for me to derive a meaningful user-facing action.

web/docs/setup/tune.md
```diff
@@ -230,45 +181,79 @@ vast:
...
+
+1. Storage is scarce, and you want to increase data retention
+2. Workloads are I/O-bound and you have available CPU
+3. Reading data with with off-the-shelf data science tools is a use case
+
+For an in-depth comparison, take a look at our [blog post on Feather and
+Parquet](/blog/parquet-and-feather-writing-security-telemetry).
+
+One final word of caution: a query over a Parquet store is currently not
+interruptible and executes over the entire store. This is an deficiency in our
```
dominiklohmann, line 227:
```suggestion
interruptible and executes over the entire store. This is a deficiency in our
```

web/docs/setup/multi-tier-indexing.excalidraw.svg
dominiklohmann, line 1: The image itself still says sparse and dense index on the left rather than catalog and partition index.

dominiklohmann (approved):
This is great, thanks for incorporating the feedback so quickly!

web/docs/setup/tune.md
dominiklohmann:
```suggestion
The `stop` command asks the server process to terminate gracefully, and returns
a zero exit code upon success, making it suitable for use in launch system scripts.
```

web/docs/setup/multi-tier-indexing.excalidraw.svg
mavam: That's on purpose. The catalog and partition are displayed in the center of the diagram. The qualifiers on the left just distinguish the type of index, it's not a name.

mavam: In this case I'd leave it as is, sorry. 🙂

PR #3036, opened 2023-03-25
Introduce the `put` operator

dominiklohmann:
The `put` operator is a replacement for both the `extend` and the `replace` operators. It takes as arguments a comma-separated list of assignments. If the assigned-to extractor resolves for the input data replaces all occurences of the field (previously `replace`), and if the extractor does not resolve adds one new field at the end (previously `extend`).

`put` is the first operator not to be available in the old pipeline operator framing, and as such is not currently reachable by the user.


libvast/builtins/operators/put.cpp
```diff
@@ -0,0 +1,228 @@
...
+        if (not inferred_type)
+          return caf::make_error(ec::logic_error,
+                                 fmt::format("failed to infer type from '{}'",
+                                             value));
+        extensions.emplace_back(extractor, value, std::move(inferred_type));
+      }
+    }
+    if (not extensions.empty())
+      result.transformations.push_back(
+        {{schema_rt.num_fields() - 1}, make_extend(std::move(extensions))});
```
dominiklohmann, line 63: This crashes if we both extend and also modify the last field via put.

jachris, line 63: Can we have `schema_rt.num_fields() == 0`?

changelog/next/features/3036--put-operator.md
```diff
@@ -0,0 +1,3 @@
+The `put` operator succeeeds the `extend` and `replace` operators. It updates
+values in-place when the assigned-to extractor exists in the input events, and
+ a new field with the assigned value ad the end otherwise.
```
jachris, line 3:
```suggestion
The `put` operator succeeeds the `extend` and `replace` operators. It updates
values in-place when the assigned-to extractor exists in the input events, and
adds a new field with the given value at the end otherwise.
```

jachris (approved):
`libvast/test/logical_pipeline.cpp` still mentions the removed operators.

libvast/builtins/operators/put.cpp
dominiklohmann: Record types by definition must have at least one field currently.

libvast/builtins/operators/put.cpp
```diff
@@ -0,0 +1,228 @@
...
+    std::sort(result.transformations.begin(), result.transformations.end());
```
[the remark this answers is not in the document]
dominiklohmann: This doesn't actually make a difference: the transformations are all the same, and we actually dedup right below the highlighted line. `transform_columns` has a precondition that states that every column must only be modified exactly once.

libvast/builtins/operators/put.cpp
dominiklohmann: I fixed this by maintaining two separate lists of transformations.

libvast/builtins/operators/put.cpp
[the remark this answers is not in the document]
dominiklohmann: Actually there's an edge case I didn't see that we just talked about in Slack, namely `put x=123 x=456`. Leaving this open until I know the desired behavior.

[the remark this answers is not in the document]
dominiklohmann: Should be fixed in https://github.com/tenzir/vast/pull/3036/commits/f23f29bbced3541b5f7373719d9d2aa94e3372f6

libvast/builtins/operators/put.cpp
```diff
@@ -0,0 +1,254 @@
...
+        auto replacement
+          = std::find_if(result.replacements.begin(), result.replacements.end(),
+                         [&](const auto& replacement) {
+                           return replacement.index == index;
+                         });
+        if (replacement == result.replacements.end()) {
+          result.replacements.push_back({index, make_replace(value)});
+        } else {
+          ctrl.warn(caf::make_error(
+            ec::invalid_argument, fmt::format("put operator assignment {}={}' "
```
jachris, line 61:
```suggestion
            ec::invalid_argument, fmt::format("put operator assignment '{}={}' "
```

PR #3039, opened 2023-03-28
Revise pipeline operator API 

jachris:
This PR makes some changes to the implementation of RFC-001 (#3004). The main goal is to simplify operator definitions and internal execution details.

- The instantiation state is now only stored by the instantiated generator.
- There is no more `physical_operator`/`logical_operator` separation in the code base, because we have only one intermediate representation for both (though we can still use these terms in explanations).
- Operators can be copied without a `parse(to_string())` round-trip.
- Schema-specific dispatching is an implementation detail of the operator, but we provide a class that makes this convenient. This drastically simplifies the execution engine.
- Polymorphic operators are now possible. For instance, `head` could be made to work on both event and byte streams.
- The empty pipeline is therefore now also correctly typed, same as `pass`.

Pipeline serialization is left for a follow-up PR.


libvast/src/pipeline.cpp
```diff
@@ -0,0 +1,187 @@
...
+    ops.push_back(std::move(*op));
+    repr = remaining_repr;
+  }
+  return pipeline{std::move(ops)};
+}
+
+auto pipeline::copy() const -> operator_ptr {
+  auto copied = std::make_unique<pipeline>();
+  copied->operators_.reserve(operators_.size());
+  for (auto& op : operators_) {
```
dominiklohmann, line 111:
```suggestion
  for (const auto& op : operators_) {
```

libvast/src/pipeline.cpp
```diff
@@ -0,0 +1,187 @@
...
+  local_control_plane ctrl;
+  auto dynamic_gen = p.instantiate(std::monostate{}, ctrl);
+  if (!dynamic_gen) {
+    co_yield std::move(dynamic_gen.error());
+    co_return;
+  }
+  auto gen = std::get_if<generator<std::monostate>>(&*dynamic_gen);
+  if (!gen) {
+    co_yield caf::make_error(ec::logic_error,
+                             "right side of pipeline is not closed");
```
dominiklohmann, line 172: This new API allows for calling `pipeline::instantiate` twice, but is that safe to do?

If yes, then `make_local_executor` should take the pipeline by const-ref.

libvast/include/vast/pipeline.hpp
```diff
@@ -0,0 +1,274 @@
...
+    static_assert(std::is_base_of_v<crtp_operator, Self>);
+    return static_cast<const Self&>(*this);
+  }
+};
+
+/// Pipeline operator with a per-schema initialization.
+///
+/// Usage: Override `initialize` and `process`, perhaps `finish`.
+template <class Self, class State, class Output = table_slice>
+class schematic_operator : public crtp_operator<Self> {
```
dominiklohmann, line 196: Name TBD; we have an open conversation in Slack to find a better one. Opening this thread so we don't forget about that one.

dominiklohmann (commented):
I went through the code; I'd like to take a second look after the feedback is incorporated.

libvast/builtins/operators/head.cpp
```diff
@@ -48,43 +49,38 @@ class head_operator : public pipeline_operator {
...
+        break;
       }
-    };
+    }
   }
 
-  [[nodiscard]] auto done() const noexcept -> bool override {
-    return remaining_ == 0;
-  }
+  // TODO: Implement this (with newline semantics).
```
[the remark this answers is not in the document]
jachris: Good point, I guess the question is whether we want to have `head -n` or `head -c` (or both?). I will change the comment to reflect this.

libvast/include/vast/pipeline.hpp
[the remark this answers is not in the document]
jachris: Good point. However, I prefer to keep this as-is, because it is also needed for `copy` (and perhaps some virtual `inspect`, though I'm not sure whether that's viable).

libvast/src/pipeline.cpp
jachris: Yes, and I think that is a feature.

We could use const-ref for `make_local_executor`, but that means that you have to keep the pipeline object alive (so probably at least const-pointer to make this obvious). Or we keep it as-is for now, we can always relax this restriction later.

libvast/builtins/operators/summarize.cpp
```diff
@@ -741,34 +740,42 @@ class summarize_operator2 : public logical_operator<events, events> {
...
+  auto process(table_slice slice, state_type& state) const
+    -> output_type override {
+    if (state) {
+      state->add(to_record_batch(slice));
+    }
+    return {};
+  }
+
+  auto finish(std::unordered_map<type, state_type> states,
+              operator_control_plane& ctrl) const
```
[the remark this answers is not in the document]
jachris: Besides `summarize`, there currently is no need for this function. A schema-specific `sort` or some `taste-tail` could use this. But I don't think we want those, so its fine by me if we remove it.

libvast/src/pipeline.cpp
dominiklohmann: Should make_local_executor then be a member function of `pipeline`?

dominiklohmann: Ah nvm. Let's just resolve this, I don't think it's worth spending more time on this.

libvast/builtins/operators/summarize.cpp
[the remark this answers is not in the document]
jachris: Actually, let's perhaps do this as part of the `summarize` rewrite, because we could also change the API to make `summarize` possible. No need to rewrite this before potentially rewriting it again.

PR #3042, opened 2023-03-28
Fix submodule checkout in the vast-plugins CI job

tobim:
This job used to work because the webfactory/ssh-agent action used to do this, but changed its behavior with v0.8.0.


dominiklohmann (approved):
We discussed this in a call.

PR #3043, opened 2023-03-30
Add Boost as a dependency

tobim:
VAST can benefit from Boost libraries in various areas, such as better stack trace support, correct resource tracking or runtime executable path lookup, while abstracting away platform details.

### DIFF
--- a/web/docs/setup/build.md
+++ b/web/docs/setup/build.md
@@ -33,7 +33,7 @@ Every [release](https://github.com/tenzir/vast/releases) of VAST includes an
 |✓|[CAF](https://github.com/actor-framework/actor-framework)|>= 0.18.7|Implementation of the actor model in C++. (Bundled as submodule.)|
 |✓|[OpenSSL](https://www.openssl.org)||Utilities for secure networking and cryptography.|
 |✓|[FlatBuffers](https://google.github.io/flatbuffers/)|>= 1.12.0|Memory-efficient cross-platform serialization library.|
-|✓|[Boost](https://arrow.apache.org)|>= 1.70.0|Required for multiple reasons.|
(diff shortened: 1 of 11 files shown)


web/docs/setup/build.md
```diff
@@ -33,6 +33,7 @@ Every [release](https://github.com/tenzir/vast/releases) of VAST includes an
 |✓|[CAF](https://github.com/actor-framework/actor-framework)|>= 0.18.7|Implementation of the actor model in C++. (Bundled as submodule.)|
 |✓|[OpenSSL](https://www.openssl.org)||Utilities for secure networking and cryptography.|
 |✓|[FlatBuffers](https://google.github.io/flatbuffers/)|>= 1.12.0|Memory-efficient cross-platform serialization library.|
+|✓|[Boost](https://arrow.apache.org)|>= 1.70.0|Required for multiple reasons.|
```
dominiklohmann, line 36: Please add this to VAST.spdx as well.

dominiklohmann, line 36: Is there any reason why you chose 1.70.0 specifically? For example, I'd be very much interested in 1.81.0 for Boost.Unordered.

dominiklohmann, line 36:
```suggestion
|✓|[Boost](https://arrow.apache.org)|>= 1.70.0|Required as a general utility library.|
```

dominiklohmann (approved):

web/docs/setup/build.md
tobim: That introduced beast. I don't have any opinion at all on this. We can go with 1.81 if you wish.

dominiklohmann: That's already changed, that was my other comment on the same line.

PR #3044, opened 2023-03-30
Catch wrongly queued partitions in the scheduler

tobim:
In case we get a partition that has no queries attached from the query queue we need to skip the scheduling.


dominiklohmann (approved):
We ruled out that this was the bug we've seen, but let's merge this anyways for two reasons:
1. It is generally good practice to program defensively and to make assumptions only very locally
2. This adds a valuable log message.

PR #3046, opened 2023-03-30
Replace indirect download links

lava:
The client-side redirects on the `vast.io` domain would send command-line clients into an endless redirect loop. Therefore, replace these links by their resolved targets.

Fixes tenzir/issues#288.


web/docs/setup/download.md
```diff
@@ -14,15 +14,15 @@ release and the current development version.
 ### Generic Linux
 
 <div align="center" class="padding-bottom--md">
-  <a class="button button--md button--primary margin-right--md" href="/download/vast-linux-static.tar.gz">Static Build (Release)</a>
-  <a class="button button--md button--info margin-left--md" href="/download/vast-linux-static-latest.tar.gz">Static Build (Development)</a>
+  <a class="button button--md button--primary margin-right--md" href="https://github.com/tenzir/vast/releases/latest/download/vast-linux-static.tar.gz">Static Build (Release)</a>
```
dominiklohmann, line 17: Do we want to keep the redirects in `web/src/pages/download/*.tsx` around, or delete them as well?

dominiklohmann (approved):

web/docs/setup/download.md
mavam: I would say remove them.

lava: That's going to break the version in `versioned_docs`, do we want to retro-actively replace the links there as well?

dominiklohmann: For this case I would say yes.

PR #3047, opened 2023-03-30
Reduce conflict potential between rebuilding and queries

dominiklohmann:
With a large value in `vast.rebuild.parallel` or `vast.automatic-rebuild` and a low value in `vast.max-queries`, VAST may end up in a state in which the state at which it is able to respond to queries is limited by the speed at which it is able to rebuild.

This fixes that issue, and then adds some additional logging that helped find this bug.

### DIFF
--- a/libvast/include/vast/query_context.hpp
+++ b/libvast/include/vast/query_context.hpp
@@ -144,13 +144,12 @@ struct query_context {
   vast::ids ids = {};
 
   struct priority {
-    constexpr static uint8_t high = 100;
-    constexpr static uint8_t normal = 25;
-    constexpr static uint8_t low = 1;
+    constexpr static uint64_t normal = 1'000;
+    constexpr static uint64_t low = 1;
   };
 
   /// The query priority.
-  uint8_t priority = priority::normal;
+  uint64_t priority = priority::normal;
 
   /// The issuer of the query.
   std::string issuer = {};
--- a/libvast/src/system/index.cpp
+++ b/libvast/src/system/index.cpp
@@ -825,20 +825,21 @@ void index_state::add_partition_creation_listener(
 
 // -- query handling ---------------------------------------------------------
 
-void index_state::schedule_lookups() {
+auto index_state::schedule_lookups() -> size_t {
   if (!pending_queries.has_work())
-    return;
+    return 0u;
   auto t = timer::start(scheduler_measurement);
   auto num_scheduled = size_t{0};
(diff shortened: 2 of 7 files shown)


patszt (approved):

libvast/src/system/index.cpp
```diff
@@ -1778,7 +1790,15 @@ index(index_actor::stateful_pointer<index_state> self,
...
-      query_context.priority = query_context::priority::high;
+      // Rebuilding a partition should not conflict with other queries, so we
+      // treat rebuilds with a low rather than the high priority that all other
+      // partition transformations get.
+      // Note that the entire index is bound to go away soon as part of the
+      // pipelines thrust, which is why this is just hardcoded here as opposed
+      // to the priority being configurable.
+      query_context.priority = query_context.issuer == "rebuild"
+                                 ? query_context::priority::low
+                                 : query_context::priority::high;
```
dominiklohmann, line 1801: One potential thing to change: We could just always use `low` here.

libvast/include/vast/query_context.hpp
```diff
@@ -144,13 +144,12 @@ struct query_context {
...
-    constexpr static uint8_t high = 100;
-    constexpr static uint8_t normal = 25;
-    constexpr static uint8_t low = 1;
+    constexpr static uint64_t normal = 1'000;
+    constexpr static uint64_t low = 1;
   };
 
   /// The query priority.
-  uint8_t priority = priority::normal;
+  uint64_t priority = priority::normal;
```
tobim, line 152: We sum up query priorities to get a partition priority for the queue. Restricting this to 0-255 was quite intentional. Can we try to find better defaults with that in mind?

tobim (changes requested):
It's good that we fix this issue, albeit with a band-aid. However I would like to avoid allowing an even bigger value space for priorities at the same time.

When we refactor the catalog we should eventually change this so rebuilds don't appear as queries any more to avoid similar problems in the future.

libvast/include/vast/query_context.hpp
dominiklohmann: I think this is a minor issue in code that is bound to disappear entirely with `from vast`, so I think for now this is absolutely alright to leave as-is.

tobim: Alright, since will be replaced in the foreseeable future...

tobim (approved):

PR #3050, opened 2023-03-31
Add `tail` operator

Dakostu:
This PR adds the `tail` operator. Currently only available as a new `crtp_operator`.

### DIFF
--- a/libvast/builtins/operators/tail.cpp
+++ b/libvast/builtins/operators/tail.cpp
@@ -1,0 +1,99 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <vast/concept/parseable/numeric/integral.hpp>
+#include <vast/concept/parseable/vast/pipeline.hpp>
+#include <vast/error.hpp>
+#include <vast/legacy_pipeline.hpp>
+#include <vast/logger.hpp>
+#include <vast/pipeline.hpp>
+#include <vast/plugin.hpp>
+#include <vast/table_slice.hpp>
+
+#include <arrow/type.h>
+
+namespace vast::plugins::tail {
+
+namespace {
+
+class tail_operator final : public crtp_operator<tail_operator> {
+public:
+  explicit tail_operator(uint64_t limit) : limit_{limit} {
+  }
+
+  auto operator()(generator<table_slice> input) const
+    -> generator<table_slice> {
+    auto buffer = std::vector<table_slice>{};
+    auto total_buffered = size_t{0};
+    for (auto&& slice : input) {
+      total_buffered += slice.rows();
+      buffer.push_back(std::move(slice));
+      if (total_buffered - buffer.front().rows() >= limit_) {
+        total_buffered -= buffer.front().rows();
+        buffer.erase(buffer.begin());
+      }
+      co_yield {};
+    }
+    if (buffer.empty())
+      co_return;
+    buffer.front() = vast::tail(buffer.front(), buffer.front().rows()
+                                                  - (total_buffered - limit_));
+    for (auto&& slice : std::move(buffer))
+      co_yield std::move(slice);
+  }
+
+  auto to_string() const -> std::string override {
+    return fmt::format("tail {}", limit_);
+  }
+
+private:
+  uint64_t limit_;
+};
+
+class plugin final : public virtual operator_plugin {
+public:
+  // plugin API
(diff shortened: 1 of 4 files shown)


libvast/builtins/operators/tail.cpp
```diff
@@ -0,0 +1,115 @@
...
+    auto remaining = limit_;
+    for (auto&& slice : input) {
+      slice = vast::tail(slice, remaining);
+      VAST_ASSERT(remaining >= slice.rows());
+      remaining -= slice.rows();
+      co_yield std::move(slice);
+      if (remaining == 0) {
+        break;
+      }
+    }
```
jachris, line 40: As far as I can tell, this does not have the semantics that we want. For example, if we receive 3 tables slices of 10 events each here, and we have `tail 15`, I would expect the last 5 events of the second `table_slice`, and all of the events of the last `table_slice` (so the events 15-30), while this here would output the first 10 events, followed by events 15-20, right? I think we need some kind of buffer.

Dakostu: Yep, indeed.

dominiklohmann: We also need the buffer because the result of tail could be heterogeneous. Imagine the following scneario:

1. slice with 15 rows of schema A
2. slice with 1 row of schema B
3. slice with 4 rows of schema A

Then your output for `tail 10` should be (in order) 

1. the last 5 rows of the first slice
2. all 1 rows of the second slice
3. all 4 rows of the last slice

Dakostu: Luckily, I already had a buffer in my `tail` `legacy_pipeline_operator` implementation.

I implemented a buffer that is order-sensitive.

libvast/builtins/operators/tail.cpp
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  auto operator()(generator<table_slice> input) const
+    -> generator<table_slice> {
+    auto remaining = limit_;
+    auto buffer = std::deque<table_slice>{};
+    auto results = std::deque<table_slice>{};
+    for (auto&& slice : input) {
+      buffer.emplace_front(slice);
+    }
```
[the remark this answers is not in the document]
Dakostu: So essentially `N` would be the upper bound for all incoming `slice`s, with later total truncating in the next loop?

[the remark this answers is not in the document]
Dakostu: Or in code terms: `vast::tail(slice, remaining)`?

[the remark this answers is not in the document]
dominiklohmann: Consider an implementation like this:

```c++
auto buffer = std::vector<table_slice>{};
auto num_buffered = size_t{0};

for (auto&& slice : input) {
  num_buffered += slice.rows();
  buffer.push_back(std::move(slice));
  if (num_buffered - buffer.front().size() >= N) {
    num_buffered -= buffer.front().size();
    buffer.erase(buffer.begin());
  }
  co_yield {};
}

for (auto first = true; auto&& slice : std::move(buffer)) {
  if (first) {
    co_yield tail(slice, num_buffered - N);
    first = false;
  } else {
    co_yield std::move(slice);
  }
}
```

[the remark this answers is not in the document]
Dakostu: I applied your deque-esque vector suggestion.

libvast/builtins/operators/tail.cpp
```diff
@@ -0,0 +1,122 @@
...
+      caf::expected<std::unique_ptr<legacy_pipeline_operator>>> override {
+    const auto* f = pipeline.begin();
+    const auto* const l = pipeline.end();
+    return {
+      std::string_view{f, l},
+      caf::make_error(ec::unimplemented,
+                      "the tail operator is not available in the old version "
+                      "of the pipeline API"),
+    };
+  }
```
dominiklohmann, line 92: I don't think you need to make this one of the old operator plugins if you're not implementing it anyways.

```suggestion
class plugin final : public virtual operator_plugin {
public:
  // plugin API
  auto initialize([[maybe_unused]] const record& plugin_config,
                  [[maybe_unused]] const record& global_config)
    -> caf::error override {
    return {};
  }

  [[nodiscard]] auto name() const -> std::string override {
    return "tail";
  };
```

web/docs/understand/language/operators/tail.md
```diff
@@ -0,0 +1,27 @@
+# tail
+
+Limits the input to the last N results.
```
dominiklohmann, line 3: This also needs a changelog entry in the feature category.

dominiklohmann (approved):
I think this is great; tests are rather minimal, which we should improve once we have this available via `vast exec`. Just two minor things remaining.

web/docs/understand/language/operators/tail.md
Dakostu: I added a changelog entry.

dominiklohmann (approved):

PR #3051, opened 2023-03-31
Add `unique` operator

jachris:


web/docs/understand/language/operators/unique.md
```diff
@@ -0,0 +1,17 @@
...
+unique
+```
+
+## Example
+
+We can remove all duplicates (including non-adjacent ones) by applying `sort` before `unique`.
+
+```
+sort | unique
+```
```
dominiklohmann, line 17: Please add an example to this, e.g., show how

```json
{"foo": 1, "bar": "a"}
{"foo": 1, "bar": "a"}
{"foo": 1, "bar": "a"}
{"foo": 1, "bar": "b"}
{"foo": null, "bar": "b"}
{"bar": "b"}
{"foo": null, "bar": "b"}
{"foo": null, "bar": "b"}
```

becomes

```json
{"foo": 1, "bar": "a"}
{"foo": 1, "bar": "b"}
{"foo": null, "bar": "b"}
{"bar": "b"}
{"foo": null, "bar": "b"}
```

libvast/test/pipeline.cpp
```diff
@@ -179,6 +179,28 @@ TEST(tail 5) {
...
+TEST(unique) {
+  auto ops = unbox(pipeline::parse("select id.orig_h | unique")).unwrap();
+  ops.insert(ops.begin(), std::make_unique<source>(std::vector<table_slice>{
+                            head(zeek_conn_log.at(0), 1), // = 1
+                            head(zeek_conn_log.at(0), 1), // + 0
+                            head(zeek_conn_log.at(0), 5), // + 4
+                            head(zeek_conn_log.at(0), 0), // + 0
+                            head(zeek_conn_log.at(0), 5), // + 4
+                            head(zeek_conn_log.at(0), 7), // + 5
+                            head(zeek_conn_log.at(0), 7), // + 6
```
dominiklohmann, line 191: Please also implement this test within one table slice rather than just at the boundaries of slices, e.g., by calling `concatenate(...)` on the same data set.

jachris: The last slice has a duplicate within itself, but I added another slice to cover the case where the duplicate is positioned differently.

libvast/src/table_slice.cpp
```diff
@@ -600,6 +600,24 @@ split(const table_slice& slice, size_t partition_point) {
   };
 }
 
+auto subslice(const table_slice& slice, size_t begin, size_t end)
+  -> table_slice {
+  VAST_ASSERT(begin <= end && end <= slice.rows());
```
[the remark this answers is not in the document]
jachris: Ah, good point, thanks :+1:

PR #3053, opened 2023-03-31
Ignore VAST v2.4.2 in regression tests

dominiklohmann:
The CI is unable to run VAST v2.4.2 because that patch release was cut after the renaming of the master branch to main, but the v2.4.x branch still assumed the default branch to be named main.


Dakostu (approved):

Dakostu:
`grep -v` successfully removed `2.4.2` from the list of versions on my machine.

PR #3054, opened 2023-04-02
Remove `tsl::robin_map` dependency

dominiklohmann:
Since we now have Boost 1.81 readily available, we can make use of the superior set and map type it contains and remove a dependency from VAST. Swapping this out was pretty straightforward, and only required minor adjusmtents.


patszt (approved):

dominiklohmann:
See for example these recent articles:
- https://martin.ankerl.com/2022/08/27/hashmap-bench-01/
- https://medium.com/@pavel.odintsov/boost-unordered-map-is-a-new-king-of-data-structures-292124d3ee2
- http://bannalia.blogspot.com/2022/11/inside-boostunorderedflatmap.html

There has also been a rather detailed analysis posted somewhere in r/cpp that I cannot find right now, but essentially this is the new best default to use, and as an added benefit we can throw out tsl with its non-standard iterator API and use something with a lot more resources and research behind it.

lava:
The big table in the first link shows `tsl::robin_map` to be faster for most of the benchmarks, or am I misreading that?

The others don't seem to include a comparison between the new boost unordered map and the robin_map.

dominiklohmann:
I didn't have the time to follow-up on this, and may revisit this in the future.

PR #3058, opened 2023-04-04
Restart the systemd service on failure

tobim:
In case of failures like OOM we want the service to restart by default.


mavam (commented):
Mind adding a changelog entry?

mavam (approved):

PR #3060, opened 2023-04-05
Handle queries against /0 subnets correctly

dominiklohmann:
This fixes 2/3 of the issues reported by @satta relating to subnet queries.

Before this change: 
- `<ip> in <ipv6>/0` crashed in Debug mode, UB in Release mode
- `<ip> in <ipv4>/0` printed an unnecessary warning for every dense index for IP addresses loaded

There is still one last issue remaining, which is that `::ffff:0:0/96` is currently not equal to `0.0.0.0/0`. This came up in a question about finding all IPv6 addresses in VAST, which as of this PR is possible with `:ip !in 0.0.0.0/0`, but not yet with `:ip !in ::ffff:0:0/96`. The fix for that is a bit more involved, so I plan to fix it in a separate PR.


libvast/src/index/ip_index.cpp
```diff
@@ -80,10 +80,12 @@ ip_index::lookup_impl(relational_operator op, data_view d) const {
...
         auto topk = x.length();
-        if (topk == 0)
-          return caf::make_error(ec::unspecified,
-                                 "invalid IP subnet length: ", topk);
         auto is_v4 = x.network().is_v4();
+        // If we're in a IPv4/0 subnet, then we can blindly return all IPv6
+        // addresses. If we're in an IPv6/0 subnet, then we can't have any
+        // results at all.
+        if (topk == 0)
+          return is_v4 ? ~v4_.coder().storage() : ids{offset(), false};
```
tobim, line 88: This doesn't pass some trivial tests for me:
```
# Ground truth for number of IPv4 addresses
> cat  libvast_test/artifacts/logs/zeek/conn.log | zeek-cut id.orig_h | grep -v '::' | grep -v '-' | wc -l
8302
# But the index doesn't agree
> vast -q count --estimate 'id.orig_h in 0.0.0.0/0'
160
# Negation is not taken into account
> vast -q count --estimate 'id.orig_h !in 0.0.0.0/0'
160
```

I'm surprised that this index has any v4 specific logic beyond handling of `::ff:x:x`.
`x.length()` should have been incremented by 96 if the address was formatted as v4 way earlier.
If `topk` is 0 here we should return all non-null values or nothing depending on the operator. This should also not be handled by a dedicated branch.

tobim (changes requested):

libvast/src/index/ip_index.cpp
dominiklohmann: We resolved this last issue together; thanks for the help!

tobim (approved):
The logic changes and new tests look right now.

libvast/src/index/ip_index.cpp
```diff
@@ -80,28 +80,30 @@ ip_index::lookup_impl(relational_operator op, data_view d) const {
...
+        // Asking for /128 membership is equivalent to an equality lookup.
+        if (topk == 128)
           return lookup_impl(op == relational_operator::in
                                ? relational_operator::equal
                                : relational_operator::not_equal,
                              x.network());
-        auto result = is_v4 ? v4_.coder().storage() : ids{offset(), true};
+        // If we're in a /96 subnet and the network can be represented as a
+        // valid IPv4 address, then we can blindly ignore all IPv4 addresses:
+        // They're irrelevant by definition.
```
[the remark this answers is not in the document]
dominiklohmann: I applied a version of this.

satta:
Thanks!

PR #3061, opened 2023-04-05
Improve operator type utilities

jachris:
This PR improves the situation where we want to inspect the type of operators and pipelines without actually instantiating them. Most of this is handled by the default implementation in `operator_base` by instantiating the operator with a dummy input. If instantiation has side-effects, the operator can override this behavior.

As a side-effect, this slightly improves the diagnostics of the query endpoint.


libvast/include/vast/tag.hpp
```diff
@@ -0,0 +1,35 @@
...
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#pragma once
+
+#include <variant>
+
+namespace vast {
```
dominiklohmann, line 13: Style nit: Empty line around namespaces, i.e., one after this and one below the matching brace at the end.

dominiklohmann (approved):
I like this a lot better than the old `test_instantiate`, so :+1: on this!

PR #3064, opened 2023-04-11
Upgrade partition transformer to new pipelines

jachris:
This PR changes the partition transformer to make use of the new pipelines. Furthermore, there are a few necessary transitive changes, as well as some small additions to the operator API.


libvast/include/vast/fwd.hpp
```diff
@@ -407,6 +408,7 @@ CAF_BEGIN_TYPE_ID_BLOCK(vast_types, first_vast_type_id)
   VAST_ADD_TYPE_ID((vast::uuid))
   VAST_ADD_TYPE_ID((vast::table_slice_column))
   VAST_ADD_TYPE_ID((vast::pipeline_ptr))
+  VAST_ADD_TYPE_ID((std::shared_ptr<vast::pipeline>))
```
dominiklohmann, line 411: Is this a short-term workaround until the inspection changes are in and pipelines are inpsectable (without the shared_ptr wrapping)?

libvast/include/vast/system/actors.hpp
```diff
@@ -468,6 +468,7 @@ CAF_ALLOW_UNSAFE_MESSAGE_TYPE(std::shared_ptr<vast_uuid_synopsis_map>)
 CAF_ALLOW_UNSAFE_MESSAGE_TYPE(vast::partition_synopsis_ptr)
 CAF_ALLOW_UNSAFE_MESSAGE_TYPE(vast::partition_synopsis_pair)
 CAF_ALLOW_UNSAFE_MESSAGE_TYPE(vast::pipeline_ptr)
+CAF_ALLOW_UNSAFE_MESSAGE_TYPE(std::shared_ptr<vast::pipeline>)
```
dominiklohmann, line 471: Same as for the comment in fwd.hpp: IIUC this is temporary, so please add a comment to remind us to remove it again.

dominiklohmann (approved):
I think the naive implementation here is fine. I left a few comments, mostly looking for clarification. if the compaction unit tests pass on this one I think it should be good to go.

libvast/include/vast/fwd.hpp
jachris: I tried to preserve the original interface (using `shared_ptr<legacy_pipeline>`) as far as possible, but given your comment, there presumably is no reason for `shared_ptr`. I now copied the changes from the other branch and removed the `shared_ptr` with 44a8eaa.

dominiklohmann: Perfect, that's a lot better!

libvast/include/vast/pipeline.hpp
```diff
@@ -179,6 +211,31 @@ class pipeline final : public operator_base {
   auto predicate_pushdown(expression const& expr) const
     -> std::optional<std::pair<expression, operator_ptr>> override;
 
+  template <class Inspector>
+  friend auto inspect(Inspector& f, pipeline& x) -> bool {
+    if constexpr (Inspector::is_loading) {
+      auto repr = std::string{};
+      if (not f.object(x)
+                .pretty_name("vast.logical-pipeline")
```
dominiklohmann, line 219: The pretty name here and below should just be `vast.pipeline`, I think, as the class is no longer named `logical_pipeline` either.

dominiklohmann (approved):
The tests all pass now, and I tested rebuilding and temporal compaction locally in a minimal setup. Just two places to go now! 🚀

PR #3066, opened 2023-04-11
Implement JSON parser

patszt:
Implementaton of a pipeline operator for parsing JSON formatted chunks.

The on demand parsing can fail after some fields/values were already added to the adaptive table slice builder. It is needed to revert these changes. This can be achieve by calling .cancel() method on an object returned by adaptive_talbe_slice_builder::push_row()

The adaptive_table_slice_builder implementation was changed so that the each individual builders are only built once we add a value to a field. Previously it would create a builder when field was created. This approach complicated the cancelling logic


libvast/builtins/formats/json.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+#include <simdjson.h>
+
+namespace vast::plugins::json {
+
+namespace {
+
+using parse_result = std::optional<simdjson::error_code>;
+
+class parser_buffer {
```
dominiklohmann, line 30: Can you move this class to `vast/detail/padded_buffer.hpp`, rename it to `padded_buffer` and take the padding size as a template argument? I think this is more generally useful than just here, and it makes sense to have it in its own place.

NB: I also think that the entire class can be made `constexpr`, which would guarantee that it works correctly.

libvast/builtins/formats/json.cpp
```diff
@@ -0,0 +1,356 @@
...
+            json_to_parse_buffer.reset();
+            emit_warning(control_plane, err);
+            continue;
+          }
+          for (auto doc : stream) {
+            if (auto parse_result = parse(doc, slice_builder)) {
+              emit_warning(control_plane, *parse_result);
+              continue;
+            }
+            // change max table slice size to be fetched from options.
```
dominiklohmann, line 286:
```suggestion
            // TODO: change max table slice size to be fetched from options.
```

libvast/builtins/formats/json.cpp
```diff
@@ -0,0 +1,356 @@
...
+    row.cancel();
+    return err;
+  }
+  return {};
+}
+
+void emit_warning(operator_control_plane& control_plane,
+                  simdjson::error_code err) {
+  control_plane.warn(caf::make_error(ec::parse_error, error_message(err)));
+}
```
dominiklohmann, line 243: This makes it so we always just forward the simdjson warnings directly to the user, which are not written in a user-facing way. Can you either change this, or write a follow-up issue to improve the error messages?

dominiklohmann (commented):
I played around with this a bit—thanks for providing such an extensive test suite—and mostly have minor things for now. I think once the feedback is incorporated we can quickly approve and merge this and then improve it based on that.

libvast/builtins/formats/json.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+parse_result
+parse_impl(simdjson::ondemand::value val, auto& pusher, size_t depth);
+
+parse_result parse_number(simdjson::ondemand::value val, auto& pusher) {
+  switch (val.get_number_type().value_unsafe()) {
+    case simdjson::ondemand::number_type::floating_point_number: {
+      auto result = val.get_double();
+      if (result.error())
+        return result.error();
```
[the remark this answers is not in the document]
patszt: Not sure. Look at the weird comment in this function.
https://github.com/simdjson/simdjson/blob/master/singleheader/simdjson.h#L11866

I would say given "2153\n" would yield double_type but would fail on the get_double()

I have tested

```
TEST(WTF)
{
  constexpr auto json = std::string_view{R"({"ble": 46363643;})"};
  simdjson::ondemand::parser p;
  simdjson::padded_string k{json};
  for (auto s : p.iterate_many(k).value())
  {
    auto obj = s.get_object().value_unsafe();
    for (auto p : obj)
    {
      if (p.error())
      {
        MESSAGE("p err");
      }
      MESSAGE(p.unescaped_key().value_unsafe());
      auto v = p.value();
      auto type = v.get_number_type().value_unsafe();
      MESSAGE((int)type);
      auto dbl = v.get_double();
      if (dbl.error())
      {
        MESSAGE("dbl error");
      }
      else
      {
      MESSAGE(dbl.value_unsafe());

      }
    }
  }
}
```

And the output is 
```
 -> [98, 108, 101] [line 345]
  -> 1 [line 348]
  -> dbl error [line 352]
```
Proving that it failed :/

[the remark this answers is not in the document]
dominiklohmann: Okay, thanks for confirming. Then let's keep this as-is. :|

libvast/builtins/formats/json.cpp
```diff
@@ -0,0 +1,356 @@
...
+      return parse_number(val, pusher);
+    case simdjson::ondemand::json_type::boolean: {
+      auto result = val.get_bool();
+      if (result.error())
+        return result.error();
+      pusher.add(result.value_unsafe());
+      return {};
+    }
+    case simdjson::ondemand::json_type::string:
+      return parse_string(val, pusher);
```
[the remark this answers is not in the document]
patszt: This is because the type() function used in switch statement is very optimistic. It only checks the first character.
https://github.com/simdjson/simdjson/blob/master/singleheader/simdjson.h#L29098

I think we can have similar examples that yield a correct type but raises an error when we try to get a value like in the double example above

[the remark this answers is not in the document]
dominiklohmann: Oh wow, that's crazy. Can you link to https://github.com/simdjson/simdjson/blob/f435fddda148bec9e798dd3bdb1988af569d40af/singleheader/simdjson.h#L29098-L29117 (permalink) in the code for the future reader?

libvast/builtins/formats/json.cpp
```diff
@@ -0,0 +1,356 @@
...
+    case simdjson::ondemand::json_type::array:
+      return parse_array(val, pusher, depth + 1);
+    case simdjson::ondemand::json_type::object:
+      return parse_object(val, pusher, depth + 1);
+  }
+  die("unhandled json object type in switch statement");
+}
+
+parse_result parse_entry(simdjson::ondemand::object object,
```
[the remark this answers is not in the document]
patszt: This is no longer needed after proper row.cancel() implementation. This method will be removed and parse_object will be used instead in the next commit

libvast/builtins/formats/json.cpp
patszt: I will write the followup

patszt: Followup story created: https://github.com/tenzir/issues/issues/274

libvast/builtins/formats/json.cpp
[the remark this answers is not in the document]
patszt: Fixed in https://github.com/tenzir/vast/pull/3066/commits/2d2fd6a1676907f87dd8fd9fc230f7b7061bd4fa

libvast/builtins/formats/json.cpp
patszt: Done. I was wondering if we want to make it more generic as it's in a common header file.
I ended up thinking that is not worth the time for now. If someone were to use it with different allocation strategy or a different value type than 'char' then he can refactor it

libvast/include/vast/detail/padded_buffer.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+#pragma once
+
+#include <algorithm>
+#include <memory>
+#include <string_view>
+
+namespace vast::detail {
+
+template <auto PaddingSize, auto PaddingValue>
```
dominiklohmann, line 17: I think this is easier to read.

```suggestion
template <size_t PaddingSize, char PaddingValue>
```

dominiklohmann (approved):
This is great, thanks for incorporating the review feedback. I have a few small things I would like to improve documentation for, but it's better to get this merged now and to do that work in follow-up PRs.

libvast/include/vast/detail/padded_buffer.hpp
patszt: Wanted to be more generic, but since the other class inernals are not generic then it may make sense to only allow size_t and char for now

libvast/builtins/formats/json.cpp
```diff
@@ -0,0 +1,340 @@
...
+
+#include <arrow/record_batch.h>
+
+#include <simdjson.h>
+
+namespace vast::plugins::json {
+
+namespace {
+
+using parse_result = simdjson::error_code;
```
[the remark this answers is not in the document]
patszt: Nope!

PR #3067, opened 2023-04-12
Add support for user-defined operator aliases

jachris:
This PR adds support for user-defined operator aliases that can be defined in the config. For example:

~~~yaml
vast:
  operators:
    anonymize_urls: |
      replace net.url="xxx"
    aggregate_flows: |
       summarize 
         pkts_toserver=sum(flow.pkts_toserver),
         pkts_toclient=sum(flow.pkts_toclient),
         bytes_toserver=sum(flow.bytes_toserver),
         bytes_toclient=sum(flow.bytes_toclient),
         start=min(flow.start),
         end=max(flow.end)
       by
         timestamp,
         src_ip,
         dest_ip
       resolution
         10 mins 
~~~

This allows composing more complex pipelines, for instance `from vast | anonymize_urls | aggregate_flows | write json to stdout`. The PR only affects the new pipelines, which will eventually be used everywhere.  The new pipeline parser also looks for definitions in `vast.pipelines` instead of `vast.operators`. However, this behavior is immediately considered deprecated and only available for compatibility.

### DIFF
--- a/libvast/builtins/endpoints/query.cpp
+++ b/libvast/builtins/endpoints/query.cpp
@@ -591,7 +591,8 @@ request_multiplexer_actor::behavior_type request_multiplexer(
           return rq.response->abort(422, "missing parameter 'query'\n",
                                     caf::error{});
         }
-        auto parse_result = pipeline::parse(*query_string);
+        // TODO
+        auto parse_result = pipeline::parse(*query_string, record{});
         if (!parse_result)
           return rq.response->abort(400, "invalid query\n",
                                     parse_result.error());
--- a/libvast/builtins/endpoints/query.cpp
+++ b/libvast/builtins/endpoints/query.cpp
@@ -453,6 +453,7 @@ struct request_multiplexer_state {
 
   system::index_actor index_ = {};
   std::unordered_map<std::string, query_manager_actor> live_queries_ = {};
+  record config_ = {};
 };
 
 query_manager_actor::behavior_type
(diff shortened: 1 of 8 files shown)


libvast/builtins/endpoints/query.cpp
```diff
@@ -720,10 +723,13 @@ class plugin final : public virtual rest_endpoint_plugin {
     return endpoints;
   }
 
-  system::rest_handler_actor
-  handler(caf::actor_system& system, system::node_actor node) const override {
-    return system.spawn(request_multiplexer, node);
+  auto handler(caf::actor_system& system, system::node_actor node) const
+    -> system::rest_handler_actor override {
+    return system.spawn(request_multiplexer, node, &config_);
```
dominiklohmann, line 728: While technically safe to do as the plugin is a singleton, can you change the config to be passed by value and kept in the state of the request multiplexer as a hard copy? Otherwise, if we ever want to support loading/unloading plugins at runtime this will become a problem.

dominiklohmann (approved):
I think this is a very small feature that is incredibly useful to have. We discussed in Slack that I'll write the documentation for this in an upcoming PR.

libvast/test/pipeline.cpp
```diff
@@ -342,5 +348,44 @@ TEST(from_read_parsing) {
   }
 }
 
+TEST(user defined operator alias) {
+  // We could detect some errors in the config file when loading the config.
+  // This test assumes that the error is only triggered when using the alias.
+  auto config_data = unbox(from_yaml(R"__(
+vast:
+  operators:
```
dominiklohmann, line 356: Please add a changelog entry. Suggestion

> The `vast.operators` section in the configuration file supersedes the now deprecated `vast.pipelines` section and more generally enables user-defined operators. Defined operators now must use the new, textual format introduced with VAST v3.0, and are available for use in all places where pipelines are supported.

PR #3069, opened 2023-04-12
Introduce a bunch of small improvements

dominiklohmann:
In no particular order, this is a bundling of changes that I amassed locally over the past week for which I didn't really know where to put them. In particular, this does the following:

- Revert the removal of the `extend` and `replace` operators now that we decided for `put` not to be a replacement of it.
- Fix the API documentation and a changelog entry for the `/query` endpoint
- Upgrade Docusaurus, and make some minor theming changes to the navbar
- Update the bundled simdjson
- Fix a few clang-tidy warnings

Sorry for bundling this all together—I didn't really know where to put all this—, but commit-by-commit this should hopefully be very easy to review.


tobim:
I checked all commits except the put-operator related ones. Green light from me, but I let someone from @tenzir/pipelines-squad do the rest.

web/docs/understand/language/operators/extend.md
```diff
@@ -0,0 +1,28 @@
+# extend
+
+Adds the configured fields with fixed values.
+
+:::warning Experimental
+We plan to merge the `extend` and `replace` operators into a single `put`
+operator in the near future, removing the need for the `extend` operator.
+:::
```
jachris, line 8: The plan changes with this PR.

web/docs/understand/language/operators/replace.md
```diff
@@ -0,0 +1,29 @@
+# replace
+
+Replaces the fields matching the configured extractors with fixed values.
+
+:::warning Experimental
+We plan to merge the `extend` and `replace` operators into a single `put`
+operator in the near future, removing the need for the `replace` operator.
+:::
```
jachris, line 8: Same here.

web/docs/understand/language/operators/replace.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Synopsis
+
+```
+replace FIELDS=VALUE[, …]
+```
+
+### Fields
+
+The fields to replace with fixed values.
```
jachris, line 18: Is this terminology accurate? `replace` still uses extractors.

web/docs/understand/language/operators/put.md
```diff
@@ -1,41 +1,47 @@
...
 ```
 
-### Extractors
+### Fields
 
-The extractors of fields to replace with a new value. For extractors that do not
-match any fields in the input events, the operator adds a new field at the end.
+The names of fields in the output data. If the right-hand side of the assignment
+is omitted, the field name is implicitly used as an extractor. If multiple
+fields match the extractor, the first matching field is used in the output.
```
jachris, line 16:
```suggestion
The names of fields in the output data. If the right-hand side of the assignment
is omitted, the field name is implicitly used as an extractor. If multiple
fields match the extractor, the first matching field is used in the output. If no
fields match, `null` is assigned instead.
```

web/docs/understand/language/operators/put.md
```diff
@@ -1,41 +1,47 @@
...
-put source_ip="REDACTED", note="masked source IP address for GDPR-compliance"
+put source_ip=src_ip, dest_ip, dest_port, payload="REDACTED"
 ```
 
-The pipeline will replace the field `source_ip` with a fixed value, and add a
-new field `note` at the end, leading to this output:
+The pipeline will always result in events with the four fields `source_ip`
+(renamed from src_ip), `dest_ip` (unchanged), `dest_port` (all null as the field
+is not available in the input data), and `payload` (set to the fixed value
+`"REDACTED"`) in that order.
```
jachris, line 38:
```suggestion
The pipeline will always result in events with the four fields `source_ip`
(renamed from `src_ip`), `dest_ip` (unchanged), `dest_port` (`null`, as the field
is not available in the input data), and `payload` (set to the fixed value
`"REDACTED"`) in that order.
```

jachris (approved):
I did not check `extend.cpp` and `replace.cpp`, but if I'm not mistaken the changes there just come from the revert. 

Also, please add `extend` and `replace` back to `libvast/test/pipeline.cpp::to_string`.

web/docs/understand/language/operators/put.md
```diff
@@ -1,41 +1,47 @@
 # put
 
-Assign new values to given fields in the input data, or add new fields if the
-assigned-to field did not exist.
+Reduce the input to the specified fields in the input data in the specified
+order with the specified values.
```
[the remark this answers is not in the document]
dominiklohmann: I rephrased this as "Returns new events with the specified fields only," which I think makes this clear combined with the rephrased synopsis below.

PR #3075, opened 2023-04-14
Implement `-` as an alias for stdout and stdin

dominiklohmann:
This implements `-` as a short form alias for the `stdout` and `stdin` connectors. This came about because I was wondering how we'd implement `suricata` and `zeek-json` as aliases for the new JSON parser, and this shows that it's really easy to do.

This also makes the local control plane emit warnings via `VAST_WARN` rather than crash, which I believe is always what we want for the local control plane.

### DIFF
--- a/libvast/builtins/connectors/dash.cpp
+++ b/libvast/builtins/connectors/dash.cpp
@@ -1,0 +1,65 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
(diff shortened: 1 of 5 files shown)


libvast/builtins/connectors/dash.cpp
```diff
@@ -0,0 +1,65 @@
...
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <vast/plugin.hpp>
+
+namespace vast::plugins::dash {
+
+// -- loader plugin -----------------------------------------------------
```
mavam, line 13:
```suggestion
```

mavam (approved):
This looks like it can work.

It'd be nice to have a small unit test.

dominiklohmann:
> It'd be nice to have a small unit test.

I provided an integration test, as the aliasing isn't all that easy to test in a unit test without replicating a full test.

PR #3076, opened 2023-04-14
Upgrade exporter to use new pipelines

jachris:


libvast/src/system/exporter.cpp
```diff
@@ -42,103 +45,107 @@ void shutdown_stream(
...
+    ++it;
+    if (it == self->state.executor.end()) {
+      VAST_DEBUG("{} has exhausted its executor", *self);
+      break;
+    }
+    auto result = *it;
+    if (!result) {
+      VAST_WARN("{} encountered an error during execution: {}", *self,
+                result.error());
+      // TODO: Inform the user? Quit actor?
```
dominiklohmann, line 115: We should quit here, otherwise we hang indefinitely.

PR #3078, opened 2023-04-14
Add `shell` operator

mavam:
The `shell` operator makes it possible to execute an arbitrary system command.


mavam:
@tobim can you help me to link against Boost Filesystem in Nix?

@dominiklohmann can you help me to update the dependencies in the Linux for us mortals?

tobim:
> @tobim can you help me to link against Boost Filesystem in Nix?

I'm not sure what the problem is here. It seems to be working in CI right now and I can also build it locally.

mavam:
> I'm not sure what the problem is here. It seems to be working in CI right now and I can also build it locally.

Oh wow, I haven't added any logic and it magically works OOTB? That's a first. Sorry for the noise.

libvast/builtins/operators/shell.cpp
```diff
@@ -0,0 +1,209 @@
...
+  auto make_operator(std::string_view pipeline) const
+    -> std::pair<std::string_view, caf::expected<operator_ptr>> override {
+    using parsers::optional_ws_or_comment, parsers::required_ws_or_comment,
+      parsers::end_of_pipeline_operator, parsers::operator_arg;
+    const auto* f = pipeline.begin();
+    const auto* const l = pipeline.end();
+    const auto p = -(required_ws_or_comment >> operator_arg)
+                   >> optional_ws_or_comment >> end_of_pipeline_operator;
+    std::string command;
+    if (not p(f, l, command)) {
```
tobim, line 191: This doesn't seem to handle arguments for the command, which is a severe limitation:
```fish
> journalctl -e -o json | head -n 1 | build/clang/debug/bin/vast exec 'load stdin | shell jq -n 4 | save stdout'
!! invalid_argument: failed to parse pipeline: !! syntax_error: invalid query: !! unspecified: failed to parse pipeline ' shell jq -n 4 | save stdout': !! syntax_error: failed to parse shell operator: ' jq -n 4 | save stdout'

> journalctl -e -o json | head -n 1 | build/clang/debug/bin/vast exec 'load stdin | shell "jq -n 4" | save stdout'                                                                                               vast-dev-env 
4
```

libvast/builtins/operators/shell.cpp
```diff
@@ -0,0 +1,209 @@
...
+namespace vast::plugins::shell {
+namespace {
+
+class shell_operator final : public crtp_operator<shell_operator> {
+public:
+  explicit shell_operator(std::string command) : command_{std::move(command)} {
+  }
+
+  auto operator()(generator<chunk_ptr> input,
+                  operator_control_plane& ctrl) const -> generator<chunk_ptr> {
```
tobim, line 35: I was hoping to be able to use this one at the start of a pipeline like this, but it doesn't work yet:
```fish
> build/clang/debug/bin/vast exec 'shell "journalctl -e -o json" | parse json | head 1 | write csv to stdout'
!! invalid_argument: cannot execute pipeline that is not closed: shell "journalctl -e -o json" | parse json | head 1 | local print csv | save stdout
```

libvast/CMakeLists.txt
```diff
@@ -318,7 +318,8 @@ target_link_libraries(libvast PUBLIC ${ARROW_LIBRARY})
 dependency_summary("Apache Arrow" ${ARROW_LIBRARY} "Dependencies")
 
 # Link against Boost.
-target_link_libraries(libvast PRIVATE Boost::headers)
+target_link_libraries(libvast PRIVATE Boost::headers Boost::filesystem)
+target_link_libraries(libvast_builtins PRIVATE Boost::headers Boost::filesystem)
```
tobim, line 322: I believe you need to add `libboost-filesystem1.81-dev` to `scripts/debian/install-dev-dependencies.sh`. Not sure what needs to be done on the macOS side.

dominiklohmann: I'll push fixes for both.

libvast/builtins/operators/shell.cpp
dominiklohmann: Allowing any argument here is dangerous from a language perspective. I would rather forbid it right now and then consider removing the required quotes in the future, but any such thing should first go through @jachris.

libvast/builtins/operators/shell.cpp
dominiklohmann: You can add a `load file -` for now before it, but that doesn't quite do the correct thing because you don't want to open stdin. The operator instead needs an additional overload for when its input is void.

libvast/builtins/operators/shell.cpp
```diff
@@ -0,0 +1,209 @@
...
+  };
+
+  auto make_operator(std::string_view pipeline) const
+    -> std::pair<std::string_view, caf::expected<operator_ptr>> override {
+    using parsers::optional_ws_or_comment, parsers::required_ws_or_comment,
+      parsers::end_of_pipeline_operator, parsers::operator_arg;
+    const auto* f = pipeline.begin();
+    const auto* const l = pipeline.end();
+    const auto p = -(required_ws_or_comment >> operator_arg)
+                   >> optional_ws_or_comment >> end_of_pipeline_operator;
```
dominiklohmann, line 189: Please apply this diff to make the parser accept a few more things.

```diff
diff --git a/libvast/builtins/operators/shell.cpp b/libvast/builtins/operators/shell.cpp
index 8468c18085..88111bf85b 100644
--- a/libvast/builtins/operators/shell.cpp
+++ b/libvast/builtins/operators/shell.cpp
@@ -181,11 +181,11 @@ public:
   auto make_operator(std::string_view pipeline) const
     -> std::pair<std::string_view, caf::expected<operator_ptr>> override {
     using parsers::optional_ws_or_comment, parsers::required_ws_or_comment,
-      parsers::end_of_pipeline_operator, parsers::qqstr;
+      parsers::end_of_pipeline_operator, parsers::operator_arg;
     const auto* f = pipeline.begin();
     const auto* const l = pipeline.end();
-    const auto p = -(required_ws_or_comment >> qqstr) >> optional_ws_or_comment
-                   >> end_of_pipeline_operator;
+    const auto p = required_ws_or_comment >> operator_arg
+                   >> optional_ws_or_comment >> end_of_pipeline_operator;
     std::string command;
     if (not p(f, l, command)) {
       return {
diff --git a/libvast/include/vast/concept/parseable/string/quoted_string.hpp b/libvast/include/vast/concept/parseable/string/quoted_string.hpp
index 78cdfa7557..c67c1dde88 100644
--- a/libvast/include/vast/concept/parseable/string/quoted_string.hpp
+++ b/libvast/include/vast/concept/parseable/string/quoted_string.hpp
@@ -59,7 +59,7 @@ public:
         return (ignore_esc >> ignore_esc) ->* [&] { x.push_back(Esc); return Esc; };
       // clang-format on
     }();
-    auto str_chr = esc_quote | esc_esc | (parsers::printable - quote);
+    auto str_chr = esc_quote | esc_esc | (parsers::any - quote);
     auto quoted_str = ignore_quote >> *str_chr >> ignore_quote;
     return quoted_str(f, l, x);
   }
```

libvast/builtins/operators/shell.cpp
tobim: The second example doesn't work either. `jq -n 4` should set the indentation level to 4, not print a single "4".

web/docs/understand/operators/transformations/shell.md
```diff
@@ -0,0 +1,68 @@
...
+```
+print json | save stdout
+```
+
+The [`print`](../transformations/print.md) operator produces raw bytes and
+[`save`](../sinks/save.md) accepts raw bytes. The `shell` operator therefore
+fits right in the middle:
+
+```
+print json | shell jq | save stdout
```
dominiklohmann, line 51: This example feels a bit useless, can you use something else here? I liked the Zeek example a lot that you posted in Discord earlier.

libvast/builtins/operators/shell.cpp
dominiklohmann: `jq -n 4` just prints `4` for me on the command line as well. If I run your example with `jq --indent 4` it works as you'd expect.

From `man jq`:

> `--indent n`:
> Use the given number of spaces (no more than 8) for indentation.

web/docs/understand/operators/transformations/shell.md
mavam: I added a more elaborate example. We could consider ditching the `jq` as well, but I do want to start with a simple one to explain the mechanics.

libvast/builtins/operators/shell.cpp
mavam: Let me try make this happen as part of this PR.

web/docs/understand/operators/transformations/shell.md
mavam: @dominiklohmann just to confirm, the function I need to implement is this one?

```cpp
  auto operator()(generator<std::monostate> input,
                  operator_control_plane& ctrl) const -> generator<chunk_ptr> {
    // TODO: implement
  }
```

I wonder how I can reuse some of the coroutine code from the other function. Maybe with a templated helper function with the same signature, but one that does `if constexpr` on the first argument? Then I could dispatch on `std::monostate` vs `chunk_ptr`. It might get convoluted, but I could do the same on the return value for a `shell` sink.

libvast/builtins/operators/shell.cpp
tobim: It seems my brain turned the `n` into a shortopt while glancing at that man page.

libvast/CMakeLists.txt
mavam: I think it needs a tad more CMake massaging:

```
-- Generating done
  The link interface of target "vast::libvast_builtins" contains:

    Boost::headers

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /__w/vast/vast/_install/lib/cmake/vast/VASTConfig.cmake:84 (include)
  CMakeLists.txt:15 (find_package)


CMake Generate step failed.  Build files cannot be regenerated correctly.
Error: Process completed with exit code 1.
```

libvast/builtins/operators/shell.cpp
mavam: @tobim can you try again? I now made `shell` a source as well.

libvast/CMakeLists.txt
dominiklohmann: I pushed a commit that should fix this.

libvast/builtins/operators/shell.cpp
tobim: Nice, it seems to be working!

libvast/CMakeLists.txt
```diff
@@ -318,8 +318,9 @@ target_link_libraries(libvast PUBLIC ${ARROW_LIBRARY})
 dependency_summary("Apache Arrow" ${ARROW_LIBRARY} "Dependencies")
 
 # Link against Boost.
-target_link_libraries(libvast PRIVATE Boost::headers Boost::filesystem)
-target_link_libraries(libvast_builtins PRIVATE Boost::headers Boost::filesystem)
+target_link_libraries(libvast PUBLIC Boost::headers Boost::filesystem)
```
tobim, line 321: Why did you change it to public? Afaik we aren't using boost in the interface.

libvast/CMakeLists.txt
mavam: Works like a charm!

libvast/CMakeLists.txt
mavam: "you" = @dominiklohmann 

Just to make sure it pops up in your feed. 🙂

dominiklohmann: That has to be because it's used in a builtin. We link the builtins against plugin unit tests so they can be used in them, and plugin unit tests may be built against an installed libvast. So that's why it's got to be public.

libvast/builtins/operators/shell.cpp
```diff
@@ -0,0 +1,292 @@
...
+    return result;
+  }
+
+  auto reading() -> bool {
+    return child_.running() && not stdout_.eof();
+  }
+
+  auto writing() -> bool {
+    return child_.running() && not stdin_.eof();
+  }
```
dominiklohmann, line 64:
```suggestion
  auto reading() const -> bool {
    return child_.running() && not stdout_.eof();
  }

  auto writing() const -> bool {
    return child_.running() && not stdin_.eof();
  }
```

libvast/builtins/operators/shell.cpp
```diff
@@ -0,0 +1,292 @@
...
+    while (not chunks.empty()) {
+      auto chk = chunks.front();
+      VAST_DEBUG("yielding chunk {}/{} with {} bytes", ++i, total, chk->size());
+      co_yield chk;
+      chunks.pop();
+    }
+  }
+
+  auto to_string() const -> std::string override {
+    return fmt::format("shell \"{}\"", command_);
```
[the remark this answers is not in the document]
jachris: You can use `escape_operator_arg` to do this.
```suggestion
    return fmt::format("shell {}", escape_operator_arg(command_));
```

libvast/CMakeLists.txt
tobim: Thanks for the explanation. I think I have an idea for what we can do about that, but I guess it's low priority.

libvast/builtins/operators/shell.cpp
mavam: Not possible, `child_.running()` is non-const. It's a Boost Process thing, don't ask me why.

mavam:
@dominiklohmann after adding your suggested integration tests, I now get:

```
2023-05-30 14:39:01,584 WARNING      !! invalid_argument: cannot execute pipeline that is not closed: version | shell xxd
```

My hunch is that I need to rebase after your open PR is merged, right?

mavam:
Actually, the problem is the pipeline itself: `version` doesn't deliver bytes, so this couldn't work in the first place. (Although it would have been nice because it doesn't require any file input.)

dominiklohmann:
> Actually, the problem is the pipeline itself: `version` doesn't deliver bytes, so this couldn't work in the first place. (Although it would have been nice because it doesn't require any file input.)

Just put a print json or so in the middle, then it will work as expected.

mavam:
Okay. But I'll switch to something other than `version` which may require a lot of baseline changes moving forward.

PR #3079, opened 2023-04-14
Expose the lower-level `load`, `parse`, `print`, and `save` operators

dominiklohmann:
This exposes the expert operators underlying `from … read …` and `write … to …`.

Fixes tenzir/issues#276
Fixes tenzir/issues#254

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -6,6 +6,7 @@ Checks: >
   -cert-dcl16-c,
   -cert-err58-cpp,
   -clang-analyzer-alpha*,
+  -cppcoreguidelines-avoid-const-or-ref-data-members,
   -cppcoreguidelines-avoid-magic-numbers,
   -cppcoreguidelines-avoid-non-const-global-variables,
   -cppcoreguidelines-macro-usage,
--- a/changelog/next/features/3064--operator-aliases.md
+++ b/changelog/next/features/3064--operator-aliases.md
@@ -1,0 +1,7 @@
+User-defined operator aliases make pipelines easier to use by enabling users to
+encapsulates pipelines into a new operator. To define a user-defined operator
+alias, add an entry to the `vast.operators` section of your configuration.
+
+Compaction now makes use of the new pipeline operators, and allows pipelines to
+be defined inline instead in addition to the now deprecated `vast.pipelines`
+configuration section.
--- a/changelog/next/features/3064--operator-aliases.md
+++ b/changelog/next/features/3064--operator-aliases.md
@@ -1,5 +1,5 @@
 User-defined operator aliases make pipelines easier to use by enabling users to
-encapsulates pipelines into a new operator. To define a user-defined operator
+encapsulate a pipelinea into a new operator. To define a user-defined operator
 alias, add an entry to the `vast.operators` section of your configuration.
 
 Compaction now makes use of the new pipeline operators, and allows pipelines to
--- a/libvast/builtins/operators/print.cpp
+++ b/libvast/builtins/operators/print.cpp
@@ -30,13 +30,24 @@ class print_operator final
   : public schematic_operator<print_operator, printer_plugin::printer,
                               generator<chunk_ptr>> {
 public:
-  explicit print_operator(const printer_plugin& printer, record config) noexcept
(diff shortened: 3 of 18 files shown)


changelog/next/features/3064--operator-aliases.md
```diff
@@ -0,0 +1,7 @@
+User-defined operator aliases make pipelines easier to use by enabling users to
+encapsulates pipelines into a new operator. To define a user-defined operator
```
Dakostu, line 2:
```suggestion
encapsulate pipelines into a new operator. To define a user-defined operator
```

jachris: This, plus `a pipeline` instead of `pipelines`.

libvast/builtins/operators/print.cpp
```diff
@@ -0,0 +1,104 @@
...
+    auto printer_name = std::string{};
+    if (!p(f, l, printer_name)) {
+      return {
+        std::string_view{f, l},
+        caf::make_error(ec::syntax_error,
+                        fmt::format("failed to parse print operator: '{}'",
+                                    pipeline)),
+      };
+    }
+    const auto* printer = plugins::find<printer_plugin>(printer_name);
```
jachris, line 85: I think you may have to test against `printer_allows_joining()` here?

.clang-tidy
```diff
@@ -6,6 +6,7 @@ Checks: >
   -cert-dcl16-c,
   -cert-err58-cpp,
   -clang-analyzer-alpha*,
+  -cppcoreguidelines-avoid-const-or-ref-data-members,
```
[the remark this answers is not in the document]
dominiklohmann: Yes, this is a new warning for LLVM 16 that's rather annoying. I ran into it while writing this code, and it doesn't make a lot of sense to me. See [clang-tidy docs](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html).

changelog/next/features/3064--operator-aliases.md
dominiklohmann: Nice catch, fixed in a15465b4a

libvast/include/vast/pipeline.hpp
```diff
@@ -174,11 +174,16 @@ class pipeline final : public operator_base {
   /// pipelines, for example `(a | b) | c` becomes `a | b | c`.
   explicit pipeline(std::vector<operator_ptr> operators);
 
-  /// Parses a logical pipeline from its textual representation. It is *not*
-  /// guaranteed that `parse(to_string())` succeeds.
+  /// Parses a pipeline from its textual representation.
   static auto parse(std::string_view repr, const vast::record& config)
     -> caf::expected<pipeline>;
```
jachris, line 179: If we want to require the round-trip property, I would mention it in the documentation.

jachris (approved):

libvast/builtins/operators/print.cpp
dominiklohmann: That's a good point, should I perhaps change the operator to check the schema and to hard error when encountering a second one?

jachris: It would be more in sync with `write ... to ...` if we reject this immediately.

dominiklohmann: I gave this some more thought, and while I can see that it would work closer to `write … to …`, I disagree because it limits the power of the lower-level operator. If we make it warn once `print_operator::initialize(...)` is called a second time we can instead also use the lower-level operator when we know we're doing the right thing.

It would also allow for removing the internal print-and-save operator in the future and replacing it with a `group-by #type (=> print <format> | save <connector>)` instead.

libvast/include/vast/pipeline.hpp
dominiklohmann: I like that. I'll add a variant of this even though we don't have `operator_location` in this branch yet.

libvast/builtins/operators/print.cpp
jachris: Yes, this is convincing. Sounds good! :rocket:

PR #3083, opened 2023-04-17
Fix partition selection for the `rebuild` command

dominiklohmann:
This fixes the `rebuild` command not to loop infinitely over all partitions when at least one not-emprty set of partitions needed to be rebuilt.

This logic bug was introduced with the VAST v3.0.2 release, and can be tracked down to the confusing meaning of the `--all` flag. In addition, this also changes the meaning of the option to do what it should, namely rebuild _all_ partitions instead of just _outdated_ ones. The `--undersized` flag remains unchanged and still adds all undersized partitions to the list of partitions considered for rebuilding, regardless of whether they're outdated or not.

### DIFF
--- a/changelog/v3.0.4/bug-fixes/3083.md
+++ b/changelog/v3.0.4/bug-fixes/3083.md
@@ -1,0 +1,5 @@
+Automatic rebuilds now correctly consider only outdated or undersized
+partitions.
+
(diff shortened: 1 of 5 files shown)


lava (approved):
I didn't test it locally, but the fix looks correct to me.

PR #3086, opened 2023-04-18
Mark some CAF types as nodiscard

dominiklohmann:
This works around actor-framework/actor-framework#1401 with a little CMake trickery.


jachris (approved):
This PR makes the CI fail, I assume that this is PCH related.

~~~
/usr/include/c++/12/bits/char_traits.h:431:56: error: 'void* __builtin_memcpy(void*, const void*, long unsigned int)' specified bound between 9223372036854775810 and 18446744073709551615 exceeds maximum object size 9223372036854775807 [-Werror=stringop-overflow=]
  431 |         return static_cast<char_type*>(__builtin_memcpy(__s1, __s2, __n));
      |                                        ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
~~~

If we cannot easily fix this, I would propose that we do not use PCH, at least for this PR. The `[[nodiscard]]` declarations should still work, because they don't have to be the first declaration (see `[dcl.attr.nodiscard]`).

Also, the CI detected some more places where the `[[nodiscard]]` is violated, would be nice if you could fix those. Thanks!

PR #3088, opened 2023-04-19
Implement `file` loader

Dakostu:
This PR implements a `file` loader with an optional `--timeout` parameter that forwards regular and UDS file input to the next pipeline operator.

This loader also replaces the `stdin` loader.


jachris:
You might want to consider overriding `operator_base::infer_type_impl` for `load_operator` as part of this PR. This is because the default implementation of `infer_type_impl` will instantiate the pipeline with a dummy generator, which calls `vast::plugins::file::plugin::make_loader` in this case. This will already try to open the file which can fail. However, perhaps this part of the pipeline is to be executed on another node where it would succeed.

For the loader, this effect is rather subtle, but for the saver you most definitely want to override it.

libvast/builtins/connectors/file.cpp
```diff
@@ -162,6 +161,9 @@ class plugin : public virtual loader_plugin {
             }
           }
         }
+        if (fd != STDIN_FILENO) {
+          ::close(fd);
```
Dakostu, line 165: I'm not so sure about this dodgy file descriptor closing.

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,262 @@
...
+          read_timeout = std::chrono::duration_cast<std::chrono::milliseconds>(
+            *parsed_duration);
+          ++i;
+        } else {
+          return caf::make_error(ec::syntax_error,
+                                 fmt::format("could not parse duration: {}",
+                                             args[i + 1]));
+        }
+      } else if (arg == "--uds") {
+        file_type = std::filesystem::file_type::socket;
```
jachris, line 58: Why can't we do this automatically by inspecting `std::filesystem::status`?

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,262 @@
...
+public:
+  static constexpr inline auto max_chunk_size = size_t{16384};
+  static constexpr inline auto stdin_path = "-";
+
+  auto
+  make_loader(std::span<std::string const> args, operator_control_plane&) const
+    -> caf::expected<generator<chunk_ptr>> override {
+    auto read_timeout = read_timeout_;
+    auto path = std::string{};
+    auto file_type = std::filesystem::file_type::regular;
```
jachris, line 38: This is not really the file type (as in `std::filesystem::file_type`, but rather a `auto is_socket = false` (this would also remove the unreachable `default` case).

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,262 @@
...
+            *parsed_duration);
+          ++i;
+        } else {
+          return caf::make_error(ec::syntax_error,
+                                 fmt::format("could not parse duration: {}",
+                                             args[i + 1]));
+        }
+      } else if (arg == "--uds") {
+        file_type = std::filesystem::file_type::socket;
+      } else if (arg == "-" || arg == "stdin") {
```
jachris, line 59: What do you think about dropping the `stdin` completely here? Most tools I know only handle `-`.

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,262 @@
...
+                                 "cannot use STDIN as UNIX "
+                                 "domain socket");
+        }
+        auto uds = detail::unix_domain_socket::connect(path);
+        if (!uds) {
+          return caf::make_error(ec::filesystem_error,
+                                 "failed to connect to UNIX domain socket at",
+                                 path);
+        }
+        fd = uds.recv_fd(); // Blocks!
```
jachris, line 122: I'm not familiar with the exact semantics around this. Is the call to `recv_fd` problematic here?

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,262 @@
...
+        break;
+      }
+      case std::filesystem::file_type::regular: {
+        if (path != stdin_path) {
+          fd = ::open(path.c_str(), std::ios_base::binary | std::ios_base::in);
```
jachris, line 127: Is `::open` guaranteed to be compatible with `std::ios_base::` flags?

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+  auto default_parser(std::span<std::string const> args) const
+    -> std::pair<std::string, std::vector<std::string>> override {
+    for (std::string_view arg : args) {
+      if (arg == "-" || arg == "stdin") {
+        break;
+      }
+      return {detail::file_path_to_parser(arg), {}};
+    }
+    return {"json", {}};
```
jachris, line 184: This always only looks at the first argument.

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,268 @@
...
+              }
+            }
+          }
+          if (eof_reached or current_data.size() == max_chunk_size) {
+            auto chunk = chunk::make(std::exchange(current_data, {}));
+            co_yield std::move(chunk);
+            if (following or not eof_reached) {
+              current_data.reserve(max_chunk_size);
+            }
+          }
```
jachris, line 166: What exactly should happen if a timeout occurs? If I read this implementation correctly, we emit an empty chunk if we did not collect any data yet, and otherwise we just go back and try reading again. That's probably not what we want to do, or is it?

libvast/builtins/connectors/file.cpp
jachris: I think this has to be RAII (because upstream generators can just be dropped). Any other reasons for why this might be dodgy?

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,268 @@
...
+          return caf::make_error(ec::filesystem_error,
+                                 "failed to connect to UNIX domain socket at",
+                                 path);
+        }
+        fd = uds.recv_fd(); // Blocks!
+        break;
+      }
+      case std::filesystem::file_type::regular: {
+        if (path != stdin_path) {
+          fd = ::open(path.c_str(), std::ios_base::binary | std::ios_base::in);
```
jachris, line 127: This is pretty much TOCTOU. Can we just try to open it without doing the access checks?

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,268 @@
...
+    const auto* read_timeout_entry
+      = caf::get_if<std::string>(&import_settings->at("read-timeout"));
+    if (!read_timeout_entry) {
+      return caf::none;
+    }
+    if (auto timeout_duration = to<vast::duration>(*read_timeout_entry)) {
+      read_timeout_ = std::chrono::duration_cast<std::chrono::milliseconds>(
+        *timeout_duration);
+    }
+    return caf::none;
```
jachris, line 210: I think this comes from `stdin.cpp`, but you could rewrite this with the new `get_if` functions.

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,268 @@
...
+private:
+  std::chrono::milliseconds read_timeout_{vast::defaults::import::read_timeout};
+};
+
+} // namespace vast::plugins::file
+
+namespace vast::plugins::stdin_ {
+
+class plugin : public virtual vast::plugins::file::plugin {
+public:
```
jachris, line 226: This is an ODR violation. We should not reopen plugin namespaces from other plugins. But I think we can now remove the `stdin.cpp` plugin, can't we?

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  auto default_parser([[maybe_unused]] std::span<std::string const> args) const
+    -> std::pair<std::string, std::vector<std::string>> override {
+    return {"json", {}};
+  }
+
+  auto name() const -> std::string override {
+    return "stdin";
+  }
```
jachris, line 242: Related: We would now have two plugins with the same name.

jachris (commented):
Nice, we are now very close to having a file loader :tada: 

Sorry for writing so many comments – most of them are nits or questions. I would like to take one final look after things are clarified.

libvast/builtins/connectors/file.cpp
jachris: I think this has to be RAII (because upstream generators can just be dropped). Any other reasons for why this might be dodgy?

libvast/builtins/connectors/file.cpp
Dakostu: You're right, but that's exactly how we specified the timeout should work, as a sort of "keepalive" for the next pipeline operator. Example: Receiving pipeline input data packets from remote connections. 

Has this specification been changed?

libvast/builtins/connectors/file.cpp
Dakostu: It would create a stronger semantic bridge from `from file` to `from stdin`.
But ultimately I have no strong opinion on this.

libvast/builtins/connectors/file.cpp
Dakostu: We specified `--uds` options for explicit `file_type`/`is_socket` setting. 
I can implement an automatic file type inspection that can get overwritten by the `--uds` option.

dominiklohmann: Inferring this automatically makes a lot of sense to me, no need to override it then or to parse the option in the first place.

jachris: > Inferring this automatically makes a lot of sense to me, no need to override it then or to parse the option in the first place.

However, the `--uds` option that @Dakostu mentioned is also used in the `from uds` syntax. We would have to remove that, or use some other mechanism to communicate that that follows must be a socket.

dominiklohmann: Can't we just remove the UDS loader in its entirety? We'd only need to distinguish in the file saver, where an option to write to a socket makes a lot of sense to have.

Dakostu: Removing the UDS loader should be no problem.

libvast/builtins/connectors/file.cpp
Dakostu: `recv_fd()` blocks until it receives the file descriptor from the socket.
I can maybe add a polling with timeout.

libvast/builtins/connectors/file.cpp
Dakostu: I removed `stdin.cpp` in https://github.com/tenzir/vast/pull/3088/commits/305e4c863f71b98e8bc9cba3db8e7da98eeb11b2

libvast/builtins/connectors/file.cpp
Dakostu: Simplified in https://github.com/tenzir/vast/pull/3088/commits/5a36784f4555eefec19eb93ce382a377c8dbd097

libvast/test/loader_plugin.cpp
```diff
@@ -140,4 +142,30 @@ TEST(stdin loader - one complete chunk) {
...
+  loader_plugin = vast::plugins::find<vast::loader_plugin>("file");
+  auto args = std::vector<std::string>{"no-file-oops"};
+  REQUIRE(loader_plugin);
+  REQUIRE_ERROR(loader_plugin->make_loader(args, control_plane));
+}
+
+TEST(file loader - unreadable file) {
+  auto current_epoch = std::time(nullptr);
+  auto unique_temp_file = fmt::format(
+    VAST_TEST_PATH "artifacts/inputs/read_restricted_{}.json", current_epoch);
```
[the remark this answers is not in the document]
Dakostu: So something like `std::filesystem::temp_directory_path()` in https://github.com/tenzir/vast/pull/3088/commits/4fb718330bfc0415e17c27fd599bbbe5159780c5??

libvast/src/detail/file_path_to_parser.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include <filesystem>
+#include <string>
+
+namespace vast::detail {
+
+constexpr inline auto fallback_parser = "json";
+
+constexpr inline auto extension_to_parser_list
+  = std::array<std::pair<std::string, std::string>, 2>{
+    {{"eve.json", "suricata"}, {".ndjson", "json"}}};
```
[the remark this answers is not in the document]
Dakostu: The CI reveals this to be a Mac Thing. I fixed the qualifiers in https://github.com/tenzir/vast/pull/3088/commits/c31f9daa3e9d2f85a3b71f4d0130fefa19129b02, hopefully.

libvast/builtins/connectors/file.cpp
Dakostu: All checks have been removed (except for `is_socket`) in https://github.com/tenzir/vast/pull/3088/commits/a3b046c402783025a72fa735ac583f17f6598ba3

libvast/builtins/connectors/file.cpp
Dakostu: Yes, this would make things simpler. I replaced the `file_type` with `is_socket` in https://github.com/tenzir/vast/pull/3088/commits/a3b046c402783025a72fa735ac583f17f6598ba3

libvast/builtins/connectors/file.cpp
Dakostu: I removed it in https://github.com/tenzir/vast/pull/3088/commits/a3b046c402783025a72fa735ac583f17f6598ba3

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,262 @@
...
+                                   "specified; currently only one "
+                                   "path per read operator call is valid"));
+        }
+        auto permissions = status.permissions();
+        auto readable = ((permissions & std::filesystem::perms::owner_read)
+                           != std::filesystem::perms::none
+                         && (permissions & std::filesystem::perms::group_read)
+                              != std::filesystem::perms::none
+                         && (permissions & std::filesystem::perms::others_read)
+                              != std::filesystem::perms::none);
```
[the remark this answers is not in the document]
Dakostu: Checks have been removed in https://github.com/tenzir/vast/pull/3088/commits/a3b046c402783025a72fa735ac583f17f6598ba3

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,262 @@
...
+        path = stdin_path;
+      } else if (arg == "-f" || arg == "--follow") {
+        following = true;
+      } else if (not arg.starts_with("-")) {
+        std::error_code err{};
+        auto status = std::filesystem::status(arg, err);
+        if (err) {
+          return caf::make_error(ec::parse_error,
+                                 fmt::format("could not parse file path {}: {}",
+                                             arg, err));
```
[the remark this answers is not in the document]
Dakostu: The message is more generic in https://github.com/tenzir/vast/pull/3088/commits/a3b046c402783025a72fa735ac583f17f6598ba3

libvast/test/loader_plugin.cpp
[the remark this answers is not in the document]
jachris: Yes, and you should be able to do `auto path = std::filesystem::temp_directory_path() / "read_restricted.json"`. But TBH, I'm not completely confident what's best practice here.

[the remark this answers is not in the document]
Dakostu: Yes, I should, but I would like the epoch as an unique file name element as to not overwrite anything. 
I will remove the `artifacts/inputs` subpath however.

libvast/builtins/connectors/file.cpp
jachris: I tried to understand what happens here, and either I don't get it, or this is not all what we want to do. What about `fd = uds.fd`?

libvast/builtins/connectors/file.cpp
Dakostu: In a Slack discussion, we agreed to make the generator always return a chunk upon hitting a timeout. https://github.com/tenzir/vast/pull/3088/commits/4158b5d2f79ca238fd72a558029d53a5f0cd2bae

libvast/builtins/connectors/file.cpp
Dakostu: I studied the `uds` code a bit, and I think `uds.fd` should suffice indeed. https://github.com/tenzir/vast/pull/3088/commits/dea1b9f7a58896bc73b8ca5a0a124eeae5629cc2

libvast/test/loader_plugin.cpp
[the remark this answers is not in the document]
Dakostu: Updated in https://github.com/tenzir/vast/pull/3088/commits/3ad3012e9da395edbc1f8b3450b6d1f86275c58f

libvast/builtins/connectors/file.cpp
Dakostu: I agree, I don't want to manually handle file descriptors at all, at best.

I could create a struct with custom destructor or a `unique_ptr` with custom destruction function that both wrap the file descriptor and then `close` it upon destruction.

But most of all I'm a bit worried about creation/destruction time. Right now the file descriptor gets created before the generator gets created, but then destroyed inside the generator. This doesn't sit right with me, because right now I would have to `move` that wrapper object into the `std::invoke` call for the generator. Does that make sense to you?

jachris: Yes, I see what you mean. You could either move the file opening into the beginning of the generator itself (this would also make `load_operator::infer_type_impl` unnecessary, but we can just keep it), but moving the RAII wrapper into the generator seems good to me.

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,203 @@
...
+    if (is_socket) {
+      if (path == stdin_path) {
+        return caf::make_error(ec::filesystem_error, "cannot use STDIN as UNIX "
+                                                     "domain socket");
+      }
+      auto uds = detail::unix_domain_socket::connect(path);
+      if (!uds) {
+        return caf::make_error(
+          ec::filesystem_error,
+          fmt::format("Unable to connect to UNIX domain socket at {}:", path));
```
jachris, line 91:
```suggestion
          fmt::format("unable to connect to UNIX domain socket at {}", path));
```

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,203 @@
...
+              break;
+            }
+            auto chunk = chunk::make(std::exchange(current_data, {}));
+            co_yield std::move(chunk);
+            if (eof_reached and not following) {
+              break;
+            }
+            if (not eof_reached or following) {
+              current_data.reserve(max_chunk_size);
+            }
```
jachris, line 130:
```suggestion
            current_data.reserve(max_chunk_size);
```
This condition is the negation of the previous one, and therefore always true.

libvast/builtins/connectors/file.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+  auto default_parser(std::span<std::string const> args) const
+    -> std::pair<std::string, std::vector<std::string>> override {
+    for (std::string_view arg : args) {
+      if (arg == "-" || arg == "stdin") {
+        break;
+      }
+      if (!arg.starts_with("-")) {
+        return {detail::file_path_to_parser(arg), {}};
+      }
```
jachris, line 149: This does not work for `load stdin --timeout 1s foo.csv`, right?

libvast/builtins/connectors/file.cpp
dominiklohmann: :+1: to removing it here, if we're referring to a file it seems rather unusual. Also, since plugins names are case-insensitive this would imply that `from STDIN` works, but `from file STDIN` does not.

libvast/builtins/connectors/file.cpp
Dakostu: I already looked into `file.hpp` and `posix.hpp` - `file` seems like a good abstraction, but right now only for regular files. `posix` has `uds` connection methods that are already used. The main issue that I see here is that `fdinbuf` in the generator only needs and uses a file descriptor to read data - basically no abstraction. Or an abstraction of all abstractions?

jachris (approved):
Great. I think this will be complete after the open points have been addressed :tada:

libvast/builtins/connectors/file.cpp
Dakostu: It shouldn't, but it does. I added a missing argument check in https://github.com/tenzir/vast/pull/3088/commits/26c17b32e7ba53a27f5027a6ab918f74b40d215c

libvast/builtins/connectors/file.cpp
Dakostu: Removed condition in https://github.com/tenzir/vast/pull/3088/commits/26c17b32e7ba53a27f5027a6ab918f74b40d215c

libvast/builtins/connectors/file.cpp
Dakostu: Lower-cased the error message in https://github.com/tenzir/vast/pull/3088/commits/26c17b32e7ba53a27f5027a6ab918f74b40d215c

libvast/builtins/connectors/file.cpp
Dakostu: I removed "stdin" as `file` argument in https://github.com/tenzir/vast/pull/3088/commits/26c17b32e7ba53a27f5027a6ab918f74b40d215c

libvast/builtins/connectors/file.cpp
Dakostu: I opted for a `unique_ptr` with custom deleters in https://github.com/tenzir/vast/pull/3088/commits/26c17b32e7ba53a27f5027a6ab918f74b40d215c

libvast/builtins/connectors/file.cpp
jachris: You missed the trailing `:` 😄

Dakostu: Not anymore :sweat_smile:

libvast/builtins/connectors/file.cpp
jachris: Sorry, I meant `from file --timeout 1s foo.csv`. This would try to infer the parser from `1s`, which leads to `[from_read.cpp:29] from/read expanded to 'load file --timeout 1s foo.csv | parse json '`. As a solution, you could refactor the argument parsing into a separate function and parse the arguments again here (and this confirms that our plugin API could really be improved).

libvast/builtins/connectors/file.cpp
jachris: This also calls `close(*fd)` if the file could not be opened but this is fine for now.

Dakostu: Added `fd` check in https://github.com/tenzir/vast/pull/3088/commits/c46fce7221b2e30b5ca1819e94496a60766aa633

libvast/builtins/connectors/file.cpp
Dakostu: I added a simpler loop in https://github.com/tenzir/vast/pull/3088/commits/e0bd40fcdb71cacdd1cc8c363179674085fabe96 -  when the API is improved, this will be changed anyway.

PR #3091, opened 2023-04-20
Remove everything relating to legacy pipelines

dominiklohmann:
We did it! This removes the old pipeline execution model.

This commit removes the (broken) support for ad-hoc import pipeline execution. The import command applied pipelines per slice, which was not correct, and also led to events being miscounted and wrong metrics being reported. The feature will return in the near future via an `import` sink operator, enabling the more flexible `vast exec '… | import'`.

Fixes tenzir/issues#199


jachris (approved):
I love deleting code :)

PR #3092, opened 2023-04-20
Implement a `cast` operator

dominiklohmann:
The semantics of this are rather straightforward: The `cast <schema>` operator takes the name of a known schema, and parses input events to it iff possible.

I developed this primarily because I needed it to be able to test casting to understand how we want to improve upon the adaptive builder, and figured that it'd make for a pretty good operator on its own.

NOTE: This is not ready to be merged; I primarily opened this as a draft PR because I didn't want the work to be lost. There are quite a few missing conversions in casting, so while the operator itself is unlikely to change it is pretty much useless right now because the underlying `cast` function needs more of our attention.


dominiklohmann:
I will bring this back in the future once we're a bit further on the casting side of things.

PR #3093, opened 2023-04-21
Implement the `measure` operator

dominiklohmann:
The `measure [--real-time]` operator emits metrics about the input into the pipeline. The operator handles both events and bytes as input, and emits the metrics as events as described by these schemas, respectively:

```
type vast.metrics.events = record  {
  timestamp: time,
  schema: string,
  schema_id: string,
  events: uint64,
}

type vast.metrics.bytes = record  {
  timestamp: time,
  bytes: uint64,
}
```

The `--real-time` option causes the operator to instantly emit the metrics at the cost of throughput.

```[tasklist]
### Tasks
- [x] Implement the operator
- [x] Agree on schema names for the metrics events
- [x] Write tests
- [x] Document the operator
- [x] Add a changelog entry
```


web/docs/understand/operators/transformations/inspect.md
```diff
@@ -0,0 +1,59 @@
...
+```title="Bytes Metrics"
+type vast.metrics.bytes = record  {
+  timestamp: time,
+  bytes: uint64,
+}
+```
+
+### `--real-time`
+
+Emit metrics immediately rather than buffering at the cost of performance.
```
mavam, line 34: As a user, I don't understand what I'm "paying" for. Can you describe details about the buffering, e.g., in terms of space or time.

mavam: Suggestion for the description:

> Emit metrics immediately with every batch, rather than buffering until the upstream operator stalls, i.e., is idle or waiting for further input.
>
> The `--real-time` option is useful when `inspect` should emit data without latency.

Writing this, I realized that I would probably want `--real-time` to be the default. Have you ever encountered an issue with this in practice?

dominiklohmann: I had a test setup with ~300M results and ran roughly the following with hardcoded timing code around the pipeline:

```
vast export json 'inspect | summarize sum(bytes) by schema'
```

With `--real-time`, the pipeline took over 100x longer because we sent batches of size 1 into the `summarize` operators.

I would argue that it's a very dangerous default, as you usually want to put the metrics into the pipeline because you want to work with them further.

mavam: I see. If most of the workloads are of this kind, then it indeed is safer to opt out of batching.

libvast/builtins/operators/inspect.cpp
```diff
@@ -0,0 +1,154 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include "vast/pipeline.hpp"
```
jachris, line 9: Nitpick: `#include <vast/...>`

web/docs/understand/operators/transformations/inspect.md
```diff
@@ -0,0 +1,59 @@
+# inspect
+
+Replaces the input with incremental metrics describing about the input.
```
jachris, line 3:
```suggestion
Replaces the input with incremental metrics describing the input.
```

jachris, line 3: For me, the word "incremental" is kind of misleading. I don't really have a great suggestion right now, but I'd prefer something like "Yields metrics about each received batch of events" and similarly for the other parts of the documentation.

jachris (approved):
One general comment: You could consider renaming this operator, because `inspect` is not the most intuitive name for its behavior. Perhaps even something like `metrics` (although we'd normally avoid nouns). But you can also keep it as-is.

libvast/builtins/operators/inspect.cpp
dominiklohmann: I think we all just copy `pass.cpp` every time we create a new operator, or so it feels... 😅

dominiklohmann:
> @jachris: You could consider renaming this operator, because inspect is not the most intuitive name for its behavior. Perhaps even something like metrics (although we'd normally avoid nouns). But you can also keep it as-is.

@mavam do you have an idea for a better name? I agree generally that `inspect` is not the ideal name, but I don't know what we should use. Potential candidates that are verbs:
1. inspect
2. measure
3. debug
4. monitor
5. view (similar to the UNIX tool `pv`)
6. metric (which according to Wiktionary can also be used as a verb)

jachris:
TBH, although I absolutely prefer verbs as well, we should not put this requirement above everything else. If there is no adequate verb, we should just use a noun. And we already do this with `from` and `to` (although you might argue that sources and sinks are special). Also, we might want to use `inspect` for something else that is more inspect-y.

mavam:
I agree that `inspect` is probably more well-suited to yield the AST of the pipeline or some other structural information. Ideas for verbs, in order of preference:

1. profile
2. measure
3. observe
4. monitor

dominiklohmann:
Of the suggested alternatives I chose to go with `measure` for now; there was no objection internally.

vast/integration/vast_integration_suite.yaml
```diff
@@ -785,3 +785,27 @@ tests:
...
+        input: data/json/files.log.json.gz
+      - command: exec 'read json | measure --real-time | measure | drop timestamp | write json'
+        input: data/json/files.log.json.gz
+
+  Measure Bytes:
+    tags: [pipelines]
+    steps:
+      - command: exec 'load stdin | measure | drop timestamp | write json'
+        input: data/json/conn.log.json.gz
+      - command: exec 'load stdin | measure | measure | drop timestamp | write json'
```
mavam, line 806: Higher-order metrics. 😜

dominiklohmann: This actually allows for seeing whether `measure --real-time` works, because all entries in `measure --real-time | measure` must show a batch size of 1.

PR #3073, opened 2023-04-14
Create a VAST version that can connect to the cloud

lava:
The main contribution is the additiona of a new `atom::proxy` handler in the node that can be used to access the REST endpoints of a VAST node. Various additional changes are made in connection to that, in particular a new `internal_http_response` class is introduced that allows usage of the REST infrastructure independentlyl of the restinio server.

Also, some infrastructure changes are made to build the closed-source `cloud` plugin as part of the new CE docker image.

PR #3094, opened 2023-04-24
Remove outgoing paths when matcher clients go down

lava:
This fixes two issues in the matcher and matcher supervisor that were found while going over the code in a bughunting session.


contrib/vast-plugins
dominiklohmann, line 1: Please add a changelog entry for this.

dominiklohmann (approved):

contrib/vast-plugins
lava: After #3073 was merged this PR now contains only the comment change, which probably doesn't warrant a changelog entry.

PR #3099, opened 2023-04-25
Refactor the Docker CI Workflow

tobim:
continuation of #2975.

The main value of this PR is a cleanup of the Docker related CI infrastructure. Docker builds are now run in a reusable workflow that is called by either the event-triggered `vast` workflow or a `manual` workflow dispatch.
The reusable workflow itself is parameterized with a JSON object that configures with targets to build, how to tag them, and which container registry each of the targets should be pushed to.

The manual workflow only exposes a subset of the available options to prevent misuse.


lava (approved):
Looks good to me; and we already tested most of the logic (except for the manual trigger as far as I can tell) in #3073

PR #3103, opened 2023-04-26
Add Feather and Parquet parsers and printers

dominiklohmann:
This changes the store plugin to implement the parser and printer plugin protocols implicitly, effectively adding Feather and Parquet parsers and printers.

```[tasklist]
- [x] Implement derived parser
- [x] Implement derived printer
- [x] Write changelog entry
- [x] Write follow-up issues to make reads and writes incremental
- [x] Write follow-up issue to support any kind of Feather or Parquet file
```

### DIFF
--- a/libvast/src/chunk.cpp
+++ b/libvast/src/chunk.cpp
@@ -218,21 +218,22 @@ chunk_ptr chunk::make_empty() noexcept {
 
 caf::expected<chunk_ptr> chunk::mmap(const std::filesystem::path& filename,
                                      size_type size, size_type offset) {
-  // Figure out the file size if not provided.
-  if (size == 0) {
-    std::error_code err{};
-    size = std::filesystem::file_size(filename, err);
-    if (size == static_cast<std::uintmax_t>(-1))
-      return caf::make_error(ec::filesystem_error,
-                             fmt::format("failed to get file size for file"
-                                         "{}: {}",
-                                         filename, err.message()));
-  }
   // Open and memory-map the file.
   const auto fd = ::open(filename.c_str(), O_RDONLY, 0644);
   if (fd == -1)
     return caf::make_error(ec::filesystem_error,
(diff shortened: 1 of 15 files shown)


mavam:
Does this build the entire store in memory first?

dominiklohmann:
> Does this build the entire store in memory first?

Yes, that's a known deficiency of the current store API that we cannot currently fix that easily. When we change the store API in the future we will get the same feature for free in the derived parsers and printers.

Currently, we always warn about this behavior:

```cpp
// TODO: Loading everything into memory here is far from ideal. We should
/ instead load passive stores incrementally. For now we at least warn the
// user that this is experimental.
VAST_WARN("the experimental {} parser does not currently load files "
          "incrementally and may use an excessive amount of memory",
          name);
```

mavam:
Understood. Let's discuss the scheduling of this. It prevents a lot of real-world use cases.

dominiklohmann (commented):
I reviewed @jachris' changes, he reviewed mine, and we agreed that this is ready to be shipped.

jachris (approved):

libvast/src/chunk.cpp
```diff
@@ -218,21 +218,22 @@ chunk_ptr chunk::make_empty() noexcept {
...
-                                         filename, err.message()));
-  }
   // Open and memory-map the file.
   const auto fd = ::open(filename.c_str(), O_RDONLY, 0644);
   if (fd == -1)
     return caf::make_error(ec::filesystem_error,
                            fmt::format("failed to open file {}", filename));
+  // Figure out the file size if not provided.
+  if (size == 0) {
+    struct stat filestat {};
```
mavam, line 228: Out of curiosity, why didn't use use `std::filesystem::file_size`?

dominiklohmann: That just does more work under the hood, so the TOCTOU window is larger.

PR #3104, opened 2023-04-26
Implement `xsv` parser & printer

Dakostu:
This PR adds an `xsv` parser and printer that enables the user to parse and print character-separated values, with specializations for `csv`, `tsv` and `ssv`.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,6 +1,7 @@
 ---
 Checks: >
   *,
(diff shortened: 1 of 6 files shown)


dominiklohmann (approved):
We reviewed this in the office together. Some open TODOs still, but ready to be merged regardless as we have follow-up issues for all outstanding items.

PR #3108, opened 2023-04-27
Add a `flatten` operator

dominiklohmann:
The `flatten` operator (and the corresponding `flatten` function) takes events and flattens them such that the outermost record is the only record and nested lists get flattened into a single list.

Consequently, the operator removes nulls from list, because they cannot be distinguished from absent values in Arrow's data model.


web/docs/understand/operators/transformations/flatten.md
```diff
@@ -0,0 +1,57 @@
...
+    "rrtype": "A",
+    "tx_id": 0,
+    "grouped": {
+      "A": ["tenzir.com", null]
+    }
+  }
+}
+```
+
+The `flatten` operator removes any nesting from the data:
```
mavam, line 40:
```suggestion
The `flatten` operator removes any nesting from the data by joining nested fields with a `.`:
```

It would be nice to tell the user *how* the flattening takes place. Perhaps already even earlier in the description.

libvast/src/table_slice.cpp
```diff
@@ -823,4 +815,255 @@ auto resolve_operand(const table_slice& slice, const operand& op)
...
+  using result_type = std::vector<
+    std::pair<struct record_type::field, std::shared_ptr<arrow::Array>>>;
+  const auto& rt = caf::get<record_type>(field.type);
+  auto struct_array
+    = std::static_pointer_cast<type_to_arrow_array_t<record_type>>(array);
+  const auto next_name_prefix
+    = fmt::format("{}{}{}", name_prefix, field.name, separator);
+  auto transformations = std::vector<indexed_transformation>{};
+  for (size_t i = 0; i < rt.num_fields(); ++i) {
+    transformations.emplace_back(
```
Dakostu, line 919: This leads to a build error on my machine with Clang 15.0.7 - `no matching function for call to 'construct_at'`. `push_back` with an initializer list works.

Using GCC 12.2.1, this line works for me.

Also on lines 1020 & 1052.

Dakostu (dismissed):
An approval for the general gist of the implementation, with some comments, the typical request for more tests and a single point of confusion.

libvast/src/table_slice.cpp
```diff
@@ -823,4 +815,266 @@ auto resolve_operand(const table_slice& slice, const operand& op)
...
+      }
+      return result;
+    },
+    [&](const list_type& lt) -> result_type {
+      return flatten_list(separator, name_prefix, std::move(list_offsets),
+                          {field.name, lt}, list_array->values());
+    },
+    [&](const record_type& rt) -> result_type {
+      return flatten_record(separator, name_prefix, list_offsets,
+                            {field.name, rt}, list_array->values());
```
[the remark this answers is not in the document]
dominiklohmann: `flatten_record` must pass the list offsets to every field, so it cannot move them. `flatten_list` only passed the list offsets to its inner values array, so it can always move them. So this is intended.

Dakostu:
Playing around some more, I noticed that `flatten` transforms `null` values into empty lists if the type is a `list`.
Consider the following example where `answers` and `TTLs` are defined as `list`s: 
image

Is this intended under the "removes nulls from list" edge case, even though there is no null to remove, because the list is missing in the first place?

libvast/src/table_slice.cpp
```diff
@@ -823,4 +815,266 @@ auto resolve_operand(const table_slice& slice, const operand& op)
...
+        },
+        array,
+      },
+    };
+  };
+}
+
+} // namespace
+
+auto flatten(table_slice slice, std::string_view separator) -> flatten_result {
```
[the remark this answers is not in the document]
dominiklohmann: The question boils down to whether we want to treat top-level nulls for list values separately from nested nulls for list values. I am honestly not sure what the right approach for this is.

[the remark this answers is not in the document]
Dakostu: Well, in this case we have an `unset` value vs. a set value that is `empty`. How differently does Zeek handle those types internally? How strong is the difference between these two value types for Zeek users? 

Or even for VAST users: Adding  something like `where answers` to the pipeline above after `flatten`ing would yield additional events that, in terms of the original information they contained, shouldn't be yielded (events where `null` has been transformed to non-null empty values).

If we can be idiomatic, we could say that we could remove non-existent values (e.g. nulls in lists), but not replace non-existent values with existent, but not useful values (e.g. `null` -> empty list).

web/docs/understand/operators/transformations/flatten.md
```diff
@@ -0,0 +1,68 @@
...
+  "src_port": 1141,
+  "dest_ip": "147.32.80.9",
+  "dest_port": 53,
+  "event_type": "dns",
+  "dns.type": "query",
+  "dns.id": 553,
+  "dns.rrname": "irc.freenode.net",
+  "dns.rrtype": "A",
+  "dns.tx_id": 0,
+  "dns.grouped.A": ["tenzir.com"]
```
mavam, line 66: Where did the `null` value go?

web/docs/understand/operators/transformations/flatten.md
```diff
@@ -0,0 +1,68 @@
...
+
+```
+flatten [<seperator>]
+```
+
+## Description
+
+The `flatten` operator removes any nested lists or records by merging lists and
+joining nested records with a separator. Flattening removes null values inside
+lists.
```
mavam, line 15: Why is flattening removing nulls? This changes the shape of the data and may lead to incorrect results, e.g., when getting the length of a list.

libvast/src/table_slice.cpp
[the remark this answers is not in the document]
mavam: Why would we touch a list value that's `null`? Flattening should only change the structure, not the values. 

> Well, in this case we have an `unset` value vs. a set value that is `empty`. How differently does Zeek handle those types internally? How strong is the difference between these two value types for Zeek users?

These are different, in the same way that the empty string is different from null.

web/docs/understand/operators/transformations/flatten.md
dominiklohmann: It necessarily has to. The purpose of the flatten operator is to bring data into a shape that makes it usable for formats or downstream tools that do not support nested structures. This includes all combinations of lists and records.

When you flatten a list of a list of a type into a list of the type, then you must remove all nulls you encounter on the way:

E.g., `[[1, null], null,  [2, 3]]` has to be flattened into `[1, 2, 3]`. Do you see another option?

mavam: I think my main problem is that record and list flattening are done in one shot here. These are quite different use cases. If it'd be possible to toggle them separately, e.g., with `-l` for only lists and `-r` for only records, I think this works.

The idea would be still stepping through lists, e.g., when you have a list of a record. Basically `-r` would only latch onto records, where ever they are.

libvast/src/table_slice.cpp
```diff
@@ -823,4 +815,266 @@ auto resolve_operand(const table_slice& slice, const operand& op)
...
+    std::pair<struct record_type::field, std::shared_ptr<arrow::Array>>>;
+
+auto flatten_list(std::string_view separator, std::string_view name_prefix,
+                  std::vector<std::shared_ptr<arrow::Array>> list_offsets,
+                  struct record_type::field field,
+                  const std::shared_ptr<arrow::Array>& array)
+  -> std::vector<
+    std::pair<struct record_type::field, std::shared_ptr<arrow::Array>>> {
+  using result_type = std::vector<
+    std::pair<struct record_type::field, std::shared_ptr<arrow::Array>>>;
```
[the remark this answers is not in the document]
dominiklohmann: It probably makes sense to have this as a member type of `indexed_transformation`.

web/docs/understand/operators/transformations/flatten.md
dominiklohmann: I'm closing this as it's the same question as in the other thread.

web/docs/understand/operators/transformations/flatten.md
dominiklohmann: I would argue that this should be use-case driven. For now, the only thing we need is a single `flatten` operation that does both, because that's what we want to have for printing csv (and similar) that is guaranteed to be usable by downstream tools.

If we ever have a concrete use case where we need both, then we could separate them out or add an option to flatten only records or only lists, but right now that seems like a lot of extra effort to me that is not properly justified.

libvast/src/table_slice.cpp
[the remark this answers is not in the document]
dominiklohmann: I thought about this over the weekend a lot. Here's the list flattening behavior that I think we may want:

1. We flatten into `null` if the outermost list is `null`.
2. We keep the null values of the innermost list.
3. We remove null values of all lists except for the innermost list.

This would give the following behavior for a `list<list<uint64>>`:

|original<br>`list<list<uint64>>`|flattened<br>`list<uint64>`|
|:-:|:-:|
|`null`|`null`|
|`[null]`|`[]`|
|`[[null]]`|`[null]`|
|`[[null, 1], null, [2, null, 3]`|`[null, 1, 2, null, 3]`|

I'd like to agree on this behavior (or a different one, for that matter) before we make any further changes to this PR. Wdyt @Dakostu @mavam?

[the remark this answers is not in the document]
mavam: Compare this to JavaScript:

```javascript
[null].flat()
// Output: [ null ]
[[null]].flat()
// Output: [ null ]
[[null, 1], null, [2, null, 3]].flat()
// Output: [ null, 1, null, 2, null, 3 ]
```

This makes more sense to me than your proposal that for some to me unclear reason strips null values (which are valid list elements).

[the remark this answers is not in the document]
dominiklohmann: The problem with that is that you're treating nulls differently depending on the nesting:

1. In `[[null: uint64]]`, the `null` is treated like `null: uint64`
2. In `[null: list<uint64>]`, the `null` is treated like `[null]: list<uint64>`.
3. In `null: list<list<uint64>>`, the `null` is treated like `null: list<list<uint64>>`.

I find the difference between (2) and (3) very unintuitive, as in one case the `null` value denotes the absence of a list, and in the other it silently converts into a list with a single `null` element. We should look closer at what other Arrow-based systems do, rather than languages with known-to-be-inconsistent type systems like JS.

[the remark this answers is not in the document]
mavam: To me, the main point still holds: `null` is a regular list element (regardless of its type) and should not be removed by a structural transformation like `flatten`.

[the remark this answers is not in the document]
Dakostu: I skimmed through some of the [projects powered by Arrow](https://arrow.apache.org/powered_by/) and the only flattening I could find that is not Python- , JS- or Rust-based is [FlattenArrays() in Cylon](https://github.com/cylondata/cylon/blob/d9a64204f9237729694df869b448fb67eed1440c/cpp/src/cylon/util/flatten_array.cpp#L289). But that function works slightly differently to our flattening; it basically [flattens multiple same-length arrays into one](https://github.com/cylondata/cylon/blob/30bc9f3b51931adca95886342e184678e6b42d77/cpp/src/cylon/util/flatten_array.hpp#L50), [combining elements](https://github.com/cylondata/cylon/blob/30bc9f3b51931adca95886342e184678e6b42d77/cpp/test/flatten_array_test.cpp#L103) and [adding empty elements in lieu of any `null` value](https://github.com/cylondata/cylon/blob/30bc9f3b51931adca95886342e184678e6b42d77/cpp/src/cylon/util/flatten_array.hpp#L57).

While I get that removing elements from flattening may seem unintuitive: The flattening of a multi-dimensional `list<list<T>` into `list<T>`  - to me - signifies that there shouldn't be any `list<T>` values in the list anymore. This also includes `null` values that denoted the absence of lists. This value that denotes absence would be "flattened" into non-existence, since the `list<T>`, in terms of dimensionality, should not be intelligent enough anymore to know about `list<list<T>>` values.

To cut a long story short: 
* `[[null]]` in `list<list<T>>` = "No  `T` value" -> flattened into `null` in `list<T>`
* `[null]` in `list<list<T>>` = "No list" -> flattened into non-existence in `list<T>`

That mental model makes sense to me, at least in a strongly typed world.

Dakostu:
Note to (maybe) myself: This PR could also implement a `flatten` call for the `zeek-tsv` format, as in [this commit](https://github.com/tenzir/vast/pull/3114/commits/c75cdf084d57a53e8222c9cd43aa103d3781bfe0#diff-722a1fc3a66d8aeaaedc0253886a7446b9db7ff4eb0dac8abc109d650f94ba95R690)

libvast/src/table_slice.cpp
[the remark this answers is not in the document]
dominiklohmann: I do have a strong opinion in the other direction: we should avoid making a design decision that does not interact well with the libraries we are using, so I think we should use the exact behavior of Arrow's list flattening function here to avoid an inconsistent behavior when we call into some Arrow functionality in the future that flattens for us implicitly.

It appears that we are stuck here then in this discussion, so I suggest we take this into a call because this thread isn't an effective use of our time.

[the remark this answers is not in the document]
mavam: If you think that both options are valuable, then make this a configuration option. I have use cases where I *don't* my list cardinality to change during analysis (counting elements is one of them). If you have others where this is desired, it's not that hard to make everyone happy.

If that's not feasible, we need a call :-).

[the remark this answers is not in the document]
dominiklohmann: We discussed this in person and agreed on the following: By default, flatten both lists and records with the Arrow-default semantics. Add an option that causes the operator to only flatten records with the Arrow-default semantics, but not lists.

I will update this PR sometime next week to implement this.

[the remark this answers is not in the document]
Dakostu: For the sake of providing basic flattening functionality, we will ship `flatten` as-is and wait for Alpha test feedback.

libvast/src/table_slice.cpp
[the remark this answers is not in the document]
Dakostu: I implemented `indexed_transformation::result_type` in https://github.com/tenzir/tenzir/pull/3108/commits/eda92596b9620071254d127fce0a5845c809c8b2

libvast/src/table_slice.cpp
```diff
@@ -891,6 +883,263 @@ auto resolve_operand(const table_slice& slice, const operand& op)
...
+  // detect this and warn if in that case.
+  for (const auto& leaf : caf::get<record_type>(slice.schema()).leaves()) {
+    if (const auto* lt = caf::get_if<list_type>(&leaf.field.type)) {
+      if (caf::holds_alternative<record_type>(lt->value_type())) {
+        VAST_WARN("flatten may produce wrong results for records inside lists "
+                  "inside records; this bug will be fixed in a future release");
+        break;
+      }
+    }
+  }
```
dominiklohmann, line 1085: This is no longer the case, I think.

libvast/src/table_slice.cpp
```diff
@@ -891,6 +883,263 @@ auto resolve_operand(const table_slice& slice, const operand& op)
...
+            {
+              fmt::format("{}{}", name_prefix, field.name),
+              type{list_type{field.type}},
+            },
+            arrow::ListArray::FromArrays(*combined_offsets, *array).ValueOrDie(),
+          },
+        };
+      },
+      [&](const list_type&) -> indexed_transformation::result_type {
+        if (opt.ignore_lists) {
```
dominiklohmann, line 1036: I think this does not work for records inside lists inside lists. Consider this:

```json5
// Input
{"foo": [[{"bar": "baz"}]]}

// Without -i
{"foo": "bar": ["baz"]}

// Actual output with -i
{"foo": [[{"bar": "baz"}]]}

// Expected output with -i
{"foo": "bar": [["baz"]]}
```

libvast/src/table_slice.cpp
Dakostu: Do you remember what we used to determine this?

For example, I tried out:
`{"foo": [{"bar": "baz", "huh": 2}]}`

`flatten` results in 
`{"foo.bar": ["baz"], "foo.huh": [2]}`

Wasn't that the "broken" behavior?

dominiklohmann: This file in the repository was broken before: vast/integration/data/suricata/rrdata-eve.json

libvast/src/table_slice.cpp
Dakostu: It "works" insofar that `ignore-lists` ignores lists including their values.
Should `ignore-lists` not ignore record values in lists, but any other scalar and list value?

dominiklohmann: Yes. We could also move this into a follow-up story, but then we should remove the option for now (maybe make that a separate commit so that can be reverted as a starting point).

libvast/src/table_slice.cpp
Dakostu: I checked the file in flattened and unflattened versions. There is no data loss between the two.

I will carefully remove that message then.

libvast/src/table_slice.cpp
Dakostu: Alright, I tried moving the `if (opt.ignore_lists)` around in the recursive call chain but that didn't achieve anything.

I removed `ignore-lists` in https://github.com/tenzir/tenzir/pull/3108/commits/a1298a1668f20052a1500a22289b0be8c3580199.

dominiklohmann:
@dakostu I cannot approve this PR myself as I created it originally, but I did enable auto-merge. If you approve it, it should get merged straight away.

Dakostu (approved):

libvast/src/arrow_table_slice.cpp
```diff
@@ -386,14 +384,15 @@ std::pair<type, std::shared_ptr<arrow::RecordBatch>> transform_columns(
...
   };
   if (transformations.empty())
-    return {schema, batch};
+    return {schema, struct_array};
   auto current = transformations.begin();
   const auto sentinel = transformations.end();
   auto layer = unpacked_layer{
     .fields = {},
-    .arrays = batch->ToStructArray().ValueOrDie()->Flatten().ValueOrDie(),
+    .arrays = struct_array->fields(),
```
dominiklohmann, line 392: This needs to be `struct_array->Flatten()`, otherwise we do not merge the null bitmaps correctly.

Dakostu: Replaced in https://github.com/tenzir/tenzir/pull/3108/commits/335341e871959d3724ea2fdcf49f5368d2ce5bb2

PR #3109, opened 2023-04-27
Migrate the Nix build to a reusable workflow

tobim:
Here we refactor the nix CI job to make it more flexible.
The API is analogous to the reusable workflow for the Docker build, but we still use a matrix for Nix because the static binary build can't share intermediate build products (yet).

The core logic is implemented in a nushell script, which simplifies the implementation thanks to its rich data model.

### DIFF
--- a/.github/workflows/nix-config-base.json
+++ b/.github/workflows/nix-config-base.json
@@ -1,0 +1,27 @@
+{
+  "editions": [
+    {
+      "name": "vast",
+      "static": true,
+      "copy-package": true,
+      "package-stores": [
+        "gcs:tenzir-public-data/static-binary-builds"
+      ],
+      "image-registries": [
+        "ghcr.io",
+        "docker.io"
+      ]
+    },
+    {
+      "name": "vast-ce",
+      "static": true,
+      "copy-package": false,
+      "package-stores": [
+        "gcs:tenzir-vast/static-binary-builds"
+      ],
+      "image-registries": [
+        "622024652768.dkr.ecr.eu-west-1.amazonaws.com"
+      ]
+    }
+  ]
+}
--- a/.github/workflows/nix-config-base.json
+++ b/.github/workflows/nix-config-base.json
@@ -3,7 +3,7 @@
     {
       "name": "vast",
       "static": true,
-      "copy-package": true,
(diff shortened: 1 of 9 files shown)


.github/workflows/vast.yaml
```diff
@@ -215,7 +216,7 @@ jobs:
           run_if_changed changelog "changelog/" "web/" "python/"
           VAST_SOURCES=(cmake/ CMakeLists.txt libvast/ libvast_test/ scripts/ schema/ tools/ vast/ vast.yaml.example version.json)
           run_if_changed vast "${VAST_SOURCES[@]}"
-          run_if_changed_default vast-plugins $run_vast "plugins/" "contrib/vast-plugins/"
+          run_if_changed vast-plugins "plugins/" "contrib/vast-plugins/"
```
tobim, line 219: This seems to be a rebase error, it should not be in this PR.

.github/workflows/nix.yaml
```diff
@@ -0,0 +1,165 @@
...
+      - name: Install Nix
+        uses: cachix/install-nix-action@v21
+        with:
+          github_access_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Setup Cachix
+        uses: cachix/cachix-action@v12
+        with:
+          name: vast
+          signingKey: "${{ secrets.CACHIX_VAST_SIGNING_KEY }}"
+          pushFilter: "vast-ee"
```
tobim, line 102: We should whitelist the open source edition here.

.github/workflows/nix-config-base.json
```diff
@@ -0,0 +1,27 @@
+{
+  "editions": [
+    {
+      "name": "vast",
+      "static": true,
+      "copy-package": true,
```
tobim, line 6: This needs a better name

```suggestion
      "upload-package-to-github": true,
```

.github/workflows/nix.yaml
tobim: Not as easy as first thought, because we'd have to whitelist all the packages we depend on as well. I came up with a better filter expression instead. The action uses `grep -vEe "$pushFilter"` internally.

PR #3110, opened 2023-04-28
Add new CEF parser plugin

jachris:
We can now `parse cef` when using the new pipelines :tada:


changelog/next/features/3110--cef-parser.md
```diff
@@ -0,0 +1 @@
+The `cef` parser allows using the CEF format with the new pipelines.
```
dominiklohmann, line 1:
```suggestion
The `cef` parser allows for using the CEF format with the new pipelines.
```

libvast/builtins/formats/cef.cpp
```diff
@@ -6,6 +6,8 @@
 // SPDX-FileCopyrightText: (c) 2022 The VAST Contributors
 // SPDX-License-Identifier: BSD-3-Clause
 
+#include "vast/to_lines.hpp"
```
dominiklohmann, line 9:
```suggestion
#include <vast/to_lines.hpp>
```

Mostly a nit; can we maybe enforce this with a clang-format rule for the builtins?

changelog/next/features/3110--cef-parser.md
dominiklohmann, line 1: Please add an integration test based on `vast exec` for this.

dominiklohmann (approved):
🚀

libvast/builtins/formats/cef.cpp
jachris: We can add a `.clang-format` file to `libvast/builtins` with `BasedOnStyle: InheritParentConfig`, but I think `clang-format` can only changes the order of includes.

dominiklohmann: There's an in-review patch here that would somewhat fix this, but nothing we can use now: https://reviews.llvm.org/D145843

PR #3115, opened 2023-05-04
Bump vast-plugins to a95e420

tobim:
This includes some bug fixes in the matcher extension.


lava (approved):

contrib/vast-plugins
dominiklohmann, line 1: This should get a changelog entry, right?

tobim: right.

PR #3117, opened 2023-05-04
A collection of minor improvements

tobim:
I accumulated most of these while investigating a few issues locally.

* Don't emit an INFO message for every plugin. I think this is no a lot less valuable compared to when it was introduced because (1) a lot more functionality is now implemented in loadable plugins, so it should be a lot more obvious if the paths are not correct, and (2) the problem that this addresses is less prevalent since we start with `plugins=all` by default.
* Add information about the resolved address to the connection log messages. This helps in situations when address name resolution does produce the assumed result.
* Fix VAST build trees for building tree plugins, i.e. building a plugin in standalone mode against a build tree instead of a vast installation.
* Print the location of VAST in find_package calls to verify that the supplied `VAST_DIR` worked.


dominiklohmann (approved):
These all make sense to me.

PR #3120, opened 2023-05-04
Write generated file names for directory saver to stdout

dominiklohmann:
This allows combining the directory saver with downstream UNIX tools, e.g., `vast exec '... | write ssv to directory foo' | xargs awk ...`.

### DIFF
--- a/libvast/builtins/connectors/directory.cpp
+++ b/libvast/builtins/connectors/directory.cpp
@@ -50,8 +50,22 @@ class plugin : public virtual saver_plugin {
         / fmt::format("{}.{}.{}", info.input_schema.name(),
                       info.input_schema.make_fingerprint(), info.format);
     auto new_args = std::vector<std::string>{file_path.string()};
-    // TODO: handle overwrite semantics
-    return saver_plugin_->make_saver(new_args, std::move(info), ctrl);
+    auto file_saver
+      = saver_plugin_->make_saver(new_args, std::move(info), ctrl);
+    if (not file_saver)
+      return std::move(file_saver.error());
+    auto guard = caf::detail::make_scope_guard([=] {
+      // We also print this when the operator fails at runtime, but then again
+      // this also means that we did create the file, so that's probably alright.
+      fmt::print(stdout, "{}\n", file_path.string());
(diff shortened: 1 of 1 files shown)


libvast/builtins/connectors/directory.cpp
```diff
@@ -51,7 +51,19 @@ class plugin : public virtual saver_plugin {
...
     auto new_args = std::vector<std::string>{file_path.string()};
     // TODO: handle overwrite semantics
-    return saver_plugin_->make_saver(new_args, std::move(info), ctrl);
+    auto file_saver
+      = saver_plugin_->make_saver(new_args, std::move(info), ctrl);
+    auto guard = caf::detail::make_scope_guard([=] {
+      fmt::print(stdout, "{}\n", file_path.string());
+    });
+    if (not file_saver)
+      return std::move(file_saver.error());
```
jachris, line 60: I think the error check should come before creating the guard.

libvast/builtins/connectors/directory.cpp
```diff
@@ -51,7 +51,19 @@ class plugin : public virtual saver_plugin {
                       info.input_schema.make_fingerprint(), info.format);
     auto new_args = std::vector<std::string>{file_path.string()};
     // TODO: handle overwrite semantics
-    return saver_plugin_->make_saver(new_args, std::move(info), ctrl);
+    auto file_saver
+      = saver_plugin_->make_saver(new_args, std::move(info), ctrl);
+    auto guard = caf::detail::make_scope_guard([=] {
+      fmt::print(stdout, "{}\n", file_path.string());
```
jachris, line 57: This unconditional printing means that we also print the path if the pipeline is aborted. But perhaps that's fine.

jachris (approved):

libvast/builtins/connectors/directory.cpp
dominiklohmann: We still create the file in that case, right?

dominiklohmann: I added a note to the code and left the behavior as-is.

PR #3121, opened 2023-05-05
Support new 'opaque' attribute in validate()

lava:
This attribute serves as a stopping point for the validation, saying that a given config field may contain arbitrary data.

One example is the `body` in a JSON request, which may contain an arbitrary payload.

### DIFF
--- a/libvast/src/validate.cpp
+++ b/libvast/src/validate.cpp
@@ -16,21 +16,35 @@ namespace vast {
 
 namespace {
 
-caf::error validate_(const vast::data& data, const vast::type& type,
-                     const enum vast::validate mode,
-                     const std::string& prefix = "", size_t depth = 0) {
+auto validate_opaque_(const vast::data& data) -> caf::error {
+  return caf::visit(vast::detail::overload{
+                      [](const record&) -> caf::error {
+                        return caf::error{};
+                      },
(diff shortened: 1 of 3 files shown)


libvast/src/validate.cpp
```diff
@@ -16,27 +16,41 @@ namespace vast {
...
-                     const std::string& prefix = "", size_t depth = 0) {
+auto validate_opaque_(const vast::data& data) -> caf::error {
+  return caf::visit(vast::detail::overload{
+                      [](const record&) -> caf::error {
+                        return caf::error{};
+                      },
+                      [](auto const&) -> caf::error {
+                        return caf::make_error(ec::invalid_argument,
+                                               "only records may have 'opaque' "
+                                               "attribute");
```
tobim, line 27: This essentially means that the "opaque" attribute may not be repurposed, which I guess is a good thing. But I suppose that means we need to document it as a reserved attribute somewhere.

lava: It's documented in `validate.hpp`. (and it's only 'reserved' with regards to this validate function, users should still be able to use freely)

tobim (approved):

lava:
To avoid being blocked on CI twice, we included the commits from this branch in #3073 instead and are closing this.

PR #3122, opened 2023-05-05
Add a new 'enabled()' method to plugin API

lava:
This PR adds a new `enabled()` virtual method to  the plugin base class, which allows plugins to decide whether or not they want to be initialized at all based on their configuration.

The default implementation checks for the presence of an `enabled` field in the plugin config, and defaults to `true` if that isn't found.

### DIFF
--- a/libvast/src/plugin.cpp
+++ b/libvast/src/plugin.cpp
@@ -357,7 +357,13 @@ const std::vector<std::filesystem::path>& loaded_config_files() {
 
 bool plugin::enabled(const record&, const record& plugin_config) const {
   auto default_value = true;
-  return get_or(plugin_config, "enabled", default_value);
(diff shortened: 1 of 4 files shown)


libvast/src/plugin.cpp
```diff
@@ -336,6 +353,19 @@ const std::vector<std::filesystem::path>& loaded_config_files() {
 
 } // namespace plugins
 
+// -- plugin base class -------------------------------------------------------
+
+bool plugin::enabled(const record&, const record& plugin_config) const {
+  auto default_value = true;
+  return get_or(plugin_config, "enabled", default_value);
```
dominiklohmann, line 360: This should probably be `try_get_or`, so that you get an error if the value is set to something other than a boolean.

tobim (approved):

PR #3126, opened 2023-05-05
Improve file saver

jachris:
This PR improves the file saver connector:
- Add home path expansion for paths like `~/foo.json`
- Use a more robust output stream type (and deprecate the old one that is somewhat broken)
- Buffer writes to to the filesystem
- Create directories to output file if they don't exist
- Call `ctrl.abort()` if closing the output file fails instead of silently discarding the error


dominiklohmann (approved):
Thanks for taking care of this. This seems to work in practice.

PR #3128, opened 2023-05-07
Implement the `import` operator

dominiklohmann:
The `import` operator is a remote event sink for a pipeline that imports events into VAST.

This current implementation is very basic, and simply forwards events to the existing `importer` singleton component in the VAST node.

In the future, we should change the `import` operator to directly create passive partitions, and to send then to the catalog directly, which will enable better fault tolerance. That, however, is a change we can make in the future, as VAST's internal APIs are not well suited for that at the current point in time.

```[tasklist]
### Tasks
- [x] Implement a simple version of the import operator.
- [x] Write follow-up issues to re-implement the operator to bypass the importer, index, and active partition actors, and instead directly communicate with the catalog.
- [x] Write documentation.
- [x] Test the operator end-to-end in integration tests.
```


libvast/src/system/execution_node.cpp
```diff
@@ -34,10 +34,14 @@ class actor_control_plane : public operator_control_plane {
     : self_{self} {
   }
 
-  auto self() noexcept -> caf::scheduled_actor& override {
+  auto self() noexcept -> system::execution_node_actor::base& override {
     return self_;
   }
 
+  auto node() noexcept -> system::node_actor override {
```
dominiklohmann, line 41: This should probably return an `expected` instead of just returning a default-initialized node actor for local actors. While the latter may be okay, there are just too many CAF APIs that do not gracefully handle nullptr destination handles.

libvast/builtins/operators/import.cpp
```diff
@@ -0,0 +1,154 @@
...
+  import_source_driver(generator<table_slice>& input, uint64_t& num_events,
+                       operator_control_plane& ctrl)
+    : input_{input}, num_events_{num_events}, ctrl_{ctrl} {
+  }
+
+  void pull(caf::downstream<table_slice>& out, size_t num) override {
+    auto it = input_.unsafe_current();
+    for (size_t i = 0; i < num; ++i) {
+      auto next = std::move(*++it);
+      if (next.rows() == 0) {
```
jachris, line 40: What if `++it == input_.end()`?

jachris (approved):
:tada:

libvast/builtins/operators/import.cpp
dominiklohmann: Good point; I think I never ran into this because num is usually 1, and then we check it after every iteration. I'll add a check to return early in that case.

dominiklohmann: Fixed in db6a764ba3

PR #3129, opened 2023-05-08
Extend `file_path_to_parser` check to also include full file path checks

Dakostu:
This PR fixes the case where `eve.json` file path checks may output the incorrect parser for all files ending with `eve.json`.

### DIFF
--- a/libvast/src/detail/file_path_to_parser.cpp
+++ b/libvast/src/detail/file_path_to_parser.cpp
@@ -9,20 +9,26 @@
 #include "vast/detail/file_path_to_parser.hpp"
 
 #include <array>
-#include <filesystem>
-#include <string>
 
 namespace vast::detail {
 
 constexpr inline auto fallback_parser = "json";
 
-const auto extension_to_parser_list
-  = std::array<std::pair<std::string, std::string>, 2>{
-    {{"eve.json", "suricata"}, {".ndjson", "json"}}};
+const auto filename_to_parser_list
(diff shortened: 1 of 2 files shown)


libvast/src/detail/file_path_to_parser.cpp
```diff
@@ -16,11 +16,19 @@ namespace vast::detail {
...
+const auto filename_to_parser_list
+  = std::array<std::pair<std::string, std::string>, 1>{
+    {{"eve.json", "suricata"}}};
+
 const auto extension_to_parser_list
-  = std::array<std::pair<std::string, std::string>, 2>{
-    {{"eve.json", "suricata"}, {".ndjson", "json"}}};
+  = std::array<std::pair<std::string, std::string>, 1>{{{".ndjson", "json"}}};
 
 auto file_path_to_parser(std::string_view path) -> std::string {
```
dominiklohmann, line 26: Semantically, shouldn't this function take a `const std::filesystem::path&`?

Dakostu: The `string_view` parameter was a carry-over from the `args` vector, but `const std::filesystem::path&` works better. Changed in https://github.com/tenzir/vast/pull/3129/commits/c190ded73fe29d25a3a41f181966cb72a3ec27fe

dominiklohmann (approved):
I think this works better. Thanks for fixing this so quickly! 🚀

PR #3130, opened 2023-05-08
Enable transparent decompression for single file inputs

tobim:
This PR replaces the IO path in the `file` loader with API functions provided by libarchive. This allows command line invocations like
```
vast -q exec 'from - read json | write json to stdout' < vast/integration/data/json/conn.log.json.gz
```
The other file loading variations work accordingly.

This currently comes at the expense of the `timeout` and `following` options, which need to be brought back in a libarchive specific way.
Doing so may be possible with a custom implementation for `archive_read_set_read_callback`, but that needs to be verified.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -86,6 +86,7 @@ RUN apt-get update && \
     apt-get -y --no-install-recommends install \
       ca-certificates \
       gnupg2 \
+      libarchive13 \
       libasan6 \
       libc++1 \
       libc++abi1 \
--- a/libvast/CMakeLists.txt
+++ b/libvast/CMakeLists.txt
@@ -349,6 +349,12 @@ else ()
 endif ()
 target_link_libraries(libvast PRIVATE FastFloat::fast_float)
 
+# Link against LibArchive.
+find_package(LibArchive REQUIRED)
+#string(APPEND VAST_FIND_DEPENDENCY_LIST "\nfind_package(LibArchive REQUIRED)")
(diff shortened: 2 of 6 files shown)


tobim:
> This currently comes at the expense of the timeout and following options, which need to be brought back in a libarchive specific way.

I looked into this and learned that `timeout` isn't possible. While you can provide your own IO callbacks there is simply no way to signal "call me later" to the caller. Once we return 0 bytes the library will switch to the finalization routine.

My intuition is to remove `timeout` and avoid buffering anything in between calls to `read` in the operator.

dominiklohmann:
> My intuition is to remove `timeout` and avoid buffering anything in between calls to `read` in the operator.

Doesn't sound like a good idea to me because this will introduce hangs on shutdown, and also increase the latency when receiving very small packets over a network. We have the read timeout for a very good reason.

Did you consider making this feature explicit instead as a bytes to bytes transformation operator called `decompress`?

tobim:
> this will introduce hangs on shutdown

In what situation? An operator that does IO should be in it's own thread, and `read` is cancellable, so the termination code path can take care of it.

> and also increase the latency when receiving very small packets over a network

I don't understand, can you elaborate?

> Did you consider making this feature explicit instead as a bytes to bytes transformation operator called `decompress`?

I did, but that makes for a worse UX. Why should we force the user to provide information that we can infer?

tobim:
This is now obsolete. The functionality has been implemented with the explicit `decompress` operator instead #3443.

PR #3134, opened 2023-05-09
Prettify INFO log when connecting to a node

dominiklohmann:


libvast/src/system/connector.cpp
```diff
@@ -140,12 +140,14 @@ std::string format_time(caf::timespan timespan) {
...
+  const auto resolved_host
+    = caf::io::network::interfaces::native_address(request.host)->first;
+  VAST_INFO("client faild to connect to remote node {}:{}{}; attempting to "
             "reconnect in {} (remaining time: {})",
-            request.host,
-            caf::io::network::interfaces::native_address(request.host)->first,
-            request.port, format_time(retry_delay),
-            format_time(remaining_time));
+            request.host, request.port,
+            resolved_host != request.host ? fmt::format(" ({})", resolved_host)
```
tobim, line 148: Small nit: we should probably `tolower()` the `request.host`, just in case an IPv6 address with `A-F` is submitted.

tobim (approved):
The change looks correct, did not bother to test locally.

libvast/src/system/connector.cpp
dominiklohmann: Implemented in 73d31a1073

PR #3137, opened 2023-05-10
Align endpoints between regular and slim Docker images

dominiklohmann:
These now both use `0.0.0.0:<default-port>` instead of `127.0.0.1:<default-port>`, unless the user explicitly sets the `VAST_ENDPOINT` environment variable. Previously, only the `tenzir/vast-slim` images did this.


mavam (approved):
Makes sense.

tobim (approved):

PR #3138, opened 2023-05-10
Introduce an experimental `sigma` operator

dominiklohmann:
The `sigma` operator takes a path to a directory or a single *.yaml or *.yml file, takes all input events, and applies the sigma rule. The resulting events contain matches for the Sigma rule and the rule itself.

### DIFF
--- a/plugins/sigma/CMakeLists.txt
+++ b/plugins/sigma/CMakeLists.txt
@@ -1,0 +1,23 @@
+cmake_minimum_required(VERSION 3.19...3.24 FATAL_ERROR)
+
+project(
+  sigma
+  DESCRIPTION "Sigma query language plugin for Tenzir"
+  LANGUAGES CXX)
+
+include(CTest)
+
+file(GLOB_RECURSE sigma_sources CONFIGURE_DEPENDS
+     "${CMAKE_CURRENT_SOURCE_DIR}/src/*.cpp"
+     "${CMAKE_CURRENT_SOURCE_DIR}/include/*.hpp")
+
+file(GLOB_RECURSE sigma_tests CONFIGURE_DEPENDS
+     "${CMAKE_CURRENT_SOURCE_DIR}/tests/*.cpp")
+
+find_package(Tenzir REQUIRED)
+TenzirRegisterPlugin(
+  TARGET sigma
+  ENTRYPOINT src/plugin.cpp
+  SOURCES ${sigma_sources}
+  TEST_SOURCES ${sigma_tests}
+  INCLUDE_DIRECTORIES include)
--- a/plugins/sigma/src/parse.cpp
+++ b/plugins/sigma/src/parse.cpp
@@ -221,9 +221,21 @@ transform_sigma_string(std::string_view str, std::string_view fmt) {
       case '}':
       case '^':
       case '$':
-      case '\\':
         rx += '\\';
         rx += c;
+        break;
+      case '\\':
+        if (f != l and (*f == '?' or *f == '*' or *f == '\\')) {
+          // Edge-case: The user intended to escape the glob character.
+          rx += '\\';
+          rx += *f++;
+          break;
+        }
+        rx += "\\\\";
+        break;
+      default:
+        rx += c;
+        break;
     }
   }
   auto result = pattern::make(rx, {.case_insensitive = true});
@@ -233,8 +245,8 @@ transform_sigma_string(std::string_view str, std::string_view fmt) {
   if (fmt.empty()) {
     return result;
   }
-  return transform_sigma_string(
-    fmt::format(TENZIR_FMT_RUNTIME(fmt), result->string()), {});
+  return pattern::make(fmt::format(TENZIR_FMT_RUNTIME(fmt), result->string()),
+                       result->options());
 }
 
 } // namespace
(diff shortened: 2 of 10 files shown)


plugins/sigma/src/plugin.cpp
```diff
@@ -8,17 +8,160 @@
...
 #include <caf/expected.hpp>
+#include <caf/typed_event_based_actor.hpp>
 #include <fmt/format.h>
 
+#include <thread>
+
 namespace vast::plugins::sigma {
 
-class plugin final : public virtual language_plugin {
+class sigma_operator final : public crtp_operator<sigma_operator> {
```
dominiklohmann, line 35: After the changes from #3223 that remove language plugins we should write the changelog entry for this PR in terms of moving Sigma from a language to an operator in our language.

plugins/sigma/README.md
mavam, line 1: This documentation can probably go.

plugins/sigma/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+cmake_minimum_required(VERSION 3.19...3.24 FATAL_ERROR)
+
+project(
+  sigma
+  DESCRIPTION "Sigma query language plugin for Tenzir"
+  LANGUAGES CXX)
```
mavam, line 6: Why is this not a builtin?

dominiklohmann: I plan to eventually make this depend on libfswatch, so I think a plugin makes more sense than a builtin.

dominiklohmann (commented):
Just leaving some notes for myself.

plugins/sigma/src/plugin.cpp
```diff
@@ -0,0 +1,211 @@
...
+                   .field_builder(0)),
+              *row);
+            TENZIR_ASSERT(append_event_result.ok());
+            const auto append_rule_result = append_builder(
+              rule_schema,
+              caf::get<arrow::StructBuilder>(
+                *caf::get<arrow::StructBuilder>(*result_builder)
+                   .field_builder(1)),
+              caf::get<view<record>>(make_view(yaml)));
+            TENZIR_ASSERT(append_rule_result.ok());
```
Dakostu, line 141: I noticed that Sigma rule hits attach the entire rule to every event. This attachment seems to copy all information from the Sigma rule file, which includes potentially useless and potentially intimate info like the description and author.

1. What was the reasoning for this again?
2. Can they be grouped in a later version of the operator? Maybe just under the rule title?

plugins/sigma/src/plugin.cpp
```diff
@@ -0,0 +1,211 @@
...
+    }
+    auto last_update = std::chrono::steady_clock::now();
+    co_yield {}; // signal that we're done initializing
+    for (auto&& slice : input) {
+      if (slice.rows() == 0) {
+        co_yield {};
+        continue;
+      }
+      if (last_update + refresh_interval_ < std::chrono::steady_clock::now()) {
+        if (auto ok = state.update(state.path); not ok) {
```
Dakostu, line 99: Should successful file updates be logged?

Dakostu (approved):
I played around with this, managed to use this operator to continuously filter data, even while changing my rule file - and approved, considering that this is experimental.

I left some comments, probably for future consideration.

plugins/sigma/src/plugin.cpp
dominiklohmann: I can add a debug log for this, but I think everything else would be a bit too noisy.

plugins/sigma/src/plugin.cpp
dominiklohmann: You typically use Sigma rules to run detections. The information which Sigma rule matched is critical here. It's not uncommon to have hundreds of rules, and sometimes you may even create multiple results for a single event if multiple rules match.

On (2), I think the idea would be to put the Sigma functionality somewhere into the matcher, once we've made that pipeline-native. We have that on the roadmap already and it's coming up. That's probably the natural evolution of this operator that currently serves as a good proof of concept.

plugins/sigma/src/plugin.cpp
```diff
@@ -0,0 +1,211 @@
...
+            TENZIR_ASSERT(append_rule_result.ok());
+          }
+          auto result = result_builder->Finish().ValueOrDie();
+          auto rb = arrow::RecordBatch::Make(
+            result_schema.to_arrow_schema(), event->rows(),
+            caf::get<arrow::StructArray>(*result).fields());
+          co_yield table_slice{rb, result_schema};
+        }
+      }
+    }
```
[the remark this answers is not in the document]
dominiklohmann: Can you elaborate? I don't understand the comment, I'm afraid.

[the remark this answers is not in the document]
Dakostu: Launch `tenzir -vvv "sigma <anyfile>" ` and you will see what I mean.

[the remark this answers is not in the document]
Dakostu: I seem to have been tricked by the implicit `load -` in local pipelines. :raised_eyebrow:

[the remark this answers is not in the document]
dominiklohmann: We figured out that this was caused by a confusing behavior of the implicit `load -` source when nothing was attached to stdin, which was waiting for an EOF. Pressing CTRL-D is one possible way to solve this. Not in the scope of this PR to change that, assuming we even want to in the first place.

plugins/sigma/src/plugin.cpp
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  struct monitor_state {
+    [[nodiscard]] auto update(const std::filesystem::path& path)
+      -> caf::expected<void> {
+      if (std::filesystem::is_directory(path)) {
+        for (const auto& entry : std::filesystem::directory_iterator(path)) {
+          auto status = update(entry.path());
+          if (not status)
+            return status;
```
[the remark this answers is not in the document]
dominiklohmann: Found it! Patterns that end in a single backslash currently fail to convert to a valid regex.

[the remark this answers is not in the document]
dominiklohmann: Fixed in a71c9127e9; this now parses the entire set of sigma rules in the SigmaHQ/sigma repository.

plugins/sigma/tests/tests.cpp
Dakostu, line 1: I think you need to realign those unit tests now.

Dakostu (approved):
Thank you for fixing the issues!
I think this warrants an approval now, any more fixes afterwards seem to be superficial now.

plugins/sigma/src/parse.cpp
```diff
@@ -0,0 +1,503 @@
...
+                auto result = transform_sigma_string(str, {});
+                if (!result)
+                  return std::move(result.error());
+                if (str == result->string()) {
+                  return str;
+                }
+                return std::move(*result);
+              },
+              [](const std::string& x) -> caf::expected<data> {
+                TENZIR_WARN("making pattern: {}", __LINE__);
```
Dakostu, line 368: This can be removed.

plugins/sigma/tests/tests.cpp
dominiklohmann: I'm on that already. Two of them actually test broken things, but one revealed one more edge case bug in the translation from glob to regex that I must fix. It's pretty straightforward, though. Will resolve this once all tests are through and then rebase and merge.

PR #3141, opened 2023-05-11
Set minimum timestamp of partitions properly

dominiklohmann:
This fixes a regression that caused partitions created by the partition transformer to have a minimum timestamp of the epoch when the transformation pipeline created new table slices, i.e., lost the original import timestamp as part of the executed pipeline.

### DIFF
--- a/libvast/src/system/partition_transformer.cpp
+++ b/libvast/src/system/partition_transformer.cpp
@@ -382,10 +382,11 @@ auto partition_transformer(
             = caf::make_copy_on_write<partition_synopsis>();
         }
         auto* unshared_synopsis = partition_data.synopsis.unshared_ptr();
-        unshared_synopsis->min_import_time
-          = std::min(slice.import_time(), unshared_synopsis->min_import_time);
-        unshared_synopsis->max_import_time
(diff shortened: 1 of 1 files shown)


tobim:
I believe we need to adjust the initial value of `state.min_import_time` to `vast::time::max()` to make it fully correct. Possible diff:
```diff
diff --git a/libvast/include/vast/system/partition_transformer.hpp b/libvast/include/vast/system/partition_transformer.hpp
index e9f8d5a770..e79184775e 100644
--- a/libvast/include/vast/system/partition_transformer.hpp
+++ b/libvast/include/vast/system/partition_transformer.hpp
@@ -103,10 +103,10 @@ struct partition_transformer_state {
   size_t events = 0ull;
 
   /// Oldest import timestamp of the input data.
-  vast::time min_import_time = {};
+  vast::time min_import_time = vast::time::max();
 
   /// Newest import timestamp of the input data.
-  vast::time max_import_time = {};
+  vast::time max_import_time = vast::time::min();
 
   /// The data of the newly created partition(s).
   std::multimap<type, active_partition_state::serialization_data> data = {};

```

tobim (approved):

PR #3147, opened 2023-05-12
Fix remaining partitions counter in the rebuilder

dominiklohmann:
This fixes a few things around rebuild statistics, in particular the `rebuilder.partitions.remaining` metric and the
`rebuild.statistics.remaining` field in the output of `vast status rebuild`:

- Fixed a bug where the rebuilder dereferenced an invalidated iterator into the list of catalog lookup results when no partitions for one schema needed to be rebuilt. This bug only affected metrics, but not the actual partitions used for the rebuilding.
- Removed a trigger for rebuilder metrics that caused metrics to be emitted twice when done, or before some partitions that did not need to be rebuilt were pruned from the list of partitions.

### DIFF
--- a/libvast/builtins/commands/rebuild.cpp
+++ b/libvast/builtins/commands/rebuild.cpp
@@ -320,7 +320,7 @@ struct rebuilder_state {
                   && result.partition_infos.front().version
                        < version::current_partition_version) {
                 // Edge case: we can't do anything if we have a single
-                // underiszed partition for a given schema.
+                // undersized partition for a given schema.
                 result.partition_infos.clear();
               }
             }
@@ -330,7 +330,7 @@ struct rebuilder_state {
                                              result.partition_infos.end());
           }
           if (run->statistics.num_total == 0) {
-            VAST_DEBUG("{} ignores requested rebuild for 0 partitions", *self);
+            VAST_DEBUG("{} ignores rebuild request for 0 partitions", *self);
             return finish({}, true);
           }
           if (run->options.automatic)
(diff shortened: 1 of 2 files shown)


libvast/builtins/commands/rebuild.cpp
```diff
@@ -323,26 +316,23 @@ struct rebuilder_state {
...
-              if (result.partition_infos.empty()) {
-                lookup_result.candidate_infos.erase(type);
-                if (lookup_result.candidate_infos.empty()) {
-                  return finish({});
-                }
+              if (result.partition_infos.size() == 1
+                  && result.partition_infos.front().version
+                       < version::current_partition_version) {
+                // Edge case: we can't do anything if we have a single
+                // underiszed partition for a given schema.
```
mavam, line 323:
```suggestion
                // undersized partition for a given schema.
```

libvast/builtins/commands/rebuild.cpp
```diff
@@ -323,26 +316,23 @@ struct rebuilder_state {
...
           }
-          const auto all_empty = std::all_of(
-            lookup_result.candidate_infos.begin(),
-            lookup_result.candidate_infos.end(),
-            [](const auto& candidate_info) {
-              return candidate_info.second.partition_infos.empty();
-            });
-          if (all_empty)
+          if (run->statistics.num_total == 0) {
+            VAST_DEBUG("{} ignores requested rebuild for 0 partitions", *self);
```
mavam, line 333:
```suggestion
            VAST_DEBUG("{} ignores rebuild request for 0 partitions", *self);
```

mavam (approved):

PR #3151, opened 2023-05-16
Replace aggregation_func_char parser with plugin_name parser

Dakostu:
This PR replaces the `aggregation_func_char` parser with the almost identical but more lenient `plugin_name` parser.


jachris (approved):
Approving as agreed upon in Slack.

PR #3160, opened 2023-05-19
Support flattened data in JSON parser

patszt:
The zeek-json parser didn't fully work due to lack of unflattening. The JSON parser supports --separator option which is hardcoded to `.` with zeek-json.

vast::type is used in series_builders to preserve attributes when constructed with a known shcema.

### DIFF
--- a/libvast/src/adaptive_table_slice_builder.cpp
+++ b/libvast/src/adaptive_table_slice_builder.cpp
@@ -27,19 +27,19 @@ auto init_root_builder(const type& start_schema, bool allow_fields_discovery)
 
 adaptive_table_slice_builder::adaptive_table_slice_builder(
   type start_schema, bool allow_fields_discovery)
-  : root_builder_{init_root_builder(start_schema, allow_fields_discovery)},
-    start_schema_{std::move(start_schema)} {
+  : root_builder_{init_root_builder(start_schema, allow_fields_discovery)} {
 }
 
 auto adaptive_table_slice_builder::push_row() -> row_guard {
   return row_guard{*this};
 }
 
-auto adaptive_table_slice_builder::finish() -> table_slice {
+auto adaptive_table_slice_builder::finish(std::string_view slice_schema_name)
+  -> table_slice {
   auto final_array = finish_impl();
   if (not final_array)
     return table_slice{};
-  auto slice_schema = get_schema();
+  auto slice_schema = get_schema(slice_schema_name);
   const auto& struct_array
     = static_cast<const arrow::StructArray&>(*final_array);
   const auto batch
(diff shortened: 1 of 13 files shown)


libvast/src/adaptive_table_slice_builder.cpp
```diff
@@ -57,25 +57,14 @@ auto adaptive_table_slice_builder::rows() const -> detail::arrow_length_type {
...
-            and caf::get<record_type>(schema)
-                  == caf::get<record_type>(start_schema_))
-          return start_schema_;
-        auto schema_name = schema.make_fingerprint();
-        return type{std::move(schema_name), std::move(schema)};
-      },
+    [slice_schema_name](const auto& builder) -> vast::type {
+      auto schema = builder.type();
+      if (slice_schema_name.empty())
+        return type{schema.make_fingerprint(), std::move(schema)};
```
dominiklohmann, line 66: UB on the order; can just remove the `std::move`, I think.

web/docs/understand/formats/zeek-json.md
```diff
@@ -1,4 +1,5 @@
 # zeek-json
 
-The `zeek-json` format is an alias for [`json`](json.md) with the argument
-`--separator=_path:zeek`.
+The `zeek-json` format is an alias for [`json`](json.md) with the arguments:
+- `--separator=_path:zeek`
```
dominiklohmann, line 4:
```suggestion
- `--selector=_path:zeek`
```

web/docs/understand/formats/json.md
```diff
@@ -36,6 +36,48 @@ For example, the [Suricata EVE JSON](suricata.md) format includes a field
...
+Sometimes the separator will stay as a part of a field name. For example the:
+```json
+{
+  "cpu" : "cpu01",
+  "cpu.logger" : "logger1"
+}
+```
+
+will not be transformed. The "cpu" field has a a value so it can't be unnested
+into a JSON object.
```
dominiklohmann, line 79: This is nice. I've got some suggestions, but it's more or less the same content:

```suggestion
### `--unnest-separator=<string>` (Parser)

A delimiter that, if present in keys, causes values to be treated as values of nested records.

A popular example of this is the [Zeek JSON](zeek-json.md) format. It includes the fields `id.orig_h`, `id.orig_p`, `id.resp_h`, and `id.resp_p` at the top-level. The data is best modeled as an `id` record with four nested fields `orig_h`, `orig_p`, `resp_h`, and `resp_p`.

Without an unnest separator, the data looks like this:

---json
{
  "id.orig_h" : "1.1.1.1",
  "id.orig_p" : 10,
  "id.resp_h" : "1.1.1.2",
  "id.resp_p" : 5
}
---

With the unnest separator set to `.`, VAST reads the events like this:

---json
{
  "id" : {
    "orig_h" : "1.1.1.1",
    "orig_p" : 10,
    "resp_h" : "1.1.1.2",
    "resp_p" : 5
  }
}
---
```

I used triple dashes over triple backticks because that otherwise breaks the suggestion mode of review comments, so please use backticks instead and also reflow the text to 80 lines.

dominiklohmann: I unresolved this as this part is still missing:

> I used triple dashes over triple backticks because that otherwise breaks the suggestion mode of review comments, so please use backticks instead and also reflow the text to 80 lines.

patszt: > I unresolved this as this part is still missing:
> 
> > I used triple dashes over triple backticks because that otherwise breaks the suggestion mode of review comments, so please use backticks instead and also reflow the text to 80 lines.

Sorry, i misunderstood what you wrote. Thought that i should be using the --- instead of ```. Then i thought that the please use backticks was a TYPO for dashes. And then i thought i was missing 80 character limit in the original commit

dominiklohmann: No worries, I should've phrased that better. That was really confusing.

PR #3162, opened 2023-05-22
A collection of minor UX improvements

tobim:


libvast/src/system/application.cpp
```diff
@@ -243,6 +243,7 @@ auto make_status_command() {
     "shows properties of a server process by component; optional positional "
     "arguments allow for filtering by component name",
     opts("?vast.status")
+      .add<int64_t>("timeout", "how long to wait for components to report")
```
dominiklohmann, line 246: This option should be a string, from which we then parse a duration. Take a look at other timeout option for examples.

dominiklohmann (approved):

PR #3113, opened 2023-05-01
Add `explore` operator and `table` format (TUI)

mavam:
This PR adds a `tui` operator that renders pipeline results in a terminal user interface (TUI).

The PR is continuation of #2838, leveraging the [FTXUI](https://arthursonzogni.github.io/FTXUI/) library.

See the docs page for details.

Screenshot:

image (2)

Review instructions:
- Review the commit that adds `record_type::depth()` separately.
- Go through the rest file-by-file, with the operator plugin as entry point.
- If you need help on understanding FTXUI, I'm happy to pair on it.
- Make sure the CMake looks right.

PR #3163, opened 2023-05-22
Bump nixpkgs to b31c968ff28

tobim:
This updates Apache Arrow to v12, and unblocks #3113.


mavam:
Arrow's CMake doesn't seem to be happy.

tobim:
Finally everything works again.

PR #3166, opened 2023-05-24
Remove old commands

dominiklohmann:
This removes a lot of the old commands. As we're moving to a pipeline-first setup, they have simply become outdated.

We have yet to replace `count`, `status`, `import`, and `export` with operators, but this already paves the way for some upcoming changes.

```[tasklist]
### Tasks
- [x] Remove stop command
- [x] Remove peer command
- [x] Remove send command
- [x] Remove kill command
- [x] Remove version command
- [x] Remove spawn command
- [x] Write changelog entries
```


tobim:
I can't reproduce the integration test failure of `exporter timeout` locally. Let's just re-run.
The Dockerfile still contains a `RUN vast version`.

tobim (approved):
Glad that we don't have to support those any more.

PR #3171, opened 2023-05-25
Fix rare crash when transforming sliced nested arrays

dominiklohmann:
This fixes a rather complicated-to-find bug that would cause transformation operators accessing specific nested arrays to produce oversized arrays when used after a slicing operation. This most commonly happened after `head` and `tail`, but could also occur after `where` or the hidden `rebatch` operator.

When we slice an array, we don't actually modify the array itself. Instead, we just modify the validity bitmap and adjust the internal offset and length values of the array. Now there were two related problems, one being an understanding and the other a bug in Apache Arrow:
1. The misunderstanding: To combine null bitmaps, nested arrays must be explicitly accessed via `arrow::StructArray::Flatten()` instead of `arrow::StructArray::fields()`. The former performs a logical and operation on the null bitmaps of the struct array and the nested array.
2. The bug: The `arrow::FieldPath::Get(...)` function returns an array pointing into the passed array or batch. The function recreates the array from the underlying child array data, thus returning the original nested array, as if the outer array were not sliced at all.

This PR fixes these two issues, and also as a bonus a bug that caused `tail` to return too many events when its input were lots of really small slices.


tobim (approved):
Pair reviewed in a call.

PR #3187, opened 2023-06-01
Introduce the `tenzir` and `tenzird` binaries

dominiklohmann:
We're renaming VAST to Tenzir. This is the first PR of a series of PRs that make this change. We want to make the change while causing as little havoc as possible, and there is an announcement blog post coming out soon that will explain what exactly is changing and why.

As a first step, this renames the `vast` binary to `tenzir`, and adds a symlink `vast` that when used prints a warning on stderr about the upcoming change.

As a second step, this adds `tenzird` and `tenzirctl` symlinks to the `tenzir` binary.

- `tenzird` is a replacement for `vast start` and starts a node
- `tenzir` is a replacement for `vast exec` and executes a pipeline
- `tenzirctl` replaces all other commands until they're migrated to pipeline operators

This makes it clearer that Tenzir has two core functionalities: nodes and pipelines.

It's an exciting time, and we at Tenzir are looking forward to making this finally making this change.


cmake/VASTRegisterPlugin.cmake
tobim, line 1: You need to rename the target on line 593 as well.

dominiklohmann: Good catch, fixed and pushed.

dominiklohmann:
@tobim can you push the necessary Nix changes on top of this PR as part of your review? I think that's easiest.

tobim:
Done and rebased to fix the conflict.

vast/CMakeLists.txt
```diff
@@ -1,25 +1,36 @@
...
+            "${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_BINDIR}/${name}"
+    DEPENDS tenzir
+    COMMENT "Creating ${name} symlink")
+  add_custom_target(
+    "${name}" ALL DEPENDS "${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_BINDIR}/${name}")
+  install(
+    CODE "
+    execute_process(COMMAND
+      \"${CMAKE_COMMAND}\" -E create_symlink \"./tenzir\"
+      \"\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/${name}\")")
```
dominiklohmann, line 24: Before merging, we must manually check that CPack handles this correctly by inspecting the CI-generated packages.

tobim (approved):

PR #3188, opened 2023-06-01
Fix reconnect attempts for remote pipelines

dominiklohmann:
To reproduce the bug, run `vast exec 'remote version'` before running `vast start` on current main, and observe a broken promise error instead of a reconnect attempt after three seconds.


Dakostu (approved):

Dakostu:
Thank you for the quick fix!

PR #3190, opened 2023-06-02
Report adaptive table slice builder errors to the control plane

patszt:
The errors are returned from adaptive_table_slice_builder field/list guards.
These errors are reported to control_plane whenever applicable in the current code base.

Also implemented early return of an error when value casting failed k  adaptive_table_slice_builder that can't change the schema.

### DIFF
--- a/libvast/include/vast/detail/adaptive_table_slice_builder_guards.hpp
+++ b/libvast/include/vast/detail/adaptive_table_slice_builder_guards.hpp
@@ -169,7 +169,10 @@ public:
 
   /// @brief Adds a value to a field.
   /// @param view View of a value to be added.
-  /// @return Error describing why the addition wasn't successful.
+  /// @return Error describing why the addition wasn't successful. TODO:
+  /// Returning a caf::error after each addition may significantly slow down the
+  /// parsing. It might be worthwhile to check if we need to optimize this in
+  /// the future.
   template <class ViewType>
     requires requires(ViewType x) { materialize(x); }
   auto add(ViewType view) -> caf::error {
--- a/libvast/include/vast/detail/adaptive_table_slice_builder_guards.hpp
+++ b/libvast/include/vast/detail/adaptive_table_slice_builder_guards.hpp
@@ -61,16 +61,18 @@ public:
   /// @brief Adds a value to a list. Use push_record and push_list to add record
   /// and list respectively.
   /// @param view View of a value to be added.
+  /// @return Error describing why the addition wasn't successful.
   template <class ViewType>
     requires requires(ViewType x) { materialize(x); }
-  auto add(ViewType view) -> void {
+  auto add(ViewType view) -> caf::error {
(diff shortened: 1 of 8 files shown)


libvast/include/vast/detail/adaptive_table_slice_builder_guards.hpp
```diff
@@ -61,16 +61,18 @@ class list_guard {
   /// @brief Adds a value to a list. Use push_record and push_list to add record
   /// and list respectively.
   /// @param view View of a value to be added.
+  /// @return Error describing why the addition wasn't successful.
   template <class ViewType>
     requires requires(ViewType x) { materialize(x); }
-  auto add(ViewType view) -> void {
+  auto add(ViewType view) -> caf::error {
```
dominiklohmann, line 67: Constructing a `caf::error` for every individual value seems like it could hurt performance a lot. Can we return something that's a bit more light-weight, like an enum? How detailed are the errors?

patszt: They are quite detailed. E.g 
```
return caf::make_error(ec::convert_error,
                             fmt::format("unable to convert {} into {}: "
                                         "the value doesn't correspond to any "
                                         "enum state",
```

Hmm a quick alternative would be to have another parameter instead of a return value.
E.g out caf::error& or some std::function error_reporter that get's called.

I could also see a try/catch as the errors are rare and they only happen when we cast. I think we agreed that casting is rare and doesn't need to be quick.

dominiklohmann: @jachris since these are effectively diagnostics and that's your domain currently, what do you think?

jachris: It depends. As a user, what would I expect to see if this error happens? What do I need to know? Can I do anything about it? Is this error rare? And if it does happen, do I get it at most once, perhaps a few times, or potentially a thousand times?

If this error is rare, happens at most a few times and performance doesn't matter, we could stick to `caf::error` here for now. The performance of the happy path is not degraded by this. As soon as we get proper diagnostics, we'd have to make some changes. However, I could also see that some parsing errors could occur thousands of times, if I have a misconfigured schema for example. Then this would probably need a better solution.

jachris: Thank you for elaborating. I'd say we can do whatever works for now and improve this later on – unless one of you already has a good idea on how to solve this situation.

patszt: @dominiklohmann Shpuld i do some benchmarks to see how these changes impact the happy path performance?

dominiklohmann: No, let's not do that now. Can you add a comment in the code for now to revisit this?

dominiklohmann (approved):

PR #3192, opened 2023-06-03
Remove `Dockerfile.extra`

dominiklohmann:
We only need a single Dockerfile if we just re-introduce the `vast` target at the very end, such that the default target does not require access to the proprietary plugins.

Additionally, this speeds up the build of the `vast-ce` and `vast-ee` targets when using Docker BuildKit.


Dockerfile
```diff
@@ -123,3 +123,94 @@ RUN vast exec version
...
+
+FROM vast-ce AS vast-ee
+
+COPY --from=compaction-plugin --chown=vast:vast /plugin/compaction /
+
+# -- vast ----------------------------------------------------------------------
+
+# As a last stage we re-introduce VAST without proprietary plugins so that it's
+# the default when not specifying a build target.
+FROM production AS vast
```
tobim, line 216: Can you switch from the `production` target to `vast` in `docker-config-base.json`, `docker.yaml`, and `docker-manual.yaml` in the workflows directory?

tobim (approved):

Dockerfile
dominiklohmann: Done in 989d4a5b40

PR #3193, opened 2023-06-05
Implement an `export` pipeline operator

Dakostu:
This change implements an `export` pipeline operator that retrieves events from a VAST node.


libvast/builtins/operators/export.cpp
```diff
@@ -0,0 +1,148 @@
...
+      = get_node_components<catalog_actor, accountant_actor, filesystem_actor>(
+        blocking_self, ctrl.node());
+    auto [catalog, accountant, fs] = std::move(*components);
+    auto expr = trivially_true_expression();
+    if (auto pushdown = predicate_pushdown(expr)) {
+      expr = std::move(*pushdown).first;
+    } else {
+      VAST_VERBOSE("no expression found: launching 'export' to retrieve all "
+                   "persisted events");
+    }
```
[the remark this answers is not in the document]
jachris: You could also make `to_string()` return `export | where <expr>`, but I think that this is quite hacky.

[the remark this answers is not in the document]
Dakostu: I probably need to also add the expression to `make_operator`...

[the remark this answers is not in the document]
jachris: We discussed this internally. The hackyness of `| where ...` will also go away with the upcoming serialization approach. Also, my mistake, I think you actually need 
~~~
    return std::pair{trivially_true_expression(),
                     std::make_unique<export_operator>(conjunction{expr_, expr})};
~~~
or something similar for full correctness. But otherwise 014416615d8e63b7a973d48b2c3c5e606735df1c looks good :+1:

web/docs/understand/operators/sources/export.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Description
+
+The `export` operator retrieves events from a VAST node.
+
+:::note Flush to disk
+Pipelines starting with the `export` operator do not access events that are not
+written to disk.
+
+We plan to change this behavior in the near future. Until then, we recommend
```
dominiklohmann, line 19: I don't think we do—it is absolutely intended that we only query passive partitions, so this second part of this note should be toned down a bit.

What we must add in the future, however, is the following:
- A mechanism to export continuously (i.e., retrieve all new partitions as they are registered at the catalog).
- Replace explicit flushing with an implicit flush in the import operator, and respect partition timeouts in the import operator.

web/docs/understand/operators/sources/export.md
```diff
@@ -0,0 +1,37 @@
...
+running `vast flush` before exporting events to make sure they're available for
```
dominiklohmann, line 20:
```suggestion
running `tenzir-ctl flush` before exporting events to make sure they're available for
```

dominiklohmann (approved):
This works really well for a first implementation of the operator!

web/docs/understand/operators/sources/export.md
Dakostu: I replaced the command in https://github.com/tenzir/vast/pull/3193/commits/45aaa7cf4310261cd85e18870d054ffae86b2149

web/docs/understand/operators/sources/export.md
Dakostu: I removed the "We plan..." sentence in https://github.com/tenzir/vast/pull/3193/commits/45aaa7cf4310261cd85e18870d054ffae86b2149.

Should I write issues for those two points?

PR #3196, opened 2023-06-06
Update the `vast-plugins` submodule pointer to include elaboration on the `managed` pipeline flag

Dakostu:
This change reflects the change in the `vast-plugin` PR: https://github.com/tenzir/vast-plugins/pull/50


rdettai (approved):
not sure I fully understand why the API docs are duplicated, but if thats the process then ✅

Dakostu:
@rdettai They are defined in separate `vast-plugin`s, then using `vast web openapi` collected into a single OpenAPI file to display in `vast.io/api`. I'd like this to be automated but that's for later.

PR #3198, opened 2023-06-07
Remove cloud documentation

mavam:
This PR removes the documentation for the AWS deployment. It is a moving target and we have plans to bring it back in a different shape.


mavam:
@rdettai if there is anything that we should strip from the repo in addition to documentation, please let me know.

rdettai:
you also have a mention here: https://github.com/tenzir/vast/blob/c3b3b54e84240e7ca6eb0a53357e6789ad57e0e2/web/docs/setup/deploy/README.md?plain=1#L9

rdettai (approved):

PR #3199, opened 2023-06-07
Update plugin documentation

mavam:
This PR updates the plugin documentation.

I've replaced the always out-of-sync Excalidraw with a Mermaid diagram. This is a lot easier to maintain. The rendered output looks like this and plugins are clickable:

image

Review instructions: please check plugin types for completeness. If I am missing something, it would be great if you could just push a commit to this branch—it's all yours!


web/docs/develop/architecture/plugins.md
```diff
@@ -24,64 +24,76 @@ static plugins. Because they are always built, we call them *builtins*.
...
+  Saver
+  click Saver "#saver"
+end
+subgraph Pipeline
+  Operator
+  click Operator "#operator"
+  AggregationFunction
+  click AggregationFunction "#aggregation-function"
+end
+```
```
dominiklohmann, line 59: This doesn't really match the theme of the documentation. I think we should just not have this image in the first place as it doesn't add any value over the left sidebar that Docusaurus already generates from the headers.

web/docs/develop/architecture/plugins.md
```diff
@@ -95,36 +107,16 @@ a custom store backend. Unlike the store plugin, the store actor plugin is
...
-:::
-
-The reader plugin adds a new format to parse input data, such as JSON (ASCII) or
-PCAP (binary).
-
-Reader plugins automatically add the subcommand `tenzir import <plugin name>`.
+### Command
 
-### Writer
+The command plugin adds a new command to the `tenzir` executable, at a
```
dominiklohmann, line 112:
```suggestion
The command plugin adds a new command to the `tenzir-ctl` executable, at a
```

web/docs/develop/architecture/plugins.md
```diff
@@ -24,64 +24,76 @@ static plugins. Because they are always built, we call them *builtins*.
...
-
-The aggregation function plugin adds a new [aggregation
-function](../../understand/operators/transformations/summarize.md#aggregate-functions)
-for the `summarize` pipeline operator that performs an incremental aggregation
-over ajj set of grouped input values of a single type.
+The parser plugin defines the output side of a
+[format](../../understand/formats/) for use in the `write FORMAT to CONNECTOR`
+and `print FORMAT` operators.
 
 ### Store
```
dominiklohmann, line 98: This should mention that every store plugin is implicitly also a parser and printer plugin.

web/docs/develop/architecture/plugins.md
```diff
@@ -24,64 +24,76 @@ static plugins. Because they are always built, we call them *builtins*.
...
+flowchart TB
+subgraph System
+  Command
+  click Command "#command"
+  Component
+  click Component "#component"
+end
+subgraph Storage
+  StoreActor
+  click StoreActor "#store"
```
dominiklohmann, line 37: The store actor plugin should not be documented and will be removed in the future.

dominiklohmann (approved):
Just minor things, other than that obviously good to go.

web/docs/develop/architecture/plugins.md
mavam: I think I'd be fine with removing it *iff* we add the intermediate headings that help with categorization.

dominiklohmann: I would suggest the following:

System: 
- Component

Format:
- Parser
- Printer
- Store

Connector:
- Loader
- Saver

Pipeline:
- Operator
- Aggregation Function

mavam: I would move Store to System or a dedicated category.

dominiklohmann: System then, but then move it to the very bottom so the user already encountered parsers and printers.

PR #3202, opened 2023-06-07
impl

patszt:


libvast/builtins/formats/json.cpp
```diff
@@ -454,6 +437,16 @@ auto make_parser_impl(generator<chunk_ptr> json_chunk_generator,
       continue;
     }
     for (auto doc_it = stream.begin(); doc_it != stream.end(); ++doc_it) {
+      // doc.error() will inherit all errors from *doc_it and get_value.
+      // No need to check after each operation.
+      auto doc = (*doc_it).get_value();
+      if (auto err = doc.error()) {
+        ctrl.warn(caf::make_error(
+          ec::parse_error, fmt::format("failed to fully parse '{}' : {}. "
+                                       "Some events can be skipped.",
```
dominiklohmann, line 446:
```suggestion
          ec::parse_error, fmt::format("skips invalid JSON '{}' : {}",
```

libvast/include/vast/detail/adaptive_table_slice_builder_guards.hpp
```diff
@@ -117,18 +117,51 @@ class list_guard {
...
+        },
+        [](const list_type&) -> caf::error {
+          die("can't add values to the list_guard with list value_type");
+        },
+        [](const map_type&) -> caf::error {
+          die("can't add values to the list_guard with map value_type");
+        },
+        [](const record_type&) -> caf::error {
+          die("can't add values to the list_guard with record value_type");
+        },
```
dominiklohmann, line 162: Calling `die` for map type is fine since that can no longer occur. For list and record, however, we don't want to crash, but rather warn and skip if we don't support it inside lists yet.

What's the missing piece to support `list<list<...>>` and `list<record { ... }>`?

dominiklohmann (approved):
I tested this by letting speeve generate input for half an hour or so, and it was able to ingest that all. I left two comments with questions inline.

libvast/include/vast/detail/adaptive_table_slice_builder_guards.hpp
patszt: Actually having die here makes sense.
The list_guard does not support doing add(list{})
The user is supposed to push_list as many times needed and then only add the non list/record values with the add method.

dominiklohmann: Can you write a follow-up story then before resolving this?

libvast/builtins/formats/json.cpp
dominiklohmann: What about this?

```suggestion
          ec::parse_error, fmt::format("failed to parse JSON and may skip some events '{}' : {}",
```

The reason being that the printed JSON in the log message could be rather lengthy—so I think we should print the outcome first.

Also, stylig-wise, please use a single sentence for warnings that starts lower-case and has no dot at the end, otherwise it's inconsistent with the remaining warnings.

libvast/include/vast/detail/adaptive_table_slice_builder_guards.hpp
patszt: Yup

PR #3204, opened 2023-06-08
Introduce a potpourri of smaller changes and fixes

dominiklohmann:
- `zeek` -> `zeek-json` to align with the docs
- `VAST` -> `node` in the node startup info message
- Compiling now works with AppleClang 15.x
- Removal of dead code from the `summarize` operator that was left over from the old YAML operator config
- Fixed bug that caused `/serve` to sometimes return too many events

### DIFF
--- a/libvast/builtins/operators/serve.cpp
+++ b/libvast/builtins/operators/serve.cpp
@@ -205,31 +205,36 @@ struct serve_request {
 /// copying data.
 auto split(std::vector<table_slice> events, uint64_t partition_point)
   -> std::pair<std::vector<table_slice>, std::vector<table_slice>> {
-  auto split_it = events.begin();
-  auto offset = uint64_t{0};
-  while (split_it != events.end()) {
-    const auto num_rows = split_it->rows();
-    ++split_it;
-    if (num_rows >= partition_point) {
-      offset = partition_point - num_rows;
-      break;
+  auto it = events.begin();
+  for (; it != events.end(); ++it) {
+    if (partition_point == it->rows()) {
+      return {
+        {events.begin(), it + 1},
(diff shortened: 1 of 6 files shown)


mavam:
> Compiling now works with AppleClang 15.x

Does this mean the Homebrew LLVM is no longer required on macOS?

libvast/builtins/operators/serve.cpp
```diff
@@ -285,7 +290,9 @@ struct managed_serve_operator {
     }
     // Cut the results buffer.
     auto results = std::vector<table_slice>{};
+    VAST_WARN("getting {}/{} events", requested, rows(buffer));
```
mavam, line 293: Should this really be warning? (Applies to comment below as well.)

dominiklohmann:
> Does this mean the Homebrew LLVM is no longer required on macOS?

Only if you use the current Xcode-beta and now how to set that up for your local development. So maybe in a few months. I just happened to test a few things with it, and since this was the only change necessary I figured I'd just make it.

libvast/builtins/operators/serve.cpp
dominiklohmann: Ah, the classic debugging left-over

libvast/builtins/operators/serve.cpp
```diff
@@ -285,7 +290,9 @@ struct managed_serve_operator {
...
     std::tie(results, buffer) = split(buffer, requested);
+    VAST_WARN("got {} events", rows(results));
```
dominiklohmann, line 295:
```suggestion
    std::tie(results, buffer) = split(buffer, requested);
```

mavam (approved):

PR #3208, opened 2023-06-08
Change Arrow extension type and metadata prefixes

dominiklohmann:
We now register extension types as `tenzir.ip`, `tenzir.subnet`, and `tenzir.enumeration` instead of `vast.address`, `vast.subnet`, and `vast.enumeration`, respectively. Type metadata now has a `Tenzir:` prefix instead of a `VAST:` prefix.

The change is mostly backwards compatible. The old type names and prefixes are still supported. However, reading old Apache Feather V2 and Apache Parquet inside the database directory files using the Python bindings will not work until a `tenzir-ctl rebuild --all` was run.

### DIFF
--- a/libvast/src/format/arrow.cpp
+++ b/libvast/src/format/arrow.cpp
@@ -164,7 +164,7 @@ reader::read_impl(size_t max_events, size_t max_slice_size, consumer& f) {
     // into crashes. A proper fix for this requires casting to VAST-compatible
     // record batch.
     const auto& metadata = *read_result->batch->schema()->metadata();
-    if (metadata.FindKey("Tenzir:name:0") == -1
+    if (metadata.FindKey("TENZIR:name:0") == -1
         && metadata.FindKey("VAST:name:0") == -1) {
       VAST_WARN("{} skips record batch with {} rows: metadata is "
                 "incomaptible with VAST",
--- a/libvast/src/format/arrow.cpp
+++ b/libvast/src/format/arrow.cpp
@@ -163,8 +163,9 @@ reader::read_impl(size_t max_events, size_t max_slice_size, consumer& f) {
     // batch, but it's a good enough heuristic to prevent users from running
     // into crashes. A proper fix for this requires casting to VAST-compatible
     // record batch.
-    if (read_result->batch->schema()->metadata()->FindKey("VAST:name:0")
-        == -1) {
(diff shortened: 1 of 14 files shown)


libvast/src/format/arrow.cpp
```diff
@@ -163,8 +163,9 @@ reader::read_impl(size_t max_events, size_t max_slice_size, consumer& f) {
     // batch, but it's a good enough heuristic to prevent users from running
     // into crashes. A proper fix for this requires casting to VAST-compatible
     // record batch.
-    if (read_result->batch->schema()->metadata()->FindKey("VAST:name:0")
-        == -1) {
+    const auto& metadata = *read_result->batch->schema()->metadata();
+    if (metadata.FindKey("Tenzir:name:0") == -1
```
mavam, line 167:
```suggestion
    if (metadata.FindKey("TENZIR:name:0") == -1
```

I find `TENZIR` or `tenzir` arguably easier to remember.

mavam (approved):
The implementation looks good.

I talked with @tobim about the need of a full rebuild. Is this really needed?

dominiklohmann:
> I talked with @tobim about the need of a full rebuild. Is this really needed?

If you want to read the files using our Python Arrow extension types, then yes, otherwise no.

libvast/src/format/arrow.cpp
dominiklohmann: I'd prefer either Tenzir or tenzir. Do you have a strong opinion on this before I merge this?

mavam: Kind of. Arrow uses `ARROW` and I would expect that we follow the naming scheme. So I'd say the most natural one would actually be `TENZIR`. What's your rationale for `Tenzir`?

mavam: I was referring to unit test output in this very PR.

dominiklohmann: Right, the undocumented and internal `ARROW:extension:{name,metadata}` keys.

I don't have too strong of an opinion on this, and it seems like you have, so I'll just make the change.

mavam: Thanks!

dominiklohmann: Implemented in 2d9ccc0148857accec408e3b3f1cdaeba4e49346

PR #3210, opened 2023-06-09
Support `TENZIR_*` environment variables

dominiklohmann:
This change makes it so we pick up `TENZIR_*` env variables with a higher precedence than `VAST_*` env variables. For backwards compatibility, we still support reading the old ones.

### DIFF
--- a/libvast/src/configuration.cpp
+++ b/libvast/src/configuration.cpp
@@ -59,7 +59,7 @@ struct has_extension_type
 /// 2. A "__" translates into the record separator '.'
 /// @pre `!prefix.empty()`
 std::optional<std::string>
-to_config_key(std::string_view key, std::string_view prefix = "VAST") {
+to_config_key(std::string_view key, std::string_view prefix) {
   VAST_ASSERT(!prefix.empty());
   // PREFIX_X is the shortest allowed key.
   if (prefix.size() + 2 > key.size())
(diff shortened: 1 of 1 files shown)


libvast/src/configuration.cpp
```diff
@@ -342,11 +360,17 @@ caf::error configuration::parse(int argc, char** argv) {
...
   if (auto it
       = std::find(command_line.begin(), command_line.end(), "--bare-mode");
-      it != command_line.end())
+      it != command_line.end()) {
     caf::put(content, "vast.bare-mode", true);
-  else if (auto vast_bare_mode = detail::getenv("VAST_BARE_MODE"))
-    if (*vast_bare_mode == "true")
+  } else if (auto vast_bare_mode = detail::getenv("TENZIR_BARE_MODE")) {
+    if (*vast_bare_mode == "true") {
+      caf::put(content, "vast.bare-mode", true);
```
mavam, line 367: Do you plan to adapt the `vast.` prefix later on?

mavam (approved):
Looks reasonable to me.

libvast/src/configuration.cpp
dominiklohmann: Yes, but that's a much bigger PR and more complex than this one.

PR #3211, opened 2023-06-09
Remove `lsvast`

dominiklohmann:
The `lsvast` utility is a blast from the past: Its main value lies in allowing for recovery of VAST databases, which can nowadays easily be done with a pipeline that reads Feather or Parquet files directly.

The introspection of databases is moving more and more into the `show` command, with early plans to move that into an operator that would allow for a deeper introspection of VAST's components.

Since the utility is no longer being actively maintained, we are removing it instead of renaming it to `lstenzir`. If we need it again in the future, then we will bring it back.


tobim (approved):
This seems reasonable to me.

PR #3212, opened 2023-06-09
Rename default database directory to `tenzir.db`

dominiklohmann:
This renames the default database directory path to `tenzir.db`. We don't expect this to break any serious deployments, which should all set the database directory manually rather than putting it somewhere in the current working directory.


tobim (approved):

dominiklohmann:
Note that this is based on #3211 which I wanted to be looked at separately, so I'm not going to merge this yet.

PR #3214, opened 2023-06-09
Remove language frontends and rewrite Sigma usage

mavam:
This PR removes language frontends from the documentation. Since the Sigma
frontend was the only one, and we actually have an open draft PR to bring back
Sigma rule execution as operator, I went ahead and wrote the documentation for
the `sigma` operator in the same vein.


dominiklohmann (approved):
This reads nicely. We should probably merge #3138 even in its currently very experimental state soon now, though, if we want to document it already.

PR #3216, opened 2023-06-09
Rewrite getting started instructions

mavam:
This PR rewrites the getting started instructions for a better onboarding experience of Tenzir.

- [x] Update getting started guide
- [x] Add content from old quickstart guide into other guides
- [ ] Complete reshaping guide


mavam:
I'm considering to merge this even though there are still todos. I plan to shift my attention to other parts in the docs before coming back, as we need to fix a few ankle biter bugs before we can complete the examples. But since the docs are heavily influx anyways, and changes only affect Next, this would work from my POV.

dominiklohmann (approved):
I really like the new separation between guides and reference, I think this is a good improvement.

Let's get this in and then do the remaining improvements based on main to avoid further merge conflicts.

PR #3219, opened 2023-06-13
Include `pipeline-manager` TTL mechanism

Dakostu:
This change includes a scaffolding for `vast-plugin`  to include a working TTL mechanism.

### DIFF
--- a/libvast/src/http_api.cpp
+++ b/libvast/src/http_api.cpp
@@ -11,10 +11,8 @@
 #include <vast/http_api.hpp>
 
 #include <fmt/format.h>
-#include <simdjson/dom/array.h>
-#include <simdjson/dom/element.h>
-#include <simdjson/dom/parser.h>
-#include <simdjson/error.h>
+
+#include <simdjson.h>
 
 namespace vast {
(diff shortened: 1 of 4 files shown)


libvast/src/http_api.cpp
```diff
@@ -43,6 +47,29 @@ auto parse_endpoint_parameters(
             return caf::make_error(ec::invalid_argument, "not a boolean value");
           return data{result};
         },
+        [&string_value](const list_type&) -> caf::expected<data> {
```
dominiklohmann, line 50: Please also check that the list type's value type is a string type, i.e.,

```
if (not caf::holds_alternative<string_type>(lt.value_type())) {
  return caf::make_error(...);
}
```

dominiklohmann (approved):

libvast/src/http_api.cpp
Dakostu: Added in https://github.com/tenzir/vast/pull/3219/commits/b8c93d423f5e4910261438079f6aa0f1f9a0cb50

PR #3220, opened 2023-06-14
Try casting builders/fields to a common type (no list/record conversions)

patszt:
The code tries to cast an input value when it's added into a field of a different type.
This casting can fail. E.g we can't convert -240 to a bool_type.

In this PR we try to cast the field's builder to the common type that can accomodate all the current values + the new added one.

Also a casting between duration_type and time_type were removed


libvast/include/vast/detail/series_builders.hpp
```diff
@@ -344,6 +339,63 @@ struct series_builder : series_builder_base {
     }
   }
 
+  template <concrete_type NewType, concrete_type CurrentType, class... Args>
+  auto change_builder(const NewType& new_value_type, auto value_to_add,
+                      auto array, const CurrentType& curr_type,
+                      type_list<Args...>) -> caf::error {
```
dominiklohmann, line 345: Please document these parameters. It's hard to understand what `Args...` is, for example.

dominiklohmann, line 345: The array should also likely be taken by const-ref. It also seems to always be a `std::shared_ptr<arrow::Array>` exactly, so why make that an inferred type?

libvast/include/vast/detail/type_list.hpp
```diff
@@ -131,4 +132,52 @@ struct tl_make<T<Ts...>> {
...
+    typename tl_remove_types_helper<tl_tail_t<TypeList>, Types...>::type,
+    tl_push_front_t<
+      typename tl_remove_types_helper<tl_tail_t<TypeList>, Types...>::type,
+      tl_head_t<TypeList>>>;
+};
+
+// Remove all of occurrences of the Types from the TypeList
+template <class TypeList, class... Types>
+  requires(is_type_list<TypeList>::value)
+using tl_remove_types_t = tl_remove_types_helper<TypeList, Types...>::type;
```
dominiklohmann, line 181: CAF already has this as `caf::detail::tl_filter_not_type`.

vast/integration/reference/use-schema-time-unit-when-converting-from-a-double-to-a-duration/step_00.ref
```diff
@@ -0,0 +1,2 @@
...ll, "SrcRate": null, "DstRate": null, "SIntPkt": "20.0ns", "SIntDist": null, "SIntPktAct": null, "SIntActDist": null, "SIntPktIdl": null, "SIntIdlDist": null, "SIntPktMax": null, "SIntPktMin": null, "DIntPkt": null, "DIntDist": null, "DIntPktAct": null, "DIntActDist": null, "DIntPktIdl": null, "DIntIdlDist": null, "DIntPktMax": null, "DIntPktMin": null, "SrcJitter": null, "SrcJitAct": null, "SrcJitIdl": null, "DstJitter": null, "DstJitAct": null, "DstJitIdl": null, "Label": null, "srcUdata": null, "dstUdata": null, "SrcWin": null, "DstWin": null, "sVlan": null, "dVlan": null, "sVid": null, "dVid": null, "sVpri": null, "dVpri": null, "SRange": null, "ERange": null, "SrcTCPBase": null, "DstTCPBase": null, "TcpRtt": null, "SynAck": null, "AckDat": null, "TcpOpt": null, "Inode": null, "Offset": null, "sMeanPktSz": null, "dMeanPktSz": null, "sPktSz": null, "sMaxPktSz": null, "dPktSz": null, "dMaxPktSz": null, "sminPktSz": null, "dminPktSz": null, "STcpMax": null, "DTcpMax": null, "dlDur": null, "dlsTime": null, "dllTime": null, "dsPkts": null, "ddPkts": null, "dsBytes": null, "ddBytes": null, "JDelay": null, "LDelay": null, "sEnc": null, "dEnc": null, "IcmpId": null, "schema": "record"}
```
dominiklohmann, line 1: Please change the test for this to only show the relevant columns by putting a `select SIntPkt` into the pipeline. It's otherwise really hard to parse this output.

dominiklohmann (approved):
Mostly minor things. Glad to see this land.

vast/integration/reference/use-schema-time-unit-when-converting-from-a-double-to-a-duration/step_00.ref
patszt: good idea

libvast/include/vast/detail/type_list.hpp
patszt: Ok, i could just tl_filter_not_type_t<tl_filter_not_type_t for two types i guess

libvast/include/vast/detail/series_builders.hpp
patszt: fixed in 3ed557e15972dbeba1297eb6b23724a0bab3132f

patszt: fixed in 3ed557e15972dbeba1297eb6b23724a0bab3132f

PR #3221, opened 2023-06-14
Remove the `prefix()` function from the REST endpoint plugin API

lava:
It turned out to be more complication than it was worth it; in particular it generated a mismatch between an endpoints external and internal name that could lead to bugs.


dominiklohmann (approved):
Thank you for the fix.

PR #3224, opened 2023-06-14
Use the vars context to access org level variables

tobim:
This should fix package uploads from `main` workflow runs.


dominiklohmann (approved):
🙈

PR #3225, opened 2023-06-14
Don't (mis)use response promises in node proxy

lava:
Don't rely on the destructor of `internal_http_response` to deliver a response promise.


libvast/src/node.cpp
```diff
@@ -563,15 +568,17 @@ node(node_actor::stateful_pointer<node_state> self, std::string /*name*/,
         ->request(handler, caf::infinite, atom::http_request_v,
                   endpoint.endpoint_id, std::move(request))
         .then(
-          []() mutable {
-            /* nop */
+          [response, rp]() mutable {
+            rp.deliver(std::exchange(*response, {}).release());
           },
-          [response](const caf::error& e) mutable {
+          [response, rp](const caf::error& e) mutable {
```
tobim, line 574: We don't need a second handle to `response`.

tobim (dismissed):
This should solve the wrong context problem.

libvast/src/node.cpp
```diff
@@ -424,9 +423,11 @@ auto node_state::get_endpoint_handler(const http_request_description& desc)
   for (auto const& endpoint : plugin->rest_endpoints())
     rest_handlers[endpoint.canonical_path()]
       = std::make_pair(handler, endpoint);
+  return rest_handlers.at(desc.canonical_path);
```
tobim, line 426: remove??

libvast/test/rest_proxy.cpp
```diff
@@ -0,0 +1,45 @@
...
+      CHECK_EQUAL(response.code(), size_t{200});
+      auto body = std::move(response).release();
+      auto parsed = vast::from_json(body);
+      CHECK(parsed);
+    },
+    [](const caf::error& e) {
+      FAIL(e);
+    });
+}
+
```
tobim, line 44: It would be good to have tests for that exercise the error paths here too.

libvast/src/node.cpp
```diff
@@ -547,34 +548,25 @@ node(node_actor::stateful_pointer<node_state> self, std::string /*name*/,
...
       auto [handler, endpoint] = self->state.get_endpoint_handler(desc);
       if (!handler)
         return caf::make_error(ec::system_error,
                                "failed to spawn rest handler");
-      auto rp = self->make_response_promise<std::string>();
-      auto response = std::make_shared<detail::internal_http_response>(rp);
+      auto rp = self->make_response_promise<rest_response>();
       auto params = parse_endpoint_parameters(endpoint, desc.params);
       if (!params)
         return caf::make_error(ec::invalid_argument, "invalid parameters");
```
tobim, line 560: I believe we should return a `rest_response` here and on line 555.

tobim (approved):
I think it is a lot easier to follow the control flow now. I have a few comments but I don't see any fundamental issues with the change (again).

libvast/src/node.cpp
tobim: Oh, and you should wait to create the response promise until all the errors are checked.

libvast/include/vast/http_api.hpp
```diff
@@ -100,48 +101,62 @@ struct rest_endpoint {
...
-public:
-  /// Data according to the type of the endpoint.
-  vast::record params;
+private:
+  size_t code_ = 200;
+
+  // TODO: Use a `vast::record` here to enforce JSON-like structure.
+  // This can only be done after we fixed the `status` handling of the
+  // node to use regular response/request messaging.
+  std::string body_ = {};
```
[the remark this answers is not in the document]
lava: Makes sense, although I think almost every other call site already has a record and converts it to a string. Probably it's better to have a constructor that takes a record and additionally a `from_json_string()` function when there's already a finished string.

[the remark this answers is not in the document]
dominiklohmann: That does sound like a really good idea.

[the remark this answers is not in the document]
tobim: Let's do this as a follow up refactoring please.

PR #3229, opened 2023-06-15
Add Kafka connector

mavam:
This PR addes a Kafka loader and saver.

The implementation leverages the official librdkafka library.

```[tasklist]
- [x] Add `kafka` loader
- [x] Add `kafka` saver
- [x] Write connector documentation page
- [x] Use new argument parser framework to parse options
- [x] Pass through librdkafka options
- [x] Fix up packaging and CI
```


mavam:
The heavy lift is done on this. All major consumer options are supported, and offset seeking now works as well. In addition, a `kafka.yaml` at the right location makes it easy to provide any librdkafka option.

The PR needs predominantly testing. @dominiklohmann I'm facing the stalling issue quite frequently, so it probably doesn't make sense to kick off the review round yet.

plugins/kafka/src/producer.cpp
```diff
@@ -0,0 +1,104 @@
...
+  std::string error;
+  result.producer_.reset(RdKafka::Producer::create(config.conf_.get(), error));
+  if (!result.producer_)
+    return caf::make_error(ec::unspecified,
+                           fmt::format("failed to create producer: {}", error));
+  result.config_ = std::move(config);
+  return result;
+}
+
+auto producer::produce(std::string topic, std::span<const std::byte> bytes,
```
dominiklohmann, line 30: Instead of the `const_cast` below, please change the API to take the proper type here already, and let the outermost context decide whether the `const_cast` is safe to do.

In this particular case, it is only safe if this operator has unique ownership of the chunk passed to it. It likely has, but if it does not then you would need to copy it first.

plugins/kafka/src/plugin.cpp
```diff
@@ -0,0 +1,410 @@
...
+  template <class Inspector>
+  friend auto inspect(Inspector& f, saver_args& x) -> bool {
+    return f.object(x)
+      .pretty_name("saver_args")
+      .fields(f.field("topic", x.topic), f.field("key", x.key),
+              f.field("timestamp", x.timestamp));
+  }
+};
+
+class kafka_saver final : public plugin_saver {
```
dominiklohmann, line 250: For the saver, you likely want batches of size 1; I'll write a follow-up story for this.

plugins/kafka/src/producer.cpp
mavam: I've provided reasoning for the `const_cast`.

This is an implementation detail and should _not_ leak out to the API.

plugins/kafka/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+    dependency_summary("librdkafka" RdKafka::rdkafka++ "Dependencies")
+  endif ()
+endif ()
+if (NOT RdKafka_FOUND)
+  if (NOT VAST_ENABLE_BUNDLED_RDKAFKA)
+    message(
+      STATUS
+        "Cannot find installed RdKafka; falling back to bundled version")
+  endif ()
+  add_subdirectory(aux/librdkafka)
```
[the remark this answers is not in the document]
mavam: No. @tobim can you vet the CMake setup and address this thread?

plugins/kafka/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+
+option(VAST_ENABLE_BUNDLED_RDKAFKA "Always use the bundled RdKafka" OFF)
+add_feature_info("VAST_ENABLE_BUNDLED_RDKAFKA" VAST_ENABLE_BUNDLED_RDKAFKA
+                 "always use the bundled RdKafka.")
+if (NOT VAST_ENABLE_BUNDLED_RDKAFKA)
+  find_package(RdKafka QUIET)
+  if (RdKafka_FOUND)
+    string(APPEND VAST_FIND_DEPENDENCY_LIST
+           "\nfind_package(RdKafka REQUIRED)")
+    dependency_summary("librdkafka" RdKafka::rdkafka++ "Dependencies")
```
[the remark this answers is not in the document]
mavam: Thanks, added.

plugins/kafka/src/producer.cpp
mavam: Does the reasoning in the comment work for you?

dominiklohmann: Yes, thank you!

plugins/kafka/src/producer.cpp
```diff
@@ -0,0 +1,118 @@
...
+        // queue.buffering.max.messages and queue.buffering.max.kbytes
+        auto ms = 1000;
+        VAST_WARN("queue full, retrying in {}ms", ms);
+        producer_->poll(ms);
+        break;
+      }
+    }
+  }
+  producer_->poll(0);
+  __builtin_unreachable();
```
jachris, line 90: This looks like you shouldn't be this way.

mavam: Thanks, fixed.

plugins/kafka/src/plugin.cpp
```diff
@@ -0,0 +1,410 @@
...
+    if (args.offset) {
+      if (!offset_parser()(args.offset->inner))
+        diagnostic::error("invalid `--offset` value")
+          .primary(args.offset->source)
+          .note("valid values are:")
+          .note("- beginning")
+          .note("- end")
+          .note("- store")
+          .note("- <value> (absolute offset)")
+          .note("- -<value> (relative offset from end)")
```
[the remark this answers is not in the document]
mavam: Much appreciate the diligence! Fixed.

mavam:
@tobim Please take it from here, I enabled auto-merge.

tobim (approved):
Reviewed the code in person and tested locally.

PR #3230, opened 2023-06-15
Make `load` and `save` implicitly detached

dominiklohmann:
All connectors we currently have should be detached. So instead of making this configurable at the connector level, we instead treat all operators wrapping connectors as detached for now, until someone adds a connector that has different requirements.

### DIFF
--- a/libvast/builtins/operators/load.cpp
+++ b/libvast/builtins/operators/load.cpp
@@ -36,6 +36,10 @@ public:
     return operator_location::local;
   }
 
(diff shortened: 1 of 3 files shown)


jachris (approved):

tobim:
Does this unblock #3130?

dominiklohmann:
Yes, although you should wait for #3223 before picking that up again because of the API changes that PR brings with it.

PR #3231, opened 2023-06-16
Implement '--disable-timestamp-tags' long-form parameter for zeek-tsv printer

Dakostu:
This change adds a new TQL parser argument `--disable-timestamp-tags` to the `zeek-tsv` printer.

### DIFF
--- a/libvast/builtins/formats/zeek_tsv.cpp
+++ b/libvast/builtins/formats/zeek_tsv.cpp
@@ -665,7 +665,7 @@ public:
     std::optional<char> set_sep;
     std::optional<std::string> empty_field;
     std::optional<std::string> unset_field;
-    bool disable_timestamp_tags;
+    bool disable_timestamp_tags = false;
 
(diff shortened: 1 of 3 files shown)


dominiklohmann:
I've changed the reviewer to @jachris because of the potential merge conflict with #3223.

jachris (approved):
Thank you!

I have only one question: This PR does not actually remove the `disable-timestamp-tag` config option. It still exists for the CLI, three config files, and in `libvast/src/format/zeek.cpp`. Is this intended?

Dakostu:
> I have only one question: This PR does not actually remove the `disable-timestamp-tag` config option. It still exists for the CLI, three config files, and in `libvast/src/format/zeek.cpp`. Is this intended?

They are still used by `tenzir-ctl export/import zeek`. Why should it be removed now?

jachris:
Okay, I thought so. Then everything is fine.

Dakostu:
These, let's call them, legacy settings will be soon gone in one big swoop, I hope.

PR #3237, opened 2023-06-19
Prevent a memory leak from disabled plugins

tobim:
Disabled plugins are erased from the plugin list, but `std::vector::erase` does not call the destructor of the exact element that gets erased, but the moved-from ones at the end. This breaks dynamic plugin management.

The correct fix for this is to clean up resources owned resources in the move assignment operator of `plugin_ptr`. Failing to do so makes them unreachable otherwise.

With that, the call to `erase` works as expected because it moves the subsequent entries on top of the one that gets erased.

### DIFF
--- a/libvast/src/plugin.cpp
+++ b/libvast/src/plugin.cpp
@@ -666,18 +666,7 @@ plugin_ptr plugin_ptr::make_builtin(plugin* instance, void (*deleter)(plugin*),
 plugin_ptr::plugin_ptr() noexcept = default;
 
 plugin_ptr::~plugin_ptr() noexcept {
-  if (instance_) {
-    VAST_ASSERT(deleter_);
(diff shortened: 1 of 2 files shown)


libvast/src/plugin.cpp
```diff
@@ -690,6 +684,8 @@ plugin_ptr::plugin_ptr(plugin_ptr&& other) noexcept
 }
 
 plugin_ptr& plugin_ptr::operator=(plugin_ptr&& rhs) noexcept {
+  // Clean up *this* to prevent leaking an instance_.
+  this->~plugin_ptr();
```
dominiklohmann, line 688: This must not call the destructor, as per UBSan's `-fsanitize=lifetime`.

dominiklohmann (approved):

PR #3240, opened 2023-06-19
Delay JSON parsing of incoming parameter data

lava:
A client forwarding incoming requests to the node will generally
    not be able to parse all request parameters to the correct
    types declared by the endpoint. Therefore, for POST requests
    we now have a way of forwarding the raw JSON body to the node.
    
This also implies the need to split up JSON body parameters and
    other parameters in the http_request_description, since the
    parsing rules for the latter are slightly different.

Also, increase test coverage of our unit tests.

### DIFF
--- a/libvast/include/vast/http_api.hpp
+++ b/libvast/include/vast/http_api.hpp
@@ -17,6 +17,7 @@
 #include <caf/expected.hpp>
 #include <caf/optional.hpp>
 
+#include <simdjson.h>
 #include <string>
 
 namespace vast {
@@ -73,8 +74,9 @@ struct rest_endpoint {
   /// Path can use the express.js conventions
   std::string path;
 
-  /// Expected parameters.
-  //  (A record_type cannot be empty, so we need an optional)
+  /// Expected parameters, if any.
+  // Note that the node will currently only forward basic types and
+  // lists of basic types as parameters.
   std::optional<vast::record_type> params;
 
   /// Version for that endpoint.
@@ -142,13 +144,6 @@ private:
   caf::error detail_ = {};
 };
 
-/// Go through the provided parameters; discard those that are not understood by
-/// the endpoint and attempt to parse the rest to the expected type.
-auto parse_endpoint_parameters(
-  const vast::rest_endpoint& endpoint,
-  const detail::stable_map<std::string, std::string>& params)
-  -> caf::expected<vast::record>;
-
 /// Used for serializing an incoming request to be able to send it as a caf
 /// message.
 class http_request_description {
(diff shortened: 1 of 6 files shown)


libvast/include/vast/http_api.hpp
```diff
@@ -17,6 +17,7 @@
 #include <caf/expected.hpp>
 #include <caf/optional.hpp>
 
+#include <simdjson.h>
```
tobim, line 20: Not necessary.

libvast/include/vast/http_api.hpp
```diff
@@ -142,34 +144,95 @@ struct rest_response {
...
 class http_request_description {
 public:
   /// Unique identification of the request endpoint.
   std::string canonical_path;
 
-  /// Request parameters. (unvalidated)
-  /// Query parameters, path parameters, and JSON body parameters
-  /// all get merged into this map.
+  /// Query parameters, path parameters, x-www-urlencoded body parameters
   detail::stable_map<std::string, std::string> params;
```
tobim, line 155: Let's comment that out as long as we aren't using it. It is currently untested and offering it gives a false assurance.

libvast/include/vast/http_api.hpp
```diff
@@ -142,34 +144,95 @@ struct rest_response {
...
+  static auto from_json(std::string_view) -> caf::expected<http_parameter_map>;
+
+  /// Access to the internal data.
+  [[nodiscard]] auto params() const
+    -> const detail::stable_map<std::string, vast::data>&;
+
+  /// Insert a new key and value.
+  auto emplace(std::string&& key, vast::data&& value) -> void;
+
+  /// Unchecked access to the internal data.
```
tobim, line 187: Add "For unit tests only!". or something.

libvast/include/vast/http_api.hpp
```diff
@@ -142,34 +144,95 @@ struct rest_response {
...
+/// Go through the provided parameters; discard those that are not understood by
+/// the endpoint and attempt to parse the rest to the expected type.
+auto parse_endpoint_parameters(const vast::rest_endpoint& endpoint,
+                               const http_parameter_map& params)
+  -> caf::expected<vast::record>;
+
+/// Parse from JSON data.
+auto parse_skeleton(simdjson::ondemand::value value, size_t depth = 0)
+  -> vast::data;
+
```
tobim, line 235: Not needed any more.

libvast/src/http_api.cpp
```diff
@@ -21,64 +21,150 @@ auto rest_endpoint::canonical_path() const -> std::string {
...
-          auto l = list{};
-          for (const auto& x : obj.value()) {
-            if (x.type() != ::simdjson::dom::element_type::STRING) {
-              return caf::make_error(ec::invalid_argument,
-                                     "currently only string values in arrays "
-                                     "are accepted");
-            }
-            l.emplace_back(std::string{x.get_string().value()});
+        [&](const list_type& lt) -> caf::expected<data> {
+          if (!is_list)
```
tobim, line 126: Let's replace that with the definition.

libvast/src/http_api.cpp
```diff
@@ -21,64 +21,150 @@ auto rest_endpoint::canonical_path() const -> std::string {
...
+            return caf::make_error(ec::invalid_argument, "expected a list");
+          auto list = caf::get<vast::list>(param_data);
```
tobim, line 128: We can also just use `caf::get_if` instead.

libvast/src/http_api.cpp
```diff
@@ -21,64 +21,150 @@ auto rest_endpoint::canonical_path() const -> std::string {
...
   for (auto const& leaf : endpoint.params->leaves()) {
     auto const& name = leaf.field.name;
     auto maybe_param = params.find(name);
     if (maybe_param == params.end())
       continue;
-    auto const& string_value = maybe_param->second;
+    auto const& param_data = maybe_param->second;
+    auto is_string = caf::holds_alternative<std::string>(param_data);
+    auto is_list = caf::holds_alternative<vast::list>(param_data);
     auto typed_value = caf::visit(
```
tobim, line 109: Lets add a `TODO` to signal that we want to allow more possible endpoint signatures and use a recursive algorithm here.

libvast/test/rest_proxy.cpp
```diff
@@ -21,12 +21,40 @@ struct fixture : public fixtures::node {
...
 
+TEST(parameter parsing) {
+  auto endpoint = vast::rest_endpoint {
+    .method = vast::http_method::post,
+    .path = "/dummy",
+    .params = vast::record_type{
+      {"id", vast::int64_type{}},
+      {"uid", vast::uint64_type{}},
+      {"timeout", vast::duration_type{}},
+      {"value", vast::string_type{}},
```
tobim, line 32: Should also have the supported list types.

tobim (approved):
We verified that this works for the existing endpoints.

PR #3247, opened 2023-06-20
Update google storage bucket names in demo node entrypoint

lava:


mavam (commented):
Should we also switch to the new `import` pipeline sink operator?

Also: in our getting started instructions we pre-filter the Suricata logs and only take alerts. See https://docs.tenzir.com/docs/next/get-started#import-data-into-a-node

tobim:
> Should we also switch to the new import pipeline sink operator?
>
> Also: in our getting started instructions we pre-filter the Suricata logs and only take alerts. See https://docs.tenzir.com/docs/next/get-started#import-data-into-a-node

Let's iterate in a different PR, this here fixes the image and it should not get delayed.

tobim (approved):
The new URLs work.

PR #3248, opened 2023-06-20
Add installer shell script

mavam:
This PR adds an installer shell script so that installing Tenzir will be as simple as:

```bash
curl -L get.tenzir.app | sh
```

This supersedes #2476, a more complex attempt that was Homebrew-inspired.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -173,7 +173,6 @@ then
   echo
   echo "  - ${cmd1}"
   echo "  - ${cmd2}"
-  echo
   read -r dummy_
   action "Installing via dpkg"
   eval "${cmd1}"
@@ -185,7 +184,6 @@ then
   echo "Press ${green}ENTER${normal} to continue with the following commands:"
   echo
   echo "  - ${cmd1}"
-  echo
   action "Unpacking tarball"
   eval "${cmd1}"
 fi
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -143,37 +143,51 @@ echo "(You can skip these steps, but then only open source features will be"
 echo "available. Visit ${bold}https://tenzir.com/pricing${normal} for a" \
   "feature comparison.)"
 echo
-echo "Press ${green}ENTER${normal} to proceed with the installation."
+echo "Press ${green}ENTER${normal} to continue."
 read -r dummy_
 
 # Check for platform configuration.
+action "Checking for existence of platform configuration"
 if ! [ -f "${config}" ]
 then
+  echo "Could not find config at ${bold}${config}${normal}."
+  action "Using open source feature set"
   open_source=1
-fi
-if [ -n "${open_source}" ]
-then
-  echo "Could not find platform config at ${bold}${config}${normal}."
-  action "Using open source edition"
+else
+  echo "Found config at ${bold}${config}${normal}."
+  action "Using complete feature set"
 fi
 
 # Trigger installation.
+action "Installing package into ${prefix}"
 if ! check sudo
 then
   echo "Could not find ${bold}sudo${normal} in \$PATH."
   exit 1
 fi
-action "Installing package into ${prefix}"
 if [ "${platform}" = "Debian" ]
 then
+  cmd1="sudo dpkg -i \"${tmpdir}/${package}\""
+  cmd2="sudo systemctl status tenzir"
+  echo "Press ${green}ENTER${normal} to continue with the following commands:"
+  echo
+  echo "  - ${cmd1}"
+  echo "  - ${cmd2}"
+  echo
+  read -r dummy_
   action "Installing via dpkg"
-  sudo dpkg -i "${tmpdir}/${package}"
+  eval "${cmd1}"
   action "Checking node status"
-  sudo systemctl status tenzir
+  eval "${cmd2}"
 elif [ "${platform}" = "Linux" ]
 then
+  cmd1="sudo tar xzf \"${tmpdir}/${package}\" -C /"
+  echo "Press ${green}ENTER${normal} to continue with the following commands:"
+  echo
+  echo "  - ${cmd1}"
+  echo
   action "Unpacking tarball"
-  sudo tar xzf "${tmpdir}/${package}" -C /
+  eval "${cmd1}"
 fi
 
 # Test the installation.
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -111,13 +111,15 @@ base="https://github.com/tenzir/tenzir/releases/latest/download"
 url="${base}/${package}"
 tmpdir="$(dirname "$(mktemp -u --tmpdir)")"
 action "Downloading ${url}"
-if check wget; then
+if check wget
+then
   wget -q --show-progress -O "${tmpdir}/${package}" "${url}"
-elif check curl; then
+elif check curl
+then
   curl --progress-bar -L -o "${tmpdir}/${package}" "${url}"
 else
   echo "Neither ${bold}wget${normal} nor ${bold}curl${normal}" \
-    "in ${bold}\$PATH${normal}, aborting"
+    "found in \$PATH."
   exit 1
 fi
 echo "Successfully downloaded ${package}"
@@ -152,17 +154,22 @@ then
 fi
 
 # Trigger installation.
+if ! check sudo
+then
+  echo "Could not find ${bold}sudo${normal} in \$PATH."
+  exit 1
+fi
 action "Installing package into ${prefix}"
 if [ "${platform}" = "Debian" ]
 then
   action "Installing via dpkg"
-  dpkg -i "${tmpdir}/${package}"
+  sudo dpkg -i "${tmpdir}/${package}"
   action "Checking node status"
-  systemctl status tenzir
+  sudo systemctl status tenzir
 elif [ "${platform}" = "Linux" ]
 then
   action "Unpacking tarball"
-  tar xzf "${tmpdir}/${package}" -C /
+  sudo tar xzf "${tmpdir}/${package}" -C /
 fi
 
 # Test the installation.
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -130,7 +130,7 @@ echo "available. Visit ${bold}https://tenzir.com/pricing${normal} for a" \
   "feature comparison.)"
 echo
 echo "Press ${green}ENTER${normal} to proceed with the installation."
-read -r
+read -r dummy_
 
 # Check for platform configuration.
 if ! [ -f "${config}" ]
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -6,6 +6,9 @@
 # Treat unset variables as an error when substituting.
 set -u
 
+# Abort on error.
+set -e
+
 #
 # Helper utilities
 #
@@ -147,7 +150,7 @@ then
   action "Installing via dpkg"
   dpkg -i "${package}"
   action "Checking node status"
-  systemctl status tenzir || exit 1
+  systemctl status tenzir
 elif [ "${platform}" = "Linux" ]
 then
   action "Unpacking tarball"
@@ -157,7 +160,7 @@ fi
 # Test the installation.
 action "Checking version"
 PATH="${prefix}:$PATH"
-tenzir -q 'version | select version | write json' || exit 1
+tenzir -q 'version | select version | write json'
 
 # Inform about next steps.
 action "Providing guidance"
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -109,7 +109,7 @@ echo "Using ${package}"
 # Download package.
 base="https://github.com/tenzir/tenzir/releases/latest/download"
 url="${base}/${package}"
-tmpdir="$(mktemp -u)"
+tmpdir="$(dirname "$(mktemp -u)")"
 action "Downloading ${url}"
 if check wget
 then
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -42,12 +42,6 @@ action() {
 # Execution
 #
 
-if [ -z "${BASH_VERSION:-}" ]
-then
-  echo "Bash is required to interpret this script."
-  exit 1
-fi
-
 # Print welcome banner.
 echo "${blue}"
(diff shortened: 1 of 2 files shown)


scripts/install.sh
```diff
@@ -0,0 +1,139 @@
...
+then
+  PACKAGE="tenzir-linux-static.deb"
+elif [ "${PLATFORM}" = "Linux" ]
+then
+  PACKAGE="tenzir-linux-static.tar.gz"
+elif [ "${PLATFORM}" = "NixOS" ]
+then
+  echo "Try Tenzir with our ${bold}flake.nix${normal}:"
+  echo
+  echo "    ${bold}nix run github:tenzir/tenzir/stable${normal}"
```
dit7ya, line 87: The https://github.com/tenzir/tenzir/tree/stable branch is 500+ commits behind, what is our plan with that @tobim ?

Also, is our cachix working? I tried 

```nix run github:tenzir/tenzir/stable```

but it is building stuff from source. (I have vast cachix enbled).

**Review by dit7ya — CHANGES_REQUESTED** [review]

**Comment by rdettai** [comment]
Following up on https://github.com/tenzir/issues/issues/472#issuecomment-1597476595:

What I would run in CI:
- a matrix with `ubuntu-22.04`, `ubuntu-20.04`, `macos-12`, `macos-11` (if all these are meant to be supported)
- in the job:
  - run the exact command you document, e.g `curl -L get.tenzir.app | sh`
  - run the commands that are documented in the script output, i.e:
```
echo "You're all set! Next steps:"
echo "  - Run a pipeline via ${blue}tenzir <pipeline>${normal}"
echo "  - Spawn a node via ${blue}tenzir-node${normal}" 
```

File: scripts/install.sh (hunk 1 above)
Comment (mavam) [reply scripts/install.sh:87]: It sounds like we want a story for this, to be addressed post alpha testing. As long as the instructions are sound, we can focus on the implementation of this separately.

Comment (dit7ya) [reply scripts/install.sh:87]: Cache not working needs a story, yes. But the correct release binary should be part of alpha.

**Comment by mavam** [comment]
> a matrix with ubuntu-22.04, ubuntu-20.04, macos-12, macos-11 (if all these are meant to be supported)

It would obviously be nice to cover all platforms. But for macOS, we don't have packages yet. Only for Linux. Can we get a Debian and non-Debian from GitHub?

> run the exact command you document, e.g `curl -L get.tenzir.app | sh`

@rdettai this won't work because the installer lives in the repo. You will never get that version that lives in the PR. That said, I don't think it's important that we go through the DNS lookup and CNAME redirect. We can simply invoke `install.sh` from CI.

**Comment by rdettai** [comment]
> Can we get a Debian and non-Debian from GitHub
I think the way to do this would be through a matrix of docker images: https://github.com/orgs/community/discussions/25079#discussioncomment-3246416

>  this won't work because the installer lives in the repo. You will never get that version that lives in the PR. That said, I don't think it's important that we go through the DNS lookup and CNAME redirect. We can simply invoke install.sh from CI

We could also decouple the two:
- have a test ensuring that the DNS setup works by just ensuring we get a 200 with something that looks like a script in it
- run install.sh

That being said, if the DNS setup is fully exogenous from the code base, it should also be probed separately. It would rather go into a sort of health bot that goes through all Tenzir resources (docs, website...) and checks their health (no dead links...). So I think that it would be enough for this PR to validate that the script works.

File: scripts/install.sh (hunk 1 above)
Comment (dominiklohmann) [reply scripts/install.sh:87]: I would assume that the cache is broken because of the repository-renaming.

We usually update the `stable` branch when we cut a release. So at least for the Alpha, we would need a different mechanism here.

**Comment by dominiklohmann** [comment]
> Does this go into the `vast.yaml` workflow? Or in a separate file? (cc @dominiklohmann)

A separate workflow because it has separate triggers.

In any case, what's the urgency of having this tested in CI? Can we add that at a later point in time? I think it's not quite so simple before we have a new release.

**Comment by mavam** [comment]
Agreed, lets' factor the CI discussion out of this PR. Get this merged in a working version, and then tackle the topic later.

File: scripts/install.sh (hunk 1 above)
Comment (tobim) [reply scripts/install.sh:87]: > Also, is our cachix working? I tried
>
> nix run github:tenzir/tenzir/stable
>
> but it is building stuff from source. (I have vast cachix enbled).

We can't use cachix for closed source code, since we are still on the free version. (poke @mavam)

Comment (mavam) [reply scripts/install.sh:88]: I'm resolving this thread, as the Cachix discussion takes place elsewhere.

**Comment by mavam** [comment]
Requesting review from @dominiklohmann for general onboarding flow and @tobim for ops details.

File: scripts/install.sh [hunk 2]
```diff
@@ -0,0 +1,178 @@
...
+  echo "Could not find platform config at ${bold}${CONFIG}${normal}."
+  action "Using open source edition"
+fi
+
+# Trigger installation.
+action "Installing package into ${PREFIX}"
+if [ "${PLATFORM}" = "Debian" ]
+then
+  action "Installing via dpkg"
+  dpkg -i "${PACKAGE}"
```
tobim, line 148: If I run this as a regular user it fails, but continues with the script, which then fails more. We should `set -e` in the script and we should `sudo` here.

scripts/install.sh
```diff
@@ -0,0 +1,178 @@
...
+echo "  1. Go to ${bold}https://app.tenzir.com${normal}"
+echo "  2. Sign up for a free account"
+echo "  3. Configure a node and download a ${bold}platform.yaml${normal} config"
+echo "  4. Move the config to ${bold}${CONFIG}${normal}"
+echo
+echo "(You can skip these steps, but then only open source features will be"
+echo "available. Visit ${bold}https://tenzir.com/pricing${normal} for a" \
+  "feature comparison.)"
+echo
+echo "Press ${green}ENTER${normal} to proceed with the installation."
```
tobim, line 129: We need to tell the user what is going to happen next here, depending on the detected platform.

Something like
```suggestion
echo "I will now run \`sudo dpkg -i tenzir-linux-static.deb\` to finish the installation."
echo "Press ${green}ENTER${normal} to proceed."
```

tobim (changes requested):
As a stylistic suggestion: Script variables that are not exported as environment variables should be all lowercase instead of ALLCAPS.

scripts/install.sh
```diff
@@ -0,0 +1,178 @@
...
+  echo "Visit ${bold}https://docs.tenzir.com${normal} for further instructions."
+  exit 1
+fi
+echo "Using ${PACKAGE}"
+
+# Download package.
+BASE="https://github.com/tenzir/tenzir/releases/latest/download"
+URL="${BASE}/${PACKAGE}"
+action "Downloading ${URL}"
+curl --progress-bar -L "${URL}" -o "${PACKAGE}" || exit 1
```
tobim, line 111: `curl` is not installed by default in Debian, but `wget` is. I suspect that is the case in some other distros as well.
```suggestion

check() {
  command -v "$1" 2>&1 >/dev/null
}

# Download package.
tmpdir="$(dirname "$(mktemp -u --tmpdir)")"
BASE="https://github.com/tenzir/tenzir/releases/latest/download"
URL="${BASE}/${PACKAGE}"
action "Downloading ${URL}"
if check quizf; then
  exit 1
elif check wget; then
  wget -q --show-progress -O "${tmpdir}/${PACKAGE}" "${URL}"
elif check curl; then
  curl --progress-bar -L -o "${tmpdir}/${PACKAGE}" "${URL}"
else
  echo "Neither 'wget' nor 'curl' in \$PATH, aborting..."
  exit 1
fi
```

scripts/install.sh
tobim, line 148:
```suggestion
  sudo dpkg -i "${tmpdir}/${PACKAGE}"
```

scripts/install.sh
```diff
@@ -0,0 +1,178 @@
...
+fi
+
+# Trigger installation.
+action "Installing package into ${PREFIX}"
+if [ "${PLATFORM}" = "Debian" ]
+then
+  action "Installing via dpkg"
+  dpkg -i "${PACKAGE}"
+  action "Checking node status"
+  systemctl status tenzir || exit 1
```
tobim, line 150:
```suggestion
  sudo systemctl status tenzir || exit 1
```

scripts/install.sh
```diff
@@ -0,0 +1,178 @@
...
+elif [ "${PLATFORM}" = "Linux" ]
+then
+  action "Unpacking tarball"
+  tar xzf "${PACKAGE}" -C /
```
tobim, line 154:
```suggestion
  sudo tar xzf "${tmpdir}/${PACKAGE}" -C /
```

scripts/install.sh
mavam: `quizf` doesn't exist on macOS. (Never heard of it before.) I think `curl` and `wget` are enough, no?

scripts/install.sh
mavam: Should we trap EXIT and delete `tmpdir`?

scripts/install.sh
mavam: I've switched to `wget` and `curl`. Feel free to resolve if that suits the bill.

mavam:
@tobim I've implemented your requests. A few things are noteworthy:

1. I simplified `mktemp` because `--tempdir` is not a thing on macOS
2. I don't think we need to announce what will happen, but rather be transparent what happens. Otherwise we'd need user input on everything.
3. I ignored `quizf` because I've never heard of it.

scripts/install.sh
tobim: That was a copy-n-paste mistake. I just wanted to make sure the code worked as intended before posting the suggestion.

scripts/install.sh
```diff
@@ -0,0 +1,197 @@
...
+then
+  action "Installing via dpkg"
+  sudo dpkg -i "${tmpdir}/${package}"
+  action "Checking node status"
+  sudo systemctl status tenzir
+elif [ "${platform}" = "Linux" ]
+then
+  action "Unpacking tarball"
+  sudo tar xzf "${tmpdir}/${package}" -C /
+fi
```
dominiklohmann, line 173: These two behave quite differently. The former should tell the user that a systemd service was set up for them, and what to do next.

scripts/install.sh
```diff
@@ -0,0 +1,197 @@
...
+    then
+        platform=NixOS
+    else
+        platform=Linux
+    fi
+elif [ "${os}" = "Darwin" ]
+then
+  platform=macOS
+fi
+echo "Found ${platform}."
```
dominiklohmann, line 75: If platform is empty at this point then you should exit with a helpful error message.

dominiklohmann (commented):
I mostly looked at the flow and had some minor things that stood out to me. I'll leave the final approval to @tobim.

tobim (approved):
> I don't think we need to announce what will happen, but rather be transparent what happens. Otherwise we'd need user input on everything.

My suggestion does not add more steps, just a little bit of transparency so the user has more information to decide whether they want to `<enter>` or `<ctrl-c>`.

scripts/install.sh
```diff
@@ -0,0 +1,197 @@
...
+  echo
+  echo "Visit ${bold}https://docs.tenzir.com${normal} for further instructions."
+  exit 1
+fi
+echo "Using ${package}"
+
+# Download package.
+base="https://github.com/tenzir/tenzir/releases/latest/download"
+url="${base}/${package}"
+tmpdir="$(mktemp -u)"
```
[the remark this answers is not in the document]
mavam: Makes sense, fixed that.

scripts/install.sh
mavam: `uname -s` doesn't return anything empty on anything reasonable we support.

mavam: We're dispatching on it in the next block and bail out if there's no appropriate package, so this is handled correctly.

scripts/install.sh
```diff
@@ -0,0 +1,197 @@
+#!/bin/sh
+#
+# This script installs the latest release of Tenzir.
+#
+
+# Treat unset variables as an error when substituting.
+set -u
+
+# Abort on error.
+set -e
```
dominiklohmann, line 10: If you can use bash rather than sh, it would be helpful to have `set -euo pipefail` here. That usually helps a lot with checking correctness.

scripts/install.sh
mavam: Yes, that's unavoidable. We have a guidance section at the end where we should provide specific instructions on what to do next. The text in the middle is not a good place to provide instructions.

I'm already mentioning there that you should start a node if you've used a tarball. I don't think we should mention systemd if it might not be there. We can assume that users can take it from here IMO.

scripts/install.sh
mavam: We're not using Bash on purpose.

scripts/install.sh
dominiklohmann: To program this a bit more defensively, could you add an `else` that prints a logic error and exits or an assertion here? That will make that immediately visible to the reader of this code.

scripts/install.sh
dominiklohmann: The difference is that one starts a node, and the other does not. Can we unify that behavior somehow?

scripts/install.sh
dominiklohmann: Why is that? `/usr/bin/env bash` is available basically everywhere.

mavam: There's no need in using Bash just for `pipefail`. POSIX Shell makes the script far more accessible.

scripts/install.sh
mavam: Added that now.

scripts/install.sh
mavam: I don't think so. It's expected to start a service after `dpkg -i`, but we can't just start a node with service manager that we don't know for the generic case. Instructions are the only way to go here.

scripts/install.sh
mavam: Resolving this for now. We can reconsider Bash if we need more advanced scripting fu.

scripts/install.sh
mavam: We'll improve the messaging over time. Let's start with what we have an wait for feedback.

PR #3252, opened 2023-06-21
Fix JSON parser --no-infer option

patszt:
It seems that --no-infer doesn't work with suricata/zeek-json parsers after pipeline parsing refactor.
This PR fixes this issue + fixes a bug in the series_builder which allowed fields to be cast into a common_type when --no-infer was enabled.

### DIFF
--- a/libvast/builtins/formats/json.cpp
+++ b/libvast/builtins/formats/json.cpp
@@ -516,6 +516,11 @@ struct config {
   }
 };
 
+auto add_common_options_to_parser(argument_parser& parser, config& cfg)
+  -> void {
+  parser.add("--no-infer", cfg.no_infer);
+}
+
 class json_parser final : public plugin_parser {
 public:
   json_parser() = default;
@@ -611,7 +616,7 @@ public:
                                           "formats/json"};
     parser.add("--selector", selector, "<selector>");
     parser.add("--unnest-separator", cfg.unnest_separator, "<separator>");
-    parser.add("--no-infer", cfg.no_infer);
+    add_common_options_to_parser(parser, cfg);
     parser.parse(p);
     if (selector) {
       cfg.selector = parse_selector(selector->inner, selector->source);
@@ -640,8 +645,13 @@ public:
 
   auto parse_parser(parser_interface& p) const
     -> std::unique_ptr<plugin_parser> override {
-    argument_parser{name()}.parse(p);
+    auto parser
+      = argument_parser{name(), fmt::format("https://vast.io/docs/next/"
+                                            "formats/{}",
+                                            name())};
     auto cfg = config{};
+    add_common_options_to_parser(parser, cfg);
+    parser.parse(p);
     cfg.selector = parse_selector(Selector.str(), location::unknown);
     cfg.unnest_separator = Separator.str();
     return std::make_unique<json_parser>(std::move(cfg));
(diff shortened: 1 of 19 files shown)


libvast/builtins/formats/json.cpp
```diff
@@ -516,6 +516,11 @@ struct config {
   }
 };
 
+auto add_common_options_to_parser(argument_parser& parser, config& cfg)
```
patszt, line 519: A ndjson option will also be supported by a suricata/zeek-json and json in the next PR. This is why i have a function that handles all common options for all json alike parsers

libvast/src/detail/series_builders.cpp
```diff
@@ -278,6 +277,7 @@ series_builder::series_builder(const vast::type& type, bool are_fields_fixed)
       },
     },
     type);
+  can_change_builder_type_ = not are_fields_fixed;
```
patszt, line 280: In the caf::visit we assign a concrete_type builder to *this. This triggers a default construction of can_change_builder_type_. This would override the false into true and allow the common_type casting.

libvast/src/detail/series_builders.cpp
```diff
@@ -330,7 +330,7 @@ auto series_builder::remove_last_row() -> void {
     [this](auto& actual) {
       if constexpr (std::is_same_v<bool, decltype(actual.remove_last_row())>) {
         if (auto all_values_are_null = actual.remove_last_row();
-            all_values_are_null) {
+            all_values_are_null and can_change_builder_type_) {
```
patszt, line 333: If we can't change the builder type, then it means we have a fixex type for a builder. There is no need to switch it back into an unknown type and allow it to infer

patszt: I will be adding additional unit tests for this branch in the next commit

libvast/builtins/formats/json.cpp
dominiklohmann: I think `suricata` and `zeek-json` should always be NDJSON. Their format is specified to be that, so there's little need to use something else as a default.

patszt: Ok. I think the function is still fine to be left as it is. We may have more common options in the future.

libvast/builtins/formats/json.cpp
```diff
@@ -640,8 +645,13 @@ class selector_parser final : public virtual parser_parser_plugin {
 
   auto parse_parser(parser_interface& p) const
     -> std::unique_ptr<plugin_parser> override {
-    argument_parser{name()}.parse(p);
+    auto parser
+      = argument_parser{name(), fmt::format("https://vast.io/docs/next/"
+                                            "formats/{}",
```
dominiklohmann, line 650: This link must be adapted to https://docs.tenzir.com/next/formats/json

dominiklohmann (approved):
These changes make sense. Just one quick change request.

libvast/src/detail/series_builders.cpp
dominiklohmann: This makes sense.

libvast/src/detail/series_builders.cpp
dominiklohmann: Another alternative would be to make this an argument of the concrete type builder, but I don't have too strong of an opinion on this.

libvast/builtins/formats/json.cpp
patszt: Ok. i Have also changed other occurrences in the code as this was no effort just replacing them

dominiklohmann: Thanks, that's fantastic!

PR #3254, opened 2023-06-21
Introduce ndjson mode for json parser

patszt:
Blocked by: https://github.com/tenzir/tenzir/pull/3252

If the input is NDJSON then we can handle malformed JSON errors in a more sophisticated way.
Given JSON
`
{"a}.?>
{"b" : "d"}
`

We would get an simdjson_error from iterate_many. The parser would give us no information about how many bytes are malformed. From it's PoV the whole input is bad as it expects the well formed JSON documents on the input. We can't always assume that end of a line is a separator of a documents. 

Parsing the first line as JSON we can discard it as it is malformed one.
Handling the second line as a separate JSON document allows us to parse this event properly.

Zeek-json and suricata parsers use --ndjson option by default.

I didn't give a much thought about optimizing the memory allocation when parsing line by line. Currently i just reuse additional padded_buffer for it. I think it could be sufficent, but maybe someone has an idea how to improve it if this is critical

### DIFF
--- a/libvast/builtins/formats/json.cpp
+++ b/libvast/builtins/formats/json.cpp
@@ -32,6 +32,56 @@ namespace vast::plugins::json {
 
 namespace {
 
+/// A variant of *to_lines* that returns a string view with additional padding
+/// bytes that are safe to read.
+inline auto to_padded_lines(generator<chunk_ptr> input)
+  -> generator<std::optional<simdjson::padded_string_view>> {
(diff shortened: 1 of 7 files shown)


dominiklohmann (approved):
As discussed in Slack, I pushed my review feedback directly to the branch. Thanks for implementing this. I was really happy to see `to_padded_lines` working so well.

PR #3255, opened 2023-06-21
Push the tenzir-ce image to ECR

tobim:


lava (approved):
It doesn't hurt to merge this, but iiuc we actually want to push these to dockerhub, otherwise our tenant-manager generated config still won't work.

PR #3259, opened 2023-06-23
Adjust Plausible website tracking

mavam:
The primary change in this PR is updating the `data-domain` for Plausible, so that we get correct statistics at now <https://plausible.io/docs.tenzir.com>.


dit7ya (approved):
We should check if everything is working as expected in a few days.

PR #3260, opened 2023-06-23
Split schemas for events with selector field but no schema

patszt:
Currently the events with no matching schema pointed by a selector field are combined together.
They should be split


jachris (approved):
Thank you, this looks good to me, the summarize test results changed as expected :+1: 

Can you rebase on top of `main` and update the tests? This should add some additional warnings to the reference files.

jachris:
Seems like the diagnostics are interleaved with the data in the test output? I'll investigate this, we might have to flush `stdout`.

patszt:
> Seems like the diagnostics are interleaved with the data in the test output? I'll investigate this, we might have to flush `stdout`.

Seems like something is off for me locally

patszt:
I think a final check by @jachris is needed.

jachris:
> Seems like something is off for me locally

Did your test runs actually produce the interleaved output or not? If they did, then this surely something that we should fix very soon (but this PR would still be fine as-is).

patszt:
> > Seems like something is off for me locally
> 
> Did your test runs actually produce the interleaved output or not? If they did, then this surely something that we should fix very soon (but this PR would still be fine as-is).

I just ran the vast exec locally and that was the output that was given to me. Could there be a difference between running it on my own VS in the test?

jachris:
> Could there be a difference between running it on my own VS in the test?

It sounds like a buffering issue, but I'm not sure why you would run into this. Do the tests still cause issues for you on `main`? If so, can you try flushing `stdout` at the start of `dump_diagnostics_to_stdout`  in `exec.cpp`?

patszt:
> 
Not sure about the tests, but i do run the test's commands in the terminal
The std::cout.flush() doesnt' help

jachris:
This was just a configuration mismatch with the integration testing setup, it's all good.

PR #3261, opened 2023-06-23
Fix target symlinks for non-relocatable builds

tobim:


vast/CMakeLists.txt
```diff
@@ -18,19 +14,26 @@ install(
 
 macro (create_tenzir_symlink name)
   add_custom_command(
-    OUTPUT "${CMAKE_BINARY_DIR}/${CMAKE_INSTALL_BINDIR}/${name}"
+    TARGET tenzir POST_BUILD
```
dominiklohmann, line 17: If you change from command output to a post-build target then you must list the created byproducts, otherwise the `clean` target cannot pick them up.

vast/CMakeLists.txt
```diff
@@ -18,19 +14,26 @@ install(
...
+        \"${CMAKE_COMMAND}\" -E create_symlink \"./tenzir\"
+        \"\$ENV{DESTDIR}\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/${name}\")"
+      COMPONENT Runtime)
+  else ()
+    install(
+      CODE "
+      execute_process(COMMAND
+        \"${CMAKE_COMMAND}\" -E create_symlink \"./tenzir\"
+        \"\$ENV{DESTDIR}${CMAKE_INSTALL_FULL_BINDIR}/${name}\")"
+      COMPONENT Runtime)
```
dominiklohmann, line 35: This should not be needed. This code should make sure that we can never run into this situation: https://github.com/tenzir/tenzir/blob/58b2be734e0b1f348afcc987bfefb643d05f6761/CMakeLists.txt#L176-L191

dominiklohmann: Should likely be `COMPONENT Runtime` though so this conflict would always be detected.

tobim: Yes it is needed. We correctly only normalize install dirs in relocatable mode, so
`{CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}` contains the prefix twice, making it fail.

dominiklohmann: I see. But that's still problematic, as it would similarly affect all other places in which we use `install(CODE …)`.

Do we have a good reason to support non-relocatable builds in the first place?

vast/CMakeLists.txt
tobim: I'll check if the handling of generator expressions in BYPRODUCT is correct, if it isn't I will leave it out.

NB: What is the use of the `clean` target when you can just delete the build tree? With the latter approach you will at least be sure that you won't get any problems from lingering cache variables.

dominiklohmann: There are some situations in which you want to `cmake --build path/to/build --clean-first` because you changed your build configuration, e.g., when removing a plugin from the list of installed files. It's generally good practice to list all created files because of this.

I think the old approach for the symlink setup is much preferred because it only runs when needed, instead of running as a side-effect of another target completing. It should also likely be `POST_LINK` rather than `POST_BUILD`, otherwise the created symlink may not be usable within CMake if used as a dependency.

vast/CMakeLists.txt
tobim: We use it in non-static Nix builds, which this PR is intended to fix.

dominiklohmann: If they could also be made relocatable builds then we could drop the CMake option, and would not need to have this complexity for all custom install scripts. Wouldn't that be a lot easier to maintain than fixing the issue?

vast/CMakeLists.txt
tobim: I actually much prefer the attach to target approach. The old way you always had to build with `--target tenzir tenzir-ctl tenzir-node` to get everything that is needed. Easy to forget and get confusing errors.

tobim: > I'll check if the handling of generator expressions in BYPRODUCT is correct, if it isn't I will leave it out.

To no surprise, it isn't:
```
CMake Error at vast/CMakeLists.txt:16 (add_custom_command):
  Error evaluating generator expression:

    $<TARGET_FILE_DIR:tenzir>

  No target "tenzir"
```

tobim: Also, there is no `POST_LINK`, `POST_BUILD` means all other build steps have completed.

vast/CMakeLists.txt
tobim: > Wouldn't that be a lot easier to maintain than fixing the issue?

The would need to investigated, and I won't spend the time on that now.
The provided patch works.

vast/CMakeLists.txt
dominiklohmann: I still don't understand the need to change this. Using the `TARGET_FILE_DIR` generator expression makes sense for maintainability, but if the goal is to just run `cmake --build path/to/build --target tenzir` to also run the `tenzir-ctl`, `tenzir-node`, and `vast` targets, then why not create an additional target for your convenience that depends on the all four? Something like this:

```cmake
add_custom_target(runtime DEPENDS tenzir tenzir-ctl tenzir-node vast)
```

vast/CMakeLists.txt
dominiklohmann: I don't want to spend too much time on this, on that front I agree. But:
- We do not test this build configuration in CI
- It is a very confusing piece of code to understand

So the very least this PR should get is a comment that explains _why_ this logic exists, and also _why_ the install-time vs. build-time prefix mismatch detection logic does not suffice for this.

vast/CMakeLists.txt
tobim: Because TARGET* generator expressions don't work in the OUTPUT field.
I found a way to make it work with the `clean` target, see the latest version.

vast/CMakeLists.txt
tobim: That detection logic is not relevant for this. I added the requested comment.

dominiklohmann (approved):

PR #3263, opened 2023-06-24
Revamp packet acquisition and parsing

mavam:
This PR revamps the PCAP plugin under the new connector/format framework.

```[tasklist]
### Tasks
- [x] Add `pcap` parser
- [x] Add `pcap` printer
- [x] Add `nic` loader
- [x] Document everything
- [x] Add `decapsulate` operator
- [x] Write changelog entry
- [x] Support externally passed PCAP file header
- [x] Add tests
```


libvast/builtins/formats/pcap.cpp
```diff
@@ -0,0 +1,510 @@
...
+      .major_version = 2,
+      .minor_version = 4,
+      .reserved1 = 0,
+      .reserved2 = 0,
+      .snaplen = snaplen,
+      .linktype = linktype,
+    };
+  }
+
+  auto instantiate(type input_schema, operator_control_plane& ctrl) const
```
mavam, line 339: @tenzir/pipelines-squad I'm struggling with this error:

```
error: !! logic_error: 'print "pcap", pcap_printer(anonymous())' does not support heterogeneous outputs; cannot initialize for 'pcap' after 'pcap.packet'
 = note: from `print "pcap", pcap_printer(anonymous())`
```

My hunch is that I cannot dispatch in `instantiate` like I do. Can you advice how I can handle two different schema types properly?

mavam (commented):
I need some help from @tenzir/pipelines-squad, per the comment inline.

libvast/builtins/formats/pcap.cpp
jachris: Your printer has `allows_joining() == false`. If the saver has `is_joining() == true` (e.g. `to file ...`), then we know that all output schemas go to the same stream, interleaved. But since your printer claims that it does not support this joining/interleaving, we can only give it events for a single schema. The error message above happens when there are events of multiple schemas.

mavam: Makes sense. I enabled joining. What are the guarantees on concurrent execution? Can I share state between printers from the same call to `instantiate`? Say I return a different generator based on the schema name, and I'd like them to share state, is that allowed? (That's what I'm doing now.) Or should I rather always return one generator that does the schema-based dispatching internally?

mavam: This is now obsolete. I found and fixed the issues.

dominiklohmann (approved):
As discussed in the office, I've simply pushed my review feedback directly onto your branch. From my side this is ready to go. Before merging, please take a look at the commits I've added.

PR #3264, opened 2023-06-26
Fix sporadic stalling of pipelines

dominiklohmann:
This is the third iteration of pipeline execution. We've started with a simple chaining of generators in a single thread just for testing, and moved on to an actor-based model that utilizes CAF's push-based streaming feature for first usage. That was convenient for getting things going, but we quickly realized some disadvantages that we now fix by switching to a message-based approach that is very close to the Volcano model:

1. Transformation or sink operators whose input is exhausted or source operator no longer sporadically stall. All execution nodes are now guaranteed to yield to the actor scheduler whenever the operator's generator advances.
2. Back pressure now regulates the number events or bytes rather than number of variable-size batches of events or bytes transferred between operators.
3. The initialization order of operators is now well-defined: The execution nodes are spawned left-to-right, and the pipeline operators are guaranteed to be initialized right-to-left. Specifically, an operator must yield once before its preceding operator is initialized.
4. Operators can now await requests if they require a request to be responded to before continuing the pipeline. This works in a scheduler-friendly way.
5. In the long-term, we want to make use async generators for operators. Plugging that in would now be a relatively easy change to make.

```[tasklist]
### Tasks
- [x] Reimplement the execution node to fix the stalling bug
- [x] Remove leftover logs from debugging
- [x] Make pipeline startup explicit, and return only once all generators have been primed
- [x] Avoid crashing when running non-closed pipelines in the pipeline manager
- [x] Move closedness checks ~and optimization~ into the pipeline executor
- [x] Fix stalling on shutdown
- [x] Change pipeline executor to return once all execution nodes have been started
- [x] Ensure that all items have been ack'ed by the next exec node for graceful shutdowns
- [x] Bring back remote pipelines
- [x] Document scheduling within an execution node
- [x] Document message flow between execution nodes
- [x] Document initialization order guarantees for pipeline executor
- [x] Consistently use diagnostics for warnings
- [x] Change `serve` not to use blocking communication anymore
- [x] Fix pipeline manager's use of the API
- [x] Fix build with gcc
- [x] Change `measure` to not yield quite as frequently
```


jachris (approved):
We looked over this together. There a few tiny issues that we want to address before merging it, but besides that, this is ready.

PR #3265, opened 2023-06-26
Split the docker image into tenzir and tenzir-node

tobim:
We now provide dedicated docker images that mirror the command line programs. You can use the `tenzir-node` image to run a node and the `tenzir` image to run ad-hoc pipelines.

The editions and `-slim` variations still exist, but are expected to be phased out in the medium term.

An image for `tenzir-ctl` is omitted intentionally, because all command functionality will be integrated into the pipelines framework in the future.


scripts/regression-tests.sh
```diff
@@ -39,7 +39,7 @@ VAST_RUN_FLAGS="-d --pull=always --rm --name vast-regression -e VAST_CONSOLE_VER
 # Pull the new version to verify database compatibility.
 docker run \
   $VAST_RUN_FLAGS \
-  ghcr.io/tenzir/tenzir:$NEW_VERSION \
+  ghcr.io/tenzir/tenzir-node:$NEW_VERSION \
   start
```
rdettai, line 43: 2 questions:
- shouldn't `tenzir-node` work without the `start` argument?
- should `tenzir-node` even work with the `start` argument? 😅

docker/compose/vast.yaml
```diff
@@ -16,6 +16,8 @@ services:
     ports:
       - ${VAST_WEB_PORT:-5160}:5160
       - ${VAST_PORT:-5158}:5158
+    entrypoint:
+      - tenzir-ctl
```
rdettai, line 20: I think this is a good workaround for now. I'm impressed/suprised the CI goes through with just this change, we probably miss some tests in the dependent compose scripts. Fully adapting the docker scripts for the new `tenzir` /  `tenzir-node` paradigm is a fair amount of work and I don't think we should undertake it right now.

scripts/regression-tests.sh
rdettai, line 1: The regression tests are comparing `tenzir/tenzir:{current_version}` to `tenzir/vast:{old_version}`. Won't that start failing as soon as we release `tenzir/tenzir:v4.0` as we won't be pushing that version to `tenzir/vast` anymore and further regression tests will still try pulling `tenzir/vast:v4.0`?

tobim: You mean `tenzir/tenzir:v4.0`? Yes, we will need to adjust the parameters for versions >= 4.0.

tobim: I didn't add that logic yet because it is a lot of effort in testing which we get for free when we get to 4.1.

scripts/regression-tests.sh
tobim: :see_no_evil:

docker/compose/vast.yaml
tobim: Yes, I'd rather keep it in place like this to keep it tested and avoid bit-rot.
The adaptation will be done in the coming weeks.

scripts/regression-tests.sh
tobim: I didn't address the second point because that is orthogonal.

dominiklohmann (approved):

PR #3268, opened 2023-06-27
Allow unsafe pipelines on the demo node

tobim:


mavam (approved):

lava (commented):
I think it makes more sense to add this variable to the service definition, so we can override it for specific demo nodes if desired.

tobim:
> I think it makes more sense to add this variable to the service definition, so we can override it for specific demo nodes if desired.

Agreed, that would be better.  Can you open a PR for that?

lava:
> Can you open a PR for that?

https://github.com/tenzir/event-horizon/pull/1750

tobim:
This is solved somewhere else in a more flexible way.

PR #3269, opened 2023-06-27
Remove handrolled `<bit>` header

mavam:
This PR removes our hand-rolled version of `<bit>`, which is now part of the standard library.


dominiklohmann (approved):
This looks pretty mechanical. Thanks!

PR #3271, opened 2023-06-27
Use "tenzir" as the main config section

tobim:


mavam (approved):
Just needs some rebasing.

libvast/src/plugin.cpp
```diff
@@ -46,12 +46,12 @@ namespace {
 detail::stable_set<std::filesystem::path>
 get_plugin_dirs(const caf::actor_system_config& cfg) {
   detail::stable_set<std::filesystem::path> result;
-  const auto bare_mode = caf::get_or(cfg, "vast.bare-mode", false);
+  const auto bare_mode = caf::get_or(cfg, "tenzir.bare-mode", false);
   // Since we do not read configuration files that were not explicitly
   // specified when in bare-mode, it is safe to just read the option
   // `vast.plugin-dirs` even with bare-mode enabled.
```
dominiklohmann, line 52: Can you grep for `'vast.` or `vast\.` as well? There's a lot of them remaining still in comments like this.

tobim: Thanks for double checking. I fixed all that I could find.

PR #3278, opened 2023-06-28
Update the release prep script for the Tenzir name

tobim:
And add an entry for the latest release candidate to the docs version selector.

PR #3279, opened 2023-06-28
Prepare repository for Tenzir v4.0.0-rc1

tobim:
The first of probably many release candidates for Tenzir v4.0.


scripts/prepare-release
```diff
@@ -78,8 +78,8 @@ if [[ true ]]; then
   new_version_name="Tenzir ${new_version_major_minor}"
   jq "map(select(. | startswith(\"${new_version_name%-*}\") | not))" < versions.json > versions.json.tmp
   mv -f versions.json.tmp versions.json
-  rm -rf "versioned_docs/version-${new_version_name%-*}"*
-  rm -rf "versioned_sidebars/version-${new_version_name%-*}"*-sidebars.json
+  rm -rf "versioned_docs/version-${new_version_name%-*}"* || true
+  rm -rf "versioned_sidebars/version-${new_version_name%-*}"*-sidebars.json || true
```
tobim, line 82: FYI: This modification is a follow up to #3278. It is a workaround for idiosyncratic behavior of zsh, which fails on a globbed `rm` if no matches are found.

mavam: This should be a comment in the code.

mavam:
@tobim could you explain how a reviewer should look at the 9k LOC diff?

tobim:
Almost all of this is created by copying from next to the versioned directories. None of the content is actually new.

mavam (approved):

mavam:
So basically you just invoked the usual script that rotated the docs?

scripts/prepare-release
tobim: I'll follow that up with a separate PR. The script should add a PR description as well.

tobim:
> So basically you just invoked the usual script that rotated the docs?

Precisely.

PR #3286, opened 2023-06-30
Fix some build scaffolding things

dominiklohmann:
This is three CMake changes in one:
- Fixed the macOS build for the Kafka plugin and removes the librdkafka submodule that was used as a fallback.
- Bumped the bundled simdjson to 3.2.0.
- Fixed a pollution of CMake's module path when building the Parquet plugin as bundled.


mavam:
@dominiklohmann Mind adding a one-liner for why the submodule cannot be bundled?

mavam:
This compiles for me. First time I'm hitting this, though:

```
[06:53:51.739] platform-client-17 failed to connect: !! unspecified: handshake error: The WebSocket handshake was declined by the remote peer; will retry in 2m
```

Not sure if we have a discussion about this already, but I think this might come up in the future and we should consider making that warning really clear, i.e., make it a diagnostic and provide a link on where to read more.

dominiklohmann:
> @dominiklohmann Mind adding a one-liner for why the submodule cannot be bundled?

We could to that, but librdkafka is available as a package on all platforms we use for development and testing, so also having the submodule is additional maintenance overhead I would like to avoid.

dominiklohmann:
> This compiles for me. First time I'm hitting this, though:
> 
> 
> 
> ```
> 
> [06:53:51.739] platform-client-17 failed to connect: !! unspecified: handshake error: The WebSocket handshake was declined by the remote peer; will retry in 2m
> 
> ```
> 
> 
> 
> Not sure if we have a discussion about this already, but I think this might come up in the future and we should consider making that warning really clear, i.e., make it a diagnostic and provide a link on where to read more.

Can you move this into a separate issue for discussion on Monday? It’s entirely independent of this PR, but I agree that it is really important.

tobim:
Can you split the simdjson version increase into a separate PR? 3.2.0 doesn't seem to include any critical fixes and we don't need the new functionality.

web/docs/developer-guides/build-from-source.md
```diff
@@ -37,7 +37,7 @@ Every [release](https://github.com/tenzir/tenzir/releases) of Tenzir includes an
 |✓|[Apache Arrow](https://arrow.apache.org)|>= 8.0.0|Required for in-memory data representation. Must be built with Compute, Zstd and Parquet enabled.|
 |✓|[re2](https://github.com/google/re2)||Required for regular expressione evaluation.|
 |✓|[yaml-cpp](https://github.com/jbeder/yaml-cpp)|>= 0.6.2|Required for reading YAML configuration files.|
-|✓|[simdjson](https://github.com/simdjson/simdjson)|>= 3.1.0|Required for high-performance JSON parsing. (Bundled as submodule.)|
+|✓|[simdjson](https://github.com/simdjson/simdjson)|>= 3.2.0|Required for high-performance JSON parsing. (Bundled as submodule.)|
```
mavam, line 40: The CMake feels like 3.1 is the minimum.

dominiklohmann: Ah right, I forgot to undo this change. I made it slightly more lenient after @tobim's comment so it's 3.1+ rather than 3.2+; before this PR it was 3.1 exactly, which caused my local 3.2 installation of simdjson to always be ignored at the cost of additional compile time.

dominiklohmann: Done and fixed.

dominiklohmann:
@tobim can you take a look at the Kafka failure for Nix that we get now?

plugins/kafka/CMakeLists.txt
```diff
@@ -18,33 +18,14 @@ VASTRegisterPlugin(
...
-  add_library(RdKafka::rdkafka++ ALIAS rdkafka++)
+  list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+  find_package(RdKafka QUIET REQUIRED)
+  list(POP_BACK CMAKE_MODULE_PATH)
 endif ()
 
-target_link_libraries(kafka PUBLIC vast::libvast vast::internal)
+dependency_summary("librdkafka" RdKafka::rdkafka++ "Dependencies")
 target_link_libraries(kafka PUBLIC RdKafka::rdkafka++)
+target_link_libraries(kafka-shared PUBLIC RdKafka::rdkafka++)
```
tobim, line 30: `kafka-shared` doesn't exist when `VAST_ENABLE_STATIC_PLUGINS` is true.

dominiklohmann: Ah damn, I totally overlooked this. That's simply a leftover from debugging, we need neither this line nor the one below.

tobim (approved):

PR #3291, opened 2023-07-03
Unroll the Zeek TSV header parsing loop

Dakostu:
This change unrolls the header parsing loop in the `zeek-tsv` parser to allow early yielding during header parsing.

### DIFF
--- a/libvast/builtins/formats/zeek_tsv.cpp
+++ b/libvast/builtins/formats/zeek_tsv.cpp
@@ -662,6 +662,12 @@ auto parser_impl(generator<std::optional<std::string_view>> lines,
       }
     }
   }
+  if (not closed) {
+    ctrl.abort(
+      caf::make_error(ec::syntax_error, fmt::format("zeek-tsv parser failed: "
+                                                    "Missing #close")));
+    co_return;
+  }
   VAST_ASSERT_CHEAP(b);
   auto finished = b->finish();
   if (metadata.output_slice_schema)
--- a/libvast/builtins/formats/zeek_tsv.cpp
+++ b/libvast/builtins/formats/zeek_tsv.cpp
@@ -493,17 +493,25 @@ auto parser_impl(generator<std::optional<std::string_view>> lines,
       metadata.sep.clear();
       metadata.sep.push_back(metadata.sep_char);
       for (const auto& prefix : metadata.prefix_options) {
-        ++it;
-        current_line = *it;
-        if (not current_line) {
-          co_yield {};
-        }
-        if (it == lines.end()) {
-          ctrl.abort(caf::make_error(ec::syntax_error,
-                                     fmt::format("zeek-tsv parser failed: "
-                                                 "header ended too "
-                                                 "early")));
-          co_return;
+        auto current_line = std::optional<std::string_view>{};
+        while (true) {
+          ++it;
+          if (it == lines.end()) {
+            ctrl.abort(caf::make_error(ec::syntax_error,
+                                       fmt::format("zeek-tsv parser failed: "
+                                                   "header ended too "
+                                                   "early")));
+            co_return;
+          }
+          current_line = *it;
+          if (not current_line) {
+            co_yield {};
+            continue;
+          }
+          if (current_line->empty()) {
+            continue;
+          }
+          break;
         }
         metadata.header = *current_line;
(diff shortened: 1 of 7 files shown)


libvast/builtins/formats/zeek_tsv.cpp
```diff
@@ -546,17 +445,152 @@ auto parser_impl(generator<std::optional<std::string_view>> lines,
...
-  std::vector<data> xs(metadata.fields.size());
+  auto sep_parser
+    = "#separator" >> ignore(+(parsers::space)) >> +(parsers::any);
+  auto sep_option = std::string{};
+  auto header_parsed = false;
+  auto closed = true;
+  auto b = std::optional<table_slice_builder>{};
+  auto xs = std::vector<data>{};
   for (; it != lines.end(); ++it) {
+    if (*it and (*it)->starts_with("#separator")) {
```
dominiklohmann, line 456: I think from a readability perspective it would be much better to just move the nullopt check to the very top of the loop, i.e.,

```c++
for (; it != lines.end(); ++it) {
  auto current_line = *it;
  if (not current_line) {
    co_yield {};
    continue;
  }
  if (current_line->empty()) {
    continue;
  }
  // work with *current_line instead of **it
}
```

Dakostu: Agreed. I added `current_line` in https://github.com/tenzir/tenzir/pull/3291/commits/d3dfa2a19440753d1d78858caf7286d7e849e792

dominiklohmann (approved):
Thanks for fixing this so quickly.

I confirmed locally that this works as expected by making the file loader return chunks of size 16 rather than 16,384, which would guarantee that I run into the error all the time. I don't anymore, so this is good to go.

libvast/builtins/formats/zeek_tsv.cpp
dominiklohmann, line 1: Please also add a changelog entry. Here's a wording suggestion:

> The `zeek-tsv` parser sometimes failed to parse concatenated Zeek TSV logs, wrongly reporting that the header ended too early. This bug no longer exists.

libvast/builtins/formats/zeek_tsv.cpp
```diff
@@ -546,27 +445,167 @@ auto parser_impl(generator<std::optional<std::string_view>> lines,
...
+  auto header_parsed = false;
+  auto closed = true;
+  auto b = std::optional<table_slice_builder>{};
+  auto xs = std::vector<data>{};
   for (; it != lines.end(); ++it) {
-    auto line = *it;
-    if (not line) {
+    auto current_line = *it;
+    if (not current_line) {
+      co_yield {};
```
dominiklohmann, line 458:
```suggestion
      co_yield {};
      continue;
```

libvast/builtins/formats/zeek_tsv.cpp
```diff
@@ -546,27 +445,167 @@ auto parser_impl(generator<std::optional<std::string_view>> lines,
...
+        if (not current_line) {
+          co_yield {};
+        }
+        if (it == lines.end()) {
+          ctrl.abort(caf::make_error(ec::syntax_error,
+                                     fmt::format("zeek-tsv parser failed: "
+                                                 "header ended too "
+                                                 "early")));
+          co_return;
+        }
```
dominiklohmann, line 506:
```suggestion
        auto current_line = std::optional<std::string_view>{};
        while (true) {
          ++it;
          if (it == lines.end()) {
            ctrl.abort(caf::make_error(ec::syntax_error,
                                       fmt::format("zeek-tsv parser failed: "
                                                   "header ended too "
                                                   "early")));
            co_return;
          }
          current_line = *it;
          if (not current_line) {
            co_yield {};
            continue;
          }
          if (current_line->empty()) {
            continue;
          }
          break;
        }
```

dominiklohmann (commented):
Forgot to actually submit my local changes before as part of the approving review, these two must get in.

libvast/builtins/formats/zeek_tsv.cpp
Dakostu: I added a changelog item in https://github.com/tenzir/tenzir/pull/3291/commits/b98d393c89ef61e22ed8cbf15f433ed8a7a9884f

PR #3295, opened 2023-07-03
Add a `--schema` option to the JSON parser

dominiklohmann:
The new option allows for setting the target schema name explicitly.


libvast/builtins/operators/from_read_load_parse.cpp
```diff
@@ -90,6 +90,18 @@ class parse_operator final : public crtp_operator<parse_operator> {
...
+  auto infer_type_impl(operator_type input) const
+    -> caf::expected<operator_type> override {
+    if (input.is<chunk_ptr>()) {
+      return tag_v<table_slice>;
+    }
+    // TODO: Fuse this check with crtp_operator::instantiate()
+    return caf::make_error(ec::type_clash,
+                           fmt::format("'{}' does not accept {} as input",
+                                       to_string(), operator_type_name(input)));
+  }
```
dominiklohmann, line 103: The additional `infer_type_impl` overrides are necessary currently, although @jachris mentioned a better approach, so we be able to get rid of this shortly.

patszt:
I think this line must be adapted to enable proper working of --no-infer.
https://github.com/tenzir/tenzir/pull/3295/files#diff-acb3f00b39defb32abdaaff714123837f32ad305710aadf4ebfcdb55a6651797L679

Should be if (no_infer and (selector or schema)) then use strict_validator.

dominiklohmann:
> I think this line must be adapted to enable proper working of --no-infer. https://github.com/tenzir/tenzir/pull/3295/files#diff-acb3f00b39defb32abdaaff714123837f32ad305710aadf4ebfcdb55a6651797L679
> 
> Should be if (no_infer and (selector or schema)) then use strict_validator.

Fantastic, that worked!

patszt (approved):
Looks good. Good that we have a test now!

libvast/builtins/operators/from_read_load_parse.cpp
jachris: I think this is good for now, we can make the mentioned change later.

vast/integration/vast_integration_suite.yaml
```diff
@@ -735,6 +735,7 @@ tests:
     tags: [pipelines]
     steps:
       - command: exec 'from file @./data/suricata/eve.json read suricata | write json to stdout'
+      - command: exec 'from file @./data/suricata/eve.json read json --schema=suricata.alert --no-infer | write json to stdout'
```
mavam, line 738: Does this test actually test that this works? The output doesn't include `"suricata.alert"`. How about something like `put #schema`?

mavam: I think we need a test where we assign the table slice simply a fixed name, independent of existing schemas.

mavam: And that name would be good to have in the test output.

dominiklohmann: That appears to be independent of the feature added by this PR. I'll add a note to a related issue to add such a test.

PR #3299, opened 2023-07-04
Use a GCS link to fetch the latest release package

tobim:
We switch from a link to the latest GitHub release to GCS. GitHub's stable link does not consider pre-releases, but we need to mark ours as such currently.

### DIFF
--- a/scripts/debian/postinst
+++ b/scripts/debian/postinst
@@ -58,6 +58,15 @@ if [ "$1" = "configure" ]; then
 fi
 
 if [ "$1" = "configure" ] ; then
+	# Delete all lines referencing /var/lib/vast if it doesn't exist. This works
+	# around restrictive behavior in systemd, which refuses to start the unit if
+	# a ReadWritePath does not exist.
+  if ! [ -d /var/lib/vast ]; then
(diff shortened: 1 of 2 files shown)


mavam (approved):

tobim:
I verified that the follow-up commit fixes the installation on systems that don't have a vast package installed.

scripts/debian/postinst
```diff
@@ -58,6 +58,15 @@ if [ "$1" = "configure" ]; then
 fi
 
 if [ "$1" = "configure" ] ; then
+	# Delete all lines referencing /var/lib/vast if it doesn't exist. This works
+	# around restrictive behavior in systemd, which refuses to start the unit if
+	# a ReadWritePath does not exist.
+  if ! [ -d /var/lib/vast ]; then
+    sed -i '/\/var\/lib\/vast/d' /opt/tenzir/lib/systemd/system/tenzir.service
```
mavam, line 65: So this is not yet `tenzir.db` I assume?

tobim: It is. But the tenzir package should be able to read `/var/lib/vast` if vast was installed before.

PR #3300, opened 2023-07-04
Improve error and diagnostic propagation

jachris:
This PR improves the diagnostics, in particular when using the pipeline manager, which had broken diagnostics if the error happens after parsing (see integration tests of the pipeline manager).

### DIFF
--- a/libvast/src/execution_node.cpp
+++ b/libvast/src/execution_node.cpp
@@ -321,8 +321,8 @@ struct exec_node_state : inbound_state_mixin<Input>,
     // Instantiate the operator with its input type.
     auto output_generator = op->instantiate(make_input_adapter(), *ctrl);
     if (not output_generator) {
-      VAST_DEBUG("{} could not instantiate operator: {}", *self,
-                 output_generator.error());
+      VAST_VERBOSE("{} could not instantiate operator: {}", *self,
+                   output_generator.error());
       return add_context(output_generator.error(),
                          "{} failed to instantiate operator", *self);
     }
@@ -334,7 +334,7 @@ struct exec_node_state : inbound_state_mixin<Input>,
     }
     instance.emplace();
     instance->gen = std::get<generator<Output>>(std::move(*output_generator));
-    VAST_DEBUG("{} calls begin on instantiated operator", *self);
+    VAST_TRACE("{} calls begin on instantiated operator", *self);
     instance->it = instance->gen.begin();
     if (abort) {
       VAST_DEBUG("{} was aborted during begin: {}", *self, op->to_string(),
@@ -342,7 +342,7 @@ struct exec_node_state : inbound_state_mixin<Input>,
       return abort;
     }
     if constexpr (std::is_same_v<Output, std::monostate>) {
-      VAST_DEBUG("{} is the sink and requests start from {}", *self,
+      VAST_TRACE("{} is the sink and requests start from {}", *self,
                  this->previous);
       auto rp = self->make_response_promise<void>();
       self
@@ -350,7 +350,7 @@ struct exec_node_state : inbound_state_mixin<Input>,
                   std::move(previous))
         .then(
           [this, rp]() mutable {
-            VAST_DEBUG("{} schedules run of sink after successful startup",
+            VAST_TRACE("{} schedules run of sink after successful startup",
                        *self);
             schedule_run();
             rp.deliver();
--- a/libvast/src/execution_node.cpp
+++ b/libvast/src/execution_node.cpp
@@ -102,18 +102,22 @@ public:
 
   auto abort(caf::error error) noexcept -> void override {
     VAST_ASSERT(error != caf::none);
-    diagnostic::error("{}", error)
-      .note("from `{}`", state_.op->to_string())
-      .emit(diagnostics());
+    if (error != ec::silent) {
+      diagnostic::error("{}", error)
+        .note("from `{}`", state_.op->to_string())
+        .emit(diagnostics());
+    }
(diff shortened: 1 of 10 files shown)


libvast/src/execution_node.cpp
```diff
@@ -316,38 +321,48 @@ struct exec_node_state : inbound_state_mixin<Input>,
...
+      VAST_DEBUG("{} could not instantiate operator: {}", *self,
+                 output_generator.error());
+      return add_context(output_generator.error(),
+                         "{} failed to instantiate operator", *self);
     }
     if (not std::holds_alternative<generator<Output>>(*output_generator)) {
-      return caf::make_error(ec::logic_error,
-                             fmt::format("{} failed to instantiate operator",
-                                         *self));
+      return caf::make_error(
```
tobim, line 330: Would `add_context(ec::logic_error, "...", ...)` work here as well?

libvast/src/execution_node.cpp
```diff
@@ -316,38 +321,48 @@ struct exec_node_state : inbound_state_mixin<Input>,
     // Instantiate the operator with its input type.
     auto output_generator = op->instantiate(make_input_adapter(), *ctrl);
     if (not output_generator) {
-      return caf::make_error(
-        ec::unspecified, fmt::format("{} failed to instantiate operator: {}",
-                                     *self, output_generator.error()));
+      VAST_DEBUG("{} could not instantiate operator: {}", *self,
```
tobim, line 324: Maybe this should be at `VAST_VERBOSE`.

libvast/src/execution_node.cpp
```diff
@@ -316,38 +321,48 @@ struct exec_node_state : inbound_state_mixin<Input>,
...
-                             fmt::format("{} failed to instantiate operator",
-                                         *self));
+      return caf::make_error(
+        ec::logic_error, fmt::format("{} expected {}, but got {}", *self,
+                                     operator_type_name<Output>(),
+                                     operator_type_name(*output_generator)));
     }
     instance.emplace();
     instance->gen = std::get<generator<Output>>(std::move(*output_generator));
+    VAST_DEBUG("{} calls begin on instantiated operator", *self);
```
tobim, line 337: I'm afraid this is too verbose. Let's remove it.

libvast/src/execution_node.cpp
```diff
@@ -316,38 +321,48 @@ struct exec_node_state : inbound_state_mixin<Input>,
...
     instance->it = instance->gen.begin();
     if (abort) {
+      VAST_DEBUG("{} was aborted during begin: {}", *self, op->to_string(),
+                 abort);
       return abort;
     }
     if constexpr (std::is_same_v<Output, std::monostate>) {
+      VAST_DEBUG("{} is the sink and requests start from {}", *self,
+                 this->previous);
```
tobim, line 346: Also too verbose. Maybe at `VAST_TRACE`.

libvast/src/execution_node.cpp
```diff
@@ -316,38 +321,48 @@ struct exec_node_state : inbound_state_mixin<Input>,
...
       auto rp = self->make_response_promise<void>();
       self
         ->request(this->previous, caf::infinite, atom::start_v,
                   std::move(previous))
         .then(
           [this, rp]() mutable {
-            VAST_DEBUG("scheduling sink after successful startup");
+            VAST_DEBUG("{} schedules run of sink after successful startup",
+                       *self);
```
tobim, line 354: Also too verbose. Move to trace.

libvast/src/pipeline_executor.cpp
```diff
@@ -52,11 +54,15 @@ void pipeline_executor_state::start_nodes_if_all_spawned() {
         finish_start();
       },
       [this](caf::error& err) mutable {
+        VAST_DEBUG("{} aborts start because execution node could not be "
+                   "started: {}",
+                   *self, err);
```
tobim, line 59: I would bump to verbose.

libvast/src/pipeline_executor.cpp
```diff
@@ -95,7 +102,8 @@ void pipeline_executor_state::spawn_execution_nodes(pipeline pipe) {
         ->request(node, caf::infinite, atom::spawn_v,
                   operator_box{std::move(op)}, input_type, diagnostics)
         .then(
-          [this, index](exec_node_actor& exec_node) {
+          [=, this](exec_node_actor& exec_node) {
+            VAST_DEBUG("{} spawned {} remotely", *self, description);
```
tobim, line 106: Also verbose.

tobim (approved):
The core change is rather small and seems fine from my side.
For the added logging, can you make it so that the sad path log statements appear on `verbose` and regular path logs go to `trace`?

libvast/src/execution_node.cpp
jachris: I think it's helpful because `.begin()` often does non-trivial initialization. For example, `from file foo.json` tries to open the fail and calls `abort()` if this fails. I'll move this down to `VAST_TRACE` if that's alright.

libvast/src/execution_node.cpp
jachris: Ha, you are right. I mean, the intended use case for `add_context` was somewhat different, but it seems like it can also be used as a replacement for `caf::make_error`. I will not make this change right now for consistency reasons, but we could use that everywhere...

libvast/src/execution_node.cpp
tobim: That's alright.

PR #3302, opened 2023-07-04
Fix potential stalling issue with the zeek-tsv parser

dominiklohmann:
This adds a missing iterator increment to the zeek-tsv parser that could lead to an infinite loop.

While debugging this, I noticed two more things that I implemented quickly:
1. The zeek-tsv parser did not yield until the schema was changed. I've now added a max batch size and a batch timeout.
2. Execution nodes that had a full outbound buffer should not ask to be scheduled until the buffer is no longer full. I've also made this change so sources are scheduled less aggressively.


libvast/builtins/formats/zeek_tsv.cpp
```diff
@@ -670,6 +666,12 @@ auto parser_impl(generator<std::optional<std::string_view>> lines,
         co_return;
       }
     }
+    if (b->rows() > defaults::import::table_slice_size
+        or last_finish + std::chrono::seconds{1}
+             < std::chrono::steady_clock ::now()) {
+      last_finish = std::chrono::steady_clock::now();
```
tobim, line 672: I don't think the compiler will be able to merge these calls. Can you change it so it only makes the call once and assigns the result?

dominiklohmann: Right, fair point.

dominiklohmann: Done in 662725466a

libvast/builtins/formats/zeek_tsv.cpp
```diff
@@ -670,6 +666,12 @@ auto parser_impl(generator<std::optional<std::string_view>> lines,
         co_return;
       }
     }
+    const auto now = std::chrono::steady_clock::now();
+    if (b->rows() >= defaults::import::table_slice_size
+        or last_finish + std::chrono::seconds{1} < now) {
```
lava, line 671: This should probably be a named constant under `defaults::`

libvast/src/execution_node.cpp
```diff
@@ -514,16 +515,21 @@ struct exec_node_state : inbound_state_mixin<Input>,
...
       this->current_demand->sink, caf::infinite, atom::push_v, std::move(lhs));
-    if (force) {
+    if (force or this->outbound_buffer_size >= defaults::max_buffered) {
+      VAST_DEBUG("{} pushes {}/{} buffered elements and suspends execution",
+                 op->name(), capped_demand, this->outbound_buffer_size);
       std::move(response_handle)
         .await(std::move(handle_result), std::move(handle_error));
     } else {
+      VAST_DEBUG("{} pushes {}/{} buffered elements", op->name(), capped_demand,
+                 this->outbound_buffer_size);
```
lava, line 532: What kind of volume do we expect for these log messages? Is there a danger that it will overwhelm the other logs on a busy node?

lava (approved):
The logic change itself makes sense to me, although I didn't test it in-depth.

We should probably take the chance and add an empty file to our zeek integration tests; and one with an empty line.

libvast/src/execution_node.cpp
dominiklohmann: Once per 250ms or 2^16 elements per operator in a running pipeline, whatever is quickest. Would you prefer this to be a trace log?

libvast/builtins/formats/zeek_tsv.cpp
dominiklohmann: We have a story for implementing this more generically for all parsers in the future.

libvast/src/execution_node.cpp
lava: Based on this description I'd say yes, iiuc this could amount to dozens to hundreds of log lines per seocnd.

dominiklohmann: I moved it to trace.

PR #3303, opened 2023-07-04
Rewrite the zeek-tsv parser from the ground up

dominiklohmann:
The `zeek-tsv` parser has been a troublemaker lately. It was one of the few parsers that we more or less translated directly from the old reader plugin interface to the new parser plugin interface without changing the parsing approach. While that was convenient initially, I personally spent way too many hours debugging things going wrong with this parser lately.

This new approach has a few advantages:
- There exists only a single place in which we advance the input generator, and as such only a single place in which we need to yield an empty on exit.
- The parser now emits helpful diagnostics when something goes wrong.
- Parsing the header is now a lot less strict, and allows for header fields to allow in any order, and for closing tags to be omitted.
- Parsing events no longer copies all individual values twice, but instead only copies them once in the builder.

### DIFF
--- a/libvast/builtins/formats/zeek_tsv.cpp
+++ b/libvast/builtins/formats/zeek_tsv.cpp
@@ -19,6 +19,7 @@
 #include "vast/detail/string_literal.hpp"
 #include "vast/detail/to_xsv_sep.hpp"
 #include "vast/detail/zeekify.hpp"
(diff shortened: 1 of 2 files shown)


tobim (approved):
Reviewed in a call. The test coverage is decent here so I'm medium confident that this doesn't bring regressions.

PR #3306, opened 2023-07-05
Remove API to dynamically disable plugins

tobim:
A plugin being enabled or disabled based on the runtime config is not useful except for component plugins.
Removing that bit of state from the generic plugin level rids us form a couple of corner cases, e.g., missing plugins from the `version` operator, or a warning that gets displayed `tenzir` connects to a `tenzir-node` with mismatching plugins.

### DIFF
--- a/libvast/include/vast/fwd.hpp
+++ b/libvast/include/vast/fwd.hpp
@@ -382,6 +382,8 @@ using writer_ptr = std::unique_ptr<writer>;
 
 } // namespace vast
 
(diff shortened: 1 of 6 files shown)


lava (approved):
Fine with me; the submodule rebase probably needs to be rebased since it seems to include some extra commits.

tobim:
Re-targeted onto `topic/error-propagation`, this should automatically switch back to `main` once the other branch is merged.

PR #3307, opened 2023-07-05
Add further performance logs to the execution node

dominiklohmann:
This makes it so we track the total time an execution node is scheduled for, which is a good indicator for performance when compared against (1) the total time spent in `advance_generator()` and (2) the total time since the execution node was started.


tobim (approved):
Went over this in a call and my suggestions were integrated live.

PR #3309, opened 2023-07-06
Fix /serve sometimes returning a token instead of null

dominiklohmann:
To reproduce, run `… | head 0 | serve foo`: This pipeline immediately finishes right before serve, but before this change set a non-null continuation token instead of null. That is now fixed, and the continuation token is now always null when the previous operator is done already.


lava (approved):
I tested it in combination with the frontend and it seems to fix the issue.

tobim (approved):
Verified that it works for both very fast and very slow upstream operators now.

PR #3320, opened 2023-07-07
Tune defaults and demo-node experience

tobim:


demo-node/entrypoint.bash
```diff
@@ -1,16 +1,27 @@
 #!/usr/bin/env bash
 
-coproc NODE { exec tenzir-node --print-endpoint; }
+export TENZIR_ACTIVE_PARTITION_TIMEOUT="${TENZIR_ACTIVE_PARTITION_TIMEOUT:-5s}"
```
dominiklohmann, line 3: I don't think this should be even lower than 30s, this will lead to too small partitions otherwise. I think 30s is fine for an interactive experience.

demo-node/entrypoint.bash
```diff
@@ -1,16 +1,27 @@
...
+export TENZIR_AUTOMATIC_REBUILD="${TENZIR_AUTOMATIC_REBUILD:-0}"
+
+coproc NODE { exec tenzir-node --print-endpoint --commands="web server --mode=dev --port=42025"; }
```
dominiklohmann, line 6: I think this should use the default port of 5160 rather than 42025

libvast/include/vast/defaults.hpp
```diff
@@ -219,7 +219,7 @@ inline constexpr size_t max_partition_size = 4'194'304; // 4 Mi
 
 /// Timeout after which an active partition is forcibly flushed.
 inline constexpr caf::timespan active_partition_timeout
-  = std::chrono::minutes{5};
+  = std::chrono::seconds{30};
```
dominiklohmann, line 222: These two also need to be adapted in the following places:
- application.cpp (help text)
- tenzir.yaml.example (default values)

And they likely also need a changelog entry.

demo-node/entrypoint.bash
tobim: I dropped this line. We can always play around with these variables by setting them in the container runtime.

dominiklohmann (approved):

PR #3326, opened 2023-07-07
Update get started guide

mavam:
This PR makes further adaptations to the get started guide.


web/docs/command-line.md
```diff
@@ -37,9 +37,9 @@ tenzir-node [opts]
 tenzir-ctl [opts] cmd1 [opts1] cmd2 [opts2] ...
 ```
 
-Both long `--long=X` and short `-s X` exist. Boolean options do not require
-explicit specification of a value, and it suffices to write `--long` and `-s`
-to set an option to true.
+We have both long `--long=X` and short `-s X` options. Boolean options do not
+require explicit specification of a value, and it suffices to write `--long` and
+`-s` to set an option to true.
```
lava, line 42: You may want to mention that the `=` is mandatory for the long option and forbidden for short options, since this is an unusual quirk.

web/docs/command-line.md
```diff
@@ -103,12 +103,12 @@ the following rules:
...
-semantics:
+`TENZIR_X__Y__Z` maps to `tenzir-ctl x y --z`. Here are two examples with
+identical semantics:
 
 ```bash
-TENZIR_IMPORT__BATCH_SIZE=42 tenzir import json < data
-tenzir import --batch-size=42 json < data
+TENZIR_IMPORT__BATCH_SIZE=42 tenzir-ctl import json < data
+tenzir-ctl import --batch-size=42 json < data
 ```
```
lava, line 112: I don't think this is true in general, e.g. `TENZIR_PLUGINS__PLATFORM__API_KEY=xxx` doesn't map to `tenzir-ctl plugins --api-key=xxx`. I'd remove this whole section.

web/docs/setup-guides/deploy-a-node/README.md
```diff
@@ -201,3 +142,10 @@ There exist two ways stop a server:
    `pkill -2 tenzir-node`.
 
 Sending the process a SIGTERM is the same as hitting CTRL+C.
+
+## Automate the deployment
+
+Read our guides on automating the deployment of a node:
+
+- [Systemd](systemd.md)
+- [Ansible](ansible.md)
```
lava, line 151: The wording seems a bit off, "automating the deployment" is strictly the realm of tools like Ansible etc., not systemd.

lava (approved):
Looks good to me; I mostly scanned for textual issues but didn't attempt to follow the guide yet since iiuc we intend to do this with multiple people in parallel afterwards.

web/docs/setup-guides/deploy-a-node/README.md
mavam: Yeah, it's a bit odd indeed. Any suggestion on how to integrate the systemd page?

web/docs/command-line.md
mavam: Okay, will remove.

web/docs/setup-guides/deploy-a-node/README.md
mavam: I spoke with @tobim about this. It's not technically correct but we'll leave it like this for now.

PR #3329, opened 2023-07-09
Use `load -` and `read json` as implicit sources

dominiklohmann:
This is the dual to the already existing feature of the `tenzir` binary that added implicit `save -` or `write json --pretty` to the pipeline if the pipeline ended in bytes or events, respectively: now we also add a source as needed.

### DIFF
--- a/changelog/next/features/3329--implicit-sink.md
+++ b/changelog/next/features/3329--implicit-sink.md
@@ -1,0 +1,3 @@
+Pipelines executed locally with `tenzir` now use `load -` and `read json` as
+implicit sources. This complements `save -` and `write json --pretty` as
+implicit sinks.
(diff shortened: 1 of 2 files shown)


jachris (approved):
I'm still confused by the loop. I think you don't need it (if you first check what the input type is). Anyway, I don't care too much, and this is very nice to have 🎉

dominiklohmann:
> (if you first check what the input type is)

I cannot do that because an operator may accept multiple different input types. I always need to check void as a possible input type first, and then opted to use bytes over events right now, i..e, `tenzir pass` becomes `tenzir 'load - | pass | save -'` over `tenzir 'read json | pass | write json --pretty' right now.

The alternative would be to check whether it's ambiguous and to not add an implicit source at all if that's the case.

jachris:
Well, yes, but you could still check for `void` first using a simple `if-else-if` chain, first for the input, and then for the output type. But let's just get this in, it doesn't matter :)

PR #3330, opened 2023-07-09
More demo node fixes

tobim:
The main fix here is to apply correct quote escaping so that the body is actually valid JSON.

The PR also contains reliability and quality of life improvements:
* Make sure that the HTTP endpoint is listening when the endpoint message is printed
* Enable unsafe pipelines in the demo node
* Make the entrypoint interruptible
* Add names to the demo data pipelines
* Filter out `suricata.stats` events

### DIFF
--- a/demo-node/entrypoint.bash
+++ b/demo-node/entrypoint.bash
@@ -22,17 +22,17 @@ while ! timeout 100 bash -c "echo > /dev/tcp/127.0.0.1/5160"; do
   sleep 1
 done
 
-suricata_pipe="shell 'bash -c \\\"curl -s -L https://storage.googleapis.com/tenzir-datasets/M57/suricata.tar.zst | tar -x --zstd --to-stdout\\\"' | parse suricata | import"
+suricata_pipe="shell 'bash -c \\\"curl -s -L https://storage.googleapis.com/tenzir-datasets/M57/suricata.tar.zst | tar -x --zstd --to-stdout\\\"' | parse suricata | where #type != \\\"suricata.stats\\\" | import"
 zeek_pipe="shell 'bash -c \\\"curl -s -L https://storage.googleapis.com/tenzir-datasets/M57/zeek.tar.zst | tar -x --zstd; cat Zeek/*.log\\\"' | parse zeek-tsv | import"
 
 curl -X POST \
   -H "Content-Type: application/json" \
-  -d "{\"definition\": \"${suricata_pipe}\", \"start_when_created\": true}" \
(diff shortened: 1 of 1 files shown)


demo-node/entrypoint.bash
```diff
@@ -10,8 +13,10 @@ read -r -u "${NODE[0]}" DUMMY
...
 export TENZIR_CONSOLE_VERBOSITY="${TENZIR_CLIENT_CONSOLE_VERBOSITY:-"info"}"
 
-suricata_pipe="shell \'bash -c \"curl -s -L https://storage.googleapis.com/tenzir-datasets/M57/suricata.tar.zst | tar -x --zstd --to-stdout\"\' | parse suricata | import"
-zeek_pipe="shell \'bash -c \"curl -s -L https://storage.googleapis.com/tenzir-datasets/M57/suricata.tar.zst | tar -x --zstd; cat Zeek/*.log\"\' | parse zeek | import"
+tenzir 'remote version | put version | write json'
+
+suricata_pipe="shell 'bash -c \\\"curl -s -L https://storage.googleapis.com/tenzir-datasets/M57/suricata.tar.zst | tar -x --zstd --to-stdout\\\"' | parse suricata | import"
+zeek_pipe="shell 'bash -c \\\"curl -s -L https://storage.googleapis.com/tenzir-datasets/M57/zeek.tar.zst | tar -x --zstd; cat Zeek/*.log\\\"' | parse zeek-tsv | import"
 
 curl -X POST \
```
dominiklohmann, line 21: Before this we need to `while ! lsof -i :5160; do sleep 1; done` to wait until the API is available.

libvast/src/start_command.cpp
```diff
@@ -147,6 +144,9 @@ caf::message start_command(const invocation& inv, caf::actor_system& sys) {
       self->send(runner, atom::run_v, *hook_invocation);
     }
   }
+  // A single line of output to publish our address for scripts.
+  if (caf::get_or(inv.options, "tenzir.start.print-endpoint", false))
+    std::cout << listen_addr << std::endl;
```
dominiklohmann, line 149: This must move back up to where it was before. We run the start commands asynchronously. For example:

```
tenzir-node --commands="exec version" --print-endpoint
```

This prints the version and the port in a non-deterministic order.

If you move it back up then the endpoint is at least guaranteed to be printed first.

dominiklohmann (approved):

demo-node/entrypoint.bash
tobim: Went for a bash native solution.

PR #3331, opened 2023-07-10
Fix predicate pushdown implementation of `local`/`remote`

jachris:
The current implementation of `local_remote_operator::predicate_pushdown` is broken because it does not re-wrap the optimized operator, which means that we are losing the `local`/`remote` modifier.

### DIFF
--- a/libvast/builtins/operators/local_remote.cpp
+++ b/libvast/builtins/operators/local_remote.cpp
@@ -35,7 +35,12 @@ public:
 
   auto predicate_pushdown(expression const& expr) const
     -> std::optional<std::pair<expression, operator_ptr>> override {
-    return op_->predicate_pushdown(expr);
+    auto result = op_->predicate_pushdown(expr);
+    if (result and result->second) {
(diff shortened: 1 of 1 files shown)


libvast/builtins/operators/local_remote.cpp
```diff
@@ -83,19 +88,6 @@ class local_remote_operator final : public operator_base {
...
-    auto allow_unsafe_pipelines
-      = try_get_or(global_config, "tenzir.allow-unsafe-pipelines", false);
-    if (not allow_unsafe_pipelines) {
-      return caf::make_error(
-        ec::invalid_configuration,
-        fmt::format("failed to parse tenzir.allow-unsafe-pipelines option: {}",
-                    allow_unsafe_pipelines.error()));
-    }
-    return {};
-  }
```
dominiklohmann, line 97: This did check whether the option was either unset or holds a boolean, which I think is not something that we want to lose.

dominiklohmann (approved):
Makes sense.

PR #3334, opened 2023-07-10
Upgrade the parser of some more operators

jachris:
We were still using the old parser in some unnecessary places. This PR will improve the parser error messages for those.


Dakostu (approved):
Approval after looking through the code. (I did not run each operator as a test, the integration tests should cover that.)

PR #3335, opened 2023-07-10
Allow list builder to change type

patszt:
In this PR we allow the some form of a common type casting inside a records and lists.

Previously changing a field inside a nested record was considered as a casting of a `record_type` so it was not implemented.
Same with the lists. A `list<int>` changing to `list<string>` is a list casting, which was not planned to be implemented yet.

It seems that the live data has a lot of `ip` -> `string` conversions that must be supported. These conversions occurr inside the lists and inside the nested record so this type of casting should be supported.

In this PR we allow:
- changing of a list_type when the added value requires the common type casting. E.g when we have a `list<ip>` we may add a "www.tenzir.com". This is inferred as a string type and the whole list must be cast to a `list<string>` in order to be compatible with the new string input.
- In case of `list<record<...>>` the field inside a record can also require the common type casting. E.g `list<record<ip_field: ip>>` can be converted to `list<record<ip_field: string>>` when we try to add some not castable data to the `ip_field`

### DIFF
--- a/libtenzir/include/tenzir/cast.hpp
+++ b/libtenzir/include/tenzir/cast.hpp
@@ -190,6 +190,38 @@ struct cast_helper<FromType, ToType> {
     else
       return caf::visit(f, from_type, to_type);
   }
+
+  static auto cast_value(const type& from_type, const data& data,
+                         const ToType& to_type) noexcept {
+    const auto f
+      = [&]<concrete_type ConcreteFromType, concrete_type ConcreteToType>(
+          const ConcreteFromType& from_type,
+          const ConcreteToType& to_type) noexcept
+      -> caf::expected<type_to_data_t<ToType>> {
+      auto v = cast_helper<ConcreteFromType, ConcreteToType>::cast_value(
+        from_type, caf::get<type_to_data_t<ConcreteFromType>>(data), to_type);
+      if constexpr (concrete_type<ToType>) {
+        return v;
+      } else {
+        if constexpr (std::is_same_v<decltype(v), caf::expected<void>>) {
+          return std::move(v.error());
+        } else if constexpr (std::is_same_v<decltype(v),
+                                            type_to_data_t<ConcreteToType>>) {
+          return v;
+        } else {
+          if (not v)
+            return std::move(v.error());
+          return std::move(*v);
+        }
+      }
+    };
+    if constexpr (concrete_type<FromType>)
+      return caf::visit(f, detail::passthrough(from_type), to_type);
+    else if constexpr (concrete_type<ToType>)
+      return caf::visit(f, from_type, detail::passthrough(to_type));
+    else
+      return caf::visit(f, from_type, to_type);
+  }
 };
 
 template <basic_type Type>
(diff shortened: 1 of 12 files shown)


libtenzir/include/tenzir/cast.hpp
```diff
@@ -261,6 +293,45 @@ struct cast_helper<list_type, list_type> {
              from_array->null_bitmap(), from_array->null_count())
       .ValueOrDie();
   }
+
+  static auto cast_value(const list_type& from_type, const list& value,
+                         const list_type& to_type) noexcept
+    -> caf::expected<type_to_data_t<list_type>> {
+    if (from_type == to_type)
+      return value;
+    auto output = to_type.construct();
```
dominiklohmann, line 302:
```suggestion
    output.reserve(value.size());
```

libtenzir/include/tenzir/cast.hpp
```diff
@@ -335,6 +406,53 @@ struct cast_helper<record_type, record_type> {
...
+        if (not index) {
+          // The field does not exist, so we insert a null.
+          ret[to_field.name] = caf::none;
+          continue;
+        }
+        // The field exists, so we can insert the cast column.
+        auto input_at_path = descend(std::addressof(in), key);
+        if (not input_at_path)
+          return std::move(input_at_path.error());
+        if (**input_at_path == caf::none) {
```
dominiklohmann, line 441: Please add an assertion that it is safe to dereference twice here.

libtenzir/include/tenzir/cast.hpp
```diff
@@ -702,6 +820,36 @@ struct cast_helper<double_type, enumeration_type> {
...
+    if (can)
+      return value;
+    return std::move(can.error());
+  }
+
+  static auto cast(const enumeration_type&,
+                   std::shared_ptr<type_to_arrow_array_t<enumeration_type>>,
+                   const enumeration_type&) noexcept
+    -> std::shared_ptr<type_to_arrow_array_t<enumeration_type>> {
+    die("enumeration to enumeration cast unimplemented");
```
dominiklohmann, line 849:
```suggestion
    TENZIR_ASSERT(can_cast(from, to));
    return array;
```

libtenzir/include/tenzir/cast.hpp
```diff
@@ -922,26 +1070,30 @@ auto cast_value(const FromType& from_type, ValueType&& value,
...
       auto status = ret->AppendNull();
       TENZIR_ASSERT(status.ok());
       continue;
     }
-    auto converted = cast_value(from_type, *v, to_type);
-    if (not converted)
-      return converted.error();
-    if constexpr (not std::is_same_v<decltype(converted), caf::expected<void>>) {
-      auto status = append_builder(to_type, *ret, *converted);
+    auto converted = cast_value(from_type, materialize(*v), to_type);
```
dominiklohmann, line 1085: This should not need to materialize. The casting API should accept views as arguments. I think this is a small change to make, as it mostly affects records and lists.

libtenzir/include/tenzir/detail/adaptive_table_slice_builder_guards.hpp
```diff
@@ -25,6 +25,9 @@ struct type_from_data
 template <class T>
 using type_from_data_t = typename type_from_data<T>::type;
 
+using parent_record_builder_provider
+  = std::function<concrete_series_builder<record_type>*()>;
```
dominiklohmann, line 29: Please document the contract of this type.

libtenzir/include/tenzir/detail/adaptive_table_slice_builder_guards.hpp
```diff
@@ -129,26 +133,34 @@ class list_guard {
...
-                                           decltype(cast_val)>) {
-            append_value_to_builder(t, *cast_val);
+          if (cast_val) {
+            // "this" is sometimes not used due to the if constexpr below.
+            // It prevents compilation warning treated as an error.
+            (void)this;
+            // We don't handle caf::expected<void> here because the cast api
+            // only returns caf::expected<void> if the operation is not
+            // supported. The caf::expected<void> will fall into common_type
+            // casting case.
```
dominiklohmann, line 152: This seems brittle to me from an API user perspective. I would expect a cast to type T to return me an `expected<type_to_data_t<T>>`, regardless of whether that cast is possible or not.

dominiklohmann (commented):
A first look, will do a second pass later.

libtenzir/src/detail/series_builders.cpp
```diff
@@ -8,8 +8,260 @@
...
+          TENZIR_ASSERT(status.ok());
+        }
+        return should_append_builder;
+      },
+      [](auto&&...) -> bool {
+        TENZIR_WARN("unexpected data/builder/type");
+        return true;
+      },
+    },
+    type, builder, data);
```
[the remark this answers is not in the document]
patszt: done in 3a3ac78a90fc52d89c6988be523478a0a4203c9e

libtenzir/include/tenzir/cast.hpp
patszt: done in 75dbddf6033f4098ed572de5ec48384401897aef

libtenzir/include/tenzir/cast.hpp
patszt: done in 12dd25984f505a5630e02fbf12bb0785043bba53

libtenzir/include/tenzir/detail/adaptive_table_slice_builder_guards.hpp
patszt: done in ce71c24dd38c9e8d377d0c6c42b1c6c5bbc1f165

libtenzir/include/tenzir/detail/adaptive_table_slice_builder_guards.hpp
patszt: done in d19a6661b1904fd4398fd4caac06f2caa10c5bd3

libtenzir/include/tenzir/cast.hpp
patszt: done in 50991a780bcf1e4322d07b81e8cecd5f56a63951

dominiklohmann (approved):
Thanks for incorporating the feedback

PR #3336, opened 2023-07-10
Minor CI cleanup and systemd service renaming

tobim:
This renames the `tenzir` service to `tenzir-node`, so the user knows which executable is used from the name. We also add a `tenzir`  convenience alias for user that want to use a shorter name, but we stick to `tenzir-node` in the docs for clarity.

This also includes 2 small CI fixes:
* Attach the SBOM to releases early, without depending on the success of other steps.
* Don't attach the dynamic build artifact to releases at all, because they are hard to use.


dominiklohmann (commented):
Ack on the CI changes, but @lava can you review the systemd changes?

dominiklohmann (approved):
I ran the deployment workflow once inside a Docker container to verify that it still works. Both `systemctl status tenzir` and `systemctl status tenzir-node` appear to work.

PR #3339, opened 2023-07-10
Fix the suricata ingest on the demo  node

tobim:
An oversight in a last minute addition on the previous PR.


dominiklohmann (approved):
Sorry for not seeing this in the review earlier 🙈

PR #3341, opened 2023-07-11
Add magic nix cache action

dit7ya:
The magic nix cache action leverages GitHub Actions cache to speed up Nix-related CI time by 30-50%.

https://github.com/DeterminateSystems/magic-nix-cache

### DIFF
--- a/.github/workflows/nix.yaml
+++ b/.github/workflows/nix.yaml
@@ -91,6 +91,11 @@ jobs:
         uses: cachix/install-nix-action@v22
         with:
           github_access_token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Use Magic Nix Cache
+        uses: DeterminateSystems/magic-nix-cache-action@v2
+        with:
+          # disables telemetry
+          diagnostic-endpoint: ""
       - name: Setup Cachix
         uses: cachix/cachix-action@v12
         with:


tobim:
This action is certainly interesting if we look at the Nix jobs in isolation, but the cache size limit of 10GB is shared between all workflows in a repository, so we have to consider the interaction.

I need to understand what exactly this does before I can judge it.
1. Does it act as a closer cache for everything that is fetched from `cache.nixos.org` and the tenzir cachix bucket?
2. Does it also cache store paths that are built as part of the workflow?

I think we should take a look at how much of the available cache space is occupied by the items produced from this PR.

dominiklohmann:
@tobim https://github.com/tenzir/tenzir/actions/caches?query=branch%3Atopic%2Fmagic-nix-cache lists no cache for this branch.

tobim:
> @tobim https://github.com/tenzir/tenzir/actions/caches?query=branch%3Atopic%2Fmagic-nix-cache lists no cache for this branch.

It isn't attached to the branch, and it should not be.

trying to get at it with the `gh` `actions-cache` extension:
```fish
> gh actions-cache list --order desc --limit 100 | grep 'nar' | cut -f 2 | tr -d ' ' | tr -d 'B' | numfmt --from=auto | awk '{s+=$1}END{print s}' | numfmt --grouping
154,633,417
```

But the tool only gives me the first 100 cache entries and no way to get the rest...
I don't have time to use `gh api` now.

tobim (approved):
Ok, this was actually a lot easier than expected:
```fish
> gh api /repos/tenzir/tenzir/actions/caches --paginate | jq -rc '.actions_caches[]' | grep nar | tenzir -qq 'read json | summarize size=sum(size_in_bytes) | write csv' | numfmt --header --grouping
size
267,664,718
```
I think we can spare 270MB.

PR #3343, opened 2023-07-11
Add colors to JSON printer

mavam:
This PR brings a few JSON printer changes:

1. Make `--pretty` the default for `write json`
2. Ditch `--pretty` in favor of the opposite `-c|--compact-output` (like `jq`)
3. Add `-C|--color-output` and `-M|--monochrome-output` to explicitly set the color.

Unlike `jq`, there's currently no auto-detection of colors. You have to provide `-C` for now. While we could do TTY detection, it's a bit more complicated from the printer's perspective, because we need to know the corresponding saver in order to make a call. If it's stdout, we can add colors, if it's a file, we can't.

```[tasklist]
### Tasks
- [x] Implement styling logic
- [x] Support clear and colored styles
- [x] Add changelog entry
- [x] Update docs
- [x] Adapt unit tests
- [x] Adapt integration tests
```


tobim:
I'm not sure I like these new defaults. I guess I don't have a strong opinion.

Can you please add support for [`NO_COLOR`](https://no-color.org/)? An equivalent Tenzir specific config option should also be available.

mavam:
> I'm not sure I like these new defaults. I guess I don't have a strong opinion.

These are `jq` and that's the de-facto UX for JSON processing. If you don't have a strong opinion, I would strongly recommend to follow their defaults.

> Can you please add support for [`NO_COLOR`](https://no-color.org/)? An equivalent Tenzir specific config option should also be available.

I added support for `NO_COLOR` in the style detection, but don't know what you mean by "config option".

tobim:
> but don't know what you mean by "config option".

Something that I can add to my `tenzir.yaml` to make it explicit. @jachris please chime in if you already considered a generic solution for configuring operators.

mavam:
> Something that I can add to my `tenzir.yaml` to make it explicit.

This is out of scope for this PR. We don't have something like that yet.

I also don't think operators should be able to configured in an external file. Then you can no longer share pipelines and assume they do what you'd expect.

jachris:
> please chime in if you already considered a generic solution for configuring operators.

Not really, but since every operator is a plugin, you _could_ use the plugin configuration for that. 

> I also don't think operators should be able to configured in an external file. Then you can no longer share pipelines and assume they do what you'd expect.

We have the same problem with environment variables like `NO_COLOR` though. But in general, I tend to agree that we want to avoid changing the observable behavior with config options (unless it leads to an error, like `allow-unsafe-pipelines`).

tobim:
I guess not providing config options for operators is fine for most cases, purely cosmetic options like `NO_COLOR` should be exempt though.

libvast/include/vast/concept/printable/vast/json.hpp
```diff
@@ -10,21 +10,63 @@
...
+}
+
+inline auto no_style() -> json_style {
+  return {};
+}
+
+inline auto default_style() -> json_style {
+  // TODO: perform TTY detection here.
+  return no_style();
+}
```
dominiklohmann, line 60: We discussed in Slack that this cannot feasibly be done in the printer, but rather needs to be done by the saver. Can you remove this, and write a follow-up story for evaluating our options for this?

mavam: @Dakostu takes the PR from here. I'll let him address this at his convenience.

Dakostu: I wrote an issue here:
https://github.com/tenzir/issues/issues/646

I will change that TODO.

Dakostu: https://github.com/tenzir/tenzir/pull/3343/commits/a97747f6c6843df9dbf0a4778a03322baad7b6a4

vast/integration/reference/zeek-tsv-pipeline-format/step_10.ref
```diff
@@ -1,2 +1,15 @@
...
+#path	zeek.conn
+#fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	proto	service	duration	orig_bytes	resp_bytes	conn_state	local_orig	local_resp	missed_bytes	history	orig_pkts	orig_ip_bytes	resp_pkts	resp_ip_bytes	tunnel_parents	community_id	_write_ts
+#types	time	string	addr	port	addr	port	string	string	interval	count	count	string	bool	bool	count	string	count	count	count	count	vector[string]	string	time
+2009-11-18T08:00:21.486539	Pii6cUUq1v4	192.168.1.102	68	192.168.1.1	67	udp	-	163.82ms	301	300	SF	-	-	0	Dd	1	329	1	328	(empty)	-	-
+2009-11-18T08:08:00.237253	nkCxlvNN8pi	192.168.1.103	137	192.168.1.255	137	udp	dns	3.78s	350	0	S0	-	-	0	D	7	546	0	0	(empty)	-	-
+2009-11-18T08:08:13.816224	9VdICMMnxQ7	192.168.1.102	137	192.168.1.255	137	udp	dns	3.75s	350	0	S0	-	-	0	D	7	546	0	0	(empty)	-	-
+2009-11-18T08:07:15.800932	bEgBnkI31Vf	192.168.1.103	138	192.168.1.255	138	udp	-	46.73s	560	0	S0	-	-	0	D	3	644	0	0	(empty)	-	-
+warning: invalid Zeek header: #empty_field
+ = note: line 3
+warning: invalid Zeek header: #unset_field
```
mavam, line 14: Is this a bug?

Dakostu: These warnings are the output of the Zeek TSV parser.

I currently don't know exactly why the integration reference expects that output to be after the output of the Zeek TSV printer, but reading erroneous Zeek TSV files leads to the warning being printed first after any other output, so I decided to leave that for now.

jachris: This reordering is a _feature_, not a bug, to prevent non-determinism in the integration test output. The behavior on the console or the explorer should still be what we want. 

We discovered some additional non-determinism that the current implementation does not take care of (when emitting diagnostics from multiple operators), but this should not be relevant here.

mavam: I was more concerned about the actual warning *content*:

```
warning: invalid Zeek header: #unset_field
```

This *is* a valid Zeek header. (There multiple other occurrences of this same comment, just posting it once here though.)

Dakostu: The warnings appear because the file that the Zeek TSV parser is parsing contains an `#unset_field` line without a value specified afterwards. Vice versa with `#empty_field`.

Is that you mean with "valid Zeek header"?

mavam: Ah, got it! Then this comment is a no-op. Sorry.

dominiklohmann (approved):
This is working well in practice. Thanks!

libvast/builtins/formats/json.cpp
```diff
@@ -709,62 +711,85 @@ class json_parser final : public plugin_parser {
...
+};
+
+struct printer_args {
+  std::optional<location> compact_output;
+  std::optional<location> color_output;
+  std::optional<location> monochrome_output;
+  std::optional<location> omit_empty;
+  std::optional<location> omit_nulls;
+  std::optional<location> omit_empty_objects;
+  std::optional<location> omit_empty_lists;
```
jachris, line 772: Just to let you know: For those where you are not using the location, you can just use `bool` here (the argument parser works with both). But it also works as-is. Your choice.

libvast/builtins/formats/json.cpp
```diff
@@ -773,15 +798,32 @@ class json_printer final : public plugin_printer {
 
   auto instantiate(type, operator_control_plane&) const
     -> caf::expected<std::unique_ptr<printer_instance>> override {
+    const auto compact = !!args_.compact_output;
+    auto style = default_style();
+    if (detail::getenv("NO_COLOR") || args_.monochrome_output)
+      style = no_style();
+    else if (args_.color_output)
+      style = jq_style();
```
jachris, line 806: I think `NO_COLOR` usually forces the default to be non-colored. But when explicitly requested, we should still color the output.

dominiklohmann: This is confirmed by https://no-color.org, so the check should move into the `default_style()` function.

One additional note from that site: `NO_COLOR=` should also not cause colors to be disabled by default, i.e., the empty string should be the same as not having the env variable set.

dominiklohmann: Fixed in 07ffd457ca74c2f77761d1aec0cdc766f14243b6

libvast/builtins/formats/json.cpp
dominiklohmann: I don't have too much of a strong opinion on this. Probably not worth changing now.

mavam:
@dominiklohmann I saw you sneaking in the `--omit-*` commit! Planned that for another PR. But oh well, @satta will be happy. 😉

satta:
I will be happy indeed :) Thanks!

BTW, are you using `isatty()` or something similar to detect whether you're printing to a terminal or not and deactivate colors in the latter case? I found that useful in order to avoid escape sequences in redirected JSON output.

dominiklohmann:
> BTW, are you using `isatty()` or something similar to detect whether you're printing to a terminal or not and deactivate colors in the latter case? I found that useful in order to avoid escape sequences in redirected JSON output.

We discussed somewhere in a thread for this PR that we actually need to pass that information up from the last operator, and wrote a follow-up story for that. For now, colors are opt-in.

E.g., if you run this:

```
tenzir '… | write json | save file foo.json'
```

Then the information of whether or not we're printing to a terminal must be passed from `save` to `write`. We talked about the how, but it's for sometime in the future because `write` cannot reliably detect that on its own.

satta:
I see, true!

PR #3344, opened 2023-07-11
Fix a crash when the JSON document is malformed

patszt:
When the JSON document is malformed it sometimes crashes when we try to continue to the next document.
It seems safer to just break here. The error message already contains the broken JSON and will be recoverable in the future when we implement "failed to parse" error channel.

### DIFF
--- a/libvast/builtins/formats/json.cpp
+++ b/libvast/builtins/formats/json.cpp
@@ -114,6 +114,9 @@ struct parser_state {
   std::string last_used_schema_name{};
   // Table slice builder used when the schema is not known.
   adaptive_table_slice_builder unknown_schema_builder{};
+  // Used to communicate a need for a co_return in the operator coroutine from
+  // the ndjson parser/default parser coroutine.
+  bool abort_requested = false;
 };
 
 template <class FieldValidator>
@@ -512,7 +515,11 @@ public:
     // After parsing one JSON object it is expected for the result to be at
     // the end. If it's otherwise then it means that a line contains more than
     // one object in which case we don't add any data and emit a warning.
-    if (not doc.at_end()) {
+    // It is also possible for a parsing failure to occurr in doc_parser. the
+    // is_alive() call ensures that the first object was parsed without errors.
+    // Calling at_end() when is_alive() returns false is unsafe and resulted in
+    // crashes.
+    if (doc.is_alive() and not doc.at_end()) {
       row.cancel();
       this->ctrl_.warn(caf::make_error(
         ec::parse_error, fmt::format("more than one JSON object in a "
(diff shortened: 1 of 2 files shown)


libvast/builtins/formats/json.cpp
```diff
@@ -555,7 +555,9 @@ class default_parser : public parser_base<FieldValidator> {
         this->ctrl_.warn(caf::make_error(
           ec::parse_error, fmt::format("skips invalid JSON '{}' : {}", view,
```
lava, line 556: The error message is incorrect now

patszt: I think it was incorrect to begin with.
I have an issue for it: https://github.com/tenzir/issues/issues/638

dominiklohmann, line 556: I don't understand how we can feasibly continue here. How can we know where the next JSON object starts?

dominiklohmann: We looked at this together in a call. `stream.truncated_bytes()` must only be called after iterating through the entire stream, which now we're no longer doing. That is, if the documentation of simdjson is correct then this must indeed be a hard-error and not a warning.

libvast/builtins/formats/json.cpp
```diff
@@ -588,7 +596,10 @@ class default_parser : public parser_base<FieldValidator> {
     // to be used with well formed JSON or truncated JSON. In this case we don't
     // know how to recover. It might be possible to use different parser or our
     // custom logic to try recover as much data as possible.
+    // TODO: revisit if this is needed now that we abort and co_return in
```
patszt, line 599: This would take some time to investigate so i'd rather leave a TODO here. Should i open an issue for it and remove the TODO from the code? Or leave the todo and create issue?

libvast/builtins/formats/json.cpp
patszt: I didn't find much on this. So it would be safe to just assume we can't call it when the iteration is not complete

patszt: Now that we abort instead of warning it is unlikely to be a wrong error message.
I still think that when we detect a schema change and yield slice just before aborting we may give a wrong error message.

E.g we might have this data in a single chunk
```json
{"pcap_cnt":83,"event_type":"alert"}
{"pcap_cnt":83,"event_type":"dns"}
{"pcap_cnt":83,"event_type":"dns"{}
```
We parse the first event successfully.
We yield the first event because the selector field for the second event changed.
We parse the second event.
We abort on third event.

I think here it might be non deterministic. Will we process the yielded slice first? Or will the call to abort stop the execution.

Notice that it happens only with read json with --selector option. It doesn't happen for zeek/suricata.
I will update the mentionted issue accordingly

libvast/builtins/formats/json.cpp
dominiklohmann: I think that we should not be able to run into this anymore now, so the comment should say something along the lines of "likely not needed anymore, need to investigate," and then in the future this should be an assertion.

libvast/builtins/formats/json.cpp
dominiklohmann: Let's tackle this in the follow-up issue.

dominiklohmann (approved):

PR #3345, opened 2023-07-11
Improve json escaping and error handling

jachris:
This mainly fixes an issue where non-breaking spaces led to an invalid json response, but also contains some additional, related improvements.


libvast/src/http_api.cpp
```diff
@@ -54,29 +54,29 @@ auto parse_skeleton(simdjson::ondemand::value value, size_t depth = 0)
...
+    http_parameter_map result;
+    if (json.empty())
+      return result;
+    auto padded_string = simdjson::padded_string{json};
+    auto parser = simdjson::ondemand::parser{};
+    auto body = simdjson::ondemand::object{};
+    auto doc = parser.iterate(padded_string).value();
+    if ([[maybe_unused]] auto error = doc.get_object().get(body))
+      return caf::make_error(ec::invalid_argument,
+                             "expected a top-level object");
```
lava, line 67: If we're switching to exceptions we can probably remove the use of the non-throwing API here as well and rely completely on them.

lava (approved):

PR #3346, opened 2023-07-11
Better enforce that the rest response contains valid JSON

lava:


libvast/src/http_api.cpp
```diff
@@ -8,12 +8,34 @@
...
 #include <simdjson.h>
 
+namespace {
+
+// Validates that the argument is a valid JSON object.
+auto validate_json(const std::string& json) -> bool {
+  auto parser = simdjson::dom::parser{};
+  auto padded = simdjson::padded_string{json};
+  auto obj = simdjson::dom::object{};
+  return parser.parse(json).get(obj) == simdjson::SUCCESS;
```
jachris, line 25: This does not work, because `simdjson` does lazy validation.

lava: Also for the `dom` parser? I thought it was only for the `ondemand` parser.

jachris: Sorry, you are correct :+1: I thought it would be `ondemand`, but you are using `dom`.

jachris (approved):

libvast/src/http_api.cpp
```diff
@@ -223,7 +254,8 @@ auto rest_response::make_error(uint16_t error_code, std::string message,
 
 auto rest_response::make_error_raw(uint16_t error_code, std::string body,
                                    caf::error detail) -> rest_response {
-  auto result = rest_response{std::move(body)};
+  auto result = rest_response{};
+  result.body_ = std::move(body);
```
jachris, line 258: I think this would also benefit from a validity check.

PR #3350, opened 2023-07-12
Add blog post on shell operator

mavam:
This PR adds a blog post about the `shell` operator, and how to profit from it using Zeek and Suricata.


web/blog/zeek-in-the-shell/index.md
```diff
@@ -0,0 +1,179 @@
...
+our favorite tools, and then perform fast in-situ analytics on them. The nice
+thing is that operators compose: a new operator automatically works with all
+existing ones.
+
+## How does it work?
+
+Under the hood, the `shell` operator forks the `tenzir` process and spawns
+`zeek` or `suricata` as child process. The operator then rewires stdin and
+stdout of the `tenzir` process such their byte streams get routed through stdin
+and stdout of the forked child.
```
jachris, line 126: And similar to one of the previous comments: I think the flow could be improved by either moving this after the "standard approach", or by moving the standard approach after the `shell` approach.

web/blog/zeek-in-the-shell/index.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+The difference stands out when you look now at the pipeline definition:
+
+```text title=Zeek
+zeek
+| where #type == "zeek.conn"
+| summarize n=count_distinct(id.orig_h) by id.resp_h
+| sort n desc
+```
```
jachris, line 107: Does this actually work as intended? `shell` is both `bytes -> bytes` and `void -> bytes`. We do not add an implicit `load -` if the pipeline can be instantiated for `void`. 

You might be able to resolve this by adding `load -` to the alias definition, but then it cannot be used anymore as an inner operator. I think we don't have a clean solution for this at the moment.

jachris, line 107: Also, we have an open issue for when shell is used as a source. As per the comment above, you maybe do not want do this here, and the issue may not apply to this situation, but just to let you know: https://github.com/tenzir/issues/issues/619

web/blog/zeek-in-the-shell/index.md
```diff
@@ -0,0 +1,179 @@
...
+![Piping Zeek to Tenzir](zeek-to-tenzir-pipe.excalidraw.svg)
+
+Using `shell`, we have this scenario:
+
+![Shelling out to Zeek](zeek-to-tenzir-shell.excalidraw.svg)
+
+The `shell` operator is bytes-to-bytes transformation. This means it can only
+interact with other operators that also work on bytes. In the above examples, we
+used the shell pipe to connect stdin and stdout of two processes. We can also
+use Tenzir's native pipe with the [`file`](/docs/connectors/file) loader:
```
jachris, line 139: But just to be clear: I think explaining it as a bytes-to-bytes transformation is still a useful simplification, even if this is not 100%, because from the POV of the user, it is not really relevant here that the pipeline starts with `void`. However, this doesn't solve the implementation issue.

web/blog/zeek-in-the-shell/index.md
```diff
@@ -0,0 +1,179 @@
...
+shell.
+
+:::info Unsafe operators
+Naturally, we must add that this is also a security risk and can lead to
+arbitrary shell access. This is why the `shell` operator and a few other ones
+with far-reaching side effects are disabled by default. If you are aware of the
+implications, you can remove this restriction by setting
+`tenzir.allow-unsafe-pipelines: true` in the `tenzir.yaml` of the respective
+node.
+:::
```
jachris, line 173: They are only disabled when using a node, but I think it's fine to ignore this detail here.

jachris (approved):
Nice blog post! 

FYI: I didn't run the pipelines, but besides what I have written in the comments, it looks like it should work.

web/blog/zeek-in-the-shell/index.md
```diff
@@ -0,0 +1,179 @@
...
+
+![Shell Operator](shell-operator.excalidraw.svg)
+
+<!--truncate-->
+
+## Purring like a Suricat
+
+In our [previous blog](/blog/zeek-and-ye-shall-pipe/) we adapted Zeek to behave
+like a good ol' Unix tool, taking input via stdin and producing output via
+stdout. Turns out you can do the same fudgery with Suricata:[^1]
```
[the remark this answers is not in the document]
jachris: Or maybe "article". I might also be mistaken, but I think calling it a "blog" (also in the PR title and description) is incorrect.

web/blog/zeek-in-the-shell/index.md
```diff
@@ -0,0 +1,179 @@
...
+
+For our use case we look at the [`shell`](/docs/operators/shell) operator in
+more detail: it lifts any Unix command line into a pipeline. Here are two
+examples that rely on our two shell scripts:
+
+```bash
+# Zeek
+zcat pcap.gz | zeekify | tenzir \
+  'read zeek-json
+   | where #type == "zeek.conn"
```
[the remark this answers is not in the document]
mavam: You can see when I drafted this blog. 😜

web/blog/zeek-in-the-shell/index.md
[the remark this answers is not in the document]
mavam: I've reworded this according to your feedback.

web/blog/zeek-in-the-shell/index.md
jachris: My comment is not about the output type of `zeek` or `shell`, but rather its input type. The pipeline above is a well-formed `void -> events` pipeline (and also a well-formed `bytes -> events` pipeline, but we "prefer" the `void -> events` type in `exec.cpp`). Unless you prefix it with another operator, we will thus not add an implicit `load -`, and your `shell` implementation is therefore instantiated for `void`. I might be mistaken, but I think this means that we do not forward the stdin of the Tenzir process to `shell`.

jachris: Yes, you can resolve this one here.

web/blog/zeek-in-the-shell/index.md
mavam: I actually removed this bytes-to-bytes thing. Let me know what you think. I feel it's easier to read now.

web/blog/zeek-in-the-shell/index.md
mavam: Ah, so the fix would be to make it a transform by appending `load -` in the UDO definition?

jachris: That would work for your examples, but then you of course can't use it anymore as an inner (non-source) operator. The only alternatives that I can think of are changing `shell_operator::operator()(operator_control_plane&)` to steal `stdin` (but I'm not sure whether we want to do that), or prefer `bytes -> void` in `exec.cpp` (such that we add an implicit `load -` if the pipeline begins with `shell` or other operators that can act both as a source and as a transformation).

mavam: I think I get it now, but I'm too far away from the code in the operator control plane. @dominiklohmann thoughts on how to proceed here? Should we factor this into a separate issue?

As user, I would like UDOs to work transparently, i.e., use them as source and transform as in this blog. Let's try to aim for this UX.

dominiklohmann: There are two ways to get that to work:
1. Change the `void -> bytes` version of `shell` to get the entire stdin of the parent process.
2. Remove the `void -> bytes` version of `shell` and always rely on the implicit `load -` source.

This is not an easy call to make:
- Option (1) will make the void -> bytes version of the operator clash with @mavam's `tui` operator, because only one operator may use stdin.
- For deployed pipelines via the pipeline manager API, you don't get implicit sources and sinks, so option (2) would require adding an explicit `load file -` for deployed pipelines that start with `shell` currently.

I think (2) is the way to go. It's a fast change to make, but also a breaking change in the API, so we should make it on the sooner side.

jachris: But (2) + implicit source has the same effect as (1) in that it breaks the `tui` operator. Yes, I can work around that with `load file /dev/null`, but I can also do that when we keep it. So the only real difference is that I have now have to add `load file -` when using the explorer.

jachris: At the same time: When using the pipeline manager, I feel like you don't want to grab `stdin` (why would you do that?). So what  preferring `bytes -> void` over `void -> void` in `exec.cpp`, and keeping the `shell` overloads as-is? As written above, I think that should work. Are there any negative side effects from this?

dominiklohmann: > But (2) + implicit source has the same effect as (1) in that it breaks the `tui` operator.

Not sure how I missed that, you are absolutely correct.


> So what preferring `bytes -> void` over `void -> void` in `exec.cpp`,

I don't think this is a good solution. If a pipeline can be considered closed then we should not be touching it.

I think that the current behavior is the behavior we want to have: If the user wants `shell` to read from stdin of the `tenzir` process, then they must explicitly say so with `load file -`.

jachris: I just noticed that you could also add a `pass` in front of the user-defined operator definition to enforce that this shell operator still needs some input... In this situation, I think this would lead to the perfect behavior, but it's quite hacky.

But yea, it's a difficult situation.

One more option (as if 4 options were not enough): We could also detect whether `stdin` is connected to a pipe to give `foo | tenzir 'shell "bar" | baz` the right behavior.

dominiklohmann: Doing any form of TTY detection is a pain for people using our software through Docker, so we should be very careful about that. I would like to avoid it unless we absolutely must.

web/blog/shelling-out-to-zeek-and-suricata/index.md
```diff
@@ -0,0 +1,179 @@
...
+   | sort n desc'
+# Suricata
+zcat pcap.gz | suricatify | tenzir \
+  'read suricata
+   | where #schema == "suricata.flow"
+   | summarize n=count_distinct(src_ip) by dst_ip
+   | sort n desc'
+```
+
+It's a bit unwiedly to write such a command line that requires an external shell
```
jachris, line 77:
```suggestion
It's a bit unwieldy to write such a command line that requires an external shell
```

web/blog/shelling-out-to-zeek-and-suricata/index.md
```diff
@@ -0,0 +1,179 @@
...
+script to work. This is where [user-defined operators](/operators/user-defined)
+come into play. In combination with the [`shell`](/operators/sources/shell)
+operator in more detail, you can write a custom `zeek` and `suricata` operator
+and ditch the shell script:
```
jachris, line 81:
```suggestion
come into play. In combination with the [`shell`](/operators/sources/shell)
operator, you can write a custom `zeek` and `suricata` operator
and ditch the shell script:
```

web/blog/shelling-out-to-zeek-and-suricata/index.md
```diff
@@ -0,0 +1,179 @@
...
+
+First, let's take a look at the standard approach where one process pipes the
+output into the next:
+
+![Piping Zeek to Tenzir](zeek-to-tenzir-pipe.excalidraw.svg)
+
+When using the `shell` operator, the `tenzir` process spawns `zeek` or
+`suricata` as child process. The operator then copies the bytes from stdin of
+the `tenzir` process to the child's stdin using a Unix pipe (and in reverse for
+stdout):
```
jachris, line 130:
```suggestion
When using the `shell` operator, the `tenzir` process spawns `zeek` or
`suricata` as child process. The operator then forwards the bytes from stdin of
the `tenzir` process to the child's stdin, and uses the child's stdout as the input 
to the subsequent `parse` operator.
```

The changes to "copies" and "using a Unix pipe" (which I think is irrelevant) are just opinionated, but "and in reverse for stdout" is objectively incorrect.

web/blog/zeek-in-the-shell/index.md
jachris: Well, then I fear we are out of options here. The only solution that checks all boxes is adding the `load file -` explicitly, but you can try out the effect on the pipelines in this post. I don't like it too much.

IMHO the best of all these non-optimal solutions is preferring `bytes -> void` in `exec.cpp`, as this should just "do the right thing", albeit at the cost of a small impurity. It's also how the shell itself works when using pipelines. But I'll let you make the call.

mavam: > Well, then I fear we are out of options here. The only solution that checks all boxes is adding the `load file -` explicitly, but you can try out the effect on the pipelines in this post. I don't like it too much.

The consequence would be that you cannot use it as transformation anymore, or combine with a different loader. I think that's a big limitation because it breaks composition—which is what operators are all about.

dominiklohmann: Jannis and I just had a call on this and agreed to make the following change: We want to let operators know whether they are allowed to acquire stdin.

We will then allow this for pipelines started via `tenzir <pipeline>`, and forbid it for all other things that start pipelines. The void-to-bytes `shell` operator changes its behavior based on this flag.

Jannis will implement this before the targeted release date of this blog post in a separate PR.

web/blog/shelling-out-to-zeek-and-suricata/index.md
```diff
@@ -0,0 +1,179 @@
...
+
+```yaml title="tenzir.yaml"
+tenzir:
+  operators:
+    zeek: >
+     shell "zeek -r - LogAscii::output_to_stdout=T
+            JSONStreaming::disable_default_logs=T
+            JSONStreaming::enable_log_rotation=F
+            json-streaming-logs"
+     | parse zeek-json
```
jachris, line 91: Opening this conversation here to remind us of the `parse -> read` change, which is scheduled to happen before this post is published.

web/blog/zeek-in-the-shell/index.md
dominiklohmann: This now works as expected, I think we can resolve this thread.

web/blog/shelling-out-to-zeek-and-suricata/index.md
dominiklohmann: Since that PR was just merged we need to replace parse with read in this blog post before merging.

web/blog/zeek-in-the-shell/index.md
mavam: Great. Did you test it yourself?

mavam: (Feel free to resolve if you think it's sufficiently addressed.)

jachris: We tested that the `shell` operator should now behave as required when in source position, but not the concrete example in this post. If this suffices for you, please resolve.

oliverrochford:
I am not sure about the title. What's the allusion to?

mavam:
> I am not sure about the title. What's the allusion to?

By "shelling out" I meant to convey that using the `shell` operator has the same UX as typing something on the command line in a shell. Here: running Zeek & Suricata from the shell.

Suggestions for better titles are always welcome.

mavam:
Oh, I wasn't aware of this connotation. Glad you brought it up.

I liked your suggestion and adapted the title accordingly. Thanks!

PR #3351, opened 2023-07-12
Perform a few CI tune-ups

mavam:
- Switch to "depedencies" label for dependabot
- Remove issue templates


mavam:
We're still getting new labels like `python` and `javascript`. I couldn't figure out to disable them. Help would be appreciated.

dominiklohmann (approved):

PR #3357, opened 2023-07-13
Add support for merge queues to the CI setup

tobim:


.github/workflows/vast.yaml
```diff
@@ -15,6 +15,9 @@ on:
     types:
       - opened
       - synchronize
+  merge_group:
+    types:
+      - checks_requested
```
dominiklohmann, line 20: This also needs to be done for the analysis, blocked, and style-check workflow files.

dominiklohmann (approved):
Reading through the docs I think this is a safe change to make. We did discuss in Slack how to handle the rev count in the version.json file, which I anticipate to clash a bit with merge queues. But we can fix that in a follow-up PR.

.github/workflows/vast.yaml
dominiklohmann, line 20: Below this we have the concurrency.cancel_in_progress section. We should consider also enabling that for merge queues.

dominiklohmann: We talked about this quickly in the office, and agreed that it's just style check we need to add. I misunderstood the trigger.

tobim: We agreed that `Tenzir` and `Style Check` are all we need.

dominiklohmann (approved):

PR #3358, opened 2023-07-13
Remove previously deprecated options

dominiklohmann:
The Tenzir v4.0 release is a major release. This change makes use of the opportunity to remove some previously deprecated options:
- Remove option `tenzir.pipelines` (replaced with `tenzir.operators`)
- Remove option `tenzir.pipeline-triggers` (no replacement)
- Remove deprecated type tokens `addr`, `count`, `int`, and `real` (replaced with `ip`, `uint64`, `int64`, and `double`, respectively)
- Remove support for reading from segment stores (read-only since VAST v2.3, migrated automatically over time to Feather and/or Parquet)


tobim (approved):
Didn't look too closely, the important stuff is covered by CI.

dominiklohmann:
> Didn't look too closely, the important stuff is covered by CI.

That actually ended up finding one last missing `list<int>` that I just changed to `list<int64>`. Pushed that, should be ready to go now.

PR #3360, opened 2023-07-13
The Big Rename

tobim:
This PR changes all remaining occurrences of the word 'vast' with 'tenzir', and is applied for the various capitalizations that exist in the code base.

TODO:
- [x] adapt the python bindings
- [ ] ~Change the text banner at the top of all source files.~ Needs some discussion.


dominiklohmann (approved):
We reviewed this in a call and the feedback was integrated directly.

PR #3363, opened 2023-07-14
Lower default amount of returned events, and return a continuation token

lava:


libtenzir/builtins/operators/serve.cpp
```diff
@@ -663,6 +666,14 @@ struct serve_handler_state {
       TENZIR_ASSERT_CHEAP(ok);
     }
     out_iter = fmt::format_to(out_iter, "}}]}}{}", '\n');
+    if (result.size() > defaults::api::max_response_size) {
+      VAST_WARN("serve-manager discards {} results because the response is too "
+                "big",
+                num_events);
+      return fmt::format(
+        R"({{"next_continuation_token":"{}","events":[], "schemas": []}})",
```
lava, line 674: We should consider adding a 'warnings' field directly to the response here so this is not happening silently from the FE perspective, but it needs some coordination on how this should be displayed.

tobim: There is also the pipeline diagnostics format. We should check whether this issue fits into that scheme.

tobim (approved):
Approving since this fixes a severe issue and we can add the missing diagnostics later.

libtenzir/builtins/operators/serve.cpp
```diff
@@ -663,6 +666,13 @@ struct serve_handler_state {
       TENZIR_ASSERT_CHEAP(ok);
     }
     out_iter = fmt::format_to(out_iter, "}}]}}{}", '\n');
+    if (result.size() > defaults::api::max_response_size) {
+      TENZIR_WARN("serve-manager discards oversized response with {} events",
+                  num_events);
+      return fmt::format(
+        R"({{"next_continuation_token":"{}","events":[], "schemas": []}})",
```
dominiklohmann: Nit: The spaces in here feel a bit arbitrary.

libtenzir/builtins/operators/serve.cpp
```diff
@@ -663,6 +666,13 @@ struct serve_handler_state {
       TENZIR_ASSERT_CHEAP(ok);
     }
     out_iter = fmt::format_to(out_iter, "}}]}}{}", '\n');
+    if (result.size() > defaults::api::max_response_size) {
+      TENZIR_WARN("serve-manager discards oversized response with {} events",
+                  num_events);
```
dominiklohmann, line 675: The user running the app does not know what serve is in any way, because they don't write the sink themselves.

Because of this I think this warning is not actionable for users. We should do one of two things:
1. Transport it back to the operator such that we can emit a warning diagnostic for it with proper context that helps the user.
2. Lower the level to debug or verbose.

libtenzir/include/tenzir/defaults.hpp
```diff
@@ -163,27 +163,30 @@ inline constexpr const size_t rotate_files = 3;
...
+/// The maximum response size for one request.
+inline constexpr size_t max_response_size = size_t{124} * 1'024; // 124 KiB
 
-/// The maximum number of results
-inline constexpr const size_t limit = 50;
+namespace serve {
 
-/// Whether to flatten nested fields.
-inline constexpr const bool flatten = false;
+/// The duration for which old results are kept after being fetched.
```
dominiklohmann, line 174: This comment is incorrect. The timeout only applies for the last result set for which the returned continuation token is null. For all other result sets we keep the results around until the next continuation token is used in a request successfully.

libtenzir/builtins/operators/serve.cpp
```diff
@@ -597,7 +598,7 @@ struct serve_handler_state {
                          .detail = std::move(detail)};
     }
     if (*timeout) {
-      if (**timeout > std::chrono::seconds{5}) {
+      if (**timeout > defaults::api::serve::max_timeout) {
         auto detail = caf::make_error(
           ec::invalid_argument, fmt::format("got timeout {}", data{**timeout}));
         return parse_error{.message = "timeout exceeds limit of 5 seconds",
```
dominiklohmann, line 608: The message should reference the default now that it is no longer hardcoded.

PR #3364, opened 2023-07-14
Create a new /ping endpoint

lava:
In order to implement HTTP health checks, we want an endpoint that always returns instantly and puts no strain on the system. The existing `/status` is not suited for this, since it can take up to 10s to respond if actors are not responsive.

The ping response contains the node version as payload in order to give a low-effort alternative to the `version` operator for developers trying to determine the version of a remote node.

### DIFF
--- a/libtenzir/builtins/endpoints/ping.cpp
+++ b/libtenzir/builtins/endpoints/ping.cpp
@@ -3,23 +3,22 @@
 //   | |/ / __ |_\ \  / /          Across
 //   |___/_/ |_/___/ /_/       Space and Time
 //
-// SPDX-FileCopyrightText: (c) 2022 The VAST Contributors
+// SPDX-FileCopyrightText: (c) 2022 The TENZIR Contributors
 // SPDX-License-Identifier: BSD-3-Clause
 
-#include <vast/builtin_rest_endpoints.hpp>
-#include <vast/node.hpp>
-#include <vast/plugin.hpp>
-#include <vast/version.hpp>
+#include <tenzir/builtin_rest_endpoints.hpp>
+#include <tenzir/plugin.hpp>
+#include <tenzir/version.hpp>
 
 #include <caf/typed_event_based_actor.hpp>
 
-namespace vast::plugins::rest_api::status {
+namespace tenzir::plugins::rest_api::ping {
 
 static auto const* SPEC_V0 = R"_(
-/version:
+/ping:
   get:
-    summary: Return node version
-    description: Returns the version number of the node
+    summary: Returns a success response
+    description: Returns a success response to indicate that the node is able to respond to requests. The response body includes the current node version.
     responses:
       200:
         description: OK.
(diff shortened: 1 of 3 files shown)


libtenzir/builtins/endpoints/ping.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+auto ping_handler(ping_handler_actor::stateful_pointer<ping_handler_state> self)
+  -> ping_handler_actor::behavior_type {
+  return {
+    [self](atom::http_request, uint64_t,
+           const tenzir::record&) -> caf::result<rest_response> {
+      TENZIR_DEBUG("{} handles /ping request", *self);
+      auto versions = retrieve_versions();
+      TENZIR_ASSERT_CHEAP(versions.contains("Tenzir"));
+      return rest_response{record{{{"version", versions.at("Tenzir")}}}};
```
dominiklohmann, line 56: Instead of this, you can also just use the `tenzir::version::version` string from the CMake-generated `tenzir/config.hpp` file.

libtenzir/builtins/endpoints/ping.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+class plugin final : public virtual rest_endpoint_plugin {
+  auto initialize([[maybe_unused]] const record& plugin_config,
+                  [[maybe_unused]] const record& global_config)
+    -> caf::error override {
+    return {};
+  }
+
+  [[nodiscard]] auto name() const -> std::string override {
+    return "api-ping";
```
dominiklohmann, line 69: Any good reason for the `api-` prefix?

dominiklohmann (approved):
This is good to go, just two small comments and a unit test that requires fixing. But the endpoint itself works as expected when tested manually.

libtenzir/builtins/endpoints/ping.cpp
tobim: Imho just `ping` would make more sense to me. I'm not sure what information is given by the extra `api-`.

lava: It's to avoid name clashes, for ping it's not so pressing because it's actually completely self-contained, but for plugin names like  `status` or `export`  these would be a bit a bit too general since they just refer to the endpoint.

Anyways, I've changed it for now.

libtenzir/builtins/endpoints/ping.cpp
```diff
@@ -0,0 +1,103 @@
...
+    TENZIR_ASSERT(result);
+    return *result;
+  }
+
+  /// List of API endpoints provided by this plugin.
+  [[nodiscard]] auto rest_endpoints() const
+    -> const std::vector<rest_endpoint>& override {
+    static const auto endpoints = std::vector<rest_endpoint>{
+      {
+        .endpoint_id = static_cast<uint64_t>(status_endpoints::status),
```
tobim, line 84: Should this not be `ping_endpoints::ping`?

tobim (commented):
Maybe this needs some  more fixes?

PR #3372, opened 2023-07-18
Fix sporadic broken promise error when pulling fails

jachris:


libtenzir/src/execution_node.cpp
```diff
@@ -834,6 +848,9 @@ auto exec_node(
     return exec_node_actor::behavior_type::make_empty_behavior();
   }
   self->state.weak_node = node;
+  self->attach_functor([name = self->state.op->name()] {
+    TENZIR_DEBUG("exec node for {} shuts down", name);
```
dominiklohmann, line 852: This should be past tense, because the functor is executed by the clock _after_ the actor was shut down, i.e., it is not safe to access `*self` anymore.

```suggestion
    TENZIR_DEBUG("exec-node for {} shut down", name);
```

dominiklohmann (approved):
Thanks for this. I am still unable to reproduce the issue locally, but the explanation for it makes sense. Let's see if this fixes the sporadic CI failures.

jachris:
> Thanks for this. I am still unable to reproduce the issue locally, but the explanation for it makes sense. Let's see if this fixes the sporadic CI failures.

I wrote a script that runs 32 Tenzir instances over and over until the issue occurs. How fast this happened also depended on the system load (makes sense, because it's triggered by the scheduler).

PR #3373, opened 2023-07-18
Pre-import demo data in the demo node

tobim:
This moves the demo data import from the container startup to the image creation phase.
I also added 2 alternative stopped pipelines that are continuous by nature.

### DIFF
--- a/demo-node/entrypoint-noingest.bash
+++ b/demo-node/entrypoint-noingest.bash
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+export TENZIR_AUTOMATIC_REBUILD="${TENZIR_AUTOMATIC_REBUILD:-0}"
+export TENZIR_ALLOW_UNSAFE_PIPELINES=true
+
+# Forward signals as SIGTERM to all children.
+trap 'trap " " SIGTERM; kill 0; wait' SIGINT SIGTERM
+
+coproc NODE { exec tenzir-node --print-endpoint --commands="web server --mode=dev"; }
+# shellcheck disable=SC2034
+read -r -u "${NODE[0]}" DUMMY
+
+# Reset the verbosity so the logs don't get spammed by dozens
+# of Tenzir processes.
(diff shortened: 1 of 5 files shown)


demo-node/entrypoint-noingest.bash
```diff
@@ -0,0 +1,35 @@
...
+# of Tenzir processes.
+export TENZIR_CONSOLE_VERBOSITY="${TENZIR_CLIENT_CONSOLE_VERBOSITY:-"info"}"
+
+# Wait until the HTTP endpoint is listening.
+while ! timeout 100 bash -c "echo > /dev/tcp/127.0.0.1/5160"; do
+  echo "Waiting for the HTTP service..."
+  sleep 1
+done
+
+nic_pipe="from nic eth0 | drop data | import"
```
mavam, line 23: Where do we provide capture capabilities to the container?

mavam (approved):
Approving the code, but I haven't tested it.

demo-node/entrypoint-noingest.bash
tobim: There is no need to. `pcap.packet` data does show up in the database, unfortunately at a very low rate, our app traffic seems to take some other path. Unfortunately `shell "ip -j a"` does not work in fargate.

demo-node/import.bash
```diff
@@ -0,0 +1,21 @@
...
+trap 'trap " " SIGTERM; kill 0; wait' SIGINT SIGTERM
+
+coproc NODE { exec tenzir-node --print-endpoint; }
+# shellcheck disable=SC2034
+read -r -u "${NODE[0]}" DUMMY
+
+tenzir 'read suricata | where #schema != "suricata.stats" | import' < Suricata/eve.json
+cat Zeek/*.log | tenzir 'read zeek-tsv | import'
+rm -rf Suricata
+rm -rf Zeek
```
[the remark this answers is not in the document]
tobim: The image size seems to be almost unaffected, so it seems fine to leave them.

PR #3375, opened 2023-07-19
Add blog on Zeek log rotation and shipping

mavam:
This PR adds a blog on Zeek's native log rotation feature and how to make it work with Tenzir.


web/blog/native-zeek-log-rotation-and-shipping/index.md
```diff
@@ -0,0 +1,195 @@
...
+
+In case you're not using Zeek's native log rotation trigger, you may observe a
+directory to which Zeek periodically writes files.
+
+Log shippers can take care of that, but your mileage may vary. For example,
+[Filebeat][filebeat] works for stock Zeek only. Every log file is hard-coded. If
+you have custom scripts or extend some logs, you're left alone. Filebeat also
+uses the stock Zeek JSON output, which has no type information. Filebeat then
+brings the typing back manually later as it converts the logs to the Elastic
+Common Schema (ECS).
```
tobim, line 30: The shortcoming of filebeat here is that the Zeek integration lacks  modularity. 

But for this tool in particular, you would not monitor the rotated logs but the live ones instead. Filebeat tracks a read offset for each log file internally and implements precautions to prevent double-shipping of log lines when the files get rotated. One of the advantages is a much lower latency compared to rotation based shipment. The lack of support for user or script defined logs is really what is hampering filebeat in particular, but a similar tool that does not insist on a particular taxonomy would certainly be a viable alternative.

I think what I'm trying to say is that filebeat is not a great example because of the problems with ECS.

tobim (changes requested):
I think the post is scoped and focused well. But we can improve on the story line. I would keep the narrow focus on pull vs push, and explain what use cases are best catered to by each approach. We can also mention that streaming logs to another service directly is another option to consider if neither of the presented approaches meet the users needs.

I think we should also rank all approaches in the categories `complexity`, `robustness`,  and `latency`, and add a comparison table as a summary.

web/blog/native-zeek-log-rotation-and-shipping/index.md
```diff
@@ -0,0 +1,195 @@
...
+running out of memory, it takes all children down along with it.
+
+This may not be a problem for one-shot trace file analysis, but live deployments
+may require higher availability and consistency. For such scenarios, we
+recommend *the opposite* strategy: **do not run the post-processing within Zeek.
+Decouple it instead.** One way to achieve this is by observing side effects. For
+example, [zeek-archiver](https://github.com/zeek/zeek-archiver) watches a
+directory for new log files that Zeek rotates in there. We have a dedicated
+[roadmap item](https://github.com/tenzir/public-roadmap/issues/51) directory
+watching.
```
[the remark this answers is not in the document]
mavam: I clarified this by saying that we plan something _similar_.

web/blog/native-zeek-log-rotation-and-shipping/index.md
```diff
@@ -0,0 +1,202 @@
...
+In case you're not using Zeek's native log rotation trigger, you may observe a
+directory to which Zeek periodically writes files. For example, the utility
+[zeek-archiver](https://github.com/zeek/zeek-archiver) does that.
+
+Generic log shippers can take care of that as well. Your mileage may vary. For
+example, [Filebeat][filebeat] works for stock Zeek only. Every log file parsing
+logic is hard-coded. If you have custom scripts or extend some logs, you're left
+alone. Filebeat also uses the stock Zeek JSON output, which has no type
+information. Filebeat then brings the typing back manually later as it converts
+the logs to the Elastic Common Schema (ECS).
```
tobim, line 32:
```suggestion
example, [Filebeat][filebeat] works for stock Zeek only. The parsing logic is
hard-coded for every log type. If you have custom scripts or extend some logs,
you're left alone. Filebeat also uses the stock Zeek JSON output, which has no
type information. Filebeat then brings the typing back manually later as it
converts the logs to the Elastic Common Schema (ECS).
```

web/blog/native-zeek-log-rotation-and-shipping/index.md
```diff
@@ -0,0 +1,202 @@
...
+
+[filebeat]: https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-module-zeek.html
+
+## Native Log Shipping (push)
+
+There's also a lesser known, push-based option using [Zeek's logging
+framework](https://docs.zeek.org/en/master/frameworks/logging.html). You can
+provide a shell script that Zeek invokes *whenever it rotates file*. The shell
```
tobim, line 40:
```suggestion
provide a shell script that Zeek invokes *whenever it a rotates file*. The shell
```

web/blog/native-zeek-log-rotation-and-shipping/index.md
```diff
@@ -0,0 +1,202 @@
...
+script receives the filename of the rotated file as argument, plus some
+additional metadata.
```
tobim, line 42:
```suggestion
script receives the filename of the rotated file plus some additional metadata
as arguments.
```

web/blog/native-zeek-log-rotation-and-shipping/index.md
```diff
@@ -0,0 +1,202 @@
...
+     }
+   ```
+
+   The above Zeek script hooks up two pipelines via the function
+   `Tenzir::postprocess`. Each pipeline executes upon log rotation and receives
+   the Zeek log file as input. The first imports all data via
+   [`import`](/operators/sinks/import) and the second writes the logs as
+   [`parquet`](/formats/parquet) files using [`to`](/operators/sinks/to).
+
+## Fate sharing
```
tobim, line 169:
```suggestion
## Reliability
```

web/blog/native-zeek-log-rotation-and-shipping/index.md
```diff
@@ -0,0 +1,202 @@
...
+This blog post shows how you can use Zeek's native log rotation feature to
+invoke an arbitrary command as soon as a log file gets rotated. This approach
+provides an attractive alternative that turns pull-based file monitoring into
+more flexible push-based delivery.
+
+|              |   Push   |     Pull     |
+| ------------ |:--------:|:------------:|
+| Trigger      | rotation | new file/dir |
+| Complexity   |   low    |    medium    |
+| Fate Sharing | coupled  |  decoupled   |
```
tobim, line 196:
```suggestion
| Reliability  |   low    |    high      |
```

tobim (approved):
I think it's good that we kept it at the same length instead of expanding on various alternative implementations. We can still publish such a post later, but then it would be an article.

I propose to remove the term "Fate Sharing", because imho that concept is only tangential to the reliability.

PR #3376, opened 2023-07-19
Expose pipeline operator metrics in execution node and pipeline executor

Dakostu:
This change implements exposure of pipeline metrics via each running operator in the execuction node.
Right now this will be needed to propagate metrics to the `pipeline-manager` via https://github.com/tenzir/tenzir-plugins/pull/86.


libtenzir/include/tenzir/pipeline.hpp
```diff
@@ -154,6 +154,42 @@ struct inspector
...
+              f.field("time_elapsed", x.time_elapsed),
+              f.field("in_unit", x.in_unit), f.field("out_unit", x.out_unit),
+              f.field("inbound_total", x.inbound_total),
+              f.field("inbound_num_batches", x.inbound_num_batches),
+              f.field("inbound_rate_per_second", x.inbound_rate_per_second),
+              f.field("outbound_total", x.outbound_total),
+              f.field("outbound_num_batches", x.outbound_num_batches),
+              f.field("outbound_rate_per_second", x.outbound_rate_per_second));
+  }
+};
```
dominiklohmann, line 191: I have one small change request: Given that we may want to add a toggle between events and estimated bytes, we may have values for both bytes and events on either side in the future. So it may make sense to model the data more like this:

```c++
struct measurement {
  std::string unit; // "events" or "bytes"
  uint64_t num_total;
  uint64_t num_batches;
};

struct operator_metric {
  std::vector<measurement> inbound_measurements;
  std::vector<measurement> outbound_measurements;
  uint64_t index;
  duration time_starting;
  duration time_running;
  duration time_scheduled;
  duration time_elapsed;
};
```

There's also a small bug fix: we should not use `size_t` here, because it is platform-specific. This can lead to subtle incompatibilities between platforms.

The rate can also be easily calculated at the receiving end, so I don't think we should include it in the data itself.

web/openapi/openapi.yaml
```diff
@@ -270,6 +311,11 @@ paths:
                         - id: 0
                           definition: export asdf
                           instrumented: false
+                      diagnostics: []
```
tobim, line 314: The diagnostics array should contain what is currently emitted for this pipeline definition.

web/openapi/openapi.yaml
```diff
@@ -260,6 +296,11 @@ paths:
                         - id: 2
                           definition: publish /bar
                           instrumented: true
+                      diagnostics: []
+                      metrics:
+                        total:
+                          in: {}
+                          out: {}
```
tobim, line 303: A real example would be very useful here.

libtenzir/test/pipeline.cpp
```diff
@@ -192,6 +194,10 @@ struct fixture : fixtures::deterministic_actor_system_and_events {
         if (not diag_error and d.severity == severity::error) {
           diag_error = caf::make_error(ec::unspecified, fmt::to_string(d));
         }
+      },
+      [&](pipeline_op_metrics&) {
+        MESSAGE("received metrics");
+        // Do nothing with metrics locally for now.
```
tobim, line 200: At least some basic assertions should be made here. Can we expect that we get data from all operators in the pipeline?

libtenzir/include/tenzir/pipeline.hpp
tobim: Instead of a `unit` filed I'd rather have
```c++
struct measurement {
  uint64_t num_events;
  uint64_t num_bytes;
  uint64_t num_batches;
};
```
I think an approximate value for `num_bytes` is fine and shouldn't be hard to calculate. For operators that deal in chunks we set `num_events` to 0.

dominiklohmann: Looking forward we also want to have series as an element type, so then you'd have every single type we have as a possible unit. That's why I think we want to have the more generic solution.

Dakostu: I agree with all points, but:

>The rate can also be easily calculated at the receiving end, so I don't think we should include it in the data itself.

I experimented with that and realized that the `rate` values can slightly differ from where the time endpoint is generated. The time endpoint when an execution node is finished/hits the metrics emission loop would be slightly different from the time endpoint when a pipeline has stopped in the pipeline manager.

When calculating metrics in the execution engine, I would prefer to calculate metrics as closest inside the execution engine as possible, then pass it to other recipients, hence the `rate` fields. WDYT?

web/openapi/openapi.yaml
jachris: I agree, but can you add the example to where you define `#/components/schemas/Metrics`? You could then consider removing the whole `example:` section of `/pipeline/list`, as this should automatically use the examples provided by the schema.

web/openapi/openapi.yaml
jachris: Should probably just be deleted (see my other comment).

libtenzir/include/tenzir/pipeline.hpp
dominiklohmann: Tobias and I just agreed on the following in a quick call:

```cpp
struct measurement {
  std::string unit;
  uint64_t num_elements;
  uint64_t num_batches;

  // Always 0 for now for events until we find a way
  // to reliably measure this and always the same as
  // num_elements for bytes.
  uint64_t num_approx_bytes;
};

struct metric {
  uint64_t operator_index;

  measurement inbound_measurement;
  measurement outbound_measurement;

  duration time_starting;
  duration time_running;
  duration time_scheduled;
  duration time_elapsed; 
};
```

dominiklohmann: > I experimented with that and realized that the rate values can slightly differ depending on where the time endpoint is generated. The time endpoint when an execution node is finished/hits the metrics emission loop would be slightly different from the time endpoint when a pipeline has stopped in the pipeline manager.

I think the rate should be the same as `{in,out}bound_measurement.num_elements / std::chrono::duration_cast<std::chrono::duration<double, std::chrono::seconds::period>>(time_elapsed)` given the data from the struct I posted. That's exactly what the execution node uses to calculate it.

Dakostu: OK, then the `rate` calculation would be easier than I suspected after all. I can remove the `rate`s then.

Dakostu: I split up `pipeline_op_metrics` in https://github.com/tenzir/tenzir/pull/3376/commits/4d9a7e8c27e1a57b626ff75ff915fc38c7cb9890

libtenzir/test/pipeline.cpp
Dakostu: >Can we expect that we get data from all operators in the pipeline?

Yes, up until the executor has finished.
I added a check in https://github.com/tenzir/tenzir/pull/3376/commits/26b4bed73aac34962561ba60dd034dad12e10be3

web/openapi/openapi.yaml
Dakostu: I added examples to the Metrics schema and removed the `/pipeline/list` examples in https://github.com/tenzir/tenzir/pull/3376/commits/324fa9d2399798c712324aa053732e7ea3a03f3e

web/openapi/openapi.yaml
Dakostu: deleted in https://github.com/tenzir/tenzir/pull/3376/commits/324fa9d2399798c712324aa053732e7ea3a03f3e

libtenzir/src/execution_node.cpp
```diff
@@ -337,8 +338,12 @@ struct exec_node_state : inbound_state_mixin<Input>,
   caf::error abort;
 
   auto start(std::vector<caf::actor> previous) -> caf::result<void> {
-    auto time_starting_guard = make_timer_guard(time_scheduled, time_starting);
+    auto time_starting_guard = make_timer_guard(current_metrics.time_scheduled,
+                                                current_metrics.time_starting);
     TENZIR_DEBUG("{} received start request for `{}`", *self, op->to_string());
+    detail::weak_run_delayed_loop(self, defaults<>::metrics_interval, [this] {
+      emit_metrics();
```
dominiklohmann, line 345:
```suggestion
      auto time_scheduled_guard = make_timer_guard(current_metrics.time_scheduled);
      emit_metrics();
```

libtenzir/src/execution_node.cpp
```diff
@@ -645,9 +662,17 @@ struct exec_node_state : inbound_state_mixin<Input>,
     }
   };
 
-  auto print_metrics() -> void {
-    const auto elapsed = std::chrono::duration_cast<duration>(
+  auto emit_metrics() -> void {
```
dominiklohmann, line 665: We need to additionally call this function in the exit handler.

changelog/next/features/3376--expose-pipeline-metrics.md
```diff
@@ -0,0 +1,3 @@
+Pipeline metrics (total ingress/egress amount and average rate per second) are
+now visible in the `pipeline-manager`, via the `metrics` field in the
+`pipeline/list` endpoint result.
```
dominiklohmann, line 3:
```suggestion
`/pipeline/list` endpoint result.
```

dominiklohmann (commented):
I only looked at the exec node changes, which look great to me.

libtenzir/src/execution_node.cpp
Dakostu: Added an exit handler with last metrics emission in https://github.com/tenzir/tenzir/pull/3376/commits/aace5e16bbc0095d05eb231ad0716f1d7a1076cc

dominiklohmann: That looks correct to me. Thanks!

jachris (commented):
I tried this out by printing the metrics locally. Seems to work just fine! :+1:

dominiklohmann (approved):
I think it makes sense to merge this now and to have any follow-up implemented in #3390. For example, that already contains an improvement upon the exit handler usage from this PR.

Let's ship it.

PR #3377, opened 2023-07-19
Improve low-load memory consumption

tobim:
This reduces the default frequency in which automatic rebuilds are triggered from 5 Minutes to 2 Hours, and decouples the option from the active partition timeout.

We also demote the number of cached query partitions to 1, to avoid holding multiple copies of the same data in memory.


dominiklohmann (approved):
Thanks for forward-porting this.

PR #3384, opened 2023-07-20
Use a distinctive marker for the final continuation token

lava:
When a pipeline is at its end, the `continuation_token` is left as an empty string.

However, that is also the expected value of the continuation token for the initial request to the serve endpoint.

So if a client repeats the initial request (i.e. because the response was dropped by the network) after a pipeline is done but before its state is ultimately removed, the serve manager will assume this is a valid initial request and assign a new continuation token, getting stuck in an endless loop until the state is finally cleaned up.

### DIFF
--- a/libtenzir/builtins/operators/serve.cpp
+++ b/libtenzir/builtins/operators/serve.cpp
@@ -262,7 +262,7 @@ struct managed_serve_operator {
     delayed_attempt.dispose();
     requested = 0;
     TENZIR_DEBUG("clearing continuation token");
-    last_continuation_token = std::exchange(continuation_token, {});
+    last_continuation_token = std::exchange(continuation_token, "__DONE__");
     last_results = results;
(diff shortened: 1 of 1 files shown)


tobim (approved):

libtenzir/builtins/operators/serve.cpp
```diff
@@ -262,7 +262,7 @@ struct managed_serve_operator {
     delayed_attempt.dispose();
     requested = 0;
     TENZIR_DEBUG("clearing continuation token");
-    last_continuation_token = std::exchange(continuation_token, {});
+    last_continuation_token = std::exchange(continuation_token, "__DONE__");
```
dominiklohmann, line 265: The OpenAPI spec needs to be adapted accordingly.

dominiklohmann: Ah, that makes a lot more sense now.

PR #3390, opened 2023-07-23
Improve metrics (and some other things)

dominiklohmann:
This PR is based on @Dakostu's excellent #3376 that laid the groundwork for metrics.

It comes with the following metrics-related changes:

- We now calculate the approximate number of bytes in execution nodes when they handle events. The calculated number is an underestimate and (1) never includes the byte size of the schema, (2) excludes some array data buffers depending on the Arrow version used. 
- `tenzir --dump-metrics '<pipeline>'` prints metrics on stderr after the pipeline finished. 
- A set of bug fixes regarding metrics collection that were still in #3376.

I also snuck in the following change that is unrelated to metrics, but that I stumbled over while developing this PR:
- Sorting now works for `ip` and `enum` fields.

This is what `--dump-metrics` looks like:

```
❯ # Using the M57 trace from the Get Started guide:
❯ cat Zeek/*.log | tenzir --dump-metrics 'read zeek-tsv | to file /dev/null'
operator #1 (load)
  elapsed: 4.02s
  scheduled: 553.13ms (13.75%)
  running: 484.85ms (12.05%)
  outbound:
    bytes: 199,898,907 at a rate of 49677650.77/s
    batches: 12,201 (16383.81 bytes/batch)
operator #2 (read)
  elapsed: 6.06s
  scheduled: 1.84s (30.34%)
  running: 1.78s (29.39%)
  inbound:
    bytes: 199,898,907 at a rate of 32995890.01/s
    batches: 12,201 (16383.81 bytes/batch)
  outbound:
    events: 953,677 at a rate of 157416.68/s
    bytes: 1,997,471 at a rate of 329708.32/s (estimate)
    batches: 54 (17660.69 events/batch)
operator #3 (write)
  elapsed: 7.96s
  scheduled: 7.67s (96.45%)
  running: 7.66s (96.24%)
  inbound:
    events: 953,677 at a rate of 119873.34/s
    bytes: 1,997,471 at a rate of 251074.02/s (estimate)
    batches: 54 (17660.69 events/batch)
  outbound:
    bytes: 541,837,222 at a rate of 68106745.97/s
    batches: 570 (950591.62 bytes/batch)
operator #4 (save)
  elapsed: 7.96s
  scheduled: 6.31ms (0.08%)
  running: 3.94ms (0.05%)
  inbound:
    bytes: 541,837,222 at a rate of 68102061.06/s
    batches: 570 (950591.62 bytes/batch)
```


libtenzir/src/execution_node.cpp
```diff
@@ -93,6 +94,32 @@ auto make_timer_guard(Duration&... elapsed) {
     });
 }
 
+// Return an underestimate for the total number of referenced bytes for a vector
+// of table slices, excluding the schema and disregarding any overlap or custom
+// information from extension types.
+auto num_approx_bytes(const std::vector<table_slice>& events) {
```
dominiklohmann, line 100: Note that the result of this function can be off by a large margin. It never overestimates, but sometimes the result is a lot smaller than you would think and also a lot smaller than it should be. 

That happens because it is not trivial for Arrow to get the referenced byte ranges from all underlying array data buffers. Sometimes, getting that information accurately is impossible.

However, I opted to use the built-in Arrow solution here hoping that it will be improved upon in the future upsrream, rather than us having to roll our own.

libtenzir/src/execution_node.cpp
```diff
@@ -93,6 +94,32 @@ auto make_timer_guard(Duration&... elapsed) {
...
+#else
+  auto result = uint64_t{};
+  for (const auto& batch : events) {
+    if (batch.rows() == 0)
+      continue;
+    auto record_batch = to_record_batch(batch);
+    TENZIR_ASSERT(record_batch);
+    // Note that this function can sometimes fail. Because we ultimately want to
+    // return an underestimate for the value of bytes, we silently fall back to
+    // a value of zero is the referenced buffer size cannot be measured.
```
Dakostu, line 115: Warning/debug message for this?

libtenzir/src/execution_node.cpp
```diff
@@ -904,9 +887,11 @@ auto exec_node(
     return exec_node_actor::behavior_type::make_empty_behavior();
   }
   self->state.weak_node = node;
-  self->attach_functor([name = self->state.op->name()] {
-    TENZIR_DEBUG("exec-node for {} shut down", name);
-  });
+  self->attach_functor(
+    [name = self->state.op->name(), metrics = self->state.metrics] {
+      TENZIR_DEBUG("exec-node for {} shut down", name);
+      metrics->emit();
```
Dakostu, line 893: This makes the functor emit metrics for pipelines that may have been deleted already, e.g. creating an immediately starting pipeline with wrong syntax in the `pipeline-manager`.

To be fair, this also already happens for diagnostics sometimes, so this is more an FYI/TODO than an immediate FIXIT.

libtenzir/builtins/commands/exec.cpp
```diff
@@ -91,6 +92,80 @@ auto add_implicit_source_and_sink(pipeline pipe) -> caf::expected<pipeline> {
...
+                      100.0 * static_cast<double>(metric.time_scheduled.count())
+                        / static_cast<double>(metric.time_elapsed.count()));
+  it = fmt::format_to(it, "{}running: {} ({:.2f}%)\n", indent,
+                      data{metric.time_running},
+                      100.0 * static_cast<double>(metric.time_running.count())
+                        / static_cast<double>(metric.time_elapsed.count()));
+  if (metric.inbound_measurement.unit != "void") {
+    it = fmt::format_to(it, "{}inbound:\n", indent);
+    it = fmt::format_to(
+      it, locale, "{}{}{}: {:L} at a rate of {:.2f}/s\n", indent, indent,
```
Dakostu, line 114:
```suggestion
      it, locale, "{0}{0}{1}: {2:L} at a rate of {3:.2f}/s\n", indent,
```

Would using indexes in this function result in more readable format strings? Up to you.

libtenzir/builtins/commands/exec.cpp
```diff
@@ -91,6 +92,80 @@ auto add_implicit_source_and_sink(pipeline pipe) -> caf::expected<pipeline> {
...
+      metric.inbound_measurement.unit, metric.inbound_measurement.num_elements,
+      static_cast<double>(metric.inbound_measurement.num_elements)
+        / std::chrono::duration_cast<
+            std::chrono::duration<double, std::chrono::seconds::period>>(
+            metric.time_elapsed)
+            .count());
+    if (metric.inbound_measurement.unit != "bytes") {
```
Dakostu, line 121: Why not `operator_type_name<chunk_ptr>()`?  I don't know if that would make things less readable, but OTOH I'm not a fan of magic strings.

Dakostu (approved):
Thank you for the addition, and the `sort` fix seems to come in just at the right time :muscle: 

I left some minor comments.

libtenzir/src/execution_node.cpp
Dakostu:
```suggestion
    // Note that this function can sometimes fail. Because we ultimately want to
    // return an underestimate for the value of bytes, we silently fall back to
    // a value of zero if the referenced buffer size cannot be measured.
```

Also this :point_up:

libtenzir/builtins/operators/sort.cpp
```diff
@@ -48,7 +39,17 @@ class sort_state {
     }
     auto batch = to_record_batch(slice);
     TENZIR_ASSERT(batch);
-    sort_keys_.push_back(path->get(*batch));
+    auto array = path->get(*batch);
+    // TODO: Sorting in Arrow using arrow::compute::SortIndices is not
+    // supported for extension types, so eventually we'll have to roll our
+    // own imlementation. In the meantime, we sort the underlying storage array,
```
Dakostu, line 45:
```suggestion
    // own implementation. In the meantime, we sort the underlying storage array,
```

libtenzir/src/execution_node.cpp
dominiklohmann: I think that's okay. It's better than not having metrics in the case of an ungraceful shutdown.

libtenzir/builtins/commands/exec.cpp
dominiklohmann: While I agree it's more readable I don't think it's worth changing right now. If we touch it again then maybe we can rewrite it.

libtenzir/src/execution_node.cpp
```diff
@@ -93,6 +94,32 @@ auto make_timer_guard(Duration&... elapsed) {
...
 }
 
+// Return an underestimate for the total number of referenced bytes for a vector
+// of table slices, excluding the schema and disregarding any overlap or custom
+// information from extension types.
+auto num_approx_bytes(const std::vector<table_slice>& events) {
+#if ARROW_VERSION_MAJOR < 12
+  // arrow::util::ReferenceBufferSize is broken for Arrow before 12.0.0 and
+  // crashes for extension types.
+  (void)events;
```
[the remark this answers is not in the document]
dominiklohmann: I wrote a follow-up issue to bump our minimum supported Arrow version to 12 instead. Let's do that instead of supporting crutches for older versions.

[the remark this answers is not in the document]
dominiklohmann: See https://github.com/tenzir/issues/issues/721. Resolving this for now.

PR #3392, opened 2023-07-23
Improve error message for operator type clashes

dominiklohmann:
Instead of showing the full internals of operators for type clashes, it makes a lot more sense to just show the name to keep the error message legible.

### DIFF
--- a/libtenzir/builtins/operators/from_load_read.cpp
+++ b/libtenzir/builtins/operators/from_load_read.cpp
@@ -57,7 +57,7 @@ protected:
     // TODO: Fuse this check with crtp_operator::instantiate()
     return caf::make_error(ec::type_clash,
                            fmt::format("'{}' does not accept {} as input",
-                                       to_string(), operator_type_name(input)));
+                                       name(), operator_type_name(input)));
   }
 
 private:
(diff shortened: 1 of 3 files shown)


jachris (approved):
If you search for `'{}' does not accept {} as input`, you'll find a few more cases of this. I want to change this mechanism soon (also for this reason), but in the meantime we should just apply your diff to every occurrence.

libtenzir/builtins/operators/from_load_read.cpp
```diff
@@ -57,7 +57,7 @@ class load_operator final : public crtp_operator<load_operator> {
     // TODO: Fuse this check with crtp_operator::instantiate()
     return caf::make_error(ec::type_clash,
                            fmt::format("'{}' does not accept {} as input",
-                                       to_string(), operator_type_name(input)));
+                                       name(), operator_type_name(input)));
```
jachris, line 60: See above.

dominiklohmann: Nice catch. I changed the remaining instances as well and double-checked via `git-grep` that I caught 'em all now.

PR #3393, opened 2023-07-24
Create a glossary that defines key terms

mavam:
This PR adds a page defining key terms in the Tenzir ecosystem.

### DIFF
--- a/web/docs/glossary.md
+++ b/web/docs/glossary.md
@@ -1,0 +1,105 @@
+# Glossary
+
+<!--
+This glossary adheres to the following template for defining terms:
+
+    ## TERM
+
+    Brief definition without using TERM.
+
+    One additional paragraphs that provide additional information to
+    understand TERM to its full extent. High-level only to understand the
+    concept, without going into details, which should be links below this
+    paragraph.
+
+    - Link to relevant material
+    - Other link to more information
+    - ...
+
+This convention is not enforced technically.
+-->
+
+This page defines central terms in the Tenzir ecosystem.
+
+:::note missing term?
+If you are missing a term, please open a [GitHub Discussion][new-discussion] or
+ping us in our [Discord chat](/discord).
+:::
+
+[new-discussion]: https://github.com/orgs/tenzir/discussions/new?category=questions-answers
+
+## Connector
+
+Manages chunks of raw bytes by interacting with a resource.
+
+A connector is either a *loader* that acquires bytes or a *saver*
+that sends bytes to an external resource. Examples include
+[`file`](./connectors/file.md), [`kafka`](./connectors/kafka.md), and
+[`nic`](./connectors/nic.md).
+
+- See all [connectors](./connectors.md)
+
+## Format
+
+Translates between bytes and events.
+
+A format is either a *parser* that converts bytes to events, or a *printer*
+that converts events to bytes.
+
+- See all [formats](./formats.md)
+
+## Node
+
+A host for [pipelines](#pipeline) and storage reachable over the network.
+
+- [Deploy a node](./setup-guides/deploy-a-node/README.md)
+
+## Operator
+
+The atomic building block of a [pipeline](#pipeline).
+
+An operator is a [source](#source), [transformation](#transformation), or
+[sink](#sink)
+
+## Partition
+
+The horizontal scaling unit of the storage attached to a [node](#node).
+
+A partition contaisn a [store](#store) and optionally a set of indexes.
+
+## Pipeline
+
+Combines a set of [operators](#operator) into a dataflow graph.
+
+- Understand the [pipeline language](./language/pipelines.md)
+- [Run a pipeline](./user-guides/run-a-pipeline/README.md)
+
+## Sink
+
+An [operator](#operator) consuming input, without producing any output.
+
+- See all [sinks](./operators/sinks/README.md)
+
+## Source
+
+An [operator](#operator) producing output, without consuming any input.
+
+- See all [sources](./operators/sources/README.md)
+
+## Store
+
+The raw data in a [partition](#partition) in Parquet or Feather.
+
+## TQL
+
+An acronym for *Tenzir Query Language*.
+
+TQL is the language in which users write [pipelines](#pipeline).
+
+- Learn more about the [language](./language/pipelines.md)
+
+## Transformation
+
+An [operator](#operator) consuming both input and producing output.
+
+- See all [transformations](./operators/transformations/README.md)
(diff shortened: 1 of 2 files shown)


web/docs/glossary.md
```diff
@@ -0,0 +1,170 @@
...
+## Node
+
+A host for [pipelines](#pipeline) and storage reachable over the network.
+
+The `tenzir-node` binary starts a node in a dedicated server process that
+listens on TCP port 5158.
+
+- [Deploy a node](./setup-guides/deploy-a-node/README.md)
+- [Use the app](./setup-guides/use-the-app/README.md) to manage a node
+- Use the [REST API](./rest-api.md) to manage a node.
```
dominiklohmann, line 94: Throughout the document you sometimes use dots at the end of bullets and sometimes not. Please choose one.

web/docs/glossary.md
```diff
@@ -0,0 +1,170 @@
...
+## Metrics
+
+Runtime statistics about pipeline and node execution.
+
+- [Collect metrics](./setup-guides/collect-metrics.md)
+- See all available [metrics](./metrics.md) details
+
+## Operator
+
+The atomic building block of a [pipeline](#pipeline).
```
dominiklohmann, line 107: The atomic is not necessarily true here, as every pipeline itself is also a valid operator again. I would just remove the word.

web/docs/glossary.md
```diff
@@ -0,0 +1,170 @@
...
+## Platform
+
+Control plane for nodes and pipelines, accessible through [app](#app) at
+[app.tenzir.com](https://app.tenzir.com).
+
+- [Use the app](./setup-guides/use-the-app/README.md)
+
+## Schema
+
+A named record type describing the top-level structure of a data frame.
```
dominiklohmann, line 134: We do not use the word data frame in our code base or in our documentation. It's a bit of an odd encounter here. Can we describe this differently?

web/docs/glossary.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Source
+
+An [operator](#operator) producing output, without consuming any input.
+
+- See all available [sources](./operators/sources/README.md)
+
+## Store
+
+The raw data in a [partition](#partition) in Parquet or Feather.
```
dominiklohmann, line 152: The store is more than just the raw data itself. It is an abstraction that provides access to the data underlying a partition.

Mentioning Parquet and Feather here is also a bit of an implementation detail that is already covered by the first bullet.

web/docs/glossary.md
```diff
@@ -0,0 +1,170 @@
...
+
+- [Deploy a node](./setup-guides/deploy-a-node/README.md)
+- [Use the app](./setup-guides/use-the-app/README.md) to manage a node
+- Use the [REST API](./rest-api.md) to manage a node.
+- [Import into a node](./user-guides/import-into-a-node.md)
+- [Export from a node](./user-guides/export-from-a-node.md)
+
+## Metrics
+
+Runtime statistics about pipeline and node execution.
```
dominiklohmann, line 100:
```suggestion
Runtime statistics about pipeline execution.
```

In the new model, metrics will only be emitted by pipelines. The old metrics page that you linked below will likely need to be rewritten or removed.

dominiklohmann (approved):
Just some minor nitpicks. Reads great otherwise. This is a nice effort, I like it! 🚀

web/docs/glossary.md
mavam: I thought I removed all dots. Glad I asked you as reviewer. 😉

web/docs/glossary.md
mavam: We actually use it in the first sentence here: https://docs.tenzir.com/next/data-model/schemas. I kept the definition synchronous. Users probably understand it, although "data frame" is not (yet) a first-class abstraction.

web/docs/glossary.md
mavam: I'm going off this diagram: https://docs.tenzir.com/next/setup-guides/tune-performance#storage-engine. There, the store is _inside_ the partition and abstracts away the raw storage. Please make a concrete suggestion if you like it phrased differently.

> Mentioning Parquet and Feather here is also a bit of an implementation detail that is already covered by the first bullet.

I don't follow. Where is it covered? It's not a detail, actually, it's critical for gaining trust in the system by understanding what our abstractions map to. This is precisely the missing link that allows people to map our concepts to existing things out there.

web/docs/glossary.md
mavam: Okay, I removed the link.

web/docs/glossary.md
dominiklohmann: The store is not just the raw data, but also the mechanism to access it, and to scan it with an expression as a filter. It also contains metadata associated with the data, e.g., type information or the import timestamp.

I propose this:

```
A mechanism to read and write data belonging to a [partition](#partition). Every partition contains exactly one store.

Tenzir ships with [Apache Feather V2](link) and [Apache Parquet](link) stores.
```

mavam: > The store is not just the raw data, but also the mechanism to access it, and to scan it with an expression as a filter.

I think it should be clear that Tenzir can access its stores efficiently. That's an implementation detail and doesn't help understanding the term.

> It also contains metadata associated with the data, e.g., type information or the import timestamp.

Isn't this part of the partition? Otherwise it wouldn't be possible to read the store with a third-party tool, unless we rely on the fact that a store has a dedicated place for metadata, which is a strong assumption.

mavam: I've slightly tweaked the description.

dominiklohmann: > > The store is not just the raw data, but also the mechanism to access it, and to scan it with an expression as a filter.
> I think it should be clear that Tenzir can access its stores efficiently. That's an implementation detail and doesn't help understanding the term.

I'm not arguing about this from a point of efficiency. It's about correctness. A store to us is not the data itself, but rather the mechanism to read and write the data. Technically, the store doesn't even know where the actual data resides, as the partition is responsible for that.

> > It also contains metadata associated with the data, e.g., type information or the import timestamp.
> 
> Isn't this part of the partition? Otherwise it wouldn't be possible to read the store with a third-party tool, unless we rely on the fact that a store has a dedicated place for metadata, which is a strong assumption.

The partition only contains aggregate information about the metadata that is required to read data, but the actual metadata is kept directly in the stores. This is by design—reading Arrow metadata is optional, and for both Feather and Parquet not reading it is as simple as opting out when creating the respective reader. We do want the stores to be self-contained, though.

mavam: We agreed on this in a call.

PR #3398, opened 2023-07-24
Tune the file loader's throughput and memory usage

dominiklohmann:
In non-scientific measurements, these two small changes have a big practical impact:
1. Increasing the max chunk size from 2^14 to 2^20 aligns it with the upper bound for a batch of bytes transferred between execution nodes for a single pull. This minimizes scheduling overhead for high-throughput pipelines. For slow-to-read files we already have a read timeout of 20ms in place that causes the file loader to yield frequently enough to support real-time use cases.
2. Measurements have shown a significant memory usage in slow readers that I was able to track down to poorly sized chunks with spare allocated capacity. Especially with the larger chunk sizes we must not keep spare capacity around for longer than necessary. Instead, we now effectively shrink chunks to fit.


libtenzir/builtins/connectors/file.cpp
```diff
@@ -234,7 +234,10 @@ class fd_wrapper {
 
 class file_loader final : public plugin_loader {
 public:
-  static constexpr auto max_chunk_size = size_t{16384};
+  // We use 2^20 for the upper bound of a chunk size, which exactly matches the
+  // upper limit defined by execution nodes for transporting events.
+  // TODO: Get the backpressure-adjusted value at runtime from the execution node.
+  static constexpr size_t max_chunk_size = 1 << 20;
```
mavam, line 240:
```suggestion
  static constexpr size_t max_chunk_size = 1_MiB;
```

mavam (approved):

PR #3399, opened 2023-07-25
Fix a disk IO race in the demo Zeek import

tobim:
Immediately reading files after decompressing is dangerous, as they may still be partially buffered up to some time after the write "completed" in user space. This frequently triggered bad reads in the demo node Zeek pipeline, causing it to abort.
Calling sync after extracting the archive to disk effectively prevents this issue.

### DIFF
--- a/demo-node/entrypoint.bash
+++ b/demo-node/entrypoint.bash
@@ -14,16 +14,20 @@ read -r -u "${NODE[0]}" DUMMY
 # of Tenzir processes.
 export TENZIR_CONSOLE_VERBOSITY="${TENZIR_CLIENT_CONSOLE_VERBOSITY:-"info"}"
 
-tenzir 'remote version | put version | write json'
-
 # Wait until the HTTP endpoint is listening.
 while ! timeout 100 bash -c "echo > /dev/tcp/127.0.0.1/5160"; do
   echo "Waiting for the HTTP service..."
   sleep 1
 done
 
+# The shell operator decompresses the data and writes it to `read suricata` on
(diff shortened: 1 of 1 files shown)


demo-node/entrypoint.bash
```diff
@@ -14,16 +14,14 @@ read -r -u "${NODE[0]}" DUMMY
...
-
 # Wait until the HTTP endpoint is listening.
 while ! timeout 100 bash -c "echo > /dev/tcp/127.0.0.1/5160"; do
   echo "Waiting for the HTTP service..."
   sleep 1
 done
 
 suricata_pipe="shell 'bash -c \\\"curl -s -L https://storage.googleapis.com/tenzir-datasets/M57/suricata.tar.zst | tar -x --zstd --to-stdout\\\"' | read suricata | where #schema != \\\"suricata.stats\\\" | import"
-zeek_pipe="shell 'bash -c \\\"curl -s -L https://storage.googleapis.com/tenzir-datasets/M57/zeek.tar.zst | tar -x --zstd; cat Zeek/*.log\\\"' | read zeek-tsv | import"
+zeek_pipe="shell 'bash -c \\\"curl -s -L https://storage.googleapis.com/tenzir-datasets/M57/zeek.tar.zst | tar -x --zstd; sync; cat Zeek/*.log\\\"' | read zeek-tsv | import"
```
mavam, line 24: Can you add a comment about `sync`, as it transcends the average person's knowledge?

mavam (approved):

PR #3404, opened 2023-07-26
Write server.log log files when running 'tenzir-node'

lava:


tenzir/tenzir.cpp
```diff
@@ -84,18 +84,22 @@ int main(int argc, char** argv) {
...
-  }
+  bool is_server = false;
+  if (app_name == "tenzir-node")
+    is_server = true;
+  if (app_name == "vast" || app_name == "tenzir-ctl")
+    is_server = invocation->full_name == "start"
+                || caf::get_or(cfg.content, "tenzir.node", false);
+  else if (app_name == "tenzir")
+    is_server = invocation->full_name == "exec"
+                || caf::get_or(cfg.content, "tenzir.node", false);
```
tobim, line 95: This can be shortened, and I believe `tenzir exec` does not create a node, so that condition was wrong.
```suggestion
    const bool is_server = app_name == "tenzir-node"
                         || caf::get_or(cfg.content, "tenzir.node", false)
                         || ((app_name == "vast" || app_name == "tenzir-ctl")
                             && invocation->full_name == "start");
```

tobim (approved):
I'm ok with the boolean parameter for this. Great to have this fixed!

tenzir/tenzir.cpp
lava: Makes sense, I slightly rearranged it to avoid the nested boolean expressions.

PR #3414, opened 2023-07-28
Add `show` operator

mavam:
This PR adds a `show` operator to introspect a node.

### DIFF
--- a/libtenzir/builtins/operators/show.cpp
+++ b/libtenzir/builtins/operators/show.cpp
@@ -1,0 +1,159 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The Tenzir Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <tenzir/actors.hpp>
+#include <tenzir/argument_parser.hpp>
+#include <tenzir/atoms.hpp>
+#include <tenzir/catalog.hpp>
+#include <tenzir/error.hpp>
+#include <tenzir/logger.hpp>
+#include <tenzir/node_control.hpp>
+#include <tenzir/partition_synopsis.hpp>
+#include <tenzir/plugin.hpp>
+#include <tenzir/table_slice_builder.hpp>
+#include <tenzir/uuid.hpp>
+
+#include <caf/scoped_actor.hpp>
+#include <caf/timespan.hpp>
+
+#include <string>
+#include <vector>
+
+namespace tenzir::plugins::show {
+
+auto table_type() -> type {
+  return type{
+    "tenzir.table",
+    record_type{
+      {"uuid", string_type{}},
+      {"memory_usage", uint64_type{}},
+      {"min_import_time", time_type{}},
+      {"max_import_time", time_type{}},
+      {"version", uint64_type{}},
+    },
+  };
+}
+
+struct operator_args {
+  located<std::string> aspect;
+
+  friend auto inspect(auto& f, operator_args& x) -> bool {
+    return f.object(x)
+      .pretty_name("operator_args")
+      .fields(f.field("aspect", x.aspect));
+  }
+};
+
+class show_operator final : public crtp_operator<show_operator> {
+public:
+  show_operator() = default;
+
+  explicit show_operator(operator_args args) : args_{std::move(args)} {
+  }
+
+  auto operator()(operator_control_plane& ctrl) const
+    -> generator<table_slice> {
+    if (args_.aspect.inner == "tables") {
+      // TODO: Some of the the requests this operator makes are blocking, so we
+      // have to create a scoped actor here; once the operator API uses async we
+      // can offer a better mechanism here.
+      auto blocking_self = caf::scoped_actor(ctrl.self().system());
+      auto components
+        = get_node_components<catalog_actor>(blocking_self, ctrl.node());
+      if (!components) {
+        ctrl.abort(std::move(components.error()));
+        co_return;
+      }
+      co_yield {};
+      auto [catalog] = std::move(*components);
+
+      auto synopses = std::vector<partition_synopsis_pair>{};
+      auto error = caf::error{};
+      ctrl.self()
+        .request(catalog, caf::infinite, atom::get_v)
+        .await(
+          [&synopses](std::vector<partition_synopsis_pair> result) {
+            synopses = std::move(result);
+          },
+          [&error](caf::error err) {
+            error = std::move(err);
+          });
+      co_yield {};
+      if (error) {
+        ctrl.abort(std::move(error));
+        co_return;
+      }
+      auto builder = table_slice_builder{table_type()};
+      // FIXME: ensure that we do no use more most 2^15 rows.
+      for (const auto& synopsis : synopses) {
+        // TODO: add schema
+        if (not(builder.add(fmt::to_string(synopsis.uuid))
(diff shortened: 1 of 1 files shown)


libtenzir/builtins/operators/show.cpp
```diff
@@ -0,0 +1,264 @@
...
+        }
+      }
+      co_yield builder.finish();
+    } else if (args_.aspect.inner == "operators") {
+      auto builder = table_slice_builder{operator_type()};
+      for (const auto* plugin : plugins::get<operator_parser_plugin>()) {
+        // TODO: figure out how we can get the operator type. Ideally also it's
+        // arguments.
+        if (not(builder.add(plugin->name()) && builder.add(false)
+                && builder.add(false) && builder.add(false))) {
```
mavam, line 149: @jachris is this information easily available?

jachris: Not really. The possible input types can depend on the arguments (e.g., `remote ...`), so in general, you cannot decide this purely by using the operator plugin. For the arguments, we wanted to have them as data anyway, but this is currently not a priority and requires quite a bit of effort to get right.

mavam: Just to be clear: with "operator type" I meant whether it's a source, sink, or transform. Not the input/output types.

jachris: Ah, I see, but the same applies here (`remote` can be a source, sink, or transform, but we don't know in advance).

mavam: So then all apply, no?

The point is to give the user more information on how to use operators. For many, it's only 1 or 2, so this is very helpful information to have.

jachris: > So then all apply, no?

Depends on your definition of "source", etc. But sure, you could say so.

> What I am asking: can we make plugin API changes so that we can query this?

We can. But then we should make the plugins return the possible type signatures. When using the `crtp_operator`, we can even infer most of this automatically. How important is this to you?

mavam: > How important is this to you?

This is probably a continuous effort. I would like to get something (marked as experimental) through the door quickly to enable our partners so that they can use these introspection capabilities, even if they are incomplete or have warts.

Anything that's scheduling-worth effort should be a strict follow-up. Within the scope of this PR, I'd just like to ship a something-is-better-than-nothing outcome.

mavam:
Note to the reviewer:
- `show types` is rough, but we marked the operator as "experimental" so we can make changes
- I want a `show tables` ultimately, in addition to `show partitions`. In fact, I think `show schemas` is not what we want. Nearly all DBs have an equivalent of `show tables`.

libtenzir/builtins/operators/from_load_read.cpp
```diff
@@ -248,6 +256,10 @@ class load_plugin final : virtual public operator_plugin<load_operator> {
 
 class read_plugin final : virtual public operator_plugin<read_operator> {
 public:
+  auto signature() const -> operator_signature override {
+    return {.source = true};
```
jachris, line 260:
```suggestion
    return {.transformation = true};
```

libtenzir/builtins/operators/show.cpp
```diff
@@ -0,0 +1,451 @@
...
+  };
+}
+
+/*
+/// Adds data to a field or list guard.
+void add(auto& guard, const data& x) {
+  auto f = detail::overload{
+    [&](const auto&) {
+      auto err = guard.add(make_view(x));
+      (void)err;
```
jachris, line 139: Is it intended that this is left here?

libtenzir/builtins/operators/show.cpp
```diff
@@ -0,0 +1,451 @@
...
+    return "show";
+  }
+
+  auto detached() const -> bool override {
+    return true;
+  }
+
+  auto location() const -> operator_location override {
+    return operator_location::local;
+  }
```
jachris, line 406: You can return `operator_location::remote` here for `show partitions` and `show types`.

libtenzir/builtins/operators/show.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+  auto optimize(expression const& filter, event_order order) const
+    -> optimize_result override {
+    (void)order;
+    (void)filter;
+    return optimize_result{std::nullopt, event_order::unordered, copy()};
```
jachris, line 412:
```suggestion
    return do_not_optimize(*this);
```

Yours might also work, but (strictly speaking) it breaks the contract of `optimize`.

libtenzir/builtins/operators/to_write_save.cpp
```diff
@@ -148,6 +148,10 @@ class write_operator final : public crtp_operator<write_operator> {
 
 class write_plugin final : public virtual operator_plugin<write_operator> {
 public:
+  auto signature() const -> operator_signature override {
+    return {.sink = true};
```
jachris, line 152:
```suggestion
    return {.transformation = true};
```

libtenzir/builtins/operators/show.cpp
```diff
@@ -0,0 +1,451 @@
...
+        die("invalid catalog interaction");
+      }
+    } else {
+      die("unchecked aspect");
+    }
+  }
+
+  auto name() const -> std::string override {
+    return "show";
+  }
```
jachris, line 398: I think this clashes with the `show` command plugin, no? This should really cause an error, but for me it silently fails and makes the `show` operator unavailable.

What can we do here to resolve this situation?

web/docs/operators/sources/show.md
```diff
@@ -0,0 +1,52 @@
...
+
+Describes the part of Tenzir to look at.
+
+Available aspects:
+
+- `connectors`: shows all available [connectors](../../connectors.md).
+- `formats`: shows all available [formats](../../formats.md).
+- `operators`: shows all available [operators](../../operators.md).
+- `partitions`: shows all partitions of a (remote) node.
+- `types`: shows all known type definitions.
```
jachris, line 31: `show types` can also only be run remotely.

libtenzir/builtins/operators/show.cpp
```diff
@@ -0,0 +1,451 @@
...
+        auto builder = adaptive_table_slice_builder{type_type()};
+        for (const auto& type : types) {
+          auto row = builder.push_row();
+          auto err = row.push_field("name").add(type.name());
+          auto structure = row.push_field("structure").push_record();
+          auto f = detail::overload{
+            [&](const auto&) {
+              auto err
+                = structure.push_field("basic").add(fmt::to_string(type));
+              (void)err;
```
jachris, line 334: Uh, discarding errors. Perhaps you can assert?

jachris (approved):

libtenzir/builtins/operators/show.cpp
mavam: Can we simply rip out the `show` command? I don't think anyone relies on it, and we're about to port the information over to the corresponding operator anyway.

libtenzir/builtins/operators/show.cpp
mavam: Not really. I thought I'll use it later on, but I now removed it with a dedicated commit so that it's easy to bring back.

libtenzir/builtins/operators/show.cpp
```diff
@@ -0,0 +1,752 @@
...
+  };
+}
+
+/// A type that describes all fields in a schema.
+auto schema_type() -> type {
+  return type{
+    "tenzir.schema",
+    record_type{
+      {"schema", string_type{}},
+      {"fingerprint", string_type{}},
```
dominiklohmann, line 127: This should be called `schema_id`, because that's the name we use everywhere else.

PR #3415, opened 2023-07-30
Add fixed route to SBOM

mavam:
This PR creates a static route to our SBOM so that we can link it reliably.

### DIFF
--- a/web/src/pages/sbom.tsx
+++ b/web/src/pages/sbom.tsx
@@ -1,0 +1,12 @@
+import ExternalLink from "@site/src/components/ExternalLink";
+import React from "react";
+
+const Element = () => {
+  return (
+    <ExternalLink
+      url={"https://github.com/tenzir/tenzir/releases/latest/download/Tenzir.spdx"}
+    />
+  );
+};
+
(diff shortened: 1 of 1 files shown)


web/src/pages/sbom.tsx
```diff
@@ -0,0 +1,12 @@
+import ExternalLink from "@site/src/components/ExternalLink";
+import React from "react";
+
+const Element = () => {
+  return (
+    <ExternalLink
+      url={"https://github.com/tenzir/tenzir/releases/latest/download/Tenzir.spdx"}
```
jachris, line 7: This link is 404 for me.

jachris (approved):

web/src/pages/sbom.tsx
mavam: Right. The reason is that VAST v3.1 was the `latest` release. RCs do not count as `latest`, so this link will only work as of Tenzir v4.0.

mavam: If you take a look at the current RCs, `Tenzir.spdx` already exists there.

PR #3422, opened 2023-08-02
Implement a fallback parser mechanism for extensions that don't have …

Dakostu:
This change adds a fallback `json` parser for files whose extensions do not have a default parser support.

### DIFF
--- a/libtenzir/src/detail/file_path_to_parser.cpp
+++ b/libtenzir/src/detail/file_path_to_parser.cpp
@@ -8,6 +8,8 @@
 
 #include "tenzir/detail/file_path_to_parser.hpp"
 
+#include "tenzir/plugin.hpp"
+
 #include <array>
 
 namespace tenzir::detail {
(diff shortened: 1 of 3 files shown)


libtenzir/src/detail/file_path_to_parser.cpp
```diff
@@ -32,11 +34,19 @@ auto file_path_to_parser(const std::filesystem::path& path) -> std::string {
...
   }
-  return fallback_ext.string().substr(1);
+  fallback_ext = fallback_ext.substr(1);
+  if (plugins::find<parser_parser_plugin>(fallback_ext)) {
+    return fallback_ext;
+  }
+  TENZIR_VERBOSE("Could not find default parser for path {} - falling back "
+                 "to "
+                 "{}",
+                 path, fallback_parser);
```
dominiklohmann, line 48: Style nit for consistency with other log messages:

```suggestion
  TENZIR_VERBOSE("failed to find default parser for '{}'; falling back "
                 "to {}",
                 path, fallback_parser);
```

dominiklohmann (approved):
Thanks for this! 🙏

PR #3423, opened 2023-08-02
Require Arrow 12.0

tobim:
This drops compatibility with older versions of Apache Arrow.


dominiklohmann (approved):
As discussed in-person, I pushed the remaining required changes myself. Thanks for this effort!

PR #3432, opened 2023-08-04
Fixup broken unit tests from layout change

tobim:
Follow up to #3430.


tobim:
Actually not necessary.

PR #3433, opened 2023-08-05
Fix hang in XSV parser for small inputs

dominiklohmann:
This fixes a hang in the XSV parser when the header was in the middle of a chunk boundary. This usually happens for really small inputs.


tobim (approved):
I tested this change with the this input:
```csv
procs_waiting,procs_sleep,memory_virtual
1,0,268
```

PR #3435, opened 2023-08-07
Fix Suricata TLS SNI field name in concepts

satta:
We noticed today that Tenzir ships a `net.domain` concept referencing the EVE JSON field `suricata.tls.sni` which does not exist in the schema. It should be `suricata.tls.tls.sni` according to the schema definition:

```
type suricata.tls = suricata.component.common + record {
  tls: record {
    sni: string,
    session_resumed: bool,
    subject: string,
    issuerdn: string,
    serial: string,
    fingerprint: string,
    ja3: suricata.component.ja3,
    ja3s: suricata.component.ja3,
    notbefore: time,
    notafter: time
  }
}
```
For the sake of completeness we also add it to the `net.hostname` concept, as it serves a similar purpose as `suricata.http.http.hostname`  which is also in that concept.

### DIFF
--- a/schema/concepts/suricata.yaml
+++ b/schema/concepts/suricata.yaml
@@ -216,6 +216,7 @@
       - suricata.smb.host
       - suricata.smb.smb.ntlmssp.host
       - suricata.ssh.host
+      - suricata.tls.tls.sni
 - concept:
     name: net.domain
     fields:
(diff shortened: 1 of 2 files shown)


tobim:
This is a minor fix that I verified manually. I will add a commit to update the target rev-count for the next tag and then merge it directly.

satta:
Thanks!

PR #3436, opened 2023-08-07
Group summarize groups when building the output

jachris:
This change speeds up the `summarize` operator by combining multiple output rows into the same table slice if they have the same output schema.


lava (approved):
While I didn't test in-depth, the high-level idea seems sound and the integration tests seem to cover the whole code path.

PR #3444, opened 2023-08-08
Flush `shell` operator's child process stdout

dominiklohmann:
This fixes a race condition in the shutdown of the `shell` operator. For example, `tenzir 'shell "echo hi"'` when run repeatedly sometimes printed nothing before this change. The fix is simple: always explicitly flush the child process' stdout on shutdown of the operator.

### DIFF
--- a/libtenzir/builtins/operators/shell.cpp
+++ b/libtenzir/builtins/operators/shell.cpp
@@ -108,12 +108,15 @@ public:
     return caf::none;
   }
 
-  void close_stdin() {
+  void close_stdinout() {
     // See https://github.com/boostorg/process/issues/125 for why we
     // seemingly double-close the pipe.
     TENZIR_DEBUG("sending EOF to child's stdin");
     stdin_.close();
     stdin_.pipe().close();
+    TENZIR_DEBUG("sending EOF to child's stdout");
+    stdout_.close();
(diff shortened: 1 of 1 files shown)


libtenzir/builtins/operators/shell.cpp
```diff
@@ -207,7 +210,7 @@ class shell_operator final : public crtp_operator<shell_operator> {
     {
       // Coroutines require RAII-style exit handling.
       auto at_exit = caf::detail::make_scope_guard([&] {
-        child->close_stdin();
+        child->close_stdinout();
         TENZIR_DEBUG("joining thread");
         thread.join();
       });
```
jachris, line 216: Also, what about the behavior when `shell` is the source?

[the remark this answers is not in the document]
dominiklohmann: I'm not 100% sure on the order, maybe this does introduce another race. But we definitely do need to flush the child process' stdout, and closing is one way of achieving that.

dominiklohmann: Should be a no-op if I understand the code correctly.

dominiklohmann:
Closing this as it's not the correct solution.

PR #3449, opened 2023-08-09
Add example pipeline that imports live CVEs

jachris:


mavam:
This is good candidate for contextualization.

dominiklohmann:
I think this would be nice to put into a package full of example snippets. But it's stale, so I'm closing it for now.

PR #3455, opened 2023-08-11
Implement `show config`

dominiklohmann:
The `config` aspect returns the current configuration as an event.

### DIFF
--- a/libtenzir/builtins/aspects/config.cpp
+++ b/libtenzir/builtins/aspects/config.cpp
@@ -1,0 +1,115 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The Tenzir Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <tenzir/adaptive_table_slice_builder.hpp>
+#include <tenzir/argument_parser.hpp>
+#include <tenzir/plugin.hpp>
+
+namespace tenzir::plugins::config {
+
+namespace {
+
+template <class Guard>
+struct recursive_add {
+  explicit recursive_add(Guard& guard) noexcept : guard{guard} {
+  }
+
+  auto operator()(const record& x) {
+    auto r = guard.push_record();
+    for (const auto& [key, value] : x) {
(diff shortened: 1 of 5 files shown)


libtenzir/builtins/aspects/config.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <tenzir/adaptive_table_slice_builder.hpp>
+#include <tenzir/argument_parser.hpp>
+#include <tenzir/plugin.hpp>
+
+namespace tenzir::plugins::config {
+
+namespace {
+
+template <class Guard>
+struct recursive_add {
```
mavam, line 18: That's what I need for `show pipelines` as well.

mavam: I.e., could you move it into some header file?

libtenzir/builtins/aspects/config.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+  auto operator()(const map&) {
+    die("unreachable");
+  }
+
+  template <class T>
+    requires(not std::is_same_v<T, data>)
+  auto operator()(const T& x) {
+    // FIXME
+    (void)guard.add(x);
```
mavam, line 62: Maybe return `caf::error` here? If you know this will never fail, then fall back to assert.

libtenzir/builtins/aspects/config.cpp
```diff
@@ -0,0 +1,115 @@
...
+          // We skip CAF's configuration as it adds a lot of noise for very
+          // little value.
+          continue;
+        }
+        auto field = row.push_field(key);
+        auto add = recursive_add{field};
+        caf::visit(
+          [&](const auto& value) {
+            add(value);
+          },
```
mavam, line 100:
```suggestion
          add,
```

mavam (approved):
Cool!

libtenzir/builtins/aspects/config.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+private:
+  record config_ = {};
+};
+
+} // namespace
+
+} // namespace tenzir::plugins::config
+
+TENZIR_REGISTER_PLUGIN(tenzir::plugins::config::plugin)
```
mavam, line 115: Please also update the `show` docs and mention this aspect.

mavam, line 115: ...and a changelog entry please.

PR #3456, opened 2023-08-11
Add support for reading and writing YAML documents

dominiklohmann:
The `yaml` format is especially nice when combined with the new `show config`:

image


mavam:
This looks like it's basically ready, modulo a few integration tests.

dominiklohmann:
Basically yes. I just didn't get back to this yet, but I'm hoping to find a quiet 30 minutes ahead of the next release to squeeze this in.

libtenzir/src/detail/file_path_to_parser.cpp
```diff
@@ -17,11 +17,15 @@ namespace tenzir::detail {
...
+  = std::array<std::pair<std::filesystem::path, std::string>, 1>{{
+    {"eve.json", "suricata"},
+  }};
 
 const auto extension_to_parser_list
-  = std::array<std::pair<std::string, std::string>, 1>{{{".ndjson", "json"}}};
+  = std::array<std::pair<std::string, std::string>, 2>{{
+    {".ndjson", "json"},
+    {".yml", "yml"},
+  }};
```
jachris, line 28: What about `.yaml`? :)

jachris: Also, the parser name `"yml"` must be changed to `"yaml"`.

dominiklohmann: .yaml works implicitly because we first try to match the file extension. I'll change the second string to yaml, that's a good catch. I had it the other way around first, and then must've adapted this incorrectly.

libtenzir/builtins/formats/yaml.cpp
```diff
@@ -0,0 +1,286 @@
...
+  auto row = builder.push_row();
+  try {
+    auto node = YAML::Load(document);
+    if (not node.IsMap()) {
+      diagnostic::error("document is not a map").emit(ctrl.diagnostics());
+      return;
+    }
+    for (const auto& element : node) {
+      auto field = row.push_field(element.first.as<std::string>());
+      parse_node(field, element.second, ctrl);
```
dominiklohmann, line 94: There's a stack-use-after-scope somewhere in the adaptive builder here that this code reveals; since @jachris is working on fixing that for other reasons anyways, I think it makes sense to wait with this PR until after that has been fixed.

web/docs/formats/yaml.md
```diff
@@ -0,0 +1,48 @@
...
+## Synopsis
+
+Parser:
+
+```
+yaml
+```
+
+Printer:
+
```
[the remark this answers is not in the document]
mavam: Since they are identical, one synopsis suffices.

dominiklohmann:
> We should write the section for the next release blog post ideally as part of this PR.

I won't get that part in before vacation, but other than that this PR is ready, so unfortunately whoever is owning the next release process (read: very likely me anyways) will have to write that part.

PR #3457, opened 2023-08-11
Make `show pipelines` a thing

mavam:
Example output of `show pipelines`:

```json
{
  "id": "0",
  "name": "Example eth0 Import",
  "definition": "from nic eth0 | decapsulate data | import",
  "state": "stopped",
  "error": "",
  "restart_with_node": false
}
```


dominiklohmann (approved):
🚀

PR #3458, opened 2023-08-14
In-sidebar operator classification

mavam:
This PR attempts to classsify operators in the sidebar for better understanding
their input and output types.


mavam:
Example screenshot:

image

@dit7ya do you see an evolutionary path here to create *multiple* tags, i.e., one for input and one for output? This would have the advantage that we can add color to make type transitions clearer.

web/src/css/custom.scss
jachris, line 1: How does these tags interact with polymorphic operators?

jachris, line 1: Given that people are presumably looking for `event -> event` operators most of the time, it could make sense to group these together in the navigation pane (instead of the current `Transformation` page).

mavam: Not sure yet, but I think we would want to make that part of the "signature". For example, "BE → E" could mean that this operator transforms both bytes and events to events.

mavam: Doable, but then we'd have to manually provide a sidebar entry for every operator. Example: https://www.cpp-lang.net/docs/std/containers/strings/string/

mavam:
Closing this as we're not pursuing this idea not further right now.

PR #3461, opened 2023-08-15
Add Fluent Bit source and sink operators

tobim:
This will add fluent-bit loaders and savers.

This PR is in a very early stage right now, the actual functionality is not implemented yet.

We include a custom `fluent-bit-minimal.h` from https://github.com/fluent/fluent-bit/issues/7165#issuecomment-1544711546 because the public API of fluent-bit is currently unusable. This will hopefully be included upstream in the near future.

```[tasklist]
### Business Logic
- [x] Implement first scaffold to test fluent-bit library
- [x] Write Fluent Bit input plugin
- [x] Implement `fluentbit` sink operator
- [x] Implement `fluentbit` source operator
- [x] Add logic that converts Fluent Bit JSON to table slices
- [x] Apply decoding heuristic
- [x] Fix crash on invalid sink, e.g., `fluent-bit xxx`
- [x] Add integration tests
```

```[tasklist]
### Scaffolding
- [x] Get initial scaffold compiling
- [x] Reconsider rolling our own input plugin, as opposed to pushing JSON through `flb_lib_push`
- [x] Fix Docker build
- [x] Static libraries are currently not installed and scattered across many archive files. We need to manually pass them to the linker command line.
- [x] The few dependencies that are not vendored need to be linked in manually for the static build.
```


mavam:
That build system work is quite a feat!

I had started something local using `crtp_operator`. Is saver/loader the right abstractions? Aren't we going from *void* to *events* directly? I haven't studied the fluent-bit API in depth, perhaps there's a MsgPack parser/printer here as well.

```cpp
struct operator_args {
  located<std::vector<std::string>> args;

  template <class Inspector>
  friend auto inspect(Inspector& f, operator_args& x) -> bool {
    return f.object(x)
      .pretty_name("operator_args")
      .fields(f.field("args", x.args));
  }
};

class fluent_bit_operator final : public crtp_operator<fluent_bit_operator> {
public:
  fluent_bit_operator() = default;

  explicit fluent_bit_operator(operator_args args) : args_{std::move(args)} {
  }

  auto operator()(operator_control_plane& ctrl) const
    -> generator<table_slice> {
    // TODO: implement
  }

  auto name() const -> std::string override {
    return "fluent_bit";
  }

  auto detached() const -> bool override {
    return false;
  }

  auto location() const -> operator_location override {
    return operator_location::local;
  }

    -> optimize_result override {
    (void)order;
    (void)filter;
  }

  friend auto inspect(auto& f, fluent_bit_operator& x) -> bool {
    return f.apply(x.args_);
  }

private:
  operator_args args_;
};

class plugin final : public operator_plugin<fluent_bit_operator> {
public:
  auto signature() const -> operator_signature override {
    return {.source = true};
  }

  auto parse_operator(parser_interface& p) const -> operator_ptr override {
    auto parser = argument_parser{
      name(),
      fmt::format("https://docs.tenzir.com/docs/connectors/{}", name())};
    auto args = operator_args{};
    parser.parse(p);
    return std::make_unique<fluent_bit_operator>(std::move(args));
  }

  auto name() const -> std::string override {
    return "fluent-bit";
  }

private:
  record config_;
};
```

tobim:
> I had started something local using crtp_operator. Is saver/loader the right abstractions? Aren't we going from void to events directly? I haven't studied the fluent-bit API in depth, perhaps there's a MsgPack parser/printer here as well.

I just assumed that we would go to bytes first, but looking at the included minimal API it seems to make more sense to go to directly to events.

This is an implementation choice that you should make as the main implementer or the feature.

mavam:
> This is an implementation choice that you should make as the main implementer or the feature.

Alrighty, can I take it from here?

mavam:
@tobim I've split the outstanding todos in two tasklists for clearer responsibilities. In particular, please review da417f10e5f1eb11aca7ce825c160f90b2463836, which I needed to get things to link on macOS.

I'm surprised that the Docker build fails, as the script that installs the dev deps runs through just fine, and the plain Debian CI build works. Here I have no intuition how to proceed and would probably not invest any time myself. Any help would be appreciated.

plugins/fluent-bit/src/plugin.cpp
```diff
@@ -0,0 +1,153 @@
...
+    -> generator<table_slice> {
+    auto ctx = flb_create();
+    auto destroyer = caf::detail::make_scope_guard([ctx] {
+      flb_destroy(ctx);
+    });
+    if (ctx == nullptr)
+      diagnostic::error("failed to create fluent bit context")
+        .hint("flb_create returned nullptr")
+        .emit(ctrl.diagnostics());
+    // Set services properties.
```
jachris, line 52: You probably want a `return` whenever an error is emitted.

mavam: I sprinkled `co_return` statements across the call sequence.

plugins/fluent-bit/src/plugin.cpp
```diff
@@ -0,0 +1,209 @@
...
+    if (ctx == nullptr)
+      return std::nullopt;
+    return engine{ctx};
+  }
+
+  ~engine() {
+    stop();
+    TENZIR_ASSERT(ctx_ != nullptr);
+    // FIXME: destroying the library context currently yields a segfault.
+    // flb_destroy(ctx_);
```
mavam, line 49: This fails with:

```
tenzir(11088,0x16d0ef000) malloc: *** error for object 0x16d0ef000: pointer being freed was not allocated
tenzir(11088,0x16d0ef000) malloc: *** set a breakpoint in malloc_error_break to debug
tenzir-v4.0.1-51-gafef3fd966-dirty: Error: signal 6 (Abort trap: 6)
0   libtenzir.2816.1.dylib              0x0000000106a6e1a8 _ZN6tenzir6detail9backtraceEv + 40
1   libtenzir.2816.1.dylib              0x0000000106ad78cc fatal_handler + 80
2   libsystem_platform.dylib            0x00000001a1f2ea24 _sigtramp + 56
3   libsystem_pthread.dylib             0x00000001a1effc28 pthread_kill + 288
4   libsystem_c.dylib                   0x00000001a1e0dae8 abort + 180
5   libsystem_malloc.dylib              0x00000001a1d2ee28 malloc_vreport + 908
6   libsystem_malloc.dylib              0x00000001a1d326e8 malloc_report + 64
7   libsystem_malloc.dylib              0x00000001a1d3ef20 find_zone_and_free + 308
8   libfluent-bit.dylib                 0x000000010ebbe878 flb_free + 24
9   libfluent-bit.dylib                 0x000000010ebbfdd0 flb_output_exit + 288
fish: Job 1, 'tenzir fluentbit' terminated by signal SIGABRT (Abort)
```

plugins/fluent-bit/src/plugin.cpp
```diff
@@ -0,0 +1,402 @@
...
+      std::vector<std::pair<std::string, std::string>> kvps;
+      if (!kvp_parser()(options->inner, kvps))
+        diagnostic::error("invalid list of key=value pairs")
+          .primary(options->source)
+          .throw_();
+      for (auto& [key, value] : kvps)
+        args.options.emplace(std::move(key), std::move(value));
+    }
+    // Parse the remainder: <plugin> [<key=value>...]
+    auto plugin = p.accept_shell_arg();
```
mavam, line 371: @jachris turns out this doesn't work. Do you see what I'm after? Any ideas on how to make it work?

jachris: `argument_parser::parse(...)` checks that it consumed all arguments. The proper solution to this is to expand the argument parser to either support what you want to do, or to at least provide a way to handle unparseable arguments. The argument parser is a bit inspired by Python's `argparse`, so this could be done similar to https://docs.python.org/3.4/library/argparse.html#partial-parsing. I wanted to extend it anyway (to support n-ary arguments and so on), but did not get around to it. You can also manually parse all your arguments for now, dropping the `argument_parser` entirely.

mavam: Got it. I switched to manual parsing for now. This is not that painful because there aren't many options here.

plugins/fluent-bit/src/plugin.cpp
mavam: Fixed by now.

plugins/fluent-bit/src/plugin.cpp
```diff
@@ -0,0 +1,447 @@
...
+        //
+        // Speaking with Dominik, we could use simdjson directly here to get the
+        // second array element of the Fluent Bit message [timestamp, {..}],
+        // using something like on_demand::dom::parse. Once we're ready to
+        // extract the contained object, we should go to Jannis as he's
+        // currently reworking the JSON parser. What we need is a library
+        // utility that bakes us a table slice given a JSON object and builder.
+        result = {};
+        return true;
+      };
```
mavam, line 306: @jachris maybe this is helpful for you to know while reworking the JSON parser.

mavam:
I managed to get rid of any custom Fluent Bit plugins. This now works with a stock Fluent Bit package that you get on your system.

The consequence is that we now *have* to go through JSON when using Fluent Bit as sink (Fluent Bit *input*), because the library function `flb_lib_push` only supports JSON. We're also restricted to JSON when using Fluent Bit as source (= Fluent Bit *output*), although here Fluent Bit already provides us with the option of parsing the raw MsgPack.

Net effect: we have the potential to tune performance as source operator, but not as sink operator.

I've asked @edsiper in Slack:

> [E]duardo, would you accept a patch that enhances `in_lib` and `flb_lib_push` to accept MsgPack in addition to JSON? It's already possible to consume MsgPack with `out_lib`, but not yet `in_lib`. This would make the library API symmetric.
>
> I'm envisioning no API change, just making this work:
> ```c
> in_ffd = flb_input(ctx, "lib", NULL);
> flb_input_set(ctx, in_ffd, "format", "msgpack", NULL);
> flb_lib_push(ctx, in_ffd, msgpack_buf, msgpack_buf_len);
> ```

With a bit of luck, this will be a welcome future contribution as it doesn't change anything about the public API.

plugins/fluent-bit/src/plugin.cpp
jachris: It sounds like this should be easy to implement with a bit of refactoring.

plugins/fluent-bit/cmake/Findfluentbit.cmake
mavam, line 1: @tobim @dominiklohmann please let me know if this is bad CMake practice. The official Debian package from Fluent Bit gets installed there, so I thought we should look there as well by default.

I did this because our Docker build cannot find the dependency, even though it's actually installed.

plugins/fluent-bit/src/plugin.cpp
mavam: May I ask you to ping me once you have completed the adaptive builder changes?

plugins/fluent-bit/cmake/Findfluentbit.cmake
mavam: FTR, this didn't work because the headers and libraries were not installed.

```
root@c1158166fa2f:/tmp/tenzir# tree /opt/fluent-bit
/opt/fluent-bit
`-- bin
    `-- fluent-bit

2 directories, 1 file
```

mavam: Since we probably have to [build it from source](https://docs.fluentbit.io/manual/installation/sources/build-and-install) anyways, it's probably a moot point.

tobim: Using hints like this should be fine. We should do whatever works for here.

libtenzir/src/tql/parser.cpp
```diff
@@ -40,8 +40,11 @@ template <>
 struct lexer_traits<identifier> {
   static auto parser() {
     // TODO: This clearly shows the weakness of the current approach.
-    return ((parsers::alpha | parsers::chr{'_'})
-            >> *(parsers::alnum | parsers::chr{'_'}))
+    // TODO: The plugin parser in
+    //       `include/tenzir/concept/parseable/tenzir/identifier.hpp`
+    //       needs to get aligned with this.
+    using namespace parsers;
+    return ((alpha | chr{'_'}) >> *(alnum | chr{'_'} | chr{'-'})) //
```
jachris, line 47: You could alternatively change `parse_operator()` to use `accept_shell_arg()` or something similar. Anyway, this lexing implementation is only a crutch and will hopefully be improved soon as a whole.

tobim: I'd rather keep this minimal change until we have a real lexer.

tobim:
> I'm surprised that the Docker build fails, as the script that installs the dev deps runs through just fine, and the plain Debian CI build works. Here I have no intuition how to proceed and would probably not invest any time myself. Any help would be appreciated.

The regular debian CI job does not build any plugins except for `web`. Plugins are built in a dedicated job matrix to which I just added `fluent-bit`.

The issue in the docker build was quite hairy and took me longer to fix than the static build. Behold: https://github.com/tenzir/tenzir/pull/3461/files#diff-4f95d2e348b638a037c80301070df229e817aaba51d9cd3b868976c7ae96c695R37-R69

mavam:
CI error:
```
CMake Error at CMakeLists.txt:29 (dependency_summary):
  Unknown CMake command "dependency_summary".
```

@tobim IIRC we don't have that function in plugins. @dominiklohmann once told me to just remove it.

tobim:
> @tobim IIRC we don't have that function in plugins. @dominiklohmann once told me to just remove it.

This has been bugging us often enough. I refactored the functionality so it can be reused for standalone plugins in the latest commit.

mavam:
We got it all green! What a feat, @tobim.

plugins/fluent-bit/src/plugin.cpp
```diff
@@ -0,0 +1,505 @@
...
+      // to represent events. This format is still supported for reading input
+      // event streams.
+      //
+      // We are parsing this into a table with the following fields:
+      //
+      // 1. timestamp: time (timestamp alias type)
+      // 2. metadata: record (inferred)
+      // 3. message: record (inferred)
+      //
+      // TODO: do 👆 as opposed to doing 👇
```
mavam, line 339: @jachris I completed the table slice building to the point where two things are missing:

1. Replacing the old adaptive builder with the new one
2. Using simdjson to parse the above structure, i.e., accept both `[[TIMESTAMP, METADATA], MESSAGE]` and `[TIMESTAMP, MESSAGE]`. These are new-line delimited *array*, but that's valid JSONL.

I have neither worked with the API of (1) or (2), and would therefore be thankful if you  could help me with that last bit.

mavam: I've used the new series builder successfully to do this. @jachris beautiful API!

mavam:
This PR is now feature complete and ready for testing. First tests showed that it works, e.g.,

```
fluent-bit random
```

Produces an infinite stream of messages. I was expecting that this works as well:

```bash
yes | head | tenzir 'fluent-bit stdin'
```

However, it takes too long for the Fluent Bit engine to ramp up before the input is already exhausted:

```
❯ yes | head | tenzir 'fluent-bit -X log_level=INFO stdin'
[2023/09/23 07:00:08] [ info] [fluent bit] version=2.1.9, commit=b1bf014171, pid=10812
[2023/09/23 07:00:08] [ info] [storage] ver=1.4.0, type=memory, sync=normal, checks[2023/09/23 07:00:08] [ info] [ctraces ] version=0.3.1
[2023/09/23 07:00:08] [ info] [input:stdin:stdin.0] initializing
[2023/09/23 07:00:08] [ info] [ctraces ] version=0.3.1
[2023/09/23 07:00:08] [ info] [input:stdin:stdin.0] initializing
[2023/09/23 07:00:08] [ info] [input:stdin:stdin.0] storage_strategy='memory' (memory only)
[2023/09/23 07:00:08] [ info] [sp] stream processor started
[2023/09/23 07:00:08] [ warn] [input:stdin:stdin.0] end of file (stdin closed by remote end)
[2023/09/23 07:00:09] [ info] [engine] service has stopped (0 pending tasks)
```

I don't have an answer for that yet.

mavam:
For the record, I reported a bug to the Fluent Bit devs that comes up when destroying an un-started engine. Then we crash in:

```
239           if (ctx->config->is_running == FLB_TRUE) {
-> 240               flb_engine_shutdown(ctx->config);
   241           }
```

Which comes from `flb_output.c`:

```
   479       params = FLB_TLS_GET(out_flush_params);
   480       if (params) {
-> 481           flb_free(params);   <---------- segfault 💥
   482       }
   483   }
```

This thread-local state never got initialized. The internal `do_start` function is the place where this state gets initialized in `flb_output_prepare`, via `pthread_once(&flb_lib_once, flb_init_env)`. However, in our scenario the output got never correctly initialized and the the user hasn't invoked `flb_start` yet, so this state is never initialized. That is, we free a pointer that is never allocated, which is exactly the crash output:

```
tenzir(63774,0x16c21b000) malloc: *** error for object 0x16c21b000: pointer being freed was not allocated
```

We worked around this by calling `flb_init_env` that initializes the thread-local state as side effect.

Dakostu (approved):
I love this operator!

I left some comments below, but I approve the general gist of the implementation.

changelog/next/features/3461-fluent-bit.md
```diff
@@ -0,0 +1,3 @@
+The new `fluent-bit` source and sink operator provide and interface to the
```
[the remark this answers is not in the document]
mavam: It's not a connector, though. Otherwise I'd be able to use `from fluent-bit`. It is its own operator because Fluent Bit has an internal structured format.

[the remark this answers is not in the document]
Dakostu: Yeaah, you're right.
I guess the phrase "source and sink operator" made me suspicious.

plugins/fluent-bit/src/plugin.cpp
```diff
@@ -0,0 +1,608 @@
...
+    if (not result->output(args.plugin, args.args))
+      return nullptr;
+    if (not result->start())
+      return nullptr;
+    return result;
+  }
+
+  ~engine() {
+    if (ctx_ != nullptr) {
+      stop();
```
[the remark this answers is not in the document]
mavam: This call will block until the engine stops, and if it fails, we still try to relinquish all resources afterwards. This strikes me as the correct behavior for what we can do here.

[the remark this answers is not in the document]
mavam: I'd say let's get some experience with this if it's insufficient.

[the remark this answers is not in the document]
Dakostu: Alright.

plugins/fluent-bit/fluent-bit.yaml.example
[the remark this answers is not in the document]
mavam: Turns out this file is needed by the CMake scaffold.

plugins/fluent-bit/src/plugin.cpp
```diff
@@ -0,0 +1,608 @@
...
+      = make_engine(plugin_config, args.poll_interval, args.service_properties);
+    if (not result)
+      return nullptr;
+    if (not result->input("lib"))
+      return nullptr;
+    if (not result->output(args.plugin, args.args))
+      return nullptr;
+    if (not result->start())
+      return nullptr;
+    return result;
```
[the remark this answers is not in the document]
mavam: Mind taking a look at e600c91ccc9881948e0745e40e8af942039ce2dc? This exposes the errors better.

Can we pair on the leak?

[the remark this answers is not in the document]
Dakostu: https://github.com/tenzir/tenzir/commit/e600c91ccc9881948e0745e40e8af942039ce2dc improved the error display.

We paired on the leak.
The aforementioned leak seems to be caused by the AWS SDK _again_ because I was using `fluent-bit s3` as a sink. The AWS SDK had a memory leak problem that I already encountered when implementing the `s3` connector.

mavam:
CI is not looking good:
```
/__w/tenzir/tenzir/_build/lib/tenzir/plugins/libtenzir-plugin-fluent-bit.so /usr/lib/fluent-bit/libfluent-bit.so: cannot allocate memory in static TLS block
```

@tobim we may have to trouble-shoot this together.

Dakostu:
> ```
> /__w/tenzir/tenzir/_build/lib/tenzir/plugins/libtenzir-plugin-fluent-bit.so /usr/lib/fluent-bit/libfluent-bit.so: cannot allocate memory in static TLS block
> ```

Looks like you have to add something like:

```
export LD_PRELOAD=/usr/lib/fluent-bit/libfluent-bit.so
```
maybe together with `/__w/tenzir/tenzir/_build/lib/tenzir/plugins/libtenzir-plugin-fluent-bit.so` to the CI machine setup.
Or maybe even just your `tests.sh` script?

mavam:
So this is just an error that occurs when loading the shared lib? I would imagine that this requires adjusting some CMake that does the linking properly so that this is not needed. 

I would argue that the `LD_PRELOAD` monkey-patching is just a band-aid for an incorrectly linked shared lib. If CI runs into this, our users might as well. So we should fix it properly.

mavam:
I've [found](https://github.com/microsoft/mimalloc/issues/147) that adding `-ftls-model=local-dynamic` might be a more appropriate fix.

tobim:
The main issue is that libfluent-bit.so contains an embedded jemalloc with static thread local storage, and that is not supported with `dlopen()`.
We could switch to build libfluent-bit.so ourselves on linux, but that would set this PR back several days, and I would also expect a few hours per month of maintenance that would have to go into this. Let's see if we can work around the problem in our CI first and try to push a fix upstream directly.

mavam:
@tobim only with the dynamically linked library we get some errors:

```
ERROR: ld.so: object '/usr/lib/gcc/x86_64-linux-gnu/12/libasan.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
ERROR: ld.so: object '/usr/lib/fluent-bit/libfluent-bit.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
```

The error `cannot open shared object file` hints that the file doesn't exist.

PR #3464, opened 2023-08-17
Implement show tables

tobim:
The tables aspect returns the outermost structure of the catalog. Each unique type that has (or had) data is emitted as an event.

### DIFF
--- a/libtenzir/builtins/aspects/types.cpp
+++ b/libtenzir/builtins/aspects/types.cpp
@@ -52,11 +52,23 @@ auto type_type() -> type {
   };
 }
 
+/// A type that describes a table in the catalog.
+auto table_type() -> type {
+  return type{
+    "tenzir.table",
+    record_type{
+      {"events", uint64_type{}},
+      {"min_import_time", time_type{}},
+      {"max_import_time", time_type{}},
+      {"schema", type_type()},
+    },
+  };
+}
+
 // TODO: harmonize this with type::to_definition. The goal was to create a
 // single fixed schema, but it's still clunky due to nested records.
 /// Adds one type definition per row to a builder.
-auto add_type(auto& builder, const type& t) -> caf::error {
-  auto row = builder.push_row();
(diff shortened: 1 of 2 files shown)


libtenzir/builtins/aspects/types.cpp
```diff
@@ -52,11 +52,23 @@ auto type_type() -> type {
...
+  return type{
+    "tenzir.table",
+    record_type{
+      {"events", uint64_type{}},
+      {"min_import_time", time_type{}},
+      {"max_import_time", time_type{}},
+      {"schema", type_type()},
+    },
+  };
+}
```
dominiklohmann, line 66: Is `show tables` just `show partitions | summarize events=sum(events), min_import_time=min(min_import_time), max_import_time=max(max_import_time) by schema`?

If my understand is correct, then two questions:
1. Shouldn't this also group by `schema_id`?
2. Do we need this as an aspect of its own in the first place?

tobim: No it is different. `show partitions` gives the schema names, but this gives the schema definition instead. I don't think changing `show partitions` to give definitions is a good idea, as that would bloat the produced batches quite a lot.

As indicated by how it is implemented, this aspect shares much more similarity with `show types`. My preferred way would be to change `show types` to `show types --all` and `show tables` to `show types`. Types that don't have any data would have `0` events and `null` for the import times.

mavam: > My preferred way would be to change `show types` to `show types --all` and `show tables` to `show types`. Types that don't have any data would have `0` events and `null` for the import times.

I think this is dangerous, as you'd render any record type definition as a table. A schema definition is not a table. A schema definition is really just a *hint*, but nothing more. Schema evolution is happening continuously.

dominiklohmann:
Closing this; this will need to be redone for the catalog revamp in a major way.

PR #3466, opened 2023-08-18
Support building against yaml-cpp 0.8.0

dominiklohmann:


jachris (approved):
I can confirm that this works! :+1:

PR #3468, opened 2023-08-21
Remove newline in serve operator docs

dit7ya:
For the codemirror-lang-tql docs to tooltips processor script to work without lots of edge-case handling, the short description of an operator should be written in one single line.


dominiklohmann (changes requested):
I'm pretty sure this will break again in the near future per our .editorconfig that adds a linebreak at 80 characters.

dit7ya:
I see. Closing it for now.

PR #3470, opened 2023-08-22
Tweak the execution node behavior

dominiklohmann:
This makes the following changes to execution nodes:
- Reduce batch timeout from 250ms to 10ms with exponential backoff to 2s. Execution nodes are now always scheduled after the timeout hits, assuming that the previous operator does not block indefinitely.
- Remove the never correctly implemented logic to advance generators multiple times per scheduled run of an execution node.

Additionally, this fixes the operator location of some aspects.

### DIFF
--- a/changelog/next/bug-fixes/3470--exec-node-tweak.md
+++ b/changelog/next/bug-fixes/3470--exec-node-tweak.md
@@ -1,0 +1,6 @@
+Pipeline operators that create output independent of their input now emit their
+output instantly instead of waiting for receiving further input. This makes the
+`shell` operator more reliable.
+
+The `show <aspect>` operator wrongfully required unsafe pipelines to be allowed
+for some aspects. This is now fixed.
--- a/libtenzir/builtins/aspects/build.cpp
+++ b/libtenzir/builtins/aspects/build.cpp
@@ -21,7 +21,7 @@ public:
   }
 
   auto location() const -> operator_location override {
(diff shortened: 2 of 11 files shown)


mavam:
> * Reduce batch timeout from 250ms to 10ms with exponential backoff to 2s. Execution nodes are now always scheduled after the timeout hits, assuming that the previous operator does not block indefinitely.

Can you explain the use of exponential backoff? This algorithm is usually deployed in a closed-loop feedback system (e.g., collision on a shared medium), and I wonder what the feedback here is.

dominiklohmann:
The exponential backoff is used to throttle an execution node in case it has neither consumed input, nor produced output, nor has further output requested. The new dev-only metrics introduced with this PR shows this working quite well. 

Basically, if you have a pipeline `A | B | C | D`, and `B` is a blocking operator like `sort`, then `C` and `D` will—after a while—be scheduled less often. If, however, `C` produces output independent of its input, it will still be allowed to run and the throttling will not kick in. Before the changes from this PR, `C` and `D` would've not been scheduled at all until `B` produced output.

Additionally, the lowered default timeout makes it so that the minimum time-to-first-result for an infinitely running pipeline with $`n`$ operators is now $`(n - 1) * 10ms`$ rather than $`(n - 1) * 250ms`$, assuming perfect scheduling.

mavam:
Great explanation, thanks a lot!

PR #3478, opened 2023-08-30
Use pinned nixpkgs for nushell script deps

tobim:
Without adding `--inputs-from` we always get the latest versions of the dependencies that are currently available in `nixpkgs`. Explicitly using the revision from the flake is an easy way to pin and avoid unexpected breakage.


dit7ya (approved):

dit7ya:
I hope this fixes the build failure. Nushell isn't broken on nixpkgs-23.05 or nixpkgs-master - so I wonder what caused the issue though? Is it a breaking change in nushell itself?

tobim:
> Is it a breaking change in nushell itself?

That is correct.

PR #3485, opened 2023-09-01
Fix the deprecated operator in the Linux node install script

Dakostu:
This change is a quick fix for the Linux node install script that still used deprecated operators.


dominiklohmann (approved):
👀

PR #3488, opened 2023-09-02
[Draft] Build arm64 images as well

netantho:
Creating arm64 images as Ampere Altra Max systems and Graviton are becoming popular

PR #3489, opened 2023-09-04
Build linux/arm64 Docker images in CI

dominiklohmann:
This cherry-picks the two commits from #3488; our CI does not build the Docker images when running from a fork.


tobim:
Given that the workflow needs more than 6 hours to run with the changes proposed here, I think we need to close this and figure out a different way to support arm64 images.

PR #3496, opened 2023-09-05
Implement the s3 connector

Dakostu:
This change implements an S3 saver and loader.

### DIFF
--- a/web/docs/connectors/s3.md
+++ b/web/docs/connectors/s3.md
@@ -1,0 +1,60 @@
+# s3
+
+Loads bytes from an Amazon S3 object. Saves bytes to an Amazon S3 object.
+
+## Synopsis
+
+Loader:
+
+```
+s3 [--anonymous] <uri>
+```
+
+Saver:
+
+```
+s3 [--anonymous] <uri>
+```
+
+## Description
+
+The `s3` loader connects to an S3 bucket to acquire raw bytes from an S3
+object. The `s3` saver writes bytes to an S3 object in an S3 bucket.
+
+The connector tries to retrieve the appropriate credentials using AWS's
+[default credentials provider
+chain](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-chain.html).
+
+### `<uri>` (Loader, Saver)
+
+The path to the S3 object.
+
+The syntax is
+`s3://<bucket-name>/<full-path-to-object>(?<options>)`.
+
+Options can be appended to the path as query parameters, as per
+[Arrow](https://arrow.apache.org/docs/r/articles/fs.html#connecting-directly-with-a-uri):
+
+> For S3, the options that can be included in the URI as query parameters are
+`region`, `scheme`, `endpoint_override`, `access_key`, `secret_key`,
+`allow_bucket_creation`, and `allow_bucket_deletion`.
+
+### `--anonymous` (Loader, Saver)
+
+Ignore any predefined credentials and try to load/save with anonymous
+credentials.
+
+## Examples
+
+Read CSV from an object `obj.csv` in the bucket `examplebucket`:
+
+```
+from s3 s3://examplebucket/obj.csv read csv
+```
+
+Read JSON from an object `test.json` in the bucket `examplebucket`, but using a
+different, S3-compatible endpoint:
+
+```
+from s3 s3://examplebucket/test.json?endpoint_override=s3.us-west.mycloudservice.com
+```
--- a/web/docs/developer-guides/build-from-source.md
+++ b/web/docs/developer-guides/build-from-source.md
@@ -34,7 +34,7 @@ Every [release](https://github.com/tenzir/tenzir/releases) of Tenzir includes an
 |✓|[OpenSSL](https://www.openssl.org)||Utilities for secure networking and cryptography.|
 |✓|[FlatBuffers](https://google.github.io/flatbuffers/)|>= 1.12.0|Memory-efficient cross-platform serialization library.|
 |✓|[Boost](https://www.boost.org)|>= 1.81.0|Required as a general utility library.|
-|✓|[Apache Arrow](https://arrow.apache.org)|>= 12.0.0|Required for in-memory data representation. Must be built with Compute, Zstd and Parquet enabled.|
+|✓|[Apache Arrow](https://arrow.apache.org)|>= 13.0.0|Required for in-memory data representation. Must be built with Compute, Zstd, Filesystem, S3 and Parquet enabled.|
 |✓|[re2](https://github.com/google/re2)||Required for regular expressione evaluation.|
 |✓|[yaml-cpp](https://github.com/jbeder/yaml-cpp)|>= 0.6.2|Required for reading YAML configuration files.|
 |✓|[simdjson](https://github.com/simdjson/simdjson)|>= 3.1.0|Required for high-performance JSON parsing. (Bundled as submodule.)|
--- a/web/docs/developer-guides/build-from-source.md
+++ b/web/docs/developer-guides/build-from-source.md
@@ -34,7 +34,7 @@ Every [release](https://github.com/tenzir/tenzir/releases) of Tenzir includes an
 |✓|[OpenSSL](https://www.openssl.org)||Utilities for secure networking and cryptography.|
 |✓|[FlatBuffers](https://google.github.io/flatbuffers/)|>= 1.12.0|Memory-efficient cross-platform serialization library.|
 |✓|[Boost](https://www.boost.org)|>= 1.81.0|Required as a general utility library.|
-|✓|[Apache Arrow](https://arrow.apache.org)|>= 13.0.0|Required for in-memory data representation. Must be built with Compute, FIlesystem, S3, Zstd and Parquet enabled. For the `gcs` plugin, GCS needs to be enabled. |
+|✓|[Apache Arrow](https://arrow.apache.org)|>= 13.0.0|Required for in-memory data representation. Must be built with Compute, Filesystem, S3, Zstd and Parquet enabled. For the `gcs` plugin, GCS needs to be enabled. |
 |✓|[re2](https://github.com/google/re2)||Required for regular expressione evaluation.|
 |✓|[yaml-cpp](https://github.com/jbeder/yaml-cpp)|>= 0.6.2|Required for reading YAML configuration files.|
 |✓|[simdjson](https://github.com/simdjson/simdjson)|>= 3.1.0|Required for high-performance JSON parsing. (Bundled as submodule.)|
--- a/Dockerfile
+++ b/Dockerfile
@@ -110,7 +110,7 @@ RUN apt-get update && \
     apt-get -y --no-install-recommends install \
       ./apache-arrow-apt-source-latest-$(lsb_release --codename --short).deb && \
     apt-get update && \
-    apt-get -y --no-install-recommends install libarrow1200 libparquet1200 && \
+    apt-get -y --no-install-recommends install libarrow1300 libparquet1300 && \
     rm -rf /var/lib/apt/lists/*
 
 USER tenzir:tenzir
--- a/changelog/next/features/3496--s3-connector.md
+++ b/changelog/next/features/3496--s3-connector.md
@@ -1,0 +1,2 @@
+The new `s3` connector enables the user to import/export file data from/to S3
+buckets.
--- a/libtenzir/CMakeLists.txt
+++ b/libtenzir/CMakeLists.txt
@@ -62,10 +62,10 @@ TenzirCompileFlatBuffers(
 # -- arrow ---------------------------------------------------------------------
 
 find_package(Arrow REQUIRED CONFIG)
-if (Arrow_VERSION_MAJOR LESS 12)
+if (Arrow_VERSION_MAJOR LESS 13)
   message(
     FATAL_ERROR
-      "Tenzir requires at least Arrow version 12.0, but found ${ARROW_VERSION}")
+      "Tenzir requires at least Arrow version 13.0, but found ${ARROW_VERSION}")
 endif ()
 mark_as_advanced(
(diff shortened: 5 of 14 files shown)


plugins/s3/src/plugin.cpp
```diff
@@ -0,0 +1,277 @@
...
+          auto finalized = arrow::fs::FinalizeS3();
+          if (not finalized.ok()) {
+            diagnostic::error("failed to close Arrow S3 filesystem: {}",
+                              finalized.ToString())
+              .emit(ctrl.diagnostics());
+          }
+        });
+        auto fs = arrow::fs::S3FileSystem::Make(
+          (args.anonymous) ? arrow::fs::S3Options::Anonymous()
+                           : arrow::fs::S3Options::Defaults());
```
netantho, line 68: Is it maybe possible to take a full URI here?
This would allow the user to change the region, scheme and endpoint.

I have following working:
```
auto fs = arrow::fs::FileSystemFromUri(args.uri.inner, &path);
```

Then for most of the rest of the file using `path` instead of `args.uri.inner`.

The pipeline usage then looks like for S3-compatible storage:
```
from s3 s3://my_bucket/my_file?endpoint_override=s3.us-west-002.backblazeb2.com
```

For S3:
```
from s3 s3://my_bucket/my_file
```

Inspiration taken from https://github.com/apache/arrow/blob/cf0cf9bdaaad89d603fe0d6b1daa8a482f3670b7/cpp/src/arrow/filesystem/s3fs_test.cc#L292C12-L292C88

Dakostu: It looks possible, but `FileSystemFromUri` would also create `Arrow::Filesystem`s for other URIs (e.g. `file://` or `gcs://`), so I would have to implement an additional `arrow::fs::S3FileSystem` check. 

But yeah, it is worth a try. Let me try.

Dakostu: BTW: Your example is showing `S3Options::FromUri` - since the `S3Options` is the one struct that holds credential info (with overrides), we may not even need a `FileSystemFromUri` call - we can just call `S3FileSystem::Make` with the options from `S3Options::FromUri`. But I would need to try that first.

OK?

netantho: Yeah, it's actually the one I tried first but `S3Options::FromUri` returns a `Result<S3Options>`. Before I could find how to convert that into just a `S3Options` (didn't search further), I found `FileSystemFromUri` and got the `FileSystemFromUri` version to work.

Dakostu: That's alright, I found a way to convert Arrow's fancy `Result`s.
Now I just need to test around a bit.

Dakostu: I refreshed the PR with an implementation that uses `S3Options::FromUri` with `S3Options::Defaults()` as a fallback.

netantho: Thanks, I've verified that it does work for non-AWS endpoints too :-)

Dakostu: Thank you for your support! :pray:

web/docs/connectors/s3.md
```diff
@@ -0,0 +1,45 @@
...
+The connector tries to retrieve the appropriate credentials using AWS's
+[default credentials provider
+chain](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-chain.html).
+
+### `<uri>` (Loader, Saver)
+
+The path to the S3 object.
+
+The syntax is `s3://<bucket-name>/<full-path-to-object>`.
+
```
netantho, line 33: Addition suggestion:

If you want to use non-AWS endpoints that are S3-compatible, you can pass the alternative endpoint in the URL like that: `s3://<bucket-name>/<full-path-to-object>?endpoint_override=<endpoint_url>`. [More information](https://arrow.apache.org/docs/python/filesystems.html#s3) about URL parameters.

Dakostu: I added `?endpoint_override=<endpoint_url>` to the documentation.

Dakostu: Actually, Arrow's S3 filesystem has many other such options. I extended the documentation.

web/docs/connectors/s3.md
Dakostu, line 1: TODO: Write about S3 in blog post, then resolve this.

plugins/s3/src/plugin.cpp
```diff
@@ -0,0 +1,307 @@
...
+    return std::make_unique<s3_loader>(std::move(args));
+  }
+
+  auto parse_saver(parser_interface& p) const
+    -> std::unique_ptr<plugin_saver> override {
+    auto parser = argument_parser{
+      name(),
+      fmt::format("https://docs.tenzir.com/docs/next/connectors/{}", name())};
+    auto args = s3_args{};
+    parser.add("-x,--anonymous", args.anonymous);
```
dominiklohmann, line 294: I talked about the `-x` short option with Rémi and Jannis at the retreat sometime late in the night, and we considered it to be confusing. Our vote was to remove the short option entirely, and to only support the spelled-out `--anonymous`.

web/docs/developer-guides/build-from-source.md
```diff
@@ -34,7 +34,7 @@ Every [release](https://github.com/tenzir/tenzir/releases) of Tenzir includes an
 |✓|[OpenSSL](https://www.openssl.org)||Utilities for secure networking and cryptography.|
 |✓|[FlatBuffers](https://google.github.io/flatbuffers/)|>= 1.12.0|Memory-efficient cross-platform serialization library.|
 |✓|[Boost](https://www.boost.org)|>= 1.81.0|Required as a general utility library.|
-|✓|[Apache Arrow](https://arrow.apache.org)|>= 12.0.0|Required for in-memory data representation. Must be built with Compute, Zstd and Parquet enabled.|
+|✓|[Apache Arrow](https://arrow.apache.org)|>= 13.0.0|Required for in-memory data representation. Must be built with Compute, Zstd and Parquet enabled. For the `s3` plugin, Filesystem also needs to be enabled.|
```
dominiklohmann, line 37: This needs to also be adjusted in CMake:

https://github.com/tenzir/tenzir/blob/afd5c18d757cc6783f063547ace736bbc74a9429/libtenzir/CMakeLists.txt#L65

plugins/s3/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+  DESCRIPTION "S3 plugin for Tenzir"
+  LANGUAGES CXX)
+
+include(CTest)
+
+find_package(Tenzir REQUIRED)
+
+TenzirRegisterPlugin(
+  TARGET s3
+  ENTRYPOINT src/plugin.cpp)
```
dominiklohmann, line 14: This either needs to be enabled in CI, or should become a builtin instead as previously discussed for GCS.

dominiklohmann (approved):
I verified this to work well with a local MinIO instance.

web/docs/developer-guides/build-from-source.md
Dakostu: Adjusted now, I hope.

plugins/s3/src/plugin.cpp
Dakostu: I removed `-x`.

web/docs/connectors/s3.md
Dakostu: Wrote my first version of the S3 chapter in this PR. @mavam

plugins/s3/CMakeLists.txt
Dakostu: I'll make this a builtin tomorrow.

plugins/s3/src/plugin.cpp
```diff
@@ -0,0 +1,307 @@
...
+                            initialized.ToString())
+            .emit(ctrl.diagnostics());
+          co_return;
+        }
+        auto guard = caf::detail::make_scope_guard([&ctrl]() {
+          auto finalized = arrow::fs::FinalizeS3();
+          if (not finalized.ok()) {
+            diagnostic::error("failed to close Arrow S3 filesystem: {}",
+                              finalized.ToString())
+              .emit(ctrl.diagnostics());
```
dominiklohmann, line 76:
```
error: !! filesystem_error: failed to initialize Arrow S3 functionality: Invalid: Attempt to initialize S3 after it has been finalized
```

I can reproduce this error reliably if I let a managed pipeline writing to or reading from S3 run through, and then try to start a second managed pipeline that does the same. I think we don't want to close the S3 filesystem here yet, but rather do it at the level of the S3 plugin and in its destructor. Similarly, we probably want to initialize it only once in the plugin's `initialize(...)` function.

I assume that the same applies to the GCS plugin, but have not yet verified that.

mavam: I can see this being necessary. Have we thought about what this means wrt. to multiple S3 pipelines using different authentication data? If we initialize S3 only once, it's available only with one global configuration and not per pipeline.

dominiklohmann: It's not a problem if you pass the parameters to the URI, e.g., `load s3 s3://bucket/file.json?endpoint_override=http://localhost:9000&secret_key=secret`.

As a user I would also not expect environment variables to be read more than once, so that's also not a problem.

Dakostu: Yeah, I crossed  _another_ Arrow stumbling block with this one.

I aligned S3 initialization time with the connector initialization time.

plugins/s3/CMakeLists.txt
Dakostu: Made this a builtin now.

plugins/s3/src/plugin.cpp
mavam: Note for the future: putting credentials in the pipeline definition is as bad as putting them on the command line. We need to think about how to factor this, e.g., with a config file as stop gap and potentially things like a secret store later on.

dominiklohmann: I think that's usually not a problem. You choose one S3-compatible store, and then put the credentials in the config file or the env variables that the AWS SDK expects. Doing overrides on a per-pipeline basis is not the common case.

Dakostu: I think Arrow already handles those stop gaps, by using [AWS's credentials provider chain](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-chain.html).

We are trying to get options from the URI, with a fallback to default credentials, from our side and Arrow's side:
https://github.com/apache/arrow/blob/1adc745f67791dde1d9e2574b9de52a8d00ca7e8/cpp/src/arrow/filesystem/s3fs.cc#L364

Dakostu: > I assume that the same applies to the GCS plugin, but have not yet verified that.
The Arrow GCS Filesystem does not have such initialization; you can `Make()` multiple GCS filesystems, and destruction will not affect further initialization.

This one-time initialization thing is actually more like an AWS SDK peculiarity than Arrow's.

netantho: > Note for the future: putting credentials in the pipeline definition is as bad as putting them on the command line. We need to think about how to factor this, e.g., with a config file as stop gap and potentially things like a secret store later on.

My tested workaround for now is to specify the endpoint on the URL but the credentials in environment variables.

Dakostu: > I aligned S3 initialization time with the connector initialization time.

This doesn't actually work. Arrow's S3 deinitialization gets called before the `plugin` destructor, which leads to segmentation faults.

@dominiklohmann I can maybe implement a `plugins::deinitialize` loop in Tenzir's `main` function, or just (de-)initialize S3 there since `s3` is going to be a builtin anyway.

dominiklohmann: I think another easy way to fix this is to manually destroy all plugins at the end of the main loop.

Right before the `return EXIT_SUCCESS` at the very end of `main`, destroy all plugins by calling `plugins::get_mutable().clear()`.

Dakostu: `clear()`ed.

dominiklohmann (approved):

plugins/s3/src/plugin.cpp
Dakostu: This `clear()`ing seems to cause a segmentation fault in the macOS CI step. @dominiklohmann can you reproduce/debug this?

web/docs/developer-guides/build-from-source.md
```diff
@@ -34,7 +34,7 @@ Every [release](https://github.com/tenzir/tenzir/releases) of Tenzir includes an
 |✓|[OpenSSL](https://www.openssl.org)||Utilities for secure networking and cryptography.|
 |✓|[FlatBuffers](https://google.github.io/flatbuffers/)|>= 1.12.0|Memory-efficient cross-platform serialization library.|
 |✓|[Boost](https://www.boost.org)|>= 1.81.0|Required as a general utility library.|
-|✓|[Apache Arrow](https://arrow.apache.org)|>= 13.0.0|Required for in-memory data representation. Must be built with Compute, Zstd, Filesystem, S3 and Parquet enabled.|
+|✓|[Apache Arrow](https://arrow.apache.org)|>= 13.0.0|Required for in-memory data representation. Must be built with Compute, FIlesystem, S3, Zstd and Parquet enabled. For the `gcs` plugin, GCS needs to be enabled. |
```
mavam, line 37:
```suggestion
|✓|[Apache Arrow](https://arrow.apache.org)|>= 13.0.0|Required for in-memory data representation. Must be built with Compute, Filesystem, S3, Zstd and Parquet enabled. For the `gcs` plugin, GCS needs to be enabled. |
```

plugins/s3/src/plugin.cpp
Dakostu: Replaced with `plugin` deinitialization. Thank you @dominiklohmann

PR #3497, opened 2023-09-05
Implement a ZeroMQ connector

mavam:
This PR adds a [ZeroMQ](https://zeromq.org/) connector with the following socket types:

- Loader: `SUB`
- Saver: `PUB`

The assumption is that the pipeline sink *publishes* messages and that the pipeline source *subscribes* to messages. Independent of that is the socket direction: you can use `--bind` or `--connect` to control the socket establishment. By default, the sink binds as publisher and source connects as subscriber.


.github/workflows/tenzir.yaml
```diff
@@ -846,6 +852,8 @@ jobs:
           - {setup: {os: macos-latest}, plugin: {name: Netflow}}
           - {setup: {os: macos-latest}, plugin: {name: Pipeline Manager}}
           - {setup: {os: macos-latest}, plugin: {name: Sigma}}
+          - {setup: {os: macos-latest}, plugin: {name: Kafka}}
+          - {setup: {os: macos-latest}, plugin: {name: ZeroMQ}}
```
dominiklohmann, line 856: Nit: both of the lists here used to be alphabetically sorted.

plugins/zmq/src/plugin.cpp
```diff
@@ -0,0 +1,415 @@
...
+
+private:
+  static void parse(parser_interface& p, connector_args& args) {
+    auto parser
+      = argument_parser{"zmq", "https://docs.tenzir.com/docs/connectors/zmq"};
+    auto endpoint = std::optional<located<std::string>>{};
+    parser.add(endpoint, "<endpoint>");
+    parser.add("-b,--bind", args.bind);
+    parser.add("-c,--connect", args.connect);
+    parser.add(endpoint, "<endpoint>");
```
dominiklohmann, line 398: You add the positional argument to the parser twice here.

web/blog/tenzir-v4.2/index.md
```diff
@@ -0,0 +1,52 @@
+---
+title: Tenzir v4.2
+authors: [dominiklohmann]
+date: 2023-09-12
+tags: [release, pipelines, connectors, zmq]
```
dominiklohmann, line 5: `draft: true`

plugins/zmq/src/plugin.cpp
```diff
@@ -0,0 +1,415 @@
...
+                                       error.num()));
+  }
+
+  static auto poll(::zmq::socket_t& socket, short flags,
+                   std::optional<std::chrono::milliseconds> timeout = {})
+    -> bool {
+    auto items = std::array<::zmq::pollitem_t, 1>{
+      {{socket.handle(), 0, flags, 0}},
+    };
+    auto ms = timeout ? *timeout : std::chrono::milliseconds(-1);
```
dominiklohmann, line 207: Can you add a comment here that explains what -1ms indicates?

dominiklohmann (commented):
I played with this and can confirm it works well in practice. We'll have to figure out the fix for detached operators sometime soon, though, and this still needs a more thorough review than me scrolling through the code for 20 minutes.

plugins/zmq/src/plugin.cpp
mavam: Fixed in d7b5490d85.

.github/workflows/tenzir.yaml
mavam: 34a3e812af

plugins/zmq/src/plugin.cpp
mavam: 45e726759e

web/blog/tenzir-v4.2/index.md
```diff
@@ -0,0 +1,67 @@
...
+interacting with blob storage and [ZeroMQ][zeromq] for writing distributed
+multi-hop pipelines.
+
+[zeromq]: https://zeromq.org/
+
+<!--truncate-->
+
+## S3 Saver & Loader
+
+TODO
```
Dakostu, line 19: I should probably do this. 
After this PR has been merged - then I add my S3 stuff to this file in my PR? @mavam

mavam: Exactly what I was thinking.

plugins/zmq/src/plugin.cpp
```diff
@@ -0,0 +1,447 @@
...
+  auto instantiate(operator_control_plane& ctrl) const
+    -> std::optional<generator<chunk_ptr>> override {
+    auto engine = engine::make_source(args_);
+    if (not engine) {
+      TENZIR_ERROR(engine.error());
+      return std::nullopt;
+    }
+    auto make = [&ctrl](class engine engine) mutable -> generator<chunk_ptr> {
+      if (auto message = engine.receive())
+        co_yield *message;
```
Dakostu, line 311: I "tapped into the raw feed" of my `nic` pipeline, as per the blog post, to receive data, and my `load zmq` pipeline always finishes after receiving the first message. Is this the tapping behavior that you'd expect?

I changed this conditional to a `while` loop, and I got the full-on continuous tap with `nic`. But transmitting non-stream data, e.g. via `from file`, won't terminate the reading pipeline anymore. 

Tired train ride brainstorming: Is there a "terminating message" that the `zmq_saver` can send and the `zmq_loader` only terminates upon receiving that message? Maybe the `engine` can fulfill such behavior?

NOTE: We should probably think about this if a non-Tenzir user wants that kind of feature.

Dakostu (approved):
I played around with this a bit, and I approve. This can be :ship: ped.

I did leave some comments though.

web/docs/connectors/zmq.md
```diff
@@ -0,0 +1,66 @@
...
+Publish the list of TQL operators as CSV and switch from binding to connecting:
+
+```
+show operators | to zmq -c write csv
+```
+
+Subscribe to a specific ZeroMQ endpoint:
+
+```
+from zmq 1.2.3.4:56789
```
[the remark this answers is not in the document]
mavam: Good point. I reworded the example in 84c338d171 to make it clearer what's going on.

web/blog/tenzir-v4.2/index.md
```diff
@@ -0,0 +1,67 @@
...
+connecting socket? No problem:
+
+```bash
+# Bind instead of connect with the ZeroMQ SUB socket:
+tenzir 'from zmq --bind'
+```
+
+These examples show the power of composability: Tenzir operators can work with
+bytes but also events, allowing for in-flight reshaping, format conversation, or
+simply flexibling data shipping.
```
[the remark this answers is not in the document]
mavam: 204ae259d6

plugins/zmq/src/plugin.cpp
mavam: > I "tapped into the raw feed" of my nic pipeline, as per the blog post, to receive data, and my load zmq pipeline always finishes after receiving the first message. Is this the tapping behavior that you'd expect?

No, that's not what I'd expect. I'm glad you tested this. I can repro it and will work on a fix.

When I tested this in more depth, I also realized that a disconnected 0mq socket doesn't stop the source from producing after it started once. This needs more accurate connection management.

> I changed this conditional to a `while` loop, and I got the full-on continuous tap with `nic`. But transmitting non-stream data, e.g. via `from file`, won't terminate the reading pipeline anymore.
>
> Tired train ride brainstorming: Is there a "terminating message" that the `zmq_saver` can send and the `zmq_loader` only terminates upon receiving that message? Maybe the `engine` can fulfill such behavior?

This idea goes in the right direction. The proper fix would be processing the 0mq meta events from monitor. There are basically two socket, the monitoring socket and the data socket. If we get an "disconnect" event from the monitoring socket, we should suspend execution.

A `while` loop can't work, but we need the equivalent of a "main loop" that checks both channels. I'll take a stab at it.

mavam: @Dakostu I was able to fix the issue by fixing the main loop equivalent in saver/loader land. Mind trying it again?

If it works now, just resolve the issue; I've enabled auto-merge.

Dakostu: OK, so now I can tap into `nic` data, but receiving pipelines never terminate, which also includes finite data `from file` delivered via `zmq`.

The `zmq_loader`'s `make` loop does not seem to handle the case where this "disconnect" event happens - the `num_peers` does get decreased in the `poll_monitor`, but with no `num_peers`, the generator is just stuck `co_yield`ing empty chunks. We also can't just `co_return`, it would make receiving pipelines non-deterministic; it already runs into that case while establishing connection.

You probably need a second counter `num_disconnected_peers` that increases with each disconnect event - then in the `make` loop, if `num_peers` is 0, check if `num_disconnected_peers` is not 0, then `co_return`, else `co_yield` empty chunks. Good idea?

Dakostu: I quickly implemented that idea and it looks like it works now. Should I push it?

mavam: I would argue that receiving pipelines should never terminate, or at least not by default. The reason is pub/sub semantics: publishers and subscribes come and go. What's stable is the channel in the middle.

I think you're proposing a flag that's similar to `kafka`'s `-e` to exit, right?

mavam: We discussed this issue in a all and agreed that we are going to leave the issue untouched for this PR, i.e., ship as-is.

The reasoning is that we'd be making strong use case assumptions to terminate a subscriber when the peer count drops to 0, which may work for some but break other use cases. We'd rather wait for user feedback before preemptively adding a `-e` flag.

PR #3498, opened 2023-09-06
Implement the GCS connector plugin

Dakostu:
This change implements a `gcs` plugin, consisting of a `gcs` saver and loader.

### DIFF
--- a/plugins/gcs/CMakeLists.txt
+++ b/plugins/gcs/CMakeLists.txt
@@ -1,15 +1,15 @@
 cmake_minimum_required(VERSION 3.19...3.24 FATAL_ERROR)
 
+project(
+  gcs
+  DESCRIPTION "GCS plugin for Tenzir"
+  LANGUAGES CXX)
+
 find_package(Arrow QUIET REQUIRED CONFIG)
 if (NOT ARROW_GCS OR NOT ${ARROW_CS} STREQUAL "ON")
   message(WARNING "Disabling gcs plugin because the ARROW_GCS flag is not found/enabled. You probably need to enable ARROW_GCS when installing Apache Arrow.")
   return()
 endif()
-
-project(
-  gcs
-  DESCRIPTION "GCS plugin for Tenzir"
-  LANGUAGES CXX)
 
 include(CTest)
 
--- a/plugins/gcs/CMakeLists.txt
+++ b/plugins/gcs/CMakeLists.txt
@@ -1,0 +1,20 @@
+cmake_minimum_required(VERSION 3.19...3.24 FATAL_ERROR)
+
+find_package(Arrow QUIET REQUIRED CONFIG)
+if (NOT ARROW_GCS OR NOT ${ARROW_CS} STREQUAL "ON")
+  message(WARNING "Disabling gcs plugin because the ARROW_GCS flag is not found/enabled. You probably need to enable ARROW_GCS when installing Apache Arrow.")
+  return()
+endif()
+
+project(
+  gcs
+  DESCRIPTION "GCS plugin for Tenzir"
+  LANGUAGES CXX)
+
+include(CTest)
+
+find_package(Tenzir REQUIRED)
+
+TenzirRegisterPlugin(
+  TARGET gcs
+  ENTRYPOINT src/plugin.cpp)
--- a/plugins/gcs/src/plugin.cpp
+++ b/plugins/gcs/src/plugin.cpp
@@ -1,0 +1,279 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The Tenzir Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <tenzir/argument_parser.hpp>
+#include <tenzir/location.hpp>
+#include <tenzir/plugin.hpp>
+
+#include <arrow/filesystem/filesystem.h>
+#include <arrow/filesystem/gcsfs.h>
+#include <arrow/filesystem/type_fwd.h>
+#include <arrow/io/api.h>
+#include <arrow/util/uri.h>
+#include <fmt/core.h>
+
+#include <string>
+
+namespace tenzir::plugins::gcs {
+
+namespace {
+
+struct gcs_args {
+  bool anonymous;
+  located<std::string> uri;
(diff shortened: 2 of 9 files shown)


plugins/gcs/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+  DESCRIPTION "GCS plugin for Tenzir"
+  LANGUAGES CXX)
+
+include(CTest)
+
+find_package(Tenzir REQUIRED)
+
+TenzirRegisterPlugin(
+  TARGET gcs
+  ENTRYPOINT src/plugin.cpp)
```
dominiklohmann, line 14: Afaict this adds no additional dependencies. Should this be a builtin instead of a plugin to keep the build process lean?

(The same applies to S3.)

Dakostu: Sure, but then the Arrow CMake sub-dependencies `ARROW_FILESYSTEM`, `ARROW_S3` and `ARROW_GCS` become mandatory then.
OK?

dominiklohmann: I'd say yes. @tobim are they available in Nix?

Dakostu: GCS also adds the `google-cloud-sdk` dependency.

plugins/gcs/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+cmake_minimum_required(VERSION 3.19...3.24 FATAL_ERROR)
+
+find_package(Arrow REQUIRED CONFIG)
+if (NOT ${ARROW_GCS} OR NOT ${ARROW_CS} STREQUAL "ON")
+  message(WARNING "Disabling gcs plugin because the ARROW_GCS flag is not found/enabled. You probably need to enable ARROW_GCS when installing Apache Arrow.")
+  return()
+endif()
+
+project(
+  gcs
```
dominiklohmann, line 10: This needs to be enabled in CI (just search for `parquet` in `.github/workflows/tenzir.yaml`) and also in the Nix build (see `nix/tenzir/default.nix`).

dominiklohmann (approved):
I can confirm this working well in practice, and just have some minor comments left. Feel free to merge after addressing them.

plugins/gcs/src/plugin.cpp
```diff
@@ -0,0 +1,279 @@
...
+    return std::make_unique<gcs_loader>(std::move(args));
+  }
+
+  auto parse_saver(parser_interface& p) const
+    -> std::unique_ptr<plugin_saver> override {
+    auto parser = argument_parser{
+      name(),
+      fmt::format("https://docs.tenzir.com/docs/next/connectors/{}", name())};
+    auto args = gcs_args{};
+    parser.add("-x,--anonymous", args.anonymous);
```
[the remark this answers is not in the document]
Dakostu: Removed.

plugins/gcs/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+
+include(CTest)
+
+find_package(Tenzir REQUIRED)
+
+TenzirRegisterPlugin(
+  TARGET gcs
+  ENTRYPOINT src/plugin.cpp)
+
+set_target_properties(gcs PROPERTIES LINKER_LANGUAGE CXX)
```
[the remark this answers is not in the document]
Dakostu: I need this on my system, else my CMake won't find the linker language.
But I usually don't commit this line, so it probably snuck in there. Removed.

web/docs/connectors/gcs.md
```diff
@@ -0,0 +1,60 @@
...
+```
+from gcs gs://examplebucket/logs/log.json
+```
+
+Read JSON from an object `test.json` in the bucket `examplebucket`, but using a
+different, GCS-compatible endpoint:
+
+```
+from gcs
+gc://examplebucket/test.json?endpoint_override=gcs.mycloudservice.com
```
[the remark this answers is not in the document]
Dakostu: Fixed.

plugins/gcs/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+cmake_minimum_required(VERSION 3.19...3.24 FATAL_ERROR)
+
+find_package(Arrow REQUIRED CONFIG)
```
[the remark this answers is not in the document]
Dakostu: Fixed.

plugins/gcs/CMakeLists.txt
Dakostu: Added `gcs` to CI & Nix.

plugins/gcs/CMakeLists.txt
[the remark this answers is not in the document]
Dakostu: I need this on my system, else CMake won't be able to determine the linker language.
But I usually don't commit that line, it snuck in there. 

Fixed.

plugins/gcs/CMakeLists.txt
tobim: It is `google-cloud-cpp`, and that needs to be patched to build correctly in static only-mode.

plugins/gcs/CMakeLists.txt
Dakostu: The GCS CI step fails because it can't find Arrow: https://github.com/tenzir/tenzir/actions/runs/6172780703/job/16755081478?pr=3498#step:17:29

Dakostu: I probably have to add an Arrow installation step to `workflows/tenzir.yaml`, right?

plugins/gcs/CMakeLists.txt
tobim: I got `google-cloud-cpp` to build and install correctly in static-only mode. Is it OK if I push to this branch?

Dakostu: YES!

plugins/gcs/CMakeLists.txt
Dakostu: So instead of finding the package - can we expose the `ARROW_GCS` variable from the `libtenzir` `find_package(ARROW)` result somehow?

tobim: Let's figure out why it fails before we come up with a solution.

Dakostu: We found out why it failed - using `find_package` before a `project` declaration makes CMake not search for CMake package in Linux, since CMake probably  can't deduct the system at that point.

Should be fixed now.

PR #3499, opened 2023-09-07
Add libcurl connectors

mavam:
This PR adds a round of loaders based on libcurl:

- `ftp`
- `ftps`
- `http`
- `https`

### DIFF
--- a/libtenzir/CMakeLists.txt
+++ b/libtenzir/CMakeLists.txt
@@ -251,6 +251,15 @@ if (TENZIR_ENABLE_JOURNALD_LOGGING)
   dependency_summary("libsystemd" systemd::systemd "Dependencies")
 endif ()
 
+# Link against curl.
+find_package(CURL REQUIRED)
+if (NOT CURL_FOUND)
+  message(FATAL_ERROR "CURL >= 3.1 required")
+endif ()
+string(APPEND TENZIR_FIND_DEPENDENCY_LIST "\nfind_package(CURL)")
+target_link_libraries(libtenzir PUBLIC CURL::libcurl)
+dependency_summary("CURL" CURL::libcurl "Dependencies")
+
 # Link against fmt.
 find_package(fmt 7.1.3 REQUIRED)
 string(APPEND TENZIR_FIND_DEPENDENCY_LIST "\nfind_package(fmt 7.1.3 REQUIRED)")
--- a/libtenzir/builtins/connectors/curl.cpp
+++ b/libtenzir/builtins/connectors/curl.cpp
@@ -1,0 +1,216 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The Tenzir Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <tenzir/argument_parser.hpp>
+#include <tenzir/detail/string_literal.hpp>
+#include <tenzir/location.hpp>
+#include <tenzir/plugin.hpp>
+
+#include <curl/curl.h>
+
+#include <filesystem>
+#include <system_error>
+
+using namespace std::chrono_literals;
+
+namespace tenzir::plugins {
+
+namespace {
+
+/// Options for the curl wrapper.
+struct curl_options {
+  std::string default_protocol;
+  std::string url;
+};
+
+/// A wrapper around the C API.
+class curl {
+  static auto to_error(CURLcode number) -> caf::error {
+    if (number == CURLE_OK)
+      return {};
+    const auto* str = curl_easy_strerror(number);
+    return caf::make_error(ec::unspecified, fmt::format("curl: {}", str));
+  }
+
+  static auto to_error(CURLMcode number) -> caf::error {
+    if (number == CURLM_OK)
+      return {};
+    const auto* str = curl_multi_strerror(number);
+    return caf::make_error(ec::unspecified, fmt::format("curl: {}", str));
+  }
+
+  static auto write_callback(void* ptr, size_t size, size_t nmemb,
+                             void* user_data) -> size_t {
+    TENZIR_ASSERT(size == 1);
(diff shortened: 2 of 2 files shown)


libtenzir/CMakeLists.txt
```diff
@@ -251,6 +251,15 @@ if (TENZIR_ENABLE_JOURNALD_LOGGING)
   dependency_summary("libsystemd" systemd::systemd "Dependencies")
 endif ()
 
+# Link against curl.
+find_package(CURL REQUIRED)
+if (NOT CURL_FOUND)
+  message(FATAL_ERROR "CURL >= 3.1 required")
```
mavam, line 257: @tobim @dominiklohmann what version should we require?

mavam:
@tobim CI in Nix fails during link time with `undefined reference to `idn2_free'. Do I need to add `libidn2`? If so, how?

web/docs/connectors/http.md
```diff
@@ -0,0 +1,111 @@
...
+
+The `http` loader performs a HTTP request and returns the bytes of the HTTP
+response body.
+
+We modeled the `http` loader after [HTTPie](https://httpie.io/), which comes
+with an expressive and intuitive command-line syntax. We recommend to study the
+[HTTPie documentation](https://httpie.io/docs/cli/examples) to understand the
+full extent of the command-line interface. In many cases, you can perform an
+*exact* copy of the HTTPie command line and use it drop-in with the HTTP loader,
+e.g..,  the invocation
```
Dakostu, line 21:
```suggestion
e.g., the invocation
```

web/docs/connectors/http.md
```diff
@@ -0,0 +1,111 @@
...
+http PUT pie.dev/put X-API-Token:123 foo=bar
+```
+
+becomes
+
+```bash
+tenzir 'from http PUT pie.dev/put X-API-Token:123 foo=bar'
+```
+
+More generally, if you HTTPie command line is `http X` then you can write `from
```
Dakostu, line 33:
```suggestion
More generally, if your HTTPie command line is `http X` then you can write `from
```

Dakostu (approved):
I played around with it -  I was able to curl random files via HTTPS, tested FTP and was even able to control the pipeline manager with this operator:

image

<sup><sub>(Unfortunately, I seem to have hit the limits of our parser with multi-operator pipelines)</sub></sup>

So I'd approve with the general gist of this PR, but some points remain.

libtenzir/src/curl.cpp
```diff
@@ -0,0 +1,173 @@
...
+  TENZIR_ASSERT(not err);
+  auto ms = detail::narrow_cast<int>(timeout.count());
+  auto still_running = int{0};
+  do {
+    auto mc = curl_multi_perform(multi_, &still_running);
+    if (mc == CURLM_OK && still_running != 0)
+      mc = curl_multi_poll(multi_, nullptr, 0u, ms, nullptr);
+    if (mc != CURLM_OK) {
+      TENZIR_ERROR(to_error(mc));
+      co_return;
```
Dakostu, line 159: When I enter an unresolvable URL with correct syntax, `curl` will report:
```
Could not resolve host
```
but `mc` will still be `CURLM_OK`.

I think just relying on `CURLMCode` for error reporting is a bit too naive to be sustainable (see also the URL parsing comment) .

libtenzir/src/curl.cpp
```diff
@@ -0,0 +1,173 @@
...
+      }
+    }
+    if (auto err = set(CURLOPT_HTTPHEADER, headers_))
+      return err;
+  }
+  return {};
+}
+
+auto curl::download(std::chrono::milliseconds timeout) -> generator<chunk_ptr> {
+  // curl_easy_setopt(easy_, CURLOPT_VERBOSE, 1L);
```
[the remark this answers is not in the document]
mavam: I'm almost inclined to add an undocumented `--debug` flag. The output on stderr is incredibly valuable.

Any objections?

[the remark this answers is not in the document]
Dakostu: Fine for now - the endgame for me would be having this verbose (and maybe also `s3` and `gcs` log) output as part of `TENZIR_VERBOSE`.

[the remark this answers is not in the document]
mavam: Okay. It's [possible](https://curl.se/libcurl/c/CURLOPT_STDERR.html) to redirect to a `FILE*` instead of stderr, but that would require some interfacing to our logging framework.

I'll go with `--debug` for now.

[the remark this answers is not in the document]
mavam: 2d0e376a1b

libtenzir/include/tenzir/curl.hpp
```diff
@@ -0,0 +1,87 @@
...
+class curl {
+  static auto to_error(CURLcode number) -> caf::error;
+
+  static auto to_error(CURLMcode number) -> caf::error;
+
+  static auto write_callback(void* ptr, size_t size, size_t nmemb,
+                             void* user_data) -> size_t;
+
+public:
+  static auto parse_url(const std::string& str) -> bool;
```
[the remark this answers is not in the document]
Dakostu: We agreed to replace this dead code with a TODO/FIXME.

libtenzir/src/curl.cpp
Dakostu: We will leave this as-is for now and wait for user feedback.

mavam: I lifted the error handling in e83f661aec, but even though DNS resolution fails, `curl_multi_perform` still returns "okay". I'm not sure how to proceed, as a quick Google search didn't yield any insight.

Let's revisit this when it becomes a bigger problem.

libtenzir/include/tenzir/curl.hpp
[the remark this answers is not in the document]
mavam: Turns out the FIXME exits already, so I just removed the dead code in 61c8bcd59c.

libtenzir/src/curl.cpp
jachris: As someone who didn't know about this behavior and just stumbled upon it, I was very confused. I created https://github.com/tenzir/issues/issues/986 for this.

PR #3500, opened 2023-09-07
Add `from <uri>` and `load <uri>`

jachris:
This PR makes it so that we resolve `from <scheme>://<path>` to `from <scheme> <path>`, which will enable `from https://example.org/...`, for example.

- [X] Implement `from <uri> [options...] [read <parser>]`
- [ ] Implement `from <path>` for `from file <path>`
- [ ] Consider some loader restrictions + remapping (e.g., for `gcs`)
- [ ] Same for `load`
- [ ] Cleanup

### DIFF
--- a/libtenzir/builtins/operators/from_load_read.cpp
+++ b/libtenzir/builtins/operators/from_load_read.cpp
@@ -16,6 +16,82 @@
 namespace tenzir::plugins::from {
 namespace {
 
+class prepend_token final : public parser_interface {
+public:
+  prepend_token(located<std::string> token, parser_interface& next)
+    : token_{std::move(token)}, next_{next} {
+  }
+
+  auto accept_shell_arg() -> std::optional<located<std::string>> override {
+    if (token_) {
+      return std::exchange(token_, std::nullopt);
+    }
+    return next_.accept_shell_arg();
+  }
+
+  auto peek_shell_arg() -> std::optional<located<std::string>> override {
+    if (token_) {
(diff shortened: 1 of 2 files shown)


libtenzir/builtins/operators/from_load_read.cpp
```diff
@@ -183,20 +259,44 @@ class from_plugin final : public virtual operator_parser_plugin {
...
+        auto sub = x.inner.substr(pos, count);
+        auto sub_source = location::unknown;
+        if (x.source && (x.source.end - x.source.begin) == x.inner.size()) {
+          sub_source.begin = x.source.begin + pos;
+          sub_source.end = sub_source.begin + sub.length();
+        }
+        return {std::move(sub), sub_source};
+      };
+      auto scheme = substr(*l_name, 0, split);
+      auto path = substr(*l_name, split + 3);
```
dominiklohmann, line 285: Both CAF and Arrow have their own [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986)-compliant URI parser. It probably makes a lot of sense to use it so that we can correctly the default parser for URIs like `s3://foo/bar.json?endpoint_override=localhost:3000`.

mavam: And soon we have one from libcurl as well. 😉

eliaskosunen:
Superseded by https://github.com/tenzir/tenzir/pull/3608.

PR #3501, opened 2023-09-11
Fix nix/nixos instructions in the installer script

tobim:


dit7ya (approved):
This works as expected. Although I wonder why do we not provide `tenzir-node` as an output under `apps.<system>`?

PR #3511, opened 2023-09-13
Add `lines` parser

mavam:
This PR adds a `lines` parser that chops up its input at newlines.


libtenzir/builtins/formats/lines.cpp
```diff
@@ -0,0 +1,148 @@
...
+        //        or last_finish + defaults::import::batch_timeout < now)
+        //
+        // To be discussed in review.
+        if (builder.rows() == 10) {
+          co_yield builder.finish();
+          builder = make_builder();
+        }
+      }
+      if (builder.rows() > 0)
+        co_yield builder.finish();
```
dominiklohmann, line 107:
```suggestion
        co_yield builder.finish("tenzir.lines");
```

For consistency with other operators and formats that have a schema without a name.

libtenzir/builtins/formats/lines.cpp
```diff
@@ -0,0 +1,148 @@
...
+};
+
+class plugin final : public virtual parser_plugin<lines_parser> {
+public:
+  auto parse_parser(parser_interface& p) const
+    -> std::unique_ptr<plugin_parser> override {
+    auto parser = argument_parser{
+      name(), fmt::format("https://docs.tenzir.com/docs/formats/{}", name())};
+    auto args = parser_args{};
+    parser.add("-f,--field-name", args.field_name, "<string>");
```
dominiklohmann, line 129: I'm not sure this should be configurable at this level. Rename and put likely can already do this, so it's more or less duplicate functionality.

libtenzir/builtins/formats/lines.cpp
dominiklohmann: Ah nvm, you're not using the adaptive builder and instead have a fixed schema above.

libtenzir/builtins/formats/lines.cpp
```diff
@@ -0,0 +1,148 @@
...
+            .hint("{}", *line)
+            .emit(ctrl.diagnostics());
+          co_return;
+        }
+        // FIXME: make configurable. Like in PCAP?
+        //
+        //    if (builder.rows() >= defaults::import::table_slice_size
+        //        or last_finish + defaults::import::batch_timeout < now)
+        //
+        // To be discussed in review.
```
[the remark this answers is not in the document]
dominiklohmann: We have a separate story for generalizing timeout and size limit handling with builders in operators. The simple answer is that we just don't know yet. But it's also not yet something we need to tackle.

libtenzir/builtins/formats/lines.cpp
mavam: No strong opinion here. What should be the default then? I can think of:

1. `line`
2. `data`

libtenzir/builtins/formats/lines.cpp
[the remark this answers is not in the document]
mavam: Should I copy the logic from the `pcap` parser for now?

libtenzir/builtins/formats/lines.cpp
dominiklohmann: I like `line` as well.

libtenzir/builtins/formats/lines.cpp
```diff
@@ -0,0 +1,148 @@
...
+            co_yield {};
+            continue;
+          }
+        } else {
+          ++num_non_empty_lines;
+        }
+        auto num_lines = skip_empty ? num_non_empty_lines
+                                    : num_non_empty_lines + num_empty_lines;
+        if (not builder.add(*line)) {
+          diagnostic::error("failed add line")
```
dominiklohmann, line 89:
```suggestion
          diagnostic::error("failed to add line")
```

dominiklohmann (approved):
Approving modulo the already open comments.

libtenzir/builtins/formats/lines.cpp
```diff
@@ -0,0 +1,148 @@
...
+  template <class Inspector>
+  friend auto inspect(Inspector& f, parser_args& x) -> bool {
+    return f.object(x)
+      .pretty_name("parser_args")
+      .fields(f.field("field_name", x.field_name),
+              f.field("skip_empty", x.skip_empty));
+  }
+};
+
+auto line_type(std::string field_name = "data") -> type {
```
jachris, line 37:
```suggestion
auto line_type(std::string field_name) -> type {
```

jachris (commented):
I can see that this can be quite useful to have!

libtenzir/builtins/formats/lines.cpp
jachris: We could also just assert in such cases. I mean, this should never fail, and we don't have to do so much bookkeeping then.

mavam: What's the maximum line size?

libtenzir/builtins/formats/lines.cpp
mavam: Removed the configurability of the field name in d4a7c856df and renamed the field to `line`.

libtenzir/builtins/formats/lines.cpp
jachris: Good question. Right now, it's more or less unlimited. But that's probably fine? If the input has huge lines, this is what you want. And if it has no newlines at all, you are screwed anyway.

libtenzir/builtins/formats/lines.cpp
```diff
@@ -0,0 +1,148 @@
...
+    if (args.field_name and args.field_name->inner.empty())
+      diagnostic::error("field name must not be empty")
+        .primary(args.field_name->source)
+        .throw_();
+    return std::make_unique<lines_parser>(std::move(args));
+  }
+
+  auto name() const -> std::string override {
+    return "lines";
+  }
```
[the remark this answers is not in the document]
mavam: Oh, right. Thanks! Removed in dcaa2c9de9.

libtenzir/builtins/formats/lines.cpp
```diff
@@ -0,0 +1,148 @@
...
+};
+
+auto line_type(std::string field_name = "data") -> type {
+  TENZIR_ASSERT(not field_name.empty());
+  return type{
+    "tenzir.line",
+    record_type{
+      {std::move(field_name), string_type{}},
+    },
```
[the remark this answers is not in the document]
mavam: With series, this should probably be a `series<string>`. Until then, the best we have is a record with a single string. So yeah, I think this means "I agree"?!

Feel free to resolve. 🙂

libtenzir/builtins/formats/lines.cpp
mavam: So should we then keep the error handling, if it *could* fail beyond 2GB lines?

I don't think it's realistic, but the error handling is there now.

jachris: Seems like Arrow just does a `static_cast<int32_t>(str.size())`, so... If we want to handle this, we must add some additional logic to `append_builder(...)`.

dominiklohmann: The limit Arrow has here is actually 2 GiB per array. This applies to all arrays. Individual arrays are stored in FlatBuffers tables, which have a 32 bit offset. FlatBuffers recently added a 64-bit builder, but that'd be a breaking change in Arrow's binary format.

mavam: Welp, I'm still missing guidance on what to actually do here now.

libtenzir/builtins/formats/lines.cpp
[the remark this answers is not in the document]
mavam: Implemented in 3751e47c16.

libtenzir/builtins/formats/lines.cpp
dominiklohmann: Keep the error handling as-is, and fix the grammar issue. You could add an `[[unlikely]]` if you want to prematurely optimize, but that doesn't matter much.

jachris: Agreed. Well, almost: You should remove (or truncate) the ` .hint("{}", *line)`.

mavam: Ah, right. I removed it.

PR #3515, opened 2023-09-15
Do not drop the `data` field in `decapsulate`

dominiklohmann:
This changes the `decapsulate` operator to keep the `data` field in its incoming packet events that rather than dropping it. It does so without copying the data.

### DIFF
--- a/libtenzir/builtins/operators/decapsulate.cpp
+++ b/libtenzir/builtins/operators/decapsulate.cpp
@@ -372,8 +372,9 @@ public:
           ctrl.diagnostics().emit(std::move(*diag));
       }
       // Add back the untouched data column at the end before yielding.
+      auto new_slice = builder.finish();
       auto transformation = indexed_transformation{
-        .index = {slice.columns() - 1},
+        .index = {caf::get<record_type>(new_slice.schema()).num_fields() - 1},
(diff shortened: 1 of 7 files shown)


libtenzir/builtins/operators/decapsulate.cpp
```diff
@@ -370,7 +371,25 @@ class decapsulate_operator final : public crtp_operator<decapsulate_operator> {
         if (auto diag = parse(row, raw_frame, inferred_type))
           ctrl.diagnostics().emit(std::move(*diag));
       }
-      co_yield builder.finish();
+      // Add back the untouched data column at the end before yielding.
+      auto transformation = indexed_transformation{
+        .index = {slice.columns() - 1},
```
mavam, line 376: Is this the correct index? I tried the sample data and get a segfault:

```
zcat < tenzir/integration/data/pcap/example.pcap.gz | tenzir 'read pcap | decapsulate'
../libtenzir/src/arrow_table_slice.cpp:401: assertion failed 'index out of bounds'
```

mavam: @dominiklohmann fixed this in 8330237f45.

Please resolve after having taken a look at it.

PR #3517, opened 2023-09-15
Support `show nics` to see network interfaces

mavam:
This PR makes `show nics` possible by turning the `nic` plugin into an aspect plugin.


dominiklohmann (approved):
I can confirm this works both locally on macOS and via Docker.

NB: Easiest way to test is this:

```
❯ docker run ghcr.io/tenzir/tenzir:topic-show-nics 'show nics'
```

The code looks good to me, it's a nice small addition.

PR #3518, opened 2023-09-15
Extend the 'http_method' enum with more common verbs

lava:


dominiklohmann (approved):
This seems harmless enough, so I think this can go in before the v4.2 release.

lava:
The "tenzir_features" part was added after the initial approval according to the designs we discussed in Slack, I went over it in a video call with @Dakostu

PR #3519, opened 2023-09-16
Fix the documentation CI job

tobim:
This PR updates the quarto Dockerfile which started to fail due to remote changes.

It also switches to the new official installation method of NodeJS.


dominiklohmann (approved):
🙏

PR #3521, opened 2023-09-18
Show processes and sockets

mavam:
This PR is a POC that we can expose detailed endpoint-centric information. The new `show processes` aspect provides a list of running processes.

Example:

```bash
tenzir '
    show processes
    | select pid, ppid, name, utime, stime
    | head 3
    | write json -C
    '
```

Possible output:

```json
{
  "name": "tenzir",
  "pid": 85558,
  "ppid": 73678,
  "utime": "213.75ms",
  "stime": "30.87ms"
}
{
  "name": "mdworker_shared",
  "pid": 85496,
  "ppid": 1,
  "utime": "49.08ms",
  "stime": "22.42ms"
}
{
  "name": "mdworker_shared",
  "pid": 85495,
  "ppid": 1,
  "utime": "59.11ms",
  "stime": "38.93ms"
}
```


mavam:
@tenzir/engineering does anyone have an idea why this is not performing the conditional include as I'd expect?

```cpp
#if defined(TENZIR_MACOS)
#  include <mach/mach_time.h>
#endif
```

CI gives this error:

```
/__w/tenzir/tenzir/libtenzir/include/tenzir/os.hpp:17:12: fatal error: mach/mach_time.h: No such file or directory
   17 | #  include <mach/mach_time.h>
      |            ^~~~~~~~~~~~~~~~~~
```

jachris:
> ```c++
> #if defined(TENZIR_MACOS)
> #  include <mach/mach_time.h>
> #endif
> ```

You have to use `#if TENZIR_MACOS` instead, because `TENZIR_MACOS` is _always_ defined (but either to `0` or `1`).

mavam:
@jachris thanks 🙈

mavam:
@tobim can you take this from here and push it over the finishing line? *Only* CMake work for pfs is remaining. 😬

changelog/next/features/3521.md
dominiklohmann, line 1: The documentation for `show` describes the operator like this:

> Returns meta information about Tenzir and nodes.

I think we should keep the scope of it to that exactly, and move both of these operators of their own. The same probably applies to `show nics`, which likewise could be made an operator of its own.

dominiklohmann: There is actually an additional argument in favor of this: Currently, all aspects except for `processes`, `sockets`, and `nics` have the location `remote` or `anywhere`. As such, running `show processes` through the API is considered unsafe. I would argue that _all_ aspects should be considered `remote` so the location does not come as a surprise to the user. We should probably have an additional `version` operator to get the local version information. That way, we'd also have `tenzir version` back.

web/docs/developer-guides/build-from-source.md
```diff
@@ -47,6 +47,7 @@ Every [release](https://github.com/tenzir/tenzir/releases) of Tenzir includes an
 ||[librdkafka](https://github.com/confluentinc/librdkafka)||Required for building the `kafka` plugin.|
 ||[http-parser](https://github.com/nodejs/http-parser)||Required for building the `web` plugin.|
 ||[cppzmq](https://github.com/zeromq/cppzmq)||Required for building the `zmq` plugin.|
+||[pfs](https://github.com/dtrugman/pfs)||Required for `show` aspects on Linux.|
```
mavam, line 50:
```suggestion
||[pfs](https://github.com/dtrugman/pfs)||Required for `processes` and `sockets` on Linux.|
```

web/docs/operators/sources/processes.nd
mavam, line 1: Off-by-one in the file extension.

web/docs/operators/sources/processes.nd
```diff
@@ -0,0 +1,30 @@
...
+
+## Synopsis
+
+```
+processes
+```
+
+## Description
+
+The `processes` operator shows a snapshot of all currently open processes.
```
mavam:
```suggestion
The `processes` operator shows a snapshot of all currently running processes.
```

libtenzir/builtins/operators/show.cpp
```diff
@@ -98,6 +98,18 @@ class plugin final : public virtual operator_plugin<show_operator> {
...
+      // We moved `show version` to `version` when we made the `show` operator
+      // always report a remote location for consistency. This makes it less of
+      // a breaking change.
+      if (aspect->inner == "version") {
+        auto op = pipeline::internal_parse_as_operator("version");
+        if (not op) {
+          diagnostic::error("failed to parse `version` operator: {}",
+                            op.error())
+            .throw_();
+        }
```
mavam, line 110: Shouldn't this be an assertion?

dominiklohmann: You can run into this with this config file:

```yaml
tenzir:
  disable-plugins:
    - version
```

dominiklohmann (approved):

PR #3526, opened 2023-09-18
Fix program startup of the static binary

tobim:
The recent addition of the s3 connector causes the linker to add symbols from the AWS crypto abstraction layer library. This lib uses the `dlopen` and `dlclose` functions even when built as a static library - which is possible. What's more, the return value of those functions is asserted, causing the tenzir executables to abort during initialization.

The good news is that this can be worked around by disabling the problematic assert. For that we pull a patch that does exactly that from an open upstream pull request.

Fixes https://github.com/tenzir/issues/issues/885.

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -21,6 +21,18 @@ in {
           ./google-cloud-cpp/0001-Use-pkg-config-to-find-CURL.patch
         ];
       });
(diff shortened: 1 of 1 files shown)


dit7ya (approved):
I can confirm this fixes the linked issue.

PR #3531, opened 2023-09-19
Fix installer on debian docker

rdettai:
Follow up to https://github.com/tenzir/tenzir/pull/3502 that fixes the failing CI: https://github.com/tenzir/tenzir/actions/runs/6230415620/job/16926081974

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -172,8 +172,14 @@ then
 fi
 if [ "${platform}" = "Debian" ]
 then
+  if ! check adduser
+  then
+    echo "Could not find ${bold}adduser${normal} in \$PATH,"
+    echo "it is required for the Debian package install."
+    exit 1
+  fi
   cmd1="sudo dpkg -i \"${tmpdir}/${package}\""
-  cmd2="sudo systemctl status tenzir-node"
+  cmd2="sudo systemctl status tenzir-node || [ ! -d /run/systemd/system ]"
   echo "This script is about to run the following commands:"
   echo
   echo "  - ${cmd1}"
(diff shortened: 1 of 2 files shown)


scripts/install.sh
```diff
@@ -172,8 +172,14 @@ then
 fi
 if [ "${platform}" = "Debian" ]
 then
+  if ! check adduser
+  then
+    echo "Could not find ${bold}adduser${normal} in \$PATH,"
+    echo "it is required for the Debian package install."
```
mavam, line 178:
```suggestion
    echo "Could not find ${bold}adduser${normal} in \$PATH."
```

mavam (approved):

scripts/install.sh
mavam: It's clear if we need if we err. Given that we don't say that for other errors, I'd just remove it here.

rdettai: Yeah I hesitated. It's not really obvious why `adduser` would be needed here (compared to sudo or curl). I thought adding some rational in the error would be at the same time a hint for the user and for the script maintainer.

mavam: But you're not saying *why* it's needed. That would help. But it's obvious that it's needed if we bail out.

rdettai: haha, I actually don't know _why_, I only know _who_ 😄

rdettai: I can blame but not explain nor solve!

mavam: User or developer first? Your call. 😉

rdettai: developer relationship as a comment, user relationship to stdout!

PR #3533, opened 2023-09-19
Change type of `version` in `suricata.quic` to `string`

jachris:
See https://docs.suricata.io/en/latest/rules/quic-keywords.html#quic-cyu-string.


mavam (approved):

netantho:
Thanks for the fix! :bow:

PR #3535, opened 2023-09-20
Fix processing of http arguments

mavam:
This PR fixes a bug with the `http` loader. Adding customer headers didn't work prior to this fix.


dominiklohmann (approved):
Makes sense to me.

PR #3536, opened 2023-09-21
Disable 0mq socket lingering

mavam:
This PR disables 0mq socket lingering to enable instant freeing of resources.


dominiklohmann (approved):
I can confirm that this works as expected. We may want to expose this as an option, but it definitely shouldn't be on by default, I think.

plugins/zmq/src/plugin.cpp
dominiklohmann, line 1: This should probably get a changelog entry in the `changes` category.

mavam: Added one.

mavam:
I'd keep it internal until a user wants to have explicit control over lingering. It's and advanced option anyways.

PR #3537, opened 2023-09-21
Implement serialization_plugins for store_plugins

Dakostu:
This PR fixes `store_plugin`s being unusable due to assertions being fired during normal usage.

### DIFF
--- a/libtenzir/include/tenzir/plugin.hpp
+++ b/libtenzir/include/tenzir/plugin.hpp
@@ -699,7 +699,8 @@ public:
 /// A base class for plugins that add new store backends.
 class store_plugin : public virtual store_actor_plugin,
                      public virtual parser_parser_plugin,
-                     public virtual printer_parser_plugin {
+                     public virtual printer_parser_plugin,
+                     public virtual printer_serialization_plugin {
 public:
   /// Create a store for passive partitions.
   [[nodiscard]] virtual caf::expected<std::unique_ptr<passive_store>>
@@ -709,6 +710,12 @@ public:
   /// @param tenzir_config The tenzir node configuration.
   [[nodiscard]] virtual caf::expected<std::unique_ptr<active_store>>
   make_active_store() const = 0;
+
+  [[nodiscard]] auto serialize(inspector& f, const plugin_printer& x) const
+    -> bool override;
+
+  auto deserialize(inspector& f, std::unique_ptr<plugin_printer>& x) const
+    -> void override;
 
 private:
   [[nodiscard]] caf::expected<builder_and_header>
--- a/libtenzir/include/tenzir/plugin.hpp
+++ b/libtenzir/include/tenzir/plugin.hpp
@@ -700,6 +700,7 @@ public:
 class store_plugin : public virtual store_actor_plugin,
                      public virtual parser_parser_plugin,
                      public virtual printer_parser_plugin,
+                     public virtual parser_serialization_plugin,
                      public virtual printer_serialization_plugin {
 public:
   /// Create a store for passive partitions.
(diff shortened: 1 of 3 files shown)


libtenzir/builtins/stores/feather.cpp
```diff
@@ -309,6 +309,54 @@ class plugin final : public virtual store_plugin {
...
+        return r.get().object(x).fields();
+      },
+      [&x](std::reference_wrapper<caf::binary_deserializer> r) {
+        return r.get().object(x).fields();
+      },
+      [&x](std::reference_wrapper<caf::detail::stringification_inspector> r) {
+        return r.get().object(x).fields();
+      }};
+    auto obj = std::visit(o, f);
+    return f.apply(obj);
```
jachris, line 331: `[&](auto f) { return f.get().object(x).fields(); }`

libtenzir/builtins/stores/feather.cpp
jachris, line 1: I think you should be able to implement this in a generic way, in `plugin.cpp`. No need to define it for every store individually then.

libtenzir/include/tenzir/plugin.hpp
```diff
@@ -726,6 +728,70 @@ class store_plugin : public virtual store_actor_plugin,
     -> std::unique_ptr<plugin_printer> final;
 };
 
+/// A printer instance class that implements the finer details of store plugin
+/// printing.
+class store_printer_impl : public printer_instance {
```
jachris, line 733: It should be possible to leave this in the `.cpp`, I think.

Dakostu: The `store_printer` needs a `store_printer_impl` definition for `return std::make_unique<store_printer_impl>(std::move(*store), ctrl);`
And as of now the `store_plugins` need the `store_printer` for `deserialize`.

libtenzir/builtins/stores/feather.cpp
Dakostu: Simplified the overloads.

libtenzir/builtins/stores/feather.cpp
Dakostu: Alright, I figured it out. 
`store_plugin`s `override` serialization and deserialization at the definition stage now.
Should be fixed now.

libtenzir/include/tenzir/plugin.hpp
Dakostu: Now the `store_printer`s  are back in the `.cpp`.

libtenzir/src/plugin.cpp
```diff
@@ -548,12 +542,40 @@ class store_printer final : public plugin_printer {
...
+  auto o = detail::overload{[&](auto g) {
+    return g.get().object(x).fields();
+  }};
+  auto obj = std::visit(o, f);
+  if (f.apply(obj)) {
+    x = std::make_unique<store_printer>(this);
+  } else {
+    x = nullptr;
+  }
+}
```
jachris, line 578: Much better! 

Nit: In the first call to `object(x)`, `x` is a `plugin_printer`, while in the second call, it's a `std::unique_ptr<plugin_printer>`. I don't think it matters for the inspectors that we are using, but in general, it would be best if both were the same type, perhaps even `store_printer` (deserialization then needs `x = std::make_unique` before inspecting `*x`, but that way the types are as accurate as possible).

libtenzir/include/tenzir/plugin.hpp
```diff
@@ -699,7 +699,8 @@ class store_actor_plugin : public virtual plugin {
 /// A base class for plugins that add new store backends.
 class store_plugin : public virtual store_actor_plugin,
                      public virtual parser_parser_plugin,
-                     public virtual printer_parser_plugin {
+                     public virtual printer_parser_plugin,
+                     public virtual printer_serialization_plugin {
```
jachris, line 703: We are missing `parser_serialization_plugin` here.

jachris (approved):
Yes, that's the way!

libtenzir/include/tenzir/plugin.hpp
Dakostu: Implemented in https://github.com/tenzir/tenzir/pull/3537/commits/3df56ef98d8221b26dbe8a8d9ef09f48dbf5c487.

`read`ing works now, too.

PR #3541, opened 2023-09-26
Add pipeline label support to the main repository

Dakostu:
This PR is a submodule pointer bump + `http_api` support for pipeline labels in HTTP requests.

### DIFF
--- a/changelog/next/features/3541.md
+++ b/changelog/next/features/3541.md
@@ -1,0 +1,1 @@
+The pipeline manager now supports user-provided labels for pipelines.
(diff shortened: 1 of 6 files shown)


dominiklohmann (approved):

contrib/tenzir-plugins
dominiklohmann, line 1: Please also add a feature flag for pipeline labels in version.cpp so that the app can conditionally render them.

Dakostu: Added the feature flag.

PR #3542, opened 2023-09-26
Provide meaningful error when encountering PCAPng

mavam:
This PR gives a nicer error message when encountering PCAPng instead of PCAP dataflows.

### DIFF
--- a/libtenzir/include/tenzir/pcapng.hpp
+++ b/libtenzir/include/tenzir/pcapng.hpp
@@ -1,0 +1,22 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The Tenzir Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#pragma once
+
+#include <cstdint>
+
+/// PCAPng utilities and data structures as defined in the IETF draft at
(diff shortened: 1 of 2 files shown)


mavam:
Here's how things look like now:

```
❯ tenzir 'read pcap' < data.pcapng 

error: PCAPng currently unsupported
 = hint: use `shell "tshark -F pcap -r - -w -"` to convert to PCAP
 = note: visit https://github.com/tenzir/public-roadmap/issues/75
```

libtenzir/include/tenzir/pcapng.hpp
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-License-Identifier: BSD-3-Clause
+
+#pragma once
+
+#include <cstdint>
+
+/// PCAPng utilities and data structures as defined in the IETF draft at
+/// https://datatracker.ietf.org/doc/draft-tuexen-opsawg-pcapng/. Visit
+/// https://pcapng.com/ for a high-level overview about the PCAPng format.
+namespace tenzir::pcapng {
```
lava, line 16: A separate header + namespace is maybe a tiny little bit of overkill? ;)

This probably could have just been `pcapng_magic_number` in `pcap.hpp`, but on the other hand having one more header doesn't hurt anyone so feel free to merge as-is.

lava (approved):

libtenzir/include/tenzir/pcapng.hpp
mavam: I hear you. The reason why I did this is because we're going to populate that header as we add PCAPng incrementally. Overkill for what we have hear, but it sets the framing.

PR #3544, opened 2023-09-29
Fix newly discovered ansible linting errors

Dakostu:
This change fixes Ansible linting errors that (probably a new version of) the Ansible linter discovered.


dominiklohmann (approved):

tobim:
These are actually false positives, see https://github.com/ansible/ansible-lint/issues/3544.
I recommend to disable that check temporarily.

tobim:
I'll take this over.

tobim (approved):
It seems I have to backtrack. Role variables do not leak into neighboring roles and the playbook scope since https://github.com/ansible/ansible/pull/69040, but a role and its nested roles share a single namespace, and the latest assignment simply wins.

Considering this unfortunate situation the proposed solution is the least evil, otherwise users trying to nest our role are in for a nasty surprise.

PR #3547, opened 2023-10-02
Print a log message upon platform reconnect

tobim:
This adds a log message to inform the user when a disconnected node re-established the connection to the Tenzir Platform.

### DIFF
--- a/nix/tenzir/plugins/source.json
+++ b/nix/tenzir/plugins/source.json
@@ -2,7 +2,7 @@
   "name": "tenzir-plugins",
   "url": "git@github.com:tenzir/tenzir-plugins",
   "ref": "main",
-  "rev": "ded6b00df2d7e411b1bba149a8d510c39d61a1c4",
+  "rev": "86efa8fa6d1e567011a2d10b9d547ac7c1669384",
   "submodules": true,
   "shallow": true
 }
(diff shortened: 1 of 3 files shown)


dominiklohmann (approved):

nix/tenzir/plugins/source.json
```diff
@@ -2,7 +2,8 @@
   "name": "tenzir-plugins",
   "url": "git@github.com:tenzir/tenzir-plugins",
   "ref": "main",
-  "rev": "ded6b00df2d7e411b1bba149a8d510c39d61a1c4",
+  "rev": "267b65fca17d3483b9c3ba3ed5ca80b103784e6c",
   "submodules": true,
-  "shallow": true
+  "shallow": true,
+  "allRefs": true
```
dominiklohmann, line 8: When rebasing this onto master, can I ask you to also pick up tenzir/tenzir-plugins#116 in addition to tenzir/tenzir-plugins#115?

tobim: I simply updated to https://github.com/tenzir/tenzir-plugins/pull/116. You can keep the cherry picked commit. I'll let the current CI run finish then we should be good to merge.

dominiklohmann: Great. I also verified manually that (1) the reconnect messages work as expected and that (2) the handshake is no longer being denied.

PR #3552, opened 2023-10-04
Disable dense indexes

dominiklohmann:
A recent release already disabled string indexes, which were using memory very inefficiently. This now generally disables all dense indexes, which have lost most of their appeal with columnar expression evaluation becoming more and more efficient.

The recent addition of the null type and empty records forced us to take another look at dense indexes, and we internally made the unbureaucratic decision not to want to maintain them any longer.

Fixes tenzir/issues#917


mavam (approved):
Less is more!

mavam (approved):

PR #3555, opened 2023-10-08
Beef up pipeline docs

mavam:
This PR adds an FAQs page to the docs and rewrites the description about
pipelines.


web/docs/faqs.md
```diff
@@ -0,0 +1,204 @@
+# FAQs
+
+This page provides answers to frequently asked questions (FAQs) about Tenzir.
+
+## What is Tenzir?
+
+Tenzir is a platform that pioneers open source security data pipelines.
```
dominiklohmann, line 7:
```suggestion
Tenzir is a platform that pioneers open-source security data pipelines.
```

web/docs/faqs.md
```diff
@@ -0,0 +1,204 @@
...
+infrastructure-as-code practices and we are performing security audits to ensure
+that our code quality meets the highest standards.
+
+We have plans to make this a single, end-to-end encrypted channel, so that we
+no longer have the theoretical ability to interpose on the data transfer between
+app and node.
+
+## Does Tenzir run on-premise?
+
+Our current version of the platform runs in the public cloud (AWS in Europe),
```
dominiklohmann, line 60:
```suggestion
Our current version of the platform runs in the public cloud (AWS in US),
```

@rdettai or @lava to confirm.

web/docs/pipelines.md
```diff
@@ -0,0 +1,216 @@
...
+
+A Tenzir **pipeline** is a chain of **operators** that represents a dataflow.
+Operators are the atomic building blocks that produce, transform, or consume
+data. Think of them as UNIX or Powershell commands where output from one command
+is input to the next:
+
+![Pipeline Chaining](pipeline-chaining.excalidraw.svg)
+
+There exist three types of operators: **sources** that produce data, **sinks**
+that consume data, and transformations that do both:
```
dominiklohmann, line 15:
```suggestion
that consume data, and **transformations** that do both:
```

web/docs/pipelines.md
```diff
@@ -0,0 +1,216 @@
...
+When composing pipelines out of operators, the type of adjacent operators have
+to match. Otherwise the pipeline is malformed. Here's an example pipeline with
+matching operators:
+
+![Typed Pipeline](typed-pipeline.excalidraw.svg)
+
+We call any void-to-void operator sequence a **closed pipeline**. Only closed
+pipelines can execute. If a pipeline does not have a source and sink, it would
+"leak" data. If a pipeline is open, the engine auto-completes a source/sink when
+possible or rejects it.
```
dominiklohmann, line 36: We should consider mentioning here that the default source/sink depends on how you use Tenzir. In the app, we automatically show the results in the explorer if the sink is missing. On the command-line, we read/write JSON from/to stdin.

web/docs/pipelines.md
```diff
@@ -0,0 +1,216 @@
...
+
+Zooming out in the above type table makes the operator types apparent:
+
+![Operator Types](operator-types.excalidraw.svg)
+
+In fact, we can define the operator type as a function of its input and output types:
+
+```
```
dominiklohmann, line 44: Not sure if Docusaurus supports it, but GitHub Markdown recently added support for math blocks by setting the language of the code block to `math`.

```suggestion
```math
```

web/docs/pipelines.md
```diff
@@ -0,0 +1,216 @@
...
+simplicity, at the cost of performance. In the spectrum of performance and ease
+of use, Tenzir therefore [fills a gap](/why-tenzir):
+
+![Structured vs. Document-Oriented](structured-vs-document-oriented.excalidraw.svg)
+
+:::info Eclectic & Super-structured Data
+[Zed](https://amyousterhout.com/papers/zed_cidr23.pdf) has a type system similar
+to Tenzir, with the difference that Zed associates types *with every single
+value*. Unlike Zed, Tenzir uses a "data frame" abstraction and relies on
+homogeneous Arrow record batches of up to 65,535 rows.
```
dominiklohmann, line 68: Mostly an FYI, not sure if we need to rewrite this: The limit of $`2^{16}`$ is arbitrarily chosen by us. Some operators can go above it, actually. With the plans for dynamic rebatching as an under-the-hood optimization we actually should go well above it in some cases.

web/docs/pipelines.md
```diff
@@ -0,0 +1,216 @@
...
+:::
+
+If the schema in a pipeline changes, we simply create new batch of events. The
```
dominiklohmann, line 71:
```suggestion
If the schema in a pipeline changes, we simply create a new batch of events. The
```

web/docs/pipelines.md
```diff
@@ -0,0 +1,216 @@
...
+worst case for Tenzir is a ordered stream of schema-switching events, with every
+event having a new schema than the previous one. That said, even for those data
+streams we can efficiently build homogeneous batches when the inter-event order
+does not matter significantly. Similar to predicate pushdown, Tenzir operators
+support "orderness pushdown" to signal to upstream operators that the event
```
dominiklohmann, line 76:
```suggestion
support "ordering pushdown" to signal to upstream operators that the event
```

At least that's how we call it internally.

web/docs/pipelines.md
```diff
@@ -0,0 +1,216 @@
...
+order only matters intra-schema but not inter-schema. In this case we
+transparently demultiplex a heterogeneous stream into *N* homogeneous streams,
+each of which yields batches of up to 65k events. The `import` operator is an
+example of such an operator, and it pushes the its orderness upstream so that we
```
dominiklohmann, line 80:
```suggestion
example of such an operator, and it pushes its ordering expectations upstream so that we
```

web/docs/pipelines.md
```diff
@@ -0,0 +1,216 @@
...
+can efficiently parse, say, a diverse stream of NDJSON records, such as
+Suricata's EVE JSON or Zeek's streaming JSON.
+
+You could call multi-schema dataflows *multiplexed* and there exist dedicated
+operators to demultiplex a stream. As of now, this is hard-coded per operator.
+For example, the [`directory`](/connectors/directory) operator demultiplexes a
```
dominiklohmann, line 86: `directory` is a saver, not an operator. This should probably be rewritten to say `to directory`.

web/docs/pipelines.md
```diff
@@ -0,0 +1,216 @@
...
+
+The diagram below illustrates the multi-schema aspect of dataflows for schemas
+A, B, and C:
+
+![Multi-schema Example](multi-schema-example.excalidraw.svg)
+
+Some operators only work with exactly one instance per schema internally, such
+as [`parquet`](/formats/parquet), [`feather`](/formats/parquet), or
+[`csv`](/formats/csv). These must be combined with a demultiplexing operator,
+such as the `directory` saver.
```
dominiklohmann, line 99: All three mentioned examples are formats, not operators. I would recommend rewriting like this:

```suggestion
Some operators only work with exactly one instance per schema internally, such
as the [`write`](/operators/transformations/write) when combined with the
[`parquet`](/formats/parquet), [`feather`](/formats/feather), or
[`csv`](/formats/csv) formats. These formats cannot handle multiple input
schemas at once. A demultiplexing operator like `to directory write parquet`
removes this limitation by writing one file per schema instead.
```

web/docs/pipelines.md
```diff
@@ -0,0 +1,216 @@
...
+
+We are having ideas to make this schema (de)multiplexing explicit with a
+`per-schema` [operator modifier](/operators/modifier) that you can write in
+front of every operator. Similarly, we are going to add union types in the
+future, making it possible to convert a heterogeneous stream of structured data
+into a homogeneous one.
```
dominiklohmann, line 105: Note that such a `fuse` operator would still be blocking, so it's not a good solution to this problem for long-running pipelines.

dominiklohmann (approved):
This reads nicely overall, I especially like the new images. I left some comments inline. Feel free to merge after addressing them.

web/docs/faqs.md
rdettai: no it's still europe, what you saw earlier was just to run the POC along the existing infra

dominiklohmann: Okay, then I'll just resolve this. Thanks.

web/docs/pipelines.md
mavam: There is [support for Math equations](https://docusaurus.io/docs/3.0.0-beta.0/markdown-features/math-equations), but it requires setting up KaTeX. So I'd say we do this at some point in the future.

web/docs/pipelines.md
mavam: Changed to:

> If a pipeline is open, the engine auto-completes a source/sink when
possible and rejects the pipeline otherwise. Auto-completion is context
dependent: on the command line we read JSON from stdin and write it stdout. In
the app we only auto-complete a missing sink with
[`serve`](operators/sinks/serve.md) to display the result in the browser.

web/docs/pipelines.md
mavam: I made this `to directory /tmp/dir write parquet`.

web/docs/pipelines.md
mavam: Right. I think it really depend on the use case here and don't want to go into details.

web/docs/pipelines.md
dominiklohmann: The only limit enforced by Arrow is 2 GiB per array, and that's only because migrating to FlatBuffers64 would be a breaking change. There is basically no limit on record batches themselves, and neither is there a limit on our chunks or table slices. We usually cut at the latest after 64 Ki events, but that's more by convention than a technical necessity.

PR #3556, opened 2023-10-09
Add a `velociraptor` operator

mavam:
This PR adds a `velociraptor` operator to interact with a Velociraptor server.


mavam:
@tobim @dominiklohmann this works minus the Protobuf & gRPC code gen, where I'm currently experiencing a linker error. This needs mostly CMake attention before a larger review. Here are resources I used to get to the point I am in the CMakeLists.txt of the plugin:

- https://medium.com/3yourmind/generate-grpc-source-files-using-cmake-ec3acd246b45
- https://github.com/protocolbuffers/protobuf/blob/main/docs/cmake_protobuf_generate.md

I tested locally that compiling and linking the CMake-generated protobuf and gRPC code works. It's really only the link-time bug that I can't wrap my head around.

tobim:
I'll look into the linker error.

mavam:
@tobim I can confirm that your latest commit works on macOS. May I also ask you to take over the remaining work to get CI green?

For example, I'm not sure why the Debian CI fails.

mavam:
Looks like the Docker image deps are not (yet) installed?

```
#22 2.711 CMake Error at CMakeLists.txt:52 (_find_package):
#22 2.711   Could not find a package configuration file provided by "protobuf" with any
#22 2.711   of the following names:
#22 2.711 
#22 2.711     protobufConfig.cmake
#22 2.711     protobuf-config.cmake
#22 2.711 
```

tobim:
> For example, I'm not sure why the Debian CI fails.

That should be fixed by the rebase just now.

> Looks like the Docker image deps are not (yet) installed?

Turns out Debian ships an older protobuf version that doesn't have the CMake convenience tooling.

mavam:
Nice, almost there!

```
!! system_error: failed to load plugin /opt/tenzir/lib/tenzir/plugins/libtenzir-plugin-velociraptor.so libgrpc++.so.1.51: cannot open shared object file: No such file or directory
```

tobim:
All green now.

dominiklohmann (approved):

PR #3557, opened 2023-10-09
Drop workaround for oversized responses in serve

dominiklohmann:
This is no longer needed, as the platform can now correctly handle responses over 128 KiB in size.

### DIFF
--- a/libtenzir/builtins/operators/serve.cpp
+++ b/libtenzir/builtins/operators/serve.cpp
@@ -119,10 +119,6 @@ constexpr auto SPEC_V0 = R"_(
                   type: string
                   description: A token to access the next pipeline data batch, null if the pipeline is completed.
                   example: "340ce2j"
-                dropped:
-                  type: integer
-                  description: The serve endpoint will drop responses that exceed the configured maximum message size. The number of events dropped this way is returned here. Should be 0 most of the time.
-                  example: 0
                 schemas:
                   type: array
                   items:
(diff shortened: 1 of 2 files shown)


libtenzir/builtins/operators/serve.cpp
```diff
@@ -119,10 +119,6 @@ constexpr auto SPEC_V0 = R"_(
                   type: string
                   description: A token to access the next pipeline data batch, null if the pipeline is completed.
                   example: "340ce2j"
-                dropped:
-                  type: integer
-                  description: The serve endpoint will drop responses that exceed the configured maximum message size. The number of events dropped this way is returned here. Should be 0 most of the time.
-                  example: 0
```
lava, line 125: I think we should be careful about backwards-incompatible changes like this, I'd propose leaving the fields, but removing the mention of message size from the description and having it always set to zero. Then we can also repurpose it if we need to drop messages for some other reason in the future.

lava (approved):

libtenzir/builtins/operators/serve.cpp
dominiklohmann: Did anyone ever use this? It existed only because of a temporary insufficiency of our platform that is now fixed.

dominiklohmann: A quick internal chat showed that we did not even make use of this internally: https://tenzir.slack.com/archives/C7QBE8WL9/p1696849415522809

I will just remove the field for now, if someone complains about it we can always add it back. Let's move fast and break things, or so.

PR #3560, opened 2023-10-09
Migrate the fluent-bit tests to bats

tobim:


plugins/fluent-bit/functional-test/reference/tests/test_read_stdin_via_fluent-2dbit/step_00.ref
mavam, line 1: Is this really the correct reference? It looks that this sorts the output. I don't think we want this here.

plugins/fluent-bit/functional-test/tests/setup_suite.bash
mavam, line 1: Is this something every plugin needs?

plugins/fluent-bit/functional-test/tests/tests.bats
```diff
@@ -0,0 +1,40 @@
...
+
+setup() {
+  bats_load_library bats-support
+  bats_load_library bats-assert
+  bats_load_library bats-tenzir
+}
+
+teardown() {
+  :
+}
```
mavam, line 13: Is this the common header?

mavam (approved):
I like that we have bats, but I also expected a bit more shell-style tests. For example:

```bash
check -c \
    "tenzir 'show operators | head | fluent-bit counter' | cut -d , -f 2"
```

This puts everything in a single string, which is quite opaque. Can't we run `tenzir` here first, and then perform a checking function?

plugins/fluent-bit/functional-test/reference/tests/test_read_stdin_via_fluent-2dbit/step_00.ref
dominiklohmann: Note that it was not the default for the old integration test suite either, that was only the case for commands starting with `tenzir-ctl export`. Anything going through `tenzir-ctl exec` was not sorted automatically.

plugins/fluent-bit/functional-test/tests/tests.bats
tobim: For some reason we can't load the libraries in the `setup_suite` file, but we might be able to extract and `source` it. I'll look into it.

plugins/fluent-bit/functional-test/reference/tests/test_read_stdin_via_fluent-2dbit/step_00.ref
tobim: `check` has a `--sort` flag now.

PR #3562, opened 2023-10-09
Fix `serve` exiting prematurely

dominiklohmann:
> [!IMPORTANT]
> We discussed internally that we want to merge this fix only after the Tenzir v4.3 release, as we want to have some time to double-check whether there are any other fallouts of the change.

Pipelines ending with `serve` should wait until all events are fetched from the corresponding `/serve` endpoint. This was not always the case because of an incorrectly set scope in the operator.

Fixing that scope, however, led to the operator deadlocking the node under some circumstances. This revealed a bug at the very core of CAF's scheduling logic, which incorrectly blocked not just ordinary but also system messages from arriving when a request response was awaited. I got that bug fixed upstream and backported the fix to our bundled fork of CAF 0.18. That in turn revealead us relying on the broken behavior in our execution nodes accidentally. I've included a fix for that as well.


libtenzir/src/execution_node.cpp
```diff
@@ -652,23 +652,21 @@ struct exec_node_state : inbound_state_mixin<Input>,
...
-      run();
-    };
     run_scheduled = true;
-    self->clock().schedule(self->clock().now(),
-                           caf::make_action(std::move(action),
-                                            caf::action::state::waiting),
-                           caf::weak_actor_ptr{self->ctrl()});
+    // It is critical that we send an _ordinary_ message to the scheduler here
+    // for the internal run loop, as opposed to a _system_ message, e.g., by
+    // scheduling an action directly on the clock. Otherwise, an operator using
```
lava, line 658: Tbf, I think scheduling directly on the clock is something that we will *never* want to do after the CAF updates, so it's probably not necessary to call this out in this particular part of the code.

However, the previous list of advantages still seems to be valid, so maybe we can leave them in as explanation.

libtenzir/builtins/operators/serve.cpp
```diff
@@ -759,39 +759,37 @@ class serve_operator final : public crtp_operator<serve_operator> {
...
-          },
-          [&](const caf::error& err) { //
-            ctrl.abort(caf::make_error(
-              ec::logic_error,
-              fmt::format("failed to register at serve-manager: {}", err)));
-          });
-      co_yield {};
-    }
+    // Step 1: Get a handle to the SERVE MANAGER actor.
+    auto blocking = caf::scoped_actor{ctrl.self().system()};
```
lava, line 763: Should we note here that the serve manager monitors `blocking`, so it has to be kept alive until the end of the function scope?

lava (approved):
Looks good to me from the logical perspective, and we did a lot of manual testing to verify that this is indeed the source of the error.

libtenzir/builtins/operators/serve.cpp
dominiklohmann: Good idea, done in b4f7863164.

libtenzir/src/execution_node.cpp
dominiklohmann: I removed the comment in 3e61cda7d0; in hindsight I think it adds more confusion than it helps.

PR #3569, opened 2023-10-10
Add default formatter with debug writer

jachris:
This PR adds a mechanism to automatically implement stringifcation based on the `inspect()` implementation. This removes the need for a `to_string()` implementation (which previously often contained a half-baked implementation of constructing a parseable represenation of the operator). We can also use this to derive `fmt::formatter` implementations when `inspect()` is available, which prevents them from going out of sync, as it leads to a "single source of truth". 

### Examples

~~~
$ tenzir --dump-ast "from a/b/c.json | where xyz == 123 | to foo.csv"
pipeline [
  load file {
    path: "a/b/c.json" @ 5..15
  },
  read json {
    unnest_separator: "",
    use_ndjson_mode: false,
    preserve_order: true,
    raw: false,
    arrays_of_objects: false
  }
] @ 0..15
where (xyz == 123 @ 24..35) @ 18..35
pipeline [
  write xsv {
    name: "csv",
    field_sep: 44,
    list_sep: 59,
    null_value: "",
    allow_comments: false
  },
  save file {
    path: "foo.csv" @ 40..47
  }
] @ 37..47

~~~

_(Note: This does not really show the AST, but already the intermediate representation we currently get directly after parsing. This is also why we can still see the nested pipelines produced by `from` and `csv`. The output seen here will likely change in the future)._

### DIFF
--- a/changelog/next/features/3723--show-udos.md
+++ b/changelog/next/features/3723--show-udos.md
@@ -1,0 +1,2 @@
+`show operators` now shows user-defined operators in addition to operators
+that ship with Tenzir or as plugins.
--- a/libtenzir/builtins/aspects/config.cpp
+++ b/libtenzir/builtins/aspects/config.cpp
@@ -34,7 +34,7 @@ public:
   }
 
   auto location() const -> operator_location override {
-    return operator_location::local;
+    return operator_location::anywhere;
   }
 
   auto show(operator_control_plane&) const -> generator<table_slice> override {
--- a/libtenzir/builtins/aspects/operators.cpp
+++ b/libtenzir/builtins/aspects/operators.cpp
@@ -8,27 +8,21 @@
 
 #include <tenzir/argument_parser.hpp>
 #include <tenzir/plugin.hpp>
-#include <tenzir/table_slice_builder.hpp>
+#include <tenzir/series_builder.hpp>
 
 namespace tenzir::plugins::operators {
 
 namespace {
 
-/// A type that represents an operator.
-auto operator_type() -> type {
-  return type{
-    "tenzir.operator",
-    record_type{
-      {"name", string_type{}},
-      {"source", bool_type{}},
-      {"transformation", bool_type{}},
-      {"sink", bool_type{}},
-    },
-  };
-}
-
 class plugin final : public virtual aspect_plugin {
 public:
+  auto initialize(const record& plugin_config, const record& global_config)
+    -> caf::error override {
+    (void)plugin_config;
+    udos_ = get_or(global_config, "tenzir.operators", udos_);
+    return {};
+  }
+
   auto name() const -> std::string override {
     return "operators";
   }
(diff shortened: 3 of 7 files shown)


libtenzir/include/tenzir/detail/default_formatter.hpp
```diff
@@ -0,0 +1,155 @@
...
+        return ++it;
+      } else if (*it == '#') {
+        auto next = std::next(it);
+        if (next != ctx.end() && *next == '?') {
+          mode = pretty_debug;
+          return ++next;
+        }
+      }
+    }
+    // TODO: Do we want make `{}` use `inspect` if there is no `to_string`?
```
jachris, line 118: This question basically boils down to whether we want to print the debug representation when using `{}` with an object that does not have `to_string()`, or whether we want to reject this at compile-time. The advantage of rejecting it is that we do not return debug representations on accident.

tenzir/integration/reference/parse-operators/step_04.ref
```diff
@@ -1,3 +1,12 @@
...
+    path: "a/b/c.json" @ 10..20
+  },
+  read json {
+    unnest_separator: "",
+    use_ndjson_mode: false,
+    preserve_order: true,
+    raw: false,
+    arrays_of_objects: false
+  }
+] @ 0..20
```
eliaskosunen, line 12: This is so much cleaner!

libtenzir/src/tql/parser.cpp
```diff
@@ -620,8 +620,8 @@ auto parse(std::string source, diagnostic_handler& diag)
 auto parse_internal(std::string source) -> caf::expected<pipeline> {
   auto [pipe, diags] = parse_internal_with_diags(std::move(source));
   if (not pipe) {
-    return caf::make_error(ec::parse_error,
-                           fmt::format("could not parse pipeline: {}", diags));
+    return caf::make_error(
+      ec::parse_error, fmt::format("could not parse pipeline: {::?}", diags));
```
eliaskosunen, line 624:
```suggestion
      ec::parse_error, fmt::format("could not parse pipeline: {:?}", diags));
```

unless I'm missing something

eliaskosunen (approved):

libtenzir/src/tql/parser.cpp
jachris: `{:?}` would apply the `?` to `std::vector`, which doesn't know what to do with it. If we use `::?`, then `std::vector` gets `:?`, which is defined to trigger the formatter of the element type with `?`.

libtenzir/include/tenzir/detail/default_formatter.hpp
```diff
@@ -0,0 +1,155 @@
...
+constexpr bool can_stringify
+  = not std::is_same_v<decltype(tenzir::detail::stringify(
+                         std::declval<fmt::format_context::iterator>(),
+                         std::declval<T>())),
+                       void>;
+
+template <class T>
+constexpr bool can_inspect = not std::is_same_v<
+  decltype(caf::inspect_access_type<tenzir::debug_writer, T>()),
+  caf::inspector_access_type::none>;
```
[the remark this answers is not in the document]
jachris: Good point.

libtenzir/include/tenzir/detail/default_formatter.hpp
```diff
@@ -0,0 +1,155 @@
...
+/// The "default formatter" provides an `fmt::formatter` for types that might
+/// have `to_string` functionality and those that can used with CAF's `inspect`.
+/// Depending on the availability of both, the default formatter supports the
+/// format strings `{}` (for user-friendly stringification), `{:?}` (for concise
+/// debug information), and `{:#?}` (for multi-line debug information).
+///
+/// To enable the default formatter for a type `T`, supply a specialization of
+/// `enable_default_formatter` with a `::value` member that is `true`.
+template <class T>
+struct enable_default_formatter : std::false_type {};
```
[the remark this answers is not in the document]
jachris: You would still need this type for specialization purposes, no? We could do something like `enable_default_formatter_v`, but the `::value` is only used in one place, so I don't think this matters for our case here (but in general, I agree).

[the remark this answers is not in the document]
eliaskosunen: Nothing's preventing you from specializing the variable template. Hence the parallel with `enable_borrowed_range`.

```
template <class T>
inline constexpr bool enable_default_formatter = false;
// ...
template <class T>
inline constexpr bool enable_default_formatter<use_default_formatter<T>> = true;
// ...
template <>
inline constexpr bool enable_default_formatter<diagnostic_annotation> = true;
// etc.
```

PR #3570, opened 2023-10-11
Detect and report incomplete object in JSON parser

jachris:
Detect and report incomplete JSON objects as an error, for example:

~~~
{"a": 42}
{"b":
~~~


dominiklohmann (approved):
I think you could add a bug fix changelog entry for this as well. Makes a lot of sense to have like this.

PR #3572, opened 2023-10-14
Fix download link to static binary from main

mavam:
This PR fixes the download link for static binaries from the `main` branch.


tobim (approved):
Works!

PR #3573, opened 2023-10-15
Add example and context to Velociraptor blog post

lo-chr:
This PR adds an example to the Velociraptor blog post and adds some context to the description of the "subscribe" use case.
Additionally it removes references to wizardry (🧙).

PR #3574, opened 2023-10-15
Complete Velociraptor blog post and add a few fix-ups

mavam:
This PR is basically #3573, but unfortunately we currently can't run CI from external contributors.


dominiklohmann (approved):
LGTM 🫣

PR #3576, opened 2023-10-16
Omit null fields by default and add `--emit-null-fields`

jachris:
This PR makes it so that `write json` does not print fields with `null` values anymore, at least by default. To get the old behavior, use the new `--emit-null-fields` option. Also, the option `--omit-nulls`, which previously omitted both null fields and null items in lists, has been renamed to `--omit-null-items`. The rationale behind splitting this it that omitting `null` fields is more common and well-behaved than omitting `null` values in lists.

### Alternative

We could also print null fields by default, and add `--omit-null-fields` instead (which would still be nice to have). Please let me know if you have an opinion on this. The advantage of omitting null fields by default is that we often insert `null` values even if a field is not present in an object (but was there previously).

#### Input
~~~json
{"a": 42, "b": null}
{"c": [null]}
~~~

#### Before
~~~json
{"a": 42, "b": null, "c": null}
{"a": null, "b": null, "c": [null]}
~~~

#### After
~~~json
{"a": 42}
{"c": [null]}
~~~

### Todo

- [ ] Make sure that this is actually what we want.
- [ ] Rename config option.
- [ ] Consider providing deprecated compatibility aliases.


mavam:
@satta what's your take on this?

satta:
> @satta what's your take on this?

Agreed for fields and lists; the thing that I'd be more interested in is this case:

```json
{
  "a": 42,
  "b": {
    "c": null,
    "d": 23,
    "e": {
      "f": {
        "g": null
      },
      "h": null
    }
  }
}
```
where I'd expect the result to be:
```json
{
  "a": 42,
  "b": {
    "d": 23
  }
}
```
i.e. all fields whose values are sub-objects that only contain either only `null` values or further sub-objects with only `null` values should be skipped in the output.
This is important for schemas with a large potential set of fields, but which are only sparsely populated with actual values. In short, a transitive behaviour across multiple levels of nesting.

I think we once discussed that and it led to the `--omit-empty-records` option. Is that also going to become the default?

jachris:
Good points, thank you for sharing. 

For me, one of the reasons why omitting nulls by default makes a lot of sense is the following: If I know what overall shape my event has, they are truly unnecessary. For example, if my events normally look like `{"foo": {"bar": <int>}}`, and I get `{}`, I know that the original data is `{"foo": null}`. There is no other option. However, if we also omit empty records, then the original data could have been `{"foo": null}`, but also `{"foo": {"bar": null}}`. This goes even further when omitting empty lists, as it is now unclear whether the list was present at all, or whether it contained only nulls, or perhaps it contained some nulls and some empty objects. We are no longer able to tell. 

For some applications, these distinctions do not matter, so I can see why something like `--omit-empty-objects` (or even `--omit-empty`) makes sense. But I'd prefer it to not be the default.

Also, just as a small note: In previous versions of Tenzir, it could occasionally happen that a `null` value of record type was transformed to a record of nulls for every field. This should no longer be the case, so I hope that you would get the situation of records with many nested null fields not that often anymore (only if the input looks the same way).

dominiklohmann:
I'm a bit conflicted on this. On the one hand, it can be confusing to see nulls in data that did not have nulls in it before the import. We mostly want to treat nulls and absent values the same way. However, for other formats, we have no option like this. I see four options:
1. Remove nulls by default.
2. Remove nulls, empty lists, and empty records by default.
3. Remove nulls by default, and remove empty records created by implicitly removed nulls.
4. Keep things as they are.

We may also want to consider an operator that removes nulls or empty lists or empty records, although that can be pretty expensive as this tends to make batching a lot worse, so using it early on in a pipeline can cause performance to suffer a lot.

satta:
> For me, one of the reasons why omitting nulls by default makes a lot of sense is the following: If I know what overall shape my event has, they are truly unnecessary.

I was completely assuming that there is a schema for each event to be ingested that describes the full superset of all fields that could ever be in that type, regardless of they're null in the data seen by the node or not. In fact, that's why the omit-nulls feature is so important for me: For some schemas (e.g. ECS based, see https://gist.github.com/satta/78ab02e1986b10ccb1bfb1e88c955fb7 autogenerated from https://github.com/elastic/ecs/blob/main/generated/ecs/ecs_nested.yml), a lot of fields are allowed for a type, but can only be populated if the data is actually gathered in that level of detail. In the case of Packetbeat, for example, it is not -- no user or machine information is present, hence these fields will always be null.

jachris:
We won't do this for now.

PR #3579, opened 2023-10-17
Add an optional content type to chunk

dominiklohmann:
This addition allows for transferring metadata about chunks in-stream. Specifically, this adds a metadata field for the content type, which is the missing piece for tenzir/tenzir#3539.

The alternative to this approach is to forbid `write <format> | save <connector>` in favor of always requiring the combined `to <connector> write <format>`. However, the latter does not compose as freely as transferring the content type in-stream.

What's left is to make use of the content type, but I will leave that to the linked PR adding HTTP(S) savers.

### DIFF
--- a/libtenzir/include/tenzir/chunk.hpp
+++ b/libtenzir/include/tenzir/chunk.hpp
@@ -27,6 +27,16 @@
 
 namespace tenzir {
 
+struct chunk_metadata {
+  std::optional<std::string> content_type = {};
+
+  friend auto inspect(auto& f, chunk_metadata& x) -> bool {
+    return f.object(x)
+      .pretty_name("tenzir.chunk_metadata")
+      .fields(f.field("content_type", x.content_type));
+  }
+};
+
 /// A reference-counted contiguous block of memory. A chunk supports custom
 /// deleters for custom deallocations when the last instance goes out of scope.
 class chunk final : public caf::ref_counted {
@@ -58,15 +68,19 @@ public:
   /// @param data The raw byte data.
   /// @param size The number of bytes *data* points to.
   /// @param deleter The function to delete the data.
+  /// @param metadata The metadata associated with the chunk.
   /// @returns A chunk pointer or `nullptr` on failure.
   static chunk_ptr
-  make(const void* data, size_type size, deleter_type&& deleter) noexcept;
+  make(const void* data, size_type size, deleter_type&& deleter,
+       chunk_metadata metadata = {}) noexcept;
 
   /// Constructs a chunk of particular size and pointer to data.
   /// @param view The span holding the raw data.
   /// @param deleter The function to delete the data.
+  /// @param metadata The metadata associated with the chunk.
   /// @returns A chunk pointer or `nullptr` on failure.
-  static chunk_ptr make(view_type view, deleter_type&& deleter) noexcept;
+  static chunk_ptr make(view_type view, deleter_type&& deleter,
+                        chunk_metadata metadata = {}) noexcept;
 
   /// Constructs an empty chunk
   static chunk_ptr make_empty() noexcept;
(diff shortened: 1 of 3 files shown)


tobim (commented):
I'd argue that this should even be mandatory instead of optional if we want to use it for essential operators like `to`.

mavam:
> I'd argue that this should even be mandatory instead of optional if we want to use it for essential operators like `to`.

But we are using chunks for various other purposes. It's not needed in many of these other use cases. Forcing every chunk instance to provide a content type feels too much. The optionality strikes me as the best choice, especially when we want to add more metadata in the future.

tobim:
How do you guarantee that a chunk that is "used for a different" purpose doesn't and up in the data stream somehow? A dedicated type would be one way, but that also means that we'd have to lift chunks into the data stream at some place, and you might not know the content type there. On the other hand we should know the type where the chunk is created, thanks to it's immutable nature.

mavam:
> How do you guarantee that a chunk that is "used for a different" purpose doesn't and up in the data stream somehow?

We mmap files in the storage engine with chunks as well, for example. Chunks are general-purpose mechanism, not just used to transport bytes between operators.

Lifting a `chunk` into something like this could also make sense for the element type in pipelines:

```cpp
struct bytes {
  metadata ...;
  chunk_ptr chunk;
};
```

But that'd be a rather big change.

tobim:
> We mmap files in the storage engine with chunks as well, for example. Chunks are general-purpose mechanism, not just used to transport bytes between operators.

Every chunk has content, so why not also a content-type? It doesn't matter how it was created.

Besides, `string` has a built-in null-like state representation, so what use is the wrapping in an extra optional? All places that use the value will need some kind of "default" path, the `optional` just adds one more way to reach that.

mavam:
> Every chunk has content, so why not also a content-type? It doesn't matter how it was created.

As long as the content type can be empty, I have no objections. But forcing users to add a string if they really just want a buffer of bytes feels too strict.

> Besides, `string` has a built-in null-like state representation, so what use is the wrapping in an extra optional? All places that use the value will need some kind of "default" path, the `optional` just adds one more way to reach that.

Agreed, I think an empty string can be considered equivalent to "no content type". Stripping the `optional` seems reasonable.

jachris:
After https://github.com/tenzir/tenzir/pull/3539 is changed to _not_ stream byte batches anymore (at least by default), I'm not sure whether this is still the correct mechanism, as the content type must stay constant then. Also, I think we should try to not make byte streams a leaky abstraction. It also breaks down the moment somebody does `write json | shell "jq ..." | save http ...`. But I am fine with this if we acknowledge that this is potentially only a temporary solution, as I can see that it can be useful.  

---

And joining the `optional` discussion: In general, an empty string is of course not the same as no string, and there might be different behavior between having the associated header with an empty value and not having it. Also, it enforces that the recipient of the content type acknowledges that it can be absent, which is easily lost with an empty string (so from a developer POV, these two are _not_ equivalent). That being said, a set-but-empty content type might not make sense, so do whatever you prefer.

tobim:
> After https://github.com/tenzir/tenzir/pull/3539 is changed to not stream byte batches anymore (at least by default), I'm not sure whether this is still the correct mechanism, as the content type must stay constant then.

I think it's the same no matter if it's chunked encoding or not. Either way we will only set the content type once. In case `save http` gets different content types it should probably produce an error.

> That being said, a set-but-empty content type might not make sense, so do whatever you prefer.

Indeed, an empty content type would be ill-formed: https://www.w3.org/Protocols/rfc1341/4_Content-Type.html.

My argument for making the member non-optional is that it adds incentive to set it to a sensible value.

tobim (approved):
In further discussion with @dominiklohmann we considered using "application/octet-stream" as a default instead of the empty string, but decided against it.

We also discussed future additions to the metadata that may actually require the optional wrapping. While thinking about that I changed my mind about using optional, because that will guarantee a more consistent API into the metadata object in the future.

mavam:
@tobim great, does that mean merge as is? What is left to do before this can ship?

PR #3580, opened 2023-10-17
Add `events` field to output of `show partitions`

dominiklohmann:
One simple use case for this is wanting to see how many events are stored at the node per schema.

```
show partitions
| summarize events=sum(events) by schema
| sort events desc
```

### DIFF
--- a/libtenzir/builtins/aspects/partitions.cpp
+++ b/libtenzir/builtins/aspects/partitions.cpp
@@ -7,34 +7,19 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 #include <tenzir/actors.hpp>
-#include <tenzir/adaptive_table_slice_builder.hpp>
 #include <tenzir/argument_parser.hpp>
 #include <tenzir/catalog.hpp>
 #include <tenzir/node_control.hpp>
 #include <tenzir/partition_synopsis.hpp>
 #include <tenzir/plugin.hpp>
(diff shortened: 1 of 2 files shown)


libtenzir/builtins/aspects/partitions.cpp
```diff
@@ -76,29 +61,32 @@ class plugin final : public virtual aspect_plugin {
...
+      event.field("version").data(synopsis.synopsis->version);
+      event.field("schema").data(synopsis.synopsis->schema.name());
+      event.field("schema_id")
+        .data(synopsis.synopsis->schema.make_fingerprint());
       if ((i + 1) % max_rows == 0) {
-        co_yield builder.finish();
-        builder = std::exchange(builder, table_slice_builder{partition_type()});
+        for (auto result : builder.finish_as_table_slice("tenzir.partition")) {
+          co_yield std::move(result);
+        }
```
jachris, line 82: Nitpick: `for (auto& ...)` is one less copy, and same below (although it's cheap anyway).

jachris (approved):

libtenzir/builtins/aspects/partitions.cpp
jachris: Exactly, that's what we want to do. But if we leave it as-is, then we copy them first and _then_ move them out, right?

PR #3581, opened 2023-10-17
Add `blob` type for arbitrary binary data

jachris:
We currently use the `string` type both for printable strings and for arbitrary binary data. However, this violates Arrow's UTF-8 requirements and leads to several other issues down the road (for example: it's not possible to print arbitrary blobs as JSON strings). Thus, this PR introduces a new `blob` type, that is designed to handle arbitrary binary data.

Internally, this binary data is directly stored as an arbitrary sequence of bytes (using Arrow's `BinaryType` for table slices). However, when the data enters or leaves Tenzir, we must have some special handling, depending on the format. For example, assume we want to print the blob `b"\x00\xFF"` as JSON. This blob cannot be represented as an escaped string. Thus, we base64 encode it, and print `"AP8="`. When parsing JSON, we cannot assume that all base64-valid strings are binary payloads. However, when a schema with a `blob` type is given, we base64 decode the blob to get `b"\x00\xFF"` back. Other formats might use different strategies to encode and decode blobs.


libtenzir/include/tenzir/concept/printable/tenzir/view.hpp
```diff
@@ -42,6 +44,19 @@ struct string_view_printer : printer_base<string_view_printer> {
...
+
+  template <class Iterator>
+  bool print(Iterator& out, const attribute& x) const {
+    // TODO: Check that this works as expected.
+    static auto escaper = detail::make_extra_print_escaper("\"");
+    static auto p = "b\"" << printers::escape(escaper) << '"';
+    return p(out, std::string_view{reinterpret_cast<const char*>(x.data()),
+                                   x.size()});
+  }
+};
```
mavam, line 58: Just throwing in a thought to consider: most tools I know use Base64 encoding when displaying binary data in strings. (We have routines for that in the code base already.)

jachris: When the data actually leaves our systems (JSON, YAML, ...), the current state of the PR already prints it as Base64. This is for internal use, where it makes sense to somehow communicate that this is not a string, but a blob (currently printing e.g. `b"hello\x00\xFF"`). Also, we probably want to have a literal for blobs later (we do not want automagically parse base64-valid strings as blobs), so they need a dedicated representation anyway (again, e.g. `b"hello\x00\xFF"`)

mavam: Sorry, I missed the right commenting location for this. Glad to see you've already went down that path!

Indeed, binary literals would be a great addition (for later).

schema/types/suricata.schema
```diff
@@ -76,10 +76,10 @@ type suricata.alert = suricata.component.common + record {
     }
   },
   flow: suricata.component.flow,
-  payload: string,
+  payload: blob,
   payload_printable: string,
   stream: uint64,
-  packet: string,
+  packet: blob,
```
jachris, line 82: I probably missed a bunch of these, but I wanted to test that it works. Also, do we actually already want to change the schemas or do we wait with that? It can break queries such as `where "abc" in payload`.

dominiklohmann: My hunch: Yes, just make the change.

@mavam can you confirm?

mavam: Confirming, I'm not aware of anyone who does `"abc" in payload` on binary data.

tenzir/integration/reference/node-suricata-alert/step_01.ref
```diff
@@ -1,4 +1,4 @@
...
+<2011-08-12T12:52:57.716360, 1031464864740687, 83, null, null, 147.32.84.165, 1181, 78.40.125.4, 6667, "TCP", "alert", null, null, "allowed", 1, 2017318, 4, "ET CURRENT_EVENTS SUSPICIOUS IRC - PRIVMSG *.(exe|tar|tgz|zip)  download command", "Potentially Bad Traffic", 2, null, null, null, null, 27, 35, 2302, 4520, 2011-08-12T12:47:24.357711, null, null, null, null, null, b"PRIVMSG #zarasa48 : smss.exe (368)\x0D\x0A", "PRIVMSG #zarasa48 : smss.exe (368)\x0D\x0A", 0, b"\x00\x1EI\xDB\x19\xC3\x08\x00'\xB5\xB7\x19\x08\x00E\x00\x00L\x19^@\x00\x80\x06.\\x93 T\xA5N(}\x04\x04\x9D\x1A\x0B\xF0\xD1\x0B\xC0\xC5E\x87uP\x18\xFA\xB0\x8B\xE2\x00\x00PRIVMSG #zarasa48 : smss.exe (368)\x0D\x0A", 1>
```
jachris, line 1: Not sure if this is what want, but we probably do not care too much. This is from `export ascii`.

dominiklohmann: We never ported that internal format to our new parser/printer framework (and don't have any plans to), so it's not worth paying much attention to this. I think this is okay for now.

schema/types/suricata.schema
jachris: Then let's keep it. If I missed any `string` fields that should really be `blob`, we can update it as we go.

libtenzir/builtins/formats/pcap.cpp
```diff
@@ -437,21 +436,19 @@ auto to_packet_record(auto row) -> std::pair<packet_record, uint32_t> {
       pkt.header.captured_packet_length = caf::get<uint64_t>(value);
     else if (key == "original_packet_length")
       pkt.header.original_packet_length = caf::get<uint64_t>(value);
-    else if (key == "data")
-      pkt_data = caf::get<std::string_view>(value);
-    else
+    else if (key == "data") {
+      pkt.data = caf::get<view<blob>>(value);
```
dominiklohmann, line 440: The potential type confusion here can now easily be run into when users store packets and write them here.

I think we basically must use `get_if` here already.

libtenzir/builtins/formats/zeek_tsv.cpp
```diff
@@ -391,6 +394,34 @@ struct zeek_printer {
       return true;
     }
 
+    auto operator()(view<blob> x) noexcept -> bool {
+      if (x.empty()) {
+        // TODO: Is this actually correct? An empty blob is not unset.
```
dominiklohmann, line 399: Zeek has placeholder text for empty and unset values. You want to print the empty value placeholder here instead.

libtenzir/include/tenzir/arrow_table_slice.hpp
```diff
@@ -292,6 +295,9 @@ value_at(const Type& type, const std::same_as<arrow::Array> auto& arr,
 
 data_view value_at(const type& type, const std::same_as<arrow::Array> auto& arr,
                    int64_t row) noexcept {
+  TENZIR_WARN("type = {}, id = {}, arr_id = {}", type,
+              static_cast<int>(type.to_arrow_type()->id()),
+              static_cast<int>(arr.type_id()));
```
dominiklohmann, line 300: This should be removed before merging.

libtenzir/include/tenzir/concept/printable/tenzir/blob.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+namespace tenzir {
+
+template <class T>
+struct generic_blob_printer : printer_base<generic_blob_printer<T>> {
+  using attribute = T;
+
+  template <class Iterator>
+  bool print(Iterator& out, const attribute& x) const {
+    // TODO: Check that this works as expected.
```
dominiklohmann, line 23: The TODO can be removed, it does.

libtenzir/include/tenzir/concept/printable/tenzir/blob.hpp
```diff
@@ -0,0 +1,31 @@
...
+#pragma once
+
+#include "tenzir/concept/printable/core.hpp"
+#include "tenzir/concept/printable/string/escape.hpp"
+#include "tenzir/detail/escapers.hpp"
+
+namespace tenzir {
+
+template <class T>
+struct generic_blob_printer : printer_base<generic_blob_printer<T>> {
```
dominiklohmann, line 18: Does a `data` containing a `blob` survive a print-parse roundtrip?

libtenzir/include/tenzir/concept/printable/tenzir/json.hpp
```diff
@@ -87,6 +88,11 @@ struct json_printer : printer_base<json_printer> {
       return true;
     }
 
+    auto operator()(view<blob> x) noexcept -> bool {
+      // TODO: Do we want to prefix the encoded string, e.g., `base64:...`?
```
dominiklohmann, line 92: I don't think so, at least not for now.

libtenzir/include/tenzir/detail/base64.hpp
```diff
@@ -61,6 +61,7 @@ size_t encode(void* dst, const void* src, size_t len);
 /// @param str The string to encode.
 /// @returns The Base64-encoded version of *str*.
 std::string encode(std::string_view str);
+std::string encode(std::basic_string_view<std::byte> str);
```
dominiklohmann, line 64: Shouldn't this be a `std::span<const std::byte>` for the argument? I don't think `std::char_traits<std::byte>` is well-defined.

libtenzir/include/tenzir/type.hpp
```diff
@@ -1340,6 +1340,35 @@ class record_type final : public stateful_type_base {
...
+// -- blob_type --------------------------------------------------------------
+
+/// An sequence of bytes.
+/// @relates type
+class blob_type final {
+public:
+  /// Returns the type index.
+  static constexpr uint8_t type_index = 16;
+
+  // TODO: Or LargeBinaryType?
```
dominiklohmann, line 1352: Regular is fine I think, at least for now.

libtenzir/include/tenzir/type.hpp
```diff
@@ -1340,6 +1340,35 @@ class record_type final : public stateful_type_base {
...
+  /// The corresponding Arrow DataType.
+  using arrow_type = arrow::BinaryType;
+
+  /// Returns a view of the underlying binary representation.
+  friend std::span<const std::byte> as_bytes(const blob_type&) noexcept;
+
+  // TODO: Or chunk_ptr?
```
dominiklohmann, line 1359: The advantage of using `chunk_ptr` for blobs is that we don't need to copy (as we can slice with a shared lifetime), and that it can be both view and data at the same time. However, I think it has the precondition to remove `chunk_ptr` and to always make `chunk` have shared pointer semantics, and I don't think that's a good use of our time right now.

libtenzir/src/format/zeek.cpp
```diff
@@ -534,6 +537,25 @@ class zeek_printer : public printer_base<zeek_printer<Iterator>> {
     return true;
   }
 
+  bool operator()(Iterator& out, const view<blob>& x) const {
```
dominiklohmann, line 540: So much of the work in this PR is for the old formats; at some point we should really remove them. I've reviewed these parts only very broadly.

tenzir/integration/integration.py
```diff
@@ -258,7 +258,7 @@ def run_step(
                 baseline_dir.mkdir(parents=True)
         else:
             LOGGER.debug("comparing step output to baseline")
-        with open(stdout) as out_handle:
+        with open(stdout, "rb") as out_handle:
```
dominiklohmann, line 261: Does this also need to be adjusted for the bats-based "functional tests?"

dominiklohmann (commented):
This is really nice, thanks for hacking this together so quickly. I have some questions on the code and some answers to questions you left in the code. I'll play around with this in practice some more to test it further and will then approve it afterwards.

libtenzir/builtins/formats/pcap.cpp
jachris: I agree (see https://github.com/tenzir/issues/issues/946). I can fix this with a fallback to `string` for this special case and ignoring the data otherwise, and would leave deeper refactorings for a future PR.

libtenzir/builtins/formats/zeek_tsv.cpp
jachris: Ah, right. I copied this from `string_type`, but it seemed wrong to me. I'll fix this for both types then.

libtenzir/include/tenzir/arrow_table_slice.hpp
jachris: Right. Thanks!

libtenzir/include/tenzir/concept/printable/tenzir/blob.hpp
jachris: No, because `blob` currently does not have a parseable representation in TQL. This printer is just for internal purposes. Does this cause any issues?

libtenzir/include/tenzir/detail/base64.hpp
jachris: I was also not so sure about that `std::char_traits<std::byte>`, but we are currently using `std::basic_string_view<std::byte>` for `view<blob>`, and this is not implicitly convertible to `std::span<const std::byte>`. I can add a generic version that also works with `std::span<const std::byte>` if that's preferrable.

jachris: And the reason for using `basic_string[_view]` over `vector` and `span` is that many other places in the code base assume that many types are addable, comparable, etc.

libtenzir/include/tenzir/type.hpp
jachris: Alright, then let's leave it as it is for now. We can always change this later.

tenzir/integration/integration.py
jachris: cc @tobim

libtenzir/src/format/zeek.cpp
dominiklohmann: That's when we effectively break Threat Bus and should also stop offering the `vast` symlink for backwards compatibility. It shouldn't be done as part of this thrust, but I do agree that we should do it soon.

libtenzir/include/tenzir/detail/base64.hpp
dominiklohmann: I trust your judgement on this. I honestly don't know what's better.

tenzir/integration/integration.py
dominiklohmann: Feel free to also just put this into a follow-up issue.

libtenzir/builtins/formats/pcap.cpp
dominiklohmann: The fallback makes sense for now.

libtenzir/include/tenzir/concept/printable/tenzir/blob.hpp
dominiklohmann: I don't think it does, but if it isn't much work it generally makes sense to add it to avoid surprises.

dominiklohmann (approved):
This works well enough in practice to be approved. I think once the outstanding discussions are resolved we can merge this and be done with it.

libtenzir/include/tenzir/concept/printable/tenzir/blob.hpp
jachris: Do we want to address this in a follow-up PR? We first have to agree on syntax, and then also have to address semantics in expressions (because it would then be surprising if these literals would misbehave in queries).

dominiklohmann: Makes sense. Can you write one, mentioning also that the printer currently uses `b"…"`, and if we decide on another representation would have to change that as well?

tenzir/integration/integration.py
jachris: It looks like it should work to me. And if it does not, it will most likely tell us about the invalid UTF-8. Hence, I think this is fine without a potentially no-op issue.

netantho:
This PR solves an issue of stalled pipeline when importing tshark data in (nd)JSON, thanks!

PR #3582, opened 2023-10-17
Upgrade remaining usages of the adaptive table slice builder

jachris:
This could have still waited a bit, but it will make https://github.com/tenzir/tenzir/pull/3581 easier.


libtenzir/builtins/formats/xsv.cpp
```diff
@@ -253,15 +255,22 @@ auto parse_impl(generator<std::optional<std::string_view>> lines,
...
-        ctrl.warn(std::move(err));
-      // TODO: Check what add() does with strings.
+      auto parsed = data{};
+      if (not(parsers::data - parsers::pattern)(value, parsed)) {
+        parsed = value;
+      }
+      auto result = row.field(field).try_data(parsed);
+      if (not result) {
+        ctrl.warn(result.error());
+      }
```
jachris, line 267: This should be the only change in behavior, so please take a look. Previously, inputs such as `42` were parsed as a string, but I don't think that this is what we want.

dominiklohmann: We actually track this for CEF and XSV under tenzir/public-roadmap#94; I think this change should get a changelog entry.

Related: I think CEF still uses the very old table slice builder that has implicit rollover to the next event where you add one leaf at a time; how much effort do you think it would be to remove that builder as well in favor of the new one?

jachris: It already has a changelog entry (although I categorized it as a bug fix, but I can change that to a feature). As an alternative, we can restore the old behavior and later think about the conversions that we want to allow here (this PR allows almost everything, as you can see).

Converting the CEF parser should be minimal effort, I think. In fact, let me just do that right away, while we are at it.

dominiklohmann: I think it's okay to consider this a bug fix rather than a feature. If you do CEF as well then we can just scrap that roadmap item, which is nice. I'll take a look at the remaining code right away.

dominiklohmann (approved):
All the changes look really obvious to me. Feels good to see so much code deleted. 🎉

libtenzir/builtins/formats/xsv.cpp
jachris: cf477c888a6f26994d66177ccb690170c1868cdf is the CEF upgrade: I changed the parsing mechanism to always reuse the same builder (which can add extra fields with null values). Previously, we started a new batch when the extension record type changed in any way. Is the new behavior what want? It leads to larger batch sizes, especially if the shape of the data changes frequently.

e3e187d775c2c4ed04cdaf796c2696bfd64d488d is where I wanted to parse the incoming data for schema inference. But it turns out that we already have that, so this is just a small bug fix instead.

dominiklohmann: The commits both look good to me from glancing at them. The larger batch sizes are definitely beneficial. Please also add a separate changelog entry, then feel free to merge this one.

This PR may also conflict with https://github.com/tenzir/tenzir/pull/3580, for which auto-merge is enabled.

PR #3585, opened 2023-10-18
Implement extended deployment options

Dakostu:
This change is a submodule bump and adds functionality for the extended deployment options in the pipeline manager.

### DIFF
--- a/libtenzir/src/http_api.cpp
+++ b/libtenzir/src/http_api.cpp
@@ -240,10 +240,17 @@ auto parse_endpoint_parameters(const tenzir::rest_endpoint& endpoint,
           auto result = record{};
           for (const auto& x : *parsed) {
             auto parse_result = false;
-            if (!parsers::boolean(caf::get<std::string>(x.second),
-                                  parse_result)) {
+            const auto* str = caf::get_if<std::string>(&x.second);
+            if (not str) {
               return caf::make_error(ec::invalid_argument,
-                                     "currently only boolean values in records "
+                                     "currently only non-null boolean values "
+                                     "in nested records "
+                                     "are supported");
+            }
+            if (!parsers::boolean(*str, parse_result)) {
+              return caf::make_error(ec::invalid_argument,
+                                     "currently only non-null boolean values "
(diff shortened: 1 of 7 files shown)


jachris:
Is this ready for review?

Dakostu:
> Is this ready for review?
I guess so, yes.

libtenzir/src/http_api.cpp
```diff
@@ -225,6 +225,23 @@ auto parse_endpoint_parameters(const tenzir::rest_endpoint& endpoint,
...
+          for (const auto& x : *parsed) {
+            auto parse_result = false;
+            if (!parsers::boolean(caf::get<std::string>(x.second),
+                                  parse_result)) {
+              return caf::make_error(ec::invalid_argument,
+                                     "currently only boolean values in records "
+                                     "are supported");
+            }
+            result[x.first] = parse_result;
+          }
```
jachris, line 242: I was surprised to see that we get a string here instead of a bool, but we seems that we pass bools as their raw string representation for these endpoints...

However, you want `caf::get_if<std::string>` here, as it can still be something else besides `std::string` (and `caf::get` on `caf::variant` with the wrong type is sadly UB...)

jachris (approved):

web/openapi/openapi.yaml
jachris, line 1: There is still a "restart_with_node" in this file.

libtenzir/src/http_api.cpp
Dakostu: Yep, a segfault can be triggered by passing `{{key : null}}` as an HTTP endpoint parameter. I made use of `caf::get_if` in https://github.com/tenzir/tenzir/pull/3585/commits/2c75d49eae971c37db2f883e0384567d5b0f5ec5

web/openapi/openapi.yaml
Dakostu: Removed in https://github.com/tenzir/tenzir/pull/3585/commits/2c75d49eae971c37db2f883e0384567d5b0f5ec5

PR #3590, opened 2023-10-19
Fix predicate pushdown for `export | where …`

dominiklohmann:
> [!NOTE]
> This fixes a high-severity performance bug, and we should consider cutting the next release a bit earlier because of it. I've requested @tobim for review as we've identified this bug together when looking at a performance problem.

This fixes a regression that snuck in between the v4.2 and v4.3 releases. We changed slightly how predicate pushdown works for the `export` operator, which caused unnecessary predicates with trivially true expressions to appear after predicate pushdownn. This exposed a bug in the catalog, which did not bind expressions to the schemas before evaluating them. The consequence of this was a too large set of candidate partitions.

This pull request fixes both the unnecessary predicates in the optimization phase of the `export` operator, and causes the catalog (1) to normalize and validate expressions, and (2) to bind expressions exactly once per schema, ignoring partitions for which tailoring fails.

Additionally, this properly sets the UUID for catalog lookup requests in the `export` operator, which makes tracking the query's lifetime in trace and debug logs easier.

### DIFF
--- a/libtenzir/src/catalog.cpp
+++ b/libtenzir/src/catalog.cpp
@@ -90,13 +90,30 @@ catalog_state::lookup(const expression& expr) const {
   auto start = stopwatch::now();
   auto total_candidates = catalog_lookup_result{};
   auto num_candidates = size_t{0};
-  auto pruned = prune(expr, unprunable_fields);
+  auto normalized = normalize_and_validate(expr);
+  if (not normalized) {
+    return caf::make_error(ec::lookup_error,
+                           fmt::format("{} failed to normalized and validate "
+                                       "expression {}: {}",
+                                       *self, expr, normalized.error()));
+  }
+  auto pruned = prune(*normalized, unprunable_fields);
   for (const auto& [type, _] : synopses_per_type) {
     auto resolved = resolve(taxonomies, pruned, type);
     if (!resolved) {
       return resolved.error();
     }
-    auto candidates_per_type = lookup_impl(*resolved, type);
+    auto tailored = tailor(*resolved, type);
+    if (not tailored) {
+      // Failing to tailor is not an error.
+      TENZIR_DEBUG("{} skips expression {} for schema {} because it cannot be "
+                   "tailored",
+                   *self, expr, type);
+      continue;
+    }
+    TENZIR_TRACE("{} starts lookup for schema {} with expression {}", *self,
+                 type, *tailored);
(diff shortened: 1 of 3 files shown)


libtenzir/src/catalog.cpp
```diff
@@ -356,6 +373,13 @@ catalog_state::lookup_impl(const expression& expr, const type& schema) const {
           }();
           return result;
         },
+        [&](const data_extractor& lhs,
+            const data& d) -> catalog_lookup_result::candidate_info {
+          if (compatible(lhs.type, x.op, d)) {
+            return all_partitions();
+          }
+          return {};
+        },
```
tobim, line 382: Afaik `data_extractor`s are only produced by the `resolve(expression, type)` function, which is not used in the catalog. I suggest to trigger an assertion if we reach this overload, then we can at least catch it in the unit tests.

libtenzir/src/passive_partition.cpp
```diff
@@ -158,9 +158,9 @@ indexer_actor passive_partition_state::indexer_at(size_t position) const {
     indexer = self->spawn(passive_indexer, id, std::move(value_index));
     return indexer;
   }
-  TENZIR_WARN("passive-partition ({}) failed to deserialize value index for "
-              "field {}",
-              id, qualified_index->field_name()->string_view());
+  TENZIR_DEBUG("passive-partition ({}) failed to deserialize value index for "
```
tobim, line 161: Why did you demote this?

libtenzir/src/catalog.cpp
dominiklohmann: `data_extractor` is produced by `tailor`, i.e., when we bind an expression to a schema.

libtenzir/src/passive_partition.cpp
dominiklohmann: We recently disabled creation of dense indexes. We issued this warning also for dense indexes that were disabled, and it's hard with how we serialize them to actually differentiate between intentionally disabled indexes and those we failed to deserialize. We should probably remove this altogether.

libtenzir/src/catalog.cpp
tobim: I see. But then this overload effectively disables the sparse index prefiltering.

As far as I understand we should be able to fix the regression with the changes to `export` alone. Let's move the rest into another PR where we can discuss the changes more in depth.

dominiklohmann:
Closing in favor of #3599

PR #3591, opened 2023-10-19
Improve error in `/serve` endpoint for expired pipelines

dominiklohmann:
This changes the `/serve` endpoint to keep track of previously known but expired serve ids to display a richer error message. Previously, this was indistinguishable from unknown serve ids.

Additionally, this silences a bogus warning for pipelines with an expired TTL.


Dakostu (approved):
Nice!
Does also adding the same error message to `stop()` and `put()` make sense?

dominiklohmann:
> Does also adding the same error message to `stop()` and `put()` make sense?

I think those are only used internally from the serve operator and not so much by the user through the API, and that should already keep track of the lifetime properly. So I don't think we should do this.

PR #3592, opened 2023-10-20
Introduce an experimental `python` pipeline operator

lava:
This experimental operator is intended for situations where development velocity is more important than performance; it allows the user to run arbitrary code against the (converted-to-python) contents of a table slice.

Example:
```
$ tenzir 'show build | python "import math; sanitizers.root = math.sqrt(2*len(tree_hash) + len(type))"'

{
  "type": "Debug",
  "tree_hash": "0a178da187a13feefa10c1f18f7f20e9",
  "assertions": true,
  "sanitizers": {
    "address": true,
    "undefined_behavior": false,
    "root": 8.306623862918075
  }
}
```

### Remaining Work
- [x] Make use of the `extra`s feature to avoid installing false dependencies for the operator venvs.
- [x] Put the venvs under a suitable cache location.
- [x] Synchronize venv creation to prevent write races.
- [x] Implement a "mutable pytenzir" config option.
- [x] Wire python errors into the operator API.
- [x] Extend the test coverage.
- [x] Clean up the commit history.
- [x] Improve the docs.


tenzir/functional-test/tests/setup_suite.bash
```diff
@@ -16,9 +16,9 @@ export BATS_LIB_PATH=${BATS_LIB_PATH:+${BATS_LIB_PATH}:}${SCRIPT_DIR}/..
 BATS_SUITE_DIRNAME="${BATS_TEST_DIRNAME}"
 export BATS_SUITE_DIRNAME
 
-export PYTHONPATH=$PYTHONPATH${PYTHONPATH:+:}$SCRIPT_DIR/../../../python
+# Normalize the environment, but pass through the pytenzir operator helper.
+unset $(printenv | grep -v '^TENZIR_PLUGINS__PYTHON__IMPLICIT_REQUIREMENTS=' | grep -o '^TENZIR[^=]*' | paste -s -)
```
tobim, line 20: Consider doing this dependent on some IMPURE env variable.

libtenzir/src/format/arrow.cpp
```diff
@@ -109,6 +111,7 @@ arrow_istream_wrapper::Read(int64_t nbytes) {
 }
 
 arrow_fd_wrapper::arrow_fd_wrapper(int fd) : fd_{fd} {
+  set_mode(::arrow::io::FileMode::READ);
```
tobim, line 114: Not sure if we really need this.

python/pytenzir/tools/python_operator_executor.py
```diff
@@ -277,6 +277,9 @@ def main():
             batch_out = _execute_user_code(batch_in, code)
             writer = pa.ipc.RecordBatchStreamWriter(ostream, batch_out.schema)
             writer.write(batch_out)
+            # TODO: Consider adding writer.close() here and reading the closing
+            # null sequence on the c++ side. Experimental evidence suggests we
+            # don't need that but in principle we should.
```
tobim: We should actually do this on both ends.

libtenzir/src/format/arrow.cpp
lava: We noticed that `READ` is the default value for the mode anyways

tenzir.yaml.example
```diff
@@ -18,6 +18,14 @@ tenzir:
   # The file system path used for persistent state.
   db-directory: "tenzir.db"
 
+  # The file system path used for persistent state.
+  # Defaults to one of the following paths, selecting the first that is
+  # available:
+  #   - $XDG_CACHE_HOME
+  #   - $XDG_HOME_DIR/.cache/tenzir or
```
dominiklohmann, line 25: - If XDG_HOME_DIR is not set, you should probably use HOME instead.
- On macOS, caches should go under `$HOME/Library/Caches/tenzir` instead, unless either of the XDG variables are defined.

lava: Are the `XDG_` variables usually defined on mac, or would that be the exception?

dominiklohmann: They are undefined, some people define them because some software respects it, but it's far from standard behavior.

tobim (approved):
This has been a long time coming. Thanks to everyone for their input and help. Especially @lava and @jachris.

PR #3593, opened 2023-10-21
Make metrics formatting locale agnostic

tobim:
We can't assume that `en_US.UTF-8` is available at runtime. We format metrics in a locale-agnostic way instead.

### DIFF
--- a/libtenzir/builtins/commands/exec.cpp
+++ b/libtenzir/builtins/commands/exec.cpp
@@ -95,7 +95,6 @@ auto add_implicit_source_and_sink(pipeline pipe) -> caf::expected<pipeline> {
 auto format_metric(const metric& metric) {
   auto result = std::string{};
   auto it = std::back_inserter(result);
-  const auto locale = std::locale("en_US.UTF-8");
   constexpr auto indent = std::string_view{"  "};
   it = fmt::format_to(it, "operator #{} ({})\n", metric.operator_index + 1,
                       metric.operator_name);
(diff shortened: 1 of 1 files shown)


netantho:
This solves https://github.com/orgs/tenzir/discussions/3586 for me, thanks! 🙇

libtenzir/builtins/commands/exec.cpp
```diff
@@ -95,7 +95,7 @@ auto add_implicit_source_and_sink(pipeline pipe) -> caf::expected<pipeline> {
 auto format_metric(const metric& metric) {
   auto result = std::string{};
   auto it = std::back_inserter(result);
-  const auto locale = std::locale("en_US.UTF-8");
+  const auto& locale = std::locale::classic();
```
dominiklohmann, line 98: I think you may as well delete the entire line and remove the usage of the locale in this function. The classic locale has no thousands separators, and that was the only thing this was used for.

dominiklohmann (approved):

libtenzir/builtins/commands/exec.cpp
tobim: I think we should rather keep it to avoid confusion in case the global one ever gets imbued with something else.

dominiklohmann: If you remove the `:L` from the format strings then the locale isn't use in any ways, which I think would be the better way to go about this rather than overriding the locale.

tobim: That's also fine with me.

PR #3594, opened 2023-10-21
Add a `yara` operator

mavam:
This PR adds a `yara` operator that makes it possible to match YARA rules on byte chunks.

```[tasklist]
### Tasks
- [x] Implement the `yara` operator
- [x] Add docs
- [x] Understand double quotes in builder conflict
- [x] Add changelog entry
- [x] Add integration tests
- [x] Add blog post
```


web/docs/operators/transformations/yara.md
```diff
@@ -0,0 +1,147 @@
...
+{
+  "rule": {
+    "identifier": "test",
+    "namespace": "default",
+    "string": "foo",
+    "tags": [],
+    "meta": [
+      {
+        "key": "string",
+        "value": "\"string meta data\""
```
mavam, line 84: @jachris I think I ran into something in the series builder when it promotes a type to a common type. The `value` field has a string with inner double quotes that—I think—should not be there. A verified that the original data doesn't have them by trimming the string, but that actually removed characters from the content.

Then I looked into `conflict_builder` and found this TODO, which seems to be the reason for this issue:

```cpp
      // TODO: We chose the strategy of always serializing conflicting types as
      // JSON strings. This means that we can always get the original data by
      // parsing the resulting strings as JSON. However, we could also use
      // different strategies here. For example, we could directly add string
      // values to the builder without JSON, such that they are preserved
      // without extra quotes.
```

Defaulting to JSON is safe, but as a user, I'm not sure if I want that. Of course, we all know that sum types are the proper solution, so I'm fine with waiting ultimately. But intuitively, it's a lot harder to *use* this value if I need strip double quotes, as opposed to, say, working with a boolean `"true"` string. Mostly because it's hard to understand that this inner value is in fact JSON.

Have you considered a simple `to_string` here, trading preservation with usability?

plugins/yara/functional-test/tests/tests.bats
```diff
@@ -0,0 +1,24 @@
...
+  bats_load_library bats-assert
+  bats_load_library bats-tenzir
+}
+
+@test "generate matches for source rule" {
+  echo 'foo barbar baz' | check tenzir "load stdin | yara ${RULE_DIR}/test.yara"
+}
+
+@test "generate matches for compiled rule" {
+  yarac ${RULE_DIR}/test.yara test.yarac
```
mavam, line 18: @tobim invoking `yarac` compiles the rule and generates a file `test.yarac` in the current directory. Is there the equivalent of a "cwd" for bats? When I run it in the source tree, it ends up there. I guess I'm looking for the equivalent of `CMAKE_CURRENT_BINARY_DIRECTORY` or so.

plugins/yara/functional-test/tests/setup_suite.bash
```diff
@@ -0,0 +1,23 @@
...
+export BATS_SUITE_DIRNAME
+
+TENZIR_DIR="$(realpath "$(dirname "$(command -v tenzir)")")"
+export BATS_LIB_PATH=${BATS_LIB_PATH:+${BATS_LIB_PATH}:}${TENZIR_DIR}/../share/tenzir/functional-test:${BATS_SUITE_DIRNAME}/../../../../tenzir/functional-test
+
+unset "${!TENZIR@}"
+# Enable bare mode so settings in ~/.config/tenzir or the build configuration
+# have no effect.
+export TENZIR_BARE_MODE=1
+export TENZIR_PLUGINS="yara"
```
mavam, line 23: @tobim I copied this file over and only changed the very last line to mention `yara`. Is that the way to do it?

web/docs/operators/transformations/yara.md
jachris: We can make this behavior configurable, but I first want to see whether you'd perhaps prefer one of two other solutions:

1. Use `to_string()` before giving the metadata to the series builder. You will not get double-quoting (because there is no type conflict). Your proposed behavior and this solution only differ in the case where there is no type conflict. I would argue that the consistency of always outputting a string is better than occasionally having the correct type here, as adding metadata to an existing yara rule can otherwise change the semantics of existing queries (due to `false` -> `"false`", which would do not have the same behavior).
2. Serialize the metadata as a record instead of a list. This way, you can fully preserve types and its easier to address the individual contents.

mavam: Both options you listed strike me as superior to what I have started with naively. The record is the cleanest, at a slight cost of schema churn. Whether that's an actual problem we still have to learn, so I'm leaning towards that route.

mavam: Implemented in [18b684f](https://github.com/tenzir/tenzir/pull/3594/commits/18b684f570467dec9c4fde34d17f263b67e25531).

plugins/yara/functional-test/tests/setup_suite.bash
tobim: Yes, that should do it.

scripts/macOS/install-dev-dependencies.sh
```diff
@@ -32,4 +32,5 @@ brew install \
     tcpdump \
     xxhash \
     yaml-cpp \
+    yara \
```
dominiklohmann, line 35: We don't build this plugin in CI on macOS so we shouldn't install it here.

plugins/yara/src/plugin.cpp
```diff
@@ -0,0 +1,423 @@
...
+            reinterpret_cast<const char*>(bytes.data()), bytes.size()};
+          match_rec.field("data").data(str);
+          match_rec.field("base").data(match->base);
+          match_rec.field("offset").data(match->offset);
+          match_rec.field("match_length")
+            .data(detail::narrow_cast<uint64_t>(match->match_length));
+          // TODO: Once we can upgrade to newer versions of libyara, uncomment
+          // the line below. YR_MATCH::xor_key is not available in the version
+          // we get on Debian.
+          // match_rec.field("xor_key").data(uint64_t{match->xor_key});
```
dominiklohmann, line 173: You can also do this with if-constexpr-requires-requires:

```cpp
if constexpr (requires requires { match->xor_key; }) {
  // ...
}
```

plugins/yara/src/plugin.cpp
```diff
@@ -0,0 +1,423 @@
...
+        }
+        YR_MATCH* match = nullptr;
+        auto list = rec.field("matches").list();
+        yr_string_matches_foreach(context, string, match) {
+          auto match_rec = list.record();
+          match_rec.field("identifier").data(string->identifier);
+          auto bytes = std::span<const std::byte>{
+            reinterpret_cast<const std::byte*>(match->data),
+            detail::narrow_cast<size_t>(match->data_length)};
+          // TODO: switch to blob type once available.
```
dominiklohmann, line 162: That now exists.

plugins/yara/src/plugin.cpp
```diff
@@ -0,0 +1,423 @@
...
+  compiler(compiler&& other) noexcept : compiler_{other.compiler_} {
+    other.compiler_ = nullptr;
+  }
+
+  auto operator=(compiler&& other) noexcept -> compiler& {
+    std::swap(compiler_, other.compiler_);
+    return *this;
+  }
+
+  ~compiler() {
```
dominiklohmann, line 228:
```suggestion
  ~compiler() noexcept {
```

web/blog/matching-yara-rules-in-byte-pipelines/index.md
```diff
@@ -0,0 +1,196 @@
...
+title: Matching Yara Rules in Byte Pipelines
+authors:
+   - mavam
+date: 2023-10-24
+tags: [yara, operator, dfir, detection engineering]
+comments: true
+---
+
+The new [`yara`][yara-operator] operator matches [YARA][yara] rules on byte
+pipelines, producing a structured match output to conveniently integrate
```
dominiklohmann, line 11: I think the term "byte pipeline" is a bit confusing. The entire thing is the pipeline; the `yara` operator is matching Yara rules on bytes, but not on pipelines.

web/docs/operators/transformations/yara.md
```diff
@@ -0,0 +1,141 @@
...
+```
+load file --mmap evil | yara rule.yara
+```
+
+:::caution
+Note the `--mmap` flag: we need it to deliver a single block of data to the
+`yara` operator. Without `--mmap`, the [`file`](../../connectors/file.md) loader
+will generate a stream of byte chunks and feed them incrementally to `yara`,
+attempting to match `rule.yara` individualy on *every chunk*. This will cause
+false negatives when rule matches would span chunk boundaries.
```
dominiklohmann, line 59: It may be better to handle this like the `feather` and `parquet` readers do: Merge the chunks and only then match on them. Warn if you receive multiple.

dominiklohmann (approved):
Lots of smaller comments; generally this works, but I think the default to match on individual chunks is dangerous and should be made an option instead. The default behavior of an operator should be what users want most of the time, and that isn't the case here as indicated by the warning in the documentation.

I'm approving because the required changes are really small and should not need another review. Please make them before merging, or let's postpone the blog post and discuss.

web/blog/matching-yara-rules-in-byte-pipelines/index.md
mavam: Can you make a suggestion? I'm not seeing the confusion here and don't know in what direction to correct.

dominiklohmann: I think just "on bytes" is better than "on byte pipelines."

plugins/yara/src/plugin.cpp
mavam: Neat. Does that work with CI? I.e., are we building it exactly on one platform?

dominiklohmann: Ah, it may be problematic for the integration tests, so it would probably require a `drop xor_key` in them if you're on Debian. I forgot about that.

Not sure how important this value is, feel free to just ignore this and leave the TODO as-is.

mavam: Yeah, let's leave the TODO in there until we get a user feedback.

web/blog/matching-yara-rules-in-byte-pipelines/index.md
mavam: Thanks, I used that instead.

web/docs/operators/transformations/yara.md
mavam: Okay, I think that would require a bit of restructuring but it's possible. How would we do this in streaming mode then? With an outer `group` operator?

plugins/yara/functional-test/tests/tests.bats
tobim: There is `BATS_TEST_DIRNAME`, maybe that helps?

web/docs/operators/transformations/yara.md
dominiklohmann: Yes, you'd need some nested pipeline that actually terminates. With some hypothetical operators:

```
ls path/to/dir
| eval {
  load {{file}}
  | yara
}
| …

Comment (dominiklohmann) [reply web/docs/operators/transformations/yara.md:59]: Another advantage of doing this: operators like `shell` and connectors like `s3` will magically work with `yara` as well, which I don't believe upstream Yara can do. That's a pretty cool thing.

File: plugins/yara/functional-test/tests/tests.bats (hunk 2 above)
Comment (tobim) [reply plugins/yara/functional-test/tests/tests.bats:18]: Or maybe you actually want to use `TENZIR_DB_DIRECTORY` for that one. There is no higher level test runtime state dir at the moment.

File: web/docs/operators/transformations/yara.md (hunk 9 above)
Comment (mavam) [reply web/docs/operators/transformations/yara.md:59]: True, but we also don't know yet how many users have the need to match YARA rules on large S3 buckets. I think it's not uncommon, but I do lack data.

Comment (dominiklohmann) [reply web/docs/operators/transformations/yara.md:59]: In any case, it's better to have a default that just works and does not depend on some internal batching parameters that we may want to change without breaking a user's setup.

File: plugins/yara/src/plugin.cpp (hunk 6 above)
Comment (mavam) [reply plugins/yara/src/plugin.cpp:162]: Since YARA rules are also frequently used for plain text, I think we should add an option to display this either as string or as bytes. Default to bytes, but opt-in via `--string-data` to use a string. Otherwise `"bar"` gets represented as `YmFy` in Base64, which may not be helpful at all.

Comment (dominiklohmann) [reply plugins/yara/src/plugin.cpp:162]: You could also do UTF-8 validation for it (simdjson should have something for this) to decide this automatically.

Comment (mavam) [reply plugins/yara/src/plugin.cpp:162]: Uh, that feels like too much magic for me. A user would want to process *either* downstream.

Comment (dominiklohmann) [reply plugins/yara/src/plugin.cpp:162]: > Or perhaps better: make that an option in the respective printer?

I'd like to hear @jachris' thoughts on this. I'm not entirely opposed to it, but we would need to maintain that for every individual printer.

Comment (mavam) [reply plugins/yara/src/plugin.cpp:162]: Another option: just make this a function so that users can do `yara <args> | replace data=f(data)` for an appropriately named function `f`.

For now, I'll just use the `blob` type as it's the most correct one.

Comment (dominiklohmann) [reply plugins/yara/src/plugin.cpp:162]: Makes sense to me. We discussed something similar a while ago for `--numeric-durations`, which is an option for the JSON printer currently. That, too, should be a function that converts from duration to double.

Comment (mavam) [reply plugins/yara/src/plugin.cpp]: 43f5ce853f has the delta. For our text example, it's no longer optimal, but it should work.

File: web/docs/operators/transformations/yara.md (hunk 9 above)
Comment (mavam) [reply web/docs/operators/transformations/yara.md:59]: > Merge the chunks and only then match on them. Warn if you receive multiple.

I don't understand the second sentence: isn't the idea to buffer all chunks and then match on "one giant" one?

FWIW, the YARA C API has mode where you can give a fixed set of memory blocks before doing a scan. So simply buffering all chunks might actually work. Then we'd execute the scan once we're out of chunks.

File: plugins/yara/src/plugin.cpp (hunk 6 above)
Comment (jachris) [reply plugins/yara/src/plugin.cpp]: I tend to value consistency a lot, but I guess it depends on what people usually do afterwards with the data. It would be quite surprising to suddenly change types when you don't expect it, so this should be opt-in if anything. I'm not against providing additional encoding options for JSON. We could also encode something like `b"hello\xFF"` as `"hello\\xFF"` in that case (double backslash is required), but base64 still seems to be the most common solution.

> replace data=f(data)

Agreed, this would be similar to `data = data.decode()` in Python (which can fail).

File: web/docs/operators/transformations/yara.md (hunk 9 above)
Comment (jachris) [reply web/docs/operators/transformations/yara.md:59]: > > Merge the chunks and only then match on them. Warn if you receive multiple.
> 
> I don't understand the second sentence: isn't the idea to buffer all chunks and then match on "one giant" one?

The idea is to emit a warning about this buffering behavior if we receive multiple chunks. You would still continue and execute the rules on the combined content. Here is the warning that we emit for `feather`:

~~~
// TODO: Loading everything into memory here is far from ideal. We should
// instead load passive stores incrementally. For now we at least warn the
// user that this is experimental.
ctrl.warn(caf::make_error(
    ec::unspecified, fmt::format("the experimental {} parser does "
                                 "not currently load files "
                                 "incrementally and may use an "
                                 "excessive amount of memory; consider "
                                 "using 'from file --mmap'",
                                 name)));
~~~

But that is more of a temporary thing, as we eventually want to have an incremental version of this. I'm not sure if it's applicable here. Unless YARA offers a way to incrementally match and notify us when it's done with a chunk?

> FWIW, the YARA C API has mode where you can give a fixed set of memory blocks before doing a scan. So simply buffering all chunks might actually work. Then we'd execute the scan once we're out of chunks.

That would be nice indeed.

File: plugins/yara/src/plugin.cpp (hunk 6 above)
Comment (mavam) [reply plugins/yara/src/plugin.cpp]: > We could also encode something like `b"hello\xFF"` as `"hello\\xFF"` in that case (double backslash is required), but base64 still seems to be the most common solution.

Indeed, Base64 is the fallback. But I've seen "hex-encoding" pretty frequently. Zeek logs, for example, always use that instead. As a user of those, I find it highly valuable to see the ASCII characters and only "get blinded" for non-printable bytes.

@jachris what do you think of defaulting to Base64 and provide hex-encoding as an opt-in?

Comment (dominiklohmann) [reply plugins/yara/src/plugin.cpp]: For Zeek it seems to be hex-encoding, but that actually leads to a lot of trouble since we would need to validate whether it's UTF-8 when parsing. Suricata by default encodes binary payloads as base64.

Ultimately, I see this as yet another use-case for functions. We should still agree on a default, and ideally before the next release when we include the blob type.

NB: I unresolved this since we're actively discussing still, the thread is hard to find otherwise.

Comment (jachris) [reply plugins/yara/src/plugin.cpp]: We always use hex enconding that for Zeek logs, as you can see [here](https://github.com/tenzir/tenzir/blob/bca0cd07f8b6c619a8e01245ca2f7180eefc2150/libtenzir/builtins/formats/zeek_tsv.cpp#L404-L421). 

But for JSON, I don't think there is a common way to do this besides base64. We can of course add it anyways, but this is what makes me hesitate. Or have you seen JSON like this in practice?

For the explorer, this is also not a problem. We can of course show `b"hello\xFF"` there (see https://github.com/tenzir/app/issues/333).

File: web/docs/operators/transformations/yara.md (hunk 9 above)
Comment (mavam) [reply web/docs/operators/transformations/yara.md:59]: > The idea is to emit a warning about this buffering behavior if we receive multiple chunks. You would still continue and execute the rules on the combined content.

I'm struggling with this idea, because it breaks composability. Say I have one file to-be-scanned via Kafka. That's a continuously running pipeline. I *want* to match on every chunk. I think opt-in is fine for this, and that accumulating should still be the default.

> Unless YARA offers a way to incrementally match and notify us when it's done with a chunk?

I'm investigating this right now.

Comment (jachris) [reply web/docs/operators/transformations/yara.md:59]: > I want to match on every chunk [individually].

We can absolutely do this. However, it would imply that, semantically, we no longer deal with bytestreams, but a rather a stream of byte chunks. That's in and of itself fine, and we also seem to be going in the direction with the introduction of the content type. But this is not the mental model we are currently communicating, and it could create issues in the future. Perhaps it would be wiser to separate the "here are multiple blobs" content from the "here is a stream of bytes".

Comment (mavam) [reply web/docs/operators/transformations/yara.md:59]: That's a good observation. The streaming semantics are indeed more about chunks.

For this operator, I am going to provide two usage modes:
1. `... | yara <rule>` buffers merges all chunks in a single one, performing one `memcpy` into an accumulating `chunk` with every new chunk arriving
2. `... | yara --incremental <rule>` doesn't do any buffering but applies `rule` to every chunk individually.

Users would do (2) only if they know that they get a stream of semantically well-defined chunks where chunk overlap wouldn't exist by design (e.g., one file per Kafka message).

Comment (jachris) [reply web/docs/operators/transformations/yara.md:59]: I think this is a good approach for now. (2) might be broken if we would internally rebatch. With a separate type, you would not have to manually specify anything by the way, as this could be automatically detected (and you could therefore not forget it in the Kafka case). But let's stick to your solution for now.

File: web/blog/matching-yara-rules-in-byte-pipelines/index.md [hunk 10]
```diff
@@ -0,0 +1,200 @@
...
+
+```bash
+tenzir 'load file --mmap /tmp/test.txt | yara /tmp/test.yara'
+```
+
+Note the `--mmap` flag: we need it to deliver a single block of data to the
+`yara` operator. Without `--mmap`, the `file` loader will generate a stream of
+byte chunks and feed them incrementally to the `yara` operator, attempting to
+match `rule.yara` individually on *every chunk*. This will cause false negatives
+when rule matches span chunk boundaries.
```
thomaspatzke, line 156: From the usability/safety point of view it would be better to make the `yara` operator failing if it receives a stream of byte chunks and adding an option like `--match-stream` to it if this is really needed.

web/docs/operators/transformations/yara.md
mavam: I've now switched to a `--blockwise` flag to *opt out* of incremental processing. This is what @thomaspatzke recommended, as it avoids footguns.

The good thing is that I was able to use YARA's scanner API that supports cross-chunk matching!

Regarding `--blockwise`: I needed to find a name that doesn't conflict with the YARA CLI, and since "blocks" are the way YARA users think of memory chunks, I opted for that terminology. Also, `-B` as shorthand is not in conflict with any of the `yara` CLI options.

I'm resolving this thread. You may want to review the `memory_block_vector` class separately that implements the incremental matching.

web/blog/matching-yara-rules-in-byte-pipelines/index.md
mavam: Good suggestion. I removed the need to provide `--mmap`. It's now only an optimization.

The default is now "buffer all memory blocks and initiate a scan at the end". I verified that the YARA scanner is capable of matching across block boundaries as follows:

```bash
# Spin up the scanner that matches on "$foo and $bar".
tenzir 'load zmq | yara rule.yara'

# Feed two blocks into the scanner:
echo foo | tenzir 'load stdin | to zmq'
echo bar | tenzir 'load stdin | to zmq'
```

A new `--blockwise` option allows to opt out of the accumulation and match on each byte chunk in isolation. This can run potentially forever without leaking memory.

plugins/yara/src/plugin.cpp
mavam: >But for JSON, I don't think there is a common way to do this besides base64. We can of course add it anyways, but this is what makes me hesitate. Or have you seen JSON like this in practice?

No, which is why I think the current default is good.

I am resolving this thread, as the discussion should probably take place in a different issue that's more discoverable.

plugins/yara/functional-test/tests/tests.bats
mavam: Can you take a look at 55b64a8baf? I used `BATS_TEST_DIRNAME` here, but that puts generated file into the source tree. `TENZIR_DB_DIRECTORY` feels weird. Maybe `mktemp`?

PR #3598, opened 2023-10-24
Update Docker Compose file to use configurable image tag

gitryder:
Updates the Docker Compose file to utilize a configurable image tag, enabling the image to be specified using a .env file for running nodes with changes from a particular Tenzir branch.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -2,7 +2,7 @@
 # file for setup instructions.
 services:
   tenzir-node:
-    image: tenzir/tenzir:main
+    image: ${TENZIR_IMAGE:-tenzir/tenzir:main}
     build:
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -2,7 +2,7 @@
 # file for setup instructions.
 services:
   tenzir-node:
-    image: tenzir/tenzir:main
+    image: ${TENZIR_IMAGE:-tenzir/tenzir:main}
```
dominiklohmann, line 5: Please also document this in the `env.example` file in the repository root.

dominiklohmann (approved):
Approving as-is, just one minor request and then I think this is ready to go.

PR #3599, opened 2023-10-25
Fix predicate pushdown in `export` and other small fixes

dominiklohmann:
This is five fixes in one PR:
1. Predicate pushdown did not work in v4.3; this supersedes my previous attempt in #3590
2. The catalog expected expressions to be already normalized and validated, which for non-normalized expressions caused all partitions to be returned
3. We introduced a race in #3562 that caused operators to sometimes hang on shutdown; this mostly occurred in release builds with high-volume pipelines at the process boundary (no changelog entry for this one as we never released with this bug)
4. The long option `--skip-empty` for `read lines` now works as expected
5. We now include the feature flags in `show build`

### DIFF
--- a/libtenzir/src/catalog.cpp
+++ b/libtenzir/src/catalog.cpp
@@ -730,7 +730,7 @@ catalog(catalog_actor::stateful_pointer<catalog_state> self,
         self->state.merge(uuid, std::move(partition_synopsis));
       return atom::ok_v;
     },
-    [self](atom::candidates, tenzir::query_context& query_context)
+    [self](atom::candidates, const tenzir::query_context& query_context)
       -> caf::result<catalog_lookup_result> {
       TENZIR_TRACE_SCOPE("{} {}", *self, TENZIR_ARG(query_context));
       if (not query_context.ids.empty()) {
--- a/libtenzir/src/catalog.cpp
+++ b/libtenzir/src/catalog.cpp
@@ -31,6 +31,7 @@
 #include "tenzir/logger.hpp"
 #include "tenzir/modules.hpp"
 #include "tenzir/partition_synopsis.hpp"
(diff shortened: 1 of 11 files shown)


tobim (approved):
The changes look correct. The CI failure needs to be investigated still.

libtenzir/src/catalog.cpp
```diff
@@ -86,17 +87,32 @@ void catalog_state::erase(const uuid& partition) {
...
   for (const auto& [type, _] : synopses_per_type) {
-    auto resolved = resolve(taxonomies, pruned, type);
-    if (!resolved) {
-      return resolved.error();
+    auto resolved = resolve(taxonomies, expr, type);
+    if (not resolved) {
+      return caf::make_error(ec::invalid_argument,
+                             fmt::format("{} failed to resolve epxression {}: "
+                                         "{}",
+                                         *self, expr, resolved.error()));
```
dominiklohmann, line 112: I noticed a few very peculiar things about this as part of my investigation into the failing integration test.

1. `resolve` should be called _before_ `prune`, as otherwise `prune` may be ineffective if the expression contains concepts.
2. `resolve` only resolves for the first type that it encounters. That is, if I import a `zeek.dns` before a `suricata.flow`, then `net.src.ip` resolves to `zeek.dns.id.orig_h` even for `suricata.flow` partitions.

(1) is a missed optimization, (2) is a hard bug. I think this only worked previously because always returned all candidate partitions here, and the evaluation logic works correctly.

@tobim can you take this over and then merge this, since I won't have time to take another look at this until the end of the week?

dominiklohmann: I managed to find (2) myself, this was a shadowing mistake. (1) I fixed as well.

PR #3600, opened 2023-10-26
Bump the bundled simdjson to v3.6.0

dominiklohmann:


Dakostu (approved):
Alright.

Should the minimum `simdjson` version in the "Build from source" documentation also be updated?

image

dominiklohmann:
The minimum can stay at v3.1, we don't use any of the newer APIs.

PR #3608, opened 2023-10-31
Add `from/load/to/save <uri/file>`

eliaskosunen:
Supersedes https://github.com/tenzir/tenzir/pull/3500, related to https://github.com/tenzir/issues/issues/328

In this PR:
 - `OP <uri>` is mapped to `OP <uri-scheme> <rest-of-uri>`, where `OP` is one of `(from, load, to, save)`, and `<uri>` is a RFC 3986 -compliant URI.
 - `OP <something-else>` is mapped to `OP file <something-else>` if `<something-else>` isn't a possible plugin name (doesn't conform to the pattern `[A-Za-z0-9-_]*`)

This effectively means, that the following mappings are implemented:
 - `from https://example.com/foo` -> `from https example.com/foo`
 - `from gs://foo/bar` -> `from gcs gs://foo/bar`
 - `from file:///tmp/file` -> `from file /tmp/file`
 - `from /tmp/file` -> `from file /tmp/file`
 - `from ./local-file` -> `from file ./local-file`
 - `from local-file` -> error (no loader `local-file` found)
 - `from https` -> error (no URL given to `https` loader)

The logic implemented in this PR is as follows:
image

This current implementation accepts `from https:foo.bar/baz` as `from https foo.bar/baz`. It's strictly not RFC 3986 -compliant, but I don't think it's that big of an issue.

- [X] Implement `from <uri> [options...] [read <parser>]`
- [X] Implement `from <path>` for `from file <path>`
- [x] Same for `load`
- [x] Same for `to` and `save`
- [x] Consider some loader restrictions + remapping (e.g., for `gcs`)

### DIFF
--- a/libtenzir/builtins/operators/from_load_read.cpp
+++ b/libtenzir/builtins/operators/from_load_read.cpp
@@ -16,6 +16,82 @@
 namespace tenzir::plugins::from {
 namespace {
 
+class prepend_token final : public parser_interface {
+public:
+  prepend_token(located<std::string> token, parser_interface& next)
+    : token_{std::move(token)}, next_{next} {
+  }
+
+  auto accept_shell_arg() -> std::optional<located<std::string>> override {
+    if (token_) {
+      return std::exchange(token_, std::nullopt);
+    }
+    return next_.accept_shell_arg();
+  }
+
+  auto peek_shell_arg() -> std::optional<located<std::string>> override {
+    if (token_) {
+      return token_;
+    }
+    return next_.peek_shell_arg();
+  }
+
+  auto accept_identifier() -> std::optional<identifier> override {
+    TENZIR_ASSERT(not token_);
+    return next_.accept_identifier();
+  }
+
+  auto peek_identifier() -> std::optional<identifier> override {
+    TENZIR_ASSERT(not token_);
+    return next_.peek_identifier();
+  }
+
+  auto accept_equals() -> std::optional<location> override {
+    TENZIR_ASSERT(not token_);
+    return next_.accept_equals();
+  }
+
+  auto accept_char(char c) -> std::optional<location> override {
+    TENZIR_ASSERT(not token_);
+    return next_.accept_char(c);
+  }
+
+  auto parse_operator() -> located<operator_ptr> override {
+    TENZIR_ASSERT(not token_);
+    return next_.parse_operator();
+  }
+
+  auto parse_expression() -> tql::expression override {
+    TENZIR_ASSERT(not token_);
+    return next_.parse_expression();
+  }
+
+  auto parse_legacy_expression() -> located<expression> override {
+    TENZIR_ASSERT(not token_);
+    return next_.parse_legacy_expression();
+  }
+
+  auto parse_extractor() -> tql::extractor override {
+    TENZIR_ASSERT(not token_);
+    return next_.parse_extractor();
+  }
+
+  auto at_end() -> bool override {
+    return not token_ && next_.at_end();
+  }
+
+  auto current_span() -> location override {
+    if (token_) {
+      return token_->source;
+    }
+    return next_.current_span();
+  }
+
+private:
+  std::optional<located<std::string>> token_;
+  parser_interface& next_;
+};
+
 class load_operator final : public crtp_operator<load_operator> {
 public:
   load_operator() = default;
@@ -183,20 +259,44 @@ public:
   auto parse_operator(parser_interface& p) const -> operator_ptr override {
     auto usage = "from <loader> <args>... [read <parser> <args>...]";
     auto docs = "https://docs.tenzir.com/next/operators/sources/from";
-    auto l_name = p.accept_shell_arg();
-    if (!l_name) {
+    auto q = until_keyword_parser{"read", p};
+    auto l_name = q.accept_shell_arg();
+    if (not l_name) {
       diagnostic::error("expected loader name")
-        .primary(p.current_span())
+        .primary(q.current_span())
         .usage(usage)
         .docs(docs)
         .throw_();
     }
-    auto l_plugin = plugins::find<loader_parser_plugin>(l_name->inner);
-    if (!l_plugin) {
-      throw_loader_not_found(*l_name);
+    auto loader = std::unique_ptr<plugin_loader>{};
+    if (auto split = l_name->inner.find("://"); split != std::string::npos) {
+      auto substr
+        = [](located<std::string> x, size_t pos,
+             size_t count = std::string::npos) -> located<std::string> {
+        auto sub = x.inner.substr(pos, count);
+        auto sub_source = location::unknown;
+        if (x.source && (x.source.end - x.source.begin) == x.inner.size()) {
+          sub_source.begin = x.source.begin + pos;
+          sub_source.end = sub_source.begin + sub.length();
+        }
+        return {std::move(sub), sub_source};
+      };
+      auto scheme = substr(*l_name, 0, split);
+      auto path = substr(*l_name, split + 3);
+      auto l_plugin = plugins::find<loader_parser_plugin>(scheme.inner);
+      if (!l_plugin) {
+        throw_loader_not_found(scheme);
+      }
+      auto r = prepend_token{std::move(path), q};
+      loader = l_plugin->parse_loader(r);
+      TENZIR_DIAG_ASSERT(r.at_end());
+    } else {
+      auto l_plugin = plugins::find<loader_parser_plugin>(l_name->inner);
+      if (!l_plugin) {
+        throw_loader_not_found(*l_name);
+      }
+      loader = l_plugin->parse_loader(q);
     }
-    auto q = until_keyword_parser{"read", p};
-    auto loader = l_plugin->parse_loader(q);
     TENZIR_DIAG_ASSERT(loader);
     TENZIR_DIAG_ASSERT(q.at_end());
     auto parser = std::unique_ptr<plugin_parser>{};
--- a/libtenzir/builtins/operators/from_load_read.cpp
+++ b/libtenzir/builtins/operators/from_load_read.cpp
@@ -270,6 +270,7 @@ public:
     }
     auto loader = std::unique_ptr<plugin_loader>{};
     if (auto split = l_name->inner.find("://"); split != std::string::npos) {
+      // Has "://", assuming to be a URI
       auto substr
         = [](located<std::string> x, size_t pos,
              size_t count = std::string::npos) -> located<std::string> {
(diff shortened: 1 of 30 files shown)


libtenzir/src/detail/loader_saver_resolver.cpp
```diff
@@ -123,6 +123,18 @@ struct try_plugin_by_uri_result {
   }
 };
 
+auto plugin_name_for_scheme(std::string_view scheme) -> std::string_view {
```
dominiklohmann, line 126: Should this be a property of the loader and saver plugins?

eliaskosunen: Maybe? On an architectural/conceptual level, it might make more sense that way. I'll try out what it'd look like.

eliaskosunen: It might be better that way. I made the mapping 1-to-1: each plugin can support only a single URI scheme, which defaults to `plugin::name()`.

libtenzir/builtins/operators/from_load_read.cpp
```diff
@@ -170,6 +170,23 @@ auto parse_default_parser(std::string definition)
...
+    diagnostic::error("expected loader name")
+      .primary(p.current_span())
+      .usage(usage)
+      .docs(docs)
+      .throw_();
+  }
+  auto [loader, name] = detail::resolve_loader(p, *l_name);
+  if (not loader) {
+    throw_loader_not_found(name);
+  }
```
jachris, line 186: My Arrow installation does not have the component for the `gcs` plugin right now (so I can't easily check it), but I think this might create a very confusing error message for `load gcs://...`:

~~~
error: loader `gcs` could not be found
 --> <input>:1:6
  |
1 | load gcs://...
  |      ^^^ 
  |
  = hint: must be one of ..., gcs, ...
  = docs: https://docs.tenzir.com/next/connectors
~~~

Perhaps you need a dedicated error path for this.

libtenzir/builtins/operators/from_load_read.cpp
jachris, line 1: Take a look at this:

~~~
$ tenzir 'load gcs bucket/logs/log.json'
error: unexpected positional argument
 --> <input>:1:10
  |
1 | load gcs bucket/logs/log.json
  |          ^^^^^^^^^^^^^^^^^^^^ 
  |
  = usage: file <path> [-f|--follow] [-m|--mmap] [-t|--timeout <duration>]
  = docs: https://docs.tenzir.com/next/connectors/file
~~~

I don't have the `gcs` plugin installed, but any invalid name would do, such as a misspelled plugin name. I think we should only delegate to the `file` loader if the file (or perhaps the path) exists. If not, then the error message should communicate that we expected either a valid file path or a loader name.

This is also relevant for `save` and `to`: `load gcs | save gcs` will actually create an empty file called `gcs` on my system. Now, that might be quite a silly pipeline, but what about about `... | to kafka`? If there is a `kafka` plugin, this is a valid pipeline. But if not, this will create a file called `kafka` and write JSON to it. This seems like hard edge case to solve, in particular because I really like being able to do something `from foo.json | ...`, and you can make an argument for symmetry here. 

What do you think about all of this?

jachris, line 1: This PR allows arguments after the URI/path. This might be want we want, but I'm not 100% sure. What do you think?

jachris, line 1: Can you create an issue that spells out the missing pieces to consider the URI feature complete? Thanks!

jachris (approved):
Sorry that the review took so long. Had too many things to do, but it's very nice to get this feature finally!

libtenzir/builtins/operators/from_load_read.cpp
eliaskosunen: That's certainly peculiar. I think it all boils down to how, in general, there's really no way of differentiating between a name of a plugin we don't have loaded, and a filename without an extension (like `gcs` or `kafka`). I can see four possible solutions to this.

1. No change (modulo maybe some error reporting adjustments). The logic is quite clear: the argument given to the operator is either a plugin name, a URI with a schema, or a filename. The simple case is simple: `load foo.json`.
2. Check if the file exists only when reading (`from`/`load`), and error if it doesn't. Solves `load gcs`, but not `save gcs`, because it's easy to imagine that we'd want to write to a file that doesn't exist. TOCTOU issues are possible, and the symmetry between reads and writes would be lost.
3. If the operator is given two positional arguments, assume that the first one is meant to be a plugin name, and the second one a path. So, `load gcs bucket/foo.json` would always be understood as giving `bucket/foo.json` to `gcs`, even if no such plugin exists. This would lead to a clearer error message: "No `gcs` plugin found", rather than "unexpected positional argument (for `file`)". This would not solve the ambiguity with `load/to gcs` when no such plugin is loaded.
4. Require the argument to be a plugin name, or a URI. Force the use of `file` or `file://` to explicitly disambiguate: `load foo.json` -> `load file foo.json` or `load file://foo.json`. Avoids the aforementioned issues, but makes the simple case a tad bit more complex.

I think (3) should probably be adopted no matter what. Then, in addition, I think it's between (1) and (4): (1) allows for a cleaner simple case, while (4) is never ambiguous. From an engineering perspective, (4) is thus more robust.

eliaskosunen: I think `from foo.json -m` should be treated as `load file foo.json -m`, and `from https://foo.bar header:value` as `load https foo.bar header:value`, i.e. how this PR is currently implemented.

eliaskosunen: I think I'll unlink this PR and modify the original issue.

jachris: There is one more knob that I want to throw into the discussion: Writing to a file that is also a valid plugin name should be pretty rare (assuming `.` cannot be contained in plugins). So we can make `to [a-zA-Z]` always refer to the saver plugin, and if it is not a valid plugin name (such as `to foo.json`), then we interpret it as a file path. And the same for loaders.

dominiklohmann, line 1: Please add a changelog entry and update the documentation of the changed operators.

eliaskosunen: After discussion, doing it so that `[A-Za-z0-9-_]*` is assumed to be a plugin name, and a filename/path otherwise.

eliaskosunen: I think I've done that, now.

libtenzir/builtins/operators/from_load_read.cpp
eliaskosunen: I modified the logic in `throw_loader_not_found` and `throw_saver_not_found`: I think the error messages are better now.

libtenzir/src/detail/loader_saver_resolver.cpp
```diff
@@ -0,0 +1,229 @@
...
+    TENZIR_DIAG_ASSERT(r.at_end());
+    return {std::move(parsed), make_located_string_view(uri.scheme)};
+  } else if (uri.valid_uri_parsed()) {
+    // Valid URI, but no matching plugin name -> error
+    return {nullptr, make_located_string_view(uri.scheme)};
+  }
+  // Check if `name` could be a valid plugin name
+  auto plugin_name_pattern = pattern::make("[A-Za-z0-9-_]*");
+  TENZIR_DIAG_ASSERT(plugin_name_pattern);
+  if (plugin_name_pattern->match(name.inner)) {
```
jachris, line 191: There is also the option to use the plugin name parser here:
~~~cpp
if (parsers::plugin_name(name.inner)) {
  ...
}
~~~

web/docs/operators/sources/load.md
```diff
@@ -36,8 +48,19 @@ Read bytes from stdin:
...
+from https example.com/file.json
+```
+
 Read bytes from the file `path/to/eve.json`:
 
 ```
+from path/to/eve.json
 from file path/to/eve.json
 ```
+
```
jachris, line 64: This has to be `load` instead of `from` (I blame the existing example).

libtenzir/builtins/operators/from_load_read.cpp
jachris: Perhaps it's even better if you replace ``loader `{}` could not be found`` with ``loader for `gs` scheme could not be found`` or something like that.

libtenzir/src/detail/loader_saver_resolver.cpp
eliaskosunen: That's better, thanks

libtenzir/builtins/operators/from_load_read.cpp
eliaskosunen: Modified again, also to take uri-ness properly into account.

web/docs/connectors/uri.md
```diff
@@ -0,0 +1,42 @@
+# URI
```
dominiklohmann, line 1: @mavam Do you think it makes more sense to document URIs as a connector, or to expand the synopsis of operators working with URIs like this?

```
from <uri>
from <file>
from <connector>
from <connector> read <format>
```

I am undecided between the two, I don't know what's better.

mavam: I don't think it makes sense as separate connector. If we did, I'd expect something like `from uri <uri>`. Rather, I think we should lift it to the parent context, similar to what we did with modifiers.

In particular, we could put this into `/connectors` and `/formats` where we already write:

```
from <connector> [read <format>]
to <connector> [write <format>]
```

The only thing I'm still thinking about is the duplication of content here.

jachris: I agree that `URI` should not be documented as a proper connector, and that we should not duplicate the path/URI/loader/saver description four times. Adding it to the synopses seems fine though, although it might be a bit inconsistent with how we use `<connector>`. But I would be okay with that (though I would probably use `<path>` instead of `<file>` to prevent confusion with `file`). `/connectors` seems like a good place to document this once, and we can link to that from the other operators. `/formats` could even stay as-is. There should not be much duplication then, right?

jachris (commented):
The improved UX is very nice! Thanks.

web/docs/connectors/uri.md
mavam: I like that approach!

eliaskosunen: I did some changes. Is it aligned with what you had in mind, now?

jachris: That's more or less the opposite direction. The suggestion was to move the URI and path docs to `/connectors`, that is: `web/docs/connectors.md`. The goal is to only have "actual" connectors (which URI and path are not) in `web/docs/connectors/*.md`.

eliaskosunen: Ah, got it. I got the directory and the file confused. Let's try again!

jachris: Yes, that's better.

jachris:
Unless there are any remaining issues, I think we can merge this :rocket:

cc @dominiklohmann @mavam

dominiklohmann:
I think we can merge this, yes.

PR #3611, opened 2023-11-02
Do not assume YARA can match across chunks

mavam:
This PR simplifies the chunk management, switching away from the incremental interface and simply accumulating the input in a buffer to ensure that we perform a scan on a single chunk.


dominiklohmann (approved):
Thank you for fixing this so quickly. I think the reworded blog post makes sense. Good that we caught this before the release.

PR #3612, opened 2023-11-02
Add a --live option to the export operator

tobim:
Running `export --live` captures all data that is ingested to
the internal database. In this mode, the stream never ends.


tobim:
Note that this would be a temporary solution until the `publish` and `subscribe` operators are ready. But this was pretty easy to implement and lifts some important restrictions.

nix/tenzir/default.nix
mavam, line 1: Did this mean to go in here?

tobim: I hoped it would just work and I could sneak it in. But it needs an additional fix, so it goes into another PR.

libtenzir/builtins/operators/export.cpp
```diff
@@ -23,23 +23,170 @@
...
+        return self->state.rp;
+      }
+      auto result = std::move(self->state.buffer.front());
+      self->state.buffer.pop();
+      return result;
+    },
+  };
+};
+
+// class export_live_operator final : public crtp_operator<export_live_operator>
```
dominiklohmann, line 91: To be deleted.

libtenzir/builtins/operators/export.cpp
```diff
@@ -132,6 +279,8 @@ class export_operator final : public crtp_operator<export_operator> {
   auto optimize(expression const& filter, event_order order) const
     -> optimize_result override {
     (void)order;
+    if (live_)
```
dominiklohmann, line 282: Can't comment on the code, but `detached()` should return false in live-mode.

libtenzir/builtins/operators/export.cpp
dominiklohmann, line 1: Please add a changelog entry.

dominiklohmann (approved):
Approving, this is a mighty-fine addition. 🙏

libtenzir/builtins/operators/export.cpp
dominiklohmann: Suggestion:

> Get all imported events as they come in with the new `--live` option for the `export` operator.

PR #3613, opened 2023-11-03
Fix a deprecation warning when using Arrow >= 14.0

dominiklohmann:
Arrow's `arrow::io::BufferReader` complains when being constructed from a `arrow::Buffer` instead of a `std::shared_ptr<arrow::Buffer>` since the recent release of Arrow 14.0.0. The change is entirely backwards compatible.

### DIFF
--- a/plugins/parquet/parquet.cpp
+++ b/plugins/parquet/parquet.cpp
@@ -286,9 +286,9 @@ std::shared_ptr<arrow::Schema> parse_arrow_schema_from_metadata(
   if (!arrow_metadata.ok())
     return {};
   auto decoded = detail::base64::decode(*arrow_metadata);
-  auto schema_buf = arrow::Buffer(decoded);
-  arrow::ipc::DictionaryMemo dict_memo;
-  arrow::io::BufferReader input(schema_buf);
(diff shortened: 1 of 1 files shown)


plugins/parquet/parquet.cpp
```diff
@@ -286,9 +286,9 @@ std::shared_ptr<arrow::Schema> parse_arrow_schema_from_metadata(
   if (!arrow_metadata.ok())
     return {};
   auto decoded = detail::base64::decode(*arrow_metadata);
-  auto schema_buf = arrow::Buffer(decoded);
-  arrow::ipc::DictionaryMemo dict_memo;
-  arrow::io::BufferReader input(schema_buf);
+  auto schema_buf = std::make_shared<arrow::Buffer>(std::move(decoded));
+  auto dict_memo = arrow::ipc::DictionaryMemo{};
+  auto input = arrow::io::BufferReader{schema_buf};
```
jachris, line 291: Nitpick:

```suggestion
  auto input = arrow::io::BufferReader{std::move(schema_buf)};
```

jachris (approved):

PR #3616, opened 2023-11-06
Use prefix matching instead of suffix matching

jachris:
This PR enables `select <field_of_record_type>` and similar examples by using prefix matching instead of suffix matching. This is a breaking change because it also makes it so that `select bar` does not select the inner field of `{"foo": {"bar": 42}}` anymore. However, we wanted to make this change anyway in the near future.

Some uses of `record_type::resolve_key_suffix` still remain for now. In particular, expressions are still resolved using suffix semantics.

### DIFF
--- a/changelog/next/changes/3616--prefix-matching.md
+++ b/changelog/next/changes/3616--prefix-matching.md
@@ -1,0 +1,3 @@
+The operators `drop`, `pseudonymize`, `put`, `extend`, `replace`, `rename` and
+`select` were converted from suffix matching to prefix matching and can
+therefore address records now.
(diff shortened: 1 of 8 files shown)


mavam:
Since it's a breaking change, will this require cutting v5.0?

If we don't want to cut v5.x, then we can do one of these things:

1. Add a feature flag and make your implementation the new default to avoid cutting v5.x
2. Label this a *breaking change* and argue that operator UX may change in minor versions
3. Label this only as *change* and move on

dominiklohmann (approved):
I tested various combinations. Nothing too surprising in the code.

image

dominiklohmann:
> Since it's a breaking change, will this require cutting v5.0?
> 
> If we don't want to cut v5.x, then we can do one of these things:
> 
> 1. Add a feature flag and make your implementation the new default to avoid cutting v5.x
> 2. Label this a _breaking change_ and argue that operator UX may change in minor versions
> 3. Label this only as _change_ and move on

I wouldn't consider this a breaking change. The only place where we ever documented the suffix matching behavior is for expressions, and that doesn't change here.

PR #3619, opened 2023-11-06
Add `export --internal` to access metrics

jachris:
WIP


libtenzir/include/tenzir/import_helper.hpp
```diff
@@ -0,0 +1,124 @@
...
+      out.push(std::move(next));
+    }
+  }
+
+  auto done() const noexcept -> bool override {
+    return input_.unsafe_current() == input_.end();
+  }
+
+  void finalize(const caf::error& error) override {
+    // TODO: Why do we catch `unreachable` here?
```
dominiklohmann, line 56: You get an unreachable error from CAF when an actor handle's refcount drops to zero, so this is not really an error.

libtenzir/fbs/table_slice.fbs
```diff
@@ -10,6 +10,9 @@ table v2 {
   /// A timestamp assigned on import, representing nanoseconds since the UNIX
   /// epoch.
   import_time: long;
+
+  /// TODO
+  internal: bool;
```
dominiklohmann, line 15: This is neat, I really like this way of doing it. Much better than checking whether the schema name begins with `tenzir.`.

dominiklohmann (commented):
👀

jachris:
I still have to write the docs for the actual metrics events, but besides that, this should be ready for review.

libtenzir/builtins/operators/import.cpp
```diff
@@ -76,45 +30,48 @@ class import_operator final : public crtp_operator<import_operator> {
...
+      if (slice.rows() == 0) {
+        co_yield {};
+        continue;
+      }
+      total_events += slice.rows();
+      stream.enqueue(std::move(slice));
+      while (stream.enqueued() > 0) {
+        co_yield {};
+      }
+    }
```
jachris, line 56: I'm not sure if this is ideal. Previously, we had the following:

~~~
while (input.unsafe_current() != input.end()) {
  if (source->generate_messages()) {
    source->out().emit_batches();
  }
  co_yield {};
}
~~~

But IIRC, this was mostly to work around a bug. Still, this refactoring changes the logic around stalling and backpressure quite a bit, so this needs a second critical look.

web/docs/metrics.md
```diff
@@ -1,106 +1,43 @@
...
+  /// True if the associated pipeline was created by running an open pipeline in
+  /// the explorer. If false, then the pipeline has a name and is visible in the
+  /// overview.
+  hidden: bool,
+  /// Measurement of the incoming stream (see below).
+  input: tenzir.metrics.operator.measurement,
+  /// Measurement of the outgoing stream (see below).
+  output: tenzir.metrics.operator.measurement,
+}
+```
```
dominiklohmann, line 30: Did you try putting these in a table?

|key|type|description|
|-|-|-|
|`pipeline_id`|`string`|The ID of the pipeline where the associated operator is from.|
|...|...|...|
|`output.unit`|`string`|The type of the output elements; either `void`, `bytes`, or `events`|
|`output.elements`|`uint64`|The number of bytes or events that were seen during the collection period. Non-cumulative.|
|`output.approx_bytes`|`uint64`|If `output.unit` is `events`, then this is an estimate of the memory footprint of the associated events. Otherwise, this is the same as `output.elements`.|

dominiklohmann (commented):
Two initial comments on the docs to start a discussion; still reviewing the code changes.

web/docs/metrics.md
```diff
@@ -1,106 +1,43 @@
 # Metrics
 
-Tenzir collects various metrics during execution. This reference describes what
-metrics are available and what they mean. We describe how to [collect
-metrics](setup-guides/collect-metrics.md) in the corresponding user guide.
+:::warning Experimental
+This flag and the schema of the returned events are subject to change.
```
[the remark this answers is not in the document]
jachris: Sorry, perhaps I should have positioned the Todolist more prominently. I'm still working on `web/docs/metrics.md`, please do not review this yet.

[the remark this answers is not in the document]
jachris: Docs should be ready now.

libtenzir/include/tenzir/import_stream.hpp
```diff
@@ -0,0 +1,218 @@
...
+
+private:
+  std::shared_ptr<import_source_state> state_;
+  std::function<void(caf::error)> on_done_;
+};
+
+} // namespace detail
+
+// TODO: This is probably not the right abstraction.
+class import_stream {
```
jachris, line 75: I'm now pretty sure that is a bad abstraction. Shall we proceed anyway and revisit this once there is more clarity regarding how this should work?

dominiklohmann (approved):

PR #3623, opened 2023-11-06
Demo node improvements

rdettai:
- Push the demo node image to DockerHub because it can be handy to play with the Demo node locally. 
- Change the import to use the http operator. To make that possible we also switch to the compressed file data instead of the tarballs.
- Remove the intermediate image layer with data files

### DIFF
--- a/demo-node/entrypoint.bash
+++ b/demo-node/entrypoint.bash
@@ -45,7 +45,7 @@ curl -X POST \
 
 # !! Not started because the data has already been imported while building the image.
 # Newlines improve readability in the app.
-suricata_pipe="tenzir 'load https https://storage.googleapis.com/tenzir-datasets/M57/suricata.json.zst \n| decompress zstd \n| read suricata \n| import'"
+suricata_pipe="tenzir 'load https https://storage.googleapis.com/tenzir-datasets/M57/suricata.json.zst \n| decompress zstd \n| read suricata \n| where #schema != \\\"suricata.stats\\\" \n| import'"
 curl -X POST \
   -H "Content-Type: application/json" \
   -d "{\"name\": \"M57 Suricata Import\", \"definition\": \"${suricata_pipe}\", \"start_when_created\": false}" \
--- a/demo-node/entrypoint.bash
+++ b/demo-node/entrypoint.bash
@@ -29,11 +29,13 @@ done
 #   http://127.0.0.1:5160/api/v0/pipeline/create
 
 # Continuously import system load data from `vmstat -a -n 1`.
-stat_pipe="shell /demo-node/csvstat.sh | read csv | replace #schema=\"vmstat.all\" | unflatten | import"
+stat_pipe="shell /demo-node/csvstat.sh | read csv | replace #schema=\\\"vmstat.all\\\" | unflatten | import"
+echo $stat_pipe
 curl -X POST \
   -H "Content-Type: application/json" \
   -d "{\"name\": \"System Load\", \"definition\": \"${stat_pipe}\", \"start_when_created\": true}" \
   http://127.0.0.1:5160/api/v0/pipeline/create
+echo
 
 # Ingest CVEs from https://services.nvd.nist.gov/rest/json/cves/2.0.
 # !! Currently disabled because of a scheduling bug.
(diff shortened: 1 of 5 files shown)


demo-node/entrypoint.bash
```diff
@@ -43,20 +43,16 @@ curl -X POST \
...
   -H "Content-Type: application/json" \
   -d "{\"name\": \"M57 Suricata Import\", \"definition\": \"${suricata_pipe}\", \"start_when_created\": false}" \
   http://127.0.0.1:5160/api/v0/pipeline/create
 
-# The shell operator "streams" the data into a `Zeek/` directory on disk, calls
-# `sync` to flush the OS write buffers, and finally sends the concatenation of
-# the logs into `read zeek-tsv`.
 # !! Not started because the data has already been imported while building the image.
-zeek_pipe="shell 'bash -c \\\"curl -s -L https://storage.googleapis.com/tenzir-datasets/M57/zeek.tar.zst | tar -x --zstd; sync; cat Zeek/*.log\\\"' | read zeek-tsv | import"
+zeek_pipe="tenzir 'load https https://storage.googleapis.com/tenzir-datasets/M57/zeek-all.log.zst | decompress zstd | read zeek-tsv | import'"
```
dominiklohmann, line 54: Since we soon will show the definition on multiple lines, can you `s/ | /\\n| /g` for both this and the Suricata pipeline?

demo-node/entrypoint.bash
```diff
@@ -43,20 +43,16 @@ curl -X POST \
 #  -d "{\"name\": \"Live CVE Notifications from the NIST API\", \"definition\": \"${cve_pipe}\", \"start_when_created\": true}" \
 #  http://127.0.0.1:5160/api/v0/pipeline/create
 
-# The shell operator decompresses the data and writes it to `read suricata` on
-# the fly.
 # !! Not started because the data has already been imported while building the image.
-suricata_pipe="shell 'bash -c \\\"curl -s -L https://storage.googleapis.com/tenzir-datasets/M57/suricata.tar.zst | tar -x --zstd --to-stdout\\\"' | read suricata | where #schema != \\\"suricata.stats\\\" | import"
+suricata_pipe="tenzir 'load https https://storage.googleapis.com/tenzir-datasets/M57/suricata.json.zst | decompress zstd | read suricata | import'"
```
dominiklohmann, line 47: This one doesn't contain the `where #schema != suricata.stats` that the one from `import.bash` has.

demo-node/entrypoint.bash
rdettai: you mean replacing with 
```
tenzir 'load https https://storage.googleapis.com/tenzir-datasets/M57/zeek-all.log.zst \n| decompress zstd \n| read zeek-tsv \n| import'
```
?

demo-node/entrypoint.bash
rdettai: should we not get rid of that filter now that we support larger events?

dominiklohmann: I think we should keep it. The events are rather meaningless.

demo-node/entrypoint.bash
dominiklohmann: Yes, although I think you can remove the space before the newline as well.

rdettai: doesn't hurt much and it helps readability of the code so I would rather leave them

dominiklohmann: Fine, I don't feel too strongly about them.

rdettai:
@dominiklohmann @tobim do you mind if I merge [RUN /demo-node/load.bash](https://github.com/tenzir/tenzir/blob/6980b5456a4268a813628a5e5672eddf2a56756f/Dockerfile#L226) with `RUN /demo-node/import.bash` and don't store the raw files locally anymore. Downloading them into a layer increases the image size by ~1GB which is actually quite a lot 😅. It wouldn't even involve that much more overhead at build time as the layer is invalidated anytime the base image/code changes anyway.

dominiklohmann:
I think it'd be even better if you spin up a node, start the two pipelines using the API, and then wait for them to complete; that way, the demo node also shows the two pipelines as completed and you don't need to have them twice in the file.

rdettai:
I'll just wait for Tobi's input as he's the original author and there might be some non-obvious reason for the extra layer

demo-node/load.bash
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
 
-curl -L https://storage.googleapis.com/tenzir-datasets/M57/suricata.tar.zst | tar -x --zstd
-curl -L https://storage.googleapis.com/tenzir-datasets/M57/zeek.tar.zst | tar -x --zstd
+set -e
+
+curl -L https://storage.googleapis.com/tenzir-datasets/M57/suricata.json.zst | unzstd > suricata.json
```
mavam, line 5: Why not this?

```bash
tenzir 'load https storage.googleapis.com/tenzir-datasets/M57/suricata.json.zst
       | decompress zstd
       | save stdout' > suricata.json
```

rdettai: I think we should rather get rid of the layer entirely and do the load and import within the same pipeline. Waiting for Tobi's confirmation on this.

rdettai: got verbal confirmation from Tobi

tobim:
> I'll just wait for Tobi's input as he's the original author and there might be some non-obvious reason for the extra layer. In all cases I'm quite confident this spared 1GB might reduce the demo node startup of 10s or more.

The download size itself should be less than 1GB thanks to compression.
The original use case for it was to test loading the data into the image as part of the build.
I don't think we need this mode any more, so we can just as well get rid of that layer.

demo-node/entrypoint.bash
```diff
@@ -44,20 +46,21 @@ curl -X POST \
...
 # !! Not started because the data has already been imported while building the image.
-# Newlines improve readability in the app.
-zeek_pipe="tenzir 'load https https://storage.googleapis.com/tenzir-datasets/M57/zeek-all.log.zst \n| decompress zstd \n| read zeek-tsv \n| import'"
-
+zeek_pipe_formatted=$(echo $ZEEK_PIPE | sed -r 's/ \| /\\n| /g' | sed -r 's/"/\\\\\\\"/g' )
+echo $suricata_pipe_formatted
 curl -X POST \
   -H "Content-Type: application/json" \
-  -d "{\"name\": \"M57 Zeek Import\", \"definition\": \"${zeek_pipe}\", \"start_when_created\": false}" \
+  -d "{\"name\": \"M57 Zeek Import\", \"definition\": \"tenzir '${zeek_pipe_formatted}'\", \"start_when_created\": false}" \
```
dominiklohmann, line 63: Would it make sense to change this to actually start the pipeline, and then to wait until it is completed? That way, the pipeline will show up as completed rather than as created.

rdettai: To me, the very best would even be not to preload the data at build time and let the user see the import of the Zeek and Suricata datasets happening live. This way the demo nodes would connect even faster because they would be more lightweight in every sense. But unfortunately if I remember correctly there were some OOM issues with that.

dominiklohmann: You can probe whether all pipelines are done like this:

```bash
curl -X POST http://127.0.0.1:5160/api/v0/pipeline/list |
  jq -e '.pipelines | map(.state) | all(. == "completed")'
```

Which you can use in a loop with some sleep to wait until both pipelines are done.

Afterwards, you can issue a `tenzir-ctl rebuild --all` to optimize the DB.

Unless I'm missing something, this would be the strictly better option.

> To me, the very best would even be not to preload the data at build time and let the user see the import of the Zeek and Suricata datasets happening live.

I'd also be in favor of that, but it would require expanding the tour a fair bit. It would, however, allow us to remove the special image entirely.

rdettai:
I think this PR is derailing a bit. My original intent was just to push the demo image to DockerHub 😅 . I propose we merge this as is and open follow up items:
- add back a continuous pipeline (needs some product design first)
- preload the data using the pipeline API instead of the current combination of "CLI build time load + unstarted pipeline in the entrypoint "

dominiklohmann:
Fine by me—please create the issues on the housekeeping board before merging.

rdettai:
Done, still missing an approval though.

dominiklohmann (approved):

PR #3630, opened 2023-11-07
Add an `api` source operator

dominiklohmann:
This operator makes it possible to use Tenzir's REST API without going through the `web` plugin's locally spun up server. This is mostly motivated by a testing use case: Right now, all of our internal tests use `curl` to test API endpoints, which can be very hard to test, and always adds an implicit dependency to the `web` plugin.

The syntax is very simple:

```
api <endpoint> [<key=value>...]
```

For example, to create a pipeline in the pipeline manager API:

```
api /pipeline/create '{"name": "Suricata Import", "definition": "load file /tmp/eve.sock | read suricata | import"}'
```

The operator parses the API response and returns it as an event, which makes it possible to interact with it just like any other data:

```
api /pipeline/list
| write yaml
```


libtenzir/builtins/operators/api.cpp
```diff
@@ -0,0 +1,147 @@
...
+        .throw_();
+    }
+    // TODO: We're using a `caf::settings` here over `record` because that
+    // supports inserting `autostart.created=true` as a nested value directly.
+    // Since we're just working with strings here there is no loss of precision,
+    // but it's still not ideal to have to do this roundtrip.
+    auto request_body = caf::settings{};
+    while (auto arg = p.accept_shell_arg()) {
+      auto kvp = detail::split_escaped(arg->inner, "=", "\\", 1);
+      TENZIR_ASSERT(kvp.size() == 2);
```
mavam, line 127: This should probably be a throwing diagnostic.

web/docs/operators/sources/api.md
```diff
@@ -0,0 +1,46 @@
+# api
+
+Use Tenzir's REST API directly from within a pipeline.
```
mavam, line 3:
```suggestion
Use Tenzir's REST API directly from a pipeline.
```

web/docs/operators/sources/api.md
```diff
@@ -0,0 +1,46 @@
...
+web server, making all APIs accessible from within pipelines.
+
+:::info OpenAPI
+Visit [Tenzir's REST API specification](/api) to see a list of all available
+endpoints.
+:::
+
+### `<endpoint>`
+
+The endpoint to query, e.g., `/pipeline/list` to list all pipelines created
```
mavam, line 23:
```suggestion
The node to query, e.g., `/pipeline/list` to list all pipelines created
```

mavam (approved):

libtenzir/builtins/operators/api.cpp
dominiklohmann: I ended up changing the logic back to just take the JSON body as an argument. Passing lists of records as arguments as for example needed to set a label is otherwise not easily possible.

web/docs/operators/sources/api.md
dominiklohmann: This doesn't make sense, it's an endpoint we're sending a request to. I changed the verb to _request_ to clarify this.

web/docs/operators/sources/api.md
dominiklohmann: Fixed in 68752d1bb8

Dockerfile
```diff
@@ -213,19 +213,18 @@ ENTRYPOINT ["tenzir-node"]
...
+    apt-get -y --no-install-recommends install jq && \
     rm -rf /var/lib/apt/lists/*
+
+USER tenzir:tenzir
 COPY demo-node /demo-node
-ENV SURICATA_PIPE='load https https://storage.googleapis.com/tenzir-datasets/M57/suricata.json.zst | decompress zstd | read suricata | where #schema != "suricata.stats" | import'
-ENV ZEEK_PIPE='load https https://storage.googleapis.com/tenzir-datasets/M57/zeek-all.log.zst | decompress zstd | read zeek-tsv | import'
+env TENZIR_AUTOMATIC_REBUILD=0
+env TENZIR_ALLOW_UNSAFE_PIPELINES=true
 RUN /demo-node/import.bash
```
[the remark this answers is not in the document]
rdettai: ci also fails: https://github.com/tenzir/tenzir/actions/runs/6853071273/job/18633180149?pr=3630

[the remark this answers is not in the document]
dominiklohmann: Ah damn, this was the last change I made locally after testing because I found the script name to no longer be appropriate. 🙈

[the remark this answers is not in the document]
dominiklohmann: Fixed in cd0a0e45fc

demo-node/setup.bash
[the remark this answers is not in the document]
dominiklohmann: I considered that as well, but then making changes to that script also causes the entire Docker image to be rebuilt. We'd need to find a better solution for not copying the entire scripts directory over during the build of the base image then.

I think this is for a follow-up PR if you want to do this.

Dockerfile
```diff
@@ -220,11 +220,15 @@ RUN apt-get update && \
...
+ENV TENZIR_AUTOMATIC_REBUILD=0 \
+    TENZIR_ALLOW_UNSAFE_PIPELINES=true \
+    TENZIR_DB_DIRECTORY="/var/lib/tenzir-demo"
+COPY --from=tenzir-ce --chown=tenzir:tenzir /var/lib/tenzir/ /var/lib/tenzir
+COPY --from=tenzir-ce --chown=tenzir:tenzir /var/lib/tenzir/ /var/lib/tenzir-demo
 RUN /demo-node/setup.bash
+ENV TENZIR_DB_DIRECTORY="/var/lib/tenzir"
 
-ENTRYPOINT ["tenzir-node"]
+ENTRYPOINT ["sh", "-c", "2>&1 echo Copying demo data... && cp -rf /var/lib/tenzir-demo/. /var/lib/tenzir/ && tenzir-node"]
```
rdettai, line 231: why reset demo nodes on restart?

rdettai: ah, I guess it's already what it does. Did you have a use-case where you want to mount a persistent volume? If not I would rather leave it as is instead of adding this additional complexity and weight.

dominiklohmann: I tried to use the demo node image in our Docker Compose setup, and that didn't work without this change because it mounts by default. This makes it work transparently.

rdettai: which docker compose are you referring to exactly?

dominiklohmann: The one at the root of this repository.

rdettai: If you really have a use for it, lets leave it. But if we can spare it I would really not add this extra complexity.

rdettai: If the main goal is to test the demo node image, I really think we should have a dedicated setup for that instead of adding a workaround that impacts all prod runs 😄

rdettai:
I tested a build of the latest demo node against the platform integration test and it failed, I need a moment to understand why

rdettai (approved):
Regarding the failing tests on the platform side, it's just that we were using a custom entrypoint that was overriding this new one with copying.

Dockerfile
rdettai: Last remark: if we keep this copying logic, I would just move it back to an entrypoint script for readability and extensibility. Calling `exec` is also always nicer to have the `tenzir-node` process be the root process.

dominiklohmann: I can move it back out, I don't mind. Would you prefer that?

dominiklohmann: Then you can also keep the custom entrypoint.

dominiklohmann: Changed in 04c662bccae42c912266603fa13dcc5bf065e339

PR #3631, opened 2023-11-08
Build arm64 macOS packages

tobim:
This PR makes it possible to build macOS packages that only need standard system libraries to run.

Our approach uses the same Nix dependency management as for the static Linux package under the hood.

### DIFF
--- a/.github/workflows/nix.yaml
+++ b/.github/workflows/nix.yaml
@@ -13,7 +13,7 @@ permissions:
 
 jobs:
   tenzir:
-    name: Tenzir (${{ runner.os }} ${{ runner.arch }})
+    name: Tenzir
     runs-on: ${{ fromJSON(inputs.config).editions[0].gha-runner-labels }}
     steps:
       - name: Prevent pushing protected editions to public repos
--- a/.github/workflows/nix.yaml
+++ b/.github/workflows/nix.yaml
@@ -13,15 +13,15 @@ permissions:
 
 jobs:
   tenzir:
-    name: Tenzir
-    runs-on: ubuntu-20.04
+    name: Tenzir (${{ runner.os }} ${{ runner.arch }})
+    runs-on: ${{ fromJSON(inputs.config).editions[0].gha-runner-labels }}
     steps:
       - name: Prevent pushing protected editions to public repos
         run: |
           # Create 'name registry' pairs
           jq -r '.editions[] | select(."image-registries" | length > 0) | .name as $name | ."image-registries"[]? // ."image-registries" | [ $name, . ] | join(" ")' \
             <<< '${{ inputs.config }}' | while read -r name registry; do
-            # Explicitly allow open source images.
+            # Explicitly allow blessed images.
             [[ "${name}" =~ ^tenzir$ ]] && continue
             # Explicitly allow private registries.
             [[ "${registry}" == "622024652768.dkr.ecr.eu-west-1.amazonaws.com" ]] && continue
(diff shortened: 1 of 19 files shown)


tenzir/CMakeLists.txt
```diff
@@ -90,7 +93,7 @@ elseif (${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
       \"Installing: \$ENV{HOME}/Library/LaunchAgents/com.tenzir.tenzir.plist\")
     configure_file(
       \"${CMAKE_CURRENT_SOURCE_DIR}/services/launchd/com.tenzir.tenzir.plist.in\"
-      \"\$ENV{HOME}/Library/LaunchAgents/com.tenzir.tenzir.plist\" @ONLY)"
+      \"${CMAKE_INSTALL_DATADIR}/Library/LaunchAgents/com.tenzir.tenzir.plist\" @ONLY)"
```
dominiklohmann, line 96: This will break the launch agent, as it has to be installed in `~/Library/LaunchAgents` (per-user installation) or `/Library/LaunchAgents` (system-wide installation) or `/System/Library/LaunchAgents` (default system-wide installation).

tobim: Is it possible to create symlinks in those directories instead?

dominiklohmann: I just tried this out, and unfortunately it makes it so the file won't be picked up.

We can also just make this optional in CMake and disable it in CI.

.github/workflows/nix.yaml
```diff
@@ -14,14 +14,14 @@ permissions:
 jobs:
   tenzir:
     name: Tenzir
```
dominiklohmann, line 16:
```suggestion
    name: Tenzir (${{ runner.os }} ${{ runner.arch }})
```

tobim: Unfortunately the workflow fails with this change applied. It seems the runner context is not available in this scope.

dominiklohmann: Just leave it out then.

dominiklohmann (approved):
Approving with some open questions left to address. Let's give this a spin for the next release before we work on codesigning.

nix/tenzir/default.nix
```diff
@@ -222,9 +271,16 @@
           maintainers = with maintainers; [tobim];
         };
       }
+      # Buggy on darwin?
+      // lib.optionalAttrs (isStatic && stdenv.isLinux) {
+        allowedRequisites = ["out"];
+      }
```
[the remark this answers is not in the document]
tobim: Not really. This setting makes the build fail if the output contains any reference to the build time dependencies. If it does the static build would not work and we want to notice that early.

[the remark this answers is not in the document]
tobim: I improved the comment.

CMakePresets.json
```diff
@@ -11,6 +11,7 @@
         "CMAKE_FIND_USE_SYSTEM_PACKAGE_REGISTRY": "OFF",
         "CMAKE_FIND_USE_PACKAGE_REGISTRY": "OFF",
         "CMAKE_OSX_SYSROOT": "",
+        "_MAC_DEPENDENCY_PATHS": "DISABLED",
```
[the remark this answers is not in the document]
tobim: It is a workaround that prevents cmake from searching fink / macports / homebrew paths when it is supposed to only use dependencies provided by Nix.
Essentially disables `TenzirMacDependencyPaths.cmake`.

nix/tenzir/default.nix
```diff
@@ -71,9 +72,13 @@
...
         "plugins/velociraptor"
         "plugins/web"
-        "plugins/yara"
         "plugins/zmq"
       ]
+      # Temporarily disable yara on the static mac build because of issues
+      # building protobufc.
+      ++ lib.optionals (!(stdenv.isDarwin && isStatic)) [
+        "plugins/yara"
+      ]
```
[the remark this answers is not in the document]
tobim: Created one in Housekeeping.

nix/tenzir/default.nix
```diff
@@ -161,21 +168,63 @@
...
 
+        # TODO: Fix LTO on darwin by passing these commands by their original
+        # executable names "llvm-ar" and "llvm-ranlib". Should work with
+        # `readlink -f $AR` to find the correct ones.
+        #preConfigure = lib.optionalString stdenv.isDarwin ''
+        #  cmakeFlagsArray+=("-DCMAKE_C_COMPILER_AR=$(command -v $AR)")
+        #  cmakeFlagsArray+=("-DCMAKE_CXX_COMPILER_AR=$(command -v $AR)")
+        #  cmakeFlagsArray+=("-DCMAKE_C_COMPILER_RANLIB=$(command -v $RANLIB)")
+        #  cmakeFlagsArray+=("-DCMAKE_CXX_COMPILER_RANLIB=$(command -v $RANLIB)")
+        #'';
```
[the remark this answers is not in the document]
tobim: I created one in Housekeeping.

PR #3636, opened 2023-11-13
Fix `http` loader use-after-free

jachris:
This is an oversight from https://github.com/tenzir/tenzir/pull/3584.


dominiklohmann (approved):

eliaskosunen (approved):

mavam:
Also fixed in https://github.com/tenzir/tenzir/pull/3539, but not yet visible.

PR #3638, opened 2023-11-14
Flush implicitly in the `import` operator

dominiklohmann:
This is a two-fold PR:
1. This fixes a race condition in `tenzir-ctl flush` that was introduced with `export --live` between the importer and the index, where events that were in the outbound buffer of the importer stream stage when the index was asked to flush were ignored.
2. This changes `import` to always flush before returning.


libtenzir/include/tenzir/actors.hpp
```diff
@@ -234,7 +234,9 @@ using importer_actor = typed_actor_fwd<
   auto(stream_sink_actor<table_slice>)
     ->caf::result<caf::outbound_stream_slot<table_slice>>,
   // Register a FLUSH LISTENER actor.
-  auto(atom::subscribe, atom::flush, flush_listener_actor)->caf::result<void>>
+  auto(atom::subscribe, atom::flush, flush_listener_actor)->caf::result<void>,
+  // Decomissions all active partitions, effectively flushing them to disk.
```
tobim, line 238: The importer doesn't know about partitions, so we should not talk about them here.
```suggestion
  // Push buffered slices downstream to make the data available.
```

tobim (approved):
Looks OK. Did not test locally.

PR #3639, opened 2023-11-14
Always enable time and bool synopses

dominiklohmann:
This changes the catalog to always build time and bool synopses. These are really compact, so there is no downside to having them always enabled.

### DIFF
--- a/changelog/next/changes/3639--build-time-and-bool-synopses.md
+++ b/changelog/next/changes/3639--build-time-and-bool-synopses.md
@@ -1,0 +1,2 @@
+Sparse indexes for time and bool fields are now always enabled, accelerating
+lookups against them.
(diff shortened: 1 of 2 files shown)


mavam:
What about a min-max sketch for durations?

dominiklohmann:
> What about a min-max sketch for durations?

We don't have that in the code base right now and it's not in scope for this PR. Adding that is a bit more work.

mavam:
> We don't have that in the code base right now and it's not in scope for this PR

I don't follow. If we have it for `time`, we necessarily have it for `duration` because `time` is just a `duration` with an epoch.

dominiklohmann:
@mavam I opened #3641 to implement that separately. It's hard to test right now, and while I've empirically verified it I'm not sure we should rely on the old synopsis infrastructure too much. Its serialization is a mysterious blackbox. It seems to work from some basic tests, but we should test that PR carefully in larger setup to verify its backwards compatibility.

mavam (approved):

PR #3640, opened 2023-11-14
Improve `summarize` result for empty inputs

jachris:
If the `summarize` operator has no `by` clause, it now returns a result even if there is no input. For example, `summarize num=count(.)` returns an event with `{"num": 0}`. Aggregation functions which do not have a single default value, for example because it would depend on the input type, return `null`.

### DIFF
--- a/libtenzir/builtins/aggregation-functions/distinct.cpp
+++ b/libtenzir/builtins/aggregation-functions/distinct.cpp
@@ -109,6 +109,10 @@ class plugin : public virtual aggregation_function_plugin {
     };
     return caf::visit(f, input_type);
   }
+
+  auto aggregation_default() const -> data override {
+    return caf::none;
+  }
 };
 
 } // namespace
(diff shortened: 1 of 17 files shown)


jachris:
I added 220cda301fed39d40df7c3ec61a81371f30e3aa1 to make `summarize` use the proper `null_type`, now that we have it. I don't think we need a changelog for this.

libtenzir/builtins/aggregation-functions/distinct.cpp
```diff
@@ -109,6 +109,10 @@ class plugin : public virtual aggregation_function_plugin {
     };
     return caf::visit(f, input_type);
   }
+
+  auto aggregation_default() const -> data override {
+    return caf::none;
```
dominiklohmann, line 114: Should this be an empty list?

dominiklohmann (approved):
Thanks for the two fixes. The implementation looks good to me, just one small question.

libtenzir/builtins/aggregation-functions/distinct.cpp
jachris: The problem is that you don't know the exact type here. The `series_builder` makes a `list<null>` out of the empty list. I'm not sure whether that's the best outcome, but I'll make this change anyway as it sounds reasonable.

dominiklohmann: I agree that it is reasonable to have.

PR #3641, opened 2023-11-14
Add min-max sketches for duration and numeric types

dominiklohmann:
This adds and enables min-max sketches for all duration, double, uint64 and int64 fields by default.

This patch is a bit repetitive for historical serialization reasons in the code base.


mavam (approved):
Looks straight-forward to me.

PR #3642, opened 2023-11-14
Add an operator blocklist

dominiklohmann:
The `tenzir.disable-plugins` option is a list of names of plugins and builtins to explicitly forbid from being used in Tenzir.

For example, adding `export` will prohibit use of the `export` operator builtin, and adding `kafka` will prohibit use of the `kafka` connector plugin. This allows for a more fine-grained control than the `tenzir.allow-unsafe-pipelines` option.

```[tasklist]
### Tasks
- [x] Implement the blocklist
- [x] Write a changelog entry
- [x] Write documentation
```

### DIFF
--- a/changelog/next/features/3642--operator-blocklist.md
+++ b/changelog/next/features/3642--operator-blocklist.md
@@ -1,0 +1,5 @@
+The `tenzir.disable-plugins` option is a list of names of plugins and builtins
+to explicitly forbid from being used in Tenzir. For example, adding `export`
+will prohibit use of the `shell` operator builtin, and adding `shell` will
+prohibit use of the `kafka` connector plugin. This allows for a more
+fine-grained control than the `tenzir.allow-unsafe-pipelines` option.
--- a/libtenzir/src/application.cpp
+++ b/libtenzir/src/application.cpp
@@ -83,6 +83,10 @@ void add_root_opts(command& cmd) {
     "plugins to load at startup; the special values 'bundled' "
     "and 'all' enable autoloading of bundled and all plugins "
     "respectively.");
+  cmd.options.add<caf::config_value::list>(
+    "?tenzir", "disable-plugins",
+    "plugins and builtins to explicitly disable; use to forbid use of "
+    "operators, connectors, or formats by policy.");
   cmd.options.add<std::string>("?tenzir", "aging-frequency",
                                "interval between two aging "
                                "cycles");
--- a/libtenzir/src/configuration.cpp
+++ b/libtenzir/src/configuration.cpp
@@ -454,7 +454,8 @@ caf::error configuration::parse(int argc, char** argv) {
   // command line. This needs to happen before the regular parsing of the
   // command line since plugins may add additional commands and schemas.
   auto is_not_plugin_opt = [](auto& x) {
-    return !x.starts_with("--plugins=") && !x.starts_with("--plugin-dirs=")
+    return !x.starts_with("--plugins=") && !x.starts_with("--disable-plugins=")
+           && !x.starts_with("--plugin-dirs=")
            && !x.starts_with("--schema-dirs=");
   };
   auto plugin_opt = std::stable_partition(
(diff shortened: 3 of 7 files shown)


libtenzir/src/application.cpp
```diff
@@ -83,6 +83,10 @@ void add_root_opts(command& cmd) {
     "plugins to load at startup; the special values 'bundled' "
     "and 'all' enable autoloading of bundled and all plugins "
     "respectively.");
+  cmd.options.add<caf::config_value::list>(
+    "?tenzir", "disabled-plugins",
```
dominiklohmann, line 87: @tobim @jachris before I write documentation for this, let's agree on the name for this option.

jachris: `disable-plugins` would fit better with other option names, but I don't care too much TBH.

libtenzir/src/application.cpp
```diff
@@ -83,6 +83,10 @@ void add_root_opts(command& cmd) {
...
+    "plugins and builtins to explicitly disable; use to forbid use of "
+    "operators, connectors, or formats by policy.");
```
jachris, line 89: I'm not sure if this would be the best way to disable operators etc. But it doesn't have to be perfect as we can introduce other ways later, so please feel free to go ahead with this.

changelog/next/features/3642--operator-blocklist.md
```diff
@@ -0,0 +1,5 @@
+The `tenzir.disabled-plugins` option is a list of names of plugins and builtins
+to explicitly forbid from being used in Tenzir. For example, adding `export`
+will prohibit use of the `export` operator builtin, and adding `kafka` will
+prohibit use of the `kafka` connector plugin. This allows for a more
+fine-granular control than the `tenzir.allow-unsafe-pipelines` option.
```
jachris, line 5:
```suggestion
prohibit use of the `kafka` connector plugin. This allows for a more
fine-grained control than the `tenzir.allow-unsafe-pipelines` option.
```

tenzir.yaml.example
```diff
@@ -85,6 +85,13 @@ tenzir:
   # to load the example plugin.
   plugins: []
 
+  # Names of plugins and builtins to explicitly forbid from being used in
+  # Tenzir. For example, adding `export` will prohibit use of the `export`
+  # operator builtin, and adding `kafka` will prohibit use of the `kafka`
+  # connector plugin. This allows more fine-granular control than the
+  # `tenzir.allow-unsafe-pipelines` option.
```
jachris, line 92:
```suggestion
  # connector plugin. This allows more fine-grained control than the
  # `tenzir.allow-unsafe-pipelines` option.
```

libtenzir/src/configuration.cpp
```diff
@@ -464,10 +465,18 @@ caf::error configuration::parse(int argc, char** argv) {
...
+  } else if (auto vast_plugin = detail::getenv("VAST_PLUGINS")) {
+    plugin_args.push_back(fmt::format("--plugins={}", *vast_plugin));
+  }
+  if (auto tenzir_disabled_plugins
+      = detail::getenv("TENZIR_DISABLED_PLUGINS")) {
+    plugin_args.push_back(
+      fmt::format("--plugins={}", *tenzir_disabled_plugins));
+  } else if (auto vast_disabled_plugins
+             = detail::getenv("VAST_DISABLED_PLUGINS")) {
+    plugin_args.push_back(fmt::format("--plugins={}", *vast_disabled_plugins));
```
jachris, line 479: This should probably be `--disable(d)-plugins={}`?

jachris (approved):

libtenzir/src/application.cpp
dominiklohmann: I chose this because it's the only layer of abstraction that equally works for operators, connectors, and formats. The error messages this causes are suboptimal (e.g., "unknown operator `shell`") but it enabled the use case quickly.

jachris: Yes, I understand — and this is also nice to have in general.

libtenzir/src/application.cpp
dominiklohmann: I changed it accordingly.

PR #3644, opened 2023-11-15
Bring back dark/light mode toggle

mavam:
This PR brings back the dark/light mode toggle in the docs.

When sharing screens or giving demos, it's arguably easier to quickly click a button than changing the system settings.

### DIFF
--- a/web/docusaurus.config.js
+++ b/web/docusaurus.config.js
@@ -188,7 +188,7 @@ async function createConfig() {
           indexName: 'vast',
         },
         colorMode: {
-          disableSwitch: true,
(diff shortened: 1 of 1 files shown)


mavam:
The only downside is that it's not possible to adjust the position of the toggle. It will look like this:

image

Unless we invest into React component swizzling, which doesn't have a justifiable ROI.

dominiklohmann (approved):
I personally much prefer not having the toggle, as having the toggle makes it so the system color scheme is only respected exactly once, but I don't have a strong enough opinion to argue on this. 🤷‍♂️

PR #3645, opened 2023-11-15
Add `syslog` parser

eliaskosunen:
This PR expands the `syslog` format already implemented, to add support for RFC 3164 -compliant(ish) syslogs, called `legacy_syslog` in the code. RFC 5424 support was already implemented.

Alongside that, a `syslog` plugin is added, which can be used to parse syslogs as a part of a pipeline.

```
# RFC 5424 (schema = syslog.rfc5424):
$ echo "<30>1 2020-03-02T19:44:50.224836+01:00 parallels-Parallels-Virtual-Platform packagekitd 1370 - -  Parent finished..."
  | tenzir 'from stdin read syslog'
{
  "facility": 3,
  "severity": 6,
  "version": 1,
  "ts": "2020-03-02T18:44:50.224835",
  "hostname": "parallels-Parallels-Virtual-Platform",
  "app_name": "packagekitd",
  "process_id": "1370",
  "message_id": "",
  "message": " Parent finished..."
}


# RFC 3164 (schema = syslog.rfc3164):
$ echo "Nov 16 14:55:56 mymachine/10.1.1.16 PROGRAM: Freeform message"
  | tenzir 'from stdin read syslog'
{
  "facility": null,
  "severity": null,
  "timestamp": "Nov 16 14:55:56",
  "host": "mymachine/10.1.1.16",
  "tag": "PROGRAM",
  "content": "Freeform message"
}

# Malformed (schema = syslog.unknown):
$ echo "foobar" | tenzir 'from stdin read syslog'
{
  "syslog_message": "foobar"
}
```

Possible extensions (to do here, or in a separate PR):
 * Specify RFC version as an argument (note, that 3164 and 5424 are never ambiguous, and can always be separated)
 * Opt-in to require well-formed input
 * Parse timestamp into an actual datetime (Probably opt-in, with an option to specify the year)


eliaskosunen:
I'll still need to add a changelog entry, and a new format to the docs.

libtenzir/builtins/formats/syslog.cpp
```diff
@@ -0,0 +1,111 @@
...
+    if (auto parser = format::syslog::message_parser{}; parser(f, l, msg)) {
+      builder.emplace(format::syslog::make_syslog_type());
+      if (not builder->add(msg.hdr.facility, msg.hdr.severity, msg.hdr.version,
+                           msg.hdr.ts, msg.hdr.hostname, msg.hdr.app_name,
+                           msg.hdr.process_id, msg.hdr.msg_id, msg.msg)) {
+        diagnostic::error(
+          "syslog parser (RFC 5242) failed to produce table slice")
+          .hint(fmt::format("line: `{}`", *line))
+          .emit(ctrl.diagnostics());
+        co_yield {};
```
jachris, line 45: You almost always want to `co_return` after emitting an error.

libtenzir/include/tenzir/format/syslog.hpp
```diff
@@ -250,6 +254,134 @@ struct message_parser : parser_base<message_parser> {
...
+    const auto prival = ignore(
+      integral_parser<uint16_t, 3>{}.with(is_prival)->*to_facility_and_severity);
+    const auto priority_parser = '<' >> prival >> '>';
+    const auto word = +(parsers::printable - space);
+    const auto ws = +space;
+    const auto wsignore = ignore(ws);
+    const auto is_month = [](const std::string& mon) {
+      return mon == "Jan" || mon == "Feb" || mon == "Mar" || mon == "Apr"
+             || mon == "May" || mon == "Jun" || mon == "Jul" || mon == "Aug"
+             || mon == "Sep" || mon == "Oct" || mon == "Nov" || mon == "Dec";
```
jachris, line 294: Do we want such strong validation (not only for months, but in general)? This is not a rhetoric question, I'm not sure what's best here. I guess we would need some kind of heuristic if for example the date is not in a format that we expect (e.g., everything after priority and up to the second-to-last word before the colon). And that might also break if we want to transform this to an actual timestamp using the current year.

libtenzir/src/format/syslog.cpp
jachris, line 1: You don't need to change this. This is only used in the legacy interface.

tenzir/integration/tests.yaml
```diff
@@ -330,6 +330,12 @@ tests:
       - command: -N import syslog
         input: data/syslog/syslog.log
       - command: '-N export ascii "where #schema == /syslog.*/"'
+  Import legacy syslog:
+    tags: [syslog, import]
+    steps:
+      - command: -N import syslog
+        input: data/syslog/syslog-rfc3164.log
+      - command: '-N export ascii "where #schema == /syslog.*/"'
```
jachris, line 338: This tests the legacy code path. We don't use this anymore.

web/docs/formats/syslog.md
```diff
@@ -0,0 +1,66 @@
+# syslog
+
+Reads syslogs.
```
jachris, line 3: I think we should stick to the singular "syslog".

jachris (commented):
Thank you for continuing this :pray: Below are just some first reactions.

libtenzir/include/tenzir/format/syslog.hpp
eliaskosunen: It's a fine line. We need to do _some_ validation, because the lenient parser we had initially

```cpp
auto timestamp = word >> ws >> word >> ws >> word;
auto p = ~(priority >> wsignore)
         >> timestamp >> wsignore
         >> word >> wsignore        // host
         >> word >> ':'             // tag
         >> *parsers::printable;
```

would accept inputs that are clearly not conformant to the old-rfc format, which was caught by our integration tests:

```
# Malformed new-rfc syslog message
2003-10-11T22:14:15.003Z mymachineexamplecom su asd ID47: - Message

# Would nevertheless be parsed as the following old-rfc message:
{
  "facility": null,
  "severity": null,
  "timestamp": "2003-10-11T22:14:15.003Z mymachineexamplecom su",
  "host": "asd",
  "tag": "ID47",
  "content": "- Message"
}
```

I think it's better that those messages are filed under the `syslog.unknown` schema, with everything being in one `"syslog_message"` field, rather than the result being poorly structured like above.

jachris: That's a great argument :+1:

libtenzir/include/tenzir/format/syslog.hpp
```diff
@@ -250,6 +254,134 @@ struct message_parser : parser_base<message_parser> {
...
+  };
+}
+
+inline type make_legacy_syslog_type() {
+  return type{"syslog.rfc3164", record_type{{{"facility", uint64_type{}},
+                                             {"severity", uint64_type{}},
+                                             {"timestamp", string_type{}},
+                                             {"host", string_type{}},
+                                             {"tag", string_type{}},
+                                             {"content", string_type{}}}}};
```
jachris, line 375: Is there a reason to not choose the same name for the `timestamp`and `host` fields?

libtenzir/builtins/formats/syslog.cpp
```diff
@@ -0,0 +1,108 @@
...
+      }
+    } else {
+      builder.emplace(format::syslog::make_unknown_type());
+      if (not builder->add(*line)) {
+        diagnostic::error(
+          "syslog parser (unknown format) failed to produce table slice")
+          .hint(fmt::format("line: `{}`", *line))
+          .emit(ctrl.diagnostics());
+        co_return;
+      }
```
jachris, line 67: I can see why this might be useful. In the JSON parser, we took a different approach and discard unparseable events instead, but emit a warning. The problem is that most downstream operators will not expect this format. But perhaps that's okay until we figure out a proper story for handling malformed input?

libtenzir/builtins/formats/syslog.cpp
```diff
@@ -0,0 +1,108 @@
...
+      co_yield {};
+      continue;
+    }
+    if (line->empty()) {
+      co_yield {};
+      continue;
+    }
+    const auto* f = line->begin();
+    const auto* const l = line->end();
+    std::optional<table_slice_builder> builder{};
```
jachris, line 33: You create (and flush) a new builder for every iteration here. This means that the outgoing batches only have 1 event each. Can you rewrite this so that the builder is only flushed when the event type changes?

web/docs/formats/syslog.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Synopsis
+
+```
+syslog
+```
+
+## Description
+
+[Syslogs](https://datatracker.ietf.org/wg/syslog/charter/) are a standard format for message logging.
```
jachris, line 13:
```suggestion
[Syslog](https://datatracker.ietf.org/wg/syslog/charter/) is a standard format for message logging.
```

Although I'm not sure why we link to the WG here?

jachris (approved):

web/docs/formats/syslog.md
jachris: There are a few more remaining occurrences of "syslogs". You can also leave it in, but based on my quick searches the plural form is almost never used.

libtenzir/include/tenzir/format/syslog.hpp
eliaskosunen: Nothing major, but since these fields are not in the same format as in the newer RFC, I picked different names for them. New-RFC `ts` is a `time_type`, old-RFC `timestamp` is a `string_type` containing a wonky timestamp. `hostname` is strictly defined by the new-RFC, but old-RFC `host` seems to also sometimes be wonky (like the `hostname/a.b.c.d` we had in our example data). I can also change them to have the same name, since the schema name is already a differentiating factor.

libtenzir/builtins/formats/syslog.cpp
eliaskosunen: I think this is fine for now. It's hard to decipher whether something is a oddly-structured syslog message, or something that really isn't a syslog message at all. Since it seems that those former cases are quite prevalent, it'd be a shame to throw those in the trash.

libtenzir/builtins/formats/syslog.cpp
eliaskosunen: Sure.

web/docs/formats/syslog.md
```diff
@@ -0,0 +1,66 @@
...
+With this input, the parser will produce the following output, with the schema name `syslog.rfc3164`:
+
+```json
+{
+  "facility": 4,
+  "severity": 2,
+  "timestamp": "Nov 16 14:55:56",
+  "host": "mymachine",
+  "tag": "PROGRAM",
+  "content": "Freeform message"
```
dominiklohmann, line 56: I'm curious: Is there a reason why it's "content" for RFC 3164, and "message" RFC 5424?

eliaskosunen: That's following the RFCs: in 3164, the `message` field is divided into two parts, `tag` and `content`. In 5424, the information previously conveyed by the tag is given in the header. To quote Wikipedia:

> In RFC 3164, the message component (known as MSG) was specified as having these fields: TAG, which should be the name of the program or process that generated the message, and CONTENT which contains the details of the message. 
> 
> Described in RFC 5424, "MSG is what was called CONTENT in RFC 3164. The TAG is now part of the header, but not as a single field. The TAG has been split into APP-NAME, PROCID, and MSGID. This does not totally resemble the usage of TAG, but provides the same functionality for most of the cases."

eliaskosunen: That's following the RFCs: in 3164, the `message` field is divided into two parts, `tag` and `content`. In 5424, the information previously conveyed by the tag is given in the header. To quote Wikipedia:

> In RFC 3164, the message component (known as MSG) was specified as having these fields: TAG, which should be the name of the program or process that generated the message, and CONTENT which contains the details of the message. 
> 
> Described in RFC 5424, "MSG is what was called CONTENT in RFC 3164. The TAG is now part of the header, but not as a single field. The TAG has been split into APP-NAME, PROCID, and MSGID. This does not totally resemble the usage of TAG, but provides the same functionality for most of the cases."

dominiklohmann: Thanks for the explanation, it does make sense to follow the RFCs here. Should we perhaps add an info box to the docs explaining this (I'm not sure)?

libtenzir/builtins/formats/syslog.cpp
```diff
@@ -0,0 +1,117 @@
...
+          .hint(fmt::format("line: `{}`", *line))
+          .emit(ctrl.diagnostics());
+        co_return;
+      }
+    }
+    if (prev_builder != builder && prev_builder)
+      co_yield prev_builder->finish();
+  }
+  if (builder)
+    co_yield builder->finish();
```
jachris, line 80: There is one more thing to consider here: If you get a long stream of let's say only RFC 5242 events, then you will only flush the builders when the incoming data ends. And if it is a continuous stream that does not end (e.g., TCP stream), then you will _never_ flush. We use 2 mechanisms to solve this problem: (1) A maximum batch size (`defaults::import::table_slice_size`), (2) and a maximum batch timeout (`defaults::import::batch_timeout`). You can take some inspiration from the other parsers.

eliaskosunen: I did, in fact, take inspiration here from the CEF parser, where this is also not implemented, and where I copied the TODO comment from :smile: But sure, let's make this work properly without any TODOs

libtenzir/include/tenzir/format/syslog.hpp
jachris: You bring up some valid points. I'm not sure if changing types is better with or without changing field names. Perhaps we have to see how this will play out in practice. I'd assume most people will only receive a single syslog type (though it's great to have the functionality for accepting both!). Having said that, I think renaming to `hostname` is fine, given that the RFC calls this field `HOSTNAME` (but I agree that it is wonky). I would also rename the new-RFC `ts` to `timestamp`. But that's just my opinion, go with whatever you prefer :+1:

eliaskosunen: Yeah, I'll do that.

libtenzir/builtins/formats/syslog.cpp
dominiklohmann: I think the cleanest implementation of this is in the zeek-tsv parser if you're looking for inspiration

libtenzir/include/tenzir/format/syslog.hpp
```diff
@@ -250,6 +254,131 @@ struct message_parser : parser_base<message_parser> {
...
+      {"version", uint64_type{}},
+      {"timestamp", time_type{}},
+      {"hostname", string_type{}},
+      {"app_name", string_type{}},
+      {"process_id", string_type{}},
+      {"message_id", string_type{}},
+      {"structured_data",
+       map_type{type{"id", string_type{}},
+                type{"params", map_type{type{"key", string_type{}},
+                                        type{"value", string_type{}}}}}},
```
jachris, line 360: You can leave the re-introduction of `structure_data` to a follow-up PR, in particular because the `map` type should not be used anymore, see https://github.com/tenzir/issues/issues/25. So this either requires upgrading to the new `series_builder` or using a `{key: ..., value: ...}` workaround.

libtenzir/builtins/formats/syslog.cpp
mavam: NB: CEF over syslog is pretty common. Might be worth adding as integration test.

eliaskosunen: I added a test case with a CEF payload. I couldn't really figure out how to actually parse the CEF as a part of the pipeline: `read syslog` gives me a record with the message in one of its fields, but I can't exactly pipe that to `read cef`. Might be something for the `parse` operator to do.

mavam: > Might be something for the `parse` operator to do.

This. @jachris and I agreed on the UX for `parse` yesterday. The idea is that you'll be able to do:

```
load tcp://1.2.3.4
| read syslog
| parse message cef
```

Exact syntax TBD. The corollary is that `parse` takes a parser as argument, and that `kv`, `dissect`, and `grok` should also become parsers.

PR #3646, opened 2023-11-15
Add `kv` parser

jachris:
The `kv` parser splits strings into key-value pairs.

### DIFF
--- a/libtenzir/builtins/formats/kv.cpp
+++ b/libtenzir/builtins/formats/kv.cpp
@@ -1,0 +1,188 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The Tenzir Contributors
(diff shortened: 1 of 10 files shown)


libtenzir/builtins/formats/kv.cpp
```diff
@@ -0,0 +1,197 @@
...
+    TENZIR_ASSERT_CHEAP(regex_->NumberOfCapturingGroups() == 1);
+    auto group = std::string_view{};
+    // Ideally, we would just pass `&group` to re2. However, this does not
+    // compile depending on the version. Same goes for `absl::string_view*`.
+    auto group_arg = re2::RE2::Arg{
+      &group,
+      [](const char* str, size_t n, void* dest) -> bool {
+        *static_cast<std::string_view*>(dest) = {str, n};
+        return true;
+      },
```
eliaskosunen, line 72: `re2::StringPiece` will work regardless of re2+absl version (see `grok.cpp`).

eliaskosunen (approved):

libtenzir/builtins/formats/kv.cpp
jachris: Thanks! :pray:

PR #3650, opened 2023-11-16
Support `show`-ing all aspects at once

dominiklohmann:
This makes the aspect in `show <aspect>` optional. When the aspect is not specified, all aspects are used automatically.

image

### DIFF
--- a/libtenzir/builtins/operators/show.cpp
+++ b/libtenzir/builtins/operators/show.cpp
@@ -20,7 +20,7 @@ namespace tenzir::plugins::show {
 namespace {
 
 struct operator_args {
-  located<std::string> aspect;
+  std::optional<located<std::string>> aspect;
 
   friend auto inspect(auto& f, operator_args& x) -> bool {
     return f.object(x)
@@ -39,7 +39,18 @@ public:
 
   auto operator()(operator_control_plane& ctrl) const
     -> generator<table_slice> {
-    return get()->show(ctrl);
+    if (auto plugin = get()) {
+      return plugin->show(ctrl);
+    }
+    return
+      [](operator_control_plane& ctrl,
+         std::vector<const aspect_plugin*> plugins) -> generator<table_slice> {
+        for (const auto* plugin : plugins) {
+          for (auto&& slice : plugin->show(ctrl)) {
+            co_yield std::move(slice);
+          }
+        }
+      }(ctrl, get_all());
   }
 
   auto name() const -> std::string override {
@@ -51,7 +62,10 @@ public:
   }
 
   auto location() const -> operator_location override {
-    return get()->location();
+    if (const auto* plugin = get()) {
+      return plugin->location();
+    }
+    return operator_location::remote;
   }
 
   auto optimize(expression const& filter, event_order order) const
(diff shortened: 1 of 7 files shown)


web/docs/operators/sources/show.md
```diff
@@ -45,6 +45,8 @@ look at what's going on:
   available from the `/serve` API endpoint.
 - `types`: shows all known types at a remote node.
 
+When no aspect is not specified, all are shown.
```
jachris, line 48:
```suggestion
When no aspect is specified, all are shown.
```

web/docs/operators/sources/show.md
```diff
@@ -72,3 +74,9 @@ Show the version of a remote node:
 ```
 remote show version
 ```
+
+Show all aspects of a remote node:
+
+```
+show
+```
```
jachris, line 82: This nicely illustrates the confusion.

libtenzir/builtins/operators/show.cpp
```diff
@@ -51,7 +62,10 @@ class show_operator final : public crtp_operator<show_operator> {
   }
 
   auto location() const -> operator_location override {
-    return get()->location();
+    if (const auto* plugin = get()) {
+      return plugin->location();
+    }
+    return operator_location::remote;
```
jachris, line 68: I find this a bit confusing.

jachris (approved):
+1

web/docs/operators/sources/show.md
dominiklohmann: I think the problem more generally is that an operator should not set its location based on its arguments. Changing this to always be remote would simplify the operator a lot, but it breaks `tenzir 'show version'` without a node—which I think would be fine.

@mavam wdyt?

mavam: >I think the problem more generally is that an operator should not set its location based on its arguments.

I lack understanding to comment on that, but...

> Changing this to always be remote would simplify the operator a lot, but it breaks tenzir 'show version' without a node—which I think would be fine.

...I find it more helpful to show the version of the `tenzir` binary than the remote node.

In general, for the CLI to be remain flexible, I would argue in the opposite direction: if possible, do *not* require a node execution context.

mavam:
Nice. A great example for the multi-schemaness in our pipelines!

web/docs/operators/sources/show.md
dominiklohmann: The problem is that it's hard to document and to explain the expected behavior to the user. When you use `show`, the version aspect is run remotely. When you use `show version`, it is run locally. If I understood Jannis correctly, then the both of us think that this is a confusing user experience.

dominiklohmann: Operator locations themselves are already a hard-to-grok topic, but if an operator can dynamically change its location like the `show` operator than it's hard to understand what's going on.

mavam: I agree that this is confusing, and (now) also that location should be independent of arguments.

I still find it essential to not lose the ability to show the local version. Would anything speak against introducing `--version` to the executable? The expectation is very clear hear: just show the version of the binary.

jachris: I find it okay-ish that the location depends on the argument. That `show version` and `show` use different locations to retrieve the version is what's confusing to me, and the same goes for every other anywhere/remote aspect. I guess most of the times the result is the same. However, `show operators` for example can now often change with `disable-plugins`.

dominiklohmann: Are there any aspects other than `version` for which we would need a local alternative?

jachris: If you use `tenzir` without a node, then you'd probably want `show operators` to show the local operators, for example.

dominiklohmann: My current intuition is to just merge this as-is. I think as soon as we work on merging results of multiple operators into one (e.g., via nested pipelines) this topic will come up again anyways.

Wdyt, yes/no?

jachris: Perhaps you just remove the "show the version of a remote node" example and we'll just glance over the fact that this is a bit inconsistent.

dominiklohmann: Done and pushed. Resolving this for now.

PR #3651, opened 2023-11-16
Add `yield` operator

jachris:
The remaining todos are mostly something for the language revamp. Also, the operator is intentionally a bit under-specified because we are not sure yet how exactly nulls, unknown fields, etc., will be handled eventually.

### DIFF
--- a/libtenzir/builtins/operators/yield.cpp
+++ b/libtenzir/builtins/operators/yield.cpp
@@ -1,0 +1,217 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2021 The Tenzir Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <tenzir/argument_parser.hpp>
+#include <tenzir/pipeline.hpp>
+#include <tenzir/plugin.hpp>
+
+#include <arrow/api.h>
+
+#include <ranges>
+#include <string_view>
+
+namespace tenzir::plugins::yield {
+
+namespace {
+
+struct unnest {};
+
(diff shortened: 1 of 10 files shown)


web/docs/operators/transformations/yield.md
```diff
@@ -0,0 +1,31 @@
...
+## Examples
+
+The schema `suricata.dns` provides a list of answers for DNS queries. We want
+to extract all answers for `CNAME` records.
+
+```
+from eve.json |
+where #schema == "suricata.dns" |
+yield dns.answers[] |
+where rrtype == "CNAME"
```
mavam, line 30:
```suggestion
from eve.json
| where #schema == "suricata.dns"
| yield dns.answers[]
| where rrtype == "CNAME"
```

I know why you put the pipe last, but this is consistent with the rest of the examples. 😉

libtenzir/builtins/operators/yield.cpp
```diff
@@ -0,0 +1,217 @@
...
+public:
+  yield_operator() = default;
+
+  explicit yield_operator(parser_interface& p) {
+    auto parser = argument_parser{"yield", "https://docs.tenzir.com/next/"
+                                           "operators/transformations/yield"};
+    auto extractor = located<std::string>{};
+    // TODO: This assumes that the given extractor can be parsed as a shell-like
+    // argument, e.g., spaces must be quoted.
+    parser.add(extractor, "<extractor>");
```
dominiklohmann, line 40: I think we need to discuss whether we want `.'foo bar'.baz` or `.["foo bar"].baz`. Both cases are also a strong argument for requiring fields to start with a `.` to make parsing easier.

I think we can postpone this and merge the operator regardless, though.

dominiklohmann (approved):
This is great, thank you.

libtenzir/builtins/operators/yield.cpp
jachris: Agreed. Let's make this part of the language revamp.

PR #3652, opened 2023-11-17
Fix exporter metrics (again)

dominiklohmann:


tobim (approved):
This should finally be watertight.

PR #3653, opened 2023-11-17
Add file extension detection to `from`/`to`

eliaskosunen:
This PR implements file extension detection for the `from` and `to` operators, with support for all loaders/savers (not just `file`).

 * If the file path looks like it has an extension which would indicate compression, `(de)compress` is automatically added to the pipeline.
 * If no `read`/`write` argument is given, the file extension is used to determine the parser/printer to use. If that doesn't work, we'll fall back on the loader/saver's default (the old behavior).

In practice, for example `from file.json.gz | to otherfile.yaml.zst` becomes:
```
load file.json.gz
| decompress gzip
| read json
| write yaml
| compress zstd
| save otherfile.yaml.zst
```


jachris (commented):
Thank you, this is great! ✨

changelog/next/features/3653--file-extension-loader.md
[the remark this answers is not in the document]
eliaskosunen: Looks like my IDE was configured for 120 columns for `.md` files. Fixing.

libtenzir/src/detail/loader_saver_resolver.cpp
```diff
@@ -180,48 +180,182 @@ auto resolve_impl(parser_interface& parser, const located<std::string>& name,
...
     if (auto plugin = plugins::find<ParserPlugin>(name.inner)) {
       // Matches a plugin name, use that
-      return {parse(plugin, parser), make_located_string_view(name), false};
+      auto arg = parser.accept_shell_arg();
+      if (!arg) {
+        // No argument after the plugin name
+        return {parse(plugin, parser), name, {}, false};
+      }
+      auto r = prepend_token{*arg, parser};
+      return {parse(plugin, r), name, *arg, false};
```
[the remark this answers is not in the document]
eliaskosunen: Fixed. I needed to do some refactoring to avoid a regression, but it should be good, now.

libtenzir/src/detail/file_path_to_plugin_name.cpp
```diff
@@ -0,0 +1,53 @@
...
+    {"eve.json", "suricata"},
+  }};
+
+constexpr auto extension_to_plugin_map
+  = std::array<std::pair<std::string_view, std::string_view>, 3>{{
+    {".ndjson", "json"},
+    {".yml", "yaml"},
+    // FIXME: Is this a good default?
+    // .log is quite broad
+    {".log", "zeek-tsv"},
```
jachris, line 28: I think this is too broad.

libtenzir/test/loader_plugin.cpp
```diff
@@ -172,20 +172,6 @@ TEST(stdin loader - one complete chunk) {
...
-              "json");
-  CHECK_EQUAL(make_loader("file", "--timeout 1s eve.json")->default_parser(),
-              "suricata");
-  CHECK_EQUAL(make_loader("file", "--timeout 1s -")->default_parser(), "json");
-  CHECK_EQUAL(make_loader("file", "--timeout 1s foo.txt")->default_parser(),
-              "json");
-  CHECK_EQUAL(make_loader("file", "--timeout 1s foo.")->default_parser(),
-              "json");
-}
-
```
jachris, line 188: This is still supposed to work, no? Maybe it does again after your latest change?

web/docs/formats.md
```diff
@@ -19,8 +19,8 @@ Formats appear as an argument to the [`read`](operators/sources/from.md) and
 [`write`](operators/sinks/to.md) operators:
 
 ```
-read <format> [from <connector>]
-write <format> [to <connector>]
+from <connector> [read <format>]
+to <connector> [write <format>]
```
jachris, line 23: Good catch (the trailing `from/to` doesn't work anymore), but the example does not fit the text above, and the text has wrong links. The example should probably show `from <connector> [read <format>]`, as well as `read <format>` (and the same for `to/write`).

libtenzir/test/loader_plugin.cpp
eliaskosunen: Oh, right, it should.

web/docs/operators/sources/from.md
```diff
@@ -15,9 +15,18 @@ from <connector> [read <format>]
...
 together a [connector][connectors] and a [format][formats].
 
-All connectors have a default format. This enables a shorter syntax, e.g.,
-`from stdin` uses the `json` format, while `from file foo.csv` uses the `csv`
-format.
+If given something that looks like a path to a file, the connector can pick
+out a format automatically based on the file extension or the file name.
+This enables a shorter syntax, e.g., `from https://example.com/file.yml`
+uses the `yaml` format. All connectors also have a default format,
+which will be used, if the format can't be determined by the path.
```
jachris, line 22:
```suggestion
which will be used if the format can't be determined by the path.
```

jachris (approved):
This seems to work perfectly fine :+1:

changelog/next/features/3653--file-extension-loader.md
```diff
@@ -0,0 +1,10 @@
+When using `from` and `to` without an explicit `read`/`write` argument,
+the default format is determined by the file extension for all loaders
+and savers, not just when using `file`, if possible.
```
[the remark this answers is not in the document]
eliaskosunen: I clarified the wording somewhat, but yes.

PR #3654, opened 2023-11-19
Update getting started docs and use cases

mavam:
This PR makes several changes to the getting started docs.


web/docs/user-guides.md
```diff
@@ -1,12 +1,12 @@
...
-Tenzir in practice.
+The user guides walk you through hands-on examples that illustrate how to use
+Tenzir.
 
 ## Datasets
 
-Throughout our guides, we use publicly available datasets for a reproducible
-experience.
+Throughout our guides, we use publicly available datasets so that you can follow
+along.
```
dominiklohmann, line 9: Not necessarily for this PR, but the instructions below should be rewritten. No need to download the files first.

dominiklohmann (approved):
This reads nicely given the target audience. I think most of the team should read this.

I just have a few smaller comments that I left inline.

I'm still undecided whether we should move the section up a bit.

web/docs/user-guides.md
mavam: Yes, good point. Also that it's a follow-up PR.

web/docs/use-cases/execute-detections-and-analytics.md
```diff
@@ -0,0 +1,171 @@
...
+The sheer volume of telemetry from endpoint, network, and cloud data sources
+makes executing detections a non-trivial challenge. If you collect petabytes of
+data, performing a full scan over the data is not just inefficient, but
+cost-prohibitive when doing so for every detection. Understanding the workloads
+that detections generate is a prerequisite to effective execution on large
+datasets.
+
+For example, an engine must handle the continuous churn of observables
+that describe the ever-changing threat landscape, and use that dataset for
+continuous live and retro matching. This "background radiation" alone puts a
```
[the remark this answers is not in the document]
mavam: Changed to "detection".

PR #3661, opened 2023-11-21
Remove unused files in docker/ directory

lava:
These were not maintained any longer, and in the process of turning into dead code.


dominiklohmann (approved):
If CI likes this then I like it. 🧹

PR #3664, opened 2023-11-21
Implement a TCP loader

dominiklohmann:
This adds a TCP loader so that we can acquire data over TCP via `load tcp://<query>`.


web/docs/connectors/tcp.md
```diff
@@ -0,0 +1,77 @@
+# tcp
+
+Loads bytes from a TCP or TLS connection.
+
+## Synopsis
+
+```
+tcp [--listen] [-k] [--tls] [--certfile] [--keyfile] <uri>
```
mavam, line 8: We should streamline this with the `zmq` connector, which uses `--connect` and `--bind`. Either this becomes `--bind` or `zmq` should be `--listen`.

tobim (approved):
Approving this a-priori, but I would like to see the docs examples as functional tests.

web/docs/connectors/tcp.md
lava: Hm, `--bind` as a boolean flag doesn't make sense to me, an argument called `--bind` should have an address as argument. Using `-l`/`-k` also has the advantage that the interface is analogous to `nc`, which is a good mental model to have for this connector.

So I tend towards renaming the `zmq` option to `--listen`, but it probably makes sense to have this in a separate PR?

PR #3665, opened 2023-11-21
Add `parse` operator

jachris:
!!! WIP !!!


libtenzir/include/tenzir/plugin.hpp
```diff
@@ -461,6 +461,16 @@ class plugin_parser {
...
 
+  /// Apply the parser to an array of strings.
+  ///
+  /// The default implementation of creates a new parser with `instantiate()`
+  /// for every single string.
+  ///
+  /// @post `input->length() == result_array->length()`
+  auto parse_strings(std::shared_ptr<arrow::StringArray> input,
+                     operator_control_plane& ctrl) const
+    -> std::vector<std::pair<type, std::shared_ptr<arrow::Array>>>;
```
eliaskosunen, line 472: Is this non-`virtual` on purpose?

jachris: Seems like I committed the fix to this in my key-value parser branch instead of this one. Fixed! :+1:

libtenzir/builtins/operators/parse.cpp
```diff
@@ -0,0 +1,213 @@
...
+        total += result->length();
+      }
+      TENZIR_ASSERT_CHEAP(total == string_array->length());
+      const auto children
+        = batch->ToStructArray().ValueOrDie()->Flatten().ValueOrDie();
+      auto next = int64_t{0};
+      for (auto& [result_ty, result] : results) {
+        auto new_fields = collect(schema.fields());
+        // TODO: Some fields could be overwritten. We currently do not check for
+        // this, which can lead to having the multiple fields with the same name.
```
dominiklohmann, line 124: One alternative to this is to (by default) replace the input field with the parsed output. Then you don't run into this problem at all, and don't need to handle this as a one-off for this operator.

jachris: Yes, I have also thought about that. We can do it that way, but (1) it seems like inserting into the root record is a common choice, and (2) this would not work if the input is not a field (although it currently must be one). But yes, I'm still very much open to changing this. @eliaskosunen What do you think makes most sense?

libtenzir/src/plugin.cpp
```diff
@@ -636,6 +639,84 @@ auto aspect_plugin::aspect_name() const -> std::string {
...
 
+// -- parser plugin ------------------------------------------------------------
+
+auto plugin_parser::parse_strings(std::shared_ptr<arrow::StringArray> input,
+                                  operator_control_plane& ctrl) const
+  -> std::vector<std::pair<type, std::shared_ptr<arrow::Array>>> {
+  // TODO: Collecting finished table slices here is very bad for performance.
+  // For example, we have to concatenate new table slices. But there are also
+  // many questions with regards to semantics. This should be either completely
+  // rewritten or replaced with a different mechanism after the revamp.
```
eliaskosunen, line 650: Agree, it looks bad, but if the semantics are at least workable for now, I'd say it's fine.

libtenzir/src/plugin.cpp
```diff
@@ -636,6 +639,84 @@ auto aspect_plugin::aspect_name() const -> std::string {
...
+      append_null();
+      continue;
+    }
+    auto slices = collect(std::move(*instance));
+    slices.erase(std::ranges::remove_if(slices,
+                                        [](table_slice& x) {
+                                          return x.rows() == 0;
+                                        })
+                   .begin(),
+                 slices.end());
```
eliaskosunen, line 694: `std::erase_if(slices, [](const table_slice& x) { return x.rows() == 0; });`

Free function `std::erase(_if)` does the remove+erase dance automatically (C++20)

tenzir/functional-test/data/syslog/cef-over-syslog.log
```diff
@@ -0,0 +1,10 @@
+Nov 13 16:59:59 host123 FOO: CEF:0|FORCEPOINT|Firewall|6.6.1|0|Generic|0|deviceExternalId=Master FW node 1 dvc=10.1.1.40 dvchost=10.1.1.40 msg=log server connection established deviceFacility=Logging System rt=Jan 17 2020 08:52:10
```
eliaskosunen, line 1: I don't really know which is more prevalent, but at least here: https://help.deepsecurity.trendmicro.com/12_0/aws/Events-Alerts/syslog-parsing.html the `"CEF"` is working as the syslog TAG, so there's no `"FOO: "` before the CEF body.

The `syslog` parser deals with this case correctly, but `read syslog | parse content --into message cef` wouldn't work, because `message` doesn't contain the `CEF:`. This is something to be dealt with elsewhere, nevertheless.

eliaskosunen (approved):
I left a couple of comments, but for an experimental operator, this is good. Ship it!

Sorry for taking a bit.

libtenzir/builtins/operators/parse.cpp
jachris: We decided to change this after some user feedback.

libtenzir/src/plugin.cpp
```diff
@@ -636,6 +639,84 @@ auto aspect_plugin::aspect_name() const -> std::string {
...
+    last
+      = concatenate({std::move(last), table_slice{null_batch, last.schema()}});
+  };
+  for (auto str : values(string_type{}, *input)) {
+    if (not str) {
+      append_null();
+      continue;
+    }
+    auto bytes = std::span<const std::byte>{
+      reinterpret_cast<const std::byte*>(str->data()), str->size()};
```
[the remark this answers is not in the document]
jachris: `std::as_bytes` doesn't seem to work with `std::string_view`, but `tenzir::as_bytes` does 🙂

libtenzir/src/plugin.cpp
jachris: Ah, yes, I remember now. Thanks :pray:

libtenzir/builtins/operators/parse.cpp
```diff
@@ -0,0 +1,213 @@
...
+        }
+        parser_name_ = p.accept_shell_arg();
+        if (not parser_name_) {
+          diagnostic::error("expected parser name")
+            .primary(p.current_span())
+            .throw_();
+        }
+      } else {
+        parser_name_ = std::move(into_or_name);
+      }
```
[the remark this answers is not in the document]
jachris: Agreed. We'll see what can be done with the language revamp :+1:

PR #3666, opened 2023-11-22
Add 'min_events' parameters to /serve endpoint

lava:
This adds a new parameter 'min_events' to the `/serve` endpoint, with the semantics that the endpoint will return a result as soon as `min_events` events have arrived, regardless of the configured `timeout` and `max_events`.

The idea is to allow long polling, by specifying a long timeout and a low value for `min_events`. The defaults for the endpoint are adjusted accordingly.

### DIFF
--- a/libtenzir/builtins/operators/serve.cpp
+++ b/libtenzir/builtins/operators/serve.cpp
@@ -103,6 +103,10 @@ constexpr auto SPEC_V0 = R"_(
                 type: integer
                 example: 50
                 description: The maximum number of events returned. If unset, the number is unlimited.
+              min_events:
+                type: integer
+                example: 50
+                description: Wait for this number of events before returning. If unset, the number is unlimited.
               timeout:
                 type: string
                 example: "100ms"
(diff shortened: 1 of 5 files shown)


libtenzir/builtins/operators/serve.cpp
```diff
@@ -103,6 +103,10 @@ constexpr auto SPEC_V0 = R"_(
                 type: integer
                 example: 50
                 description: The maximum number of events returned. If unset, the number is unlimited.
+              min_events:
+                type: integer
+                example: 50
+                description: Wait for this number of events before returning. If unset, the number is unlimited.
```
dominiklohmann, line 109: Please also regenerate openapi.yaml:

```bash
tenzir-ctl web openapi > web/openapi/openapi.yaml
```

dominiklohmann, line 109: Please add a changelog entry.

dominiklohmann, line 109: Can you add an example to the `serve` operator documentation that uses the new parameter?

dominiklohmann (approved):
This is fantastic. I've tested it with the app and the behavior is much better now and results arrive quicker with this approach. There's one change to the loading behavior of the app we need to make, but that's orthogonal to this change.

PR #3667, opened 2023-11-22
Allow rest plugins to define their own openAPI schemas

lava:
Previously, all schemas used in a plugins' endpoint description had to be globally defined in the unrelated web plugin.

This PR adds a new optional interface method that allows endpoint plugins to define their schemas locally.


libtenzir/include/tenzir/plugin.hpp
```diff
@@ -640,13 +640,22 @@ class language_plugin : public virtual plugin {
...
     = 0;
 
+  /// OpenAPI description of the schemas used by the plugin endpoints, if any.
+  /// @returns A `tenzir::data` object that is a record containing entries for
+  /// the `paths` element of an OpenAPI spec.
+  [[nodiscard]] virtual auto
+  openapi_schemas(api_version /*version*/ = api_version::latest) const
+    -> std::optional<data> {
+    return std::nullopt;
+  }
```
Dakostu, line 657: This is one of the rare methods in `plugins.hpp` that isn't pure virtual - should we make it pure virtual and delegate that "optionalness" to all `rest_endpoint_plugin` implementations by returning that `std::nullopt`? Or would that be too much boilerplate code?

Dakostu (approved):
Looks OK to me. I left 1 question though.

libtenzir/include/tenzir/plugin.hpp
dominiklohmann: We have quite a few methods with default implementations for plugins, especially for those plugin types that implement other plugin types. For example, the `store` plugin implements everything for `parser` and `printer` plugins.

One question I have: `data` is already nullable, so what's the difference between returning a `std::nullopt` and a `data{}` here?

plugins/web/src/specification_command.cpp
```diff
@@ -12,294 +12,24 @@ namespace tenzir::plugins::web {
...
   auto paths = record{};
+  auto schemas = record{};
   for (auto const* plugin : plugins::get<rest_endpoint_plugin>()) {
-    auto spec = plugin->openapi_specification();
+    auto spec = plugin->openapi_endpoints();
     TENZIR_ASSERT_CHEAP(caf::holds_alternative<record>(spec));
     for (auto& [key, value] : caf::get<record>(spec))
       paths.emplace(key, value);
+    if (auto schemas_spec = plugin->openapi_schemas()) {
+      TENZIR_ASSERT_CHEAP(caf::holds_alternative<record>(*schemas_spec));
```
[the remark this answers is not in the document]
lava: Good point, there was some reason we originally switched from `record` to `data`, but I can't remember what it was so I assume it became irrevelant with subsequent refactorings.

libtenzir/include/tenzir/plugin.hpp
lava: Imho `std::optional<data>` is explicitly saying "this may be empty" on a type level; but I've now switched to `record` so I guess the point is moot anyways.

PR #3672, opened 2023-11-24
Add enrichment blog post

mavam:
This PR adds the first blog post in our contextualization series, scheduled to be published on Monday.


dominiklohmann (approved):
I think this reads nicely. I'm a lot more excited about where this series is headed than this first entry, though.

mavam:
> I think this reads nicely. I'm a lot more excited about where this series is headed than this first entry, though.

Then you should read [part two](https://github.com/tenzir/tenzir/pull/3673) 😉

PR #3681, opened 2023-11-28
Support comments in xsv parser

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1021

Add a new flag to `xsv` (incl. `csv`, `tsv` and `ssv`): `--allow-comments`. When enabled, lines starting with the `'#'` character are skipped. Comments at any other point in a line but the beginning, and comments marked with characters other than `'#'` are not supported.

To prevent a regression, this behavior is opt-in.


dominiklohmann (approved):
🙏

PR #3682, opened 2023-11-28
Build plugins in bundled mode on macOS

tobim:
We often run into problems when a plugin job runs on a different macOS action runner image than the main tenzir build. In that case some of the linked dependencies may have a different `soversion`, which causes the tests to fail.

This change moves all plugins to build in bundled mode, sidestepping the problem.

### DIFF
--- a/.github/workflows/tenzir.yaml
+++ b/.github/workflows/tenzir.yaml
@@ -609,6 +609,19 @@ jobs:
             cxx: g++-12
             dependencies-script-path: scripts/debian/install-dev-dependencies.sh
             cmake-extra-flags: -DTENZIR_ENABLE_BUNDLED_CAF:BOOL=ON
+            bundled-plugins:
+              # We enable the web and fluent-bit plugins here, because:
+              #  * web: Is needed as a dependency of the pipeline-manager, and
+              #    plugin dependency graphs cannot the modeled with the current
+              #    plugin API.
+              #  * fluent-bit: The plugin library links to libfluent-bit.so, which
+              #    is built by upstream without support for `dlopen()`, the
+              #    integrated build method used here works around that by linking
+              #    libfluent-bit.so to libtenzir.so directly. The alternative
+              #    workaround of using LD_PRELOAD for the standalone plugin build
+              #    does not work in the GitHub Actions Runner.
+              - plugins/web
+              - plugins/fluent-bit
           - os: macos-latest
             container: null
             name: macOS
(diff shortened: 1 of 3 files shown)


mavam (approved):

.github/workflows/tenzir.yaml
```diff
@@ -617,6 +620,20 @@ jobs:
             cxx: clang++
             dependencies-script-path: scripts/macOS/install-dev-dependencies.sh
             cmake-extra-flags: -DTENZIR_ENABLE_BUNDLED_CAF:BOOL=ON
+            bundled-plugins:
+              # rabbitmq-c is not packaged in homebrew.
+              #- plugins/amqp
```
dominiklohmann, line 625:
```
❯ brew info rabbitmq-c
==> rabbitmq-c: stable 0.13.0 (bottled), HEAD
C AMQP client library for RabbitMQ
https://github.com/alanxz/rabbitmq-c
/opt/homebrew/Cellar/rabbitmq-c/0.13.0 (35 files, 779.7KB) *
  Poured from bottle on 2023-11-06 at 16:43:12
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/r/rabbitmq-c.rb
License: MIT
==> Dependencies
Build: cmake ✔, pkg-config ✔, xmlto ✘
Required: openssl@3 ✔, popt ✔
==> Options
--HEAD
	Install HEAD version
==> Analytics
install: 176 (30 days), 587 (90 days), 1,650 (365 days)
install-on-request: 132 (30 days), 441 (90 days), 1,327 (365 days)
build-error: 0 (30 days)
```

.github/workflows/tenzir.yaml
```diff
@@ -609,6 +609,9 @@ jobs:
             cxx: g++-12
             dependencies-script-path: scripts/debian/install-dev-dependencies.sh
             cmake-extra-flags: -DTENZIR_ENABLE_BUNDLED_CAF:BOOL=ON
+            bundled-plugins:
+              - plugins/web
+              - plugins/fluent-bit
```
dominiklohmann, line 614: I know why for the web plugin, but why is the fluent-bit plugin bundled?

dominiklohmann (approved):

.github/workflows/tenzir.yaml
tobim: I forgot to move the comment around.

PR #3683, opened 2023-11-28
Add `grok` parser

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1022.

Adds a new parser, `grok`, reachable by the `parse` operator, based on the Logstash plugin of Elasticsearch of the same name: https://www.elastic.co/guide/en/logstash/current/plugins-filters-grok.html. Its syntax is
```
grok <input_pattern> [--pattern_definitions <patterns>] [--indexed_captures] [--include-nulls] [--include-unnamed]
```
where `<input_pattern>` is the grok-pattern used for parsing. Other options:
 * `--pattern-definitions` can be used to define additional patterns for the use by `<input_pattern>`, alongside the [built-in ones](https://github.com/logstash-plugins/logstash-patterns-core/tree/main/patterns/ecs-v1), that are included as-is.
 * `--indexed-captures` can be used to include all captures in the result, not just the named ones.
 * `--include-unnamed` includes unnamed captures (like `%{URI}`) in the output.

Examples:
```
# GREEDYDATA is a built-in pattern, meaning (.*)
> show version | parse version grok "%{GREEDYDATA:version}"
{
  "version": {
    "version": "v4.5.0-71-gae887a0ca3-dirty"
  }
}

# Complex custom pattern, with casting (%{...:int})
> show version
> | parse version grok
>      "v%{INT:major:int}\.%{INT:minor:int}\.%{INT:patch:int}(-%{INT:tweak:int}-%{DATA:ref}(-%{DATA:extra})?)?"
{
  "version": {
    "major": 4,
    "minor": 5,
    "ref": "gae887a0ca3",
    "extra": "dirty",
    "patch": 0,
    "tweak": 71
  }
}

# Additional pattern definitions + regex named captures
> show version
> | parse version grok
>      --pattern-definitions "VERSION_TRIPLET %{INT:major:int}\.%{INT:minor:int}\.%{INT:patch:int}"
>      "v%{VERSION_TRIPLET}(-(?<leftovers>.*))?"
{
  "version": {
    "major": 4,
    "minor": 5,
    "patch": 0,
    "leftovers: "71-gae887a0ca3-dirty"
  }
}

# URI is a complicated built-in pattern
# None of the captures inside %{URI} are named, so adding --include-unnamed
> ... | put url="https://example.com/test.txt?foo=bar"
>     | parse url grok --include-unnamed "%{URI}"
{
  "uri": {
    "URIQUERY": "foo=bar",
    "URIHOST": "example.com",
    "HOSTNAME": "example.com",
    "URI": "https://example.com/hello.txt?foo=bar",
    "IPORHOST": "example.com",
    "URIPATH": "/hello.txt",
    "URIPROTO": "https"
  }
}
```

### DIFF
--- a/libtenzir/builtins/operators/grok.cpp
+++ b/libtenzir/builtins/operators/grok.cpp
@@ -1,0 +1,524 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The Tenzir Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <tenzir/argument_parser.hpp>
+#include <tenzir/arrow_table_slice.hpp>
+#include <tenzir/plugin.hpp>
+#include <tenzir/series_builder.hpp>
+
+// Both Boost.Regex and RE2 are used:
+//  - Boost.Regex is used for actual grokking
+//  - RE2 is used for parsing the patterns we're given
+//
+// RE2 can't be used for both, because it doesn't support all the regex features
+// the built-in patterns need.
+//
+// Boost.Regex _could_ be used for both, but it's slow, so we're using RE2 where
+// we can.
+#include <boost/regex.hpp>
+#include <caf/make_copy_on_write.hpp>
+#include <re2/re2.h>
+
+#include <algorithm>
+#include <ranges>
+#include <span>
+
+namespace caf {
+template <>
+struct inspector_access<boost::regex> : inspector_access_base<boost::regex> {
+  template <typename Inspector>
+  static auto apply(Inspector& f, boost::regex& x) {
+    auto str = x.str();
+    auto result = f.apply(str);
+    if constexpr (Inspector::is_loading)
+      x.assign(str);
+    return result;
+  }
+};
+} // namespace caf
+
+namespace tenzir::plugins::grok {
+
+namespace {
+
+constexpr std::string_view builtin_patterns_strings[] =
+#include "./grok-patterns/patterns.inc"
+  ;
+
+enum class capture_type {
+  // The three options below (string, int, float)
+  // are grok standard options
+
+  // %{::string}, default
+  string,
+  // %{::int}
+  integer,
+  // %{::float}
+  floating,
+
+  // Replacement fields without a NAME, only a SYNTAX
+  unnamed,
+
+  // "implicit" named regex capture group,
+  // without %{...},
+  // but with (?<NAME>...) or (?'NAME'...)
+  // Used as a tag: in terms of type conversion, this is the same as `string`
+  implicit,
+};
+
+template <typename Inspector>
+bool inspect(Inspector& f, capture_type& x) {
+  return detail::inspect_enum_str(
+    f, x, {"string", "integer", "floating", "unnamed", "implicit"});
+}
+
+struct pattern_store;
+
+struct pattern {
+  pattern() = default;
+
+  explicit pattern(std::string p) : raw_pattern(std::move(p)) {
+  }
+
+  // Resolve this pattern, using `patterns` pattern store.
+  //
+  // Replaces %{replacement fields} in `raw_pattern` with the corresponding
+  // pattern found in `patterns`. Stores the complete regex in
+  // `resolved_pattern`, and all the found named captures (both implicit regex
+  // ones like `(?<NAME>EXPRESSION)` and replacement fields) in `named_captures`.
+  void resolve(const pattern_store& patterns, bool allow_recursion);
+
+  friend auto inspect(auto& f, pattern& x) -> bool {
+    return f.object(x)
+      .pretty_name("grok_pattern")
+      .fields(f.field("raw", x.raw_pattern),
+              f.field("resolved", x.resolved_pattern),
+              f.field("named_captures", x.named_captures));
+  }
+
+  // The grok pattern itself
+  std::string raw_pattern;
+  // Resolved regex
+  std::optional<boost::regex> resolved_pattern{std::nullopt};
+  // List of all the named captures in `resolved_pattern`
+  std::vector<std::pair<std::string, capture_type>> named_captures{};
+};
+
+struct pattern_store : public caf::ref_counted {
+  pattern_store() = default;
+
+  explicit pattern_store(std::span<const std::string_view> input) {
+    add(input);
+  }
+
+  auto copy() const -> pattern_store* {
+    return new pattern_store{patterns};
+  }
+
+  void add(std::span<const std::string_view> input) {
+    for (auto&& in : input) {
+      for (auto&& line : detail::split(in, "\n")) {
+        parse_line(line);
+      }
+    }
+    resolve_all();
+  }
+
+  void add(std::string_view input) {
+    for (auto&& line : detail::split(input, "\n")) {
+      parse_line(line);
+    }
+    resolve_all();
+  }
+
+  void resolve_all() {
+    for (auto& [name, pattern] : patterns) {
+      if (pattern.resolved_pattern)
+        continue;
+      pattern.resolve(*this, true);
+    }
+  }
+
+  friend auto inspect(auto& f, pattern_store& x) -> bool {
+    return f.apply(x.patterns);
+  }
+
+  std::unordered_map<std::string, pattern> patterns{};
+
+private:
+  pattern_store(const std::unordered_map<std::string, pattern>& other)
+    : patterns(other) {
+  }
+
+  void parse_line(std::string_view line);
+};
+
+void pattern::resolve(const pattern_store& patterns, bool allow_recursion) {
+  auto is_escaped = []<typename T>(const T* pattern_begin,
+                                   const T* match_begin) {
+    int backslash_count = 0;
+    auto end = std::make_reverse_iterator(pattern_begin);
+    for (auto it = std::make_reverse_iterator(match_begin); it != end; ++it) {
+      if (*it != '\\')
+        break;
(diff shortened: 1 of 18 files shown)


libtenzir/builtins/operators/grok_patterns.inc
jachris, line 1: Where were these patterns copied from? 

Seems like https://github.com/logstash-plugins/logstash-patterns-core/ is the original source, which is licensed under Apache 2.0 and thus requires attribution.

eliaskosunen: Yes, I'll need to do that.

tenzir/integration/reference/parse-with-grok/step_00.ref
```diff
@@ -0,0 +1,5 @@
+{
+  "version": {
+    "version": "v4.5.0-71-gae887a0ca3-dirty"
```
mavam, line 3: Won't this fail in future versions and only work with this particular commit?

eliaskosunen: There's a `replace version` after the `show version`, there, just so this test is reproducible.

mavam: Ah, I was just curious because I'm commenting on a checked-in baseline step.

libtenzir/builtins/operators/grok.cpp
```diff
@@ -0,0 +1,518 @@
...
+        builder.null();
+        continue;
+      }
+      boost::cmatch matches{};
+      if (not boost::regex_match(string->begin(), string->end(), matches,
+                                 *input_pattern_.resolved_pattern)) {
+        diagnostic::warning("pattern could not be matched")
+          .hint("input: `{}`", string)
+          .hint("pattern: `{}`", input_pattern_.resolved_pattern->str())
+          .emit(ctrl.diagnostics());
```
jachris, line 413: This could be very noisy (in particular if you change the `return` to `continue`, which it arguably should be), but I guess that's fine for now, given that we currently only focus on the happy path of `parse`. You could perhaps even make this an `error`.

libtenzir/builtins/operators/grok.cpp
```diff
@@ -0,0 +1,518 @@
...
+        return {};
```
jachris, line 414: The implementation of `parse` assumes that the total length of the returned arrays is exactly the length of the input array. Again, this is on the not-so-happy path, but we should probably do _something_ to not run into the assertion. Perhaps check for empty result and error in `parse`, WDYT?

libtenzir/builtins/operators/grok.cpp
```diff
@@ -0,0 +1,518 @@
...
+    : patterns_(get_builtin_pattern_store()) {
+    auto parser
+      = argument_parser{"grok", "https://docs.tenzir.com/next/operators/"
+                                "transformations/grok"};
+    parser.add(input_pattern_.raw_pattern, "<input_pattern>");
+    std::optional<std::string> pattern_definitions{};
+    parser.add("--pattern-definitions", pattern_definitions, "<patterns>");
+    parser.add("--indexed-captures", indexed_captures_);
+    parser.add("--include-unnamed", include_unnamed_);
+    parser.add("--include-nulls", include_nulls_);
```
jachris, line 374: I'm not sure whether the `--include-nulls` option makes sense. Nulls would be included anyway if at least one of the strings in a batch matches the pattern. This is also one of the reasons why we generally want to treat `null` and absence the same.

libtenzir/builtins/operators/grok.cpp
jachris, line 1: Can you add some lightweight docs?

jachris (commented):
This is awesome!

libtenzir/builtins/operators/grok.cpp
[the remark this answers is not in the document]
eliaskosunen: Yeah, that happens whenever the pattern doesn't match. It's the `parse` operator firing the assertion as the pattern not matching is essentially a parse error, which we don't currently handle properly.

libtenzir/builtins/operators/grok.cpp
eliaskosunen: Yeah, sure, let's make it an error to avoid everything blowing up in `parse`.

jachris: If you add a `null` it should not blow up.

libtenzir/builtins/operators/grok.cpp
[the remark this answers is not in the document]
jachris: Yep, seems like this has the same underlying reason as the other conversation. Let's resolve this one :+1:

libtenzir/builtins/operators/grok.cpp
```diff
@@ -0,0 +1,518 @@
...
+
+  // Replacement fields without a NAME, only a SYNTAX
+  unnamed,
+
+  // "implicit" named regex capture group,
+  // without %{...},
+  // but with (?<NAME>...) or (?'NAME'...)
+  // Used as a tag: in terms of type conversion, this is the same as `string`
+  implicit,
+};
```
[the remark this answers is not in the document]
eliaskosunen: I don't know. I intentionally did no modifications to the built-in patterns, to ensure portability, and to prevent myself from going down an endless rabbit hole of minute adjustments. I also don't think we should be special-casing any built-in patterns.

Possible solutions:
 1. Add an option called `infer` (or `auto`, or something else), which could try parsing a string, and see if it could be represented better as another type. This could maybe also be made the default, which would be a breaking change compared to Logstash grok, which defaults to `string`. It should also be careful: the text "null" should probably not be treated as a null value.
 2. Add an option for (some subset of) our own types, in addition to what's already there, so `bool`, `int64`, `uint64`, `double`, `duration`, `time`, `ip`, `subnet`. (I skipped `none`, `pattern`, `list` and `record`).
 3. Do both?

With (1), `%{IP:addr}` would output an `ip` if we made `infer` the default, otherwise `%{IP:addr:infer}` would do the trick. With (2), `%{IP:addr}` would keep producing a string, and `%{IP:addr:ip}` would yield an `ip`.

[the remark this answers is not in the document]
eliaskosunen: It's also somewhat difficult to programmatically ascertain what type the pattern would map to. For example, the pattern `%{DATE_EU}` (defined as `%{MONTHDAY}[./-]%{MONTHNUM}[./-]%{YEAR}`) clearly _means_ a date, but without additionally telling the parser that it's a date in the format `DD(.|/|-)MM(.|/|-)YYYY`, there's no way to parse it reliably.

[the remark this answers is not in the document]
eliaskosunen: This is why the optimal solution would probably be to only do limited type inference/casting in `grok`, and leave the rest for other operators. Our revamped language is starting to sound more grandiose by the second.

```
# extremely hypothetical
... | parse grok --include-unnamed "%{DATE_EU}" | parse timestamp f"{self.YEAR}-{self.MONTHNUM}-{self.MONTHDAY}"
```

libtenzir/builtins/operators/grok.cpp
eliaskosunen: I guess it should be placed under `formats/`, where all other formats (=parsers) live, even though it feels kind of weird.

jachris: Agreed. I'm not a fan of this. In general, I think streaming parsers and parsers like this should be more separate (and they might be separated with the language revamp). I guess we could also add the docs for `kv` and `grok` to the `parse` docs for now?

libtenzir/builtins/operators/grok.cpp
```diff
@@ -0,0 +1,517 @@
...
+    return "grok";
+  }
+
+  auto
+  instantiate(generator<chunk_ptr> input, operator_control_plane& ctrl) const
+    -> std::optional<generator<table_slice>> override {
+    (void)input;
+    diagnostic::error("`{}` cannot be used here", name())
+      .emit(ctrl.diagnostics());
+    return {};
```
dominiklohmann, line 395: I expected this to just work line-by-line, returning `tenzir.grok` events. 

If `parse_strings` takes a `generator<string_view>` and you do the `values(string_type{}, *arr)` in the calling context for it, then you basically get this for free.

eliaskosunen: This piece is actually copied from the `kv` PR: https://github.com/tenzir/tenzir/pull/3646

If we want to make the design decision that `grok` is usable outside of `parse` as a standalone operator, I can make that change.

jachris: `kv` will have to be adapted in the same way. I think this change makes sense, so +1 from me :+1:

dominiklohmann: I'd be fine with having it in a separate PR, just generally wanted to note it down. I think we should do that before including the two parsers in a release. Otherwise we'd have to document that the parsers are not usable with the `read` operator.

libtenzir/builtins/operators/grok.cpp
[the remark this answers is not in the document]
jachris: Let's leave this to a potential follow-up PR.

jachris (approved):

libtenzir/builtins/operators/grok.cpp
eliaskosunen: In that case, I'll move the docs for this under `formats/`, too.

web/docs/formats/grok.md
```diff
@@ -0,0 +1,71 @@
+# `grok`
+
+Parses a string using a `grok`-pattern, backed by regular expressions.
+
+:::caution Experimental
+This operator is experimental and may change at any time. Stay tuned!
+:::
```
mavam, line 7: I think we can remove this. Many of our operators are. 😅

PR #3687, opened 2023-11-29
Support lists and null values and empty strings in XSV parser

dominiklohmann:
This change adapts the XSV parser (and by extension the CSV, TSV, and SSV) parsers to support the same options as their respective printers for lists, null values, and empty strings—neither of which were previously supported.


libtenzir/builtins/formats/xsv.cpp
```diff
@@ -185,41 +250,41 @@ auto parse_impl(generator<std::optional<std::string_view>> lines,
...
+                  break;
+              }
+              ++f;
+              return true;
+            };
+            return detail::unescape(in, unescaper);
+          })
+          .with([](const std::string& in) {
+            return !in.empty();
+          })
```
jachris, line 278: I know that this check is not new, but I don't think it makes much sense. If we parse quoted strings, then `""` should also be allowed, otherwise `""` gets parsed as `"\"\""`.

libtenzir/builtins/formats/xsv.cpp
```diff
@@ -248,22 +314,51 @@ auto parse_impl(generator<std::optional<std::string_view>> lines,
...
     for (const auto& [field, value] : detail::zip(fields, values)) {
+      // TODO: Instead of the data parser, parse just non-container types, and
+      // respect the list separator and null value options for creating lists.
       auto parsed = data{};
-      if (not(parsers::data - parsers::pattern)(value, parsed)) {
-        parsed = value;
+      const auto value_parser
+        = (parsers::str{args.null_value} >> parsers::eoi).then([](std::string) {
+            return data{};
+          })
```
jachris, line 336: This check treats the quoted null string as a null, which I don't think we want to do?

jachris (approved):
Thank you, this is a much-needed improvement :+1:

libtenzir/builtins/formats/xsv.cpp
dominiklohmann: `parsers::str` is not a quoted string, that would be `parsers::qqstr`. For CSV, `"-"` parses as the string `-`, and `-` parses as `null`.

libtenzir/builtins/formats/xsv.cpp
dominiklohmann: I rebased after merging https://github.com/tenzir/tenzir/pull/3681 and have made this change as well.

libtenzir/builtins/formats/xsv.cpp
dominiklohmann: Ah, I now see what you mean. There was a double-parsing going on, because we used the parser for the header line to parse the values. I changed this so that we parse the values directly now, and added another test with edge cases in 5fe21af882125c0c569d81eff152f5e262d2532c.

image

jachris: Nice, thanks :pray:

PR #3690, opened 2023-11-30
"URI" -> "URL" in docs and loader_saver_resolver.cpp

eliaskosunen:
Related discussion https://github.com/tenzir/tenzir/pull/3675#discussion_r1410691025

The `s3` connector still uses the term "URI", both in the docs, and in the implementation, even though "URL" would be more precise. I left that out of this drive-by fix, since they have been there for longer than this newfangled URL-connector stuff.


lava (approved):
:ship:

PR #3692, opened 2023-12-01
Add support for macOS-style syslog messages

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1055

macOS syslog messages are otherwise RFC 3164 -compliant, except they use an ISO 8601 timestamp. This PR adds a fallback parser for timestamps to `legacy_message_parser`, which uses `time_parser`. That ought to catch a lot of cases where the timestamp is not RFC-compliant, but still quite clearly a timestamp.

```
2023-12-01 09:20:32.300854+0100  localhost iTerm2[91860]: (CoreFoundation) [com.apple.defaults:User Defaults] found no value for key AllowClipboardAccess in CFPrefsSearchListSource<0x60000304b0f0> (Domain: com.googlecode.iterm2, Container: (null))
```

Additionally, the TAG field is split into `app_name` and `process_id`, for additional user convenience. While this isn't exactly RFC-compliant, we weren't so before either, and neither is anyone else, really. A `message` field is also added, to escape the additional parsing we do to the message (splitting it into `app_name`, `process_id` and `content`, while removing whitespace and special characters between them).

These two changes essentially fix the CEF-over-syslog problem described here: https://github.com/tenzir/tenzir/pull/3665#discussion_r1410669284.


libtenzir/include/tenzir/format/syslog.hpp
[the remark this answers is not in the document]
eliaskosunen: I made an issue for multiline syslog messages: https://github.com/tenzir/issues/issues/1068. Even though technically they'd be needed for full macOS support, I think adding support for them should be done separately. Good point nevertheless, and your suggested implementation approach might be the one to take.

tenzir/integration/reference/syslog-format/step_00.ref
```diff
@@ -54,7 +54,14 @@
...
 {
-  "syslog_message": "2003-10-11T22:14:15.003Z mymachineexamplecom evntslog - ID47 [exampleSDID@32473 iut=\"5\" eventSource=\"Applic\\\\ation\" eventID=\"1011\"][examplePriority@32473 class=\"high\"]"
+  "facility": null,
+  "severity": null,
+  "timestamp": "2003-10-11T22:14:15.003000",
+  "hostname": "mymachineexamplecom",
+  "app_name": null,
+  "process_id": null,
+  "content": "evntslog - ID47 [exampleSDID@32473 iut=\"5\" eventSource=\"Applic\\\\ation\" eventID=\"1011\"][examplePriority@32473 class=\"high\"]",
+  "message": "evntslog - ID47 [exampleSDID@32473 iut=\"5\" eventSource=\"Applic\\\\ation\" eventID=\"1011\"][examplePriority@32473 class=\"high\"]"
```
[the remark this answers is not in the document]
eliaskosunen: Yeah, that's triggered by the removal of the requirement for a terminal colon `:` in the TAG. The results look reasonable, there, so I don't really have an issue with it: no data is in the wrong place or parsed incorrectly.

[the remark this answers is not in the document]
jachris: We won't attempt to parse the "structured data" (once we add this back in), but I think that could be considered acceptable given that this message is non-conforming.

tenzir/integration/tests.yaml
```diff
@@ -1223,6 +1225,10 @@ tests:
         input: data/syslog/cef-over-syslog.log
       - command: exec 'read syslog | parse content cef'
         input: data/syslog/cef-over-syslog.log
+      - command: exec 'read syslog'
+        input: data/syslog/cef-over-syslog-without-tags.log
+      - command: exec 'read syslog | parse message cef'
```
jachris, line 1230: One final thing: Can you help me understand whether we only need `message` (as opposed to having just `content`) for this very use-case, where the `CEF` tag is needed to parse it as CEF? Or do you think that this is also convenient to have in general? The duplication bothers me a bit, in particular if the "full" message can always be reconstructed from the components. What do you think?

jachris (approved):
:rocket:

tenzir/integration/tests.yaml
eliaskosunen: `message` can't strictly be reconstructed from the constituent parts: any whitespace after the colon is lost. For a `message` `FOO: bar`, `app_name` is `FOO`, and `content` is `bar`, which is the same as for a `message` `FOO:bar`.

I feel inclined to have both `message`, and `app_name`+`process_id`+`content`, at least for now, to not accidentally lose any other information.

jachris: Losing some initial whitespace seems fine to me, but I'll leave this up to you. You could also take a look at how other syslog parsers represent their results.

eliaskosunen: It seems like, in general, other syslog parsers tend to not have a `message`, but `app_name`+`process_id`+`content`. I can just remove it, it shouldn't be too hard to add it back later.

PR #3700, opened 2023-12-01
Disable regression tests for v4.6.0 and v4.6.1

dominiklohmann:
These were bad releases that did not upload correctly in CI, so the regression tests fail for them.


mavam (approved):
A short comment (like in the PR description) would help the reader of this script in the future.

PR #3701, opened 2023-12-02
Implement `timeshift` and `delay` operators

dominiklohmann:
These two operators combined allow for replaying a data set in real-time—which is really useful when giving demos.

The `timeshift [--replay-speed <replay-speed>] <field>` operator adjusts a given series of time values $`X_0, ..., X_n`$ to $`X_n' = T + \frac{X_n - X_0}{P}`$, where $`T`$ is the time where the first value arrives at the operator, and $`P`$ is a factor describing the replay speed that defaults to $`1.0`$.

The `delay <field>` operator delays events until the current time is greater or equal to the value of the last entry in the given time series.

For example, the following pipeline replays all Zeek conn logs from the M57 trace used as an example in our documentation at 100x speed.

```
from https://storage.googleapis.com/tenzir-datasets/M57/zeek-all.log.zst read zeek-tsv
| where #schema == "zeek.conn"
| timeshift --replay-speed 100 ts
| delay ts
```


mavam:
This is great and `timeshift` and `delay` work nicely together in your example.

Only request: let's make $T$ configurable so that user can anchor a time series at a given point.

dominiklohmann:
> Two observations: […]

I implemented both of these, and also a `--relative-to-first` option for `delay` that uses the first time value encountered as a baseline. Mind giving this a review?

mavam:
> I implemented both of these, and also a `--relative-to-first` option for `delay` that uses the first time value encountered as a baseline. Mind giving this a review?

I don't understand "relative to first". Here's how I'd like to think of this: my mental model is that I would like to wait by the **inter-arrival times**, which are the durations $T_{i+1} - T_i$ for every event $i$. It sounds like we're still thinking about timestamps in absolute terms? My expectation would be that I don't need to think about the first event. Conceptually, I want to "hold up" the current/last event by the time delta to it's predecessor.

If this is what actually happens, then that'd be best case and we only need to rename the option to, e.g., `--inter-arrival-times` to switch from the points on the timeline to the space between the points.

dominiklohmann:
>  Here's how I'd like to think of this: my mental model is that I would like to wait by the **inter-arrival times**, which are the durations $`T_{i+1} - T_i`$ for every event $`i`$.

That mental model has a bug, as you're actually delaying every event by $`min(0, T_{i} - T_{now} + delta)`$, where $`delta`$ is one of the following:
1. $`T_{now} - X`$ if `--relative-to X` is used
2. $`T_{now} - T_{0}`$. if `--relative-to-first` is used
3. $`0`$ otherwise.

Let's review this in a call together tomorrow, I think that's gonna be easier.

dominiklohmann:
@mavam the edit token is yours for docs.

mavam (changes requested):
This almost works, but we need one change: `--start now` is not the default for `timeshift`.

dominiklohmann:
> This almost works, but we need one change: `--start now` is not the default for `timeshift`.

If I recall correctly then we discussed this already. If we do this, then we'd need a new `--no-start` option to return to the current behavior so that you can only apply the relative shift without also shifting by an absolute value.

Right now, the options match exactly for `delay` and `timeshift`: If `--start` is not set, the first time value is used as the starting time. I'd like to keep it that way.

web/docs/operators/timeshift.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+## Description
+
+The `timeshift` operator adjusts a series of time values by anchoring them
+around a given start time.
+
+If you do not provide a start time with `--start`, the operator will anchor the
+timestamps in `field` to begin with the current wall clock time, as if you
+provided `--start now`.
```
dominiklohmann, line 24: As I explained in the comment on the thread, this behavior is incompatible with only applying a speedup. Omitting `--start` should behave just like for `delay`, where the start time is then the first non-null time value encountered.

mavam:
> Right now, the options match exactly for `delay` and `timeshift`: If `--start` is not set, the first time value is used as the starting time. I'd like to keep it that way.

That's where I am coming from as well, so that's good. 🙂 

During my testing, I invoked `timeshift ts` and I realized that it's a no-op. That's what primarily prompted my feedback. That is, `delay ts` does something meaningful whereas `timeshift ts` does not.

But I also understand that we'd need `--start-at-first` again if we want the to be able to control the speedup separately. We cannot have both.

web/docs/operators/timeshift.md
mavam: Is it correctly described in 8a0a805194?

mavam (approved):
This is ready from my perspective. Waiting for you final approval on docs and diagrams before merging, @dominiklohmann.

web/docs/operators/delay.md
```diff
@@ -0,0 +1,94 @@
...
+![Delay](delay.excalidraw.svg)
+
+The options `--start` and `--speed` work independently, i.e., you can use them
+separately or both together.
+
+### `--start <time>`
+
+The timestamp to anchor the time values around.
+
+Defaults to `now`.
```
dominiklohmann, line 44: Defaults to the first non-null timestamp in `field`.

libtenzir/builtins/operators/timeshift.cpp
```diff
@@ -0,0 +1,160 @@
...
+                                               "operators/timeshift"};
+    parser.add("--speed", speed, "<factor>");
+    parser.add("--start", start, "<time>");
+    parser.add(field, "<field>");
+    parser.parse(p);
+    if (speed and *speed <= 0.0) {
+      diagnostic::error("`--speed` must be greater than 0")
+        .note("from `{}`", name())
+        .throw_();
+    }
```
dominiklohmann, line 150: We can error here if `not start and not speed` to make the user aware that it's a no-op. I don't have a strong opinion on that.

dominiklohmann (commented):
Docs read nicely. Found just one small mistake.

libtenzir/builtins/operators/timeshift.cpp
mavam: I'd say leave it as is now.

PR #3703, opened 2023-12-03
Implement the `slice` operator

dominiklohmann:
The `slice [--begin <begin>] [--end <end>]` operator cuts events in an interval $`[begin, end)`$. A negative index counts from the end rather than from the beginning, which is a syntax familiar to users with experience in Python, jq, and many other languages.

Either `begin` or `end` can be omitted, and default to the 0 and the size of the input, respectively.

`head N` is equivalent to `slice :N`.

`tail N` is equivalent to `slice -N:`.


web/docs/operators/transformations/tail.md
```diff
@@ -13,6 +13,8 @@ tail [<limit>]
 The semantics of the `tail` operator are the same of the equivalent Unix tool:
 consume all input and only display the last *N* events.
 
+`tail <limit>` is a short-hand notation for [`slice --begin -<limit>`](slice.md).
```
jachris, line 16: Same here.

web/docs/operators/transformations/head.md
```diff
@@ -14,6 +14,8 @@ The semantics of the `head` operator are the same of the equivalent Unix tool:
 process a fixed number of events from the input. The operator terminates
 after it has reached its limit.
 
+`head <limit>` is a short-hand notation for [`slice --end <limit>`](slice.md).
```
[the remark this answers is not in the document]
dominiklohmann:
```suggestion
`head <limit>` is a shorthand notation for [`slice --end <limit>`](slice.md).
```

web/docs/operators/transformations/tail.md
dominiklohmann:
```suggestion
`tail <limit>` is a shorthand notation for [`slice --begin -<limit>`](slice.md).
```

PR #3714, opened 2023-12-06
Use systemd provided default paths if available

tobim:
We recently introduced a bug in the systemd unit that caused tenzir to discard the `db-directory` and `cache-directory` settings from `tenzir.yaml` config files.

The change here restores the precedence while making sure that the paths provided by systemd will be used as defaults in case they are not set otherwise.


libtenzir/src/configuration.cpp
```diff
@@ -444,6 +444,30 @@ caf::error configuration::parse(int argc, char** argv) {
   *config = flatten(*config);
   if (auto err = merge_environment(*config))
     return err;
+  // Fallback handling for system default paths that may be provided by the
+  // runtime system and should win over the build-time defaults but loose if set
+  // via any other method.
+  if (!config->contains("tenzir.db-directory")) {
+    // Provided by systemd when STATE_DIRECTORY= is set in the unit.
```
lava, line 451:
```suggestion
    // Provided by systemd when StateDirectory= is set in the unit.
```

libtenzir/src/configuration.cpp
```diff
@@ -444,6 +444,30 @@ caf::error configuration::parse(int argc, char** argv) {
...
+    if (auto state_directory = detail::getenv("STATE_DIRECTORY")) {
+      (*config)["tenzir.db-directory"] = std::string{*state_directory};
+    }
+  }
+  if (!config->contains("tenzir.log-directory")) {
+    // Provided by systemd when LOG_DIRECTORY= is set in the unit.
```
lava, line 457:
```suggestion
    // Provided by systemd when LogsDirectory= is set in the unit.
```

libtenzir/src/configuration.cpp
```diff
@@ -444,6 +444,30 @@ caf::error configuration::parse(int argc, char** argv) {
...
+    if (auto log_directory = detail::getenv("LOG_DIRECTORY")) {
+      (*config)["tenzir.log-file"]
+        = fmt::format("{}/server.log", *log_directory);
+      (*config)["tenzir.client-log-file"]
+        = fmt::format("{}/client.log", *log_directory);
+    }
+  }
+  if (!config->contains("tenzir.cache-directory")) {
+    // Provided by systemd when CACHE_DIRECTORY= is set in the unit.
```
lava, line 466:
```suggestion
    // Provided by systemd when CacheDirectory= is set in the unit.
```

lava (approved):
We should also adjust the documentation and/or `example.yaml` to mention the new fallbacks and add a changelog entry, but otherwise looks good to me.

PR #3718, opened 2023-12-07
Fix RFC3164 (legacy-syslog) parser expecting spaces after `<PRI>`

eliaskosunen:
Accept inputs as described in Example 1 of the RFC: https://datatracker.ietf.org/doc/html/rfc3164#section-5.4

```
<34>Oct 11 22:14:15 mymachine su: 'su root' failed for lonvick on /dev/pts/8
```

The `<34>` in the beginning (called the `PRI`-field) doesn't need to be followed by spaces, as was previously expected by the parser implementation.


changelog/next/bug-fixes/3718--legacy-syslog-pri-spaces.md
```diff
@@ -0,0 +1,3 @@
+RFC 3164 syslog parser no longer requires the `PRI`-field
+(part in <angle brackets> in the beginning of a message)
+to be followed by whitespace.
```
tobim, line 3:
```suggestion
The RFC 3164 syslog parser no longer requires a whitespace after the `PRI`-field
(part in <angle brackets> in the beginning of a message).
```

libtenzir/include/tenzir/format/syslog.hpp
```diff
@@ -328,13 +328,11 @@ struct legacy_message_parser : parser_base<legacy_message_parser> {
...
     const auto tag_parser = +(parsers::printable - ':');
     const auto content_parser = ~(ignore(*space) >> *parsers::printable);
-    // clang-format off
-    const auto p = ~(priority_parser >> wsignore)
-                   >> timestamp_parser >> wsignore    // timestamp
-                   >> word >> wsignore                // host
-                   >> tag_parser >> ':'               // tag
-                   >> content_parser;                 // content
-    // clang-format on
+    const auto p = ~(priority_parser >> ignore(*space)) // priority
```
tobim, line 331: Nit (stylistic): Whitespaces can be assumed to be `ignore`d, I would rename `wsignore` to just `ws` and create a `maybe_ws` for this one.

tobim (approved):
Thanks for the quick fix!

libtenzir/include/tenzir/format/syslog.hpp
eliaskosunen: `ws` is used without `ignore` (on purpose) above, in `timestamp_parser`. Leaving as-is, at least for now.

PR #3723, opened 2023-12-07
Include UDOs in `show operators`

dominiklohmann:
This adds user-defined operators to the output of `show operators`.


jachris (approved):
:+1:

PR #3727, opened 2023-12-11
Add TCP saver

mavam:
This PR adds a saver for the `tcp` connector.

```[tasklist]
### Tasks
- [x] Update README and fixate desired UX
- [x] Implement chunk writing in TCP bridge
- [x] Implement saver
- [x] Write changelog entry
```


dominiklohmann (approved):

tenzir/functional-test/tests/tcp.bats
```diff
@@ -20,7 +20,7 @@ setup() {
     -nodes \
     -subj "/C=US/ST=Denial/L=Springfield/O=Dis/CN=www.example.com" >/dev/null 2>/dev/null
   check --bg listen \
-    tenzir "load tcp://127.0.0.1:4000 --listen --tls --certfile ${key_and_cert} --keyfile ${key_and_cert}"
+    tenzir "load tcp://127.0.0.1:4000 --tls --certfile ${key_and_cert} --keyfile ${key_and_cert}"
   timeout 10 bash -c 'until lsof -i :4000; do sleep 0.2; done'
   echo foo | openssl s_client 127.0.0.1:4000
   wait_all "${listen[@]}"
```
lava, line 26: We should also add new tests for the saving functionality.

libtenzir/builtins/connectors/tcp.cpp
```diff
@@ -385,49 +548,60 @@ class plugin final : public virtual loader_plugin<loader> {
...
-    parser.add("-l,--listen", listen);
-    parser.add("-k,--keep-listening", keep_listening);
-    parser.add("--certfile", certfile, "TLS certificate");
-    parser.add("--keyfile", keyfile, "TLS private key");
+    if constexpr (std::is_same_v<Args, loader_args>) {
+      parser.add("-c,--connect", args.connect);
+    } else if constexpr (std::is_same_v<Args, saver_args>) {
+      parser.add("-l,--listen", args.listen);
+    }
+    parser.add("-o,--once", args.once);
```
lava, line 574: You should probably add a check that `--once` is only used in combination with `--listen`, since it's not obvious from the name anymore.

lava (commented):
Some things I noticed at a glance

libtenzir/builtins/connectors/tcp.cpp
dominiklohmann: I previously suggested calling the option `--listen-once` and having it imply `--listen`. Do you think that would be more intuitive?

tenzir/functional-test/tests/tcp.bats
mavam: Added that.

libtenzir/builtins/connectors/tcp.cpp
lava: Yes, that sounds better

mavam: We fixed the issue. We can reconsider the naming at a later point.

dominiklohmann: Since it already caused confusion for at least two people, i.e.. Benno and myself, why not rename it now?

mavam: Okay, I'll go with `--listen-once`.

mavam: See e65814f.

PR #3730, opened 2023-12-11
Restructure operator docs

mavam:
This PR changes the top-level structure of the operator docs.

```[tasklist]
### Tasks
- [x] Move operators
- [x] Fix blog post links
- [x] Fix C++ links of builtins and plugins
- [x] Consider auxiliary navigation for sources, sinks, and transformations
```


mavam:
@dan, summarizing our discussion here. TL;DR, we want something like this:

image

Details:
- We're lifting the operators into a single top-level category, to align them with how we present connectors and formats.
- We'd like to provide a visual clue in the sidebar to help users understand what types and operator is (e.g., any combination of source, sink, or transformation).
- Visually, these icons should align on the right side and be rather unobtrusive.
- I'm also attaching three example SVGs, one per operator type. We can exchange those later
- In f23119c2cd, I've added labels to the 
frontmatter of a few operators to illustrate how the docs writer would want to do this.
- I tried swizzling via `yarn docusaurus swizzle @docusaurus/theme-classic DocSidebarItem/Html`, but didn't get very far.

operator-source
operator-transformation
operator-sink

mavam:
This is technically ready to be merged, ideally sooner than later given that it streamlines the contextualization docs.

@dan it wouldn't hurt to start a new PR if this gets merged before you get the sidebar improvement.

mavam:
The categorization now looks like this:

image

web/src/theme/DocSidebarItem/Link/IconSink.svg
jachris, line 1: Perhaps you want to adjust the icons such that their "visual weight" is the same? They have the same dimensions technically, but they look differently sized.

mavam: I thought about this. Intuitively, I thought that scaling by area is optimal.

jachris (approved):

PR #3734, opened 2023-12-12
Make the catalog lookup incremental

dominiklohmann:
> [!WARNING]
> This is still under active development. Don't waste time looking at the code for this pull request yet.

This fundamentally changes how the catalog operators:
1. The catalog now maintains a single list of partitions, sorted reverse-chronologically by max-import-time instead of maintaining this list by schema. The lookup, insert, and replace operators changed accordingly.
2. The catalog no longer independently manages types and schemas, and instead infers this information from the partitions it manages. This is a first step to supporting schema evolution, as it allows the catalog to drop schemas that it no longer needs to maintain.
3. A new mechanism allows for receiving catalog lookup results one partition at a time, reducing the time to first result. We make use of this mechanism in the `export` operator; other code interfacing with the catalog still uses the old lookup mechanism, which is now implemented in terms of the new mechanism.


dominiklohmann:
I have integrated the changes from this PR into the upcoming plugin that replaces the current catalog in a cleaner way. Closing this for now.

PR #3735, opened 2023-12-13
Fix implicit discovery of named capture groups in Grok

eliaskosunen:
In regular expressions, named capture groups don't match the pattern `\(\?<.+?>`, but `\(\?<\w+>`. The former would erroneously match `(?<=` and `(?<!`, which mean lookbehind. This manifested as regex lookbehind directives being included as named capture groups.


dominiklohmann (approved):
Thanks for the additional test; the explanation for the fix makes sense as well. That was quick! 🚀

PR #3736, opened 2023-12-13
Add node health metrics

lava:


libtenzir/include/tenzir/healthmetrics_collector.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+  // list of health checks to run
+  std::unordered_map<std::string, typename health_metrics_plugin::collector>
+    collectors;
+
+  // time between checks
+  caf::timespan collection_interval = std::chrono::seconds{30};
+
+  // output stream
+  std::unique_ptr<import_stream> importer = nullptr;
```
jachris, line 42: FWIW: Feel free to replace the `import_stream` abstraction with something better if it gets in your way. If not, we can do it a later time.

libtenzir/src/healthmetrics_collector.cpp
```diff
@@ -0,0 +1,57 @@
...
+    auto collector = plugin->make_collector();
+    self->state.collectors[name] = collector;
+  }
+  return {
+    [self](atom::run) -> caf::result<void> {
+      // Start measurement loop
+      detail::weak_run_delayed_loop(
+        self, self->state.collection_interval, [self]() {
+          auto b = series_builder{
+            type{"tenzir.metrics.health", record_type{}, {{"internal", ""}}}};
```
tobim, line 38: In my opinion `tenzir.metrics.system` would make more sense for the name.

libtenzir/src/healthmetrics_collector.cpp
```diff
@@ -0,0 +1,57 @@
...
+          auto r = b.record();
```
tobim, line 39: This should include a timestamp.

libtenzir/src/healthmetrics_collector.cpp
```diff
@@ -0,0 +1,57 @@
...
+  TENZIR_ASSERT_CHEAP(stream); // FIXME error handling
+  self->state.importer = std::make_unique<import_stream>(std::move(*stream));
+  self->state.collection_interval = collection_interval;
+  for (auto const* plugin : plugins::get<health_metrics_plugin>()) {
+    auto name = plugin->name();
+    auto collector = plugin->make_collector();
+    self->state.collectors[name] = collector;
+  }
+  return {
+    [self](atom::run) -> caf::result<void> {
```
tobim, line 33: I think is suffices to start this automatically with the actor itself. One change I would make here is to run the collection once immediately before entering the loop.

libtenzir/builtins/health-metrics/memory.cpp
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  auto make_collector() const -> collector override {
+    return get_raminfo;
+  }
+
+  static auto metrics_layout() -> record_type /*override*/ {
+    return record_type{{
+      {"total_bytes", uint64_type{}},
+      {"available_bytes", uint64_type{}},
```
tobim, line 43: As long as we don't have an easy and fast way to do arithmetic in the language I would also add `used-bytes`.

tobim, line 43: Use dashes.

libtenzir/builtins/health-metrics/disk.cpp
```diff
@@ -0,0 +1,62 @@
...
+    return [dbdir = dbdir_]() {
+      return get_diskspace_info(dbdir);
+    };
+  }
+
+  static auto metrics_layout() {
+    return record_type{{
+      {"name", string_type{}},
+      {"total_bytes", uint64_type{}},
+      {"used_bytes", uint64_type{}},
```
tobim, line 50: Same here. Dashes and add a `free-bytes` column for completeness.

tobim (commented):
I have some comments after taking this for an initial spin.

libtenzir/include/tenzir/healthmetrics_collector.hpp
lava: For this use case it seems pretty well suited so far, the only crutch is that the constructor may fail so we should run it delayed, but that's more of a general CAF interface issue.

libtenzir/builtins/health-metrics/memory.cpp
jachris: The resulting events should surely use underscores instead of dashes. Thus, I think it's more consistent to go the other direction and use underscores everywhere.

libtenzir/builtins/health-metrics/cpu.cpp
dominiklohmann, line 1: Do you have a few example pipelines that make for good example charts?

libtenzir/builtins/health-metrics/process.cpp
```diff
@@ -0,0 +1,65 @@
...
+    return get_raminfo;
+  }
+
+  auto metric_name() const -> std::string override {
+    return "process";
+  }
+
+  auto metric_layout() const -> record_type override {
+    return record_type{{
+#ifdef TENZIR_LINUX
```
jachris, line 48: These platform thingies are always def-ed, either as `0` or `1`.

libtenzir/src/healthmetrics_collector.cpp
```diff
@@ -0,0 +1,101 @@
...
+        return stream.error();
+      }
+      self->state.importer
+        = std::make_unique<import_stream>(std::move(*stream));
+      // Do a one-off import immediately.
+      self->state.collect_and_import_metrics();
+      // Start measurement loop.
+      detail::weak_run_delayed_loop(
+        self, self->state.collection_interval,
+        [handle = detail::weak_handle(self)]() {
```
dominiklohmann, line 84: The weak loop will not be called if self has quit already, so you don't need to capture a weak handle additionally here.

libtenzir/src/healthmetrics_collector.cpp
```diff
@@ -0,0 +1,101 @@
...
+namespace tenzir {
+
+namespace {
+
+auto collect_metrics(
+  const healthmetrics_collector_state::collectors_map& collectors)
+  -> table_slice {
+  auto now = std::chrono::system_clock::now();
+  auto b = series_builder{
+    type{"tenzir.metrics.system", record_type{}, {{"internal", ""}}}};
```
dominiklohmann, line 26: Can you use `metrics_name()` here in the schema name and document the available options just like [`tenzir.metrics.operator`](https://docs.tenzir.com/metrics#tenzirmetricsoperator)?

libtenzir/src/healthmetrics_collector.cpp
lava: Where is this checked? Internally in caf?

dominiklohmann: Yeah, it's modeled as an action on the actor clock with a weak handle to the node, and it's only being invoked when the weak handle locks successfully. So you can safely just capture self here.

libtenzir/src/healthmetrics_collector.cpp
lava: You mean switch to one import stream per metric, as opposed to building the combined format?

dominiklohmann: My understanding is that the import stream can handle heterogeneous events, is that incorrect?

dominiklohmann: Yes, I think that's preferable, both for documentation and for usage in charts.

dominiklohmann: Unresolving as this is still missing:

> [...] and document the available options just like [`tenzir.metrics.operator`](https://docs.tenzir.com/metrics#tenzirmetricsoperator)?

web/docs/metrics.md
```diff
@@ -118,3 +118,180 @@ head 3
...
+
+### Examples
+
+Get current memory usage
+
+~~~c
+export --internal
+| where #schema == "tenzir.metrics.memory"
+| sort timestamp desc
+| head 1
```
dominiklohmann, line 285: This should likely just be `tail 1` instead.

dominiklohmann (approved):

PR #3737, opened 2023-12-14
A couple of small fixes for the python operator

tobim:
* Persist venvs between muliple `docker run tenzir '... | python "..."' invocations
* Trim python error messages
* Forward venv creation error messages in the diagnostics.


libtenzir/src/execution_node.cpp
```diff
@@ -181,6 +181,7 @@ class exec_node_control_plane final : public operator_control_plane {
       diagnostic::error("{}", error)
         .note("from `{}`", state_.op->name())
         .emit(diagnostics());
+      return;
```
dominiklohmann, line 184: This can go when you rebase, I rewrote the abort and diagnostic-emitting functions in the blocking operators fix.

Dockerfile
```diff
@@ -127,7 +134,7 @@ USER tenzir:tenzir
 
 EXPOSE 5158/tcp
 WORKDIR /var/lib/tenzir
-VOLUME ["/var/lib/tenzir"]
+VOLUME ["/var/cache/tenzir", "/var/lib/tenzir"]
```
dominiklohmann, line 137: This also needs to be added to the Docker Compose file, both in this repository and for the platform-generated one.

libtenzir/builtins/operators/python.cpp
```diff
@@ -128,10 +130,15 @@ class python_operator final : public crtp_operator<python_operator> {
...
         // list of requirements is installed correctly?
         if (!exists(venv_path, ec)) {
-          if (bp::system(python_executable, "-m", "venv", venv)) {
-            ctrl.abort(caf::make_error(ec::system_error,
-                                       "failed to create virtualenv "
-                                       "(see log for details)"));
+          // The default size of the pipe buffer is 64k on Linux, we assume
+          // (hope) that this is enough. A possible solution would be to wrap
+          // the invocation in a script that drains the pipe continuously but
+          // only forwards the first n bytes.
```
dominiklohmann, line 136: What happens if the buffer is too small?

tobim: The subprocess blocks in a call to `write`.

dominiklohmann: What does it mean for the operator? Does it stall entirely, fail, or is it just blocking the thread for a while?

tobim: I think it's very unlikely that either process writes that much to stderr.

dominiklohmann: Right, but what would that mean for the operator? In case a process does this, do we hang, crash, or fail? I think that should be explained in the comment.

dominiklohmann (approved):

libtenzir/builtins/operators/python.cpp
jachris: I wrote an issue for this: https://github.com/tenzir/issues/issues/1079

PR #3738, opened 2023-12-14
Add time parser

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1051

Adds a new parser, `time`, based on POSIX `strptime`.
```
parse <field> time [--components] [--strict] <format>
```

`<format>` is a format string compatible with `strptime`. The output is a `tenzir::time`, except if the `--components` flag is passed, the output is a record. There's some additional logic for missing fields and what their default values are, which can be controlled with `--strict`.

Deals with timezones properly, which letmetellya is something one should not call trivial.


web/docs/formats/time.md
```diff
@@ -0,0 +1,69 @@
+# `time`
```
dominiklohmann, line 1:
```suggestion
# time
```

We don't put the title into code blocks for the other formats either.

dominiklohmann, line 1: This file needs the new YAML frontmatter for categorization.

libtenzir/builtins/operators/time.cpp
```diff
@@ -0,0 +1,553 @@
...
+    if (tm_zone) {
+      add_field_if_set("utc_offset", tm_gmtoff);
+      builder.field("tz", tm_zone);
+    } else if (tm_gmtoff) {
+      builder.field("utc_offset", tm_gmtoff);
+      add_field_if_set("tz", tm_zone);
+    } else {
+      builder.field("utc_offset", 0);
+      builder.field("tz", "UTC");
+    }
```
dominiklohmann, line 342:
```suggestion
    if (tm_zone) {
      add_field_if_set("utc_offset", tm_gmtoff);
      builder.field("tz", *tm_zone);
    } else if (tm_gmtoff) {
      builder.field("utc_offset", *tm_gmtoff);
      add_field_if_set("tz", tm_zone);
    } else {
      builder.field("utc_offset", 0);
      builder.field("tz", "UTC");
    }
```

dominiklohmann (approved):
The tests look good to me. I skimmed through the code and didn't find anything major, so assuming that the tests now pass on macOS this is also ready to go.

libtenzir/builtins/operators/time.cpp
```diff
@@ -0,0 +1,567 @@
...
+namespace {
+
+constexpr auto docs = "https://docs.tenzir.com/docs/next/formats/time";
+
+std::string strerror_nonbroken(int errnum) {
+  // strerror is not thread safe, strerror_r is
+  std::string buf(64, '\0');
+  ::strerror_r(errnum, buf.data(), buf.size());
+  return buf;
+}
```
jachris, line 37: We also have `detail::describe_errno`, which does some additional error handling and resizing.

libtenzir/builtins/operators/time.cpp
```diff
@@ -0,0 +1,567 @@
...
+        //  - but --components is NOT: error, as there's not enough information
+        //                             to create a timestamp
+        const bool year_set = time->tm_year.has_value();
+        auto today_beginning = partial_timestamp::today_beginning();
+        if (not time->enrich(today_beginning, diag))
+          return {};
+        TENZIR_ASSERT(time->is_complete());
+        // A special case:
+        // If `strptime_partial` didn't set the year, but the resulting enriched
+        // time would be in the future, subtract 1 from the year.
```
jachris, line 496: I think we should allow the timestamp to be a bit in the future. Other systems do this as well (with varying allowed differences). We should also make sure that this behavior is "smooth".

libtenzir/builtins/operators/time.cpp
```diff
@@ -0,0 +1,567 @@
...
+    return f.object(x)
+      .pretty_name("time_parser")
+      .fields(f.field("format", x.format_),
+              f.field("components", x.components_),
+              f.field("strict", x.strict_));
+  }
+
+private:
+  std::string format_;
+  bool components_{false}, strict_{false};
```
jachris, line 552: Nit: We normally separate declarations.

jachris (commented):
This looks nice!

libtenzir/builtins/operators/time.cpp
```diff
@@ -0,0 +1,607 @@
...
+          if (val)
+            builder.field(name, mod(*val));
+          else
+            builder.field(name, caf::none);
+        };
+    add_field_if_set("sec", tm_sec);
+    add_field_if_set("min", tm_min);
+    add_field_if_set("hour", tm_hour);
+    add_field_if_set("day", tm_mday);
+    add_field_if_set("month", tm_mon, {[](int mon) {
```
dominiklohmann, line 384: Should we use
1. `sec`, `min`, and `mday`, or
2. `second`, `minute`, and `day`?

Not sure which is better, but right now we appear to be using a mix of spelled-out and what the underlying fields are called, which is a bit inconsistent.

eliaskosunen: Definite no for `mday`. The field is called that in `std::tm` to distinguish it from `wday` (month-day as opposed to week-day), but we have no need for that. `mday` would also indicate that the valid range would be [0, 30], and not [1, 31].

As for `sec` and `min` vs. `second` and `minute`, I have no strong feelings either way.

jachris: Then I would put my vote in for `second` and `minute`, such that all of these are complete words. 

(Also maybe `tz` -> `timezone`).

web/docs/formats/time.md
```diff
@@ -0,0 +1,75 @@
...
+### `<format>`
+
+`time` is backed by POSIX
+[`strptime`](https://man7.org/linux/man-pages/man3/strptime.3.html),
+and uses the same format string syntax, with the `"C"` locale.
+
+### `--components`
+
+Instead of a timestamp, returns a record with fields for
+`sec`, `min`, `hour`, `day`, `month`, `year`, `utc_offset`, and `tz`.
```
dominiklohmann, line 31: This still needs the full names that we switched to.

dominiklohmann (approved):
This works quite well. Thank you!

PR #3741, opened 2023-12-15
Streamline Setup and User Guide

mavam:
This PR polishes the onboarding docs.


web/docs/integrations/suricata.md
```diff
@@ -0,0 +1,44 @@
+# Suricata
```
mavam, line 1: @dominiklohmann we could add a section at the end how on simulating Suricata with Speeve, like from your demo:

```bash
~/go/bin/speeve spew --persec 100 --profile <SPEEVE_PROFILE_PATH> | docker compose run --no-TTY  tenzir 'remote save file -a /tmp/eve.sock'
```

WDYT?

dominiklohmann: I think that is better suited for a package in the library rather than as part of this integration. I'd like to focus on the tool only.

dominiklohmann (commented):
I feel like it's been way too long since we first talked about wanting to have that integration section. Really happy to finally see it come to light.

dominiklohmann (approved):

PR #3742, opened 2023-12-15
Allow empty field names

jachris:
Previously, the record `{"": 42}` was not allowed. This PR removes this restriction.

### DIFF
--- a/libtenzir/src/type.cpp
+++ b/libtenzir/src/type.cpp
@@ -61,7 +61,7 @@ constexpr size_t reserved_string_size(std::string_view str) {
   // table. It adds an extra byte because strings in FlatBuffers tables are
   // always zero-terminated, and then rounds up to a full four bytes because of
   // the included padding.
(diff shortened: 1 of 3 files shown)


libtenzir/src/type.cpp
```diff
@@ -61,7 +61,7 @@ constexpr size_t reserved_string_size(std::string_view str) {
   // table. It adds an extra byte because strings in FlatBuffers tables are
   // always zero-terminated, and then rounds up to a full four bytes because of
   // the included padding.
-  return str.empty() ? 0 : (((str.size() + 1 + 3) / 4) * 4);
+  return ((str.size() + 1 + 3) / 4) * 4;
```
dominiklohmann, line 64: Can you add a small unit test for a record with an empty field name? That should crash in CI if this isn't correct.

dominiklohmann (approved):
Thank you!

PR #3743, opened 2023-12-15
Prevent delays for blocking operators

dominiklohmann:
This change removes the outbound buffer of execution nodes, changing them to fulfill demand incrementally by immediately sending out partial results. This reduces latency, and avoids a problem with operators that block during execution, which sometimes held up already buffered partial results at the execution node instead.


jachris (approved):
We reviewed this in a call together. Great work! 🎉

PR #3758, opened 2023-12-20
Downtone push behavior of pipelines

dominiklohmann:
This PR fixes two bugs:
1. This reduces memory usage for fast operators right in front of very slow operators. Previously, the fast operator overfulfilled the demand of the slow operator, causing the inbound buffer of the slow operator to overfill. This no longer happens as drastically, and is limited to at most one batch.
2. This fixes a potential hang on shutdown due to a bug in CAF by sending an explicit exit message to operators preceding an operator that itself down early (e.g., `head`). I accidentally dropped this fix as part of the blocking operators fix.


mavam (approved):
I don't understand the actor code, but I tested that the fix resolves the issue for me.

PR #3764, opened 2023-12-21
Fix the `tenzir/tenzir:latest-slim` image

dominiklohmann:
This was missing the `TENZIR_CACHE_DIRECTORY`, causing the slim variant to fail on startup.


tobim (approved):
Thanks!

PR #3765, opened 2023-12-22
Fix blob parsing with padding

jachris:
When reading Base64-encoded JSON string uses the `blob` type, padding with `=` is now accepted.


dominiklohmann (approved):
Please add a test before merging.

PR #3771, opened 2023-12-29
Improve robustness of TCP connector

mavam:
This PR streamlines the response promise handling in the TCP connector, avoiding potential double delivery.

### DIFF
--- a/libtenzir/builtins/connectors/tcp.cpp
+++ b/libtenzir/builtins/connectors/tcp.cpp
@@ -111,13 +111,12 @@ auto make_tcp_bridge(tcp_bridge_actor::stateful_pointer<tcp_bridge_state> self)
           | boost::asio::ssl::verify_fail_if_no_peer_cert);
         self->state.tls_socket.emplace(*self->state.socket,
                                        *self->state.ssl_ctx);
-        if (SSL_set1_host(self->state.tls_socket->native_handle(),
-                          hostname.c_str())
-            != 1)
+        auto tls_handle = self->state.tls_socket->native_handle();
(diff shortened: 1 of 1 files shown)


libtenzir/builtins/connectors/tcp.cpp
```diff
@@ -131,24 +130,26 @@ auto make_tcp_bridge(tcp_bridge_actor::stateful_pointer<tcp_bridge_state> self)
...
                     ec::system_error,
                     fmt::format("connection failed: {}", ec.message())));
+                }
                 if (self->state.tls_socket) {
                   self->state.tls_socket->handshake(
                     boost::asio::ssl::stream<
                       boost::asio::ip::tcp::socket>::client,
                     ec);
+                  if (ec) {
+                    return self->state.connection_rp.deliver(caf::make_error(
```
tobim, line 144: IIuc this `return` is the only non-cosmetic change, is that correct?

tobim (commented):
Can you provide an example pipeline that now behaves differently? Ideally in the form of a test ;).

libtenzir/builtins/connectors/tcp.cpp
mavam: Some `return`s are cosmetic transformations for consistency throughout a single function, others not. For example, a failed TLS handshake caused a double delivery.

mavam:
> Can you provide an example pipeline that now behaves differently? Ideally in the form of a test ;).

Alas, no. It's a bit stabbing in the dark. At least one potential bug is fixed with this PR, the rest are transformations that result in a more robust control flow.

tobim (approved):

PR #3777, opened 2024-01-04
Update the manual static build workflow

tobim:
The actual builder workflow now takes an additional parameter named "gha-runner-labels". The workflow broke when we added support for building on darwin.


mavam (approved):
Blindly approving, but seems like a plausible fix.

PR #3781, opened 2024-01-08
Use Inter and JetBrains Mono fonts on docs site

dit7ya:
Closes tenzir/issues#1080.

Docusaurus and Infima does a lot of CSS abstractions that don't map one-on-one to Tailwind (and thus our design system on the app and the site). For example, Tailwind sets line-height for each of `text-base, text-lg` etc, while docusaurus just uses a base and a header line-height variables. Docusaurus also conditionally changes the font-size for small screens - which will apply when the iframe width (the main motivation for this PR) is small, but not always.

Having considered all of these, I think we should just pluck the low-hanging fruit of changing the font-families on the docs site.

### DIFF
--- a/web/src/css/custom.scss
+++ b/web/src/css/custom.scss
@@ -4,6 +4,9 @@
  * work well for content-centric websites.
  */
 
+/* Load fonts from Bunny Fonts */
+@import url(https://fonts.bunny.net/css?family=inter:100,200,300,400,500,600,700,800,900|jetbrains-mono:100,200,300,400,500,600,700,800);
+
 /* You can override the default Infima variables here. */
 :root {
   --ifm-color-primary: #00a4f1;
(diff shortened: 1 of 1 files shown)


mavam:
Do you have a screenshot that shows how this looks like? I did this in the past and found that it requires *a lot* more CSS adjustments to look reasonable.

ClickHouse has a similar design system, also based on Inter. Their [docs site](https://clickhouse.com/docs/) is also Docusaurus-based. I went through their custom.css once, and things like line-height and a bunch of other things were adapted to make it look as solid as it is right now.

dominiklohmann:
Here's a side-by-side comparison. I think this looks good enough to be merged as-is.

image

mavam:
> I think this looks good enough to be merged as-is.

Agreeing.

mavam (approved):

PR #3783, opened 2024-01-09
Make the schema definition easier to parse programmatically

dominiklohmann:
This is an experimental branch for @dit7ya to play with: It changes the schema definition included with the response in `/serve` to a simpler to parse format for the app hosted at app.tenzir.com.

### DIFF
--- a/libtenzir/builtins/operators/serve.cpp
+++ b/libtenzir/builtins/operators/serve.cpp
@@ -714,8 +714,7 @@ struct serve_handler_state {
       first = false;
       out_iter = fmt::format_to(out_iter, R"("schema_id":"{}","definition":)",
                                 schema.make_fingerprint());
-      const auto ok
-        = printer.print(out_iter, schema.to_definition(/*expand*/ false));
+      const auto ok = printer.print(out_iter, schema.to_definition2());
       TENZIR_ASSERT_CHEAP(ok);
     }
     out_iter = fmt::format_to(out_iter, R"(}}]}}{})", '\n');
--- a/libtenzir/builtins/operators/serve.cpp
+++ b/libtenzir/builtins/operators/serve.cpp
@@ -720,7 +720,7 @@ struct serve_handler_state {
     }
     // Write schemas
     if (seen_schemas.empty()) {
-      out_iter = fmt::format_to(out_iter, R"(}}],"schemas":[]}}{})", '\n');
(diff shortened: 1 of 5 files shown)


dit7ya:
Some preliminary thoughts

1. I think this is shipping way more data which is making the network call take longer. `export | taste` ships 200kb compared to demo node's 50kb.
2. Any reason for the `type` being a string? In this structure the string is not much usable from the frontend anyway.
```
{
	"12": {
		"name": "alert",
		"kind": "record",
		"type": "record {app_proto: string, action: enum {allowed: 0, blocked: 1}, gid: uint64 #index=hash, signature_id: uint64 #index=hash, rev: uint64, signature: string, category: string, severity: uint64, source: record {ip: ip, port: port}, target: record {ip: ip, port: port}}",
}
```

3. The listness is not indicated in the type or kind fields. For example, the schema `suricata.anomaly` has `uint64` for both `kind` and `type`, while it is actually `list<uint64>`.

dit7ya:
I have started to integrate this, you can play with this at tenzir/app#460.

mavam:
Just curious, what's the reason for not using https://arrow.apache.org/docs/js/? Extension types?

dominiklohmann:
Three reasons, to be precise

- Incomplete extension type support
- Lack of ability to have heterogeneous streams
- All we'd be doing in the app is converting back to JSON—at least with the current table implementation that's row-based—which is just a lot faster to do in the node

dominiklohmann:
We just discussed this in a quick call.

> Some preliminary thoughts […]

1. I will dedup schema definitions across calls to serve for the same serve id to cut it down again a bit.
2. Is a bug, I will fix that. It should exactly match the kind for unnamed types.
3. Is a bug as well, will fix.

dominiklohmann:
> 1\. I will dedup schema definitions across calls to serve for the same serve id to cut it down again a bit.

I decided against after some consideration as it's not simple to implement with the current behavior of storing the last reply in-memory so that it can be retrieved again. That, unfortunately, is currently implemented not by keeping the actual response, but by keeping the underlying events. Making the proposed change without adjusting this implementation could lead to schemas not arriving in the app.

> 2\. Is a bug, I will fix that. It should exactly match the kind for unnamed types.
> 3\. Is a bug as well, will fix.

Both are fixed. Please take a look again @dit7ya.

This is guarded behind the feature flag `serve_use_simple_format_option` and the `use_simple_format: bool` parameter for `/serve` now.

PR #3784, opened 2024-01-09
Improve handling of open file descriptors

lava:
Attempt to improve the behavior of nodes that are running into the maximum number of open files:

  * Increase the upper limit imposed by our systemd unit files to 64Ki
  * Add monitoring of open file descriptors as a health metric so that leaks can be detected by looking at the rate of change over time.

### DIFF
--- a/ansible/roles/tenzir/tenzir-node.service
+++ b/ansible/roles/tenzir/tenzir-node.service
@@ -28,6 +28,9 @@ CapabilityBoundingSet=
 AmbientCapabilities=
 RestrictSUIDSGID=yes
 
+# limits
+LimitNOFILE=65535
+
 # system access
 ExecPaths=/opt/tenzir/libexec/tenzir-df-percent.sh
 ProtectSystem=strict
--- a/changelog/next/features/3784--open-fds.md
+++ b/changelog/next/features/3784--open-fds.md
@@ -1,0 +1,3 @@
+On linux systems, the process metrics now have an additional
+value `open_fds` showing the number of file descriptors
+opened by the node.
--- a/changelog/next/features/3784--open-fds.md
+++ b/changelog/next/features/3784--open-fds.md
@@ -1,3 +1,3 @@
-On linux systems, the process metrics now have an additional
+On Linux systems, the process metrics now have an additional
 value `open_fds` showing the number of file descriptors
 opened by the node.
--- a/libtenzir/src/detail/process.cpp
+++ b/libtenzir/src/detail/process.cpp
@@ -75,18 +75,6 @@ static record get_status_proc() {
   return result;
 }
 
-static int64_t get_process_fds() {
-  auto path = "/proc/self/fd";
-  try {
-    // Empirically this uses about 1 syscall per 1500 directory entries.
-    auto begin = std::filesystem::directory_iterator(path);
-    auto end = std::filesystem::directory_iterator{};
-    return std::distance(begin, end);
-  } catch (const std::filesystem::filesystem_error& e) {
-    return -1;
-  }
-}
-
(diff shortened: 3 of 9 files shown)


ansible/roles/tenzir/tenzir-node.service
```diff
@@ -28,6 +28,9 @@ CapabilityBoundingSet=
 AmbientCapabilities=
 RestrictSUIDSGID=yes
 
+# limits
+LimitNOFILE=65535
```
lava, line 32: I considered going all the way to `LimitNOFILE=infinity`, but in the case of a file descriptor leak it seems better to have the node fail at some point rather than the system hitting the maximum number of global open files.

mavam: Agreed. Could you expand on the comment a bit here to inform users about this decision?

libtenzir/src/detail/process.cpp
```diff
@@ -75,6 +75,18 @@ static record get_status_proc() {
   return result;
 }
 
+static int64_t get_process_fds() {
```
mavam, line 78: We have such functionality already in the code base: https://github.com/tenzir/tenzir/blob/main/libtenzir/src/os.cpp#L252

I don't think we want it twice.

mavam (commented):
Mind adding a changelog entry?

It'd be great if we can reuse logic from `os.cpp` instead of hand-rolling it again. Otherwise this good.

ansible/roles/tenzir/tenzir-node.service
lava: I think it's the wrong audience; for the end users it doesn't matter *why* we set the limit this way. I'll copy the comment in the commit description instead.

changelog/next/features/3784--open-fds.md
```diff
@@ -0,0 +1,3 @@
+On linux systems, the process metrics now have an additional
```
mavam, line 1:
```suggestion
On Linux systems, the process metrics now have an additional
```

libtenzir/src/os.cpp
```diff
@@ -413,7 +452,7 @@ auto socket_state_to_string(auto proto, auto state) -> std::string_view {
 
 auto darwin_os::fetch_sockets() -> std::vector<socket> {
   auto result = std::vector<socket>{};
-  for (const auto& proc : fetch_processes()) {
+  for (const auto& proc : fetch_processes(std::nullopt)) {
```
mavam, line 455:
```suggestion
  for (const auto& proc : fetch_processes()) {
```

mavam: If you make `= {}` a default arg, the API looks a bit nicer, as `nullopt` as arg is not very semantic.

libtenzir/src/os.cpp
```diff
@@ -173,8 +198,12 @@ auto linux_os::fetch_processes() -> std::vector<process> {
           .startup = {},
           .vsize = static_cast<uint64_t>(stat.vsize),
           .rsize = static_cast<uint64_t>(stat.rss * getpagesize()),
+          .peak_mem = static_cast<uint64_t>(status.vm_hwm),
+          .swap = static_cast<uint64_t>(status.vm_swap),
+          .open_fds = open_fds,
           .utime = std::chrono::seconds{stat.utime / state_->clock_tick},
           .stime = std::chrono::seconds{stat.stime / state_->clock_tick},
+
```
mavam, line 206:
```suggestion
```

mavam (approved):
Nice work. This doubles down on our effort to centralize OS-specific aspects.

Just minor nits in the review comments.

PR #3787, opened 2024-01-10
Fix serialization of `pseudonymize` operator

jachris:
This bug was caused by `configuration::seed_bytes` being computed by the operator constructor, while at the same time, it was not listed in the `inspect()` implementation. Hence, serializing this operator made the seed all-zero.


dominiklohmann (approved):
Whoops 🙈

PR #3788, opened 2024-01-10
Display failing pipeline diagnostics in `/serve`

dominiklohmann:
Fixes tenzir/issues#1099

This changes `/serve` to show pipeline diagnostics when the pipeline is no longer available instead of a generic error. To make this possible I had to change the routing of down messages between execution nodes and the pipeline executor, and also very slightly change how diagnostics are emitted.


libtenzir/src/error.cpp
```diff
@@ -95,10 +98,28 @@ std::string render(caf::error err) {
...
       render_default_ctx(oss, err.context());
       break;
-    case caf::type_id_v<tenzir::ec>:
-      oss << to_string(static_cast<tenzir::ec>(err.code()));
-      render_default_ctx(oss, err.context());
+    case caf::type_id_v<tenzir::ec>: {
+      const auto code = static_cast<tenzir::ec>(err.code());
+      if (code == ec::diagnostic) {
+        auto printer = make_diagnostic_printer("<unknown>", "<unknown>",
+                                               color_diagnostics::yes, oss);
```
jachris, line 105: This means that errors with location information cannot be printed (you might even run into assertions in the diagnostic printer, which assumes that all locations are valid). We will probably change how locations are associated with files and sources in the future, but we have to do something here now.

dominiklohmann: Can I just use a 0:0 location?

libtenzir/include/tenzir/diagnostics.hpp
```diff
@@ -32,6 +33,20 @@
 
 namespace tenzir {
 
+namespace {
+
+auto trim_and_truncate(std::string& str) {
```
jachris, line 38: You probably want to make this `inline` instead of `static`.

libtenzir/src/error.cpp
jachris: You mean you want to patch all locations in the embedded diagnostic?

dominiklohmann: I'm afraid I don't understand the argument passed to this function correctly then. Is it the entire pipeline?

jachris: We solved this (for now) by making adding an option in the printer to not print any location information. This is not the optimal solution at all, and we shall improve this in in the future when we change how locations work.

jachris (approved):

PR #3789, opened 2024-01-10
Add an additional control flow for unspecified msgpack types in the fluent-bit plugin

Dakostu:
This change fixes a warning during building by adding a more explicit edge case handling for unknown MsgPack types in the `fluent-bit` plugin.

### DIFF
--- a/plugins/fluent-bit/src/plugin.cpp
+++ b/plugins/fluent-bit/src/plugin.cpp
@@ -22,8 +22,12 @@
 
 #include <cstring>
 #include <queue>
+#include <stdexcept>
 
 #include <fluent-bit/fluent-bit-minimal.h>
+
+// Tag type for when we are retrieving MsgPack objects with an unknown type.
+struct unknown_msgpack_type {};
 
 template <>
 struct fmt::formatter<msgpack_object_type> : fmt::formatter<std::string_view> {
@@ -131,6 +135,7 @@ auto visit(auto f, const msgpack_object& object) {
     case MSGPACK_OBJECT_EXT:
       return f(object.via.ext);
   }
+  return f(unknown_msgpack_type{});
 }
 
 /// RAII-style wrapper around `msgpack_unpack`.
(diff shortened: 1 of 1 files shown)


plugins/fluent-bit/src/plugin.cpp
```diff
@@ -130,6 +130,8 @@ auto visit(auto f, const msgpack_object& object) {
       return f(to_bin(object));
     case MSGPACK_OBJECT_EXT:
       return f(object.via.ext);
+    default:
+      return f(ec::unspecified);
```
eliaskosunen, line 134: Having a `default:` label here, instead of having this outside the `switch`, has the effect that the compiler won't warn about any potentially missing enumerators. This enum isn't probably going to get new enumerators any time soon, but that's something to consider.

eliaskosunen (approved):

plugins/fluent-bit/src/plugin.cpp
```diff
@@ -131,6 +131,7 @@ auto visit(auto f, const msgpack_object& object) {
     case MSGPACK_OBJECT_EXT:
       return f(object.via.ext);
   }
+  return f(ec::unspecified);
```
mavam, line 134: Why are dispatching to a CAF error code? There's no CAF at all nearby.

Dakostu: We're dispatching `tenzir::ec` values as to catch them in upstream methods that call `msgpack::visit()`, such as `add()` (pretty much the only method that calls `msgpack::visit()` anyway).

Would you prefer a semantically cleaner type, such as `std::monostate`?

mavam: I don't want to bikeshed on this too much, but anything that smells like an error structure from `std*` would feel more natural.

Dakostu: I replaced the error type with an `std::runtime_error`.

Dakostu: Basically, yes, a non-`msgPack` tag type.

Is there one you'd recommend?

Dakostu: I implemented `unknown_msgpack_type`.

PR #3790, opened 2024-01-10
Fix `export --live` and introduce `metrics`

dominiklohmann:
The `metrics [--live]` source operator is a short-hand notation for accessing all metrics events via `export --internal`.

Additionally, this PR fixes `export --live`, which was broken in many ways.

### DIFF
--- a/changelog/next/features/3790--metrics-operator.md
+++ b/changelog/next/features/3790--metrics-operator.md
@@ -1,0 +1,3 @@
+The `metrics` operator returns internal metrics events generated in a Tenzir
+node. Use `metrics --live` to get a feed of metrics as they are being generated
+instead of past metrics.
--- a/changelog/next/features/3790--metrics-operator.md
+++ b/changelog/next/features/3790--metrics-operator.md
@@ -1,3 +1,2 @@
 The `metrics` operator returns internal metrics events generated in a Tenzir
-node. Use `metrics --live` to get a feed of metrics as they are being generated
-instead of past metrics.
+node. Use `metrics --live` to get a feed of metrics as they are being generated.
--- a/libtenzir/builtins/operators/export.cpp
+++ b/libtenzir/builtins/operators/export.cpp
@@ -23,8 +23,6 @@
 #include <tenzir/uuid.hpp>
 
 #include <arrow/type.h>
-#include <caf/attach_stream_sink.hpp>
-#include <caf/attach_stream_source.hpp>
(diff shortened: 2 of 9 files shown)


changelog/next/features/3790--metrics-operator.md
```diff
@@ -0,0 +1,3 @@
+The `metrics` operator returns internal metrics events generated in a Tenzir
+node. Use `metrics --live` to get a feed of metrics as they are being generated
+instead of past metrics.
```
mavam, line 3:
```suggestion
node. Use `metrics --live` to get a feed of metrics as they are being generated.
```

libtenzir/builtins/operators/metrics.cpp
```diff
@@ -0,0 +1,51 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2023 The Tenzir Contributors
```
mavam, line 6:
```suggestion
// SPDX-FileCopyrightText: (c) 2024 The Tenzir Contributors
```

libtenzir/builtins/operators/export.cpp
```diff
@@ -37,45 +35,36 @@
...
-               })
-        .inbound_slot();
+    [self](table_slice slice) {
+      auto filtered = filter(std::move(slice), self->state.expr);
+      if (not filtered)
+        return;
+      if (self->state.rp.pending()) {
+        self->state.rp.deliver(std::move(*filtered));
+      } else {
+        self->state.buffer.push(std::move(*filtered));
```
tobim, line 61: I think we need to limit the buffer size here. Otherwise we're going to get memory issues when the downstream operator isn't pulling from the `export --live`.

tobim: In principle it depends on how much is available. If we had topics with bounded buffers it would already be solvable dynamically at the pipeline definition level, but for now I'd say even 100K events should be enough, and be less of a burden for small deployments.

dominiklohmann: We agreed in a call to use a higher buffer size. I chose 4Mi for now, which is also our partition size.

tobim (approved):
:pray:

PR #3539, opened 2023-09-25
Add savers for curl connectors

mavam:
This PR adds savers for the curl-based connectors, making the API symmetric.

```[tasklist]
### Tasks
- [x] Implement saver
- [x] Fix PUT and POST on empty body
- [x] Support URL-encoded form submission
- [x] Update docs
- [x] Set Content-Type based on chosen format
```

PR #3791, opened 2024-01-10
Fix handling of escapes of trailing separators in `detail::split_escaped`

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1128

The handling of escaped separators after `max_splits` is reached was inconsistent in `detail::split_escaped`. This PR makes the behavior consistent, so that escape characters are always removed:

```cpp
// Before:
// Only first separator after `max_splits` was reached
// had its escape sequence removed in the output
auto ret = detail::split_escaped("abc\\==def==ghi\\==jkl\\==mno", "==", "\\", 1);
// ret == ["abc==def", "ghi==jkl\\==mno"]

// After:
// Escape sequences always removed
auto ret = detail::split_escaped("abc\\==def==ghi\\==jkl\\==mno", "==", "\\", 1);
// ret == ["abc==def", "ghi==jkl==mno"]
```


libtenzir/src/detail/string.cpp
jachris, line 1: The change itself looks good, but what is the motivation here? I'm not sure what the correct behavior is TBH (the previous one is not necessarily wrong, as it for example allows you to iterate with `max_splits = 1` to split everything).

eliaskosunen: I think either behavior (remove no escapes, or remove all of them) makes sense, as long as it's consistent. The need for this behavior originally arose here: https://github.com/tenzir/tenzir/pull/3539/files#diff-a8954be45ce4d0cb841f5429c74fa09a6716d2d161c37b57a543dc01f632390e, where the remove-all behavior is currently expected.

I guess you're right, that the remove-all behavior can be emulated with remove-none with `detail::join(detail::split_escaped(remainder, separator, escape), separator)`, but the other way around isn't really possible.

eliaskosunen:
To elaborate, this PR expands upon the band-aid fix to `split_escaped` in #3539, fixing the consistency issue detailed in the PR description.

Current main:
```cpp
auto a = split_escaped("foo=bar\\=", "=", "\\", 1);
// a == ["foo"]
// !!! "bar\\=" just disappeared

auto b = split_escaped("foo=bar\\=baz\\=quux", "=", "\\", 1);
// b == ["foo, "bar=baz=quux"]
// Escapes are removed even after max_splits is reached...

auto c = split_escaped("foo=bar=baz\\=quux", "=", "\\", 1);
// c == ["foo", "bar=baz\\=quux"]
// ...except if there's an unescaped separator before it?
```

With #3539 "Add savers for curl connectors"
```cpp
auto a = split_escaped("foo=bar\\=", "=", "\\", 1);
// a == ["foo", "bar="]
// This is fixed with a minimal change to split_escaped

auto b = split_escaped("foo=bar\\=baz\\=quux", "=", "\\", 1);
// b == ["foo, "bar=baz=quux"]

auto c = split_escaped("foo=bar=baz\\=quux", "=", "\\", 1);
// c == ["foo", "bar=baz\\=quux"]
// This is still lingering
```

With this PR
```cpp
auto a = split_escaped("foo=bar\\=", "=", "\\", 1);
// a == ["foo", "bar="]

auto b = split_escaped("foo=bar\\=baz\\=quux", "=", "\\", 1);
// b == ["foo, "bar=baz=quux"]

auto c = split_escaped("foo=bar=baz\\=quux", "=", "\\", 1);
// c == ["foo", "bar=baz=quux"]
// This is also fixed
```

libtenzir/src/detail/string.cpp
mavam: I'm with @jachris on this.

eliaskosunen: Ack.

PR #3796, opened 2024-01-11
Fix ccache upload behavior

lava:
* Ensure that `ccache` can actually read the downloaded ccache during the build
* Split upload and download cache steps to upload before integration tests

### DIFF
--- a/.github/workflows/tenzir.yaml
+++ b/.github/workflows/tenzir.yaml
@@ -611,8 +611,13 @@ jobs:
       CCACHE_ABSSTDERR: true
       CCACHE_COMPRESS: true
       CCACHE_COMPRESSLEVEL: 6
-      CCACHE_DIR: "${{ github.workspace }}/.ccache"
-      CCACHE_HASH_DIR: true
+      # We're intentionally placing the cache dir outside of `${{ github.workspace }}`
+      # because that has a weird issue where it switches between `/home/runner` and
+      # `/__w/` depending on the current context. See https://github.com/actions/checkout/issues/785
+      # and https://github.com/actions/runner/issues/2058.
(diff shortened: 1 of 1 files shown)


.github/workflows/tenzir.yaml
```diff
@@ -718,6 +726,11 @@ jobs:
           ccache --show-stats
           # Print statistics about cache compression.
           ccache --show-compression
+      - name: Save ccache Cache
+        uses: actions/cache/save@v3
```
tobim, line 730: Should we add
```
        if: always()
```
here?

.github/workflows/tenzir.yaml
```diff
@@ -718,6 +726,11 @@ jobs:
...
+        with:
+          path: ${{ env.CCACHE_DIR }}
+          key: ${{ github.workflow }}-${{ matrix.tenzir.name }}-${{ matrix.tenzir.compiler }}-${{ needs.configure.outputs.head-ref-slug }}-${{ github.sha }}
```
tobim, line 733: Would it be possible to de-dup the value for `key`, to prevent it going out of sync?

tobim (approved):
Great work!

.github/workflows/tenzir.yaml
lava: I'll create a separate PR for that so we can already start building up the ccache now

.github/workflows/tenzir.yaml
lava: It's actually a bit more complicated, because key matched here will almost never be the primary key but one of the prefix-matched restore keys. Additionally the path has to match as well as one of the keys. I'll just keep it for now and hope that no one will dare to touch it again :D

PR #3808, opened 2024-01-12
Override lookup-table context entries for duplicate keys

dominiklohmann:
Previously, values with an already existing entry for the same key were just ignored.


jachris (approved):
Makes sense.

PR #3812, opened 2024-01-12
Support concepts in more places

dominiklohmann:
This change adds support for concepts in almost all operators and contexts. In particular, this affects the following contexts or operators:
- `context`
- `delay`
- `drop`
- `enrich`
- `extend`
- `geoip`
- `hash`
- `lookup`
- `lookup-table`
- `parse`
- `pseudonymize`
- `put`
- `rename`
- `replace`
- `select`
- `sort`
- `summarize`
- `timeshift`
- `lookup`

Note that `put net.src.ip` creates a field `net.src.ip`, but `select net.src.ip` keeps the original field name.


jachris (approved):
👍

PR #3814, opened 2024-01-12
Remove restriction for unflattening into empty field names

dominiklohmann:
This restriction was put in place when record field names were not allowed to be empty. That is no longer the case, so we can safely remove it.

There is still some more work to do before this becomes really useful for GELF, as that still requires moving from that top-level empty-named field into the top-level record itself, but this is an easy step forward.


mavam (approved):
Makes sense

PR #3819, opened 2024-01-16
Add SQS Connector

mavam:
This PR adds an AWS SQS connector.

```[tasklist]
- [x] Implement loader
- [x] Implement saver
- [x] Support long polling
- [x] Support queue creation
- [x] Write docs
- [x] Add integration tests
- [x] Integrate into CMake (fix linking)
```


mavam:
This one needs review from @tobim regarding deeper integration into the build scaffold.

While I was able to build it locally, it's not yet there in CI. Things we need to figure out if Arrow links the SDK privately. It's already a dependency, so I thought we can reuse it. But only if linked publicly, otherwise we risk ODR violations.

mavam:
After doing the necessary setup on the CLI, I got this working as follows:

Generate data:
```bash
tenzir  'version | repeat 100 | enumerate | to sqs tenzir'
```

Retrieve it:

```bash
tenzir 'from sqs tenzir'
```

If you have suggestions how to test it in CI, let me know, but I unless we have local mock instance of SQS, we'll have to live without integration tests.

tobim:
> If you have suggestions how to test it in CI, let me know, but I unless we have local mock instance of SQS, we'll have to live without integration tests.

You could try https://github.com/softwaremill/elasticmq.

mavam:
This works locally. Surprisingly well. I know too little about BATS though to integrate it properly. Can you show me how to take it from here?

Here's my `elasticmq.conf`:

```
include classpath("application.conf")

queues {
  tenzir {
    defaultVisibilityTimeout = 10 seconds
    delay = 5 seconds
    receiveMessageWait = 0 seconds
    fifo = false
    contentBasedDeduplication = false
  }
}
```

And this is my BATS file:

```bash
: "${BATS_TEST_TIMEOUT:=3}"

ELASTICMQ_CONF="$(dirname "$BATS_SUITE_DIRNAME")/elasticmq.conf"

setup_elasticmq() {
  coproc SERVER {
    docker run -p 9324:9324 -p 9325:9325 \
      -v ${ELASTICMQ_CONF}:/opt/elasticmq.conf \
      softwaremill/elasticmq-native
  }
}

setup() {
  bats_load_library bats-support
  bats_load_library bats-assert
  bats_load_library bats-tenzir
  setup_elasticmq
  export AWS_ENDPOINT_URL="http://localhost:9324"
}

@test "SQS read and write" {
  check aws sqs list-queues
  check tenzir "version | put foo=42 | repeat 10 | enumerate n | to sqs://test"
  check tenzir "from sqs://test | sort n desc"
}
```

mavam:
I got a bit further in https://github.com/tenzir/tenzir/pull/3819/commits/b8c6620bff7cd9d68c207046bb430020eef323b4, but now I'm facing an AWS SDK issue. When I'm using dummy `AWS_*` values for my local endpoint, commands like `tenzir from sqs://tenzir` always return this error:

```
The security token included in the request is invalid.
```

@rdettai or @tenzir/platform-squad any thoughts what I need to do? I wiped my `~/.aws` directory and just set these env vars:

```
export AWS_DEFAULT_REGION="us-west-2"
export AWS_ACCESS_KEY_ID="dummy"
export AWS_SECRET_ACCESS_KEY="dummy"
export AWS_ENDPOINT_URL="http://localhost:9324"
```

With "real" creds that I configure via `aws configure sso`, things work. I just can't get the local setup to work. My hunch is that it's [an MFA issue](https://stackoverflow.com/a/47098678) and that I need to do something with `AWS_SESSION_TOKEN `—but I'm not sure.

rdettai:
Feels like the AWS client is doing something ElasticMQ doesn't like. From the docs https://github.com/softwaremill/elasticmq?tab=readme-ov-file#using-the-amazon-boto-python-to-access-an-elasticmq-server looks like `use_ssl=False` is required. Not sure how the C++ SDK supports this.

mavam:
> Do you manage to interact with the local endpoint using Boto3 or the AWS CLI?

It works with the AWS CLI:

```bash
# Sending works.
aws sqs send-message --queue-url http://localhost:9324/000000000000/tenzir --message-body '{"foo": 42}'

# And so does receiving.
aws sqs receive-message --queue-url http://localhost:9324/000000000000/tenzir
```

The interesting part is that it *does* work when I use my actual SSO creds and only adapt `AWS_ENDPOINT_URL`. This makes me believe the the AWS SDK we use here requires something special. I just don't know how to bend it, since it's inside Tenzir and I haven't futzed with it yet.

mavam:
@rdettai and I paired on this and we found the issue: the `tenzir` binary does not respect `$AWS_ENDPOINT_URL`.

The AWS SDK [introduced support for this variable in mid '23](https://github.com/aws/aws-sdk-cpp/issues/2587), though, so it should work in theory. We now have to look if our SDK is simply too old.

mavam:
Since we pushed the integration tests back until we have Arrow 15, the only outstanding issue is now fixing the linking and packaging. Waiting for @tobim or @dominiklohmann for that.

.github/workflows/tenzir.yaml
```diff
@@ -616,7 +616,7 @@ jobs:
               # - The http-parser package is disabled on Hoembrew, which means
               #   we cannot build the web plugin on macOS anymore. Last checked
               #   on 2024-01-31.
-              - plugins/[^(parquet)(web)]*
+              - plugins/[^(parquet)(web)(sqs)]*
```
dominiklohmann, line 619: This could really use a comment explaining _why_ we disable sqs on macOS, similar to the comments above for parquet and web.

mavam: Does that work for you?

plugins/sqs/integration/tests/sqs.bats
```diff
@@ -0,0 +1,43 @@
...
+export AWS_ACCESS_KEY_ID="dummy"
+export AWS_SECRET_ACCESS_KEY="dummy"
+export AWS_ENDPOINT_URL="http://localhost:9324"
+
+setup() {
+  DATADIR="${BATS_TEST_DIRNAME}/../data"
+  bats_load_library bats-support
+  bats_load_library bats-assert
+  bats_load_library bats-tenzir
+  container_id=$(docker run -d -p 9324:9324 \
```
dominiklohmann, line 11: You want something like this here in setup, in teardown and for all tests:

```
command -v docker 2>/dev/null || skip "Docker is not available"
```

mavam:
> I'll add the missing changelog entries in a new PR.

Pushed a change already!

PR #3820, opened 2024-01-16
Fix handling of escapes in detail::split_escaped, attempt 2

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1128, old PR https://github.com/tenzir/tenzir/pull/3791

```cpp
// Before:
// Only first separator after `max_splits` was reached
// had its escape sequence removed in the output
auto ret = detail::split_escaped("abc\\==def==ghi\\==jkl\\==mno", "==", "\\", 1);
// ret == ["abc==def", "ghi==jkl\\==mno"]

// After:
// Escape sequences not removed after `max_splits` is reached
auto ret = detail::split_escaped("abc\\==def==ghi\\==jkl\\==mno", "==", "\\", 1);
// ret == ["abc==def", "ghi\\==jkl\\==mno"]
```

TODO: sign commits

### DIFF
--- a/libtenzir/src/detail/string.cpp
+++ b/libtenzir/src/detail/string.cpp
@@ -137,7 +137,7 @@ split_escaped(std::string_view str, std::string_view sep, std::string_view esc,
   auto it = str.begin();
   std::string current{};
   size_t splits = 0;
-  while (it != str.end()) {
+  while (it != str.end() && splits != max_splits) {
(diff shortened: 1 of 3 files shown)


jachris (approved):
Some of the tests still need to be adjusted (and the commit signed), but otherwise this looks good to go! 🚀

PR #3821, opened 2024-01-16
Add `xsv --no-header`

eliaskosunen:
Done to support https://github.com/tenzir/issues/issues/1135

Adds a new option to the `xsv`-family of printers, `--no-header`, with the expected behavior.


dominiklohmann (approved):
Nice, thank you!

PR #3823, opened 2024-01-16
Migrate integration tests to BATS

lava:


mavam:
The 500k LOC addition seems massive. Is there a chance we can reduce the taste baselines? I would challenge that we need thousands of lines of baseline for most tests.

lava:
> Is there a chance we can reduce the taste baselines?

Adding `TENZIR_PLUGINS__JSON__COMPACT_OUTPUT=true` to the default flags would likely reduce this quite a bit, but I'd like to see the tests pass once before starting to change things up again.

mavam:
> but I'd like to see the tests pass once before starting to change things up again.

Okay, so the strategy is first getting everything green, and then starting to work on reducing the size of the diff, and then merge? That would make sense to me.

tenzir/functional-test/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+To run all tests in a specific test file, use e.g. `bats tests/version.test`
+
+To run the set of tests that failed during the last run of bats,
+the `--filter-status failed` option can be used. This can also
+be useful in combination with the `UPDATE` option below when updating
+tests after implementing a breaking change.
+
+To run one specific test, add the `bats:focus` tag to that test and
+run the test file.
```
mavam, line 23: How do I add a tag?

tenzir/functional-test/data/reference/python/test_empty_output_throws/step_00.ref
jachris, line 1: This file contains escape sequences. This should not happen with `--dump-diagnostics` (which should probably be enabled everywhere).

tenzir/functional-test/README.md
tobim: See https://bats-core.readthedocs.io/en/stable/writing-tests.html#tagging-tests.

TLDR:
```
# bats test_tags=tag:1, tag:2, tag:3
@test "first test" {
  # ...
}
```

tenzir/functional-test/tests/python.bats
```diff
@@ -116,32 +109,34 @@ END
 END
   check grep -v '\(Host:\|User-Agent:\)' "${data}"
   truncate -s 0 "${data}"
-}
 
+  # Teardown
+  kill -9 "$SRV_PID"
+  rm "${conn}" "${data}"
```
tobim, line 115: If one of the commands in the test fails the rest of it won't be executed any more and the teardown here won't run, leaving the `socat` running. I believe bats copies the test contents into a temporary file and runs that with `bash -e`, so you could try to set a `trap` after creating the `coproc` instead.

lava: Bats recommends against setting traps, because they might interfere with the bats-internal traps.

Instead, I moved the kill back into the global teardown, and moved the directories into the temporary directory of this test so that they don't need to be removed manually.

tobim (approved):
Approving now, but the todolist needs to be removed before merging.

PR #3833, opened 2024-01-19
Add support for Cisco Firepower syslog messages

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1162

The RFC 3164 syslog parser couldn't handle (non-standard) years in timestamps. Now it can.


jachris (approved):
Looks good to me :+1:

libtenzir/include/tenzir/format/syslog.hpp
dominiklohmann, line 1: This bug fix still needs a changelog entry.

PR #3834, opened 2024-01-19
Add Bloom filter context

mavam:
This PR adds a Bloom filter context.

```[tasklist]
- [x] Implement DCSO Bloom filter
- [x] Scaffold `bloom-filter` context
- [x] Agree on UX
- [x] Write documentation
- [x] Add tests
```

### DIFF
--- a/libtenzir/builtins/contexts/bloom_filter.cpp
+++ b/libtenzir/builtins/contexts/bloom_filter.cpp
@@ -1,0 +1,218 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2024 The VAST Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <tenzir/arrow_table_slice.hpp>
+#include <tenzir/concept/parseable/numeric/bool.hpp>
+#include <tenzir/data.hpp>
+#include <tenzir/dcso_bloom_filter.hpp>
+#include <tenzir/detail/range_map.hpp>
+#include <tenzir/expression.hpp>
+#include <tenzir/fbs/data.hpp>
+#include <tenzir/flatbuffer.hpp>
+#include <tenzir/fwd.hpp>
+#include <tenzir/operator.hpp>
+#include <tenzir/plugin.hpp>
+#include <tenzir/project.hpp>
+#include <tenzir/series_builder.hpp>
+#include <tenzir/table_slice.hpp>
+#include <tenzir/table_slice_builder.hpp>
+#include <tenzir/type.hpp>
+#include <tenzir/typed_array.hpp>
+
+#include <arrow/array.h>
+#include <arrow/array/array_base.h>
+#include <arrow/array/builder_primitive.h>
+#include <arrow/type.h>
+#include <caf/error.hpp>
+
+#include <chrono>
+#include <memory>
+#include <string>
+
+namespace tenzir::plugins::bloom_filter {
+
+namespace {
+
+class bloom_filter_context final : public virtual context {
+public:
+  bloom_filter_context() noexcept = default;
+
+  explicit bloom_filter_context(dcso_bloom_filter bloom_filter) noexcept
+    : bloom_filter_{std::move(bloom_filter)} {
+  }
+
+  bloom_filter_context(uint64_t n, double p) : bloom_filter_{n, p} {
+  }
+
+  /// Emits context information for every event in `slice` in order.
+  auto apply(table_slice slice, context::parameter_map parameters) const
+    -> caf::expected<std::vector<typed_array>> override {
+    auto resolved_slice = resolve_enumerations(slice);
+    auto field_name = std::optional<std::string>{};
+    for (const auto& [key, value] : parameters) {
+      if (key == "field") {
+        if (not value) {
+          return caf::make_error(ec::invalid_argument,
+                                 "invalid argument type for `field`: expected "
+                                 "a string");
+        }
+        field_name = *value;
+        continue;
+      }
+    }
+    if (not field_name) {
+      return caf::make_error(ec::invalid_argument, "missing argument `field`");
+    }
+    auto field_builder = series_builder{};
+    auto column_offset = slice.schema().resolve_key_or_concept(*field_name);
+    if (not column_offset) {
+      for (auto i = size_t{0}; i < slice.rows(); ++i) {
+        field_builder.null();
+      }
+      return field_builder.finish();
+    }
+    auto [type, slice_array] = column_offset->get(resolved_slice);
+    for (const auto& value : values(type, *slice_array)) {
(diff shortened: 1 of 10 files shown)


tenzir/functional-test/tests/bloom-filter-context.bats
```diff
@@ -0,0 +1,25 @@
+export TENZIR_PLUGINS="context,enrich"
```
mavam, line 1: @lava or @tobim: how do get this to work?

I'm getting errors in CI:

```
not ok 1 Bloom filter context in 3167ms
# (from function `setup_node' in file /__w/tenzir/tenzir/tenzir/functional-test/tests/../bats-tenzir/src/fixtures.bash, line 22,
#  from function `setup' in test file /__w/tenzir/tenzir/tenzir/functional-test/tests/bloom-filter-context.bats, line 9)
#   `setup_node' failed
# !! invalid_configuration: failed to find the context plugin as libtenzir-plugin-context.so in ["/__w/tenzir/tenzir/build/lib/tenzir/plugins/"]
# /__w/tenzir/tenzir/tenzir/functional-test/tests/../bats-tenzir/src/fixtures.bash: line 26: kill: (16875) - No such process
```

libtenzir/builtins/contexts/bloom_filter.cpp
```diff
@@ -0,0 +1,240 @@
...
+    if (not parsers::real(*fpp, p)) {
+      return caf::make_error(ec::invalid_argument, "--fp-prob is not a double");
+    }
+    if (n == 0) {
+      return caf::make_error(ec::invalid_argument, "--capacity must be > 0");
+    }
+    if (p < 0.0 || p > 1.0) {
+      return caf::make_error(ec::invalid_argument, "--fp-prob not in (0,1)");
+    }
+    return std::make_unique<bloom_filter_context>(n, p);
```
dominiklohmann, line 223: Please iterate over all key-value pairs in `parameters` so that you can return a meaningful error when an option is provided that may not be used with the bloom-filter context.

libtenzir/builtins/contexts/bloom_filter.cpp
```diff
@@ -0,0 +1,240 @@
...
+    }
+    auto p = double{0.0};
+    if (not parsers::real(*fpp, p)) {
+      return caf::make_error(ec::invalid_argument, "--fp-prob is not a double");
+    }
+    if (n == 0) {
+      return caf::make_error(ec::invalid_argument, "--capacity must be > 0");
+    }
+    if (p < 0.0 || p > 1.0) {
+      return caf::make_error(ec::invalid_argument, "--fp-prob not in (0,1)");
```
dominiklohmann, line 221:
```suggestion
      return caf::make_error(ec::invalid_argument, "--fp-prob not in [0,1]");
```

libtenzir/builtins/contexts/bloom_filter.cpp
```diff
@@ -0,0 +1,240 @@
...
+  auto name() const -> std::string override {
+    return "bloom-filter";
+  }
+
+  auto make_context(context::parameter_map parameters) const
+    -> caf::expected<std::unique_ptr<context>> override {
+    auto capacity = parameters["capacity"];
+    if (not capacity)
+      return caf::make_error(ec::parse_error, "no --capacity provided");
+    auto fpp = parameters["fp-prob"];
```
dominiklohmann, line 205: Looking at other bloom-filter implementations, it seems to me that this option is most commonly called `fp-probability`. I recommend switching to that.

libtenzir/include/tenzir/dcso_bloom_hasher.hpp
```diff
@@ -0,0 +1,123 @@
...
+        // For compound values, we assume that users provide values in JSON to
+        // Bloom.
+        auto json = to_json(materialize(value));
+        TENZIR_ASSERT(json);
+        h(*json);
+      } else {
+        // By default, we convert to string. For cases that may have multiple
+        // representations, such as durations, we need to have a dialogue with
+        // our users before committing to a fixed string representation.
+        h(to_string(value));
```
dominiklohmann, line 100: Please add another `if constexpr` case here that lists the remaining options, and then make the `else` case use `static_assert(always_false_v<decltype(value)>, "...")`. This will make it so that we are automatically made aware of this code when we add to the data variant.

web/docs/contexts/bloom-filter.md
```diff
@@ -0,0 +1,64 @@
...
+| `n` | Capacity | The maximum number of unique elements that guarantee the configured false-positive probability
+| `m` | Size | The number of bits that the Bloom filter occupies
+| `p` | False positive probability | The probability of erroneously reporting an element to be in the set
+
+The Bloom filter implementation is a C++ rebuild of DCSO's
+[bloom](https://github.com/DCSO/bloom) library. It is binary-compatible and uses
+the exact same method for FNV1 hashing and parameter calculation, making it a
+drop-in replacement for `bloom` users.
+
+### `-n|--capacity <capacity>`
```
dominiklohmann, line 44: There are no short-options for context parameters, so they also should not be in the docs. This also applies to the false-positive probability below.

dominiklohmann (approved):
This is nice to have. I left some comments inline that should be rather easy to address, so I'm already approving.

tenzir/functional-test/tests/bloom-filter-context.bats
dominiklohmann: You can't write these tests here. The context plugin is closed-source, so you will have to add these tests to the context plugin instead.

I recommend waiting for https://github.com/tenzir/tenzir/pull/3886 to do this, which should very soon be merged.

libtenzir/builtins/contexts/bloom_filter.cpp
mavam: Depends on whether you prefer the ]0,1[ or (0,1) syntax for half-open intervals, but I think this was correct beforehand, as neither 0.0 or 1.0 is a valid value.

dominiklohmann: Then you probably want to adjust the if to exclude 0 and 1 as well.

tenzir/functional-test/tests/bloom-filter-context.bats
mavam: I removed the tests and will migrate them as a follow-up.

libtenzir/builtins/contexts/bloom_filter.cpp
mavam: Solved in [15e245a](https://github.com/tenzir/tenzir/pull/3834/commits/15e245a58d6294ece98c1e20b55e85366f72f0c7).

libtenzir/include/tenzir/dcso_bloom_hasher.hpp
mavam: Please take another look at 53f907f4e7. I think this does what you wanted.

dominiklohmann: Yes, precisely. Thank you!

dominiklohmann (approved):

PR #3836, opened 2024-01-19
Print Zeek TSV metadata when schema changes

dominiklohmann:
Fixes tenzir/issues#587

This addresses two issues in the `zeek-tsv` printer:
1. Previously, the printer returned opening and closing tags once per batch rather than when the schema changes. That is not technically incorrect, but bloated the generated file sizes by a lot for small batches.
2. The printer was incorrectly marked as not allowing joining the output. Zeek TSV can be printer interleaved, so things like `to stdout write zeek-tsv` should not be forbidden.

### DIFF
--- a/libtenzir/builtins/formats/zeek_tsv.cpp
+++ b/libtenzir/builtins/formats/zeek_tsv.cpp
@@ -816,17 +816,22 @@ public:
       auto input_type = caf::get<record_type>(input_schema);
       auto array
         = to_record_batch(resolved_slice)->ToStructArray().ValueOrDie();
-      auto should_print_close = static_cast<bool>(*last_schema);
-      auto first = std::exchange(*last_schema, input_schema) != input_schema;
+      auto first = true;
+      auto is_first_schema = static_cast<bool>(*last_schema);
+      auto did_schema_change
+        = *last_schema != input_schema and not is_first_schema;
+      *last_schema = input_schema;
       for (const auto& row : values(input_type, *array)) {
         TENZIR_ASSERT_CHEAP(row);
         if (first) {
-          if (should_print_close) {
-            printer.print_closing_line(out_iter);
+          if (did_schema_change) {
+            if (not is_first_schema) {
+              printer.print_closing_line(out_iter);
+            }
+            printer.print_header(out_iter, input_schema);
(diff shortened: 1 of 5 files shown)


libtenzir/builtins/formats/zeek_tsv.cpp
```diff
@@ -801,19 +801,29 @@ class zeek_tsv_printer final : public plugin_printer {
...
       auto buffer = std::vector<char>{};
       auto out_iter = std::back_inserter(buffer);
       auto resolved_slice = flatten(resolve_enumerations(slice)).slice;
       auto input_schema = resolved_slice.schema();
       auto input_type = caf::get<record_type>(input_schema);
       auto array
         = to_record_batch(resolved_slice)->ToStructArray().ValueOrDie();
-      auto first = true;
+      auto should_print_close = static_cast<bool>(*last_schema);
+      auto first = std::exchange(*last_schema, input_schema) != input_schema;
```
mavam, line 820: Incredibly hard to read...can we please simplify this?

mavam (approved):

libtenzir/builtins/formats/zeek_tsv.cpp
dominiklohmann: Would it be easier to read if I renamed `first` to `did_schema_change`?

mavam: Oh, I meant the `std::exchange` part. This reads very cryptic. Can't we make this easier to read?

But yeah, renaming `should_print_close` could be `schema_changed`.

dominiklohmann: I personally find the exchange to be quite readable, but I suppose it's something that one would need to get used to. I can move the swap to the line below, as in this particular case it doesn't make any difference.

mavam: Unless there's a performance reason to do this, I would love to keep things simple for those with less deep C++ knowledge. That's my only motivation. And to be honest, I would have to look up the `std::exchange` idiom, even though that we would be know the 3rd time or so.

dominiklohmann: I rewrote this to be a bit more readable.

PR #3840, opened 2024-01-19
Support setting a single `TENZIR_TOKEN` for platform auth

dominiklohmann:
The new `tenzir.token` option has the format
`<tenant-id>:<api-key>[:<control-endpoint>]`, and has a higher precedence than the existing variables.

### DIFF
--- a/tenzir.yaml.example
+++ b/tenzir.yaml.example
@@ -7,6 +7,10 @@ tenzir:
   # The host and port to listen at and connect to.
   endpoint: "localhost:5158"
 
+  # The platform auth token for app.tenzir.com. This is an alternative to
+  # the dedicated platform.yaml file. The key has the following format:
+  #token: <tenant-id>:<api-key>[:<control-endpoint>]
+
   # The timeout for connecting to a Tenzir server. Set to 0 seconds to wait
   # indefinitely.
   connection-timeout: 5m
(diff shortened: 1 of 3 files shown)


tenzir.yaml.example
```diff
@@ -7,6 +7,10 @@ tenzir:
   # The host and port to listen at and connect to.
   endpoint: "localhost:5158"
 
+  # The platform auth token for app.tenzir.com. This is an alternative to
+  # the dedicated platform.yaml file. The key has the following format:
+  #token: <tenant-id>:<api-key>[:<control-endpoint>]
```
mavam, line 12:
```suggestion
  #platform-token: <tenant-id>:<api-key>[:<control-endpoint>]
```

WDYT?

dominiklohmann: I prefer the brevity of `TENZIR_TOKEN` over `TENZIR_PLATFORM_TOKEN` to be honest. But it's not a very strong opinion. Would be curious to hear what the @tenzir/platform-squad thinks first, since I kind of made this change without a lot of discussion.

mavam: > But it's not a very strong opinion.

FWIW, me neither. It just felt more intuitive, and I wanted to share my first gut reaction.

dominiklohmann:
Closing this; I'll bring this back in the future.

PR #3842, opened 2024-01-20
Fail properly when transfer breaks

mavam:
This PR improves error handling when something like `load http://does.not.exist` fails. Previously, such a pipeline ran through successfully.

NB: I removed the `tenzir::uri` abstraction, which was not used outside `tenzir::transfer`. We have already too many URI abstractions and our parser was everything but complete. It doesn't make sense to re-invent the wheel here. The one from libcurl is battle-tested, so is the one from Boost. This removal also fixes an issue where our parser was too strict and required trailing slashes after the host in the URI.

Best reviewed commit-by-commit.


dominiklohmann (approved):
Tested this locally; it works quite well, and I'm happy to see us going down from five URI parsers (Arrow, Boost, CAF, CURL, Tenzir) in the codebase to "just" four. 

The change overall makes sense. Good to see that this also removes the need for the trailing slash.

PR #3847, opened 2024-01-22
Add `lines` printer

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1135

Adds a new printer, `lines`, which is essentially an alias to `ssv --no-header`.

Doesn't allow specifying an extractor as an argument, even though that was specified in the issue.


libtenzir/builtins/formats/lines.cpp
```diff
@@ -104,8 +105,49 @@ class lines_parser final : public plugin_parser {
...
+
+  auto instantiate(type input_schema, operator_control_plane& ctrl) const
+    -> caf::expected<std::unique_ptr<printer_instance>> override {
+    TENZIR_ASSERT_CHEAP(inner_);
+    return inner_->instantiate(std::move(input_schema), ctrl);
+  }
+
+  auto allows_joining() const -> bool override {
+    TENZIR_ASSERT_CHEAP(inner_);
+    return inner_->allows_joining();
```
dominiklohmann, line 132: Since we don't print schema information we can always join the output across multiple schemas.

dominiklohmann (approved):

tenzir/integration/tests.yaml
```diff
@@ -1129,6 +1129,10 @@ tests:
         input: data/cef/checkpoint.log
       - command: exec 'read lines -s | summarize n=count(.) | write json -c'
         input: data/cef/checkpoint.log
+      - command: exec 'read json | write lines'
+        input: data/json/all-types.json
+      - command: exec 'read json | put e | write lines'
+        input: data/json/all-types.json
```
dominiklohmann, line 1135: cc @lava: new tests in the old format.

libtenzir/builtins/formats/lines.cpp
eliaskosunen: Making this return `true` unconditionally will cause an assertion failure in `xsv_printer`.

eliaskosunen: `allows_joining()` causes the `write` operator to choose a different code path, which `xsv_printer` doesn't support.

dominiklohmann: Does it make sense to generally allow the `xsv_printer` to join between schemas when not printing a header?

eliaskosunen: That would mean some refactoring in the `xsv_printer`. I'll write an issue about that, and follow up separately.

tenzir/integration/tests.yaml
lava: Can we please rewrite this as a BATS tests? Otherwise there's a danger that these new tests will get accidentally swallowed, since the BATS pr is going to remove the whole `tenzir/integration` directory.

PR #3848, opened 2024-01-22
Support setting profiles in the Velociraptor config

dominiklohmann:
Fixes tenzir/issues#1449

### DIFF
--- a/web/docs/operators/velociraptor.md
+++ b/web/docs/operators/velociraptor.md
@@ -108,7 +108,7 @@ Specifies the configuration profile for the Velociraptor instance. This enables
 connecting to multiple Velociraptor instances from the same Tenzir node.
 
 To use profiles, edit your `velociraptor.yaml` configuration like this, where
-`<config>` refers to the configuration file created by Velociraptor, and
+`<config>` refers to the contents of the configuration file created by Velociraptor, and
 `<profile>` to the desired profile name.
 
 ```yaml
--- a/web/docs/operators/velociraptor.md
+++ b/web/docs/operators/velociraptor.md
@@ -121,6 +121,8 @@ profiles:
     <config>
 ```
 
+If profiles are defined, the operator defaults to the first profile.
+
 ## Examples
 
 Show all processes:
--- a/web/docs/operators/velociraptor.md
+++ b/web/docs/operators/velociraptor.md
@@ -14,6 +14,7 @@ Submits VQL to a Velociraptor server and returns the response as events.
 velociraptor [-n|--request-name <string>] [-o|--org-id <string>]
              [-r|--max-rows <uint64>] [-s|--subscribe <artifact>]
              [-q|--query <vql>] [-w|--max-wait <duration>]
+             [--profile <profile>]
 ```
 
 ## Description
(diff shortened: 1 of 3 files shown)


web/docs/operators/velociraptor.md
```diff
@@ -101,6 +102,25 @@ Controls how long to wait before releasing a partial result set.
 
 Defaults to `1 sec`.
 
+### `--profile <profile>`
+
+Specifies the configuration profile for the Velociraptor instance. This enables
+connecting to multiple Velociraptor instances from the same Tenzir node.
+
+To use profiles, edit your `velociraptor.yaml` configuration like this, where
+`<config>` refers to the configuration file created by Velociraptor, and
```
mavam, line 111:
```suggestion
To use profiles, edit your `velociraptor.yaml` configuration like this, where
`<config>` refers to the contents of the configuration file created by Velociraptor, and
```

web/docs/operators/velociraptor.md
```diff
@@ -101,6 +102,25 @@ Controls how long to wait before releasing a partial result set.
...
+`<profile>` to the desired profile name.
+
+```yaml
+# before
+<config>
+
+# after
+profiles:
+  <profile>:
+    <config>
```
mavam, line 121:
```suggestion
# after
<profile>:
  <config>
```

IIRC we discussed something like this initially. Was that not possible or undesired?

mavam: I think this would be even fine in the default case of exactly one Velo instance.

mavam (commented):
Looks good, but didn't take out for a spin yet.

I'm sure @lo-chr is digging this one.

web/docs/operators/velociraptor.md
dominiklohmann: I considered it, but it'd lead to worse error messages if there's conflict between top-level keys in the Velociraptor configuration and the profile name. The solution with the single top-level section makes this a lot less likely to happen (unless Velociraptor introduces a top-level `profiles` section).

I think being able to report good error messages is more important than one level of indentation in a config file that's write-once-read-never-again.

mavam: I see what you see, but I'm trying to be more radical: do not make profiles an opt-in, but *require* them. Then you will *always* have the profile name as top-level record key and adding additional profiles will be a breeze.

dominiklohmann: I'm not a fan of that option. It just makes the setup for first-time users harder, and it also makes the usage harder as a profile name would always need to be specified. It'd also be a breaking change for current users of the operator.

mavam: I think your proposal is too complex because there are now two ways of doing things. My maxim is keeping variance as low possible for users. I don't see the extra indent causing any trouble. You're basically just naming your Velo server, which seems good practice anyway.

My proposal would be: you have to provide a `--name` or `--profile` as soon as there is more than a single server defined in your config, i.e., the number of fields in that records is greater than one. The setup cost is negligible compared to getting the Velo server up and running. It's a one-time setup cost anyway that quickly amortizes. The net result is that the overall configuration is stable and simple.

(Since there are only *very* few users, the breaking change argument doesn't have a lot of weight.)

dominiklohmann: We discussed this in a call and came to an agreement to keep the top-level profiles key, but to use the first configured profile by default if profiles are used.

tobim (approved):

PR #3854, opened 2024-01-23
Enable the AMQP plugin in the static binary

dominiklohmann:
Whoops.


tobim (approved):

tobim (approved):
I verified that the debian package dependency works as expected.

PR #3857, opened 2024-01-23
Add a CI job that ensures our OpenAPI spec is up-to-date

dominiklohmann:
This has historically gone out of sync quite often. This ensures it cannot happen anymore.

Fixes tenzir/issues#976


tobim (approved):
I trust that this works as expected.

PR #3860, opened 2024-01-24
Run functional tests when building the static binary

tobim:
The Nix package expression is refactored into 2 derivations. The first
builds the binary and package as before, and the newly introduced second
derivation runs the functional tests. This allows us to iterate on the
tests without having to rebuild c++ code with every change.

The untested binary can still be accessed with the the `unchecked`
attribute, e.g. `...#tenzir-static.unchecked`, `...#tenzir-de.unchecked`
and so on.

### DIFF
--- a/plugins/fluent-bit/integration/tests/tests.bats
+++ b/plugins/fluent-bit/integration/tests/tests.bats
@@ -20,7 +20,18 @@ setup() {
 }
 
 @test "read stdin via fluent-bit" {
-  echo '{"foo": {"bar": 42}}' | check tenzir 'fluent-bit stdin | drop timestamp'
+  # fluent-bit has an internal race condition that can cause the loss of
+  # events during startup. We have to accept this for now and avoid
+  # test flakyness with a bit of trickery.
+  result="$(
+    while :; do
+      echo '{"foo": {"bar": 42}}'
+      sleep 1
+    done |
+      tenzir 'fluent-bit stdin | drop timestamp | head 1' |
+      grep -v 'kevent'
+  )"
+  check echo "$result"
 }
 
 @test "fluent-bit works as a sink" {
--- a/tenzir/integration/tests/tcp.bats
+++ b/tenzir/integration/tests/tcp.bats
@@ -1,4 +1,4 @@
-: "${BATS_TEST_TIMEOUT:=3}"
+: "${BATS_TEST_TIMEOUT:=10}"
 
 setup() {
   bats_load_library bats-support
--- a/.github/workflows/nix.nu
+++ b/.github/workflows/nix.nu
@@ -130,7 +130,7 @@ export def run [
   # Run local effects by building all requested editions.
   let targets = ($cfg.editions | each {|e| $".#(attribute_name $e)" })
   print $"::notice building ($targets)"
-  nix --accept-flake-config --print-build-logs build --no-link $targets
+  nix --accept-flake-config --print-build-logs build --no-link ...$targets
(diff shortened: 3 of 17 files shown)


plugins/fluent-bit/integration/tests/tests.bats
```diff
@@ -20,7 +20,11 @@ setup() {
 }
 
 @test "read stdin via fluent-bit" {
-  echo '{"foo": {"bar": 42}}' | check tenzir 'fluent-bit stdin | drop timestamp'
+  # fluent-bit has an internal race condition that can cause the loss of
+  # events during startup. We have to accept this for now and avoid
+  # test flakyness with a bit of trickery.
+  { echo '{"foo": {"bar": 42}}'; sleep 1; echo '{"foo": {"bar": 42}}'; } \
+    | check tenzir 'fluent-bit stdin | drop timestamp | tail 1'
```
dominiklohmann, line 27: Given your explanation of the race condition, shouldn't this suffice?

```suggestion
  { sleep 1; echo '{"foo": {"bar": 42}}'; } \
    | check tenzir 'fluent-bit stdin | drop timestamp'
```

tenzir/integration/tests/tcp.bats
```diff
@@ -1,4 +1,4 @@
-: "${BATS_TEST_TIMEOUT:=3}"
+: "${BATS_TEST_TIMEOUT:=10}"
```
dominiklohmann, line 1: Is there a good reason why we specify a test timeout for this suite that separate from the globally configured one in the first place?

dominiklohmann (approved):

plugins/fluent-bit/integration/tests/tests.bats
tobim: I don't know. It depends on whether the init logic is triggered on startup or when the first input arrives.

dominiklohmann: The engine is created before the first yield, i.e., as part of the operator's startup.

tenzir/integration/tests/tcp.bats
tobim: The test sometimes ran into the 3 second timeout when they are run in parallel. I don't see a problem in upping the threshold a bit for that.

dominiklohmann: My suggestion was to axe the entire line and to just use the default instead of setting a custom one in the first place, assuming a default value exists.

plugins/fluent-bit/integration/tests/tests.bats
tobim: I was referring to what happens in fluent-bit internally.
In any case, I observed the problem once now even with the change as is.

tenzir/integration/tests/tcp.bats
tobim: In that case the timeout for every test would be dictated by the one test with the highest expected runtime.
I believe setting per-file thresholds provides a better testing UX.

plugins/fluent-bit/integration/tests/tests.bats
tobim: I updated the test once more. It doesn't look as clean as in the beginning, but it was to only way to make it reliable without adding redirects directly to the binary.

tenzir/integration/tests/tcp.bats
dominiklohmann: Makes sense. I don't have too strong of an opinion on this, so let's just keep it as-is then.

PR #3861, opened 2024-01-24
Lift single-schema requirement from `write xsv --no-header`

dominiklohmann:
... and thus implicitly from `write csv --no-header`, `write ssv --no-header`, `write tsv --no-header` , and `write lines`.

Fixes tenzir/issues#1471


eliaskosunen:
Integrated into https://github.com/tenzir/tenzir/pull/3847. Closing.

PR #3864, opened 2024-01-24
Force braces around statements with clang-format

dominiklohmann:
We've changed our coding style to always using braces almost two years ago now, and it turns out that there's actually a clang-format option to automatically rewrite code to have statements around braces for improved legibility.


jachris (approved):
🤩

PR #3865, opened 2024-01-24
Schedule idle operators less aggressively

dominiklohmann:
This reduces the CPU usage of `load tcp://localhost:3000 | discard` where no data arrives over TCP drastically by re-implementing exponential backoff for the execution node's internal run loop.

Fixes tenzir/issues#1483

### DIFF
--- a/libtenzir/src/execution_node.cpp
+++ b/libtenzir/src/execution_node.cpp
@@ -661,7 +661,7 @@ struct exec_node_state {
     } else if (demand.has_value()) {
       // If we shouldn't continue, but there is an upstream demand, then we may
       // be in a situation where the operator has internally buffered events and
-      // needs to be polled until some oeprator-internal timeout expires before
+      // needs to be polled until some operator-internal timeout expires before
       // it yields the results. We use exponential backoff for this with 25%
       // increments.
       schedule_run(true);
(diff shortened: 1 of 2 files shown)


libtenzir/src/execution_node.cpp
```diff
@@ -621,16 +641,23 @@ struct exec_node_state {
...
     //   a. The upstream operator has completed.
-    //   b. The operator has issued demand that is not yet answered.
-    //   c. The operator can produce output independently from receiving input.
-    //   d. The operator has input it can consume.
-    const auto should_continue = instance->it != instance->gen.end()  // (1)
-                                 and (not previous                    // (2a)
-                                      or issue_demand_inflight        // (2b)
-                                      or op->input_independent()      // (2c)
-                                      or not inbound_buffer.empty()); // (2d)
+    //   b. The operator has upstream demand and can produce output
```
tobim, line 644: You mean "downstream" right?

dominiklohmann: Yes, absolutely. There is no such thing as upstream demand.

```suggestion
    //   b. The operator has downstream demand and can produce output
```

jachris (approved):

PR #3869, opened 2024-01-25
Update the main repository to include timestamped pipelines

Dakostu:
This change is another submodule bump with a small OpenAPI update.


contrib/tenzir-plugins
dominiklohmann, line 1: Please add a changelog entry for this, and when it's merged create a follow-up story to include this information in the app.

dominiklohmann (approved):

contrib/tenzir-plugins
Dakostu: Changelog entry added.

changelog/next/features/3869.md
```diff
@@ -0,0 +1,7 @@
+The new `created_at` field tracks a managed pipeline's creation time.
+
+The new `last_modified` field tracks a managed pipeline's last manual
+modification time.
+
+Pipelines that were created before this change will have the creation and
+modification time set to the pipeline manager launch time.
```
dominiklohmann, line 7: Paragraphs render as individual changelog entries. Here's a rewording suggestion that fixes that and I think reads a bit better.

```suggestion
`show pipelines` and the `/pipeline` API endpoints now include `created_at` and `last_modified` fields that track the pipeline's creation and last manual modification time, respectively. Pipelines created with older versions of Tenzir will use the start time of the node as their creation time.
```

PR #3871, opened 2024-01-26
Add structured_data to syslog output

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1072

Previously, the `structured_data` field in new-RFC syslog messages was parsed, but wasn't included in the output. This was chiefly because the data was in a `tenzir::map`, which our data model doesn't support well. This PR changes it to be a `list` of key-value-pair-`record`s, that are included in the output.


web/docs/formats/syslog.md
```diff
@@ -40,15 +40,25 @@ With this input, the parser will produce the following output, with the schema n
...
+  ],
   "message": "Event log entry"
 }
 ```
 
-:::note Structured Data
-The STRUCTURED-DATA field defined in RFC 5424, delimited by square brackets `[]` in the input,
-is currently not included in the resulting schema.
-:::
-
```
dominiklohmann, line 51: The examples below need to be updated to include the `structured_data` field.

dominiklohmann (approved):
This is a really straightforward change. Just one small request for the docs.

web/docs/formats/syslog.md
eliaskosunen: `structured_data` is only applicable for RFC 5424 syslog messages, so there isn't going to be a field for them in 3164 messages.

dominiklohmann: Makes sense. 🙈

PR #3872, opened 2024-01-26
Remove dead code: `inventory` and `netflow`

dominiklohmann:
Neither of these plugins are supported anymore. The `inventory` plugin has been phased out a while anymore and wasn't included in any builds, and the `netflow` plugin has never been ported to the new parser framework.


mavam (approved):
Good to detach them know, we'll bring 'em back once interest grows stronger again.

PR #3874, opened 2024-01-27
Gracefully handle misaligned header and values in `xsv` parser

dominiklohmann:
The `xsv` parser (and by extension the `csv`, `tsv`, and `ssv` parsers) previously skipped lines that had a mismatch between the number of values contained and the number of fields defined in the header. This is now handled in a more lenient way in both directions:
1. If the header contains more fields than a row contains data, the remaining fields are filled in as `null` values.
2. If a row contains more values than the header had fields defined for, we dynamically expand the header with fields nammed `unnamed1`, `unnammed2`, and so forth.

These two examples show the new behavior:

```
❯ echo "1,2,3,4,5" | tenzir -q 'read csv --header "foo,bar,baz"'
{
  "foo": 1,
  "bar": 2,
  "baz": 3,
  "unnamed1": 4,
  "unnamed2": 5
}
```

```
❯ echo "1" | tenzir -q 'read csv --header "foo,bar,baz"'
{
  "foo": 1,
  "bar": null,
  "baz": null
}
```

Fixes https://github.com/tenzir/issues/issues/1500


mavam:
I liked this.

One idea re. naming: Instead of starting the unnamed field at 1, we could start at their field number. In your first example, that'd be:

```json
{
  "foo": 1,
  "bar": 2,
  "baz": 3,
  "field4": 4,
  "field5": 5
}
```

dominiklohmann:
> ```json
>   "field4": 4,
> ```

This breaks as soon as there's a collision. The current implementation skips over field names that already exist in the data:

```
❯ echo '1,2,3,4,5' | tenzir 'read csv --header "foo,unnamed2,bar" | write csv'
foo,unnamed2,bar,unnamed1,unnamed3
1,2,3,4,5
```

libtenzir/builtins/formats/xsv.cpp
```diff
@@ -365,21 +364,26 @@ auto parse_impl(generator<std::optional<std::string_view>> lines,
...
+        continue;
+      }
+      auto result = row.field(fields[i]).try_data(values[i]);
       if (not result) {
         diagnostic::warning(result.error())
           .note("from `{}` parser", args.name)
           .emit(ctrl.diagnostics());
       }
+      continue;
     }
```
jachris, line 387:
```suggestion
    for (size_t i = 0; i < fields.size(); ++i) {
      if (i >= values.size()) {
        row.field(fields[i]).null();
        continue;
      }
      auto result = row.field(fields[i]).try_data(values[i]);
      if (not result) {
        diagnostic::warning(result.error())
          .note("from `{}` parser", args.name)
          .emit(ctrl.diagnostics());
      }
    }
```

libtenzir/builtins/formats/xsv.cpp
jachris, line 1: I'm not sure whether such leniency a good default behavior. A mismatch in the number of expected fields and actual fields usually hints to some kind of problem, or at least should be something the user should be made aware of. Maybe it's better as an opt-in. What do you think?

jachris (approved):

libtenzir/builtins/formats/xsv.cpp
dominiklohmann: I think that adding nulls for missing fields should just be the default without an option. No reason not to have it.

I think that we should consider hiding adding fields for additional values behind a flag.

changelog/next/bug-fixes/3874--xsv-leniency.md
dominiklohmann, line 1: After https://github.com/tenzir/tenzir/pull/3823 is merged, we should add some tests for this. I didn't do this yet because I didn't want to have the inevitable merge conflict.

libtenzir/builtins/formats/xsv.cpp
mavam: > I'm not sure whether such leniency a good default behavior. 

We try to follow the rationale "be lenient in what you accept and be strict in what you send." So I think it's in line if we try to iron out kinks in the data. This refers to filling in null values when a record is too short.

> I think that we should consider hiding adding fields for additional values behind a flag.

When a record is too long, this indeed can be a major problem. I would opt-in here as well and add an `--auto-extend` field to signal that the XSV parser will extend the schema if need be.

dominiklohmann: I like the name `auto-extend` for the option. I'll go with that for now.

PR #3875, opened 2024-01-28
Improve the Style Check CI

dominiklohmann:
This commit improves the _Style Check_ CI workflow. It contains the following changes:
- Every style check now runs in a new job, causing them to run in parallel. This allows developers to see at a glance which checks failed, rather than having to re-run after having fixed one failure only to see that the next style check also fails.
- For BATS test definitions, this adds formatting checks via `shfmt`.
- Where possible, the style check now also runs for submodules.


tenzir/functional-test/tests/vast.bats
```diff
@@ -198,20 +183,10 @@ teardown() {
...
 
-  # We need subsecond precision here because `date` rounds
-  # down otherwise. Also, we need `tr` because `date` uses
-  # a comma for subsecond precision by default but tenzir
-  # requires a dot. (ISO 8601 allows both)
-  NOW=$(date -Ins | tr ',' '.')
-  check tenzir "export | where #import_time > ${NOW}"
-  check tenzir "export | where #import_time <= ${NOW} | sort timestamp"
+  check tenzir "export | where #import_time > now"
+  check tenzir "export | where #import_time <= now | sort timestamp"
```
lava, line 189: I think these are nice additional test cases, but it's not quite the same as the original test which verifies that the assigned import time is consistent with the system time.

lava (approved):

tenzir/functional-test/tests/vast.bats
dominiklohmann: I pushed a fix for this so that it runs correctly on macOS.

PR #3876, opened 2024-01-28
Allow using libtenzir FlatBuffers schemas from plugins

dominiklohmann:
With this change, plugins using `TenzirCompileFlatBuffers` can now use the `include "<file>";` directive to include FlatBuffers files from libtenzir like `type.fbs` or `data.fbs`.

I came across this deficiency while working on another plugin, and I think this is best reviewed separately, so I split it out into a PR of its own. Right now, this is just a build scaffolding change that is only user-visible in that the installed package now includes `*.fbs` files.

Fixes https://github.com/tenzir/issues/issues/1498

### DIFF
--- a/cmake/TenzirRegisterPlugin.cmake
+++ b/cmake/TenzirRegisterPlugin.cmake
@@ -198,6 +198,22 @@ function (TenzirCompileFlatBuffers)
     INTERFACE $<BUILD_INTERFACE:${output_prefix}>
               $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 
+  if ("${CMAKE_PROJECT_NAME}" STREQUAL "Tenzir")
+    set(FBS_PATH "${CMAKE_SOURCE_DIR}/libtenzir/fbs")
+  else ()
+    string(TOUPPER "${TENZIR_CMAKE_BUILD_TYPE}" build_type_uppercase_)
+    get_target_property(FBS_PATH tenzir::tenzir
+                        IMPORTED_LOCATION_${build_type_uppercase_})
+    unset(build_type_uppercase_)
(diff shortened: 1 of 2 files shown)


cmake/TenzirRegisterPlugin.cmake
```diff
@@ -245,13 +263,14 @@ function (TenzirCompileFlatBuffers)
...
+    add_custom_target("compile-flatbuffers-schema-${FBS_TARGET}-${basename}"
                       DEPENDS "${desired_file}")
-    add_dependencies(${FBS_TARGET} "compile-flatbuffers-schema-${basename}")
+    add_dependencies(${FBS_TARGET}
+                     "compile-flatbuffers-schema-${FBS_TARGET}-${basename}")
     set_property(
       TARGET ${FBS_TARGET}
       APPEND
-      PROPERTY PUBLIC_HEADER "${desired_file}")
+      PROPERTY PUBLIC_HEADER "${desired_file}" "${schema}")
```
dominiklohmann, line 273: This is a bit of a hack to install the `*.fbs` files—we could also just install the entire directory into a location other than `<includedir>/tenzir/include/fbs` with a bit more effort, if that's preferred.

cmake/TenzirRegisterPlugin.cmake
```diff
@@ -198,6 +198,22 @@ function (TenzirCompileFlatBuffers)
     INTERFACE $<BUILD_INTERFACE:${output_prefix}>
               $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 
+  if ("${CMAKE_PROJECT_NAME}" STREQUAL "Tenzir")
```
dominiklohmann, line 201: If you want to test this working, just add `include "data.fbs";` at the top of any `*.fbs` file in any closed-source plugin. That used to fail.

eliaskosunen (approved):

PR #3883, opened 2024-01-30
Update the main repository to include the pipeline run ID

Dakostu:
This change is another submodule pointer + OpenAPI docs bump.

### DIFF
--- a/changelog/next/features/3883.md
+++ b/changelog/next/features/3883.md
@@ -1,0 +1,3 @@
+Managed pipelines now contain a new `total_runs` parameter that counts all
+started runs. The new `run` field is available in the events delivered by the
+`metrics` and `diagnostics` operators.
--- a/web/docs/operators/diagnostics.md
+++ b/web/docs/operators/diagnostics.md
@@ -35,6 +35,7 @@ Contains detailed information about the diagnostic.
 |Field|Type|Description|
 |:-|:-|:-|
 |`pipeline_id`|`string`|The ID of the pipeline that created the diagnostic.|
+|`run`|`uint64`|The number of the run, starting at 1 for the first run.|
 |`timestamp`|`time`|The exact timestamp of the diagnostic creation.|
 |`message`|`string`|The diagnostic message.|
 |`severity`|`string`|The diagnostic severity.|
--- a/web/docs/operators/metrics.md
+++ b/web/docs/operators/metrics.md
@@ -66,6 +66,7 @@ operator instantiation.
 |Field|Type|Description|
 |:-|:-|:-|
 |`pipeline_id`|`string`|The ID of the pipeline where the associated operator is from.|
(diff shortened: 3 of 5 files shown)


changelog/next/features/3883.md
```diff
@@ -0,0 +1,3 @@
+Managed pipelines now contain a new `total_runs` parameter that counts all
+started runs. This counter also serves as the new `run_id` for the
```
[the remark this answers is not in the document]
Dakostu: Well, our current pipeline IDs are numbers, too...
Internally the `run_id` for diagnostics and metrics is a string, just like `pipeline_id` - I can change the name, but then we'd have a naming misalignment between `pipeline_id` and `run`. IDK if that is a big problem though.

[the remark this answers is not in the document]
dominiklohmann: I think `pipeline_id` should stay a string. That it always contains a number happened more by accident than by design. At some point in the future, we should just use something like `fmt::to_string(uuid::random())` for the id when we create a new pipeline.

`run`/`run_id` should be a number, as otherwise the user cannot perform arithmetic operations with it—which they're supposed to do if they want to compare it to `total_runs` for the same pipeline.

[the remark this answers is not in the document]
Dakostu: I implemented `run` as a number now.

web/docs/operators/metrics.md
```diff
@@ -66,6 +66,7 @@ operator instantiation.
 |Field|Type|Description|
 |:-|:-|:-|
 |`pipeline_id`|`string`|The ID of the pipeline where the associated operator is from.|
+|`run`|`uint64`|A unique identifier for the run the associated metric is from.|
```
dominiklohmann, line 69:
```suggestion
|`run`|`uint64`|The number of the run, starting at 1 for the first run.|
```

web/docs/operators/diagnostics.md
```diff
@@ -35,6 +35,7 @@ Contains detailed information about the diagnostic.
 |Field|Type|Description|
 |:-|:-|:-|
 |`pipeline_id`|`string`|The ID of the pipeline that created the diagnostic.|
+|`run`|`uint64`|A unique identifier for the run the associated diagnostic is from.|
```
dominiklohmann, line 38:
```suggestion
|`run`|`uint64`|The number of the run, starting at 1 for the first run.|
```

dominiklohmann (approved):

web/docs/operators/metrics.md
Dakostu: Applied your suggestion.

web/docs/operators/diagnostics.md
Dakostu: Applied your suggestion.

dominiklohmann (approved):

PR #3886, opened 2024-01-30
BATS transition follow-up cleanup

lava:
Some further improvements to our bats setup:

* Always use `TENZIR_BARE_MODE=true` for tests
* Further simplify the `setup_suite.bash` boilerplate by moving datadir computations into the library
* Refactor naming of the `setup_node*` library functions
* Move the last integration suites for the context and compaction plugins to bats


tenzir/functional-test/lib/bats-tenzir/src/fixtures.bash
```diff
@@ -1,10 +1,26 @@
...
 
 export_default_node_config() {
-  export TENZIR_DB_DIRECTORY="${BATS_TEST_TMPDIR}/db"
+  # Normalize the environment unless `BATS_TENZIR_KEEP_ENVIRONMENT` is set.
+  if [[ ! -n ${BATS_TENZIR_KEEP_ENVIRONMENT} ]]; then
+    unset $(printenv | grep -o '^TENZIR[^=]*' | paste -s -)
+  fi
+
   export TENZIR_BARE_MODE=true
+  export TENZIR_DB_DIRECTORY="${BATS_TEST_TMPDIR}/db"
```
dominiklohmann, line 23: I think this needs to be a new directory every time the function runs. Otherwise, we risk problems with tests running in parallel that use import or export.

tenzir/functional-test/tests/setup_suite.bash
```diff
@@ -2,29 +2,25 @@ setup_suite() {
...
-export BATS_TENZIR_DATADIR="${BATS_TEST_DATADIR%%/functional-test/*}/data"
-
-export INPUTSDIR="${BATS_TENZIR_DATADIR}/inputs"
-export QUERYDIR="${BATS_TENZIR_DATADIR}/queries"
-export MISCDIR="${BATS_TENZIR_DATADIR}/misc"
+# Try to load the `bats-tenzir` library either relative to
+# the current directory (for in-tree builds) or relative to the
+# tenzir binary (for builds against an installed tenzir)
+TENZIR_DIR="$(realpath "$(dirname "$(command -v tenzir)")")"
+export BATS_LIB_PATH=${BATS_LIB_PATH:+${BATS_LIB_PATH}:}${TENZIR_DIR}/../share/tenzir/functional-test/lib:${BATS_TEST_DIRNAME}/../../../../../tenzir/functional-test/lib
```
dominiklohmann, line 26: Can you add a comment here explaining why we step out 5 times?

dominiklohmann (approved):
Thank you!

tenzir/functional-test/lib/bats-tenzir/src/fixtures.bash
lava: `BATS_TEST_TMPDIR` is a bats built-in variable that is unique per test run (according to the docs at least, i didn't actually try running in parallel yet)

We could move it to `setup_node()` so that its impossible to accidentally forget to reset it, but on the other hand then it would also be impossible to overwrite it for custom tests.

dominiklohmann: Ah, I assumed that it was unique per suite. Makes sense, then we don't need to do anything here.

tenzir/functional-test/tests/setup_suite.bash
dominiklohmann: When I commented this I did not know that we have this copied over for every plugin. Let's revisit this later.

lava: There is one in line 24, but I guess i should reword it to be more obvious :D

tobim:
Can you add the following diff to fix the python tests:
```diff
diff --git a/python/tests/test_tenzir.py b/python/tests/test_tenzir.py
index 67db62d2e4..0f0c085cc9 100644
--- a/python/tests/test_tenzir.py
+++ b/python/tests/test_tenzir.py
@@ -56,7 +56,7 @@ async def tenzir_exec(endpoint, pipeline: str):
 
 def integration_data(path):
     dir_path = os.path.dirname(os.path.realpath(__file__))
-    return os.path.normpath(f"{dir_path}/../../tenzir/functional-test/data/{path}")
+    return os.path.normpath(f"{dir_path}/../../tenzir/functional-test/data/inputs/{path}")
 
 
 @pytest.mark.asyncio
```

PR #3887, opened 2024-01-30
Add more details to OpenWEC section

satta:
This PR adds some more details about setting up an OpenWEC server to ingest Windows events into Tenzir. It distills the most important steps down so they can be concisely shown within the integration documentation.
There are also small typographical fixes.

### DIFF
--- a/web/docs/integrations/windows-event-logs.md
+++ b/web/docs/integrations/windows-event-logs.md
@@ -456,12 +456,125 @@ Refer to the [OpenWEC getting started
 guide](https://github.com/cea-sec/openwec/blob/main/doc/getting_started.md) for
 setup instructions.
 
+For running OpenWEC on non-Windows machines that are likely not joined to a
+Windows domain, it is most useful to configure TLS client authentication. The
+OpenWEC documentation (see above) recommends to use [a script collection from
+NXLog](https://gitlab.com/nxlog-public/contrib/-/tree/master/windows-event-forwarding)
+that  creates keys/certificates that can immediately be used to configure both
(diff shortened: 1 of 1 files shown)


mavam (approved):
Awesome, thanks a bunch for contributing these details!

I pushed one commit with smaller touch ups.

One remaining question: it looks like you configured OpenWEC to send the logs to `tcp://localhost:1514`. Is that correct? the subsequent section on configuring the Tenzir pipeline uses 10.0.0.1. I wonder whether we need to adapt this.

satta:
Thanks for the changes. I didn't know the engine you are using to render your documentation could also include filenames as hints!

I intended to include the full output of the scripts though so readers do not expect an empty response from these scripts. If you're not doing it like that anywhere else in the documentation, then I agree that should be skipped here too.

> One remaining question: it looks like you configured OpenWEC to send the logs to `tcp://localhost:1514`. Is that correct? the subsequent section on configuring the Tenzir pipeline uses 10.0.0.1. I wonder whether we need to adapt this.

Right, just noticed that too. That indeed needs to be consistent, let's use the same IP here. Can you just make that quick change please?

mavam:
> I intended to include the full output of the scripts though so readers do not expect an empty response from these scripts. If you're not doing it like that anywhere else in the documentation, then I agree that should be skipped here too.

Fair point. So far we haven't done this in the docs anywhere. If we include (copious) output, we use the `<details><summary>..` tags to wrap it. I think we could do this here as well if you think the output is helpful.

> Right, just noticed that too. That indeed needs to be consistent, let's use the same IP here. Can you just make that quick change please?

Will do.

PR #3888, opened 2024-01-30
Use the new M1 runners for the macOS CI

dominiklohmann:
See also: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

### DIFF
--- a/.github/workflows/tenzir.yaml
+++ b/.github/workflows/tenzir.yaml
@@ -1020,7 +1020,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.10"]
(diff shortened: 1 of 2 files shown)


mavam (approved):
🍎

tobim:
We have 2 CI errors here:
* `pipx` is missing on the new runner.
* `http_parser` is marked as deprecated in Homebrew and fails to install. We can either try to script `brew edit` to remove the deprecation or update to `restinio` 0.7.1. But I think a package for `llhttp` is missing in Homebrew. I'd try to go with the first option for a quick unblock.

PR #3889, opened 2024-01-30
Replace `tenzir.db-directory` with `tenzir.state-directory`

dominiklohmann:
The option `tenzir.db-directory` is quite old, and comes from a time when Tenzir wasn't all about pipelines, but rather mostly about its storage engine VAST. Nowadays, we store lots of state in this directory that is not part of a "database" in the classical sense. As such, the option is now called `tenzir.state-directory`, with the old option name still being supported for the foreseeable future in a backwards-comaptible way.

Additionally, this change makes it possible to set the `tenzir.cache-directory` on the command line when starting the node.

Fixes tenzir/issues#1109

### DIFF
--- a/tenzir.yaml.example
+++ b/tenzir.yaml.example
@@ -20,7 +20,16 @@ tenzir:
   # available:
   #   - $STATE_DIRECTORY
   #   - $PWD/tenzir.db
-  #db-directory:
+  #state-directory:
+
+  # The file system path used for persistent state.
+  # Defaults to one of the following paths, selecting the first that is
+  # available:
+  #   - $CACHE_DIRECTORY
+  #   - $XDG_CACHE_HOME/tenzir
+  #   - $HOME/.cache/tenzir (Linux)
+  #   - $HOME/Library/Caches/tenzir (macOS)
+  #cache-directory:
 
   # The file system path used for persistent state.
   # Defaults to one of the following paths, selecting the first that is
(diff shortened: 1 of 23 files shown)


tenzir.yaml.example
```diff
@@ -38,7 +47,7 @@ tenzir:
   # Defaults to one of the following paths, selecting the first that is
   # available:
   #   - $LOGS_DIRECTORY/server.log
-  #   - <db-directory>/server.log
+  #   - <state-directory>/server.log
```
mavam, line 50: Might be a follow-up, but IIRC we also wanted something like a `log-directory`?

dominiklohmann: Ideally yes, but I'm not sure it's worth the effort of figuring out how to make that backwards-compatible with `tenzir.log-file`. Having a clearer split between persistent state and cache directories was something we actually have a need for.

mavam: Would it then make sense to move the logs over into the cache directory, given that they are non-essential?

dominiklohmann: I think they are very much essential.

mavam: I don't have a strong opinion here. I would imagine that users that care about the logs feed them into their o11y tool of choice anyway.

dominiklohmann: Some day we'll have the `logs` operator and then users can do whatever they want with it.

mavam: Hmmm, maybe. Before that, we may need an OTel logging sink and adopt structured logging.

mavam (approved):

PR #3891, opened 2024-01-30
Drop unit tests for pipelines for good

dominiklohmann:
All of these tests are better covered by the integration tests written with BATS. Writing unit tests for pipelines has always been a hassle, so nobody really invested time into maintaining them.

Fixes tenzir/issues#955


eliaskosunen (approved):
Many lines removed, none added, I like it

PR #3893, opened 2024-01-31
Implement a context content dumping mechanism

Dakostu:
This change implements a context content dumping mechanism for debugging purposes.

### DIFF
--- a/libtenzir/builtins/contexts/bloom_filter.cpp
+++ b/libtenzir/builtins/contexts/bloom_filter.cpp
@@ -111,7 +111,8 @@ public:
     if (bloom_filter_.parameters().k) {
       params.field("k", *bloom_filter_.parameters().k);
     }
-    co_yield entry_builder.finish_assert_one_slice(dump_event_name());
+    co_yield entry_builder.finish_assert_one_slice(
+      fmt::format("tenzir.{}.info", context_type()));
   }
 
   /// Updates the context.
--- a/libtenzir/builtins/contexts/bloom_filter.cpp
+++ b/libtenzir/builtins/contexts/bloom_filter.cpp
@@ -91,6 +91,29 @@ public:
     };
   }
 
+  auto dump() -> generator<table_slice> override {
+    auto ptr = reinterpret_cast<const std::byte*>(bloom_filter_.data().data());
+    auto size = bloom_filter_.data().size();
+    auto data = std::basic_string<std::byte>{ptr, size};
+    auto entry_builder = series_builder{};
+    auto row = entry_builder.record();
+    row.field("num_elements", bloom_filter_.num_elements());
+    auto params = row.field("parameters").record();
+    if (bloom_filter_.parameters().m) {
+      params.field("m", *bloom_filter_.parameters().m);
+    }
+    if (bloom_filter_.parameters().n) {
+      params.field("n", *bloom_filter_.parameters().n);
+    }
+    if (bloom_filter_.parameters().p) {
+      params.field("p", *bloom_filter_.parameters().p);
+    }
+    if (bloom_filter_.parameters().k) {
+      params.field("k", *bloom_filter_.parameters().k);
+    }
+    co_yield entry_builder.finish_assert_one_slice(dump_event_name());
+  }
+
   /// Updates the context.
   auto update(table_slice slice, context::parameter_map parameters)
     -> caf::expected<update_result> override {
--- a/libtenzir/builtins/contexts/lookup_table.cpp
+++ b/libtenzir/builtins/contexts/lookup_table.cpp
@@ -103,11 +103,13 @@ public:
       row.field("key", key);
       row.field("value", value);
       if (entry_builder.length() >= context::dump_batch_size_limit) {
-        co_yield entry_builder.finish_assert_one_slice(dump_event_name());
+        co_yield entry_builder.finish_assert_one_slice(
+          fmt::format("tenzir.{}.info", context_type()));
       }
     }
     // Dump all remaining entries that did not reach the size limit.
-    co_yield entry_builder.finish_assert_one_slice(dump_event_name());
+    co_yield entry_builder.finish_assert_one_slice(
+      fmt::format("tenzir.{}.info", context_type()));
   }
 
   /// Updates the context.
--- a/libtenzir/builtins/contexts/lookup_table.cpp
+++ b/libtenzir/builtins/contexts/lookup_table.cpp
@@ -96,6 +96,14 @@ public:
     return record{{"num_entries", context_entries.size()}};
   }
 
+  auto dump() const -> list override {
+    auto l = list{};
+    for (const auto& [key, value] : context_entries) {
+      l.emplace_back(record{{"key", key}, {"value", value}});
+    }
+    return l;
+  }
+
   /// Updates the context.
   auto update(table_slice slice, context::parameter_map parameters)
     -> caf::expected<update_result> override {
--- a/libtenzir/builtins/contexts/lookup_table.cpp
+++ b/libtenzir/builtins/contexts/lookup_table.cpp
@@ -96,12 +96,26 @@ public:
     return record{{"num_entries", context_entries.size()}};
   }
 
-  auto dump() const -> list override {
-    auto l = list{};
+  auto dump_data_entries() const -> generator<table_slice> {
+    auto entry_builder = series_builder{};
     for (const auto& [key, value] : context_entries) {
-      l.emplace_back(record{{"key", key}, {"value", value}});
+      entry_builder.data(record{{"key", key}, {"value", value}});
+      auto slices = entry_builder.finish_as_table_slice();
+      for (auto&& slice : slices) {
+        co_yield slice;
+      }
     }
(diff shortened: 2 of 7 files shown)


libtenzir/builtins/contexts/lookup-table.cpp
```diff
@@ -125,6 +125,28 @@ class ctx final : public virtual context {
...
 
+  auto dump_data_entries() const -> generator<table_slice> {
+    auto entry_builder = series_builder{};
+    for (const auto& [key, value] : context_entries) {
+      entry_builder.data(record{{"key", key}, {"value", value}});
+      auto slices = entry_builder.finish_as_table_slice();
+      for (auto&& slice : slices) {
+        co_yield slice;
+      }
+    }
```
dominiklohmann, line 136: This does two things we shouldn't do:
- It creates batches of size 1.
- It allocates a new record, only to then transpose it back into a columnar structure when adding it to the builder.

I'd suggest rewriting the loop (1) to add key and value separately to the builder, instead of allocating an intermediate row-major representation and (2) to only finish every $`2^{16}`$ entries.

libtenzir/builtins/contexts/lookup-table.cpp
```diff
@@ -125,6 +125,28 @@ class ctx final : public virtual context {
...
+  auto dump() -> caf::expected<std::vector<table_slice>> override {
+    if (not entry_dumper) {
+      entry_dumper = dump_data_entries();
+    }
+    for (const auto&& x : *entry_dumper) {
+      return std::vector<table_slice>{x};
+    }
+    entry_dumper.reset();
+    return std::vector<table_slice>{};
+  }
```
dominiklohmann, line 148: I don't understand what this function does.

libtenzir/include/tenzir/plugin.hpp
```diff
@@ -798,6 +798,9 @@ class context {
   /// Inspects the context.
   virtual auto show() const -> record = 0;
 
+  /// Dumps the context content or a dumping error.
+  virtual auto dump() -> caf::expected<std::vector<table_slice>> = 0;
```
dominiklohmann, line 802: How does this "stream" events to the `context inspect` operator when it returns a single vector as opposed to a generator?

dominiklohmann (changes requested):

libtenzir/include/tenzir/plugin.hpp
Dakostu: The generator right now is inside the context implementation. Launching `dump()` advances that generator and returns the yielded element in a vector.

This would probably be way easier (and more correct) if I could just straight up return a `generator<table_slice>` to the operator. But how would I make a `generator` inspectable?

dominiklohmann: You can't transfer the generator to the operator, but you can at least return a generator from the context and hold the instantiated coroutines in the context actor instead of the context itself.

Dakostu: That is what I have been actually working on right now. Seems like a good compromise.

dominiklohmann: FWIW, if we ever manage to actually upgrade to CAF 0.19 then we can transfer generators between actors, but unfortunately it's not possible until then. Once the database plugin is done we can start migrating to the newest version of CAF because then we'll finally have gotten rid of all the places where we currently use CAF streaming.

libtenzir/builtins/contexts/lookup-table.cpp
Dakostu: This will be gone.

libtenzir/include/tenzir/plugin.hpp
Dakostu: I made use of storing `generator`s in the `context_actor`.

libtenzir/builtins/contexts/bloom_filter.cpp
```diff
@@ -91,6 +91,31 @@ class bloom_filter_context final : public virtual context {
...
+    if (bloom_filter_.parameters().n) {
+      params.field("n", *bloom_filter_.parameters().n);
+    }
+    if (bloom_filter_.parameters().p) {
+      params.field("p", *bloom_filter_.parameters().p);
+    }
+    if (bloom_filter_.parameters().k) {
+      params.field("k", *bloom_filter_.parameters().k);
+    }
+    row.field("data", data);
```
dominiklohmann, line 114: This does not belong in here. We cannot just put a potentially gigabyte-large string into an event. There's `context save` if the user wants to have this.

libtenzir/builtins/contexts/bloom_filter.cpp
```diff
@@ -91,6 +91,31 @@ class bloom_filter_context final : public virtual context {
...
+    co_yield entry_builder.finish_assert_one_slice(dump_event_name());
+    co_return;
```
dominiklohmann, line 116:
```suggestion
```

dominiklohmann (approved):
Some remaining notes, but approving already.

libtenzir/builtins/contexts/bloom_filter.cpp
Dakostu: Removed in https://github.com/tenzir/tenzir/pull/3893/commits/99652cea4c33742e6f0b7e36b1046368493dc474

libtenzir/builtins/contexts/bloom_filter.cpp
Dakostu: Removed the `data` field in https://github.com/tenzir/tenzir/pull/3893/commits/99652cea4c33742e6f0b7e36b1046368493dc474

libtenzir/builtins/contexts/lookup_table.cpp
```diff
@@ -225,6 +249,7 @@ class ctx final : public virtual context {
 
 private:
   tsl::robin_map<data, data> context_entries;
+  std::optional<generator<table_slice>> entry_dumper;
```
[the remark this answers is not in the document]
Dakostu: Yes.. old dumper from a previous approach. Removed in https://github.com/tenzir/tenzir/pull/3893/commits/99652cea4c33742e6f0b7e36b1046368493dc474

libtenzir/builtins/contexts/lookup_table.cpp
```diff
@@ -96,6 +96,30 @@ class ctx final : public virtual context {
...
+        }
+      }
+    }
+    if (entry_builder.length() > 0) {
+      // Dump all remaining entries that did not reach the size limit.
+      for (auto&& slice :
+           entry_builder.finish_as_table_slice(dump_event_name())) {
+        co_yield slice;
+      }
+    }
```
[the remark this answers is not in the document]
Dakostu: Changed in https://github.com/tenzir/tenzir/pull/3893/commits/99652cea4c33742e6f0b7e36b1046368493dc474

libtenzir/include/tenzir/plugin.hpp
```diff
@@ -731,13 +733,19 @@ class context {
 
   virtual auto context_type() const -> std::string = 0;
 
+  virtual auto dump_event_name() const -> std::string {
+    return fmt::format("tenzir.{}.info", context_type());
+  }
```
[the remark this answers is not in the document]
Dakostu: I removed `dump_event_name()` in https://github.com/tenzir/tenzir/pull/3893/commits/4f8a412aa555db5fd615412ac9410a4506726a32

libtenzir/builtins/contexts/lookup_table.cpp
```diff
@@ -96,6 +96,30 @@ class ctx final : public virtual context {
...
   }
 
+  auto dump() -> generator<table_slice> override {
+    auto entry_builder = series_builder{};
+    auto size = 0;
+    for (const auto& [key, value] : context_entries) {
+      auto row = entry_builder.record();
+      row.field("key", key);
+      row.field("value", value);
+      ++size;
```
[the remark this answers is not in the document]
Dakostu: Removed unnecessary counter in https://github.com/tenzir/tenzir/pull/3893/commits/99652cea4c33742e6f0b7e36b1046368493dc474

libtenzir/builtins/contexts/lookup_table.cpp
```diff
@@ -96,6 +96,30 @@ class ctx final : public virtual context {
...
+      if (size % context::dump_batch_size_limit == 0) {
+        auto slices = entry_builder.finish_as_table_slice(dump_event_name());
```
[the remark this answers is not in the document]
Dakostu: Changed in https://github.com/tenzir/tenzir/pull/3893/commits/99652cea4c33742e6f0b7e36b1046368493dc474

libtenzir/builtins/contexts/lookup_table.cpp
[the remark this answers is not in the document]
Dakostu: And https://github.com/tenzir/tenzir/pull/3893/commits/fb8b31385841791cd4dc18a072477c10aa56de65

PR #3894, opened 2024-01-31
Do the integration switcharoo

dominiklohmann:
This removes our old integration framework, and renames the functional test framework to integration.


mavam (approved):
Back to where we've been a while ago, nice. 😉

PR #3897, opened 2024-02-01
Introduce the shared diagnostics handler

dominiklohmann:
This introduces a diagnostics handler that is safe to use from other threads, allowing operators, connectors, and formats that spawn threads through third-party libraries or other actors to emit diagnostics.

To create a shared diagnostics handler, call `ctrl.shared_diagnostics()` instead of `ctrl.diagnostics()` on the operator's control plane.

Unlike for the regular diagnostics handler, errors emitted through the shared handler do not shut down the operator's execution node immediately. Diagnostics sent through the shared diagnostics handler may arrive out of order.

This is a first step towards being able to emit errors in the context API. For the `shell` transformation operator, this fixes an unsafe use of the regular diagnostic handler from another thread.

Fixes tenzir/issues#1073

### DIFF
--- a/libtenzir/builtins/operators/shell.cpp
+++ b/libtenzir/builtins/operators/shell.cpp
@@ -206,14 +206,15 @@ public:
     // queue such that to this coroutine can yield them.
     auto chunks = std::queue<chunk_ptr>{};
     auto chunks_mutex = std::mutex{};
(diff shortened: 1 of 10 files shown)


jachris (approved):
Thanks! :pray:

PR #3899, opened 2024-02-01
Remove reader, writer, and language plugin types

dominiklohmann:
Now that the old tests have all been ported, there is no more use for the reader and writer plugin types. They have long been superseded by the parser and printer plugin type, respectively.

While looking at this I also identified that the language plugin has effectively been dead code for quite a while now, so I've removed that as well.


contrib/tenzir-plugins
jachris, line 1: Not sure whether you want to update the plugins to this commit.

jachris (approved):
+1 for sneaking in some additional changes.

contrib/tenzir-plugins
dominiklohmann: I had to because I needed to remove an unused include of utils.hpp from the compaction tests.

dominiklohmann: I will wait with merging this until after the run counter change landed, then this bump should go away.

PR #3900, opened 2024-02-01
Handle nested fields and integers as selectors in JSON parser

jachris:
This makes it so that `read json --selector=foo.bar` searches the `foo` record for a field `bar`. Previously, we looked for a top-level field `foo.bar`.


libtenzir/builtins/formats/json.cpp
```diff
@@ -482,14 +482,18 @@ class doc_parser {
...
-  if (auto err = type.error()) {
-    if (err != simdjson::error_code::NO_SUCH_FIELD)
+  if (auto err = object.error()) {
+    if (err != simdjson::error_code::NO_SUCH_FIELD) {
       return caf::make_error(ec::parse_error, error_message(err));
+    }
     return std::string{unknown_entry_name};
   }
-  auto maybe_schema_name = type.value_unsafe().get_string();
+  auto maybe_schema_name = object.value_unsafe().get_string();
```
dominiklohmann, line 496: Haven't tried it yet, but I think this will crash if it's not a string, as that isn't checked above.

dominiklohmann (approved):

libtenzir/builtins/formats/json.cpp
jachris: I think this is fine, because `get_string()` should return an error if it is not a string, which is checked right below.

Hence having for example an integer gives us a `warning: INCORRECT_TYPE: The JSON element does not have the requested type.`. But we could additionally try to convert at least integers to strings here.

dominiklohmann: For windows event logs we probably needs that as the selector field to use is likely the event id.

jachris: I added an integer -> string conversion as an additional feature to this PR.

PR #3901, opened 2024-02-01
Add new '--file' option to the python operator

lava:


tenzir/integration/tests/setup_suite.bash
```diff
@@ -19,5 +19,6 @@ if ! command -v tenzir; then
 fi
 
 libpath_relative_to_binary="$(realpath "$(dirname "$(command -v tenzir)")")/../share/tenzir/integration/lib"
-libpath_relative_to_pwd="${BATS_TEST_DATADIR%%/integration/*}/lib"
+libpath_relative_to_pwd="${BATS_TEST_DIRNAME%%/integration/*}/integration/lib"
```
dominiklohmann, line 22: I don't understand this change, can you explain? Wasn't this something we needed to copy over to every single plugin?

libtenzir/builtins/operators/python.cpp
```diff
@@ -354,15 +378,28 @@ class plugin final : public virtual operator_plugin<python_operator> {
...
                                             "operators/transformations/python"};
     parser.add("-r,--requirements", requirements, "<requirements>");
+    parser.add("-f,--file", filename, "<filename>");
     parser.add(command, "<command>");
     parser.parse(p);
+    if (!filename.empty() == command.has_value()) {
+      diagnostic::error(
+        "Must have exactly one of the '--file' argument or inline code")
+        .throw_();
+    }
```
dominiklohmann, line 394: If both are present, you probably want to have a different error message that refers to the location of both arguments. You can get them by making the arguments `std::optional<located<std::string>>` instead.

Generally, the error messages start lower-case, and we use backticks rather than single quotes for referencing code.

libtenzir/builtins/operators/python.cpp
```diff
@@ -77,19 +77,43 @@ class python_operator final : public crtp_operator<python_operator> {
...
+              code_chunk->size()};
+          },
+          [](std::string inline_code) -> caf::expected<std::string> {
+            return inline_code;
+          }},
+        code_);
+      if (!maybe_code) {
+        diagnostic::error("failed to obtain code")
+          .note(fmt::format("error: {}", maybe_code.error()))
+          .emit(ctrl.diagnostics());
```
dominiklohmann, line 111:
```suggestion
        diagnostic::error(maybe_code.error())
          .note("failed to obtain code")
          .emit(ctrl.diagnostics());
```

libtenzir/builtins/operators/python.cpp
```diff
@@ -77,19 +77,43 @@ class python_operator final : public crtp_operator<python_operator> {
...
+        co_return;
+      }
+      auto code = *maybe_code;
+      TENZIR_INFO("got code {}", code);
```
dominiklohmann, line 115: This looks like a leftover from debugging.

dominiklohmann (approved):
Lots of error handling suggestions, but the functionality works, so approving this already.

tenzir/integration/tests/setup_suite.bash
lava: `BATS_TEST_DATADIR` straight up doesn't exist, the other plugins all already have the correct `BATS_TEST_DIRNAME` so I think it was just an oversight here that was just never noticed because the other path already contained the bats libraries.

tenzir/integration/tests/vast.bats
```diff
@@ -70,7 +70,7 @@ teardown() {
   check tenzir "export | where resp_h == 216.240.189.196 | sort ts"
 }
 
-# bats test_tags=node,import,export,zeek
+# bats test_tags=node,import,export,zeek,bats:focus
```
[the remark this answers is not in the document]
lava: Right, good thing that feature has an automatic "fail CI" safeguard built in :D

libtenzir/builtins/operators/python.cpp
```diff
@@ -354,15 +377,32 @@ class plugin final : public virtual operator_plugin<python_operator> {
...
+    parser.add("-f,--file", filename, "<filename>");
     parser.add(command, "<command>");
     parser.parse(p);
+    if (!filename && !command) {
+      diagnostic::error("must have either the `--file` argument or inline code")
+        .throw_();
+    }
+    if (filename && command) {
+      diagnostic::error(
+        "cannot have `--file` argument together with inline code")
```
dominiklohmann, line 395: Here you can use `.primary(*filename).primary(*command)` to highlight both visually in the generated diagnostic. Both are primary since they both equally contribute to the diagnostic.

tenzir/integration/tests/setup_suite.bash
dominiklohmann: Ah, that makes sense. Thanks!

PR #3902, opened 2024-02-01
Remove wrong value updates for paused pipeline metrics

Dakostu:
This change fixes time delta calculation for paused pipelines: Dismiss any deltas in metrics for a paused pipeline and add a `time_paused_total` metric to fix further duration-based calculations.

### DIFF
--- a/changelog/next/bug-fixes/3902--remove-wrong-values-for-paused-pipeline-metrics.md
+++ b/changelog/next/bug-fixes/3902--remove-wrong-values-for-paused-pipeline-metrics.md
@@ -1,0 +1,2 @@
+Durations in pipeline metrics now correctly exclude time that a pipeline was
+paused.
--- a/libtenzir/src/execution_node.cpp
+++ b/libtenzir/src/execution_node.cpp
@@ -70,16 +70,6 @@ struct defaults<chunk_ptr> : defaults<> {
 
 namespace {
 
-template <class... Duration>
-  requires(std::is_same_v<Duration, duration> && ...)
-auto make_timer_guard(Duration&... elapsed) {
-  return caf::detail::make_scope_guard(
-    [&, start_time = std::chrono::steady_clock::now()] {
-      const auto delta = std::chrono::steady_clock::now() - start_time;
-      ((void)(elapsed += delta, true), ...);
-    });
-}
-
 // Return an underestimate for the total number of referenced bytes for a vector
 // of table slices, excluding the schema and disregarding any overlap or custom
 // information from extension types.
@@ -188,9 +178,11 @@ auto size(const chunk_ptr& chunk) -> uint64_t {
 }
 
 struct metrics_state {
-  auto emit() -> void {
-    values.time_total = std::chrono::duration_cast<duration>(
-      std::chrono::steady_clock::now() - start_time);
+  auto emit(bool exec_node_paused) -> void {
+    if (not exec_node_paused) {
+      values.time_total = std::chrono::duration_cast<duration>(
+        std::chrono::steady_clock::now() - start_time - total_time_paused);
+    }
     caf::anon_send(metrics_handler, values);
   }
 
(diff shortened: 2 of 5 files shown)


libtenzir/src/execution_node.cpp
```diff
@@ -269,18 +264,32 @@ struct exec_node_state {
...
   }
 
+  template <class... Duration>
+    requires(std::is_same_v<Duration, duration> && ...)
+  auto make_timer_guard(Duration&... elapsed) {
+    return caf::detail::make_scope_guard(
+      [&, start_time = std::chrono::steady_clock::now()] {
+        if (paused) {
+          // Do not add a delta to paused pipeline metrics.
+          return;
```
dominiklohmann, line 280: This only affects the `pause` and `resume` message handlers: technically, they should still count towards scheduled time.

libtenzir/src/execution_node.cpp
```diff
@@ -188,16 +178,21 @@ auto size(const chunk_ptr& chunk) -> uint64_t {
...
 
 struct metrics_state {
-  auto emit() -> void {
-    values.time_total = std::chrono::duration_cast<duration>(
-      std::chrono::steady_clock::now() - start_time);
+  auto emit(bool exec_node_paused) -> void {
+    if (not exec_node_paused) {
+      values.time_total = std::chrono::duration_cast<duration>(
+        std::chrono::steady_clock::now() - start_time - total_time_paused);
+    }
```
dominiklohmann, line 185: Did you consider leaving the total time the wall-clock time, and adding the paused time as a new value in addition to it? What are the pros and cons for how we utilize the metrics?

libtenzir/src/execution_node.cpp
Dakostu: So the `pause` and `resume` handlers should force a delta? Or should every `time_scheduled` delta be recorded, regardless of whether the node has been paused before, since `time_scheduled` also records the time for pushing and pulling?

libtenzir/src/execution_node.cpp
Dakostu: Do you mean a new metric value `time_paused`, while leaving `time_total` as-is?

For the app, this would mean that it would probably have to do the `total - paused` calculation itself to display the proper runtime in the overview page.

Simply calling a `metrics` pipeline for a pipeline that has been paused would then, for those paused durations, display metrics with a `time_total`, but no other noteworthy delta. Someone who would work with the `metrics` output would most likely also have to apply that subtraction.

I guess it depends on what `time_total` actually _means_ - the total _run_ time? If we did that, we would unify the semantics of this particular metric. But I wouldn't mind turning that into the total _life_ time metric, too.

changelog/next/bug-fixes/3902--remove-wrong-values-for-paused-pipeline-metrics.md
```diff
@@ -0,0 +1,2 @@
+Paused pipelines will not deliver metrics with increasing duration values
+anymore. These durations will not increase during the paused state anymore.
```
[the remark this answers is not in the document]
Dakostu: I updated the changelog item.

libtenzir/src/execution_node.cpp
dominiklohmann: I think we can ignore this for now. It's not a big difference, not worth putting much time into this now.

libtenzir/src/execution_node.cpp
Dakostu: From our internal team-wide discussion, it looks like there is a slight consensus towards a separate `time_paused` metric. I'll go ahead and implement that.

Dakostu: I implemented a `time_paused_total` metric in https://github.com/tenzir/tenzir/pull/3902/commits/ab740247d7484622b0096064b91c9b7d20203fa0

Dakostu:
https://github.com/tenzir/tenzir/pull/3940 supersedes this.

PR #3903, opened 2024-02-02
Add support for commas in seconds in the time data parser

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1520

Makes `double_parser` a template, that accepts the expected radix separator as a parameter. `parsers::real` is `double_parser<'.'>`, and the new `parsers::real_comma` is `double_parser<','>`.

`parsers::real_detect_sep` is also added, that tries both `real` and `real_comma`, and yields whichever matched more characters. The seconds parser in `parsers::time` is changed to use that.

The end result is, that both dots and commas are accepted as separators for subsecond values in `parsers::time`, as specified by ISO 8601. I'm not 100% happy with the implementation, but it works.

### DIFF
--- a/libtenzir/include/tenzir/concept/parseable/tenzir/time.hpp
+++ b/libtenzir/include/tenzir/concept/parseable/tenzir/time.hpp
@@ -168,7 +168,7 @@ struct ymdhms_parser : tenzir::parser_base<ymdhms_parser> {
       [](auto x) { return x >= 0 && x <= 23; });
     auto min = integral_parser<int, 2, 2>{}.with(
       [](auto x) { return x >= 0 && x <= 59; });
-    auto sec = parsers::real.with([](auto x) {
+    auto sec = parsers::real_detect_sep.with([](auto x) {
       return x >= 0.0 && x <= 60.0;
     });
     auto time_divider = '+'_p | 'T' | ' ';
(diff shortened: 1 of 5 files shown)


libtenzir/include/tenzir/concept/parseable/tenzir/time.hpp
```diff
@@ -168,7 +168,7 @@ struct ymdhms_parser : tenzir::parser_base<ymdhms_parser> {
       [](auto x) { return x >= 0 && x <= 23; });
     auto min = integral_parser<int, 2, 2>{}.with(
       [](auto x) { return x >= 0 && x <= 59; });
-    auto sec = parsers::real.with([](auto x) {
+    auto sec = parsers::real_detect_sep.with([](auto x) {
```
dominiklohmann, line 171: This should probably get a changelog entry in the `bug-fixes` category.

tenzir/integration/tests/vast.bats
```diff
@@ -188,14 +188,12 @@ teardown() {
...
+    NOW=$(gdate -Ins)
   else
-    NOW=$(date -Ins | tr ',' '.')
+    NOW=$(date -Ins)
   fi
 
-  # We need subsecond precision here because `date` rounds down otherwise. Also,
-  # we need `tr` because `date` uses a comma for subsecond precision by default
-  # but tenzir requires a dot. (ISO 8601 allows both)
+  # We need subsecond precision here because `date` rounds down otherwise.
```
dominiklohmann, line 196: Can you move this comment up to before the first NOW? I think it makes a lot more sense there.

dominiklohmann (approved):
I think this works. Thank you!

PR #3906, opened 2024-02-05
Remove the -cm edition

tobim:
This relaxes the version string requirements to `MAJOR.MINOR.PATCH{DEV_SUFFIX}`.

In detail the changes are:
* No more optional `TWEAK` position. We never used it in the two years since it was added, so at this point it makes sense to get rid of the extra complexity.
* The rigid requirement of a `git describe` style version is replaced with a free-form `DEV_SUFFIX`. This is used in 2 locations: 
   - Docker builds in Pull Requests now use the PR number instead of the commit hash of the merge base. This is an improvement because it is less misleading and more stable against rebases of the feature branch.
   - The Nix build now uses its own internal output hash. Thanks to that, the package won't have to be rebuilt any more when none of the actual source files that go into the build change.


dominiklohmann:
> * We never used it in the two years since it was added, so at this point it makes sense to get rid of the extra complexity.

We actually use it in the app for version comparison, and I think that may need to be relaxed accordingly. We should double-check.

tobim:
> > * We never used it in the two years since it was added, so at this point it makes sense to get rid of the extra complexity.
> 
> We actually use it in the app for version comparison, and I think that may need to be relaxed accordingly. We should double-check.

I checked, and we use the npm package `semver-parser`. This led me to take another look at the [semver spec](https://semver.org/) and I discovered that it doesn't mention "tweak" at all.
While reading, I realized that `git describe` style versions are actually not valid semver. I added another commit here with the necessary changes to make our version strings compliant.

version.json
dominiklohmann, line 1: We also have a VERSIONING.md file in the repository that appears to be wholly outdated. Can you make a quick pass over that?

libtenzir/builtins/operators/version.cpp
```diff
@@ -22,10 +22,10 @@ class version_operator final : public crtp_operator<version_operator> {
     auto builder = series_builder{};
     auto event = builder.record();
     event.field("version", tenzir::version::version);
-    event.field("major", std::stoull(tenzir::version::major));
-    event.field("minor", std::stoull(tenzir::version::minor));
-    event.field("patch", std::stoull(tenzir::version::patch));
-    event.field("tweak", std::stoull(tenzir::version::tweak));
+    event.field("build", tenzir::version::build_metadata);
```
dominiklohmann, line 25: More of a nit for consistency, feel free to ignore this if you think it's not worth the effort.

```suggestion
    event.field("build", tenzir::version::build);
```

dominiklohmann (dismissed):
This is fantastic, thank you.

libtenzir/builtins/operators/version.cpp
tobim: I tried to implement this change, but it clashes with the `tenzir::version::build` namespace that is defined in `config.hpp.in`. I think it makes most sense to keep the longer name as it is used in https://semver.org/.

dominiklohmann (approved):
This is fantastic. Thank you!

PR #3908, opened 2024-02-05
Implement `context load`, `context save`, and `context reset`

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1523, companion PR: https://github.com/tenzir/tenzir-plugins/pull/175

 * `context update` was previously overloaded on its input type. Now it only accepts events. The bytes-overload is removed (it wasn't implemented by contexts), and the void-overload is called `context reset`.
 * `context load` and `context save` are added, that load from and serialize to bytes, respectively.
   * They are transformations on bytes<->events, and don't perform I/O themselves: the most common use case is probably that `load <serialized-binary-file>` is piped to `context load`, and `context save` is piped to `save <serialized-binary-file>`. A carbon-copy of a context can be created with `context save foo | context load bar`.
   * The contexts already implemented this functionality, the necessary plumbing to call them just wasn't there, yet

As a table:
| Before                             | After                         | Description                       |
|:-----------------------------------|:------------------------------|:----------------------------------|
| `<events> \| context update <ctx>` | (unchanged)                   | Update context with events        |      
| `<bytes> \| context update <ctx>`  | (removed, wasn't implemented) | Update context with bytes         |
| `context update <ctx> [flags]`     | `context reset <ctx> [flags]` | Reset context state               |
| N/A                                | `context save \| <bytes>`     | Serialize context save into bytes |
| N/A                                | `<bytes> \| context load`     | Load context state from bytes     |


Dakostu (commented):
Left some comments.

libtenzir/include/tenzir/plugin.hpp
```diff
@@ -727,6 +727,8 @@ class context {
 
   virtual ~context() noexcept = default;
 
+  virtual auto name() const -> std::string = 0;
```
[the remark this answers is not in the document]
eliaskosunen: I opted for `context_type()`.

libtenzir/builtins/contexts/geoip.cpp
```diff
@@ -378,14 +382,13 @@ class ctx final : public virtual context {
                            "geoip context can not be updated with events");
   }
 
-  auto update(chunk_ptr, context::parameter_map)
-    -> caf::expected<update_result> override {
-    return caf::make_error(ec::unimplemented, "geoip context can not be "
-                                              "updated with bytes");
+  auto make_query() -> caf::expected<make_query_type> override {
+    return caf::make_error(ec::unimplemented,
+                           "geoip context doesn't support retroactive lookups");
```
Dakostu, line 387:
```suggestion
    return {};
```

This error gets swallowed by the post-load query expression emittance loop. It will never be displayed.

Contexts that do not support retroactive lookups should return an empty function.

If anything, this error should appear when you want to _only_ create a retrospective lookup for a context that doesn't support them, but that's not in the scope of this PR.

libtenzir/builtins/contexts/bloom_filter.cpp
```diff
@@ -163,18 +167,17 @@ class bloom_filter_context final : public virtual context {
...
-    -> caf::expected<update_result> override {
-    // TODO: We're getting chunks piece-meal currently and therefore cannot
-    // determine the boundary of the stream. In theory, we should accumulate the
-    // chunks until the upstream pipeline is complete. What we need in the
-    // context API is a signal that we're done.
-    return ec::unimplemented;
+  auto make_query() -> caf::expected<make_query_type> override {
+    return caf::make_error(ec::unimplemented,
+                           "bloom filter doesn't support retroactive lookups "
+                           "after insertion");
```
Dakostu, line 173:
```suggestion
    return {};
```

This error gets swallowed by the post-load query expression emittance loop. It will never be displayed.

Contexts that do not support retroactive lookups should return an empty function.

If anything, this error should appear when you want to _only_ create a retrospective lookup for a context that doesn't support them, but that's not in the scope of this PR.

Dakostu (approved):
I preemptively approve now, with some small suggestions remaining.

Dakostu:
@eliaskosunen Please also add a changelog item regarding the new subcommands.

PR #3909, opened 2024-02-05
Improve the `export` operator

dominiklohmann:
This brings back the `--low-priority` option for the `export` operator. It was previously available for the `tenzir-ctl export` command. Low-priority exports may take longer than other exports to complete, and in return interfere less with other exports.

Additionally, this PR fixes a potential indefinite hang in the second to last execution node of a pipeline caused by a race condition. This sometimes caused pipelines to get get stuck until manually stopped.


lava (approved):
We looked at this together in a video call.

PR #3910, opened 2024-02-05
Update the repository to include lookup lifetime fixes

Dakostu:
This change is another submodule pointer bump, with a change to the "missing field" lookup error that hopefully makes it more usable.

### DIFF
--- a/libtenzir/builtins/contexts/bloom_filter.cpp
+++ b/libtenzir/builtins/contexts/bloom_filter.cpp
@@ -149,7 +149,8 @@ public:
       auto column = params["field"];
       if (not column) {
         return caf::make_error(ec::invalid_argument,
-                               "missing 'field' parameter");
+                               "missing 'field' parameter for lookup in "
+                               "bloom-filter");
       }
       return expression{
         predicate{
--- a/libtenzir/builtins/contexts/lookup_table.cpp
+++ b/libtenzir/builtins/contexts/lookup_table.cpp
@@ -182,7 +182,8 @@ public:
       auto column = params["field"];
       if (not column) {
         return caf::make_error(ec::invalid_argument,
-                               "missing 'field' parameter");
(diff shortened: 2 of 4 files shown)


libtenzir/builtins/contexts/bloom_filter.cpp
```diff
@@ -149,7 +149,8 @@ class bloom_filter_context final : public virtual context {
       auto column = params["field"];
       if (not column) {
         return caf::make_error(ec::invalid_argument,
-                               "missing 'field' parameter");
+                               "missing 'field' parameter for bloom "
+                               "filter-based lookup");
```
dominiklohmann, line 153: The hyphens are a bit off in this one. How about this?

> missing 'field' parameter for lookup in bloom-filter

libtenzir/builtins/contexts/lookup_table.cpp
```diff
@@ -182,7 +182,8 @@ class ctx final : public virtual context {
       auto column = params["field"];
       if (not column) {
         return caf::make_error(ec::invalid_argument,
-                               "missing 'field' parameter");
+                               "missing 'field' parameter for lookup "
+                               "table-based lookup");
```
dominiklohmann, line 186: The hyphens are a bit off in this one as well. How about this?

> missing 'field' parameter for lookup in lookup-table

dominiklohmann (approved):

libtenzir/builtins/contexts/bloom_filter.cpp
Dakostu: Applied your suggestion.

libtenzir/builtins/contexts/lookup_table.cpp
Dakostu: Applied your suggestion.

PR #3911, opened 2024-02-05
Fix logger deadlock in python tests

lava:


tenzir.yaml.example
```diff
@@ -77,6 +77,10 @@ tenzir:
   # Maximum number of log messages in the logger queue.
   log-queue-size: 1000000
 
+  # Behavior if the logger queue runs full. Possible values are 'overrun_oldest'
+  # and 'block'.
+  log-overflow-policy: overrun_oldest
```
dominiklohmann, line 82: Should this really be configurable? Why would you ever use `block`?

libtenzir/src/application.cpp
```diff
@@ -248,7 +247,6 @@ auto make_root_command(std::string_view name) {
   root->add_subcommand(make_count_command());
   root->add_subcommand(make_export_command());
   root->add_subcommand(make_import_command());
-  root->add_subcommand(make_start_command());
```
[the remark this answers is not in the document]
lava: Shouldn't we see it in integration tests if it would break that way? Just from the code in `application.cpp` it looks to me like the behavior of `tenzir-node` would not be changed:

```
    auto cmd = make_start_command();
    cmd->name = "";
    add_root_opts(*cmd);
    return {
      std::move(cmd),
      command::factory{
        {"", start_command},
      },
    };
```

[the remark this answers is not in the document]
dominiklohmann: Ah, I misunderstood it then. You didn't remove the `start_command` itself, just the call to register it as a subcommand. That makes sense.

tenzir.yaml.example
lava: Just from the behavior `block` actually makes more sense to me, since it doesn't lose any log data. In theory "block" shouldn't mean "deadlock", it should just pause the process until the logger thread has had time to catch up. I didn't investigate why exactly that didn't happen in our case.

But the real reason why I made it configurable is that we're changing the default, and this way we can easily revert to the previous behavior if the logging starts to break mysteriously afterwards.

changelog/next/features/3899--async-policy.md
```diff
@@ -0,0 +1,3 @@
+We introduced a new `tenzir.log-overflow-policy` settings key which
+allows the user to control the behavior when the log queue runs full.
+Accepted values are `block` or `overrun_oldest`.
```
dominiklohmann, line 3: This should mention the default value, especially if we want to change it from the old default of `block`

tenzir.yaml.example
dominiklohmann: Then asking the question the other way around: Why would a user not want to use `block`?

tobim: That was already answered above:

> In theory "block" shouldn't mean "deadlock", it should just pause the process until the logger thread has had time to catch up. I didn't investigate why exactly that didn't happen in our case.

Empirically `block` causes the server to lock up when the queue size limit gets hit.

dominiklohmann: Exposing this option seems a bit like shifting the responsibility to the user. For such system internals, how would a user know what value to choose?

I think we should not expose an option to our users that we don't have tests for and know to sometimes cause hangs.

libtenzir/src/logger.cpp
```diff
@@ -264,10 +264,22 @@ bool setup_spdlog(bool is_server, const tenzir::invocation& cmd_invocation,
     file_sink->set_pattern(file_format);
     sinks.push_back(file_sink);
   }
+  auto overflow_policy = caf::get_or(cfg_file, "tenzir.log-overflow-policy",
+                                     defaults::logger::overflow_policy);
```
dominiklohmann, line 268: There's related documentation at https://docs.tenzir.com/setup-guides/tune-performance#logging which should probably be updated to reflect the new default behavior.

tenzir.yaml.example
tobim: The reason for exposing it was also already stated:

> But the real reason why I made it configurable is that we're changing the default, and this way we can easily revert to the previous behavior if the logging starts to break mysteriously afterwards.

But I agree with your reasoning as well. Wdyt of *not* documenting this knob and removing it again after ~2 releases?

dominiklohmann: I'd rather remove it entirely, document the changed behavior, and if a user asks for it to be made configurable we bring the option back.

tobim: If things don't go as expected we'd have to roll out and distribute new builds in an emergency. Having one option less does not justify that risk.

tobim: We decided to keep it as a hidden option temporarily.

libtenzir/src/logger.cpp
tobim: We decided to make this option hidden and remove it again after the dust settled.

PR #3913, opened 2024-02-06
Change data model of syslog `structured_data`

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1542

This PR changes the schema of the `structured_data` field in RFC 5424 syslog messages. Previously, they were stored as lists of records, with fields `key` and `value`. With this PR, it's a simple record.

Because now the field names are not predetermined (`key` and `value`), but are determined by the input data, the type can't be described statically in a `type` definition. Consequently, the implementation for building one needed to be changed to use `series_builder` (instead of `table_slice_builder`), and be able to handle differences in the schemas. This also leads to extra `null`s being inserted into events in the same batch, as is evident by the integration test reference files.

No changelog entry is needed, as this feature hasn't been released yet, and the existing entry doesn't talk about the exact schema.


libtenzir/builtins/formats/syslog.cpp
```diff
@@ -10,41 +10,156 @@
...
+  auto finish() -> std::vector<table_slice> {
+    return {builder_.finish()};
+  }
+
+private:
+  table_slice_builder builder_;
+};
+
+template <typename T>
+struct builder_tag {};
```
jachris, line 99: We already have this with `tag<T>` and `tag_v<T>`.

libtenzir/builtins/formats/syslog.cpp
```diff
@@ -10,41 +10,156 @@
...
 namespace tenzir::plugins::syslog {
 
 namespace {
 
+struct syslog_builder {
+public:
+  using message_type = format::syslog::message;
+
+  syslog_builder(const type& ty) : builder_(ty) {
+  }
```
jachris, line 26: Unless I'm missing something: Are the types passed to all three constructors not effectively constant? If so, I would just remove this parameter and use the types directly in these classes. This would also simplify the initialization code below.

libtenzir/builtins/formats/syslog.cpp
```diff
@@ -10,41 +10,156 @@
...
+                        msg.app_name, msg.process_id, msg.content);
+  }
+
+  auto rows() -> size_t {
+    return builder_.rows();
+  }
+
+  auto finish() -> std::vector<table_slice> {
+    return {builder_.finish()};
+  }
```
jachris, line 69: This and the one below needs to check whether there are any rows at all, as we'll otherwise yield empty table slices in the logic below.

libtenzir/builtins/formats/syslog.cpp
```diff
@@ -61,13 +176,11 @@ auto impl(generator<std::optional<std::string_view>> lines,
     format::syslog::message msg{};
     format::syslog::legacy_message legacy_msg{};
     if (auto parser = format::syslog::message_parser{}; parser(f, l, msg)) {
-      if (auto slice = change_builder_schema(syslog_type); slice.rows() > 0) {
+      for (auto slices = change_builder(builder_tag<syslog_builder>{});
+           auto&& slice : slices) {
         co_yield std::move(slice);
       }
```
jachris, line 182:
```suggestion
      for (auto&& slice : change_builder(builder_tag<syslog_builder>{})) {
        co_yield std::move(slice);
      }
```

libtenzir/builtins/formats/syslog.cpp
```diff
@@ -77,14 +190,11 @@ auto impl(generator<std::optional<std::string_view>> lines,
...
     } else if (auto legacy_parser = format::syslog::legacy_message_parser{};
                legacy_parser(f, l, legacy_msg)) {
-      if (auto slice = change_builder_schema(legacy_syslog_type);
-          slice.rows() > 0) {
+      for (auto slices = change_builder(builder_tag<legacy_syslog_builder>{});
+           auto&& slice : slices) {
```
jachris, line 194: Same here.

libtenzir/builtins/formats/syslog.cpp
```diff
@@ -93,10 +203,11 @@ auto impl(generator<std::optional<std::string_view>> lines,
         co_return;
       }
     } else {
-      if (auto slice = change_builder_schema(unknown_type); slice.rows() > 0) {
+      for (auto slices = change_builder(builder_tag<unknown_syslog_builder>{});
+           auto&& slice : slices) {
```
jachris, line 207: And here.

libtenzir/builtins/formats/syslog.cpp
```diff
@@ -106,8 +217,8 @@ auto impl(generator<std::optional<std::string_view>> lines,
       }
     }
   }
-  if (builder && builder->rows() > 0) {
-    co_yield builder->finish();
+  for (auto slices = finish(); auto&& slice : slices) {
+    co_yield std::move(slice);
```
jachris, line 221: And here.

web/docs/formats/syslog.md
```diff
@@ -40,21 +40,12 @@ With this input, the parser will produce the following output, with the schema n
...
-        },
-        {
-          "key": "eventID",
-          "value": "1011"
-        }
-      ]
+  "structured_data": {
+    "exampleSDID@32473": {
+      "eventSource": "Application",
+      "eventID": "1011"
```
jachris, line 46: Do we want to do type inference in the syslog parser?

jachris (approved):

PR #3918, opened 2024-02-07
Enhance node sizing docs

mavam:
This PR adds more details on the node sizing docs.

### DIFF
--- a/web/docs/setup-guides/size-a-node.md
+++ b/web/docs/setup-guides/size-a-node.md
@@ -23,8 +23,8 @@ profile.
 
 A typical [shaping](../user-guides/shape-data/README.md) operation is applying
 filters on the input values, e.g., by specifying a regular expression to only
-keep matching events. Filtering is a **CPU**-intensive operation. Unless you use
-numerous complex regular expressions, filtering is a cheap.
+keep matching events. Filtering only incurs **CPU** load. Unless you use
+numerous, complex regular expressions, filtering is cheap.
   
 #### Aggregation
 
--- a/web/docs/setup-guides/size-a-node.md
+++ b/web/docs/setup-guides/size-a-node.md
@@ -4,8 +4,69 @@ sidebar_position: 0
 
 # Size a node
 
-To better understand what resources you need to run a node, we provide the
-following calculator to derive **CPU**, **RAM**, and **storage** requirements.
+To better understand what resources you need to run a node, we provide guidance
+on sizing and a [calculator](#calculator) to derive concrete **CPU**, **RAM**,
+and **storage** requirements.
+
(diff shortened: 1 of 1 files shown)


web/docs/setup-guides/size-a-node.md
```diff
@@ -4,8 +4,69 @@ sidebar_position: 0
...
+
+Depending on what you do with pipelines, you may generate a different resource
+profile.
+
+#### Filtering
+
+A typical [shaping](../user-guides/shape-data/README.md) operation is applying
+filters on the input values, e.g., by specifying a regular expression to only
+keep matching events. Filtering is a **CPU**-intensive operation. Unless you use
+numerous complex regular expressions, filtering is a cheap.
```
tobim, line 27: The last two sentences seem to be unfinished.

web/docs/setup-guides/size-a-node.md
```diff
@@ -4,8 +4,69 @@ sidebar_position: 0
...
+either as events per second or bytes per day, is helpful for sizing your node
+proportionally.
+
+### Retention
+
+When you leverage the node's built-in storage engine by
+[importing](../user-guides/import-into-a-node/README.md) and
+[exporting](../user-guides/export-from-a-node/README.md) data, you need
+persistent **storage**. To assess your retention span, you need to understand
+your data volume and your capacity.
```
tobim, line 67: We should also mention that memory is needed for optimal query and automated lookup performance.

tobim (approved):
Useful explanations! GTM after the comments are addressed.

web/docs/setup-guides/size-a-node.md
mavam: Would "historical queries" be a separate workload then?

mavam: Or maybe "import/export"

tobim: Something like
```
### Lookups and Interactive Queries

You can improve the performance of historical query workloads with in-memory indexes.
In case you need to perform historical investigations regularly or want to use automatic lookups for IoCs you may want to provide additional memory to make use of these indexes.

mavam: So far all the headings in this document are a *one-to-one* mapping to the headings in the calculator. I'm hesitant to add a new heading that is not represented in the calculator.

mavam: Makes sense, will add that.

mavam:
@tobim can you take a look again at the new https://tenzir-node-sizing.streamlit.app/ and whether you think that works wrt. the text?

PR #3919, opened 2024-02-08
Test binary package installation in every PR

tobim:
Contrary to the existing installer test, this one also verifies that the `tenzir-node` service comes up as desired.

### DIFF
--- a/.github/workflows/nix.yaml
+++ b/.github/workflows/nix.yaml
@@ -148,8 +148,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
       - name: Download Tenzir artifact
         id: download
         uses: actions/download-artifact@v4
--- a/.github/workflows/nix.yaml
+++ b/.github/workflows/nix.yaml
@@ -12,8 +12,8 @@ permissions:
   packages: write
 
 jobs:
-  tenzir:
-    name: Tenzir
+  build:
+    name: Build
     runs-on: ${{ fromJSON(inputs.config).editions[0].gha-runner-labels }}
     steps:
       - name: Prevent pushing protected editions to public repos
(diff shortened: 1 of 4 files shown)


docker/systemd-test/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM eniocarboni/docker-ubuntu-systemd:22.04
+
+RUN apt-get update && \
+    apt-get -y --no-install-recommends install \
+    curl && \
+    rm /usr/sbin/policy-rc.d
```
eliaskosunen, line 6: Nit: add `&& rm -rf /var/lib/apt/lists/*` for smaller container size

.github/workflows/nix.yaml
```diff
@@ -131,3 +131,44 @@ jobs:
...
+    needs:
+      - build
+    if: ${{ contains(fromJSON(inputs.config).editions[0].gha-runner-labels, 'ubuntu-latest') }}
+    name: Install
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
eliaskosunen, line 152: Is `fetch-depth` of 0, i.e. the entire history, needed here?

eliaskosunen (approved):

docker/systemd-test/Dockerfile
tobim: The installer calls `apt-get install` internally. That would break if we remove this directory.

.github/workflows/nix.yaml
tobim, line 152:
```suggestion
```

tobim: No, I thought this was needed to fetch only 1 commit...

PR #3920, opened 2024-02-08
Implement more malleable lookup data for contexts

Dakostu:
This change implements more malleable lookup result data for the specific context implementations, as dictated by https://github.com/tenzir/tenzir-plugins/pull/178.


libtenzir/builtins/contexts/bloom_filter.cpp
```diff
@@ -50,47 +50,19 @@ class bloom_filter_context final : public virtual context {
   }
 
   /// Emits context information for every event in `slice` in order.
-  auto apply(table_slice slice, context::parameter_map parameters) const
+  auto apply(typed_array array) const
```
dominiklohmann, line 53: If you rebase this onto current main you'll notice a compiler error here: `typed_array` is now called `series`. The call to the free function `values(...)` below can now also just be replaced with a call to the new member function `values()` of `series`.

eliaskosunen (approved):

PR #3922, opened 2024-02-08
Introduce a most barebones `series` type

dominiklohmann:
This is far from being what we'd like to have eventually, but it's a decent enough starting point to start iterating on.

### DIFF
--- a/libtenzir/include/tenzir/series.hpp
+++ b/libtenzir/include/tenzir/series.hpp
@@ -8,7 +8,8 @@
 
 #pragma once
 
-#include "tenzir/table_slice.hpp"
+#include "tenzir/arrow_table_slice.hpp"
+#include "tenzir/offset.hpp"
 #include "tenzir/type.hpp"
 
 #include <arrow/array.h>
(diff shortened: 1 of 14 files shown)


libtenzir/include/tenzir/series.hpp
```diff
@@ -19,11 +20,15 @@
 namespace tenzir {
 
 /// A temporary series representation (until we have a proper one).
```
jachris, line 22:
```suggestion
```

tenzir/CMakeLists.txt
```diff
@@ -125,6 +125,7 @@ add_custom_target(
     ${CMAKE_COMMAND} -E env
     PATH="$<TARGET_FILE_DIR:tenzir::tenzir>:\$\$PATH:${CMAKE_CURRENT_SOURCE_DIR}/integration/lib/bats/bin"
     bats "-r" "-T" "${CMAKE_CURRENT_SOURCE_DIR}/integration/tests"
+  OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/integration-core.dep"
```
jachris, line 128: On purpose?

libtenzir/include/tenzir/series.hpp
```diff
@@ -19,11 +20,15 @@
...
-struct typed_array {
-  typed_array() = default;
+struct series {
```
jachris, line 23: Maybe this should be 

```cpp
template<class Type = type>
struct series { ... }
```

But we can also keep this as-is for now. We will surely see how we want to do it when we build the evaluator.

jachris (approved):

tenzir/CMakeLists.txt
dominiklohmann: Whoops, definitely not!

libtenzir/include/tenzir/series.hpp
dominiklohmann: I tried that. The problem is that we need to assign a type ID to id so that we can send it with CAF, and that requires being able to forward-declare it. It's a mess from a code-organization perspective that I'd like to avoid for now.

libtenzir/include/tenzir/series.hpp
dominiklohmann: Instead of removing, I chose to replace this with a meaningful description instead.

PR #3930, opened 2024-02-12
Skip `import_stream` and go directly to importer actor

jachris:


libtenzir/src/importer.cpp
```diff
@@ -206,6 +213,16 @@ importer(importer_actor::stateful_pointer<importer_state> self,
     self->state.index = std::move(index);
     self->state.stage->add_outbound_path(self->state.index);
   }
+  for (const auto& plugin : plugins::get<analyzer_plugin>()) {
+    // We can safely assert that the analyzer was already initialized. The
+    // pipeline API guarantees that remote operators run after the node was
+    // successfully initialized, which implies that analyzers have been
+    // initialized as well.
+    // TODO: Is this actually true?
```
dominiklohmann, line 221: I think so. The node itself does not allow any incoming messages before initializing components.

libtenzir/src/metrics_collector.cpp
```diff
@@ -70,15 +63,12 @@ auto metrics_collector(
...
-        return stream.error();
-      }
-      self->state.importer
-        = std::make_unique<import_stream>(std::move(*stream));
+      auto scoped_self = caf::scoped_actor{self->system()};
+      auto components
+        = get_node_components<importer_actor>(scoped_self, self->state.node);
+      TENZIR_ASSERT_CHEAP(components);
+      auto [importer] = std::move(*components);
+      self->state.importer = std::move(importer);
```
dominiklohmann, line 71: We can get the importer directly from the node state when we spawn the metrics collector actor in `make_component` here: https://github.com/tenzir/tenzir/blob/4ebca1698b1b3ed4374b85ca639aecde38c0c5cf/libtenzir/builtins/components/metrics_collector.cpp#L44-L47

dominiklohmann (approved):
Just two minor things.

PR #3938, opened 2024-02-13
Fix non-records list path with use_simple_format

dit7ya:
We want to include `-1` in the path only when we are dealing with lists of records, and not for lists of other types.

The code was actually authored by @dominiklohmann.


dominiklohmann (approved):
Slowly but surely turning you into a C++ developer 🧑‍💻

PR #3939, opened 2024-02-13
Check for contiguity and uniqueness of values in the `chart` operator

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1546


libtenzir/builtins/operators/chart.cpp
```diff
@@ -133,87 +178,198 @@ class chart_operator final : public crtp_operator<chart_operator> {
...
+    return result;
+  }
+
+  auto
+  verify_single_field(const configuration_item& item, const table_slice& slice,
+                      const arrow::StructArray& struct_array,
+                      const record_type& record_schema,
+                      std::string_view field_name,
+                      previous_values_type& previous_values,
+                      operator_control_plane& ctrl) const -> bool {
```
dominiklohmann, line 243: For checking the requirements in this function, should we check the requirements per schema id, or over all the data globally, i.e., do we care about ordering with every schema or ordering in the entire pipeline?

eliaskosunen: I don't know. Can the frontend handle wildly different schemas when charting? If it can, per-schema checking makes sense.

dominiklohmann: Right now, all results rendering in the frontend is per-schema. We eventually want to add a view that's across schemas, but that doesn't exist yet.

I think it's fine if the check is across schemas. It's slightly more strict than it needs to be right now, but there's no harm in having it, I think.

dominiklohmann (approved):

PR #3941, opened 2024-02-15
Test debian package install in the build job

tobim:
Before this change the enterprise edition install job actually used the community edition build. This sometimes led to a failure in case the latter was not uploaded yet.

We fix both problems by merging both into a joint "Build & Install" job.

### DIFF
--- a/.github/workflows/nix.nu
+++ b/.github/workflows/nix.nu
@@ -48,18 +48,18 @@ def upload_packages [
       do $run "tarball" $tgzs
     }
   }
+  print $"::notice copying artifacts to /packages/{debian,tarball}"
+  mkdir ./packages/debian ./packages/tarball
+  for deb in $debs {
+    cp -v $deb ./packages/debian
+  }
+  for pkg in $pkgs {
+    cp -v $pkg ./packages/macOS
+  }
+  for tgz in $tgzs {
+    cp -v $tgz ./packages/tarball
+  }
   if $copy {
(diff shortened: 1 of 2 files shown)


.github/workflows/nix.nu
```diff
@@ -48,18 +48,18 @@ def upload_packages [
       do $run "tarball" $tgzs
     }
   }
+  print $"::notice copying artifacts to /packages/{debian,tarball}"
+  mkdir ./packages/debian ./packages/tarball
```
lava, line 52: Do we have to include `./packages/macOS` here?

.github/workflows/nix.yaml
```diff
@@ -118,49 +118,41 @@ jobs:
...
       - name: Upload Debian Package to Github
-        if: steps.cfg.outputs.upload-to-github == 'true'
+        if: >-
+          runner.os == 'Linux' &&
+          steps.cfg.outputs.upload-to-github == 'true'
         uses: actions/upload-artifact@v4
         with:
           name: debian
           path: "./packages/debian/*.deb"
           if-no-files-found: ignore
```
lava, line 137: Shouldn't we error here now if now files are found, since we rely on the presence of `./packages/debian/*.deb` later on in the install step anyways?

lava (approved):
Makes sense to me.

.github/workflows/nix.yaml
tobim: Valid point.

PR #3942, opened 2024-02-15
fix: change the primary color in docs for better contrast

asmitbm:
As I was exploring [Tenzir docs](https://docs.tenzir.com/get-started) from a design point of view,  It appears that the links in light mode are not very accessible [contrast](https://webaim.org/resources/contrastchecker/) wise. The minimum contrast ratio for large text is 4.5:1 and for normal text, it is 7:1.

### Light Mode

Screenshot 2024-02-15 133205

### Dark Mode

Screenshot 2024-02-15 133554

## Fix

For light mode and dark mode, I changed the primary color and improved the color contrast ratio. Now it passes AA and AAA color contrast checks

### Light Mode

Screenshot 2024-02-15 144236

### Dark Mode

Screenshot 2024-02-15 144316

## How will this change affect the docs?

In both the modes, the primary color will be replaced in areas where it is used, such as links or the side navigation bar.

This adjustment is only temporary and is crucial for those currently reading the documentation. Following a discussion with Dominik, I discovered that there isn't a consistent design system in place yet.

### DIFF
--- a/web/src/css/custom.scss
+++ b/web/src/css/custom.scss
@@ -9,7 +9,7 @@
 
 /* You can override the default Infima variables here. */
 :root {
-  --ifm-color-primary: #0377ad;
+  --ifm-color-primary: #005E8A;
   --ifm-color-primary-dark: #0094d9;
   --ifm-color-primary-darker: #008bcd;
   --ifm-color-primary-darkest: #0073a9;
@@ -37,7 +37,7 @@
 
 /* For readability concerns, you should choose a lighter palette in dark mode. */
 [data-theme='dark'] {
-  --ifm-color-primary: #00a4f1;
+  --ifm-color-primary: #0EB0FB;
   --ifm-color-primary-dark: #0094d9;
   --ifm-color-primary-darker: #008bcd;
   --ifm-color-primary-darkest: #0073a9;
--- a/web/src/css/custom.scss
+++ b/web/src/css/custom.scss
@@ -9,7 +9,7 @@
 
 /* You can override the default Infima variables here. */
 :root {
-  --ifm-color-primary: #00a4f1;
+  --ifm-color-primary: #0377ad;
   --ifm-color-primary-dark: #0094d9;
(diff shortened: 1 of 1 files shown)


dominiklohmann:
I think this introduces a new problem: While the contrast to the white background is now good enough the links no longer stand out enough compared to the black text:

Before:

image

After:

image

asmitbm:
> I think this introduces a new problem: While the contrast to the white background is now good enough the links no longer stand out enough compared to the black text:

I agree with that, and to further enhance it, we can consider:

1. Defaulting to underlined links, with a thicker underline in the hover state.
2. Exploring a new primary and secondary color system.

mavam:
Thank you for the initiative! I can see that the new primary color indeed fixes the low contrast. However, we actually have a design system in place. When we make such impactful changes like these, we should always consider them across the entire brand, e.g., https://tenzir.com/, https://app.tenzir.com/, and the docs here. 

We're also big fans of the [ClickHouse Docs](https://clickhouse.com/docs). They really tweaked the hell out of their Docusaurus setup. Unfortunately we currently don't have the bandwidth to do something similar.

asmitbm:
> Thank you for the initiative! I can see that the new primary color indeed fixes the low contrast. However, we actually have a design system in place. When we make such impactful changes like these, we should always consider them across the entire brand, e.g., https://tenzir.com/, https://app.tenzir.com/, and the docs here.

Welcome! I wasn't aware of the existing design system, but I totally agree that consistency is key across all the products. I'd be happy to make adjustments based on that, or if the design system isn't public yet, this issue can be raised internally. 

> We're also big fans of the [ClickHouse Docs](https://clickhouse.com/docs). They really tweaked the hell out of their Docusaurus setup. Unfortunately we currently don't have the bandwidth to do something similar.

Their documentation looks amazing! I've got some experience working with custom Docusaurus components. If there's an open issue down the road, count me in – I'd love to contribute!

dominiklohmann:
Closing as this is stale.

PR #3943, opened 2024-02-15
Add `--progress` option to `save` and `to`

dominiklohmann:
WIP, don't look at this yet.


libtenzir/src/execution_node.cpp
```diff
@@ -348,8 +348,8 @@ struct exec_node_state {
       }
       if (not std::holds_alternative<generator<Output>>(*output_generator)) {
         return caf::make_error(
-          ec::logic_error, fmt::format("{} expected {}, but got {}", *self,
-                                       operator_type_name<Output>(),
+          ec::logic_error, fmt::format("{} {} expected {}, but got {}", *self,
+                                       op->name(), operator_type_name<Output>(),
```
eliaskosunen, line 352: This error message would need to be reformatted. "exec-node internal-save-without-progress failed to start" is not very user-friendly.

libtenzir/builtins/operators/to_write_save.cpp
```diff
@@ -333,13 +422,14 @@ class write_and_save_operator final
   }
 
   auto name() const -> std::string override {
-    return "internal-write-save";
+    return fmt::format("internal-write-save-{}-progress",
+                       ShowProgress == show_progress::yes ? "with" : "without");
```
eliaskosunen, line 426: Do we actually want/need the `internal-` prefix here? It's going to be visible to the user via error messages and `--dump-ast`, so it's not really internal.

libtenzir/builtins/operators/to_write_save.cpp
```diff
@@ -361,6 +455,10 @@ class write_and_save_operator final
 private:
   std::unique_ptr<plugin_printer> printer_;
   std::unique_ptr<plugin_saver> saver_;
+
+  // This will anger @jachris if he sees it, but @eliaskosunen said it was okay.
+  // -- @dominiklohmann, half jokingly, during a Hackathon.
```
mavam, line 460: Muha

libtenzir/src/curl.cpp
```diff
@@ -126,8 +126,10 @@ auto easy::set_http_header(std::string_view name, std::string_view value)
       break;
     }
   }
-  auto header = fmt::format("{}: {}", name, value);
-  headers_.append(header);
+  if (not value.empty()) {
+    auto header = fmt::format("{}: {}", name, value);
+    headers_.append(header);
+  }
```
mavam, line 132: Did this bite you?

The reason why empty values were supported was to make it possible to *unset* header values. It's a cURL idiom to use an empty value for that. This ability is now gone.

libtenzir/builtins/connectors/curl.cpp
```diff
@@ -26,6 +26,7 @@ namespace tenzir::plugins {
 namespace {
 
 struct http_options {
+  bool octet_stream;
```
mavam, line 29: Why does this need its own flag? To set the Content Type? (If so, you typically simply set the header for this.)

mavam (commented):
Mind adding also the missing documentation to `--verbose` to the HTTP connector? 🙏

libtenzir/builtins/connectors/curl.cpp
dominiklohmann: This is simply an artifact from debugging because we couldn't get the upload to S3 to work from the pre-signed URL. I will split everything except for new option for save and to into a separate PR.

libtenzir/builtins/operators/to_write_save.cpp
dominiklohmann: I don't consider `--dump-ast` to be user-facing, so for that I wouldn't wanna do it.

Do you have a suggestion for an alternative name that makes more sense? It can't be `write-save`, and it also can't be `to`, as those names are taken already.

libtenzir/builtins/operators/to_write_save.cpp
dominiklohmann: Resolving this quickly so he doesn't see.

libtenzir/src/curl.cpp
dominiklohmann: Yes; the fix for this doesn't work either, though, so the reason why we're unable to override the content type likely lies elsewhere. We temporarily worked around this with `--octet-stream`, but that should go away again.

libtenzir/builtins/operators/to_write_save.cpp
jachris: 🫨

... 

Adding this to my to-be-deleted list 😉

libtenzir/builtins/operators/to_write_save.cpp
eliaskosunen: Just `write-save-{}-progress`, without the `internal-` prefix?

libtenzir/src/execution_node.cpp
eliaskosunen: This error is not only shown on internal logic errors, but also when the user writes an invalid pipeline, as is evident by the integration test reference files:

```
> from file:///foo.json read json

error: the file `/foo.json` does not exist
 --> <input>:1:13
  |
1 | from file:///foo.json read json
  |             ^^^^^^^^^ 
  |
error: pipeline-executor aborts start because an execution node failed to start
 = note: exec-node internal-save-without-progress failed to start
 = note: exec-node load failed to instantiate operator
 = note: could not instantiate loader
 = note: pipeline failed to start
```

jachris: This is a fallout from the recent error propagation changes. We should make it so that the second error is not shown at all.

dominiklohmann: This problem only exists for loaders and savers right now as they have a separate error path still. I'm not sure what to do about that.

web/docs/operators/save.md
```diff
@@ -23,6 +23,11 @@ The `save` operator operates on raw bytes.
 Notably, it cannot be used after an operator that emits events, but rather only
 with operators that emit bytes, e.g., [`write`](write.md) or [`load`](load.md).
 
+### `--progress`
+
+Emit events describing the progress of the save operation. Note that this turns
+the `save` operator into a transformation rather than a sink.
+
```
jachris, line 30: As suggested during the demo: I would recommend to remove this from the docs and from diagnostics.

dominiklohmann:
Closing this for now; once we continue working on the download button outside of the scope of a Hackathon we can pick this up again.

PR #3944, opened 2024-02-15
Introduce a CI job to automatically update submodule main branch

lava:
On a push to main, this will check if the commit that the tenzir-plugins submodule points to is the current master commit of the tenzir/tenzir-plugins repository. If not, it will automatically merge the corresponding plugins PR, update the nix settings, and push a new commit to main.

### DIFF
--- a/.github/workflows/tenzir.yaml
+++ b/.github/workflows/tenzir.yaml
@@ -394,13 +394,13 @@ jobs:
         uses: webfactory/ssh-agent@v0.9.0
         with:
           ssh-private-key: ${{ secrets.TENZIR_PLUGINS_DEPLOY_KEY }}
-      - name: Require contrib/tenzir-plugins to be on a main branch commit
+      - name: Require contrib/tenzir-plugins to be downstream of current `main`
         run: |
           git submodule update --init contrib/tenzir-plugins
           git -C contrib/tenzir-plugins fetch origin main
           git -C contrib/tenzir-plugins merge-base --is-ancestor \
-            $(git -C contrib/tenzir-plugins rev-parse HEAD) \
-            $(git -C contrib/tenzir-plugins rev-parse origin/main)
+            $(git -C contrib/tenzir-plugins rev-parse origin/main) \
+            $(git -C contrib/tenzir-plugins rev-parse HEAD)
       - name: Install Nix
         uses: cachix/install-nix-action@v25
         with:
--- a/.github/workflows/tenzir.yaml
+++ b/.github/workflows/tenzir.yaml
@@ -83,10 +83,60 @@ jobs:
       run-python: ${{ steps.configure.outputs.run-python }}
       run-python-package: ${{ steps.configure.outputs.run-python-package }}
     steps:
+      - name: Configure ssh-agent
+        uses: webfactory/ssh-agent@v0.8.0
+        with:
+          ssh-private-key: ${{ secrets.TENZIR_PLUGINS_DEPLOY_KEY }}
       - name: Checkout
         uses: actions/checkout@v4
         with:
+          submodules: true
           fetch-depth: 0
+          persist-credentials: false
+      - name: Generate a token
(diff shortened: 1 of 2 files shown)


.github/workflows/tenzir.yaml
```diff
@@ -83,10 +83,60 @@ jobs:
...
+            gh pr --repo tenzir/tenzir-plugins merge $pr_number --merge --match-head-commit ${plugins_sha}
+
+            # Create a new commit for the updated submodule.
+            git checkout main # TODO: Is this necessary or are we already on main?
+            git -C contrib/tenzir-plugins checkout main
+            new_plugins_main_sha=$(git -C contrib/tenzir-plugins rev-parse --short origin/main)
+            nix/update-plugins.sh
+            git add contrib/tenzir-plugins
+            git add nix/
+            git commit -m "Automatic submodule bump from $plugins_sha_short to $new_plugins_main_sha"
```
mavam, line 137:
```suggestion
            git commit -m "Bump submodule from $plugins_sha_short to $new_plugins_main_sha"
```

Just to keep our git commit conventions in check, i.e., use verbs to denote the primary action.

.github/workflows/tenzir.yaml
```diff
@@ -83,10 +83,60 @@ jobs:
...
+          if [ "$plugins_sha" != "$plugins_main_sha" ]; then
+            # Auto-merge the plugins PR and update 'tenzir-plugins/main'.
+            # gh pr --repo tenzir/tenzir-plugins list
+            pr_number=$(gh pr --repo tenzir/tenzir-plugins list --search $plugins_sha --json number -q .[0].number)
+            echo $pr_number
+            [ $pr_number ] # Assert the pr number was non-empty
+            gh pr --repo tenzir/tenzir-plugins merge $pr_number --merge --match-head-commit ${plugins_sha}
+
+            # Create a new commit for the updated submodule.
+            git checkout main # TODO: Is this necessary or are we already on main?
```
mavam, line 131: Does this need to be sorted out before merging?

.github/workflows/tenzir.yaml
```diff
@@ -83,10 +83,60 @@ jobs:
...
+          persist-credentials: false
+      - name: Generate a token
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        id: generate_token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ vars.TENZIR_TESTAPP_APP_ID }}
+          private-key: ${{ secrets.TENZIR_TESTAPP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+      - name: Test submodule update
```
mavam, line 104: The name seems to be very innocuous compared what it actually does. Maybe make it a bit clearer?

.github/workflows/tenzir.yaml
lava: We can only really test if after it runs on main. But in the worst case it is going to be a no-op, so it shouldn't matter too much either way.

.github/workflows/tenzir.yaml
```diff
@@ -83,10 +83,60 @@ jobs:
...
+          persist-credentials: false
+      - name: Generate a token
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        id: generate_token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ vars.TENZIR_TESTAPP_APP_ID }}
+          private-key: ${{ secrets.TENZIR_TESTAPP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+      - name: Bump plugins submodule
```
tobim, line 104: Why is all of this implemented as extra steps of the `configure` job?
May I politely ask to move this logic into its own dedicated workflow?

tobim (changes requested):

dominiklohmann (approved):

PR #3945, opened 2024-02-15
Context versioning

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1454, plugins-PR: https://github.com/tenzir/tenzir-plugins/pull/196

This PR implements versioned loading of serialized contexts.
- The `context` base class is refactored, to support multiple `context_loader`s, each associated with a version
  - Each context is refactored accordingly, with a single loader corresponding to "v1"
- When persisted on disk, the version number is stored in the filename: `<name>.<type>.<version>.ctx` (previously `<name>.<type>.ctx`). No version number in the filename is assumed to mean "v1".
- Saving is not materially changed:
  - Only a single saver per context is still supported
  - `context::save` also returns a version number
  - When persisted, the version number is marked in the filename
- When starting a node, the context directory is iterated through, and the newest supported version of each context is picked for loading. If the load succeeds, all persisted versions of that context with an older version number are removed, to prevent bloating.
  - Because saving only uses the newest supported version, this essentially means that on the happy path, contexts are auto-migrated to newer versions on node startup, with no older stale versions being left lingering around
- `context load` gains an optional `--version` flag, as it doesn't have the filename to infer version information from. The default is to use the newest supported version.

### DIFF
--- a/changelog/next/changes/3945--context-versioning.md
+++ b/changelog/next/changes/3945--context-versioning.md
@@ -1,0 +1,3 @@
+The binary format used by contexts for saving on disk on node shutdown
+is now versioned. A node can support loading of multiple different versions,
+and automigrate between them.
--- a/libtenzir/builtins/contexts/bloom_filter.cpp
+++ b/libtenzir/builtins/contexts/bloom_filter.cpp
@@ -169,20 +169,37 @@ public:
     return show();
   }
 
-  auto save() const -> caf::expected<chunk_ptr> override {
+  auto save() const -> caf::expected<save_result> override {
     std::vector<std::byte> buffer;
     if (auto err = convert(bloom_filter_, buffer)) {
       return add_context(err, "failed to serialize Bloom filter context");
     }
-    return chunk::make(std::move(buffer));
+    return save_result{.data = chunk::make(std::move(buffer)), .version = 1};
   }
 
 private:
   dcso_bloom_filter bloom_filter_;
 };
 
+struct v1_loader : public context_loader {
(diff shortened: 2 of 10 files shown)


web/docs/operators/context.md
```diff
@@ -59,6 +59,10 @@ See the [list of available context types](../contexts.md).
 
 Context-specific options in the format `--key value` or `--flag`.
 
+### `--version <version>`
+
+Context-specific version number for the binary data to load.
+
```
dominiklohmann, line 65: I know that this was straightforward to implement, but from a user-perspective, what's the use case for it? Does the user ever need to think about versioning?

dominiklohmann (approved):
I tried this out locally by creating a v2 loader for the lookup table, and that transitioned automatically on startup for me as expected. I have one question about the necessity of `--version`, other than that I think it's good to go.

web/docs/operators/context.md
eliaskosunen: You might be right, the real use cases for this feature are very slim, and it probably isn't worth the increased API surface. It's not every day one has to `context load` something that was `context save`d with a different context binary version.

dominiklohmann: I'd be fine with leaving it in as well, but we should probably have a note that explains what it's for in the docs then.

eliaskosunen: Eh, I'll rip it out. It's easy to add if there's customer or developer demand. The version isn't user-queryable anyway, so it's difficult for a user to find out what exact value to put there.

PR #3947, opened 2024-02-15
Fix restart on failure

dominiklohmann:
Really helps to call a function called `on_pipeline_failure` when the pipeline fails every time the state changes to failed...

We added a new way that pipelines could fail with v4.8, and that did not trigger restarts on failure correctly.


Dakostu (approved):
I approve, especially since I contributed to the fix.

PR #3950, opened 2024-02-16
Improve submodule bump workflow

lava:
* Require an open PR for any changes to the plugins submodule
* Remove some leftover debug code
* Rename the used secrets away from `TESTAPP`
* Fix a bug where the submodule would be bumped to an incorrect state.

### DIFF
--- a/.github/workflows/tenzir.yaml
+++ b/.github/workflows/tenzir.yaml
@@ -349,8 +349,24 @@ jobs:
           git submodule update --init contrib/tenzir-plugins
           git -C contrib/tenzir-plugins fetch origin main
           git -C contrib/tenzir-plugins merge-base --is-ancestor \
-            $(git -C contrib/tenzir-plugins rev-parse origin/main) \
-            $(git -C contrib/tenzir-plugins rev-parse HEAD)
+            $(git -C contrib/tenzir-plugins rev-parse HEAD) \
+            $(git -C contrib/tenzir-plugins rev-parse origin/main)
+      - name: Generate a token
+        id: generate_token
+        uses: actions/create-github-app-token@v1
+        with:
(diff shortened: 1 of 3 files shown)


dominiklohmann (approved):
The changes here all make sense to me.

.github/workflows/tenzir.yaml
```diff
@@ -349,8 +349,24 @@ jobs:
...
+          app-id: ${{ vars.TENZIR_AUTOBUMPER_APP_ID }}
+          private-key: ${{ secrets.TENZIR_AUTOBUMPER_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+      - name: Require an open PR for the submodule bump
+        run: |
+          plugins_sha=$(git -C contrib/tenzir-plugins rev-parse HEAD)
+          plugins_main_sha=$(git -C contrib/tenzir-plugins rev-parse origin/main)
+
+          if [ "$plugins_sha" != "$plugins_main_sha" ]; then
+            pr_number=$(gh pr --repo tenzir/tenzir-plugins list --search $plugins_sha --json number -q .[0].number)
```
dominiklohmann, line 367: Can we also check here if all comments on the PR are resolved?

lava: As far as I can tell there's no easy way to do this using `gh` (or the github API for that matter).

I'm not even sure how to check that all tests have passed, there is a `mergeable` field but that only checks if there would be a merge conflict in the PR.

I think at least initially we'll have to rely on reviewers and authors to not merge until comments from both PRs are fully resolved.

dominiklohmann: Fine by me.

Dakostu (approved):

PR #3954, opened 2024-02-19
Use thread-safe `strerror` everywhere

jachris:
`::strerror` is not thread-safe, and we already have `describe_errno()`, which is. It also fixes a bug within `describe_errno()` which was caused by `::strerror_r` returning a `char*` instead of `int` on some systems.


dominiklohmann (approved):

libtenzir/src/detail/posix.cpp
```diff
@@ -21,24 +20,32 @@
...
+        return fmt::format("<errno = {}>", err);
+      }
+      buffer.erase(std::ranges::find(buffer, '\0'), buffer.end());
+      return buffer;
+    } else {
+      static_assert(std::same_as<decltype(result), char*>);
+      TENZIR_ASSERT(result);
+      return std::string(result);
+    }
+  }(result);
```
jachris, line 48: @dominiklohmann Please take a look at this once more and resolve if it's good to go. 

I had to change this again because the `syserr_list` approach is not portable, and `strerror_r` returns a `char*` on Linux, which passes type checking with `!= 0`, but has totally wrong semantics... Seems like getting textual descriptions of errors is still a hard problem in 2024.

dominiklohmann: Oh wow, the promotion from the 0 literal to nullptr striking again.

I did not test this locally, but the code makes sense to me. It's easy enough to follow. Feel free to resolve.

PR #3964, opened 2024-02-20
Slightly relax rules for policy enforcement

lava:
As observed by @jachris , if a PR does not change the plugins, but some other PR does in the meantime, that the branch protections (of the PR that does not change plugins) do not pass. This means that all PRs have to be updated, even though they don't necessarily conflict.

This PR aims to solve this by allowing submodule commits that are either direct ancestors or direct descendants of `origin/main`.

### DIFF
--- a/.github/workflows/bump-plugins-submodule.yaml
+++ b/.github/workflows/bump-plugins-submodule.yaml
@@ -48,8 +48,8 @@ jobs:
           set -x
 
           # Setup git
-          git config --global user.name 'CI Bot'
-          git config --global user.email 'tenzir-ci-app@users.noreply.github.com'
+          git config --global user.name 'tenzir-bot'
+          git config --global user.email 'engineering@tenzir.com'
           git remote set-url origin https://x-access-token:$GITHUB_APP_TOKEN@github.com/tenzir/tenzir.git
           git -C contrib/tenzir-plugins fetch -q origin main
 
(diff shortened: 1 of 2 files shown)


.github/workflows/tenzir.yaml
```diff
@@ -348,9 +348,14 @@ jobs:
         run: |
           git submodule update --init contrib/tenzir-plugins
           git -C contrib/tenzir-plugins fetch origin main
+          # Check that the plugins submodule commit is either already merged
+          # into `main` or branches off directly from current `main`.
```
tobim, line 352: The wording "branches off directly from current main" suggests that the commit that the submodule is pointing to does not include any fork points with other HEADs after the merge-base with main.
I suggest to change the phrase:
```suggestion
          # Check that the plugins submodule commit is either already merged
          # into `main` or descends from current `main`.
```

tobim (dismissed):

.github/workflows/bump-plugins-submodule.yaml
```diff
@@ -34,6 +34,11 @@ jobs:
           app-id: ${{ vars.TENZIR_AUTOBUMPER_APP_ID }}
           private-key: ${{ secrets.TENZIR_AUTOBUMPER_APP_PRIVATE_KEY }}
           owner: ${{ github.repository_owner }}
+      - uses: crazy-max/ghaction-import-gpg@v6
+        with:
+          gpg_private_key: ${{ secrets.TENZIR_BOT_GPG_SIGNING_KEY }}
```
dominiklohmann, line 39: How does this work if it's the app that creates the commit and this is the gpg private key of the `tenzir-bot` account?

lava: Right, we also need to set the user/email so that it fits with the gpg key.

(the app should be orthogonal since its only used for github permissions, but the signing is for the commit object in git itself)

tobim (approved):

PR #3965, opened 2024-02-21
Type-independent numeric lookup for `lookup-table`

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1087, plugins PR: https://github.com/tenzir/tenzir-plugins/pull/203


dominiklohmann (approved):
I played around with this a fair bit today. I think it's a good enough solution. It's a bit unfortunate that it doesn't solve the problem for `data` generally, but it wouldn't be reasonable to spend the time now to do that.

PR #3968, opened 2024-02-22
Implement multi-field lookups for context implementations

Dakostu:
This change adds a multi-field lookup functionality to the context and context implementations.


libtenzir/include/tenzir/plugin.hpp
```diff
@@ -732,7 +732,8 @@ class context {
   virtual auto context_type() const -> std::string = 0;
 
   /// Emits context information for every event in `array` in order.
-  virtual auto apply(series array) const -> caf::expected<std::vector<series>>
+  virtual auto apply(std::vector<series> array) const
+    -> caf::expected<std::vector<std::vector<series>>>
```
dominiklohmann, line 736: Looking at the diff in this PR, all the implementations just loop over the list of series to do the same thing as before. Did this function need to change?

Dakostu: Yes, if we want to prevent multiple runs of set-up code for each `series` before applying the values inside the context implementation (although right now, the implementations don't set up much).

Would removing this dimension and calling `apply` for each `series` be good for now?

dominiklohmann: I would say yes, as for the current contexts there is no need for this optimization at all.

Dakostu: Removed the dimension in https://github.com/tenzir/tenzir/pull/3968/commits/b5ab0f369b4c36c472c34c4b5f4465343f7c365a

dominiklohmann (approved):

Dakostu:
@dominiklohmann LGTM.

PR #3972, opened 2024-02-22
Add a helper to copy GHA cache entries

tobim:
Github does not offer an API to download entries from the actions cache for inspection. This workflow offers a way around that limitation.


lava (approved):
Looks good to me.

As usual with Github Actions workflows we can't test it in advance, but since it's just a workflow_dispatch it can't break anything anyways.

PR #3977, opened 2024-02-23
Clear GHA caches of closed branches

tobim:
Taken from:
https://github.com/actions/cache/blob/main/tips-and-workarounds.md#force-deletion-of-caches-overriding-default-cache-eviction-policy


lava (approved):
Should be safe since it's a straight copy from the official github docs.

PR #3979, opened 2024-02-23
Run the installer script more flexibly as root

mavam:
This PR improves `sudo` usage in the installer.


Dakostu (approved):
Tested this with a custom Debian VM.

Installing with `sudo` only passed the root check ✅
Installing with  `doas` only passed the root check ✅
Trying to install without `sudo` or `doas` failed accordingly ✅

I'll approve this.

PR #3982, opened 2024-02-24
Improve metrics collection

dominiklohmann:
This rewrites the _metrics collector_ component from the ground up to be much simpler. Metrics plugins can now specify the frequency at which they want to emit metrics, defaulting to one second, and the metrics are collected every thirty seconds. Previously, metrics were collected every ten seconds and then immediately sent out in batches of exactly one event.

Additionally, this implements memory metrics on macOS.


mavam (approved):
Paired on this with Dominik.

libtenzir/include/tenzir/detail/weak_run_delayed.hpp
```diff
@@ -35,15 +35,20 @@ auto weak_run_delayed(caf::scheduled_actor* self, caf::timespan delay,
...
 /// @param function The action to run with the signature () -> void.
+/// @param run_immediately Whether to run the function immediately.
 template <class Function>
   requires std::is_invocable_r_v<void, std::remove_reference_t<Function>&>
 void weak_run_delayed_loop(caf::scheduled_actor* self, caf::timespan delay,
-                           Function&& function) {
+                           Function&& function, bool run_immediately = true) {
+  if (run_immediately) {
+    std::invoke(function);
+  }
```
jachris, line 45: I checked all uses of this function whether they can handle being run immediately, as this might be dangerous during the setup. But it _seems_ like they do.

Not sure whether this is a good default. In most cases it's unnecessary. For collecting health metrics it makes sense of course. 

(But I don't care too much, please just resolve if you want to keep this as-is.)

jachris (approved):
Tested this on macOS. Looks good!

libtenzir/include/tenzir/detail/weak_run_delayed.hpp
dominiklohmann: I also checked all the cases when I made this change, and I found that almost all of them actually want to be run immediately. I considered creating a second function instead, but then we wouldn't use the old one anymore.

PR #3987, opened 2024-02-27
Allow listen mode UDS in from file

tobim:
This adds the `--uds` flag to the file input connector. If specified, a listening socket is created at at the path if no socket exists yet.


mavam:
How do you switch between listening and connecting? Is hard-coding the former the primary use case?

tobim:
> How do you switch between listening and connecting? Is hard-coding the former the primary use case?

The idea is to connect if the socket already exists, or listen otherwise.

tobim:
This is not working as expected in local testing yet.
Putting this on hold until it gets a higher prio.

PR #3989, opened 2024-02-27
Introduce multi-arch Docker images

dominiklohmann:
This makes our Docker images multi-arch images, running natively on linux/amd64 and on linux/arm64/v8.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -12,9 +12,19 @@ permissions:
   packages: write
 
 jobs:
-  tenzir:
-    name: Tenzir
-    runs-on: ubuntu-20.04
+  build:
+    name: Build (${{ matrix.build.tag }})
+    runs-on: ${{ matrix.build.runner }}
+    strategy:
+      fail-fast: true
+      matrix:
+        build:
+          - runner: ubuntu-20.04
+            arch: linux/amd64
+            tag: amd64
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yaml
```diff
@@ -107,14 +87,19 @@ jobs:
...
-              docker tag "tenzir/${name}:latest" "${registry}/tenzir/${name}:${tag}"
-              docker push "${registry}/tenzir/${name}:${tag}"
-            done
+          cmake_version_args="-DTENZIR_VERSION_BUILD_METADATA:STRING=${{ fromJSON(inputs.config).version-build-metadata }}"
+          echo "::info Building the Developer Edition Image"
+          jq -r '.editions[] | .name as $name | .target as $target | .registries[]? | [ $name, $target, . ] | join(" ")' \
+            <<< '${{ inputs.config }}' | while read -r name target registry; do
+            docker buildx build . \
+              -t "${registry}/tenzir/${name}:latest" --target "${target}" \
+              --platform linux/amd64,linux/arm64 --push \
```
dominiklohmann, line 101: Note to future self: This exceeds the 6 hour timeout that GitHub Actions workflows have. We need to instead build two separate images on their respective native platforms, and then in merge the two with `docker manifest create` manually.

tobim (approved):
Verified that the new workflow works

PR #3990, opened 2024-02-29
Count nulls and absent values in `top` and `rare`

dominiklohmann:
This makes the handling of null and absent values in `top` and `rare` more consistent with the remaining language.

Previously, the `top` and `rare` operators added a single event to their output with a count of zero when the counted field was not present or held a null value in any of the input fields. Now, they correctly count nulls and absent values.


jachris (approved):
As previously discussed.

PR #3991, opened 2024-02-29
Add TQLv2 parser

jachris:
This PR adds a new `--tql2` option (for developer use only). If present, a completely new and separate code path is taken, which processes the input according to the new and upcoming version of the Tenzir Query Language. This PR contains the new lexer, parser and AST definitions for most of the new language. The implementation stops before name resolution. In particular, pipelines cannot be executed yet. However, the resulting AST can be dumped with `--dump-ast`. 

As this is a solely developer feature for now, there is no changelog entry.


dominiklohmann (approved):
We agreed in a call to do a post-hoc review of this. This should land sooner rather than later to avoid merge conflicts, and the new code is not user-reachable yet.

PR #3993, opened 2024-03-01
Allow using `and`, `or`, and `not` in expressions

dominiklohmann:
This change makes it so that `X or not (Y and Z)` is equivalent to `X || !(Y && Z)`. Since we want to move to that anyways with TQL2 and I'm constantly mistyping this already, I figured it'd make sense to introduce this four line diff that's entirely backwards-compatible to allow using the new tokens for conjunctions, disjunctions, and negations already.

This also removes the undocumented and untested `[+`, `[-`, `+]`, and `-]` relational operators that were aliases for `in`, `not in`, `ni`, and `not ni`, respectively. It's odd when you find code after over four years of contributing to a project that's just silently been there and that you've never seen. I doubt anyone is going to miss them.


libtenzir/src/concept/parseable/tenzir/expression.cpp
```diff
@@ -219,10 +220,10 @@ auto make_predicate_parser() {
...
     | "ni"_p  ->* [] { return relational_operator::ni; }
     | "!ni"_p ->* [] { return relational_operator::not_ni; }
-    | "[+"_p  ->* [] { return relational_operator::in; }
-    | "[-"_p  ->* [] { return relational_operator::not_in; }
-    | "+]"_p  ->* [] { return relational_operator::ni; }
-    | "-]"_p  ->* [] { return relational_operator::not_ni; }
+    | "not"_p >> required_ws_or_comment >> "in"_p 
+      ->* [] { return relational_operator::not_in; }
+    | "not"_p >> required_ws_or_comment >> "ni"_p 
+      ->* [] { return relational_operator::not_ni; }
```
jachris, line 226:
```suggestion
    | "not"_p >> required_ws_or_comment >> "in"_p 
      ->* [] { return relational_operator::not_in; }
```

Should be `ni ton` instead of `not ni`... Only slightly joking. But since it won't be valid in the new language, there's no need to add this, right?

libtenzir/src/concept/parseable/tenzir/expression.cpp
jachris, line 1: By the way, this parser parses `!foo == bar` as `!(foo == bar)`. But we can't do anything about that, as we cannot represent `!foo`, just letting you know as a fun curiosity.

jachris (approved):

libtenzir/src/concept/parseable/tenzir/expression.cpp
dominiklohmann: It really should be contains rather than ni, but yeah, I do agree.

libtenzir/src/concept/parseable/tenzir/expression.cpp
```diff
@@ -279,13 +280,13 @@ static auto make_expression_parser() {
     ;
   group
     = '(' >> optional_ws_or_comment >> ref(expr) >> optional_ws_or_comment >> ')'
-    | '!' >> optional_ws_or_comment >> pred_expr ->* negate_expr
-    | '!' >> optional_ws_or_comment >> '(' >> optional_ws_or_comment >> (ref(expr) ->* negate_expr) >> optional_ws_or_comment >> ')'
+    | ('!'_p | "not") >> optional_ws_or_comment >> pred_expr ->* negate_expr
+    | ('!'_p | "not") >> optional_ws_or_comment >> '(' >> optional_ws_or_comment >> (ref(expr) ->* negate_expr) >> optional_ws_or_comment >> ')'
```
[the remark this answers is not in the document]
dominiklohmann: I used identifier instead of alnum, which I think is slightly more correct.

PR #3998, opened 2024-03-05
Prevent tcp socket inheritance to child processes

tobim:
Child processes created via `shell` or `python` inherit all currently open file descriptors from the node, which prevent the TCP connector form getting the same port upon restart, because it was now blocked by the children, because of the inherited open file descriptors.

To prevent this, we now create our own socket with `FN_CLOEXEC` that we assign to the ASIO abstraction where possible, and use `fcntl` where not.


dominiklohmann (approved):
We went over this together in a call. This is good to go modulo a changelog entry.

PR #4001, opened 2024-03-07
Fix partial specialization of S3 configuration in URL

dominiklohmann:
This fixes partial overrides of the S3 configuration in the arguments to the URL. Previously, when providing any arguments to the S3 connector as part of the URL, it failed to respect the default credentials from the configuration file.

Additionally, this makes it possible to specify the S3 secrets in the S3 plugin's configuration file directly.

### DIFF
--- a/libtenzir/builtins/connectors/s3.cpp
+++ b/libtenzir/builtins/connectors/s3.cpp
@@ -21,28 +21,49 @@ namespace tenzir::plugins::s3 {
 
 namespace {
 
+struct s3_config {
+  std::string access_key = {};
+  std::string secret_key = {};
+  std::string session_token = {};
+
+  friend auto inspect(auto& f, s3_config& x) -> bool {
+    return f.object(x)
+      .pretty_name("s3_config")
+      .fields(f.field("access-key", x.access_key),
+              f.field("secret-key", x.secret_key),
+              f.field("session-token", x.session_token));
+  }
+};
+
 struct s3_args {
-  bool anonymous;
-  located<std::string> uri;
+  bool anonymous = {};
+  located<std::string> uri = {};
+  std::optional<s3_config> config = {};
 
   template <class Inspector>
   friend auto inspect(Inspector& f, s3_args& x) -> bool {
     return f.object(x).pretty_name("s3_args").fields(
-      f.field("anonymous", x.anonymous), f.field("uri", x.uri));
+      f.field("anonymous", x.anonymous), f.field("uri", x.uri),
+      f.field("config", x.config));
   }
 };
 
 auto get_options(const s3_args& args) -> arrow::fs::S3Options {
-  auto opts = arrow::fs::S3Options::Anonymous();
-  if (not args.anonymous) {
-    auto opts_from_uri = arrow::fs::S3Options::FromUri(args.uri.inner);
-    if (not opts_from_uri.ok()) {
-      opts = arrow::fs::S3Options::Defaults();
-    } else {
-      opts = *opts_from_uri;
-    }
+  auto opts = arrow::fs::S3Options::FromUri(args.uri.inner);
+  if (not opts.ok()) {
+    diagnostic::error("failed to parse S3 options: {}",
+                      opts.status().ToString())
+      .throw_();
   }
-  return opts;
+  if (args.anonymous) {
+    opts->ConfigureAnonymousCredentials();
+  } else if (args.config) {
+    opts->ConfigureAccessKey(args.config->access_key, args.config->secret_key,
+                             args.config->session_token);
+  } else {
+    opts->ConfigureDefaultCredentials();
+  }
+  return opts.MoveValueUnsafe();
 }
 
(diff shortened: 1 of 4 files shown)


libtenzir/builtins/connectors/s3.cpp
```diff
@@ -21,28 +21,49 @@ namespace tenzir::plugins::s3 {
...
+  if (args.anonymous) {
+    opts->ConfigureAnonymousCredentials();
+  } else if (args.config) {
+    opts->ConfigureAccessKey(args.config->access_key, args.config->secret_key,
+                             args.config->session_token);
+  } else {
+    opts->ConfigureDefaultCredentials();
   }
-  return opts;
+  return opts.MoveValueUnsafe();
```
dominiklohmann, line 66: This is the bug fix. The idea is as follows:
- Always parse the options from the URI
- If explicitly asked to, use the anonymous credentials
- Otherwise, if the new config file exists, use the credentials from it
- Otherwise, use the default credentials if they exist

web/docs/connectors/s3.md
```diff
@@ -42,6 +42,16 @@ that user, meaning that the AWS credentials must also be configured for that
 user. The directory `~/.aws` must be readable for the `tenzir` user.
 :::
 
+If a config file `<prefix>/etc/tenzir/plugin/s3.yaml` or
+`~/.config/tenzir/plugin/s3.yaml` exists, it is always preferred over the
+default AWS credentials. The configuration file must have the following format:
+
+```yaml
+secret-key: your-secret-key
+access-key: your-access-key
```
Dakostu, line 51: The AWS documentation puts the `access-key` before the `secret-key` - maybe we should follow that order, too.
```suggestion
access-key: your-access-key
secret-key: your-secret-key
```

libtenzir/builtins/connectors/s3.cpp
Dakostu: OK, but `arrow::fs::S3Options::FromUri()` already calls `S3Options::ConfigureDefaultCredentials()`, no?
See here: https://github.com/apache/arrow/blob/f589e08a2e86e0562e68ed3aaa8e0c8c1e93c5d1/cpp/src/arrow/filesystem/s3fs.cc#L369

This change would actually remove support for the `s3://access_key:secret_key@bucket-name/path/to/file` syntax that Arrow specifies [here](https://arrow.apache.org/docs/r/articles/fs.html#s3-authentication)  - if you specify the `access_key:secret_key` pair in the URI, and no other credentials elsewhere in the steps you mentioned, the final `ConfigureDefaultCredentials()` call here will overwrite the credentials from the URI key pair with empty values.

So that `ConfigureDefaultCredentials()` call most likely needs to go, if we want to keep that support.

Dakostu (changes requested):
Alright, so I tested around with config files, env variables, and `credential` files. The general mechanism works, but there are some edge cases that we need to reconsider.

libtenzir/builtins/connectors/s3.cpp
Dakostu: Also I noticed that the `ConfigureDefaultCredentials()` call takes a noticeable time on my machine. I don't know why, let's call it Arrow black magic, but removing that call makes credential set up faster in Debug builds at least (~10 seconds with the call vs. 1 second without)

dominiklohmann: That's a nice catch, thank you!

libtenzir/builtins/connectors/s3.cpp
```diff
@@ -221,17 +243,45 @@ class s3_saver final : public plugin_saver {
...
                                          "functionality: {}",
                                          initialized.ToString()));
     }
-    return caf::none;
+    if (plugin_config.empty()) {
+      {};
+    }
+    config_.emplace();
+    for (const auto& [key, value] : plugin_config) {
+#define X(opt, var)                                                            \
```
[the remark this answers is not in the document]
dominiklohmann: That'd be a lot less readable here than the macro, given that I then couldn't return directly. I think this is fine as-is.

libtenzir/builtins/connectors/s3.cpp
```diff
@@ -21,28 +21,49 @@ namespace tenzir::plugins::s3 {
...
-      opts = arrow::fs::S3Options::Defaults();
-    } else {
-      opts = *opts_from_uri;
-    }
+  auto opts = arrow::fs::S3Options::FromUri(args.uri.inner);
+  if (not opts.ok()) {
+    diagnostic::error("failed to parse S3 options: {}",
+                      opts.status().ToString())
+      .throw_();
+  }
```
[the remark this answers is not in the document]
dominiklohmann: I changed this around so that it now shows the proper error message to the user. I also wrote a follow-up story to render thrown diagnostics in exec nodes properly.

Dakostu (approved):
OK, looks good to me now.

PR #4005, opened 2024-03-07
Support ZLIB compressed data in the kafka connector

tobim:
This fixes a build issue with the static binary that caused our `librdkafka` dependency to be built without ZLIB support.


dominiklohmann (approved):
We went over this together in a call.

PR #4006, opened 2024-03-08
Add configurable pipelines

tobim:
This addition allows defining pipelines in `tenzir.yaml`. Configured pipelines will be started automatically with the node.

Here is an example for how to use this functionality from `tenzir.yaml`:

```
pipelines:
  my-fancy-pipeline:
    definition: show version | write json -c | save stdout
    autostart:
      failed: false
  json-import:
    name: JSON Import
    definition: from tcp://0.0.0.0:34343 read json | import
    autostart:
      failed: true
```

### DIFF
--- a/changelog/next/features/4006--configurable-pipelines.md
+++ b/changelog/next/features/4006--configurable-pipelines.md
@@ -1,0 +1,2 @@
+We made it possible to set pipelines declaratively in the `tenzir.yaml`
+configuration file.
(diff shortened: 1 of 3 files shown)


contrib/tenzir-plugins
dominiklohmann, line 1: This feature definitely deserves a changelog entry.

Dakostu:
I checked the PR in `tenzir-plugins` and I'll approve for now, only things missing are integration tests.

Dakostu (approved):

PR #4010, opened 2024-03-11
Fix crash in `sigma` operator for non-existent file

dominiklohmann:
This fixes a potential crash in the `sigma` operator when reading a non-existent file.


mavam (approved):
I tested the fix locally. It works.

PR #4012, opened 2024-03-11
Fix panic on parsing invalid syslog messages

eliaskosunen:
Reported on Discord.

This PR fixes a logic bug in the parser, that previously caused a panic, when encountering an invalid syslog message. This is caused by a multiline message, where the latter lines are not valid messages themselves. We don't currently provide proper support for multiline syslog messages, but now we no longer crash.

Lines that don't look like a syslog message on their own are parsed as `syslog.unknown`, with the entire line contained in a field called `syslog_message`, as can be seen in the test reference files.

### DIFF
--- a/changelog/next/bug-fixes/4012--multiline-syslog.md
+++ b/changelog/next/bug-fixes/4012--multiline-syslog.md
@@ -1,0 +1,2 @@
+Parsing an invalid syslog message (using the schema `syslog.unknown`)
+no longer causes a crash.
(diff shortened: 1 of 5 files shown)


changelog/next/bug-fixes/4012--multiline-syslog.md
```diff
@@ -0,0 +1 @@
+Parsing a multi-line syslog message no longer causes a crash.
```
jachris, line 1: This crash is not caused by multi-line messages, but whenever `unknown_syslog_builder` is used (including single-line messages).

jachris (approved):

PR #4026, opened 2024-03-12
Handle presence of simdjson number type `big_integer`

jachris:
New releases of `simdjson` include `simdjson::ondemand::number_type::big_integer`. This PR makes it so that we don't have a non-exhaustive `switch` (if it exists), while still being compatible with older `simdjson` versions. 

This can be removed once we require a newer `simdjson` version, but it was easy enough to support.

### DIFF
--- a/libtenzir/builtins/formats/json.cpp
+++ b/libtenzir/builtins/formats/json.cpp
@@ -377,8 +377,35 @@ private:
         }
         return add_value(builder, result.value_unsafe());
       }
+      default: {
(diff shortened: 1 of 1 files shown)


libtenzir/builtins/formats/json.cpp
dominiklohmann, line 1: Please bump the bundled simdjson to v3.8.0 as well, then this can completely supersede https://github.com/tenzir/tenzir/pull/4023

PR #4028, opened 2024-03-12
Ignore events in `lookup` that did not bind

dominiklohmann:
This fixes two issues in `lookup`:
1. The operator did not ignore internal events, and
2. the operator let events through for which the provided fields did not bind.


Dakostu (approved):
Approved, with 1 question in the subrepo.

PR #4031, opened 2024-03-12
Improve HTTP transfer abstraction

mavam:
This PR fixes a few internals when setting up HTTP transfers and using them in a saver context.


mavam:
@lava this now works for me:

```
tenzir 'version | write lines | save http -v PUT https://tenzir-tzcp-dev-responses-... Content-Type:'
```

I've correctly implemented two disjoint paths for processing PUT and POST bodies. The former using `CURLOPT_UPLOAD` and the latter `CURLOPT_POST`. This also solves a related refactoring in the the platform plugin that I'll get back to after merging this one.

PR #4033, opened 2024-03-13
Fix loading of large contexts

eliaskosunen:
This PR fixes a logic bug in `context load`, where each chunk of the input was attempted to be loaded individually, rather than loading the entire input. This error was only encountered, when loading contexts that were large enough to be split into multiple chunks.

Plugins PR https://github.com/tenzir/tenzir-plugins/pull/210


jachris (approved):
Thanks for the quick fix :pray:

contrib/tenzir-plugins
jachris, line 1: (to be resolved once plugin PR conversations are resolved)

PR #4036, opened 2024-03-13
Isolate environment in python operator

jachris:
This PR fixes an issue with the current usage of `exec` and opens up the possibility to de-uglify the python operator implementation.


lava (approved):
Looks good to me, the only possible concern is that previously code could implictly use the existing imports as in

```
python "self.x = os.environ['foo']"
```

But since we currently only import `os` and `sys` from the standard library, the chance seems pretty low that this will break any existing user pipelines.

satta:
From a user perspective I would have expected to `import os` anyway in this case.

satta:
> the only possible concern is that previously code could implictly use the existing imports as in
> 
> ```
> python "self.x = os.environ['foo']"
> ```

From a user perspective I would have expected to need to `import os` anyway in this case.

PR #4040, opened 2024-03-14
Introduce `--replace`, `--separate`, and `--yield` for contexts

dominiklohmann:
This adds `--yield` and `--separate` options to the `lookup` and `enrich` operators, and the `--replace` option for `enrich`.

The `--replace` option makes the `enrich` operator replace the input value with the context as opposed to adding it to the output, and makes the output a lot leaner.

The `--separate` option causes the operators duplicate the input events for each matched field alongside making the nesting structure a lot less deep for the context.

The `--yield <field>` option causes the operators to add only the specified subset of the context.

Additionally, this changes the context structure slightly to add the meta information per context as opposed to once per event. This makes it easier to work with the results of the two operators in downstream tooling.

This also fixes a bug in the `geoip` context, which sometimes returned bogus values, and changes the `bloom-filter` context to just return a boolean instead of a record containing a single field containing a boolean.

### DIFF
--- a/changelog/next/changes/4040--context-metadata.md
+++ b/changelog/next/changes/4040--context-metadata.md
@@ -1,0 +1,3 @@
+The `enrich` and `lookup` operators now include the metadata in every context
+object to accomodate the new `--replace` and `--separate` options. Previously,
+the metadata was available once in the output field.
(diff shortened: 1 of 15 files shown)


changelog/next/changes/4040--context-metadata.md
jachris, line 1: Changelog reminder for `mode`. Also, might want to consider to mark these changes as breaking.

jachris (approved):
LGTM :+1:

PR #4042, opened 2024-03-15
Add line comments to legacy parser

jachris:


dominiklohmann (approved):
🫣

jachris:
Actually, let's not do this right now. This PR makes multiple tests fail where the empty pattern literal is used `//` and also the rules around `//` have to be slightly changed because of arguments that contain `//` somewhere. Both things can be solved, but it's not worth our time right now.

PR #4043, opened 2024-03-15
Fix `from <url>` with username

jachris:


dominiklohmann (approved):
Awesome, thanks for the really quick fix!

PR #4044, opened 2024-03-15
Fix a regression in `from <uri>`

dominiklohmann:
This fixes an oversight from #4043 that led to a CI failure; we hastily merged it earlier to get the fix in for S3 with explicit credentials, but overlooked that the URI parser from libcurl rejects schemes it does not support. The fix is to simply switch to the Arrow URI parser instead, which is a lot more forgiving.


jachris (approved):
Thank you for quickly fixing the fix :pray:

PR #4050, opened 2024-03-19
Introduce the `every` operator modifier

dominiklohmann:
`every <interval> <operator> [<args...>]` runs the source operator `<operator>` every `<interval>`.

Fixes tenzir/issues#1670


Dakostu (approved):
@dominiklohmann and I went over this in a call. Looks good to me now.

PR #4051, opened 2024-03-19
Support IP-in-subnet queries in lookup tables

mavam:
This PR enhances lookup tables when their key type is a `subnet` value. In this case, it is now possible to probe the lookup table with a value of type `ip` and perform a *prefix match* on all contained subnets.


dominiklohmann:
@mavam Out of curiosity: Did you see https://github.com/zeek/pysubnettree before and consider using the patricia trie from it?

mavam:
Yep, this is exactly the implementation I'm using now.

libtenzir/builtins/contexts/lookup_table.cpp
```diff
@@ -276,11 +296,25 @@ class ctx final : public virtual context {
 
   /// Inspects the context.
   auto show() const -> record override {
-    return record{{"num_entries", context_entries.size()}};
+    // TODO: there's not size() function for the patricia trie, so we need to
+    // track the number of elements externally and integrate it into this
+    // output.
```
mavam, line 301: We could track insertions/deletions and then compute the number of elements in the subnet tree.

libtenzir/builtins/contexts/lookup_table.cpp
```diff
@@ -379,9 +419,11 @@ class ctx final : public virtual context {
 
   auto reset() -> caf::expected<void> override {
     context_entries.clear();
+    subnet_entries.clear();
     return {};
   }
 
+  // TODO: persist patricia trie.
```
mavam, line 426: @tenzir/pipelines-squad does this requires bumping the Flatbuffer version?

libtenzir/builtins/contexts/lookup_table.cpp
```diff
@@ -415,8 +457,10 @@ class ctx final : public virtual context {
 
 private:
   map_type context_entries;
+  detail::subnet_tree subnet_entries;
 };
 
+// TODO: do we need a v2 loader now with the subnet optimization?
```
mavam, line 463: @tenzir/pipelines-squad what about this? 👆

libtenzir/builtins/contexts/lookup_table.cpp
dominiklohmann: Why would you need to persist it at all? All the data is already present in the lookup table, so when loading you can just recreate the trie.

libtenzir/builtins/contexts/lookup_table.cpp
dominiklohmann: Same as above—if you avoid persisting the patricia trie in the first place this is irrelevant.

libtenzir/builtins/contexts/lookup_table.cpp
```diff
@@ -324,8 +358,14 @@ class ctx final : public virtual context {
     while (key_it != key_values.end()) {
       TENZIR_ASSERT(context_it != context_values.end());
       auto materialized_key = materialize(*key_it);
-      context_entries.insert_or_assign(materialized_key,
-                                       materialize(*context_it));
+      // Subnets never make it into the regular map of entries.
```
dominiklohmann, line 361: I think this makes everything a lot more complicated than it needs to be. You can remove a lot of the code in this file if you just add it to both.

mavam: But that 2x'es the space utilization.

dominiklohmann: I think this is irrelevant until someone complains about it. Let's wait with improving this until someone wants to add a few million subnets.

mavam: Okay. YOLO but fine with me for the first version.

dominiklohmann: Even then, if you really want to optimize it, you could still keep it separate in memory but persist it together as one lookup table. No need to change the on-disk format either way.

PR #4057, opened 2024-03-21
Add the `set` operator for upserting fields

dominiklohmann:
This is the sweet spot between `extend` and `replace`.

### DIFF
--- a/web/docs/operators/set.md
+++ b/web/docs/operators/set.md
@@ -1,0 +1,41 @@
+---
+sidebar_custom_props:
+  operator:
+    transformation: true
+---
+
+# set
+
+Upserts fields in events.
+
+## Synopsis
+
+```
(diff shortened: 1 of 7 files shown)


web/docs/operators/set.md
```diff
@@ -0,0 +1,40 @@
...
+## Synopsis
+
+```
+set <field=operand>...
+```
+
+## Description
+
+The `set` operator appends a specified list of fields to the input. If any of
+the fields already exist in the input event, they are replaced again.
```
tobim, line 20:
```suggestion
The `set` operator sets a list of fields to the given values. It overwrites old
values of fields matching the `field` expression, or creates new fields of a
given name otherwise.
```

tobim (approved):

PR #4067, opened 2024-03-24
Add UDP connector

mavam:
This PR adds a UDP connector.

```[tasklist]
- [x] Add loader
- [x] Support hostname resolution
- [x] Support IPv6
- [x] Write docs
- [x] Fix off-by-one hang
- [x] Add saver
- [x] Add integration tests
- [x] Add changelog entry
```


libtenzir/builtins/connectors/udp.cpp
```diff
@@ -0,0 +1,260 @@
...
+  } else {
+    TENZIR_DEBUG("binding to {}", args.url);
+    if (sock.bind(*endpoint) < 0) {
+      diagnostic::error("failed to bind to socket")
+        .note("error: {}", detail::describe_errno())
+        .emit(ctrl.diagnostics());
+      co_return;
+    }
+  }
+  while (true) {
```
dominiklohmann, line 183: You want to `co_yield {}` here once to signal that the operator started successfully.

libtenzir/builtins/connectors/udp.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+  auto parse_saver(parser_interface& p) const
+    -> std::unique_ptr<plugin_saver> override {
+    auto parser = argument_parser{
+      name(),
+      fmt::format("https://docs.tenzir.com/docs/connectors/{}", name())};
+    auto endpoint = located<std::string>{};
+    auto args = saver_args{};
+    parser.add(endpoint, "<endpoint>");
+    parser.add("-n,--insert-newlines", args.insert_newlines);
```
dominiklohmann: This doesn't make sense.

libtenzir/include/tenzir/as_bytes.hpp
```diff
@@ -13,43 +13,54 @@
...
+template <size_t Size>
+auto as_bytes(const void* data) noexcept {
+  return std::span<const std::byte, Size>{
+    reinterpret_cast<const std::byte*>(data), Size};
+}
+
+template <size_t Size>
+auto as_writeable_bytes(void* data) noexcept {
+  return std::span<std::byte, Size>{reinterpret_cast<std::byte*>(data), Size};
+}
```
dominiklohmann, line 28: This is a really unnecessary optimization that bloats the binary, just use the other two. They'll be inlined by the compiler anyways if you pass in a compile-time constant for the second argument.

web/docs/connectors/udp.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+Use a shell to test the UDP loader with netcat:
+
+```bash
+# Shell 1
+tenzir 'from udp://127.0.0.1:56789'
+# Shell 2
+jq -n '{foo: 42}' | nc -u 127.0.0.1 56789
+```
```
dominiklohmann, line 70: This looks a lot nicer if you split it into two code blocks that each have a `title="Shell X"` set.

The second example should also use `socat` rather than `netcat`, because netcat is not portable.

dominiklohmann (approved):
This works well in practice. I left a few comments inline that should be addressed, but they're of minor nature only.

web/docs/connectors/udp.md
mavam: > This looks a lot nicer if you split it into two code blocks that each have a `title="Shell X"` set.

Will adjust.

> The second example should also use `socat` rather than `netcat`, because netcat is not portable.

There's a bug in `socat` 1.8 that actually prevents this from working on macOS, so I opted for netcat instead.

libtenzir/include/tenzir/as_bytes.hpp
mavam: I tried undoing this but I realized that's actually not redundant. I `span<T, N>` is very much different from a `span<T, dynamic_extent>`. The other overloads do not supporting a fixed-size span, which I need here.

PR #4080, opened 2024-03-27
Add multi-line syslog message support

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1068

This PR adds support for multi-line syslog messages. This is achieved through a simple assumption, where if a line of input couldn't be parsed as a syslog message, it's assumed to be a continuation of a message in a preceding line (if any). This is technically a change in behavior, which is reflected in the changelog.

### DIFF
--- a/libtenzir/builtins/formats/syslog.cpp
+++ b/libtenzir/builtins/formats/syslog.cpp
@@ -14,102 +14,228 @@
 #include <tenzir/table_slice_builder.hpp>
 #include <tenzir/to_lines.hpp>
 
+#include <ranges>
+
 namespace tenzir::plugins::syslog {
 
 namespace {
 
+template <typename Message>
+struct syslog_row {
+  syslog_row(Message msg, size_t line_no)
+    : parsed(std::move(msg)), starting_line_no(line_no) {
+  }
+
+  void emit_diag(std::string_view parser_name, diagnostic_handler& diag) const {
+    if (line_count == 1) {
+      diagnostic::error("syslog parser ({}) failed", parser_name)
+        .note("input line number {}", starting_line_no)
+        .emit(diag);
+      return;
+    }
+    diagnostic::error("syslog parser ({}) failed", parser_name)
+      .note("input lines number {} to {}", starting_line_no,
+            starting_line_no + line_count - 1)
+      .emit(diag);
+  }
+
+  Message parsed;
+  size_t starting_line_no;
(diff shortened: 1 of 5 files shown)


libtenzir/builtins/formats/syslog.cpp
```diff
@@ -18,88 +18,160 @@ namespace tenzir::plugins::syslog {
...
+        {"message", std::move(msg.msg)},
+      };
+      if (not builder.try_data(r)) {
+        row.emit_diag("RFC 5242", diag);
+        return std::nullopt;
+      }
+    }
+    rows_.clear();
+    return builder.finish_assert_one_slice();
   }
```
jachris, line 95: The call to `finish_assert_one_slice` is dangerous because you might get more than one slice if there is a type mismatch. For example, try this:
```
<165>8 2003-10-11T22:14:15.003Z foo bar - ID47 [example@32473 iut="5"]
<165>8 2003-10-11T22:14:15.003Z foo bar - ID47 [example@32473 iut="foo"]
```

libtenzir/builtins/formats/syslog.cpp
```diff
@@ -149,8 +221,10 @@ auto impl(generator<std::optional<std::string_view>> lines,
     if (rows() >= defaults::import::table_slice_size
         or last_finish + defaults::import::batch_timeout < now) {
       last_finish = now;
-      for (auto&& slice : finish()) {
-        co_yield std::move(slice);
+      if (auto slice = finish(); not slice) {
+        co_return;
+      } else if (slice->rows() > 0) {
+        co_yield std::move(*slice);
```
jachris, line 227: There is a small bug hidden here: If the last message was a multi-line one and we either reach the timeout or have collected `defaults::import::table_slice_size` messages, then we will emit it without its continuation, and will produce an extra `syslog.unknown` message instead.

jachris (approved):
I think the strategy of continuing the previous message if the next line is not a valid one will work out very nicely in practice!

libtenzir/builtins/formats/syslog.cpp
jachris: Unresolving, as it seems that this is not fixed yet? Perhaps I'm mistaken, please let me know if I'm missing something.

eliaskosunen: Should really be fixed now.

PR #4084, opened 2024-04-02
Add the `accept` operator

dominiklohmann:
The `accept <endpoint> read <format>` operator accepts multiple TCP connections on the specified endpoint, and for each creates a parser of the specified format, returning the events into the pipeline.


tobim:
When someone tries to connect to a TLS socket with the wrong handshake the pipeline terminates.
I believe this should not be the case.
```
tenzir "from tcp://127.0.0.1:4000 --tls --certfile ss.crt --keyfile ss.crt read json"
error: wrong version number (SSL routines)
 = note: TLS handshake failed
```
client:
```
echo '{"foo":"bar"}' | socat - TCP4:127.0.0.1:4000
```

libtenzir/builtins/operators/tcp-listen.cpp
```diff
@@ -0,0 +1,512 @@
...
+      return {};
+    }
+    if (not SSL_set_tlsext_host_name(tls_handle,
+                                     self->state.args.hostname.c_str())) {
+      diagnostic::error("failed to set SNI")
+        .emit(self->state.ctrl->diagnostics());
+      return {};
+    }
+    auto ec = boost::system::error_code{};
+    self->state.ssl_ctx.emplace(boost::asio::ssl::context::tls_server);
```
lava, line 155: Aren't we unconditionally overwriting the previous context from line 136 here?

dominiklohmann: Yes, but I think we may actually have the same bug in `tcp.cpp` then because that's what this is copied from.

lava: Let's take the opportunity to fix it in both places then :D

tobim (approved):

PR #4085, opened 2024-04-02
Make `python` operator not discard fields that start with an underscore

jachris:
Fields starting with `_` were previously discarded due to some implementation details. However, following #4036, this is no longer necessary. I also used this opportunity to clean up the surrounding code.


changelog/next/changes/4085--python-operator-underscore.md
dominiklohmann, line 1: I think from a user-perspective this should be consider a bug fix, especially since it was undocumented behavior.

dominiklohmann (approved):

PR #4086, opened 2024-04-02
Fix `python` deadlock for empty input

jachris:
This PR fixes a deadlock with `version | python ""`, which is caused by a blocking read on a pipe. Note that the approach with a single `read(...)` also has other issues (for example, it doesn't handle slicing). But since this did not occur yet in practice, I opted for this simple fix instead.


tenzir/integration/data/reference/python/test_python_operator_empty_input/step_00.ref
```diff
@@ -0,0 +1 @@
+{"version": "4.11.2+gf8c6f3452d-dirty", "build": "gf8c6f3452d-dirty", "major": 4, "minor": 11, "patch": 2}
```
dominiklohmann, line 1: Please use something other than `version` for testing.

dominiklohmann (approved):
Makes sense.

tenzir/integration/data/reference/python/test_python_operator_empty_input/step_00.ref
jachris: Ha, that's a _very_ good point. Thank you for catching this. 

(Kept `version`, but added `put x=42` afterwards.)

PR #4087, opened 2024-04-02
Fix JSON printer handling of `inf` and `nan`

jachris:
The JSON printer previously printed invalid JSON for `inf` and `nan`, which means that `serve` could sometimes emit invalid JSON, which is not handled well by platform/app. Instead, we now emit `null`.

```diff
- {"inf": inf.0, "nan": nan, "subnormal": 2.2250738585072e-310, "zero": -0.0}
+ {"inf": null, "nan": null, "subnormal": 2.2250738585072e-310, "zero": -0.0}
```


mavam (approved):
You may want to add a changelog entry for this bugfix.

PR #4088, opened 2024-04-02
Support zero-copy reading in new Feather parser

dominiklohmann:
This change builds on top of @balavinaithirthan's upcoming Feather format and adds zero-copy support to it. This is a vast improvement when reading Feather from a memory-mapped file, where copies of data can now be avoided entirely until we need to do IPC or exchange any columns.

### DIFF
--- a/libtenzir/include/tenzir/make_byte_reader.hpp
+++ b/libtenzir/include/tenzir/make_byte_reader.hpp
@@ -14,76 +14,94 @@
 namespace tenzir {
 
 /// Returns a stateful function that retrieves a given number of bytes in a
+/// contiguous buffer from a generator of chunks. The last chunk is underful,
+/// i.e., smaller than the number of bytes requested, and zero-sized if the
+/// input boundaries are aligned. The function returns nullptr whenever it
+/// merges buffers from multiple chunks. This does not indicate completion.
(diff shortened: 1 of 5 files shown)


libtenzir/include/tenzir/make_byte_reader.hpp
```diff
@@ -14,76 +14,94 @@
...
+        TENZIR_ASSERT(buffer->size() <= num_bytes);
+        return buffer;
+      }
+      auto result = buffer->slice(offset, num_bytes);
+      offset = buffer->size();
+      TENZIR_ASSERT(result->size() <= num_bytes);
+      return result;
+    }
+    auto chunk = std::move(*current);
+    ++current;
```
balavinaithirthan, line 56: Are we allowed to modify current after std::moved?

balavinaithirthan (approved):
Looks great!

libtenzir/include/tenzir/make_byte_reader.hpp
dominiklohmann: We don't move the iterator, but rather the value pointed to by the iterator, as indicated by the dereferencing. We can't access `*current` again before we increment the iterator, but we don't do that here.

PR #4091, opened 2024-04-04
Treat missing fields as nulls in `deduplicate`

eliaskosunen:
Fixes https://github.com/tenzir/issues/issues/1697 by explicitly adding `null`s for every missing field in the projected input. This means, that missing fields and `null`s are treated equivalently. These `null`s are not visible in the operator output, but are only used for comparing whether two inputs are duplicates of one another.

Also contains a drive-by fix for a parsing issue when `deduplicate` without arguments is the last operator in a pipeline.

Adds an undocumented internal `--project-only` flag to `deduplicate`, useful for debugging the projection logic.

Also fixes dependence on input field ordering in `deduplicate`. For example, `{"a": 1, "b": 2}` and `{"b": 2, "a": 1}` are now treated as duplicates of one another.

As a _feature_, because after this PR the projected inputs are flattened automatically for comparison purposes, `{"foo": {"bar": 123}}` and `{"foo.bar": 123}` are also treated as duplicates of one another.


libtenzir/builtins/operators/deduplicate.cpp
```diff
@@ -277,9 +403,11 @@ class plugin final : public virtual operator_plugin<deduplicate_operator> {
     std::optional<int64_t> limit{};
     std::optional<int64_t> distance{};
     std::optional<duration> timeout{};
+    bool project_only{false};
     parser.add("--limit", limit, "<count>");
     parser.add("--distance", distance, "<count>");
     parser.add("--timeout", timeout, "<duration>");
+    parser.add("--project-only", project_only);
```
dominiklohmann, line 410: This isn't really undocumented as listed in the PR description, as it shows up when trying to parse a wrong option. I.e., `deduplicate --help` will list this option. Do we have a good solution for this?

cc @jachris

dominiklohmann (commented):
Just a question.

libtenzir/builtins/operators/deduplicate.cpp
eliaskosunen: I could just rip it out, but it was really convenient when developing this PR, and I could imagine it being useful for development in the future

jachris: We could add a parameter to make it hidden.

libtenzir/builtins/operators/deduplicate.cpp
```diff
@@ -122,33 +136,143 @@ class deduplicate_operator final : public crtp_operator<deduplicate_operator> {
...
+        indices.erase(uniq.begin(), uniq.end());
+      }
+      return cached_projection{
+        .indices = std::move(indices),
+        .transformation_factory = {},
+      };
+    }
+    // Some fields were missing.
+    // Construct a `transformation_factory`, that'll add these fields in, with a
+    // value of `null`.
```
jachris, line 178: IIUC, this also modifies the events returned by the operator, right? Not sure how I feel about this. Did you consider the alternative of not modifying the incoming events?

libtenzir/builtins/operators/deduplicate.cpp
```diff
@@ -122,33 +136,143 @@ class deduplicate_operator final : public crtp_operator<deduplicate_operator> {
...
+        {num_fields - 1},
+        [rows = slice.rows(),
+         &transformation_callback](struct record_type::field input_field,
+                                   std::shared_ptr<arrow::Array> input_array) {
+          return transformation_callback(static_cast<int64_t>(rows),
+                                         std::move(input_field),
+                                         std::move(input_array));
+        },
+      };
+    };
```
jachris, line 218: I have a feeling this doesn't handle nested fields correctly. Can you confirm/debunk this?

libtenzir/builtins/operators/deduplicate.cpp
dominiklohmann: I don't feel too strongly about it and we can also leave it in.

libtenzir/builtins/operators/deduplicate.cpp
eliaskosunen: The events returned by the operator are not modified -- the outputs are yielded in `deduplicate()`, and are only ever subslices of the original input. This null-trickery is part of the logic for comparing whether two events are duplicates of one another. Essentially, heavily simplified, the logic is:
```
// inside deduplicate::operator():
std::unordered_map<data, ...> matches{};
for (auto&& slice : input) {
  auto projected_slice = project(slice);

  size_t begin_row = 0;
  for (size_t current_row = 0; current_row < slice.rows(); ++current_row) {
    // .row_at() is pseudocode
    auto match_it = matches.find(projected_slice.row_at(current_row));
    if (match_it == matches.end()) {
      std::tie(match_it, std::ignore)
        = matches.emplace(projected_slice.row_at(current_row), ...);
    }
    if (is_duplicate(match_it)) {
      co_yield subslice(slice, begin_row, current_row);
      begin_row = current_row + 1;
    }
  }
  co_yield subslice(slice, begin_row, slice.rows());
}
```
The output of `project` is used for lookup into `matches`, i.e., to find out whether the row is a duplicate. The actual output contains only subslices of the original input.

libtenzir/builtins/operators/deduplicate.cpp
eliaskosunen: Oh right it doesn't, as this is semi-directly pulled from `put`/`extend`. I added a `flatten` call before any projection logic, which should make this work.

libtenzir/builtins/operators/deduplicate.cpp
jachris: Thank you for the explanation :pray:

libtenzir/builtins/operators/deduplicate.cpp
```diff
@@ -122,33 +136,145 @@ class deduplicate_operator final : public crtp_operator<deduplicate_operator> {
...
+        missing_fields.emplace_back(field);
       }
     }
-    std::sort(indices.begin(), indices.end());
-    indices.erase(std::unique(indices.begin(), indices.end()), indices.end());
-    return indices;
+    if (missing_fields.empty()) {
+      // Every field in `cfg_.fields` was found in `slice.schema()`.
+      // Clean up the indices and return.
+      std::ranges::sort(indices);
```
jachris, line 166: It seems like we incorrectly depend on the order of the fields. Perhaps this was also the case before this PR, I didn't check that.

~~~
$ cat test.json
{"a": 1, "b": 2}
{"b": "reset"}
{"b": 2, "a": 1}

$ tenzir "from test.json | deduplicate a, b"
{
  "a": 1,
  "b": 2
}
{
  "b": "reset"
}
{
  "b": 2,
  "a": 1
}
~~~

(The middle event is needed to convince the JSON reader to use a different field ordering).

eliaskosunen: Oh darn it, we do, and also did before

eliaskosunen: I think this is now fixed. This PR is really starting to blow up in size, and in turn, so is `deduplicate` in complexity!

libtenzir/builtins/operators/deduplicate.cpp
eliaskosunen: I'll just leave it in, then.

libtenzir/builtins/operators/deduplicate.cpp
eliaskosunen: Nested fields are now handled correctly.

libtenzir/builtins/operators/deduplicate.cpp
```diff
@@ -98,72 +268,89 @@ class deduplicate_operator final : public crtp_operator<deduplicate_operator> {
...
-      for (auto idx : schema.resolve(field)) {
-        indices.emplace_back(std::move(idx));
-      }
-    }
-    std::sort(indices.begin(), indices.end());
-    indices.erase(std::unique(indices.begin(), indices.end()), indices.end());
-    return indices;
-  }
+    = std::unordered_map<sorted_flat_data, match_type, sorted_flat_data_hash,
+                         std::equal_to<>>;
```
jachris, line 315: You could consider using `robin_map` here instead.

jachris (approved):
We agreed to address the open conversations at a later point in time and move forward with this now, as it already represents a significant improvement upon the current behavior of `deduplicate`.

libtenzir/builtins/operators/deduplicate.cpp
```diff
@@ -13,24 +13,179 @@
...
 namespace tenzir::plugins::deduplicate {
 
 namespace {
 
 using std::chrono::steady_clock;
 
+/// Returns `true` if
+///  - `d` is a flat record (depth of 0 or 1), and
+///  - the keys of `d` are sorted.
+/// Requires `d` to be a record.
```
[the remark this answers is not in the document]
eliaskosunen: Yeah, you're right, it can never really be a non-record, so let's just encode that in the type system.

libtenzir/builtins/operators/deduplicate.cpp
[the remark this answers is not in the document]
eliaskosunen: Type extractors are admittedly a little messy with this, and the concept of correctness is not as clear. I'm also 100% sure if we need type extractors with this, given how the semantics are a little blurry. (Are `{foo: 1.2.3.4}` and `{foo: 1.1.1.1, bar: 1.2.3.4}` duplicates, if `{foo: 1.2.3.4}` and `{bar: 1.2.3.4}` are/aren't?)

libtenzir/builtins/operators/deduplicate.cpp
```diff
@@ -98,72 +268,89 @@ class deduplicate_operator final : public crtp_operator<deduplicate_operator> {
...
     for (size_t row = 0; row < slice.rows(); ++row) {
-      auto element_view = value_at(ty, elements, static_cast<int64_t>(row));
-      auto match_it = matches.find(element_view);
+      auto projected_value_view = value_at(projected_type, projected_elements,
+                                           static_cast<int64_t>(row));
+      auto match_it = matches.find(projected_value_view);
       if (match_it == matches.end()) {
         std::tie(match_it, std::ignore)
-          = matches.emplace(materialize(element_view), match_type{});
+          = matches.emplace(materialize(projected_value_view), match_type{});
```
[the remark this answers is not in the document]
eliaskosunen: The initial idea was to do as much as possible beforehand on a per-schema basis, and cache these mappings/projections to be applied when the data came in. In retrospect, I see this approach may have been misguided (or would've required some other implementation decisions to be made).

To try to simplify this at least a little, I removed `sorted_flat_data_view`.

libtenzir/builtins/operators/deduplicate.cpp
eliaskosunen: I couldn't, without breaking const-correctness: the iterator returned by `robin_map::find` doesn't allow modifying `->second`, which is needed in `deduplicate()`

eliaskosunen:
I pushed a final simplification commit, and I'm going to keep the commit history without squashing. Enabling auto-merge, this is all yours now.

PR #4092, opened 2024-04-04
Fix HTTP PUT with empty request body

mavam:
This PR fixes a bug when attempting to send a HTTP PUT request with an empty request body.

### DIFF
--- a/libtenzir/src/transfer.cpp
+++ b/libtenzir/src/transfer.cpp
@@ -50,29 +50,42 @@ auto transfer::prepare(http::request req) -> caf::error {
     if (auto err = to_error(easy.set(CURLOPT_POST, 1))) {
       return err;
     }
+    // We set the POST body size here, even if the request body is empty, i.e.,
+    // the size is 0. This allows us to send POST requests that are empty, which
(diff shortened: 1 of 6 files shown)


libtenzir/src/transfer.cpp
```diff
@@ -75,6 +68,26 @@ auto transfer::prepare(http::request req) -> caf::error {
...
+      // handle.
+      if (auto err = to_error(easy.set(CURLOPT_COPYPOSTFIELDS, req.body))) {
+        return err;
+      }
+    } else if (req.method == "PUT") {
+      if (auto err = set(easy, chunk::make(std::move(req.body)))) {
+        return err;
+      }
+    } else {
+      TENZIR_WARN("unexpected request body with HTTP {} method", req.method);
```
mavam, line 88: Note to the reviewer: this will fire on HTTP PATCH. But it's so incredibly rare that I didn't go the extra mile to see whether PATCH would go in to POST or PUT branch.

dominiklohmann (approved):
Approving modulo the missing feature flag.

PR #4093, opened 2024-04-04
Fix shutdown of connected pipelines alongside node

dominiklohmann:
In some cases, pipelines created with a `tenzir` process connected to a `tenzir-node` process did not shut down correctly when the `tenzir-node` process was terminated. This change makes it so that remotely spawned operators are always shut down reliably.

Fixes tenzir/issues#1712


mavam (approved):
Paired the review with Dominik.

PR #4096, opened 2024-04-05
Minor cleanups and simplifications

tobim:
These are a bunch of minor improvements that I found during other recent work. They are not directly related or necessary for any planned work, but it is still valuable to use opportunities for cleanup and simplifications to make our lives a little easier in the long run.

To the reviewer:
CI should catch possible errors, please don't spend more than 2 minutes on this. If you find an issue with one of the changes I will just drop that.

### DIFF
--- a/tenzir/integration/lib/bats-tenzir/src/fixtures.bash
+++ b/tenzir/integration/lib/bats-tenzir/src/fixtures.bash
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 export_default_paths() {
-  BATS_TEST_DATADIR="$(realpath $(dirname ${BATS_TEST_DIRNAME}))"
+  BATS_TEST_DATADIR="$(realpath "$(dirname "${BATS_TEST_DIRNAME}")")"
   # Look for the first folder called `integration/` that is a parent
   # of the location of the test file.
   export BATS_TENZIR_DATADIR="${BATS_TEST_DATADIR%%/integration/*}/data"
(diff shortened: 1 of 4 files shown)


dominiklohmann (approved):

cmake/TenzirRegisterPlugin.cmake
```diff
@@ -222,27 +222,12 @@ function (TenzirCompileFlatBuffers)
...
-    set(output_file
-        "${output_prefix}/${FBS_INCLUDE_DIRECTORY}/${basename}_generated.h")
-    # The path that we want.
-    set(desired_file
+    # The expected output path.
+    set(output_path
         "${output_prefix}/${FBS_INCLUDE_DIRECTORY}/${basename}.hpp")
-    # Hackish way to patch generated FlatBuffers schemas to support our naming.
-    set("rename_${basename}"
-        "${CMAKE_CURRENT_BINARY_DIR}/flatbuffers_strip_suffix_${basename}.cmake"
```
dominiklohmann, line 233: As we discussed last week in Slack, I think you need to drop the commit that changes this, as it doesn't work with older versions of FlatBuffers that we still rely on in our Dockerfile and CI.

tenzir/integration/lib/bats-tenzir/src/fixtures.bash
```diff
@@ -27,34 +27,32 @@ export_default_node_config() {
 
 bats_tenzir_initialize() {
   # Normalize the environment unless `BATS_TENZIR_KEEP_ENVIRONMENT` is set.
-  if [[ ! -n ${BATS_TENZIR_KEEP_ENVIRONMENT} ]]; then
-    unset $(printenv | grep -o '^TENZIR[^=]*' | paste -s -)
+  if [[ -z ${BATS_TENZIR_KEEP_ENVIRONMENT} ]]; then
+    unset "$(printenv | grep -o '^TENZIR[^=]*' | paste -s -)"
   fi
```
dominiklohmann, line 32: This doesn't seem to work as expected.

tobim: Yeah, adding the quotes was wrong. It's fixed now.

PR #4100, opened 2024-04-06
Add support for void-to-void operators

dominiklohmann:
These operators are effectively commands, i.e., they cannot actually be combined with any other operator, except for operator modifiers.

The rationale for this is an upcoming change to `context save`, `context update`, and `context create` that turns the three operators into sinks. But since `context create` already is a source operator this means that we now had to implement support for operators that are both a source and a sink at the same time. Luckily this was rather easy to implement.

For testing, I modified the `version` soucce operator to return a `generator<std::monostate>` and to print the event on the stderr instead. The only code outside of the execution nodes that I needed to modify was the `every` operator modifier that ran into an assertion for void-to-void operators.

### DIFF
--- a/libtenzir/builtins/operators/every.cpp
+++ b/libtenzir/builtins/operators/every.cpp
@@ -119,8 +119,11 @@ public:
     if (output->is<table_slice>()) {
       return run<table_slice>(op_->copy(), interval_, std::move(input), ctrl);
     }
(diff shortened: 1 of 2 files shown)


jachris (approved):
Looks good to me, though I did not really test this as we have no non-trivial operator that already makes use of this.

PR #4102, opened 2024-04-06
Add targets to CMake for updating test references

dominiklohmann:
Purely a convenience thing, because explaining that for every integration test target `integration*` there is a corresponding `update-integration*` is a lot easier than remembering which environment variable to set for updating test references.

### DIFF
--- a/tenzir/integration/README.md
+++ b/tenzir/integration/README.md
@@ -8,7 +8,7 @@ the test output against a known reference file.
 
 # Running integration tests
 
-To run all tests, execute the `integration` target from cmake.
+To run all tests, execute the `integration` target from CMake.
 
 To run all tests in this directory, use `bats tests/`.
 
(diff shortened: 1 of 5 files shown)


tenzir/integration/README.md
```diff
@@ -8,7 +8,7 @@ the test output against a known reference file.
 
 # Running integration tests
 
-To run all tests, execute the `integration` target from cmake.
+To run all tests, execute the `integration` target from CMake.
```
mavam, line 11: Would it be target inflation if we had one target per test file? If not, I would love to switch away from manually invoking bats entirely. But I often only care for a single suite.

mavam (approved):

tenzir/integration/README.md
dominiklohmann: I think having one target per test suite is an excellent idea. I've written down a note to do that some time in the future.

PR #4107, opened 2024-04-10
Adjust `chart` syntax from `--key value` to `key=value`

eliaskosunen:
Done in preparation for https://github.com/tenzir/issues/issues/1637, which specifies a `key=value` syntax instead of `--key value`, which we have currently.

Simultaneously, also removes the `--title` option from `chart`, closing https://github.com/tenzir/issues/issues/1653


jachris:
This was superseded by #4119.

PR #4109, opened 2024-04-10
Generalize `every` to work with all operators

dominiklohmann:
This enables streaming aggregations by turning blocking operators into operators that stream their results at a fixed interval. This is best explained on an example. The following pipeline returns the total ingress as bytes per minute per pipeline:

```
metrics --live
| where #schema == "tenzir.metrics.operator"
| where source == true
| every 1min summarize ingress=sum(output.approx_bytes), duration=sum(duration) by pipeline
```

Similarly, this pipeline returns for every managed pipeline how many diagnostics were encountered, grouped by severity (warning or error):

```
diagnostics --live
| every 10s count=count(.) by severity
```

### DIFF
--- a/libtenzir/builtins/operators/every.cpp
+++ b/libtenzir/builtins/operators/every.cpp
@@ -74,25 +74,50 @@ public:
     return result;
   }
 
-  template <class T>
+  template <class Input, class Output>
   static auto run(operator_ptr op, duration interval, operator_input input,
-                  operator_control_plane& ctrl) -> generator<T> {
-    TENZIR_ASSERT(std::holds_alternative<std::monostate>(input));
+                  operator_control_plane& ctrl) -> generator<Output> {
     auto alarm_clock = ctrl.self().spawn(make_alarm_clock);
-    auto next_run = time::clock::now() + interval;
+    auto now = time::clock::now();
+    auto next_run = now + interval;
     co_yield {};
+    auto make_input = [input = std::move(input), &now, &next_run]() mutable {
+      if constexpr (std::is_same_v<Input, std::monostate>) {
+        (void)now;
+        (void)next_run;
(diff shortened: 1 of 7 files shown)


web/docs/language/operator-modifiers.md
```diff
@@ -8,7 +8,7 @@ Operator modifiers are keywords that may occur before an operator.
 
 ## Scheduled Executions
 
-The special keyword `every` enables scheduled execution of a source operator.
+The special keyword `every` enables scheduled execution of an operator.
```
dominiklohmann, line 11: @mavam with this change we should think about putting operator modifiers into a more prominent place. My current thinking is that we may just want to document them as operators and stop distinguishing between operators and operator modifiers in the documentation in the first place.

It's a similar situation to void-to-void operators in that operator modifiers are hard to categorize as either source, transformation or sink, which our current docs sidebar layout enforces too strictly.

libtenzir/builtins/operators/every.cpp
jachris, line 1: TODO: Changelog.

jachris, line 1: The case where the inner operator is a `pipeline` does not look entirely right under the new semantics. Perhaps we just want to reject it for now?

jachris (approved):

libtenzir/builtins/operators/every.cpp
dominiklohmann: I had to revert this change, as it's really hard to forbid `every … top …` but not `every … from …`. And the latter is definitely needed.

I think it's better to live with the slightly surprising behavior of `every … top …` for now. We'll fix this with nested pipeline execution in the future.

I've enabled auto-merge, please resolve if you agree.

PR #4111, opened 2024-04-11
Add value grouping to `chart`

eliaskosunen:
Closes https://github.com/tenzir/issues/issues/1637. Depends on https://github.com/tenzir/tenzir/pull/4107

This is quite messy and somewhat rushed, but it seems to be working.

The vast majority of the diff is in there to support list arguments in `chart` (like `y=[foo, bar]`). To add the `position` argument, only the small additions to `chart_definitions` are required.


libtenzir/builtins/operators/chart.cpp
```diff
@@ -453,7 +567,93 @@ class chart_argument_parser {
...
+              fmt::format("{}, ", escape_operator_arg(val->inner.substr(
+                                    0, val->inner.size() - 1))));
+            continue;
+          }
+          if (val->inner.ends_with(']') && val->inner.size() > 1) {
+            values_str.append(
+              fmt::format("{}]", escape_operator_arg(val->inner.substr(
+                                   0, val->inner.size() - 1))));
+            break;
+          }
```
dominiklohmann, line 622: This we need to take a deeper look at, not sure if this is correct.

jachris:
The commit was cherry-picked and is part of #4119.

PR #4114, opened 2024-04-12
Remove metrics from `/pipeline/list`

dominiklohmann:
We recommend getting metrics from the `metrics` operator for good reason, and it has been long enough now that we can safely remove the metrics from the endpoint. They were problematic from a performance perspective, as they required the `/pipeline` API to inspect operator metrics.

### DIFF
--- a/web/openapi/openapi.yaml
+++ b/web/openapi/openapi.yaml
@@ -374,12 +374,6 @@ paths:
                       error: null
                       diagnostics:
                         []
-                      metrics:
-                        total:
-                          ingress:
-                            {}
-                          egress:
-                            {}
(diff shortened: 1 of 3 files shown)


web/openapi/openapi.yaml
Dakostu, line 1: The `/pipeline/list` example in this file still contains `metrics`.

Dakostu (approved):
Less code = Always better.
 I found a small oversight, but otherwise good to go.

web/openapi/openapi.yaml
dominiklohmann: Fixed in 99f5ff2358df433f5f2bb197619ec1db11c4cc43

PR #4115, opened 2024-04-12
Refactor platform CLI docs

lava:
This mostly moves around the different sections of the platform CLI docs, to more clearly separate on-premise exclusive commands from generally available ones, and to move simple commands to the beginning and complex commands to the end.

* The high-level is now subdivided in 4 blocks, "Authentication", "Manage Nodes", "Manage Workspaces" and "Configure Access Rules", two of which are exclusive to on-prem deployments.
* The `workspace list` and `workspace select` commands are moved into the Authentication section, the rules-related commands are moved out.
* Added some additional text to some of the auth-rule related documentation

### DIFF
--- a/web/docs/platform-cli.md
+++ b/web/docs/platform-cli.md
@@ -1,7 +1,7 @@
 # Platform CLI
 
-The *Tenzir Platform CLI* manages authentication, workspaces, and nodes for the
-*Tenzir Platform*.
+The *Tenzir Platform CLI* allows users to interact with the *Tenzir Platform* from the command-line
+to manage their workspaces and nodes.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


web/docs/platform-cli.md
```diff
@@ -18,150 +18,181 @@ pip install tenzir-platform
...
 
-### Description
+#### Description
 
 The `tenzir-platform workspace admin list-global-workspaces`, `tenzir-platform
 admin create-workspace`, and `tenzir-platform admin delete-workspace` commands
 list, create, or delete workspaces, respectively.
 
-#### `<owner_namespace>`
+##### `<owner_namespace>`
```
dominiklohmann, line 111: This looks really off compared to our operator docs when rendered, where we use h3 for all parameter descriptions. Do we need the extra nesting level?

lava: Not necessarily, I split it into two separate sections now.

dominiklohmann (approved):

PR #4117, opened 2024-04-12
Support 0mq inproc sockets

mavam:
As a poorman's pub/sub, we want to use our existing `zmq` connector in combination with `bitz`. This requires support for [inproc sockets](http://api.zeromq.org/4-2:zmq-inproc) sockets, which this PR adds.

### DIFF
--- a/plugins/zmq/src/plugin.cpp
+++ b/plugins/zmq/src/plugin.cpp
@@ -36,13 +36,14 @@ struct saver_args {
   std::optional<located<std::string>> endpoint;
   std::optional<location> connect;
   std::optional<location> listen;
+  std::optional<location> monitor;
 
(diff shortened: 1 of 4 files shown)


dominiklohmann (approved):
This works well in practice. I have some minor review comments, but I already realized them in code locally, so after you've merged this I'll just open a follow-up PR.

plugins/zmq/src/plugin.cpp
dominiklohmann, line 1: I just found one issue with this after approval that we should look into, either in this PR or in a follow-up:

When the publishing side ends, the subscribing side should either attempt to reconnect (preferred) or stop (ok, but not ideal). Right now it just hangs and does nothing.

PR #4119, opened 2024-04-15
Add value grouping to `chart` and remove `--title`

jachris:
This PR is basically #4111 without #4107 plus a few changes to make it work. The syntax for specifying multiple fields becomes `-y foo,bar,baz`. The appearance can be configured with `--position stacked` and `--position grouped`.


changelog/next/changes/4111--chart-implicit-values.md
dominiklohmann, line 1: Can you merge the two changelog entries for changes into a single file with two paragraphs? The prepare-release script does not handle this well otherwise—which should be considered a bug, but that really doesn't have any priority.

libtenzir/builtins/operators/chart.cpp
```diff
@@ -38,12 +44,12 @@ struct schema_name {
...
 };
 struct field_name {
-  std::string field;
+  std::vector<std::string> fields;
 
   friend auto inspect(auto& f, field_name& x) -> bool {
     return f.object(x)
-      .pretty_name("field_name")
-      .fields(f.field("field", x.field));
+      .pretty_name("field_names")
```
dominiklohmann, line 51: Shouldn't this match the name of the struct?

dominiklohmann (approved):
Thanks for taking this over. I went through the code and did try it briefly, and things seem to work expected.

PR #4120, opened 2024-04-15
Fix `every` operator when used as a source

jachris:
The current implementation never updates `now` when `every` is used as a source. As a consequence, the calculated `delta` grows bigger and bigger and the effective delay is multiplied by `n` for the `n`-th iteration. This PR makes sure that `now` is always updated.

As the bug was only recently introduced (#4109), no changelog is necessary.


dominiklohmann (approved):
Whoops. 🙈

PR #4123, opened 2024-04-16
Fix a race condition in `/serve`

dominiklohmann:
This fixes a very odd race condition in the bridge between the `serve` operator and the `/serve` API endpoint. When a pipeline completes, we keep its last result set around so that it can be retrieved in case the last request failed. However, when a second pipeline using serve in precisely that time frame was started and ran into a timeout on its first request, we accidentally leaked resources for the first pipeline and had the request for the second pipeline go unanswered, leading to a 504 error in the explorer.

The cuplrit behind this was in a wrong comparison in the handler that delivers underful results when a timeout hit that compared using the continuation token as opposed to the serve id. Since both the initial and the final continuation tokens are null, it found the wrong serve request, and then failed to answer for it as that was already done. The serve request it was actually supposed to answer was left unanswered.

The fix for this is rather simple. I also rewrote some checks into assertions that would've caught this issue a lot earlier.

### DIFF
--- a/libtenzir/builtins/operators/serve.cpp
+++ b/libtenzir/builtins/operators/serve.cpp
@@ -260,9 +260,7 @@ struct managed_serve_operator {
   /// not enough results are buffered.
   /// @returns Whether the results were delivered.
   auto try_deliver_results(bool force_underful) -> bool {
-    if (not get_rp.pending()) {
-      return false;
-    }
+    TENZIR_ASSERT(get_rp.pending());
     // If we throttled the serve operator, then we can continue its operation
     // again if we have less events buffered than desired.
     if (put_rp.pending() and rows(buffer) < std::max(buffer_size, requested)) {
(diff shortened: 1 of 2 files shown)


jachris (approved):

libtenzir/builtins/operators/serve.cpp
```diff
@@ -486,21 +485,24 @@ struct serve_manager_state {
...
           = std::find_if(ops.begin(), ops.end(), [&](const auto& op) {
-              return op.continuation_token == continuation_token;
+              return op.serve_id == serve_id;
             });
         if (found == ops.end()) {
           TENZIR_DEBUG("unable to find serve request after timeout expired");
           return;
         }
+        TENZIR_ASSERT(not found->done);
+        TENZIR_ASSERT(found->continuation_token == continuation_token);
```
lava, line 502: Wouldn't this trigger if a user successfully retries the `/serve` request before the timeout has run out?

dominiklohmann: No, in that case the delayed run would be disposed.

PR #4124, opened 2024-04-16
Introduce the `timeout` operator modifier

dominiklohmann:
... and remove the one-off timeout option that we added to the `batch` operator.

### DIFF
--- a/libtenzir/builtins/operators/every_timeout.cpp
+++ b/libtenzir/builtins/operators/every_timeout.cpp
@@ -19,7 +19,7 @@
 #include <arrow/type.h>
 #include <caf/typed_event_based_actor.hpp>
 
-namespace tenzir::plugins::every {
+namespace tenzir::plugins::every_timeout {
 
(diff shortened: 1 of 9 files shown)


libtenzir/builtins/operators/every_timeout.cpp
jachris, line 1: Unless I'm mistaken, the semantics here seem very different from `batch 10000 --timeout 10s`, which would guarantee a batch every `10s`. However, `timeout 10s batch 10000` can wait for more than 27 hours before yielding the first batch. This is not what we want, right?

dominiklohmann:
@jachris and I discussed this and given our internal use cases for this being still a bit out we decided to close this for now and to revisit it later. Ultimately, having timeout primitives like this is really valuable, e.g., `deduplicate` can be expressed as `group … timeout … head …`, and we also will need this for the Feather store in the future, but then we probably want to allow counting both from the first and the last event, respectively.

PR #4125, opened 2024-04-17
Fix Pipelines as Code examples

dominiklohmann:
These were missing a level of indentation, as reported by a user via Discord.


jachris (approved):
Oops!

PR #4126, opened 2024-04-17
Add configurable contexts

tobim:
Similar to configurable pipelines, we can now also configure contexts in `tenzir.yaml` files:
```
tenzir:
  contexts:
    lt:
      type: lookup-table
    geoip:
      type: geoip
      arguments:
        db-path: /path/to/GeoLite2-ASN.mmdb
    bf:
      type: bloom-filter
      arguments:
        capacity: 100
        fp-probability: 0.1

```

### DIFF
--- a/changelog/next/features/4126--configurable-contexts.md
+++ b/changelog/next/features/4126--configurable-contexts.md
@@ -1,0 +1,2 @@
+You can now define contexts and their creation parameters in the
+`tenzir.contexts` section of the configuration file.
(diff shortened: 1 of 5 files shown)


contrib/tenzir-plugins
dominiklohmann, line 1: Please add a changelog entry for this.

dominiklohmann (approved):
We went over this together. Please address the comments in the PR in the submodule, add a changelog entry, and then feel free to merge.

PR #4128, opened 2024-04-18
Fix links to docs throughout the code

dominiklohmann:
This does one fell swoop over the entire codebase to fix all links to our documentation page.


tobim (approved):
I did not test these explicitly, but looking at the diff the new link targets look ok.

PR #4129, opened 2024-04-18
Fix configured pipeline autostart

jachris:
This PR contains some fixes to actually make configured pipelines autostart.


dominiklohmann (approved):
The changes all make sense to me. Thanks for fixing this!

Looks like there is still one test failure that should be investigated, did you miss an update to a reference file?

jachris:
> Looks like there is still one test failure that should be investigated, did you miss an update to a reference file?

The behavior of the existing tests was unstable. I modified them accordingly now.

dominiklohmann (approved):
I looked at the additional commits, the changes make sense to me.

PR #4130, opened 2024-04-19
Make publish and subscribe internal operators

dominiklohmann:
Let's not count these as ingress or egress.


mavam (approved):

dominiklohmann:
Eh, just noticed that the approach I've taken here can't work because of the current implementation. I'll write a story to address this properly in the near future.

PR #4133, opened 2024-04-19
Update Suricata schema with additional flow fields

satta:
This PR adds some more detail to the `suricata.flow` record type in the Suricata schema by including fields that were previously missing. This is aimed at reducing the number of non-schema fields encountered in potential input from recent Suricata versions, which would then lead to a larger number of schema layouts that need to be inferred based on the observed event content.

Please review and -- if necessary -- ensure that reference files in the tests are updated.

### DIFF
--- a/schema/types/suricata.schema
+++ b/schema/types/suricata.schema
@@ -10,6 +10,18 @@ type suricata.component.common = record {
   dest_port: port,
   proto: string,
   event_type: string,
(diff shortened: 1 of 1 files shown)


dominiklohmann:
I'll merge this into a new branch and will open a PR for that with the reference file updates. Thanks for the contribution!

PR #4134, opened 2024-04-19
Fix HTTP saver Content-Length computation

mavam:
When using the `http` saver, the `Content-Length` header wasn't computed correctly.

(NB: The bug was that we set `CURLOPT_POSTFIELDSIZE` to 0 when preparing the request and used `CURLOPT_INFILESIZE` to set it again when we got the first chunk of data to submit. Since it depends on the request method what size libcurl looks at, the wrong size got picked for POST requests and the right one for, say, PUT requests. We're now setting *both* sizes when receiving chunks.)


dominiklohmann (approved):

mavam:
@dominiklohmann I re-requested a review because I added a bunch of integration tests for the `http` connector.

tenzir/integration/tests/http.bats
```diff
@@ -1,39 +1,54 @@
...
+  if ! command -v python; then
+    skip "python executable must be in PATH"
   fi
+  check --bg server python "${MISCDIR}/scripts/webserver.py"
+  tenzir "$@"
+  wait_all ${server[@]}
+}
+
+@test "load HTTP POST" {
+  load 'load http://localhost:8888 foo=42'
```
dominiklohmann, line 28: These tests do not correctly run in parallel because they all use the same port. 8888 is also a rather commonly used port, and because CI runners may be shared this is likely to cause sporadic CI failures in CI.

Other than that this is all good—just please use a port that's less commonly used and use different ports for each test.

dominiklohmann (approved):

PR #4135, opened 2024-04-21
Add docs page that compares Tenzir to Cribl

mavam:
This PR adds a page to our docs that compares Tenzir with Cribl.


dominiklohmann (approved):
I think this is a good starting point that we can continuously improve upon. Since we want to release tomorrow, I suggest that we merge this and then put any follow-up changes into a new PR.

PR #4137, opened 2024-04-22
Fix verification of large FlatBuffers tables

dominiklohmann:
This fixes a bug when loading large lookup tables or other data persisted in a FlatBuffers table from disk. The default behavior of the FlatBuffers table verifier is to blame here.

### DIFF
--- a/libtenzir/include/tenzir/flatbuffer.hpp
+++ b/libtenzir/include/tenzir/flatbuffer.hpp
@@ -140,9 +140,16 @@ public:
     if (not result) {
       return std::move(result.error());
     }
+    // FlatBuffers defaults to erroring out after 1M table entries in the
+    // verifier. This was chosen rather randomly and for historic reasons, they
+    // cannot change it. We use the much saner default of not erroring out for
+    // large tables here.
+    auto options = flatbuffers::Verifier::Options{};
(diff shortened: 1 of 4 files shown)


libtenzir/include/tenzir/flatbuffer.hpp
```diff
@@ -140,9 +140,16 @@ class flatbuffer final {
...
+    // verifier. This was chosen rather randomly and for historic reasons, they
+    // cannot change it. We use the much saner default of not erroring out for
+    // large tables here.
+    auto options = flatbuffers::Verifier::Options{};
+    options.max_tables = std::numeric_limits<flatbuffers::uoffset_t>::max();
     const auto* const data
       = reinterpret_cast<const uint8_t*>(result->chunk()->data());
-    auto verifier = flatbuffers::Verifier{data, result->chunk()->size()};
+    auto verifier
+      = flatbuffers::Verifier{data, result->chunk()->size(), options};
```
jachris, line 152: Thanks for the accompanying explanation!

We also use `flatbuffers::Verifier` for `type` and `table_slice`. For `type` is probably doesn't matter, but perhaps we need the same change for `table_slice`?

jachris (approved):

PR #4139, opened 2024-04-22
Find syntax errors before record batch reading

balavinaithirthan:
1. Instead of compiling and executing the code at every record batch, compile only once at the beginning. Therefore, syntax errors can be identified even in a streaming context where the first record batch has not arrived.

### DIFF
--- a/libtenzir/builtins/operators/python.cpp
+++ b/libtenzir/builtins/operators/python.cpp
@@ -265,13 +265,13 @@ public:
       }
       codepipe.close();
       ::close(errpipe.pipe().native_sink());
-      TENZIR_WARN("Done with set up");
       co_yield {}; // signal successful startup
-      TENZIR_WARN("Hi");
       for (auto&& slice : input) {
         if (!child.running()) {
           auto python_error = drain_pipe(errpipe);
-          diagnostic::error("{}", python_error).emit(ctrl.diagnostics());
+          diagnostic::error("{}", python_error)
+            .note("python process exited with exit-code 1")
+            .emit(ctrl.diagnostics());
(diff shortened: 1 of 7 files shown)


python/tenzir/tools/python_operator_executor.py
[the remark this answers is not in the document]
balavinaithirthan: Thanks, was thinking about a way to test this!

tenzir/integration/data/reference/python/test_python_operator_invalid_syntax/step_00.ref
jachris, line 1: Can you add the definitions of these new tests to the PR?

libtenzir/builtins/operators/python.cpp
```diff
@@ -265,6 +267,13 @@ class python_operator final : public crtp_operator<python_operator> {
       ::close(errpipe.pipe().native_sink());
       co_yield {}; // signal successful startup
       for (auto&& slice : input) {
+        if (!child.running()) {
+          auto python_error = drain_pipe(errpipe);
+          diagnostic::error("{}", python_error)
+            .note("python process exited with exit-code 1")
+            .emit(ctrl.diagnostics());
```
jachris, line 274: Nitpick: You don't know that the exit-code was 1. You can use the actual exit code instead (same below).

libtenzir/builtins/operators/python.cpp
jachris, line 1: Can you create a changelog entry for the improved behavior?

jachris (approved):
This works great, thank you!

PR #4143, opened 2024-04-23
Remove events output from many context operators

dominiklohmann:
This change removes the events output from quite a few context-related operators, and also makes less of them take generic arguments in favor of fixed arguments that produce better diagnostics when they're misused.


tobim (approved):
Went over this in a review call.

dominiklohmann:
Closing in favor of https://github.com/tenzir/tenzir/pull/4146, as the CI got stuck here for hours now and I don't really know what's going on.

PR #4145, opened 2024-04-23
Fix an off-by-one in the `slice` operator

dominiklohmann:
This fixes `batch X | slice --begin Y` where $`X = Y - 1`$.


jachris (approved):
Thanks for fixing this.

PR #4147, opened 2024-04-23
Add an operator for dumping schemas

dominiklohmann:
This is pretty helpful to get started with writing schemas, as it allows one to go from inferred events over to having a manually specified schema without needing to write everything from scratch.

The operator name is still a TBD and this definitely needs documentation _iff_ we want to keep this. This originated in a discussion in the community Discord, where a new user suggested this workflow.

### DIFF
--- a/libtenzir/builtins/operators/schemaof.cpp
+++ b/libtenzir/builtins/operators/schemaof.cpp
@@ -1,0 +1,78 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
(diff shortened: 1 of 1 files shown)


lava (approved):
I applied this locally and verified that it works as expected.

However, the string format is a bit awkward to use, in order to do anything useful with the schema afterwards (ie. search for the presence of a field with some specific type) one has to resort to substring search.

I think it would be worth considering to turn the `definition` in a nested field, or as a compromise maybe in JSON format so that it can be easily postprocessed by other tools.

dominiklohmann:
Closing this for now, I think this makes much more sense to have as a function in TQL2.

PR #4148, opened 2024-04-23
Keep changelog .gitkeep files when cutting a release

dominiklohmann:
This is a small fix for the prepare-release script, which before this required manual intervention to move around the changelog .gitkeep files, which while not a lot of effort was easy to overlook.


jachris (approved):
Not tested, but looks good.

PR #4149, opened 2024-04-24
Consolidate community and enterprise edition builds

tobim:


mavam (approved):

.github/workflows/docker-manual.yaml
```diff
@@ -89,15 +85,6 @@ jobs:
                 }
               else
                 empty
-              end,
-              if ${{ inputs.push-ee-to-ecr }} then
-                {
-                  "name": "tenzir-ee", "target": "tenzir-ee", "registries": [
-                   "622024652768.dkr.ecr.eu-west-1.amazonaws.com"
```
dominiklohmann, line 96: @tobim doesn't this break things? We don't push the other image to ECR.

tobim: This file is only used for the manual dispatched workflow. The only image that is automatically pushed to ECR is `tenzir-demo`, and that is unchanged by this PR.

nix/overlay.nix
```diff
@@ -384,24 +384,14 @@ in {
...
-        pname = "tenzir-ee";
-      };
     in
       pkg.withPlugins (ps: [
         ps.compaction
         ps.context
         ps.matcher
         ps.pipeline-manager
         ps.platform
+        ps.vast
```
dominiklohmann, line 394: Why do we need to define all the plugins here again when they're already defined here? https://github.com/tenzir/tenzir/blob/topic/unify-builds/nix/tenzir/plugins/names.nix

tobim: Now that we have the complete list here we can do that, but we couldn't before when we had individual edition builds. This code could be refactored to reflect that change, but I didn't have the time to do it yet.

dominiklohmann: Okay, I'll write a story for this.

PR #4154, opened 2024-04-24
Make the `publish` operator unordered

dominiklohmann:
... and introduce the `unordered` modifier.

### DIFF
--- a/libtenzir/builtins/operators/unordered.cpp
+++ b/libtenzir/builtins/operators/unordered.cpp
@@ -1,0 +1,103 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2024 The Tenzir Contributors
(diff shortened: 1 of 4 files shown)


libtenzir/builtins/operators/unordered.cpp
```diff
@@ -0,0 +1,103 @@
...
+      op_ = std::move(op->op_);
+    }
+    TENZIR_ASSERT(not dynamic_cast<const unordered_operator*>(op_.get()));
+  }
+
+  auto optimize(const expression& filter, event_order order) const
+    -> optimize_result override {
+    (void)order;
+    return op_->optimize(filter, event_order::unordered);
+  }
```
jachris, line 34: The `event_order::unordered` argument allows `op` to assume that the behavior of the pipeline that comes after it does not depend on the order of its input. I think you want to modify the returned `optimize_result::order` instead perhaps? Or perhaps even both?

jachris (approved):
Even though we might want to change this eventually, this solves a problem right now, so I'm completely fine with it.

web/docs/language/operator-modifiers.md
jachris, line 1: And perhaps you can also document the unorderedness of `publish`.

PR #4158, opened 2024-04-25
Add new loading mechanism for GeoIP context

balavinaithirthan:
1. We added an additional loader to the GeoIP plugin to allow for users to create an empty context, load data in later, and replace contexts through data loads.  The loader allows for the original V1 loader to still work because it loads the V1 context into memory and on save converts it to a V2 context. The `context save context_name` should not be used alone because an implicit writing sink is added to the end; instead it should continue being used to pipe data into `context load`.

Questions:
1. Is it possible to rebase onto main with the tensor-plugins submodule conflict?


libtenzir/builtins/contexts/geoip.cpp
```diff
@@ -531,9 +551,53 @@ struct v1_loader : public context_loader {
...
+  }
+
+  auto load(chunk_ptr serialized) const
+    -> caf::expected<std::unique_ptr<context>> {
+    const auto* cache_dir
+      = get_if<std::string>(&global_config_, "tenzir.cache-directory");
+    const auto current_time = std::time(nullptr);
+    std::string temp_file_name = *cache_dir + std::to_string(current_time);
+    TENZIR_WARN(temp_file_name);
+    auto myfile
```
mavam, line 570:
```suggestion
    auto mmdb_file
```

The "my" prefix is often used in example, but we don't use it in production code. Recommendation: give it simply the most descriptive name you can think of, so that readers ideally know what it is by its name and type.

balavinaithirthan: Sorry, I opened for review on accident. Will do!

libtenzir/builtins/contexts/geoip.cpp
```diff
@@ -475,6 +490,11 @@ class ctx final : public virtual context {
   }
 
   auto reset() -> caf::expected<void> override {
+    if (!mmdb_) {
+      return caf::make_error(ec::lookup_error,
+                             fmt::format("no GeoIP data currently exists for "
+                                         "this context"));
```
dominiklohmann, line 496: This function should just be a no-op and always return `{}` in all cases, as the resetting has no effect.

libtenzir/builtins/contexts/geoip.cpp
```diff
@@ -490,18 +510,19 @@ class ctx final : public virtual context {
...
-    auto geoip_data = geoip_builder.Finish();
-    fbs::context::geoip::FinishGeoIPDataBuffer(builder, geoip_data);
-    return save_result{.data = tenzir::fbs::release(builder), .version = 1};
+    if (!mmdb_) {
+      return caf::make_error(ec::lookup_error,
+                             fmt::format("no GeoIP data currently exists for "
+                                         "this context"));
+    }
+    return save_result{.data = std::move(chunk::mmap(db_path_).value()),
+                       .version = latest_version};
```
dominiklohmann, line 519: Missing error handling for `chunk::mmap(...)`.

libtenzir/builtins/contexts/geoip.cpp
```diff
@@ -531,9 +552,59 @@ struct v1_loader : public context_loader {
...
+  }
+
+  auto version() const -> int {
+    return 2;
+  }
+
+  auto load(chunk_ptr serialized) const
+    -> caf::expected<std::unique_ptr<context>> {
+    const auto* cache_dir
+      = get_if<std::string>(&global_config_, "tenzir.cache-directory");
```
dominiklohmann, line 567: Since we know that the cache directory option is always set, we should express that clearly to the reader of the code.

```suggestion
    const auto* cache_dir
      = get_if<std::string>(&global_config_, "tenzir.cache-directory");
    TENZIR_ASSERT(cache_dir);
```

libtenzir/builtins/contexts/geoip.cpp
```diff
@@ -531,9 +552,59 @@ struct v1_loader : public context_loader {
...
+    std::filesystem::create_directories(*cache_dir);
+    const auto current_time = std::time(nullptr);
+    std::string temp_file_name = *cache_dir + std::to_string(current_time);
+    auto temp_file = std::fstream(temp_file_name, std::ios_base::out);
```
dominiklohmann, line 571: Please use something like `{cache_dir}/plugins/geoip/{random_uuid}` for the filename.

libtenzir/builtins/contexts/geoip.cpp
```diff
@@ -490,18 +510,19 @@ class ctx final : public virtual context {
...
+      return caf::make_error(ec::lookup_error,
+                             fmt::format("no GeoIP data currently exists for "
+                                         "this context"));
+    }
+    return save_result{.data = std::move(chunk::mmap(db_path_).value()),
+                       .version = latest_version};
   }
 
 private:
   std::string db_path_;
```
dominiklohmann, line 523: We can remove the file path here and replace it with a memory-mapped chunk of the file contents for use in `save`.

libtenzir/builtins/contexts/geoip.cpp
```diff
@@ -6,6 +6,9 @@
 // SPDX-FileCopyrightText: (c) 2023 The VAST Contributors
 // SPDX-License-Identifier: BSD-3-Clause
 
+#include "tenzir/detail/posix.hpp"
+#include "tenzir/logger.hpp"
+
```
dominiklohmann, line 11: "" -> <>

libtenzir/builtins/contexts/geoip.cpp
```diff
@@ -531,9 +552,59 @@ struct v1_loader : public context_loader {
...
+                                         "on data load: {}",
+                                         detail::describe_errno()));
+    }
+    temp_file.close();
+    if (!temp_file) {
+      return caf::make_error(ec::filesystem_error,
+                             fmt::format("failed close the temp file: {}",
+                                         detail::describe_errno()));
+    }
+    auto mmdb = make_mmdb(temp_file_name);
```
dominiklohmann, line 592: To guarantee that the temp file is still alive, it should only be closed after the call to `make_mmdb` so that there's always at least one reference to the file.

libtenzir/builtins/contexts/geoip.cpp
tobim: I second the request to move away from `std::time` for the intermediate file.
I'm not sure we should put it in `tenzir.cache-directory`, because looking at the `ctx` class I get the impression that the file gets read sometimes even after the context was created. A cache on the other hand should not be required for correctness.

balavinaithirthan: The logic has been changed so that the directory is purely a cache now. It never needs to be accessed later.

dominiklohmann (approved):
Awesome, thanks for this! I think this is a very solid improvement to the geoip context usability.

PR #4166, opened 2024-04-29
Fix shutdown of `every <interval> <transformation|sink>`

dominiklohmann:
`every` when used with a transformation or a sink did not shut down properly before this change.

### DIFF
--- a/libtenzir/builtins/operators/every.cpp
+++ b/libtenzir/builtins/operators/every.cpp
@@ -80,8 +80,9 @@ public:
     auto alarm_clock = ctrl.self().spawn(make_alarm_clock);
     auto now = time::clock::now();
     auto next_run = now + interval;
+    auto done = false;
     co_yield {};
-    auto make_input = [input = std::move(input), &next_run]() mutable {
+    auto make_input = [&, input = std::move(input)]() mutable {
       if constexpr (std::is_same_v<Input, std::monostate>) {
         (void)next_run;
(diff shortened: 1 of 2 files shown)


libtenzir/builtins/operators/every.cpp
jachris, line 1: Strictly speaking, the new behavior "does not work" for transformations that can have output even when there is no input. But I feel like this is an edge case that we do not need to address.

jachris (approved):

libtenzir/builtins/operators/every.cpp
dominiklohmann: I think this does actually work, as this waits for both the input to be depleted and the current output to be generated entirely.

jachris: Right, but something like `version | write json | every 1s 'shell "echo okay"' | read lines` can theoretically continue after both input and one of the output generators is depleted. But obviously this is a very weird example, so let's just merge this.

dominiklohmann: Isn't that a bug in `shell`, which should let its generate deplete only after the child process existed and all of its output was forwarded?

jachris: I think this is what already happens. But we instantiate a new child process `every 1s`.

dominiklohmann: Ah. I would consider that the expected behavior to be honest.

PR #4169, opened 2024-04-29
Add `--precise` mode to JSON parser

jachris:
With this PR, the `json` parser gains a new `--precise` flag, which ensures that the layout of the emitted events precisely match the input. For example, it guarantees that no additional `null` fields will be added, which otherwise happens because the events are packed into the same batch.

This is especially useful with formats such as GELF, where the set fields can vary significantly between events and the user is interested in the exact fields from the input. 

Overall, the performance depends a lot on the input. When testing this with a GELF file that contains many type conflicts between IP and string, I observed a speedup from 30s to 18s (because `--precise` separates events by their types), which is a solid improvement (the majority of the time is then spent in the `data` parsers). However, when also using `--raw`, which also prevents this type conflict, `--precise` slows us down from 9s to 12s (due to being 2-pass instead of 1-pass). However, in the case of GELF, the output of `--precise` is much closer to what the user wants, so I would deem this slowdown acceptable.    

The implementation uses a custom "data signature" approach to avoid calling `type::infer` on the parsed `data`, because it turns out that the latter has unacceptable performance implications for this use-case. Also, I added a `TRY`/`TENZIR_TRY` macro with a customization point for types such as `simdjson_result`, `expected`, `optional`, `arrow::Status`, etc.


libtenzir/builtins/formats/json.cpp
jachris, line 1: I'm very much open to naming suggestions better than `--precise`.

libtenzir/builtins/formats/json.cpp
```diff
@@ -728,40 +888,78 @@ class ndjson_parser final : public parser_base {
...
+        co_return;
+      }
+      signature_.clear();
+      append_signature(event, signature_);
+      auto it = state.precise_map.find(signature_);
+      if (it == state.precise_map.end()) {
+        // TODO: We should eventually garbage collect this.
+        auto index = state.entries.size();
+        state.entries.emplace_back("tenzir.json", std::nullopt);
+        it = state.precise_map.emplace_hint(it, signature_, index);
```
jachris, line 914: I would consider this a follow-up.

libtenzir/builtins/formats/json.cpp
dominiklohmann: We could also consider generally making this the default, and adding an option that re-enables the old behavior.

jachris: I'm almost wondering whether there should be a default at all.

dominiklohmann (approved):
This works well. I went through the code in detail, and it's a good improvement overall.

Before merging, we should come to an agreement on the naming of the option and the default behavior.

libtenzir/builtins/formats/json.cpp
dominiklohmann: I think it's clear that we want to have a more generic implementation for this that we can also use for Fluent Bit, CEF, and LEEF. We should do that for that implementaiton.

jachris: Alright, resolving then. (Also, see tenzir/issues#1809).

libtenzir/builtins/formats/json.cpp
jachris: We'll leave this as-is for now.

PR #4170, opened 2024-04-30
Fix reopening ports while subprocesses are open

tobim:
This fixes an unwanted interaction between the `http` and `tcp` connectors and the `python` and `shell` operators. When a connector opens a listening port the forked subprocesses would inherit the file descriptors representing the resource. That prevented the connector from being recreated while the subprocesses were still running.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -16,6 +16,14 @@ in
           pkgs.shfmt
           pkgs.poetry
           pkgs.python3Packages.spdx-tools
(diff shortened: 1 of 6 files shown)


shell.nix
jachris, line 1: 👆 Is this meant to be part of this PR? If yes, please just resolve.

jachris (approved):
Can you add a small description of why this change is necessary to the PR description before you merge it (for future reference in case we need it)? Thanks!

PR #4171, opened 2024-04-30
Chain the execution node start response promise to the node lifetime

Dakostu:
This change cleans up the response promise lifetime semantics a bit and also fixes a missing retry delay mechanism.

### DIFF
--- a/libtenzir/src/pipeline_executor.cpp
+++ b/libtenzir/src/pipeline_executor.cpp
@@ -62,10 +62,6 @@ void pipeline_executor_state::start_nodes_if_all_spawned() {
         finish_start();
       },
       [this](const caf::error& err) mutable {
-        if (err == caf::sec::broken_promise) {
-          abort_start(ec::silent);
-          return;
-        }
         abort_start(err);
       });
 }
--- a/changelog/next/bug-fixes/4171--retry-delay-fix-for-aborted-pipelines.md
+++ b/changelog/next/bug-fixes/4171--retry-delay-fix-for-aborted-pipelines.md
@@ -1,0 +1,1 @@
+The retry delay now works for pipelines that fail during startup.
--- a/libtenzir/src/execution_node.cpp
+++ b/libtenzir/src/execution_node.cpp
@@ -239,6 +239,8 @@ struct exec_node_state {
   std::optional<struct demand> demand = {};
   bool issue_demand_inflight = {};
 
+  caf::typed_response_promise<void> start_rp = {};
+
   /// Exponential backoff for scheduling.
   static constexpr duration min_backoff = std::chrono::milliseconds{10};
   static constexpr duration max_backoff = std::chrono::milliseconds{1000};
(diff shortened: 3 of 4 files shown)


contrib/tenzir-plugins
dominiklohmann, line 1: This needs a bug fix changelog entry.

libtenzir/src/pipeline_executor.cpp
```diff
@@ -62,10 +62,7 @@ void pipeline_executor_state::start_nodes_if_all_spawned() {
         finish_start();
       },
       [this](const caf::error& err) mutable {
-        if (err == caf::sec::broken_promise) {
-          abort_start(ec::silent);
-          return;
-        }
+        TENZIR_ASSERT(err != caf::sec::broken_promise);
```
dominiklohmann, line 65: That's too harsh—what if an operator exits with a broken promise?

contrib/tenzir-plugins
dominiklohmann, line 1: @tobim can you assess this part? I give my approval on the exec node and executor changes, but I'm not deep enough in this right now.

libtenzir/src/pipeline_executor.cpp
tobim: Agreed.

FWIW: I still get the `broken promise` diagnostic in the pipeline overview with this branch, the actual error is printed on stderr:
```
*** unhandled exception: [id: 362, name: user.scheduled-actor, exception typeid: boost.wrapexcept]: bind: Address already in use [system:98 at include/boost/asio/detail/reactive_socket_service.hpp:161:33 in function 'boost::system::error_code boost::asio::detail::reactive_socket_service<boost::asio::ip::tcp>::bind(implementation_type &, const endpoint_type &, boost::system::error_code &) [Protocol = boost::asio::ip::tcp]']
```
Pretty ugly, but at least it contains "Address already in use".

tobim: Hm, I think this is a local problem in `tcp-listen`.

dominiklohmann: I'll create a separate PR to generally catch exceptions in operators to turn them into diagnostics.

contrib/tenzir-plugins
tobim: The retry delay itself is fixed now.

tobim (approved):

contrib/tenzir-plugins
Dakostu: I added a changelog entry.

libtenzir/src/pipeline_executor.cpp
Dakostu: I removed the assetion.

dominiklohmann: Great, think this can be merged then. I'll open a follow-up PR shortly.

contrib/tenzir-plugins
Dakostu: I added an `early silent abort` integration test and one more edge case check. Should work now.

PR #4174, opened 2024-04-30
Render exceptions from operators as diagnostics

dominiklohmann:
This vastly improves error handling when interfacing with third-party libraries that throw an unexpected exception.

### DIFF
--- a/libtenzir/src/execution_node.cpp
+++ b/libtenzir/src/execution_node.cpp
@@ -746,6 +746,23 @@ auto exec_node(
     return exec_node_actor::behavior_type::make_empty_behavior();
   }
   self->state.weak_node = node;
+  self->set_exception_handler(
+    [self](std::exception_ptr exception) -> caf::error {
+      try {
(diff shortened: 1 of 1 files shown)


libtenzir/src/execution_node.cpp
```diff
@@ -746,6 +746,23 @@ auto exec_node(
...
+      } catch (const std::exception& err) {
+        diagnostic::error("{}", err.what())
+          .note("unhandled exception in {} {}", *self, self->state.op->name())
+          .emit(self->state.ctrl->diagnostics());
+        return {};
+      }
+      return diagnostic::error("unhandled exception in {} {}", *self,
+                               self->state.op->name())
+        .to_error();
+    });
```
dominiklohmann, line 765: @Dakostu there really isn't a good way to test this right now, as usually we would consider an uncaught exception a bug. I verified that this works manually by modifying an operator to throw one.

Dakostu (approved):

PR #4178, opened 2024-05-02
Add LEEF parser

mavam:
This PR adds a [LEEF](https://www.ibm.com/docs/en/dsm?topic=overview-leef-event-components) parser.


dominiklohmann (approved):
Nice work! I've played with this for ~30 min and went through the code. Just one small changelog request, and then I think this is ready to be shipped.

PR #4179, opened 2024-05-02
Remove the --clear parameter for lookup table contexts

Dakostu:
This change removes the lookup table-only `--clear` parameter since `context reset` supersedes it and the implementation of that flag is broken anyway.


dominiklohmann (approved):
Thanks for picking this up so quickly!

PR #4180, opened 2024-05-02
Projection pushdown optimization

balavinaithirthan:
Here, we implement projection pushdown, allowing the select operator to get pushed up the pipeline and to allow for optimization in feather and parquet parse. In this PR, we modified every inherited optimize function to include a selection type, we allowed for feather and parquet to read only selected columns, and considered cases where select can be pushed up. This is a good starting point for the summarize and put operators because the framework for moving selection information has been implemented. 
Semantically, an empty columnar selection means that no selection should be performed. Thus, special care must be taken when dealing with a null selection. Here, a null selection acts as a blocker. Further, when returning an optimize result, std::null opt, the selection is blocked. Thus, order_invariant and do_not_optimize are blockers

Currently, only slice, batch, and pass allow the selection to move upstream. More should be added carefully.


libtenzir/src/pipeline.cpp
```diff
@@ -207,14 +217,60 @@ auto pipeline::optimize(expression const& filter, event_order order) const
...
+      }
+      if (opt.selection->do_not_optimize_selection
+          && opt.selection->fields_of_interest) { // is blocking
+        std::string fields_as_string;
+        for (auto&& field : *current_selection.fields_of_interest) {
+          fields_as_string += field + ", ";
+        }
+        fields_as_string.erase(fields_as_string.end() - 2);
+        auto pipe
+          = tql::parse_internal(fmt::format("select {}", fields_as_string));
```
dominiklohmann, line 238:
```suggestion     
        auto pipe
          = tql::parse_internal(fmt::format("select {}", fmt::join(*current_selection.fields_of_interest, ", ")));
```

libtenzir/src/pipeline.cpp
```diff
@@ -207,14 +217,60 @@ auto pipeline::optimize(expression const& filter, event_order order) const
...
+    // order_invarient: should it block optimization? I don't think so, should
+    // pass up original
+    if (opt.selection) {
+      if (opt.selection->selection_finished) {
+        current_selection = columnar_selection::no_columnar_selection();
+      } else {
+        current_selection = std::move(*opt.selection);
+      }
+      if (opt.selection->do_not_optimize_selection
+          && opt.selection->fields_of_interest) { // is blocking
```
dominiklohmann, line 231: nullopt and empty vector currently mean the same thing, and they really shouldn't.

libtenzir/include/tenzir/columnar_selection.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include <caf/none.hpp>
+#include <caf/variant.hpp>
+
+#include <optional>
+#include <string>
+#include <vector>
+
+namespace tenzir {
+
+class columnar_selection {
```
dominiklohmann, line 22: `select_optimization`?

libtenzir/include/tenzir/selection.hpp
```diff
@@ -0,0 +1,44 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2016 The Tenzir Contributors
```
dominiklohmann, line 6: 2024

libtenzir/include/tenzir/selection.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include <caf/default_sum_type_access.hpp>
+#include <caf/detail/type_list.hpp>
+#include <caf/none.hpp>
+#include <caf/variant.hpp>
+
+#include <memory>
+#include <string>
+#include <type_traits>
+#include <vector>
```
dominiklohmann, line 30: Please clean up headers in all newly added files.

libtenzir/include/tenzir/selection.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace tenzir {
+
+class selection {
+public:
+  std::optional<std::vector<std::string>> fields;
```
dominiklohmann, line 36: Please explain the difference between dead and empty states.

libtenzir/include/tenzir/selection.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+  selection() = default;
+  selection(std::optional<std::vector<std::string>> fields)
```
dominiklohmann, line 39:
```suggestion
  explicit selection(std::optional<std::vector<std::string>> fields)
```

Single-argument constructors should always be explicit, and please also follow the rule of five (or zero).

libtenzir/include/tenzir/plugin.hpp
```diff
@@ -385,6 +386,15 @@ class plugin_parser {
     (void)order;
     return nullptr;
   }
+
+  virtual auto optimize(expression const& filter, event_order order,
```
dominiklohmann, line 390: Is the old one above unused now?

libtenzir/include/tenzir/pipeline.hpp
```diff
@@ -317,8 +321,8 @@ class operator_base {
   /// which is implied by `sink <=> OPT | sink`. If `order = schema`, this
   /// resolves to `sink <=> interleave | pass | sink`, which follows from what
   /// we may assume about `sink`.
-  virtual auto optimize(expression const& filter, event_order order) const
```
dominiklohmann, line 320: The comment above this should be expanded to include select optimizations.

libtenzir/src/type.cpp
balavinaithirthan, line 1: I can discard type.cpp changes

libtenzir/include/tenzir/plugin.hpp
```diff
@@ -378,13 +379,20 @@ class plugin_parser {
...
+                        select_optimization const& selection)
+    -> std::unique_ptr<plugin_parser> {
+    (void)filter;
+    (void)selection;
     (void)order;
     return nullptr;
   }
+  bool selection_optimized = false;
+  bool expression_optimized = false;
+  bool order_optimized = false;
```
balavinaithirthan, line 395: I had to add these to distinguish which type of optimization had occurred.

libtenzir/builtins/operators/select.cpp
```diff
@@ -77,10 +77,29 @@ class select_operator final
...
+      std::sort(selection_fields.begin(), selection_fields.end());
+      std::vector<std::string> intersection;
+      std::set_intersection(config_fields.begin(), config_fields.end(),
+                            selection_fields.begin(), selection_fields.end(),
+                            std::back_inserter(intersection));
+      if (intersection.empty()) {
+        return optimize_result{filter, order, copy(), std::nullopt};
+      }
+      return optimize_result{filter, order, nullptr,
+                             select_optimization(intersection)};
```
jachris, line 99: It seems like this assumes that we are free to reorder fields as part of the optimization, right? I would need to be convinced that this is a good choice, as field order is very visible to the user when looking at the output. The more conservative choice would be to require that the optimization does not change the order of fields.

libtenzir/include/tenzir/pipeline.hpp
```diff
@@ -317,7 +318,43 @@ class operator_base {
...
+  /// The selection is a projection pushdown that is represented as a vector of
+  /// columns to select. This vector can be moved around a pipeline similarly to
+  /// an expression. The implementation must ensure that the selected columns
+  /// are preserved in the correct order throughout the pipeline processing
+  /// stages. For an otherwise unknown pipeline `sink`, the following
+  /// equivalences must hold:
+  ///
+  /// ~~~
+  /// this | select columns | sink
+  /// <=> select columns | this | sink
```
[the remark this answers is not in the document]
balavinaithirthan: **Does this sound better (italics is new part)?**


#Projection Pushdown

The selection is a projection pushdown represented as a vector of
columns to select. This vector can be moved around a pipeline similar to an
expression. The implementation must ensure that the selected columns are
preserved in the correct order throughout the pipeline processing stages. For an
otherwise unknown pipeline `sink,` the following equivalences must hold :
  ~~~
  this  | select columns | sink <=> select columns | this | sink
  ~~~

_The input columns will differ from the output columns(i.e., if columns 0 and 4 are
selected, then 1, 2, and 3 remain), but the
individual parse operator must handle this functionality. Optimize allows a modified parser with specified
columns to be returned._

If the operator modifies the schema of the events, the
selection must be adjusted accordingly to ensure that the specified columns are
correctly mapped and preserved in the resulting schema. _The pipeline
implementation can block the upward movement of the select operator by
passing a std::nullopt into the optimized result. Semantically, an empty vector
of selection means no selection was provided, whereas a nullopt selection means
the prior selection is being blocked._

#Example

For an operator that filters and then selects columns,
the optimized output must ensure that the filter is applied before the
selection to maintain the correct semantics :

  ~~~
  where expr | OPT | select columns | sink <==> where expr | select columns |
OPT | sink
  ~~~

  This ensures that filtering and selection are treated as distinct operations,
  and the correct behavior is preserved in the optimized pipeline.

libtenzir/builtins/operators/select.cpp
```diff
@@ -77,10 +77,29 @@ class select_operator final
...
+    if (!selection.fields_of_interest.empty()) {
+      auto config_fields = config_.fields;
+      auto selection_fields = selection.fields_of_interest;
+      std::sort(config_fields.begin(), config_fields.end());
+      std::sort(selection_fields.begin(), selection_fields.end());
+      std::vector<std::string> intersection;
+      std::set_intersection(config_fields.begin(), config_fields.end(),
+                            selection_fields.begin(), selection_fields.end(),
+                            std::back_inserter(intersection));
+      if (intersection.empty()) {
```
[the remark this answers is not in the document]
dominiklohmann: What's your plan for it?

libtenzir/builtins/operators/select.cpp
```diff
@@ -77,10 +77,29 @@ class select_operator final
...
+    }
+    if (!selection.fields_of_interest.empty()) {
+      auto config_fields = config_.fields;
+      auto selection_fields = selection.fields_of_interest;
+      std::sort(config_fields.begin(), config_fields.end());
+      std::sort(selection_fields.begin(), selection_fields.end());
+      std::vector<std::string> intersection;
+      std::set_intersection(config_fields.begin(), config_fields.end(),
+                            selection_fields.begin(), selection_fields.end(),
+                            std::back_inserter(intersection));
```
dominiklohmann, line 94: The intersection works incorrectly when using type extractors. Consider this example:

```
…
| select :ip
| select src_ip
```

You cannot know the type of the field `src_ip` ahead of time. The intersection is only empty when `src_ip` does not exist in the input event, or there are no IP type fields, or when `src_ip` is not of type IP.

I think we need to back to the drawing board to exactly figure out whether we can optimize this at all for TQL1.

libtenzir/builtins/operators/where.cpp
```diff
@@ -6,6 +6,8 @@
 // SPDX-FileCopyrightText: (c) 2021 The Tenzir Contributors
 // SPDX-License-Identifier: BSD-3-Clause
 
+#include "tenzir/select_optimization.hpp"
```
dominiklohmann, line 9: Please include with angle brackets for external library includes (which libtenzir is here, as `where` is a plugin).

libtenzir/include/tenzir/select_optimization.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <caf/variant.hpp>
+
+#include <string>
+#include <vector>
+
+namespace tenzir {
+
+class select_optimization {
+public:
+  std::vector<std::string> fields_of_interest{};
```
dominiklohmann, line 23:
```suggestion
struct select_optimization {
  std::vector<std::string> fields_of_interest{};
```

This expresses more clearly to the reader that all members are public intentionally.

libtenzir/include/tenzir/select_optimization.hpp
```diff
@@ -0,0 +1,41 @@
...
+  select_optimization() = default;
+  explicit select_optimization(std::vector<std::string> fields_of_interest)
+    : fields_of_interest{std::move(fields_of_interest)} {
+  }
```
dominiklohmann, line 27: I don't think these are needed.

libtenzir/include/tenzir/select_optimization.hpp
dominiklohmann, line 23: Should this just be named `fields`?

libtenzir/include/tenzir/select_optimization.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include <caf/detail/type_list.hpp>
+#include <caf/none.hpp>
+#include <caf/variant.hpp>
+
+#include <string>
+#include <vector>
+
+namespace tenzir {
+
+class select_optimization {
```
dominiklohmann, line 21: Please add a comment explaining what this is for and how it is used.

libtenzir/include/tenzir/select_optimization.hpp
```diff
@@ -0,0 +1,41 @@
...
+//
+// SPDX-FileCopyrightText: (c) 2024 The Tenzir Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#pragma once
+
+#include <caf/default_sum_type_access.hpp>
+#include <caf/detail/type_list.hpp>
+#include <caf/none.hpp>
+#include <caf/variant.hpp>
```
dominiklohmann, line 14: These seem to be unused.

libtenzir/include/tenzir/select_optimization.hpp
```diff
@@ -0,0 +1,41 @@
...
+    return select_optimization(std::vector<std::string>{});
+  }
+};
+
+template <class Inspector>
+auto inspect(Inspector& f, select_optimization& x) {
+  return f.object(x)
+    .pretty_name("expression")
+    .fields(f.field("node", x.fields_of_interest));
+}
```
dominiklohmann, line 39: Please make this a friend function of the struct to make it a bit more ADL-friendly.

```suggestion
friend auto inspect(auto& f, select_optimization& x) -> bool {
  return f.object(x)
    .fields(f.field("fields", x.fields_of_interest));
}
```

libtenzir/builtins/operators/from_load_read.cpp
```diff
@@ -82,22 +85,24 @@ class read_operator final : public crtp_operator<read_operator> {
...
+  auto optimize(expression const& filter, event_order order,
+                select_optimization const& selection) const
     -> optimize_result override {
     (void)filter;
-    if (order == event_order::ordered) {
+    auto parser_opt = parser_->optimize(filter, order, selection);
+    if (!parser_opt.replacement) {
       return do_not_optimize(*this);
     }
-    // TODO: We could also propagate `where #schema == "..."` to the parser.
```
dominiklohmann, line 91: You removed this comment, but as far as I can tell you didn't actually address it.

libtenzir/builtins/operators/from_load_read.cpp
```diff
@@ -82,22 +85,24 @@ class read_operator final : public crtp_operator<read_operator> {
...
   }
 
-  auto optimize(expression const& filter, event_order order) const
+  auto optimize(expression const& filter, event_order order,
+                select_optimization const& selection) const
     -> optimize_result override {
     (void)filter;
-    if (order == event_order::ordered) {
+    auto parser_opt = parser_->optimize(filter, order, selection);
+    if (!parser_opt.replacement) {
```
dominiklohmann, line 93: What happens when `parser_opt.selection_optimized` is true and `parser_opt.replacement` is false?

libtenzir/builtins/operators/from_load_read.cpp
```diff
@@ -6,6 +6,8 @@
 // SPDX-FileCopyrightText: (c) 2023 The Tenzir Contributors
 // SPDX-License-Identifier: BSD-3-Clause
 
+#include "tenzir/pipeline.hpp"
```
dominiklohmann, line 9:
```suggestion
#include <tenzir/pipeline.hpp>
```

libtenzir/builtins/formats/json.cpp
```diff
@@ -1193,10 +1193,15 @@ class json_parser final : public plugin_parser {
...
+  auto optimize(expression const& filter, event_order order,
+                select_optimization const& selection)
+    -> optimize_parser_result override {
+    (void)filter;
+    (void)selection;
     auto args = args_;
     args.preserve_order = order == event_order::ordered;
-    return std::make_unique<json_parser>(std::move(args));
+    return {std::make_unique<json_parser>(std::move(args)), false, false,
+            false};
```
dominiklohmann, line 1204: As far as I understand, one of the `false, false, false` indicates that the order was not optimized. But it was optimized right above this line.

libtenzir/builtins/formats/feather.cpp
```diff
@@ -285,22 +287,108 @@ class callback_listener : public arrow::ipc::Listener {
...
+  auto stream_decoder = arrow::ipc::StreamDecoder(listener, read_options);
+  for (auto&& buffer : buffered_schema_data) {
+    auto decode_result = stream_decoder.Consume(buffer);
+    if (!decode_result.ok()) {
+      diagnostic::error("{}", decode_result.ToStringWithoutContextLines())
+        .note("failed to decode the byte stream into a record batch")
+        .emit(ctrl.diagnostics());
+      co_return;
+    }
+  }
```
dominiklohmann, line 391: This code is almost exactly duplicated from right below. Can we make this easier to maintain?

libtenzir/src/pipeline.cpp
```diff
@@ -177,24 +177,38 @@ auto pipeline::optimize_into_filter() const -> std::pair<expression, pipeline> {
...
+      current_selection = *opt.selection;
+    } else if (!current_selection.fields_of_interest.empty()) {
+      auto pipe = tql::parse_internal(fmt::format(
+        "select {}", fmt::join(current_selection.fields_of_interest, ", ")));
+      TENZIR_ASSERT(pipe);
+      auto ops = std::move(*pipe).unwrap();
+      TENZIR_ASSERT(ops.size() == 1);
+      result.push_back(std::move(ops[0]));
+      current_selection = select_optimization::no_select_optimization();
+    }
```
dominiklohmann, line 211: Did you consider an alternative implementation where there is no replacement operator, i.e., `select` never gets elided and also never added back in, because generally `select` is a cheap operator? We could still push up the information so that it could become a no-op in optimized cases.

Not sure if this is worth it, I think you can judge the impact of that best.

dominiklohmann (commented):
Thanks for this massive undertaking. This is a great piece of work! 🚀

I went through everything in code once now and have left a bunch of comments that need addressing. I think it's only minor things from here on out.

libtenzir/builtins/operators/select.cpp
jachris: `event_order` is about the order of _events,_ not _fields_. I was asking about the field order. IIUC, your code does not guarantee that the field order is the same before and after optimization.

balavinaithirthan: Ya, I see the issue. We want to keep the ordering of the fields the same:
ie `select foo bar baz | select bar foo baz` should be `foo baz` and not `baz foo`

libtenzir/builtins/operators/select.cpp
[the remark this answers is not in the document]
balavinaithirthan: The current plan is to not do any optimization if it is an empty intersection. I could also return a warning.

libtenzir/include/tenzir/pipeline.hpp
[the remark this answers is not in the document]
jachris: It would be best if you could write the exact requirements down succinctly. It might be easiest to do this semi-formally, as we have done for predicate pushdown.

Let me try my luck, and you tell me whether what I wrote captures what you are doing. I did not take a close look at the implementation yet, so what I write might also be incorrect. Just let me know. I will limit myself to the selection optimization. The full interaction between predicate, selection and ordering optimizations also needs to be described somewhere, but this is a starting point (please do not use it directly).

---

If `opt.selection` has a value (where `opt` is the returned `optimize_result`), it must hold that

```
this | select selection | sink 
<=> select opt.selection | opt.replacement | sink
``` 

Otherwise: 

```
this | select columns | sink 
<=> opt.replacement | select columns | sink
```

If `opt.replacement` is `nullptr`, then we interpret it as `pass` above.  

---

Is this correct? 

If so, can you extend this to describe how it interacts with the predicate optimization? In particular, I am interested the order of `select` and `where` in the requirements.

libtenzir/builtins/operators/select.cpp
jachris: That's right. And `select a, b | select b, a` needs to be the same as `select b, a`.

libtenzir/builtins/operators/select.cpp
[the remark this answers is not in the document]
jachris: If you write down the exact requirements for the non-empty case (see other conversation), it'll probably tell you what to do in the empty case.

libtenzir/builtins/operators/select.cpp
jachris: Actually, I spoke to @dominiklohmann. We want to proceed under the assumption that `select` reorders (even though it does not atm), to make this future proof.

libtenzir/builtins/operators/from_load_read.cpp
balavinaithirthan: This was an old comment from before I started. Should I add the comment back.

dominiklohmann: You removed the TODO. That would indicate that it either doesn't apply anymore or that you addressed it. But it very much still applies, and you didn't address it.

libtenzir/builtins/operators/from_load_read.cpp
balavinaithirthan: This shouldn't happen because the parser (feather, parquet etc.) shouldn't know the selection information before the `optimize` function. Thus, if there is no replacement, the selection information was never used and the `parser` class is the same.

dominiklohmann: I see. The problem is then that the return value of `parser_->optimize(...)` does not encode this invariant as part of its data structure. It is rather easy to write an optimize function for a parser with the current design that breaks your invariant. It really shouldn't be possible to do that in the first place.

libtenzir/src/pipeline.cpp
balavinaithirthan: I see, so the idea would be to keep the `select` in place in addition to pushing up a copy of the `select`. I can see the benefit for code accuracy since rejecting optimization means the select stays in place, but I can see this proving difficult when combining `select` + `select` or `select` + `filter`. I don't think it is worth the effort especially since the `do_not_optimize` does this simply enough.

libtenzir/builtins/operators/select.cpp
balavinaithirthan: Considering the select | select logic needs more discussion, should I not combine selections for now (stop the later selections from getting pushed up).

libtenzir/builtins/formats/feather.cpp
dominiklohmann: I tried for a few minutes, but I couldn't come up with something better. So maybe just leave it as-is, as annoying as that is.

libtenzir/include/tenzir/pipeline.hpp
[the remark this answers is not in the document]
balavinaithirthan: Thoughts?

If `opt.selection` has a value (where `opt` is the returned `optimize_result`), it must hold that

```
this | select columns | sink 
<=> select opt.selection | this | sink
``` 

If `opt.selection` has no value and the `current_selection`, representing the last select seen call, is not empty, it must hold that
```
this | sink 
<=> this | select columns | sink
```
Here, the previous `select` is inserted as a block. In all other cases where`opt.selection` has no value, it must hold that no `select` operator is being pushed up the pipeline. 

---

If `opt.filter` and `opt.selection` can never contain values simultaneously. They must push up their optimization information, or the filter/selection information will never be recorded.

Note: An alternative to not allowing both filter and select to be pushed together up the pipeline is to force this combination to have one semantic meaning, i.e., `opt.filter` and `opt.selection` means `select | this | where filter | sink`. 

---

`opt.replacement` operates in isolation from `opt.selection ` and `opt.filter`. For example, in the do_not_optimize() function, opt.replacement() is a copy of the original operator.

```
this | select selection | sink 
<=> select opt.selection | opt.replacement | sink
``` 

If `opt.replacement` is `nullptr`, then it must hold that no changes were made to the original operator in the pipeline.

jachris:
Here is a quick rundown of what I think the actual requirements should be, including all optimizations (ordering, replacement, predicate pushdown, and selection pushdown). This probably need some further elaboration and also needs to be checked again as to whether this is actually what we want.

---

For every pipeline `sink` and result `opt` from the call to `this->optimize(filter, order, selection)`, we define
```python
if order == ordered:
  SINK = sink 
elif unordered:
  SINK = shuffle | sink
elif order == schema:
  SINK = interleave | sink

if opt.order == ordered:
  OPT = opt.replacement
elif opt.order == schema:
  OPT = interleave | opt.replacement
elif opt.order == unordered:
  OPT = shuffle | opt.replacement
```
where `opt.replacement == nullptr` means `pass`. The implementation of `optimize` must then guarantee that `this | where filter | select selection | SINK` is equivalent to:
```python
if opt.filter and opt.selection:
  where opt.filter | select opt.selection | OPT | SINK
if opt.filter and not opt.selection:
  where opt.filter | OPT | select selection | SINK
if not opt.filter and opt.selection:
  select opt.selection | OPT | where filter | SINK
if not opt.filter and not opt.selection:
  OPT | where filter | select selection | SINK
```

dominiklohmann:
Closing this for now. @jachris will re-implement this for TQL2. Thanks for doing the foundational work in this already, @balavinaithirthan, that will make Jannis' job much easier.

PR #4181, opened 2024-05-02
Add the `show index` aspect

dominiklohmann:
This is temporary and for debugging purposes in a production setup, so I don't think this needs documentation. This will be removed again.

### DIFF
--- a/libtenzir/builtins/aspects/index.cpp
+++ b/libtenzir/builtins/aspects/index.cpp
@@ -1,0 +1,67 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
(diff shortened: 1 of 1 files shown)


libtenzir/builtins/aspects/index.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+namespace tenzir::plugins::index {
+
+namespace {
+
+class plugin final : public virtual aspect_plugin {
+public:
+  auto name() const -> std::string override {
+    return "index";
+  }
```
tobim, line 25: Nit: The name `index` is a bit of a lie, in reality this would be `index-aspect` or `show-index` and have a separate `aspect_name`.

tobim (approved):

libtenzir/builtins/aspects/index.cpp
dominiklohmann: We do the same for all the other plugins, where we only use a plugin name that's different from the operator/format/connector/aspect/component name when it would otherwise clash. It's easy to change then. Let's just keep this as-is now.

PR #4183, opened 2024-05-02
Bump nixpkgs to 94035b482

tobim:
This bumps us to Arrow 16.0.0 and spdlog 3.9.1.


mavam:
@tobim looks like your nushell script broke with an update.

jachris (approved):

PR #4185, opened 2024-05-03
Add unordered batch optimization

IyeOnline:
The batch operator now has an unordered mode, where it keeps a separate buffer for each schema. 

The import operator now implicitly batches.

Fixes https://github.com/tenzir/issues/issues/1792


libtenzir/builtins/operators/batch.cpp
```diff
@@ -6,6 +6,8 @@
 // SPDX-FileCopyrightText: (c) 2023 The Tenzir Contributors
 // SPDX-License-Identifier: BSD-3-Clause
 
+#include "tenzir/detail/assert.hpp"
+
```
dominiklohmann, line 10: This seems like an IWYU leftover, it should already be included many times over transitively.

dominiklohmann (approved):
Welcome to Tenzir! This is a great first contribution. 🚀

libtenzir/builtins/operators/batch.cpp
```diff
@@ -21,63 +23,88 @@ namespace tenzir::plugins::batch {
...
-        num_buffered = 0;
+      // Check all current buffers to see if we have hit a timeout.
+      for (auto it = buffers.begin(); it != buffers.end(); ++it) {
+        auto& entry = it->second;
+        if (now - entry.last_yield > timeout_ and entry.num_buffered > 0) {
+          TENZIR_ASSERT(entry.num_buffered < limit_);
+          entry.last_yield = now;
+          co_yield concatenate(std::exchange(entry.events, {}));
+          it = buffers.erase(it);
+          entry.num_buffered = 0;
```
dominiklohmann, line 55: Entry may be invalidated here.

dominiklohmann (approved):

PR #4189, opened 2024-05-06
Make python venv creation independent from the user

tobim:
A permission error caused python operator to fail when it was previously used by another system user with the same set of requirements.


libtenzir/builtins/operators/python.cpp
dominiklohmann, line 1: Please add a changelog entry for the bug fix.

dominiklohmann (dismissed):
Nice catch! 👀

tobim:
I did quite a bit of testing on this and unfortunately the change as proposed so far does not work. The problem is that semaphore creation is subject to the processes `umask`, which defaults to 022 on most systems. The naive approach would be to temporarily set the umask to 000, but that that is highly unsafe in a multithreaded program.

A complete fix should still be possible by wrapping all the venv creation steps in an extra python subprocess (which would be single threaded) and use that to manage the inter process synchronization. This would have the nice side effect of securing the sync against SIGKILLs, SEGFAULTS, and the like.

dominiklohmann:
There's also the option of creating separate cache directories per user.

libtenzir/builtins/operators/python.cpp
```diff
@@ -162,14 +163,16 @@ class python_operator final : public crtp_operator<python_operator> {
...
         // in terms of wait time. As a workaround we yield early, so that the
         // pipeline appears as created and do the actual venv setup on first
         // call. At that point the delay is not problematic any more because
         // `/serve` takes care of it gracefully.
         co_yield {};
+        auto permissions = boost::interprocess::permissions{0666};
         auto sem = boost::interprocess::named_semaphore{
-          boost::interprocess::open_or_create, sem_name.c_str(), 1u};
+          boost::interprocess::open_or_create, sem_name.c_str(), 1u,
+          permissions};
```
dominiklohmann, line 175: I think Tobias' commit should be reverted again as that doesn't work.

changelog/next/bug-fixes/4189--python-permission-issues.md
```diff
@@ -0,0 +1,3 @@
+A permission error caused `python` operator to fail when it was previously used
+by another system user with the same set of requirements. There now is a one
+python environment per user and set of requirements.
```
dominiklohmann, line 3:
```suggestion
Python environment per user and set of requirements.
```

dominiklohmann (approved):
Thanks for the fix!

PR #4194, opened 2024-05-08
Fix a panic on shutdown

dominiklohmann:
I recently introduced this alongside the component lifetime logging. For some old components, we do still use the labels that I thought we didn't use anymore. Ignoring them led to the components being tracked wrongly and the assertions `num_erased == 1` right below the diff to be triggered.


IyeOnline (approved):
Looks good!

PR #4195, opened 2024-05-08
Fix CSV/XSV format printing the header once for each batch

jachris:
The XSV format printed the header once for each batch. This PR makes it so that the header is printed only once.

The actual diff can be seen in the second commit.


mavam (approved):
Might be worth adding a changelog entry, otherwise LGTM.

PR #4196, opened 2024-05-08
Allow editing definitions of managed pipelines

Dakostu:
This is another submodule bump with an API docs update.


tobim (approved):
We pair reviewed and improved the submodule changes.

PR #4197, opened 2024-05-08
Undocument fictive schema syntax

mavam:
This effort never materialized.


dominiklohmann (approved):
I suppose not having it documented is better than having the wrong thing documented.

PR #4198, opened 2024-05-08
Make tests independent of Accept-Encoding header

mavam:
This PR makes the HTTP integration tests independent of the value of the `Accept-Encoding` header.


mavam:
@tobim can you try whether that fixes the issue with your Nix setup?

dominiklohmann (approved):

PR #4200, opened 2024-05-10
Fix the lifetime tracking of node components

dominiklohmann:
The old code had a race condition that sometimes triggered an assertion failure in CI. Now the tracking is properly implemented through monitoring, which should fix the issue we've been seeing.


IyeOnline (approved):
I'll just trust you on this one :)

PR #4203, opened 2024-05-10
Add an option to `export` both past and future events

IyeOnline:
* The `export` operators now features a `--retro`
flag. This flag will make the operators export past events, even when `--live`
is set: 
  * Specifying *both* `--retro --live` will first export all past events and
  then go into live mode.
  * Specifying *only* `--live`  behaves like it did previously. It only exports
  new events.
  * Specifying *only* `--retro` behaves as if the flag was not set, just like
  previously without any flag.
  * Specifying none of the flags is also unchanged.

  **The combination of both flags is experimental**. It has an obvious issue with events that happen live while exporting previous events.

* The `diagnostics` and `metrics` operators also feature these flags, with the
same behavior.

* A one batch delay in `export --live` has been fixed.


changelog/next/features/4203--export-past-and-live.md
dominiklohmann, line 1: This won't render correctly. All paragraphs in the individual changelog entry files get merged into a single entry, so they need to be text rather than notes.

dominiklohmann (approved):
I think this can and should be merged—even with the gap in the data that we currently have between the two modes this still adds quite some value.

changelog/next/bug-fixes/4203--export delay.md
dominiklohmann, line 1: Please write this in a bit more user-facing way, for example:

> `export --live` no longer buffers the last batch of event that was imported, and instead immediately returns all imported events.

PR #4206, opened 2024-05-10
Adjust to breaking change in Google's GitHub Actions

dominiklohmann:
See googleapis/release-please-action#980


dominiklohmann:
Nvm, they didn't actually move all—they just moved one.

PR #4207, opened 2024-05-11
Remove the old format API

dominiklohmann:
Now that `tenzir-ctl import` and `tenzir-ctl export` are gone, we can finally remove the old reader and writer APIs. This removes a bunch of code and tests of effectively dead code.


mavam (approved):
Glorious.

PR #4208, opened 2024-05-11
Add a `mean` aggregation function

dominiklohmann:
I needed this for an ingress/egress chart and was surprised we didn't have that yet. This was easy enough to implement, so I just went ahead and did it.


jachris (approved):

mavam:
Could you also add something for median and standard deviation, assuming it's as easy? They often go together.

dominiklohmann:
> Could you also add something for median and standard deviation, assuming it's as easy? They often go together.

Bit harder to do, but certainly not a problem. If I find some more time to hack on code this weekend I'll consider it, otherwise feel free to just use `mean` as a template.

dominiklohmann:
> Could you also add something for median and standard deviation, assuming it's as easy? They often go together.

Done. Mind giving this another review given that there were major changes since Jannis' initial approval?

changelog/next/features/4208--more-aggregation-functions.md
```diff
@@ -0,0 +1,11 @@
+The new `mean` aggregation function computes the mean of grouped numeric values.
+
+The new `approximate_median` aggregation function computes an approximate median
+of grouped numeric values using a T-Digest algorithm.
```
mavam, line 4:
```suggestion
The new `approximate_median` aggregation function computes an approximate median
of grouped numeric values using a t-digest algorithm.
```

libtenzir/builtins/aggregation-functions/approximate_median.cpp
```diff
@@ -0,0 +1,114 @@
...
+    }
+    return data{static_cast<type_to_data_t<Type>>(tdigest_.Quantile(0.5))};
+  }
+
+  arrow::internal::TDigest tdigest_;
+};
+
+class plugin : public virtual aggregation_function_plugin {
+  auto name() const -> std::string override {
+    return "approximate_median";
```
mavam, line 72:
```suggestion
    return "approx_median";
```

Anything longer than that is going to be painful to write.

libtenzir/builtins/aggregation-functions/collect.cpp
```diff
@@ -0,0 +1,79 @@
...
+private:
+  auto output_type() const -> type override {
+    return type{list_type{input_type()}};
+  }
+
+  auto add(const data_view& view) -> void override {
+    using view_type = tenzir::view<type_to_data_t<Type>>;
+    if (caf::holds_alternative<caf::none_t>(view)) {
+      return;
+    }
```
mavam, line 33: We may want to make treating nulls a runtime option, but we can also wait until the first user requests that.

libtenzir/builtins/aggregation-functions/mean.cpp
```diff
@@ -0,0 +1,111 @@
...
+      return;
+    }
+    const auto x = static_cast<double>(caf::get<view_type>(view));
+    if constexpr (std::is_same_v<Type, double_type>) {
+      if (std::isnan(x)) {
+        return;
+      }
+    }
+    count_ += 1;
+    mean_ += (x - mean_) / count_;
```
mavam, line 41: Is this done typically like that? I could imagine that tracking the sum has overflow issues, which is why you resorted to this approach.

libtenzir/builtins/aggregation-functions/stddev_variance.cpp
```diff
@@ -0,0 +1,126 @@
...
+      return data{};
+    }
+    const auto variance = mean_squared_ - (mean_ * mean_);
+    return data{mode_ == mode::stddev ? std::sqrt(variance) : variance};
+  }
+
+  double mean_ = {};
+  double mean_squared_ = {};
+  size_t count_ = {};
+  mode mode_ = {};
```
mavam, line 79: Why is this not a template parameter to the class?

mavam (approved):
Great work!

Approving modulo missing integration tests.

I'm a bit worried about the error bounds of using t-digest for incremental median computation, but having even an approximate value is already very helpful. IIRC the error bounds do not hold upon when composing t-digest sketches, but that's not an issue in our case.

libtenzir/builtins/aggregation-functions/approximate_median.cpp
dominiklohmann: I've chosen to use the same names used by Pandas, Polars, and Arrow. I'm reluctant to change this.

libtenzir/builtins/aggregation-functions/stddev_variance.cpp
dominiklohmann: It's looked at exactly once, so it doesn't matter from a runtime performance perspective. Let's avoid bloating our binary where we can.

libtenzir/builtins/aggregation-functions/approximate_median.cpp
mavam: Good reason.

libtenzir/builtins/aggregation-functions/mean.cpp
dominiklohmann: That's just an implementation of Welford's algorithm. It's what you'd typically use to avoid overflows.

libtenzir/builtins/aggregation-functions/collect.cpp
dominiklohmann: I don't think it'd be particularly useful tbh, so let's definitely wait until someone asks for it.

dominiklohmann:
> Approving modulo missing integration tests.

I'll add some before merging. For now I mostly verified that this works as expected by manually comparing values. That's why I implemented `collect` as well.

PR #4209, opened 2024-05-11
Add timeout options to `summarize`

dominiklohmann:
This adds two new options to the `summarize` operator:
- `timeout <duration>` specifies the maximum lifetime of a bucket from the time it was created at.
- `update-timeout <duration>` specifies the maximum lifetime of a bucket from the time it was last updated.

These options basically enable streaming aggregations. The `timeout` option is useful to guarantee that events are held back no more than the specified duration. The `update-timeout` is useful when operating on a stream of events where it's known that events that would be sorted into the same group arrive at near the same time.


mavam:
Nice, even though `group` will supersede this in TQL2. I hope this is not distracting us too much.

dominiklohmann:
> Nice, even though `group` will supersede this in TQL2. I hope this is not distracting us too much.

Don't think it's distracting in any way, as we'll just delete this implementation of `summarize` when the `group` operator is ready.

libtenzir/builtins/operators/summarize.cpp
```diff
@@ -788,6 +845,13 @@ class summarize_operator final : public crtp_operator<summarize_operator> {
     return "summarize";
   }
 
+  auto input_independent() const -> bool override {
+    if (config_.created_timeout or config_.update_timeout) {
+      return true;
+    }
+    return false;
+  }
```
jachris, line 853:
```suggestion
  auto input_independent() const -> bool override {
    return config_.created_timeout || config_.update_timeout;
  }
```

Just because I have nothing else to say (:

jachris (approved):
LGTM (temporary syntax anyway)

PR #4211, opened 2024-05-12
Change the syntax of the `slice` operator

dominiklohmann:
This changes the `slice` operator to use a more familiar syntax for slicing events: `<begin>:<end>`, `<begin>:`, or `:<end>`.

This makes room for adding strides to the operator in the famililar Python syntax.


jachris (approved):
Not sure whether the syntax stays this way, but this is good for now :+1:

PR #4212, opened 2024-05-12
Treat AMQP library exceptions as actual errors

satta:
This ensures that the `amqp` operator actually fails in cases such as network disconnects, which otherwise would not cause pipelines with amqp connectors to restart on failure.

### DIFF
--- a/plugins/amqp/src/plugin.cpp
+++ b/plugins/amqp/src/plugin.cpp
@@ -350,8 +350,15 @@ public:
     if (ret.library_error == AMQP_STATUS_TIMEOUT)
       return chunk_ptr{};
     // Now we're leaving the happy path.
-    if (ret.reply_type == AMQP_RESPONSE_LIBRARY_EXCEPTION
-        && ret.library_error == AMQP_STATUS_UNEXPECTED_STATE) {
+    TENZIR_DEBUG("reply type is {}, library error {} ({})", ret.reply_type,
+                 ret.library_error, amqp_error_string2(ret.library_error));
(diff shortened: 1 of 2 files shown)


plugins/amqp/src/plugin.cpp
```diff
@@ -350,8 +350,16 @@ class amqp_engine {
     if (ret.library_error == AMQP_STATUS_TIMEOUT)
       return chunk_ptr{};
     // Now we're leaving the happy path.
-    if (ret.reply_type == AMQP_RESPONSE_LIBRARY_EXCEPTION
-        && ret.library_error == AMQP_STATUS_UNEXPECTED_STATE) {
+    TENZIR_DEBUG("leaving the AMQP happy path...");
```
dominiklohmann, line 353: I don't think this log message adds any value over the one right below. Can you remove it?

plugins/amqp/src/plugin.cpp
dominiklohmann, line 1: Please add a changelog entry for the bug fix in `changelog/next/bug-fixes/4212-satta.md` that contains a single paragraph of Markdown describing in active voice and present tense what you fixed.

dominiklohmann (approved):

plugins/amqp/src/plugin.cpp
satta: Sure. That was a leftover from @mavam's original

plugins/amqp/src/plugin.cpp
satta: Will do.

PR #4213, opened 2024-05-12
Fix rendering of AMQP error code

dominiklohmann:
This fixes an oversight from tenzir/tenzir#4112 that caused the Nix build to fail.

### DIFF
--- a/plugins/amqp/src/plugin.cpp
+++ b/plugins/amqp/src/plugin.cpp
@@ -350,8 +350,12 @@ public:
     if (ret.library_error == AMQP_STATUS_TIMEOUT)
       return chunk_ptr{};
     // Now we're leaving the happy path.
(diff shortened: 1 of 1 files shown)


mavam (approved):

plugins/amqp/src/plugin.cpp
```diff
@@ -351,7 +351,9 @@ class amqp_engine {
       return chunk_ptr{};
     // Now we're leaving the happy path.
     TENZIR_DEBUG("reply type is {}, library error {} ({})", ret.reply_type,
-                 ret.library_error, amqp_error_string2(ret.library_error));
+                 static_cast<std::underlying_type_t<amqp_response_type_enum>>(
+                   ret.library_error),
+                 amqp_error_string2(ret.library_error));
```
mavam, line 356: There's also a `to_error` utility function defined above.

PR #4214, opened 2024-05-13
Remove stream managers when decommissioning partitions

dominiklohmann:
We noticed that in some deployments, the CPU usage of the `tenzir.index` thread would rise over time. Most time was spent in `std::_Rb_tree_increment`—which is an internal of `std::map` or `std::set`. But we didn't use that anywhere in the index actor, or so we thought: it turned out that CAF itself uses a `std::map` for its stream managers, creating a new one for every active partition. When decommissioning an active partition into a passive partition, that entry in the map was not removed, making further interactions with the map increasingly expensive.

This change makes it so we manually erase stale stream managers when decommissioning an active partition. This really should be considered a bug in CAF, but since upstream has replaced the streaming functionality entirely it doesn't make much sense to report or fix it there.


mavam (approved):
YOLO-approving this one.

libtenzir/src/index.cpp
```diff
@@ -1371,6 +1372,10 @@ index(index_actor::stateful_pointer<index_state> self,
     // Flush buffered batches and end stream.
     self->state.stage->stop();
   });
+  detail::weak_run_delayed_loop(self, std::chrono::seconds{10}, [self] {
+    TENZIR_VERBOSE("{} has {} active stream managers", *self,
+                   self->stream_managers().size());
+  });
```
tobim, line 1378:
```suggestion
```

By itself, this message provides too little information to aid understanding. I'd say it's rather adding more confusion for everybody that was not present during our conversation.
What would help more is printing all `stream_managers` that do not belong to the list of current active partitions if there are any. But since we know that we are the only ones creating those outbound stream slots I wouldn't bother.

tobim (approved):
I double checked that the outbound stream slot that we receive when we add the path for an active partition is internally created by emplacing into the same stream manager map.

PR #4222, opened 2024-05-15
Remove the superfluous path separators when using S3 or GS connectors

Dakostu:
Arrow 16 tightened their URI parsing: Double slashes for post-bucket path separators are now forbidden. We previously generated S3 and GCSs paths out of a parsed `arrow::util::Uri`, using a `"{}/{}", uri.host(), uri.path()"` format string to compensate for missing slashes. This exposed an unnecessary additional slash that this change removes.

See this Arrow PR diff: https://github.com/apache/arrow/commit/0148db4f97bfcf80237b265ee0d9a4b61194ffde


lava (approved):
Nice fix before I even finished reproducing the issue :D

PR #4224, opened 2024-05-15
Handle loading of configured and non-configured contexts with the same name

Dakostu:
This change adds order to the chaos of loading contexts.
Configured contexts will be loaded first, and non-configured contexts that happen to share the same name will be renamed to avoid non-deterministic loading conflicts.

### DIFF
--- a/changelog/next/bug-fixes/4224--configured-context-loading-precedence.md
+++ b/changelog/next/bug-fixes/4224--configured-context-loading-precedence.md
@@ -1,0 +1,4 @@
+Configured and non-configured contexts with the same name will not cause
+non-deterministic behavior upon loading anymore. Non-configured contexts that
+share the same name with configured contexts will have a random suffix added to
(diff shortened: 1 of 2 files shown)


dominiklohmann (approved):
Thanks for fixing this!

contrib/tenzir-plugins
dominiklohmann, line 1: There's a use-after-free in this that still needs to be addressed (which should be trivial to address), but the general approach is definitely the right one. Approving already.

dominiklohmann: @Dakostu can you rebase this PR so that the CI runs?

Dakostu: Done.

tobim, line 1: We went over the submodule changes and all comments are already addressed.

tobim (approved):

PR #4227, opened 2024-05-17
Wipe the cache directory on node restart

lava:
Currently the cache directory is never cleared and can grow without bounds. Additionally, after node updates it can contain stale entries produced by older node versions.

As a brutal yet straightforward workaround, remove the cache directory on every node restart.


lava:
Closing this in favor of a more detailed discussion in https://github.com/tenzir/issues/issues/1828

PR #4237, opened 2024-05-21
Remove unused functionality from the catalog

dominiklohmann:


web/docs/operators/show.md
```diff
@@ -17,7 +17,7 @@ show [<aspect>]
 ## Description
 
 The `show` operator offers introspection capabilities to look at various
-*aspects* of a Tenzir node.
+_aspects_ of a Tenzir node.
```
mavam, line 20: Since when do we use underscores for emphasis?

mavam (approved):

web/docs/operators/show.md
dominiklohmann: We don't, I think this was a byproduct of some config changes I made following the neovim 0.10 update.

PR #4238, opened 2024-05-22
Increase maximum JSON object size to 10MB

jachris:
This PR increases the maximum size of a single object/event when *not* using the NDJSON mode (which does not have such a restriction). If a fixed maximum becomes problematic in the future, we can use a dynamic approach instead, but in the meantime, this should work with most documents.

### DIFF
--- a/libtenzir/builtins/formats/json.cpp
+++ b/libtenzir/builtins/formats/json.cpp
@@ -39,6 +39,11 @@ namespace tenzir::plugins::json {
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


libtenzir/builtins/formats/json.cpp
jachris, line 1: Not sure whether this needs a changelog as it just tweaks some numbers.

dominiklohmann: Don't think it does.

dominiklohmann (approved):

PR #4239, opened 2024-05-22
Make TQLv2 executable

jachris:
Following up on #3991, this PR adds everything that is needed to execute TQLv2. All features are gated behind `--tql2`, thus no changelog is necessary.

It contains the following new functionality:

- Name resolution and binding of operators, functions and methods.
- A prototype of `let` binding substitution.
- Compilation of `ast::pipeline` objects to `pipeline`.
- A small collection of loaders, parsers, printers and savers.
- A base library of 25 operators, 16 functions and 6 methods.
- A basic implementation of `fork { ... }` and `if { ... } else { ... }`.
- A revamped `summarize` operator with implicit grouping.
- An evaluator for actual expressions, making TQLv2 much more powerful than TQLv1.
- Various utilities to implement all the above.

 This PR does not yet include the integration for the web app, documentation or an actual test suite.


libtenzir/builtins/formats/json.cpp
```diff
@@ -1494,6 +1498,139 @@ class selector_parser final : public virtual parser_parser_plugin {
...
+
+class write_json_plugin final
+  : public virtual operator_inspection_plugin<write_json>,
+    public virtual operator_factory_plugin {
+public:
+  auto make_operator(invocation inv, session ctx) const
+    -> operator_ptr override {
+    // TODO
+    auto args = printer_args{};
+    argument_parser2{"https://docs.tenizr.com/operators/write_json"}.parse(inv,
```
mavam, line 1628:
```suggestion
    argument_parser2{"https://docs.tenzir.com/operators/write_json"}.parse(inv,
```

libtenzir/src/tql2/parser.cpp
```diff
@@ -434,19 +434,34 @@ class parser {
...
       auto ident = expect(tk::identifier).as_identifier();
       auto kind = static_cast<enum meta_extractor::kind>(0);
       if (ident.name == "tag") {
         kind = meta_extractor::schema;
+      } else if (ident.name == "import_time") {
+        kind = meta_extractor::import_time;
+      } else if (ident.name == "internal") {
+        kind = meta_extractor::internal;
+      } else if (ident.name == "schema") {
+        diagnostic::error("use `{}tag` instead", is_at ? "@" : "meta.")
```
mavam, line 454: I like `@` a lot as symbol. How is `meta.` not ambiguous with actual values in the data?

jachris: I implemented both for now, so that we can play around with it and see what's better.

It's not ambiguous in the same way that `this`/`self`/`if`/`match`/any other keyword is not ambiguous: You are just not allowed to use that syntax if your field name is a keyword. You'd use `this["if"]` as a fallback.

mavam: > You are just not allowed to use that syntax

So I'd get an error if I use `meta.timestamp` and a field with name `meta` exists also?

libtenzir/src/tql2/eval_binary.cpp
```diff
@@ -27,25 +27,34 @@ namespace tenzir {
 namespace {
 
 [[maybe_unused]] constexpr auto is_arithmetic(ast::binary_op op) -> bool {
+  using enum ast::binary_op;
   switch (op) {
-    using enum ast::binary_op;
     case add:
     case sub:
     case mul:
     case div:
```
mavam, line 35: Is there a case for `mod` as well? I could imagine that it's useful for load-balancing.

libtenzir/builtins/functions/string.cpp
```diff
@@ -0,0 +1,91 @@
...
+public:
+  starts_with() : starts_or_ends_with(true) {
+  }
+};
+
+class ends_with final : public virtual starts_or_ends_with {
+public:
+  ends_with() : starts_or_ends_with(false) {
+  }
+};
```
mavam, line 84: Any reason why the boolean is not a template parameter and a member instead?

jachris: I don't really like the template pattern for deriving multiple plugins when there is nothing gained by making the parameter part of the type system. So I was just trying something else, but this here is not that much better. I would rather be able to pass arguments in `TENZIR_REGISTER_PLUGIN`.

tobim (approved):

PR #4241, opened 2024-05-23
Add an RPM tab in the deploy-a-node documentation

tobim:
This adds one missing piece from #4188, namely the addition of an `RPM` tab in the manual node installation instructions.

### DIFF
--- a/web/docs/setup-guides/deploy-a-node.md
+++ b/web/docs/setup-guides/deploy-a-node.md
@@ -64,6 +64,18 @@ dpkg -i tenzir-static-amd64-linux.deb
 [tenzir-debian-package]: https://github.com/tenzir/tenzir/releases/latest/download/tenzir-static-amd64-linux.deb
 
 </TabItem>
+<TabItem value="rpm_based" label="RPM based (RedHat / OpenSUSE / ...)">
+
+Download the latest [RPM package][tenzir-rpm-package] and install it via
+`rpm`:
+
+```bash
+yum -y localinstall tenzir-static-amd64-linux.rpm
+```
+
+[tenzir-rpm-package]: https://github.com/tenzir/tenzir/releases/latest/download/tenzir-static-amd64-linux.rpm
+
+</TabItem>
 <TabItem value="nix" label="Nix">
 
 Use our `flake.nix`:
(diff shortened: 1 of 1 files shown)


web/docs/setup-guides/deploy-a-node.md
```diff
@@ -63,6 +63,18 @@ dpkg -i tenzir-static-amd64-linux.deb
 
 [tenzir-debian-package]: https://github.com/tenzir/tenzir/releases/latest/download/tenzir-static-amd64-linux.deb
 
+</TabItem>
+<TabItem value="rpm_based" label="RPM based (RedHat / OpenSUSE / ...)">
```
dominiklohmann, line 67:
```suggestion
<TabItem value="rpm_based" label="RPM-based (RedHat, OpenSUSE, Fedora)">
```

I think those are the main three.

web/docs/setup-guides/deploy-a-node.md
```diff
@@ -63,6 +63,18 @@ dpkg -i tenzir-static-amd64-linux.deb
...
+
+Download the latest [RPM package][tenzir-rpm-package] and install it via
+`rpm`:
```
dominiklohmann, line 70:
```suggestion
`dnf` or `yum`:
```

I think `dnf` is actually the more modern version.

web/docs/setup-guides/deploy-a-node.md
```diff
@@ -63,6 +63,18 @@ dpkg -i tenzir-static-amd64-linux.deb
...
+
+```bash
+yum -y localinstall tenzir-static-amd64-linux.rpm
```
dominiklohmann, line 73: With `dnf` it should be `sudo dnf localinstall tenzir-static-amd64-linux.rpm` if I Googled the right thing. I think we should document that as well.

Should we add the sudo for `yum` as well?

dominiklohmann (approved):
Awesome. Kinda makes me want to cut a new release this week again, even though it's only minor new things.

web/docs/setup-guides/deploy-a-node.md
tobim: I think we should actually use `rpm -i` because that should work across all distros, whereas the high level commands `yum`, `dnf`, `zypper`,... are distro dependent.

dominiklohmann: Even better.

web/docs/setup-guides/deploy-a-node.md
tobim: We also didn't add the `sudo` in the other tabs, and I think either way is as good as the other, as long as it is consistent.

dominiklohmann: Makes sense.

PR #4242, opened 2024-05-23
Support multiple fields in `sort`

dominiklohmann:
This is a rather trivial implementation that's certainly not optimal in terms of performance, but I'd argue this is better than not having it.


mavam:
How does that interplay with the `asc` and `desc` modifiers? Are they per field or globally?

dominiklohmann:
> How does that interplay with the `asc` and `desc` modifiers? Are they per field or globally?

Per field, that's the only thing that really makes sense.

web/docs/operators/sort.md
```diff
@@ -64,3 +68,9 @@ Arrange by field `foo` in descending order and put null values first:
 ```
 sort foo desc nulls-first
 ```
+
+Sort by the field `src_ip` and for matching values sort by `dest_ip`.
+
+```
+sort src_ip, dest_ip
```
tobim, line 75: So if I want to specify the order I would write something like:
```
sort src_ip asc, src_bytes desc 
```
?

tobim (approved):
Looks pretty straight-forward.

web/docs/operators/sort.md
dominiklohmann: Yes. Let me add an example.

PR #4244, opened 2024-05-24
Suppress noise when we are not on an RPM distro

mavam:
This PR reduces confusing output when we are not on an RPM distro.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -69,7 +69,7 @@ platform=
 os=$(uname -s)
 if [ "${os}" = "Linux" ]
 then
-    if [ -n "$(rpm -qa)" ] 2>/dev/null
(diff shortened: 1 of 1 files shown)


dominiklohmann (approved):

scripts/install.sh
```diff
@@ -69,7 +69,7 @@ platform=
 os=$(uname -s)
 if [ "${os}" = "Linux" ]
 then
-    if [ -n "$(rpm -qa)" ] 2>/dev/null
+    if check rpm && [ -n "$(rpm -qa 2>/dev/null)" ]
```
mavam, line 72: @tobim I haven't checked whether the redirect into the `rpm` call causes issues. Please confirm and resolve if that seems okay.

tobim: I verified that this silences the unintended output.

PR #4246, opened 2024-05-24
Minor installer improvements

tobim:
* Clear the destination path of the package download so `curl` does not abort.
* Drop the check for `adduser`, because it is out of place in the installer.
* Add `/opt/tenzir/bin` to the system default path for ease of use.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -164,6 +164,7 @@ echo "Using ${package}"
 # Download package.
 tmpdir="$(dirname "$(mktemp -u)")"
 action "Downloading ${package_url}"
+rm -f "${tmpdir}/${package}"
 # Wget does not support the file:// URL scheme.
 if check wget && beginswith "${package_url}" "https://"
 then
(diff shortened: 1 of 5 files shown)


scripts/install.sh
```diff
@@ -209,6 +210,7 @@ fi
 
 # Trigger installation.
 action "Installing package into ${prefix}"
+insert_path="$sudo echo 'export PATH=\$PATH:/opt/tenzir/bin' > /etc/profile.d/tenzir.sh"
```
tobim, line 213: I'm not sure how to achieve this on macOS.

mavam: There's no equivalent. Shells have their specific way of doing this, e.g., via `.bash_profile`.

mavam: Can we create a symlink that points from `/usr/local/bin/tenzir` to `/opt/tenzir/bin/tenzir`?

mavam:
@tobim are you moving the `/etc/profile.d` actions to the RPM/DEB package in this PR?

PR #4248, opened 2024-05-26
Support disabling TLS in `https` connector

mavam:
To disable TLS peer and hostname verification, the `https` connector now gains
the new `--skip-peer-verification` and `--skip-hostname-verification` options as
already known from the `email` saver.


web/docs/connectors/https.md
```diff
@@ -12,10 +12,37 @@ Loads bytes via HTTPS.
...
+
+When negotiating TLS and SSL connections, the server sends a certificate
+indicating its identity. By default, that certificate must indicate that the
+server is the server to which you meant to connect, or the connection fails.
+That is, the server has to have the same name in the certificate as is in the
+URL you operate against. We consider the server the intended one when the
+*Common Name* field or a *Subject Alternate Name* field in the certificate
+matches the hostname in the URL.
+
+Providing this flag skips this check, but it makes the connection insecure.
```
mavam, line 48: This section could arguably go into the http connector, as https is only an alias. Thoughts?

dominiklohmann (approved):

web/docs/connectors/https.md
dominiklohmann: I think either works.

PR #4249, opened 2024-05-27
Add an option for writing JSON arrays of objects

dominiklohmann:
The new `write json --arrays-of-objects` option writes all events arriving within the same batch into a JSON array. This is convenient when interfacing with REST APIs through the `http` connector.

Fixes https://github.com/tenzir/issues/issues/1855

### DIFF
--- a/web/docs/formats/json.md
+++ b/web/docs/formats/json.md
@@ -15,7 +15,8 @@ Parser:
 
 ```
 json [--schema <schema>] [--selector <field[:prefix]>] [--unnest-separator <string>]
-     [--no-infer] [--ndjson] [--precise] [--raw] [--arrays-of-objects]
+     [--no-infer] [--ndjson] [--precise] [--raw]
+     [--arrays-of-objects]
 ```
 
 Printer:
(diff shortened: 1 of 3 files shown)


web/docs/formats/json.md
```diff
@@ -233,6 +234,39 @@ With `--omit-empty-lists`, this example becomes:
 
 This options combines all other `--omit-*` options.
 
+### `--arrays-of-objects` (Printer)
```
mavam, line 237:
```suggestion
### `--arrays-of-objects`
```

web/docs/formats/json.md
```diff
@@ -233,6 +234,39 @@ With `--omit-empty-lists`, this example becomes:
...
+
+Prints one array of objects per batch of events arriving at the printer as
+opposed to printing one object per event.
```
mavam, line 240: You should probably split the content here: "For the parser, this option does X. For the printer, it does Y."

mavam (approved):

mavam:
...and a changelog entry.

web/docs/formats/json.md
dominiklohmann: We discussed this quickly at the office and prefer the separate entries because they follow the order of the synopsis.

PR #4250, opened 2024-05-27
Improve integration test robustness

tobim:
This PR addresses a bit of flakyness in the integration tests:
* The `read stdin via fluent-bit` test is now skipped because it sometimes fails to produce any output.
* The tests of the `lookup` operator now install a test timeout and add some waiting time for initializing background lookups before continuing with subsequent steps.

This PR also adds opt-in CI steps to connect to the Tenzir tailnet for debugging.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -46,6 +46,15 @@ jobs:
           git submodule update --init --recursive libtenzir
           git submodule update --init --recursive plugins
           git submodule update --init --recursive tenzir
+      - name: Tailscale
+        if: false
+        uses: tailscale/github-action@v2
+        with:
+          oauth-client-id: ${{ secrets.TS_OAUTH_CLIENT_ID }}
+          oauth-secret: ${{ secrets.TS_OAUTH_SECRET }}
+          tags: tag:ci
+          version: 1.66.4
+          args: --ssh
       - name: Configure
         id: cfg
         run: |
--- a/plugins/fluent-bit/integration/tests/tests.bats
+++ b/plugins/fluent-bit/integration/tests/tests.bats
@@ -20,6 +20,11 @@ setup() {
 }
 
 @test "read stdin via fluent-bit" {
+  # In general we deem the fluent-bit stdin functionality as too unreliable.
+  # Please refrain if you came here with the intention to re-enable this test,
+  # we only keep it to document that the operator should not be tested this
+  # way.
+  skip "This test often fails to produce any output when run on GitHub Actions Runners"
   # fluent-bit has an internal race condition that can cause the loss of
   # events during startup. We have to accept this for now and avoid
   # test flakyness with a bit of trickery.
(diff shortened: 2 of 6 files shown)


plugins/fluent-bit/integration/tests/tests.bats
```diff
@@ -20,13 +20,14 @@ setup() {
 }
 
 @test "read stdin via fluent-bit" {
+  skip "This test often fails to produce any output when run on GitHub Actions Runners"
```
mavam, line 23: Can we just ditch the test entirely?

plugins/fluent-bit/integration/tests/tests.bats
```diff
@@ -20,13 +20,14 @@ setup() {
...
   # fluent-bit has an internal race condition that can cause the loss of
   # events during startup. We have to accept this for now and avoid
   # test flakyness with a bit of trickery.
   result="$(
     while :; do
       echo '{"foo": {"bar": 42}}'
-      sleep 1
+      sleep 1 || exit 1
     done |
```
dominiklohmann, line 31: For the exit 1 to take any effect it's need to come after `done`, right? As-is, the while loop swallows the exit code.

plugins/fluent-bit/integration/tests/tests.bats
dominiklohmann: From past experience I would say: ditch it. We know we can't make this work reliably with the Fluent Bit library, and it's not worth sinking time into this test.

contrib/tenzir-plugins
dominiklohmann, line 1: The changes in here make me think that we want a version of

```
check --bg tenzir <pipeline>
```

that only returns once the pipeline has actually started.

One way to already do this is using the API, as spawning a pipeline via the API behaves like that. However, we then require a node and cannot check for stdout anymore. Is there perhaps another mechanism that we can utilize for this?

.github/workflows/docker.yaml
```diff
@@ -46,6 +46,15 @@ jobs:
...
           git submodule update --init --recursive tenzir
+      - name: Tailscale
+        if: false
+        uses: tailscale/github-action@v2
+        with:
+          oauth-client-id: ${{ secrets.TS_OAUTH_CLIENT_ID }}
+          oauth-secret: ${{ secrets.TS_OAUTH_SECRET }}
+          tags: tag:ci
+          version: 1.66.4
+          args: --ssh
```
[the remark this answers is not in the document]
dominiklohmann: Resolving this; I have a weak preference for removing this, but we discussed it, and @tobim wants to keep it in here, so I'll yield.

plugins/fluent-bit/integration/tests/tests.bats
tobim: I would not remove it entirely, but extend the explanation for the skip to tell the tempted developer to not try to fix this test. That way the knowledge that `fluent-bit stdin` is not reliable enough for this kind of testing is preserved, and nobody wastes time reimplementing a similar test in case we get a bug report for that feature.

dominiklohmann (approved):

PR #4251, opened 2024-05-29
Treat `publish`, `subscribe`, and `version` as internal

dominiklohmann:
Now that we render metrics from node-internal operators grayed-out, it became quickly apparent that these three were not categorized correctly.


jachris (approved):
Looks complete to me now.

PR #4265, opened 2024-06-03
Print Operator

balavinaithirthan:
1. The new `print` operator allows for printing an event field as a string. The operator constructs a `print` with the given printer type and target field. The operator loops through each event and turns the corresponding field into a string. This only works if the field is a record_type.


libtenzir/builtins/operators/print.cpp
```diff
@@ -0,0 +1,198 @@
...
+        auto builder = series_builder();
+        auto printer_instance = printer_->instantiate(field.type, ctrl);
+        auto row_as_slice = to_record_batch(row);
+        auto chunks = std::vector<chunk_ptr>{};
+        std::ranges::copy(printer_instance->get()->process(row),
+                          std::back_inserter(chunks));
+        std::ranges::copy(printer_instance->get()->finish(),
+                          std::back_inserter(chunks));
+        if (chunks.empty()) {
+          return {};
```
dominiklohmann, line 120: This should add the empty string instead.

libtenzir/builtins/operators/print.cpp
```diff
@@ -0,0 +1,198 @@
...
+            .primary(input_.source)
+            .throw_();
+        }
+        field.type = type{"dummy_head", field.type};
+        auto rb = arrow::RecordBatch::Make(
+          field.type.to_arrow_schema(), array->length(),
+          static_cast<const arrow::StructArray&>(*array).Flatten().ValueOrDie());
+        auto row = table_slice{rb, field.type};
+        auto builder = series_builder();
+        auto printer_instance = printer_->instantiate(field.type, ctrl);
```
dominiklohmann, line 112: After this we need to check whether an error was emitted through the control plane, and if it was, throw it.

libtenzir/builtins/operators/print.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+        return {{
+          {field.name, series.type},
+          series.array,
+        }};
+      };
+      auto transformations = std::vector<indexed_transformation>{
+        {std::move(*target_index), transform},
+      };
+      for (size_t i = 0; i < slice.rows(); i++) {
```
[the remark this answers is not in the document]
dominiklohmann: Yes, we do need to move this in.

libtenzir/builtins/operators/print.cpp
```diff
@@ -0,0 +1,198 @@
...
+            std::string_view(data_as_char, data_as_char + data.size()));
+        }
+        auto series = builder.finish_assert_one_array();
+        auto check_utf_8
+          = arrow::compute::CallFunction("utf8_is_printable", {series.array});
+        if (not check_utf_8.ok()) {
+          diagnostic::error("{}. {} does not print UTF8. ", printer_->name(),
+                            check_utf_8.status().ToString())
+            .throw_();
+        }
```
dominiklohmann, line 141: This is too strict, we want to allow all UTF8 results. We should check before we add to the builder.

dominiklohmann: Consider experimenting with an API change like this that we can check upfront:

```diff
M libtenzir/include/tenzir/plugin.hpp
@@ -454,6 +454,10 @@ public:
   /// Returns whether the printer allows for joining output streams into a
   /// single saver.
   virtual auto allows_joining() const -> bool = 0;
+
+  /// Returns whether the printer has a text-based output formatted that is
+  /// guaranteed to be (1) valid UTF8 and (2) supposed to be human-readable.
+  virtual auto text_output() const -> bool = 0;
 };
 
 /// @see operator_parser_plugin
```

dominiklohmann (approved):

libtenzir/builtins/operators/print.cpp
```diff
@@ -0,0 +1,217 @@
...
+            });
+          } catch (diagnostic diag) {
+            std::move(diag)
+              .modify()
+              .severity(severity::warning)
+              .emit(ctrl.diagnostics());
+            builder.null();
+            continue;
+          } catch (...) {
+            std::rethrow_exception(std::current_exception());
```
jachris, line 142:
```suggestion
            throw;
```

👆 Absolute nitpick, but that's more idiomatic (and there are subtle differences). Can also ignore ^^

dominiklohmann: TIL, I didn't know this was a thing.

jachris: Although I just noticed that you could also just omit the entire `catch (...)` clause then... 😅  

Well, whatever, I won't block CI once more just because of that.

libtenzir/builtins/operators/print.cpp
```diff
@@ -0,0 +1,217 @@
...
+          }
+          auto validate_and_add = [&](std::string_view str) {
+            TENZIR_ASSERT_EXPENSIVE(arrow::util::ValidateUTF8(str));
+            builder.data(str);
+          };
+          if (chunks.empty()) {
+            builder.data("");
+          }
+          if (chunks.size() == 1) {
```
jachris, line 155: Sorry, I found one more bug.
```suggestion
          } else if (chunks.size() == 1) {
```
Otherwise we add to the builder twice if it's empty.

dominiklohmann: Don't say sorry for having spotted bugs :)

PR #4266, opened 2024-06-03
Improve formatting rules for lambdas

dominiklohmann:
```cpp
// before
foo([] {
      // …
    });

// after
foo([] {
  // …
});
```

I think this shouldn't be controversial, but I figured I'd open a separate PR anyways instead of sneaking this into the one where I actually made this change because I was annoyed by the old formatting rules.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -45,6 +45,7 @@ IndentWidth: 2
 IndentWrappedFunctionNames: false
 InsertBraces: true
 KeepEmptyLinesAtTheStartOfBlocks: false
+LambdaBodyIndentation: OuterScope
 Language: Cpp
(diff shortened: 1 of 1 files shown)


Dakostu (approved):

jachris (changes requested):
This leads to bad formatting:

~~~cpp
  for (auto& arg : inv.args) {
      arg.match(
        [&](ast::assignment& x) {
        check_assignment(x, ctx);
        assignments.push_back(std::move(x));
      },
        [&](auto&) {
        diagnostic::error("expected assignment")
          .primary(arg.get_location())
          .usage(usage)
          .docs(docs)
          .emit(ctx.dh());
      });
    }
~~~

dominiklohmann:
> I think this shouldn't be controversial

Famous last words 🙈

PR #4267, opened 2024-06-03
Log leftover actor ids when shutting down

tobim:
This is an experimental change to the actor system registry in CAF. Instead of maintaining a simple running actor counter, it now keeps all ids of the running actors. This change should allow us to debug actor termination problems.

### DIFF
--- a/tenzir/tenzir.cpp
+++ b/tenzir/tenzir.cpp
@@ -237,5 +237,11 @@ auto main(int argc, char** argv) -> int {
     render_error(*root, run_error, std::cerr);
     return EXIT_FAILURE;
   }
+  while (sys.registry().running() > 1) {
+    TENZIR_VERBOSE("remaining actors: [{}], named: [{}]",
+                   fmt::join(sys.registry().running_ids(), ", "),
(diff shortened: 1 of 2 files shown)


tenzir/tenzir.cpp
```diff
@@ -237,5 +237,11 @@ auto main(int argc, char** argv) -> int {
     render_error(*root, run_error, std::cerr);
     return EXIT_FAILURE;
   }
+  while (sys.registry().running() > 1) {
+    TENZIR_VERBOSE("remaining actors: [{}], named: [{}]",
+                   fmt::join(sys.registry().running_ids(), ", "),
+                   fmt::join(sys.registry().named_actors(), ", "));
+    std::this_thread::sleep_for(std::chrono::seconds(2));
```
dominiklohmann, line 244: Isn't the sleep actively harmful here? If we enter this loop at least once (which is likely) then the shutdown in the correct case takes another two seconds.

dominiklohmann:
Closing this; we fixed the hang so this isn't needed right now, and if we do need it again we can always choose to bring it back.

PR #4273, opened 2024-06-04
Add percentile aggregation functions

dominiklohmann:
This adds aggregation functions `p99`, `p95`, `p90`, `p75`, and `p50`, and renames `approximate_median` to `median`.

With TQL2 we probably want this to just be `percentile(0.99, foo)` rather than `p99(foo)`, but for now this is the best we can do and I had a use case for these.

### DIFF
--- a/libtenzir/builtins/aggregation-functions/quantile.cpp
+++ b/libtenzir/builtins/aggregation-functions/quantile.cpp
@@ -7,19 +7,22 @@
 // SPDX-License-Identifier: BSD-3-Clause
 
 #include <tenzir/aggregation_function.hpp>
+#include <tenzir/detail/string_literal.hpp>
 #include <tenzir/plugin.hpp>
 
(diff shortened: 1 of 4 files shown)


libtenzir/builtins/aggregation-functions/quantile.cpp
```diff
@@ -107,8 +118,20 @@ class plugin : public virtual aggregation_function_plugin {
   }
 };
 
+using median_plugin = plugin<"median", 50>;
+using p50_plugin = plugin<"p50", 50>;
+using p75_plugin = plugin<"p75", 75>;
+using p90_plugin = plugin<"p90", 90>;
+using p95_plugin = plugin<"p95", 95>;
+using p99_plugin = plugin<"p99", 99>;
```
jachris, line 126: Perhaps a feature changelog for this?

jachris (approved):

PR #4274, opened 2024-06-04
Enable selective deletion of lookup table entries

dominiklohmann:
For `lookup-table` contexts, the `context update` operator now sports a new `--erase` flag, which entries to be removed rather than aded.


mavam (approved):
We paired on this, LGTM.

PR #4279, opened 2024-06-06
Allow pip to write to stdout in venv creation

tobim:
Pip sometimes wants to draw a spinner on `stdout` during package installation. This currently leads to setup failures, because we're closing `stdout` before execing pip and pip does not check if `stdout` is `None` when probing if it is a TTY.

### DIFF
--- a/libtenzir/builtins/operators/python.cpp
+++ b/libtenzir/builtins/operators/python.cpp
@@ -285,7 +285,7 @@ public:
           TENZIR_VERBOSE("installing python modules with: '{}'",
                          fmt::join(pip_invocation, "' '"));
           if (bp::system(pip_invocation, env, bp::std_err > std_err,
-                         detail::preserved_fds{{STDERR_FILENO}},
+                         detail::preserved_fds{{STDOUT_FILENO, STDERR_FILENO}},
                          boost::process::limit_handles)
               != 0) {
             auto pip_error = drain_pipe(std_err);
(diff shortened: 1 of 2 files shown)


libtenzir/builtins/operators/python.cpp
dominiklohmann, line 1: Doesn't this break things like `tenzir '… | python … | to stdout' > file.json`?

tobim: No, because `pip` guards these prints against !isatty.

dominiklohmann: Could we also fix this then by letting pip never think that it has a TTY?

dominiklohmann (approved):
Approving, but let's agree on the approach before merging.

libtenzir/builtins/operators/python.cpp
tobim: I tested the case you brought up and it works as expected.

> Could we also fix this then by letting pip never think that it has a TTY?

There is no explicit option for that. `-q` only affects the logger. I think pip's `isatty()` check should be fine.
A more robust alternative might be to re-route stdout to `void`/`/dev/null`, but I'm not sure it's worth the effort.

dominiklohmann: Makes sense to me.

Mind adding a changelog entry, then merging this?

PR #4284, opened 2024-06-07
Implement the `context rename` operator

Dakostu:
This change implements context renaming while the node is running via the `context rename` operator.

Also fixes a few bugs along the way.

### DIFF
--- a/web/docs/operators/context.md
+++ b/web/docs/operators/context.md
@@ -20,6 +20,7 @@ context reset   <name>
 context save    <name>
 context load    <name>
 context inspect <name>
+context rename  <old-name> <new-name> [--force]
 ```
 
 ## Description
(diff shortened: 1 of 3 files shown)


web/docs/operators/context.md
```diff
@@ -46,6 +47,8 @@ The `context` operator manages [context](../contexts.md) instances.
 - The `inspect` command dumps a specific context's user-provided data, usually
   the context's content.
 
+- The `rename` operator renames a context.
```
dominiklohmann, line 50: Can you explain what `--force` does here?

Dakostu: Added an explanation.

dominiklohmann:
Closing this; we'll reconsider this in the future, for now we don't have a need for this.

PR #4285, opened 2024-06-10
Correct the URL for RPMs in the installer

tobim:


lava (approved):
I verified that I can download an rpm from the updated URL.

PR #4288, opened 2024-06-11
Fix start abort error message

jachris:
This changes the error message for something like `version | version` from
~~~
error: failed to run
error: 
~~~
to
~~~
error: 'version' does not accept events as input
 = note: failed type inference
~~~
I don't think the resulting diagnostic is good, but this PR is just about restoring the behavior that we had previously (until it was broken by one of the diagnostic propagation PRs).


dominiklohmann (approved):

dominiklohmann (approved):
This took a bit of back-and-forth, but I think except for the open question of whether we want to/can remove `ec::silent` entirely now we're good.

PR #4289, opened 2024-06-11
Fix using `summarize … by x` when `x` is of type `null`

dominiklohmann:
This is a two-fold change:
1. The `put`, `set`, `extend`, and `replace` operators when used to assign a field the value `null` set the string type rather than the null type, because they predated the null type. This is now fixed.
2. The `summarize` operator no longer crashes when grouping by a null type.

Fixes tenzir/issues#1900

### DIFF
--- a/libtenzir/builtins/operators/summarize.cpp
+++ b/libtenzir/builtins/operators/summarize.cpp
@@ -396,6 +396,10 @@ public:
             // already warned and can ignore it.
             continue;
           }
+          if (not other->type) {
+            // We can skip `null_type` as that is compatible with every other.
+            continue;
+          }
           if (existing.is_dead()) {
             continue;
           }
@@ -470,7 +474,7 @@ public:
       auto new_bucket = std::make_shared<bucket>();
       new_bucket->group_by_types.reserve(bound.group_by_columns.size());
       for (auto&& column : bound.group_by_columns) {
-        if (column) {
+        if (column and column->type) {
           new_bucket->group_by_types.push_back(
             group_type::make_active(column->type));
         } else {
--- a/changelog/next/bug-fixes/4289--summarize-by-null.md
+++ b/changelog/next/bug-fixes/4289--summarize-by-null.md
@@ -1,0 +1,3 @@
+The `summarize` operator no longer crashes when grpuping by a field of type
+`null`, i.e., a field whose type could not be inferred because all of its values
+were `null`.
--- a/libtenzir/src/table_slice.cpp
+++ b/libtenzir/src/table_slice.cpp
@@ -872,11 +872,9 @@ auto resolve_operand(const table_slice& slice, const operand& op)
     }
     inferred_type = *tmp_inferred_type;
     if (not inferred_type) {
-      inferred_type = type{string_type{}};
-      // Tenzir has no N/A type equivalent for Arrow, so we just use a string
-      // type here.
+      inferred_type = type{null_type{}};
       auto builder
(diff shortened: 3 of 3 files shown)


libtenzir/builtins/operators/summarize.cpp
jachris, line 1: I think this works for now. Ideally, `null` and column-does-not-exist would lead to the same state here (IIUC they don't with this change), but I guess that's okay, given that TQLv2 already has a different `summarize` implementation anyway.

jachris (approved):

libtenzir/builtins/operators/summarize.cpp
dominiklohmann: Absolutely, that was my thinking as well. This is basically a throwaway bug fix with TQL2 having the proper fix.

jachris: I just noticed that this should be a one-liner. What do you think?
~~~
diff --git a/libtenzir/builtins/operators/summarize.cpp b/libtenzir/builtins/operators/summarize.cpp
index 9e9d1dd47e..350f04d9d5 100644
--- a/libtenzir/builtins/operators/summarize.cpp
+++ b/libtenzir/builtins/operators/summarize.cpp
@@ -470,7 +470,7 @@ public:
       auto new_bucket = std::make_shared<bucket>();
       new_bucket->group_by_types.reserve(bound.group_by_columns.size());
       for (auto&& column : bound.group_by_columns) {
-        if (column) {
+        if (column && column->type) {
           new_bucket->group_by_types.push_back(
             group_type::make_active(column->type));
         } else {

~~~

dominiklohmann: Let my try it out, if it works then it's obviously better because it skips the grouping.

jachris: It does not skip the grouping. It just makes it so that we can upgrade the type later. Try it with mixed null-type/non-null-type if you want to.

jachris: I noticed the change above doesn't really suffice, because there is another callsite of `.set_active` with `type`. Those would also need to do some type upgrading I believe.

dominiklohmann: Take another look, I think I covered both now.

jachris: There is a `not` missing, or:

~~~
diff --git a/libtenzir/builtins/operators/summarize.cpp b/libtenzir/builtins/operators/summarize.cpp
index 9e9d1dd47e..9c09d82397 100644
--- a/libtenzir/builtins/operators/summarize.cpp
+++ b/libtenzir/builtins/operators/summarize.cpp
@@ -396,6 +396,10 @@ public:
             // already warned and can ignore it.
             continue;
           }
+          if (not other->type) {
+            // We can skip `null_type` as that is compatible with every other.
+            continue;
+          }
           if (existing.is_dead()) {
             continue;
           }
@@ -470,7 +474,7 @@ public:
       auto new_bucket = std::make_shared<bucket>();
       new_bucket->group_by_types.reserve(bound.group_by_columns.size());
       for (auto&& column : bound.group_by_columns) {
-        if (column) {
+        if (column && column->type) {
           new_bucket->group_by_types.push_back(
             group_type::make_active(column->type));
         } else {
~~~

dominiklohmann: You're right. 🙈

PR #4295, opened 2024-06-14
Allow components to depend on other components

tobim:
This extends the component_plugin interface so that can provide a list of other components that have to be instantiated before the plugins own component can be spawned.

### DIFF
--- a/libtenzir/builtins/components/metrics_collector.cpp
+++ b/libtenzir/builtins/components/metrics_collector.cpp
@@ -63,7 +63,9 @@ struct metrics_collector_state {
         return ok.error();
       }
     }
-    TENZIR_ASSERT(not instances.empty());
+    if (instances.empty()) {
+      return {};
+    }
     detail::weak_run_delayed_loop(
       self, std::chrono::seconds{30},
       [this] {
@@ -139,9 +141,6 @@ public:
 
   auto make_component(node_actor::stateful_pointer<node_state> node) const
     -> component_plugin_actor override {
-    if (collect(plugins::get<metrics_plugin>()).empty()) {
-      return {};
-    }
     auto [importer] = node->state.registry.find<importer_actor>();
     return node->spawn<caf::linked>(metrics_collector, std::move(importer));
   }
--- a/libtenzir/src/node.cpp
+++ b/libtenzir/src/node.cpp
@@ -284,6 +284,56 @@ auto node_state::get_endpoint_handler(const http_request_description& desc)
   return result->second;
 }
 
+auto spawn_components(node_actor::stateful_pointer<node_state> self) -> void {
+  using component_plugin_map = std::unordered_map<std::string, const component_plugin*>;
+  // 1. Collect all component_plugins into a name -> plugin* map:
+  component_plugin_map todo = {};
+  for (const auto* component : plugins::get<component_plugin>()) {
+    todo.emplace(component->component_name(), component);
+  }
+  // 2. Calculate an ordered loading sequnce based on the wanted_components of
+  //    each plugin.
+  std::vector<const component_plugin*> sequenced_components = {};
+  std::unordered_set<std::string> done = {};
+  auto derive_sequence = [&](auto derive_sequence, const std::string& name) {
+    auto entry = todo.find(name);
+    if (entry == todo.end()) {
+      return;
+    }
+    const auto* plugin = entry->second;
+    todo.erase(entry);
+    if (done.contains(name)) {
+      return;
+    }
+    for (auto& wanted : plugin->wanted_components()) {
+      derive_sequence(derive_sequence, wanted);
+    }
+    done.insert(name);
+    sequenced_components.push_back(plugin);
+  };
+  while (not todo.empty()) {
+    derive_sequence(derive_sequence, todo.begin()->second->component_name());
+  }
+  // 3. Load all components in order.
+  for (const auto* plugin : sequenced_components) {
+    auto name = plugin->component_name();
+    auto handle = plugin->make_component(self);
+    if (!handle) {
+      diagnostic::error("{} failed to create the {} component", *self, name)
+        .throw_();
+    }
+    self->system().registry().put(fmt::format("tenzir.{}", name), handle);
+    if (auto err = register_component(self, caf::actor_cast<caf::actor>(handle),
+                                      name)) {
+      diagnostic::error(err)
+        .note("{} failed to register component {} in component registry", *self,
+              name)
+        .throw_();
+    }
+    self->state.ordered_components.push_back(handle);
+  }
+}
+
 node_actor::behavior_type
 node(node_actor::stateful_pointer<node_state> self, std::string /*name*/,
      std::filesystem::path dir) {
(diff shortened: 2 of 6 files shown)


libtenzir/src/node.cpp
```diff
@@ -330,6 +372,23 @@ node(node_actor::stateful_pointer<node_state> self, std::string /*name*/,
...
+    } catch (...) {
+      auto x = std::current_exception();
+      err = caf::scheduled_actor::default_exception_handler(self, x);
+      if (err) {
+        self->call_error_handler(err);
+      }
+    }
+    self->call_error_handler(err);
+    return err;
+  });
```
tobim, line 391: Not sure if we still need this, isn't there a global default that does the same thing now?

libtenzir/src/node.cpp
```diff
@@ -284,6 +284,48 @@ auto node_state::get_endpoint_handler(const http_request_description& desc)
...
+    self->system().registry().put(fmt::format("tenzir.{}", needed), handle);
+    if (auto err = register_component(self, caf::actor_cast<caf::actor>(handle),
+                                      needed)) {
+      diagnostic::error(err)
+        .note("{} failed to register component {} in component registry", *self,
+              needed)
+        .throw_();
+    }
+    loaded.insert(std::move(*entry));
+    not_loaded.erase(entry);
```
tobim, line 322: This line needs to be executed always.

libtenzir/src/node.cpp
```diff
@@ -284,6 +284,48 @@ auto node_state::get_endpoint_handler(const http_request_description& desc)
...
+  using component_plugin_map = std::unordered_map<std::string, const component_plugin*>;
+  component_plugin_map loaded = {};
+  component_plugin_map not_loaded = {};
+  for (const auto* component : plugins::get<component_plugin>()) {
+    not_loaded.emplace(component->component_name(), component);
+  }
+  auto load_recursive = [&](auto load_recursive, const std::string& needed) {
+    if (loaded.contains(needed)) {
+      return;
+    }
```
tobim, line 297: The variable name `needed` is confusing. We should rename it to `named`.
```suggestion
  auto load_recursive = [&](auto load_recursive, const std::string& needed) {
    if (loaded.contains(needed)) {
      return;
    }
```

libtenzir/src/node.cpp
```diff
@@ -284,6 +284,48 @@ auto node_state::get_endpoint_handler(const http_request_description& desc)
   return result->second;
 }
 
+auto spawn_components(node_actor::stateful_pointer<node_state> self) -> void {
+  using component_plugin_map = std::unordered_map<std::string, const component_plugin*>;
+  component_plugin_map loaded = {};
+  component_plugin_map not_loaded = {};
+  for (const auto* component : plugins::get<component_plugin>()) {
+    not_loaded.emplace(component->component_name(), component);
```
tobim, line 292: Assert that this is not a duplicate.

libtenzir/src/node.cpp
```diff
@@ -284,6 +284,48 @@ auto node_state::get_endpoint_handler(const http_request_description& desc)
...
+    if (entry == not_loaded.end()) {
+      TENZIR_WARN("{} is trying to load the unknown component {}, expect "
+                  "reduced functionality.",
+                  *self, needed);
+      return;
+    }
+    for (auto& required : entry->second->required_components()) {
+      load_recursive(load_recursive, required);
+    }
+    auto handle = entry->second->make_component(self);
```
tobim, line 308: It might make sense to write the component into a serialized vector first and load it later in a dedicated loop. This would make it easier to implement a unit test for the algorithm.

tobim, line 308: The shutdown order is currently untouched. This is based on the assumption that these dependencies are only relevant during component initialization (for example for starting configured pipelines) but have no effect for the shutdown, but this might not be right.

Imho the correct way to model the shutdown of components would be to keep a data structure that models the dependency graph in the node and holds actor handles to the components to increase the ref-count. Once the shutdown is triggered the components that are not referenced by others in the graph can be removed, the rest will be removed step by step in the down_handler when they don't have any live dependencies any more.

tobim: The simple solution would be to keep the sequential loading order and use the `terminator` to shut the components down in the reverse order, also sequentially.

tobim: Implemented the simple approach for now.

libtenzir/src/node.cpp
tobim: We still need it.

libtenzir/builtins/components/metrics_collector.cpp
```diff
@@ -139,9 +139,6 @@ class plugin final : public virtual component_plugin {
 
   auto make_component(node_actor::stateful_pointer<node_state> node) const
     -> component_plugin_actor override {
-    if (collect(plugins::get<metrics_plugin>()).empty()) {
-      return {};
-    }
```
dominiklohmann, line 144: The assertion in L66 is going to fire when you run into this, so that would now need to be handled gracefully.

libtenzir/src/node.cpp
```diff
@@ -284,6 +284,59 @@ auto node_state::get_endpoint_handler(const http_request_description& desc)
...
+  // 2. Calculate an ordered loading sequnce based on the wanted_components of
+  //    each plugin.
+  std::vector<const component_plugin*> sequenced_components = {};
+  std::unordered_set<std::string> done = {};
+  auto derive_sequence = [&](auto derive_sequence, const std::string& name) {
+    auto entry = todo.find(name);
+    if (entry == todo.end()) {
+      TENZIR_WARN("{} is trying to load the unknown component {}, expect "
+                  "reduced functionality.",
+                  *self, name);
```
dominiklohmann, line 303: I think this should either be a hard error or a debug message, but nothing in-between makes sense to me.

libtenzir/src/node.cpp
```diff
@@ -330,6 +383,23 @@ node(node_actor::stateful_pointer<node_state> self, std::string /*name*/,
       }
     }
   });
+  self->set_exception_handler([=](std::exception_ptr& ptr) -> caf::error {
+    caf::error err;
+    try {
+      std::rethrow_exception(ptr);
+    } catch (const diagnostic& diag) {
+      err = diag.to_error();
+      TENZIR_ERROR("{}", err);
```
dominiklohmann, line 392: If the node actor shuts down with an error we already print that in the `start_command` which started the node and waits for it to exit. This is going to cause diagnostics to be rendered twice.

libtenzir/src/node.cpp
```diff
@@ -330,6 +383,23 @@ node(node_actor::stateful_pointer<node_state> self, std::string /*name*/,
...
+    } catch (...) {
+      auto x = std::current_exception();
+      err = caf::scheduled_actor::default_exception_handler(self, x);
+      if (err) {
+        self->call_error_handler(err);
+      }
+    }
+    self->call_error_handler(err);
```
dominiklohmann, line 400: This could call the error handler twice, no?

tobim: I'm not even sure why we need to call this. We don't in the other custom `exception_handler`s and neither do the CAF unit tests.

libtenzir/builtins/components/metrics_collector.cpp
```diff
@@ -63,7 +63,6 @@ struct metrics_collector_state {
         return ok.error();
       }
     }
-    TENZIR_ASSERT(not instances.empty());
```
dominiklohmann, line 66: You could return early here.

libtenzir/src/node.cpp
```diff
@@ -354,6 +420,12 @@ node(node_actor::stateful_pointer<node_state> self, std::string /*name*/,
     // Core components are terminated in a second stage, we remove them from the
     // registry upfront and deal with them later.
     std::vector<caf::actor> core_shutdown_handles;
+    for (const auto& name :
+         std::ranges::reverse_view(self->state.ordered_components)) {
```
dominiklohmann, line 424: Could be `self->state.ordered_components | std::ranges::views::reverse`, which I think is a bit more idiomatic.

dominiklohmann (approved):

PR #4297, opened 2024-06-17
Tone down execution node backoff behavior

dominiklohmann:
This turns down the execution node backoff behavior. Instead of going from 10ms to 1s with 1.25x increases, we now go from 250ms to 1 min with 2x increases.

Additionally, this causes execution nodes that explicitly ask not to be scheduled to be exempt from backoff and not to be scheduled at all.

### DIFF
--- a/libtenzir/src/execution_node.cpp
+++ b/libtenzir/src/execution_node.cpp
@@ -236,8 +236,9 @@ struct exec_node_state {
   caf::typed_response_promise<void> start_rp = {};
 
   /// Exponential backoff for scheduling.
-  static constexpr duration min_backoff = std::chrono::milliseconds{10};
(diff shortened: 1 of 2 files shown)


libtenzir/src/execution_node.cpp
```diff
@@ -647,21 +649,25 @@ struct exec_node_state {
     advance_generator();
     // We can continue execution under the following circumstances:
     // 1. The operator's generator is not yet completed.
-    // 2. The operator has one of the three following reasons to do work:
+    // 2. The operator did not signal that we're supposed to wait.
+    // 3. The operator has one of the three following reasons to do work:
```
Avaq, line 653:
```suggestion
    // 3. The operator has one of the four following reasons to do work:
```

jachris (approved):

PR #4299, opened 2024-06-17
Remove built-in type aliases

dominiklohmann:
This removes the `timestamp` and `port` type aliases in bundled schemas and built-in parsers and source operator. They were aliased to `time` and `uint64`, respectively.

We're making this change because many operators do not handle aliased types well, and because type inference cannot infer them. While we do want to have a notion of an "event timestamp" eventually, the `timestamp` failed to deliver on its promise. We will add a new concept for this in the future that works better.


mavam (approved):

tenzir/integration/data/reference/vast/test_node_zeek_dns_log/step_02.ref
dominiklohmann, line 1: Need to adjust this test.

PR #4300, opened 2024-06-17
Make the parallel level in `lookup` configurable

dominiklohmann:
This adds a new `--parallel <level>` option to the `lookup` operator that controls how many partitions it opens at once for retrospective lookups. Cranking this number up makes lookups significantly faster, but also causes its resource usage to rise accordingly.


mavam (approved):

Dakostu (approved):
OK - I left a comment in the plugins submodule repo.

PR #4307, opened 2024-06-18
Implement lookup metric events

Dakostu:
This change implements metric events for the `lookup` operator.

Each second, a new metric will expose:
* The amount of {live,retro,snapshot} matches within that second.
* The amount of context updates within that round.
* The total partition queue size.


Dakostu:
Closed in favor of https://github.com/tenzir/tenzir/pull/4339

PR #4312, opened 2024-06-19
Fix HTTP header parsing logic

mavam:
Trailing `=` in a Base64 header caused the wrong branch to be selected.

### DIFF
--- a/libtenzir/src/http.cpp
+++ b/libtenzir/src/http.cpp
@@ -38,12 +38,24 @@ auto message::header(const std::string& name) const -> const struct header* {
 }
 
 auto request_item::parse(std::string_view str) -> std::optional<request_item> {
+  auto is_valid_header_name = [](std::string_view name) {
+    for (char c : name) {
+      if (std::isalnum(static_cast<unsigned char>(c)) == 0 and c != '-'
+          and c != '_') {
+        return false;
+      }
+    }
+    return true;
+  };
   auto xs = detail::split_escaped(str, ":=@", "\\", 1);
   if (xs.size() == 2)
     return request_item{.type = file_data_json, .key = xs[0], .value = xs[1]};
   xs = detail::split_escaped(str, ":=", "\\", 1);
   if (xs.size() == 2)
     return request_item{.type = data_json, .key = xs[0], .value = xs[1]};
+  xs = detail::split_escaped(str, ":", "\\", 1);
+  if (xs.size() == 2 and is_valid_header_name(xs[0]))
+    return request_item{.type = header, .key = xs[0], .value = xs[1]};
   xs = detail::split_escaped(str, "==", "\\", 1);
   if (xs.size() == 2)
     return request_item{.type = url_param, .key = xs[0], .value = xs[1]};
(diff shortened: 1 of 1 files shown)


dominiklohmann (approved):

libtenzir/src/http.cpp
```diff
@@ -50,15 +59,15 @@ auto request_item::parse(std::string_view str) -> std::optional<request_item> {
   xs = detail::split_escaped(str, "=@", "\\", 1);
   if (xs.size() == 2)
     return request_item{.type = file_data, .key = xs[0], .value = xs[1]};
+  xs = detail::split_escaped(str, ":", "\\", 1);
+  if (xs.size() == 2)
```
dominiklohmann, line 63:
```suggestion
  if (xs.size() == 2 and is_valid_header_name(xs[0]))
```

You forgot to call your new function 🙈

dominiklohmann: Or this:

```diff
M libtenzir/src/http.cpp
@@ -39,13 +39,10 @@ auto message::header(const std::string& name) const -> const struct header* {
 
 auto request_item::parse(std::string_view str) -> std::optional<request_item> {
   auto is_valid_header_name = [](std::string_view name) {
-    for (char c : name) {
-      if (std::isalnum(static_cast<unsigned char>(c)) == 0 and c != '-'
-          and c != '_') {
-        return false;
-      }
-    }
-    return true;
+    return std::ranges::all_of(name, [](char c) {
+      return std::isalnum(static_cast<unsigned char>(c)) == 0 or c == '-'
+             or c == '_';
+    });
   };
   auto xs = detail::split_escaped(str, ":=@", "\\", 1);
   if (xs.size() == 2)
@@ -53,15 +50,16 @@ auto request_item::parse(std::string_view str) -> std::optional<request_item> {
   xs = detail::split_escaped(str, ":=", "\\", 1);
   if (xs.size() == 2)
     return request_item{.type = data_json, .key = xs[0], .value = xs[1]};
+  xs = detail::split_escaped(str, ":", "\\", 1);
+  if (xs.size() == 2 and is_valid_header_name(xs[0])) {
+    return request_item{.type = header, .key = xs[0], .value = xs[1]};
+  }
   xs = detail::split_escaped(str, "==", "\\", 1);
   if (xs.size() == 2)
     return request_item{.type = url_param, .key = xs[0], .value = xs[1]};
   xs = detail::split_escaped(str, "=@", "\\", 1);
   if (xs.size() == 2)
     return request_item{.type = file_data, .key = xs[0], .value = xs[1]};
-  xs = detail::split_escaped(str, ":", "\\", 1);
-  if (xs.size() == 2)
-    return request_item{.type = header, .key = xs[0], .value = xs[1]};
   xs = detail::split_escaped(str, "@", "\\", 1);
   if (xs.size() == 2)
     return request_item{.type = file_form, .key = xs[0], .value = xs[1]};
```

mavam: Dang, I had amended the commit locally but didn't force-push properly.

mavam: But your diff is not correct. The `isalnum` call should have been an inequality. 😉

dominiklohmann: The header check still isn't high enough, isn't it? Should be before all other arguments that do not start with a colon in the separator.

mavam: We need to try to parse `:=` first. Otherwise we will always split on `:`.

dominiklohmann: That's what I'm doing in my diff, but it definitely needs to be before the split on `==`.

mavam: I moved it before `==`. That'll do for now. 😅

PR #4315, opened 2024-06-20
Extend the component actor interface with online reconfiguration

tobim:
All component actors now have to implement a handler that receives the configuration in the same way as their parent plugins during initialization.


libtenzir/include/tenzir/atoms.hpp
```diff
@@ -61,6 +61,8 @@ CAF_BEGIN_TYPE_ID_BLOCK(tenzir_atoms, caf::id_block::tenzir_types::end)
   TENZIR_ADD_ATOM(move, "move")
   TENZIR_ADD_ATOM(next, "next")
   TENZIR_ADD_ATOM(normalize, "normalize")
+  TENZIR_ADD_ATOM(package_add, "padd")
+  TENZIR_ADD_ATOM(package_remove, "premove")
```
tobim, line 65: I don't think that every atom needs to be verb form, as long as every handler receives exactly one verb-atom. We already have `add` and `erase`, why not add a `package` atom and combine as necessary?

dominiklohmann:
Closing as superseded by #4344

PR #4317, opened 2024-06-21
Add missing `-X` option for `kafka` saver

dominiklohmann:
This was advertised to exist in the docs, but was actually only implemented for the ´kafka` loader. The fix is mostly a carbon copy of the loader code.

Fixes tenzir/issues#1889


IyeOnline (approved):
Looks good. Now I don't have to do it :)

PR #4322, opened 2024-06-21
Revert "Allow components to depend on other components (#4295)"

dominiklohmann:
This reverts commit 7b612d3ae3a7e37f0a50cf7db5cd441f7d82ec4e, reversing changes made to 49f0c603f54208b3b66ebe325591abe02ec2d8c3.


lava (approved):
I didn't read through the code in detail, but we observed that nodes fail to start with this patch applied and can successfully boot after the revert.

PR #4329, opened 2024-06-24
Add an operator for partition candidate checks

dominiklohmann:
The `partitions [<expr>]` operator supersedes the `show partitions` operator. The new and optional positional argument is a filter expression for doing a candidate check in the catalog component. This has two primary use cases:
1. Seeing which schemas would be considered for `export | where <expr>` by running `partitions <expr> | deduplicate schema`.
2. Quickly calculating an upper bound for the number of results in `export | where <expr>` by running `partitions <expr> | summarize events=sum(events)`.

Fixes https://github.com/tenzir/issues/issues/1532

### DIFF
--- a/libtenzir/include/tenzir/actors.hpp
+++ b/libtenzir/include/tenzir/actors.hpp
@@ -200,8 +200,14 @@ using catalog_actor = typed_actor_fwd<
     ->caf::result<atom::ok>,
   // Merge a set of partition synopses.
   auto(atom::merge, std::vector<partition_synopsis_pair>)->caf::result<atom::ok>,
-  // Get *ALL* partition synopses stored in the catalog.
+  // Get *ALL* partition synopses stored in the catalog, optionally filtered
+  // with an expression to filter the candidate set.
+  // Note that this returns a pointer into the catalog's internal data
+  // structures, which is inherently unsafe to transfer between processes. The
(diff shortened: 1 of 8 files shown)


libtenzir/include/tenzir/actors.hpp
```diff
@@ -202,6 +202,8 @@ using catalog_actor = typed_actor_fwd<
   auto(atom::merge, std::vector<partition_synopsis_pair>)->caf::result<atom::ok>,
   // Get *ALL* partition synopses stored in the catalog.
   auto(atom::get)->caf::result<std::vector<partition_synopsis_pair>>,
+  // Get *ALL* partition synopses stored in the catalog for a given predicate.
+  auto(atom::get, expression)->caf::result<std::vector<partition_synopsis_pair>>,
```
tobim, line 206: I'm not a fan of giving out pointers to internal memory from the catalog here, but it seems to be nothing new. Could you please add a comment to warn the next coder to never write to the structure returned by this handler and the one above?

tobim (approved):
I have only one minor remark about missing in-code documentation. I suggest you let the current CI run finish before addressing that, and then you can force-merge to save some time.

libtenzir/include/tenzir/actors.hpp
dominiklohmann: Done.

PR #4333, opened 2024-06-25
Fix two concurrency issues related to child process creation

tobim:
This fixes "Bad file descriptor" errors from the python operator and shell operators.
Additionally, a thread-safety issue in the python operators environment cloning routine is worked around.

### DIFF
--- a/libtenzir/builtins/operators/python.cpp
+++ b/libtenzir/builtins/operators/python.cpp
@@ -242,7 +242,7 @@ public:
           if (bp::system(python_executable, "-m", "venv", venv,
                          bp::std_err > std_err,
                          detail::preserved_fds{{STDERR_FILENO}},
-                         boost::process::limit_handles)
+                         bp::detail::limit_handles_{})
               != 0) {
             auto venv_error = drain_pipe(std_err);
             // We need to delete the potentially broken venv here to make sure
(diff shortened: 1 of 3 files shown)


libtenzir/builtins/operators/python.cpp
```diff
@@ -445,13 +445,15 @@ class python_operator final : public crtp_operator<python_operator> {
 
 private:
   config config_ = {};
+  mutable bp::environment env_ = {};
```
dominiklohmann, line 448: This will always be default-initialized in the operator when you actually run it. As part of the optimization process, the operator goes through an inspect roundtrip, and this just doesn't make it past that.

The call operator of `crtp_operator` is also marked `const` for a very good reason. If the state must be mutable then it should be on the coroutine stack, not outside of it.This will always be default-initialized in the operator when you actually run it. As part of the optimization process, the operator goes through an inspect roundtrip, and this just doesn't make it past that.

The call operator of `crtp_operator` is also marked `const` for a very good reason. If the state must be mutable then it should be kept on the coroutine stack, not outside of it.

tobim: I forgot about the inspect roundtrip. Fixing that would mean writing an overload for `boost::process::environment`. I opted to implement a slightly more expensive but much simpler approach now.

lava (approved):
Makes sense to me

PR #4334, opened 2024-06-26
Fix configured pipelines causing a crash when they contain a syntax error

Dakostu:
Configured pipelines with retry-on-error enabled that caused a failure upon launching were triggering an assertion due to the starting callback lambda capturing an old version of the `stored_pipeline_info` with a missing `start_time`. This PR fixes that.


tobim (approved):
Reviewed in the submodule repo.

PR #4338, opened 2024-07-01
Make patterns consider newlines when encountering the dot metacharacter

Dakostu:
This change makes the dot metacharacter in patterns consider newlines by default to make processing of quasi-structured field values easier.


mavam (approved):

dominiklohmann:
We've opted not to make this change for now.

PR #4339, opened 2024-07-01
Implement the metrics handler

Dakostu:
This change implements a `metric_handler` with a way to emit additional metrics for operators. Also, some operators now emit these custom metrics.


libtenzir/include/tenzir/metrics.hpp
tobim, line 1: Align this filename with the class name.

libtenzir/src/exec_pipeline.cpp
```diff
@@ -246,6 +246,9 @@ auto exec_pipeline(std::string content,
         [&](diagnostic& d) {
           diag->emit(std::move(d));
         },
+        [&](std::string&, record&) {
+          // drop custom metrics.
```
tobim, line 250: Treat similar to the handler below.

libtenzir/include/tenzir/metrics.hpp
Dakostu: Done in https://github.com/tenzir/tenzir/pull/4339/commits/a12adcc553be0237f7a61c845214cc7098e42431

libtenzir/src/exec_pipeline.cpp
Dakostu: I added custom metric handling in https://github.com/tenzir/tenzir/pull/4339/commits/26c37425433535201598f1644dbf473fdd7d27d4

contrib/tenzir-plugins
dominiklohmann, line 1: Can you document the new metrics here? https://docs.tenzir.com/operators/metrics#schemas

Dakostu: Yes, there is no documentation at all right now.

Dakostu: Added in https://github.com/tenzir/tenzir/pull/4339/commits/c61f1c35f6f41c43b6263a337846dfe832213fed

web/docs/operators/metrics.md
```diff
@@ -109,6 +109,86 @@ Contains a measurement of the amount of memory used by the `tenzir-node` process
...
+Contains a measurement of data that goes through the `publish` operator, for each data batch.
+
+|Field|Type|Description|
+|:-|:-|:-|
+|`timestamp`|`time`|The time when this event was emitted.|
+|`operator_index`|`uint64`|The index of the `publish` operator in the pipeline.|
+|`topic`|`string`|The topic name.|
+|`schema`|`string`|The schema name of the batch.|
+|`schema_id`|`string`|The schema ID of the batch.|
+|`events`|`uint64`|The amount of events that were published to the `topic`.|
```
dominiklohmann, line 147: One thing I just noticed: If values are optional, or negative, it can be easy to infer the wrong type here. Did you consider an approach like this where you seed the series builder with a type explicitly?

```cpp
auto mh = ctrl.metrics_handler(type{"tenzir.metrics.publish", record_type{
  {"topic", string_type{}},
  ...
}});

// ...

mh.emit(record{
  {"topic", topic},
  ...
});

Comment (Dakostu) [reply web/docs/operators/metrics.md:147]: I looked into defining something like `metric_handler::emit(const type& t, record&& r) `, but was not satisfied with having to forward a type to `emit()` calls every time.
Your snippet suggests that `ctrl` should create a `metric_handler` for that specific type first, then we can just keep `emit`ting record. I'll look into whether that works.

Comment (Dakostu) [reply web/docs/operators/metrics.md:147]: OK, it does work. It just requires some awkward `type` transformations (adding `timestamp`/`operator_id` to all metrics, and `pipeline_id`/`run`/`hidden` for metrics entering the pipeline manager) for the builders to accept the events.

I also aligned the `series_builder` creation mechanism for different metric types in the pipeline manager - the generic metrics do not require a separate `series_builder` in the `analytics_collector` anymore but are part of the operator-specific metric builder map.

Comment (dominiklohmann) [reply web/docs/operators/metrics.md:147]: Great, I think this change is for the better. Having the output schema be inferred can lead to quite a few nasty surprises, I think.

File: CMakePresets.json
Comment (dominiklohmann) [inline CMakePresets.json:1]: I assume that these changes were pushed accidentally? There's also CMakeUserPresets.json, which is a git-ignored file with equivalent behavior and higher precedence than CMakePresets.json.

File: libtenzir/src/metric_handler.cpp [hunk 3]
```diff
@@ -0,0 +1,47 @@
...
+    }),
+  }});
+  metric_type_ = type{metric_type.name(), *params_type, {{"internal", ""}}};
+}
+
+auto metric_handler::emit(record&& r) -> void {
+  // Explicitly create a data-from-time cast here to support macOS builds.
+  r["timestamp"] = data{time{time::clock::now()}};
+  r["operator_id"] = op_index_;
+  caf::anon_send(receiver_.lock(), metric_type_, std::move(r));
```
dominiklohmann, line 40: There's one problem with this approach: Sending the type along every single time is cheap when the operator is in the same process as the diagnostic handler, and absurdly expensive when it isn't.

Do you wanna pair sometime today to make this cheaper and push this PR over the finish line?

CMakePresets.json
Dakostu: Oh, yes. I'll rebase the PR to remove this change from https://github.com/tenzir/tenzir/pull/4339/commits/da06eeacd5c9450b348ace2d5b6ab797f78ad19b.

libtenzir/src/metric_handler.cpp
Dakostu: OK; after I prepared my other PR.

CMakePresets.json
Dakostu: The file is gone from the PR now.

dominiklohmann (approved):

PR #4341, opened 2024-07-02
Remove the zero-size check in the split_at_null() input loop

Dakostu:
The zero-size check in the `split_at_null()` input loop was causing the loop to invalidate chunks if the first character was a null character. This is easily triggerable by using the `gelf` parser on `gralog` data. The NDJSON events are separated by null characters, and the parser was ignoring events if such an invalidation occured.

### DIFF
--- a/libtenzir/builtins/formats/json.cpp
+++ b/libtenzir/builtins/formats/json.cpp
@@ -107,9 +107,6 @@ inline auto split_at_null(generator<chunk_ptr> input, char split)
         continue;
       }
       const auto size = static_cast<size_t>(current - begin);
-      if (size == 0) {
-        continue;
-      }
       const auto capacity = static_cast<size_t>(end - begin);
       if (buffer.empty() and capacity >= size + simdjson::SIMDJSON_PADDING) {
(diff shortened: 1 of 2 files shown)


jachris (approved):

libtenzir/builtins/formats/json.cpp
jachris, line 1: Do you want to add a changelog for this bug fix?

Dakostu: Yes, I am about to add one. 
This is easily user-facing, especially if you check the `diagnostics`.

jachris: It would not be "a", but "an" NDJSON parser, although from a user POV this has nothing to do with NDJSON. Also, in my opinion, it's good to qualify such statements w.r.t. to their severity/frequency. Something like:

> We fixed a rarely occurring issue in the `gelf` parser that led to parsing errors for some events.

Dakostu: I applied your suggestion

PR #4342, opened 2024-07-02
Allow plugins to be registered with arguments

dominiklohmann:
Just a refactoring, not a functional change. Idea shamelessly stolen from https://github.com/tenzir/tenzir/pull/4239#discussion_r1659739896.

### DIFF
--- a/libtenzir/include/tenzir/plugin.hpp
+++ b/libtenzir/include/tenzir/plugin.hpp
@@ -1056,21 +1056,23 @@ extern const char* TENZIR_PLUGIN_VERSION;
       ::tenzir::plugin_ptr::make_builtin(__VA_ARGS__, {})
 #  endif
 
-#  define TENZIR_REGISTER_PLUGIN(name)                                         \
-    template <class>                                                           \
+#  define TENZIR_REGISTER_PLUGIN(...)                                          \
+    template <auto>                                                            \
     struct auto_register_plugin;                                               \
     template <>                                                                \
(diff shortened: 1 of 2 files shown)


libtenzir/include/tenzir/plugin.hpp
```diff
@@ -1056,21 +1056,23 @@ extern const char* TENZIR_PLUGIN_VERSION;
...
-    struct auto_register_plugin<name> {                                        \
+    struct auto_register_plugin<[] {}> {                                       \
       auto_register_plugin() {                                                 \
         static_cast<void>(flag);                                               \
       }                                                                        \
       static auto init() -> bool {                                             \
+        /* NOLINTBEGIN(cppcoreguidelines-owning-memory) */                     \
         ::tenzir::plugins::get_mutable().push_back(TENZIR_MAKE_PLUGIN(         \
-          new (name), /* NOLINT(cppcoreguidelines-owning-memory) */            \
+          new (name)(__VA_ARGS__),                                             \
```
jachris, line 1070: I think you can enable `TENZIR_REGISTER_PLUGIN(tenzir::plugins::quantile::plugin{"median", 0.5})` if you make this `TENZIR_REGISTER_PLUGIN(...)` and `new __VA_ARGS__`.

jachris (approved):
Nice! I think this may or may not work with operators because their name needs to be known after default-construction, but this is still good to have.

PR #4344, opened 2024-07-02
Add a package manager

lava:


libtenzir/include/tenzir/atoms.hpp
```diff
@@ -61,6 +61,8 @@ CAF_BEGIN_TYPE_ID_BLOCK(tenzir_atoms, caf::id_block::tenzir_types::end)
   TENZIR_ADD_ATOM(move, "move")
   TENZIR_ADD_ATOM(next, "next")
   TENZIR_ADD_ATOM(normalize, "normalize")
+  TENZIR_ADD_ATOM(package_add, "padd")
+  TENZIR_ADD_ATOM(package_remove, "premove")
```
tobim, line 65: Please use only a `package` atom and combine it with `add` and `remove` atoms as needed.

libtenzir/include/tenzir/fwd.hpp
```diff
@@ -506,6 +512,7 @@ CAF_BEGIN_TYPE_ID_BLOCK(tenzir_types, first_tenzir_type_id)
   // TODO: Make list, record, and map concrete typs to we don't need to do
   // these kinda things. See tenzir/aliases.hpp for their definitions.
   TENZIR_ADD_TYPE_ID((std::vector<tenzir::data>))
+  TENZIR_ADD_TYPE_ID((std::vector<tenzir::package>))
```
tobim, line 515: Can be removed.

changelog/next/4344--package-manager.md
```diff
@@ -0,0 +1,3 @@
+In the community edition, users can now add and remove packages
```
tobim, line 1: "You can now ..."

changelog/next/4344--package-manager.md
```diff
@@ -0,0 +1,3 @@
...
+using the new 'package' operator. A package is a combination of
+pipelines and contexts that is deployed to a node as a set.
```
dominiklohmann, line 3: We don't usually mention the editions in the changelog.

changelog/next/4344--package-manager.md
```diff
@@ -0,0 +1,3 @@
+In the community edition, users can now add and remove packages
+using the new 'package' operator. A package is a combination of
```
dominiklohmann, line 2:
```suggestion
using the new `package` operator. A package is a combination of
```

libtenzir/include/tenzir/actors.hpp
```diff
@@ -297,6 +297,8 @@ using filesystem_actor = typed_actor_fwd<
   auto(atom::write, std::filesystem::path, chunk_ptr)->caf::result<atom::ok>,
   // Reads a chunk of data from a given path and returns the chunk.
   auto(atom::read, std::filesystem::path)->caf::result<chunk_ptr>,
+  // Reads all files from a given directory and returns their content as a 'blob'.
+  auto(atom::read, atom::recursive, std::filesystem::path)->caf::result<record>,
```
dominiklohmann, line 301: This looks like a mis-use of the `record` type to me. Why not return a `map` from string to `chunk_ptr`?

libtenzir/include/tenzir/actors.hpp
```diff
@@ -471,6 +490,7 @@ CAF_BEGIN_TYPE_ID_BLOCK(tenzir_actors, caf::id_block::tenzir_atoms::end)
   TENZIR_ADD_TYPE_ID((tenzir::node_actor))
   TENZIR_ADD_TYPE_ID((tenzir::partition_actor))
   TENZIR_ADD_TYPE_ID((tenzir::partition_creation_listener_actor))
+  TENZIR_ADD_TYPE_ID((tenzir::package_listener_actor))
```
dominiklohmann, line 493: As far as I can tell we don't send this in messages, so why does this need a type id?

libtenzir/include/tenzir/fwd.hpp
```diff
@@ -506,7 +512,10 @@ CAF_BEGIN_TYPE_ID_BLOCK(tenzir_types, first_tenzir_type_id)
...
   TENZIR_ADD_TYPE_ID((tenzir::detail::stable_map<std::string, tenzir::data>))
+  TENZIR_ADD_TYPE_ID(
+    (std::vector<tenzir::detail::stable_map<std::string, tenzir::data>>))
```
dominiklohmann, line 518: Is this a `vector<record>`? Where do we send that?

libtenzir/include/tenzir/package.hpp
```diff
@@ -0,0 +1,190 @@
...
+  : public detail::flat_map<std::string, package_context> {
+  using super = detail::flat_map<std::string, package_context>;
+  using super::super;
+};
+
+using package_snippets_list = std::vector<package_snippet>;
+
+struct package final {
+  std::string id = {};   // required
+  std::string name = {}; // required
```
dominiklohmann, line 156: If the other fields are not required, shouldn't they be modeled as `std::optional<std::string>`?

libtenzir/include/tenzir/plugin.hpp
```diff
@@ -145,7 +145,10 @@ class component_plugin : public virtual plugin {
 
   /// Components that should be created before the current one so initialization
   /// can succeed.
-  /// Defaults to empty list.
+  /// Note that the *only* guarantee made is that components are able to
+  /// retrieve actor handles of the wanted components from the registry, but
+  /// there is no guarantee that the initialization routine is already
+  /// completed. Defaults to empty list.
```
dominiklohmann, line 151: I don't understand this comment. I assume that with "initialization routine" you're referring to the code up to returning the behavior. But that runs in its own thread, and any message sent to the actor handle will only arrive after the initial behavior was returned. So what is this warning the developer about?

libtenzir/src/posix_filesystem.cpp
```diff
@@ -44,6 +44,39 @@ posix_filesystem_state::rename_single_file(const std::filesystem::path& from,
   return atom::done_v;
 }
 
+auto read_recursive(const std::filesystem::path& root,
+                    /*const std::filesystem::path& current_path,*/ size_t&
```
dominiklohmann, line 48: Looks like the `current_path` parameter can be removed?

tenzir.yaml.example
```diff
@@ -270,6 +270,11 @@ tenzir:
       # - Set the option to a valid duration to enable with a custom delay.
       restart-on-error: 1 minute
 
+
+  # The path from which the node attempts to read statically configured packages.
+  # This setting is ignored unless the pipeline manager plugin is enabled.
+  # packages-dir: <sysconfir>/packages
+
```
dominiklohmann, line 277: Instead of making this configurable, can you make this a fixed `<configdir>/packages`, and please also support the other config directories beside the sysconfdir?

web/docs/operators/package.md
```diff
@@ -0,0 +1,68 @@
...
+management support of Tenzir. All interfaces are subject to change,
+and user-facing documentation of the package format is still in progress.
+:::
+
+## Synopsis
+
+```
+package add
+package remove <package_id>
+show packages
```
dominiklohmann, line 24: This should be documented in `show.md`.

web/docs/operators/package.md
```diff
@@ -0,0 +1,68 @@
...
+
+- The `add` command adds a new package on the node by
+  running all pipelines and contexts defined in the package.
+
+- The `remove` command removes an existing package.
+
+- The `show packages` command generates a list of successfully
+  installed packages.
+
+Note: When using TQLv2, the operators are called `package_add` and `package_remove` respectively.
```
dominiklohmann, line 39: Please leave out the TQL2 docs for now, that will come all at once.

web/docs/operators/package.md
```diff
@@ -0,0 +1,68 @@
...
+### `<package_id>`
+
+The unique id of the package, as found in the package definition.
+
+## Examples
+
+Add a package from a public URL:
+
+```
+from https://github.com/tenzir/library/raw/main/feodo/package.yaml read yaml
```
dominiklohmann, line 50: No need to specify this, the format is inferred from the file extension.

```suggestion
from https://github.com/tenzir/library/raw/main/feodo/package.yaml
```

web/docs/operators/package.md
```diff
@@ -0,0 +1,68 @@
...
+
+```
+package add
+package remove <package_id>
+show packages
+```
+
+## Description
+
+The `packages` operator manages packages.
```
dominiklohmann, line 29: Can you explain briefly here what a `package` is?

dominiklohmann, line 29:
```suggestion
The `package` operator manages packages.
```

web/docs/operators/package.md
```diff
@@ -0,0 +1,68 @@
...
+sidebar_custom_props:
+  operator:
+    source: true
+    transformation: false
+    sink: true
+---
+
+# package
+
+Manages the packages installed on a given node.
```
dominiklohmann, line 11: The node isn't a parameter, so "on a given node" doesn't quite make sense. Maybe this?

```suggestion
Manages packages at a node.
```

libtenzir/include/tenzir/detail/fanout_counter.hpp
```diff
@@ -99,4 +100,24 @@ auto make_fanout_counter(size_t expected, Continuation&& then,
                                    std::forward<ErrorContinuation>(error));
 }
 
+// template <typename State>
+// auto make_fanout_counter(size_t expected, caf::typed_response_promise<State>
+// rp) {
+//   return make_fanout_counter<State>(expected,
+//     [rp](State&& state) mutable { rp.deliver(std::move(state)); },
+//     [rp](State&&, caf::error&& e) mutable { rp.deliver(e); });
+// }
```
dominiklohmann, line 109: dead code.

libtenzir/include/tenzir/detail/fanout_counter.hpp
```diff
@@ -99,4 +100,24 @@ auto make_fanout_counter(size_t expected, Continuation&& then,
...
+
+inline auto
+make_fanout_counter(size_t expected, caf::typed_response_promise<void> rp) {
+  return make_fanout_counter(
+    expected,
+    [rp]() mutable {
+      rp.deliver();
+    },
+    [rp](caf::error&& e) mutable {
+      rp.deliver(e);
```
dominiklohmann, line 119:
```suggestion
    [rp](caf::error e) mutable {
      rp.deliver(std::move(e));
```

libtenzir/include/tenzir/detail/fanout_counter.hpp
```diff
@@ -99,4 +100,24 @@ auto make_fanout_counter(size_t expected, Continuation&& then,
...
+// template <typename State>
+// auto make_fanout_counter(size_t expected, caf::typed_response_promise<State>
+// rp) {
+//   return make_fanout_counter<State>(expected,
+//     [rp](State&& state) mutable { rp.deliver(std::move(state)); },
+//     [rp](State&&, caf::error&& e) mutable { rp.deliver(e); });
+// }
+
+inline auto
+make_fanout_counter(size_t expected, caf::typed_response_promise<void> rp) {
```
dominiklohmann, line 112:
```suggestion
make_fanout_counter(size_t expected, const caf::typed_response_promise<void>& rp) {
```

Saves one copy, since you copy twice below.

changelog/next/4344--package-manager.md
dominiklohmann, line 1: Unrelated to this file, but `.github/workflows/tenzir.yaml` must be adapted to also run the tests for the new plugin.

contrib/tenzir-plugins
dominiklohmann, line 1: I also left a lot of small comments in the submodule.

libtenzir/include/tenzir/actors.hpp
lava: A record preserves the directory structure, with a map we'd have to reconstruct it from the keys in a post-processing step (which might not even be unambigously possible, if there are filenames containing slashes)

We could make a nested map, but I don't see the benefit of introducing a second recursive data structure when we already have one that can do the job.

libtenzir/include/tenzir/fwd.hpp
lava: In the context manager and in the package manager.

tenzir.yaml.example
lava: Is there a use case for this? If not I'd prefer not to, since it's a lot of additional work since we'd have to collect the set of all packages from all  configdirs, handle conflicts, etc.

Setting the package dir directly has also proven useful so far, e.g. for installing some commit of the whole library for testing purposes, or for easily overriding it in tests.

libtenzir/include/tenzir/actors.hpp
dominiklohmann: Ah, it was not clear to me from the comment that this keeps the directory structure by returning nested record fields for directories. Then let's just keep this as-is for now, although I do think it'd be better to change this eventually.

tenzir.yaml.example
dominiklohmann: > Setting the package dir directly has also proven useful so far, e.g. for installing some commit of the whole library for testing purposes, or for easily overriding it in tests.

I'm not against having it configurable, but even then it should probably work similarly to `--config`, `--plugin-dirs` and `--schema-dirs`, where it simply adds a directory with the highest precedence instead of disabling the default directories, and disabling the default directories requires setting `--bare-mode`.

lava: How about just *not* setting a default value then? Then we can avoid the whole issue and for the case where we actually want to use this for deploying packages via IaC we're controlling the node config anyways and can just set the directory.

dominiklohmann: But why have this behave different from the other config subdirectories?

libtenzir/include/tenzir/plugin.hpp
lava: If there is a request happening before the behavior is returned, there is no guarantee that the request is already in the receiving actors inbox. Or in other words, if there are two components who depend on each other there's no guarantee that such requests from the one that is first in the list are arriving before the second.

I'll try to reword the comment to that effect.

libtenzir/include/tenzir/package.hpp
dominiklohmann: From the app's perspective there is a difference, I think, whether there is an author or not. And `show packages | where author != null` does also not work if the author is rendered as an empty string.

.github/workflows/tenzir.yaml
```diff
@@ -823,6 +827,9 @@ jobs:
           - name: Pipeline Manager
             target: pipeline-manager
             path: contrib/tenzir-plugins/pipeline-manager
+          - name: Packages
```
dominiklohmann, line 830: We probably want to remove the Context and Pipeline Manager plugins from the list above, they're going to fail to build now that they're bundled.

PR #4345, opened 2024-07-03
Make the python operator usable in TQL2

tobim:


libtenzir/builtins/operators/python.cpp
jachris, line 1: I took the liberty to push to this branch (improve error messages and worked around on some limitations of the current implementation, e.g., w.r.t. throwing). 

I hope that's okay. Please resolve if my edits are fine.

jachris (approved):
Thanks!

libtenzir/builtins/operators/python.cpp
dominiklohmann: Is it intended that you use emit for one error and throw for the other?

jachris: Oops, not at all. Thank you!

PR #4348, opened 2024-07-03
Prepare the pipeline manager for TQLv2

jachris:
This is an internal change that does not have any visible impact (yet).

### DIFF
--- a/libtenzir/src/diagnostics.cpp
+++ b/libtenzir/src/diagnostics.cpp
@@ -13,6 +13,7 @@
 #include "tenzir/shared_diagnostic_handler.hpp"
 
 #include <boost/algorithm/string.hpp>
+#include <boost/functional/hash.hpp>
 #include <caf/message_handler.hpp>
 #include <fmt/color.h>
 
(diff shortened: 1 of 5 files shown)


libtenzir/src/diagnostics.cpp
```diff
@@ -242,4 +243,26 @@ void diagnostic_builder::emit(const shared_diagnostic_handler& diag) && {
   return diag.emit(std::move(result_));
 }
 
+auto diagnostic_deduplicator::insert(const diagnostic& d) -> bool {
+  // We remember whether we have seen a diagnostic by storing its main message
+  // and the locations of its annotations.
+  // TODO: Improve this.
+  auto locations = std::vector<location>{};
+  for (auto& annotation : d.annotations) {
+    locations.push_back(annotation.source);
```
dominiklohmann, line 252: Should the locations be sorted here?

jachris: This would only matter if there is a diagnostic with multiple annotations where the order is derived from something else or even is non-deterministic. And then it's not clear to me whether you'd want to consider those duplicates.

dominiklohmann: Makes sense.

PR #4353, opened 2024-07-05
Tie pipeline autocreation & autodeletion to the hidden flag

Dakostu:
This change tightens the enforcement of requirements for `hidden` pipelines. They must always start automatically, and autodelete in all stopping cases. Instead of making the user enforce that (and even missing an enforcement for `autodeletion` flags), we simply overwrite the flags in the request.


dominiklohmann (approved):
Thanks for the quick fix, the changes make sense to me. Didn't even think about the update restriction you put in, that is a nice catch.

PR #4357, opened 2024-07-06
Improve and fix TQLv2 error propagation

jachris:
This PR improves error propagation by using `failure_or<T>` in more places, thus statically guaranteeing that an error has been emitted if no `T` is returned. However, functions can also emit an error and _still_ return a `T`, as long as the calling code would be safe to proceed with it. This enables better error messages. Because of this, we need an explicit check before we start the actual execution of a parsed pipeline, since we only want to do that if no errors have been emitted.


dominiklohmann (approved):
:shipit:

PR #4359, opened 2024-07-08
Honor proxy settings in the SQS connector

tobim:


plugins/sqs/src/plugin.cpp
```diff
@@ -51,6 +51,10 @@ class sqs_queue {
     if (auto endpoint_url = detail::getenv("AWS_ENDPOINT_URL_SQS")) {
       config.endpointOverride = *endpoint_url;
     }
```
tobim, line 53: Can we actually remove this now? Not sure why this should be tied to an Arrow version instead of AWS_SDK_CPP.

mavam: We tried removing it once when it already was supposed to be fixed and it didn't work.  

I'm happy to try again.

mavam (approved):

plugins/sqs/src/plugin.cpp
tobim: Turns out this is surprisingly difficult to test, so we now just leave it in.

PR #4364, opened 2024-07-09
Handle complexity errors in `grok` parser

jachris:
Some grok/regex patterns lead to catastrophic backtracking on some strings. Instead of aborting the pipeline, we decided to produce a warning instead, although such patterns should be simplified/optimized if possible.

### DIFF
--- a/libtenzir/builtins/formats/grok.cpp
+++ b/libtenzir/builtins/formats/grok.cpp
@@ -422,6 +422,7 @@ public:
 
   auto parse_strings(const arrow::StringArray& input,
                      diagnostic_handler& dh) const -> std::vector<series> {
+    auto too_complex = false;
     auto builder = series_builder{type{record_type{}}};
     for (auto&& string : values(string_type{}, input)) {
       if (not string) {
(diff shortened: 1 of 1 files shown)


libtenzir/builtins/formats/grok.cpp
```diff
@@ -422,19 +422,29 @@ class grok_parser final : public plugin_parser {
...
+            .hint("pattern: `{}`", input_pattern_.resolved_pattern->str())
+            .emit(dh);
+          builder.null();
+          continue;
+        }
+      } catch (const boost::regex_error& e) {
+        if (e.code() != boost::regex_constants::error_complexity) {
+          throw;
+        }
+        too_complex = true;
```
tobim, line 447: I don't understand why we only emit a single warning for each batch in this case but one warning per value in the case above?

Now that I think about it, I would love to have an opt-out "pattern could not be matched" when know my data is not clean and I don't care about non-matches.

For the complexity case it would be nice to have count of how many times it happened and the first input on which it happened in the warning.

jachris: > I don't understand why we only emit a single warning for each batch in this case but one warning per value in the case above?

The problem with once per event is that you can get millions of warnings for high-velocity streams. I agree that this should be consistent, but I did not do this yet because it goes beyond the scope of the issue, and there are many more cases where this happens. Also, we would not show all strings for which the match failed, but either none or only one "example" string. But if you want I can still do this as part of this PR. 

> Now that I think about it, I would love to have an opt-out "pattern could not be matched" when know my data is not clean and I don't care about non-matches.

I think we previously talked about a workaround such as `"(?:<my_pattern>)|.*"`, although that will yield `null` fields instead of making the whole record `null`. Can we make this a separate issue? This needs some consideration on if/how we want to enable the use case.

> For the complexity case it would be nice to have count of how many times it happened and the first input on which it happened in the warning.

The reason I'm including this information is that we are currently introducing a deduplication strategy. I already talked a bit with @dominiklohmann about this, but basically every warning is presented at most once. For example, when a referenced field does not exists, the user sees exactly one warning for this. This reduces diagnostic fatigue, but of course also has some downsides. Currently, diagnostic notes are exempt from deduplication, so you can add an "example", but the count would be meaningless if only shown once. I added a diagnostic note for the input string.

tobim: > The reason I'm including this information is that we are currently introducing a deduplication strategy. I already talked a bit with @dominiklohmann about this, but basically every warning is presented at most once. For example, when a referenced field does not exists, the user sees exactly one warning for this. This reduces diagnostic fatigue, but of course also has some downsides. Currently, diagnostic notes are exempt from deduplication, so you can add an "example", but the count would be meaningless if only shown once. I added a diagnostic note for the input string.

It's good to hear that we finally get rid of the million diagnostics issue.
For the count note in particular I would still consider it valuable if presented as "N times in M events".

> I think we previously talked about a workaround such as "(?:<my_pattern>)|.*", although that will yield null fields instead of making the whole record null. Can we make this a separate issue? This needs some consideration on if/how we want to enable the use case.

I'll create a story for it.

tobim (approved):

PR #4365, opened 2024-07-09
Revamp the `export` operator

dominiklohmann:
This is a complete rewrite of the `export` operator that brings a few changes:
1. The operator is no longer detached, but rather cooperatively scheduled. This makes the operator cheaper to start.
2. The operator now emits metrics with the name `tenzir.metrics.export` once per second and on shutdown.
3. The operator now schedules up to three partitions at once for retroactive queries, pipelining them for an improved performance, especially when using smaller partitions. A new `--parallel <level>` option allows for overriding this setting.
4. The gap between retro and live exports is now drastically reduced, as the live export starts immediately and no longer waits for the retro export to finish. Note that this doesn't eliminate the gap window entirely, as there is still a gap between the importer and passive partitions being ack'd by the catalog.
5. The previously deprecated `--low-priority` option for the operator no longer exists. The option was a no-op for quite some time now.

### DIFF
--- a/libtenzir/src/argument_parser2.cpp
+++ b/libtenzir/src/argument_parser2.cpp
@@ -171,6 +171,21 @@ auto argument_parser2::parse(const ast::entity& self,
           return;
         }
         auto cast = caf::get_if<T>(&*value);
+        if constexpr (std::same_as<T, uint64_t>) {
+          if (not cast) {
+            auto other = caf::get_if<int64_t>(&*value);
+            if (other) {
+              if (*other < 0) {
+                emit(diagnostic::error("expected positive integer, got `{}`",
+                                       *other)
+                       .primary(expr));
+                return;
+              }
+              value = static_cast<uint64_t>(*other);
+              cast = caf::get_if<T>(&*value);
+            }
+          }
+        }
         if (not cast) {
           // TODO: Attempt conversion.
           emit(diagnostic::error("expected argument of type `{}`, but got `{}`",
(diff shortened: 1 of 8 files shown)


libtenzir/src/argument_parser2.cpp
```diff
@@ -171,6 +171,21 @@ auto argument_parser2::parse(const ast::entity& self,
...
+                emit(diagnostic::error("expected positive integer, got `{}`",
+                                       *other)
+                       .primary(expr));
+                return;
+              }
+              value = static_cast<uint64_t>(*other);
+              cast = caf::get_if<T>(&*value);
+            }
+          }
+        }
```
dominiklohmann, line 188: @jachris this is more or less a copy of the handling of the same situation in another place in the same file. Feel free to unify the two. I ran into this because I had a `std::optional<located<uint64_t>>` in an `argument_parser2`, which didn't convert as I had expected it to.

jachris: Thanks :+1: 

Side note: The type promotion thing is still has a bit to be desired in general, as it also becomes relevant when you define functions etc... Doing everything manually for now until we get a feeling for what we want to do.

web/docs/operators/export.md
```diff
@@ -11,7 +11,7 @@ Retrieves events from a Tenzir node. The dual to [`import`](import.md).
 ## Synopsis
 
 ```
-export [--live] [--internal] [--low-priority]
+export [--live] [--retro] [--internal] [--paralllel <level>]
```
jachris, line 14:
```suggestion
export [--live] [--retro] [--internal] [--parallel <level>]
```

jachris (approved):

PR #4366, opened 2024-07-09
Add missing time format specifiers for the static binary 

tobim:
This adds the `%F`, `%g`, `%G`, `%u`, `%V`, `%z`, and `%Z` conversions. The `%s` specifier (seconds since epoch) is now also accepted but has no effect on the output.


nix/overlay.nix
dominiklohmann, line 1: This should get a bug fix changelog entry.

dominiklohmann (approved):
Thank you!

PR #4367, opened 2024-07-09
Allow immediate restarts of the TCP listen connector

tobim:
A pipeline with `from tcp://...` could not be restarted seamlessly because the listening port was blocked for a certain amount of time after the pipeline was stopped.

This change adds the necessary flags so the port can be reused immediately.


libtenzir/builtins/operators/tcp-listen.cpp
```diff
@@ -361,6 +361,8 @@ auto make_connection_manager(
   self->state.endpoint = endpoints.begin()->endpoint();
   self->state.acceptor = boost::asio::ip::tcp::acceptor(*self->state.io_context,
                                                         *self->state.endpoint);
+  boost::asio::ip::tcp::acceptor::reuse_address reuse_address(true);
```
mavam, line 364:
```suggestion
  auto reuse_address = boost::asio::ip::tcp::acceptor::reuse_address{true};
```

mavam (approved):

PR #4370, opened 2024-07-09
Raise the default and max timeouts for `/serve`

dominiklohmann:
This brings up the default and maximum long-polling timeouts for `/serve` from 2s and 5s to 5s and 10s, respectively.

I've made this change because we've started polling a lot of metrics and diagnostics from the app, and this drastically reduces the load on the API when there are no diagnostics.

We can't really raise it much higher because of AWS' lambda and websocket gateway timeouts, but this I think we can safely do.


lava (approved):
This should be safe, we're currently imposing an 11 seconds timeout in the tenant-manager which would need to be modified as well if we want to increase the maximum any further.

PR #4379, opened 2024-07-11
Do not optimize `deduplicate`

dominiklohmann:
I noticed that this is broken with the following pipeline:

```
metrics platform
| where timestamp > 1w ago
| sort timestamp
| deduplicate connected --distance 1
| where connected == false
| put timestamp, message = "disconnected from platform"
```

The `where connected == false` in this pipeline got pushed through `deduplicate` incorrectly.

### DIFF
--- a/libtenzir/builtins/operators/deduplicate.cpp
+++ b/libtenzir/builtins/operators/deduplicate.cpp
@@ -185,10 +185,18 @@ public:
     return "deduplicate";
   }
 
-  auto optimize(expression const& filter, event_order order) const
+  auto optimize(const expression& filter, event_order order) const
     -> optimize_result override {
     (void)order;
-    return optimize_result{filter, event_order::schema, copy()};
+    if (cfg_.distance < std::numeric_limits<int64_t>::max()) {
+      // When `--distance` is used, we're not allowed to optimize at all. Here's
+      // a simple example that proves this:
+      //   metrics platform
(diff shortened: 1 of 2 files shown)


libtenzir/builtins/operators/deduplicate.cpp
```diff
@@ -187,8 +187,9 @@ class deduplicate_operator final : public crtp_operator<deduplicate_operator> {
 
   auto optimize(expression const& filter, event_order order) const
     -> optimize_result override {
+    (void)filter;
     (void)order;
-    return optimize_result{filter, event_order::schema, copy()};
+    return do_not_optimize(*this);
```
jachris, line 192: I think you could conditionally propagate the filter, if there is no `--limit` set, right? But `event_order::schema` looks incorrect in either case.

jachris (approved):

libtenzir/builtins/operators/deduplicate.cpp
dominiklohmann: Right, that's a good call. I didn't actually notice that this is only broken because of `--limit`.

jachris: Eh, I (and probably you) mean `--distance`, not `--limit`, right?

dominiklohmann: Yes. 🙈

PR #4380, opened 2024-07-11
Allow disabling node-to-node connections

dominiklohmann:
This change makes it so that setting the value `false` for the `tenzir.endpoint` option causes the node not to expose the node as a remote actor for `tenzir` processes or other nodes to connect to it.

### DIFF
--- a/web/docs/setup-guides/deploy-a-node.md
+++ b/web/docs/setup-guides/deploy-a-node.md
@@ -137,15 +137,22 @@ tenzir-node
 
           v4.0.0-rc6-0-gf193b51f1f
 Visit https://app.tenzir.com to get started.
+
+[16:50:26.741] node listens for node-to-node connections on tcp://127.0.0.1:5158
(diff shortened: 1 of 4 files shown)


mavam (approved):

web/docs/setup-guides/deploy-a-node.md
```diff
@@ -139,14 +139,18 @@ tenzir-node
...
+via the `tenzir.endpoint` option, e.g., bind to an IPv6 address:
 
 ```bash
 tenzir-node --endpoint=[::1]:42000
 ```
 
+If you want to use the node through the Tenzir Platform only, we recommend
+disabling node-to-node connections by setting the `tenzir.endpoint` option to
+`false`.
+
```
mavam, line 153:
```suggestion
Set `tenzir.endpoint` to `false` to disable the endpoint to access the node exclusively through the platform. This prevents connections from other `tenzir` or `tenzir-node` processes.
```

PR #4385, opened 2024-07-12
Port `metrics` and `diagnostics` to TQL2

dominiklohmann:


libtenzir/builtins/operators/diagnostics.cpp
```diff
@@ -50,6 +50,32 @@ class plugin final : public virtual operator_parser_plugin {
...
+          .add("retro", retro)
+          .parse(inv, ctx));
+    if (not live) {
+      retro = true;
+    }
+    const auto definition
+      = fmt::format("export --internal{}{} | where #schema "
+                    "== \"tenzir.diagnostic\"",
+                    live ? " --live" : "", retro ? " --retro" : "");
+    auto result = pipeline::internal_parse_as_operator(definition);
```
dominiklohmann, line 69: @jachris I'm sure there has to be a better way not than this one where I can actually pass through the session. This works, but I'd like to learn how to do this properly with the new APIs.

jachris: You have two other options:

1. Use `operator_factory_plugin::make`.
2. Don't use `operator_factory_plugin::make`.

This might sound a bit dumb, but what I mean is: For some things, I find that the plugin interface might not be the best choice. It might be _too_ decoupled. Might be best to discuss this a bit in Slack.

dominiklohmann: I understand. See 74d05dd677a.

jachris: Yep, that would be one way to do it ☺️

jachris (approved):

PR #4386, opened 2024-07-12
Fix TQL2 metadata assignment

jachris:
Since schema name and the internal flag are reflected by Arrow's schema, we have to keep both in sync. The import time does not need this because it's not part of the Arrow schema.


dominiklohmann (approved):
This makes sense to me.

PR #4387, opened 2024-07-12
Add `else if` as a better alternative to nested `if`s

jachris:
Turns out this TQL2 feature was rather easy to implement.


Dakostu (approved):
Approving to confirm that it works for my use case.

PR #4388, opened 2024-07-13
Add `community_id` TQL function

mavam:
This PR adds a `community_id` function to TQL.

### DIFF
--- a/libtenzir/builtins/functions/community_id.cpp
+++ b/libtenzir/builtins/functions/community_id.cpp
@@ -1,0 +1,140 @@
+//    _   _____   __________
+//   | | / / _ | / __/_  __/     Visibility
+//   | |/ / __ |_\ \  / /          Across
+//   |___/_/ |_/___/ /_/       Space and Time
+//
+// SPDX-FileCopyrightText: (c) 2024 The Tenzir Contributors
+// SPDX-License-Identifier: BSD-3-Clause
+
+#include <tenzir/as_bytes.hpp>
+#include <tenzir/community_id.hpp>
+#include <tenzir/concept/parseable/tenzir/si.hpp>
+#include <tenzir/detail/narrow.hpp>
+#include <tenzir/flow.hpp>
+#include <tenzir/tql2/arrow_utils.hpp>
+#include <tenzir/tql2/plugin.hpp>
+
+#include <caf/fused_downstream_manager.hpp>
+
+namespace tenzir::plugins::community_id {
+
+namespace {
+
+struct arguments {
+  ast::expression src_ip;
+  ast::expression dst_ip;
+  ast::expression dst_port;
+  ast::expression src_port;
+  ast::expression proto;
+};
+
+class plugin final : public function_plugin {
+public:
+  auto name() const -> std::string override {
+    return "tql2.community_id";
+  }
+
+  auto make_function(invocation inv, session ctx) const
+    -> failure_or<function_ptr> override {
+    auto args = arguments{};
+    TRY(argument_parser2::function("community_id")
+          .add(args.src_ip, "<source ip>")
+          .add(args.dst_ip, "<destination ip>")
+          .add(args.src_port, "<source port>")
+          .add(args.dst_port, "<destination port>")
+          .add(args.proto, "<transport protocol>")
+          .parse(inv, ctx));
+    return function_use::make([args = std::move(args)](evaluator eval,
+                                                       session ctx) -> series {
+      auto null_series = [&] {
+        return series::null(string_type{}, eval.length());
+      };
+      auto src_ip_series = eval(args.src_ip);
+      if (not caf::holds_alternative<ip_type>(src_ip_series.type)) {
+        diagnostic::warning("`community_id` expected `ip` as 1st argument")
+          .primary(args.src_ip)
+          .emit(ctx);
+        return null_series();
+      }
+      auto dst_ip_series = eval(args.dst_ip);
+      if (not caf::holds_alternative<ip_type>(dst_ip_series.type)) {
+        diagnostic::warning("`community_id` expected `ip` as 3rd argument")
+          .primary(args.dst_ip)
+          .emit(ctx);
+        return null_series();
+      }
+      auto src_port_series = eval(args.src_port);
+      if (not caf::holds_alternative<int64_type>(src_port_series.type)) {
+        diagnostic::warning("`community_id` expected `integer` as 2nd argument")
+          .primary(args.src_port)
+          .emit(ctx);
+        return null_series();
+      }
+      auto dst_port_series = eval(args.dst_port);
+      if (not caf::holds_alternative<int64_type>(dst_port_series.type)) {
+        diagnostic::warning("`community_id` expected `integer` as 4th argument")
+          .primary(args.dst_port)
+          .emit(ctx);
+        return null_series();
+      }
+      auto proto_series = eval(args.proto);
+      if (not caf::holds_alternative<string_type>(proto_series.type)) {
+        diagnostic::warning("`community_id` expected `string` as 5th argument")
+          .primary(args.dst_port)
+          .emit(ctx);
+        return null_series();
+      }
+      auto b = arrow::StringBuilder{};
+      check(b.Reserve(eval.length()));
+      auto src_ips = src_ip_series.as<ip_type>();
+      auto dst_ips = dst_ip_series.as<ip_type>();
+      auto src_ports = src_port_series.as<int64_type>();
+      auto dst_ports = dst_port_series.as<int64_type>();
+      auto protos = proto_series.as<string_type>();
+      for (auto i = int64_t{0}; i < eval.length(); ++i) {
+        if (src_ips->array->IsNull(i) or src_ports->array->IsNull(i)
+            or dst_ips->array->IsNull(i) or dst_ports->array->IsNull(i)
+            or protos->array->IsNull(i)) {
+          check(b.AppendNull());
+          continue;
+        }
+        const auto* src_ip_ptr = src_ips->array->storage()->GetValue(i);
+        const auto* dst_ip_ptr = dst_ips->array->storage()->GetValue(i);
+        auto src_ip = ip::v6(as_bytes<16>(src_ip_ptr));
+        auto dst_ip = ip::v6(as_bytes<16>(dst_ip_ptr));
+        auto src_port = src_ports->array->GetView(i);
+        auto dst_port = dst_ports->array->GetView(i);
+        auto proto = protos->array->GetView(i);
+        auto type = port_type::unknown;
+        if (proto == "tcp") {
+          type = port_type::tcp;
+        } else if (proto == "udp") {
+          type = port_type::udp;
+        } else if (proto == "icmp") {
+          type = port_type::icmp;
+        } else if (proto == "icmp6") {
+          type = port_type::icmp6;
+        } else {
+          diagnostic::warning("`community_id` expected `tcp`, `udp`, `icmp`, "
+                              "or `icmp6` as protocol")
+            .primary(args.proto)
+            .emit(ctx);
+          return null_series();
+        }
+        auto flow
+          = make_flow(src_ip, dst_ip, detail::narrow<uint16_t>(src_port),
+                      detail::narrow<uint16_t>(dst_port), type);
+        auto seed = uint16_t{0}; // TODO: make an optional parameter.
+        check(
+          b.Append(tenzir::community_id::compute<policy::base64>(flow, seed)));
+      }
+      return series{string_type{}, finish(b)};
+    });
+  }
+};
+
+} // namespace
+
+} // namespace tenzir::plugins::community_id
+
+TENZIR_REGISTER_PLUGIN(tenzir::plugins::community_id::plugin)
--- a/libtenzir/builtins/functions/community_id.cpp
+++ b/libtenzir/builtins/functions/community_id.cpp
@@ -119,7 +119,8 @@ public:
                               "or `icmp6` as protocol")
             .primary(args.proto)
             .emit(ctx);
-          return null_series();
+          check(b.AppendNull());
+          continue;
         }
         auto flow
           = make_flow(src_ip, dst_ip, detail::narrow<uint16_t>(src_port),
--- a/libtenzir/builtins/functions/community_id.cpp
+++ b/libtenzir/builtins/functions/community_id.cpp
@@ -21,9 +21,12 @@ namespace tenzir::plugins::community_id {
 namespace {
 
 struct arguments {
-  ast::expression src_ip;
-  ast::expression dst_ip;
-  ast::expression proto;
+  // TODO: make src_ip, dst_ip, and proto required named arguments. We currently
+  // check this requirement manually, but should remove the std::optional<T> in
+  // the future.
+  std::optional<ast::expression> src_ip;
+  std::optional<ast::expression> dst_ip;
+  std::optional<ast::expression> proto;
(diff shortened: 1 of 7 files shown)


mavam:
@jachris I tried this but get an assertion failure:

```
source {cid: community_id(10.0.0.1, 10.0.0.2, 42221, 22, "tcp")}
```

This fails in:

```cpp
      auto src_ips = src_ip_series.as<ip_type>();
```

Any suggestion what I'm doing wrong?

libtenzir/builtins/functions/community_id.cpp
```diff
@@ -0,0 +1,140 @@
...
+        } else if (proto == "icmp") {
+          type = port_type::icmp;
+        } else if (proto == "icmp6") {
+          type = port_type::icmp6;
+        } else {
+          diagnostic::warning("`community_id` expected `tcp`, `udp`, `icmp`, "
+                              "or `icmp6` as protocol")
+            .primary(args.proto)
+            .emit(ctx);
+          return null_series();
```
dominiklohmann, line 122: I don't think you want to return a whole batch of nulls if one protocol in the batch is invalid, but rather just add a null here and emit a warning.

mavam: Yes, thanks. Fixed.

jachris:
> @jachris I tried this but get an assertion failure:
> 
> ```
> source {cid: community_id(10.0.0.1, 10.0.0.2, 42221, 22, "tcp")}
> ```
> 
> This fails in:
> 
> ```c++
>       auto src_ips = src_ip_series.as<ip_type>();
> ```
> 
> Any suggestion what I'm doing wrong?

This unexpectedly fails the dynamic cast to `ip_type::array_type`, even though its `typeid(...)` is as expected (equal to the target type). I previously encountered this problem with a different type and had to apply a workaround. This is an issue with how builtin plugins are setup.

libtenzir/builtins/functions/community_id.cpp
```diff
@@ -0,0 +1,141 @@
...
+          .add(args.dst_port, "<destination port>")
+          .add(args.proto, "<transport protocol>")
+          .parse(inv, ctx));
+    return function_use::make([args = std::move(args)](evaluator eval,
+                                                       session ctx) -> series {
+      auto null_series = [&] {
+        return series::null(string_type{}, eval.length());
+      };
+      auto src_ip_series = eval(args.src_ip);
+      if (not caf::holds_alternative<ip_type>(src_ip_series.type)) {
```
jachris, line 53: This check fails for `null_type`. Thus you'd get a warning here, while the behavior for "properly-typed" `null` would be different. I think we want to avoid that.

libtenzir/builtins/functions/community_id.cpp
```diff
@@ -0,0 +1,141 @@
...
+          .emit(ctx);
+        return null_series();
+      }
+      auto b = arrow::StringBuilder{};
+      check(b.Reserve(eval.length()));
+      auto src_ips = src_ip_series.as<ip_type>();
+      auto dst_ips = dst_ip_series.as<ip_type>();
+      auto src_ports = src_port_series.as<int64_type>();
+      auto dst_ports = dst_port_series.as<int64_type>();
+      auto protos = proto_series.as<string_type>();
```
jachris, line 93: FWIW: This is a checked cast, so you don't have to check with `caf::holds_alternative` before.

libtenzir/builtins/functions/community_id.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+  auto make_function(invocation inv, session ctx) const
+    -> failure_or<function_ptr> override {
+    auto args = arguments{};
+    TRY(argument_parser2::function("community_id")
+          .add(args.src_ip, "<source ip>")
+          .add(args.dst_ip, "<destination ip>")
+          .add(args.src_port, "<source port>")
+          .add(args.dst_port, "<destination port>")
+          .add(args.proto, "<transport protocol>")
```
dominiklohmann, line 45: The arguments in this are all positional, when they could also all be named. What do we want to default to here?

I personally think that named arguments are better.

mavam: Agreed, in this case there is no obvious argument order. I would argue only then it's a better UX to have positional arguments.

jachris: The given order here seems pretty established across the spec and various implementations. So I think it might be fine? It's also hard to misuse without at least getting a warning.

libtenzir/builtins/functions/community_id.cpp
```diff
@@ -0,0 +1,141 @@
...
+      auto src_ports = src_port_series.as<int64_type>();
+      auto dst_ports = dst_port_series.as<int64_type>();
+      auto protos = proto_series.as<string_type>();
+      for (auto i = int64_t{0}; i < eval.length(); ++i) {
+        if (src_ips->array->IsNull(i) or src_ports->array->IsNull(i)
+            or dst_ips->array->IsNull(i) or dst_ports->array->IsNull(i)
+            or protos->array->IsNull(i)) {
+          check(b.AppendNull());
+          continue;
+        }
```
jachris, line 100: It looks like the ports are optional, at least when the protocol is not UDP or TCP. We could also add that later.

libtenzir/builtins/functions/community_id.cpp
```diff
@@ -0,0 +1,141 @@
...
+                              "or `icmp6` as protocol")
+            .primary(args.proto)
+            .emit(ctx);
+          check(b.AppendNull());
+          continue;
+        }
+        auto flow
+          = make_flow(src_ip, dst_ip, detail::narrow<uint16_t>(src_port),
+                      detail::narrow<uint16_t>(dst_port), type);
+        auto seed = uint16_t{0}; // TODO: make an optional parameter.
```
jachris, line 128: You could make that an optional named argument, if you want to. 

~~~
community_id(10.0.0.1, 10.0.0.2, 42221, 22, "tcp", seed=42)
~~~

But it's also fine to add this later.

mavam: Exactly what I want. I just haven't figured out how to do it in TQL2.

libtenzir/builtins/functions/community_id.cpp
mavam: Even though the spec has a fixed order, the reality is that networking monitoring tools present flows in different ways. I'd say we should switch to named arguments.

libtenzir/builtins/functions/community_id.cpp
mavam: For ICMP, the message types are used as ports. These might be difficult to get from log files. Zeek, for example, encodes them in source and destination port. But that's okay. This is something the user of this function should know. I would prefer not to default to `tcp` just to remain safe and explicit.

libtenzir/builtins/functions/community_id.cpp
dominiklohmann: You just swap the two arguments to `add`, so instead of `add(seed, "<seed>")` (positional argument) you use `add("seed", seed)` (named argument).

mavam: Now implemented.

libtenzir/builtins/functions/community_id.cpp
```diff
@@ -0,0 +1,180 @@
...
+          continue;
+        }
+        auto flow
+          = make_flow(src_ip, dst_ip, detail::narrow<uint16_t>(src_port),
+                      detail::narrow<uint16_t>(dst_port), type);
+        auto seed = uint16_t{0};
+        if (seeds and not seeds->array->IsNull(i)) {
+          // TODO: Perform a bounds check. There are probably already utilities
+          // for this available.
+          seed = detail::narrow_cast<uint16_t>(seeds->array->GetView(i));
```
dominiklohmann, line 166: @jachris here's your need for `try_narrow` again 😅

mavam:
I am struggling with a problem here: the ports can be either of `int64_type` or `uint64_type`. What I'd like to have is something along the lines of `series<variant<int64, uint64>>`, because all I care for is an integer within [0, 2^16-1].

It seems that I must now check for both signed and unsigned integer everywhere. Has anyone thought about this already?

dominiklohmann:
> It seems that I must now check for both signed and unsigned integer everywhere. Has anyone thought about this already?

There's an open TODO about a similar situation in the argument parser, which has a similar confusion between number types: https://github.com/tenzir/tenzir/blob/073d3e8639aac75cb705ac9353433828d8f012ef/libtenzir/src/argument_parser2.cpp#L101-L118

libtenzir/builtins/functions/community_id.cpp
jachris: Yeah, we should definitely add this. 

What should happen if the port is out of range? Add a `null`, and emit a warning (one per batch)?

libtenzir/builtins/functions/community_id.cpp
```diff
@@ -0,0 +1,180 @@
...
+          type = port_type::icmp;
+        } else if (proto == "icmp6") {
+          type = port_type::icmp6;
+        } else {
+          diagnostic::warning("`community_id` expected `tcp`, `udp`, `icmp`, "
+                              "or `icmp6` as protocol")
+            .primary(args.proto)
+            .emit(ctx);
+          check(b.AppendNull());
+          continue;
```
jachris, line 157: This will emit one warning per entry. It will get deduplicated eventually, but a more optimal solution would be to emit this warning once at the end of the function. Unless we don't want to bother with that.

libtenzir/builtins/functions/community_id.cpp
jachris: I'm aware of that, but the reference implementation strongly suggests that ports are optional: 
https://github.com/corelight/pycommunityid/blob/5748e0c4808de6910abcf67b426cf8284af2da74/communityid/algo.py#L450-L457

> I would prefer not to default to tcp just to remain safe and explicit.

I fully agree with that :+1:

mavam: > the reference implementation strongly suggests that ports are optional

Interesting, I learned something new. Thanks for being persistent!

jachris:
> It seems that I must now check for both signed and unsigned integer everywhere. Has anyone thought about this already?

I expected this to come up eventually. I think the optimal solution depends a bit on the use-case. In your case, you could use a `arrow::UInt64Array -> arrow::Int64Array` conversion function, that inserts `null` when conversion fails (and provides a way to notice that this happened so that we can emit a warning), or you can use some custom getter to convert this on the fly (in a way where we prevent emitting code for this 4x). That could either go through some variant, or directly through `arrow::Array`.

dominiklohmann:
I think it'd be nice to have `series.cast<T>(const T& t)` in addition to `series.as<T>()` that does this for you as best possible.

mavam:
> I think it'd be nice to have `series.cast<T>(const T& t)` in addition to `series.as<T>()` that does this for you as best possible.

Yes! Or let `cast<T>()` return a `std::optional<basic_series<T>>`, similar to `as<T>()`.

jachris:
But you would not want to emit `null` for a whole batch just because one single row was out-of-range. So it can't return `std::optional<series>`.

---

> You could use a arrow::UInt64Array -> arrow::Int64Array conversion function, that inserts null when conversion fails (and provides a way to notice that this happened so that we can emit a warning)

I meant something more along the lines of `basic_series<T>::cast<U>() -> std::pair<basic_series<U>, uint64_t>`.

libtenzir/builtins/functions/community_id.cpp
mavam: I'd say let's wait with adding complexity until someone runs into this.

libtenzir/builtins/functions/community_id.cpp
mavam: I've implemented a bounds check manually.

libtenzir/builtins/functions/community_id.cpp
mavam: Fixed by now.

libtenzir/builtins/functions/community_id.cpp
mavam: All arguments are now named arguments.

libtenzir/builtins/functions/community_id.cpp
```diff
@@ -0,0 +1,251 @@
...
+          seed_series = std::nullopt;
+        }
+      }
+      auto src_ips = src_ip_series.as<ip_type>();
+      if (not src_ips) {
+        diagnostic::warning("`community_id` expected `ip` as 1st argument")
+          .primary(*args.src_ip)
+          .emit(ctx);
+        return null_series();
+      }
```
jachris, line 118: Since they are now named, the "1st argument" doesn't make sense (same below). 

You can do this instead:

```cpp
diagnostic::error("expected argument of type `ip`, but got `{}`", src_ip_series.type.kind())
```

(Side note: We should probably factor this out at some point to provide templates for the common error messages)

libtenzir/builtins/functions/community_id.cpp
```diff
@@ -0,0 +1,251 @@
...
+      auto src_ports = std::optional<basic_series<int64_type>>{};
+      auto dst_ports = std::optional<basic_series<int64_type>>{};
+      if (src_port_series) {
+        src_ports = src_port_series->as<int64_type>();
+        if (not src_ports) {
+          diagnostic::warning("`community_id` got an argument type mismatch")
+            .primary(*args.src_port)
+            .note("expected argument `src_port` to be of type `int64`")
+            .emit(ctx);
```
jachris, line 141:
```suggestion
          diagnostic::warning("expected argument of type `int64`, but got `{}`", src_port_series->type.kind())
            .primary(*args.src_port)
            .emit(ctx);
```

Same below.

libtenzir/builtins/functions/community_id.cpp
```diff
@@ -0,0 +1,251 @@
...
+          seed = detail::narrow_cast<uint16_t>(value);
+        }
+        auto have_src_port = src_ports and not src_ports->array->IsNull(i);
+        auto have_dst_port = dst_ports and not dst_ports->array->IsNull(i);
+        if (have_src_port and have_dst_port) {
+          auto src_port = src_ports->array->GetView(i);
+          auto dst_port = dst_ports->array->GetView(i);
+          auto flow
+            = make_flow(src_ip, dst_ip, detail::narrow<uint16_t>(src_port),
+                        detail::narrow<uint16_t>(dst_port), proto_type);
```
jachris, line 213: This narrowing needs to be checked. This once more shows the need for some abstractions again, but it doesn't hurt to do this a few times manually so that we know what we want.

libtenzir/builtins/functions/community_id.cpp
```diff
@@ -0,0 +1,251 @@
...
+          check(b.Append(
+            tenzir::community_id::make(src_ip, dst_ip, proto_type, seed)));
+        }
+      }
+      if (emit_seed_warning) {
+        diagnostic::warning("`community_id` got invalid seed")
+          .primary(*args.seed)
+          .note("`seed` exceeded maximum value of 65,535")
+          .hint("use a seed that's less then the maximum value")
+          .emit(ctx);
```
jachris, line 229:
```suggestion
        diagnostic::warning("`seed` must be between 0 and 65535")
          .primary(*args.seed)
          .emit(ctx);
```

YMMV, but I prefer this. Overall information content should be about the same (and I added the 0 lowerbound). You can make similar changes below, if you want to.

jachris (approved):
Thanks! :rocket:

PR #4392, opened 2024-07-15
Make `libtenzir_builtins` an object library

jachris:
This is a test.


libtenzir/CMakeLists.txt
```diff
@@ -176,7 +176,7 @@ add_library(libtenzir ${libtenzir_sources})
 TenzirTargetEnableTooling(libtenzir)
 add_library(tenzir::libtenzir ALIAS libtenzir)
 
-add_library(libtenzir_builtins STATIC ${libtenzir_builtins_sources})
+add_library(libtenzir_builtins OBJECT ${libtenzir_builtins_sources})
```
dominiklohmann, line 179: There's also the option of using the approach from https://github.com/tenzir/tenzir/pull/3754 (which was for reasons I still don't understand reverted in https://github.com/tenzir/tenzir/pull/3766). That has the advantage of also massively reducing linker time.

jachris: I think the overall effect should be similar, no? No additional linking should take place here.

dominiklohmann:
@jachris can we close this one?

jachris:
@dominiklohmann One of the reasons for this PR was to solve the problem where we cannot reliably `dynamic_cast` across plugins, which led to hacky workarounds at least twice. That issue is still there, so I think there still might be value in seeing whether using object libraries solves it. But I did have other priorities in the meantime, not sure when I'll be able to pick this up again.

dominiklohmann:
Then let's keep it around for a while.

PR #4399, opened 2024-07-16
Break back pressure in hidden subscribers

dominiklohmann:
This change makes it so that `subscribe` no longer propagates back pressure when run as part of a hidden pipeline, i.e., a pipeline that is not visible on the Tenzir Platform.

Without this change, it was surprisingly easy to break publishers by simply subscribing to their topic and piping the data into an operator that did nothing, e.g., `serve` with a very small buffer size.

### DIFF
--- a/web/docs/operators/subscribe.md
+++ b/web/docs/operators/subscribe.md
@@ -21,6 +21,12 @@ The `subscribe` operator subscribes to events from a channel with the specified
 topic. Multiple `subscribe` operators with the same topic receive the same
 events.
 
+Subscribers propagate back pressure to publishers. If a subscribing pipeline
+fails to keep up, all publishers will slow down as well to a matching speed to
+avoid data loss. This mechanism is disabled for pipelines that are not visible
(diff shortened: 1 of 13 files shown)


web/docs/operators/subscribe.md
```diff
@@ -21,6 +21,12 @@ The `subscribe` operator subscribes to events from a channel with the specified
 topic. Multiple `subscribe` operators with the same topic receive the same
 events.
 
+Subscribers propagate back pressure to publishers. If a subscribing pipeline
+fails to keep up, all publishers will slow down as well to a matching speed to
+avoid data loss. This mechanism is disabled for pipelines that are not visible
+on the overview page on app.tenzir.com, which drop data rather than slow down
+their publishers.
```
jachris, line 28:
```suggestion
on the overview page on [app.tenzir.com](https://app.tenzir.com), which drop data
rather than slow down their publishers.
```

Otherwise it's not shown as a link (you could also change the text if you want to, not sure how consistent we are when referencing the app).

jachris (approved):

PR #4403, opened 2024-07-17
Add `{...<expr>}` to spread records

jachris:
For example: 
```
foo = {...foo.bar, baz: 42, ...qux(quux)}
```


libtenzir/src/tql2/eval_impl.cpp
```diff
@@ -15,19 +15,25 @@ namespace tenzir {
...
-  for (auto& item : x.content) {
+  for (auto& item : x.items) {
     item.match(
       [&](const ast::record::field& field) {
         auto val = eval(field.expr);
-        auto [_, inserted] = fields.emplace(field.name.name, std::move(val));
-        if (not inserted) {
-          diagnostic::warning("todo: overwrite existing?")
-            .primary(field.name)
+        // TODO: Do we want to keep the order of overwritten fields?
```
mavam, line 22: I would say the user expects that the original order of the fields remains, also if they get overwritten by other values.

jachris: I think so too, and JS seems to agree:
```js
>>> { a: 1, b: 2, ...{a: 3} }
{ a: 3, b: 2 }
```

dominiklohmann (approved):
❤️

